PR #1873, opened 2019-02-22
Utility method for combining protos naively

stephanwlee:
Graph plugin soon will support FunctionGraphs in the RunMetadata.
Because it is a tree of GraphDefs, for our frontend, we need to be
able to compose one GraphDef.

### DIFF
--- a/tensorboard/plugins/graph/graph_util.py
+++ b/tensorboard/plugins/graph/graph_util.py
@@ -14,9 +14,7 @@
 # limitations under the License.
 # ==============================================================================
 """Utilities for graph plugin."""
-from tensorboard.compat.proto.graph_pb2 import GraphDef
-from tensorboard.compat.tensorflow_stub import dtypes
-
+from tensorboard.compat.proto import function_pb2
 
 def combine_graph_defs(to_proto, from_proto):
   """Combines two GraphDefs by adding nodes from from_proto into to_proto.
--- a/tensorboard/plugins/graph/graph_util.py
+++ b/tensorboard/plugins/graph/graph_util.py
@@ -16,6 +16,7 @@
 """Utilities for graph plugin."""
 from tensorboard.compat.proto import function_pb2
 
+
 def combine_graph_defs(to_proto, from_proto):
   """Combines two GraphDefs by adding nodes from from_proto into to_proto.
 
--- a/tensorboard/plugins/graph/graph_util.py
+++ b/tensorboard/plugins/graph/graph_util.py
@@ -15,14 +15,19 @@
 """Utilities for graph plugin."""
 from tensorboard.compat.proto import function_pb2
 
-def _is_present(key_to_proto, proto, get_key, error_msg):
-  """Checks for key in a dict and raises if content in the dict differ from proto.
+
+def _safe_copy_proto_list_values(to_proto_list, from_proto_list, get_key, error_msg):
+  """Safely copies the value from from_proto_list to to_proto_list.
+
+  Copies if to_proto_list does not contain an item with the same key. In case an item
+  is found with the same key, it checks whether contents match and, if not, it raises
+  a ValueError.
 
   Args:
-    key_to_proto: A dict that maps key to a proto.
+    to_proto_list: A dict that maps key to a proto.
     proto: A proto.
     get_key: A lambda that returns a string key from a proto.
-    error_msg: Error message.
+    error_msg: User friendly error message.
 
   Raises:
     ValueError when there is the key in the dict but contents mismatch.
@@ -30,10 +35,18 @@ def _is_present(key_to_proto, proto, get_key, error_msg):
   Returns:
     True if key from proto is present in the key_to_proto.
   """
-  key = get_key(proto)
-  if key in key_to_proto and proto != key_to_proto.get(key):
-    raise ValueError(error_msg + ': %s' % key)
-  return key in key_to_proto
+  key_to_proto = {}
+  for proto in to_proto_list:
+    key = get_key(proto)
+    key_to_proto[key] = proto
+
+  for proto in from_proto_list:
+    key = get_key(proto)
+    if key in key_to_proto:
+      if proto != key_to_proto.get(key):
+        raise ValueError(error_msg + ': %s' % key)
+    else:
+      to_proto_list.add().CopyFrom(proto)
 
 
 def combine_graph_defs(to_proto, from_proto):
@@ -53,41 +66,23 @@ def combine_graph_defs(to_proto, from_proto):
   if from_proto.version != to_proto.version:
     raise ValueError('Cannot combine GraphDefs of different versions.')
 
-  node_name_to_nodedef = {}
-  for node in to_proto.node:
-    node_name_to_nodedef[node.name] = node
+  _safe_copy_proto_list_values(
+      to_proto.node,
+      from_proto.node,
+      lambda n: n.name,
+      'Cannot combine GraphDefs because nodes share a name but are different')
 
-  func_name_to_func = {}
-  for func in to_proto.library.function:
-    func_name_to_func[func.signature.name] = func
+  _safe_copy_proto_list_values(
+      to_proto.library.function,
+      from_proto.library.function,
+      lambda n: n.signature.name,
+      'Cannot combine GraphDefs because nodes share a name but are different')
 
-  gradient_name_to_def = {}
-  for gradient_def in to_proto.library.gradient:
-    gradient_name_to_def[gradient_def.gradient_func] = gradient_def
-
-  for from_node in from_proto.node:
-    if not _is_present(
-        node_name_to_nodedef,
-        from_node,
-        lambda n: n.name,
-        'Cannot combine GraphDefs because nodes share a name but are different'):
-      to_proto.node.add().CopyFrom(from_node)
-
-  for from_function in from_proto.library.function:
-    if not _is_present(
-        func_name_to_func,
-        from_function,
-        lambda f: f.signature.name,
-        'Cannot combine GraphDefs because functions share a name but are different'):
-      to_proto.library.function.add().CopyFrom(from_function)
-
-  for from_gradient in from_proto.library.gradient:
-    if not _is_present(
-        gradient_name_to_def,
-        from_gradient,
-        lambda g: g.gradient_func,
-        ('Cannot combine GraphDefs because gradients share a gradient_func name '
-        'but maps to a different function')):
-      to_proto.library.gradient.add().CopyFrom(from_gradient)
+  _safe_copy_proto_list_values(
+      to_proto.library.gradient,
+      from_proto.library.gradient,
+      lambda g: g.gradient_func,
+      ('Cannot combine GraphDefs because gradients share a gradient_func name '
+      'but maps to a different function'))
 
   return to_proto
--- a/tensorboard/plugins/graph/graph_util.py
+++ b/tensorboard/plugins/graph/graph_util.py
@@ -31,7 +31,16 @@ def combine_graph_defs(to_proto, from_proto):
   if from_proto.version != to_proto.version:
     raise ValueError('Cannot combine GraphDefs of different versions.')
 
+  node_names = {}
+
+  for node in to_proto.node:
+    node_names[node.name] = True
+
   for from_node in from_proto.node:
+    if from_node.name in node_names:
+      raise ValueError(
+          'Cannot combine GraphDefs when node names collide: %s' % from_node.name)
+
     to_node = to_proto.node.add()
     to_node.CopyFrom(from_node)
 
--- a/tensorboard/plugins/graph/graph_util.py
+++ b/tensorboard/plugins/graph/graph_util.py
@@ -15,12 +15,33 @@
 """Utilities for graph plugin."""
 from tensorboard.compat.proto import function_pb2
 
+def _is_present(key_to_proto, proto, get_key, error_msg):
+  """Checks for key in a dict and raises if content in the dict differ from proto.
+
+  Args:
+    key_to_proto: A dict that maps key to a proto.
+    proto: A proto.
+    get_key: A lambda that returns a string key from a proto.
+    error_msg: Error message.
+
+  Raises:
+    ValueError when there is the key in the dict but contents mismatch.
+
+  Returns:
+    True if key from proto is present in the key_to_proto.
+  """
+  key = get_key(proto)
+  if key in key_to_proto and proto != key_to_proto.get(key):
+    raise ValueError(error_msg + ': %s' % key)
+  return key in key_to_proto
+
 
 def combine_graph_defs(to_proto, from_proto):
   """Combines two GraphDefs by adding nodes from from_proto into to_proto.
 
   All GraphDefs are expected to be of TensorBoard's.
-  It assumes node names are unique across GraphDefs.
+  It assumes node names are unique across GraphDefs if contents differ. The
+  names can be the same if the NodeDef content are exactly the same.
 
   Args:
     to_proto: A destination TensorBoard GraphDef.
@@ -32,25 +53,41 @@ def combine_graph_defs(to_proto, from_proto):
   if from_proto.version != to_proto.version:
     raise ValueError('Cannot combine GraphDefs of different versions.')
 
-  node_names = set([node.name for node in to_proto.node])
+  node_name_to_nodedef = {}
+  for node in to_proto.node:
+    node_name_to_nodedef[node.name] = node
+
+  func_name_to_func = {}
+  for func in to_proto.library.function:
+    func_name_to_func[func.signature.name] = func
+
+  gradient_name_to_def = {}
+  for gradient_def in to_proto.library.gradient:
+    gradient_name_to_def[gradient_def.gradient_func] = gradient_def
 
   for from_node in from_proto.node:
-    if from_node.name in node_names:
-      raise ValueError(
-          'Cannot combine GraphDefs when node names collide: %s' % from_node.name)
+    if not _is_present(
+        node_name_to_nodedef,
+        from_node,
+        lambda n: n.name,
+        'Cannot combine GraphDefs because nodes share a name but are different'):
+      to_proto.node.add().CopyFrom(from_node)
 
-    to_node = to_proto.node.add()
-    to_node.CopyFrom(from_node)
+  for from_function in from_proto.library.function:
+    if not _is_present(
+        func_name_to_func,
+        from_function,
+        lambda f: f.signature.name,
+        'Cannot combine GraphDefs because functions share a name but are different'):
+      to_proto.library.function.add().CopyFrom(from_function)
 
-  if from_proto.library:
-    if not to_proto.library:
-      to_proto.library = function_pb2.FunctionDefLibrary()
-
-    for from_function in from_proto.library.function:
-      to_function = to_proto.library.function.add()
-      to_function.CopyFrom(from_function)
-    for from_gradient in from_proto.library.gradient:
-      to_gradient = to_proto.library.gradient.add()
-      to_gradient.CopyFrom(from_gradient)
+  for from_gradient in from_proto.library.gradient:
+    if not _is_present(
+        gradient_name_to_def,
+        from_gradient,
+        lambda g: g.gradient_func,
+        ('Cannot combine GraphDefs because gradients share a gradient_func name '
+        'but maps to a different function')):
+      to_proto.library.gradient.add().CopyFrom(from_gradient)
 
   return to_proto
--- a/tensorboard/plugins/graph/graph_util.py
+++ b/tensorboard/plugins/graph/graph_util.py
@@ -39,7 +39,8 @@ def _safe_copy_proto_list_values(dst_proto_list, src_proto_list, get_key, error_
       that two nodes share a key but are distinct.
 
   Raises:
-    ValueError when there is the key in the dict but contents mismatch.
+    ValueError with error_msg when there is an item with the same key
+    but contents mismatch.
   """
   key_to_proto = {}
   for proto in dst_proto_list:
--- a/tensorboard/plugins/graph/graph_util.py
+++ b/tensorboard/plugins/graph/graph_util.py
@@ -20,6 +20,7 @@ def combine_graph_defs(to_proto, from_proto):
   """Combines two GraphDefs by adding nodes from from_proto into to_proto.
 
   All GraphDefs are expected to be of TensorBoard's.
+  It assumes node names are unique across GraphDefs.
 
   Args:
     to_proto: A destination TensorBoard GraphDef.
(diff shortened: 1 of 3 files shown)


tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,48 @@
...
+    from_proto: a GraphDef to copy contents from.
+
+  Returns:
+    to_proto
+  """
+  if from_proto.version != to_proto.version:
+    raise ValueError('Cannot combine GraphDefs of different versions.')
+
+  for from_node in from_proto.node:
+    to_node = to_proto.node.add()
```
wchargin: Correct me if I’m wrong: from our discussions, I took away that the
purpose of this function is to construct a union of _disjoint_ graphs so
that they may be visualized in the UI. But this function makes no
attempt to ensure or validate uniqueness of names: the result of this
function may be a `GraphDef` that has multiple node entries with the
same name, even though no graphs emitted by TensorFlow have that
property.

Does it make sense for graphs to have multiple nodes with the same name,
and will our code support that? (Does none of our data manipulation or
frontend UI logic depend on keying things uniquely by name?)

If the answer is something like, “this function is only expected to be
called with `to_proto` and `from_proto` representing function graphs
from the same TensorFlow run, so we know that their key sets are
disjoint”. In that case, we must verify that this assumption holds, and
raise a `ValueError` if the key sets are not disjoint.

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,48 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Utilities for graph plugin."""
+from tensorboard.compat.proto import function_pb2
+
+
+def combine_graph_defs(to_proto, from_proto):
+  """Combines two GraphDefs by adding nodes from from_proto into to_proto.
```
wchargin: Related to below comment, but: what does it mean to “combine” two
`GraphDef`s? If I have graphs

```
a -> b
```

and

```
a -> c -> b
```

is the result

```
 .----> b
a       ^
 `-> c -'
```

or what?

tensorboard/plugins/graph/graph_util_test.py
```diff
@@ -0,0 +1,108 @@
...
+          op: "Input"
+        }
+        node {
+          name: "Y"
+          op: "MatMul"
+          input: "X"
+          input: "W"
+        }
+        node {
+          name: "X"
```
wchargin: For instance, I don’t understand this expected output. The node `X`
appears twice. Which node is the input to `Y`? which is the input to
`Q`? are the two `X`s actually the “same” node? if so, what happens if
one of them has `attr`s that the other doesn’t?

tensorboard/plugins/graph/graph_util_test.py
```diff
@@ -0,0 +1,108 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+from google.protobuf import text_format
+import tensorflow as tf
+
+from tensorboard.compat.proto.graph_pb2 import GraphDef
+from tensorboard.plugins.graph import graph_util
+from tensorflow.python.platform import test
```
wchargin: unused?

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,48 @@
...
+
+  Args:
+    to_proto: a destination GraphDef.
```
wchargin: It may be worth documenting whether these are the TensorFlow protos or
the TensorBoard protos. It looks to me like the code expects TB protos,
but will actually work if the `to_proto` is a TF proto and the
`from_proto` has no nodes and no library, or the `to_proto` is a TB
proto and the `from_proto` is a TF proto with no nodes and an
absent-or-empty library.

wchargin: nit: Capitalize *a* in “a destination […]”, “a GraphDef […]”.

tensorboard/plugins/graph/graph_util.py
stephanwlee: Not sure if this directly answers your question but we need something that can either render tree of graphs or combine the graphs perfectly. This is the first very naive attempt and we intend to iterate on it when we face issues after the alpha.

Now more concrete answers and assumptions:
- Akshay tells me that although nothing enforces and guarantees the global uniqueness, he says we are making every op's name globally unique. Perhaps this can change in the future and we will have to devise much more sophisticated strategy in that case. I have not found anything (did not spend time reading the code) that corroborates Akshay so I decided not to write that down in comment or anything.
- Graphlib (I may be wrong) seems to take the last instance of the node if the names (used as ID) collides. To answer your general question and your question in comment below here,
```
GraphDefA {
  nodeA {}
  nodeB {
    input: "node_a"
  }
}
GraphDefB {
  nodeA {}
  nodeC {
    input: "node_a"
  }
  nodeB {
    input: "node_c"
  }
}
```
will be combined like this
```
GraphDefComb {
  nodeA {}
  nodeB {
    input: "node_a"
  }
  nodeA {}
  nodeC {
    input: "node_a"
  }
  nodeB {
    input: "node_c"
  }
}
```
and it will visually render
```
nodeA {}
nodeC {
  input: "node_a"
}
nodeB {
  input: "node_c"
}
a -> c -> b
```

> In that case, we must verify that this assumption holds, and
raise a `ValueError` if the key sets are not disjoint.

That indeed is my assumption but I am a bit scared of adding invariants just yet since I am operating on very weakly proved assumptions and iterate on this.

wchargin: > > In that case, we must verify that this assumption holds, and
> > raise a `ValueError` if the key sets are not disjoint.
> 
> That indeed is my assumption but I am a bit scared of adding
> invariants just yet since I am operating on very weakly proved
> assumptions and iterate on this.

Then imho this is the _perfect_ place to assert that invariant: we have
an assumption and we want it to hold but we’re not actually sure if it
does. If it doesn’t, we sure want to know about it! The alternative is
for the code to silently not just “do the wrong thing” but “answer a
question for which we don’t even understand want the right answer
_should_ be”, leaving us to debug this down the line with far less
information.

wchargin: To be clear, the intent is for the graphs to always be rendered as
disjoint, right?

stephanwlee: > To be clear, the intent is for the graphs to always be rendered as
disjoint, right?

If there are multiple of them, yes (unless there is a use case we need not consider)

stephanwlee: Added a basic invariance. I honestly dunno what is supposed to happen for Function Libraries so I just kept that without any check.

tensorboard/plugins/graph/graph_util.py
wchargin: Resolved: key sets must be disjoint.

wchargin:
> * Akshay tells me that although nothing enforces and guarantees the
> global uniqueness

For what it’s worth, the proto definition explicitly says that node
names are “[u]nique within a single GraphDef”:
<https://github.com/tensorflow/tensorboard/blob/638b3664f32e8f74660abc33460e2589586ca687/tensorboard/compat/proto/node_def.proto#L12-L15>

This suggests to me that this is _not_ likely to change in the future
without warning, as it would be a breaking change. It also means that we
really should not emit `GraphDef`s that violate this invariant.

> I honestly dunno what is supposed to happen for Function Libraries so
> I just kept that without any check.

This worries me. How can we be confident that this code is correct if we
don’t even know what the correct behavior is? Can we find some
real-world examples that have large trees of graphs including nontrivial
function libraries and see what properties they have? Can we ask around?

From looking at [`function.proto`], there indeed is no explicit
disclaimer like in `graph.proto` as discussed above. But the
`GradientDef` is clearly keyed against function names, and I don’t see
how this makes sense unless function names are also unique. I also don’t
see how it could make sense for there to be two `GradientDef`s with the
same `function_name`.

[`function.proto`]: https://github.com/tensorflow/tensorboard/blob/638b3664f32e8f74660abc33460e2589586ca687/tensorboard/compat/proto/function.proto

stephanwlee:
>> * Akshay tells me that although nothing enforces and guarantees the
>>   global uniqueness
> For what it’s worth, the proto definition explicitly says that node
> names are “[u]nique within a single GraphDef”:

I meant global uniqueness across GraphDefs within an model execution (even across multiple "session.run"s)

tensorboard/plugins/graph/graph_util.py
stephanwlee: I was incorrect. Our hierarchy builder or graphlib (cannot tell who is doing this atm) seems to combine the NodeDef like this:
image

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,88 @@
...
+    to_proto_list: A dict that maps key to a proto.
+    proto: A proto.
+    get_key: A lambda that returns a string key from a proto.
+    error_msg: User friendly error message.
+
+  Raises:
+    ValueError when there is the key in the dict but contents mismatch.
+
+  Returns:
+    True if key from proto is present in the key_to_proto.
```
wchargin: Please remove this `Returns` clause.

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,88 @@
...
+  a ValueError.
+
+  Args:
+    to_proto_list: A dict that maps key to a proto.
+    proto: A proto.
+    get_key: A lambda that returns a string key from a proto.
+    error_msg: User friendly error message.
+
+  Raises:
+    ValueError when there is the key in the dict but contents mismatch.
```
wchargin: Please mention that this includes the provided `error_msg`.

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,88 @@
...
+
+  Returns:
+    True if key from proto is present in the key_to_proto.
+  """
+  key_to_proto = {}
+  for proto in to_proto_list:
+    key = get_key(proto)
+    key_to_proto[key] = proto
```
wchargin: Is it a precondition of this function that no single input list contain
two elements with the same key? (As I understand the context, we should
be able to maintain this invariant.) If so, let’s document and enforce
that.

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,88 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Utilities for graph plugin."""
+from tensorboard.compat.proto import function_pb2
+
+
+def _safe_copy_proto_list_values(to_proto_list, from_proto_list, get_key, error_msg):
```
wchargin: Suggest using `dst` and `src` instead of `to` and `from` (but I don’t
feel too strongly about this).

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,88 @@
...
+  key_to_proto = {}
+  for proto in to_proto_list:
+    key = get_key(proto)
+    key_to_proto[key] = proto
+
+  for proto in from_proto_list:
+    key = get_key(proto)
+    if key in key_to_proto:
+      if proto != key_to_proto.get(key):
+        raise ValueError(error_msg + ': %s' % key)
```
wchargin: It’s not common to mix string concatenation and string formatting (not
least because you have to think about the precedence). Suggest:

```py
raise ValueError("%s: %s" % (error_msg, key))
```

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def _safe_copy_proto_list_values(to_proto_list, from_proto_list, get_key, error_msg):
+  """Safely copies the value from from_proto_list to to_proto_list.
+
+  Copies if to_proto_list does not contain an item with the same key. In case an item
+  is found with the same key, it checks whether contents match and, if not, it raises
+  a ValueError.
+
+  Args:
```
wchargin: These args don’t look right? Maybe:

```python
  """...

  Args:
    dst_proto_list: A `RepeatedCompositeContainer` or
      `RepeatedScalarContainer` into which values should be copied.
    src_proto_list: A container holding the same kind of values as in
      `dst_proto_list` from which values should be copied.
    get_key: A function that takes an element of `dst_proto_list` or
      `src_proto_list` and returns a key, such that if two elements have
      the same key then it is required that they be deep-equal. For
      instance, if `dst_proto_list` is a list of nodes, then `get_key`
      might be `lambda node: node.name` to indicate that if two nodes
      have the same name then they must be the same node. All keys must
      be hashable.
    error_msg: A user-friendly error message to be raised in the case
      that two nodes share a key but are distinct.
  """
```

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,88 @@
...
+      to_proto.node,
+      from_proto.node,
+      lambda n: n.name,
+      'Cannot combine GraphDefs because nodes share a name but are different')
+
+  _safe_copy_proto_list_values(
+      to_proto.library.function,
+      from_proto.library.function,
+      lambda n: n.signature.name,
+      'Cannot combine GraphDefs because nodes share a name but are different')
```
wchargin: Is this meant to say, “[…] functions share a name […]”?

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,88 @@
...
+
+  _safe_copy_proto_list_values(
+      to_proto.library.gradient,
+      from_proto.library.gradient,
+      lambda g: g.gradient_func,
+      ('Cannot combine GraphDefs because gradients share a gradient_func name '
+      'but maps to a different function'))
```
wchargin: nit: add one space of hanging indentation

tensorboard/plugins/graph/graph_util_test.py
```diff
@@ -0,0 +1,488 @@
...
+      versions {
+        producer: 21
+      }
+    ''', graph_def_b)
+
+    self.assertProtoEquals(
+        expected_proto,
+        graph_util.combine_graph_defs(graph_def_a, graph_def_b))
+
+  def test_combine_graph_defs_name_collided_differnt_content(self):
```
wchargin: nit: *different* (sp.)

tensorboard/plugins/graph/graph_util_test.py
```diff
@@ -0,0 +1,488 @@
...
+            input_arg {
+              name: "x"
+              type: DT_HALF
+            }
+            output_arg {
+              name: "identity"
+              type: DT_HALF
+            }
+          }
+          node_def {
```
wchargin: Not super salient to the test, but is this a valid function definition?
The actual node def takes `x: f16` and `y: f16` to `Add: f16`, but the
signature is `(x: f16) -> identity: f16`.

tensorboard/plugins/graph/graph_util_test.py
```diff
@@ -0,0 +1,488 @@
...
+            input: "y"
+          }
+        }
+      }
+    ''', graph_def_b)
+
+    with six.assertRaisesRegex(
+        self,
+        ValueError,
+        r'Cannot combine GraphDefs because functions share a name but are different. foo'):
```
wchargin: I don’t understand how this test is passing: the string “functions share
a name” doesn’t appear anywhere in the code under test. What’s going on?

wchargin: nit: Also, please wrap to within 80.

tensorboard/plugins/graph/graph_util.py
wchargin: (If you don’t like the use of `Repeated*Container`, feel free to
substitute a more generic term like “proto repeated field container”.)

tensorboard/plugins/graph/graph_util_test.py
stephanwlee: Structure-wise it is a valid but it would never be produced by TF :)
Did not focus too much on TF correctness.

tensorboard/plugins/graph/graph_util_test.py
stephanwlee: I had a spec with the same name and second one ran but not this one.

tensorboard/plugins/graph/graph_util.py
stephanwlee: Not sure if I want to add that invariance atm. I will refactor later to raise different error in that invariance violence presumably with different messages (and tests).

tensorboard/plugins/graph/graph_util_test.py
wchargin: Ugh, *that* particular nasty corner of Python. Good catch; thanks.

We should make that a lint error. It looks like we don’t have any such
violations in HEAD:

```
$ git ls-files '*_test.py' "$filename" | while read filename; do sed -n 's/.*\(def test[a-zA-Z0-9_]*\)(.*/\1/p; /\bclass\b/p' "$filename" | awk -v "filename=${filename}" '$1 == "class" { for (k in c) c[k] = 0; next; } $1 == "def" { if (c[$2]++) printf "%s: %s\n", filename, $2; next } { printf "unknown token: %s\n", $1 > "/dev/stderr"; exit 2; }'; done | diff -u /dev/null -
```

tensorboard/plugins/graph/graph_util_test.py
wchargin: Cool; that is fine with me.

stephanwlee:
Chatted offline and shared my assumption around the implementation. We decided to add invariants so when violated, users can file bugs and we can address them later.

@wchargin PTAL.

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,145 @@
...
+    get_key: A function that takes an element of `dst_proto_list` or
+      `src_proto_list` and returns a key, such that if two elements have
+      the same key then it is required that they be deep-equal. For
+      instance, if `dst_proto_list` is a list of nodes, then `get_key`
+      might be `lambda node: node.name` to indicate that if two nodes
+      have the same name then they must be the same node. All keys must
+      be hashable.
+
+  Raises:
+    _ProtoListNonUniqueKeyError: a proto_list contains items with duplicate
```
wchargin: nit: Capitalize the first word after a colon in a docstring per Google
style. (also next line)

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,145 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Utilities for graph plugin."""
+from tensorboard.compat.proto import function_pb2
+
+
+class _ProtoListNonUniqueKeyError(Exception):
```
wchargin: IMO, the phrase “duplicate” is easier to immediately understand than
“non-unique”.

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,145 @@
...
+      be hashable.
+
+  Raises:
+    _ProtoListNonUniqueKeyError: a proto_list contains items with duplicate
+      keys.
+    _SameKeyDiffContentError: an item with the same key has different contents.
+  """
+
+  def _assert_proto_container_unique_keys(proto_list, get_key):
+    """Asserts whether proto_list only contains unique keys.
```
wchargin: nit: You assert *that* it contains only unique keys (a judgment), not
*whether* it does (a boolean).

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Args:
```
wchargin: Given that this is a nested helper function in a private function, and
that the meanings of its parameters are clear from context, I think it’s
fine to omit the “Args:” section. If you prefer to include it, please
actually fill it out.

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class _SameKeyDiffContentError(Exception):
+  pass
+
+
+def _safe_copy_proto_list_values(dst_proto_list, src_proto_list, get_key):
+  """Safely copies the value from src_proto_list to dst_proto_list.
+
+  First checks whether keys in both dst_proto_list and src_proto_list are
```
wchargin: As written, this paragraph defines _how_ the function is implemented,
not _what_ its contract is. The person reading these docs doesn’t care
what happens first or second (or they’d look at the code); they need to
know what the function _requires_ about its arguments and what it
_ensures_ about its result.

Suggest something like:

> ```python
>   """Safely merge values from `src_proto_list` into `dst_proto_list`.
>
>   Each element in `dst_proto_list` must be mapped by `get_key` to a key
>   value that is unique within that list; likewise for `src_proto_list`.
>   If an element of `src_proto_list` has the same key as an existing
>   element in `dst_proto_list`, then the elements must also be equal.
>
>   ...
>   """
> ```

The precise exceptions raised don’t need to be specified in the first
paragraph; they’re already in the “Raises” section.

tensorboard/plugins/graph/graph_util.py
```diff
@@ -0,0 +1,145 @@
...
+        to_proto.library.gradient,
+        from_proto.library.gradient,
+        lambda g: g.gradient_func)
+  except _ProtoListNonUniqueKeyError as exc:
+    raise ValueError(
+      'A GraphDef contains non-unique gradient function names: %s' % exc)
+  except _SameKeyDiffContentError as exc:
+    raise ValueError(
+      ('Cannot combine GraphDefs because gradients share a gradient_func name '
+       'but maps to a different function: %s') % exc)
```
wchargin: nit: s/but maps to different a function/but map to different functions/

tensorboard/plugins/graph/graph_util_test.py
```diff
@@ -214,7 +214,84 @@ def test_combine_graph_defs_name_collided_different_content(self):
...
+      }
+      versions {
+        producer: 21
+      }
+    ''', graph_def_b)
+
+    with six.assertRaisesRegex(
+        self,
+        ValueError,
+        ('A GraphDef contains non-unique node names: X')):
```
wchargin: The inputs in this test case are problematic for two reasons: the first
graph has multiple nodes called “X”, and the second graph has a node “X”
that is different from the “X” nodes in the first graph. It seems to me
that the `combine_graph_defs` function would be within its rights to
raise _either_ of the relevant exceptions. Can we amend this test so
that the inputs have exactly one problem? (e.g., remove “X” from the
second graph)

wchargin: nit: Spurious parentheses.

PR #1875, opened 2019-02-22
Docs scalars

manivaradarajan:
This is a very quick introduction to viewing default scalar summaries when developing a Keras model in TensorFlow. It's intended to be very simple to show that virtually no extra code is required to get started using TensorBoard to see how loss is changing during training.

To some extent this repeats some of the material in the "Quick Start", but since it's very focussed I don't think the repetition is a bad thing.

I'm going to followup with a similar notebook showing how to use the Scalar Summary API itself, when using GradientTape.

I'm asking for a preliminary review to make sure this style of notebook tutorial is on the right track. I will updat the yaml in a followon.


manivaradarajan:
Sent incorrectly. I'll re-request this in a bit.

PR #1879, opened 2019-02-22
Graph plugin populates graph and RunMeta v2.

stephanwlee:
This change pipes more data into the graph plugin that is persisted
on the TensorFlow's side with the new summary ops.

Also refactored the test a bit but it is hard to test anything
because the nightlies do not contain the required changes yet.
In a follow up changes, we will add test using more public APIs.

The PR as is contains #1873. Please review the last commit starting from 
face82a9.

### DIFF
--- a/tensorboard/plugins/graph/graphs_plugin.py
+++ b/tensorboard/plugins/graph/graphs_plugin.py
@@ -18,13 +18,18 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import json
 import six
 from werkzeug import wrappers
 
 from tensorboard.backend import http_util
 from tensorboard.backend import process_graph
 from tensorboard.backend.event_processing import plugin_event_accumulator as event_accumulator  # pylint: disable=line-too-long
+from tensorboard.compat.proto import config_pb2
+from tensorboard.compat.proto import graph_pb2
 from tensorboard.plugins import base_plugin
+from tensorboard.plugins.graph import graph_util
+from tensorboard.plugins.graph import keras_util
 from tensorboard.util import tb_logging
 
 logger = tb_logging.get_logger()
@@ -34,8 +39,12 @@ _PLUGIN_PREFIX_ROUTE = 'graphs'
 # The Summary API is implemented in TensorFlow because it uses TensorFlow internal APIs.
 # As a result, this SummaryMetadata is a bit unconventional and uses non-public
 # hardcoded name as the plugin name. Please refer to link below for the summary ops.
+# https://github.com/tensorflow/tensorflow/blob/11f4ecb54708865ec757ca64e4805957b05d7570/tensorflow/python/ops/summary_ops_v2.py#L757
+_PLUGIN_NAME_RUN_METADATA = 'graph_run_metadata'
 # https://github.com/tensorflow/tensorflow/blob/11f4ecb54708865ec757ca64e4805957b05d7570/tensorflow/python/ops/summary_ops_v2.py#L788
 _PLUGIN_NAME_RUN_METADATA_WITH_GRAPH = 'graph_run_metadata_graph'
+# https://github.com/tensorflow/tensorflow/blob/565952cc2f17fdfd995e25171cf07be0f6f06180/tensorflow/python/ops/summary_ops_v2.py#L825
+_PLUGIN_NAME_KERAS_MODEL = 'graph_keras_model'
 
 
 class GraphsPlugin(base_plugin.TBPlugin):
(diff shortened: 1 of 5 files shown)


tensorboard/plugins/graph/graphs_plugin_v2_test.py
```diff
@@ -0,0 +1,104 @@
...
+import os.path
+
+from google.protobuf import text_format
+import tensorflow as tf
+
+from tensorboard.compat.proto import graph_pb2
+from tensorboard.plugins.graph import graphs_plugin_test
+
+
+tf.enable_v2_behavior()
```
nfelt: Is this line necessary?  If this test will pass under TF 1.x as well that's ideal, because then when we set up tests against TF 1.x and 2.0 the tests will run using the default behavior in both cases (as opposed to being forced into 2.0 behavior in both).

tensorboard/plugins/graph/graphs_plugin_v2_test.py
```diff
@@ -0,0 +1,104 @@
...
+            write_graph=include_graph)])
+
+  def _get_graph(self, *args, **kwargs):
+    """Fetch and return the graph as a proto."""
+    (graph_pbtxt, mime_type) = self.plugin.graph_impl(*args, **kwargs)
+    self.assertEqual(mime_type, 'text/x-protobuf')
+    return text_format.Parse(graph_pbtxt, graph_pb2.GraphDef())
+
+  def test_info(self):
+    raise self.skipTest('tf-nightly==20190222 does not write a conceptual graph.')
```
nfelt: Can you add a TODO to enable this?  Maybe e.g. as

`raise self.skipTest('TODO: enable this after tf-nightly writes a conceptual graph')`

tensorboard/plugins/graph/graphs_plugin_v2_test.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
```
nfelt: Not necessary?

tensorboard/plugins/graph/graphs_plugin_v2_test.py
stephanwlee: Argh, I realized why Keras was not writing the conceptual graph. It is implemented in callback_v2 and I am not sure doing `enable_v2_behavior` will load that symbol. Will report back what I can do.

nfelt: Can you just refer to `tf.compat.v2.keras.callbacks.TensorBoard` instead?

tensorboard/plugins/graph/graphs_plugin.py
```diff
@@ -108,10 +139,25 @@ def add_row_item(run, tag=None):
...
-  def graph_impl(self, run, limit_attr_size=None, large_attrs_key=None):
+  def graph_impl(self, run, tag, conceptual, limit_attr_size=None, large_attrs_key=None):
     """Result of the form `(body, mime_type)`, or `None` if no graph exists."""
     try:
-      graph = self._multiplexer.Graph(run)
+      if conceptual:
+        tensor_events = self._multiplexer.Tensors(run, tag)
+        # Take the first event if there are multiple events written from different
+        # steps.
+        keras_model_config = json.loads(tensor_events[0].tensor_proto.string_val[0])
```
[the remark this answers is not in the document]
stephanwlee: I hope to approach this holistically later since our exception handling on our backend/frontend generally lack. As is, it will result in HTTP 500 with some logs in the console.

tensorboard/plugins/graph/graphs_plugin_v2_test.py
stephanwlee: Doh, I am quite dumb. I did check the source of callbacks.py in the virtualenv and confirmed that the new code did not land in the tf-nightly.

tensorboard/plugins/graph/graphs_plugin.py
```diff
@@ -170,10 +174,14 @@ def run_metadata_impl(self, run, tag):
     try:
       run_metadata = self._multiplexer.RunMetadata(run, tag)
     except ValueError:
+      # TODO(stephanwlee): Should include whether FE is fetching for v1 or v2 RunMetadata
+      # so we can remove this try/except.
       tensor_events = self._multiplexer.Tensors(run, tag)
-      # Take the first event if there are multiple events written from different
-      # steps.
-      run_metadata = RunMetadata.FromString(tensor_events[0].tensor_proto.string_val[0])
+      if len(tensor_events) > 0:
```
nfelt: This needs an else branch that also sets run_metadata otherwise it will fail at the `if run_metadata is None` because the variable won't be defined.  Either that or you can check `if not tensor_events: return None`.

PR #1883, opened 2019-02-22
Test build_pip_package.sh on Travis

nfelt:
This PR adds `bazel run //tensorboard/pip_package:build_pip_package` to our Travis CI testing script.

This will allow us to ensure the pip package builds and passes basic smoke tests as a presubmit check, which in turn unblocks adding minimal TF + TB integration testing for the tf.summary module (in a future PR).  The smoke tests run quickly - less than 30 seconds on my machine - and seem pretty stable, though they do launch the TB webserver very briefly so that's a potential source of flakiness.  If it proves too flaky we can look into maybe running just the non-webserver parts of the smoke testing.

In order to get the smoke test embedded in build_pip_package.sh to run appropriately in our Travis environment, using the Travis env's TF version and python version, I added some arguments controlling the smoke tests to the script and generally cleaned it up (e.g. moving the smoke test the end; adding some messages when the tests are running; etc.).

Test plan: we'll find out when I create this PR if the Travis part works :)  For the script changes, I've tested that the generated pip wheel files have identical contents before and after this change by unzipping them into a temporary git repo, committing them, and diffing the commits (it's not sufficient to just hash the .whl because zip files include timestamps).

### DIFF
--- a/tensorboard/pip_package/build_pip_package.sh
+++ b/tensorboard/pip_package/build_pip_package.sh
@@ -24,7 +24,7 @@ else
 fi
 
 tf_version="tf-nightly"
-if [ -n "$TF_VERSION" ]; then
+if [ -n "${TF_VERSION}" ]; then
   tf_version="tensorflow==${TF_VERSION}"
 fi
 smoke="all"
@@ -116,32 +116,33 @@ python setup.py bdist_wheel --python-tag py2 >/dev/null
 python setup.py bdist_wheel --python-tag py3 >/dev/null
 
 smoke() {
-  pymajorversion="$1"
-  if [ -z "${pymajorversion}" ]; then
-    pymajorversion="$(python -c 'import sys; print(sys.version_info[0])')"
+  py_major_version="$1"
+  if [ -z "${py_major_version}" ]; then
+    py_major_version="$(python -c 'import sys; print(sys.version_info[0])')"
   fi
-  smokepython="python$1"
-  smokevenv="smoke-venv$1"
-  smoketf="$2"
+  smoke_python="python$1"
+  smoke_venv="smoke-venv$1"
+  smoke_tf="$2"
   set +x
   printf '\n\n%70s\n' | tr ' ' '='
-  printf "Smoke testing with ${smokepython} and ${smoketf}...\n\n"
+  echo "Smoke testing with ${smoke_python} and ${smoke_tf}..."
+  printf '\n'
   set -x
-  command -v "${smokepython}" >/dev/null
-  virtualenv -qp "${smokepython}" "${smokevenv}"
-  cd "${smokevenv}"
+  command -v "${smoke_python}" >/dev/null
+  virtualenv -qp "${smoke_python}" "${smoke_venv}"
+  cd "${smoke_venv}"
   . bin/activate
   pip install -qU pip
-  pip install -qU "${smoketf}"
-  pip install -qU ../dist/*"py${pymajorversion}"*.whl >/dev/null
+  pip install -qU "${smoke_tf}"
+  pip install -qU ../dist/*"py${py_major_version}"*.whl >/dev/null
   # Test TensorBoard application
   [ -x ./bin/tensorboard ]  # Ensure pip package included binary
   mkfifo pipe
   tensorboard --port=0 --logdir=smokedir 2>pipe &
   perl -ne 'print STDERR;/http:.*:(\d+)/ and print $1.v10 and exit 0' <pipe >port
-  curl -fs http://localhost:$(cat port) >index.html
+  curl -fs "http://localhost:$(cat port)" >index.html
   grep '<tf-tensorboard' index.html
-  curl -fs http://localhost:$(cat port)/data/logdir >logdir.json
+  curl -fs "http://localhost:$(cat port)/data/logdir" >logdir.json
   grep 'smokedir' logdir.json
   kill $!
   # Test TensorBoard APIs
@@ -155,7 +156,7 @@ import tensorboard.summary._tf.summary as tf_summary
 "
   deactivate
   cd ..
-  rm -rf "${smokevenv}"
+  rm -rf "${smoke_venv}"
 }
 
 case "${smoke}" in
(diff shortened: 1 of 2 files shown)


tensorboard/pip_package/build_pip_package.sh
```diff
@@ -23,60 +23,40 @@ else
   sedi="sed -i"
 fi
 
-run_smoke_test=1
+tf_version="tf-nightly"
+if [ -n "$TF_VERSION" ]; then
```
wchargin: nit: `${TF_VERSION}` for consistency and per style:
<https://google.github.io/styleguide/shell.xml>

tensorboard/pip_package/build_pip_package.sh
```diff
@@ -135,9 +115,66 @@ pip install -qU wheel 'setuptools>=36.2.0'
...
+  pymajorversion="$1"
+  if [ -z "${pymajorversion}" ]; then
+    pymajorversion="$(python -c 'import sys; print(sys.version_info[0])')"
+  fi
+  smokepython="python$1"
+  smokevenv="smoke-venv$1"
+  smoketf="$2"
+  set +x
+  printf '\n\n%70s\n' | tr ' ' '='
+  printf "Smoke testing with ${smokepython} and ${smoketf}...\n\n"
```
wchargin: These should be format specifiers rather than part of the format string:

```sh
printf 'Smoke testing with %s and %s...\n' "${smokepython}" "${smoketf}"
```

tensorboard/pip_package/build_pip_package.sh
```diff
@@ -135,9 +115,66 @@ pip install -qU wheel 'setuptools>=36.2.0'
...
-  smoke 2
-  smoke 3
-fi
+smoke() {
+  pymajorversion="$1"
+  if [ -z "${pymajorversion}" ]; then
+    pymajorversion="$(python -c 'import sys; print(sys.version_info[0])')"
+  fi
+  smokepython="python$1"
+  smokevenv="smoke-venv$1"
```
wchargin: As fun as it is to always parse these names as smock-ev-env and
smocket-eff (a new addition to the socket(2) family?), please add
underscores per style.

tensorboard/pip_package/build_pip_package.sh
```diff
@@ -135,9 +115,66 @@ pip install -qU wheel 'setuptools>=36.2.0'
...
+  . bin/activate
+  pip install -qU pip
+  pip install -qU "${smoketf}"
+  pip install -qU ../dist/*"py${pymajorversion}"*.whl >/dev/null
+  # Test TensorBoard application
+  [ -x ./bin/tensorboard ]  # Ensure pip package included binary
+  mkfifo pipe
+  tensorboard --port=0 --logdir=smokedir 2>pipe &
+  perl -ne 'print STDERR;/http:.*:(\d+)/ and print $1.v10 and exit 0' <pipe >port
+  curl -fs http://localhost:$(cat port) >index.html
```
wchargin: As you rewrite this, please fix it: `$(cat port)` must be quoted
(personally I’d quote the whole thing). (Here and below.)

tensorboard/pip_package/build_pip_package.sh
nfelt: Honest question: why is that better?  It's just harder to keep in sync both from a code standpoint and logically since you need to match the %s with the arguments, and it's more verbose.  When you have native string interpolation, why not use it?

tensorboard/pip_package/build_pip_package.sh
nfelt: Ok, but FWIW, all of these lines are only moved, not rewritten.

tensorboard/pip_package/build_pip_package.sh
wchargin: If you prefer

    printf '%s\n\n' "Smoke testing with ${smokepython} and ${smoketf}..."

then that’s also totally fine with me. But you don’t interpolate into
format strings (e.g., `smokepython` or `smoketf` could have percent
signs, which would screw everything up).

tensorboard/pip_package/build_pip_package.sh
nfelt: Done.

tensorboard/pip_package/build_pip_package.sh
nfelt: Fair enough, I wasn't considering them containing format specifiers.  Changed to use echo.

tensorboard/pip_package/build_pip_package.sh
nfelt: Done.

tensorboard/pip_package/build_pip_package.sh
nfelt: Done.

tensorboard/pip_package/build_pip_package.sh
wchargin: Okay. Just to share where I’m coming from: I personally always avoid
`echo` because it is so ill-defined. The `echo -n` behavior is
unportable, but you can’t use `--` in the way that you can with standard
utilities. The behavior of `echo -e '\x1b'` is unspecified (in fact, the
behavior is implementation-defined if _any_ argument contains a
backslash), but the behavior of `echo -e foo` _is_ specified and most
implementations get it wrong. Implicitly inserting single spaces is not
the end of the world, but I’d prefer to do that explicitly. Et cetera.

I am much happier sticking to `printf` for all my printing needs, but as
a rule I don’t usually request that others do the same.

wchargin: Oh, TIL that the POSIX spec actually recommends that new applications
use `printf` instead of `echo`:
<https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html#tag_20_37_16>

nfelt: Feel free to update it to use `printf '%s\n'` instead.  I think shell scripts are the wrong tool for the job if we're getting really concerned about how we handle pathological inputs, so I'm not inclined to worry about this much further myself.

PR #1887, opened 2019-02-25
Stop setting PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION env var

### LINKED ISSUE #1566: ImportError: cannot import name '_message' from 'google.protobuf.pyext' (/usr/lib/python3.7/site-packages/google/protobuf/pyext/__init__.py)
- TensorBoard version: tensorboard-1.12.0
 - GNU/Linux, Manjaro
 - Python version: 3.7.1

Issue: Just calling tensorboard fails to load. A bug with protobuf?

```
  File "/usr/bin/tensorboard", line 11, in <module>
    load_entry_point('tensorboard==1.12.0', 'console_scripts', 'tensorboard')()
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 484, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2725, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2343, in load
    return self.resolve()
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2349, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python3.7/site-packages/tensorboard/main.py", line 45, in <module>
    from tensorboard import default
  File "/usr/lib/python3.7/site-packages/tensorboard/default.py", line 34, in <module>
    import tensorflow as tf
  File "/usr/lib/python3.7/site-packages/tensorflow/__init__.py", line 22, in <module>
    from tensorflow.python import pywrap_tensorflow  # pylint: disable=unused-import
  File "/usr/lib/python3.7/site-packages/tensorflow/python/__init__.py", line 52, in <module>
    from tensorflow.core.framework.graph_pb2 import *
  File "/usr/lib/python3.7/site-packages/tensorflow/core/framework/graph_pb2.py", line 6, in <module>
    from google.protobuf import descriptor as _descriptor
  File "/usr/lib/python3.7/site-packages/google/protobuf/descriptor.py", line 47, in <module>
    from google.protobuf.pyext import _message
ImportError: cannot import name '_message' from 'google.protobuf.pyext' (/usr/lib/python3.7/site-packages/google/protobuf/pyext/__init__.py)

```

nfelt:
This should fix #1566 by reverting #1489, which was unfortunately a misguided change on my part.  I thought that the env var was necessary in order to opt in to the C++ implementation because not all protobuf python distributions have the C++ implementation activated by default even if they include the appropriate extension.  Apparently this either was never true, or certainly hasn't been true since protobuf 3.5.1 (and we're already on 3.6.0).  And setting the env var is actively harmful when protobuf doesn't include the C++ extension because then it fails with the error in #1566.  (Protobuf should really provide a safe way to opt into the C++ implementation only if it is present, but they currently don't.)

I think I thought this was necessary because apparently when you `bazel build tensorboard` the resulting binary uses the bazel-provided `protobuf-python` and that does seem to include the C++ extension but still default to the python implementation (which is why this env var made a difference in my limited initial testing; I just didn't realize I needed to test it against the pip package version).

We should figure out a way to make the locally built `tensorboard` fast that doesn't require setting this env vars unconditionally, but I'm fixing this first to unbreak users in #1566.

Tested by building pip package and confirming that the selected protobuf implementation is still CPP and that when profiling event loading, it shows the C++ extension signatures in the output.

### DIFF
--- a/tensorboard/main.py
+++ b/tensorboard/main.py
@@ -33,11 +33,6 @@ import os
 #   https://github.com/tensorflow/tensorboard/issues/1225
 # This must be set before the first import of tensorflow.
 os.environ['GCS_READ_CACHE_DISABLED'] = '1'
-
-# Use fast C++ implementation of Python protocol buffers. See:
-# https://github.com/protocolbuffers/protobuf/blob/v3.6.0/python/google/protobuf/pyext/README
-os.environ['PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION'] = 'cpp'
-os.environ['PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION_VERSION'] = '2'
 # pylint: enable=g-import-not-at-top
 
 import sys


nfelt:
Filed https://github.com/tensorflow/tensorboard/issues/1888 for making `bazel build tensorboard` use the fast C++ protobuf impl, without having to set these env vars.

PR #694, opened 2017-11-02
Add animated gif

olgabot:
Adds an animated GIF to the README so newcomers to the package can get a sense of what TensorBoard is without having to install it.

PR #1890, opened 2019-02-25
Update the index page description and add a GIF

GalOshri:
Update the TensorBoard docs index page:
- Description includes brief itemization of some of TensorBoard's capabilities
- GIF shows some of TensorBoard's capabilities
- Adding a link to a Google AI blog on TensorBoard's WIT plugin

### DIFF
--- a/docs/_index.yaml
+++ b/docs/_index.yaml
@@ -13,7 +13,7 @@ landing_page:
         <ul style="padding-left: 20px; margin: 20px 0;">
           <li>Tracking and visualizing metrics such as loss and accuracy</li>
           <li>Visualizing the model graph (ops and layers)</li>
-          <li>Viewing histograms and distributions of Tensors (e.g. weights 
+          <li>Viewing histograms and distributions of Tensors (e.g., weights 
           and biases) to track how they change over time</li>
           <li>Projecting embeddings to a lower dimensional space</li>
           <li>Displaying images, text, and audio data</li>
--- a/docs/_index.yaml
+++ b/docs/_index.yaml
@@ -8,22 +8,36 @@ landing_page:
     items:
     - classname: devsite-landing-row-50
       description: >
-        The computations you'll use TensorFlow for—like training a massive deep
-        neural network—can be complex and confusing. To make it easier to
-        understand, debug, and optimize TensorFlow programs, we've included a
-        suite of visualization tools called TensorBoard.
+        TensorBoard provides the visualization and tooling needed for 
+        machine learning experimentation. It enables:
+        <ul style="padding-left: 20px; margin: 20px 0;">
+          <li>Tracking and visualizing metrics such as loss and accuracy</li>
+          <li>Visualizing the model graph (ops and layers)</li>
+          <li>Viewing histograms and distributions of Tensors (e.g. weights 
+          and biases) to track how they change over time</li>
+          <li>Projecting embeddings to a lower dimensional space</li>
+          <li>Displaying images, text, and audio data</li>
+          <li>Profiling TensorFlow programs</li>
+          <li>And much more</li>
+        </ul>
 
-      image_path: /images/graph_vis_animation.gif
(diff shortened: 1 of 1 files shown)


lamberta:
@MarkDaoust 
There's a ~2mb limit for images before we need to host them directly in tensorflow.org. Mark to confirm

MarkDaoust:
Most copybara configs have 2Mb set as a limit.
But AFAIK it's just good practice for keeping your repo-size down.

You can submit it (internally) to site-tensorflow/en/images/tensorboard if you want to avoid ballooning the repo.

nfelt:
Justine had some tips for compressing a video file for gif generation back on this old PR: https://github.com/tensorflow/tensorboard/pull/694 (which unfortunately I don't think was ever properly followed up on).

Beyond reducing framerate we might also be able to downsample the color palette, which for non-photographic stuff is generally not going to be very noticeable anyway.

GalOshri:
Thank you for catching this. How do people feel about a 500 KB GIF?

nfelt:
500k seems fine to me personally.  It's certainly below the 2MB limit.

stephanwlee:
> Most copybara configs have 2Mb set as a limit.

But it says 2Mb not 2MB.

nfelt:
I think we're all talking about megabytes here.

MarkDaoust:
Yeah, that was a typo.  MB.

GalOshri:
I updated the GIF.

> Yeah, that was a typo. MB.

Great pun if intended.

docs/_index.yaml
```diff
@@ -8,22 +8,36 @@ landing_page:
...
+          <li>Viewing histograms and distributions of Tensors (e.g. weights 
+          and biases) to track how they change over time</li>
+          <li>Projecting embeddings to a lower dimensional space</li>
+          <li>Displaying images, text, and audio data</li>
+          <li>Profiling TensorFlow programs</li>
+          <li>And much more</li>
+        </ul>
 
-      image_path: /images/graph_vis_animation.gif
+      image_path: /images/tensorboard.gif
```
wchargin: We use `/images/*` here but `/resources/images/*` below; is this right?
(I see the `/resources/images/*` files within google3 but I don’t see
where they come from or how our `./images` might be merged into the same
path.)

docs/_index.yaml
```diff
@@ -8,22 +8,36 @@ landing_page:
...
-        The computations you'll use TensorFlow for—like training a massive deep
-        neural network—can be complex and confusing. To make it easier to
-        understand, debug, and optimize TensorFlow programs, we've included a
-        suite of visualization tools called TensorBoard.
+        TensorBoard provides the visualization and tooling needed for 
+        machine learning experimentation. It enables:
+        <ul style="padding-left: 20px; margin: 20px 0;">
+          <li>Tracking and visualizing metrics such as loss and accuracy</li>
+          <li>Visualizing the model graph (ops and layers)</li>
+          <li>Viewing histograms and distributions of Tensors (e.g. weights
```
wchargin: nit: Add comma after *e.g.*.

docs/_index.yaml
GalOshri: I am not sure. I think the references to `/resources/images/*` were in the original addition from  @lamberta. I'm happy to move the GIF to another folder if it makes sense. Do the other images need to be moved to this repo or would they be added as part of the website publishing?

lamberta: For images you want to keep within your site, creating a local `./images` directory is fine and use a relative path to include it.
For large files, we need to put them in our main site images directory in g3. We can create an `$SITE_ROOT/images/tensorboard/...` directory for you

GalOshri: Per offline conversation, leaving the resource card image as is for now. We'll update this a bit later.

docs/_index.yaml
GalOshri: Never knew this. Updated.

PR #1891, opened 2019-02-26
Make step parameter optional in summary V2 API

nfelt:
This makes the step parameter in summary API V2 functions default to None, which will make them fall back to the to-be-introduced `tf.summary.experimental.get_step()` or raise an exception if that is unset.

See cl/235060769 internally.

Note that this PR won't pass Travis until the aforementioned changes make it into tf-nightly.


nfelt:
Yes, I will hold off until tomorrow and re-run Travis then.  Thanks!

nfelt:
FYI, the TF change didn't make last night's nightly so this will have to wait another day.

nfelt:
PTAL, tests pass locally against latest nightly now so hopefully they'll pass on Travis too.  Minor updates to the tests to reflect the test helper changes in #1908.

PR #1894, opened 2019-02-26
add capture-profile into profile plugin

qqfish:
Allow sending capture profile request via TB

* Screenshots of UI changes
image

### DIFF
--- a/tensorboard/plugins/profile/profile_plugin.py
+++ b/tensorboard/plugins/profile/profile_plugin.py
@@ -27,6 +27,7 @@ import six
 import tensorflow as tf
 from werkzeug import wrappers
 
+from tensorflow.python.eager import profiler_client
 from tensorboard.backend import http_util
 from tensorboard.backend.event_processing import plugin_asset_util
 from tensorboard.plugins import base_plugin
@@ -43,6 +44,7 @@ PLUGIN_NAME = 'profile'
 DATA_ROUTE = '/data'
 TOOLS_ROUTE = '/tools'
 HOSTS_ROUTE = '/hosts'
+CAPTURE_ROUTE = '/capture_profile'
 
 # Available profiling tools -> file name of the tool data.
 _FILE_NAME = 'TOOL_FILE_NAME'
@@ -432,9 +434,35 @@ class ProfilePlugin(base_plugin.TBPlugin):
       return http_util.Respond(request, '404 Not Found', 'text/plain', code=404)
     return http_util.Respond(request, data, 'application/json')
 
+  @wrappers.Request.application
+  def capture_route(self, request):
+    service_addr = request.args.get('service_addr')
+    try:
+      duration = int(request.args.get('duration'))
+    except TypeError:
+      return http_util.Respond(request, 'Invalid duration', 
+                               'text/plain', code=400)
+    is_tpu_name = request.args.get('is_tpu_name') == 'true'
+    if is_tpu_name:
+      tpu_cluster_resolver = (
+          tf.distribute.cluster_resolver.TPUClusterResolver([service_addr]))
+      service_addr = tpu_cluster_resolver.get_master()
+      # TPU cluster resolver always returns port 8470. Replace it with 8466
+      # on which profiler service is running.
+      service_addr = service_addr.replace('grpc://', '').replace(':8470', ':8466')
+    try:
+      profiler_client.start_tracing(service_addr, self.logdir, duration)
+      return http_util.Respond(
+          request, {'result': 'Capture profile successfully. Please refresh'},
+          'application/json')
+    except tf.errors.UnavailableError:
+      return http_util.Respond(request, 'Empty trace result', 
+                               'text/plain', code=404)
+
   def get_plugin_apps(self):
     return {
         TOOLS_ROUTE: self.tools_route,
         HOSTS_ROUTE: self.hosts_route,
         DATA_ROUTE: self.data_route,
+        CAPTURE_ROUTE: self.capture_route,
     }
--- a/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
+++ b/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
@@ -59,6 +59,7 @@ profile run can have multiple tools that present the performance profile as diff
     </p>
   </div>
 </template>
+<paper-toast id="toast" text="" always-on-top></paper-toast>
 <paper-dialog id="captureProfileDialog" modal>
   <paper-input label="Profiler Service URL or TPU name"
                always-float-label
@@ -73,13 +74,13 @@ profile run can have multiple tools that present the performance profile as diff
   </paper-radio-group>
   <paper-input label="Profiling Duration (milliseconds)"
                always-float-label
-               type=number
+               type="number"
                min=1
                value="{{_profileDuration}}"></paper-input>
   <div class="buttons">
     <paper-button dialog-confirm raised
-                  on-tap="captureProfile"
-                  disabled$="[[shouldDisableCaptureProfileButton(profilerServiceAddress, _profileDuration)]]">Capture</paper-button>
+                  on-tap="_captureProfile"
+                  disabled$="[[_shouldDisableCaptureProfileButton(profilerServiceAddress, _profileDuration)]]">Capture</paper-button>
     <paper-button dialog-confirm>Close</paper-button>
   </div>
 </paper-dialog>
@@ -137,7 +138,7 @@ profile run can have multiple tools that present the performance profile as diff
     <div class="allcontrols">
       <div class="sidebar-section">
         <paper-button raised
-                      on-tap="openCaptureProfileDialog">Capture Profile</paper-button>
+                      on-tap="_openCaptureProfileDialog">Capture Profile</paper-button>
       </div>
       <div class="sidebar-section">
         <div class="title">Runs <span class="counter">([[_datasets.length]])</span></div>
@@ -220,7 +221,6 @@ profile run can have multiple tools that present the performance profile as diff
     </template>
   </div>
   </tf-dashboard-layout>
-<paper-toast id="toast" text="" always-on-top></paper-toast>
 </template>
 <style include="dashboard-style"></style>
 
@@ -439,13 +439,13 @@ profile run can have multiple tools that present the performance profile as diff
     detached: function() {
       this.set('_isAttached', false);
     },
-    openCaptureProfileDialog: function() {
+    _openCaptureProfileDialog: function() {
       this.$.captureProfileDialog.open();
     },
-    shouldDisableCaptureProfileButton: function(profilerServiceAddress, profileDuration) {
+    _shouldDisableCaptureProfileButton: function(profilerServiceAddress, profileDuration) {
       return !profilerServiceAddress || profileDuration <= 0;
     },
-    captureProfile: function() {
+    _captureProfile: function() {
       const captureProfileUrl = tf_backend.addParams(
           tf_backend.getRouter().pluginRoute('profile', '/capture_profile'),
           {
--- a/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
+++ b/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
@@ -66,8 +66,7 @@ profile run can have multiple tools that present the performance profile as diff
                placeholder="localhost:6009"
                value="{{_profilerServiceAddress}}"></paper-input>
   <label id="profiler-address-type-label">Address Type:</label>
-  <paper-radio-group selected="ip-address"
-                     selected="{{_profilerServiceAddressType}}"
+  <paper-radio-group selected="{{_profilerServiceAddressType}}"
                      aria-labelledby="profiler-address-type-label">
     <paper-radio-button name="ip-address">IP Address</paper-radio-button>
     <paper-radio-button name="tpu-name">TPU Name</paper-radio-button>
--- a/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
+++ b/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
@@ -20,6 +20,8 @@ limitations under the License.
 <link rel="import" href="../paper-item/paper-item.html">
 <link rel="import" href="../paper-menu/paper-menu.html">
 <link rel="import" href="../paper-progress/paper-progress.html">
+<link rel="import" href="../paper-radio-group/paper-radio-group.html">
+<link rel="import" href="../paper-toast/paper-toast.html">
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../tf-backend/tf-backend.html">
 <link rel="import" href="../tf-dashboard-common/tf-dashboard-layout.html">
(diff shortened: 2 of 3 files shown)


qqfish:
Screenshot:
image

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -327,12 +329,32 @@ def data_route(self, request):
...
+  @wrappers.Request.application
+  def capture_route(self, request):
+    service_addr = request.args.get('service_addr')
+    duration = int(request.args.get('duration'))
+    is_tpu_name = request.args.get('is_tpu_name') == 'true'
+    if is_tpu_name:
+      tpu_cluster_resolver = (
+        tf.contrib.cluster_resolver.TPUClusterResolver([service_addr]))
+      service_addr = tpu_cluster_resolver.get_master()
+    service_addr = service_addr.replace('grpc://', '').replace(':8470', ':8466')
```
stephanwlee: Sorry for my ignorance but (1) where does this port number come from? (2) can we use urlparse to strip out the scheme or at least match with regex like `^grpc://`?

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -327,12 +329,32 @@ def data_route(self, request):
...
+    try:
+      profiler_client.start_tracing(service_addr, self.logdir, duration)
+      return http_util.Respond(
+          request, {'result': 'Capture profile successfully. Please refresh'},
+          'application/json')
+    except tf.errors.UnavailableError:
+      return http_util.Respond(request, {'result': 'Failed to capture profile'},
```
stephanwlee: Hmm please use proper response status code for such error. 404 would be appropriate AFAICT

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -43,6 +45,7 @@
 
 <dom-module id="tf-profile-dashboard">
 <template>
+<paper-toast id="toast" text="" always-on-top></paper-toast>
```
stephanwlee: DOM-wise, you might want at the end so it shows up on top of other sibling nodes.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -77,6 +80,33 @@ <h3>No profile data was found.</h3>
   <tf-dashboard-layout>
   <div class="sidebar">
     <div class="allcontrols">
+      <div class="sidebar-section">
+        <paper-button raised onclick="captureProfileDialog.open()">Capture Profile</paper-button>
```
stephanwlee: I might be misremembering but please refrain from using JS in an attribute like this. It is more prone for mistake that at the worst can cause XSS (and harder to statically analyze -- i don't want HTML AST parser just for this!).

First, please use `on-tap` (or `on-click`; [Polymer specific] tap is preferred because it works both for touch and mouse devices!) with a callback as string (https://polymer-library.polymer-project.org/1.0/docs/devguide/events)

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -327,6 +368,25 @@ <h3>No profile data was found.</h3>
...
+    shouldDisableCaptureProfileButton: function(profilerServiceAddress, profileDuration) {
+      return !profilerServiceAddress || profileDuration < 0;
+    },
+    captureProfile: function() {
+      const captureProfileUrl = tf_backend.addParams(
+        tf_backend.getRouter().pluginRoute('profile', '/capture_profile'),
+        {
+          service_addr: this.profilerServiceAddress,
+          is_tpu_name: this.profilerServiceAddressType == 'tpu-name',
+          uration: this.profileDuration,
```
stephanwlee: probably a typo: `uration` sounds wrong

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -265,6 +295,17 @@ <h3>No profile data was found.</h3>
         type: String,
         value: "/trace_viewer_index.html",
       },
+      profilerServiceAddress: {
```
stephanwlee: It looks like these are not a public API of this component. Please prefix these props with underscore the privateness.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -327,6 +368,25 @@ <h3>No profile data was found.</h3>
     detached: function() {
       this.set('_isAttached', false);
     },
+    shouldDisableCaptureProfileButton: function(profilerServiceAddress, profileDuration) {
+      return !profilerServiceAddress || profileDuration < 0;
+    },
+    captureProfile: function() {
+      const captureProfileUrl = tf_backend.addParams(
+        tf_backend.getRouter().pluginRoute('profile', '/capture_profile'),
```
stephanwlee: stylistic nit: +2 whitespace for https://google.github.io/styleguide/jsguide.html#formatting-indent

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -327,12 +329,32 @@ def data_route(self, request):
       return http_util.Respond(request, '404 Not Found', 'text/plain', code=404)
     return http_util.Respond(request, data, 'application/json')
 
+  @wrappers.Request.application
+  def capture_route(self, request):
+    service_addr = request.args.get('service_addr')
+    duration = int(request.args.get('duration'))
```
stephanwlee: if the arg `duration` is missing, this will raise TypeError. What is an expected outcome when certain inputs are missing?

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -25,6 +25,7 @@
 
 import tensorflow as tf
 
+from tensorflow.python.eager import profiler_client
```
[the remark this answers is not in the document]
qqfish: I don't want to make it a public api yet, since it will only be used by TensorBoard

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
qqfish: But it is updated by html similar to other public properties like selectedToolIndex, sellectedHostIndex, etc.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
qqfish: Done.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
qqfish: It should be 'duration'..

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
qqfish: Done

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
stephanwlee: Either those props are used by parent or they are mistakes too :)

tensorboard/plugins/profile/profile_plugin.py
qqfish: We don't allow user click "capture" if duration is missing.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
qqfish: But it can be modified by html similar to other public properties.

tensorboard/plugins/profile/profile_plugin.py
qqfish: I am not sure this part as well. Copy from capture_tpu_profile. +qiuminxu

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -327,12 +329,32 @@ def data_route(self, request):
...
     return http_util.Respond(request, data, 'application/json')
 
+  @wrappers.Request.application
+  def capture_route(self, request):
+    service_addr = request.args.get('service_addr')
+    duration = int(request.args.get('duration'))
+    is_tpu_name = request.args.get('is_tpu_name') == 'true'
+    if is_tpu_name:
+      tpu_cluster_resolver = (
+        tf.contrib.cluster_resolver.TPUClusterResolver([service_addr]))
```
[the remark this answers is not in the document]
qqfish: +@qiuminxu 

We want it to work on Cloud TPU as well. If user uses it in tf 2.0, it should just return an error.

tensorboard/plugins/profile/profile_plugin.py
stephanwlee: Should the server blow up in that case? Please handle the input properly.

tensorboard/plugins/profile/profile_plugin.py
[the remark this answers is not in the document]
stephanwlee: > If user uses it in tf 2.0, it should just return an error

TensorBoard will soon advertise TF2.0 compatness. If possible, let us not do that.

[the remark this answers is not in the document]
qqfish: @qiuminxu what is your preference? Submit this as it is and move it out from conrib? Or delete this and add tpu supports later?

tensorboard/plugins/profile/profile_plugin.py
qqfish: An exception will not blow up the server. It will just return an error response.

tensorboard/plugins/profile/profile_plugin.py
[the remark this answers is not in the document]
qiuminxu: I believe Tf 2.0 will have the tpu_cluster_resolver here https://github.com/tensorflow/tensorflow/tree/master/tensorflow/python/distribute/cluster_resolver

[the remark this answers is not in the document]
qqfish: Updated

tensorboard/plugins/profile/profile_plugin.py
qiuminxu: The port number is fixed by the tpu tfrc runtime. The 8470 is the port to run a TPU model, while the 8466 is the port to tpu profiler service. Match with regex is probably cleaner.

qqfish: Moved it under condition is_tpu_name and added comments.

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -327,12 +329,34 @@ def data_route(self, request):
...
+      service_addr = tpu_cluster_resolver.get_master()
+      # TPU cluster resolver always returns port 8470. Replace it with 8466
+      # on which profiler service is running.
+      service_addr = service_addr.replace('grpc://', '').replace(':8470', ':8466')
+    try:
+      profiler_client.start_tracing(service_addr, self.logdir, int(duration))
+      return http_util.Respond(
+          request, {'result': 'Capture profile successfully. Please refresh'},
+          'application/json')
+    except tf.errors.UnavailableError:
```
stephanwlee: may want to handle the case when duration is None and respond with appropriate status code as well as a message.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -57,6 +59,30 @@ <h3>Profiling isn’t supported in Colab yet.</h3>
...
+  <label id="profiler-address-type-label">Address Type:</label>
+  <paper-radio-group selected="ip-address"
+                     selected="{{_profilerServiceAddressType}}"
+                     aria-labelledby="profiler-address-type-label">
+    <paper-radio-button name="ip-address">IP Address</paper-radio-button>
+    <paper-radio-button name="tpu-name">TPU Name</paper-radio-button>
+  </paper-radio-group>
+  <paper-input label="Profiling Duration (milliseconds)"
+               always-float-label
+               type=number
```
stephanwlee: Nit, please put quotes around. `type="number"`

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -399,6 +439,30 @@ <h3>No profile data was found.</h3>
     detached: function() {
       this.set('_isAttached', false);
     },
+    openCaptureProfileDialog: function() {
+      this.$.captureProfileDialog.open();
+    },
+    shouldDisableCaptureProfileButton: function(profilerServiceAddress, profileDuration) {
```
stephanwlee: your changes in this file: we prefix private methods (since JS, not TS, does not have concept of private method, that's the naming convention we use). Would you be able to make that change? Thanks!

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -432,9 +434,35 @@ def data_route(self, request):
...
+    is_tpu_name = request.args.get('is_tpu_name') == 'true'
+    if is_tpu_name:
+      tpu_cluster_resolver = (
+          tf.distribute.cluster_resolver.TPUClusterResolver([service_addr]))
+      service_addr = tpu_cluster_resolver.get_master()
+      # TPU cluster resolver always returns port 8470. Replace it with 8466
+      # on which profiler service is running.
+      service_addr = service_addr.replace('grpc://', '').replace(':8470', ':8466')
+    try:
+      profiler_client.start_tracing(service_addr, self.logdir, duration)
```
stephanwlee: Oh interesting, didn't realize there is no way to stop and have to wait for that duration. Just curious, why?

stephanwlee: Separately, I am pondering the implication of using `self.logdir` here. Are profile data tagged with timestamp or anything? Would one be able to make any sense of the name when user profiles multiple times?

qqfish: 1) I think it is fine to block the thread since python server uses different multiple threads to handle request.
2) Profile data are tagged with timestamp.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -399,6 +439,30 @@ <h3>No profile data was found.</h3>
...
+    },
+    _captureProfile: function() {
+      const captureProfileUrl = tf_backend.addParams(
+          tf_backend.getRouter().pluginRoute('profile', '/capture_profile'),
+          {
+            service_addr: this._profilerServiceAddress,
+            is_tpu_name: this._profilerServiceAddressType == 'tpu-name',
+            duration: this._profileDuration,
+          });
+      this._requestManager.request(captureProfileUrl).then((json) => {
```
stephanwlee: Hmm I did not realize this soon enough. I generally am scared of holding request open for at least a duration of the profile. Any thoughts on changing this contract so FE starts, polls (presumably with some status about the profile like memory usage), and stop?

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -337,6 +368,15 @@ <h3>No profile data was found.</h3>
...
         value: "/trace_viewer_index.html",
       },
+      _profilerServiceAddress: String,
+      _profilerServiceAddressType: {
+        type: String,
+        value: "ip-address",
+      },
+      _profileDuration: {
+        type: Number,
+        value: 1000
```
stephanwlee: super nit: trailing comma for consistency :)

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
qqfish: I think it could be a future improvement. Created an issue #2013 to track it.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -57,6 +59,31 @@ <h3>Profiling isn’t supported in Colab yet.</h3>
...
   </div>
 </template>
+<paper-toast id="toast" text="" always-on-top></paper-toast>
+<paper-dialog id="captureProfileDialog" modal>
+  <paper-input label="Profiler Service URL or TPU name"
+               always-float-label
+               placeholder="localhost:6009"
+               value="{{_profilerServiceAddress}}"></paper-input>
+  <label id="profiler-address-type-label">Address Type:</label>
+  <paper-radio-group selected="ip-address"
```
stephanwlee: sorry again, it looks like you have duplicate property `selected` here. I believe this one can be omitted.

qqfish: Removed.

PR #1904, opened 2019-02-26
build: upgrade to Bazel 0.22, rules_closure HEAD

### LINKED ISSUE #1903: Bazel 0.23 breaks TensorBoard build
Bazel 0.23 was released roughly 4hrs ago. 
https://github.com/bazelbuild/bazel/releases 

```sh
$ bazel run tensorboard -- --logdir ~/data/foo

ERROR: /private/var/tmp/_bazel_stephanlee/e1e7b89094e7ae98436218ba85a1f221/external/io_bazel_rules_closure/closure/stylesheets/closure_css_library.bzl:43:17: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_stephanlee/e1e7b89094e7ae98436218ba85a1f221/external/io_bazel_rules_closure/closure/stylesheets/closure_css_library.bzl", line 39
		rule(implementation = _closure_css_li..., ...")})
	File "/private/var/tmp/_bazel_stephanlee/e1e7b89094e7ae98436218ba85a1f221/external/io_bazel_rules_closure/closure/stylesheets/closure_css_library.bzl", line 43, in rule
		attr.label_list(cfg = "data", allow_files = True)
Using cfg = "data" on an attribute is a noop and no longer supported. Please remove it. You can use --incompatible_disallow_data_transition=false to temporarily disable this check.
ERROR: error loading package '': in /private/var/tmp/_bazel_stephanlee/e1e7b89094e7ae98436218ba85a1f221/external/io_bazel_rules_closure/closure/defs.bzl: Extension file 'closure/stylesheets/closure_css_library.bzl' has errors
ERROR: error loading package '': in /private/var/tmp/_bazel_stephanlee/e1e7b89094e7ae98436218ba85a1f221/external/io_bazel_rules_closure/closure/defs.bzl: Extension file 'closure/stylesheets/closure_css_library.bzl' has errors
```

wchargin:
Summary:
Fixes #1903. Googlers are now locked to Bazel 0.23, which breaks
`rules_closure` unless we pull in HEAD (they’ve not yet had a release).
Conversely, `rules_closure` HEAD is broken on Bazel 0.21, so our CI
needs to be upgraded to match; Bazel 0.22 is the minimum working
version. Also, `rules_closure` had a breaking change in a package
called `private` whose default visibility is `public`, so our
`clutz.bzl` needs to be updated to work around that.

Test Plan:
Running `bazel build //tensorboard/... && bazel test //tensorboard/...`
passes on Bazel versions 0.22.0 and 0.23.0, failing on version 0.21.0.

Verified that the shasum matches the properly signed Bazel 0.22 binary:

```
$ gpg --verify ./bazel-0.22.0-linux-x86_64{.sig,}
gpg: Signature made Mon 28 Jan 2019 05:30:32 AM PST
gpg:                using RSA key 71A1D0EFCFEB6281FD0437C93D5919B448457EE0
gpg: Good signature from "Bazel Developer (Bazel APT repository key) <bazel-dev@googlegroups.com>" [full]
$ shasum -a 256 --check ./bazel-0.22.0-linux-x86_64.sha256
bazel-0.22.0-linux-x86_64: OK
$ shasum -a 256 ./bazel-0.22.0-linux-x86_64
8474ed28ed4998e2f5671ddf3a9a80ae9e484a5de3b8b70c8b654c017c65d363  ./bazel-0.22.0-linux-x86_64
```

wchargin-branch: bazel-0.23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ env:
   #
   # Grab the BAZEL_SHA256SUM from the Bazel releases page; e.g.:
   # bazel-0.20.0-linux-x86_64.sha256
-  - TF_VERSION_ID=tf-nightly BAZEL=0.21.0 BAZEL_SHA256SUM=7a2f347d779c4c772cdb318f0435f8d489a238be42e7a9bf1e2bad94b1c3094f
+  - TF_VERSION_ID=tf-nightly BAZEL=0.22.0 BAZEL_SHA256SUM=8474ed28ed4998e2f5671ddf3a9a80ae9e484a5de3b8b70c8b654c017c65d363
 
 cache:
   directories:
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -16,15 +16,17 @@ http_archive(
 load("@bazel_skylib//lib:versions.bzl", "versions")
 # Keep this version in sync with the BAZEL environment variable defined
 # in our .travis.yml config.
-versions.check(minimum_bazel_version = "0.16.1")
+versions.check(minimum_bazel_version = "0.22.0")
 
 http_archive(
     name = "io_bazel_rules_closure",
-    sha256 = "b29a8bc2cb10513c864cb1084d6f38613ef14a143797cea0af0f91cd385f5e8c",
-    strip_prefix = "rules_closure-0.8.0",
+    sha256 = "0e6de40666f2ebb2b30dc0339745a274d9999334a249b05a3b1f46462e489adf",
+    # The changes that we need for Bazel 0.23 compatibility are not in
+    # any release, so we pin to HEAD as of 2019-02-22.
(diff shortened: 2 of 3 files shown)


wchargin:
WANT_LGTM=any

wchargin:
> If this forces a new minimum version of 0.22, we need to update the
> versions.check in WORKSPACE

Yes; good point.

> and our nightly build machine setup as well, right?

Also true, though this can’t be done atomically.

> Maybe we can get rules_closure to release a version that is
> back-compatible to at least 0.21 and forward to 0.23? It's not great
> to only be able to build with a single version of Bazel.

I agree, but I think that it is more important to unbreak the build
first and expand the supported version range later.

nfelt:
Fair enough, though after being broken in this way by forced bazel updates I switched to a local bazel install which is currently on 0.19.2, so this PR will in fact unbreak your build but break mine until I update :)

PR #1908, opened 2019-02-27
Fix flaky eventfile usage in summary V2 tests

nfelt:
This PR avoids flaky behavior in the way that eventfiles get read in the test cases for V2 summary ops, which in turn unblocks https://github.com/tensorflow/tensorboard/pull/1891 which otherwise needs to hack around the flakiness.

The implemented change is to fully encapsulate the creation of an eventfile as an implementation detail of the test helpers that construct V2 summary API `Summary` protos, and ensure that at most 1 eventfile is on disk at any given time by deleting it after we read it in.

---

Motivation: the summary API tests all use helpers that attempt to encapsulate the process of converting the API call into a resulting `Summary` protobuf.  For the V1 ops, this is easy because they're designed to produce this output directly.  For the V2 ops, there is no intermediate way to collect the protos before they hit disk, so we write the single event of output to an eventfile and read it back in.

The issue is that some of the tests for convenience call the API several times with different parameters when testing related behavior, which writes multiple separate eventfiles to disk.  Previously this was a non-issue; we just read from the last file found, since the new file was always last in sort order - either by overwriting the previous file, or having a later UNIX timestamp.  Astute readers might note that this is not quite guaranteed because the sort uses the stringified timestamp, not its integer value, and this only works for numbers with the same digit count.  That being said, all unix timestamps from 2001 to 2286 will have 10 digits, so this is not a problem in practice.

However, the recent change in https://github.com/tensorflow/tensorflow/commit/826027dbd4277a2636fc2935ed245700fd01e7cd changes the eventfile naming strategy to avoid precisely these collisions (which are usually undesirable) by baking the PID followed by a per-TF-process UID into the filename after the timestamp.  The UID is basically an autoincrement int so UIDs in the 3-4 digit range are typical, and unfortunately for us, it's much easier to encounter the numeric != lexicographic sort issue with these small numbers.  This means that seemingly unrelated behavior (e.g. inserting new test methods) caused the filenames written by these tests to no longer sort in the order of creation, and thus the wrong data got read and the tests would fail.  (Deterministically, at least in my testing, but I don't think that's actually guaranteed since the UID can be incremented across threads.)

### DIFF
--- a/tensorboard/plugins/histogram/summary_test.py
+++ b/tensorboard/plugins/histogram/summary_test.py
@@ -164,33 +164,41 @@ class SummaryV2OpTest(SummaryBaseTest, tf.test.TestCase):
       self.skipTest('v2 summary API not available')
 
   def histogram(self, *args, **kwargs):
+    return self.histogram_event(*args, **kwargs).summary
+
+  def histogram_event(self, *args, **kwargs):
     kwargs.setdefault('step', 1)
     writer = tf2.summary.create_file_writer(self.get_temp_dir())
     with writer.as_default():
       summary.histogram(*args, **kwargs)
     writer.close()
-    return self.read_single_event_from_eventfile().summary
-
-  def read_single_event_from_eventfile(self):
     event_files = sorted(glob.glob(os.path.join(self.get_temp_dir(), '*')))
     self.assertEqual(len(event_files), 1)
     events = list(tf.compat.v1.train.summary_iterator(event_files[0]))
     # Expect a boilerplate event for the file_version, then the summary one.
     self.assertEqual(len(events), 2)
+    # Delete the event file to reset to an empty directory for later calls.
+    os.remove(event_files[0])
     return events[1]
+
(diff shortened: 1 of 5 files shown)


tensorboard/plugins/histogram/summary_test.py
```diff
@@ -173,7 +173,8 @@ def histogram(self, *args, **kwargs):
 
   def read_single_event_from_eventfile(self):
     event_files = sorted(glob.glob(os.path.join(self.get_temp_dir(), '*')))
-    events = list(tf.compat.v1.train.summary_iterator(event_files[-1]))
+    self.assertEqual(len(event_files), 1)
+    events = list(tf.compat.v1.train.summary_iterator(event_files[0]))
```
wchargin: Do we not want to delete the event file here?

nfelt: Done.  I missed updating this file, thanks for catching this.

nfelt:
Okay, gotcha.  Yes, sounds good!

Issue #1705, opened 2018-12-19
Fix broken tests when running with tf 2.0


stephanwlee:
There are tests that break when running under TF 2.0 for various reasons:

- [x] tensorboard/plugins/debugger/debug_graphs_helper_test.py - breaks because tf.add creates nested nodes in TF 2.0 (#2583)
- [x] tensorboard/util/encoder_test.py - Audio summary using deprecated tf.contrib.ffmpeg.encode_audio. Must find replacement. (#2557)
- [x] tensorboard/data_compat_test.py - Audio summary using deprecated tf.contrib.ffmpeg.encode_audio. Must find replacement. (#2557)
- [x] tensorboard/backend/event_processing/plugin_event_accumulator_test.py - Audio summary using deprecated tf.contrib.ffmpeg.encode_audio. Must find replacement. (#2558)
- [x] tensorboard/plugins/core/core_plugin_test.py - tf.contrib.summary.* usage
- [x] tensorboard/plugins/scalar/scalars_plugin_test.py:
    - [x] tf.placeholder usage (#2522)
    - [x] tf.contrib.summary.* usage for multiplexer tests (#2522)
    - [x] tf.contrib.summary.* usage for DB writer tests (#3539)
- [x] tensorboard/plugins/debugger/interactive_debugger_plugin_test.py (#2583)
- [x] tensorboard/plugins/debugger/session_debug_test.py server fails to come up in TFv2. (#2583)
- [x] tensorboard/plugins/beholder/beholder_test.py
- [x] tensorboard/plugins/mesh/mesh_plugin_test.py (#2560)
- [x] tensorboard/plugins/mesh/summary_test.py (#2560)
- [x] tensorboard/data_compat_test.py (#2559)

(To generate this list, change `run_v1_only` to a no-op, then run all tests in TF 2.x.)

wchargin:
I have some fixes pending. After these pull requests are merged, the
stragglers will fall into two categories:

  - debugger-related tests:
      - `tensorboard/plugins/debugger/debug_graphs_helper_test.py`
      - `tensorboard/plugins/debugger/interactive_debugger_plugin_test.py`
      - `tensorboard/plugins/debugger/session_debug_test.py`
  - users of `tf.contrib.summary.create_db_writer`:
      - `tensorboard/plugins/core/core_plugin_test.py`
      - `tensorboard/plugins/scalar/scalars_plugin_test.py`
  - beholder:
[...]

PR #1911, opened 2019-02-27
Test against TF2 preview in CI

stephanwlee:
We would like to not regress for TensorFlow v2.

Add an environment variable entry to trigger the build/test against the tf-nightly-2.0-preview.
In case TF2.0 becomes mainstream, until we decide to drop implicit support for TF1, we should test against TF1.

* Screenshots of UI changes
N/a

This PR :)

?

### DIFF
--- a/tensorboard/data_compat_test.py
+++ b/tensorboard/data_compat_test.py
@@ -30,8 +30,8 @@ from tensorboard.plugins.image import summary as image_summary
 from tensorboard.plugins.scalar import metadata as scalar_metadata
 from tensorboard.plugins.scalar import summary as scalar_summary
 from tensorboard.util import tensor_util
+from tensorboard.util import test_util
 
-tf.compat.v1.disable_v2_behavior()
 
 
 class MigrateValueTest(tf.test.TestCase):
@@ -61,22 +61,24 @@ class MigrateValueTest(tf.test.TestCase):
     self.assertEqual(original_pbtxt, value.SerializeToString())
 
   def test_scalar(self):
-    old_op = tf.compat.v1.summary.scalar('important_constants', tf.constant(0x5f3759df))
-    old_value = self._value_from_op(old_op)
-    assert old_value.HasField('simple_value'), old_value
-    new_value = data_compat.migrate_value(old_value)
+    with tf.compat.v1.Session():
+      old_op = tf.compat.v1.summary.scalar('important_constants', tf.constant(0x5f3759df))
+      old_value = self._value_from_op(old_op)
+      assert old_value.HasField('simple_value'), old_value
+      new_value = data_compat.migrate_value(old_value)
 
-    self.assertEqual('important_constants', new_value.tag)
-    expected_metadata = scalar_metadata.create_summary_metadata(
-        display_name='important_constants',
-        description='')
-    self.assertEqual(expected_metadata, new_value.metadata)
-    self.assertTrue(new_value.HasField('tensor'))
-    data = tensor_util.make_ndarray(new_value.tensor)
-    self.assertEqual((), data.shape)
-    low_precision_value = np.array(0x5f3759df).astype('float32').item()
-    self.assertEqual(low_precision_value, data.item())
+      self.assertEqual('important_constants', new_value.tag)
+      expected_metadata = scalar_metadata.create_summary_metadata(
+          display_name='important_constants',
+          description='')
+      self.assertEqual(expected_metadata, new_value.metadata)
+      self.assertTrue(new_value.HasField('tensor'))
+      data = tensor_util.make_ndarray(new_value.tensor)
+      self.assertEqual((), data.shape)
+      low_precision_value = np.array(0x5f3759df).astype('float32').item()
+      self.assertEqual(low_precision_value, data.item())
 
+  @test_util.run_v1_only('v1 audio summary uses contrib')
   def test_audio(self):
     audio = tf.reshape(tf.linspace(0.0, 100.0, 4 * 10 * 2), (4, 10, 2))
     old_op = tf.compat.v1.summary.audio('k488', audio, 44100)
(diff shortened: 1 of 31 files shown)


stephanwlee:
Seems like the build is passing! Please hide whitespace changes to less confuse you :)

tensorboard/util/test_util.py
```diff
@@ -273,3 +278,93 @@ def ensure_tb_summary_proto(summary):
...
+      decorator the can be applied to a function. If `func` is not None this
+      returns the decorator applied to `func`.
+
+  Returns:
+    Returns a decorator that will conditionally skip the decorated test method.
+  """
+
+  def decorator(f):
+    if inspect.isclass(f):
+      setup = f.__dict__.get("setUp")
```
wchargin: I’m not sure exactly what to make of this `setUp` patching. On the one
hand, `setUp` only runs if an actual test case runs, and we patch all
the test cases, so this seems a bit superfluous. On the other hand, due
to test case inheritance, (a) you could have more test cases on a
subclass, and (b) `__dict__` is not sufficient to patch `setUp` because
the `setUp` in question could be defined on a superclass (possibly using
subclass-dependent method implementations). I _think_ we’re safe here
because logically there should be an independence principle (if your
test class extends a test class that’s V1-only, then you must also be
V1-only). Is this consistent with your reasoning? If not, can you
explain why you believe that this mechanism for `setUp` patching is
correct?

tensorboard/util/test_util.py
```diff
@@ -273,3 +278,93 @@ def ensure_tb_summary_proto(summary):
...
+        value = getattr(f, name)
+        if (callable(value) and
+            name.startswith(unittest.TestLoader.testMethodPrefix)):
+          setattr(f, name, decorator(value))
+
+      return f
+
+    def decorated(self, *args, **kwargs):
+      if not tf2.enabled():
+        self.skipTest("Skipping test for TF v2.")
```
wchargin: No use of `reason`?

tensorboard/plugins/debugger/debug_graphs_helper_test.py
```diff
@@ -133,6 +134,22 @@ def testExtractGatedGrpcTensorsFoundNoGatedGrpcOps(self):
       self.assertEqual([], gated_debug_ops)
 
 
+# [1]: Below graph creates different ops
+#   a = tf.Variable([1.0], name='foo/a')
```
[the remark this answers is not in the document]
stephanwlee: I don't follow. This is from L157. Are you saying we don't need foo/ and bar/ to exemplify the problem thus should be removed?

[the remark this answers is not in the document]
wchargin: Yeah, that’s what I was asking: “Can the problem be reproduced without
the `foo/` and `bar/` name scopes? If so, wouldn’t it be simpler to
remove them?”

tensorboard/plugins/graph/keras_util_test.py
```diff
@@ -24,14 +24,22 @@
...
-class GraphUtilTest(tf.test.TestCase):
+class BaseGraphUtilTest(object):
   def assertGraphDefToModel(self, expected_proto, model):
     model_config = json.loads(model.to_json())
 
     self.assertProtoEquals(
         expected_proto, keras_util.keras_model_to_graph_def(model_config))
+  
+  def lstm_node_name(self):
+    raise NotImplementedError
```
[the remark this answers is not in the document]
stephanwlee: Hmm wonder why flake8 didn't catch this.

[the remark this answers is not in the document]
wchargin: At a glance, it doesn’t appear to have a rule for this at all. Even if
it did, we only run `flake8` with `--select=E901,E999,F821,F822,F823`.

tensorboard/util/test_util.py
stephanwlee: This was borrowed code in verbatim from TF so I did not have any rationale in doing so.
>setUp only runs if an actual test case runs

If this is true, I do agree that this patching is quite superfluous  and I wonder if TF ever had a case when a class is v1 only and one method is v2 only. This seems very unpredictable (how does test author know if setUp gets run? and also, why is setUp skipped but not tearDown?) and should be warned against (since v2 only will be ineffective AFAICT).

Anyways, I will remove this.

stephanwlee:
Reopening the PR since I have time now and think this is important to prevent any regression.

tensorboard/plugins/debugger/debug_graphs_helper_test.py
```diff
@@ -68,6 +68,7 @@ def _createTestGraphAndRunOptions(self, sess, gated_grpc=True):
                          debug_urls=self.debug_server_url)
     return z, run_options
 
+  @test_util.run_v1_only('Ops differ. Similar to [1].')
```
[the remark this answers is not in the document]
stephanwlee: Updated to contain name of the file and moved the comment block to module doc so it is more prominent and visible.

tensorboard/plugins/debugger/session_debug_test.py
```diff
@@ -40,8 +40,10 @@
 
 from tensorboard.plugins.debugger import constants
 from tensorboard.plugins.debugger import debugger_server_lib
+from tensorboard.util import test_util
 
 
+@test_util.run_v1_only("Server fails to come up. Requires study.")
```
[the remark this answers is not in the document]
stephanwlee: Updated the tracking bug but I decided not to have circular reference; GitHub issue just having list of file names and code is all littered with `TODO(#ABC)`. Instead, I decided to just write why certain block of test warrants v1-ness.

tensorboard/plugins/graph/keras_util_test.py
```diff
@@ -285,11 +293,11 @@ def test_keras_model_to_graph_def_lstm_model(self):
       attr {
         key: "keras_class"
         value {
-          s: "LSTM"
+          s: "%s"
```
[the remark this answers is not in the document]
stephanwlee: Letting you know: instead of using LSTM which for some reason had changed its name, I decided to use `SimpleRNN` (or I can use any other layer technically; most AFAICT didn't change in name unlike the LSTM).

While moving this to proto, I found myself confused a little. 

Is proto construction more readable and is it easier to reason whether our utility is doing the right thing or do you think pbtxt is much more readable at glance (when it is small in size, of course)? If string interpolation makes pbtxt less appealing, it should be "addressed" by using SimpleRNN. I realized I hate the `lstm_class_name` I previously took and it is unavoidable even with proto construction.

tensorboard/plugins/graph/keras_util_test.py
```diff
@@ -525,7 +533,7 @@ def test_keras_model_to_graph_def_functional_multi_inputs(self):
         }
       }
     }
-    """
+    """ % (self.lstm_node_name(), self.lstm_keras_class_name(), self.lstm_node_name(), self.lstm_node_name())
```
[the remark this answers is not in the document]
stephanwlee: TIL! Thanks for letting me know. I have seen bunch of example using `%s` or `{}` but not seen the `%-dict`. 

I removed the string interpolation by using different layer that did not change in name between v1 and v2. LSTM seems to be the odd-ball.

tensorboard/util/encoder_test.py
```diff
@@ -55,6 +56,7 @@ def test_encodes_png_with_alpha(self):
     self._check_png(data)
 
 
+@test_util.run_v1_only('Uses contrib')
```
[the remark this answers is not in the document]
stephanwlee: This encoder_test.py is captured in #1705.

tensorboard/util/test_util.py
wchargin: > This seems very unpredictable (how does test author know if setUp gets
> run? and also, why is setUp skipped but not tearDown?)

I must be missing something. It seems very predictable to me. The logic is:

 1. `setUpClass` runs
 2. for each test case:
     1. `setUp` runs (skip next two steps if this fails)
     2. the test runs
     3. `tearDown` runs
 3. `tearDownClass` runs

It follows naturally that `setUp` only runs if a test case runs, just
from the semantics of what `setUp` means (“runs before each test”).

tensorboard/data_compat_test.py
```diff
@@ -61,22 +61,24 @@ def _assert_noop(self, value):
...
-        display_name='important_constants',
-        description='')
-    self.assertEqual(expected_metadata, new_value.metadata)
-    self.assertTrue(new_value.HasField('tensor'))
-    data = tensor_util.make_ndarray(new_value.tensor)
-    self.assertEqual((), data.shape)
-    low_precision_value = np.array(0x5f3759df).astype('float32').item()
-    self.assertEqual(low_precision_value, data.item())
-
+    with tf.compat.v1.Session():
```
nfelt: I'm a bit confused about using a session context manager here but not actually including `as sess` - does anything actually use this session in the code below?

It seems like perhaps this is being used just as a way to enable a default graph, but then I think `with tf.compat.v1.Graph.as_default()` would be clearer?

Also, it would be great to scope these contexts to just where they're really needed.  In this case I suspect it's just the first couple lines; the rest of these aren't executing TF code.

stephanwlee: It seems like we need then fix the problem elsewhere too.
Since I am more interested in enabling v2 tests right now, I will let future commits fix things like `tensorboard.compat.tf2` (or is that version correct as is? Even considering the env variable?)

tensorboard/util/test_util.py
```diff
@@ -25,11 +25,17 @@
...
 import os
+import six
 import sqlite3
 import threading
+import unittest
+from unittest import loader as _unittest_loader
+
 
 import tensorflow as tf
+from tensorflow.python import tf2
```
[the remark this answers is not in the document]
stephanwlee: It seems like we need then fix the problem elsewhere too.
Since I am more interested in enabling v2 tests right now, I will let future commits fix things like `tensorboard.compat.tf2` (or is that version correct as is? Even considering the env variable?)

tensorboard/data_compat_test.py
stephanwlee: Done.

tensorboard/util/test_util.py
```diff
@@ -280,91 +277,37 @@ def ensure_tb_summary_proto(summary):
...
+  def _impl(reason=None):
+    if reason is None:
+      if default_reason is None:
+        raise ValueError('%s requires a reason for skipping.' % name)
+      reason = default_reason
+    return unittest.skipUnless(guard(), reason)
+
+  return _impl
+
+_is_tf2 =  tf.__version__.startswith('2.')
```
nfelt: Can you add a TODO here for this to work with internal TF?  Can assign to me if you want.

tensorboard/util/test_util.py
```diff
@@ -273,3 +275,40 @@ def ensure_tb_summary_proto(summary):
     return summary
 
   return summary_pb2.Summary.FromString(summary.SerializeToString())
+
+
+def _run_conditionally(guard, name, default_reason=None):
+  """Creates a decorator that skips a test when guard returns False.
```
wchargin: nit: *Create*, not *Creates* (!pep 257)

wchargin: This function doesn’t actually create a decorator. It creates a function
that returns decorators. (This is a decorator factory factory.) The
distinction is important; “create a decorator” suggests that it would be
valid to write

```python
@run_v1_only  # not invoked!
def foo():
  pass
```

which is not the case.

tensorboard/util/test_util.py
```diff
@@ -273,3 +275,40 @@ def ensure_tb_summary_proto(summary):
...
+
+    Args:
+      guard: A lambda that returns True if a test should be executed.
+      name: A human readable name for the decorator for an error message.
+      default_reason: A string describing why a test should be skipped. If it
+          is None, the decorator will make sure the reason is supplied by the
+          consumer of the decorator. Default is None.
+
+    Raises:
+      ValueError when both reason and default_reason are None.
```
wchargin: No, this function never raises `ValueError`; it returns a _function_
that may raise `ValueError`. You can tell that this must be the case
because this function has no argument `reason`.

tensorboard/plugins/debugger/session_debug_test.py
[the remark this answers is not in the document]
wchargin: Makes sense; thanks.

tensorboard/util/test_util.py
stephanwlee: Hmm what is the right way to capture that in pydoc? It sounds like there should be a separate pydoc for the `_impl`?

wchargin: Good question. I don’t have a definitive answer. It would be fine me to
just clearly document the semantics of the returned function in the
“Returns” clause of the top-level function’s docs (like: “Returns: A
function that takes these arguments, raises under these conditions, and
returns a decorator.”). You might also consider adding docs to `_impl`
itself, which will appear if a user runs `help(run_v1_only)` or
`run_v1_only.__doc__` in an interactive session.

Issue #1913, opened 2019-02-27
profile/dynamic plugins should work in Colab


wchargin:
The profile dashboard currently does not work in Colab, because the
trace viewer is embedded via an iframe, which cannot traverse Colab’s
service worker tunnel for security reasons (service workers never proxy
iframe content).

Ideally, the profile dashboard should work in Colab, which we could by
including the trace viewer directly rather than using an iframe.\* At
minimum, we should display a helpful error instead of just failing to
load the content.

\* @stephanwlee says that this is not trivial, because even though the
trace viewer uses web components as well, it’s built via a different
mechanism that’s specific to Chrome, and so we’d need some translation
layer for the dependency structure (possibly manual).

Edit: 2019-06-27
Dynamic plugins also employ iframe to render the plugin content which
is not compatible with Colab. We have plans to use new Colab's API,
`proxyPort`, to mitigate the issue but it is not ready for consumption yet.

wchargin:
Now that #1914 has merged, I’m demoting this to a feature request.

stephanwlee:
Retroactively cleaning issues. 

#2798 should fix the issue and the fix should be available in tb-nightly (not 2.0.1).

PR #1914, opened 2019-02-28
profile: disable dashboard inside Colab

wchargin:
Summary:
The profile dashboard now shows a “disabled” message on the frontend
when we detect that we’re running in a Colab notebook. See #1913.

This commit does _not_ cause the profile dashboard to be marked inactive
by the main TensorBoard UI. We might want to do that, but it’s less
straightforward: currently, the backend is responsible for telling the
frontend which plugins are active, and the backend can’t know whether
we’re running in Colab. As a quick hack, we could add special-case logic
in `tf-tensorboard.html` to check for the profile plugin specifically.
As a longer-term solution, we could let the `tf_tensorboard.Dashboard`
interface specify an `isActive` callback, defaulting to `() => true`.

![Screenshot of the “Profiling isn’t yet supported in Colab” message][s]

[s]: https://user-images.githubusercontent.com/4317806/53541886-b83b1880-3ad0-11e9-9501-299f5e671d7a.png

The implementation includes a small refactor to the profile dashboard to
make it easier to introduce a new state.

Test Plan:
Evaluate `!!(window.TENSORBOARD_ENV || {}).IN_COLAB` in both standalone
TensorBoard and Colab `%tensorboard`, and note that each evaluates to
the appropriate boolean.

Run TensorBoard locally with the profile plugin’s demo data, and note
that its behavior is unchanged. To trigger the loading state, it
suffices to load the dashboard, then throttle your network to “Slow 3G”,
then change the selected tool. To trigger the “no data” state, relaunch
TensorBoard pointing at a different log directory.

Then, test the Colab behavior by either pointing TensorBoard at a log
directory that does not contain profile data, then executing

```js
window.TENSORBOARD_ENV = window.TENSORBOARD_ENV || {};
window.TENSORBOARD_ENV["IN_COLAB"] = true;
```

in the main TensorBoard pane before switching to the inactive profile
dashboard. Note that this is the same setup JavaScript as is injected
into the Colab output frame by `notebook.py`.

wchargin-branch: profile-disable-colab

### DIFF
--- a/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
+++ b/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
@@ -43,13 +43,23 @@ profile run can have multiple tools that present the performance profile as diff
 
 <dom-module id="tf-profile-dashboard">
 <template>
-<template is="dom-if" if="[[_isNotComplete(progress)]]">
+<template is="dom-if" if="[[_isState(_topLevelState, 'IN_COLAB')]]">
+  <div style="max-width: 540px; margin: 80px auto 0 auto;">
+    <h3>Profiling isn’t yet supported in Colab.</h3>
+    <p>
+      Please see
+      <a href="https://github.com/tensorflow/tensorboard/issues/1913">GitHub issue #1913</a>
+      for more information.
+    </p>
+  </div>
+</template>
+<template is="dom-if" if="[[_isState(_topLevelState, 'LOADING')]]">
   <div id="progress-bar">
     <div id="progress-msg">[[progress.msg]]</div>
     <paper-progress value="[[progress.value]]"></paper-progress>
   </div>
 </template>
-<template is="dom-if" if="[[_dataNotFound]]">
+<template is="dom-if" if="[[_isState(_topLevelState, 'DATA_NOT_FOUND')]]">
   <div style="max-width: 540px; margin: 80px auto 0 auto;">
     <h3>No profile data was found.</h3>
     <p>To collect a profile, you need to run your model on Google Cloud TPUs and
@@ -73,7 +83,7 @@ profile run can have multiple tools that present the performance profile as diff
     and consider filing an issue on GitHub.</p>
   </div>
 </template>
-<template is="dom-if" if="[[!_dataNotFound]]">
+<template is="dom-if" if="[[_isState(_topLevelState, 'ACTIVE')]]">
   <tf-dashboard-layout>
   <div class="sidebar">
     <div class="allcontrols">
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -534,6 +584,24 @@ <h3>No profile data was found.</h3>
       // Otherwise, remove the seperator, e.g. "host1_" => "host1".
       return host.slice(0, -1);
     },
+    _computeTopLevelState: function(inColab, dataNotFound, progress) {
+      if (inColab)
+        return "IN_COLAB";
+      if (dataNotFound)
+        return "DATA_NOT_FOUND";
+      if (!progress || progress.value < 100)
```
nfelt: I think we can now remove `_isNotComplete()` since its logic was inlined here and it's not called anywhere else

wchargin: Good point; done.

GalOshri:
nit on the error message: "Profiling isn't yet supported in Colab" sounds less natural to me than "Profiling isn't supported in Colab yet".

Relevant question: I think the profiling (creating the right logs) is supported, right? It's just displaying them in the dashboard that isn't.

Additionally, does this affect only Colab or also Jupyter?

wchargin:
> nit on the error message: "Profiling isn't yet supported in Colab"
> sounds less natural to me than "Profiling isn't supported in Colab
> yet".

Sure; can change.

> Relevant question: I think the profiling (creating the right logs) is
> supported, right? It's just displaying them in the dashboard that
> isn't.

Correct; we generate the logs (though the autoprofiler, triggered from
TensorBoard, of course can’t run if TensorBoard can’t show the
profiler). I think that the wording is _fine_, because the profiling
_experience_ is certainly broken given that you can’t consume what you
produce. But I’d accept a PR to change it.

> Additionally, does this affect only Colab or also Jupyter?

Colab only. Jupyter puts all of TensorBoard in an iframe with direct
network access; another iframe within that iframe is fine because there
is no service worker proxy to contend with. (This also means that
hparams should work fine with POSTs.)

wchargin:
Proof of life for Jupyter:

Screenshot of a happy trace viewer in Jupyter.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -235,6 +274,17 @@ <h3>No profile data was found.</h3>
...
+      _inColab: {
+        type: Boolean,
+        value: () => !!(window.TENSORBOARD_ENV || {}).IN_COLAB,
+        readOnly: true,
+      },
+      /** @type {TopLevelState} */
+      _topLevelState: {
+        type: String,
+        computed: '_computeTopLevelState(_inColab, _dataNotFound, progress)',
+        readOnly: true,
```
stephanwlee: computed is readOnly. https://polymer-library.polymer-project.org/1.0/docs/devguide/properties

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -235,6 +274,17 @@ <h3>No profile data was found.</h3>
         type: Array,
         observer: '_activeHostsChanged'
       },
+      _inColab: {
+        type: Boolean,
+        value: () => !!(window.TENSORBOARD_ENV || {}).IN_COLAB,
```
stephanwlee: Instead of `!!`, use `Boolean((window.TENSORBOARD_ENV || {}).IN_COLAB)` though I understand both with ease.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
wchargin: Good to know; thanks.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
wchargin: http://go/tsstyle explicitly says that `!!` is okay; jsstyle has no mention.

PR #1916, opened 2019-02-28
notebook: change default height to 800px

wchargin:
Summary:
The previous default height of 600px is okay for some dashboards but
particularly cramped for hparams. Through rigorous testing, we have
determined that 800px is a larger number.

Before and after screenshots of the hparams dashboard:
Screenshots

Test Plan:
Build a Pip package, upload it to Colab, run the hparams demo, take
screenshots above.

wchargin-branch: notebook-enlarge-default-height

### DIFF
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -218,7 +218,7 @@ def display(port=None, height=None):
       launched TensorBoard.
     height: The height of the frame into which to render the TensorBoard
       UI, as an `int` number of pixels, or `None` to use a default value
-      (currently 600).
+      (currently 800).
   """
   _display(port=port, height=height, print_message=True, display_handle=None)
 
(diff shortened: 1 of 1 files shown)


tensorboard/notebook.py
```diff
@@ -235,7 +235,7 @@ def _display(port=None, height=None, print_message=False, display_handle=None):
       render TensorBoard.
   """
   if height is None:
-    height = 600
+    height = 800
```
nfelt: Please add a TODO to load the page at `height = 10000`, screenshot it, feed that through a convnet running on tf.js, output a predicted optimal height value, and set the height to that :wink:

wchargin: You’re on to something, but I’m skeptical that that proposal has the
right architecture. I think I’ll compile `javax.swing.*` to JavaScript
then use reinforcement learning to learn to learn\* to faithfully embed
DOM models into Swing `LayoutManager`s. Then it’s trivial: we’ve reduced
the problem to simply calling `frame.pack()`, discarding the result, and
proclaiming that 800px should be enough for anyone.

\* (you _have_ to learn to learn; zeroth-order learning is so passé
these days)

PR #1924, opened 2019-02-28
Rewrite the tf.summary import logic, this time with tests

nfelt:
This rewrite fixes the issue where `import tensorflow.compat.v2.summary as tf_summary` produced a module that included only the tensorboard-defined symbols and none of the TF-defined ones.  See the comments in the rewritten code for the detailed explanation of the new approach.

This PR also adds tests for importing `tf.summary` to our pip package smoke test (which runs on Travis as of #1883), so that there's at least some test coverage for this cross-package import quagmire.

image

### DIFF
--- a/tensorboard/pip_package/build_pip_package.sh
+++ b/tensorboard/pip_package/build_pip_package.sh
@@ -164,12 +164,12 @@ tb.notebook.start  # don't invoke; just check existence
     import_attr="import tensorflow as tf; a = tf${1}.summary; a.write; a.scalar"
     import_as="import tensorflow${1}.summary as b; b.write; b.scalar"
     import_from="from tensorflow${1} import summary as c; c.write; c.scalar"
-    printf "${import_attr}\n${import_as}\n${import_from}" | python -
-    printf "${import_attr}\n${import_from}\n${import_as}" | python -
-    printf "${import_as}\n${import_attr}\n${import_from}" | python -
-    printf "${import_as}\n${import_from}\n${import_attr}" | python -
-    printf "${import_from}\n${import_attr}\n${import_as}" | python -
-    printf "${import_from}\n${import_as}\n${import_attr}" | python -
+    printf '%s\n' "${import_attr}" "${import_as}" "${import_from}" | python -
+    printf '%s\n' "${import_attr}" "${import_from}" "${import_as}" | python -
+    printf '%s\n' "${import_as}" "${import_attr}" "${import_from}" | python -
+    printf '%s\n' "${import_as}" "${import_from}" "${import_attr}" | python -
+    printf '%s\n' "${import_from}" "${import_attr}" "${import_as}" | python -
+    printf '%s\n' "${import_from}" "${import_as}" "${import_attr}" | python -
   }
   test_tf_summary '.compat.v2'
   is_tf_2() {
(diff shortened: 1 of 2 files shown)


nfelt:
I've confirmed that this works internally as well by replicating the pip package test logic.

tensorboard/summary/_tf/summary/__init__.py
```diff
@@ -18,53 +18,94 @@
...
+  ]
+  # If we aren't sure we're on V2, don't use tf.summary since it could be V1.
+  # Note there may be false positives since the __version__ attribute may not be
+  # defined at this point in the import process.
+  if not getattr(tf, '__version__', '').startswith('2.'):  # noqa: F821
+    packages.remove('tensorflow')
+
+  def dynamic_wildcard_import(module):
+    """Implements the logic of "from module import *" for the given module."""
+    symbols = getattr(module, '__all__', None)
```
wchargin: Discussed a bit offline.

It was not obvious to me that `module.__dict__` is sufficient. For
instance, if `type(module)` is not `types.ModuleType` but rather a
subtype of that type, then values defined on `type(module)` will not
appear in `module.__dict__`. We _do_ use custom module types for
TensorBoard `LazyModule`s, though they don’t export additional symbols.

But this does in fact appear to be consistent with the native behavior
of `from … import *`, per `Python/ceval.c:import_all_from` in the
CPython source:
<https://github.com/wchargin/cpython/blob/1f512b9a344b5053c013cf0a116b3562d426ef2b/Python/ceval.c#L5099-L5100>

So, while a little strange, this appears to be correct.

tensorboard/pip_package/build_pip_package.sh
```diff
@@ -145,15 +147,39 @@ smoke() {
...
+    printf "${import_attr}\n${import_as}\n${import_from}" | python -
+    printf "${import_attr}\n${import_from}\n${import_as}" | python -
+    printf "${import_as}\n${import_attr}\n${import_from}" | python -
+    printf "${import_as}\n${import_from}\n${import_attr}" | python -
+    printf "${import_from}\n${import_attr}\n${import_as}" | python -
+    printf "${import_from}\n${import_as}\n${import_attr}" | python -
+  }
+  test_tf_summary '.compat.v2'
+  is_tf_2() {
+    python -c "import tensorflow as tf; assert tf.__version__[:2] == '2.'" \
```
wchargin: Heh, clever `assert`.

tensorboard/pip_package/build_pip_package.sh
```diff
@@ -145,15 +147,39 @@ smoke() {
...
-import tensorboard.summary._tf.summary as tf_summary
 "
+
+  # Exhaustively test various sequences of importing tf.summary.
+  test_tf_summary() {
+    # First argument is subpath to test, e.g. '' or '.compat.v2'.
+    import_attr="import tensorflow as tf; a = tf${1}.summary; a.write; a.scalar"
+    import_as="import tensorflow${1}.summary as b; b.write; b.scalar"
+    import_from="from tensorflow${1} import summary as c; c.write; c.scalar"
+    printf "${import_attr}\n${import_as}\n${import_from}" | python -
```
wchargin: To avoid incorrect behavior in case one of these has a percent sign, you
can use

```sh
    printf '%s\n' "${import_attr}" "${import_as}" "${import_from}" | python -
    printf '%s\n' "${import_attr}" "${import_from}" "${import_as}" | python -
    printf '%s\n' "${import_as}" "${import_attr}" "${import_from}" | python -
    printf '%s\n' "${import_as}" "${import_from}" "${import_attr}" | python -
    printf '%s\n' "${import_from}" "${import_attr}" "${import_as}" | python -
    printf '%s\n' "${import_from}" "${import_as}" "${import_attr}" | python -
```

PR #1930, opened 2019-03-01
Graph plugin: show graph fetch error properly

stephanwlee:
Few issues:
1. fetchPbTxt or XMLHTTPRequest invokes `onload` even in case of error.
2. Promise handling was weird
3. Graph plugin try/catched and returned None instead of letting the
   route handle the error by responding with 400.

Test changes to catch these issues will come in follow up changes.

* Screenshots of UI changes
image

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_common/parser.ts
+++ b/tensorboard/plugins/graph/tf_graph_common/parser.ts
@@ -43,8 +43,9 @@ export function fetchPbTxt(filepath: string): Promise<ArrayBuffer> {
       // Fetch does not reject for 400+.
       if (res.ok) {
         res.arrayBuffer().then(resolve, reject);
+      } else {
+        res.text().then(reject, reject);
       }
-      res.text().then(reject, reject);
     });
   });
 }
@@ -79,7 +80,7 @@ export function fetchAndParseMetadata(path: string, tracker: ProgressTracker) {
 export function fetchAndParseGraphData(path: string, pbTxtFile: Blob,
     tracker: ProgressTracker) {
   return tf.graph.util
-      .runTask(
+      .runAsyncPromiseTask(
           'Reading graph pbtxt', 40,
           () => {
             if (pbTxtFile) {
@@ -95,7 +96,7 @@ export function fetchAndParseGraphData(path: string, pbTxtFile: Blob,
           },
           tracker)
       .then((arrayBuffer: ArrayBuffer) => {
-        return tf.graph.util.runTask('Parsing graph.pbtxt', 60, () => {
+        return tf.graph.util.runAsyncPromiseTask('Parsing graph.pbtxt', 60, () => {
           return parseGraphPbTxt(arrayBuffer);
         }, tracker);
       });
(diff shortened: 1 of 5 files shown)


tensorboard/plugins/graph/tf_graph_common/parser.ts
```diff
@@ -38,15 +38,14 @@ function parseValue(value: string): string|number|boolean {
...
-    const request = new XMLHttpRequest();
-    request.open('GET', filepath);
-    request.responseType = 'arraybuffer';
-
-    request.onerror = () => reject(request.status);
-    request.onload = () => resolve(request.response);
-
-    request.send(null);
+  return new Promise((resolve, reject) => {
+    fetch(filepath).then(res => {
```
wchargin: style:

> Tip: It is a good practice to use parentheses even for single-argument
> arrows, since the code may still parse reasonably (but incorrectly) if
> the parentheses are forgotten when an additional argument is added.

tensorboard/plugins/graph/tf_graph_common/util.ts
```diff
@@ -106,14 +113,25 @@ module tf.graph.util {
...
+        // Update the progress value.
+        tracker.updateProgress(incProgressValue);
+      }
       // Return the result to be used by other tasks.
       return result;
     } catch (e) {
       // Errors that happen inside asynchronous tasks are
       // reported to the tracker using a user-friendly message.
       tracker.reportError('Failed ' + msg, e);
+      throw e;
```
wchargin: Why throw? The whole point of these utilities is to have nicely
formatted errors rather than dumping stack traces to the console, right?

tensorboard/plugins/graph/tf_graph_common/util.ts
```diff
@@ -106,14 +113,25 @@ module tf.graph.util {
     // UI to update.
     try {
       let result = tf.graph.util.time(msg, task);
-      // Update the progress value.
-      tracker.updateProgress(incProgressValue);
+
+      if (result instanceof Promise) {
```
wchargin: This code doesn’t make any sense to me. `runTask` is used for
synchronous computations. It computes the result synchronously, it
updates the progress synchronously, and it returns the result. Changing
this function such that if the result _happens_ to be a promise then the
time-updating behavior is deferred seems wrong to me. It’s like saying,
“we’ll compute this value for you, and if it happens to be a string then
we’ll also print it to the console.”

We have a separate `runAsyncPromiseTask`, which already supports
updating the progress and logging the timing data asynchronously. If
callers are using `runTask` where they should be using
`runAsyncPromiseTask`, then they are wrong and should be fixed. We
shouldn’t try to shoehorn “sometimes-synchronous sometimes-async”
behavior into an existing function that has a clear contract.

tensorboard/plugins/graph/tf_graph_common/util.ts
stephanwlee: Was related to the "Promise shenanigans". Since we never swallow Promise rejection, I wanted to take it congruent. Removed after removing those Promise checks.

tensorboard/plugins/graph/tf_graph_common/util.ts
stephanwlee: I don't understand the need for two different methods, really, but that really is a separate task we can visit later. Done.

PR #1932, opened 2019-03-01
Fix regression PD plots points order

jameswex:

In some case, regression model PD plot line charts were having x-axis points out-of-order, leading to incorrect plots. Fixed by adding missing sort step that already existed for bar charts.

* Screenshots of UI changes

Broken UI:
broken

Fixed UI:
fixed

Ran colab notebook that showed the problem and verified that with this fix, the colab notebook no longer showed the problem.


jameswex:
@tolga-b please review

jameswex:
@stephanwlee please review, thanks

tensorboard/plugins/interactive_inference/utils/inference_utils.py
```diff
@@ -596,6 +596,7 @@ def make_json_formatted_for_single_chart(mutant_features,
         x_label: value,
         y_label: sum(y_list) / float(len(y_list))
       })
+    list_of_points.sort(key=lambda p: p[x_label])
```
stephanwlee: Can it ever have the same x_label? If so, what breaks the tie? Will sort be indeterministic? Thanks!

jameswex: If two points had the same x-label, the order would be based on the order encountered in the original list. But in how we do PD plots, it isn't possible to have 2 x_labels given we split the range of min to max for 10 x labels, unless min and max x value were set to the same. In that case it wouldn't matter the order of the points as the plot would always looks the same as there is no motion in the x direction.

PR #1935, opened 2019-03-01
Adding collapsible content

erzel:
Adding collapsible-content element and used it in the scale-and-color-controls of par-coords and scatter-plot vizs.
@stephanwlee PTAL.

### DIFF
--- a/tensorboard/components/tf_collapsible/tf-collapsible.html
+++ b/tensorboard/components/tf_collapsible/tf-collapsible.html
@@ -1,0 +1,110 @@
+<!--
+@license
+Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<link rel="import" href="../paper-button/paper-button.html">
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../iron-collapse/iron-collapse.html">
+<link rel="import" href="../iron-icon/iron-icon.html">
+
+<!--
+  An element that shows a button that can be used to collapse / expand
+  arbitrary content found below.
+
+  Example:
+  <tf-collapsible title="Click to expand">
+    Put content here to be expanded / collapsed
+  </tf-collapsible>
+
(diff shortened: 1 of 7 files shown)


tensorboard/components/tf_collapsible/tf-collapsible.html
```diff
@@ -0,0 +1,110 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../iron-collapse/iron-collapse.html">
+<link rel="import" href="../iron-icon/iron-icon.html">
+
+<!--
+  An element that shows a button that can be used to collapse / expand
```
stephanwlee: Can you kindly tell me what does not work with https://www.webcomponents.org/element/@polymer/iron-collapse?
If there is a compelling reason, would you mind writing it down as the comment so we can later know what to use when? Thanks

tensorboard/plugins/hparams/tf_hparams_scale_and_color_controls/tf-hparams-scale-and-color-controls.html
```diff
@@ -28,71 +28,88 @@
...
+            </template>
           </template>
-        </template>
+        </div>
       </div>
-    </div>
+    </tf-collapsible>
   </template>
-
+
```
stephanwlee: extraneous whitespace

tensorboard/components/tf_collapsible/tf-collapsible.html
```diff
@@ -0,0 +1,110 @@
...
+    </style>
+  </template>
+
+  <script>
+    Polymer({
+      is: "tf-collapsible",
+
+      properties: {
+        // The title that appear on the button
+        title: {
```
stephanwlee: If title is the point, what does not work for `tf-category-pane`?

stephanwlee:
Would really appreciate a screenshot for better review. Thanks!

erzel:
Here is a screenshot:
image

GalOshri:
I thought we were planning to go with the approach of moving the axis scale selection to be on the same line as the "color by" selection. That way it still takes up only one line but the user can modify what they need without collapsing/expanding a section.

So the top pane would just be:

Color by: <dropdown>      initial temperature axis scale: <dropdown>       ambient temperature axis scale: <dropdown> ...

erzel:
> I thought we were planning to go with the approach of moving the axis scale selection to be on the same line as the "color by" selection. That way it still takes up only one line but the user can modify what they need without collapsing/expanding a section.
> 
> So the top pane would just be:
> 
> Color by: initial temperature axis scale: ambient temperature axis scale: ...

As far as I recall, we were talking about both options: ('I think another option that mahimap@ suggested is to use the dropdown elements from the what-if tool. If easier, we could make the axis selections look like the color-by option for now. This will take up much less space.'),

But, I think you make a good point about taking less space without collapsing (though we would likely have a horizontal scroll bar if the number of axes is not small). We can change this.

PR #1942, opened 2019-03-02
profile: update demo data to include trace-only run

qqfish:
Summary:
We make this change because we have a bug in profile plugin that will be
triggered if logdir only contains trace data.

Test Plan:
Remove any old profile demo data, then regenerate the data and launch
TensorBoard:

```
$ rm -rf /tmp/profile_demo/
$ bazel run //tensorboard/plugins/profile:profile_demo
$ bazel run //tensorboard -- --logdir /tmp/profile_demo
```

Prior to this commit, this would have _appeared_ to work; now, it hangs
on the “Processing datasets” screen. Fix to come shortly.


wchargin:
Thanks. I’ll send out a fix for the relevant bug presently.

wchargin:
Edited description to include test plan, per the PR template.

PR #1957, opened 2019-03-05
Improve tensorboard docs subsite navigation

nfelt:
This makes some adjustments to the navigation for the TensorBoard docs subsite.  See the preview on staging internally.

The changes I made are:
- Rename the nav titles to be more short and sweet and consistent
- Fix a broken missing `"` in the hparams title
- Re-order the alpha nav listing to put the "tensorboard in notebooks" tutorial last (really this is more like an "Advanced notebook usage" tutorial, since all the guides cover basic notebook usage already...)
- add "TensorBoard" as a prefix to the landing page title
- add actionable buttons to the landing page directing users either to current docs or the alpha
- remove a word from the Scalars guide so the title doesn't spill over to a 2nd line
- fix trailing whitespace to satisfy yaml linting

This also adds a Travis check for .yaml file linting within the docs subdirectory.


docs/_index.yaml
```diff
@@ -4,25 +4,30 @@ description: <!--no description-->
...
           <li>Projecting embeddings to a lower dimensional space</li>
           <li>Displaying images, text, and audio data</li>
           <li>Profiling TensorFlow programs</li>
           <li>And much more</li>
         </ul>
-
       image_path: /tensorboard/images/tensorboard.gif
-
+      buttons:
+      - label: Learn More
```
stephanwlee: Just wondering, what's the rule for putting quotes? Do these need to be quoted like the titles?

GalOshri: The "learn more" here is a bit confusing. I likely landed here by clicking on "learn more" from some other page. Maybe we explicitly refer to visiting the r1 and r2 docs?

wchargin: This is YAML; quotes are roughly optional (subject to a ton of caveats,
because YAML).

nfelt: I guess that didn't seem like an issue to me - I think it would be a continuation of the "learning more" flow since it's still progressively more detail about TensorBoard, and this is basically just an interstitial in that case.  I was trying to avoid framing it just as "r1 and r2" docs since 1) I was trying to make the alpha docs sound more exciting to encourage users to go there for the new stuff and 2) it makes them sound like parallel experiences, even though the content they have is quite different.

I also wasn't sure that we wanted to specifically say just e.g. "View the Docs" in the call-to-action buttons since it sounds relatively dry.

Is there a specific wording you have in mind?

GalOshri: I agree. Keeping it like this seems fine. 
Maybe another idea is to swap the order and have TF 2.0 alpha first to really emphasize that's the exciting stuff, but maybe that just adds to the confusion.

nfelt: Hmm, I tried that on staging but looking at it, I think that might add to the confusion since it's less clear then that "Learn More" is generally about TensorBoard versus about the action to explore the alpha?

It also seemed a little more consistent to keep the alpha on the right in both the navbar and the buttons, and kind in keeping with it being the "upcoming" thing.  My thinking was sort of that we have a standard "learn more" continuation action, but then include the alpha as a teaser button to pull the reader's eye off the beaten track towards the new and exciting thing.  Maybe I'm overthinking it though.

GalOshri: Yeah, I think the current approach sounds like a great start.

nfelt:
FYI, I added a Travis check for .yaml file linting within the docs directory using the `yamllint` package, which has nice config options to make it accept the devsite standard .yaml style.  This should help catch any future issues with string quoting and other YAML gotchas.

PR #1958, opened 2019-03-05
add argument to histograms_demo.py

arp95:

In the histograms_demo.py, we need an argument to let user define the number of iterations for which summary are to be seen.

I have passed the number of iterations with default value of 400 as used earlier, as an argument to the main function of histograms_demo.py file/


wchargin:
> This doesn't seem all that useful without a command line argument
> since you'd still have to edit the file to change this parameter, but
> if it helps you, sure.

These modules are designed to be imported; `run_all` is intentionally
public. The original idea was to have a single demo script that imports
and runs all the demos, saving their output to a single directory
tree—that’s also why they take `logdir` as an argument to `run_all` and
have a consistent interface.

PR #1959, opened 2019-03-05
Fix parser when GraphDef is empty

stephanwlee:
When GraphDef is empty, due to the bug in streamParse, we failed to
resolve or reject the promise.

Added a test and improved test a bit to add malformed pbtxt case, which
in turn required some additional code changes.

### Screenshot
image

Will work on further UI improvements in follow up changes.

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_common/parser.ts
+++ b/tensorboard/plugins/graph/tf_graph_common/parser.ts
@@ -145,7 +145,7 @@ export function streamParse(
       };
       file.readAsText(nextChunk);
     }
-    
+
     readChunk('', '', 0);
   });
 }
--- a/tensorboard/plugins/graph/tf_graph_common/parser.ts
+++ b/tensorboard/plugins/graph/tf_graph_common/parser.ts
@@ -115,62 +115,38 @@ export function streamParse(
     arrayBuffer: ArrayBuffer, callback: (string) => void,
     chunkSize: number = 1000000, delim: string = '\n'): Promise<boolean> {
   return new Promise<boolean>(function(resolve, reject) {
-    let offset = 0;
-    let data = '';
+    function readChunk(oldData: string, newData: string, offset: number) {
+      const doneReading = offset >= arrayBuffer.byteLength;
+      const parts = newData.split(delim);
+      parts[0] = oldData + parts[0];
 
-    function readHandler(str: string) {
-      offset += chunkSize;
-      let parts = str.split(delim);
-      let first = data + parts[0];
-      if (parts.length === 1) {
-        data = first;
-        readChunkIfContent(offset, chunkSize);
-        return;
-      }
-      data = parts[parts.length - 1];
-      try {
-        callback(first);
-        for (let i = 1; i < parts.length - 1; i++) {
-          callback(parts[i]);
-        }
-      } catch (e) {
-        reject(e);
-        return;
-      }
-      readChunkIfContent(offset, chunkSize);
-    }
+      // The last part may be part of a longer string that got cut off
+      // due to the chunking.
+      const remainder = doneReading ? '' : parts.pop();
 
-    /**
-     * Reads a chunk of data from the ArrayBuffer starting from offset.
-     *
-     * It terminates (resolves the promise) if there is no more content to be
-     * read.
-     * @param offset Starting index of content to read from the Buffer.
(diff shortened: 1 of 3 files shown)


stephanwlee:
Assigning review to Nick to spread the load and info.

tensorboard/plugins/graph/tf_graph_common/parser.ts
```diff
@@ -109,50 +109,68 @@ export function fetchAndParseGraphData(path: string, pbTxtFile: Blob,
...
-    function readChunk(offset: number, size: number) {
+    /**
+     * Reads a chunk of data from the ArrayBuffer starting from offset.
+     *
+     * It terminates (resolves the promise) if there is no more content to be
+     * read.
+     * @param offset Starting index of content to read from the Buffer.
+     * @param size The size of each read chunk. (optional)
+     */
+    function readChunkIfContent(offset: number, size: number) {
```
nfelt: The code here feels a lot more complicated than necessary with two different functions that each invoke the other, and the callback being invoked in three different places.  Also the chunk size doesn't need to be passed around since it's a constant within any invocation of `streamParse()`.  Could we simplify to something like this:

```ts
function readChunk(oldData, newData, offset) {
  let doneReading = offset >= arrayBuffer.bytelength;
  let parts = newData.split(delim);
  parts[0] = oldData + parts[0];
  let remainder = '';
  if (!doneReading) {
    remainder = parts.pop();
  }
  for (let part of parts) {
    try {
      callback(part);
    } catch (e) {
      reject(e):
      return;
    }
  }
  if (doneReading) {
    resolve(true);
    return;
  }
  const nextChunk = new Blob([arrayBuffer.slice(offset, offset + chunkSize)]);
  const file = new FileReader();
  file.onload = (e: any) => readChunk(remainder, e.target.result, offset + chunkSize);
  file.readAsText(nextChunk);
}

readChunk('', '', 0);
```

stephanwlee: Good idea! The code was hard to read as is. The new code is an improvement!

tensorboard/plugins/graph/tf_graph_common/parser.ts
```diff
@@ -109,50 +109,68 @@ export function fetchAndParseGraphData(path: string, pbTxtFile: Blob,
...
     let data = '';
 
-    function readHandler(str) {
+    function readHandler(str: string) {
       offset += chunkSize;
       let parts = str.split(delim);
       let first = data + parts[0];
       if (parts.length === 1) {
         data = first;
-        readChunk(offset, chunkSize);
```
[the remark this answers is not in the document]
stephanwlee: Resolved this offline. Reading an empty blob did not result in `file.onload` callback to be invoked. Not sure if this is by spec but this is how we did not have an infinite loop, though, promise never resulted as I originally indicated.

tensorboard/plugins/graph/tf_graph_common/test/parser-test.ts
```diff
@@ -16,84 +16,122 @@ limitations under the License.
...
+              () => {
+                // happy!
+              });
+      });
+
+      it('"parses" pbtxt without newlines.', () => {
+        const pbtxt = tf.graph.test.util.stringToArrayBuffer(
+            `node { name: "Q" op: "Input" } node { name: "A" op: "Input" }`);
+        return tf.graph.parser.parseGraphPbTxt(pbtxt).then((graph) => {
+          assert.notProperty(graph, 'node');
```
nfelt: So this is a failure case, right?  As in this shouldn't behave this way, but it does?  Would be helpful to make that clearer in the test case description, e.g. 'parses pbtxt without newlines as erroneously empty'

tensorboard/plugins/graph/tf_graph_common/test/parser-test.ts
```diff
@@ -16,84 +16,122 @@ limitations under the License.
...
+        node { name: "X" op: "MatMul" input: "Q" input: "W" }
+        node {
+          name: A"
+          op: "Input"
+        }`);
+        return tf.graph.parser.parseGraphPbTxt(pbtxt)
+            .then(
+              () => assert.fail('Should NOT resolve'),
+              () => {
+                // happy!
```
nfelt: maybe clearer as e.g. `// Expected to fail and reject the promise`

Issue #1906, opened 2019-02-26
1.13.1 cherry-pick tracking


stephanwlee:
Below PRs should be CPed into 1.13.1 release.

- https://github.com/tensorflow/tensorboard/pull/1898
- https://github.com/tensorflow/tensorboard/pull/1900
- https://github.com/tensorflow/tensorboard/pull/1914
- https://github.com/tensorflow/tensorboard/pull/1922
- https://github.com/tensorflow/tensorboard/pull/1945

nfelt:
Obsolete.

PR #1900, opened 2019-02-26
manager: use portable epoch time computation

wchargin:
Summary:
Fixes #1895 by replacing the (apparently unportable) `%s` format
specifier with an explicit subtraction from epoch.

Test Plan:
I don’t have easy access to a Windows machine with a Bazel/TensorBoard
setup, but I verified that this expression can be evaluated by itself:

```
Python 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 24 2018, 00:16:47) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license()" for more information.
>>> import datetime
>>> dt = datetime.datetime.now()
>>> dt.strftime("%s")
Traceback (most recent call last):
  File "<pyshell#2>", line 1, in <module>
    dt.strftime("%s")
ValueError: Invalid format string
>>> int((dt - datetime.datetime.fromtimestamp(0)).total_seconds())
1551203889
```

Running `git grep strftime.*%s` returns no matches as of this commit.

wchargin-branch: windows-strftime

Issue #1895, opened 2019-02-26
Error launching TensorBoard on Windows due to `%s` strftime specifier


Origamichen:
TensorBoard version: 1.13.0
Tensorflow version: tensorflow-gpu 1.13.0rc2
OS Platform and version: Windows10 1803
Python version: 3.7

Hi,
Here's a minor error occurs when I run tensorboard 1.13.0. It is fixed by changing **line 50** in tensorboard/manager.py
`serialize=lambda dt: int(dt.strftime("%s")),` to `serialize=lambda dt: int(dt.strftime("%S")),`
The "%s" is changed to "%S". 

> Traceback (most recent call last):
  File "d:\programme\anaconda\Lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "d:\programme\anaconda\Lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "D:\Workspace\Python\TensorflowVenv\TfEnv\Scripts\tensorboard.exe\__main__.py", line 9, in <module>
  File "d:\workspace\python\tensorflowvenv\tfenv\lib\site-packages\tensorboard\main.py", line 57, in run_main
    app.run(tensorboard.main, flags_parser=tensorboard.configure)
  File "d:\workspace\python\tensorflowvenv\tfenv\lib\site-packages\absl\app.py", line 300, in run
    _run_main(main, args)
  File "d:\workspace\python\tensorflowvenv\tfenv\lib\site-packages\absl\app.py", line 251, in _run_main
    sys.exit(main(argv))
  File "d:\workspace\python\tensorflowvenv\tfenv\lib\site-packages\tensorboard\program.py", line 228, in main
    self._register_info(server)
  File "d:\workspace\python\tensorflowvenv\tfenv\lib\site-packages\tensorboard\program.py", line 274, in _register_info
    manager.write_info_file(info)
  File "d:\workspace\python\tensorflowvenv\tfenv\lib\site-packages\tensorboard\manager.py", line 268, in write_info_file
    payload = "%s\n" % _info_to_string(tensorboard_info)
  File "d:\workspace\python\tensorflowvenv\tfenv\lib\site-packages\tensorboard\manager.py", line 128, in _info_to_string
    for k in _TENSORBOARD_INFO_FIELDS
  File "d:\workspace\python\tensorflowvenv\tfenv\lib\site-packages\tensorboard\manager.py", line 128, in <dictcomp>
[...]

Issue #1794, opened 2019-01-31
Loading a csv file in what-if tool to explore data


ellehoej:
Hi, 
I'm trying to use the functionality in the What-If Tool, as described in the Readme file, of loading a .csv file in order to just explore and visualise the data. 
However, when I specify the file path to the csv (in this case to "adult.csv" which is in the Tensorboard logdir folder) in "Path to examples", I keep getting a RequestNetworkError: 

`Request failed: RequestNetworkError: RequestNetworkError: 400 at /data/plugin/whatif/examples_from_path?examples_path=adult.csv&max_examples=1000&sampling_odds=1&sequence_examples=false
`

I have tried to change the formatting and the location of the file with no success. 

Is it possible for you to elaborate on the process of loading a .csv file or perhaps give an example? 

Thanks,
Mads

jameswex:
Please try using the full path to the file on your machine, as opposed to a relative path from the logdir. I believe that will work, and if it does, I will update the README to be more explicit about the csv path.

tomalbrecht:
Same Problem here with tf1.12 and full path.

`Request failed: RequestNetworkError: RequestNetworkError: 400 at /data/plugin/whatif/examples_from_path?examples_path=%2FUsers%2Ftom%2Ftitanic.csv&max_examples=1000&sampling_odds=1&sequence_examples=false`

jameswex:
Does the device/container on which the tensorboard server is executing have access to that file path? The file is loaded by the server itself and the path must be something accessible by the server, not the client/browser connecting to the tensorboard. The documentation isn't clear on this, and I will fix it to be clear. And perhaps we should enable CSV upload from a client so the server doesn't need to be able to load the CSV directly.

Locally I've verified that if I build and run tensorboard from HEAD on my machine ("bazel run tensorboard:tensorboard -- --logdir /tmp") and then navigate to the WIT tab on that tensorboard, and provide a CSV file that has been downloaded to that local machine as the path to examples ("/Users/jwexler/Downloads/test.csv"), that the data points from the CSV load correctly in the tool.

Another thing to check out is the new Jupyter/colab notebook mode where you use WIT directly inside a notebook instead of TensorBoard. In that case, you can load a CSV and convert it to a list of tf.Examples for use in WIT, as shown in this example notebook: https://colab.research.google.com/github/tensorflow/tensorboard/blob/master/tensorboard/plugins/interactive_inference/What_If_Tool_Notebook_Usage.ipynb

[...]

PR #1963, opened 2019-03-05
Release 1.13.1

wchargin:
**Merge this pull request ONLY via rebase, NOT via squash or merge-commit.**

This PR cherry-picks #1898, #1900, #1914, #1922, and #1945 on top of the
1.13 tracking branch to form 1.13.1, per our plan in #1906.

Tested in Python 3 on Windows and in Python 2 on Linux: TensorBoard
starts, and the profile dashboard works (all tools) and is disabled if
you set `window.TENSORBOARD_ENV = {IN_COLAB: true}`, in which case the
help link opens in a new tab.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,12 @@
+# Release 1.13.1
+
+## Bug fixes
+- #1895 - Fix `strftime`-related launch error on Windows (PR #1900)
+- #1794 - Fix What-If Tool loading examples without inference (PR #1898)
+- #1914 - Disable the profile dashboard inside Colab, where it doesn’t work
+- #1945 - Fix profile dashboard loading behavior
+
+
 # Release 1.13.0
 
 The 1.13 minor series tracks TensorFlow 1.13.
--- a/tensorboard/manager.py
+++ b/tensorboard/manager.py
@@ -47,7 +47,8 @@ _FieldType = collections.namedtuple(
 _type_timestamp = _FieldType(
     serialized_type=int,  # seconds since epoch
     runtime_type=datetime.datetime,  # microseconds component ignored
-    serialize=lambda dt: int(dt.strftime("%s")),
+    serialize=lambda dt: int(
+        (dt - datetime.datetime.fromtimestamp(0)).total_seconds()),
     deserialize=lambda n: datetime.datetime.fromtimestamp(n),
 )
 _type_int = _FieldType(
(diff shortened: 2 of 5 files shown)


RELEASE.md
```diff
@@ -1,3 +1,11 @@
+# Release 1.13.1
+
+## Bug fixes
+- Fix launch error on Windows (#1900)
```
nfelt: If these fixed a bug that was actually filed, I list that first, and then put the PR fixing it as `(PR #NNNN)` after that.  If no bug I list the PR first as being the documentation for the bug.  See 1.13.0 for the usual style.

So in this case that would be #1895 fixed by #1900, and #1794 fixed by #1898.  For the profile dashboard I guess we could use #1913 but probably makes sense to just reference #1914.

I think #1945 can be broken out as a separate fix with no attached bug, since it fixes an issue that's not inherently related to Colab and was observable before (by compressing the viewport enough).

Finally, "launch error" is pretty vague - perhaps "Fix strftime-related launch error on Windows"?

wchargin:
Additional notes:

  - I used Bazel 0.22 to build the Pip packages for testing, rather than
    cherry-picking in the BUILD changes for Bazel 0.23 compatibility.

  - When testing on Windows, I did verify that I could reproduce #1895
    on the live version of TensorBoard (`tensorboard==1.13.0`).

  - These were built and tested with `tensorflow==1.13.1`.

RELEASE.md
```diff
@@ -1,3 +1,11 @@
...
+- Fix What-If Tool loading examples without inference (#1898)
+- Disable the profile dashboard inside Colab, where it doesn’t work (#1914, #1922, #1945)
```
nfelt: I've been wrapping to 80 chars except for long URLs per the internal markdown style.

RELEASE.md
wchargin: > If these fixed a bug that was actually filed, I list that first, and
> then put the PR fixing it […] If no bug I list the PR first as being
> the documentation for the bug

Ah, so the first token is _either_ a PR or an issue depending on the
line item—okay, I understand now. Done. (Rest done as well.)

RELEASE.md
wchargin: Done.

RELEASE.md
nfelt: Yes... I won't argue it's the most straightforward precedent ever, but that is what I've been doing.  I'm open to alternative suggestions though as long as we commit to being consistent with some convention.  Mostly the thing I'd like is to include the bug reference itself in the text when fixing a numbered bug so that it's clear from the relnotes without having to visit the PR first.

An alternative convention that's maybe clearer would just be to insert the bug number in the prose text somewhere, like:

```
- Fix #ISSUE strftime-related startup error on Windows (#PR)
- Disable profile dashboard in Colab where it doesn't work yet (#PR)
```

wchargin:
5dc725c74fa75cbf5dbab42ca37812ece5e11a21

RELEASE.md
wchargin: Yep, no worries; consistency is good, and this is fine with me.

PR #1853, opened 2019-02-15
Change graph plugin backend for v2 compatibility

stephanwlee:
# Motivation for features / changes
With TensorFlow v2, graph will be part of RunMetadata and will be
tagged. In order to keep the backwards compatibility, we established new
contract between graph plugin backend and frontend.

# Screenshots of UI changes
### Before
image

### After
image
image

Reviewers: Due to changes in backend APIs, both BE and FE have to be updated at the same time. However, I have broken the changes down into two commits, one for BE and one for FE, for "easier review".

PR #1970, opened 2019-03-06
Fixed tf-graph-app to use the new API

stephanwlee:
PR #1853 made changes to support tags in the graph plugin. When making
the change, we broke the tf-graph-app forgetting to update to use the
new API.

Confirmed the fix by:
1. running the demo and uploading a GraphDef pbtxt 
2. running the TensorBoard with dataset that contains op graph, conceptual graph, and profile and uploaded pbtxt manually.

Disclaimer and thoughts:
- Do not exactly understand use case or how tf-graph-app works.
- If tf-graph-app is truly important, there should be a test at least running the demo app.

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_app/tf-graph-app.html
+++ b/tensorboard/plugins/graph/tf_graph_app/tf-graph-app.html
@@ -16,6 +16,7 @@ limitations under the License.
 -->
 
 <link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../tf-backend/tf-backend.html">
 <link rel="import" href="../tf-graph-board/tf-graph-board.html">
 <link rel="import" href="../tf-graph-controls/tf-graph-controls.html">
 <link rel="import" href="../tf-graph-loader/tf-graph-loader.html">
@@ -98,7 +99,9 @@ tf-graph-controls {
           selected-node="{{selectedNode}}"
           selected-file="{{selectedFile}}"
       ></tf-graph-controls>
-      <tf-graph-loader id="loader"
+      <tf-graph-loader
+          id="loader"
+          router="[[_router]]"
           out-graph-hierarchy="{{graphHierarchy}}"
           out-graph="{{graph}}"
           out-stats="{{stats}}"
@@ -126,6 +129,7 @@ tf-graph-controls {
 
 <script>
 (function(){
+class NotImplementedError extends Error {}
 
 Polymer({
   is: 'tf-graph-app',
@@ -166,6 +170,35 @@ Polymer({
 
     _renderHierarchy: Object,
     _progress: Object,
+    _router: {
+      type: Object,
+      value: function() {
+        return {
+          pluginRoute: (pluginName, route, params) => {
+            if (this.pbtxtFileLocation) {
+              return this.pbtxtFileLocation;
+            }
+            return tf_backend.getRouter().pluginRoute(
+                pluginName, route, params);
+          },
+          environment: () => {
+            throw new NotImplementedError();
+          },
+          experiments: () => {
+            throw new NotImplementedError();
+          },
+          pluginsListing: () => {
+            throw new NotImplementedError();
+          },
+          runs: () => {
+            throw new NotImplementedError();
+          },
+          runsForExp: () => {
+            throw new NotImplementedError();
+          },
+        };
+      },
+    },
   },
   _updateToolbar: function() {
     this.$$('.container').classList.toggle('no-toolbar', !this.toolbar);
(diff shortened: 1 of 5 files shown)


tensorboard/plugins/graph/tf_graph_app/tf-graph-app.html
```diff
@@ -181,10 +214,20 @@
...
+          profile: false,
+        }],
       }];
-      this.$.loader.set('selectedDataset', 0);
+      this.$.loader.selection = {
+        run: this.pbtxtFileLocation,
+        tag: null,
+        type: tf.graph.SelectionType.OP_GRAPH,
+      };
     } else if (this.pbtxt) {
```
nfelt: I think this branch also needs fixes.  In particular the private method below was renamed in #1853 from `_parseAndConstructHierarchicalGraph()` to `_fetchAndConstructHierarchicalGraph()`.  Not sure if anything else needs to change?

tensorboard/plugins/graph/tf_graph_app/tf-graph-app.html
```diff
@@ -166,6 +170,35 @@
 
     _renderHierarchy: Object,
     _progress: Object,
+    _router: {
+      type: Object,
+      value: function() {
+        return {
+          pluginRoute: (pluginName, route, params) => {
```
nfelt: I'm not sure it's a good idea to recreate the whole Router API here as a dummy implementation.  It seems unlikely to stay up to date as we modify the actual Router, unless TypeScript could enforce it for us, and I'm guessing it can't at least in this context.

But more broadly it seems like Router isn't really the right abstraction here.  IMO one of the goals of the graph visualizer is that it's a chunk of self-contained visualization code that doesn't require TensorBoard APIs to work correctly and can be hosted outside TensorBoard; any TensorBoard-specific logic should added just in `<tf-graph-dashboard>`.

It looks like #1853 introduced the dependency from `<tf-graph-loader>` to the Router.  Could we rework that change so we use a smaller, graph-visualizer-specific abstraction instead?  E.g. we could create a simple graph-specific variation on Router that just provides paths for the `/graph` and `/run_metadata` endpoints (and nothing else).  Or maybe we could have the loader take pre-constructed URLs within its `selection` rather than constructing them on-the-fly in `_load()`.

stephanwlee: I thought about those but decided not to do that for following reasons:

There should be a separation of concern between the graph-dashboard (or graph-app) and the graph-loader. The contract between these two components is that the upper component should provide `what` to fetch (which has another contract with the graph-control but it doesn't matter where that comes from in loader's point of view) and graph-loader will do whatever it can to fetch the resources and provide the `outGraph`, `outGraphHierarchy`, and `outStats`. 

With that in mind,
1. Can we pass only subset of the Router instead of implementing all Router interface?
This was the most tempting route but I dissuaded myself by arguing that the graph-dashboard or the graph-app should not be aware of how the URL is formed or which route is used by the loader. As long as, for sake of satisfying odd requirement of the graph-app, the router is passed, the upper component can be oblivious to how the data is being fetched.

Now a counter argument can be that an upper component generally are more business logic dense and should be aware of capability of the loader, thus, should kinda know which API end points will be used. This is fair point but I think both approaches are valid.

2. Can we form URLs in the `selection` or somewhere and pass it to the loader like pre-#1853 days?
I think this violates the principle I tried to follow. Upper component should be as ignorant as possible about _how_ the data is fetched as long as it can define _what_ data it wants to fetch.

3. graph-specific Router?
This again makes the upper component aware of capability + implementation details of how the data is loaded.

tensorboard/plugins/graph/tf_graph_app/tf-graph-app.html
stephanwlee: Didn't think a private method would be used like this. Thanks for catching this and fixed!

tensorboard/plugins/graph/tf_graph_app/tf-graph-app.html
nfelt: It may be easiest to discuss in person, but my initial thoughts are:

> This was the most tempting route but I dissuaded myself by arguing that the graph-dashboard or the graph-app should not be aware of how the URL is formed or which route is used by the loader.

I see this differently.  It makes sense to me that tf-graph-dashboard and tf-graph-app should have more direct control over the data sources used by the graph loader, because they draw on very different data sources, and the current `Router` abstraction is a poor fit for the tf-graph-app data source.  The graph app is used in contexts that are totally unrelated to TensorBoard and don't use the TensorBoard URL structure, and in those contexts it doesn't really make sense for the loader to load from a `pluginRoute` because there is no graph "plugin" at all.  (Additional evidence for this is how the fake Router here returns a constant `pbtxtFileLocation` regardless of the `pluginName` or `route` parameters.)

I see the conceptual job of the graph loader as taking a given dataset, including the data sources (whether blobs or URLs), and implementing just the loading of that data - fetching from the URL if needed, and then parsing the result.  It doesn't seem necessary or really feasible to me for the graph loader to fully encapsulate what data sources it will actually need, because for the tf-graph-app use case you need to explicitly provide the data sources up front anyway - it can't just be a private API between the loader and the plugin backend because there is no plugin backend.

PR #1972, opened 2019-03-06
update development.md

arp95:

While setting up the environment to run tensorboard on my local system, I found certain statements missing which can help the future new developers.

Updated development.md with required statements

* Screenshots of UI changes

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -14,7 +14,9 @@ TensorBoard builds are done with [Bazel](https://bazel.build), so you may need t
 You can build and run TensorBoard via Bazel (from within the TensorFlow nightly virtualenv) as follows:
 
 ```sh
+(tf)$ cd tensorboard/
 (tf)$ bazel build tensorboard
+(tf)$ cd && cd tf/
 (tf)$ ./bazel-bin/tensorboard/tensorboard --logdir path/to/logs
 # Or combine the above steps as:
 (tf)$ bazel run //tensorboard -- --logdir /path/to/logs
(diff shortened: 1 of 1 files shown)


arp95:
@wchargin yes that will do. Making those changes.

arp95:
@wchargin please check now.

wchargin:
> To install TensorFlow nightly within the virtualenv, you can simply
> run the following commands at the root of the Tensorboard Git
> repository:

But you can create the virtualenv anywhere you want, and indeed it’s
better _not_ to put it in the Git work tree because then you have to add
it to your `.git/info/exclude`.

I don’t understand—what problem were you hitting that could be resolved
by creating the virtualenv inside the repository rather than elsewhere?

arp95:
@wchargin the problem was, I created the virtualenv at the home location instead of inside the Git repo. Once I did this, the statements didnt execute as needed and I was not able to follow the steps.

arp95:
If I create the virtualenv at home location, then the commands bazel build tensorboard dont make sense. As at that point I am at home location and need to move to the tensorboard repo to do a bazel build.

Although, it is resolved for me, I can close this if you want.

wchargin:
> If I create the virtualenv at home location, then the commands bazel
> build tensorboard dont make sense. As at that point I am at home
> location and need to move to the tensorboard repo to do a bazel build.

You do indeed need to build TensorBoard inside the TensorBoard repo.

That is totally independent of where the virtualenv lives.

For instance, I personally set up TensorBoard like this:

```
$ cd ~/git/  # where I put repositories
$ git clone tensorboard
$ virtualenv ~/virtualenv/tf-1.13.1
$ . ~/virtualenv/tf-1.13.1/bin/activate
(tf-1.13.1) $ pip install -q tensorflow==1.13.1
(tf-1.13.1) $ cd tensorboard
(tf-1.13.1) $ bazel run //tensorboard -- --logdir ~/tmp/tb_data
```

As you can see, my virtualenv is _not_ in the repository, and I only
need to cd into the repository when I actually want to build from it.

arp95:
@wchargin i mentioned the step "cd tensorboard" to make the process more clear. Anyways should i close this now?

wchargin:
Yes, let’s go ahead and close this.

PR #1974, opened 2019-03-06
Change maximum smoothing factor from 1 to 0.999.

flostim:
1 as a smoothing factor always leads to a constant value of 0 being plotted.
0.999 is the actual highest value you can select for meaningful smoothing.

1.0 smoothing
0.999 smoothing

### DIFF
--- a/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-dashboard.html
+++ b/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-dashboard.html
@@ -84,7 +84,7 @@ limitations under the License.
             weight="{{_smoothingWeight}}"
             step="0.001"
             min="0"
-            max="1"
+            max="0.999"
           ></tf-smoothing-input>
         </div>
         <div class="sidebar-section">


flostim:
@stephanwlee PTAL

GalOshri:
Thank you for submitting this PR @flostim! 

This seems reasonable as a smoothing factor of 1 is indeed not useful (are there any scenarios where it might be?). One question I have is whether users will be confused about the slider not going all the way to 1. Additionally, while we're on this topic, do we really need 3 decimal places here? 

Out of curiosity, are you specifically testing different smoothing factors or are you just trying to make it easier to set it to the maximum reasonable value quickly?

flostim:
Thanks for having a look! 

> This seems reasonable as a smoothing factor of 1 is indeed not useful (are there any scenarios where it might be?).

I don't know of any scenarios where it would be useful.

> One question I have is whether users will be confused about the slider not going all the way to 1. 

Hmm, that's a fair point but I guess if they know that TB is using exponential smoothing they could see that the smoothing factor has to be <1. 

>Additionally, while we're on this topic, do we really need 3 decimal places here?

I wouldn't want to remove functionality here and I definitely sometimes want to have stronger smoothing than 0.99 to get almost an overall average of the scalar. 

> Out of curiosity, are you specifically testing different smoothing factors or are you just trying to make it easier to set it to the maximum reasonable value quickly?

It's mainly that I sometimes want to go to a really high smoothing factor, e.g. to see an overall trend and then it gets set to 1 and I have to go back to have effective smoothing. Very minor annoyance, but as I don't see a reasonable use-case for smoothing factor 1 I thought it could be avoided.

wchargin:
> 1 as a smoothing factor always leads to a constant value of 0 being
> plotted.

Just to provide a bit of context: this was known at the time of
implementation of the current smoothing algorithm. See #639 for various
comparisons to the previous smoothing algorithm, which plotted a
straight line through the first data point rather than through zero for
smoothing=1.0.

flostim:
> Just to provide a bit of context: this was known at the time of
> implementation of the current smoothing algorithm. See #639 for various
> comparisons to the previous smoothing algorithm, which plotted a
> straight line through the first data point rather than through zero for
> smoothing=1.0.
Interesting. I can kind of see a reason to have smoothing=1.0 for the old algorithm but in the new case it just overlays the x-axis.

flostim:
Thanks for the review!

flostim:
Hey @GalOshri,

can you merge this, as I don't have write access. 

Thanks!

GalOshri:
Yep, was just waiting to see if anyone else had questions. Thank you for making this change!

PR #1975, opened 2019-03-07
Use HTTP GET in HParams in colab, POST otherwise.

erzel:

* Screenshots of UI changes

### DIFF
--- a/tensorboard/components/tf_backend/requestManager.ts
+++ b/tensorboard/components/tf_backend/requestManager.ts
@@ -233,8 +233,8 @@ export class RequestManager {
 }
   
 function buildXMLHttpRequest(methodType: HttpMethodType, url: string,
-                             withCredentials? : boolean,
-                             contentType? : string): XMLHttpRequest {
+                             withCredentials?: boolean,
+                             contentType?: string): XMLHttpRequest {
   const req = new XMLHttpRequest();
   req.open(methodType, url);
   if (withCredentials) {
--- a/tensorboard/components/tf_backend/requestManager.ts
+++ b/tensorboard/components/tf_backend/requestManager.ts
@@ -15,17 +15,6 @@ limitations under the License.
 namespace tf_backend {
 
 
-/*==============================================================================
-
-  Please do not use RequestManager for new code.
-
-  We've generally found code that uses XMLHttpRequest without promises is
-  easier to understand and maintain. This API also makes it difficult to use
-  the HTTP protocol in an idiomatic RESTful manner.
-
-==============================================================================*/
-
-
 interface ResolveReject {
   resolve: Function;
   reject: Function;
@@ -44,6 +33,16 @@ export class RequestCancellationError extends Error {
   public name = 'RequestCancellationError';
 }
 
+export class InvalidRequestOptionsError extends Error {
+  public name = 'InvalidRequestOptionsError';
+  constructor(msg : string) {
+    super(msg);
+    // The following is needed due to a limitation of TypeScript when
+    // extending 'Error'. See: https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work
+    Object.setPrototypeOf(this, InvalidRequestOptionsError.prototype);
+  }
+}
+
 export class RequestNetworkError extends Error {
   public name: string;
   public req: XMLHttpRequest;
@@ -57,17 +56,22 @@ export class RequestNetworkError extends Error {
     this.url = url;
   }
 }
-
+ 
+/** The HTTP method-type to use. Currently only 'GET' and 'POST' are
+ * supported.
+ */
+export enum HttpMethodType {
+  GET = 'GET',
+  POST = 'POST',
+}
+  
 /**
  * Holds options that can be used to configure the HTTP request.
  */
 export class RequestOptions {
-  /** The HTTP method-type to use. Currently only 'GET' and 'POST' are
-   * supported.
-   */
-  public methodType: string;
+  public methodType: HttpMethodType;
   
-  /** The content-type request header to use. */
+  /** The content-type request header to use. Cannot be set for a GET request.*/
   public contentType?: string;
 
   /** The request body to use. This is the object that is passed to the 
@@ -80,6 +84,19 @@ export class RequestOptions {
    * XMLHttpRequest.withCredentials property.
    */
   public withCredentials?: boolean;
+
+  // Validates this object. Throws InvalidRequestOptionsError on error.
+  public validate() {
+    if (this.methodType === HttpMethodType.GET) {
+      // We don't allow a body for a GET.
+      if (this.body) {
+        throw new InvalidRequestOptionsError(
+          'body must be missing for a GET request.');
+      }
+    }
+    // We allow body-less or contentType-less POSTs even if they don't
+    // make much sense.
+  }
 }
   
 export class RequestManager {
@@ -102,18 +119,13 @@ export class RequestManager {
    */
   public request(url: string, postData?: {[key: string]: string}):
       Promise<any> {
-    const requestOptions = new RequestOptions();
-    if (postData) {
-      requestOptions.methodType = 'POST';
-      requestOptions.body = this.formDataFromDictionary(postData);
-    } else {
-      requestOptions.methodType = 'GET';
-    }
+    const requestOptions = requestOptionsFromPostData(postData);
     return this.requestWithOptions(url, requestOptions);
   }
 
-  private requestWithOptions(url: string, requestOptions: RequestOptions):
+  public requestWithOptions(url: string, requestOptions: RequestOptions):
       Promise<any> {
+    requestOptions.validate();
     const promise = new Promise((resolve, reject) => {
         const resolver = {resolve: resolve, reject: reject};
         this._queue.push(resolver);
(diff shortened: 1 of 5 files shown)


erzel:
@stephanwlee please take a look.

tensorboard/plugins/hparams/tf_hparams_backend/tf-hparams-backend.html
```diff
@@ -29,37 +29,50 @@
...
+          return this._requestManager.request(url);
+        }
+        /* Use POST */
+        const requestOptions = new tf_backend.RequestOptions();
+        requestOptions.withCredentials = true;
+        requestOptions.methodType = 'POST';
+        requestOptions.contentType = 'text/plain';
+        requestOptions.body = JSON.stringify(request_proto);
+        const url = this._apiUrl + '/' + methodName;
+        return this._requestManager.requestWithOptions(url, requestOptions);
```
stephanwlee: You are invoking a private method here. Please consider making this a public API (which I think is reasonable).

tensorboard/components/tf_backend/requestManager.ts
```diff
@@ -58,6 +58,30 @@ export class RequestNetworkError extends Error {
...
 }
 
+/**
+ * Holds options that can be used to configure the HTTP request.
+ */
+export class RequestOptions {
+  /** The HTTP method-type to use. Currently only 'GET' and 'POST' are
+   * supported.
+   */
+  public methodType: string;
```
stephanwlee: Few things: 
1. This can be an enum with string value.
```ts
export enum MethodType {
  GET = 'GET',
  POST = 'POST,
}
```

2. It seems like there should be some preconditions. For instance, for POST request, body should be required (or does POST without body makes sense?) and there should be an explicit setting of Content-Type when body is present. Do you think these are good ideas? If so, would you mind implementing it?

3. Since this is one of the core library code, would you mind adding some test around the new public API?

erzel: Thanks Stephan!
I've changed the string to an enum, and added some preconditions.
I've added tests and discovered issue #1990 along the way.
PTAL!

tensorboard/plugins/hparams/tf_hparams_backend/tf-hparams-backend.html
erzel: Yes that was my intention. The private was a mistake.

tensorboard/components/tf_backend/requestManager.ts
```diff
@@ -80,6 +84,19 @@ export class RequestOptions {
...
+  // Validates this object. Throws InvalidRequestOptionsError on error.
+  public validate() {
+    if (this.methodType === HttpMethodType.GET) {
+      // We don't allow a body for a GET.
+      if (this.body) {
+        throw new InvalidRequestOptionsError(
+          'body must be missing for a GET request.');
+      }
+    }
+    // We allow body-less or contentType-less POSTs even if they don't
```
stephanwlee: Some validation on POST would be nice. For instance, the body cannot Form when the ContentType is JSON and vice a versa.

tensorboard/components/tf_backend/requestManager.ts
```diff
@@ -221,18 +230,44 @@ export class RequestManager {
...
+  if (!postData) {
+    result.methodType = HttpMethodType.GET;
+    return result;
+  }
+  result.methodType = HttpMethodType.POST;
+  result.body = formDataFromDictionary(postData);
+  return result;
+}
+  
+export function formDataFromDictionary(postData: {[key: string]: string}) {
```
stephanwlee: This does not to be export AFAICT.

tensorboard/components/tf_backend/requestManager.ts
```diff
@@ -221,18 +230,44 @@ export class RequestManager {
...
-      }
+export function requestOptionsFromPostData(postData?: {[key: string]: string}):
+    RequestOptions {
+  const result = new RequestOptions();
+  if (!postData) {
+    result.methodType = HttpMethodType.GET;
+    return result;
+  }
+  result.methodType = HttpMethodType.POST;
+  result.body = formDataFromDictionary(postData);
```
stephanwlee: Can we set the contentType explictly to multipart form data?

tensorboard/components/tf_backend/test/requestManagerTests.ts
```diff
@@ -289,6 +290,83 @@ describe('backend', () => {
...
+           return (rm.request("my_url") as any)
+             .finally(()=>{
+               chai.assert.lengthOf(server.requests, 1);
+               chai.assert.equal(server.requests[0].url, "my_url");
+               chai.assert.equal(server.requests[0].requestBody, null);
+               chai.assert.equal(server.requests[0].method, HttpMethodType.GET);
+               chai.assert.notProperty(server.requests[0].requestHeaders,
+                                       "Content-Type");
+             }).finally(()=>{
+               server.restore();
```
stephanwlee: awesome that you are using sinon fake server. AFAIK, in mocha, `afterEach` gets called after every spec. Is there a reason why you had to manually call this here?

tensorboard/components/tf_backend/test/requestManagerTests.ts
```diff
@@ -289,6 +290,83 @@ describe('backend', () => {
...
+
+      it('builds correct XMLHttpRequest when requestWithOptions is called',
+         function() {
+           const rm = new RequestManager();
+           const requestOptions = new RequestOptions();
+           requestOptions.methodType = HttpMethodType.POST;
+           requestOptions.contentType = "text/plain;charset=utf-8";
+           requestOptions.body = "the body";
+           return (rm.requestWithOptions("my_url", requestOptions) as any)
+             .finally(()=>{
```
stephanwlee: JW, does `.then` better capture what you think?

tensorboard/components/tf_backend/requestManager.ts
erzel: I think this done automatically (from my experiments), and I'd rather not change the behavior from what it used to be.

tensorboard/components/tf_backend/test/requestManagerTests.ts
erzel: Thanks for catching. This was anachronistic.

tensorboard/components/tf_backend/test/requestManagerTests.ts
erzel: I want the assertions to be called even if the promise rejects.

tensorboard/components/tf_backend/requestManager.ts
erzel: Thanks. Done.

tensorboard/components/tf_backend/test/requestManagerTests.ts
stephanwlee: Make sense in perspective of the server.

but that sounds like there should be a test around result (whether the promise resolves or not) and add a case when request fails to resolve too.

tensorboard/components/tf_backend/requestManager.ts
erzel: Hmm. I think it's better if we don't become responsible for enforcing the HTTP protocol standards here, but rather pass the properties opaquely to the XMLHttpRequest.

erzel:
Thanks @stephanwlee . PTAL!

tensorboard/components/tf_backend/requestManager.ts
stephanwlee: Ok, fair!

tensorboard/components/tf_backend/test/requestManagerTests.ts
erzel: Actually, thinking some more about this, I agree that 'then' is better'. Changed.

erzel:
Thanks again @stephanwlee. PTAL!

tensorboard/components/tf_backend/requestManager.ts
```diff
@@ -187,9 +222,52 @@ export class RequestManager {
...
+      }
+      else {
+        req.send();
+      }
     });
   }
 }
+  
+function buildXMLHttpRequest(methodType: HttpMethodType, url: string,
+                             withCredentials? : boolean,
```
stephanwlee: Sorry for one last stylistic nit (we really should have a linter :( ): at least for consistency within this file, can you remove the whitespace before colon? `withCredentials?: boolean`

tensorboard/components/tf_backend/test/requestManagerTests.ts
```diff
@@ -209,7 +209,7 @@ describe('backend', () => {
       // and then launch remaining requests in queue (i.e. this one)
       r1.then((success) => done(), (failure) => done(new Error(failure)));
     });
-
+
```
stephanwlee: Remove me please

tensorboard/components/tf_backend/test/requestManagerTests.ts
```diff
@@ -289,6 +289,81 @@ describe('backend', () => {
...
+           return rm.request("my_url",
+                              {"key1": "value1", "key2": "value2"})
+             .then(() => {
+               chai.assert.lengthOf(server.requests, 1);
+               chai.assert.equal(server.requests[0].url, "my_url");
+               chai.assert.equal(server.requests[0].method,
+                                 HttpMethodType.POST);
+               chai.assert.instanceOf(server.requests[0].requestBody, FormData);
+               chai.assert.sameDeepMembers(
+                 Array.from((server.requests[0].requestBody as any).entries()),
```
stephanwlee: does a `FormData` not have `entries` according to TS? That's odd :(

tensorboard/components/tf_backend/test/requestManagerTests.ts
```diff
@@ -289,6 +289,81 @@ describe('backend', () => {
         // resolving the first request triggers finishTheTest
       });
     });
+
+    it('throws InvalidRequestOptionsError on bad RequestOptions', function() {
+      const rm = new RequestManager();
+        const badOptions = new RequestOptions();
```
stephanwlee: wrong indentation

tensorboard/components/tf_backend/test/requestManagerTests.ts
```diff
@@ -289,6 +289,81 @@ describe('backend', () => {
         // resolving the first request triggers finishTheTest
       });
     });
+
+    it('throws InvalidRequestOptionsError on bad RequestOptions', function() {
```
stephanwlee: I had to think a bit reading this. Can you please change this to say `throws an error when a GET request has a body` instead? (or something to that effect)

tensorboard/components/tf_backend/test/requestManagerTests.ts
wchargin: `FormData` does have `entries`, and `server` is typed as `any` anyway.
This cast is not required.

erzel:
@stephanwlee, if you don't have more comments, can you merge this? I don't have write access. Thanks!

PR #1610, opened 2018-11-14
graph: fix “Download PNG” for large images

wchargin:
Summary:
When viewing large graphs in Chrome, clicking the “Download PNG” button
could the error message, “Failed - Network error”. This is because
Chrome does not support downloading large data-URLs. Chrome does,
however, support downloading large object URLs, so this patch makes the
appropriate change to use that API instead.

Test Plan:
Generate the `audio_demo` data and open the `wave:01;sine_wave` graph in
TensorBoard in Chrome. Transitively expand all node groups. Then, click
“Download PNG”. Before this patch, this consistently failed to download.
After this patch, it downloads and image with the correct contents.

Also tested in Firefox.

wchargin-branch: graph-download-png-blob

PR #1979, opened 2019-03-07
scalar: use ObjectURL for SVG link for durability

stephanwlee:
Data uri in the link for download has given us some trouble with
encoding and length limitations. We want to use the ObjectURL like
PR #1610 that is more durable to aforementioned limitations.

Tested this on Chrome 72 & 74, Firefox 65, and Safari.

### DIFF
--- a/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
+++ b/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
@@ -316,7 +316,7 @@ limitations under the License.
       },
       _updateDownloadLink() {
         const svgStr = this.$$('tf-line-chart-data-loader').exportAsSvgString();
-        const bytes = new Uint8Array(svgStr.length).map((_, i) => svgStr.charCodeAt(i));
+        const bytes = new TextEncoder().encode(svgStr);
         const blob = new Blob([bytes], {type: 'image/svg+xml'});
         this.$$('#svgLink').setAttribute('href', URL.createObjectURL(blob));
       },
--- a/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
+++ b/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
@@ -316,13 +316,6 @@ limitations under the License.
       },
       _updateDownloadLink() {
         const svgStr = this.$$('tf-line-chart-data-loader').exportAsSvgString();
-<<<<<<< HEAD
-        // The SVG code string may include hash characters, such as an
-        // attribute `clipPath="url(#foo)"`. Thus, we base64-encode the
-        // data so that such a hash is not interpreted as a fragment
-        // specifier, truncating the SVG. (See issue #1874.)
-        this.$$('#svgLink').href = `data:image/svg+xml;base64,${btoa(svgStr)}`;
-=======
         const bytes = new Uint8Array(svgStr.length).map((_, i) => svgStr.charCodeAt(i));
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
```diff
@@ -314,11 +316,13 @@
       },
       _updateDownloadLink() {
         const svgStr = this.$$('tf-line-chart-data-loader').exportAsSvgString();
-        // The SVG code string may include hash characters, such as an
-        // attribute `clipPath="url(#foo)"`. Thus, we base64-encode the
-        // data so that such a hash is not interpreted as a fragment
-        // specifier, truncating the SVG. (See issue #1874.)
-        this.$$('#svgLink').href = `data:image/svg+xml;base64,${btoa(svgStr)}`;
+        const bytes = new Uint8Array(svgStr.length).map((_, i) => svgStr.charCodeAt(i));
```
wchargin: This will silently do the wrong thing if `svgStr` contains code points
outside of Latin-1. This could result in incorrect data or a corrupt
SVG: e.g., an attribute `x="my • tag"` will become `x="my " tag"`, which
is clearly problematic.

Please use `new TextEncoder().encode(svgStr)` instead.

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
```diff
@@ -314,11 +316,13 @@
...
+        const blob = new Blob([bytes], {type: 'image/svg+xml'});
+        this.$$('#svgLink').setAttribute('href', URL.createObjectURL(blob));
+      },
+      _removeDownloadLink() {
+        URL.revokeObjectURL(this.$$('#svgLink').href);
```
wchargin: Is it possible to use `this.$.svgLink` here? Or does that not work due
to the template structure?

stephanwlee: It does not work because of wrapping `dom-if`.

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
stephanwlee: Oh wonderful! JW, I guess this is not relevant to graph because it is taking PNG from the canvas?

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
wchargin: Right, okay. Thanks.

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
wchargin: The code in `minimap.ts` is fine because its data comes from an
immediately preceding `atob`, which only emits code points in 0..255.
(Conversely, `btoa` rejects arguments that have code points outside of
Latin-1.) That said, it would be fine to use `TextEncoder` there anyway.

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
stephanwlee: For posterity: dom-if (at least in Polymer v1) does not mount the DOM until the first truthy to `if` property. As a result, at the shadowRoot creation time (kinda speculating based on experience), the DOM does not exist and thus `this.$` is not populated.

PR #1980, opened 2019-03-07
lazy: check sentinel identity with `is`, not `==`

wchargin:
Summary:
This is not likely to be a problem in practice, but we might as well do
the safe thing when comparing to a sentinel.

Test Plan:
Unit test added, for good measure.

wchargin-branch: lazy-sentinel-is

### DIFF
--- a/tensorboard/lazy_test.py
+++ b/tensorboard/lazy_test.py
@@ -87,6 +87,26 @@ class LazyTest(unittest.TestCase):
     with six.assertRaisesRegex(self, ValueError, expected_message):
       bad.day
 
+  def test_loads_only_once_even_when_result_equal_to_everything(self):
+    # This would fail if the implementation of `_memoize` used `==`
+    # rather than `is` to check for the sentinel value.
+    class EqualToEverything(object):
+      init_count = 0
+
+      def __init__(self):
+        EqualToEverything.init_count += 1
+
+      def __eq__(self, other):
+        return True
+
+    @lazy.lazy_load("foo")
+    def foo():
+      return EqualToEverything()
+
+    dir(foo)
+    dir(foo)
+    self.assertEqual(EqualToEverything.init_count, 1)
+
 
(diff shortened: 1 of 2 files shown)


tensorboard/lazy_test.py
```diff
@@ -87,6 +87,26 @@ def bad():
     with six.assertRaisesRegex(self, ValueError, expected_message):
       bad.day
 
+  def test_loads_only_once_even_when_result_equal_to_everything(self):
+    # This would fail if the implementation of `_memoize` used `==`
+    # rather than `is` to check for the sentinel value.
+    class EqualToEverything(object):
+      init_count = 0
```
nfelt: It seems a little better to just have a counter within `foo()` itself?  Since what we are really tracking is that `foo()` should be called just once, regardless of what it does.  That way we separate concerns so that EqualToEverything is only doing what it says on the label and not also being a counter.

This does require an indirection to get around python read-only closures, e.g. `foo.counter = getattr(foo, 'counter', 0) + 1`.

wchargin: I originally had the standard `count_box = [0]` trick, which I can
reinstate. You can’t just store the value on `foo` itself because that
will refer to the wrapped `LazyModule`; i.e., using

```python
    @lazy.lazy_load("foo")
    def foo():
      foo.counter = getattr(foo, "counter", 0) + 1
      return EqualToEverything()
```

raises “Circular import when resolving LazyModule 'foo'”. You can get around this with

```python
    @lazy.lazy_load("foo")
    def foo():
      foo.counter += 1
      return EqualToEverything()
    foo.counter = 0
```

but given that it’s this subtle I’m inclined to stick to the much
simpler close-over-`[0]`, which is at least somewhat idiomatic.

nfelt: Okay, I was being too clever.  The boxing singleton list is fine, sure.

Issue #1613, opened 2018-11-16
Support uploading graph in pbtxt format when Tensorboard is started with empty log dir


iganichev:
When Tensorboard is started with an empty log dir, the "Graphs" tab does not have a way to upload a graph to be visualized. It would be nice to have this button in this case.
![tb](https://user-images.githubusercontent.com/9123400/48637321-3dffe180-e982-11e8-8dcc-ef9b3a58ed0b.png)

arames:
As noted in  #1985, this has regressed. I have not bisected which version regressed, but the feature is working in 2.0.2, but not 2.2.1.
A workaround is to install a specific (working) version via `pip`.

bondhugula:
> As noted in #1985, this has regressed. I have not bisected which version regressed, but the feature is working in 2.0.2, but not 2.2.1.
> A workaround is to install a specific (working) version via `pip`.

But trying to install the older version 2.0.2 will lead to:
```
ERROR: tensorflow 2.2.0 has requirement tensorboard<2.3.0,>=2.2.0, but you'll have tensorboard 2.0.2 which is incompatible.
```
It's not immediately clear which older version of tensorflow to install to go with the above.
```
pip install --user 'tensorflow==2.0.2'
ERROR: Could not find a version that satisfies the requirement tensorflow==2.0.2 (from versions: 2.2.0rc1, 2.2.0rc2, 2.2.0rc3, 2.2.0rc4, 2.2.0, 2.3.0rc0, 2.3.0rc1)
ERROR: No matching distribution found for tensorflow==2.0.2
$ pip install --user 'tensorflow==2.0'
ERROR: Could not find a version that satisfies the requirement tensorflow==2.0 (from versions: 2.2.0rc1, 2.2.0rc2, 2.2.0rc3, 2.2.0rc4, 2.2.0, 2.3.0rc0, 2.3.0rc1)
ERROR: No matching distribution found for tensorflow==2.0
$ pip install --user 'tensorflow==2.0.0'
ERROR: Could not find a version that satisfies the requirement tensorflow==2.0.0 (from versions: 2.2.0rc1, 2.2.0rc2, 2.2.0rc3, 2.2.0rc4, 2.2.0, 2.3.0rc0, 2.3.0rc1)
ERROR: No matching distribution found for tensorflow==2.0.0
$ pip install --user 'tensorflow==2.1.0'
[...]

PR #1985, opened 2019-03-08
Make graph plugin usable without Graph data

stephanwlee:
Graph plugin is possible to use without any graph data on event file by
using its file selector. However, the left panel was hidden when the
event file is missing and thus not possible.

While making the change, I decided to update some pointer to v1 docs for
the v2 preparedness. If TensorBoard team decides to maintain v1.x branch
for 1.14+, we may want a different copy.

This change addresses #1613.

The major changes to look out for:
1. Moved dom-if to the body instead of the whole dashboard so the left-nav is visible always
2. `_datasetsEmpty` is now `_canLoadGraph`. The rationale is, `selectedFile` is like a `datasets` -- it is something the loader understands and something the loader can load graph from. Thus, the tf-graph-board and tf-graph-app should be rendered when selectedFile is present.
3. the `no data found` page content changes

image

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.html
+++ b/tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.html
@@ -846,7 +846,7 @@ Polymer({
       notify: true,
     },
     /**
-     * @type {!Selection}
+     * @type {?Selection}
      */
     selection: {
       type: Object,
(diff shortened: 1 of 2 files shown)


arames:
Note that this has regressed. The left panel is unavailable when starting TensorBoard 2.2.1 with an empty logdir. I was using this to visualize dumped `.pbtxt` files manually. I'll look for a workaround.

arames:
> Note that this has regressed. The left panel is unavailable when starting TensorBoard 2.2.1 with an empty logdir. I was using this to visualize dumped `.pbtxt` files manually. I'll look for a workaround.

I installed TensorBoard 2.0.2 via pip, and that feature is available.

PR #1987, opened 2019-03-08
program: initialize server objects exactly once

wchargin:
Summary:
This resolves a piece of technical debt that was convenient but not
robust to implementation changes in Werkzeug and the Python standard
library. When searching for ports, instead of calling superclass
initializers multiple times until they succeed, we create new server
objects (each with a single port) multiple times until one is
successfully initialized. Each object’s initializer chain is invoked
only once.

The mechanism provides a `with_port_scanning` factory that is agnostic
to the actual server implementation.

This commit is easier to review when ignoring whitespace changes.

Test Plan:
Spawn 11 TensorBoard processes without specifying a port; e.g.:

```
$ bazel build //tensorboard
$ for in in {1..11}; do
> ./bazel-bin/tensorboard/tensorboard --logdir /tmp/whatever &
> done; sleep 2
```

Observe that the first ten bind to successive ports, and that the last
one gives “TensorBoard could not bind to any port around 6006 (tried 10
times)”).

Then, run

```
./bazel-bin/tensorboard/tensorboard --logdir /tmp/whatever --port 6006
```

to see that you still get the “could not bind to port 6006, it was
already in use” error.

Finally, you can use `jobs -p | xargs kill` to kill all extant jobs.

wchargin-branch: program-single-init

### DIFF
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -422,7 +422,7 @@ def with_port_scanning(cls):
 
 
 @with_port_scanning
-class WerkzeugServer(serving.ThreadedWSGIServer):
+class WerkzeugServer(serving.ThreadedWSGIServer, TensorBoardServer):
   """Implementation of TensorBoardServer using the Werkzeug dev server."""
 
   # ThreadedWSGIServer handles this in werkzeug 0.12+ but we allow 0.11.x.
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -341,13 +341,9 @@ class TensorBoardServerException(Exception):
     self.msg = msg
 
 
-class WerkzeugServer(serving.ThreadedWSGIServer, TensorBoardServer):
+class WerkzeugServer(TensorBoardServer):
   """Implementation of TensorBoardServer using the Werkzeug dev server."""
-  # ThreadedWSGIServer handles this in werkzeug 0.12+ but we allow 0.11.x.
-  daemon_threads = True
-
   def __init__(self, wsgi_app, flags):
-    self._flags = flags
     host = flags.host
 
     # base_port: what's the first port to which we should try to bind?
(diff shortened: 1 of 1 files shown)


tensorboard/program.py
```diff
@@ -364,56 +360,77 @@ def __init__(self, wsgi_app, flags):
...
+
+  def get_url(self):
+    return self._server.get_url()
+
+
+class _TensorBoardFailedToBindError(TensorBoardServerException):
+  pass
+
+
+class _TensorBoardWsgiServer(serving.ThreadedWSGIServer):
```
nfelt: It's probably worth a comment either here or on WerkzeugServer or both explaining what the relationship of these two "Server" classes is, since at first glance it's definitely not obvious.

Also, it seems like a subpar naming situation that WerkzeugServer is an instance of TensorBoardServer but strictly speaking contains no Werkzeug-specific logic, while _TensorBoardWsgiServer is an instance of werkzeug.serving.ThreadedWSGIServer yet not an instance of TensorBoardServer.

This could be resolved in a few ways.  One that might actually make sense is turning this class into the original WerzeugServer (by having it inherit from TensorBoardServer), and then renaming the class above to something like WerzeugScanningServer, and just changing the default server class used in the TensorBoard constructor.  (Strictly speaking the class above is more generic than that, and we could make it e.g. ScanningServer and then compose those two yet again to produce WerzeugScanningServer, but I'm not sure that's worth it.)

tensorboard/program.py
```diff
@@ -364,56 +360,77 @@ def __init__(self, wsgi_app, flags):
...
+        if attempt_index < max_attempts - 1:
+          continue
+        elif should_scan:
           raise TensorBoardServerException(
-              'TensorBoard must be run as superuser to bind to port %d' %
+              'TensorBoard could not bind to any port around %s '
+              '(tried %d times)'
+              % (base_port, max_attempts))
+        else:
+          raise TensorBoardServerException(e.msg)
```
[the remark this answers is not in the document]
wchargin: We could. I was aiming to make `_TensorBoardFailedToBindError` an
implementation detail. But I suppose that I’m not opposed to just
re-raising.

tensorboard/program.py
```diff
@@ -364,56 +360,77 @@ def __init__(self, wsgi_app, flags):
...
+            wsgi_app=wsgi_app,
+            host=host,
+            port=port,
+            path_prefix=flags.path_prefix,
+        )
         break
-      except socket.error as e:
-        if hasattr(errno, 'EACCES') and e.errno == errno.EACCES:
+      except _TensorBoardFailedToBindError as e:
+        if attempt_index < max_attempts - 1:
```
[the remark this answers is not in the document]
wchargin: With the `for`–`else` you would need to be careful to use `six.reraise`
and capture the exception info explicitly.

But your second suggestion looks quite attractive! Done.

tensorboard/program.py
wchargin: > It's probably worth a comment either here or on WerkzeugServer or both
> explaining what the relationship of these two "Server" classes is,
> since at first glance it's definitely not obvious.

Yes, agreed. Will do.

> Also, it seems like a subpar naming situation that WerkzeugServer is
> an instance of TensorBoardServer but strictly speaking contains no
> Werkzeug-specific logic, while _TensorBoardWsgiServer is an instance
> of werkzeug.serving.ThreadedWSGIServer yet not an instance of
> TensorBoardServer.

Agreed, mostly. `WerkzeugServer` contains Werkzeug-specific logic in
that it delegates to `_TensorBoardWsgiServer`, which is
Werkzeug-specific.

In your estimation, what are the public API guarantees that we make
around `WerkzeugServer`? I know that `tensorboard.program` is explicitly
a public module.

(Will look at this more carefully tomorrow.)

nfelt: In practice I don't think anyone has really used WerkzeugServer; it's not the primary API we're intending to expose from `tensorboard.program`.  But at any rate I don't think that's actually an issue here since all we'd be changing in any case is the implementation/behavior, not the API per se.

I guess more broadly my point was that certainly the proposed _TensorboardWsgiServer seems more Werkzeug-specific than the  proposed WerkzeugServer so if the latter is going to maintain Werkzeug in its name, it seems like the former should too.

Another way to rework this would be to convert the scanning logic into a factory function on the original WerkzeugServer (aka current _TensorBoardWsgiServer), since that's really more like what it is than a wholly separate class.  We could then retool the `TensorBoard` class a bit so that instead of taking only a server_class it really takes a server_factory function.  If you wanted to avoid an actual API change (which I think it likely harmless) we could add it as a new keyword argument that's mutually exclusive with server_class.

tensorboard/program.py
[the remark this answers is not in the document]
nfelt: If you'd prefer to reraise as the public TensorBoardServerException that's fine, though it seems fine to expose the exception class too, especially if it's still underscore-prefixed to indicate it's not for general use.  (We could also just make it part of the public API.)

wchargin:
That’s true—it could be an ML-style functor that takes one server
implementation (that does not do port scanning) and uses that to
construct a new server implementation (that does do port scanning). That
would be pretty elegant, I agree.

I’d be totally happy to make that change—would this be useful for the
changes that you’d like to make on top of this, or would you prefer that
I just merge roughly as is (maybe modulo renaming symbols)?

nfelt:
If you're up for trying the approach that just adds port scanning generically to any single-port server implementation, I'm happy to wait for that.  I do think it's more elegant and I think it would provide a good resolution to the naming issue.

wchargin:
Done. It’s quite nice.

Pushed as a new commit, but the change is actually easier to review if
you just look at the overall diff ignoring whitespace changes.

tensorboard/program.py
```diff
@@ -341,79 +342,128 @@ def __init__(self, msg):
     self.msg = msg
 
 
+class TensorBoardFailedToBindError(TensorBoardServerException):
```
nfelt: Given that this error actually shouldn't be raised for all failures to bind, maybe it's better to just make this TensorBoardPortInUseError?  That's the only case that we actually want to retry.  Or at a minimum something like TensorBoardRetryableBindError.

tensorboard/program.py
```diff
@@ -341,79 +342,128 @@ def __init__(self, msg):
...
+  This should be raised when it is expected that binding to another
+  similar port would succeed---for instance, if the underlying error is
+  that the address is in use, not that the address family is
+  unsupported. This is used as a signal to indicate that automatic port
+  searching should continue rather than aborting.
+  """
+  pass
+
+
+def with_port_scanning(cls):
```
nfelt: The class decorator is nifty, but it seems to have some functional disadvantages versus just having `with_port_scanning` return `__init__` rather than a wrapper class, and then define `create_port_scanning_werkzeug_server = with_port_scanning(WerkzeugServer)` and making that the default value for `TensorBoard.server_class` (possibly renamed to server_factory).

- it's more conceptually complicated since class decorators that implement functional behavior themselves are somewhat exotic
- it breaks super()
- it requires re-implementing pass-through versions of server_forever() and get_url() that we'd otherwise get for free
- it makes the original WerkzeugServer inaccessible to e.g. test code or other uses

If you're particularly enthusiastic about this approach it's ok, but personally it seems better to keep it simple?

tensorboard/program.py
```diff
@@ -341,79 +342,128 @@ def __init__(self, msg):
...
-        # probably by switching composing around a Werkzeug server
-        # rather than inheriting from it.
-        super(WerkzeugServer, self).__init__(host, port, wsgi_app)
-        break
-      except socket.error as e:
-        if hasattr(errno, 'EACCES') and e.errno == errno.EACCES:
-          raise TensorBoardServerException(
-              'TensorBoard must be run as superuser to bind to port %d' %
+    try:
+      serving.ThreadedWSGIServer.__init__(self, host, port, wsgi_app)
```
nfelt: Comment for why this isn't a call to super()?

tensorboard/program.py
```diff
@@ -466,7 +516,7 @@ def server_bind(self):
         # since that's expected if IPv4 isn't supported at all (IPv6-only).
         if hasattr(errno, 'EAFNOSUPPORT') and e.errno != errno.EAFNOSUPPORT:
           logger.warn('Failed to dual-bind to IPv4 wildcard: %s', str(e))
-    super(WerkzeugServer, self).server_bind()
+    serving.ThreadedWSGIServer.server_bind(self)
```
nfelt: Comment for why this isn't a call to super()?

tensorboard/program.py
```diff
@@ -341,79 +342,128 @@ def __init__(self, msg):
...
+    try:
+      serving.ThreadedWSGIServer.__init__(self, host, port, wsgi_app)
+    except socket.error as e:
+      if hasattr(errno, 'EACCES') and e.errno == errno.EACCES:
+        raise TensorBoardServerException(
+            'TensorBoard must be run as superuser to bind to port %d' %
+            port)
+      elif hasattr(errno, 'EADDRINUSE') and e.errno == errno.EADDRINUSE:
+        if port == 0:
+          raise TensorBoardFailedToBindError(
```
nfelt: We don't actually want to retry in this case, do we?

tensorboard/program.py
wchargin: Both good suggestions, thanks. I’ll go with “port in use” but have no
strong preference.

tensorboard/program.py
wchargin: > return `__init__` rather than a wrapper class

Good point. That should work. Let me try it.

tensorboard/program.py
wchargin: Good catch; thanks.

tensorboard/program.py
wchargin: Done. Kept the name `server_class` because that would technically be a
breaking change, and the functionality didn’t actually change (you could
have passed a factory the whole time).

tensorboard/program.py
wchargin: No longer needed.

tensorboard/program.py
wchargin: No longer needed.

PR #1781, opened 2019-01-26
Make tensorboard.compat.{tf,tf2} lazily loaded

nfelt:
This PR makes `tensorboard.compat.tf` into a lazy-loading module for accessing the TF API (either real or stub), and introduces a similar `tensorboard.compat.tf2` lazy-loading module for accessing the TF 2.0 API (either as `tf` or `tf.compat.v2`). 

The first change (lazifying `tensorboard.compat.tf`) is necessary so that that the `tensorboard.summary.v2` summary API can be imported without immediately importing TensorFlow, since we want TF to be able to import that without triggering a circular dependency.  Previously, the V2 summary API transitively imports various pieces from under `tensorboard.compat`, and that module's `__init__.py` file was opportunistically importing `tensorflow` at module load time to provide `tensorboard.compat.tf`).  Now that import is deferred until `tf` is actually used, and we avoid using it from within the V2 summary APIs.  I've added `summary_dep_test.py` to enforce this so we don't have a regression.

To implement this lazy-loading behavior, I refactored LazyLoader to let you customize the module loading with your own function, and made it usable as a decorator for convenience.  I also refactored the existing LazyLoader uses in `tensorboard/__init__.py` to use this, and replaced `tensorboard.compat.import_tf_v2()` with a lazy loading `tensorboard.compat.tf2`, which in turn eliminates the need to call `import_tf_v2()` on-demand in the various places that use it (necessary because we don't always have a TF 2.0 API available).

PR #2005, opened 2019-03-13
init: fix top-level module reloading

### LINKED ISSUE #1989: After reloading `tensorboard` module, public APIs are broken
The public APIs `notebook`, `program`, and `summary` are exposed from
the `tensorboard` module as `LazyModule`s. If the `tensorboard` module
is reloaded after one of these modules has been dereferenced, then
attempting to dereference that module again will blow the stack on a
recursion error.

In a Python 2 virtualenv:

```
$ python2
Python 2.7.14+ (default, Dec  5 2017, 15:17:02) 
[GCC 7.2.0] on linux2
>>> import tensorboard
>>> tensorboard.program.TensorBoard
<class 'tensorboard.program.TensorBoard'>
>>> import imp  # Python 2
>>> imp.reload(tensorboard)
<module 'tensorboard' from '<snip>'>
>>> tensorboard.program.TensorBoard
  File "<snip>/tensorboard/lazy.py", line 65, in __getattr__
    return getattr(load_once(self), attr_name)
  <snip>
  File "<snip>/tensorboard/lazy.py", line 65, in __getattr__
    return getattr(load_once(self), attr_name)
RuntimeError: maximum recursion depth exceeded
```

Or in a Python 3 virtualenv:

```
$ python3
Python 3.6.5 (default, Mar 31 2018, 05:34:57) 
[GCC 7.3.0] on linux
>>> import tensorboard
>>> tensorboard.program.TensorBoard
<class 'tensorboard.program.TensorBoard'>
>>> import importlib  # Python 3
>>> importlib.reload(tensorboard)
<module 'tensorboard' from '<snip>'>
>>> tensorboard.program.TensorBoard
  File "<snip>/tensorboard/lazy.py", line 65, in __getattr__
    return getattr(load_once(self), attr_name)
  <snip>
  [Previous line repeated 329 more times]
RecursionError: maximum recursion depth exceeded
```

Note that using the IPython directive `%reload_ext EXTENSION` will
reload the module specified by `EXTENSION`, so this precludes us from
exposing `%load_ext tensorboard`. (Reloading a module directly via
`imp`/`importlib` may not be super common, but using `%reload_ext` is.)

As I understand it, the execution flow is:

  - `tensorboard` is initialized. `tensorboard.program` is a LazyModule.
  - `tensorboard.program` is forced, importing the `program` module for
    the first time, which effects the following:
      - `sys.modules["tensorboard.program"]` is initialized to an empty
        module object.
      - The `program` module is initialized.
      - The `program` attribute on the `tensorboard` module object is
        set to the `program` module object.
  - `tensorboard` is reloaded. `tensorboard.program` is reset to a lazy
    module.
  - `tensorboard.program` is forced, importing the `program` module.
    Because this module already exists in `sys.modules`, it is not
    reinitialized, _and_ the `program` attribute on the `tensorboard`
    module object is not affected. Thus, the LazyModule initializer
    resolves to the lazy module itself, hitting a recursion error.

I believe that it should suffice to change

```py
  import tensorboard.program as module  # pylint: disable=g-import-not-at-top
  return module
```

to

```py
  return importlib.import_module("tensorboard.program")
```

with the caveat that this would require changes in the Google-internal
sync process (and I’m not exactly thrilled about these changes:
rewriting `import` syntax forms is one thing, but rewriting function
calls that happen to be of the form `importlib.import_module` is in
principle neither sound nor complete…—but, if it’s the only thing that
works, then so be it).

cc resident module spelunking expert @nfelt for confirmation

wchargin:
Summary:
Fixes #1989 (also described in a new comment in `__init__.py`).

This will require a Google-internal sync change: <http://cl/238144308>.

Test Plan:
Regression test added. It fails before this change in Python 2 and 3,
and passes after this change in both as well. The <http://cl/238144308>
test plan also passes.

wchargin-branch: reload-tensorboard

### DIFF
--- a/tensorboard/lib_test.py
+++ b/tensorboard/lib_test.py
@@ -21,7 +21,7 @@ import sys
 import unittest
 
 
-class StopshipTest(unittest.TestCase):
+class ReloadTensorBoardTest(unittest.TestCase):
 
   def test_functional_after_reload(self):
     self.assertNotIn("tensorboard", sys.modules)
--- a/tensorboard/lib_test.py
+++ b/tensorboard/lib_test.py
@@ -1,0 +1,43 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from six import moves
+import sys
+import unittest
+
+
+class StopshipTest(unittest.TestCase):
+
+  def test_functional_after_reload(self):
+    self.assertNotIn("tensorboard", sys.modules)
+    import tensorboard as tensorboard  # it makes the Google sync happy
+    submodules = ["notebook", "program", "summary"]
+    dirs_before = {
+        module_name: dir(getattr(tensorboard, module_name))
+        for module_name in submodules
+    }
+    tensorboard = moves.reload_module(tensorboard)
+    dirs_after = {
+        module_name: dir(getattr(tensorboard, module_name))
+        for module_name in submodules
+    }
+    self.assertEqual(dirs_before, dirs_after)
+
+
+if __name__ == '__main__':
+  unittest.main()
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -82,7 +82,6 @@ py_test(
     ],
 )
 
-
 py_library(
     name = "manager",
     srcs = ["manager.py"],
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -73,6 +73,7 @@ py_library(
 
 py_test(
     name = "lib_test",
+    size = "small",
     srcs = ["lib_test.py"],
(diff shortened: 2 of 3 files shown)


tensorboard/lib_test.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from six import moves
+import sys
+import unittest
+
+
+class StopshipTest(unittest.TestCase):
```
nfelt: did you perhaps mean to rename this?

wchargin: I did! Thank you. :-)

wchargin: (Should have called it `Do_Not_Submit_Test`, I suppose.)

wchargin:
> check if the named module already exists in sys.modules

Presuming that you mean the argument to `lazy.lazy_load`… this seems
like an opportunity for confusion/surprise. These names aren’t required
to correspond to actual modules at all: e.g., the `tensorboard.compat`
LazyModules have module-like names, but no such modules exist. For this
string to sometimes be used decoratively and sometimes be used
semantically would be an unnecessary inconsistency, imo. If it were to
_always_ behave semantically, then that would be fine—but as of #1781
that is deliberately not the case.

nfelt:
Fair enough.  I think it is fair to require that the names passed to the decorator match the actual importable name of the decorated symbol, which is all that's really required for this to work (if there is no module in sys.modules, you fall back to the lazy function, regardless of whether there ever is such a module).  But what you've done is fine too.

wchargin:
> if there is no module in sys.modules, you fall back to the lazy
> function, regardless of whether there ever is such a module

Yep, understood. It only really breaks if you have a `lazy_load("foo")`
and a module `foo` but the lazy module loads a _different_ module, which
seems like a bad idea all around.

> I think it is fair to require that the names passed to the decorator
> match the actual importable name of the decorated symbol, which is all
> that's really required for this to work

This already doesn’t hold within Google, though. :-)

PR #2006, opened 2019-03-13
Use isinstance rather than type to support CorePlugin subclass

orionr:

Was unable to subclass CorePlugin so we could update the log directory path via a URL class.

Use isinstance instead of type.

cc @nfelt


orionr:
Thanks @wchargin. Something we're exploring internally is changing the log directory via a URL so multiple people can use the same server instance over time (from a pool of them). For this to work we need to make a number of changes, but this seemed like a small bug. Sounds like it wasn't!  We'd talked about posting an issue around this "on demand" model for TensorBoard, but wanted to talk with everyone on Friday before doing that.

PR #2009, opened 2019-03-14
Use console logs to determine web test results

### LINKED ISSUE #1960: tf-web-test reports success when there is a failure
Repro case:

1. Add below code to tensorboard/plugins/graph/tf_graph_common/test/parser-test.ts
```ts
  it('fails', () => {
    const pbtxt = tf.graph.test.util.stringToArrayBuffer(``);
    return tf.graph.parser.parseGraphPbTxt(pbtxt)
        .then(() => assert.fail('Should NOT resolve'));
  })
```
2. [Optional] Confirm that it fails when run with `bazel run tensorboard/plugins/graph/tf_graph_common/test:test_web_library`
3. Run `bazel test tensorboard/plugins/graph/tf_graph_common/...`
4. Expected failure but it passes.

erzel:
Fixes #1960.

### DIFF
--- a/tensorboard/functionaltests/wct_test_driver.py
+++ b/tensorboard/functionaltests/wct_test_driver.py
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""WebDriver for running TypeScript and Polymer unit tests."""
+"""WebDriver for running TypeScript and Polymer unit tests.
+
+Figures out the result by scanning the browser log.
+"""
 
 from __future__ import absolute_import
 from __future__ import division
@@ -46,6 +49,30 @@ def create_test_class(binary_path, web_path):
     A new subclass of `unittest.TestCase`. Bind this to a variable in
     the test file's main module.
   """
+
+  class BrowserLogIndicatesResult(object):
+    def __init__(self):
+      self.passed = False
+    def __call__(self, driver):
+      # Scan through the log entries and search for a text indicating whether
+      # the test passed or failed.
+      messages = [entry['message'] for entry in driver.get_log('browser')]
+      if self._log_contains(messages, 'failing test'):
+        self.passed = False
+        return True
+      if self._log_contains(messages, 'test suite passed'):
+        self.passed = True
+        return True
+      # Here, we still don't know.
+      return False
+
+    def _log_contains(self, messages, text):
+      for message in messages:
+        if text in message:
+          # Print log message as an aid for debugging.
+          print("Found: %s in log entry message: %s" % (text, message))
+          return True
+      return False
 
   class WebComponentTesterTest(unittest.TestCase):
     """Tests that a family of unit tests completes successfully."""
@@ -90,13 +117,14 @@ def create_test_class(binary_path, web_path):
       cls.process.wait()
 
     def test(self):
-      driver = webtest.new_webdriver_session()
+      driver = webtest.new_webdriver_session(
+          capabilities={'loggingPrefs':{'browser':'ALL'}}
+      )
       url = "http://localhost:%s%s" % (self.port, web_path)
       driver.get(url)
-      wait.WebDriverWait(driver, 10).until(
-          expected_conditions.title_contains("test"))
-      title = driver.title
-      if "failing test" in title or "passing test" not in title:
-        self.fail(title)
+      browser_log_indicates_result = BrowserLogIndicatesResult()
+      wait.WebDriverWait(driver, 10).until(browser_log_indicates_result)
+      if not browser_log_indicates_result.passed:
+        self.fail()
 
   return WebComponentTesterTest
--- a/tensorboard/functionaltests/wct_test_driver.py
+++ b/tensorboard/functionaltests/wct_test_driver.py
@@ -134,11 +134,11 @@ def create_test_class(binary_path, web_path):
       finally:
         # Print log as an aid for debugging.
         log = browser_log_indicates_result.log + driver.get_log("browser")
-        self._printLog(log)
+        self._print_log(log)
 
-    def _printLog(self, entries):
-      print ("Browser log follows:")
-      print ("--------------------")
+    def _print_log(self, entries):
+      print("Browser log follows:")
+      print("--------------------")
       print(" | ".join(entries[0].keys()))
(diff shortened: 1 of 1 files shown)


tensorboard/functionaltests/wct_test_driver.py
```diff
@@ -47,6 +50,30 @@ def create_test_class(binary_path, web_path):
...
+        self.passed = True
+        return True
+      # Here, we still don't know.
+      return False
+
+    def _log_contains(self, messages, text):
+      for message in messages:
+        if text in message:
+          # Print log message as an aid for debugging.
+          print("Found: %s in log entry message: %s" % (text, message))
```
wchargin: Note that the message is just `"%c1 failing test"`, so I’m not sure how
much this actually helps debugging (e.g., it doesn’t give you the name
of the failing test case), but this is fine with me.

tensorboard/functionaltests/wct_test_driver.py
```diff
@@ -90,13 +117,14 @@ def tearDown(cls):
...
       url = "http://localhost:%s%s" % (self.port, web_path)
       driver.get(url)
-      wait.WebDriverWait(driver, 10).until(
-          expected_conditions.title_contains("test"))
-      title = driver.title
-      if "failing test" in title or "passing test" not in title:
-        self.fail(title)
+      browser_log_indicates_result = BrowserLogIndicatesResult()
+      wait.WebDriverWait(driver, 10).until(browser_log_indicates_result)
+      if not browser_log_indicates_result.passed:
```
wchargin: `self.assertTrue(browser_log_indicates_result.passed)` is simpler, yes?

tensorboard/functionaltests/wct_test_driver.py
```diff
@@ -47,6 +50,30 @@ def create_test_class(binary_path, web_path):
...
+    def __init__(self):
+      self.passed = False
+    def __call__(self, driver):
+      # Scan through the log entries and search for a text indicating whether
+      # the test passed or failed.
+      messages = [entry['message'] for entry in driver.get_log('browser')]
+      if self._log_contains(messages, 'failing test'):
+        self.passed = False
+        return True
+      if self._log_contains(messages, 'test suite passed'):
```
wchargin: Do you have a reference that indicates that the driver will only print
“test suite passed” once _all_ tests have passed? For instance, it’s not
obvious to me that this waits for all top-level `describe` blocks—I
checked, and it appears to, but I don’t see any guarantee of that.

tensorboard/functionaltests/wct_test_driver.py
erzel: I had that initially, but then the error message for a passing test is something like 'False' is 'True', which I didn't like so much. It's only slightly better with 'fail'.

tensorboard/functionaltests/wct_test_driver.py
erzel: I'm not a 100% certain, but if I look at console.js in chrome's debugger there's a function called logSummary that seems to be called only at the end of the mocha run that logs 'test suite passed'.

tensorboard/functionaltests/wct_test_driver.py
erzel: Good point. Changed to output entire log at end of test.

erzel:
Thanks @wchargin !
PTAL.

tensorboard/functionaltests/wct_test_driver.py
wchargin: Okay, no problem—if we find out that this is insufficient then we can
update this test code again. :-)

tensorboard/functionaltests/wct_test_driver.py
wchargin: Got it; thanks.

tensorboard/functionaltests/wct_test_driver.py
```diff
@@ -90,13 +121,26 @@ def tearDown(cls):
...
+      try:
+        wait.WebDriverWait(driver, 10).until(browser_log_indicates_result)
+        if not browser_log_indicates_result.passed:
+          self.fail()
+      finally:
+        # Print log as an aid for debugging.
+        log = browser_log_indicates_result.log + driver.get_log("browser")
+        self._printLog(log)
+
+    def _printLog(self, entries):
```
wchargin: `_print_log`, not `_print_log`

tensorboard/functionaltests/wct_test_driver.py
```diff
@@ -90,13 +121,26 @@ def tearDown(cls):
...
+      print ("Browser log follows:")
```
wchargin: remove space between `print` and parenthesis

tensorboard/functionaltests/wct_test_driver.py
```diff
@@ -90,13 +121,26 @@ def tearDown(cls):
...
-        self.fail(title)
+      browser_log_indicates_result = BrowserLogIndicatesResult()
+      try:
+        wait.WebDriverWait(driver, 10).until(browser_log_indicates_result)
+        if not browser_log_indicates_result.passed:
+          self.fail()
+      finally:
+        # Print log as an aid for debugging.
+        log = browser_log_indicates_result.log + driver.get_log("browser")
+        self._printLog(log)
```
wchargin: Nice; thanks! It looks like this prints the assertion failures
(“expected 0, got 1”), which is definitely better than the previous
behavior even though the test case name isn’t logged.

tensorboard/functionaltests/wct_test_driver.py
erzel: Done.

tensorboard/functionaltests/wct_test_driver.py
erzel: Oops. Done.

tensorboard/functionaltests/wct_test_driver.py
erzel: Yep. It would be nice to try to get that as well, but I'll stop here for now.
PTAL. Thanks!

erzel: Note that I don't have write-access and cannot merge.

PR #2011, opened 2019-03-14
Use tensorboard.compat for tf in projector plugin

orionr:
Not sure how I missed this, but we need to use `tensorboard.compat` for the TensorFlow dependency in the projector plugin. Saw this failure when attempting to use internally.

cc @nfelt @wchargin

### DIFF
--- a/tensorboard/plugins/projector/__init__.py
+++ b/tensorboard/plugins/projector/__init__.py
@@ -28,7 +28,7 @@ from __future__ import print_function
 import os
 
 from google.protobuf import text_format as _text_format
-import tensorflow as tf
+from tensorboard.compat import tf2 as tf
 from tensorboard.plugins.projector import projector_plugin as _projector_plugin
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/projector/__init__.py
```diff
@@ -28,7 +28,7 @@
 import os
 
 from google.protobuf import text_format as _text_format
-import tensorflow as tf
+from tensorboard.compat import tf2 as tf
```
nfelt: This should probably just be `import tf`.  Using `tf2` won't ever use the stub because the stub doesn't include the TF 2.0 only APIs, and I'm guessing your intent with using compat was to allow use of the stub?  Also it shouldn't be necessary to use tf2 since `tf.io.gfile` exists in 1.13+ as well as 2.0.

orionr: I wasn't clear on the difference between `tf` and `tf2` in this case - makes sense. Fixing.

PR #2012, opened 2019-03-14
Graph: refactor Keras util for readability

stephanwlee:
Keras util has to make some improvements to address bugs but this less
than a month old code was quite hard to read. In order to improve the
readability, we:
1. formalized the Keras data types as namedtuple.
2. normalize Sequential model as Functional model so we do not have to
   special case and maintain the Sequential model flows.
3. Instead of passing dict around when collecting metadata in a single
   iteration, we now "walk" three times to normalize model, collect
   metadata and build the GraphDef.

Note that this change tried to minimize extraneous changes not related to the refactor.

TDD (test covers cases reasonably complex models) and exercised the code against the TensorBoard graph tutorial

### DIFF
--- a/tensorboard/plugins/graph/keras_util.py
+++ b/tensorboard/plugins/graph/keras_util.py
@@ -57,7 +57,7 @@ TensorKeyWithArgs = collections.namedtuple(
 
 # A `TensorKey` represents an entry in either `input_layers` or `output_layers`
 # fields of a Keras functional model's `config` data. See module docstring for
-# more nformation. Alternatively, see below link for TensorFlow.js type
+# more information. Alternatively, see below link for TensorFlow.js type
 # definition.
 # https://github.com/tensorflow/tfjs-layers/blob/4106f4290ef18e7bde7baaa9995b82b6024ea2ef/src/keras_format/model_serialization.ts#L19-L20
 TensorKey = collections.namedtuple(
@@ -118,7 +118,7 @@ def _is_model(layer):
   Returns:
     bool: True if layer is a model.
   """
-  return layer.get('config').get('layers') is not None
+  return layer.get('config', {}).get('layers') is not None
 
 
 def _get_inbound_nodes(tensor_key_with_args_array):
@@ -126,19 +126,19 @@ def _get_inbound_nodes(tensor_key_with_args_array):
 
   Keras model serialization records input to a layer in a property,
   `tensor_key_with_args_array`, and it can be of below forms:
-  - [['name_1', 1, 0], ['name_2', 1, 1]]: two invocations witn single input.
+  - [['name_1', 1, 0], ['name_2', 1, 1]]: two invocations with a single input.
   - [
       [['name_1', 0, 0], ['name_2', 1, 0]],
       [['name_3', 0, 0], ['name_4', 1, 0]]]: two invocations with multiple
     inputs. Multiple inputs are only applicable to limited keras layers like
     keras.layers.add, keras.layers.concat, and etc...
 
-  Note that there cannot be fixture of types as a layer that takes multiple
-  inputs have precondition on size of the input.
+  Note that there cannot be a mixture of two forms where a layer can take both
+  a single input and multiple inputs.
 
   Args:
-    tensor_key_with_args_array: Value of `tensor_key_with_args_array` of a Keras
-    model serialization.
+    tensor_key_with_args_array: Value of `inbound_nodes` of a Keras model
+        serialization.
 
   Raises:
     ValueError: when the `tensor_key_with_args_array` violate our assumption
@@ -147,16 +147,14 @@ def _get_inbound_nodes(tensor_key_with_args_array):
   Returns:
     List of invocations that has a list of inputs.
   """
-  normalized_key_array_list = tensor_key_with_args_array \
-      if tensor_key_with_args_array else []
+  normalized_key_array_list = tensor_key_with_args_array or []
 
   is_single_input_mode = isinstance(
       normalized_key_array_list[0][0], six.string_types) \
           if normalized_key_array_list else False
   invocations = []
   for key_array in normalized_key_array_list:
-    if (is_single_input_mode and not
-        isinstance(key_array[0], six.string_types)):
+    if (is_single_input_mode != isinstance(key_array[0], six.string_types)):
       raise ValueError('Expected all inbound nodes to have the same type')
 
     if is_single_input_mode:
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/graph/keras_util.py
```diff
@@ -41,18 +40,41 @@
...
+import six
+
 from tensorboard.compat.proto.graph_pb2 import GraphDef
 from tensorboard.compat.tensorflow_stub import dtypes
 
 
+# Please refer to below link for more information.
+# https://github.com/tensorflow/tensorflow/blob/1020739e1724346ef0088186f0253defaad3edb6/tensorflow/python/keras/engine/network.py#L1079-L1090
+InboundNodes = collections.namedtuple(
+    'InboundNodes', ['inbound_layer', 'node_index', 'tensor_id', 'kwargs'])
```
wchargin: What types of values are `node_index` and `tensor_id`? From the names,
I’d assume “int” and “locally unique string” respectively… but the
linked code in Keras actually calls them both `*_id`, whereas other code
in Keras calls them both `*_index`.

tensorboard/plugins/graph/keras_util.py
```diff
@@ -41,18 +40,41 @@
...
+import collections
+import six
+
 from tensorboard.compat.proto.graph_pb2 import GraphDef
 from tensorboard.compat.tensorflow_stub import dtypes
 
 
+# Please refer to below link for more information.
+# https://github.com/tensorflow/tensorflow/blob/1020739e1724346ef0088186f0253defaad3edb6/tensorflow/python/keras/engine/network.py#L1079-L1090
+InboundNodes = collections.namedtuple(
```
wchargin: Is `InboundNodesEntry` or `InboundEdge` or similar a more appropriate
name here? Nodes themselves are never really “inbound”, right?

tensorboard/plugins/graph/keras_util.py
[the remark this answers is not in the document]
stephanwlee: Honestly, I generally am confused about nature of the fourth data. https://github.com/tensorflow/tensorflow/blob/1020739e1724346ef0088186f0253defaad3edb6/tensorflow/python/keras/engine/network.py#L1086
As you may have noticed, it is set to kwargs and, no, I have yet to find an example.

stephanwlee: I could see why that is confusing and I am confused in general too. Keras does not define the term firmly and like you pointed out earlier, some code refers to data as index and others refer to it as id.
I honestly dunno what to call them except I want to be consistent with _something_. Would you prefer index like the definition of `iterate_inbound`?
https://github.com/tensorflow/tensorflow/blob/1020739e1724346ef0088186f0253defaad3edb6/tensorflow/python/keras/engine/network.py#L1081

stephanwlee:
> For NodeData, I’m a bit more lost. I can’t easily guess what part of
the Keras ontology this corresponds to—or is it a new notion that you
introduce? I read the linked code snippet, but have no idea what it
refers to; it appears to be a polymorphic internal data conversion
function, not a definition of a “node data” structure. What part of the
model JSON are we talking about?

Fair point. I tried to infer nature of the data from this line https://github.com/tensorflow/tensorflow/blob/1020739e1724346ef0088186f0253defaad3edb6/tensorflow/python/keras/engine/network.py#L1110
What is your recommendation? I am lacking creativity here. Perhaps this is where "wrong abstraction is more expensive than copypasta"?

> Similarly for ModelMetadata. My guess is that this is your invention
for the convenience of this module, but I’m not sure what it represents.
I’d assume that there is exactly one ModelMetadata per model, but the
type includes a model_name_to_outputs map—are the keys names of
“sub-models” or something?

Yes, ModelMetadata is my invention. It is just a collection of dicts from traversing *models* and *layers* and contain information about weak connection between models/layers. There will be only one ModelMetadata for one root level model. Now, do you think it is a bad idea to create a namedtuple for this?

tensorboard/plugins/graph/keras_util.py
wchargin: I care most that we pick one and use it consistently. I tested this a
bit more and it _looks_ to me like both of these values are always
integers, and also TFJS calls them `NodeIndex` and `TensorIndex`
respectively, so maybe a slight preference for `index` over `id`?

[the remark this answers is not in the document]
wchargin: Okay, I found an example:

```python
>>> import tensorflow as tf
>>> inputs = tf.keras.layers.Input(shape=(32, 32, 3))
>>> dropout = tf.keras.layers.Dropout(0.2)(inputs, training=True)
>>> model = tf.keras.models.Model(inputs=inputs, outputs=dropout)
>>> import json
>>> print(json.dumps(json.loads(model.to_json())["config"]["layers"][1], indent=4))
{
    "name": "dropout",
    "class_name": "Dropout",
    "config": {
        "name": "dropout",
        "trainable": true,
        "dtype": "float32",
        "rate": 0.2,
        "noise_shape": null,
        "seed": null
    },
    "inbound_nodes": [
        [
            [
                "input_1",
                0,
                0,
                {
                    "training": true
                }
            ]
        ]
    ]
}
```

As you can see, these `kwargs` are additional arguments passed to the
layer at `call` invocation time rather than layer construction time.
Only certain layers admit extra parameters here.

stephanwlee:
Assuming that TF.js team is using correct terminology based on their discussions with the Keras team, I decided to follow their vocabularies. I believe, this should resolve some of the high level concerns you had.

tensorboard/plugins/graph/keras_util.py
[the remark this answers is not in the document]
stephanwlee: Duh, I by mistake passed it onto the constructor without reading the code carefully. Thanks for the example.

tensorboard/plugins/graph/keras_util.py
```diff
@@ -92,89 +121,171 @@ def _is_model(layer):
...
 
 
-def _norm_to_list_of_layers(maybe_layers):
-  """Normalizes to a list of layers.
+def _get_inbound_nodes(tensor_key_with_args_array):
+  """Return normalized list of TensorKeyWithArgs.
+
+  Keras model serialization records input to a layer in a property,
+  `tensor_key_with_args_array`, and it can be of below forms:
+  - [['name_1', 1, 0], ['name_2', 1, 1]]: two invocations witn single input.
```
wchargin: sp.: *witn* → *with*

tensorboard/plugins/graph/keras_util.py
```diff
@@ -92,89 +121,171 @@ def _is_model(layer):
...
+  - [
+      [['name_1', 0, 0], ['name_2', 1, 0]],
+      [['name_3', 0, 0], ['name_4', 1, 0]]]: two invocations with multiple
+    inputs. Multiple inputs are only applicable to limited keras layers like
+    keras.layers.add, keras.layers.concat, and etc...
+
+  Note that there cannot be fixture of types as a layer that takes multiple
```
wchargin: What is a “fixture of types”, and what does it have to do with the size
of the input? Like, the input data shape and batch size? I don’t
understand this proviso.

tensorboard/plugins/graph/keras_util.py
```diff
@@ -92,89 +121,171 @@ def _is_model(layer):
...
+    tensor_key_array: Value from 'input_layers' or 'output_layers' from Keras
+        model configuraiton.
+
+  Returns:
+    A list of TensorKey.
   """
-  return (maybe_layers if isinstance(maybe_layers[0], (list,))
-          else [maybe_layers])
+  normalized_key_array_list = ([tensor_key_array] if isinstance(
+      tensor_key_array[0], six.string_types) else tensor_key_array)
```
wchargin: On empty input, we want to return `[]` rather than crashing, right?

tensorboard/plugins/graph/keras_util.py
```diff
@@ -92,89 +121,171 @@ def _is_model(layer):
...
+
+  A model configuration has fields 'input_layers' and 'output_layers' which can
+  be of below forms:
   - ['in_layer_name', 0, 0]
   - [['in_layer_is_model', 1, 0], ['in_layer_is_model', 1, 1]]
-  The data inside the list seems to describe [name, size, index].
+
+  Args:
+    tensor_key_array: Value from 'input_layers' or 'output_layers' from Keras
+        model configuraiton.
```
wchargin: sp.: *configuraiton* → *configuration*

tensorboard/plugins/graph/keras_util.py
```diff
@@ -92,89 +121,171 @@ def _is_model(layer):
...
+              node_args=input_layer[3]))
+    normalized_key_array_list.append(inbound_invocation)
+  return normalized_key_array_list
+
+
+def _get_tensor_key(tensor_key_array):
+  """Return normalized list of TensorKeyArray and returns list of TensorKeys.
+
+  A model configuration has fields 'input_layers' and 'output_layers' which can
+  be of below forms:
```
wchargin: If I understand correctly, these two forms correspond to “just one
input” and “multiple inputs” respectively. In particular, in the first
case, the length is a constant (3), where in the second case the length
is variable but the length of each element is the same constant (3). Is
that right?

If so, then could you please note this? You’ve defined enough vocabulary
to express it succinctly:

```python
  """...

  A model configuration has fields `input_layers` and `output_layers`.
  Each is either a raw `TensorKey` or a list of raw `TensorKey`s. This
  function normalizes such a value to always be a list of `TensorKey`s,
  forming a singleton list in the case where the input is a single raw
  `TensorKey`.

  ...

  Returns:
    A list of `TensorKey` values.
  """
```

tensorboard/plugins/graph/keras_util.py
```diff
@@ -92,89 +121,171 @@ def _is_model(layer):
...
+  Returns:
+    A list of TensorKey.
   """
-  return (maybe_layers if isinstance(maybe_layers[0], (list,))
-          else [maybe_layers])
+  normalized_key_array_list = ([tensor_key_array] if isinstance(
+      tensor_key_array[0], six.string_types) else tensor_key_array)
 
+  tensor_keys = []
+  for node in normalized_key_array_list:
```
wchargin: Consider using a list comprehension here?

tensorboard/plugins/graph/keras_util.py
```diff
@@ -92,89 +121,171 @@ def _is_model(layer):
   return layer.get('config').get('layers') is not None
```
wchargin: Unrelated, but this `get`-chaining isn’t correct, because `None.get` is
an error:

```
>>> {}.get('config').get('layers')
AttributeError: 'NoneType' object has no attribute 'get'
```

Presumably you want `layer.get('config', {}).get('layers')`.

PR #2016, opened 2019-03-15
Fix with long words in projector metadata card

makseq:
Bug with long strings presentation in metadata card. 

Only two css classes were changed. 

* Screenshots of UI changes
Before fix:
image

After fix:
image

Just put long strings in metadata

### DIFF
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-metadata-card.html
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-metadata-card.html
@@ -46,6 +46,7 @@ limitations under the License.
   line-height: 24px;
   padding: 12px 12px 8px;
   width: 230px;
+  word-break: break-word;
 }
 
 #metadata-table {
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/projector/vz_projector/vz-projector-metadata-card.html
```diff
@@ -46,6 +46,7 @@
   line-height: 24px;
   padding: 12px 12px 8px;
   width: 230px;
+  word-break: break-word;
```
stephanwlee: Ah! I remember why I did not use `word-break: break-word` in the past! It does not work in Firefox!
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
According to the MDN, `overflow-wrap` should work

makseq: No problem : ) Let's check this one.

Issue #1711, opened 2018-12-20
Replace use of TF PyRecordReader


nfelt:
TensorBoard currently relies on tf.pywrap_tensorflow.PyRecordReader_New() to read tfrecords files (aka event files).  Not only is this bad because it's not strictly speaking a public API symbol, it's also being hidden in TF 2.0.

There is a reimplementation of record reading logic in pure python in our tensorflow_stub here:
https://github.com/tensorflow/tensorboard/blob/3f055ba9c057ad2fbacfbd2c61b6e27471a51396/tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py

I haven't benchmarked this, but I expect that it's considerably slower than the C++ version from TensorFlow.  If so and if we can't get the performance to be comparable to the C++ version, then when TF is available we should prefer the C++ code.  To use that code, however, we'll need to expose it under a real public python API.

What exists today is:

* [`tf.train.summary_iterator`](https://www.tensorflow.org/versions/r1.12/api_docs/python/tf/train/summary_iterator) - deprecated, will go away entirely in TF 2.0

* [`tf.TFRecordReader`](https://www.tensorflow.org/versions/r1.12/api_docs/python/tf/TFRecordReader) - deprecated queue-based reader being replaced by tf.data

* [`tf.io.tf_record_iterator`](https://www.tensorflow.org/versions/r1.12/api_docs/python/tf/io/tf_record_iterator) - deprecated in favor of tf.data.TFRecordDataset for tf.data, and not enough for our needs

* [`tf.data.TFRecordDataset`](https://www.tensorflow.org/versions/r1.12/api_docs/python/tf/data/TFRecordDataset) - tf.data-compatible Dataset, but ultimately still an iterator, and not enough for our needs

* [`tf.io.TFRecordWriter`](https://www.tensorflow.org/versions/r1.12/api_docs/python/tf/io/TFRecordWriter) - writer side (so not immediately useful,  but an example of what does still exist)

An iterator API isn't powerful enough for us because we need the ability to read to the end of a file and then wait for new data, preserving our read offset.  As far as I can tell, all the above iterator approaches will abort iteration when they reach the end of the file, so reading new data would require re-reading all the previously consumed data.  Also, I believe we currently rely on being able to handle cases where a record isn't fully flushed (e.g. the file appears to end in a corrupted record, but it's really just that the data isn't all there yet, and the appropriate response is to wait and retry), which the iterator APIs also don't allow.

[...]

PR #2024, opened 2019-03-19
Remove dead code in db.py, loader.py, and related utilities and tests

nfelt:
As I understand it, the db.py and loader.py code was introduced as an earlier iteration of what ended up becoming tf.contrib.summary.create_db_writer().  None of it is currently used [1] and there's no real plan for using it; the current minimal support for ingesting a logdir to a db via `--db_import` uses its own loading implementation which is perhaps less optimized but considerably more straightforward, and it's unclear whether we will end up wanting to optimize that path.  If we do find a need for this logic in the future, it'll be in the revision history.

In the mean time, removing it means no longer worrying about maintenance (e.g. removing the PyRecordReader reference per #1711) and means we can also remove code in utils that only existed to support those files.

[1] The one exception is the db connection wrapper, which is a proxy object and we can just replace it with the regular sqlite3 connection object.


nfelt:
```
[🥝🥝🥝🥝🥝🥝🥝🥝🥝🥝🥝☠__________________________________] 33%
2017             2019              ???            profit
```

wchargin:
Of course, as soon as you delete `FakeClock`, I end up wanting it. :-)

PR #2029, opened 2019-03-20
Minor changes to make integration tests work.

erzel:
I also fixed a bug in tf-runs-selector that caused flakiness in the integration tests:
The 'dataLocation' property wasn't being updated, which caused the scalars PageObject to fail to find
the 'data-location' div.
@stephanwlee, @wchargin please take a look.

Thanks.

### DIFF
--- a/tensorboard/components/tf_runs_selector/tf-runs-selector.html
+++ b/tensorboard/components/tf_runs_selector/tf-runs-selector.html
@@ -160,14 +160,14 @@ Properties out:
       });
       this.set('runs', tf_backend.runsStore.getRuns());
 
-      this._envStoreListener = tf_backend.runsStore.addListener(() => {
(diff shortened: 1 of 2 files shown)


tensorboard/components/tf_runs_selector/tf-runs-selector.html
```diff
@@ -160,14 +160,14 @@ <h3 id="tooltip-help" class="tooltip-container">Runs</h3>
       });
       this.set('runs', tf_backend.runsStore.getRuns());
 
-      this._envStoreListener = tf_backend.runsStore.addListener(() => {
+      this._envStoreListener = tf_backend.environmentStore.addListener(() => {
```
stephanwlee: Oh wow. Thanks for fixing this. :\ I must have caused the regression.

PR #2031, opened 2019-03-20
docs: fix list rendering

wchargin:
Summary:
The Google-internal Markdown processor, which renders our docs site,
requires that lists be preceded by a blank line. This is [nonstandard],
so the previews shown in GitHub and even Colab itself look fine, but the
docs site does not:

![Example of a botched ordered list][ex-ol]

![Example of a botched unordered list][ex-ul]

This patch fixes up our notebooks to accommodate to the stricter style.

Test Plan:
Regression test added. Before this change, running the new test outputs

```
docs/r2/graphs.ipynb:413:        "*   Define and annotate a function with `tf.function`\n",
docs/r2/graphs.ipynb:416:        "*   With a Summary file writer, call `tf.summary.trace_export()` to save the log data\n",
docs/r2/hyperparameter_tuning_with_hparams.ipynb:85:        "1. Experiment setup and HParams summary\n",
docs/r2/hyperparameter_tuning_with_hparams.ipynb:510:        "- Filter which hyperparameters/metrics are shown in the dashboard\n",
docs/r2/hyperparameter_tuning_with_hparams.ipynb:525:        "* The **Table View** lists the runs, their hyperparameters, and their metrics.\n",
```

and exits 123. After this change, it outputs nothing and exits 0.

[ex-ol]: https://user-images.githubusercontent.com/4317806/54706565-75cb9100-4afc-11e9-81a3-b91ab1f96087.png
[ex-ul]: https://user-images.githubusercontent.com/4317806/54706613-90056f00-4afc-11e9-9cb6-84f83470bf29.png
[nonstandard]: https://spec.commonmark.org/dingus/?text=No%20newline%20after%20me%0A*%20and%20yet%20a%20fine%20list%20here

wchargin-branch: docs-newline

### DIFF
--- a/tensorboard/tools/docs_list_format_test.sh
+++ b/tensorboard/tools/docs_list_format_test.sh
@@ -1,0 +1,37 @@
+#!/bin/sh
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+# Check that any Markdown lists in IPython notebooks are preceded by a
+# blank line, which is required for the Google-internal docs processor
+# to interpret them as lists.
+
+set -e
+
+if ! [ -f WORKSPACE ]; then
+    printf >&2 'fatal: no WORKSPACE file found (are you at TensorBoard root?)\n'
+    exit 2
+fi
+
+git ls-files -z '*.ipynb' |
+xargs -0 awk '
+    { is_list = /"([-*]|[0-9]+\.) / }
+    is_list && !last_list && !last_blank {
+        printf "%s:%s:%s\n", FILENAME, FNR, $0;
+        status = 1;
+    }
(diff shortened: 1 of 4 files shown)


tensorboard/tools/docs_list_format_test.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
stephanwlee: :stuck_out_tongue:  it is 2019!

tensorboard/tools/docs_list_format_test.sh
```diff
@@ -0,0 +1,37 @@
...
+set -e
+
+if ! [ -f WORKSPACE ]; then
+    printf >&2 'fatal: no WORKSPACE file found (are you at TensorBoard root?)\n'
+    exit 2
+fi
+
+git ls-files -z '*.ipynb' |
+xargs -0 awk '
+    { is_list = /"([-*]|[0-9]+\.) / }
```
stephanwlee: Maybe the trick is hidden in the `"` but I presume this cannot catch the case like below, right?
```md
   * foo\n
* bar\n
\n
```

tensorboard/tools/docs_list_format_test.sh
wchargin: So it is! Thanks.

tensorboard/tools/docs_list_format_test.sh
wchargin: Correct, it won’t catch that. This is all a big approximation that is
neither sound nor complete (grepping over JSON strings representing
Markdown text in a serialized notebook file…) so I’m not *too* worried
about it being imperfect. If it becomes problematic, we can patch the
script or ditch it completely, but I figure that it’s better than
nothing for now.

stephanwlee: :+1: iterates the best :)

PR #2032, opened 2019-03-21
Add popup dialog for plugins that require TensorFlow

### LINKED ISSUE #2027: Add clean UI for plugins requiring TensorFlow when not installed
As a part of supporting non-TensorFlow installs of TensorBoard, we need to disable UI for plugins that require TensorFlow and can't be loaded. One option is to take inspiration from the debugger plugin and show a popup with the following:

* Title: This plugin is disabled without TensorFlow
* Message: To enable this plugin in TensorBoard, install TensorFlow with <code>pip install tf-nightly</code>

Once we've done that we can also remove the requirement in https://github.com/tensorflow/tensorboard/pull/1796 that non-TF needs to be run with

```
bazel build tensorboard
TENSORBOARD_NO_TF=1 ./bazel-bin/tensorboard/tensorboard --logdir ~/local/sample_runs/runs
```
and instead will simply be runnable with

```
bazel build tensorboard
./bazel-bin/tensorboard/tensorboard --logdir ~/local/sample_runs/runs
```

orionr:
Adding a popup dialog for when plugins are not created. Note that this is different from being "inactive", which can be due to data, etc.

Fixes #2027 

Everything should be good to go and I've tested on my Mac with the following:

```
conda create --name tfpy36 python=3.6
conda activate tfpy36
pip install --upgrade pip
pip install tf-nightly-2.0-preview
conda install protobuf
conda install absl-py
conda install numpy
bazel build tensorboard
./bazel-bin/tensorboard/tensorboard --logdir ~/local/sample_runs/runs
# Navigate to the hparams tab and verify a popup is NOT shown

conda create --name notf python=2.7
conda activate notf
pip install --upgrade pip
pip install scipy
conda install protobuf
conda install absl-py
conda install numpy
bazel build tensorboard
./bazel-bin/tensorboard/tensorboard --logdir ~/local/sample_runs/runs
# Navigate to the hparams tab and verify a popup is shown
```

cc @lanpa, @nfelt

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -42,6 +42,7 @@ limitations under the License.
 <link rel="import" href="../tf-dashboard-common/tf-option-selector.html">
 <link rel="import" href="../tf-storage/tf-storage.html">
 <link rel="import" href="./tf-inference-panel.html">
+<link rel="import" href="../tf-tensorboard/plugin-dialog.html">
 <link rel="import" href="../tf-tensorboard/registry.html">
 <link rel="import" href="../vz-bar-chart/vz-bar-chart.html">
 <link rel="import" href="../vz-line-chart/vz-line-chart.html">
@@ -1181,6 +1182,7 @@ limitations under the License.
         --paper-toggle-button-label-color: #3C4043;
       }
     </style>
+    <tf-plugin-dialog id="initialDialog"></tf-plugin-dialog>
     <paper-dialog id="inferencesettings" class="inference-settings" opened="[[!local]]">
       <tf-inference-panel
           inference-address="{{inferenceAddress}}"
@@ -2133,7 +2135,7 @@ limitations under the License.
       properties: {
 
         // TensorBoard plugin standard.
-        requestManager: {
+        _requestManager: {
           type: Object,
           value: () => new tf_backend.RequestManager(),
         },
@@ -2553,7 +2555,23 @@ limitations under the License.
           self.$.dive.$.vis._onIronResize();
         });
 
-        resizer.call(dragResize);
+        // Check if the plugin was created
+        this._requestManager
+            .request(tf_backend.getRouter().pluginsListing())
+            .then(this._canceller.cancellable(result => {
+              if (result.cancelled) {
+                return;
+              }
+              if (!('whatif' in result.value)) {
+                // The plugin was not created
+                this.$.inferencesettings.close();
+                this.$.initialDialog.openNoTensorFlowDialog();
+              } else {
+                // The plugin was created
+                this.$.initialDialog.closeDialog();
+                resizer.call(dragResize);
+              }
+            }));
       },
 
       settingsClicked_: function() {
@@ -4447,7 +4465,7 @@ limitations under the License.
             thenDoFn(result);
           }
         });
-        this.requestManager.request(url, postData).then(wrapperFn)
+        this._requestManager.request(url, postData).then(wrapperFn)
             .catch(reason => {
               this.exampleStatusStr = 'Request failed';
               this.showToast_('Request failed: ' + reason);
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -2571,6 +2571,9 @@ limitations under the License.
                 // The plugin was not created
                 this.$.inferencesettings.close();
                 this.$.initialDialog.openNoTensorFlowDialog();
+              } else {
+                // The plugin was created
+                this.$.initialDialog.closeDialog();
               }
             });
       },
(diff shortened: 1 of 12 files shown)


orionr:
@nfelt if tests pass this should be good for review. Thanks!

orionr:
And tests pass! We're good.

orionr:
@nfelt and @wchargin, a gentle ping for review on this. Thanks.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -433,7 +439,23 @@ <h3>No profile data was found.</h3>
...
+      this._requestManager
+          .request(tf_backend.getRouter().pluginsListing())
+          .then(this._canceller.cancellable(result => {
+            if (result.cancelled) {
+              return;
+            }
+            if (!('profile' in result.value)) {
+              // The plugin was not created
+              this.$.initialDialog.openNoTensorFlowDialog();
+              this.set('_isAttached', false);
```
nfelt: Let's change this to `_isAvailable` like the other dashboards, but then we should also update _maybeInitializeDashboard which reads the value in two places:

- Observer registration up on line 437 should change to the following https://github.com/tensorflow/tensorboard/blob/678e6b90e0341c26691441a77c09223708012037/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html#L437
```javascript
_maybeInitializeDashboard(_isAttached, _isAvailable)
````

- Observer definition down on line 487 should change to the following https://github.com/tensorflow/tensorboard/blob/678e6b90e0341c26691441a77c09223708012037/tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html#L487-L492

```javascript
_maybeInitializeDashboard: function(isAttached, isAvailable) {
  if (this._initialized || !isAttached || !isAvailable) {
    // Either this dashboard is already initialized ... or we are not yet
    // ready to initialize.
    return;
  }
```

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -2133,7 +2135,7 @@ <h2>Show similarity to selected datapoint</h2>
       properties: {
 
         // TensorBoard plugin standard.
-        requestManager: {
+        _requestManager: {
```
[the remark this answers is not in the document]
orionr: I can revert if you want, but this standardizes across plugins.

orionr:
@nfelt (and @wchargin) all comments should be addressed. I needed to use a function scope to add `PLUGIN_NAME` but that seemed worth it. Also kept the `_requestManager` change, since it is consistent with the other plugins. Let me know if I should revert that particular change. Tested in a non-TensorFlow and  TensorFlow environment locally on my Mac.

orionr:
@nfelt any thoughts on the build failures? Seems unrelated to my changes, but not 100% sure.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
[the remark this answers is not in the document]
nfelt: Okay, that's fine.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -2553,7 +2559,20 @@ <h2>Show similarity to selected datapoint</h2>
           self.$.dive.$.vis._onIronResize();
         });
 
-        resizer.call(dragResize);
+        // Check if the plugin was created
```
nfelt: Hmm, I realized this is in the `ready` callback instead of `attached` like the others.  Could we instead put this in an `attached` callback as well?  That way we can avoid making network requests unless the dashboard is attached to the DOM.  (I think the `resizer.call(dragResize)` line could probably stay in `ready` as it was originally.)

cc @jameswex just FYI in case you anticipate any issues here

orionr: Unfortunately I'm not sure how to block the UI from loading / calling bad APIs without injecting it here. Can we call this good enough for now and have @jameswex or others move it with more expert knowledge?

jameswex: If this code is moved to attached, what bad APIs are called? This plugin shouldn't be making any calls in the backend until the user clicks accept in the this.$.inferencesettings dialog, which I imagine can't happen with this code, regardless of it is in atttached or ready callbacks.

orionr: Yup - I was wrong about the code flow. Fixing.

orionr:
@nfelt ready for final review. Thanks.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
nfelt: Looks good, thanks!  Perhaps though we should we still include the `else` clause like the other plugins?  The body would just be the initialDialog.closeDialog() call.

orionr:
One last else clause coming - stay tuned.

orionr:
@nfelt made that one last fix - thanks.

orionr:
@nfelt, please merge when ready. Thank you. :)

PR #2036, opened 2019-03-21
Fix `TFRecordCompressionType` test failures

### LINKED ISSUE #2035: Tests failing on master since tf-nightly-2.0-preview-20190319
`__import__("tensorflow").io.TFRecordCompressionType` was removed; this breaks `//tensorboard/plugins/interactive_inference:interactive_inference_plugin_test`.

wchargin:
Summary:
tensorflow/tensorflow#26676 removed `tf.io.TFRecordCompressionType`,
which our tests used; `tf-nightly-2.0-preview==2.0.0.dev20190319` was
the first version without this symbol.

Fixes #2035.

Test Plan:
All tests pass on `tf-nightly-2.0-preview==2.0.0.dev20190321` and also
on `tf-nightly==1.14.1.dev20190321`.

wchargin-branch: fix-tf26676-failures

### DIFF
--- a/tensorboard/plugins/interactive_inference/utils/platform_utils.py
+++ b/tensorboard/plugins/interactive_inference/utils/platform_utils.py
@@ -128,9 +128,9 @@ def example_protos_from_path(path,
 
   filenames = filepath_to_filepath_list(path)
   compression_types = [
-      tf.io.TFRecordCompressionType.NONE,
-      tf.io.TFRecordCompressionType.GZIP,
-      tf.io.TFRecordCompressionType.ZLIB,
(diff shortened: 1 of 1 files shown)


wchargin:
> Any thoughts on using `tf.compat.v1.python_io.TFRecordCompressionType`?

I had `tf.compat.v1.io.TFRecordCompressionType` originally, but replaced
that with the V2 version because my understanding is that we want to
limit our use of `compat` modules where possible. This code works with
both V1 and V2 natively, which seems like a material benefit even though
the hard-coded string literals are annoying.

PR #2037, opened 2019-03-22
Added public visibility to binary HParams targets

erzel:
Needed to allow "packaging" these public binaries from other BUILD packages.  
@wchargin please take a look.
Thanks!


tensorboard/plugins/hparams/BUILD
```diff
@@ -94,6 +94,7 @@ py_binary(
     name = "hparams_demo",
     srcs = ["hparams_demo.py"],
     srcs_version = "PY2AND3",
+    visibility = ["//visibility:public"],
```
wchargin: Nix this one and you’ll have my vote. http://cl/239833315

erzel: Done.

Issue #2017, opened 2019-03-16
Work around `fromtimestamp(0)` bug in Windows Python 3.6.0…3.6.6


wchargin:
See #1976.

wchargin:
For the record, I’ve attempted to do this a few times and given up each
time, because I can’t find a good way to implement `strftime("%s")`
portably. Using `(dt - EPOCH).total_seconds()` (the current code) is not
correct because it incorporates daylight savings time…

```
>>> fts = datetime.datetime.fromtimestamp
>>> (fts(9971999) - fts(0)).total_seconds()
9971999.0
>>> (fts(9971999 + 1) - fts(0)).total_seconds()
9975600.0
```

In Python 3, `datetime.timezone.utc` looks like it should at least
provide the foundations to solve these issues, but without adding a
`pytz` dependency we don’t have a Python 2 equivalent, and the code to
backport is not trivial (I checked).

I might just throw up my hands and replace everything in Python memory
with bare ints-since-epoch. At least then we just have `time.time()` to
get the real value and `str`/`int` for serialization—no syscalls, no OS
errors. I guess I’m just not Dutch enough for this one. \*sigh\*

stefanbschneider:
Actually, this was a quick and easy solution for me: https://github.com/tensorflow/tensorboard/issues/1976#issuecomment-471834147

```
pip install tensorboard==1.12.2
```

wchargin:
@stefanbschneider: That downgrades you to a version of TensorBoard prior
to the notebook integration features. You should be able to safely
upgrade to TensorBoard 1.14, which has both the features and the fix.

stefanbschneider:
Perfect, thanks!

Issue #1976, opened 2019-03-07
OSError: [Errno 22] Invalid argument


IamWangYunKai:
When I updated tensorboard from 1.12 to 1.13.1(because of svg export problem) and updated numpy to 1.16.2, I found my tensorboard didn's work. I reinstalled tensorflow and tensorboard but it still didn't work.
My system: Win7, python version:3.6.5(in Anaconda3).
**Error code:**
`bash
$ tensorboard --logdir=.
TensorBoard 1.13.1 at http://Myname:6006 (Press CTRL+C to quit)
Traceback (most recent call last):
  File "f:\anaconda3\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "f:\anaconda3\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "F:\Anaconda3\Scripts\tensorboard.exe\__main__.py", line 9, in <module>
  File "f:\anaconda3\lib\site-packages\tensorboard\main.py", line 57, in run_main
    app.run(tensorboard.main, flags_parser=tensorboard.configure)
  File "f:\anaconda3\lib\site-packages\absl\app.py", line 300, in run
    _run_main(main, args)
  File "f:\anaconda3\lib\site-packages\absl\app.py", line 251, in _run_main
    sys.exit(main(argv))
  File "f:\anaconda3\lib\site-packages\tensorboard\program.py", line 228, in main
    self._register_info(server)
  File "f:\anaconda3\lib\site-packages\tensorboard\program.py", line 274, in _register_info
    manager.write_info_file(info)
  File "f:\anaconda3\lib\site-packages\tensorboard\manager.py", line 269, in write_info_file
    payload = "%s\n" % _info_to_string(tensorboard_info)
  File "f:\anaconda3\lib\site-packages\tensorboard\manager.py", line 129, in _info_to_string
    for k in _TENSORBOARD_INFO_FIELDS
  File "f:\anaconda3\lib\site-packages\tensorboard\manager.py", line 129, in <dictcomp>
    for k in _TENSORBOARD_INFO_FIELDS
  File "f:\anaconda3\lib\site-packages\tensorboard\manager.py", line 51, in <lambda>
    (dt - datetime.datetime.fromtimestamp(0)).total_seconds()),
OSError: [Errno 22] Invalid argument
`
Does anyone have same issue or any idea how to solve it?

IamWangYunKai:
[...]

PR #2043, opened 2019-03-23
Update manager.py

Fieldhunter:
Fifty-one lines of code in the source code have BUG. OSError: [Errno 22] Invaild argumentation error will be reported on Windows 10 system. The modified code can run tensorboard normally.

### DIFF
--- a/tensorboard/manager.py
+++ b/tensorboard/manager.py
@@ -47,8 +47,7 @@ _FieldType = collections.namedtuple(
 _type_timestamp = _FieldType(
     serialized_type=int,  # seconds since epoch
     runtime_type=datetime.datetime,  # microseconds component ignored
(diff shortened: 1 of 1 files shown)


wchargin:
Hi @Fieldhunter! Thanks for the contribution. The bug that you are
running into is #1976, which is actually a bug in Python 3.6.0…3.6.6 and
not in TensorBoard. Please upgrade to Python 3.6.7 or a later version of
Python to resolve this issue.

(Your proposed fix does not actually work, because `%S` is a format
string for the number of seconds into the current minute, not the
current epoch time. You will note that we have tests that catch this
failure.)

Even though this is a problem in Python, we plan to work around it for
users who still have buggy Python installations. This is tracked in
#2017.

PR #2054, opened 2019-03-25
manager: switch from `datetime` to bare timestamps

### LINKED ISSUE #2017: Work around `fromtimestamp(0)` bug in Windows Python 3.6.0…3.6.6

wchargin:
Summary:
Fixes #2017. I’ve spent multiple hours on multiple weekends trying to
figure out how to get Python `datetime`s to properly (a) count seconds
since the epoch and (b) serialize and deserialize properly on all
platforms and all Python versions. I have failed. If nothing else, I can
trust in `int`s.

This is a backward-compatible change: the data format on disk is
unchanged, and the only APIs with incompatible changes were always
explicitly marked as private.

Test Plan:
In addition to running `//tensorboard:manager_e2e_test`, I built a Pip
package and exercised the full notebook workflow in Jupyter (magic
invocation plus `tensorboard.notebook` API) in Python 2 and 3.

wchargin-branch: bare-int-times

### DIFF
--- a/tensorboard/manager.py
+++ b/tensorboard/manager.py
@@ -394,7 +394,7 @@ def start(arguments, timeout=datetime.timedelta(seconds=60)):
 
   (stdout_fd, stdout_path) = tempfile.mkstemp(prefix=".tensorboard-stdout-")
   (stderr_fd, stderr_path) = tempfile.mkstemp(prefix=".tensorboard-stderr-")
-  start_time = time.time()
+  start_time_seconds = time.time()
   try:
     p = subprocess.Popen(
         ["tensorboard"] + arguments,
@@ -406,8 +406,8 @@ def start(arguments, timeout=datetime.timedelta(seconds=60)):
     os.close(stderr_fd)
 
   poll_interval_seconds = 0.5
-  end_time = start_time + timeout.total_seconds()
-  while time.time() < end_time:
+  end_time_seconds = start_time_seconds + timeout.total_seconds()
+  while time.time() < end_time_seconds:
     time.sleep(poll_interval_seconds)
     subprocess_result = p.poll()
     if subprocess_result is not None:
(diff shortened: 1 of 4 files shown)


tensorboard/manager_test.py
```diff
@@ -71,11 +71,12 @@ def test_roundtrip_serialization(self):
...
 
   def test_serialization_rejects_bad_types(self):
-    info = _make_info()._replace(start_time=1549061116)  # not a datetime
+    bad_time = datetime.datetime.fromtimestamp(1549061116)  # not an int
+    info = _make_info()._replace(start_time=bad_time)
     with six.assertRaisesRegex(
         self,
         ValueError,
-        "expected 'start_time' of type.*datetime.*, but found: 1549061116"):
+        "expected 'start_time' of type.*int.*, but found: datetime\."):
```
stephanwlee: noob question: why do you need a leading blackslash in front of the "."? It does not look like a regex or it requires any escaping.

tensorboard/manager.py
```diff
@@ -409,8 +406,8 @@ def start(arguments, timeout=datetime.timedelta(seconds=60)):
     os.close(stderr_fd)
 
   poll_interval_seconds = 0.5
-  end_time = start_time + timeout
-  while datetime.datetime.now() < end_time:
+  end_time = start_time + timeout.total_seconds()
```
stephanwlee: optional: any thoughts on putting suffix of "_s" on `end_time` to denote the unit?

tensorboard/manager_test.py
wchargin: It’s a regex: note `with six.assertRaisesRegex`, and note the
metacharacters earlier in the string (`type.*int.*`).

tensorboard/manager.py
wchargin: Sure. The unit is really “seconds since epoch”, not “seconds”, but
adding a `_seconds` suffix to these local variables probably can’t hurt.

PR #1801, opened 2019-02-03
[WIP] tensorboard-notf for pytorch (re-submit of #1799)

lanpa:
This modification let https://github.com/pytorch/pytorch/pull/16196/commits/00b8330ea75707fdc2ac9664f6693c94c7389858 use the code inside tensorboard installed by `whl` file build by bazel.

The generated event file can be rendered correctly by web server build from:
https://github.com/tensorflow/tensorboard/pull/1663

TODO:
- [x] rename tensorboardX to something else (keep this name to avoid namespace confusion)
- [ ] correct way to handle plugin protos (now copy that directly)
- [ ] unittests
- [x] fill license information to pass the test
- [x] re-enable python2 build/test

PR #2065, opened 2019-03-27
Add TensorBoard writer without TensorFlow

lanpa:
This PR addresses discussions in #1801.

- [ ] move writer files to tensorboard/writer
- [ ] correct way to handle plugin protos (use softlink)
- [ ] unittests
- [ ] rewrite the writer in case of license issue.

cc @orionr


orionr:
Can you change the title for this  PR to "Add TensorBoard writer without TensorFlow"? Thanks.

tensorboard/compat/BUILD
```diff
@@ -44,6 +44,7 @@ py_library(
     deps = [
         ":compat",
         "//tensorboard/compat/tensorflow_stub",
+        "//tensorboard/writer",
```
orionr: Let's not have this dependency here. Instead it should be in https://github.com/tensorflow/tensorboard/blob/master/tensorboard/BUILD#L45 like

```
    deps = [
        ":lazy",
        ":notebook",
        ":program",
        "//tensorboard/summary",
        "//tensorboard/writer",
    ],
```

tensorboard/compat/proto/layout.proto
```diff
@@ -0,0 +1 @@
+../../plugins/custom_scalar/layout.proto
```
orionr: @nfelt what's your thought on symlinks here?

tensorboard/writer/BUILD
```diff
@@ -0,0 +1,42 @@
...
+package(default_visibility = ["//tensorboard:internal"])
+
+licenses(["notice"])  # Apache 2.0
+
+exports_files(["LICENSE"])
+
+py_library(
+    name = "writer",
+    srcs = glob([
+        "*.py",
```
orionr: Can we do an explicit  list here? Looks like it would be:

```
    srcs = [
        "__init__.py",
        "crc32c.py",
        "event_file_writer.py",
        "record_writer.py",
    ],

File: tensorboard/writer/BUILD [hunk 4]
```diff
@@ -0,0 +1,42 @@
...
+
+
+py_test(
+    name = "event_file_writer_test",
+    size = "small",
+    srcs = ["event_file_writer_test.py"],
+    main = "event_file_writer_test.py",
+    srcs_version = "PY2AND3",
+    deps = [
+        ":event_file_writer",
```
orionr: I would first try just making this

```
    deps = [
        ":writer",
        "//tensorboard:expect_tensorflow_installed",
    ]
```

and then test to see what breaks with

```
bazel test //tensorboard/... --test_output=errors
```

tensorboard/writer/BUILD
```diff
@@ -0,0 +1,42 @@
...
+        # "//tensorboard:expect_tensorflow_installed",
+        # "//tensorboard/backend:application",
+        # "//tensorboard/backend/event_processing:event_accumulator",
+        # "//tensorboard/backend/event_processing:event_multiplexer",
+        # "//tensorboard/plugins:base_plugin",
+        # "//tensorboard/util:test_util",
+        # "@org_pocoo_werkzeug",
+        # "@org_pythonhosted_six",
+    ],  # TODO: I am not sure what to put here
+)
```
orionr: Please add a newline at the end

tensorboard/writer/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Compatibility interfaces for TensorBoard."""
+
+from __future__ import absolute_import
```
orionr: I think you can actually remove all  the contents of this `___init__.py` file - just have it be empty.

tensorboard/writer/event_file_writer.py
```diff
@@ -0,0 +1,231 @@
...
+from __future__ import print_function
+
+import os.path
+import socket
+import threading
+import time
+
+import six
+
+from ..proto import event_pb2
```
orionr: Might not want to do relative path. Instead can you do

```
from  tensorboard.compat.proto import  event_pb2
from  tensorboard.writer. record_writer import  RecordWeriter, directory_check

File: tensorboard/writer/event_file_writer.py [hunk 8]
```diff
@@ -0,0 +1,231 @@
...
+            self._py_recordio_writer.close()
+        return return_value
+
+
+class EventFileWriter(object):
+    """Writes `Event` protocol buffers to an event file.
+    The `EventFileWriter` class creates an event file in the specified directory,
+    and asynchronously writes Event protocol buffers to the file. The Event file
+    is encoded using the tfrecord format, which is similar to RecordIO.
+    @@__init__
```
orionr: Can we match the rest of TensorBoard for this docstring? I think you can actually just remove them but @nfelt might have a stronger opinion.

tensorboard/writer/record_writer.py
```diff
@@ -0,0 +1,158 @@
...
+        bucket, path = self.bucket_and_path()
+        upload_buffer = copy.copy(self.buffer)
+        upload_buffer.seek(0)
+        s3.upload_fileobj(upload_buffer, bucket, path)
+
+    def close(self):
+        self.flush()
+
+
+class S3RecordWriterFactory(object):
```
orionr: Let's remove S3 support for now and we'll add that by extending gfile in the compat layer in a later PR.

tensorboard/writer/record_writer.py
```diff
@@ -0,0 +1,158 @@
...
+    '''Open a writer for outputting event files.'''
+    try:
+        prefix = path.split(':')[0]
+        factory = REGISTERED_FACTORIES[prefix]
+        return factory.open(path)
+    except KeyError:
+        return open(path, 'wb')
+
+
+class S3RecordWriter(object):
```
orionr: Let's remove S3 support for now and we'll add that by extending gfile in the compat layer in a later PR.

tensorboard/compat/proto/BUILD
```diff
@@ -25,9 +25,12 @@ tb_proto_library(
         "event.proto",
         "function.proto",
         "graph.proto",
+        "layout.proto",
```
orionr: I don't think these protos are required  for the writer implementation. Can we just remove those additions from this PR for now?

lanpa: layout.proto is needed by `custom_scalar` plugin

tensorboard/compat/proto/layout.proto
nfelt: I still don't see any reason why the plugin protos need to be reparented under tensorboard/compat/proto.

tensorboard/compat/proto/BUILD
nfelt: @lanpa Yes, but why is that relevant to the writer implementation?

tensorboard/writer/event_file_writer.py
nfelt: Retaining the docstrings seems useful, but yes, please follow the rest of TensorBoard in style.

tensorboard/compat/proto/BUILD
lanpa: @nfelt @orionr  Oh, I think I got your point. Do you mean these protos are not relevant to the writer and should be added in another PR?

tensorboard/writer/BUILD
lanpa: nothing breaks with the test.  ( with `deps = [":writer",]`  )
Does that mean I can remove other deps safely?

tensorboard/compat/proto/BUILD
orionr: Exactly right - thanks.

tensorboard/writer/BUILD
orionr: That should do it.

tensorboard/writer/BUILD
```diff
@@ -0,0 +1,45 @@
...
+
+py_test(
+    name = "event_file_writer_test",
+    size = "small",
+    srcs = ["event_file_writer_test.py"],
+    main = "event_file_writer_test.py",
+    srcs_version = "PY2AND3",
+    deps = [
+        ":writer",
+        # "//tensorboard:expect_tensorflow_installed",
```
orionr: Can you remove these commented-out deps?

tensorboard/writer/crc32c.py
```diff
@@ -0,0 +1,138 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import array
+
+
+CRC_TABLE = (
```
orionr: Given that this is only used internally, I think TensorBoard convention is to start with an underscore like `_CRC_TABLE`. Can you change? Thanks.

tensorboard/writer/crc32c.py
```diff
@@ -0,0 +1,138 @@
...
+    0x69e9f0d5, 0x9b8273d6, 0x88d28022, 0x7ab90321,
+    0xae7367ca, 0x5c18e4c9, 0x4f48173d, 0xbd23943e,
+    0xf36e6f75, 0x0105ec76, 0x12551f82, 0xe03e9c81,
+    0x34f4f86a, 0xc69f7b69, 0xd5cf889d, 0x27a40b9e,
+    0x79b737ba, 0x8bdcb4b9, 0x988c474d, 0x6ae7c44e,
+    0xbe2da0a5, 0x4c4623a6, 0x5f16d052, 0xad7d5351,
+)
+
+
+CRC_INIT = 0
```
orionr: Same for this - `_CRC_INIT`

tensorboard/writer/event_file_writer.py
```diff
@@ -0,0 +1,228 @@
...
+        """Creates an _EventLoggerThread.
+        Args:
+          queue: A Queue from which to dequeue events.
+          ev_writer: An event writer. Used to log brain events for
+           the visualizer.
+          flush_secs: How often, in seconds, to flush the
+            pending file to disk.
+        """
+        threading.Thread.__init__(self)
+        self.daemon = True
```
orionr: Nit - can we call this `self._daemon` to be consistent?

tensorboard/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,204 @@
...
+
+# tf.compat.v1.disable_v2_behavior()
+
+import tensorflow as tf
+
+class HistogramsPluginTest(tf.test.TestCase):
+  def set_up_with_runs(self, run_names):
+    return
+
+# class HistogramsPluginTest(tf.test.TestCase):
```
orionr: I think we need to get some version of these tests back before we land. :-/

tensorboard/writer/record_writer.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class RecordWriter(object):
+    """write encoded protobuf along with its checksum.
+
+    """
+    def __init__(self, path):
+        self._name_to_tf_name = {}
+        self._tf_names = set()
+        self.path = path
```
orionr: Nit - can we make this `self._path`?

tensorboard/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,204 @@
...
+# from tensorboard.backend.event_processing import plugin_event_accumulator as event_accumulator  # pylint: disable=line-too-long
+# from tensorboard.backend.event_processing import plugin_event_multiplexer as event_multiplexer  # pylint: disable=line-too-long
+# from tensorboard.plugins import base_plugin
+# from tensorboard.plugins.histogram import histograms_plugin
+# from tensorboard.plugins.histogram import summary
+# from tensorboard.util import test_util
+
+# tf.compat.v1.disable_v2_behavior()
+
+import tensorflow as tf
```
orionr: If you keep this make sure the test deps in BUILD includes `"//tensorboard:expect_tensorflow_installed",`

tensorboard/writer/event_file_writer.py
lanpa: `daemon` is defined in the parent class: `threading.Thread`, so I think we can't.

tensorboard/writer/record_writer.py
```diff
@@ -0,0 +1,73 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""
+To write tf_record into file. Here we use it for tensorboard's event writting.
```
orionr: I think this comment needs to be removed and the code rewritten.

tensorboard/writer/record_writer.py
```diff
@@ -0,0 +1,52 @@
...
+            raise OSError('file writer is missing')
+
+    def close(self):
+        if self._writer is not None:
+            self._writer.close()
+        else:
+            raise OSError('file writer is missing')
+
+
+def masked_crc32c(data):
```
orionr: And you can remove masked_crc32c and  u32 here.

tensorboard/writer/record_writer.py
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import struct
+from tensorboard.compat.tensorflow_stub.pywrap_tensorflow import crc32c
```
orionr: Change this to

```
from tensorboard.compat.tensorflow_stub.pywrap_tensorflow import masked_crc32c
```

orionr:
@nfelt and @wchargin this should be ready for your review. Thanks.

tensorboard/writer/event_file_writer.py
```diff
@@ -0,0 +1,229 @@
...
+            the filename of the event file.
+          filename_suffix: The string that will be appended to
+            the filename of the event file.
+        '''
+        self._file_name = file_prefix + ".out.tfevents." + str(time.time())[:10] + "." +\
+            socket.gethostname() + filename_suffix
+        self._num_outstanding_events = 0
+        self._py_recordio_writer = RecordWriter(self._file_name)
+        # Initialize an event instance.
+        self._event = event_pb2.Event()
```
nfelt: We should also set the `file_version` field to `"brain.Event:2"` to match this line:
https://github.com/tensorflow/tensorflow/blob/3109adb52056659e1756002771e3593975804f15/tensorflow/core/util/events_writer.cc#L91

tensorboard/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,55 @@
...
+# limitations under the License.
+# ==============================================================================
+
+# """Integration tests for the Writer."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+# tf.compat.v1.disable_v2_behavior()
```
nfelt: This can be removed

tensorboard/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+import glob
+import os
+import tensorflow as tf
```
nfelt: Rather than depending on actual tensorflow, for tf.test.TestCase we can use the test case @orionr was adding in the other PR once that's ready, and instead of `tf.train.summary_iterator()` we could just use the `PyRecordReader_New` implementation from the TF stub.

That way this test can still run in no-TF mode.

tensorboard/writer/record_writer.py
```diff
@@ -0,0 +1,51 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import os
+import struct
+from tensorboard.compat.tensorflow_stub.pywrap_tensorflow import masked_crc32c
+
+
+def directory_check(path):
```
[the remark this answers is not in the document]
lanpa: It is only called once for each new `summarywriter` instance created. So I make it inlined now for more simplicity.

tensorboard/writer/event_file_writer.py
lanpa: Yup, write this information by event_file_writer makes more sense. (I was put it in the pytorch side)

tensorboard/writer/event_file_writer.py
```diff
@@ -0,0 +1,229 @@
...
+import threading
+import time
+
+import six
+
+from tensorboard.compat.proto import event_pb2
+from tensorboard.writer.record_writer import RecordWriter, directory_check
+
+
+class EventsWriter(object):
```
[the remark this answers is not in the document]
lanpa: I agree. BTW, we have some issues regarding the python threading library (which seldom happens). It looks like the only solution is changing the underlying writing procedure from `threading` to `multiprocess`. Considering the trade-offs, do you recommend switching to multiprecessing? ref. https://github.com/lanpa/tensorboardX/issues/123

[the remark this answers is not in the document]
nfelt: Hmm, it sounds like this issue happens only when using a single `Writer` from multiple processes at once, which this doesn't work because `fork()` is broken for multi-threaded code (see e.g. https://thorstenball.com/blog/2014/10/13/why-threads-cant-fork/).

I don't think we really want to support that use case.  If somebody is using multiprocessing to parallelize training, they should use a separate Writer instance in each process.  Otherwise they will have far more problems than just deadlocks as described in the issue - for example, the same Writer across multiple forked processes is presumably using the same file descriptor, but writes to it aren't synchronized (because locks don't have any effect for multiprocess code) so it would be possible to end up with conflicting or overlapping writes to the same file, resulting in data corruption.

tensorboard/writer/event_file_writer_test.py
orionr: Work in progress on that new TestCase class over in https://github.com/tensorflow/tensorboard/pull/2089

orionr: And this has landed, so feel free to do 

```
from tensorboard import test as tb_test
```

tensorboard/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,179 @@
...
+import time
+from tensorboard.writer.event_file_writer import EventFileWriter
+from tensorboard.writer.event_file_writer import _AsyncWriter
+from tensorboard.writer.event_file_writer import _AsyncWriterThread
+from tensorboard.compat.proto import event_pb2, summary_pb2
+from tensorboard.compat.proto.summary_pb2 import Summary
+from google.protobuf import json_format
+
+
+class EventFileWriterTest(tf.test.TestCase):
```
orionr: This becomes
```
class EventFileWriterTest(tb_test.TestCase):
```

tensorboard/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def get_copy_by_OS(oldfilename):
+  newfilename = tempfile.NamedTemporaryFile().name
+  shutil.copy(oldfilename, newfilename)
+  return newfilename
+
+
+if __name__ == '__main__':
+  tf.test.main()
```
orionr: This becomes

```
  tb_test.main()
```

tensorboard/writer/BUILD
```diff
@@ -0,0 +1,48 @@
...
+    srcs = ["event_file_writer_test.py"],
+    main = "event_file_writer_test.py",
+    srcs_version = "PY2AND3",
+    deps = [
+        ":writer",
+        "//tensorboard:expect_tensorflow_installed",
+    ],
+)
+
+py_test(
```
orionr: Change this from

```
py_test(
    name = "record_writer_test",
    size = "small",
    srcs = ["record_writer_test.py"],
    main = "record_writer_test.py",
    srcs_version = "PY2AND3",
    deps = [
        ":writer",
        "//tensorboard:expect_tensorflow_installed",
    ],
) 
```
to
```
py_test(
    name = "record_writer_test",
    size = "small",
    srcs = ["record_writer_test.py"],
    main = "record_writer_test.py",
    srcs_version = "PY2AND3",
    tags = ["support_notf"],
    deps = [
        ":writer",
        "//tensorboard:test",
    ],
) 
```

tensorboard/writer/event_file_writer.py
```diff
@@ -0,0 +1,229 @@
...
+        # The first event will be flushed immediately.
+        self._next_event_flush_time = 0
+        self._has_pending_events = False
+        self._shutdown_signal = object()
+
+    def stop(self):
+        self._queue.put(self._shutdown_signal)
+        self.join()
+
+    def run(self):
```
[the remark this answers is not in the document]
lanpa: @nfelt I would like to have a benchmark comparison on this. Can we make this a follow PR?

orionr:
@nfelt, confirmed with @lanpa that this is ready for review! The optimization changes can hopefully be done as a followup, but let us know.

orionr:
@nfelt and @wchargin, a gentle ping for review on this. Thanks.

tensorboard/writer/event_file_writer.py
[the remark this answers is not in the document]
nfelt: Fine to do the optimization in a follow-up, although some of the suggested changes were also to make it more testable (see other comments).

tensorboard/summary/writer/event_file_writer.py
```diff
@@ -0,0 +1,215 @@
...
+        need the summary writer anymore.
+        """
+        self._async_writer.flush()
+        self._async_writer.close()
+
+
+class _AsyncWriter(object):  # _AsyncWriter
+    '''Writes bytes to an event file.'''
+
+    def __init__(self, writerinstance, max_queue_size=20, flush_secs=120, dummy_delay=False):
```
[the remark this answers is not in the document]
lanpa: It is used to simulate the testing for flush()

tensorboard/summary/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,182 @@
...
+    w.write(random_bytes)
+    w.close()
+
+    w.write(random_bytes)
+    # nothing is written to the file after close
+    with open(logfile, 'rb') as f:
+      assert f.read() == random_bytes
+
+
+def get_copy_by_OS(oldfilename):
```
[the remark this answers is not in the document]
lanpa: It is used by `test_async_writer_auto_flushing`. It copies the source file (which is still opened by async writer) to the destination as a snapshot. Then the following code will check that if the auto-flushed content (snapshot) is long enough with tolerance.

tensorboard/summary/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,182 @@
...
+from tensorboard.summary.writer.event_file_writer import EventFileWriter
+from tensorboard.summary.writer.event_file_writer import _AsyncWriter
+from tensorboard.summary.writer.event_file_writer import _AsyncWriterThread
+from tensorboard.compat.proto import event_pb2, summary_pb2
+from tensorboard.compat.proto.summary_pb2 import Summary
+from tensorboard.compat.tensorflow_stub.pywrap_tensorflow import PyRecordReader_New
+from tensorboard import test as tb_test
+
+class EventFileWriterTest(tb_test.TestCase):
+  def __init__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
lanpa: I was facing `__init__() takes 1 positional argument but 2 were given` and I fixed with https://stackoverflow.com/questions/17353213/init-for-unittest-testcase . Both unittest.testcase and tb_test.TestCase has same error.

tensorboard/summary/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,182 @@
...
+    w = _AsyncWriter(open(logfile, 'wb'), max_queue_size=1, dummy_delay=True)
+    random_bytes = bytearray(os.urandom(64))
+    repeat = 10  # faster
+    for i in range(repeat):
+      w.write(random_bytes)
+    w.close()
+    with open(logfile, 'rb') as f:
+      assert f.read() == random_bytes * repeat
+
+  # write    ...................................
```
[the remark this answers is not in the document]
lanpa: Just an illustration. dot, exponent, dash means write, flush, idle respectively.

tensorboard/summary/writer/record_writer.py
```diff
@@ -13,39 +13,37 @@
...
-    if not os.path.exists(path):
-        os.makedirs(path)
-
 class RecordWriter(object):
+    """Write encoded protobuf to a file with packing defined in tensorflow"""
     def __init__(self, logfile):
+        """Open a file to keep the tensorboard records.
+
+        Args:
+        logfile: (string) The location where the file will be opened.
```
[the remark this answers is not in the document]
lanpa: So that we may pass a writer instance from different writer class as long as it provides `write`, `flush` and `close` in the future. Am I right?

tensorboard/summary/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,182 @@
...
+      r = PyRecordReader_New(event_files[0])
+      r.read()
+      events = r.event_strs
+      assert len(events) == 1
+      s = event_pb2.Event()
+      s.ParseFromString(events[0])
+      assert s.file_version == "brain.Event:2"
+
+  def test_async_writer_write_once(self):
+    logfile = tempfile.NamedTemporaryFile().name
```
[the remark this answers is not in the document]
lanpa: Can `six.BytesIO` be closed and reopened again? Or we just omit the test of `w.close()`?

tensorboard/summary/writer/event_file_writer_test.py
```diff
@@ -0,0 +1,182 @@
...
+
+  # write    ...................................
+  # flush    ---------^---------^---------^
+  # numelinQ 12345678901234567890  (expected, because the IO overhead)
+  # Make strict comparion for the flushing result is possible, but it requires accessing
+  # the queue inside the async writer. So I write the test to simulate real write and flush.
+  # In my experiment, the tolerance can be set as high to roughly to 0.95.
+  # I set 0.9 here in case the CI is too slow.
+
+  def test_async_writer_auto_flushing(self):
```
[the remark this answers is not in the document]
lanpa: Yes, we should made this a follow up PR.

lanpa:
@nfelt Thank you for the precious comments. Now I have more sense about how to write unittests. (tests should be deterministic) Please have a look at those unresolved conversations for my questions. ps. The CI fails on tf-nightly only starting from bcbb1ac, It should be fixable by `Merge remote-tracking branch 'upstream/master' into no-tf-sep-writer` again before PR landing. Btw, since many `asyncThreadTest` related tests will be follow PR, I mark conversation resolved. 
cc @orionr

tensorboard/summary/writer/event_file_writer.py
```diff
@@ -0,0 +1,226 @@
...
+    def close(self):
+        """Performs a final flush of the event file to disk, stops the
+        write/flush worker and closes the file. Call this method when you do not
+        need the summary writer anymore.
+        """
+        self._async_writer.close()
+
+
+class _AsyncWriter(object):
+    '''Writes bytes to an file.'''
```
orionr: Nit: `Writes bytes to an file.` -> `Writes bytes to a file.`

tensorboard/summary/writer/event_file_writer.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def __init__(self, record_writer, max_queue_size=20, flush_secs=120):
+        """Writes bytes to an file asynchronously.
```
orionr: Nit: `Writes bytes to an file asynchronously.` -> `Writes bytes to a file asynchronously.`

tensorboard/summary/writer/event_file_writer_test.py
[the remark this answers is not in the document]
orionr: I think @nfelt means you can just remove this `def __init__(self, *args, **kwargs):` method completely. Give that a try and see how it goes.

tensorboard/summary/writer/event_file_writer_test.py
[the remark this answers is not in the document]
orionr: Given that there is only one use, can  we just inline this code? We can then remove it later.

tensorboard/summary/writer/record_writer_test.py
```diff
@@ -0,0 +1,71 @@
...
+import tempfile
+from tensorboard.summary.writer.record_writer import RecordWriter
+from tensorboard.compat.proto import event_pb2, summary_pb2
+from tensorboard.compat.proto.summary_pb2 import Summary
+from tensorboard.compat.tensorflow_stub.pywrap_tensorflow import PyRecordReader_New
+from tensorboard import test as tb_test
+
+class RecordWriterTest(tb_test.TestCase):
+  def __init__(self, *args, **kwargs):
+    super(RecordWriterTest, self).__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
orionr: Yup - please just remove this method.

tensorboard/summary/writer/event_file_writer_test.py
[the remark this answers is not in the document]
nfelt: Okay, so can we remove the more complex flushing tests from this PR and add them back in the follow-up?

That would be `test_async_writer_auto_flushing` (along with the comment/diagram) and `test_async_writer_flush_before_flush_secs` now that the latter has gotten more complicated.

tensorboard/summary/writer/event_file_writer_test.py
```diff
@@ -49,127 +49,138 @@ def test_event_file_writer_roundtrip(self):
...
+
+class AsyncWriterTest(tb_test.TestCase):
+  def __init__(self, *args, **kwargs):
+    super(AsyncWriterTest, self).__init__(*args, **kwargs)
 
   def test_async_writer_without_write(self):
     for i in range(100):
-      logdir = tempfile.mkdtemp()
+      logdir = self.get_temp_dir()
       w = EventFileWriter(logdir)
```
nfelt: This test seems misplaced - it's using `EventFileWriter` so it should go in the other TestCase above.  Also I'm not sure wrapping it in the `for i in range(100)` adds anything - can we just remove the repetition?

It might also make sense to just add the lines

```python
s = event_pb2.Event.FromString(r.record())
self.assertEqual(s.file_version, "brain.Event:2")
```
to the existing `test_event_file_writer_roundtrip` above, after the first `r.GetNext()`.

orionr:
@nfelt, should be ready for review. Thanks @lanpa!

orionr:
Also heads up that this landing will be critical for our 1.1 release, which we'll patch ASAP!

orionr:
@nfelt, please merge when ready. Thank you. :)

PR #2068, opened 2019-03-27
graph: sync parser config to proto counterpart

### LINKED ISSUE #2062: Eager graph fails to display (Failed Normalizing names)
Code to reproduce:

```python
from __future__ import absolute_import


import shutil

import tensorflow as tf


if tf.version.VERSION.startswith('1.'):
  tf.compat.v1.enable_eager_execution()
  tf = tf.compat.v2


LOGDIR = './logs'


def main():
  shutil.rmtree(LOGDIR, ignore_errors=True)
  model = tf.keras.models.Sequential([
    tf.keras.layers.Dense(1, input_shape=(1,)),
  ])
  model.compile(loss='mse', optimizer='adagrad')
  callback = tf.keras.callbacks.TensorBoard(LOGDIR)
  model.fit(x=[[0]], y=[[0]], callbacks=[callback])


if __name__ == '__main__':
  main()
```

Expected behavior: Running this and then `tensorboard --logdir ./logs`
provides a graph visualizer that shows the Keras implementation graph.

Actual behavior: On versions of TensorFlow from 2019-03-23 onward, the
“Default” graph in the graph visualizer gives the error, “Graph: Failed
Normalizing names”.

Screenshot of error

The stack trace in the JS console is:

```
(index):10466 TypeError: d.list.shape.map is not a function
    at d ((index):10270)
    at new <anonymous> ((index):10276)
    at d ((index):10286)
    at (index):10285
    at arrayEach ((index):80)
    at Function.forEach ((index):242)
    at a ((index):10285)
    at arrayEach ((index):80)
    at Function.forEach ((index):242)
    at (index):10286
```

(On all versions, the `batch_1` graph gives the error, [“Error: The
graph is empty. \[…\]”][1], and the Keras conceptual graph works fine.)

The problem appears to be due to a change in the emitted graphs, not a
change in TensorBoard; the above-described behavior occurs independent
of the TensorBoard version.

[1]: https://github.com/tensorflow/tensorboard/blob/6de38d858b4ec19ed5e4088e863a8eba737292e1/tensorboard/plugins/graph/tf_graph_loader/tf-graph-loader.ts#L193-L197

The behavior is the same under tf-nightly and tf-nightly-2.0-preview
(but removing the `enable_eager_execution()` call causes the code to
work fine in tf-nightly).

cc @stephanwlee; not sure whether this is a TensorBoard error or a
TensorFlow problem.

stephanwlee:
A repeated field in Graph proto definition requires an entry in the
parser.ts to parse the text proto correctly. It was not only missing
entries for FunctionDefLibrary but also missed ones for few properties.

Do note that we did not encounter issues because we do not make use of
most of the properties.

Fixes #2062

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_common/parser.ts
+++ b/tensorboard/plugins/graph/tf_graph_common/parser.ts
@@ -156,23 +156,45 @@ export function streamParse(
  * to be repeated. This list is used in parsing time to convert repeated
  * attributes into arrays even when the attribute only shows up once in the
  * object.
+ * Repeated fields have to be in sync with graph.proto and all of its
+ * dependencies.
+ * See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/graph.proto
  */
 const GRAPH_REPEATED_FIELDS: {[attrPath: string]: boolean} = {
   'library.function': true,
   'library.function.node_def': true,
+  'library.function.node_def.input': true,
+  'library.function.node_def.attr': true,
+  'library.function.node_def.attr.value.list.b': true,
+  'library.function.node_def.attr.value.list.f': true,
+  'library.function.node_def.attr.value.list.func': true,
+  'library.function.node_def.attr.value.list.i': true,
+  'library.function.node_def.attr.value.list.s': true,
+  'library.function.node_def.attr.value.list.shape': true,
+  'library.function.node_def.attr.value.list.shape.dim': true,
+  'library.function.node_def.attr.value.list.tensor': true,
+  'library.function.node_def.attr.value.list.type': true,
+  'library.function.node_def.attr.value.shape.dim': true,
+  'library.function.node_def.attr.value.tensor.string_val': true,
+  'library.function.node_def.attr.value.tensor.tensor_shape.dim': true,
   'library.function.signature.input_arg': true,
   'library.function.signature.output_arg': true,
   'library.versions': true,
   'node': true,
   'node.input': true,
   'node.attr': true,
+  'node.attr.value.list.b': true,
+  'node.attr.value.list.f': true,
+  'node.attr.value.list.func': true,
+  'node.attr.value.list.i': true,
+  'node.attr.value.list.s': true,
+  'node.attr.value.list.shape': true,
+  'node.attr.value.list.shape.dim': true,
+  'node.attr.value.list.tensor': true,
   'node.attr.value.list.type': true,
   'node.attr.value.shape.dim': true,
   'node.attr.value.tensor.string_val': true,
   'node.attr.value.tensor.tensor_shape.dim': true,
-  'node.attr.value.list.shape': true,
-  'node.attr.value.list.shape.dim': true,
-  'node.attr.value.list.s': true
 };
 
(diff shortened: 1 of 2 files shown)


wchargin:
Thanks! Did you somehow generate these automatically or was it manual?

If we felt so inclined, we could fairly easily write a Python script to
crawl the proto DAG and identify all the repeated fields. We could then
write a test that ensures that they’re exactly in sync.

Quick proof-of-concept:

```py
# Prints out repeated fields for messages reachable from GraphDef.
from __future__ import absolute_import

import tensorflow as tf

from google.protobuf import descriptor


def visit(cache, message_descriptor):
  """Add `message_descriptor`'s repeated fields to the cache.
  
  Args:
    cache: A dict mapping `MessageDescriptor` objects to `frozenset`s of
      strings containing the names of all repeated fields on the
      message.
    message_descriptor: A descriptor to explore.
  """
  if message_descriptor in cache:
    return
  result = set()
  subdescriptors = set()
  for field in message_descriptor.fields:
    if field.label == descriptor.FieldDescriptor.LABEL_REPEATED:
      result.add(field.name)
    subdescriptor = field.message_type
    if subdescriptor:
      subdescriptors.add(subdescriptor)
  cache[message_descriptor] = frozenset(result)
  for subdescriptor in subdescriptors:
    visit(cache, subdescriptor)


def main():
  result = {}
  visit(result, tf.GraphDef.DESCRIPTOR)
  for d in sorted(result, key=lambda x: x.name):
    if result[d]:
      print("%s: %s" % (d.name, list(result[d])))


if __name__ == "__main__":
  main()
```

```
ExperimentalDebugInfo: ['original_node_names']
FunctionDef: ['ret', 'attr', 'control_ret', 'node_def']
FunctionDefLibrary: ['gradient', 'function']
GraphDef: ['node']
ListValue: ['shape', 'i', 'func', 'b', 'tensor', 's', 'f', 'type']
NameAttrList: ['attr']
NodeDef: ['input', 'attr']
OpDef: ['attr', 'control_output', 'input_arg', 'output_arg']
TensorProto: ['uint32_val', 'resource_handle_val', 'half_val', 'int_val', 'variant_val', 'scomplex_val', 'string_val', 'int64_val', 'bool_val', 'float_val', 'dcomplex_val', 'uint64_val', 'double_val']
TensorShapeProto: ['dim']
VariantTensorDataProto: ['tensors']
VersionDef: ['bad_consumers']
```

stephanwlee:
It was hand-tweaked for now though your script in follow-up change sound nice. Do note that you need similar or AST parser for TypeScript as well :)

wchargin:
> Do note that you need similar or AST parser for TypeScript as well :)

I imagined that the Python script would write out a JSON file that would
be consumed by a TypeScript test that would check for equality, rather
than inspecting the actual TypeScript source.

wchargin:
@stephanwlee Is merging this on your radar?

stephanwlee:
Sorry, this fell out of my radar until I tried to render Keras graph today. I actually like your script and it feels like that should be part of our test suite soon. I created another issue for tracking this.

wchargin:
No worries; thanks, and thanks for opening #2094.

Issue #1998, opened 2019-03-11
Design a public Python API for the hparams plugin


wchargin:
To use the hparams dashboard, users currently have to manually construct
[hparams-specific protocol buffers][proto] and send them to file
writers (see [the tutorial notebook][nb] for an example; it takes a few
dozen lines of Python code). The protobuf bindings are not particularly
idiomatic Python, and are less than pleasant to use. We should
investigate possible simplifications to this API.

[proto]: https://github.com/tensorflow/tensorboard/blob/726c2f22921adb43ea062ce3ceb43523c55aad22/tensorboard/plugins/hparams/api.proto
[nb]: https://github.com/tensorflow/tensorboard/blob/726c2f22921adb43ea062ce3ceb43523c55aad22/docs/r2/hyperparameter_tuning_with_hparams.ipynb#L265-L300

For example, we could streamline the construction of the `ListValue`s
for the discrete domains by allowing the user to pass Python lists, and
we can also infer the data types from the types of the elements of the
domain\* (which also lets us require that the list is homogeneously
typed):

```py
def create_experiment_summary():
  api = hparams_summary  # for brevity
  hparams = [
      api.hparam("num_units", domain=api.discrete([16, 32])),
      api.hparam("dropout_rate", domain=api.interval(min=0.1, max=0.2)),
      api.hparam("optimizer", domain=api.discrete(["adam", "sgd"])),
  ]
  metrics = [
      api.metric("accuracy"),
      api.metric("xent", display_name="cross-entropy"),
  ]
  experiment = api.experiment(hparams=hparams, metrics=metrics)
  experiment.write(logdir=os.path.join("logs", "hparam_tuning"))
```

This is just a sketch, but it’s already three times shorter than the
current demo without (imho) any loss of utility.

\* It’s fine to prohibit empty domains here. If a hyperparameter has
empty domain, then the whole hyperparameter space is empty, so there can
be no runs; thus, allowing empty domains is not actually useful.

omoindrot:
Thanks for this awesome plugin, this is really a useful addition to Tensorboard.

After using it in my project, I have some comments:

1. Would it be possible to support int types? There is currently only `api_pb2.DATA_TYPE_FLOAT64`, I think it would make sense to add `api_pb2.DATA_TYPE.INT64`.
It would just be a little better for the display.

[...]

PR #2076, opened 2019-04-01
Added datatype int64

shashvatshahi1998:
* I came across one comment of @omoindrot in #1998 

* I tried to add int 64 dtpe in api.proto but I am not at all sure about correct way of doing that.

* I added dtpe int64 in the the list of Datatypes in api.proto file present tensorboard/plugins/hparams

### DIFF
--- a/tensorboard/plugins/hparams/api.proto
+++ b/tensorboard/plugins/hparams/api.proto
@@ -101,6 +101,7 @@ enum DataType {
   DATA_TYPE_STRING = 1;
   DATA_TYPE_BOOL = 2;
   DATA_TYPE_FLOAT64 = 3;
+  DATA_TYPE_INT64 = 4;
 }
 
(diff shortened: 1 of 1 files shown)


shashvatshahi1998:
@stephanwlee can you please guide me, what else can I do for adding a float dtype?

stephanwlee:
hparams/api.proto is pretty new and I am not aware of the design choice. However, what I do know is that @wchargin is designing and actively working to make the API for defining hparams better _without_ using proto as an API. I do not think improving the api.proto at this time is the right thing and I would hold off from it.

wchargin:
Indeed. Adding int-type support to hparams is a cross-cutting effort,
and involves more than simply changing this proto definition. It also
interacts with the ongoing design work in #1998, so I think that it’d be
best if we hold off from making such a change until we know where we
want to go with that API.

PR #2077, opened 2019-04-01
Add GPU version of witwidget pip package

### LINKED ISSUE #2070: Witwidget problem with GPU
Hi, 

I am trying to apply the What If Tool on my model on Jupyter Notebook with the library wit widget.
However, it seems that I cannot use tensorflow-gpu with wit widget: this library automatically install another tensorflow which does not recognize any gpu.

Is there a way to use the What If Tool with tensorflow-gpu?

Thank you!

jameswex:

Fixes #2070 

For users of tensorflow-gpu, add a witwidget-gpu pip package that depends on GPU versions of TF and TF Serving API.

Followed the same building and pip package naming strategies as the GPU versions of TF and TF Serving API, which also have their own GPU-specific versions of pip packages, with the suffix "-gpu".

Updates the setup.py for witwidget in the same style as tensorflow-serving-api's setup.py, to build both cpu and gpu versions. The package_name parsing and prereq-setting logic is taken directly from that setup.py.

Update the build pip package script to build both versions.

Updated WIT readme to document the -gpu version of the package.

> pip install witwidget-gpu
Saw that it also installed as prereqs the GPU versions of TF and TF Serving API.

Considered having witwidget pip package no longer depend at all on TF or TF Serving API packages, instead requiring them to be independently installed by the user, for their specific version. But for witwidget this isn't ideal as it isn't immediately obvious that witwidget would require a user to also install tensorflow-serving-api (or -gpu version). So this solution with the correct deps in the pip package is the better user experience and in-line with other TF-based pip packages.


tensorboard/plugins/interactive_inference/README.md
```diff
@@ -288,6 +288,13 @@ pip install witwidget
 jupyter nbextension install --py --symlink --sys-prefix witwidget
 jupyter nbextension enable --py --sys-prefix witwidget
 ```
+Note that if you use TensorFlow with GPU support (tensorflow-gpu), then you
+should instead install the GPU-compatible version of witwidget:
+```sh
+pip install widwidget-gpu
```
wchargin: Confirmed; these instructions work for me.

PR #2080, opened 2019-04-01
Don't automatically run projection on tab change

cannoneyed:
It can be really jarring to click the "UMAP" or "t-SNE" tab and have the UX freeze while the projection begins running (computing knn to start). This is particularly problematic in the case of UMAP, where it's impossible to cancel the projection once it starts running. Additionally, it's not possible to set hyperparameters before running the projection.

This PR prevents the automatic running of the projection methods when the corresponding tab is selected (except for PCA, which is automatically run on startup)

### DIFF
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-projections-panel.ts
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-projections-panel.ts
@@ -481,11 +481,7 @@ export class ProjectionsPanel extends ProjectionsPanelPolymer {
         new Projection('tsne', accessors, dimensionality, dataSet);
     this.projector.setProjection(projection);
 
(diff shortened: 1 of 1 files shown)


cannoneyed:
Hey William!

Thanks for the review here - I agree with your take on the full UX being the actual problem, I just wanted to try the simplest solution first. 

I think that it's going to be better long-term to figure out a way to make it easy to exit the knn computation / UMAP optimization from the "Busy" dialog, so that one can change the hyperparameters or switch to a different projection without having to wait or reload the page.

I'll look into making this UI and see how difficult it'll be.

Also, good idea on the knn computation idea - Obviously it depends on how fast that computation is, but if it works out I think that would be a nice improvement to the experience.

PR #2075, opened 2019-03-31
Add separate non-TensorFlow test run in Travis CI

orionr:
Continuation of work for https://github.com/tensorflow/tensorboard/issues/2027 and ultimately removing the `TENSORBOARD_NO_TF=1` environment variable.

Plugins tests aren't run yet, but this confirms that build and limited tests run without TensorFlow installed.

PR #2089, opened 2019-04-03
Enable more tests in notf case for Travis CI

orionr:
Continuation of work for #2027 and ultimately removing the TENSORBOARD_NO_TF=1 environment variable.

Pulled out of https://github.com/tensorflow/tensorboard/pull/2075 and still need to fix some Python 3 vs 2 issues.

Tested with

```
bazel test //tensorboard:manager_test --test_output=errors
bazel test //tensorboard/backend:application_test --test_output=errors
```

### DIFF
--- a/tensorboard/manager_test.py
+++ b/tensorboard/manager_test.py
@@ -23,7 +23,6 @@ import errno
 import json
 import os
 import re
-import shutil
 import tempfile
 
 import six
--- a/tensorboard/manager_test.py
+++ b/tensorboard/manager_test.py
@@ -23,10 +23,11 @@ import errno
 import json
 import os
 import re
+import shutil
 import tempfile
 
 import six
-import tensorflow as tf
+import unittest
 
 try:
   # python version >= 3.3
@@ -60,7 +61,7 @@ def _make_info(i=0):
   )
 
 
-class TensorBoardInfoTest(tf.test.TestCase):
+class TensorBoardInfoTest(unittest.TestCase):
   """Unit tests for TensorBoardInfo typechecking and serialization."""
 
   def test_roundtrip_serialization(self):
@@ -168,7 +169,7 @@ class TensorBoardInfoTest(tf.test.TestCase):
     self.assertEqual(manager.data_source_from_info(info), "db sqlite:~/bar")
 
 
-class CacheKeyTest(tf.test.TestCase):
+class CacheKeyTest(unittest.TestCase):
   """Unit tests for `manager.cache_key`."""
 
   def test_result_is_str(self):
@@ -254,16 +255,21 @@ class CacheKeyTest(tf.test.TestCase):
     self.assertEqual(with_list, with_tuple)
 
 
-class TensorBoardInfoIoTest(tf.test.TestCase):
+class TensorBoardInfoIoTest(unittest.TestCase):
   """Tests for `write_info_file`, `remove_info_file`, and `get_all`."""
 
   def setUp(self):
     super(TensorBoardInfoIoTest, self).setUp()
-    patcher = mock.patch.dict(os.environ, {"TMPDIR": self.get_temp_dir()})
-    patcher.start()
-    self.addCleanup(patcher.stop)
+    self.temp_dir = tempfile.mkdtemp()
+    self.patcher = mock.patch.dict(os.environ, {"TMPDIR": self.temp_dir})
+    self.patcher.start()
     tempfile.tempdir = None  # force `gettempdir` to reinitialize from env
     self.info_dir = manager._get_info_dir()  # ensure that directory exists
+
+  def tearDown(self):
+    self.patcher.stop()
+    shutil.rmtree(self.temp_dir)
+    tempfile.tempdir = None  # force `gettempdir` to reinitialize from env
 
   def _list_info_dir(self):
     return os.listdir(self.info_dir)
@@ -352,4 +358,4 @@ class TensorBoardInfoIoTest(tf.test.TestCase):
 
 
 if __name__ == "__main__":
-  tf.test.main()
+  unittest.main()
--- a/tensorboard/manager_test.py
+++ b/tensorboard/manager_test.py
@@ -265,9 +265,6 @@ class TensorBoardInfoIoTest(test_case.TestCase):
     tempfile.tempdir = None  # force `gettempdir` to reinitialize from env
     self.info_dir = manager._get_info_dir()  # ensure that directory exists
 
-  def tearDown(self):
-    tempfile.tempdir = None  # force `gettempdir` to reinitialize from env
-
   def _list_info_dir(self):
     return os.listdir(self.info_dir)
 
--- a/tensorboard/manager_test.py
+++ b/tensorboard/manager_test.py
@@ -330,19 +330,19 @@ class TensorBoardInfoIoTest(unittest.TestCase):
     def remove_info(i):
       with mock.patch("os.getpid", lambda: 76540 + i):
         manager.remove_info_file()
-    self.assertItemsEqual(manager.get_all(), [])
+    six.assertCountEqual(self, manager.get_all(), [])
     add_info(1)
-    self.assertItemsEqual(manager.get_all(), [_make_info(1)])
+    six.assertCountEqual(self, manager.get_all(), [_make_info(1)])
     add_info(2)
-    self.assertItemsEqual(manager.get_all(), [_make_info(1), _make_info(2)])
+    six.assertCountEqual(self, manager.get_all(), [_make_info(1), _make_info(2)])
     remove_info(1)
-    self.assertItemsEqual(manager.get_all(), [_make_info(2)])
+    six.assertCountEqual(self, manager.get_all(), [_make_info(2)])
     add_info(3)
-    self.assertItemsEqual(manager.get_all(), [_make_info(2), _make_info(3)])
+    six.assertCountEqual(self, manager.get_all(), [_make_info(2), _make_info(3)])
     remove_info(3)
-    self.assertItemsEqual(manager.get_all(), [_make_info(2)])
+    six.assertCountEqual(self, manager.get_all(), [_make_info(2)])
     remove_info(2)
-    self.assertItemsEqual(manager.get_all(), [])
+    six.assertCountEqual(self, manager.get_all(), [])
 
   def test_get_all_ignores_bad_files(self):
     with open(os.path.join(self.info_dir, "pid-1234.info"), "w") as outfile:
--- a/tensorboard/manager_test.py
+++ b/tensorboard/manager_test.py
@@ -27,7 +27,6 @@ import shutil
 import tempfile
 
 import six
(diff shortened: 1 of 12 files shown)


wchargin:
Please ping when this is ready for review.

orionr:
@wchargin ready for review - thanks.

tensorboard/manager_test.py
```diff
@@ -254,17 +255,22 @@ def test_arguments_list_vs_tuple_irrelevant(self):
...
-    patcher = mock.patch.dict(os.environ, {"TMPDIR": self.get_temp_dir()})
-    patcher.start()
-    self.addCleanup(patcher.stop)
+    self.temp_dir = tempfile.mkdtemp()
+    self.patcher = mock.patch.dict(os.environ, {"TMPDIR": self.temp_dir})
+    self.patcher.start()
     tempfile.tempdir = None  # force `gettempdir` to reinitialize from env
     self.info_dir = manager._get_info_dir()  # ensure that directory exists
 
+  def tearDown(self):
```
wchargin: Can we just add more cleanups instead of using `tearDown`? As written,
`self.temp_dir` won’t be properly cleaned up if `self.patcher.stop()`
fails for any reason.

nfelt:
Having our own `TestCase` sounds like the better approach.  We can copy what we need from `tf.test.TestCase` over (though I suspect we can get by with a simpler tmp dir creation scheme).  I deleted the old one just on the principle that it was also dead code (and the main things it was managing we still do not need), not because it's a bad idea to have one.

wchargin:
Yep, that makes sense to me, too.

@orionr: Would you like to take a stab at creating a `TestCase` with
this functionality? Given that `tensorboard.util.test_util` has a hard
(module-level) TensorFlow dependency, I’d suggest creating a separate
`tensorboard.util.test_case` module, though there are perhaps other
reasonable ways to factor this.

orionr:
Sounds good - I'll take a stab at it. @nfelt, while I have you, any chance you could look at https://github.com/tensorflow/tensorboard/pull/2065 ? :)

orionr:
@wchargin (and @nfelt) this should be ready for review now. Much appreciated.

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+    regardless of their order.
+
+    Same as assertCountEqual in Python 3 with unittest.TestCase."""
+    if hasattr(self, 'assertCountEqual'):
+        self.assertCountEqual(actual, expected, msg)
+    else:
+        super(TestCase, self).assertItemsEqual(actual, expected, msg)
+
+  def assertStartsWith(self, actual, expected_start, msg=None):
+    """Test that string first starts with string second."""
```
wchargin: These aren’t the right parameter names.

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+    if not actual.startswith(expected_start):
```
wchargin: This implementation differs from the one in `tf.test.TestCase`, in that
it uses `str`-formatting instead of `repr`-formatting for the arguments,
doesn’t include the actual/expected outputs if `msg` is provided, and
directly raises an `AssertionError` rather than delegating to `fail`.
Why these changes? (At first glance, the TF implementation seems
preferable on all counts, but maybe you have some objections?)

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+
+  return _temp_dir
+
+
+class TestCase(unittest.TestCase):
+  """TensorBoard base test class.
+  This class can lazily create a temporary directory for tests to use.
+  """
+
+  def __init__(self, method='runTest'):
```
wchargin: Should be `methodName='runTest'` for compatibility with `unittest`.

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+    super(TestCase, self).__init__(method)
+    self._method = method
```
wchargin: Why do we need to store this?

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+    self._tempdir = None
+
+  def setUp(self):
+    super(TestCase, self).setUp()
+
+  def tearDown(self):
+    super(TestCase, self).tearDown()
+
+  def assertItemsEqual(self, actual, expected, msg=None):
```
wchargin: If you want to include this method, please prefer to delegate to
`six.assertCountEqual` in the implementation rather than checking
`hasattr`.

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+
+  def assertStartsWith(self, actual, expected_start, msg=None):
+    """Test that string first starts with string second."""
+    if not actual.startswith(expected_start):
+      if msg is None:
+        msg = '{} does not start with {}'.format(actual, expected_start)
+      raise AssertionError(msg)
+
+  def get_temp_dir(self):
+    """Returns a unique temporary directory for the test to use.
```
wchargin: Missing blank line after this line.

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+    If you call this method multiple times during in a test, it will return the
+    same folder. However, across different runs the directories will be
+    different. This will ensure that across different runs tests will not be
+    able to pollute each others environment.
+    If you need multiple unique directories within a single test, you should
+    use tempfile.mkdtemp as follows:
+      tempfile.mkdtemp(dir=self.get_temp_dir()):
+    Returns:
```
wchargin: Missing blank line before this lien.

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
wchargin: nit: It’s 2019! :-)

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+      tempfile.mkdtemp(dir=self.get_temp_dir()):
+    Returns:
+      string, the path to the unique temporary directory created for this test.
+    """
+    if not self._tempdir:
+      self._tempdir = tempfile.mkdtemp(dir=get_temp_dir())
+    return self._tempdir
+
+
+def main(*args, **kwargs):
```
wchargin: Can we remove this function and just call `unittest.main()`?

imho, it seems that a test _case_ module should not need to provide a
test _running_ mechanism, and I don’t see what value this alias
provides.

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+        logger.error('Error removing %s: %s', dirname, e)
+
+    atexit.register(delete_temp_dir)
+    _temp_dir = temp_dir
+
+  return _temp_dir
+
+
+class TestCase(unittest.TestCase):
+  """TensorBoard base test class.
```
wchargin: nit: Missing blank line after this line.

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+    super(TestCase, self).setUp()
+
+  def tearDown(self):
+    super(TestCase, self).tearDown()
+
+  def assertItemsEqual(self, actual, expected, msg=None):
+    """Test that sequence first contains the same elements as second,
+    regardless of their order.
+
+    Same as assertCountEqual in Python 3 with unittest.TestCase."""
```
wchargin: nit: Newline before `"""`.

tensorboard/util/test_case.py
```diff
@@ -0,0 +1,113 @@
...
+  """TensorBoard base test class.
+  This class can lazily create a temporary directory for tests to use.
+  """
+
+  def __init__(self, method='runTest'):
+    super(TestCase, self).__init__(method)
+    self._method = method
+    self._tempdir = None
+
+  def setUp(self):
```
wchargin: Are these `setUp` and `tearDown` superfluous?

tensorboard/util/test_case.py
orionr: Surprisingly, this was the docstring style used in unittest. I'll change it, though.

tensorboard/util/test_case.py
orionr: I think I couldn't find the original  implementation. :) I'll find it.

tensorboard/util/test_case.py
orionr: I think I had some issue with delegating to six, but let me double check. That was the  original implementation.

tensorboard/util/test_case.py
orionr: Removing.

tensorboard/util/test_case.py
orionr: You're right.

tensorboard/util/test_case.py
orionr: It's primarily that callers don't don't need to also import `unittest`, but  maybe that's not enough benefit. It also matched `tf.test` semantics. Thoughts?

tensorboard/util/test_case.py
orionr: Actually, I think calling `six.assertCountEqual` caused a cycle.

orionr:
Two outstanding questions (six usage for `assertItemsEqual` and `unittest.main`), but all others should be addressed. Thanks for the review!

tensorboard/util/test_case.py
wchargin: You’ll need to pass up the `super` object:

```py
    return six.assertCountEqual(super(TestCase, self), actual, expected, msg)
```

(Tested on all the `support_notf` tests on Python 2.7.)

tensorboard/util/test_case.py
wchargin: Thanks. (Unfortunately, the Python standard library is not always a
paragon of good Python style… e.g., the `assertCamelCase` method names,
also from `unittest`.)

tensorboard/util/test_case.py
wchargin: Thinking about this a bit more: The existence of `tf.test.main` makes
sense to me because the `tf.test` module _replaces_ the `unittest`
module. In a sense, this new `test_case` module does the same thing, but
its name belies that. Perhaps, then, the right thing to do is to rename
this module to `tensorboard.test`. It doesn’t really provide “utility”
functions in the usual sense, just as TensorFlow’s `tf.test` isn’t
called `tf.test_util`.

What do you think?

(I don’t feel too strongly about this either way, so if you have a
strong preference then I’d probably be happy to accommodate it.)

orionr: I think you're right. Owning main will let us update / override that as needed to match the `tf.test` semantics. I'll change this to `tensorboard.test` - makes sense to me.

orionr:
@wchargin should be ready for one more review. Thanks.

Issue #1946, opened 2019-03-02
Chart domains should update as training progresses


GalOshri:
TensorBoard version: https://pypi.org/project/tb-nightly/1.14.0a20190301/

Repro steps: 
1. Train with `Keras.fit()` using the TensorBoard callback for 5 epochs
2. Open TensorBoard and look at the Scalars dashboard. See the loss and accuracy charts for epochs 1-5. 
3. Continue training with `Keras.fit()` using the TensorBoard callback for another 5 epochs

Expected: As long as I didn't manually change the zoom of the charts, the chart should rescale to show me the new data (of the new epochs).

Actual: The data from the continued training gets added to the chart but I can see that it goes off the chart.

Relevant code (assuming notebook environment):
```py
model.fit(x=x_train, 
          y=y_train, 
          epochs=5, 
          validation_data=(x_test, y_test), 
          callbacks=[tensorboard_callback])
```
```
%tensorboard --logdir logs
```
```py
model.fit(x=x_train, 
          y=y_train, 
          epochs=10,
          initial_epoch=5,
          validation_data=(x_test, y_test), 
          callbacks=[tensorboard_callback])
```

Image of current behavior:
![image](https://user-images.githubusercontent.com/2601900/53675567-7e8f1c80-3c4c-11e9-8969-7bdad9745ff8.png)

Note: this is not specific to `Keras.fit()`.

nfelt:
If you did manually change the zoom of the charts, would you expect it to stay fixed, or for it to scroll as the run progresses to the right?  The latter was requested in #1897.

GalOshri:
That is more ambiguous :). If I zoomed in on some data, I would not want things to suddenly shift without me being able to stop it. We could add some form of "auto-scrolling" toggle, but I would prioritize this lower than the case where I didn't change any zoom settings.

What do others think?

jheffez:
@GalOshri  suggestion for "auto-scrolling" toggle makes sense. Another option is if the "fit domain to data" button was pressed then it would be considered as "ok to auto-shift left".

pesapower:
This issue has been fixed in Tensorboard 2.1.1.

jacobaustin123:
This issue still isn't fixed for me in Tensorboard 2.2.1. I'd love to be able to watch my results live like I can in Visdom, instead of being forced to double click the chart every time I reload.

asorie:
This problem still exists in version 2.3.0. Is there any update on this?

[...]

PR #2095, opened 2019-04-05
Making session group detail reset after data change

erzel:
This is needed to solve b/111691396 as well as make the UI more deterministic which aids integration tests (without this, occaisonally and non-predictably the charts do not reset).

### DIFF
--- a/tensorboard/plugins/hparams/tf_hparams_session_group_details/tf-hparams-session-group-details.html
+++ b/tensorboard/plugins/hparams/tf_hparams_session_group_details/tf-hparams-session-group-details.html
@@ -202,6 +202,14 @@ clicks on a row in table-view or a curve in parallel-coords view.
           this._sessionGroupNameHash = tf.hparams.utils.hashOfString(
               this.sessionGroup.name);
         }
+        // Reset each scalar-card by prodding 'tag'.
+        // We do this so the card will reset its domain on the next load.
+        Polymer.dom(this.root).querySelectorAll('tf-scalar-card').forEach(
+            c => {
+              const tag = c.get("tag");
+              c.set("tag", "");
+              c.set("tag", tag);
+            });
       },
 
       _haveMetrics() {


erzel:
@wchargin please take a look. Thanks!

wchargin:
I’ll look at this today (but possibly late in the day).

wchargin:
Thanks for your patience.

tensorboard/plugins/hparams/tf_hparams_session_group_details/tf-hparams-session-group-details.html
```diff
@@ -202,6 +202,14 @@ <h3>No metrics are enabled</h3>
           this._sessionGroupNameHash = tf.hparams.utils.hashOfString(
               this.sessionGroup.name);
         }
+        // Reset each scalar-card by prodding 'tag'.
```
wchargin: You could also choose to make `_resetDomain` a public method, such that
you can just call `c.resetDomain`. But this is fine if you prefer.

erzel: Yes, that does seem more natural, right?
However, I don't think that would work here, since we need to reset the scalar-card only after it finished loading the data. Prodding the tag causes one of the implementing classes  (I don't remember which right now) to set a special flag to reset the domain once the load is completed.

erzel:
Thanks for the review @wchargin.

>Do I understand correctly that this only happens to the scalar charts
>shown in the bottom pane, not to the actual parallel coordinate axes?
>If so, then this sounds like a manifestation of #1946, correct? (And
>therefore a downside of this change is that the domain will be reset on
>load even if a user has manually zoomed the chart, which we would prefer
>to avoid.)

Yes, that only happens to the scalar charts at the bottom pane (called tf-hparams-session-group-details here). And it seems like it is a manifestation of #1946. I agree that here the graphs are more ephemeral, and it seems to me like the user would more often like to see the entire graph, rather than preserve the original zoom-setting, but I'm not really a typical user. On the other hand, if this comes up again we can revisit.

tensorboard/plugins/hparams/tf_hparams_session_group_details/tf-hparams-session-group-details.html
wchargin: Ah, got it. Thanks for the explanation.

wchargin:
Sounds good; thanks!

PR #2096, opened 2019-04-05
Update TensorBoard to import pywrap_tensorflow from tensorflow.python

annarev:

Update TensorBoard to import pywrap_tensorflow from tensorflow.python instead of tensorflow.compat.v1. We are removing pywrap_tensorflow from TF API including tf.compat.v1. Importing from tensorflow.python is a temporary solution until a better API for reading records is available for TensorBoard to use.

tensorboard/compat/__init__.py now includes a function _pywrap_tensorflow to access either actual or stub pywrap_tensorflow implementation.

bazel test //...
I verified that the only tests that fail also fail on master without my change.

### DIFF
--- a/tensorboard/compat/__init__.py
+++ b/tensorboard/compat/__init__.py
@@ -74,6 +74,7 @@ def tf2():
   raise ImportError('cannot import tensorflow 2.0 API')
 
 
+# TODO(https://github.com/tensorflow/tensorboard/issues/1711): remove this
 @_lazy.lazy_load('tensorboard.compat._pywrap_tensorflow')
 def _pywrap_tensorflow():
   """Provide pywrap_tensorflow access in TensorBoard.
--- a/tensorboard/compat/__init__.py
+++ b/tensorboard/compat/__init__.py
@@ -72,3 +72,33 @@ def tf2():
     # As a fallback, try `tensorflow.compat.v2` if it's defined.
     return tf.compat.v2
   raise ImportError('cannot import tensorflow 2.0 API')
+
+
+@_lazy.lazy_load('tensorboard.compat._pywrap_tensorflow')
+def _pywrap_tensorflow():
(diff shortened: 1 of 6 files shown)


tensorboard/plugins/projector/projector_plugin.py
```diff
@@ -423,7 +424,7 @@ def _get_reader_for_run(self, run):
     reader = None
     if config.model_checkpoint_path and _using_tf():
       try:
-        reader = tf.compat.v1.pywrap_tensorflow.NewCheckpointReader(
+        reader = _pywrap_tensorflow.NewCheckpointReader(
```
nfelt: I think we can actually just change this to `tf.compat.v1.train.NewCheckpointReader` if you don't mind.  It looks like `tf.train.NewCheckpointReader` in TF 1.x is the place this symbol was properly exposed:

https://www.tensorflow.org/api_docs/python/tf/train/NewCheckpointReader

annarev: I changed it to tf.train.load_checkpoint. It seems to be the same thing:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/checkpoint_utils.py#L43

tensorboard/compat/__init__.py
```diff
@@ -72,3 +72,33 @@ def tf2():
     # As a fallback, try `tensorflow.compat.v2` if it's defined.
     return tf.compat.v2
   raise ImportError('cannot import tensorflow 2.0 API')
+
+
+@_lazy.lazy_load('tensorboard.compat._pywrap_tensorflow')
```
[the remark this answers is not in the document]
annarev: Added comment.

tensorboard/plugins/projector/projector_plugin.py
nfelt: Even better, thank you!

PR #2101, opened 2019-04-09
Fix URL to have trailing slash when --path_prefix is used

### LINKED ISSUE #1176: --path_prefix startup URL omits trailing slash
When I execute `tensorboard --logdir storage/logs --path_prefix /test`, I get this message: ```TensorBoard 1.8.0 at http://lcswillems-ubuntu:6006/test (Press CTRL+C to quit)```. But when I click on the link, I get `Not found` page.

I precise that when I execute `tensorboard --logdir storage/logs`, it is working.

I use Tensorboard 1.8.0.

amanp592:
Fixes #1176 

@nfelt  @wchargin  please review this!!


amanp592:
@nfelt you mentioned that a test case for the same in `class WerkzeugServerTest(tf.test.TestCase):` will be written i want to ask that the test case would be for the trail slash check of something else. please tell me so that i can write the test case too.

amanp592:
@nfelt sorry was busy with my semester internals now i'll work on this and will let you know if i face any issues!

amanp592:
@nfelt i made few lines of code for testing the `path_prefix` problem it consist of three cases

1. `if the url does not ends with trailing slash` then it adds a trailing slash on its own
2. `if the url ends with two trailing slash` then it removes one trailing slash
3. `if the url has the trailing slash` no problem it passes as it is

implementing this to the program_test.py is the problem i am facing, can you guide me a littlle bit?

nfelt:
Hi @amanp592 - I did already outline how to make the test changes in my previous comment.  Basically you would copy the existing `testServerBlankHost()` test twice, rename it as appropriate, and then test two situations:

- passing in a path prefix that doesn't end in a slash, and assert that server.get_url() does end that prefix + a slash
- passing in a path prefix that does already end in a single slash, and assert that server.get_url() ends in exactly that prefix (no extra, second slash added - which your current code would do)

This is about as detailed guidance as I can give for this; if you have questions about actually writing python tests I'd recommend checking out some tutorials or looking at more existing code in the TensorBoard codebase.

amanp592:
@nfelt sir, please review this now!.

amanp592:
@nfelt now please review this!

tensorboard/program_test.py
```diff
@@ -62,6 +62,18 @@ def testMakeServerBlankHost(self):
         self._StubApplication(),
         self.make_flags(host='', port=0, path_prefix=''))
     self.assertStartsWith(server.get_url(), 'http://')
+    
+    def testPathPrefixSlash(self):
```
nfelt: This needs to be un-indented by 2 spaces

tensorboard/program_test.py
```diff
@@ -62,6 +62,18 @@ def testMakeServerBlankHost(self):
...
+    #Test that checks the path prefix ends with one trailing slash
+    server = program.WerkzeugServer(
+        self._StubApplication(),
+        self.make_flags(host='127.0.0.1', port=0, path_prefix='/test'))
```
nfelt: Rather than setting the host, let's just use `''` like the test above so this works on IPv4-only and IPv6-only stacks.

Same below.

tensorboard/program_test.py
```diff
@@ -62,6 +62,18 @@ def testMakeServerBlankHost(self):
...
+    self.assertStartsWith(server.get_url(), 'http://127.0.0.1:0/test/')
```
nfelt: Rather than assertStartsWith, we really need an `assertEndsWith` so that we can check the exact path and make sure there are no extra slashes.  And that way we can avoid asserting on the exact host and port, which aren't relevant to this test.

Can you implement an `assertEndsWith` helper in this file as a method of the TestCase class?  It should look exactly like `assertStartsWith` except that you can replace "start" with "end" everywhere:

https://github.com/tensorflow/tensorboard/blob/2aa2ba70eec793442438d66574304f64b990127b/tensorboard/test.py#L80-L85

You may need to merge the master branch into your pull request branch in order to have the `test.py` file.

tensorboard/program_test.py
amanp592: Done!

tensorboard/program_test.py
amanp592: Ohk!

tensorboard/program_test.py
amanp592: @nfelt  i updated the assertStartsWith to assertEndsWith in my code, but i lost you at merging the master branch into your pull request branch, can you help me with this??

amanp592: @nfelt i did everything else just cant get the `test.py` in my pull request branch. Please help me this!

nfelt: Here's some instructions: https://help.github.com/en/articles/merging-an-upstream-repository-into-your-fork

Basically you should:
1. check out your PR branch - https://github.com/amanp592/tensorboard/tree/patch-1 - so `git checkout patch-1` locally
2. run `git pull https://github.com/tensorflow/tensorboard.git master` and accept the merge
3. make changes to `test.py` as needed, which should now exist
4. `git push` your updated branch patch-1

amanp592:
@nfelt 
Now i updated the `test.py` also

Sorry for the dirty commits got myself stuck very bad with git!!

Now please review this

tensorboard/program_test.py
```diff
@@ -62,6 +62,18 @@ def testMakeServerBlankHost(self):
         self._StubApplication(),
         self.make_flags(host='', port=0, path_prefix=''))
     self.assertStartsWith(server.get_url(), 'http://')
+    
+  def testPathPrefixSlash(self):
+    #Test that checks the path prefix ends with one trailing slash
+    server = program.WerkzeugServer(
+        self._StubApplication(),
+        self.make_flags(host='', port=0, path_prefix='/test'))
+    self.assertEndsWith(server.get_url(), 'http://127.0.0.1:0/test/')
```
nfelt: This should just be `self.assertEndsWith(server.get_url(), '/test/')` - that way you avoid trying to match on the rest of the URL which isn't going to be `127.0.0.1:0`.

Same thing below.

amanp592: Ohk Done!

amanp592:
@nfelt Done Sir!

tensorboard/program.py
```diff
@@ -526,7 +526,7 @@ def get_url(self):
       host = self._flags.host
       display_host = (
           '[%s]' % host if ':' in host and not host.startswith('[') else host)
-    return 'http://%s:%d%s' % (display_host, self.server_port,
+    return 'http://%s:%d%s/' % (display_host, self.server_port,
```
nfelt: Okay, so now you just need to fix the fact this still appends a redundant second `/` when path_prefix contains a slash.  This is why the Travis check is failing in the second assertion in testPathPrefixSlash(), as shown:

```text
FAIL: testPathPrefixSlash (__main__.WerkzeugServerTest)
----------------------------------------------------------------------
  File "/home/travis/.bazel-output-base/sandbox/linux-sandbox/4/execroot/org_tensorflow_tensorboard/bazel-out/k8-fastbuild/bin/tensorboard/program_test.runfiles/org_tensorflow_tensorboard/tensorboard/program_test.py", line 76, in testPathPrefixSlash
    self.assertEndsWith(server.get_url(), '/test/')
  File "/home/travis/.bazel-output-base/sandbox/linux-sandbox/4/execroot/org_tensorflow_tensorboard/bazel-out/k8-fastbuild/bin/tensorboard/program_test.runfiles/org_tensorflow_tensorboard/tensorboard/test.py", line 92, in assertEndsWith
    self.fail(fail_msg)
AssertionError: 'http://travis-job-c803f8d6-a50b-4bd6-924a-5036a813e700:38810/test//' does not end with '/test/'
```

You could do this by just changing this string format call to take `self._flags.path_prefix.rstrip('/')`.

amanp592: Done! @nfelt Please See

nfelt: This didn't fix the test because rstrip() returns a new string.  Python strings are immutable so calling it on a separate line by itself does nothing.

You need to change the next line where `self._flags.path_prefix` is passed into the string formatting call.

amanp592: Did that! 
@nfelt  now look!

amanp592:
@nfelt now all checks have been passed!

i guess now it is pretty good to get merged!

amanp592:
@nfelt Thankyou fro all the help you provided! Really Appreciated!! 
Anyomore issues i should look upon then please let me know!

PR #2104, opened 2019-04-09
Support browser fetch with retry/queue 

stephanwlee:
RequestManager is a thin wrapper around XMLHttpRequest and fetch.
Unfortunately, for some convenience of plugin authors, XMLHttpRequest
version hide away certain control over the request or response. For
fetch API, we would like to keep it as true to the real browser API
while giving it feature of automatic retries and request queuing.

Motivation: it enables clients to interpret responses as array buffers,
plain text, etc., rather than only JSON. This is helpful for performance
reasons: you can send binary blobs without having to JSON-over-base64
encode them on the server.

### DIFF
--- a/tensorboard/components/tf_backend/requestManager.ts
+++ b/tensorboard/components/tf_backend/requestManager.ts
@@ -56,7 +56,7 @@ export class RequestNetworkError extends Error {
     this.url = url;
   }
 }
- 
+
 /** The HTTP method-type to use. Currently only 'GET' and 'POST' are
  * supported.
  */
@@ -64,23 +64,23 @@ export enum HttpMethodType {
   GET = 'GET',
   POST = 'POST',
 }
-  
+
 /**
  * Holds options that can be used to configure the HTTP request.
  */
 export class RequestOptions {
   public methodType: HttpMethodType;
-  
+
   /** The content-type request header to use. Cannot be set for a GET request.*/
   public contentType?: string;
 
-  /** The request body to use. This is the object that is passed to the 
+  /** The request body to use. This is the object that is passed to the
    * XMLHttpRequest.send() method. If not given the 'send' method is called
-   * without an argument. 
+   * without an argument.
    */
   public body?: any;
 
-  /** If specified, this will be the value set in the 
+  /** If specified, this will be the value set in the
    * XMLHttpRequest.withCredentials property.
    */
   public withCredentials?: boolean;
@@ -98,7 +98,7 @@ export class RequestOptions {
     // make much sense.
   }
 }
-  
+
 export class RequestManager {
   private _queue: ResolveReject[];
   private _maxRetries: number;
@@ -154,6 +154,32 @@ export class RequestManager {
           return Promise.reject(rejection);
         });
     return promise;
+  }
+
+  public fetch(url: string, fetchOptions?: object): Promise<Response> {
+    return new Promise((resolve, reject) => {
+        const resolver = {resolve: resolve, reject: reject};
+        this._queue.push(resolver);
+        this.launchRequests();
+    }).then(() => {
+      let numTries = 1;
+      return new Promise<Response>((resolve) => {
+        const retryFetch = () => {
+          fetch(url, fetchOptions).then((response) => {
+            if (!response.ok && this._maxRetries > numTries) {
+              numTries++;
+              retryFetch();
+              return;
+            }
+            resolve(response);
+            this._nActiveRequests--;
+            this.launchRequests();
+          });
+        }
+
+        retryFetch();
+      });
+    });
   }
 
   public clearQueue() {
@@ -231,7 +257,7 @@ export class RequestManager {
     });
   }
 }
-  
+
 function buildXMLHttpRequest(methodType: HttpMethodType, url: string,
                              withCredentials?: boolean,
                              contentType?: string): XMLHttpRequest {
@@ -257,7 +283,7 @@ function requestOptionsFromPostData(postData?: {[key: string]: string}):
   result.body = formDataFromDictionary(postData);
   return result;
 }
-  
+
 function formDataFromDictionary(postData: {[key: string]: string}) {
   const formData = new FormData();
   for (let postKey in postData) {
--- a/tensorboard/components/tf_backend/requestManager.ts
+++ b/tensorboard/components/tf_backend/requestManager.ts
@@ -156,7 +156,7 @@ export class RequestManager {
     return promise;
   }
 
-  public fetch(url: string, fetchOptions?: object): Promise<Response> {
+  public fetch(url: string, fetchOptions?: RequestInit): Promise<Response> {
     return new Promise((resolve, reject) => {
         const resolver = {resolve: resolve, reject: reject};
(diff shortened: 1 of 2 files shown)


tensorboard/components/tf_backend/requestManager.ts
```diff
@@ -207,14 +210,16 @@ export class RequestManager {
...
-        url,
-        requestOptions.withCredentials,
-        requestOptions.contentType);
+      const req = buildXMLHttpRequest(url, requestOptions);
       req.onload = function() {
         if (req.status === 200) {
-          resolve(JSON.parse(req.responseText));
+          // Edge and IE do not support "responseType == 'json'".
+          // Fall back to manual parsing.
+          if (req.responseType == 'json' && typeof req.response == 'string') {
```
wchargin: Won’t this try to double-parse if the result is an actual JSON string?

wchargin: I’m not sure that this test actually works in IE. On my machine:


```js
var xhr = new XMLHttpRequest();
xhr.open("GET", "http://example.com", true);
function report() {
  console.log("xhr.responseType: %s", JSON.stringify(xhr.responseType));
}

report();  // ""
xhr.responseType = "json";
report();  // ""
xhr.responseType = "blob";
report();  // "blob"
xhr.responseType = "json";
report();  // "blob"
```

(This is IE11 on Windows 10.)

stephanwlee: In case the result is an actual JSON string, the `response` will be an object (in non-IE case)

stephanwlee: Good call in a sense that I should have tried it on my Windows machine instead of assuming things based on MDN doc.

In any case (I will come back with my own experiments), this seems to be "correct". In case response CANNOT be parsed as the desired type, the `response` is an empty string. Just making sure, is your code snippet correct?

stephanwlee:
> I’m intrigued; how is responseType = "json" more efficient than
calling JSON.parse ourselves?

I really shouldn't have assumed things and sorry for using the word "efficient". My guess was based on assumption that instead of XHR returning its server response type, say, ArrayBuffer into string and our JavaScript doing `new ArrayBuffer(bufferString)`, by giving XHR an information about the type of response we are expecting, we can avoid some inefficiency. This is not substantiated so I should omit that. Will make edits soon.

tensorboard/components/tf_backend/requestManager.ts
wchargin: Hmm, doesn’t look like that to me—

```javascript
fetch("./test.json")
  .then((x) => x.json())
  .then((r) => void console.log("from fetch: %s of type %s", r, typeof r));

var xhr = new XMLHttpRequest();
xhr.open("GET", "./test.json", true);
xhr.responseType = "json";
xhr.onload = () => {
  var r = xhr.response;
  console.log("from XHR: %s of type %s", r, typeof r);
};
xhr.send();
```

```
from fetch: hello of type string
from XHR: hello of type string
```

(The server is `printf '"hello"' >test.json && python3 -m http.server`.
Client is Chrome M73 on Linux.)

wchargin: > In case response CANNOT be parsed as the desired type, the `response`
> is an empty string.

As far as I can tell:

 1. If the **response** from the server cannot be parsed as the desired
    type, then the `response` property of the XHR is `null` (at least
    when `responseType` is `"json"` on Chrome M73 on Linux).

 2. If one tries to set the `responseType` **property** is to an invalid
    response type (like `"json"` in IE11), then this property is not
    changed from its previous value. In particular, in IE11, it is not
    possible to create an XHR such that `req.responseType === "json"`.

The first point isn’t really what I’m interested in. I’m concerned that
the check `req.responseType == 'json'` will always evaluate to `false`
in IE, which means that control will never flow into the very block
that’s supposed to handle IE compatibility.

It’s totally plausible that I’m wrong about some or all of this; the XHR
spec and implementations are certainly not my expertise.

> Just making sure, is your code snippet correct?

Screenshot of code snippet in IE11

wchargin:
> My guess was based on assumption that instead of XHR returning its
> server response type, say, ArrayBuffer into string and our JavaScript
> doing `new ArrayBuffer(bufferString)`, by giving XHR an information
> about the type of response we are expecting, we can avoid some
> inefficiency

That’s totally plausible, though I would expect the cost of a `memcpy`
or two to dwindle in comparison to the rest of the work that TensorBoard
typically does.

tensorboard/components/tf_backend/requestManager.ts
```diff
@@ -156,6 +156,32 @@ export class RequestManager {
     return promise;
   }
 
+  public fetch(url: string, fetchOptions?: object): Promise<Response> {
```
wchargin: Can you use `fetchOptions?: RequestInit`?

tensorboard/components/tf_backend/test/requestManagerTests.ts
```diff
@@ -364,6 +375,112 @@ describe('backend', () => {
...
+        this.stubbedFetch.returns(Promise.resolve(new Response('Success', {status: 200})));
+        const rm = new RequestManager();
+
+        const response = await rm.fetch('foo');
+
+        expect(response).to.have.property('ok', true);
+        expect(response).to.have.property('status', 200);
+        const body = await response.text();
+        expect(body).to.equal('Success');
+
```
wchargin: nit: extraneous blank line

tensorboard/components/tf_backend/test/requestManagerTests.ts
```diff
@@ -364,6 +375,112 @@ describe('backend', () => {
...
+        this.resolvesAfter = function(value: any, timeInMs: number): Promise<any> {
+          return new Promise((resolve) => {
+            setTimeout(() => resolve(value), timeInMs);
+          });
+        }
+      });
+
+      it('resolves', async function() {
+        this.stubbedFetch.returns(Promise.resolve(new Response('Success', {status: 200})));
```
wchargin: nit: wrap before 91

tensorboard/components/tf_backend/test/requestManagerTests.ts
```diff
@@ -364,6 +375,112 @@ describe('backend', () => {
              });
          });
     });
+
+    describe('fetch', () => {
```
wchargin: Tests look great; thanks!

wchargin:
Just to note from offline discussions: one reason that something like
this was desired is that it enables clients to interpret responses as
array buffers, plain text, etc., rather than only JSON. This is helpful
for performance reasons: you can send binary blobs without having to
JSON-over-base64 encode them on the server.

tensorboard/components/tf_backend/requestManager.ts
stephanwlee: Absolutely! Did not know about that type. One day, I will learn all these types!

stephanwlee:
@wchargin thanks for writing out the motivation behind this PR. Updated the description to capture that so the commit message contains it.

PR #2107, opened 2019-04-11
Making hparams' table header break words.

erzel:


erzel:
@wchargin please take a look

Thanks!

tensorboard/plugins/hparams/tf_hparams_table_view/tf-hparams-table-view.html
```diff
@@ -108,6 +108,7 @@
       .table-header {
         /* line-break overflowing column headers */
         white-space: normal;
+        word-break: break-word;
```
stephanwlee: Note that this does not work in Firefox: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
Might want to give `text-overflow: break-word` a try instead.

Some screen shot definitely would be nice too.

erzel:
Thanks @stephanwlee.
Sorry for the delay in responding. I've been sick the last week.
I've converted the value to 'break-all' (I couldn't find a 'text-overflow:break-word' CSS statement)  which I think should be compatible with all modern browsers. PTAL.

erzel:
Here is a screenshot as you requested (using 'break-all'):
image

stephanwlee:
Argh, I meant to say `overflow-wrap` instead of `text-overflow`. https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

`break-all` can produce weirdness so I would strongly recommend using overflow-wrap.

erzel:
Done. @stephanwlee PTAL! 
Note that I can't merge.

erzel:
It doess look good! Thanks!

Issue #2109, opened 2019-04-11
update TF 2.0 summary usage to avoid deprecated/removed APIs


nfelt:
This is a cleanup issue to track removing TensorBoard's references to old versions of TF 2.0 summary APIs.  Specifically, the `tf.summary.import_event` API is removed entirely (in favor of `tf.summary.experimental.write_raw_pb`) and `tf.summary.summary_scope` is moved to `tf.summary.experimental.summary_scope`, since it's unclear that we want either API to exist in the long term and they're primarily intended for TB/TF-internal usage.

To be cross-compatible with TF versions before and after this change, we'll maintain code that can use either pathway for a little while, but this should be cleaned up before the TF 2.0 final release, which this issue tracks.

amanp592:
@nfelt i want to work on this! can you please guide me a little bit

shashvatshahi1998:
@nfelt can I also work with @amanp592 as a team on this issue?

colinshenc:
Hey, in the meantime, how to use tf.summary.import_event()?

PR #2110, opened 2019-04-11
Update TF 2.0 summary_scope()/import_event() usage

nfelt:
This updates TensorBoard's usage of `tf.summary.summary_scope()` and `tf.summary.import_event()` in TF 2.0 in anticipation of a change to those APIs, which moves the functionality under `tf.summary.experimental` and in the case of `import_event()`, also changes it to import `tf.compat.v1.Summary` proto bytes instead of `Event` protos.

Fallback logic is included so we stay compatible with TF versions that don't have this change, including current tf-nightly as of sending this PR.  I've filed https://github.com/tensorflow/tensorboard/issues/2109 to track cleaning up those fallbacks before TF 2.0 final.

Tested by confirming tests pass both before the TF change (as this PR is sent) and after (with the change made locally).  I also manually tested the `hparams_util` helper binary since the current test doesn't actually test its output.

### DIFF
--- a/tensorboard/plugins/audio/summary_v2.py
+++ b/tensorboard/plugins/audio/summary_v2.py
@@ -83,7 +83,11 @@ def audio(name,
       description=description,
       encoding=metadata.Encoding.Value('WAV'))
   inputs = [data, sample_rate, max_outputs, step]
-  with tf.summary.summary_scope(
+  # TODO(https://github.com/tensorflow/tensorboard/issues/2109): remove fallback
+  summary_scope = (
+      getattr(tf.summary.experimental, 'summary_scope', None) or
+      tf.summary.summary_scope)
+  with summary_scope(
       name, 'audio_summary', values=inputs) as (tag, _):
     tf.debugging.assert_rank(data, 3)
     tf.debugging.assert_non_negative(max_outputs)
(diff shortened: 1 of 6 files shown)


colinshenc:
how to use tf.summary.import_event() now with the latest version?

wchargin:
> how to use tf.summary.import_event() now with the latest version?

TensorFlow no longer provides `tf.summary.import_event` (which was never
a stable feature). If you were using `import_event` to import events
containing summary data, then instead of

```python
tf.summary.import_event(e.SerializeToString())
```

you should be able to use

```python
tf.summary.experimental.write_raw_pb(e.summary.SerializeToString(), step=e.step)
```

with recent releases of `tf-nightly-2.0-preview`.

wchargin:
Do we want to update the hparams tutorial, too?

```
docs/r2/hyperparameter_tuning_with_hparams.ipynb:        "  tf.summary.import_event(tf.compat.v1.Event(summary=exp_summary).SerializeToString())"
docs/r2/hyperparameter_tuning_with_hparams.ipynb:        "    tf.summary.import_event(tf.compat.v1.Event(summary=summary_start).SerializeToString())\n",
docs/r2/hyperparameter_tuning_with_hparams.ipynb:        "    tf.summary.import_event(tf.compat.v1.Event(summary=summary_end).SerializeToString())"
```

PR #2112, opened 2019-04-12
Redo add umap-js to embedding projector

cannoneyed:
I think I've got everything working with the reimport back into google3 - there are a few CLs for wchargin@ to review and then we should be good to go. Let's follow up on this next Monday.

This reverts commit 58df24b5aad3c828e8d87c3e8abbc507f8bc68b7.


third_party/js.bzl
```diff
@@ -206,6 +206,18 @@ def tensorboard_js_workspace():
       rename = {"numeric-1.2.6.js": "numeric.js"},
   )
 
+  filegroup_external(
+      name = "ai_google_pair_umap_js",
+      # no @license header
+      licenses = ["notice"],  # Apache License 2.0
+      sha256_urls = {
+          "85a2ff924f1bf4757976aca22fd0efb045d9b3854f5a4ae838c64e4d11e75005": [
+              "https://mirror.bazel.build/raw.githubusercontent.com/PAIR-code/umap-js/1.0.5/lib/umap-js.min.js",
```
wchargin: It seems a little strange to me that the 1.0.4→1.0.5 upgrade was
supposed to add license headers, yet the contents of the `.min.js` build
are identical before and after the upgrade, and the `.min.js` file
includes no license headers. Is this known?

cannoneyed: Yeah, the two bundles are exactly the same - this was to satisfy a minor nit in the reimport of some of the imported .ts files that aren't compiled in (test files).

PR #2114, opened 2019-04-16
update: module "program", removing line 384.

shenfe:
Just remove a useless line which is duplicated with another line below.

* Line 384 is not needed because of line 390.

### DIFF
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -381,7 +381,6 @@ def with_port_scanning(cls):
     # max_attempts: how many ports shall we try?
     should_scan = flags.port is None
     base_port = core_plugin.DEFAULT_PORT if flags.port is None else flags.port
-    max_attempts = 10 if should_scan else 1
 
     if base_port > 0xFFFF:
       raise TensorBoardServerException(


shenfe:
> Actually, I think that line 390 is the one that should be removed,
> because line 384 is right next to the comment on line 381 that explains
> what these variables do. Could you amend this PR to keep the earlier
> definition?

Sure, line 384 responds with line 381 indeed. However, the `if` block at line 386-389 has no business with variable `max_attempts`, and just acts as a check of variable `base_port`. That's why I chose to remove line 386.

shenfe:
> 
> 
> I don’t understand—it sounds like we agree that the line at 384 makes
> sense and the line at 390 doesn’t, and that we prefer line 384, so would
> you mind changing the patch to reinstate line 384 and remove line 390?

If we keep line 384 and the code raises an exception after then because of wrong `base_port`, variable `max_attempts` would have been declared as a waste at line 384.

wchargin:
What do you mean by “would have been declared as a waste”?

If performance is your concern, don’t worry about it. A branch and local
variable assignment to one of two integer constants is absolutely free
compared to all the other work that TensorBoard has to do. On the
contrary, colocating the variable with the comment describing it has
real advantages for readers of the code.

shenfe:
> What do you mean by “would have been declared as a waste”?
> 
> If performance is your concern, don’t worry about it. A branch and local
> variable assignment to one of two integer constants is absolutely free
> compared to all the other work that TensorBoard has to do. On the
> contrary, colocating the variable with the comment describing it has
> real advantages for readers of the code.

Alright, it seems that we have some different styles on coding. Anyway, 无伤大雅. Just go ahead and make changes by yourself. Never mind.

wchargin:
Okay, no worries—thanks anyway for the contribution and for bringing
this to our attention! :-)

PR #2126, opened 2019-04-18
hparams: add Python layer for hparams and metrics

wchargin:
Summary:
This change introduces `HParam`, `Metric`, and `Experiment` classes,
which represent their proto counterparts in a more Python-friendly way.
It similarly includes a `Domain` class hierarchy, which does not
correspond to a specific proto message, but rather unifies the domain
variants defined on the `HParamInfo` proto.

The design is roughly as in the original sketch of #1998.

The primary benefit of this change is that having first-class domains
enables clients to reuse the domain information for both the experiment
summary and the underlying tuning algorithm. We don’t provide a method
to do this out of the box, because we don’t actually provide any tuners
at this time, but it’s easy to write (e.g.) a `sample_uniform` function
like the one included in this commit. Then, sampling is as easy as

```python
    hparams = {h: sample_uniform(h.domain, rng) for h in HPARAMS}
```

It is also now more convenient to reference hparam values such that
static analysis can detect potential typos, because the `HParam` objects
themselves can be declared as constants and used as keys in a dict.
Writing `hparams["dropuot"]` fails at runtime, but `hparams[HP_DROPUOT]`
fails at lint time.

As a pleasant bonus, hparam definitions are now more compact, fitting
on one line instead of several. The demo code has net fewer lines.

Manual summary writer management is still required. A future change will
introduce a Keras callback to reduce this overhead.

Test Plan:
Some unit tests included, and the demo still works.

wchargin-branch: hparams-structured-api

### DIFF
--- a/tensorboard/plugins/hparams/BUILD
+++ b/tensorboard/plugins/hparams/BUILD
@@ -95,6 +95,7 @@ py_binary(
     srcs = ["hparams_demo.py"],
     srcs_version = "PY2AND3",
     deps = [
+        ":api",
         ":protos_all_py_pb2",
         ":summary",
         "//tensorboard:expect_absl_app_installed",
@@ -102,7 +103,6 @@ py_binary(
         "//tensorboard:expect_numpy_installed",
         "//tensorboard:expect_tensorflow_installed",
         "//tensorboard/plugins/scalar:summary",
-        "@com_google_protobuf//:protobuf_python",
         "@org_pythonhosted_six",
     ],
 )
@@ -145,6 +145,38 @@ sh_test(
         ":hparams_util",
     ],
 )
+
+py_library(
+    name = "api",
+    srcs = ["api.py"],
+    srcs_version = "PY2AND3",
+    visibility = [
+        "//visibility:public",
+    ],
+    deps = [
+        ":protos_all_py_pb2",
+        ":summary",
+        "@org_pythonhosted_six",
+    ],
+)
+
+py_test(
+    name = "api_test",
+    srcs = ["api_test.py"],
+    srcs_version = "PY2AND3",
+    visibility = [
+        "//visibility:public",
+    ],
+    deps = [
+        ":api",
+        ":metadata",
+        ":protos_all_py_pb2",
+        "//tensorboard:test",
+        "@com_google_protobuf//:protobuf_python",
+        "@org_pythonhosted_six",
+    ],
+)
+
 
 py_library(
     name = "summary",
--- a/tensorboard/plugins/hparams/BUILD
+++ b/tensorboard/plugins/hparams/BUILD
@@ -162,6 +162,7 @@ py_library(
 
 py_test(
     name = "api_test",
+    size = "small",
     srcs = ["api_test.py"],
     srcs_version = "PY2AND3",
     deps = [
@@ -173,7 +174,6 @@ py_test(
         "@org_pythonhosted_six",
     ],
 )
-
 
 py_library(
     name = "summary",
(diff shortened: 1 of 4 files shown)


wchargin:
cc @erzel

wchargin:
Reassigning to @stephanwlee to distribute load. For the other two in
particular I’d like @nfelt’s feedback on the summary APIs, but this one
is mostly summary-agnostic.

tensorboard/plugins/hparams/BUILD
```diff
@@ -146,6 +146,38 @@ sh_test(
...
+        "@org_pythonhosted_six",
+    ],
+)
+
+py_test(
+    name = "api_test",
+    srcs = ["api_test.py"],
+    srcs_version = "PY2AND3",
+    visibility = [
+        "//visibility:public",
```
stephanwlee: probably don't need to be really public, right?

tensorboard/plugins/hparams/api.py
```diff
@@ -0,0 +1,378 @@
...
+
+from tensorboard.plugins.hparams import api_pb2
+from tensorboard.plugins.hparams import summary
+
+
+class Experiment(object):
+  """A top-level experiment description.
+
+  An experiment has a fixed set of hyperparameters and metrics, and
+  consists of multiple sessions. Each session has different associated
```
stephanwlee: Here, is it okay if I interpret `session` as `tf.Session`? Or does it map better to TB's `run`s?

tensorboard/plugins/hparams/api.py
```diff
@@ -0,0 +1,378 @@
...
+    if dtype not in (int, float, bool, str):
+      raise ValueError("Unknown dtype: %r" % (dtype,))
+    self._dtype = dtype
+    for value in self._values:
+      if not isinstance(value, self._dtype):
+        raise ValueError(
+            "dtype mismatch: not isinstance(%r, %s)"
+            % (value, self._dtype.__name__)
+        )
+    self._values.sort()
```
stephanwlee: Did not think too much about this: Is this always desired and is this natural for users?

tensorboard/plugins/hparams/api.py
```diff
@@ -0,0 +1,378 @@
...
+      group=None,
+      display_name=None,
+      description=None,
+      dataset_type=None,
+  ):
+    """
+    Args:
+      tag: The tag name of the scalar summary that corresponds to this
+        metric (as a `str`).
+      group: An optional string listing the subdirectory under the
```
stephanwlee: Usage question: so I want metric to be written in arbitrary deep path, I should pass a string with the correct path separator for a specific OS?

tensorboard/plugins/hparams/hparams_demo.py
```diff
@@ -261,13 +244,7 @@ def run_all(logdir, verbose=False):
...
   session_index = 0  # across all session groups
   for group_index in xrange(flags.FLAGS.num_session_groups):
-    hparams = {
-        "conv_layers": rng.randint(1, 3),
-        "conv_kernel_size": rng.choice([3, 5]),
-        "dense_layers": rng.randint(1, 3),
-        "dropout": rng.uniform(0.1, 0.4),
-        "optimizer": rng.choice(["adam", "adagrad"])
-    }
+    hparams = {h: sample_uniform(h.domain, rng) for h in HPARAMS}
```
stephanwlee: Shouldn't it be `{h.name: sample_uniform(h.domain, rng) for h in HPARAMS}`?

 Perhaps the doc on `run` should be updated. Currently, it reads `hparams: A dict mapping hyperparameter names to values.`.

tensorboard/plugins/hparams/api.py
wchargin: It’s a fair question. Logically, a discrete domain is an unordered set,
so it seemed natural to me that the representation should be invariant
under permutation of `values`. But I could be convinced otherwise.

(Note that the frontend already displays discrete axes as sorted,
irrespective of the domain definition.)

PR #2130, opened 2019-04-19
hparams: add Keras callback

wchargin:
Summary:
A new `hparams.api.KerasCallback` class simplifies client APIs by
writing session start and end summaries automatically, with a dict of
hparams provided by the client. Cf. #1998.

This only works in TensorFlow eager mode. The stock `TensorBoard` Keras
callback works in both eager and graph modes, but to do so it must use
TensorFlow-internal symbols (`eager_mode` and `executing_eagerly` on the
context object, which we do not have access to).

Test Plan:
Unit tests included. The demo still works, generating valid data.

wchargin-branch: hparams-keras-callback


wchargin:
cc @erzel

tensorboard/plugins/hparams/api.py
```diff
@@ -412,3 +413,92 @@ def as_proto(self):
...
+    self._hparams = _normalize_hparams(hparams)
+    self._group_name = group_name if group_name is not None else ""
+    self._writer = tf.compat.v2.summary.create_file_writer(logdir)
+
+  def _write_summary(self, pb, step=None):
+    raw_pb = pb.SerializeToString()
+    with self._writer.as_default():
+      result = tf.compat.v2.summary.experimental.write_raw_pb(raw_pb, step=step)
+    sentinel = object()
+    if getattr(result, "numpy", sentinel) is sentinel:
```
nfelt: Any reason not just to use `tf.executing_eagerly()` up at the top?
https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/executing_eagerly

tensorboard/plugins/hparams/api_test.py
```diff
@@ -240,5 +244,111 @@ def test_dtype_mismatch(self):
...
+    ])
+    self.model.compile(loss="mse", optimizer=self.hparams["optimizer"])
+    self.logdir = os.path.join(self.get_temp_dir(), "logs")
+    self.callback = hp.KerasCallback(
+        self.logdir,
+        self.hparams,
+        group_name="psl27",
+    )
+
+  @test_util.run_v2_only("Requires eager mode.")
```
nfelt: You can still run this in TF 1.0 if we just set `tf.compat.v1.enable_eager_execution()` at the top of the test case?  And then you can use `with tf.compat.v1.Graph().as_default()` to get back into graph mode for the non-eager test case.

This is what the summary tests do, see https://github.com/tensorflow/tensorboard/blob/992bd022ed95c994d2f97f206a3c27bd5e351763/tensorboard/plugins/scalar/summary_test.py#L42

tensorboard/plugins/hparams/api_test.py
```diff
@@ -240,5 +244,111 @@ def test_dtype_mismatch(self):
...
+          metadata.PLUGIN_NAME,
+      )
+      plugin_data.append(value.metadata.plugin_data.content)
+
+    self.assertEqual(len(plugin_data), 2, plugin_data)
+    (start_plugin_data, end_plugin_data) = plugin_data
+    start_pb = metadata.parse_session_start_info_plugin_data(start_plugin_data)
+    end_pb = metadata.parse_session_end_info_plugin_data(end_plugin_data)
+
+    # Remove any dependence on system time.
```
nfelt: Any particular reason not to just patch time.time instead?

tensorboard/plugins/hparams/api_test.py
```diff
@@ -240,5 +244,111 @@ def test_dtype_mismatch(self):
...
+        hparams {
+          key: "dense_neurons"
+          value {
+            number_value: 8.0
+          }
+        }
+        """,
+        expected_start_pb,
+    )
+    self.assertEqual(start_pb, expected_start_pb)
```
nfelt: Optional but we might want to crib assertProtoEqual from TF for comparisons like this and put it in our TestCase class now that we have one again.  That way we can do the underlying assert on text format strings and hopefully get a more obvious diff for discrepancies.

tensorboard/plugins/hparams/api.py
```diff
@@ -412,3 +413,92 @@ def as_proto(self):
...
+      logdir,
+      hparams,
+      group_name=None,
+  ):
+    """Create a callback for logging hyperparameters to TensorBoard.
+
+    Each callback object is good for one session only.
+
+    Args:
+      logdir: The log directory for this session.
```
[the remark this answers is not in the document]
nfelt: Yes, I think del self._writer should be sufficient if we want to preserve that the callback is one-time-use only.

If we wanted the callback to be usable across sessions, I think it would be reasonable to interpret the ownership as taking no ownership of a passed-in writer (we retain a reference for re-use, and to purge the reference you must del the callback) and taking per-session ownership of a writer constructed from a logdir string.

tensorboard/plugins/hparams/api_test.py
nfelt: That's fine, I was mostly just curious.  No need to make changes, though FWIW I think just patching time.time to a constant value and asserting that both times equal that value would still be a somewhat stronger test even if it doesn't test that the two times are differentiated.

tensorboard/plugins/hparams/BUILD
```diff
@@ -172,8 +172,11 @@ py_test(
         ":api",
         ":metadata",
         ":protos_all_py_pb2",
+        "//tensorboard:expect_tensorflow_installed",
         "//tensorboard:test",
+        "//tensorboard/util:test_util",
```
wchargin: (note to self: remove unused `:test_util` dependency and import)

tensorboard/plugins/hparams/api.py
```diff
@@ -448,19 +452,25 @@ def __init__(
...
+          "hparams Keras callback cannot be reused across training sessions"
+      )
     raw_pb = pb.SerializeToString()
-    with self._writer.as_default():
-      result = tf.compat.v2.summary.experimental.write_raw_pb(raw_pb, step=step)
-    sentinel = object()
-    if getattr(result, "numpy", sentinel) is sentinel:
-      # Not an eager tensor, so the summary wasn't actually written. We
-      # don't have access to any session context to force the op. Bail.
+    if not tf.executing_eagerly():
```
nfelt: nit: might as well move this above `pb.SerializeToString()` and not serialize the PB if we're only going to fail anyway

tensorboard/plugins/hparams/api_test.py
```diff
@@ -257,16 +270,18 @@ def setUp(self):
...
+    self.callback = hp.KerasCallback(writer, self.hparams, group_name="psl27")
 
-  @test_util.run_v2_only("Requires eager mode.")
   def test_eager(self):
-    self.model.fit(x=[(1,)], y=[(2,)], callbacks=[self.callback])
+    def mock_time():
+      mock_time.time += 1
+      return mock_time.time
+    mock_time.time = 1556227801.875
+    initial_time = mock_time()
```
nfelt: might as well use `mock_time.time` here for a tighter bound?  Ditto for the end time.

tensorboard/plugins/hparams/api_test.py
```diff
@@ -288,14 +303,22 @@ def test_eager(self):
...
-    # Remove any dependence on system time.
-    start_pb.start_time_secs = 123.45
-    end_pb.end_time_secs = 234.56
+    # We're not the only callers of `time.time`; Keras calls it
+    # internally an unspecified number of times, so we're not guaranteed
+    # to know the exact values. Instead, we perform relative checks...
+    self.assertGreater(start_pb.start_time_secs, initial_time)
+    self.assertLess(start_pb.start_time_secs, end_pb.end_time_secs)
+    self.assertLess(start_pb.start_time_secs, final_time)
+    self.assertEqual(start_pb.start_time_secs % 1, initial_time % 1)
```
nfelt: Optional, but I'm not sure these last two asserts add much value, since all they really check (IMO) is that our mocked time advances 1 second at a time, which is a detail of the test, not the system under test.

wchargin: They’re intended to check that the fractional part doesn’t get dropped
in the pipeline. I can remove them if you prefer.

tensorboard/plugins/hparams/api_test.py
wchargin: Eh… sure. I have to be careful about equality strictness then. :-)

tensorboard/plugins/hparams/api.py
wchargin: Done.

tensorboard/plugins/hparams/api_test.py
nfelt: Not sure I follow?  It should be literally the same floating point value, no?

tensorboard/plugins/hparams/api_test.py
nfelt: I suppose.  Could we just test that instead by changing the mock to `+= 0.001` instead?

tensorboard/plugins/hparams/api_test.py
wchargin: Right, just `<=`/`>=` vs. `<`/`>`. Not a problem.

PR #2131, opened 2019-04-19
manager: make TensorBoardInfo directory universally accessible and useful

### LINKED ISSUE #2010: Permission denied issue when writing to `/tmp/.tensorboard-info`
- TensorBoard version (from pip package, also printed out when running `tensorboard`)
1.13.1
 - TensorFlow version if different from TensorBoard
1.13.1
 - OS Platform and version (e.g., Linux Ubuntu 16.04)
Ubuntu 16.04
 - Python version (e.g. 2.7, 3.5)
2.7
 
**Please describe the bug as clearly as possible, and if possible provide a minimal example (code, data, and/or command line) to reproduce the issue.  Thanks!**

The writing of tensorboard info files introduced in https://github.com/tensorflow/tensorboard/pull/1806 can cause permission problem under multi-user scenario. It directly create `.tensorboard-info` directory under `/tmp` as in https://github.com/tensorflow/tensorboard/blob/5fc3c8cea4b5f79c738345686a218f089b58ddba/tensorboard/manager.py#L237 . If the dir has already been created by an user, it will not be writable to other users.

image

wchargin:
Summary:
Fixes #2010.

This patch is primarily directed at Unices. On Windows, the underlying
problem should not be an issue, because the default temporary directory
seems to be user-scoped (e.g., `C:\Users\wchargin\AppData\Local\Temp`).

Test Plan:
Unit tests added. Before this change, the first test has an assertion
error (`'0o755' != '0o777'`) and the second test has an I/O error trying
to write to a non-writable directory. After this change, all tests pass.

Tested only on Linux.

wchargin-branch: info-dir-mode

### DIFF
--- a/tensorboard/manager_test.py
+++ b/tensorboard/manager_test.py
@@ -268,6 +268,24 @@ class TensorBoardInfoIoTest(tb_test.TestCase):
   def _list_info_dir(self):
     return os.listdir(self.info_dir)
 
+  def assertMode(self, path, expected):
+    """Assert that the permission bits of a file are as expected.
+
+    Args:
+      path: File to stat.
+      expected: `int`; a subset of 0o777.
+
+    Raises:
+      AssertionError: If the permissions bits of `path` do not match
+        `expected`.
+    """
+    stat_result = os.stat(path)
+    format_mode = lambda m: "0o%03o" % m
+    self.assertEqual(
+        format_mode(stat_result.st_mode & 0o777),
+        format_mode(expected),
+    )
+
   def test_fails_if_info_dir_name_is_taken_by_a_regular_file(self):
     os.rmdir(self.info_dir)
     with open(self.info_dir, "w") as outfile:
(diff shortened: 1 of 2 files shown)


wchargin:
Rats. Missed opportunity to title the commit, “Make `TensorBoardInfo`
directory universally accessible and useful”.

tensorboard/manager_test.py
```diff
@@ -268,6 +268,24 @@ def setUp(self):
...
+    Args:
+      path: File to stat.
+      expected: `int`; a subset of 0o777.
+
+    Raises:
+      AssertionError: If the permissions bits of `path` do not match
+        `expected`.
+    """
+    stat_result = os.stat(path)
+    format_mode = lambda m: "0o%03o" % m
```
nfelt: Could just use `oct()`?

wchargin: I had that, but I prefer the leading zeros (`"0o0"` is not clearly a
file mode, whereas `"0o000"` is). Also, `oct` is different on Python 2
vs. Python 3.

wchargin:
I should note that I did test that `os.chmod(dir_path, 0o777)` works
fine on Windows, without throwing an exception.

PR #2134, opened 2019-04-21
Check if tensor value only has 1 shape

chihuahua:
Previously, if a tensor value only has 1 shape, graph plugin logic would
fail to parse the list of shapes because pbtxt-parsing logic does not
differentiate between a field with a single message or a repeated field
containing a single message. This prevented an internal graph from being
parsed (with a console error).

Now, we can verify that the graph is rendered.


chihuahua:
Thanks a lot, William! I can no longer repro the previous issue, so I'm reverting this commit. Around April 10, I helped a neighbor get a graph to upload by patching this change internally.

wchargin:
Great; thanks!

PR #2137, opened 2019-04-22
Add handling of CMLE-hosted models to WitWidget

jameswex:

To support Google Cloud users, allow easy use of WIT in notebooks for models hosted by Google Cloud Machine Learning Engine (CMLE).

Added a base class for Jupyter and Colab impls of WitWidget to better organize/de-dup shared code.
Added logic to handle JSON input, which CMLE models accept. Internally we convert it to/from tf.Example to share it with WIT front-end.
Added methods to WitConfigBuilder to specify CMLE models.
Added method to specify target feature in loaded examples, as CMLE models need that feature stripped out before sending off for inference.
Added logic to perform CMLE inference when a CMLE model is specified, through use of an internally-defined custom predict function (implementation detail, this is hidden from users).
Updated README for CMLE use case and to correct existing out-of-date information.
Fixed bug in WitWidget for image-based models when using python 3, found during manual testing of this change.

Built pip package locally, installed and tested in Jupyter notebooks locally, and installed and tested in colab notebook (for both python 2 and python 3). Tested with CMLE-hosted models and our standard test notebooks as well.


jameswex:
@tolga-b please review

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
```diff
@@ -234,59 +202,10 @@ def _get_element_html(self):
...
-      infer_objs.append(inference_utils.run_inference_for_inference_results(
-          examples_to_infer, serving_bundle))
-    self.updated_example_indices = set()
-    inferences = {
-      'inferences': {'indices': indices_to_infer, 'results': infer_objs},
-      'label_vocab': self.config.get('label_vocab')}
+    base.WitWidgetBase.infer_impl(self)
     output.eval_js("""inferenceCallback('{inferences}')""".format(
-      inferences=json.dumps(inferences)))
+      inferences=json.dumps(self.inferences)))
```
tolga-b: It seems that we duplicate the memory usage of the tool for inference values by keeping it in the class. Is there any advantage to this?

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
```diff
@@ -305,70 +224,17 @@ def duplicate_example(self, index):
...
-    if 'image/encoded' in feature_list:
-      example_strings = [
-        self.json_to_proto(ex).SerializeToString()
-        for ex in self.examples]
-      encoded = base64.b64encode(
-        inference_utils.create_sprite_image(example_strings))
-      sprite = 'data:image/png;base64,{}'.format(encoded)
+    sprite = base.WitWidgetBase.create_sprite(self)
+    if sprite is not None:
+      print(sprite)
```
tolga-b: Is there a reason to print sprite here?

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
```diff
@@ -34,16 +35,24 @@ def __init__(self, examples):
...
+      These are the examples that will be displayed in WIT. If no model to
       infer these examples with is specified through the methods on this class,
       then WIT will display the examples for exploration, but no model inference
       will be performed by the tool.
     """
     self.config = {}
-    self.set_examples(examples)
     self.set_model_type('classification')
     self.set_label_vocab([])
+    if len(examples) > 0 and not (
```
tolga-b: It looks like WIT infers model's input type from the type of examples (json vs tf.Example). I think it would be good to add an extra force_json_conversion flag in set_cmle_model to enable comparison of two models where one accepts tf.Example and other json as tf.Example->json conversion is possible. json->tf.Example conversion is not doable in all cases since the data types are missing in json so we cannot make this process fully transparent hence the requirement for force_json_conversion flag.

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
```diff
@@ -441,3 +450,80 @@ def set_compare_custom_predict_fn(self, predict_fn):
...
+      self, in order to enabled method chaining.
+    """
+    self.set_compare_inference_address(project)
+    self.set_compare_model_name(model)
+    self.store('use_cmle_compare', True)
+    if version is not None:
+      self.set_compare_model_signature(version)
+    return self
+
+  def set_target_feature(self, target):
```
tolga-b: In future, we may add another function (or extend this function) to blacklist features in case people want to add extra features for slicing purposes (since cloud models crash when there are unknown extra features).

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
```diff
@@ -441,3 +450,80 @@ def set_compare_custom_predict_fn(self, predict_fn):
...
+      target: The name of the feature in the examples that represents the value
+      that the model is trying to predict.
+
+    Returns:
+      self, in order to enabled method chaining.
+    """
+    self.store('target_feature', target)
+    return self
+
+  def _set_uses_json_input(self, is_json):
```
tolga-b: This may need edits after force_json_conversion flag as the main model and compare model may have different inputs in that case.

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
jameswex: fixed

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
jameswex: leftover from debugging

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
jameswex: acknowledged, keeping simple for now

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
jameswex: This still works, as in the cmle infer method we convert from tf.example to json if their uses_json_input is true (meaning the examples were initially provided as JSON) ,or if for that model the new override argument was set to true (meaning the examples were provided as tf.Example but for this model we want to convert to json to send it to the model).

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
jameswex: done

jameswex:
@wchargin @stephanwlee could someone from TB please review this PR? Thanks!

PR #2139, opened 2019-04-22
hparams: write `hparams_config` summary directly

wchargin:
Summary:
The `Experiment` object bundled `HParam`s and `Metric`s with some
metadata that’s not actually used in the current UI. We don’t think that
it pulls its conceptual weight, so this commit replaces it with a direct
summary-writing operation.

This function will soon be extracted into a `summary_pb2` module, as
part of a [larger plan to refactor the `api` module][1]. Making this
change first minimizes churn in the demo code.

[1]: https://github.com/tensorflow/tensorboard/pull/2139#discussion_r279106791

Cf. #1998.

Test Plan:
Unit tests modified appropriately, and the demo still works.

wchargin-branch: hparams-experiment-writing


tensorboard/plugins/hparams/hparams_demo.py
```diff
@@ -215,13 +215,10 @@ def run_all(logdir, verbose=False):
...
 
-  base_writer = tf.summary.create_file_writer(logdir)
-  with base_writer.as_default():
+  with tf.summary.create_file_writer(logdir).as_default() as base_writer:
     experiment = hp.Experiment(hparams=HPARAMS, metrics=METRICS)
-    experiment_string = experiment.summary_pb().SerializeToString()
-    tf.summary.experimental.write_raw_pb(experiment_string, step=0)
-    base_writer.flush()
-  base_writer.close()
+    hp.experiment(experiment)
```
wchargin: Note: The awkwardness of `hp.experiment(experiment)`, or the equivalent
but worse `hp.experiment(hp.Experiment(…))`, is why I’d originally
proposed having `experiment.write()`. I understand why you don’t want
that. If you have suggestions for improvement here, I’m all ears.

tensorboard/plugins/hparams/api_test.py
```diff
@@ -155,6 +155,63 @@ def test_summary_pb(self):
...
+  def test_write_experiment_v1(self):
+    experiment = hp.Experiment(
+        hparams=[hp.HParam("num_units", hp.Discrete([16, 32]))],
+        metrics=[hp.Metric("accuracy")],
+    )
+    logdir = os.path.join(self.get_temp_dir(), "logs")
+    with tf.compat.v1.Session() as sess:
+      with tf.compat.v2.summary.create_file_writer(logdir).as_default() as w:
+        with tf.compat.v2.summary.record_if(True):
+          tf.contrib.summary.initialize()
```
wchargin: Note: I couldn’t find a way to get the V1 test case to work without
calling into `tf.contrib`. If there is one, I’d love to hear it.

nfelt: `sess.run(w.init())` should do the trick, if the problem was the writer not being initialized?

tensorboard/plugins/hparams/api_test.py
```diff
@@ -155,6 +155,63 @@ def test_summary_pb(self):
...
+  @test_util.run_v1_only("Requires graph-mode summary writing semantics.")
+  def test_write_experiment_v1(self):
+    experiment = hp.Experiment(
+        hparams=[hp.HParam("num_units", hp.Discrete([16, 32]))],
+        metrics=[hp.Metric("accuracy")],
+    )
+    logdir = os.path.join(self.get_temp_dir(), "logs")
+    with tf.compat.v1.Session() as sess:
+      with tf.compat.v2.summary.create_file_writer(logdir).as_default() as w:
+        with tf.compat.v2.summary.record_if(True):
```
nfelt: This line shouldn't be necessary with the V2 summary API.

tensorboard/plugins/hparams/api_test.py
```diff
@@ -155,6 +155,63 @@ def test_summary_pb(self):
...
+      summary_pb: A `summary_pb2.Summary` object.
+    """
+    files = os.listdir(logdir)
+    self.assertEqual(len(files), 1, files)
+    events_file = os.path.join(logdir, files[0])
+    for event in tf.compat.v1.train.summary_iterator(events_file):
+      if event.WhichOneof("what") != "summary":
+        continue
+      self.assertEqual(event.summary, summary_pb)
+      break
```
nfelt: nit, but you can simplify this to just `return` here and then skip the `else` clause and `self.fail()` unconditionally.  It's one line shorter and avoids introducing a slightly more novel language feature when there's a more standard alternative.

tensorboard/plugins/hparams/api_test.py
```diff
@@ -155,6 +155,63 @@ def test_summary_pb(self):
         expected_experiment_pb,
     )
 
+  def _assert_unique_summary(self, logdir, summary_pb):
+    """Test that `logdir` contains exactly one summary, `summary_pb`.
+
+    Specifically, `logdir` must be a directory containing exactly one
+    entry, which must be an events file of whose events exactly one is a
```
nfelt: Strictly speaking this doesn't verify that there aren't additional summaries after the first one.  The test passes immediately after finding the first valid summary.

It's a little less pretty but in the past I've just basically done

```python
events = list(tf.compat.v1.train.summary_iterator(events_file))
self.assertEqual(len(events), 2)
self.assertEqual(events[1].summary, summary_pb)
```

tensorboard/plugins/hparams/hparams_demo.py
wchargin: I hesitate to just call the endpoints `.hparams`, because hparams really
has three kinds of summaries, and of the three kinds of summaries, the
name `.hparams` sounds more like `session_start_pb` (logging your actual
hparam values) than `experiment_pb`.

wchargin: I hesitate to just call the endpoints `.hparams`, because hparams really
has three kinds of summaries, and of the three kinds of summaries, the
name `.hparams` sounds more like `session_start_pb` (logging your actual
hparam values) than `experiment_pb`.

I guess we _could_ provide `summary_v2` with `hparams`, `session_start`,
and `session_end` (the last two for non-Keras users), but (a) these
would look more out of place as `writer` methods (“session start? what
session?”), and (b) that’s not great if we’re trying to get rid of the
confusing “session” terminology.

I’ll try to think about this a bit more and get back to it tomorrow.

tensorboard/plugins/hparams/hparams_demo.py
```diff
@@ -215,13 +215,10 @@ def run_all(logdir, verbose=False):
...
+    base_writer.close()
```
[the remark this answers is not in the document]
wchargin: Got it; I wasn’t sure what was idiomatic. Thanks.

tensorboard/plugins/hparams/api_test.py
```diff
@@ -155,6 +155,63 @@ def test_summary_pb(self):
...
+    events_file = os.path.join(logdir, files[0])
+    for event in tf.compat.v1.train.summary_iterator(events_file):
+      if event.WhichOneof("what") != "summary":
+        continue
+      self.assertEqual(event.summary, summary_pb)
+      break
+    else:
+      self.fail("No summary data found")
+
+  @test_util.run_v2_only("Requires eager summary writing semantics.")
```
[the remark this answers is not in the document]
wchargin: Will do.

tensorboard/plugins/hparams/hparams_demo.py
nfelt: That's fair, but IMO the pretty strong reason to use `hparams` would be consistency with all the other plugins, which use their plugin name as (at least a prefix to) the summary op name when they have a dedicated summary op.  And I assume we would consider `writer.hparams_experiment()` to be too unwieldy :)

In fairness, I think part of the difficulty here is that hparams sort of invented summary formats for concepts that are more properly broader than just the hparams plugin; "experiment" and "session" are constructs that go beyond just setting hparams.  I don't expect us to resolve all of that in the short term, but what I'd maybe suggest is that we can figure out a subset of the full end-state API that we're pretty comfortable with now (and which still improves on the original API).  We don't necessarily need to expose a full Experiment construct with a name and user when there is nothing in TensorBoard today to actually visualize that data.

For example, a subset of the API could be something like:

- `hparams_config(hparams, metrics)` to log the hparam and metric definitions (optional and if not logged, TB will infer domains and use all scalar values as metrics)
- `hparams(hparams)` to log the actual values

To me that seems like it already covers virtually all of the functionality that's actually usable today in the hparams dashboard; the main exception is the session end status value, which doesn't seem like a big deal to me (and there's still the "plumbing" API for now if folks who aren't using the keras callback really want that).  Even the session group name doesn't add much value since session group aggregation isn't implemented on the frontend.

This would buy us a little more time to fine tune how we'd like to capture experiment and session information in the generalized summary context.

wchargin: Okay, I can be convinced that `Experiment` isn’t worth its conceptual
weight, and that we should remove it from the public API in favor of
directly writing `hparams_config(hparams, metrics)`.

Instead of complicating the module and dependency structure more by
introducing both `api` and `summary_v2`, I would _like_ to rename `api`
to `summary_v2`, containing:

  - `hparams(hparams)` to log a session start summary;
  - `hparams_config(hparams, metrics)` to log an experiment summary;
  - `HParam`, `Metric`, and domain descriptors, as currently;
  - `KerasCallback`, as currently.

But `KerasCallback` has a hard (module-time) dependency on `tf.keras`,
so this would preclude `summary_v2.hparams` in notf builds, which is not
great. The two options that I see right now are:

 1. omit the callback from `summary_v2` and instead put it in `api`
    (which would need to exist after all), and also let `api` forward
    the other `summary_v2` symbols; or

 2. move the callback into a separate `callback` module.

Option (1) has the downside that I can’t rename `api` to `__init__` as
I’d hoped, because you again wouldn’t be able to import `summary_v2`
without depending on `tf.keras`. Option (2) has the downside of being
ugly.

I guess option (1) is less bad. At least in that case, the vast majority
of users could just use `api` and not worry about `summary_v2` unless
they need a notf build. I’ll implement it and see if I can come up with
anything better.

wchargin: Discussed offline. We’re comfortable with this. Explicitly:

  - `hparams.api`:
      - is the one-stop user-facing shop, to be imported `as hp`
      - re-exports `*` from `hparams.summary_v2`
      - re-exports `*` from `hparams.callback`
      - transitively depends on `tf.keras`
  - `hparams.summary_v2`:
      - contains `hparams(concrete_hparam_values)`
      - contains `hparams_config(hparams, metrics)`
      - contains `HParam`, `Metric`, and domain descriptors
      - depends only on `api_pb2`, `summary`, etc.
      - is the only public API for notf users
  - `hparams.callback`:
      - contains `KerasCallback`
      - depends on `summary_v2`
      - depends on `tf.keras`
  - nothing is added to `tb.summary.v2.*`
  - nothing is currently added to `tf.summary`; in the future, either
    `tf.summary` or a unified TensorBoard `writer` may depend on
    `summary_v2` but not `callback` to avoid a `tf.keras` dependency

wchargin:
Ready for re-review when you get a chance.

PR #2140, opened 2019-04-22
hparams: add docstring describing public API

wchargin:
Summary:
This summarizes APIs introduced in #2126, #2130, #2139, and #2188. See
also #1998.

Test Plan:
None. These code snippets aren’t meant to be literally executable out of
context; that’s what the demo is for.

wchargin-branch: hparams-api-docs


wchargin:
Also taking opinions as to what the primary entry point should be:

```python
# Current:
from tensorboard.plugins.hparams import api as hp

# But could be:
from tensorboard.plugins import hparams as hp
import tensorboard.plugins.hparams as hp  # equivalent (in this case)
```

I lean toward the latter, but could be convinced to keep it as is.

tensorboard/plugins/hparams/api.py
```diff
@@ -14,7 +14,85 @@
...
+>>>
+>>> model = model_fn(hparams)
+>>> callbacks = [
+>>>     tf.keras.callbacks.TensorBoard(logdir),
+>>>     hp.KerasCallback(logdir, hparams),
+>>> ]
+>>> model.fit(..., callbacks=callbacks)
+
+To control how hyperparameters and metrics appear in the TensorBoard
+frontend, you can define `HParam` and `Metric` objects and collect them
```
stephanwlee: frontend -> ui (or omit)

The doc says `...The domains will also be reflected in the TensorBoard UI.` below.

tensorboard/plugins/hparams/api.py
```diff
@@ -14,7 +14,85 @@
...
+...         HP_NEURONS,
+...     ],
+...     metrics=[
+...         hp.Metric("xent", group="validation", display_name="cross-entropy"),
+...         hp.Metric("f1", group="validation", display_name="F&#x2081; score"),
+...         hp.Metric("loss", group="train", display_name="training loss"),
+...     ],
+... )
+>>> with tf.summary.create_file_writer(base_logdir).as_default():
+...   hp.experiment(experiment)  # write experiment summary
```
stephanwlee: Does this only support TF2? If so, is it true for the Keras callback and should we clarify that above?

tensorboard/plugins/hparams/api.py
wchargin: Good point. Consistency is good. Will use “UI”.

tensorboard/plugins/hparams/api.py
wchargin: The `hp.experiment` endpoint supports TF 1.x and 2.x both; this is
tested in #2139. The Keras callback indeed only supports TF 2.x. I’ll
clarify that above. Thanks!

wchargin: More precisely, the Keras callback only supports TF eager. You can use
it in V1 if you explicitly enable eager mode.

PR #2141, opened 2019-04-23
Update example viewer to handle long text features

jameswex:

Long text features were always clipped (will ellipsis, can scroll through them slowly) after ~20 characters in WIT's vz-example-viewer due to the use of <input> elements. Switched to iron-autogrow-textarea to always be perfect size for the content, as requested by users.

Switched from input elems to iron-autogrow-textarea elems, which perfectly solve the problem in general.

Updated styling of the text area per feedback (more style updates coming from Mahima).

Switched example viewer to always use "stacked" display setting, removed controls to switch between display styles as the controls were deemed unnecessary and "stacked" mode is the appropriate mode to always be in.

* Screenshots of UI changes

Showing handling of long text correctly, with multiple lines:
datapointlong

Showing handling of features with multiple values, including one currently being edited:
datapointtwo

Tested demo apps, adding feature values to show multiple feature values per feature.

Built pip package, installed in colab, tested toxicity colab to verify handling of long, variable length strings.


jameswex:
@tolga-b please review - moving forward with this change as Mahima is traveling this week. Her work can continue on top of this change.

tensorboard/components/vz_example_viewer/vz-example-viewer.html
```diff
@@ -101,54 +102,55 @@
...
+        border: 1px solid #FFFFFF;
       }
 
       .value-pill-stacked {
         width: 100%;
         margin: 4px 4px 4px 0;
+        padding: 2px;
       }
 
       .value-pill-grid {
```
tolga-b: We can remove this rule in future now value-pill-grid is not necessary.

jameswex: plan on keeping these rules in the file in case we re-enable the mode switching or move to another display mode by default.

jameswex:
> Thank you James, this looks good to me.
> Have you also verified that closest counterfactual mode works correctly?

yes verified correctness, thanks

tensorboard/components/vz_example_viewer/vz-example-viewer.ts
```diff
@@ -836,12 +836,14 @@ Polymer({
       this.getBoundingClientRect().right -
       inputControl.getBoundingClientRect().right + 30) + 'px';
     this.showDeleteValueButton = true;
-
+    inputControl.style.border = '1px solid #ffeb3b';
```
stephanwlee: though there are enough examples of this, can we just use the CSS pseudo-selector?
i.e.,
```css
paper-input:focus {
  border: 1px solid #ffeb3b;
}
```

I want styling logic/code to be yanked away from JS into CSS as much as possible.

jameswex: i also want this in CSS, but iron-autogrow-textarea :focus selector doesn't work as expected. even when the item is focused by user input (clicking, typing), that css selector isn't chosen. must have to do with its design (contains a div which contains a textarea). wasn't aware of a way to get the selector on that element based on properties of a grandchild element, so moved to JS code.

do you have any ideas of how it can be done just in CSS in this case?

stephanwlee:
```css
iron-autogrow-textarea[focused] {
  border: 1px solid #ffeb3b;
}
```

seems to work

jameswex: thanks that works! i was trying ":focus" in the specifier and that wasn't being picked up.

tensorboard/components/vz_example_viewer/vz-example-viewer.html
```diff
@@ -101,54 +102,60 @@
...
+      iron-autogrow-textarea {
+        --iron-autogrow-textarea: {
+          font-size: 14px;
+          color: #3C4043;
+        }
+      }
+
+      iron-autogrow-textarea.value-same {
+        --iron-autogrow-textarea: {
+          font-size: 14px;
```
stephanwlee: You probably can omit font-size declaration in this and below cases.

jameswex: it didn't inherit from the more general rule when the font-size was in the mixin. i moved it up to the main part of the  iron-autogrow-textarea rule and now it inherits correctly and i removed it from the class-specific rules.

stephanwlee:
Please feel free to merge at your convenience. Please provide more meaningful commit message when you squash (also, please send a PR with the minimal commits in the future)

PR #2144, opened 2019-04-23
Fix graph-loading for tf-graph-app and tf-debugger

stephanwlee:
tf-graph-control, being inside graph plugin, changed recently to work
better with other graph components like tf-graph-controls. More
specifically, the problem happen because the tf-graph-loader was
refactored to take `selection` as the primary input to the loader and
it used run & type of graph selected by user to fetch appropriate data.
This contrasts with previous input `datasets` which took a `path` as an
input which loader used to fetch the data from. tf-graph-app and
tf-debugger-dashboard plugin were using the datasets prop to pass in
the special endpoint where the data lives.

Before describing refactor, we should define what the tf-graph-loader
should be. The loader is a UI-less functional component (renders no DOM;
uses lifecycle method and props to provide some functionality) that is
capable of fetching various informations like graph, conceptual graph,
and profile info (RunMetadata). We have recently introduce more
complexity to it.

Interestingly, all tf-graph variants (internal to Google too) does not
make use of any other data type other than the graph and there is almost
no reason to use the tf-graph-loader. The graph utilities provide an
ability to fetch and parse GraphDef, create SlimGraph, and create a
hierarchy with ease. The core values the tf-graph-loader provides are
1. error handling and 2. progress tracker (uses Polymer API). By
reusing the component, one can provide some consistency there.

There were few ways to refactor this:
1. Behavior (mixin)
2. tf-graph-dashboard can have specialized wrapper that converts
`selection` into the old `datasets`.
3. Put bulk of fetch+parse+graph_building to util and have specialized
graph loaders

To stephanwlee, option 2 was unfavorable because the tf-graph-loader
becomes needlessly complex with different control flow depending on type
of data requested when it was only used/required by tf-graph-dashboard.
Instead, we chose to go with the option 3. Instead of inadvertently breaking
others due to dependency on tf-graph-* components, for our velocity, we
believe all variants of graph should use utility method without reusing the 
components.

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_app/tf-graph-app.html
+++ b/tensorboard/plugins/graph/tf_graph_app/tf-graph-app.html
@@ -92,7 +92,6 @@ tf-graph-controls {
     <div class="side">
       <tf-graph-controls
           color-by-params="[[colorByParams]]"
-          stats="[[stats]]"
           color-by="{{colorBy}}"
           render-hierarchy="[[_renderHierarchy]]"
           selected-node="{{selectedNode}}"
@@ -101,7 +100,6 @@ tf-graph-controls {
       <tf-graph-loader id="loader"
           out-graph-hierarchy="{{graphHierarchy}}"
           out-graph="{{graph}}"
-          out-stats="{{stats}}"
           progress="{{_progress}}"
           selected-file="[[selectedFile]]"
       ></tf-graph-loader>
@@ -110,7 +108,6 @@ tf-graph-controls {
       <tf-graph-board id="graphboard"
           graph-hierarchy="[[graphHierarchy]]"
           graph="[[graph]]"
-          stats="[[stats]]"
           progress="[[_progress]]"
           color-by="[[colorBy]]"
           color-by-params="{{colorByParams}}"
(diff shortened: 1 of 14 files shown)


tensorboard/plugins/graph/tf_graph_app/tf-graph-app.html
```diff
@@ -176,22 +176,36 @@
...
-      // Fetch a pbtxt file. The fetching will be part of the loading sequence.
-      this.$.loader.datasets = [{
-        // Just name the dataset based on the file location.
-        "name": this.pbtxtFileLocation,
-        "path": this.pbtxtFileLocation,
-      }];
-      this.$.loader.set('selectedDataset', 0);
-    } else if (this.pbtxt) {
+  _fit: function() {
+    this.$$('#graphboard').fit();
```
nfelt: This seems to be new - what does this do?  I also don't see it invoked anywhere?

stephanwlee:
@nfelt based on your feedback, I changed dashboard specific loader (which is more feature rich -- like conceptual graph and stats) to tf-graph-dashboard-loader and created a "minimal" tf-graph-loader that should address all of our existing users of tf-graph-loader.

PR #2149, opened 2019-04-25
Fix sliced confusion matrices issue

jameswex:

For multi-class confusion matrices in WIT, if any slice of data doesn't contain predictions/labels from any of the possible classes, those classes were being omitted from the confusion matrix for that slice, leading to inconsistent matrices that are hard to compare/understand.

Added argument to tf-confusion-matrix to optionally take a list of all cell labels to put into confusion matrix. If this is provided, use those are rows/columns instead of extracting rows/columns from the provided "counts" information

* Screenshots of UI changes

Notice all conf matrices are 3x3, even when there are empty rows or columns. These are the fixed visuals.

confmatrix

Ran multiclass demo, saw 3x3 conf matrix, sliced data by class, saw 3 3x3 matrices with some empty rows/columns as expected, instead of strange 1x1 and 2x2 matrices as before.

Also verified still works correctly with other binary classification demos.


jameswex:
@tolga-b plz review

jameswex:
> Thank you James. Looks good, just one minor comment. Do the polymer triggers work correctly when one changes from binary->multi-class model and the other way around without refreshing the page?

yes, i tested in iris demo by setting multiClass param on WIT to false, then true, to see how the conf matrix would update.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-confusion-matrix.html
```diff
@@ -95,26 +105,32 @@
...
-        // Get all labels from the counts dict.
-        let allLabels = {};
-        const outerLabelList = Object.keys(counts);
-        for (let i = 0; i < outerLabelList.length; i++) {
-          allLabels[outerLabelList[i]] = true;
-          const innerLabelList = Object.keys(counts[outerLabelList[i]]);
-          for (let j = 0; j < innerLabelList.length; j++) {
-            allLabels[innerLabelList[j]] = true;
+        // Get all labels from the provided object, or counts dict.
+        let allLabels = allItems;
```
stephanwlee: Two points:
1. Below, you are mutating the `allItems` without making a copy. That is dangerous.
2. What you are essentially doing here is creating an array of label list. Just use an array? If there can be dup, please use `const set = new Set()` and do `Array.from(set.values())`

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-confusion-matrix.html
```diff
@@ -82,7 +82,17 @@
...
+
+        /**
+         * Dictionary where keys are all cell labels to be used in the
+         * confusion matrix. If empty, then the cell labels are extracted
+         * from the provided counts dictionary. This is used for when a
+         * confusion matrix should have more rows/columns that just the
+         * information from counts, such as in a multiclass problem where
+         * the counts do not contain information from all possible classes.
+         */
+        allItems: {type: Object, value: {}},
```
stephanwlee: `value: {}` -> `value: () => ({})`

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-confusion-matrix.html
jameswex: changed to using Set and then converting to list as suggested, and removed mutation of provided data.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-confusion-matrix.html
jameswex: fixed

PR #2162, opened 2019-04-26
add new test for exporting TB symbols under tf.summary

nfelt:
This adds a new python-based integration test that for TF 2.0, the Tensorboard-provided symbols are successfully exported under `tf.summary`.  We already have tests for this in build_pip_package.sh, but that only runs regularly in our Travis CI, whereas this python test easily can run wherever our python tests run, including our internal CI (which I verified with a test sync).

This should help avoid issues like https://github.com/tensorflow/tensorflow/issues/28027.

### DIFF
--- a/tensorboard/summary/tf_summary_test.py
+++ b/tensorboard/summary/tf_summary_test.py
@@ -1,0 +1,55 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Integration test for exporting TB symbols under the `tf.summary` API.
+
+This checks the integration between TF and TB to verify that the mechanism we
+use to export TensorBoard symbols as part of the `tf.summary` API works.
+
+We also test this at an even higher level (and with many different import
+sequences) via our pip package build script. But that script is run only in
+our external CI, whereas this python test can run anywhere our normal tests run.
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import sys
+import unittest
+
+class TfSummaryExportTest(unittest.TestCase):
+
+  def test_tf_summary_export(self):
+    # Check as a precondition that TF wasn't already imported.
+    self.assertEqual('notfound', sys.modules.get('tensorflow', 'notfound'))
+    import tensorflow as tf
+    if not tf.__version__.startswith('2.'):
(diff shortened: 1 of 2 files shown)


tensorboard/summary/tf_summary_test.py
```diff
@@ -0,0 +1,55 @@
...
+      if hasattr(tf, 'compat') and hasattr(tf.compat, 'v2'):
+        tf = tf.compat.v2
+      else:
+        self.skipTest('TF v2 summary API not available')
+    # Check that tf.summary contains both TB-provided and TF-provided symbols.
+    expected_symbols = frozenset(
+        ['scalar', 'image', 'audio', 'histogram', 'text']
+        + ['write', 'create_file_writer', 'SummaryWriter'])
+    self.assertLessEqual(expected_symbols, frozenset(dir(tf.summary)))
+    # Ensure we can deference symbols as well.
```
wchargin: sp.: *deference* → *dereference*

tensorboard/summary/tf_summary_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+import collections
+import sys
+import unittest
+
+class TfSummaryExportTest(unittest.TestCase):
+
+  def test_tf_summary_export(self):
+    # Check as a precondition that TF wasn't already imported.
+    self.assertEqual('notfound', sys.modules.get('tensorflow', 'notfound'))
```
wchargin: Why is this needed, given that we immediately import TensorFlow?

The existing occurrence of this assertion in `summary_test.py` ensures
that TensorFlow isn’t automatically imported by accessing certain
symbols, but I don’t see why we need something like that here.

tensorboard/summary/tf_summary_test.py
```diff
@@ -0,0 +1,55 @@
...
+We also test this at an even higher level (and with many different import
+sequences) via our pip package build script. But that script is run only in
+our external CI, whereas this python test can run anywhere our normal tests run.
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
```
wchargin: Unused import.

tensorboard/summary/tf_summary_test.py
```diff
@@ -0,0 +1,55 @@
...
+import sys
+import unittest
+
```
wchargin: nit: Add blank line.

tensorboard/summary/tf_summary_test.py
nfelt: Done.

tensorboard/summary/tf_summary_test.py
nfelt: Done.

tensorboard/summary/tf_summary_test.py
nfelt: I think this is a stronger test if we ensure that it's testing the entire flow of importing TF and then using tf.summary, versus having the import happen "somewhere else".  This check enforces that we don't modify this test down the road to import TF at the top of the file or in another test method, which would muddy state going into this test.

I've elaborated the comment to explain this reasoning better.

tensorboard/summary/tf_summary_test.py
nfelt: Done.

tensorboard/summary/tf_summary_test.py
wchargin: Okay, fine with me. Thanks.

PR #2163, opened 2019-04-27
hparams: refactor to isolate TF dependencies

wchargin:
Summary:
This sequence of commits turns the current `api` module into a wrapper
around two modules, `callback` (which depends on TensorFlow and _only_
provides the Keras callback) and `summary_v2` (which does not, and
provides everything else). Thus, it is now possible in principle to
write hparams summaries without a TensorFlow dependency, though in
practice `summary_v2` is currently implemented in terms of `summary`,
which depends on the TensorFlow copy of the `Summary` proto.

This is an internal refactor; the demo code is unchanged. 

Reorganization to match the structure described here:
<https://github.com/tensorflow/tensorboard/pull/2139#discussion_r279106791>

Test Plan:
Existing unit tests retained, up to refactoring. The demo still works.

wchargin-branch: hparams-api-refactor


wchargin:
It is much easier to review these commits individually.

tensorboard/plugins/hparams/api.py
```diff
@@ -12,456 +12,28 @@
...
-    if isinstance(k, HParam):
-      k = k.name
-    if k in result:
-      raise ValueError("multiple values specified for hparam %r" % (k,))
-    result[k] = v
-  return result
+del absolute_import
+del division
+del print_function
+del summary_v2
```
nfelt: `del keras` too?

wchargin: Yes! Thanks.

PR #2164, opened 2019-04-27
feat: enable scrolling when added large image

Ir1d:

In my case the image is too big I suppose it's 1920x640.
image
After clicking:
image

I let it scroll.

* Screenshots of UI changes

image

But it's not that visually pleasing totally.

image

I'm not sure how to make it more beautiful, but the current fix works perfectly for me. Hope it helps.

### DIFF
--- a/tensorboard/plugins/image/tf_image_dashboard/tf-image-loader.html
+++ b/tensorboard/plugins/image/tf_image_dashboard/tf-image-loader.html
@@ -113,7 +113,7 @@ future for loading older images.
         height: auto;
         position: relative;
         margin: 0 15px 40px 0;
-        overflow-y: auto;
+        overflow-x: auto;
       }
 
       /** When actual size shown is on, use the actual image width. */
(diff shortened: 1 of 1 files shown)


Ir1d:
I signed it!

tensorboard/plugins/image/tf_image_dashboard/tf-image-loader.html
```diff
@@ -113,6 +113,7 @@
         height: auto;
         position: relative;
         margin: 0 15px 40px 0;
+        overflow-y: scroll;
```
stephanwlee: scroll should be changed to `auto`.

Ir1d: thx! i forgot there's an `auto` :+1:

Ir1d:
btw, do you think it should be `overflow: auto;` instead of x/y ?

tensorboard/plugins/image/tf_image_dashboard/tf-image-loader.html
```diff
@@ -113,6 +113,7 @@
         height: auto;
         position: relative;
         margin: 0 15px 40px 0;
+        overflow-y: auto;
```
stephanwlee: Actually, thanks for pointing it out, this should be `overflow-x`. In y-axis, the page grows infinitely so there should be no scrollbar. If I am understanding your screenshots correctly, you probably had `overflow-x`.

PR #2171, opened 2019-04-30
Fix issue where precomputed knn aren't being reused

### LINKED ISSUE #2082: k-nearest-neighbors caching is broken (unsound)
Missed this in my review of #1901. The cache key check is incorrect. The
code reads:

https://github.com/tensorflow/tensorboard/blob/9270699784657b69a21f022dce98c439f7602d86/tensorboard/plugins/projector/vz_projector/data.ts#L444-L447

where the `nNeighbors <= this.nearest.length` is intended to check
whether the cached KNN computation was computed with a value of
`nNeighbors` not smaller than the currently requested value. But in fact
`this.nearest.length` is the number of data points, not the number of
neighbors computed for each point.

To verify, add

```typescript
    console.log(
        `Found ${nNeighbors}-nearest: shape is ` +
        `(${this.nearest.length}, ${this.nearest[0].length})`
    );
```

below line 384 in `projectUmap`. Then, run UMAP, and subsequently re-run
UMAP with a higher value of *k*. The output is (e.g.)

```
Found 15-nearest: shape is (1024, 15)
Found 23-nearest: shape is (1024, 15)
```

which is wrong, because after attempting to find the 23-nearest
neighbors we only have 15 elements for each data point.

I’m not sure why this never hits a hard error anywhere in the
pipeline—implicit conversion of `undefined` to `0`/`NaN` somewhere?—but
it definitely causes observable effects. To observe, patch in #2080\* to
prevent t-SNE from running automatically when its tab is selected, then:

  - in one tab, load the projector page and run t-SNE with
    perplexity=100;

  - in another tab, load the projector page and run t-SNE with
    perplexity=8, then re-run it with perplexity=100.

These two tabs _should_ yield approximately the same projection, but
instead the projection in the first tab converges to a normal result
whereas the projection in the second tab diverges to points far apart
with little visible structure:

Screenshot of t-SNE okay as described above

Screenshot of t-SNE diverged as described above

\* Tested at commit b0310cd76bd84038a72b1df40c4711a12b959951.

cannoneyed:
Fixes #2082.

### DIFF
--- a/tensorboard/plugins/projector/vz_projector/data.ts
+++ b/tensorboard/plugins/projector/vz_projector/data.ts
@@ -358,17 +358,16 @@ export class DataSet {
     const knnComputation = this.computeKnn(sampledData, k)
 
     knnComputation.then(nearest => {
-      this.nearest = nearest;
       util.runAsyncTask('Initializing T-SNE...', () => {
-            this.tsne.initDataDist(this.nearest);
-          }).then(step);
+        this.tsne.initDataDist(nearest);
+      }).then(step);
     });
   }
 
   /** Runs UMAP on the data. */
   async projectUmap(
-    nComponents: number, 
-    nNeighbors: number, 
+    nComponents: number,
+    nNeighbors: number,
     stepCallback: (iter: number) => void) {
     this.hasUmapRun = true;
     this.umap = new UMAP({nComponents, nNeighbors});
@@ -376,23 +375,23 @@ export class DataSet {
     let currentEpoch = 0;
     const epochStepSize = 10;
     const sampledIndices = this.shuffledDataIndices.slice(0, UMAP_SAMPLE_SIZE);
-  
+
     const sampledData = sampledIndices.map(i => this.points[i]);
     // TODO: Switch to a Float32-based UMAP internal
     const X = sampledData.map(x => Array.from(x.vector));
 
-    this.nearest = await this.computeKnn(sampledData, nNeighbors);
-    
+    const nearest = await this.computeKnn(sampledData, nNeighbors);
+
     const nEpochs = await util.runAsyncTask('Initializing UMAP...', () => {
-      const knnIndices = this.nearest.map(row => row.map(entry => entry.index));
-      const knnDistances = this.nearest.map(row => 
+      const knnIndices = nearest.map(row => row.map(entry => entry.index));
+      const knnDistances = nearest.map(row =>
         row.map(entry => entry.dist)
       );
 
       // Initialize UMAP and return the number of epochs.
       return this.umap.initializeFit(X, knnIndices, knnDistances);
     }, UMAP_MSG_ID);
-    
+
     // Now, iterate through all epoch batches of the UMAP optimization, updating
     // the modal window with the progress rather than animating each step since
     // the UMAP animation is not nearly as informative as t-SNE.
@@ -404,9 +403,9 @@ export class DataSet {
         for (let i = 0; i < epochsBatch; i++) {
           currentEpoch = this.umap.step();
         }
-        const progressMsg = 
+        const progressMsg =
         `Optimizing UMAP (epoch ${currentEpoch} of ${nEpochs})`;
-        
+
         // Wrap the logic in a util.runAsyncTask in order to correctly update
         // the modal with the progress of the optimization.
         util.runAsyncTask(progressMsg, () => {
@@ -416,7 +415,7 @@ export class DataSet {
             const result = this.umap.getEmbedding();
             sampledIndices.forEach((index, i) => {
               const dataPoint = this.points[index];
-    
+
               dataPoint.projections['umap-0'] = result[i][0];
               dataPoint.projections['umap-1'] = result[i][1];
               if (nComponents === 3) {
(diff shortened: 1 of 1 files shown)


wchargin:
@manivaradarajan: Let’s make sure to get this in for the 1.14 release.

tensorboard/plugins/projector/vz_projector/data.ts
```diff
@@ -444,9 +443,14 @@ export class DataSet {
   private async computeKnn(
     data: DataPoint[], 
     nNeighbors: number): Promise<knn.NearestEntry[][]> {
-    if (this.nearest != null && nNeighbors <= this.nearest.length) {
-      // We found the nearest neighbors before and will reuse them.
-      return Promise.resolve(this.nearest);
+    
+    // Handle the case where we've previously found the nearest neighbors.
+    const previouslyComputedNNeighbors = this.nearest ? 
+      this.nearest[0].length : 0;
```
wchargin: In the edge case that the dataset is empty, this will crash with “Cannot
read property `length` of `undefined`”, where the correct behavior is
presumably to return `[]`.

tensorboard/plugins/projector/vz_projector/data.ts
```diff
@@ -381,11 +380,11 @@ export class DataSet {
...
     const X = sampledData.map(x => Array.from(x.vector));
 
-    this.nearest = await this.computeKnn(sampledData, nNeighbors);
+    const nearest = await this.computeKnn(sampledData, nNeighbors);
     
     const nEpochs = await util.runAsyncTask('Initializing UMAP...', () => {
-      const knnIndices = this.nearest.map(row => row.map(entry => entry.index));
-      const knnDistances = this.nearest.map(row => 
+      const knnIndices = nearest.map(row => row.map(entry => entry.index));
+      const knnDistances = nearest.map(row =>
```
wchargin: nit: Trailing whitespace on a bunch of these lines.

tensorboard/plugins/projector/vz_projector/data.ts
cannoneyed: Will `result` computed from `knn.findKNN` exist? There's a ton of `any` types floating around here, which makes it hard to reason about what the state of this  value ought to be. Anyway, I think I've changed it such that this edge issue won't sink things.

cannoneyed:
@wchargin I wound up fixing another knn-related bug in this PR (that @nsthorat found while debugging the standalone release) - You'll probably want to give it another look.

tl;dr is there was some slight drift between the `umap.d.ts` file and the umap implementation's function signature that was causing an issue with small dataset sizes.

tensorboard/plugins/projector/vz_projector/data.ts
```diff
@@ -358,41 +358,43 @@ export class DataSet {
...
-    const epochStepSize = 10;
     const sampledIndices = this.shuffledDataIndices.slice(0, UMAP_SAMPLE_SIZE);
-  
     const sampledData = sampledIndices.map(i => this.points[i]);
-    // TODO: Switch to a Float32-based UMAP internal
     const X = sampledData.map(x => Array.from(x.vector));
 
-    this.nearest = await this.computeKnn(sampledData, nNeighbors);
-    
+    nNeighbors = Math.min(X.length, nNeighbors)
```
wchargin: What are the implications of this change? If `X.length < nNeighbors`
(which can only happen on small datasets), did we previously crash at
some point down the pipeline, or yield wrong results, or something else?
Is this a separate bug?

tensorboard/plugins/projector/vz_projector/data.ts
wchargin: > Will result computed from `knn.findKNN` exist?

I’d assume that it would exist and equal `[]`, and from a glance at the
source this looks accurate (but I could be wrong). In any case, your
patch looks fine; thanks!

wchargin:
Ah, didn’t see your comment before starting the review. Thanks for the
context. Please separate the other fix into another PR.

cannoneyed:
OK cool, I'll move the other fix into a separate PR

cannoneyed:
New PR opened [here](https://github.com/tensorflow/tensorboard/pull/2200)

tensorboard/plugins/projector/vz_projector/data.ts
cannoneyed: I've actually removed this from the other PR - it's essentially a noop (umap does this internally) but it seemed like a good idea at the time.

tensorboard/plugins/projector/vz_projector/data.ts
```diff
@@ -358,41 +358,43 @@ export class DataSet {
...
-      const knnIndices = this.nearest.map(row => row.map(entry => entry.index));
-      const knnDistances = this.nearest.map(row => 
+      const knnIndices = nearest.map(row => row.map(entry => entry.index));
+      const knnDistances = nearest.map(row =>
         row.map(entry => entry.dist)
       );
 
       // Initialize UMAP and return the number of epochs.
-      return this.umap.initializeFit(X, knnIndices, knnDistances);
+      this.umap.setPrecomputedKNN(knnIndices, knnDistances);
```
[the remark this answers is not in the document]
cannoneyed: Moved to [other PR](https://github.com/tensorflow/tensorboard/pull/2200)

PR #2172, opened 2019-04-30
Polymer2 migration 

### LINKED ISSUE #251: Polymer 2.0
Hi!
I am learning how the code is constructed. I think the html is applying the tool of Polymer. It seems like Polymer 2.0 is available, is there any plan to move from the current implementation to polymer 2.0?
Thanks!

stephanwlee:
Polymer1 is ancient and we would like to migrate off to Polymer2.
The major changes with the new version are usages of v1 CustomElements  and Shadow DOM specs (Polymer1 was largely relying on [polyfills](https://github.com/webcomponents/webcomponentsjs) of v0 specs). With the v1 spec, the notable impacts are below:

### 1. CSS is encapsulated without using mucking around with class names. Some other notable changes are [here](https://polymer-library.polymer-project.org/2.0/docs/upgrade#shadow-dom-styles)
Impact
  - move `<style>` tag inside `<template>`
  - `:host.selector` -> `:host(.selector)`

### 2. Proper Shadow DOM 
Impact
  - window.querySelector cannot find elements inside shadowRoot. Replace it with `this.shadowRoot.querySelector` or `this.root.querySelector` (latter is Polymer specific). This largely impacts component that used querySelector for inter-component communication (many examples in the graph plugin).
  - eventing and targetting changed. `event.target` now points at an Custom Element instead of a specific node inside a shadowRoot. This difference made the largest impact on Plottable and its interaction. Please see [its issue](https://github.com/palantir/plottable/issues/3350) for more information.
  - child component, even when using `dom-if` is mounted onto DOM tree right away. This relates to the lifecycle changes. To exemplify what it actually looks like
```html
<component-a>
  <shadow-root>
    <!-- some component-a structure -->
    <slot name="foo"></slot> <!-- this "points" to component-b -->
  </shadow-root>
  <component-b slot="foo">
    <!-- component-b structure -->
  </component-b>
</component-a>
```
  - SVG definitions do not work across shadowRoots.

#### 3.1 Users of `<slot>` requires an attribute `slot="name"`: [Polymer doc](https://polymer-library.polymer-project.org/2.0/docs/upgrade#dom-template)  
Impact
  - Requires mechanical changes

#### 3.2 [Lifeycycle changes](https://polymer-library.polymer-project.org/2.0/docs/about_20#lifecycle-changes)
Impact
  - The most notable change here is the `attached` and `detached`. Previously, in v1, `attached` got triggered when a component is mounted onto the DOM tree. In v2, components are mounted onto DOM tree right away and the difference largely impacts component that tries to defer fetch/render based on visibility of the component. 
  - Another subtle but concrete change has to do with when observers get invoked. Previously, if a function depends on property A and B, until both A and B has a value (non-undefined), the observer did not get called. Now, in Polymer2, there is no such guarantee and setting a value to A can invoke the function before property B has a chance to get a value (through computed function). Polymer doc describes it as below:
> batched data changes, which can improve performance as well as correctness.
> Undefined dependency checks for observers, computed bindings and computed properties have been removed. These are all called once at initialization time.

-----------------------------
Tips for reviewer:
Some changes make bunch of whitespace changes. I'd recommend turning on "Ignore whitespace changes".
Also, please review commit by commit. 

Fixes #251


stephanwlee:
Assigning to @nfelt as per our rotation :). Nick, please let me know if you think others can do a better job at reviewing.

stephanwlee:
Closing it in favor of more modular PRs

PR #2175, opened 2019-05-01
Fix WIT textarea border issue

jameswex:

WIT's example viewer textarea's had incorrect border styling.

Added css rules to not use the -appearance rules for textareas and instead allow our custom styling
Reorganized existing rules around these elements to be a little simpler

* Screenshots of UI changes
KpLS84vsueC

Ran demos, saw correct visuals

### DIFF
--- a/tensorboard/components/vz_example_viewer/vz-example-viewer.html
+++ b/tensorboard/components/vz_example_viewer/vz-example-viewer.html
@@ -106,7 +106,11 @@ limitations under the License.
       }
 
       iron-autogrow-textarea {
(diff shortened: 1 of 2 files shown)


tensorboard/components/vz_example_viewer/vz-example-viewer.html
```diff
@@ -106,7 +106,11 @@
       }
 
       iron-autogrow-textarea {
+        border: 1px solid #FFFFFF;
```
stephanwlee: as per style guide, prefer `#fff`

PR #2178, opened 2019-05-02
github: update issue templates

wchargin:
Summary:
This commit adds issue templates for bug reports, feature requests, and
installation problems, with specific instructions for each.

This also resolves the nag from GitHub to move to the new issue template
workflow.

I created these by following the “issue template wizard” described here:
<https://help.github.com/en/articles/creating-issue-templates-for-your-repository>

Test Plan:
I’ve committed this to the default branch of my fork. Try drafting an
issue there: <https://github.com/wchargin/tensorboard/issues/>.

wchargin-branch: github-issue-templates


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,54 @@
...
+  - Mode of installation (e.g., pip inside virtualenv; Conda):
+  - Python version (e.g., 2.7; 3.5):
+  - For browser-related issues:
+      - Browser type and version (e.g., Chrome 64.0.3282.140):
+      - Screenshot, if it’s a visual issue:
+
+## Issue description
+
+Please describe the bug as clearly as possible. Provide a minimal
+example as a self-contained Python program that reproduces the problem
```
nfelt: It's not always going to be a python program, though.  For example in many cases it could just be launching tensorboard against a particular logdir or with a particular argument.

.github/ISSUE_TEMPLATE/installation_problem.md
```diff
@@ -0,0 +1,63 @@
+---
+name: Installation problem
+about: Get help installing or starting TensorBoard
```
nfelt: Do we want to phrase this as "get help"?  I think it might be better to stick to the clear messaging that our issue tracker should be for real bugs and feature requests, not for general assistance.  Installation issues in particular are often specific to the user, so it's hard to scale supporting those types of concerns.

.github/ISSUE_TEMPLATE/installation_problem.md
```diff
@@ -0,0 +1,63 @@
...
+      - `tensorboard`:
+      - `tb-nightly`:
+      - `tensorflow`:
+      - `tf-nightly`:
+      - `tf-nightly-2.0-preview`:
+  - OS platform and version (e.g., Linux with Ubuntu 16.04):
+  - Mode of installation (e.g., pip inside virtualenv; Conda):
+  - Python version (e.g., 2.7; 3.5):
+
+## Script to reproduce (required)
```
nfelt: Again, I think this may be a little prescriptive.  I'm not sure we would get anyone to follow these instructions who wouldn't already be the type of user who would have been careful to describe the things they tried.

I'd recommend keeping this a bit simpler and just focusing on 1) provide explicit commands you used to reproduce the problem and 2) try a fresh environment in case yours is in a bad state.

wchargin:
> Edited to add: ok, so a downside is that they wouldn't get the latest
> version of the diagnose script.

Right, that was my thinking. It’d be a one-line change to `setup.py` if
we wanted to provide a `diagnose_tensorboard` binary, but we couldn’t
keep that up to date with checks and suggestions.

I’ll merge this PR as is, and would be happy to consider further
improvements to the delivery.

PR #2188, opened 2019-05-03
hparams: add hparam writing for non-Keras users

wchargin:
Summary:
This commit adds `hparams` and `hparams_pb`, to write `SessionStartInfo`
summaries for users who do not use the Keras callback.

As with `hparams_config{,_pb}`, this “conceptually” supports notf, but
there is still a transitive dependency through `metadata`.

In the tests, we opt to duplicate some infrastructure code rather than
abstracting over it. Compared to (e.g.) `scalar/summary_test`, our test
matrix has an extra dimension: in addition to “V1 vs. V2” and “op vs.
pb”, we have “`hparams` vs. `hparams_config`”. I couldn’t find an
abstraction whose code reuse outweighed its extra complexity.

These tests caught a bug present in the existing `summary.py` code,
wherein `bool` hparam values are always serialized as `int`s, because in
Python `bool` is a subtype of `int` and so one of the conditionals was
dead. The new code properly emits `bool` hparams, and so is not strictly
“compatible” with the old code.

Test Plan:
Unit tests included. The existing demo uses the Keras callback and so
does not exercise this code.

wchargin-branch: hparams-writing

### DIFF
--- a/tensorboard/plugins/hparams/summary_v2.py
+++ b/tensorboard/plugins/hparams/summary_v2.py
@@ -22,6 +22,8 @@ from __future__ import division
 from __future__ import print_function
 
 import abc
+import hashlib
+import json
 import time
 
 import six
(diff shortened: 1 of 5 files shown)


tensorboard/plugins/hparams/summary_v2.py
```diff
@@ -101,6 +172,39 @@ def hparams_config_pb(hparams, metrics, time_created_secs=None):
...
+      raise ValueError("multiple values specified for hparam %r" % (k,))
+    result[k] = v
+  return result
+
+
+def _derive_session_group_name(hparams):
+  # Use `json.dumps` rather than `str` to ensure invariance under string
+  # type (incl. across Python versions) and dict iteration order.
+  jparams = json.dumps(hparams, sort_keys=True, separators=(",", ":"))
+  return hashlib.sha256(jparams.encode("utf-8")).hexdigest()
```
nfelt: We don't really need a full 256 bits of collision avoidance, do we?  N bits gives us about sqrt(N) uses without a collision, so it seems like 128 or fewer bits would be more than enough in our case.

Just thinking that 64 characters is rather long if this If this automatic session name gets exposed in the UI anywhere.  Although I suppose we could always attempt to heuristically detect that it's a hash digest and then abbreviate it.

tensorboard/plugins/hparams/summary_v2_test.py
```diff
@@ -43,6 +43,164 @@
...
+    self._check_logdir(self.logdir)
+
+  def test_graph_mode(self):
+    with \
+        tf.compat.v1.Graph().as_default(), \
+        tf.compat.v1.Session() as sess, \
+        tf.compat.v2.summary.create_file_writer(self.logdir).as_default() as w:
+      sess.run(w.init())
+      summ = hp.hparams(self.hparams, start_time_secs=self.start_time_secs)
+      self.assertTrue(sess.run(summ))
```
nfelt: You probably want `sess.run(w.flush())` after running the summary ops.  Without it I would expect this test to be flaky.

Or you can set `max_queue` on the writer to 0.

tensorboard/plugins/hparams/summary_v2.py
wchargin: > We don't really need a full 256 bits of collision avoidance, do we? N
> bits gives us about sqrt(N) uses without a collision, so it seems like
> 128 or fewer bits would be more than enough in our case.

$\sqrt(2^N)$, yeah. It’s true that 32 bits suffice for a 99.8% chance of
perfect hashing with up to 4096 hyperparameter configurations, which is
already pushing the performance boundaries of TensorBoard.

> Just thinking that 64 characters is rather long if this If this
> automatic session name gets exposed in the UI anywhere.

It is exposed in the table view, but I have been thinking of removing
that. All the examples that I’ve seen have either a hash or an
incrementing integer their, neither of which is useful. If we remove
that, then we don’t have to have “session groups” or “trials” or
anything in the UI at all.

Note also that these strings in the table view are already truncated
(`text-overflow: ellipsis;`), so truncating from 64- to 32-character
strings in the data wouldn’t actually make an observable difference.
Also, we really shouldn’t shove *either* 32-character hex strings or
64-character hex strings into users’ faces; cutting the length in half
doesn’t actually solve the problem imho. So, given that each session
gropu name appears on disk exactly once across all event files (unlike
tag names, for instance), I’m inclined to keep it as is.

> Although I suppose we could always attempt to heuristically detect
> that it's a hash digest and then abbreviate it.

If we ever want to show these names, this is fine with me.

tensorboard/plugins/hparams/summary_v2_test.py
wchargin: I thought that exiting the context manager would flush the writer
automatically. How is this context different from
<https://github.com/tensorflow/tensorboard/pull/2139#discussion_r278677673>?
Even though this one is in legacy graph mode, both use the `.as_default`
method on a V2 summary writer object, right?

wchargin: Hmm—from a bit of experimentation, it is _not_ guaranteed to flush when
the writer context manager exits, but _does_ always seem to flush when
the session context manager exits. That’s sufficiently non-obvious that
I’ll go ahead and add the explicit `flush()`, though I would still be
interested to understand why this differs from the situation linked in
my previous comment.

PR #2192, opened 2019-05-06
Allow WitWidget example changing

jameswex:

To support upcoming WIT integrations, made it possible to change the examples being shown by WitWidget through a method call.

- Changed WitWidget private _set_examples method to public set_examples method.
- Changed colab implementation of set_examples to use BroadcastChannel to allow it to be called in a different colab cell than the one that displays WIT. This is necessary because colab sandboxes each output cell in iframes.

* Screenshots of UI changes

N/A

Installed WitWidget with these changes in colab and jupyter and ran test notebook that displayed initial examples, then in another cell, calls set_examples to change which examples are shown. Verified correct behavior in WIT.

### DIFF
--- a/tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
+++ b/tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
@@ -158,7 +158,8 @@ WIT_HTML = """
     }};
     // BroadcastChannel to allow examples to be updated by a call from an
     // output cell that isn't the cell hosting the WIT widget.
-    const updateExampleListener = new BroadcastChannel('updateExamples');
+    const channelName = 'updateExamples' + id;
+    const updateExampleListener = new BroadcastChannel(channelName);
     updateExampleListener.onmessage = msg => {{
       window.updateExamplesCallback(msg.data);
     }};
@@ -185,6 +186,7 @@ class WitWidget(base.WitWidgetBase):
       height: Optional height in pixels for WIT to occupy. Defaults to 1000.
     """
     self._ctor_complete = False
+    self.id = WitWidget.index
     base.WitWidgetBase.__init__(self, config_builder)
     # Add this instance to the static instance list.
     WitWidget.widgets.append(self)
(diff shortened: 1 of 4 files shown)


jameswex:
@tolga-b please review

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
```diff
@@ -157,6 +156,12 @@ def infer_mutants(wit_id, details):
         window.spriteCallback(wit.localAtlasUrl);
       }}
     }};
+    // BroadcastChannel to allow examples to be updated by a call from an
+    // output cell that isn't the cell hosting the WIT widget.
+    const updateExampleListener = new BroadcastChannel('updateExamples');
```
tolga-b: It seems that every instance of WIT is listening to the same broadcast channel. In case there are multiple WITs in the same colab, this would cause calling set_examples on one of them to update examples on all windows. A fix is to append wit id to channel name 'UpdateExamples_{}'.format(id).

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
```diff
@@ -179,29 +184,43 @@ def __init__(self, config_builder, height=1000):
...
+    output.eval_js("""updateExamplesCallback({examples})""".format(
+      examples=json.dumps(self.examples)))
     self._generate_sprite()
+    self._ctor_complete = True
 
   def _get_element_html(self):
     return """
       <link rel="import" href="/nbextensions/wit-widget/wit_jupyter.html">"""
 
+  def set_examples(self, examples):
```
tolga-b: I think we need to keep the wit id in self at init so that this method can call the correct broadcast channel.

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
jameswex: great catch, fixed

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
jameswex: fixed

PR #2197, opened 2019-05-06
Fix mesh visualizer

stephanwlee:
There were two problems:
1. dataProvider is not available until component is attached
2. missing dependency to tf-mesh-loader in mesh-dashboard

One minor feature improvements:
- Runs selector now appears below the Point of view section and it now
  spans the entire view height.

image

+cc @podlipensky

### DIFF
--- a/tensorboard/plugins/mesh_visualizer/tf_mesh_dashboard/mesh-loader.js
+++ b/tensorboard/plugins/mesh_visualizer/tf_mesh_dashboard/mesh-loader.js
@@ -139,7 +139,7 @@ Polymer({
    * Function to call when component must be reloaded.
    */
   reload: function() {
-    if (!this.active) {
+    if (!this.active || !this.isAttached) {
       return;
     }
     this.set('_isMeshLoading', true);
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/mesh_visualizer/tf_mesh_dashboard/mesh-loader.js
```diff
@@ -139,7 +139,7 @@ Polymer({
    * Function to call when component must be reloaded.
    */
   reload: function() {
-    if (!this.active) {
```
wchargin: Did you consider guarding for `this._dataProvider` itself, rather than
the lifecycle methods? No change necessarily requested; mostly wondering
if there’s a reason for the indirection.

stephanwlee: I actually did think about that but I am torn.

On one hand: `reload` does make much sense without component being mounted.
On the other hand, both `active` and `_dataProvider` should be part of the observer AND guard against the undefinedness (in Polymer2+, there is no guarantee that all props are initialized before invoking the observer).

stephanwlee: Decided to take the second "hand" :)

PR #2198, opened 2019-05-06
Fix word-break issue on metadata card

cannoneyed:
Long strings in metadata were incorrectly overflowing the metadata card. Since the element is styled as a table, we need to use word-break rather than overflow-wrap.

**before**
Screen Shot 2019-05-06 at 3 14 42 PM

**after**
Screen Shot 2019-05-06 at 3 14 50 PM

### DIFF
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-metadata-card.html
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-metadata-card.html
@@ -69,7 +69,8 @@ limitations under the License.
 }
     
 .metadata-value {
-  overflow-wrap: break-word;
+  word-wrap: anywhere; /* Firefox only -- word-wrap DNE in Chrome. anywhere DNE in Chrome */
+  word-break: break-word; /* break-word DNE in Firefox */
 }
 </style>
 


stephanwlee:
> Since the element is styled as a table, we need to use word-break rather than overflow-wrap

I am not too familiar with this perhaps the `table` wasn't too popular when I actively developed. Would you have a reference handy?

Also, would setting a concrete width on the table help?

cannoneyed:
Yeah I'm not really that familiar with this styling approach either... had to do quite a bit of digging / experimenting to finally figure out how to make it work. Sadly I don't have a great reference for exactly what's going on here, but there are a few things I found that led me to this solution 9https://gist.github.com/prabhuramkumar/4491951)

Also, setting a concrete width on the table does NOT work, unfortunately (which was very frustrating)

stephanwlee:
Can you please provide me with a sample data I can use to reproduce the bug? I cannot reproduce this by simply patching the DOM, thanks!

cannoneyed:
Here's what I did to address it: Add the following to `/tensorboard/plugins/projector/vz_projector/standalone_projector_config.json`

```json
    {
      "bookmarksPath": "http://projector.tensorflow.org/oss_data/bnmt_enjako_interlingua_10k_1024_bookmarks.txt",
      "metadataPath": "http://projector.tensorflow.org/oss_data/bnmt_enjako_interlingua_10k_1024_metadata.tsv",
      "tensorName": "GNMT Interlingua",
      "tensorPath": "http://projector.tensorflow.org/oss_data/bnmt_enjako_interlingua_10k_1024_tensors.bytes",
      "tensorShape": [9978, 1024]
    }
```

tensorboard/plugins/projector/vz_projector/vz-projector-metadata-card.html
```diff
@@ -69,7 +69,7 @@
 }
     
 .metadata-value {
-  overflow-wrap: break-word;
+  word-break: break-word;
```
stephanwlee: I am rather appalled at how difficult styling a table is. I understand why but it is crazy. 
Instead of wasting more time, I am okay with going forward with this change with minor change.

Note that this does not work in Firefox. Funnily, `overflow-wrap: break-word` does not work either. Perhaps we want:
```css
word-wrap: anywhere; /* Firefox only -- word-wrap DNE in Chrome. anywhere DNE in Chrome */
word-break: break-word; /* break-word DNE in Firefox */
```

PR #2200, opened 2019-05-06
Fix issue where we werent initializing knn properly

cannoneyed:
While debugging the standalone projector app, we discovered a bug caused by the divergence between the umap-js and the `umap.d.ts` typings that we maintain in this repo. This bug caused an error when initializing very small datasets. This PR fixes the incorrect usage of `umap.initializeFit` and updates the types to match the library.


wchargin:
Missing the actual `Math.min` change?

cannoneyed:
I mentioned it in a comment on the last thread, but the `Math.min` change is technically a noop, so I opted just to not bring it along while setting up this PR.

PR #2202, opened 2019-05-07
hparams: manually flush writer in graph mode tests

wchargin:
Summary:
The test as written may be flaky, according to @nfelt’s comment here:
<https://github.com/tensorflow/tensorboard/pull/2188#discussion_r281431612>

Test Plan:
I couldn’t actually get the previous version of the test to fail (even
with `--runs_per_test=1000`), but it still passes, so that’s nice.

wchargin-branch: hparams-tests-explicit-flush


wchargin:
Okay, I see (kind of). Thanks for the elaboration.

PR #2206, opened 2019-05-07
Mesh Plugin demo app

podlipensky:
Demo app for the mesh plugin.


tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,99 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+
+import numpy as np
+import tensorflow as tf
+import trimesh
```
stephanwlee: this dependency does not exist on our project I believe. Perhaps catch the failure and prompt user to install trimesh to run this? @nfelt would have better idea than mine :)

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,99 @@
...
+tf.flags.DEFINE_string('sample_meshes', '',
+                       'Path to folder with PLY files or string of comma '
+                       'separated PLY files to visualize.')
+tf.flags.DEFINE_integer('batch_size', 1,
+                        'Size of the batch. All provided PLY files will be '
+                        'divided into batch_size groups and used at different '
+                        'steps during training.')
+FLAGS = tf.flags.FLAGS
+
+
```
stephanwlee: here, I'd add `tf.compat.v1.disable_v2_behavior()` for interop with TensorFlow 2.0 preview

wchargin:
@podlipensky, could you please make new commits from master rather than
trying to merge in 27 commits, most of which are already integrated? To
fix this PR, you can rebase and force-push.

tensorboard/plugins/mesh/mesh_demo.py
podlipensky: can we add it to tensorboard or you'd rather not?
Also do you think I need build rule for demo app?

PR #2207, opened 2019-05-07
hparams: update tutorial to use new APIs

wchargin:
Summary:
Until now, the hparams tutorial notebook has directly manipulated protos
and logged them with `experimental.write_raw_pb`. Now that the
high-level APIs have merged, we can simplify the tutorial significantly.

Test Plan:
Uploading this notebook to Colab, resetting all runtimes, and pressing
“Run All” shows that it behaves correctly. Staging to the internal dev
site shows that the images and code render correctly.

wchargin-branch: hparams-tutorial


docs/r2/hyperparameter_tuning_with_hparams.ipynb
```diff
@@ -227,113 +206,74 @@
         "2. Dropout rate in the dropout layer\n",
         "3. Optimizer\n",
         "\n",
-        "List the values to try:"
+        "List the values to try, and log an experiment configuration to TensorBoard. This step is optional: it enables you to customize how the hyperparameters and metrics appear in the UI, and which metrics to include."
```
GalOshri: Minor: is it about customizing how they appear in the UI (which implies multiple options) or about providing additional detail so the UI can filter more precisely?

docs/r2/hyperparameter_tuning_with_hparams.ipynb
```diff
@@ -343,169 +283,187 @@
...
     },
     {
+      "cell_type": "markdown",
       "metadata": {
-        "id": "u2nOgIKAXdcb",
+        "id": "2mYdW0VKLbFE",
         "colab_type": "text"
       },
+      "source": [
+        "When training Keras models, you can use callbacks instead of writing these directly:\n",
```
GalOshri: Minor: this code snippet looks a bit awkward in GitHub's notebook rendering, but maybe it's the best option.

docs/r2/hyperparameter_tuning_with_hparams.ipynb
```diff
@@ -533,22 +491,22 @@
...
+      "cell_type": "code",
       "metadata": {
         "id": "oxrSUAnCeFmQ",
         "colab_type": "code",
         "colab": {}
       },
-      "cell_type": "code",
       "source": [
         "%%bash\n",
         "wget -q 'https://storage.googleapis.com/download.tensorflow.org/tensorboard/hparams_demo_logs.zip'\n",
```
GalOshri: There is no reason to update this set of logs, right?

docs/r2/hyperparameter_tuning_with_hparams.ipynb
wchargin: I’d checked the rendering in the TensorFlow.org staging server and in Colab
before sending the PR. On the docs site, the code block is rendered the
same as other code blocks:

Screenshot of rendering in docs site

In Colab, it’s more similar to GitHub, in that the cell background isn’t
shaded as much as executable code cells:

Screenshot of rendering in Colab

GitHub, for comparison:

Screenshot of rendering in GitHub

I don’t mind any of the styles, personally. I appreciate the ones that
clearly distinguish executable code from sample-only code (i.e., Colab
and GitHub), but I think that the literal ellipsis is clear enough that
there won’t be any actual confusion.

If you have suggestions or concerns, I’m all ears.

docs/r2/hyperparameter_tuning_with_hparams.ipynb
wchargin: It’s both. The display name just customizes the UI; the domains provide
additional detail for more precise filtering. I was going for concision,
but I like your phrasing, and it’s probably wise to avoid conveying the
impression that this is _only_ for cosmetic changes. Will fix—thanks!

docs/r2/hyperparameter_tuning_with_hparams.ipynb
wchargin: That was my conclusion. The purpose of those logs is to explore the
dashboard, not to correspond exactly to the experiment code in this
notebook (and they already didn’t). It’s fine if there continue to be
minor deviations.

docs/r2/hyperparameter_tuning_with_hparams.ipynb
GalOshri: It looks good on the TensorFlow.org staging site, so I'm good with this approach. Thank you for exploring it across the different surfaces!

docs/r2/hyperparameter_tuning_with_hparams.ipynb
GalOshri: Sounds good.

PR #2208, opened 2019-05-07
mesh: explictly define public API, for consistency

wchargin:
Summary:
Amends #2205 to be consistent with other plugins.

Test Plan:
Pip package smoke test still passes.

wchargin-branch: mesh-library-target


wchargin:
@nfelt: I’d love to, but unfortunately the smoke test is currently
broken in a way that makes that test unfalsifiable. See #2211 for the
fix.

Issue #1907, opened 2019-02-26
warn if both `tensorboard` and `tb-nightly` pip packages are installed


nfelt:
This is a common source of inscrutable errors, because pip will happily install both pip packages even though they ship the tensorboard python package [1], and in doing so the second one will overwrite any files from the first one: https://github.com/pypa/pip/issues/4625.  This means you can easily end up with a garbled frankentensorboard package (and it gets even worse if you uninstall one of them but not both of them, since then you get a frankentensorboard with a lot of vital organs removed).

Until pip fixes this mess, it's a fact of life, but we can at least try to warn users about this pitfall ourselves.

[1] It's colloquial to call both of these packages, so I'm prefixing to avoid confusion.  See also https://github.com/pypa/packaging.python.org/issues/106

nfelt:
For an example of this issue happening, see https://github.com/tensorflow/tensorboard/issues/1862

nfelt:
We could warn about this at runtime if we used the `pkg_resources` module from setuptools to check active distributions (aka pip packages).  I think this would be something like `pkg_resources.find(pkg_resources.Requirement("tensorboard"))`.

Even better would be to do this check at installation time.  I think this may be possible via the `cmdclass` parameter to `setup()` by passing in a custom install command, see: 
https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools
https://blog.niteo.co/setuptools-run-custom-code-in-setup-py/

Either way we'd want to check reasonably thoroughly to make sure we don't have false positives.  It's also TBD whether to just log a warning or whether to fail entirely.  I'm inclined towards the latter but it might be good to start out with the former just in case there are any false positive installation scenarios out there.  (We could always offer some kind of escape hatch if you get a false positive, like setting an environment variable that disables the check.)

nfelt:
Another example of this happening: https://github.com/tensorflow/tensorboard/issues/2092

wchargin:
Note that you get in this state if you upgrade to the TensorFlow 2.0
alpha using the recommended incantation from the [Dev Summit talk]:

```
$ virtualenv -q ./ve
$ . ./ve/bin/activate
(ve) $ pip install tensorflow==1.13.1 >/dev/null 2>&1
(ve) $ pip install -U --pre tensorflow >/dev/null 2>&1
(ve) $ pip freeze | grep -e tensor -e tf- -e tb-
[...]

PR #2211, opened 2019-05-07
pip: fix smoke test double-TensorBoard dependency

wchargin:
Summary:
The smoke test doesn’t actually catch those things that it’s supposed to
catch, because it installs `tf-nightly` (or `tf-nightly-2.0-preview`),
which in turn installs `tb-nightly`. We then install the just-built
version of `tensorboard`, so the resulting virtualenv has both
`tensorboard` and `tb-nightly`: #1907 strikes again.

When `--tf-version ""` is passed, this is not an issue, but some smoke
checks only run when that is not passed.

It just so happens that this is currently benign: each of the direct
dependencies of `:build_pip_package` either has no smoke test, contains
an `__init__.py` file that is required to overwrite the Bazel-generated
empty one, because the smoke test references symbols directly in that
file, or is not yet in `tb-nightly`. But (a) starting with tomorrow’s
nightly, the mesh smoke test will meet none of those conditions, so its
smoke test will cease to be effective, and (b) I’d like to add an
hparams smoke test, which already falls into the same boat.

Test Plan:
Add an hparams smoke test, but remove the actual hparams dep:

```diff
diff --git a/tensorboard/pip_package/BUILD b/tensorboard/pip_package/BUILD
index c11e22e2..e63b790a 100644
--- a/tensorboard/pip_package/BUILD
+++ b/tensorboard/pip_package/BUILD
@@ -21,7 +21,6 @@ sh_binary(
         "//tensorboard:lib",  # User-facing overall TensorBoard API
         "//tensorboard:version",  # Version module (read by setup.py)
         "//tensorboard/plugins/beholder",  # User-facing beholder API
-        "//tensorboard/plugins/hparams",  # User-facing hparams API
         "//tensorboard/plugins/mesh:summary",  #  User-facing mesh summary API
         "//tensorboard/plugins/projector",  # User-facing projector API
         "//tensorboard/summary:tf_summary",  #  tf.summary API for TF 2.0
diff --git a/tensorboard/pip_package/build_pip_package.sh b/tensorboard/pip_package/build_pip_package.sh
index de778196..113724ef 100755
--- a/tensorboard/pip_package/build_pip_package.sh
+++ b/tensorboard/pip_package/build_pip_package.sh
@@ -172,6 +172,7 @@ import tensorboard as tb
 tb.summary.scalar_pb('test', 42)
 from tensorboard.plugins.beholder import Beholder, BeholderHook
 from tensorboard.plugins.mesh import summary
+from tensorboard.plugins.hparams.api import hparams_pb, KerasCallback
 "
   fi

```

Then run

    bazel run //tensorboard/pip_package:build_pip_package

and note that the smoke test passes (incorrectly) before this commit and
fails (correctly) after it.

Note also that the `pip freeze` output from the smoke test includes both
`tb-nightly` and `tensorboard` before this change (e.g., on recent
Travis logs), but only `tensorboard` after it.

wchargin-branch: pip-smoke-2dep

### DIFF
--- a/tensorboard/pip_package/build_pip_package.sh
+++ b/tensorboard/pip_package/build_pip_package.sh
@@ -141,8 +141,8 @@ smoke() {
 
   if [ -n "${smoke_tf}" ]; then
     pip install -qU "${smoke_tf}"
+    pip uninstall -qy tensorboard tb-nightly  # Drop any conflicting packages
   fi
-  pip uninstall -qy tensorboard tb-nightly  # Drop any conflicting packages
   pip install -qU ../dist/*"py${py_major_version}"*.whl >/dev/null
   pip freeze  # Log the results of pip installation
 
--- a/tensorboard/pip_package/build_pip_package.sh
+++ b/tensorboard/pip_package/build_pip_package.sh
@@ -142,6 +142,7 @@ smoke() {
   if [ -n "${smoke_tf}" ]; then
     pip install -qU "${smoke_tf}"
   fi
+  pip uninstall -qy tensorboard tb-nightly  # Drop any conflicting packages
   pip install -qU ../dist/*"py${py_major_version}"*.whl >/dev/null
   pip freeze  # Log the results of pip installation
 


tensorboard/pip_package/build_pip_package.sh
```diff
@@ -142,6 +142,7 @@ smoke() {
   if [ -n "${smoke_tf}" ]; then
     pip install -qU "${smoke_tf}"
   fi
+  pip uninstall -qy tensorboard tb-nightly  # Drop any conflicting packages
```
nfelt: You might as well do this one line up, inside the case where we install the TF package?  Otherwise this is a fresh virtualenv so there's no need to do the uninstall.

wchargin: Sure; fine with me.

PR #2213, opened 2019-05-07
hparams: avoid TensorFlow dependency in summary_v2

wchargin:
Summary:
This changes the private `metadata` module to emit TensorBoard protos
instead of TensorFlow protos, removing its dependency on TensorFlow. Its
two clients are updated: we add a reinterpret-cast to one and remove the
cast from the other.

Test Plan:
A new test target runs with `//tensorboard/compat:no_tensorflow`.
Verified that this passes in a virtualenv that actually does not have
TensorFlow installed, skipping only the expected six tests. Smoke test
updated for notf builds; the new smoke test fails before this change.

wchargin-branch: hparams-notf


wchargin:
WANT_LGTM=any

PR #2216, opened 2019-05-08
Move What-If Tool into an iframe in Jupyter widget

jameswex:

In order for WitWidget to live in a Jupyter notebook without interfering with other widgets that also use Polymer, we need to load it inside an iframe. Otherwise for examples the TFMA visualizations and WitWidget cannot live in the same notebook without failing, due to their vulcanized HTML files for their widgets both containing polymer dependencies (since the browser will fail on the loading of those deps a second time).

Moved WIT HTML element and its vulcanized dependency to inside an iframe.
Updated wit.js code to ensure WIT is fully loaded inside iframe before trying to use it.
Update WIT element to use correct font, as without this, inside iframe it reverts to default browser fonts.
Update WIT element to auto-infer when a model is set, if examples have already been loaded. Previously this only happened if a model was set and THEN examples were set. But it should auto-infer when these are set in either order.

* Screenshots of UI changes

No visual changes, but included screenshot of WIT loading inside iframe in Jupyter

witiframe

pip build/install with this PR. Run WIT inside Jupyter notebook with model/data, ensure all functionality and visual correctness.

### DIFF
--- a/tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/lib/wit.js
+++ b/tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/lib/wit.js
@@ -19,6 +19,7 @@ var widgets = require('@jupyter-widgets/base');
 // python backend.
 var WITView = widgets.DOMWidgetView.extend({
   render: function() {
+    this.setupComplete = false;
     // Load up the WIT polymer element.
     this.loadAndCreateWhatIfToolElement();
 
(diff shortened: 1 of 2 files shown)


jameswex:
@tolga-b please review

stephanwlee:
Jumping guns just reading the title: note that this may make What-If tool not usable in Colab. Colab currently cannot handle iframe based plugins.

jameswex:
> Jumping guns just reading the title: note that this may make What-If tool not usable in Colab. Colab currently cannot handle iframe based plugins.

This only effects the Jupyter version. No changes to colab version of WitWidget, which already works in the same notebook as TFMA visualizations, as colab uses iframes for each output cell already.

tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/lib/wit.js
```diff
@@ -36,35 +37,48 @@ var WITView = widgets.DOMWidgetView.extend({
    * Loads up the WIT element.
    */
   loadAndCreateWhatIfToolElement: function() {
+    const height = parseInt(
+      this.model.attributes.layout.attributes.height, 10) - 20;
+    const iframe = document.createElement('iframe');
     const templateLocation =
-        window.__webpack_public_path__ + 'wit_jupyter.html';
+      window.__webpack_public_path__ + 'wit_jupyter.html';
+    const src = `<link rel="import" href="${templateLocation}">
```
wchargin: I expect that this will break very soon. HTML imports are deprecated and
will be removed in Chrome M75 (we’re on M74 right now). TensorBoard’s
HTML imports are compiled statically, so we don’t immediately have this
problem, but it looks like this is using HTML imports at runtime. Have
you tested this on Chrome canary?

Depending on what exactly is in `wit_jupyter.html`, you may be able to
achieve the same behavior by, within the iframe, (1) `fetch`ing that
page, (2) setting the `document.body.innerHTML` to its source, and (3)
rehydrating `<script>` elements. You can see how we do that for the
`%tensorboard` Colab widget here:

<https://github.com/tensorflow/tensorboard/blob/1925d060fbfa71fa8e26d793a94f704b7ba413e3/tensorboard/notebook.py#L314-L364>

PR #2217, opened 2019-05-08
test: skip V1-only tests under TF2 behavior

### LINKED ISSUE #1996: platform: do a better detection of TF2
Even if a user is using tf-nightly-2.0-preview or 2.0 symbols are present, user may want to use the v1 behavior by explicitly calling `tf.compat.v1.disable_v2_behavior` or set `TF2_BEHAVIOR=0`. 

We should try to respect the intent.

wchargin:
Summary:
When the `TF2_BEHAVIOR` environment variable is set, TensorFlow behaves
as if `tf.compat.v1.enable_v2_behavior()` were invoked. In such a case,
we should not run V1-only tests.

We elect to test this by importing a private TensorFlow module rather
than reading the environment variable directly. Both are undocumented,
but the module import will give us a hard failure if it is changed,
while the environment variable will silently change behavior.

This fixes some Google-internal issues (see <http://cl/247294725>). It
should be a no-op externally; we do not set this environment variable.

Resolves #1996.

Test Plan:
Count the number of skipped tests before and after this change:

```
tests="$(bazel query "
    tests(rdeps(
        //tensorboard/...,
        set($(git grep -l run_v._only)) - //tensorboard/util:test_util.py
    ))
")" &&
bazel test ${tests} --test_output=all 2>&1 |
    sed -n -e 's/.*(skipped=\([0-9]\+\)).*/\1/p' |
    awk '{ c += $1 } END { print c }'
```

On TF 2.0, it prints `166` both before and after this change; on TF 1.x,
it prints `21` both before and after this change.

wchargin-branch: respect-tf2-behavior

### DIFF
--- a/tensorboard/util/test_util.py
+++ b/tensorboard/util/test_util.py
@@ -27,6 +27,8 @@ import threading
 import unittest
 
 import tensorflow as tf
+# See discussion on issue #1996 for private module import justification.
+from tensorflow.python import tf2 as tensorflow_python_tf2
 
 from tensorboard.compat.proto import event_pb2
 from tensorboard.compat.proto import graph_pb2
@@ -158,16 +160,10 @@ def _run_conditionally(guard, name, default_reason=None):
 
   return _impl
 
-# Logic from `tensorflow.python.tf2`, assuming that `enable_v2_behavior`
-# has not been manually called (which we have no way to check).
-_is_tf2 = (
-    tf.__version__.startswith('2.')
-    or os.getenv('TF2_BEHAVIOR', '0') != '0'
-)
 run_v1_only = _run_conditionally(
-    lambda: not _is_tf2,
+    lambda: not tensorflow_python_tf2.enabled(),
     name='run_v1_only')
 run_v2_only = _run_conditionally(
-    lambda: _is_tf2,
+    lambda: tensorflow_python_tf2.enabled(),
     name='run_v2_only',
     default_reason='Test only appropriate for TensorFlow v2')
--- a/tensorboard/util/test_util.py
+++ b/tensorboard/util/test_util.py
@@ -22,7 +22,6 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import os
 import threading
 import unittest
 
--- a/tensorboard/util/test_util.py
+++ b/tensorboard/util/test_util.py
@@ -22,6 +22,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import os
 import threading
 import unittest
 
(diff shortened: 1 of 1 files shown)


tensorboard/util/test_util.py
```diff
@@ -157,8 +158,12 @@ def _impl(reason=None):
 
   return _impl
 
-# TODO(#1996): Better detect TF2.0.
-_is_tf2 =  tf.__version__.startswith('2.')
+# Logic from `tensorflow.python.tf2`, assuming that `enable_v2_behavior`
```
nfelt: Can we just use `tensorflow.python.tf2.enabled()` Per https://github.com/tensorflow/tensorboard/issues/1996#issuecomment-490647019

tensorboard/util/test_util.py
```diff
@@ -157,12 +160,10 @@ def _impl(reason=None):
...
 
-# TODO(#1996): Better detect TF2.0.
-_is_tf2 =  tf.__version__.startswith('2.')
 run_v1_only = _run_conditionally(
-    lambda: not _is_tf2,
+    lambda: not tensorflow_python_tf2.enabled(),
     name='run_v1_only')
 run_v2_only = _run_conditionally(
-    lambda: _is_tf2,
+    lambda: tensorflow_python_tf2.enabled(),
```
nfelt: Optional, but you don't need the lambda wrapper; it can just be `tensorflow_python_tf2.enabled`.

tensorboard/util/test_util.py
```diff
@@ -157,12 +159,10 @@ def _impl(reason=None):
 
   return _impl
 
-# TODO(#1996): Better detect TF2.0.
-_is_tf2 =  tf.__version__.startswith('2.')
```
stephanwlee: I copied this from tensorboard/summary/tf_summary_test.py. May want similar treatment there (tensorboard/compat/__init__.py also has that line btw)

wchargin:
Please hold off on merging; need to verify that this works in google3.

tensorboard/util/test_util.py
wchargin: It looks to me like `tensorboard/compat/__init__.py` is correct: it’s
checking for the TF2 API, not TF2 behavior. Try running

```
TF2_BEHAVIOR=1 python -c 'import tensorflow as tf; print(tf.constant(1)); tf.summary.create_file_writer'
```

in both V1 and V2, and you’ll see that while both print out an eager
tensor, the V1 environment will fail to deref `create_file_writer`,
while the V2 environment will succeed.

Granted, it’s not exactly crystal-clear, but I’ll err on leaving this
alone to minimally unbreak the build unless we see a reason not to.

stephanwlee: :+1:  Sounds good

tensorboard/util/test_util.py
wchargin: Indeed, but you _do_ need the lambda wrapper above for the `not`.
Might as well be consistent, especially as the two forms are not
technically equivalent. It doesn’t matter a whole lot, in any case,
given that `run_v2_only` is never used. :-)

tensorboard/util/test_util.py
nfelt: As William mentioned, the other places check the API symbols directly because that's what they care about, rather than about whether we're actually under TF 2.0 behavior or not.  In fact those helpers are designed to fall back to `tf.compat.v2` if we aren't actually running TF 2.0 - which is the 2.0 API with 1.x behavior - because they're testing the API surface only.

PR #2220, opened 2019-05-09
Mesh plugin README.md and example colab.

podlipensky:
Helping materials that should go out asap. All reviews were done in https://github.com/tensorflow/tensorboard/pull/2215


wchargin:
In lieu of splicing these out of the previous pull request, please
simply remove the demo from the previous PR (`git rm`, `git commit`,
`git push`) and update it.

PR #2221, opened 2019-05-09
Zoom with mouse wheel only when alt key is pressed

### LINKED ISSUE #2214: Zoom charts by holding Shift
Current versions of tensorboard support zooming of charts by mouswheel.
I find it rather annoying that when scrolling down the page I sometimes accidentally zoom out charts.

I propose to support zooming charts only when pressing `Ctrl` while scolling.

stephanwlee:
Default mouse wheel to zoom is rather obnoxious since large area of the
TensorBoard is zoomable area.

Note that preferably we'd want to provide some tooltip or help to make
this feature discoverable but simple hack using `HTMLElement.title` felt very 
disturbing and poor. We, in later time, might want to consider doing what
Google Map does on mobile (when scroll event happens inside the zoomable
area, it shows shows a transparent splash describing how to zoom). Note that
none of these solutions are friendly to the screen reader softwares.

Fixes #2214.


stephanwlee:
I will work on quick follow up to add the Google Map like help when user tries to only zoom.

PR #2232, opened 2019-05-14
graph: Fix JSE when switching node selection

stephanwlee:
Repro steps:
1. Select a run that has more than one tags
2. Select a tag with RunMetadata
3. Select a node
4. Select the default tag
5. Click nodes around and see JSEs

We have nested dom-ifs to prevent `getTotalMicros` from being called.
However, this is only effective when the DOM is not mounted but this
assumption does not hold after first render (if=true).


stephanwlee:
JavaScript Exception. Maybe that acronym is not widely used

wchargin:
Ah, got it. Thanks.

PR #2236, opened 2019-05-15
Remove tailing slash in profiler tutorial

qqfish:


stephanwlee:
Seems to look nominal on app.reviewnb.com.

PR #2237, opened 2019-05-16
tools: create `diagnose_tensorboard`

wchargin:
Summary:
Users often report problems that depend on environment-specific
configuration. Rather than asking them to find all this information and
enter it into an issue template manually, we can ask them to run a
script and paste its output verbatim into the issue. Furthermore, we can
detect and suggest fixes to common problems, such as #1907 and #2010.
The script can grow as we see fit to add new diagnoses and suggestions.

Test Plan:
The script is designed to be robust to errors in each individual
diagnosis: in the worst case, it prints a stack trace and continues to
the next one. I’ve manually tested that the main framework works in
Python 2 and 3 on Linux and in Python 3 on Windows.

Automated testing of this script is possible, but would take a fair
number of CPU cycles to run tests (setting up virtualenvs and Conda,
installing and importing TensorFlow many times). Given that this script
is never a production dependency and is explicitly designed to be run in
a discussion context, light testing seems reasonable.

To simulate a bad hostname, add

```python
socket.getfqdn = lambda: b"\xc9".decode("utf-8")
```

to the top of the script.

To simulate a bad `.tensorboard-info` directory and test the quoting
behavior, run

```shell
export TMPDIR="$(mktemp -d)/uh oh/" &&
mkdir -p "${TMPDIR}/.tensorboard-info" &&
chmod 000 "${TMPDIR}/.tensorboard-info" &&
python ./tensorboard/tools/diagnose_tensorboard.py
```

To cross-check the autoidentification logic:

```
$ python tensorboard/tools/diagnose_tensorboard.py |
> awk '/version / { print $NF }'
e093841ffaea564cb2410e0b430bd0c552ada208
$ git hash-object tensorboard/tools/diagnose_tensorboard.py
e093841ffaea564cb2410e0b430bd0c552ada208
$ git rev-parse HEAD:tensorboard/tools/diagnose_tensorboard.py
e093841ffaea564cb2410e0b430bd0c552ada208
$ git cat-file blob e093841ffaea564cb2410e0b430bd0c552ada208 |
> diff -u tensorboard/tools/diagnose_tensorboard.py - | wc -l
0
```

wchargin-branch: diagnose-me

### DIFF
--- a/tensorboard/tools/diagnose_me.py
+++ b/tensorboard/tools/diagnose_me.py
@@ -23,6 +23,8 @@ from __future__ import print_function
 import collections
 import errno
 import functools
+import hashlib
+import inspect
 import logging
 import os
 import pipes
@@ -90,7 +92,7 @@ def pip(args):
       "ignore:DEPRECATION",
       ",%s" % old_pythonwarnings if old_pythonwarnings else "",
   )
-  command = [sys.executable, "-m", "pip"]
+  command = [sys.executable, "-m", "pip", "--disable-pip-version-check"]
   command.extend(args)
   try:
     os.environ[PYTHONWARNINGS_KEY] = new_pythonwarnings
@@ -113,6 +115,25 @@ def which(name):
     return subprocess.check_output([binary, name])
   except subprocess.CalledProcessError:
     return None
+
+
+@check
+def autoidentify():
+  """Print the Git hash of the current version of `diagnose_me.py`.
+
+  Given this hash, use `git cat-file blob HASH` to recover the relevant
+  version of the script.
+  """
+  module = sys.modules[__name__]
+  try:
+    source = inspect.getsource(module).encode("utf-8")
+  except TypeError:
+    logging.info("diagnose_me.py source unavailable")
+  else:
+    # Git inserts a length-prefix before hashing; cf. `git-hash-object`.
+    blob = b"blob %d\0%s" % (len(source), source)
+    hash = hashlib.sha1(blob).hexdigest()
+    logging.info("diagnose_me.py version %s", hash)
 
 
 @check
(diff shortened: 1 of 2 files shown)


manivaradarajan:
Yay, thanks for starting this.

tensorboard/tools/diagnose_me.py
```diff
@@ -0,0 +1,382 @@
...
+#     again". Should be imperative with no trailing punctuation. May
+#     contain inline Markdown.
+#   description (str): A full enumeration of the steps that the user
+#     should take to accept the suggestion. Within this string, prose
+#     should be formatted with `reflow`. May contain Markdown.
+Suggestion = collections.namedtuple("Suggestion", ("headline", "description"))
+
+
+def check(fn):
+  """Register a function as a check.
```
manivaradarajan: Can you add a line that this is intended to be used as a decorator, i.e., as @check.

tensorboard/tools/diagnose_me.py
```diff
@@ -0,0 +1,382 @@
...
+  set_up_logging()
+
+  print("### Diagnostics")
+  print()
+
+  markdown_code_fence = "``````"  # seems likely to be sufficient
+  print(markdown_code_fence)
+  suggestions = []
+  for check in CHECKS:
+    print("--- check: %s" % check.__name__)
```
stephanwlee: Consider putting one newline after to give some space between the checks (alternatively, we can use the color but that complicates things I guess).

wchargin: Okay, if you prefer. (I like the concise form, but don’t feel strongly.)
I did consider color, but decided against it—color is easy on xterm
emulators, but this script needs to work on Windows, too.

tensorboard/tools/diagnose_me.py
wchargin: Sure.

PR #2244, opened 2019-05-17
Fix datapoint id not displaying correctly for id=0

tolga-b:
Minor bug fix: Datapoint id title did not populate correctly when id is 0, because 0 == ''" is true.

* Screenshots of UI changes

old

new

I locally ran the web demos and verified that they work correctly.


tolga-b:
@jameswex please review. We can stack this with other changes if you want to as well.

PR #2245, opened 2019-05-18
build: use `mirror.tensorflow.org`, now populated

wchargin:
Summary:
It’s easier for TensorBoard team members to upload files to the
TensorFlow.org mirror than the Bazel mirror. Many of our files were not
mirrored, and fetches from GitHub can be noticeably slow. (For instance,
the Bazel 0.22.0 binary is 1.8–2.0 seconds from the TF mirror and 8.5–12
seconds from GitHub.) As of this commit, all `mirror.bazel.build` URLs
are replaced with `mirror.tensorflow.org` URLs, and all such URLs
contain valid content.

Along the way, I swapped out the defunct numericjs source to a host that
actually still resolves.

On Travis, getting Bazel and running `bazel fetch //tensorboard/...`
typically takes about 44 seconds. Let’s see if this speeds that up.

Test Plan:
Check that Bazel URLs resolve:

```
$ git grep -Pho '(?<=")https?://mirror\.tensorflow\.org/[^"]*' |
> grep -vF '${BAZEL}' | sort | uniq |
> xargs -n 1 -P 32 -- \
> sh -c 'curl -sfL "$1" >/dev/null && echo OK || echo FAIL' unused |
> sort | uniq -c
    231 OK
```

Check that the Travis URL resolves:

```
$ BAZEL=0.22.0
$ curl -sfL "http://mirror.tensorflow.org/github.com/bazelbuild/bazel/releases/download/${BAZEL}/bazel-${BAZEL}-linux-x86_64" |
> shasum -a 256
8474ed28ed4998e2f5671ddf3a9a80ae9e484a5de3b8b70c8b654c017c65d363
```

Check that we caught ’em all:

```
$ git grep mirror.bazel | wc -l
0
```

wchargin-branch: mirror-tensorflow-org

### DIFF
--- a/tensorboard/tools/import_google_fonts.py
+++ b/tensorboard/tools/import_google_fonts.py
@@ -74,7 +74,7 @@ flags.DEFINE_string('user_agent',
                        'Chrome/62.0.3202.94 '
                        'Safari/537.36',
                        'HTTP User-Agent header to send to Google Fonts')
-flags.DEFINE_string('mirror', 'https://mirror.tensorflow.org/',
+flags.DEFINE_string('mirror', 'http://mirror.tensorflow.org/',
                        'Mirror URL prefix')
 FLAGS = flags.FLAGS
 
--- a/tensorboard/tools/import_google_fonts.py
+++ b/tensorboard/tools/import_google_fonts.py
@@ -74,7 +74,7 @@ flags.DEFINE_string('user_agent',
                        'Chrome/62.0.3202.94 '
                        'Safari/537.36',
                        'HTTP User-Agent header to send to Google Fonts')
-flags.DEFINE_string('mirror', 'https://mirror.bazel.build/',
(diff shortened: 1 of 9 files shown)


wchargin:
Well, fetch down from 44s to 40s on Travis. Sure, okay.

tensorboard/tools/import_google_fonts.py
```diff
@@ -74,7 +74,7 @@
                        'Chrome/62.0.3202.94 '
                        'Safari/537.36',
                        'HTTP User-Agent header to send to Google Fonts')
-flags.DEFINE_string('mirror', 'https://mirror.bazel.build/',
+flags.DEFINE_string('mirror', 'https://mirror.tensorflow.org/',
```
stephanwlee: Should this url be `http` instead of `https`?

wchargin: Yes! Thank you.

PR #2250, opened 2019-05-20
Polymer 2: core fixes

stephanwlee:
This pull request contains three changes that are more platform-y than future pull requests.
Please see individual commit messages for better descriptions of the changes.

### DIFF
--- a/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
+++ b/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
@@ -19,6 +19,11 @@ namespace tf_dashboard_common {
  */
 export const DataLoaderBehavior = {
   properties: {
+    active: {
+      type: Boolean,
+      observer: '_loadDataIfActive',
+    },
+
     /**
      * A unique identifiable string. When changes, it expunges the data
      * cache.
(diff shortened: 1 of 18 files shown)


tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -139,10 +144,15 @@ export const DataLoaderBehavior = {
...
   },
 
+  _loadDataIfActive() {
+    if (this.active) {
+      this._loadData();
+    }
+  },
+
   _loadDataImpl() {
+    if (!this.active) return;
```
wchargin: It doesn’t look like `this.active` is ever set to `true`. What’s
happening here? How does this work?

tensorboard/plugins/debugger/tf_debugger_dashboard/tf-debugger-resizer.html
```diff
@@ -26,20 +26,20 @@
...
-      :host[is-horizontal] {
+      :host([is-horizontal]) {
         cursor: row-resize;
         height: 10px;
         left: 0;
         right: 0;
       }
 
-      :host[_is-vertical] {
+      :host([_is-olvertical]) {
```
wchargin: s/olvertical/vertical/

stephanwlee: Opps. Done.

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
stephanwlee: It is only set by parent (technically component that uses the behavior can hard code it too). 
This should gets used with `tf-category-paginted-view`. i.e.,
https://github.com/stephanwlee/tensorboard/blob/225748afd496f62387682f27b304543282bded37/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-dashboard.html#L131-L138

wchargin: Right, of course. Thanks.

PR #2255, opened 2019-05-21
Add option to select which model to use for counterfactual computation

tolga-b:
This PR adds the ability to choose which model to use for for counterfactual example computation (Previously, it was computed with respect to first model).

Added a drop-down menu to choose model name in counterfactual settings row. This panel is hidden when there is only one loaded model.

* Screenshots of UI changes

Screenshot from 2019-05-21 09-47-57

Screenshot from 2019-05-21 09-48-18

I ran all demos locally and verified that counterfactuals are computed correctly for the chosen model.

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -2683,18 +2683,6 @@ limitations under the License.
         }
       },
 
-      // nearestCounterfactualDistChanged_: function() {
-      //   if (this.showNearestCounterfactual) {
-      //     this.findClosestCounterfactual_();
-      //   }
-      // },
-
-      // nearestCounterfactualModelChanged_: function() {
-      //   if (this.showNearestCounterfactual) {
-      //     this.findClosestCounterfactual_();
-      //   }
-      // },
-
       findClosestCounterfactual_: function() {
         const selected = this.selected[0];
         const modelInferenceValueStr =
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -190,6 +190,17 @@ limitations under the License.
         }
       }
 
+      .counterfactual-dropdown {
+        display: block;
+        width: 100px;
+        min-width: 100px;
+        padding-right: 10px;
+        --paper-input-container-input: {
+          font-size: 14px;
+          color: #3C4043;
+        }
+      }
+
       .slider-label {
         margin-top: 8px;
       }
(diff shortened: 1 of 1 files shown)


tolga-b:
@jameswex please review.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -2868,6 +2898,10 @@ <h2>Show similarity to selected datapoint</h2>
         return stats && Object.keys(stats).length > 0;
       },
 
+      shouldHideCounterfactualModelSelector_: function(modelNames) {
```
jameswex: how about being dependent on "numModels" instead? Then just need to check if that num != 1.

tolga-b: I wanted to make sure that parsedModelNames is populated before creating the window element, in case we end up in a state where numModels is set but parsedModelNames is still being processed. Right now it seems like parsedModelNames is updated before numModels, so again it may cause some inconsistencies (e.g. numModels=1 but parsedModelNames,length=2).

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -2667,11 +2691,17 @@ <h2>Show similarity to selected datapoint</h2>
         }
       },
 
-      // TODO (tolgab) Update for model comparison
+      nearestCounterfactualModelChanged_: function() {
+        if (this.showNearestCounterfactual) {
+          this.findClosestCounterfactual_();
```
stephanwlee: I see a bit of repetition with `nearestCounterfactualDistChanged_`. Would it make sense to just use an observer and do
```ts
observers: [
  ...,
  'maybeFindClosestCounterFactual_(showNearestCounterfactual, nearestCounterfactualModel, nearestCounterfactualDist)'
],
```

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -2266,6 +2284,12 @@ <h2>Show similarity to selected datapoint</h2>
           value: '',
           observer: 'labelFeatureSelected_',
         },
+        // Choose which model is used for counterfactual computation
+        nearestCounterfactualModel: {
+          type: Number,
+          value: 0,
```
stephanwlee: Is this number referring to an id of a model or an index? Perhaps a change to the prop name to better depict its nature might be nice.

tolga-b: done!

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
tolga-b: Good idea, done!

PR #2256, opened 2019-05-21
Fix WIT multiclass confusion matrix

jameswex:

In certain cases the confusion matrix was missing entries.

Correctly add both the inferred label and the true label to the set of all labels to show in the multiclass confusion matrix, including coercing all labels to strings. 

* Screenshots of UI changes

fixconfmatrix

Verified continued correct working in multiclass demo, and verified correct behavior in multiclass colab notebook that displayed this issue.

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -3176,8 +3176,8 @@ limitations under the License.
                 const modelInferenceValueStr =
                   this.strWithModelName_(inferenceValueStr, modelInd);
                 let count = statsActual[item[modelInferenceValueStr]];
-                allLabels.add(item[modelInferenceValueStr]);
-                allLabels.add(item[modelInferenceValueStr]);
+                allLabels.add(String(item[modelInferenceValueStr]));
+                allLabels.add(String(item[this.selectedLabelFeature]));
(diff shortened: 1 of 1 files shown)


jameswex:
@tolga-b please review

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3176,8 +3176,8 @@ <h2>Show similarity to selected datapoint</h2>
                 const modelInferenceValueStr =
                   this.strWithModelName_(inferenceValueStr, modelInd);
                 let count = statsActual[item[modelInferenceValueStr]];
-                allLabels.add(item[modelInferenceValueStr]);
-                allLabels.add(item[modelInferenceValueStr]);
+                allLabels.add(String(item[modelInferenceValueStr]));
```
stephanwlee: Curious: what happens if you don't explicitly cast it to string?

jameswex: For users that provide "ground truth class ID" as a number to WIT, you may get a confusion matrix where rows and columns are doubled, that is that there is a row for Predicted class 0 (as an int) and a row for Prediction class 0 (as a string) separately. They both contain the same correct numbers for those confusion matrix cells, but the rows and columns are duplicated leading to an overly large and confusing confusing matrix.

PR #2257, opened 2019-05-21
Extensible plugin infrastructural changes

stephanwlee:
This is the change based on https://github.com/tensorflow/community/pull/90.

The change makes two improvements:
1.  allow dynamic plugins (as spec in the RFC; uses entry_points to discover) to be loaded
2. change `plugins_listing` contract to include metadata about plugins including their respective FE entry point.

Notes on default_test: tried to get this tested end-to-end but setuptools was a bit cumbersome to test. For specifically, `setuptools.setup` cannot be invoked programmatically and it can lead to bad side-effects when not cleaned up properly.

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -277,9 +277,14 @@ class TensorBoardWSGI(object):
     response = {}
     for plugin in self._plugins:
       start = time.time()
+      module_path = None
+      if plugin.es_module_path() is not None:
+        module_path = (self._path_prefix + DATA_PREFIX + PLUGIN_PREFIX + '/' +
+                       plugin.plugin_name + plugin.es_module_path())
+
       response[plugin.plugin_name] = {
           'enabled': plugin.is_active(),
-          'es_module_path': plugin.es_module_path() or '',
+          'es_module_path': module_path,
       }
       elapsed = time.time() - start
       logger.info(
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -277,7 +277,10 @@ class TensorBoardWSGI(object):
     response = {}
     for plugin in self._plugins:
       start = time.time()
-      response[plugin.plugin_name] = plugin.is_active()
+      response[plugin.plugin_name] = {
+          'enabled': plugin.is_active(),
+          'es_module_path': plugin.es_module_path() or '',
+      }
       elapsed = time.time() - start
       logger.info(
           'Plugin listing: is_active() for %s took %0.3f seconds',
--- a/tensorboard/components/tf_tensorboard/tf-tensorboard.html
+++ b/tensorboard/components/tf_tensorboard/tf-tensorboard.html
@@ -817,8 +817,11 @@ limitations under the License.
         return this._dashboardData
             .map((d) => d.plugin)
             .filter((dashboardName) => {
-                return this._pluginsListing[dashboardName] &&
-                    this._pluginsListing[dashboardName].enabled;
+                const maybeMetadata = this._pluginsListing[dashboardName];
+                if (typeof maybeMetadata === 'boolean') {
+                  return maybeMetadata;
+                }
+                return this._pluginsListing[dashboardName].enabled;
             });
       },
 
--- a/tensorboard/components/tf_tensorboard/tf-tensorboard.html
+++ b/tensorboard/components/tf_tensorboard/tf-tensorboard.html
@@ -817,11 +817,15 @@ limitations under the License.
         return this._dashboardData
             .map((d) => d.plugin)
             .filter((dashboardName) => {
+                // TODO(stephanwlee): Remove boolean code path when releasing
+                // 2.0.
+                // PluginsListing can be an object whose key is name of the
+                // plugin and value is a boolean indicating whether if it is
+                // enabled. This is deprecated but we will maintain backwards
+                // compatibility for some time.
                 const maybeMetadata = this._pluginsListing[dashboardName];
-                if (typeof maybeMetadata === 'boolean') {
-                  return maybeMetadata;
-                }
-                return this._pluginsListing[dashboardName].enabled;
+                if (typeof maybeMetadata === 'boolean') return maybeMetadata;
+                return maybeMetadata && maybeMetadata.enabled;
             });
       },
 
--- a/tensorboard/components/tf_tensorboard/tf-tensorboard.html
+++ b/tensorboard/components/tf_tensorboard/tf-tensorboard.html
@@ -465,17 +465,18 @@ limitations under the License.
           value: _.values(tf_tensorboard.dashboardRegistry),
         },
 
+        _pluginsListing: {
+          type: Object,
+          value: () => ({}),
+        },
+
         /**
-         * The set of currently active dashboards. `null` if not yet fetched.
-         *
-         * TODO(@wchargin): Consider templating in an initial value for
-         * this property.
-         *
+         * The set of currently active dashboards.
          * @type {Array<string>?}
          */
         _activeDashboards: {
           type: Array,
-          value: null,
+          computed: '_computeActiveDashboard(_dashboardData, _pluginsListing)'
         },
 
         /** @type {tf_tensorboard.ActiveDashboardsLoadState} */
@@ -812,15 +813,22 @@ limitations under the License.
         this._lastReloadTime = new Date().toString();
       },
 
-      _fetchActiveDashboards() {
+      _computeActiveDashboard() {
+        return this._dashboardData
+            .map((d) => d.plugin)
+            .filter((dashboardName) => {
+                return this._pluginsListing[dashboardName] &&
+                    this._pluginsListing[dashboardName].enabled;
+            });
+      },
+
+      _fetchPluginsListing() {
         this._canceller.cancelAll();
-        const updateActiveDashboards = this._canceller.cancellable(result => {
+        const updatePluginsListing = this._canceller.cancellable(result => {
           if (result.cancelled) {
             return;
           }
-          const activePlugins = result.value;
-          this._activeDashboards = this._dashboardData.map(
-              d => d.plugin).filter(p => activePlugins[p]);
+          this._pluginsListing = result.value;
           this._activeDashboardsLoadState = tf_tensorboard.ActiveDashboardsLoadState.LOADED;
         });
         const onFailure = () => {
@@ -834,7 +842,7 @@ limitations under the License.
         };
         return this._requestManager
           .request(tf_backend.getRouter().pluginsListing())
-          .then(updateActiveDashboards, onFailure);
+          .then(updatePluginsListing, onFailure);
       },
 
       _computeActiveDashboardsNotLoaded(state) {
@@ -878,7 +886,7 @@ limitations under the License.
       _reloadData() {
         this._refreshing = true;
         return Promise.all([
-          this._fetchActiveDashboards(),
+          this._fetchPluginsListing(),
           tf_backend.environmentStore.refresh(),
           tf_backend.runsStore.refresh(),
           tf_backend.experimentsStore.refresh(),
--- a/tensorboard/default.py
+++ b/tensorboard/default.py
@@ -95,22 +95,19 @@ def get_plugins():
 def get_dynamic_plugins():
   """Returns a list specifying TensorBoard's dynamically loaded plugins.
 
-	A dynamic TensorBoard plugin is specified using entry_points [1]. This will be
-  the de facto mechanism to load plugins and there no longer will be a concept
-  of the first-party plugins as depicted above in `get_plugins`.
+  A dynamic TensorBoard plugin is specified using entry_points [1] and it is
+  the primary way to integrate plugin whose code does not reside in TensorBoard
+  repository.
 
   This list can be passed to the `tensorboard.program.TensorBoard` API.
 
   Returns:
-		a list of base_plugin.TBPlugin
+    list of base_plugin.TBLoader or base_plugin.TBPlugin.
 
   [1]: https://packaging.python.org/specifications/entry-points/
   """
-  dynamic_plugins = [
-      entry_point.load()
-      for entry_point
-      in pkg_resources.iter_entry_points('tensorboard_plugins')]
-
   return [
-      get_plugin()[2]
-      for get_plugin in dynamic_plugins]
+      entry_point.load()
+      for entry_point in pkg_resources.iter_entry_points('tensorboard_plugins')
+  ]
+
--- a/tensorboard/default.py
+++ b/tensorboard/default.py
@@ -30,6 +30,7 @@ from __future__ import print_function
 
 import logging
 import os
+import pkg_resources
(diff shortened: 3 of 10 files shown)


tensorboard/http_api.md
```diff
@@ -35,13 +35,30 @@ The `logdir` argument is the path of the directory that contains events files.
 
 ## `data/plugins_listing`
 
-Returns a dict mapping from plugin name to a boolean indicating whether
+Returns a dict mapping from plugin name to an object that describes a plugin.
```
wchargin: Previously, dashboard order was determined by plugin registration order,
which was in turn hard-coded in `default-plugins.html`. But now there is
nothing to determine plugin registration order. Should this be an array
of plugin objects instead?

tensorboard/http_api.md
```diff
@@ -35,13 +35,30 @@ The `logdir` argument is the path of the directory that contains events files.
...
 the plugin is active. A plugin might be inactive, for instance, if it
 lacks relevant data. Every plugin has a key. This route allows the
 frontend to hide or deprioritize inactive plugins so that the user can
 focus on the plugins that have data. Note that inactive plugins may
 still be rendered if the user explicitly requests this.
 
+Another property `es_module_path` is an optional one that describes a path to
+the main JavaScript plugin module that will be loaded onto an iframe. For "v1"
+plugins whose JavaScript source is incorporated into webfiles.zip, the field is
+empty.
```
wchargin: Above you say that the field is optional. For V1 plugins, is the field
empty (`""`) or absent? If empty, in which cases is the field absent?

tensorboard/http_api.md
```diff
@@ -35,13 +35,30 @@ The `logdir` argument is the path of the directory that contains events files.
...
+
+Example response:
+
+    {
+      "scalars": {
+        "enabled": true,
+        "es_module_path": "/scalars/plugins/main.js"
```
wchargin: > `/scalars/plugins/main.js`

Is this realistic? Wouldn’t it be `/data/plugin/scalars/main.js`?

tensorboard/default.py
```diff
@@ -86,4 +87,30 @@ def get_plugins():
...
   """
+
   return _PLUGINS[:]
+
+# TODO(stephanwlee): Combine with get_plugins when concept of first-party
+# plugins is undone.
+def get_dynamic_plugins():
+  """Returns a list specifying TensorBoard's dynamically loaded plugins.
+
+	A dynamic TensorBoard plugin is specified using entry_points [1]. This will be
```
wchargin: Your editor appears to be inserting literal tabs (0x09). Please run
these files through `expand -t 2` and consider adding

```viml
set expandtab
autocmd FileType python setl ts=2 sw=2 sts=2
```

to your vimrc.

tensorboard/http_api.md
wchargin: nit: The rest of this section is wrapped at 72; please consider wrapping
thus for consistency. (`:setl tw=72` then `gqip` will do it.)

tensorboard/default.py
```diff
@@ -86,4 +87,30 @@ def get_plugins():
...
+  the de facto mechanism to load plugins and there no longer will be a concept
```
wchargin: It’s not just the _de facto_ mechanism; it’s the actual mechanism. (In
fact, one might say that the _de facto_ mechanism is the legacy one,
given that the new one is not yet used.)

tensorboard/default.py
```diff
@@ -86,4 +87,30 @@ def get_plugins():
...
+  of the first-party plugins as depicted above in `get_plugins`.
```
wchargin: Hmm, do we really want to ~deprecate `get_plugins`? Users who launch
TensorBoard via a custom `tensorboard.program.TensorBoard` instance may
continue to want a fixed list of plugins rather than using the entry
points mechanism—perhaps because they build/package Python programs in
such a way that entry points are not used, or perhaps because they
simply want to have clear build-time dependencies for a production
system.

tensorboard/default.py
```diff
@@ -86,4 +87,30 @@ def get_plugins():
...
+
+  This list can be passed to the `tensorboard.program.TensorBoard` API.
+
+  Returns:
+		a list of base_plugin.TBPlugin
```
wchargin: They’ll actually be either `TBPlugin`s or `TBLoader`s, right? I.e., the
return type is the same as that of `get_plugins`?

tensorboard/default.py
```diff
@@ -86,4 +87,30 @@ def get_plugins():
...
+
+  [1]: https://packaging.python.org/specifications/entry-points/
+  """
+  dynamic_plugins = [
+      entry_point.load()
```
wchargin: After a bit more thought, how about just returning this list as is?

```python
  return [
    entry_point.load()
    for entry_point in pkg_resources.iter_entry_points("tensorboard_plugins")
  ]
```

I’d originally wanted the symbol to be a function to accommodate plugins
that want to defer importing expensive modules as much as possible,
hence the `get_plugin()` indirection. But that’s already possible with
the `TBLoader` mechanism (e.g., `BeholderPluginLoader`), so it’s not
clear to me that further indirection is actually needed. The “name and
description” metadata is probably superfluous, and if we want to add it
later we can add optional fields/methods to `TBLoader`.

In this form, entry points would reference plugins/loaders directly:

```python
entry_points = {
    "tensorboard_plugins": [
        "widgets = tensorboard_plugin_widgets.widgets_plugin:WidgetsPlugin",
        "doodads = tensorboard_plugin_doodads.doodads_plugin:DoodadsLoader",
    ],
}
```

tensorboard/http_api.md
```diff
@@ -35,13 +35,30 @@ The `logdir` argument is the path of the directory that contains events files.
...
+      }
```
wchargin: Can the sample response include a representative example of each kind?
Maybe include the scalars plugin with no `es_module_path`, and a widgets
plugin with a module path?

tensorboard/plugins/base_plugin.py
```diff
@@ -75,6 +75,18 @@ def is_active(self):
     """
     raise NotImplementedError()
 
+  def es_module_path(self):
+    """Returns one of the keys in get_plugin_apps that is an entry ES module.
+
+    For a plugin that is loaded onto an iframe, a frontend entry point has to be
```
wchargin: nit: Loaded *into* an iframe, not *onto* an iframe.

tensorboard/plugins/base_plugin.py
```diff
@@ -75,6 +75,18 @@ def is_active(self):
...
+    specified. For a plugin that is bundled with TensorBoard as part of
```
wchargin: Okay—we can specify the interface more concretely by pointing to a
frontend TypeScript declaration once we have that part of the system
built. (No action required.)

tensorboard/plugins/base_plugin.py
```diff
@@ -75,6 +75,18 @@ def is_active(self):
...
+    webfiles.zip, return None.
+
+    Returns:
+      A key in the get_plugins_apps.
```
wchargin: sp.; perhaps ``A key in the result of `get_plugin_apps()`, or `None`.``?

tensorboard/backend/application.py
```diff
@@ -277,7 +277,10 @@ def _serve_plugins_listing(self, request):
     response = {}
     for plugin in self._plugins:
       start = time.time()
-      response[plugin.plugin_name] = plugin.is_active()
+      response[plugin.plugin_name] = {
+          'enabled': plugin.is_active(),
+          'es_module_path': plugin.es_module_path() or '',
```
wchargin: Shouldn’t this have the `/data/plugins/PLUGIN/` prefix?

wchargin: Hmm, identifying the empty string with “no module” seems confusing: I’d
expect the empty string to mean that the ES module is hosted at simply
`/data/plugins/PLUGINS/`. Can we just drop the `or ''` and let the field
be `null`?

tensorboard/default.py
stephanwlee: I use editorconfig plugin and it should technically take care of that though I have seen few cases when my editorconfig is not honored. retabbed.

tensorboard/http_api.md
stephanwlee: TIL about gqip! Thanks!

tensorboard/http_api.md
stephanwlee: Discussed a bit offline:
1. the order of plugins in TensorBoard UI is a UI consideration and plugins_listing probably should not care about that (sorting should be done on the frontend)
2. currently, the order in which plugins are listed is dictated by order of `registerDashboard` calls and it is defined in https://github.com/tensorflow/tensorboard/blob/master/tensorboard/components/tf_tensorboard/default-plugins.html. 
3. For dynamic plugins, the frontend entry point will not appear in aforementioned default-plugins.html. In future FE changes, we will _likely_ to introduce some business logic around how plugins are sorted in UI: plugins listed on the default-plugins will appear first then dynamic plugins will be alphabetically sorted.

tensorboard/default.py
```diff
@@ -30,6 +30,7 @@
 
 import logging
 import os
+import pkg_resources
```
wchargin: This is technically a third-party dependency, and so (a) should be in a
separate imports section, and (b) needs to be expressed in Bazel. The
second point matters for the Google import process.

tensorboard/main.py
```diff
@@ -54,8 +54,9 @@ def run_main():
     print("TensorFlow installation not found - running with reduced feature set.",
           file=sys.stderr)
 
-  tensorboard = program.TensorBoard(default.get_plugins(),
-                                    program.get_default_assets_zip_provider())
+  tensorboard = program.TensorBoard(
+			default.get_plugins() + default.get_dynamic_plugins(),
```
wchargin: nit: Tabs (0x09) spotted.

tensorboard/http_api.md
```diff
@@ -35,31 +35,34 @@ The `logdir` argument is the path of the directory that contains events files.
...
-webfiles.zip, the field is empty.
+## `data/plugins_metadata`
+
+Returns a dict mapping from plugin name to an object that describes a
+plugin.
+
+The object contains a property `es_module_path` is an optional one that
+describes a path to the main JavaScript plugin module that will be
+loaded onto an iframe. For "v1" plugins whose JavaScript source is
+incorporated into webfiles.zip, the field is empty.
```
wchargin: From the example below it looks like the field is `null`, right?

tensorboard/backend/application.py
```diff
@@ -277,19 +277,33 @@ def _serve_plugins_listing(self, request):
...
+    for plugin in self._plugins:
       module_path = None
       if plugin.es_module_path() is not None:
         module_path = (self._path_prefix + DATA_PREFIX + PLUGIN_PREFIX + '/' +
                        plugin.plugin_name + plugin.es_module_path())
 
       response[plugin.plugin_name] = {
-          'enabled': plugin.is_active(),
-          'es_module_path': module_path,
+	  'es_module_path': module_path,
```
wchargin: nit: Tab (0x09) spotted. (Your EditorConfig does not appear to be working.)

tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -812,15 +813,22 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
         this._lastReloadTime = new Date().toString();
       },
 
-      _fetchActiveDashboards() {
+      _computeActiveDashboard() {
+        return this._dashboardData
+            .map((d) => d.plugin)
+            .filter((dashboardName) => {
+                return this._pluginsListing[dashboardName] &&
```
wchargin: Did we want this frontend logic to be backward compatible with the
previous HTTP API?

The easiest way to make it backward compatible might be to transform the
response from `{k: v}` to `{k: typeof v === 'boolean' ? v : v.enabled}`
immediately after we receive it (in `updatePluginsListing`), allowing
most of the other logic to stay unchanged. Or you could store the actual
response data and case on it here, but it looks like as written this JS
will crash if given a response from the previous backend.

tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -812,15 +813,22 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
...
         this._canceller.cancelAll();
-        const updateActiveDashboards = this._canceller.cancellable(result => {
+        const updatePluginsListing = this._canceller.cancellable(result => {
           if (result.cancelled) {
             return;
           }
-          const activePlugins = result.value;
-          this._activeDashboards = this._dashboardData.map(
-              d => d.plugin).filter(p => activePlugins[p]);
+          this._pluginsListing = result.value;
```
wchargin: The renames/refactors in this commit make the `tf-tensorboard` logic
easier to follow; thanks!

tensorboard/default.py
wchargin: Got the Bazel dep, but the imports should still be:

```python
import logging
import os

import pkg_resources

from tensorboard.compat import tf
```

tensorboard/default_test.py
```diff
@@ -0,0 +1,74 @@
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
wchargin: nit: 2019

tensorboard/default.py
```diff
@@ -86,4 +87,25 @@ def get_plugins():
...
+  Returns:
+    list of base_plugin.TBLoader or base_plugin.TBPlugin.
+
+  [1]: https://packaging.python.org/specifications/entry-points/
+  """
+  return [
+      entry_point.load()
+      for entry_point in pkg_resources.iter_entry_points('tensorboard_plugins')
+  ]
+
```
wchargin: nit: trailing blank line

wchargin:
Mind rebasing/merging master to pull in #2267? Bazel 0.26 is out.

tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -817,8 +817,11 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
         return this._dashboardData
             .map((d) => d.plugin)
             .filter((dashboardName) => {
-                return this._pluginsListing[dashboardName] &&
-                    this._pluginsListing[dashboardName].enabled;
+                const maybeMetadata = this._pluginsListing[dashboardName];
+                if (typeof maybeMetadata === 'boolean') {
+                  return maybeMetadata;
+                }
+                return this._pluginsListing[dashboardName].enabled;
```
wchargin: nit: Replace `this._pluginsListing[dashboardName]` with `maybeMetadata`?

A brief comment about the reason for this business (namely, API compat)
would be reasonable imho, but I’ll leave that up to you.

PR #2259, opened 2019-05-21
hparams: add `sample_uniform`

### LINKED ISSUE #2246: Getting a random value from a Domain
It would be convenient to have a single method to get a random value from a Domain. It would make the code to do a random search shorter and prettier. 

Would that make sense? If yes, I can create a PR.
(cc @wchargin )

Examples:

```
class RealInterval(Domain):
  ...
  def rand(self):
    return random.randint(self.minvalue, self.maxvalue)
  ... 
```

```
class Discrete(Domain):
  ...
  def rand(self):
    return random.choice(self.values)
  ... 
```

wchargin:
Summary:
Resolves #2246.

Test Plan:
Unit tests added, and demo updated; the demo still works.

wchargin-branch: hparams-sample-uniform


tensorboard/plugins/hparams/summary_v2_test.py
```diff
@@ -440,6 +443,47 @@ def _get_unique_summary(self, logdir):
   return summaries[0]
 
 
+def assert_uniform(self, range_, samples, delta=0.05, alpha_bound=1e-4):
```
nfelt: I realize you already did a bunch of work here, but I'm a little bit uneasy about incorporating such sophisticated and conceptually nondeterministic [1] testing of what is ultimately very, very thin wrappers over `random`.  This feels like a case where it's better to just test against a mock implementation and ensure that the wrappers call the expected APIs, and then trust that those APIs behave as documented.  Otherwise I feel like we're picking up a not-insubstantial maintenance burden for the test code itself, which seems out of proportion to the complexity of the implementation.

The other concern I have is that this test, while elegant, wouldn't even catch many types of non-uniform behavior:
- we're only testing each method with a fixed range; non-uniformity that doesn't appear within this range goes undetected.  For example, a not-strictly-uniform implementation using `rand() % N` works for range widths that evenly divide `RAND_MAX` but not otherwise.
- we're only testing that buckets above a minimum threshold of 0.95*expected and not checking against a maximum threshold, so a non-uniformity that assigns 5% of the probability mass to a single bucket still passes, even in a case of 100 buckets where that makes the single bucket 5 times more likely than any other
- we aren't testing against the buckets being *too* uniform; for the discrete domains, deterministic uniformity (e.g. a stateful round-robin sampling approach) would pass this test with flying colors, but it's contrary to the presumed probabilistic behavior which we'd expect to actually look a little non-uniform some of the time.

So I'm worried this test gives a false sense of comprehensiveness.

[1]  Even if they are functionally deterministic in the particular invocation here due to the fixed seed, there's nothing about the logic that guarantees that; the 0 seed just "happens" to work, so the fact that the test passes can't really be ascertained via inspection alone.

wchargin: Okay; that is a defensible perspective. Done.

For the record, the previous formulation has some advantages. For
instance, using `randrange` instead of `randint` could easily pass
cursory inspection, but `randrange` actually samples from a half-open
interval. The sortedness of the argument to `choice` leaks into the
test, unless you take special precautions with `assertItemsEqual` and
`mock_calls`. And of course you have the usual mocking brittleness (the
implementation could freely change to `random.Random().uniform` from
`random.uniform`, which would be a false test failure) and the awful
Python `mock` library (don’t get me started about `assret_*` [sic]).

I’m not too worried about “exotic” nonuniformities like in your second
bullet point; it’s fair to assume that `sample_uniform` doesn’t handroll
some broken PRNG. Your first and third bullets are reasonable, imho.

tensorboard/plugins/hparams/summary_v2_test.py
```diff
@@ -512,20 +471,25 @@ def test_backward_endpoints(self):
...
   def test_sample_uniform_unseeded(self):
     domain = hp.IntInterval(2, 7)
-    for _ in xrange(100):
-      sample = domain.sample_uniform()
-      self.assertIsInstance(sample, int)
-      self.assertGreaterEqual(sample, domain.min_value)
-      self.assertLessEqual(sample, domain.max_value)
+    # Note: `randint` samples from a closed interval, which is what we
+    # want (as opposed to `randrange`).
+    with mock.patch("random.randint") as m:
```
nfelt: Slightly cleaner IMO to `import random` and then use `mock.patch.object` (this is what the internal mocking guide recommends) but up to you.

tensorboard/plugins/hparams/summary_v2_test.py
nfelt: Thanks.  I realize there are errors that the non-mocking test could catch (like randint/randrange), and mocking isn't normally the best way to test, but in this case given the simplicity of the implementations and the unlikelihood of us actually changing them, it's better to keep it simple overall.

tensorboard/plugins/hparams/summary_v2_test.py
wchargin: Done; thanks for the reference.

PR #2260, opened 2019-05-21
Add S3 write support for non-TensorFlow writer

orionr:
This adds Amazon S3 write support for the non-TensorFlow writer. It will be used with TensorBoard for PyTorch (torch.utils.tensorboard) as well as for other frameworks.

Tested with

```
conda create --name notf36 python=3.6
conda activate notf36
pip install --upgrade pip
pip install scipy
conda install -y protobuf
conda install -y absl-py
conda install -y numpy
pip install boto3
pip install moto
bazel test //tensorboard/compat/tensorflow_stub:gfile_test --test_output=errors
bazel test //tensorboard/compat/tensorflow_stub:gfile_s3_test --test_output=errors
bazel test //tensorboard/summary/writer:event_file_writer_test --test_output=errors
bazel test //tensorboard/summary/writer:event_file_writer_s3_test --test_output=errors
```

cc @nfelt @lanpa @natalialunova @wchargin


tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -109,6 +109,22 @@ def read(self, filename, binary_mode=False, size=None, offset=None):
...
+            filename: string, a path
+            file_content: string, the contents
+            binary_mode: bool, write as binary if True, otherwise text
+        """
+        mode = "wb" if binary_mode else "w"
+        if binary_mode:
+            file_content = compat.as_bytes(file_content)
+        else:
+            file_content = compat.as_text(file_content)
+        with io.open(filename, mode) as f:
```
nfelt: It'd be better to set encoding explicitly (e.g. to UTF-8) versus letting it be platform-dependent, for the text-mode case.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -228,6 +248,22 @@ def read(self, filename, binary_mode=False, size=None, offset=None):
...
+        """Writes string file contents to a file.
+
+        Args:
+            filename: string, a path
+            file_content: string, the contents
+            binary_mode: bool, write as binary if True, otherwise text
+        """
+        client = boto3.client("s3")
+        bucket, path = self.bucket_and_path(filename)
+        if binary_mode:
```
nfelt: What's the behavior of the boto3 APIs if you pass text vs binary?  Does the text get encoded in some particular encoding?  If so, can we specify it (UTF-8) so it's explicit?

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -109,6 +109,22 @@ def read(self, filename, binary_mode=False, size=None, offset=None):
...
+    def write(self, filename, file_content, binary_mode=False):
+        """Writes string file contents to a file.
+
+        Args:
+            filename: string, a path
+            file_content: string, the contents
+            binary_mode: bool, write as binary if True, otherwise text
+        """
+        mode = "wb" if binary_mode else "w"
+        if binary_mode:
```
nfelt: For a new API I'd prefer to be strict and only accept six.binary_type if binary_mode is True, otherwise only accept six.text_type.  Ditto for the S3 implementation.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -140,6 +156,10 @@ def listdir(self, dirname):
         entries = [compat.as_str_any(item) for item in entries]
         return entries
 
+    def makedirs(self, path):
+        """Creates a directory and all parent/intermediate directories."""
```
nfelt: `os.makedirs()` has the following behavior: "Raises an error exception if the leaf directory already exists or cannot be created."

That should be documented here, and the S3 version should maintain the same behavior contract - does it?  If it's difficult for S3 to do that, we could choose a different behavior (like "has no effect if the leaf directory already exists") but then we should still document it and ensure it's consistent across FS implementations.

tensorboard/compat/tensorflow_stub/io/gfile_test.py
```diff
@@ -82,6 +82,13 @@ def testListdir(self):
             expected_files,
             gfile.listdir(temp_dir))
 
+    def testMakeDirs(self):
+        temp_dir = tempfile.mkdtemp(prefix=self.base_temp_dir)
```
nfelt: Now that `tensorboard.test.TestCase` exists - thanks! - could we use that in this test file so it's hooked into the bazel test temp directory?

tensorboard/compat/tensorflow_stub/io/gfile_test.py
```diff
@@ -192,6 +199,28 @@ def testReadWithOffset(self):
             ckpt_read = f.read()
             self.assertEqual(ckpt_b_content, ckpt_read)
 
+    def testWrite(self):
```
nfelt: There should be at least one test that checks the behavior of multiple writes to a single GFile instance.  Ideally also things like an empty write and a write that overwrites an existing file.  And ideally also for the S3 implementation.

tensorboard/summary/writer/event_file_writer_s3_test.py
```diff
@@ -0,0 +1,176 @@
...
+    logdir = s3_temp_dir()
+
+    w = EventFileWriter(logdir, filename_suffix='.event_horizon')
+    w.close()
+    event_files = sorted(gfile.glob(s3_join(logdir, '*')))
+    self.assertEqual(event_files[0].split('.')[-1], 'event_horizon')
+
+  @unittest.skipIf(USING_REAL_TF, 'Test only passes when using stub TF')
+  @mock_s3
+  def test_async_writer_without_write(self):
```
nfelt: "async_writer" is misplaced in this test name, since this is a test of EventFileWriter

tensorboard/summary/writer/event_file_writer_s3_test.py
```diff
@@ -0,0 +1,176 @@
...
+    w = EventFileWriter(logdir)
+    w.close()
+    event_files = sorted(gfile.glob(s3_join(logdir, '*')))
+    r = PyRecordReader_New(event_files[0])
+    r.GetNext()
+    s = event_pb2.Event.FromString(r.record())
+    self.assertEqual(s.file_version, "brain.Event:2")
+
+
+class AsyncWriterTest(tb_test.TestCase):
```
nfelt: Factoring out AsyncWriter was supposed to allow it to be tested independent of the underlying filesystem implementation, since it doesn't auto-create a writer (only EventFileWriter does that).  So I'm not sure that it's necessary to have these tests for AsyncWriter with an S3 backend specifically.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -366,6 +415,16 @@ def read(self, n=None):
...
 
+    def write(self, file_content):
+        if not self.write_mode:
+            raise errors.OpError(None, None, "File not opened in write mode.")
+
+        # add to buffer, but wait for flush to write to filesystem
+        if self.write_buff is None:
+            self.write_buff = file_content
+        else:
+            self.write_buff += file_content
```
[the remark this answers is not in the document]
orionr: I like the tempfile idea. Going with that.

tensorboard/compat/tensorflow_stub/io/gfile_test.py
orionr: You're right! Changing.

tensorboard/summary/writer/event_file_writer_s3_test.py
orionr: event_file_writer_test.py had these and I was looking for parity in the S3 case, but I'm okay with removing.

tensorboard/compat/tensorflow_stub/io/gfile.py
orionr: I'm a little bit nervous about this since we'll be passing down from torch.utils.tensorboard. We can do this in the GFile class implementation instead if you want.

tensorboard/compat/tensorflow_stub/io/gfile.py
orionr: UTF-8 is the standard and I think everything gets stored as that in the default case.

orionr:
@nfelt this should be ready for a final review and hopefully land. Thanks!

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -203,7 +248,9 @@ def read(self, filename, binary_mode=False, size=None, offset=None):
             if offset is None:
                 offset = 0
             endpoint = '' if size is None else (offset + size)
-            args['Range'] = 'bytes={}-{}'.format(offset, endpoint)
+            if offset != 0 or endpoint != '':
```
orionr: With our new tests we actually hit a good failure case where the file is empty and we request "0-", which failed in boto. This fixes that issue.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -101,14 +108,45 @@ def read(self, filename, binary_mode=False, size=None, offset=None):
...
+            filename: string, a path
+            file_content: string, the contents to append
+            binary_mode: bool, write as binary if True, otherwise text
+        """
+        self._write(filename, file_content, "ab" if binary_mode else "a")
+
+    def _write(self, filename, file_content, mode):
+        if "b" in mode:
+            encoding = None
+            file_content = compat.as_bytes(file_content)
```
nfelt: I'd still prefer to avoid doing the auto-conversions here between string and bytes - I think it's just better to start out with the stricter behavior and if that proves too onerous we could relax it later, whereas it's backwards incompatible to go the other direction.  For torch.utils.tensorboard I think all it requires is being clear about whether you're writing text or bytes at a given callsite which should be straightforward - and I'd expect that generally you're only dealing with bytes, right?  Certainly for the EventFileWriter it's only bytes.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -395,8 +483,37 @@ def __next__(self):
...
+    def flush(self):
+        if self.write_temp is not None:
+            # read temp file from the beginning
+            self.write_temp.flush()
+            self.write_temp.seek(0)
+            chunk = self.write_temp.read()
+            if chunk is not None:
+                # use append if filesystem supports it
+                if self.fs_supports_append:
+                    if not self.write_started:
```
nfelt: I don't think you actually need to track a `write_started`, do you?  You can call append() even if the file doesn't already exist and in that case it will create it.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -366,6 +445,15 @@ def read(self, n=None):
...
         return result
 
+    def write(self, file_content):
+        if not self.write_mode:
+            raise errors.OpError(None, None, "File not opened in write mode")
+
+        # add to temp file, but wait for flush to write to final filesystem
+        if self.write_temp is None:
+            self.write_temp = tempfile.TemporaryFile("w+b")
+        self.write_temp.write(compat.as_bytes(file_content))
```
nfelt: Again I'd like to be strict here - we should open the tempfile in binary mode to match self.binary_mode, and avoid doing the auto-conversion.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -395,8 +483,37 @@ def __next__(self):
...
+                    self.write_temp.seek(len(chunk))
+
     def close(self):
-        pass
+        self.flush()
+        if  self.write_temp is not None:
+            self.write_temp.close()
+            self.write_temp = None
+            self.write_started = False
+        self.closed = True
```
nfelt: We never read the value of `closed` - we probably should to e.g. to block a write() or flush() on a closed file.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -366,6 +445,15 @@ def read(self, n=None):
 
         return result
 
+    def write(self, file_content):
```
nfelt: Please add a docstring here explaining that this always appends to the end of the file.

tensorboard/compat/tensorflow_stub/io/gfile.py
orionr: Actually, we do need this. If a file already has content (thank you for the test coverage suggestions) then we will just append to that contents on first write rather than clearing it out. This is the most efficient way I could see to handle this. Let me know your thoughts, though.

orionr: Also, this code will move up into `write()`.

tensorboard/compat/tensorflow_stub/io/gfile.py
orionr: I don't think that matches TensorFlow GFile semantics, but maybe I'm wrong? Seems like it should operate like open(filename, mode), which means it clears out the contents and then appends on each write. How about this for docstring:

```
        """Writes string file contents to file, clearing contents of the
        file on first write and then appending on subsequent calls.

        Args:
            file_content: string, the contents
        """
```

tensorboard/compat/tensorflow_stub/io/gfile.py
orionr: Let me see what I can do.

orionr:
@nfelt, changes made. Please review when you have a chance and thanks. I left comments open that  weren't directly addressed.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -446,13 +451,31 @@ def read(self, n=None):
...
-        # add to temp file, but wait for flush to write to final filesystem
-        if self.write_temp is None:
-            self.write_temp = tempfile.TemporaryFile("w+b")
-        self.write_temp.write(compat.as_bytes(file_content))
+        if self.closed:
+            raise errors.OpError(None, None, "File already closed")
+
+        if self.fs_supports_append:
+            if not self.write_started:
+                # write the first chunk
```
nfelt: Ok, I see now, but can you add "to truncate the file if it already exists" to this comment to clarify why it's needed?

tensorboard/compat/tensorflow_stub/io/gfile.py
nfelt: Unresolving since I think this is still an issue.  I think the fix is just to change the if statement here so that if we're if we're in binary mode, we just check that we got a `six.binary_type` for `file_content` and raise a TypeError if we didn't, else if we're not in binary mode, we call `compat.as_bytes(file_content)` to convert the text string to a binary string.

That way we ensure we're always passing a binary string to put_object().

tensorboard/compat/tensorflow_stub/io/gfile.py
orionr: Got it - thanks for the clarification.

tensorboard/compat/tensorflow_stub/io/gfile.py
orionr: Will do.

PR #2262, opened 2019-05-22
Add more options for TPU profiling

qiuminxu:
StartTracing API has more options to config how to profile a model. Add that to the capture profile pop up.
1) Add work_list, num_tracing_attempts and include_dataset_ops to the pop up.
2) Trim the string of the ProfilerServiceAddress to remove leading and trailing spaces.
3) Add auto-validate to the input field.
4) Fix a bug that the "_profilerServiceAddress" should be passed into the shouldDisableCaptureProfileButton instead of the "profilerServiceAddress".

* Screenshots of UI changes
capture_profile_button

bazel run tensorboard/plugins/profile:profile_demo
bazel run tensorboard:tensorboard -- --logdir=/tmp/profile_demo

### DIFF
--- a/tensorboard/plugins/profile/profile_plugin.py
+++ b/tensorboard/plugins/profile/profile_plugin.py
@@ -427,11 +427,11 @@ class ProfilePlugin(base_plugin.TBPlugin):
   @wrappers.Request.application
   def capture_route(self, request):
     service_addr = request.args.get('service_addr')
-    duration = int(request.args.get('duration'))
+    duration = int(request.args.get('duration', '1000'))
     is_tpu_name = request.args.get('is_tpu_name') == 'true'
     worker_list = request.args.get('worker_list')
     include_dataset_ops = request.args.get('include_dataset_ops') == 'true'
-    num_tracing_attempts = int(request.args.get('num_retry')) + 1
+    num_tracing_attempts = int(request.args.get('num_retry', '0')) + 1
 
     if is_tpu_name:
       try:
@@ -455,7 +455,7 @@ class ProfilePlugin(base_plugin.TBPlugin):
       self.master_tpu_unsecure_channel = master_ip
     try:
       profiler_client.start_tracing(service_addr, self.logdir, duration,
-          worker_list, include_dataset_ops, num_tracing_attempts);
+          worker_list, include_dataset_ops, num_tracing_attempts)
       return http_util.Respond(
           request, {'result': 'Capture profile successfully. Please refresh.'},
           'application/json')
--- a/tensorboard/plugins/profile/profile_plugin.py
+++ b/tensorboard/plugins/profile/profile_plugin.py
@@ -431,7 +431,7 @@ class ProfilePlugin(base_plugin.TBPlugin):
     is_tpu_name = request.args.get('is_tpu_name') == 'true'
     worker_list = request.args.get('worker_list')
     include_dataset_ops = request.args.get('include_dataset_ops') == 'true'
-    num_tracing_attempts = int(request.args.get('num_tracing_attempts'))
+    num_tracing_attempts = int(request.args.get('num_retry')) + 1
 
     if is_tpu_name:
       try:
--- a/tensorboard/plugins/profile/profile_plugin.py
+++ b/tensorboard/plugins/profile/profile_plugin.py
@@ -73,14 +73,14 @@ def process_raw_trace(raw_trace):
   trace.ParseFromString(raw_trace)
   return ''.join(trace_events_json.TraceEventsJsonStream(trace))
 
-def get_workers_list(cluster_resolver):
+def get_worker_list(cluster_resolver):
   """Parses TPU workers list from the cluster resolver."""
   cluster_spec = cluster_resolver.cluster_spec()
   task_indices = cluster_spec.task_indices('worker')
-  workers_list = [
+  worker_list = [
       cluster_spec.task_address('worker', i).split(':')[0] for i in task_indices
   ]
-  return ','.join(workers_list)
+  return ','.join(worker_list)
 
 class ProfilePlugin(base_plugin.TBPlugin):
   """Profile Plugin for TensorBoard."""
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -81,11 +83,34 @@ <h3>Profiling isn’t supported in Colab yet.</h3>
...
+    <paper-input label="Automatically retry N times when no trace event is collected"
+                 always-float-label
+                 type="number"
+                 min=1
+                 max=100
+                 auto-validate
+                 pattern="[0-9]+"
+                 error-message="Input is not an integer between 1 to 100"
+                 required
+                 value="{{_profileNumTracingAttempts}}"></paper-input>
```
qqfish: This option should be useful for GPU as well

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -81,11 +83,34 @@ <h3>Profiling isn’t supported in Colab yet.</h3>
...
                type="number"
                min=1
+               max=600000
+               auto-validate
+               pattern="[0-9]+"
+               error-message="Input is not an integer between 1 to 600000"
+               required
                value="{{_profileDuration}}"></paper-input>
+  <div hidden="{{!_shouldShowTpuOptions(_profilerServiceAddressType)}}">
+    <paper-checkbox checked="{{_profileIncludeDatasetOps}}">Trace dataset ops</paper-checkbox>
```
qqfish: What are the differences between dataset ops and other cpu activities? Why do we need to turn this on explicitly for TPU? I think for GPU, all dataset ops will be traced once turn on cpu profiling.

qiuminxu: The dataset ops was found to be dominant in the CPU traces, removing that allows us to see more TPU device events. As you said, the filtering was not done for GPU, therefore I set it as an TPU only option.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
qiuminxu: Done. Moved that to show in GPU capture profile page as well.

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -427,18 +427,16 @@ def data_route(self, request):
   @wrappers.Request.application
   def capture_route(self, request):
     service_addr = request.args.get('service_addr')
-    try:
-      duration = int(request.args.get('duration'))
-    except TypeError:
-      # Returns code=200 to show error message in UI.
-      return http_util.Respond(request, {'error': 'invalid duration.'},
-                               'application/json', code=200)
+    duration = int(request.args.get('duration'))
```
wchargin: Is `duration` intended to be an optional query parameter? If so, you’ll
need to provide a default here (`request.args.get('duration', '0')`), or
the `int(None)` conversion will be a type error. If it’s intended to be
mandatory, please use `request.args['duration']` to make that clear.

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -427,18 +427,16 @@ def data_route(self, request):
...
     is_tpu_name = request.args.get('is_tpu_name') == 'true'
-    workers_list = ''
+    worker_list = request.args.get('worker_list')
+    include_dataset_ops = request.args.get('include_dataset_ops') == 'true'
+    num_tracing_attempts = int(request.args.get('num_retry')) + 1
```
wchargin: As with `duration` above.

tensorboard/plugins/profile/profile_plugin.py
```diff
@@ -447,16 +445,17 @@ def capture_route(self, request):
...
       # on which profiler service is running.
       master_ip = master_grpc_addr.replace('grpc://', '').replace(':8470', '')
       service_addr = master_ip + ':8466'
       # Set the master TPU for streaming trace viewer.
       self.master_tpu_unsecure_channel = master_ip
     try:
-      profiler_client.start_tracing(service_addr, self.logdir,
-                                    duration, workers_list);
+      profiler_client.start_tracing(service_addr, self.logdir, duration,
+          worker_list, include_dataset_ops, num_tracing_attempts);
```
wchargin: nit: Superfluous semicolon on this line.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -388,6 +415,17 @@ <h3>No profile data was found.</h3>
         type: Number,
         value: 1000,
       },
+      _profileNumRetry: {
+        type: Number,
```
wchargin: Here `_profileNumRetry` is declared as a number, but below you compare
it for strict equality with `""`. It seems like one of these must be
wrong—does the `paper-input` provide a string or a number?

tensorboard/plugins/profile/profile_plugin.py
qiuminxu: Done, set a default duration to 1s.

tensorboard/plugins/profile/profile_plugin.py
qiuminxu: Done.

tensorboard/plugins/profile/profile_plugin.py
qiuminxu: Done, thanks for spotting this.

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
qiuminxu: Yes, it is String. Also fixed for _profileDuration

tensorboard/plugins/profile/tf_profile_dashboard/tf-profile-dashboard.html
```diff
@@ -385,9 +412,20 @@ <h3>No profile data was found.</h3>
         value: "ip-address",
       },
       _profileDuration: {
-        type: Number,
+        type: String,
         value: 1000,
       },
+      _profileNumRetry: {
+        type: String,
+        value: 3,
```
wchargin: If `type` is `String`, then `value` should be `"3"`, not `3`, right?

(Likewise for `_profileDuration` above.)

qiuminxu: Yep, done, thanks :)

PR #2265, opened 2019-05-22
Add demo application for mesh plugin

podlipensky:
Summary:
This includes a parser for simple ASCII PLY files (intentionally
restricted to those that can be parsed by a simple implementation).

Test Plan:
With TensorFlow 1.x, run

```
bazel run //tensorboard/plugins/mesh:mesh_demo -- \
    --mesh_path="${PWD}/tensorboard/plugins/mesh/test_data/icosphere.ply"
```

and then point TensorBoard at `/tmp/mesh_demo` to see some mesh data.
Drag the step slider to see different colors at every step.


tensorboard/plugins/mesh/demo_utils.py
```diff
@@ -0,0 +1,91 @@
...
+  Returns:
+    numpy `[dim_1, 3]` array of vertices, `[dim_1, 3]` array of colors and 
+    `[dim_1, 3]` array of faces of the mesh.
+  """
+  header_size = 0
+  vertices = []
+  colors = []
+  faces = []
+  with tf.gfile.Open(filename) as ply_file:
+    lines = ply_file.readlines()
```
wchargin: This doesn’t need to read the whole file into memory (which is
intentional in the PLY format design). The streaming code is less
susceptible to off-by-one errors in making sure that the ranges match
up, and imho easier to read:

```python
  with tf.gfile.Open(filename) as ply_file:
    for line in ply_file:
      if line.startswith('end_header'):
        break
      elif line.startswith('element vertex'):
        vert_count = int(line.split()[-1])
      elif line.startswith('element face'):
        face_count = int(line.split()[-1])
    vertex_data = [_parse_vertex(next(ply_file)) for _ in range(vert_count)]
    vertices = [datum[0] for datum in vertex_data]
    colors = [datum[1] for datum in vertex_data]
    faces = [_parse_face(next(ply_file)) for _ in range(face_count)]
  return (np.array(vertices), np.array(colors), np.array(faces))
```

No need to track `header_lines` or perform any indexing at all.

(You can swap out the `vertex_data` parsing for an all-in-one

```python
    (vertices, colors) = zip(
        *(_parse_vertex(next(ply_file)) for _ in range(vert_count))
    )
```

but personally I find that a bit less readable.)

tensorboard/plugins/mesh/demo_utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def _parse_vertex(vertex_row):
+  """Parses a line in a PLY file which encodes a vertex coordinates.
+
+  Args:
+    vertex_row: string with vertex coordinates and color.
+
+  Returns:
+    `[3,]` array of vertex coordinates and `[3,]` array with RGB color.
```
wchargin: Here you use “array” to mean `list`, but below you use it to instead
mean `ndarray`. Suggest rephrasing to:

```
  Returns:
    2-tuple containing a length-3 array of vertex coordinates (as
    floats) and a length-3 array of RGB color values (as ints between 0
    and 255, inclusive).
```

Speaking of which, shouldn’t the colors be returned as `int`s?

tensorboard/plugins/mesh/demo_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    # Read vertices and their colors.
+    for i in range(header_size, header_size + vert_count):
+      vertex, color = _parse_vertex(lines[i])
+      vertices.append(vertex)      
+      colors.append(color)
+    # Read faces.
+    for i in range(
+      header_size + vert_count, header_size + vert_count + face_count):
+      faces.append(_parse_face(lines[i]))
+    return np.array(vertices), np.array(colors), np.array(faces)
```
wchargin: Shouldn’t `colors` be of uint8 dtype? You’re currently relying on
implicit float-to-uint8 conversion.

tensorboard/plugins/mesh/demo_utils.py
```diff
@@ -0,0 +1,91 @@
...
+def _parse_vertex(vertex_row):
+  """Parses a line in a PLY file which encodes a vertex coordinates.
+
+  Args:
+    vertex_row: string with vertex coordinates and color.
+
+  Returns:
+    `[3,]` array of vertex coordinates and `[3,]` array with RGB color.
+  """
+  vertex = map(float, vertex_row.strip().split(' '))
```
wchargin: Consider using `.split()` instead of `.split(" ")` to split on arbitrary
whitespace sequences (which I believe valid PLY; Blender accepts it).

wchargin: This doesn’t do what you want in Python 3, because the result of `map`
is a generator and cannot be subscripted. Use of `map` is generally
discouraged in Python, anyway; prefer

```python
  vertex = [float(x) for x in vertex_row.strip().split()]
```

Likewise below with faces.

tensorboard/plugins/mesh/demo_utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+  Returns:
+    numpy `[dim_1, 3]` array of vertices, `[dim_1, 3]` array of colors and 
+    `[dim_1, 3]` array of faces of the mesh.
+  """
+  header_size = 0
+  vertices = []
+  colors = []
+  faces = []
+  with tf.gfile.Open(filename) as ply_file:
```
wchargin: The `tf.gfile` module is deprecated and removed in TF 2.0. Please use
`tf.compat.v2.io.gfile.GFile` instead of `tf.gfile.Open`.

tensorboard/plugins/mesh/README.md
```diff
@@ -92,6 +92,20 @@ camera_config = {
 Keep in mind that scene configuration is not a trainable variable (i.e., it is 
 a static attribute) and is fixed at summary-creation time.
 
+## Example
+
+Please refer to the demo application `demo_utils.py` as an example of how to
```
wchargin: The demo application is not `demo_utils.py`.

tensorboard/plugins/mesh/README.md
```diff
@@ -92,6 +92,20 @@ camera_config = {
...
+use the plugin and display a mesh with colors in TensorBoard. Note, that demo
+application will read PLY file in ASCII format and doesn't support all
+the variations of PLY format. You can find a sample mesh in PLY format [stored
```
wchargin: Users running the demo already have the test data file and don’t need to
fetch anything from the cloud. Can we point them to the `test_data/...`
file instead of or in addition to this link?

tensorboard/plugins/mesh/README.md
```diff
@@ -92,6 +92,20 @@ camera_config = {
...
+in the cloud](https://storage.googleapis.com/tensorflow-graphics/tensorboard/test_data/ShortDance07_a175_00013_2.ply).
+Here is a snippet of how to build and run the demo application:
+
+```
+bazel build tensorboard/plugins/mesh:mesh_demo
```
wchargin: This can just be 

```
bazel run //tensorboard/plugins/mesh:mesh_demo -- --mesh_path=path/to/ply/file
```

; no need for manual `bazel-bin`, which doesn’t properly populate the
environment and can only be run from the root directory of the
repository.

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,90 @@
...
+def run():
+  """Runs session with a mesh summary."""
+  
+  # Camera and scene configuration.
+  config_dict = {
+      'camera': {'cls': 'PerspectiveCamera', 'fov': 75}
+  }
+
+  # Read sample PLY file.
+  vertices, colors, faces = demo_utils.read_ascii_ply(FLAGS.mesh_path)
```
wchargin: The failure when users omit this flag is a bit obscure: it’s a traceback
from within TensorFlow whose error message is “TypeError: Expected
binary or unicode string, got None”.

Can you explicitly validate that the flag is not `None` and raise a nice
error message in case of fire? This especially because the other demos
don’t need any explicit file path, so users (like me) will be used to
just running them with no arguments.

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,90 @@
...
+  
+  # Add batch dimension.
+  vertices = np.expand_dims(vertices, 0)
+  faces = np.expand_dims(faces, 0)
+  colors = np.expand_dims(colors, 0)  
+  
+  # Create placeholders for tensors representing the mesh.
+  vertices_tensor = tf.placeholder(
```
wchargin: Why placeholders here, rather than just constant tensors?

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,90 @@
...
+
+  for i in range(10):
+    summaries = sess.run([meshes_summary], feed_dict={
+        vertices_tensor: vertices,
+        faces_tensor: faces,
+        colors_tensor: colors,
+    })
+    # Save summaries.
+    for summary in summaries:
+      writer.add_summary(summary)
```
wchargin: Missing step parameter.

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,90 @@
...
+  meshes_summary = mesh_summary.op(
+      'mesh_color_tensor', vertices=vertices_tensor, faces=faces_tensor,
+      colors=colors_tensor, config_dict=config_dict)
+
+  # Create summary writer and session.
+  writer = tf.summary.FileWriter(FLAGS.logdir)
+  sess = tf.Session()
+
+  for i in range(10):
+    summaries = sess.run([meshes_summary], feed_dict={
```
wchargin: Part of the purpose of the demo is to exercise all the functionality of
the dashboard, for testing purposes. Currently, all the time steps are
the same, so it’s hard to verify that the slider works. How about making
a simple transformation to the mesh and/or colors over these ten time
steps? For example, you could invert the colors over time:

```python
  t = step / MAX_STEPS
  transformed_colors = t * (255 - colors) + (1 - t) * colors
```

tensorboard/plugins/mesh/demo_utils.py
```diff
@@ -0,0 +1,91 @@
...
+  Args:
+    vertex_row: string with vertex coordinates and color.
+
+  Returns:
+    `[3,]` array of vertex coordinates and `[3,]` array with RGB color.
+  """
+  vertex = map(float, vertex_row.strip().split(' '))
+  # The row will contain either just coordinates or RGB/RGBA color in addition
+  # to that.
+  assert len(vertex) == 3 or len(vertex) == 6 or len(vertex) == 7
```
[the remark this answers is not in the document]
podlipensky: I think this assertion is not really necessary since this code is meant to be executed on particular .ply file.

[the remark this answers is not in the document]
wchargin: ? By design, it can be executed on any PLY file; it’s controlled by a
command-line argument.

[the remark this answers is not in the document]
wchargin: If it’s possible to fail the assertion using only the public APIs for
the module (here, `read_ascii_ply`), then it shouldn’t be an assertion.

tensorboard/plugins/mesh/demo_utils.py
podlipensky: good feedback, thank you

tensorboard/plugins/mesh/demo_utils.py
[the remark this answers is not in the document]
podlipensky: no, it wasn't intended to run on any PLY file, as I mentioned in a comment. I think adding explicit checks to this reader will just complicate code since there are quite few valid variations of PLY file. I'd let it fail with default errors.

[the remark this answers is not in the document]
wchargin: It doesn’t have to be able to parse arbitrary PLY files. It’s fine if it
raises an exception on files that other PLY files would accept. But it
cannot have an assertion failure. An assertion failure by definition
means that the _code_ is wrong.

All I’m asking is that

```python
  assert len(vertex) == 3 or len(vertex) == 6 or len(vertex) == 7
```

be changed to

```
  if len(vertex) not in (3, 6, 7):
    raise ValueError('Vertices must have 3, 6, or 7 entries; got: %r' % vertex)
```

tensorboard/plugins/mesh/mesh_demo.py
podlipensky: good idea

tensorboard/plugins/mesh/mesh_demo.py
podlipensky: could use constant, but since we going to modify color, I'll keep placeholders.

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,97 @@
...
+        faces_tensor: faces,
+        colors_tensor: transformed_colors.astype(np.uint8),
+    })
+    # Save summaries.
+    for summary in summaries:
+      writer.add_summary(summary, global_step=i)
+
+
+def main(unused_argv):
+  tf.logging.info('Saving output to %s.', FLAGS.logdir)
```
wchargin: The `tf.logging` module is removed in TF 2.0. Most of the other demos
just use `print` calls here. If you need a logging mechanism for some
other reason, please use `tb_logging.get_logger()`.

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,97 @@
...
+from tensorboard.plugins.mesh import summary as mesh_summary
+from tensorboard.plugins.mesh import demo_utils
+
+flags.DEFINE_string('logdir', '/tmp/mesh_demo',
+                    'Directory to write event logs to.')
+flags.DEFINE_string('mesh_path', None, 'Path to PLY file to visualize.')
+
+FLAGS = flags.FLAGS
+
+tf.compat.v1.disable_v2_behavior()
```
wchargin: This is not expected to work at all in TF 2.0, right? Can we explicitly fail in that case, rather than leaving the user to deduce why `tf.placeholder` does not exist?

```python
if int(tf.__version__.split(".")[0]) > 1:
  # Mesh summaries only work on TensorFlow 1.x.
  raise ImportError("TensorFlow 1.x is required to run this demo.")
```

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,97 @@
...
+from __future__ import print_function
+
+import os
+
+from absl import app
+from absl import flags
+import numpy as np
+import tensorflow as tf
+
+
```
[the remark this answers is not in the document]
podlipensky: Don't we do double spaces for inter-module functions/classes?

tensorboard/plugins/mesh/demo_utils.py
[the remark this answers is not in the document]
podlipensky: I removed assertion.

tensorboard/plugins/mesh/mesh_demo.py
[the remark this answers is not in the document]
wchargin: We do use two blank lines between functions/classes in a module, but
this double-newline is between two import blocks.

Instead of:

```python
import tensorflow as tf


from tensorboard.plugins.mesh import summary as mesh_summary
from tensorboard.plugins.mesh import demo_utils

flags.DEFINE_string(...)
```

it should be:

```python
import tensorflow as tf

from tensorboard.plugins.mesh import summary as mesh_summary
from tensorboard.plugins.mesh import demo_utils


flags.DEFINE_string(...)
```

tensorboard/plugins/mesh/BUILD
```diff
@@ -125,3 +125,51 @@ tb_proto_library(
...
+    deps = [
+        "//tensorboard:expect_numpy_installed",
+        "//tensorboard/compat:tensorflow",
+    ],
+)
+
+py_test(
+    name = "demo_utils_test",
+    size = "small",
+    srcs = ["demo_utils_test.py"],
```
wchargin: There’s still extra whitespace at the end of this line.

tensorboard/plugins/mesh/demo_utils.py
```diff
@@ -0,0 +1,86 @@
...
+  """Reads a PLY file encoded in ASCII format.
+
+  NOTE: this util method is not intended to be comprehensive PLY reader
+  and serves as part of demo application.
+
+  Args:
+    filename: path to a PLY file to read.
+
+  Returns:
+    numpy `[dim_1, 3]` array of vertices, `[dim_1, 3]` array of colors and
```
wchargin: There’s still extra whitespace at the end of this line.

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,102 @@
...
+    raise ValueError(
+      'Flag --mesh_path is required and must contain path to PLY file.')
+  # Camera and scene configuration.
+  config_dict = {
+      'camera': {'cls': 'PerspectiveCamera', 'fov': 75}
+  }
+
+  # Read sample PLY file.
+  vertices, colors, faces = demo_utils.read_ascii_ply(FLAGS.mesh_path)
+
```
wchargin: There’s still extra whitespace at the end of this line.

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  # Read sample PLY file.
+  vertices, colors, faces = demo_utils.read_ascii_ply(FLAGS.mesh_path)
+  
+  # Add batch dimension.
+  vertices = np.expand_dims(vertices, 0)
+  faces = np.expand_dims(faces, 0)
+  colors = np.expand_dims(colors, 0)
+
```
wchargin: There’s still extra whitespace at the end of this line.

tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -0,0 +1,102 @@
...
+  transformed_colors = t * (255 - colors) + (1 - t) * colors
+
+  meshes_summary = mesh_summary.op(
+      'mesh_color_tensor', vertices=vertices_tensor, faces=faces_tensor,
+      colors=transformed_colors, config_dict=config_dict)
+
+  # Create summary writer and session.
+  writer = tf.summary.FileWriter(FLAGS.logdir)
+  sess = tf.Session()
+
```
wchargin: There’s still extra whitespace at the end of this line.

tensorboard/plugins/mesh/test_data/icosphere.ply
```diff
@@ -0,0 +1,189 @@
+ply
```
wchargin: Somehow this file was changed to have CRLF line endings. Please fix it
to use standard LF-only line endings.

podlipensky: hm, I see LF-only (double checked by trying to match \r\n pattern). Are you sure it is not on your end only?

wchargin: I’m sure:

```
$ git clone -q https://github.com/podlipensky/tensorboard.git
$ cd tensorboard/
$ git checkout -q origin/tf-graphics-demo2
$ git rev-parse --verify HEAD
1073a1be70a431208f2694e9571e38de4196f3db
$ # normal file, LF only:
$ head -n 1 tensorboard/plugins/mesh/README.md | xxd
00000000: 2320 4d65 7368 2050 6c75 6769 6e0a       # Mesh Plugin.
$ # this PLY file, CRLF:
$ head -n 1 tensorboard/plugins/mesh/test_data/*.ply | xxd
00000000: 706c 790d 0a                             ply..
$ file tensorboard/plugins/mesh/test_data/*.ply
tensorboard/plugins/mesh/test_data/icosphere.ply: ASCII text, with CRLF line terminators
```

Furthermore, if you use `git show af14d2e2 '*/icosphere.ply'` to inspect
the commit in which the file was introduced, you will see that the diff
lines contain carriage returns, indicated by `^M` symbols:

```
--- /dev/null
+++ b/tensorboard/plugins/mesh/test_data/icosphere.ply
@@ -0,0 +1,189 @@
+ply^M
+format ascii 1.0^M
```

wchargin: Please apply the following patch (`git am 0001-*.txt`) and push it to
your branch:

[0001-mesh-remove-carriage-returns-from-test-data.txt](https://github.com/tensorflow/tensorboard/files/3254533/0001-mesh-remove-carriage-returns-from-test-data.txt)

PR #2266, opened 2019-05-22
vz-example-viewer: fix int feature list type

wchargin:
Summary:
Caught during a Clutz upgrade. From context, it looks like `ints` should
be an `Int64List`, as is already the case in `parseFeature` below.

Test Plan:
This preserves existing behavior. To exercise the codepath, run

```
bazel run //tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/demo:agedemoserver
```

then navigate to `/tf-interactive-inference-dashboard/age_demo.html` in
a browser. Select a data point in the plot. In the list of features for
that data point (on the left side of the screen), click the “+” button
immediately below the bottom feature. Add an `Int` feature, and set its
value to some number. Select a different data point, then select the
first data point again, and observe that the new feature and value are
still present.

wchargin-branch: vzev-ints-list-type


wchargin:
@jameswex: How should this be tested?

jameswex:
I tested this change locally by running a WIT demo, adding a new int feature in the UI, setting it to an int value, clicking away to another example, then back to the changed example and seeing the value correctly persisted.

wchargin:
Thanks!

> I tested this change locally by running a WIT demo, adding a new int
> feature in the UI, setting it to an int value, clicking away to
> another example, then back to the changed example and seeing the value
> correctly persisted.

I did this (I think), but when I set the value to `1234.5`, clicked a
different data point, and clicked back, the value was still `1234.5`
(rather than being truncated to `1234` or left absent). This occurs both
before and after this change. Is this intended?

Is there an observable difference before vs. after this change, or is it
just hygiene?

jameswex:
hygiene as since the example is passed from js back to python as json, this never exhibited any outwardly-bad behavior from what i can tell.

wchargin:
Understood; thanks.

PR #2267, opened 2019-05-22
build: fix forward to Bazel 0.26

wchargin:
Summary:
A bunch of changes here required due to `--incompatible_depset_union`
and consequent transitive upgrades.

The new depset union makes inefficient use more apparent. I opted to
change our own uses as locally and minimally as possible rather than
fixing them to be more efficient, for the sake of keeping things simple.
(Or, rather, I first tried to do things the efficient way by
accumulating lists of transitive depsets, but due to lack of types the
refactoring was too hard to keep track of to be worth it. The simple
option just requires application of one of two Vim macros at each call
site, with just a bit of thought to guess which one is right.)

The change to `Vulcanize.java` was already present within google3.

Test Plan:
Both `bazel build //tensorboard` and `bazel test //...` work with Bazel
version 0.26.0rc12.

wchargin-branch: bazel-0.26.0rc12


tensorboard/defs/web.bzl
```diff
@@ -100,7 +106,7 @@ def _tf_web_library(ctx):
       new_typings.append(src)
       new_typings_paths.append(entry[0])
       new_typings_execroot.inputs.append(entry)
-      ts_inputs += [src]
+      ts_inputs = depset([src], transitive=[ts_inputs])
```
stephanwlee: stupid question: I see two patterns of concatenation:
```py
# first
a = depset(transitive=[b, a])
# second
a = depset([b], transitive=[a])
```

Are they functionally different? If not, what is the reason to prefer one over the other?

tensorboard/components/vz_example_viewer/vz-example-viewer.ts
```diff
@@ -924,7 +924,7 @@ Polymer({
       const valueList: number[] = [];
       const ints = new FloatList();
       ints.setValueList(valueList);
-      feat.setInt64List(ints);
+      feat.setInt64List(ints as any);  // TODO(#2266): Preserves behavior.
```
stephanwlee: Is this change related to this PR?

wchargin: Yes, it is: the upgrade to Clutz causes this to be an error without the
any-cast. PR #2266 actually fixes it, but I didn’t want to introduce a
(potentially) functional change here.

The upgrade to Clutz is required by the upgrade to the Closure compiler,
which is required by the upgrade to `rules_closure`, which is required
to fix the `--incompatible_depset_union`s, which are errors in the new
Bazel version. It’s a bundle of fun.

tensorboard/defs/web.bzl
wchargin: Yes, they are different. The elements of the `transitive` list must be
depsets, whereas the elements of `items` (the first positional
parameter) must be files. If you use the wrong one, you get an error.

wchargin: (Good question, by the way.)

PR #2269, opened 2019-05-22
Fix long type not found for json input in python3

tolga-b:
long type does not exist in python 3, changed type testing to six.integer types from long to prevent errors when wit is used with json inputs.

* Screenshots of UI changes
None
Tested cloud xgboost model with python 3 kernel in colab.
None


tolga-b:
@jameswex

tolga-b:
@stephanwlee please review.

tolga-b:
You are right. We realized that BUILD file for witwidget is missing a lot of dependencies and it mainly works because the pip package depends on tensorflow which has them. We will update the witwidget BUILD file in a separate PR.

PR #2273, opened 2019-05-23
Polymer 2: changes to tf-paginated-view

stephanwlee:
Better details are captured in the commit message. Please refer to them for more details.

Now that components are "mounted" (`Polymer.prototype.isAttached` is more or less deprecated)
at the creation time (dom-if does not help), we now have issues with our strategy in defer-data-fetching.

The complexity of this lifecycle change grew especially for dashboard using tf-category-pane and tf-paginated-view together -- both components separately decide whether a page/chart will appear on
the screen. Now that data-loaders cannot rely on attached-ness to detect the activeness, we, somehow,
have to propagate information from tf-category-pane and tf-paginated-view down to these loaders.
This, however, is not as easy as just getting a reference to the component and imperatively setting
the component property.

The charts are created/owned by dashboards but they are hosted by dom-repeat inside
tf-pagianted-view. Disregarding tf-paginated-view also requiring to pass information down from
the tf-category-pane, in order to imperatively set the property, tf-paginated-view has to make
assumption on how inner content will look (i.e., imagine querySelecting for data-loader but should
paginated-view really know about these data-loaders?). The most Polymeric way of doing this would
be passing information up to the owner (dashboard) and pass it back down to the right components.
This method was attempted but caused so much code bloat that had to be repeated by many
dashboards that I decided to take another route.

While seeking solutions, I got inspired by how dom-repeat passes down information and decided
to replicate that and also decided to collapse tf-category-pane and tf-paginated-view into one
to remove complexity. Do note that I was unable to remove tf-category-pane just yet due to one
use case that does not use tf-paginated-view.


stephanwlee:
To reviewers: I would strongly recommend you to hide whitespace changes and, of course, review commit by commit. Also do note that large part of tf-category-paginated-view is a copypasta from tf-category-pane (didn't know how to tell git that a new file is a merge of two files).

tensorboard/components/tf_paginated_view/tf-paginated-view2.html
```diff
@@ -346,13 +361,17 @@
...
+      }
+
       let fragOrEl;
       const key = this.getItemKey(item);
       if (this._lruCachedItems.has(key)) {
         fragOrEl = this._lruCachedItems.get(key);
         this._lruCachedItems.delete(key);
       } else {
-        const inst = this.stamp({[this.as]: item});
+        const inst = new this.__ctor({[this.as]: item});
```
wchargin: Above you use `this._ctor`, but here you use `this.__ctor`. This seems
confusing. What’s the difference between the two?

tensorboard/components/tf_paginated_view/tf-paginated-view2.html
```diff
@@ -256,16 +256,30 @@
...
-      // Template instance props that should be excluded from forwarding
-     this._instanceProps = {
-       [this.as]: true,
-     };
-
-      this.templatize(this.$.template.querySelector('template'));
+    _ensureTemplatized() {
+      // Polymer is not ready (and props/DOM for the components are not
+      // populated)
+      if (!this.isAttached) return false;
```
wchargin: If `isAttached` is deprecated (as suggested in the PR description), do
we want to use an alternative here? Like, manage an `_attached` property
ourselves?

(I also can’t find docs for `isAttached` anywhere: neither the [API docs
for `Polymer.Base`][base] nor the [“Registration and lifecycle”
guide][ral]. What do you use for a reference?)

[base]: https://polymer-library.polymer-project.org/1.0/api/classes/Polymer.Base
[ral]: https://polymer-library.polymer-project.org/1.0/docs/devguide/registering-elements

tensorboard/components/tf_paginated_view/tf-paginated-view2.html
```diff
@@ -256,16 +256,30 @@
...
+      if (!this._ctor) {
+        const templateNode = this.$.template.assignedElements()[0];
+        this._ctor = Polymer.Templatize.templatize(templateNode, this, {
+            parentModel: true,
+            instanceProps: {
+              [this.as]: true,
+              active: true,
+            },
+            forwardHostProp: function(prop, value) {
+              // Polymer uses `call` to invoke this method.
```
wchargin: Not sure what this is supposed to mean. It looks like the function body
expects the `this` context to be the element itself; if so, can’t we
just use an arrow function and drop the comment? If I’m
misunderstanding, can you please clarify?

wchargin:
Just to write down one quick point from an extended offline discussion…

IIUC, a fair amount of the complexity in the Polymer 2 change and in
this PR particularly is because in Polymer 2, the contents of a `dom-if`
will be mounted immediately, even if the `dom-if` condition is false.

Where this is problematic, can we replace

```html
<template is="dom-if" if="[[foo]]">
  …
</template>
```

with

```html
<template is="dom-repeat" items="[[_domRepeatDomain(foo)]]" as="_">
  …
</template>
```

```javascript
function domRepeatDomain(domIfCondition) {
  return domIfCondition ? [null] : [];
}
```

?

wchargin:
Again abbreviating an offline conversation: the hack that I suggested
above does work but is not sufficient. If we ever write

```html
<tf-image-dashboard>
  <tf-category-pane>
    <tf-paginated-view><!-- some contents --></tf-paginated-view>
  </tf-category-pane>
</tf-image-dashboard>
```

then the `tf-paginated-view` _will_ be mounted on the DOM immediately,
no matter what we put in the `tf-category-pane` component. The only way
to prevent this is by using

```html
<tf-image-dashboard>
  <tf-category-pane>
    <template>
      <tf-paginated-view><!-- some contents --></tf-paginated-view>
    </template>
  </tf-category-pane>
</tf-image-dashboard>
```

but in that case we _must_ perform the prop forwarding dance, to keep
props in sync among the owner, host, and child elements.\*

So it doesn’t look like there’s any easy way out in Polymer 2.

\* Terminology and notes:

  - The **owner** is the component that lexically includes the host and
    child elements; in our example, it’s `tf-image-dashboard`.
  - The **host** is the element that does the template stamping and prop
    forwarding; in our example, it’s `tf-category-pane`.
  - The **child** is the element stamped by the host; in our example,
    it’s `tf-paginated-view`.
  - The child can receive props from both the owner _and_ the host, and
    can also notify props back up to both.

tensorboard/components/tf_paginated_view/tf-paginated-view2.html
stephanwlee: I find it super odd that `isAttached` is not present in the doc and this PR does not seem to do anything. https://github.com/Polymer/polymer/pull/4368/files but is present in code:
1. https://github.com/Polymer/polymer/blob/2.x/lib/legacy/legacy-element-mixin.html#L112
2. https://github.com/Polymer/polymer/blob/2.x/types/lib/legacy/legacy-element-mixin.d.ts#L40

Although `isAttached` has very different meaning in Polymer2, the prop is _not_ removed/deprecated.

stephanwlee:
Ok, the GitHub is showing commits in weird order (seems to be ordered by commit creation time :() but we should probably squash bd354c5 with 056cdf1 when we merge this PR. Otherwise, I don't think TensorBoard will build at bd354c5.

stephanwlee:
@wchargin I made a mistake in refactoring custom scalar -- fixed it (I wonder why I didn't notice the build failure). About What-If, they have DOMs that have classes of `tf-categroy-pane` and `tf-category-pagne-content` but it is not really using `<tf-category-pane>`. 

I don't recall, anymore, why I decided not to create a HOC that wraps both tf-category-pane and tf-paginated-view but at this point, we now have no use case of `tf-category-pane` and `tf-paginated-view` separately. If you strongly feel that it is good idea to keep those two separate (readability I think is the critical reason to do so), I will spend some time to do so.

PR #2277, opened 2019-05-24
Polymer 2: fix the line chart tooltip

stephanwlee:
Line chart took a bit different approach in terms of tooltip: because we
showed large content, we wanted it to look reasonable in whichever
orientation appearing above the sidebar using a technical called portal.

The way we portaled previously in Polymer1 no longer works in Polymer2
especially because how real shadow CSS works. Unlike ShadyCSS where it
populated global style tag that is accessible anywhere in the DOM tree,
now, the styles are properly encapsulated inside a component. As a
result, styling for the vz-chart-tooltip became quite difficult.

With this change, to get around the styling issue, the vz-chart-tooltip
now take the name of the custom element and instantiate the element
(of course, with its style defined inside) and attached it to the body
of the document. When doing it, the vz-chart-tooltip style it a bit.


tensorboard/components/vz_line_chart2/vz-line-chart-tooltip.ts
```diff
@@ -0,0 +1,21 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+namespace vz_line_chart2 {
+
+Polymer({
+  is: 'vz-line-chart-tooltip',
```
nfelt: As discussed, some comment about how this (logic-less) component is used (e.g. it's designed to be instantiated by vz-chart-tooltip, and populated by logic from vz-line-chart) would be helpful.

PR #2281, opened 2019-05-27
Fix Step In Mesh Slider

### LINKED ISSUE #2270: Mesh Plugin: slider has 3x more steps than required
Looks like total count of steps in the mesh plugin is miscalculated because of internal representation of the mesh.

abhinavsagar:
[Fixes #2270](https://github.com/tensorflow/tensorboard/issues/2270)

### DIFF
--- a/tensorboard/plugins/mesh/tf_mesh_dashboard/tf-mesh-loader.html
+++ b/tensorboard/plugins/mesh/tf_mesh_dashboard/tf-mesh-loader.html
@@ -55,7 +55,7 @@ limitations under the License.
             max="[[_getMaxStepIndex(_step)]]"
             max-markers="[[_getMaxStepIndex(_step)]]"
             snaps
-            step="1"
+            step="1/3"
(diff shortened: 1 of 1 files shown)


abhinavsagar:
@tafsiri Please review.

wchargin:
Hi @abhinavsagar! Thanks for the pull request.

This change doesn’t seem to work for me. Prior to this change, the mesh
slider indeed had an invalid range (though not exactly off by a factor
of three), but at least dragging the slider back and forth would still
change the step. As of this commit, dragging the slider immediately
locks the step to `0`, because the string “`1/3`” is [not a valid value
for `step`][step]. Changing the step to 1/3 (or 0.333, or similar) also
doesn’t seem like the right fix conceptually: you’d be able to select
“step 2.333”, for one thing, which doesn’t mean anything.

[step]: https://www.w3.org/TR/html51/sec-forms.html#the-step-attribute

Issue #2284, opened 2019-05-29
profile: pod viewer tooltip position


stephanwlee:
The tooltip displayed when hovering on image of a pod not only is positioned suboptimally but also incorrectly positions with respect to the scroll position.

Before scroll:
![image](https://user-images.githubusercontent.com/2547313/58574100-15d82180-81f4-11e9-8683-14d7cbdc376d.png)

After scroll:
![image](https://user-images.githubusercontent.com/2547313/58574132-27212e00-81f4-11e9-9b5d-a85c47b00ff5.png)

Note that although I hovered on the same TPU at col 7 row 9, the tooltip is missing. If I scroll the page up a little while maintaining the cursor on the TPU, you will find the tooltip rendered above the viewport.

![image](https://user-images.githubusercontent.com/2547313/58574149-30aa9600-81f4-11e9-9ad9-9bba09062860.png)

qiuminxu:
Sure, I will take a look on how to better position the tooltip. BTW, I've moved the folder and please use gs://cloud-tpu-tools/test in the future.

PR #2285, opened 2019-05-29
fix tooltip position

qiuminxu:
The tooltip displayed when hovering on image of a pod is positioned suboptimally, this PR position the tooltip better. #2284 

* Screenshots of UI changes
pod_viewer_new3

bazel run tensorboard:tensorboard -- --logdir=gs://cloud-tpu-tools

### DIFF
--- a/tensorboard/plugins/profile/pod_viewer/topology_graph/topology-graph.ts
+++ b/tensorboard/plugins/profile/pod_viewer/topology_graph/topology-graph.ts
@@ -356,8 +356,8 @@ Polymer({
 
             // Update the tooltip position and value
             d3.select(parent.$.tooltip)
-                .style('left', d3.select(this).attr('x') + 'px')
-                .style('top', d3.select(this).attr('y') + 'px')
+                .style('left', (~~d3.select(this).attr('x') + 50) + 'px')
+                .style('top', (~~d3.select(this).attr('y') + 200) + 'px')
                 .select('#value')
                 .text(parent._getToolTipText(d));
             d3.select(parent.$.tooltip).classed('hidden', false);
--- a/tensorboard/plugins/profile/pod_viewer/topology_graph/topology-graph.ts
+++ b/tensorboard/plugins/profile/pod_viewer/topology_graph/topology-graph.ts
@@ -356,8 +356,8 @@ Polymer({
 
             // Update the tooltip position and value
             d3.select(parent.$.tooltip)
-                .style('left', d3.event.pageX + 10 + 'px')
-                .style('top', d3.event.pageY - 10 + 'px')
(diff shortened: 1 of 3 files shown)


tensorboard/plugins/profile/pod_viewer/topology_graph/topology-graph.ts
```diff
@@ -356,8 +356,8 @@ Polymer({
 
             // Update the tooltip position and value
             d3.select(parent.$.tooltip)
-                .style('left', d3.event.pageX + 10 + 'px')
-                .style('top', d3.event.pageY - 10 + 'px')
+                .style('left', d3.select(this).attr('x') + 'px')
```
stephanwlee: Wouldn't you want some padding so it shows right (or bottom) of the current item?

qiuminxu: Done.

tensorboard/plugins/profile/pod_viewer/topology_graph/topology-graph.ts
```diff
@@ -356,8 +356,8 @@ Polymer({
 
             // Update the tooltip position and value
             d3.select(parent.$.tooltip)
-                .style('left', d3.event.pageX + 10 + 'px')
-                .style('top', d3.event.pageY - 10 + 'px')
+                .style('left', (~~d3.select(this).attr('x') + 50) + 'px')
```
stephanwlee: `~~` is a nifty trick but I had to think about it. How about just `Number(d3.select(this).attr('x'))`?

stephanwlee: Also I presume `50` is the width of the tooltip? Can we just measure the bounding client rect instead?

qiuminxu: Sure, done.

qiuminxu: The width of the tooltip is 200. I empirically adjusted the padding and find 50 works.

stephanwlee: I would advise against using such "magic" number. Is it at all impossible to calculate this value based on dimension somewhere?

qiuminxu: I would imagine d3.event.pageX and d3.event.pageY should just work or d3.select(this).attr('x') + rect.width + 5 and d3.select(this).attr('y') + rect.height + 5 should work. But neither works, except for these magic numbers.

PR #2288, opened 2019-05-29
core: always gate data selector behind query param

wchargin:
Summary:
We’d like to serve dashboard registration metadata from the backend, as
will be the case for dynamically loaded plugins. This is straightforward
except for the scalar plugin’s `useDataSelector` bit, which depends on
the query parameter. This patch changes moves the query parameter gating
from the scalar plugin to the core frontend.

Test Plan:
Verify that the following behavior is the same before and after this
change: (a) when the URL has no query parameters, neither the scalars
dashboard nor the images dashboard shows a data selector; and (b) when
the URL has `?EnableDataSelector`, the scalars dashboard has a data
selector and no run selector, while the images dashboard has a run
selector and no data selector.

wchargin-branch: top-level-data-selector-flag

### DIFF
--- a/tensorboard/components/tf_tensorboard/tf-tensorboard.html
+++ b/tensorboard/components/tf_tensorboard/tf-tensorboard.html
@@ -730,18 +730,19 @@ limitations under the License.
           return;
         }
         const dashboard = tf_tensorboard.dashboardRegistry[selectedDashboard];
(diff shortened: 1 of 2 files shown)


wchargin:
FWIW, neither before nor after this commit can I get the scalars
dashboard to actually show me any charts when the data selector is
enabled in “simple mode”. It works fine in “advanced mode”. I’m assuming
that this is a known failure.

PR #2289, opened 2019-05-30
test: ensure that build URLs are valid

wchargin:
Summary:
This tests that URLs point to the TensorFlow mirror (as opposed to the
Bazel mirror) and that they resolve to valid archives.

Test Plan:
Run `./tensorboard/tools/mirror_urls_test.sh`, and note that it passes
with no output. Then apply the following patch to add some bad URLs:

```diff
diff --git a/WORKSPACE b/WORKSPACE
index 3943b0cf..d3975dfe 100644
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -9,6 +9,9 @@ http_archive(
     urls = [
         # tag 0.7.0 resolves to commit 6741f733227dc68137512161a5ce6fcf283e3f58 (2019-02-08 18:37:26 +0100)
         "http://mirror.tensorflow.org/github.com/bazelbuild/bazel-skylib/archive/0.7.0.tar.gz",
+        "http://mirror.tensorflow.org/example.com/nonexistent.txt",
+        "http://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/archive/0.7.0.tar.gz",
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/archive/0.7.0.tar.gz",
         "https://github.com/bazelbuild/bazel-skylib/archive/0.7.0.tar.gz",
     ],
 )
```

Re-run the test, and note that it fails with exit code 1 and this text:

```
The following URLs are not properly mirrored:
  - http://mirror.tensorflow.org/example.com/nonexistent.txt
Please comment on your PR asking a TensorBoard core team member
to mirror these URLs per instructions in http://b/133880558.

The following URLs point to the legacy Bazel mirror:
WORKSPACE:13:        "http://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/archive/0.7.0.tar.gz",
WORKSPACE:14:        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/archive/0.7.0.tar.gz",
Please update them to use http://mirror.tensorflow.org/ instead.

```

Tested on Git v2.15.1, which is what Travis uses.

wchargin-branch: mirror-urls-test

### DIFF
--- a/tensorboard/tools/mirror_urls_test.sh
+++ b/tensorboard/tools/mirror_urls_test.sh
@@ -41,12 +41,12 @@ check_urls_resolve() {
     unresolved_urls_file="$(mktemp)"
     # shellcheck disable=SC2016
     check_cmd='curl -sfL "$1" >/dev/null || printf "%s\n" "$1"'
-    # shellcheck disable=SC2016
-    exclude='${version}'  # as in `ci/download_bazel.sh`
+    url_pcre='(?<=")https?://mirror\.tensorflow\.org/[^"]*'
+    exclude=':!ci/download_bazel.sh'  # uses a '${version}' format string
     # We use `git-grep` to efficiently get an initial result set, then
     # filter it down with GNU `grep` separately, because `git-grep` only
     # learned `-o` in Git v2.19; Travis uses v2.15.1.
-    git grep -Ph '(?<=")https?://mirror\.tensorflow\.org/[^"]*' \
+    git grep -Ph "${url_pcre}" "${exclude}" \
         | grep -o 'https\?://mirror\.tensorflow\.org/[^"]*' \
         | grep -vF -- "${exclude}" \
         | sort \
@@ -65,9 +65,10 @@ check_urls_resolve() {
     fi
 
     printf '%s\n' 'The following URLs are not properly mirrored:'
-    cat "${unresolved_urls_file}"
+    sed -e 's/^/  - /' "${unresolved_urls_file}"
     printf '%s\n' \
-        'Googlers, see http://b/133880558 for further instructions.' \
+        'Please comment on your PR asking a TensorBoard core team member ' \
+        'to mirror these URLs per instructions in http://b/133880558.' \
         '' \
         ;
     return 1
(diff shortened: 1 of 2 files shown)


tensorboard/tools/mirror_urls_test.sh
```diff
@@ -0,0 +1,100 @@
...
+    rm "${bazel_urls_file}" || true
+}
+trap cleanup EXIT
+
+check_urls_resolve() {
+    unresolved_urls_file="$(mktemp)"
+    # shellcheck disable=SC2016
+    check_cmd='curl -sfL "$1" >/dev/null || printf "%s\n" "$1"'
+    # shellcheck disable=SC2016
+    exclude='${version}'  # as in `ci/download_bazel.sh`
```
nfelt: Can we just exclude the specific ci/download_bazel.sh file itself?  That seems less brittle.

tensorboard/tools/mirror_urls_test.sh
```diff
@@ -0,0 +1,100 @@
...
+    # list of URLs may potentially be mangled.
+
+    if ! [ -s "${unresolved_urls_file}" ]; then
+        return 0
+    fi
+
+    printf '%s\n' 'The following URLs are not properly mirrored:'
+    cat "${unresolved_urls_file}"
+    printf '%s\n' \
+        'Googlers, see http://b/133880558 for further instructions.' \
```
nfelt: More important that instructions for Googlers I think are instructions for what to do as a non-core-TB-team member who gets this error when sending a PR.  Probably something like "please comment on your PR asking a TensorBoard core team member to mirror the URLs as per instructions in http://b/133880558".

tensorboard/tools/mirror_urls_test.sh
wchargin: Sure; done.

tensorboard/tools/mirror_urls_test.sh
wchargin: Sure; done.

PR #2290, opened 2019-05-30
Remove experimental data selector

stephanwlee:
DataSelector was an experimental feature, when using Database backed
backend, allowed improved user experience in choosing data with built-in
notion of experiments.

The team has no plans to make tangible improvements to it and want to
re-consider the data selection story more holistically.

I, personally, am ok with maintaining this code iff it does not incur 
maintenance burden but, especially, since we lack tests, that is not the 
case today. If we want to revive it, we can always revert the change.

WANT_LGTM=all


tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -146,14 +145,6 @@ <h2>Settings</h2>
...
-          <div id="data-selector">
-            <tf-collapsible-data-selector
-              selection="{{_dataSelection}}"
-              active-plugins="[[_activeDashboards]]"
-            ></tf-collapsible-data-selector>
-          </div>
-        </template>
         <div id="content">
           <template is="dom-if" if="[[_activeDashboardsFailedToLoad]]">
             <div class="warning-message">
```
wchargin: GitHub doesn’t let me comment in the right place, but:

```
$ git grep -n data.selector
tensorboard/components/tf_tensorboard/tf-tensorboard.html:314:      #data-selector {
tensorboard/components/tf_tensorboard/tf-tensorboard.html:318:      tf-collapsible-data-selector {
tensorboard/components/tf_tensorboard/tf-tensorboard.html:319:        --tf-collapsible-data-selector: {
tensorboard/components/tf_tensorboard/tf-tensorboard.html:323:        --tf-collapsible-data-selector-data-selector: {
```

stephanwlee: Thanks, I totally forgot about the css parts

nfelt:
I'm ok with this if you think it's best to avoid the maintenance overhead.  I do think it's valuable to be able to have this concrete functionality on hand as a working reference/demo when revisiting this area in the near future (whether or not we revive the literal code itself), but it's true that we can check out and build older revisions even if it's removed at head.

Basically it's up to you.

stephanwlee:
>  it's true that we can check out and build older revisions even if it's removed at head.

@nfelt yup, TB 1.13.x and probably older version should contain this.

PR #2292, opened 2019-05-30
Travis CI: Do not hardcore trusty because it is EOL

cclauss:
Trusty is beyond its normal support date so do not hardcode it in .travis.yml. @nfelt Your review please?

* Screenshots of UI changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-dist: trusty
 language: python
 python:
   - "2.7"


wchargin:
While Ubuntu Trusty (14.04) itself is indeed past EOL as of 2019-04-01,
the Trusty build environment is still supported on Travis and is in fact
the default Linux environment, per [Travis docs][beo]. This pull request
doesn’t actually change any behavior; the build logs still say, “Build
dist: trusty”. So I think the only effect of this PR is that if Travis
changes the default, our builds will change underneath us without
warning, as opposed to affording us the opportunity to manually switch
to the new default when we’re ready.

What’s the motivation for this? 

[beo]: https://docs.travis-ci.com/user/reference/overview/

cclauss:
The motivation is to avoid the situation where we are testing on expired software while our users are reporting problems that tests can not replicate.  Many Travis users are setting __dist: xenial__ (in fact that is required for Python >= 3.7) but that also hardcodes a distro which could set us up with  a similar problem a few years down the road.  I am OK if this PR is rejected or if xenial is set as the default but hardcoding an expired OS seems a strange choice to me.

cclauss:
Travis CI's default distro is now Xenial:
* https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

PR #2298, opened 2019-05-31
update author email in pip package

willnorris:
opensource@ is owned by Google's Open Source Programs Office and is used primarily for reporting code of conduct violations and open source licensing issues.  It shouldn't be used as the author email for packages.


nfelt:
I assume that this line should also be updated then: https://github.com/tensorflow/tensorboard/blob/dab66d2a07ca8cf3564015b0e82d707d756e0dae/tensorboard/pip_package/setup.py#L56

willnorris:
yep, good catch, thanks :)

PR #2299, opened 2019-06-01
core: define plugin metadata on the backend


wchargin:
Summary:
Until now, the `registry.ts` module has stored dashboard metadata (tab
name, reload behavior, etc.). This commit enables defining that metadata
on the backend in a backward-compatible way, to facilitate dynamic
plugin loading. There is no user-observable change, and legacy plugins
continue to work. (We convert a proper subset of the core plugins for
demonstration purposes.)

Test Plan:
Modify `audio_plugin.py` to add an ES module definition:

```python
  def es_module_path(self):
    # DO NOT SUBMIT
    return '/magic.js'
```

Modify `tf-image-dashboard.html` to remove the `registerDashboard` call.

Then check the behavior of the following dashboards:

  - scalars, which defines `frontend_metadata` _and_ registers itself on
    the frontend, should work properly;
  - images, which defines `frontend_metadata` but does not register
    itself on the frontend, should work properly;
  - histograms, which does not define `frontend_metadata` but does
    register itself on the frontend, should work properly;
  - audio, which defines an ES module path, should show the message,
    “iframe plugin loading not yet supported”.

wchargin-branch: send-metadata


wchargin:
I submit this for your consideration, though I should probably poke
around a bit more to make sure that I’ve fixed up all the pre-refactor
uses in `tf-tensorboard.html`. (My kingdom for a type system.)

tensorboard/backend/application_test.py
```diff
@@ -173,15 +173,30 @@ def testPluginsListing(self):
...
+                'loading_mechanism': {'type': 'NONE'},
+                'remove_dom': False,
+                'tab_name': 'foo',
+                'disable_reload': False,
+                'use_data_selector': False,
             },
             'bar': {
                 'enabled': False,
-                'es_module_path': None,
+                'loading_mechanism': {'type': 'NONE'},
```
stephanwlee: Let's make `bar` use the `element_name` and have the test.

tensorboard/backend/application.py
```diff
@@ -277,19 +277,47 @@ def _serve_plugins_listing(self, request):
...
+      elif element_name is None and es_module_handler is not None:
+        loading_mechanism = {
+            'type': 'IFRAME',
+            'module_path': ''.join([
+                self._path_prefix, DATA_PREFIX, PLUGIN_PREFIX, '/',
+                plugin.plugin_name, es_module_handler,
+            ]),
+        }
+      else:
+        # As a compatibility measure, we'll pull it from the frontend
```
stephanwlee: Do we expect to get rid of this and expect legacy plugins to have `element_name` in the future? Is this for some interoperability with plugins we do not control?

tensorboard/plugins/base_plugin.py
```diff
@@ -75,6 +103,21 @@ def is_active(self):
...
+  def frontend_metadata(self):
+    """Defines how the plugin will be displayed on the frontend.
+
+    The base implementation returns a default value. Subclasses are
+    encouraged to override this method and replace any attributes on the
+    result.
+    """
+    return FrontendMetadata(
+        tab_name=None,
+        disable_reload=False,
```
stephanwlee: Can we change this to `enable_reload`? Negative name can read funny. I know that our existing flag name was disable_reload but this might be the time to change that.

tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -830,6 +898,90 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
...
+              plugin: name,
+              loadingMechanism: loadingMechanism,
+              tabName: backendMetadata.tab_name,
+              disableReload: backendMetadata.disable_reload,
+              removeDom: backendMetadata.remove_dom,
+              useDataSelector: backendMetadata.use_data_selector,
+            };
+          }
+        }
+        console.log('Changing dashboard registry:');
```
stephanwlee: Is console.log intentional?

wchargin:
> One non-critical question: I find `_replace` below to be a bit... ugly
> (of course, subjective). Is there a prettier way to do this? :\

I’m curious what you find ugly about `_replace`. As far as I’m aware,
it’s idiomatic: this usage is [explicitly called out in the `namedtuple`
docs for `namedtuple`][docs]:

> Default values can be implemented by using `_replace()` to customize a
> prototype instance:
>
> ```python
> >>> Account = namedtuple('Account', 'owner balance transaction_count')
> >>> default_account = Account('<owner name>', 0.0, 0)
> >>> johns_account = default_account._replace(owner='John')
> ```

In case it’s the underscore-name that bothers you, note that this does
not mean that `_replace` is a private symbol:

> In addition to the methods inherited from tuples, named tuples support
> three additional methods and one attribute. To prevent conflicts with
> field names, the method and attribute names start with an underscore.

[docs]: https://docs.python.org/2/library/collections.html#collections.namedtuple

There are a couple alternate ways, but none looks prettier to me:

```python
# Use dicts instead of namedtuples; validation is manual
result = super(ScalarsPlugin, self).frontend_metadata()
assert 'element_name' in result, result
result['element_name'] = 'tf-scalar-dashboard'
return result
```

```python
# Use namedtuples but modify them as dicts
result = super(ScalarsPlugin, self).frontend_metadata()._asdict()
result['element_name'] = 'tf-scalar-dashboard'
return base_plugin.FrontendMetadata(**result)
```

```python
# Use namedtuples with inline dict updates
base = super(ScalarsPlugin, self).frontend_metadata()
return base_plugin.FrontendMetadata(
    dict(list(base.items()) + [('element_name', 'tf-scalars-dashboard')])
)
```

PR #2304, opened 2019-06-04
core: stabilize and use `plugin_metadata`

wchargin:
Summary:
This patch switches first-party plugins from calling `registerDashboard`
on the frontend to implementing `plugin_metadata` on the backend.

This requires reworking the “dashboard containers stamped” logic, whose
purpose is to introduce a Polymer data dependency from the `dom-repeat`
stamping state to the dashboard rendering. Previously, all containers
were stamped simultaneously, but this may no longer be the case; we must
instead track exactly which containers have been stamped.

Plugin ordering was also previously defined implicitly on the frontend;
we shift that ordering to the backend.

Test Plan:
Check that basic functionality still works (dashboards display as
intended and in the correct order). Click through every dashboard to
verify that the proper web component is rendered. Verify that navigating
to `/#wat` displays a “no such dashboard” message, but navigating to
`/#scalars` shows no such message even before the plugins listing loads.
Verify that TensorBoard works properly when pointed at an empty logdir.

Then, check that all plugins were migrated (except for the core plugin,
which has no frontend):

```
$ git grep -l registerDashboard
tensorboard/components/tf_tensorboard/registry.ts
$ comm -3 \
> <(git grep -l 'class .*TBPlugin' ':!*test.py' | sort) \
> <(git grep -l 'def frontend_metadata' ':!*test.py' | sort) \
> ;
tensorboard/plugins/core/core_plugin.py
```

wchargin-branch: core-use-pluginmetadata

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -23,6 +23,7 @@ from __future__ import division
 from __future__ import print_function
 
 import atexit
+import collections
 import json
 import os
 import re
@@ -274,7 +275,7 @@ class TensorBoardWSGI(object):
     Returns:
       A werkzeug.Response object.
     """
-    response = {}
+    response = collections.OrderedDict()
     for plugin in self._plugins:
       start = time.time()
       is_active = plugin.is_active()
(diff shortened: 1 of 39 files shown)


wchargin:
We could create the containers ourselves, yes. I’m not sure whether
Polymer would continue to manage the `selected$="[[…]]"` binding, or
whether we’d have to do that ourselves, but I suppose that we could also
do that.

Forcing render on the `dom-repeat` sounds viable; presumably we’d do
that at the top of `_ensureDashboardContainersStamped`?

PR #2305, opened 2019-06-04
core: exclude `CorePlugin` from `plugins_listing`

wchargin:
Summary:
The `plugins_listing` response has long included a `"core"` key. Prior
to #2299 and #2304, this did nothing; now, it properly emits a warning,
because the frontend does not know how to render such a plugin. There’s
no reason that the frontend needs to know about this plugin at all.

Test Plan:
Observe that the “Plugin has no loading mechanism and no baked-in
registry entry: core” warning no longer appears in the JS console.

wchargin-branch: core-verb


tensorboard/backend/application.py
```diff
@@ -277,6 +277,9 @@ def _serve_plugins_listing(self, request):
     """
     response = collections.OrderedDict()
     for plugin in self._plugins:
+      if type(plugin) is core_plugin.CorePlugin:  # pylint: disable=unidiomatic-typecheck
```
nfelt: Optional, but we could just do `isinstance()` - CorePlugin won't ever have subclasses but I think one could argue that if it did then it makes sense to skip those too.

wchargin: The existing uses of `type(—) is —` were intentional (see #2006). For
now, I’ll keep it consistent with the other two in the file, and we can
change them all together if we so choose.

nfelt: Oh right, I forgot about that.  Ok, sure.

PR #2307, opened 2019-06-04
Change tf_storage API to async.

stephanwlee:
We are in process of making tf_storage more asynchronous. Specifically,
we had to make `get` and `set` asynchronous and, in the process, had to
make behavior changes for others.

This change is a ground work for plugin extensibility.

### DIFF
--- a/tensorboard/components/tf_storage/test/storageTests.ts
+++ b/tensorboard/components/tf_storage/test/storageTests.ts
@@ -39,48 +39,48 @@ describe('Storage', () => {
     disposeObjectBinding()
   });
 
-  it('get/setString', () => {
-    setString('key_a', 'hello', option);
-    setString('key_b', 'there', option);
-    assert.equal('hello', getString('key_a', option));
-    assert.equal('there', getString('key_b', option));
-    assert.equal(null, getString('key_c', option));
+  it('get/setString', async () => {
+    await setString('key_a', 'hello', option);
+    await setString('key_b', 'there', option);
+    assert.equal('hello', await getString('key_a', option));
+    assert.equal('there', await getString('key_b', option));
+    assert.equal(null, await getString('key_c', option));
   });
 
-  it('get/setNumber', () => {
-    setNumber('key_a', 12, option);
-    setNumber('key_b', 3.4, option);
-    assert.equal(12, getNumber('key_a', option));
-    assert.equal(3.4, getNumber('key_b', option));
-    assert.equal(null, getNumber('key_c', option));
+  it('get/setNumber', async () => {
+    await setNumber('key_a', 12, option);
+    await setNumber('key_b', 3.4, option);
+    assert.equal(12, await getNumber('key_a', option));
+    assert.equal(3.4, await getNumber('key_b', option));
+    assert.equal(null, await getNumber('key_c', option));
   });
 
-  it('get/setObject', () => {
+  it('get/setObject', async () => {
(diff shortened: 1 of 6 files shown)


tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -620,15 +629,32 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
...
          '_activeDashboards, _selectedDashboard)'),
       ],
 
+      async attached() {
+        const tabValue = await tf_storage.getString(tf_storage.TAB);
+        this._selectedDashboard = tabValue;
+        this._selectedDashboardInitialized = true;
+      },
+
+      _persistSelectedDashboard() {
```
wchargin: Why are these analytics changes in this commit?

tensorboard/components/tf_storage/test/storageTests.ts
```diff
@@ -95,53 +95,101 @@ describe('Storage', () => {
...
+        let getStringStub = null;
+
         function setValue(key: string, value: string): void {
           if (useLocalStorage) window.localStorage.setItem(key, value);
           else setHash(`${key}=${value}`);
+          getStringStub.withArgs(key, sinon.match.any).returns(value);
+        }
+
+        /**
+         * HACK: because `get` and `set` are asynchronous, it, with above stub,
```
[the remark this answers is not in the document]
stephanwlee: `getInitializer` is plumbed into Polymer's property initializer which expects either value or a function that returns a value.

tensorboard/components/tf_storage/test/storageTests.ts
```diff
@@ -95,53 +95,101 @@ describe('Storage', () => {
...
-          const initializer = getStringInitializer('foo', options);
+          const initializer = customBinding.getInitializer('foo', options);
           const fakeScope = {prop: null};
           initializer.call(fakeScope);
+          await tick();
 
           // Simulate the hashchange.
           setValue('foo', 'changed');
           window.dispatchEvent(new Event(eventName));
+          await tick();
```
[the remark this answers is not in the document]
stephanwlee: It takes a tick for `set` to take effect. 

To describe the implementation, on data changed, `setComponentValue` first invokes a `get` which is now async then `set`s.

tensorboard/components/tf_tensorboard/tf-tensorboard.html
stephanwlee: I forgot to remove `_selectedDashboardChanged`. Thanks for catching that. 

As to the behavioral change, my rationale is that analytics have to roughly match that of the URL change. Since we now persist `selectedDashboard` prop in the storage after first _real_ value is set, I thought it makes sense to move the `ga` here too. 

Technically, this is a behavior change, as you have indicated, since we would no longer have "/" pageviews if user loaded TensorBoard with a hash in the URL. I am willing to revert the change back but I am afraid of double logging of "/".

stephanwlee: Edit: reverted back to the older behavior.

wchargin:
Please let me know when this is ready for re-review (have been holding off per offline conversation, but it looks like that might be fixed now—?).

wchargin:
(un-requesting review per request, due to test failures)

PR #2312, opened 2019-06-05
Fix pip package in colab with latest closure libraries

jameswex:

Update of TB to new closure rules broke the functioning of the witwidget pip package in colab.

Javascript published to colab when using witwidget (in colab/wit.py) was doing work in the global scope, unnecessarily creating global vars that now collide with other vars, leading to JS failures when loading WIT. This javascript should never have used the global scope. This fix moves the work into an anonymous function so as to avoid polluting the global scope.

* Screenshots of UI changes

N/A

Pip install local version with this change and verify that witwidget works again in colab.
Also test locally in Jupyter to ensure it still works there.


jameswex:
@tolga-b please review

Issue #2271, opened 2019-05-22
JS error in the mesh plugin when displaying sample mesh


podlipensky:
While running TB with mesh demo app summaries, an error is shown in Chrome browser:

```
(index):29587 Uncaught TypeError: Cannot read property 'forEach' of undefined
    at b._createPointCloud ((index):29587)
    at b._createGeometry ((index):29586)
    at b.updateScene ((index):29585)
    at HTMLElement._updateScene ((index):29639)
    at HTMLElement._complexObserverEffect ((index):1595)
    at HTMLElement._effectEffects ((index):1589)
    at HTMLElement._propertySetter ((index):1589)
    at HTMLElement.__setProperty ((index):1589)
    at HTMLElement._computeEffect ((index):1595)
    at HTMLElement._effectEffects ((index):1589)
```

PR #2315, opened 2019-06-05
Handle cancelled requests properly.

podlipensky:
This fixes issue https://github.com/tensorflow/tensorboard/issues/2271 by ignoring cancelled requests.


tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
```diff
@@ -153,6 +153,11 @@ Polymer({
         this.appendChild(this._meshViewer.getRenderer().domElement);
         this._meshViewerAttached = true;
       }
+    }).catch(function (error) {
```
wchargin: Per Google JS style, prefer arrow functions over anonymous `function`
expressions, to make it clear that there’s no dynamic `this`-binding
going on:

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
```diff
@@ -153,6 +153,11 @@ Polymer({
...
+      if (!error || !error.code || error.code != vz_mesh.ErrorCodes.CANCELLED) {
+        error = error || 'Response processing failed.';
```
wchargin: Please only throw `Error`s, not string literals. Without an `Error`, you
lose programmatic access to the stack trace, and surrounding code that
attempts to reference `e.name` and `e.message` will fail. See:
<https://google.github.io/styleguide/jsguide.html#features-exceptions>

third_party/js.bzl
```diff
@@ -515,9 +515,9 @@ def tensorboard_js_workspace():
       name = "org_tensorflow_graphics_lib",
       licenses = ["notice"],  # MIT
       sha256_urls = {
-          "82bf620fbf10af00a83754b7ebadcbcd41af7181e1aa237cdd72a2881d8004fe": [
-              "http://mirror.tensorflow.org/raw.githubusercontent.com/tensorflow/graphics/be403ab520d129fa2ad99a2dc9ae0102c57a499f/tensorflow_graphics/tensorboard/mesh_visualizer/tf_mesh_dashboard/array-buffer-data-provider.js",
-              "https://raw.githubusercontent.com/tensorflow/graphics/659021ac8d56c0d13eb0ebfa99786a12d48fbc61/tensorflow_graphics/tensorboard/mesh_visualizer/tf_mesh_dashboard/array-buffer-data-provider.js",
+          "5a9cd221f6727b7524ba6b63bdc0355843c42a238d3f83df980005c7cb270a92": [
+              "http://mirror.tensorflow.org/raw.githubusercontent.com/tensorflow/graphics/cd8c42f9ca260f77c6acfecd42e66ef01d1a3766/tensorflow_graphics/tensorboard/mesh_visualizer/tf_mesh_dashboard/array-buffer-data-provider.js",
```
wchargin: I’ve added this URL to the TensorFlow mirror, so that test should pass
when you push a new commit.

third_party/js.bzl
```diff
@@ -515,9 +515,9 @@ def tensorboard_js_workspace():
...
+              "https://raw.githubusercontent.com/tensorflow/graphics/cd8c42f9ca260f77c6acfecd42e66ef01d1a3766/tensorflow_graphics/tensorboard/mesh_visualizer/tf_mesh_dashboard/array-buffer-data-provider.js",
           ],
           "bc7e983aae707e2d8dc0ca6406e3779e4ad73c537a02bed6d3d3d40180f26904": [
               "http://mirror.tensorflow.org/raw.githubusercontent.com/tensorflow/graphics/659021ac8d56c0d13eb0ebfa99786a12d48fbc61/tensorflow_graphics/tensorboard/mesh_visualizer/tf_mesh_dashboard/mesh-viewer.js",
```
wchargin: Can you please update this entry, too, such that all the files that we
pull from `tensorflow/graphics` are synced to the same commit?

podlipensky:
> When running the demo on this commit, I still see lots of “Cannot use
> 'in' operator to search for 'config' in null” errors. Is this commit not
> intended to fix that error?

I don't see it. Is it js error?

third_party/js.bzl
podlipensky: Not sure i follow. You want mesh-viewer.js to be part of the same commit? That file wasn't changed...

wchargin: You’re fetching `array-buffer-data-provider.js` from Version A of the
codebase, and `mesh-viewer.js` from Version B. It happens to be the case
that the file contents are the same at the two versions, but that’s not
clear from the BUILD file. You can fix this by updating the URLs to
fetch from the same commit: change

```
              "http://mirror.tensorflow.org/raw.githubusercontent.com/tensorflow/graphics/659021ac8d56c0d13eb0ebfa99786a12d48fbc61/tensorflow_graphics/tensorboard/mesh_visualizer/tf_mesh_dashboard/mesh-viewer.js",
```

to

```
              "http://mirror.tensorflow.org/raw.githubusercontent.com/tensorflow/graphics/cd8c42f9ca260f77c6acfecd42e66ef01d1a3766/tensorflow_graphics/tensorboard/mesh_visualizer/tf_mesh_dashboard/mesh-viewer.js",
```

and likewise for the GitHub URL.

podlipensky:
@wchargin mirror's link must be updated somewhere else manually right?

podlipensky:
> @wchargin mirror's link must be updated somewhere else manually right?

(the reason why build fails now)

wchargin:
Thanks, yes. I’ve mirrored the URL and restarted the build.

(For future reference, http://b/133880558 has instructions for how to
update the mirror.)

PR #2319, opened 2019-06-06
build: use `tf.test.main` for `mesh:demo_utils_test`

wchargin:
Summary:
This test case already contains actual TensorFlow code, via the `GFile`
use in the code under test. Within Google, such tests must be run via
`tf.test.main()` or `absltest.main()`, and the TensorFlow-specific code
should be in a `tf.test.TestCase` as well (though this is not strictly
required).

Test Plan:
Verified that this patch fixes a test that’s broken when syncing into
Google3.

wchargin-branch: tf-test-mesh-test


wchargin:
WANT_LGTM=any

PR #2321, opened 2019-06-06
test: use `absltest` for `tensorboard.test`

wchargin:
Summary:
Within Google, tests that call certain functions must be run via either
`tf.test` or `absltest`. When possible, we’d like to use `absltest` so
that the tests can be run in notf settings. Thus, we change the backing
implementation of `tensorboard.test` from `unittest` to `absltest`.

We retain the `tensorboard.test` layer to add the `get_temp_dir` method,
which is provided by `tf.test` but not `absltest`, and which our tests
often expect to use.

This has the side benefit that we get lots of assertions for free,
including `assertStartsWith`, `assertEndsWith`, and `assertLen`, as well
as fancy ones like `assertSequenceAlmostEqual` and `assertUrlEqual`.
(Unfortunately, we still lack `tf.test`’s `assertProtoEquals`.)

Test Plan:
Existing tests pass. Googlers, see <http://cl/251757823>.

wchargin-branch: tb-test-absltest

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -159,6 +159,7 @@ py_library(
     srcs = ["test.py"],
     srcs_version = "PY2AND3",
     deps = [
+        "//tensorboard:expect_absl_testing_absltest_installed",
         "//tensorboard/util:tb_logging",
         "@org_pythonhosted_six",
     ],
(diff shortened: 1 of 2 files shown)


tensorboard/BUILD
```diff
@@ -288,6 +289,14 @@ py_library(
     visibility = ["//visibility:public"],
 )
 
+py_library(
+    name = "expect_absl_testing_absltest_installed",
+    # This is a dummy rule used as a absl-py dependency in open-source.
```
nfelt: As an aside, I wonder if perhaps we should just take an actual WORKSPACE dep on absl at this point.  We use a fair amount of it and unlike projects like TF and numpy that are either heavyweight or not natively built with bazel or both, absl-py is pretty light and already built with bazel.

wchargin: That seems reasonable. I’ll look into it.

PR #2326, opened 2019-06-06
infra: upgrade three.js to r104 from r77

stephanwlee:
### Notable changes:
- Vertex shader require related glsl code block
- other API changes in between
- WebGLRenderer.render no longer takes 4 inputs. Instead, you must set
  renderTarget manually.
- BufferAttribute.prototype.count is important and attribute.array is
  discourage in favor of setArray

### Notable no changes:
- Types are not updated. 
  - DefinitelyType no longer provide the latest
    typing of three in favor TypeScript in three.js repo. For the time
    being, we have decided not to pursue this as downloading the archive
    would fetch unnecessarily large file (examples dir is 300+MB).
  - Types in newer DefinitelyTyped and three.js seem to be incompatible
     with current build system. https://github.com/stephanwlee/tensorboard/commit/daa35e457ed64cbe1cfe871beceb901377e5ec80

### Detailed steps to verify changes work correctly (as executed by you)
Poked around projector, what-if, and mesh plugins thoroughly.

### DIFF
--- a/tensorboard/plugins/projector/vz_projector/scatterPlot.ts
+++ b/tensorboard/plugins/projector/vz_projector/scatterPlot.ts
@@ -446,7 +446,7 @@ export class ScatterPlot {
   }
 
   private add3dAxis() {
-    const axes = new THREE.AxisHelper();
+    const axes = new (THREE as any).AxesHelper();
     axes.name = 'axes';
     this.scene.add(axes);
   }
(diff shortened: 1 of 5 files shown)


stephanwlee:
@wchargin updated the PR message. Another issue that I faced was captured in this branch: https://github.com/stephanwlee/tensorboard/commit/daa35e457ed64cbe1cfe871beceb901377e5ec80. Both projector code and facets code need to be updated to work with the new type changes which I think should be done in a separate PR.

(I didn't know how to share my code in more permanent way but I will keep my branch around for awhile)

stephanwlee:
Didn't know about the TypeScript 3.5 changes. Thanks for that. I will consider upgrading that in the future.

PR #2328, opened 2019-06-07
add compatibility check badges to README

ylil93:
Hello google package maintainer,
The badges being added to the README in this PR will indicate your compatibility with other google packages. This addresses a set of user bugs which have happened when a user depends on two Cloud libraries (or runtimes that bundle libraries), say A and B, which both depend on library C. If the two libraries require different versions of C, the users can run into issues both when they pip install the libraries, and when they deploy their code. Our compatibility server checks that all libraries we make including this one are self and pairwise compatible as well as not having any deprecated dependencies. The two badges will mark the build for your project green when the latest version available on PyPI and github HEAD respectively meet all compatibility checks with itself and all other libraries. The badge target will link to a details page that elaborates on the current status. This should help you fix issues pre-release, to avoid user surprises. For more information, please take a look at our project charter at go/python-cloud-dependencies-project-charter and the badging PRD https://docs.google.com/document/d/1GYRFrfUou2ssY71AtnLkc8Sg1SD4dxqN4GzlatGHHyI/edit?ts=5c6f031d

* Screenshots of UI changes
Screen Shot 2019-06-06 at 5 47 48 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# TensorBoard [![Travis build status](https://www.travis-ci.com/tensorflow/tensorboard.svg?branch=master)](https://travis-ci.com/tensorflow/tensorboard/)
+# TensorBoard [![Travis build status](https://www.travis-ci.com/tensorflow/tensorboard.svg?branch=master)](https://travis-ci.com/tensorflow/tensorboard/) [![Compat check PyPI](https://python-compatibility-tools.appspot.com/one_badge_image?package=tensorboard)](https://python-compatibility-tools.appspot.com/one_badge_target?package=tensorboard)
 
 TensorBoard is a suite of web applications for inspecting and understanding your
 TensorFlow runs and graphs.


ylil93:
> LGTM on adding the badge itself, but what is the expectation around actually keeping this green? I rarely look at the home page of the repo for badge statuses (we generally find out in other ways for the build status). The badge approach seems more suited for informing potential users than for the maintainers of the repo.

@nfelt The expectation was for the badge to be a tool to help package maintainers stay on top of compatibility issues. There currently is no plan to implement other notifications. Although it's unfortunate that this change doesn't completely mesh with this project's workflow, at least it's a step in the right direction (in that it's better than nothing).

nfelt:
> > LGTM on adding the badge itself, but what is the expectation around actually keeping this green? I rarely look at the home page of the repo for badge statuses (we generally find out in other ways for the build status). The badge approach seems more suited for informing potential users than for the maintainers of the repo.
> 
> @nfelt The expectation was for the badge to be a tool to help package maintainers stay on top of compatibility issues. There currently is no plan to implement other notifications. Although it's unfortunate that this change doesn't completely mesh with this project's workflow, at least it's a step in the right direction (in that it's better than nothing).

Ok, fair enough.

PR #2334, opened 2019-06-07
infra: use base.setAttribute instead .href

stephanwlee:
Closure security conformance test guard against almost all instances of
setting .href. Since we will not be using the goog.safe APIs for now, we
will be using setAttribute instead.

Confirm the change with an example plugin.

### DIFF
--- a/tensorboard/components/tf_tensorboard/tf-tensorboard.html
+++ b/tensorboard/components/tf_tensorboard/tf-tensorboard.html
@@ -809,6 +809,9 @@ limitations under the License.
               const pluginBasePath = tf_backend.getRouter()
                   .pluginRoute(selectedDashboard, '/')
               const base = subdocument.createElement('base');
+              // TODO(stephanwlee): Use sanitized URL when setting the href.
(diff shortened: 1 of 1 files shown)


tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -806,10 +806,12 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
               this.scopeSubtree(iframe);
               container.appendChild(iframe);
               const subdocument = iframe.contentDocument;
-              const baseUrl = tf_backend.getRouter()
+              const pluginBasePath = tf_backend.getRouter()
                   .pluginRoute(selectedDashboard, '/')
               const base = subdocument.createElement('base');
-              base.href = baseUrl;
+              base.setAttribute(
```
nfelt: Can you add a comment explaining why the indirection is necessary?

stephanwlee: Done. Added few lines.

PR #2340, opened 2019-06-11
frontend: reinstate registration for conditional plugins

### LINKED ISSUE #2338: Plugins without backend have no frontend since 2019-06-06
Consider Stack Overflow for getting support using TensorBoard—they have
a larger community with better searchability:

https://stackoverflow.com/questions/tagged/tensorboard

Do not use this template for for setup, installation, or configuration
issues. Instead, use the “installation problem” issue template:

https://github.com/tensorflow/tensorboard/issues/new?template=installation_problem.md

To report a problem with TensorBoard itself, please fill out the
remainder of this template.

## Environment information (required)

Please run `diagnose_tensorboard.py` (link below) in the same
environment from which you normally run TensorFlow/TensorBoard, and
paste the output here:

https://raw.githubusercontent.com/tensorflow/tensorboard/master/tensorboard/tools/diagnose_tensorboard.py

For browser-related issues, please additionally specify:

  - Browser type and version (e.g., Chrome 64.0.3282.140):
  - Screenshot, if it’s a visual issue:

## Issue description

Please describe the bug as clearly as possible. How can we reproduce the
problem without additional resources (including external data files and
proprietary Python modules)?

Profile plugin is missing in tb-nightly 06-10.
There is no profile tool in the plugins dropdown on the right.
profile_plugin_1

localhost:6006/#profile will show "There's no dashboard by the name of "profile." error.
profile_plugin_2

wchargin:
Summary:
Plugins whose backends are provided via “loaders” rather than the
plugins themselves may fail to load. We still want to allow those
plugins to display a message in the frontend indicating what must be
done to enable them (e.g., update TensorFlow, start a `tfdbg` debugger),
which requires registering their frontends unconditionally.

This is a short-term hack to retain functionality prior to #2304 (with
the minor change that these plugins will now always appear at the end of
the list). We’ll want to replace this with a better mechanism when time
permits.

Fixes #2338.

Test Plan:
Launch TensorBoard without TensorFlow installed and with no flags other
than `--logdir`. Note that the beholder, debugger, hparams, what-if
tool, and profile dashboards all have valid entries in the “inactive”
dropdown, displaying appropriate help messages.

Verify that of the plugins listed in `default.py`, all those provided
via loaders now have `registerDashboard` statements.

wchargin-branch: register-conditional-plugins

### DIFF
--- a/tensorboard/plugins/beholder/tf_beholder_dashboard/tf-beholder-dashboard.html
+++ b/tensorboard/plugins/beholder/tf_beholder_dashboard/tf-beholder-dashboard.html
@@ -513,6 +513,14 @@ with MonitoredSession(..., hooks=[beholder_hook]) as sess:
       },
     });
 
+  // TODO(#2338): Remove this, and set up the "no TF" message properly.
+  // Keep this in sync with `frontend_metadata` in `beholder_plugin.py`.
+    tf_tensorboard.registerDashboard({
+      plugin: PLUGIN_NAME,
+      elementName: 'tf-beholder-dashboard',
+      shouldRemoveDom: true,
+    });
+
     })();
   </script>
 </dom-module>
--- a/tensorboard/plugins/beholder/tf_beholder_dashboard/tf-beholder-dashboard.html
+++ b/tensorboard/plugins/beholder/tf_beholder_dashboard/tf-beholder-dashboard.html
@@ -513,8 +513,8 @@ with MonitoredSession(..., hooks=[beholder_hook]) as sess:
       },
     });
 
-  // TODO(#2338): Remove this, and set up the "no TF" message properly.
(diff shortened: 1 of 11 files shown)


tensorboard/plugins/beholder/tf_beholder_dashboard/tf-beholder-dashboard.html
```diff
@@ -513,6 +513,14 @@ <h3>No Beholder data was found.</h3>
       },
     });
 
+  // TODO(#2338): Remove this, and set up the "no TF" message properly.
```
nfelt: nit: indentation not aligned

wchargin: Thanks; fixed.

PR #2341, opened 2019-06-11
Polymer 2: fix rest of the plugins

stephanwlee:
Now that we've made most of the infrastructural changes, we have to go through and address issues in all plugins. Most changes are minor (in size) so, for sake of faster code reviews, this change incorporates several plugin changes at once. The PR is meant to be merged without squash.

The change makes changes to below plugins:
- hparams
- debugger
- projector
- what-if
- image
- custom-scalars
- mesh

WANT_LGTM=any

### DIFF
--- a/tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-dashboard.html
+++ b/tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-dashboard.html
@@ -321,10 +321,10 @@ writer.add_summary(layout_summary)
         this._requestManager.request(url).then(handlePluginsListingResponse);
       },
       _reloadCharts() {
-        const charts =
-          this.querySelectorAll('tf-custom-scalar-margin-chart-card, ' +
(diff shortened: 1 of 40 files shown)


tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-dashboard.html
```diff
@@ -380,6 +382,9 @@ <h3>The custom scalars dashboard is inactive.</h3>
         const pane = event.target;
         if (pane.opened) {
           this._openedCategories[pane.category.name] = true;
+          Array.from(this._getCharts())
```
nfelt: Can you add a comment indicating why it's necessary to reload the charts here?  Basically just a summary of the commit description.

stephanwlee: We now use tf-category-paginated-view so no longer require this change. Removed it.

PR #2342, opened 2019-06-11
Make sure compat PyRecordReader_New reads new data

orionr:
This fixes https://github.com/pytorch/pytorch/issues/20477

It looks like our original compat implementation of `PyRecordReader_New` didn't continue to try and read content from a file after it had done the original read. This updates the class to continue to request data from the file and read in new events. Note that this would only be seen in the non-TensorFlow case. Here's a test case:

```
conda create --name notf36 python=3.6
conda activate notf36
pip install --upgrade pip
pip install scipy
conda install -y protobuf
conda install -y absl-py
conda install -y numpy
conda install pytorch-nightly -c pytorch  # For the train.py below
pip install tb-nightly  # For the train.py below
bazel build tensorboard --verbose_failures
./bazel-bin/tensorboard/tensorboard --logdir ../realtime_runs/runs/ --verbosity 2
bazel test //tensorboard/summary/writer:record_writer_test --test_output=errors
```

While the above is running, execute the following `train.py` script located in `../realtime_runs/`:

```
import numpy as np
import time

import torch
from torch.utils.tensorboard import SummaryWriter

writer = SummaryWriter()

for i in range(1000000):
    writer.add_scalar('data/s1', i % 100, i)
    # writer.add_scalar('data/s2', i % 200, i)
    # writer.add_scalars('data/scalar_group', {'xsinx': i * np.sin(i), 'xcosx': i * np.cos(i)})
    writer.flush()

    time.sleep(1)
```

with `python train.py`

With the old code TensorBoard won't update. With the updated code it will as expected.

cc @nfelt, @natalialunova, @wchargin, @lanpa

### DIFF
--- a/tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py
+++ b/tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py
@@ -164,13 +164,8 @@ def crc32c(data):
 
 
 class PyRecordReader_New:
-    def __init__(
-      self,
-      filename=None,
-      start_offset=0,
-      compression_type=None,
-      status=None
-    ):
+    def __init__(self, filename=None, start_offset=0, compression_type=None,
+                 status=None):
         self.filename = filename
         self.file_handle = None
         self.start_offset = start_offset
--- a/tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py
+++ b/tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py
@@ -166,29 +166,28 @@ def crc32c(data):
 class PyRecordReader_New:
     def __init__(self, filename=None, start_offset=0, compression_type=None,
                  status=None):
+        if filename is None:
+            raise errors.NotFoundError(
+                None, None, 'No filename provided, cannot read Events')
+        if not gfile.exists(filename):
+            raise errors.NotFoundError(
+                None, None,
+                '{} does not point to valid Events file'.format(filename))
+        if start_offset:
+            raise errors.UnimplementedError(
+                None, None, 'start offset not supported by compat reader')
+        if compression_type:
+            # TODO: Handle gzip and zlib compressed files
+            raise errors.UnimplementedError(
+                None, None, 'compression not supported by compat reader')
         self.filename = filename
-        self.file_handle = None
         self.start_offset = start_offset
         self.compression_type = compression_type
         self.status = status
         self.curr_event = None
+        self.file_handle = gfile.GFile(self.filename, 'rb')
 
     def GetNext(self):
-        if self.file_handle is None:
-            if self.filename is None:
-                raise errors.NotFoundError(
-                    None, None, 'No filename provided, cannot read Events')
-            if not gfile.exists(self.filename):
-                raise errors.NotFoundError(
-                    None, None,
-                    '{} does not point to valid Events file'.format(self.filename))
-            if self.compression_type:
-                # TODO: Handle gzip and zlib compressed files
-                raise errors.UnimplementedError(
-                    None, None, 'compression not supported by compat reader')
-            self.file_handle = gfile.GFile(self.filename, 'rb')
-            self.file_handle.seek(self.start_offset)
-
         # Read the header
         self.curr_event = None
         header_str = self.file_handle.read(8)
--- a/tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py
+++ b/tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py
@@ -171,75 +171,69 @@ class PyRecordReader_New:
       compression_type=None,
       status=None
     ):
-        self.event_strs = []
         self.filename = filename
+        self.file_handle = None
         self.start_offset = start_offset
         self.compression_type = compression_type
         self.status = status
-        self.done_reading = False
-        self.index = -1
-
-    def read(self):
-        if self.filename is None:
-            raise errors.NotFoundError(
-                None, None, 'No filename provided, cannot read Events')
-        if not gfile.exists(self.filename):
-            raise errors.NotFoundError(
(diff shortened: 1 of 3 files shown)


orionr:
@nfelt, hopefully we can land https://github.com/tensorflow/tensorboard/pull/2260 first and then this one.

natalialunova:
Looks good!

tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py
```diff
@@ -164,82 +164,71 @@ def crc32c(data):
...
+            if not gfile.exists(self.filename):
+                raise errors.NotFoundError(
+                    None, None,
+                    '{} does not point to valid Events file'.format(self.filename))
+            if self.compression_type:
+                # TODO: Handle gzip and zlib compressed files
+                raise errors.UnimplementedError(
+                    None, None, 'compression not supported by compat reader')
+            self.file_handle = gfile.GFile(self.filename, 'rb')
+            self.file_handle.seek(self.start_offset)
```
nfelt: TensorBoard's usage of PyRecordReader_New never passes in the `start_offset` argument, so I'd rather leave out that implementation and avoid expanding the GFile API surface we support.  We can just raise an UnimplementedError instead if `start_offset` is provided.

tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py
```diff
@@ -164,82 +164,71 @@ def crc32c(data):
...
-        if self.index >= len(self.event_strs) - 1:
+        if self.file_handle is None:
+            if self.filename is None:
+                raise errors.NotFoundError(
+                    None, None, 'No filename provided, cannot read Events')
+            if not gfile.exists(self.filename):
+                raise errors.NotFoundError(
+                    None, None,
+                    '{} does not point to valid Events file'.format(self.filename))
+            if self.compression_type:
```
nfelt: Let's handle this in the constructor instead.

tensorboard/summary/writer/record_writer_test.py
```diff
@@ -62,6 +63,21 @@ def test_record_writer_roundtrip(self):
       r.GetNext()
       self.assertEqual(r.record(), bytes_to_write)
 
+  def test_record_immediate_read(self):
+    filename = os.path.join(self.get_temp_dir(), "record_immediate_read")
+    bytes_to_write = b"hello world"
```
nfelt: I think this would be a better test if we use different bytestrings each time (for example, a stringified counter value).  Otherwise, as is, this test would pass if we were actually returning the first record over and over.

orionr:
Good point - changing test.

On Jun 17, 2019, at 3:40 PM, Nick Felt <notifications@github.com<mailto:notifications@github.com>> wrote:


@nfelt commented on this pull request.

Can you add a minimal test for the record reader that at least covers this case and ideally a few other basic ones? I think the fact that this bug persisted is evidence that the implementation is different enough from the original that this really should have tests.

________________________________

In tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py<https://github.com/tensorflow/tensorboard/pull/2342#discussion_r294545995>:

> -
-        if self.index >= len(self.event_strs) - 1:
+        if self.file_handle is None:
+            if self.filename is None:
+                raise errors.NotFoundError(
+                    None, None, 'No filename provided, cannot read Events')
+            if not gfile.exists(self.filename):
+                raise errors.NotFoundError(
+                    None, None,
+                    '{} does not point to valid Events file'.format(self.filename))
+            if self.compression_type:
+                # TODO: Handle gzip and zlib compressed files
+                raise errors.UnimplementedError(
+                    None, None, 'compression not supported by compat reader')
+            self.file_handle = gfile.GFile(self.filename, 'rb')
+            self.file_handle.seek(self.start_offset)


TensorBoard's usage of PyRecordReader_New never passes in the start_offset argument, so I'd rather leave out that implementation and avoid expanding the GFile API surface we support. We can just raise an UnimplementedError instead if start_offset is provided.

________________________________

In tensorboard/compat/tensorflow_stub/pywrap_tensorflow.py<https://github.com/tensorflow/tensorboard/pull/2342#discussion_r294546044>:

>
     def GetNext(self):
-        if not self.done_reading:
-            self.read()
-
-        if self.index >= len(self.event_strs) - 1:
+        if self.file_handle is None:
+            if self.filename is None:
+                raise errors.NotFoundError(
+                    None, None, 'No filename provided, cannot read Events')
+            if not gfile.exists(self.filename):
+                raise errors.NotFoundError(
+                    None, None,
+                    '{} does not point to valid Events file'.format(self.filename))
+            if self.compression_type:


Let's handle this in the constructor instead.

orionr:
@nfelt if tests pass this should be ready for another review. Much appreciated.

tensorboard/summary/writer/record_writer_test.py
```diff
@@ -52,30 +52,28 @@ def test_empty_record(self):
   def test_record_writer_roundtrip(self):
     filename = os.path.join(self.get_temp_dir(), "record_writer_roundtrip")
     w = RecordWriter(open(filename, 'wb'))
-    bytes_to_write = b"hello world"
-    times_to_test = 50
-    for _ in range(times_to_test):
-      w.write(bytes_to_write)
+    chunks_to_write = [ "hello world{}".format(i).encode() for i in range(10) ]
```
nfelt: nit: no whitespace inside brackets per google style, also on line 67

orionr: Fixed.

cswwp:
Hi, @nfelt, i have install tb_nightly (version 2.1.0a20191021), and restart tensorboard --log_dir target dir,  and it still shows that "TensorFlow installation not found - running with reduced feature set." I'm confused

Issue #1063, opened 2018-03-19
Support multiple file writers writing at once.


chihuahua:
## Background

TensorBoard assigns one `DirectoryWatcher` per run. The `DirectoryWatcher` watches for new events being written to the directory and loads it into TensorBoard's multiplexer. Today, [the `DirectoryWatcher` iterates through events files (for a run) in lexicographic order](https://github.com/tensorflow/tensorboard/blob/bf1e23229b8f9efbc5348ffbd7d9418c5fcb8b3c/tensorboard/backend/event_processing/directory_watcher.py#L29). After it has finished reading from an events file, it never reads from it again (and moves on to a subsequent events file).

This behavior means that TensorBoard does not support multiple file writers writing to the same run directory: TensorBoard would move on to a different events file B once it finishes reading events file A and would never go back to reading A despite how A is updated.

In turn, that problem blocks TensorFlow+TensorBoard's transition towards migrating `Estimator`, tflearn hooks, and other high-level TensorFlow constructs to using the `tf.contrib.summary`-style summaries. `tf.contrib.summary`-style summaries are written via a type of `SummaryWriter` different from the one that writes traditional-style (`tf.summary.*`) summaries, so users that intermingle the summaries will necessarily have to concurrently use 2 summary writers. Many users thoughout google3 intermingle the summaries, and the transition must happen soon.

nfelt:
Just jotting down what we discussed while it's fresh in my head.  We keep the essential structure of DirectoryWatcher - Load() spools events from the current _loader until it's exhausted, and then moves on to a new file.

However, we change the algorithm for moving on to a new file to do the following:

0. (setup) As state, we keep a "stream map" mapping from file suffixes to "stream entry" values that hold a loader instance, a timestamp, and a list of filenames
1. List the directory and get all filenames
2. Filter out any filenames that are "dead" (using a set of known "dead" files)
3. Find the next filename, per the following substeps:
3a. Order the filenames by timestamp, then by suffix.
3b. Seek ahead in the list to the first element with timestamp and suffix >= the current timestamp and suffix.
3c. Check if element.timestamp is >= to the timestamp value stored for element.suffix in the stream map.  If not, log an "out of order files" error and keep going until this is true.
3d. If this process failed to find a filename, null out the current timestamp and suffix and return.  (We'll resume with the next Load() call.)
[...]

PR #2346, opened 2019-06-13
Allow reading multiple files simultaneously from Tensorboard

bcoopers:
Enables setting multiple evaluation workers in TF estimator. Each writes to its own file. Now all are read by Tensorboard, allowing greater statistical significance and/or throughput rate of points written.

Reads from all files, publishing the event with the most recent step number. For identical step numbers publishes the earlier time.

* Screenshots of UI changes
Please reach out to me for this. I cannot share this externally I think.

Ran training, observed results on Tensorboard. Ran unit test- it passes.

None

### DIFF
--- a/tensorboard/backend/event_processing/directory_watcher.py
+++ b/tensorboard/backend/event_processing/directory_watcher.py
@@ -18,9 +18,9 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from google3.third_party.tensorboard.backend.event_processing import io_wrapper
-from google3.third_party.tensorboard.compat import tf
-from google3.third_party.tensorboard.util import tb_logging
+from tensorboard.backend.event_processing import io_wrapper
+from tensorboard.compat import tf
+from tensorboard.util import tb_logging
 
 
 logger = tb_logging.get_logger()
--- a/tensorboard/backend/event_processing/directory_watcher.py
+++ b/tensorboard/backend/event_processing/directory_watcher.py
@@ -18,14 +18,52 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import bisect
-
-from tensorboard.backend.event_processing import io_wrapper
-from tensorboard.compat import tf
-from tensorboard.util import tb_logging
+from google3.third_party.tensorboard.backend.event_processing import io_wrapper
+from google3.third_party.tensorboard.compat import tf
+from google3.third_party.tensorboard.util import tb_logging
 
 
 logger = tb_logging.get_logger()
+
+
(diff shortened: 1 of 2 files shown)


wchargin:
Hi @bcoopers—thank you for the PR! We’re definitely interested in this
kind of functionality; see #1063 for the canonical issue, and #1867 for
a shelved (but not abandoned) prior attempt.

I believe that in the past we’ve run into performance issues when
continuously reading from all files under the logdir, because the number
of such files is unbounded and access over a network file system can be
expensive. The PR description of #1867 has some discussion about this.

The best person to review this is @nfelt; he’s OOO right now, but will
be back on Monday. I’ll see if I can take a preliminary look tomorrow.

wchargin:
For Googlers, cf. <http://cl/253027277>.

bcoopers:
If you want I can add a max # of files parameter (say a thousand) and start expiring the least-recently-updated files once we have that many.

bcoopers:
Friendly ping

nfelt:
Hi and thanks for your interest in contributing, and apologies for so long to get to this PR.

I have a few reservations about this approach, relative to the approach in #1867.

1) We do need a way to limit the total number of files we're polling to avoid resource blowup.  My preference is to age out files based on last write times (like #1867) rather than simply a fixed file count, since I think this better captures what we're actually trying to achieve and avoids polling very old files unnecessarily, even if we have "capacity" to poll them.

2) Even with a mitigation for that case, turning this behavior on by default could cause significant regressions for large logdirs with many small event files.  This is why in #1867 it's a separate implementation that's off by default behind a flag.

3) Reconciling parallel streams of data across files is more complicated than just merging in sorted order by step.  For example sometimes people write two successive separate "runs" to the same directory by accident; I believe this code would merge them into a single logical run sorted by step, which could look very misleading.  In that case it's arguably better to keep them in the order they would appear in today's TensorBoard (data from first file, then second file) because at least this shows up in TensorBoard as a "zigzag" plot doubling back on itself that indicates something went wrong.  In general, I think the "reconciliation across data streams" problem is better solved at a higher level in the code where we have a little more context.

I should be able to finish off #1867 in mid-July.  I realize it's been a long time coming, and I was hoping I could get to it this week, but unfortunately I didn't and I'll be OOO next week.  If you'll bear with us until then, I think that will be the best path to fixing the underlying issue #1063.

nfelt:
I've submitted #1867, please take a look and feel free to provide feedback.  I believe it should satisfy the use case for this PR, so I will go ahead and close it, but let me know if you'd like it re-opened.

Note that our sync into Google is currently blocked on some unrelated issues, but once those are resolved this should land internally soon.

PR #2354, opened 2019-06-17
plugins: add sample dynamically loaded plugin

wchargin:
Summary:
This plugin can eventually replace the `tensorboard-plugin-example`
repository; it’ll be easier for us to keep it up to date if it’s in the
same repository as main TensorBoard. And it has tests!

Test Plan:
Run `bazel test //tensorboard/plugins/example:smoke_test` to start.
Then, modify `setup.py` to remove the `entry_points` entry, and note
that the test fails. Separately, modify line 42 of `__init__.py` to
serve the JS under `/not_index.js` instead; the test should again fail.

Test the development workflow: Follow the instructions in `README.md`.
Then, edit `widgets.js` (say, add another exclamation point) and refresh
to see changes live without restarting TensorBoard:

![screenshot of “Hello TensorBoard!” in example plugin][1]

[1]: https://user-images.githubusercontent.com/4317806/59641012-cb700380-9114-11e9-8323-c811388df31c.png

wchargin-branch: example-dynamic-plugin


tensorboard/plugins/example/setup.py
```diff
@@ -0,0 +1,33 @@
...
+from __future__ import print_function
+
+import setuptools
+
+
+setuptools.setup(
+    name="tensorboard_plugin_example",
+    version="0.1.0",
+    description="Sample TensorBoard plugin.",
+    packages=["tensorboard_plugin_example"],
```
stephanwlee: Does this guarantee that the example.js will be part of the pip package? I don't think this would be a problem for symlink case but it may not work as intended when creating a pip package out of it.

tensorboard/plugins/example/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Generated by running `python setup.py develop`.
```
stephanwlee:
```
Requires license information in below file(s):
tensorboard/plugins/example/.gitignore
```

😅

tensorboard/plugins/example/setup.py
wchargin: No, it doesn’t. Good point; we should make sure that it does.

tensorboard/plugins/example/.gitignore
wchargin: You know, I _almost_ ran that test locally! :-) Thanks.

tensorboard/plugins/example/tensorboard_plugin_example/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+from tensorboard.plugins import base_plugin
+import werkzeug
+from werkzeug import wrappers
+
+
+class ExamplePlugin(base_plugin.TBPlugin):
+  plugin_name = 'example'
+
+  def __init__(self, context):
```
nfelt: Can't this just be omitted?  Or do you have it there for demonstration purposes?  If the latter, maybe replace `pass` with a comment to that effect.

tensorboard/plugins/example/tensorboard_plugin_example/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+    pass
+
+  def is_active(self):
+    return True
+
+  def get_plugin_apps(self):
+    return {
+        "/plugin.js": self._serve_js,
```
nfelt: Stephan's #2356 calls this index.js.  Let's pick one convention to be the standard?  cc @stephanwlee

tensorboard/plugins/example/tensorboard_plugin_example/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+# limitations under the License.
+# ==============================================================================
+"""A sample plugin to demonstrate dynamic loading."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+import textwrap
```
nfelt: Unused

tensorboard/plugins/example/tensorboard_plugin_example/__init__.py
wchargin: Sure; `index.js` works.

tensorboard/plugins/example/tensorboard_plugin_example/__init__.py
wchargin: No, it can’t. If you omit this, you get an inscrutable error in the
depths of TensorBoard. I had a patch to make it possible to omit this,
but couldn’t test it due to #2355. Patch:
<https://gist.github.com/wchargin/c6f69767393dc55db653f6daa73b9b91>

tensorboard/plugins/example/tensorboard_plugin_example/__init__.py
wchargin: Thanks.

tensorboard/plugins/example/tensorboard_plugin_example/__init__.py
nfelt: Let's just add the patch, we can test against an older bazel.

wchargin:
> Can we add some kind of minimal smoke test for this example plugin to
> ensure that it continues to be usable, unlike the old example repo?

Yes.

Such a smoke test needs to depend on the TensorBoard Pip package, as
well as building the Pip package for the example plugin and installing
both into a new virtualenv.

I think that probably the cleanest way to do this is to split the
current `//tensorboard/pip_package:build_pip_package`, which both builds
the package and tests it, into an actual Bazel target for the two wheels
(a `genrule` depending on an `sh_binary`) and a second target to test
them (a `sh_test`, as currently). Then the example plugin’s test can
depend on the TensorBoard wheels rather than rebuilding them.

I’m happy to do this, though it will take a bit of doing.

tensorboard/plugins/example/tensorboard_plugin_example/__init__.py
wchargin: Done: #2360 merged.

tensorboard/plugins/example/tensorboard_plugin_example/__init__.py
wchargin: (Using `index.js` per Node convention.)

wchargin:
Thanks to both of you for pushing back on the Pip package data deps and
the testing, respectively. I’d planned to set those up in a later PR,
but now that they’re both here I’m much happier with this one.

PR #2356, opened 2019-06-18
Load projector plugin dynamically

stephanwlee:
Build projector frontend code as a separate binary and loaded that as a
soruce (not from webfiles) in projector_plugin.

Made changes to our pip packaging to include the projector FE binaries.
Made changes to tensorboard dependencies to make local development easy.

Tested:
  - local dev by running `pip setup.py develop` in source tree A then
    running TensorBoard in source tree B
  - tested pip package building and running TensorBoard

### DIFF
--- a/tensorboard/pip_package/build_pip_package.sh
+++ b/tensorboard/pip_package/build_pip_package.sh
@@ -158,6 +158,8 @@ smoke() {
   curl -fs "http://localhost:$(cat port)/data/plugin/projector/runs" >projector_runs.json
   # logdir does not contain any checkpoints and thus an empty runs.
   grep '\[\]' projector_runs.json
+  curl -fs "http://localhost:$(cat port)/data/plugin/projector/projector_binary.html" >projector_binary.html
+  grep '<vz-projector-dashboard' projector_binary.html
   kill $!
 
   # Test TensorBoard APIs
--- a/tensorboard/pip_package/build_pip_package.sh
+++ b/tensorboard/pip_package/build_pip_package.sh
@@ -155,6 +155,9 @@ smoke() {
   grep '<tf-tensorboard' index.html
   curl -fs "http://localhost:$(cat port)/data/logdir" >logdir.json
   grep 'smokedir' logdir.json
+  curl -fs "http://localhost:$(cat port)/data/plugin/projector/runs" >projector_runs.json
+  # logdir does not contain any checkpoints and thus an empty runs.
+  grep '\[\]' projector_runs.json
   kill $!
 
   # Test TensorBoard APIs
--- a/tensorboard/pip_package/setup.py
+++ b/tensorboard/pip_package/setup.py
@@ -58,6 +58,9 @@ setup(
     packages=find_packages(),
     entry_points={
         'console_scripts': CONSOLE_SCRIPTS,
+        'tensorboard_plugins': [
+            'projector_plugin=tensorboard.plugins.projector.projector_plugin:ProjectorPlugin',
+        ],
     },
     package_data={
         'tensorboard': [
@@ -65,6 +68,10 @@ setup(
         ],
         'tensorboard.plugins.beholder': [
             'resources/*',
+        ],
+        'tensorboard.plugins.projector': [
+            'tf_projector_plugin/index.js',
+            'tf_projector_plugin/projector_binary.html',
         ],
     },
     # Disallow python 3.0 and 3.1 which lack a 'futures' module (see above).
--- a/tensorboard/pip_package/setup.py
+++ b/tensorboard/pip_package/setup.py
@@ -59,7 +59,7 @@ setup(
     entry_points={
         'console_scripts': CONSOLE_SCRIPTS,
         'tensorboard_plugins': [
-            'projector_plugin=tensorboard.plugins.projector.projector_plugin:ProjectorPlugin',
+            'projector = tensorboard.plugins.projector.projector_plugin:ProjectorPlugin',
         ],
     },
     package_data={
@@ -69,6 +69,7 @@ setup(
         'tensorboard.plugins.beholder': [
             'resources/*',
         ],
+        # Must keep this in sync with tf_projector_plugin:projector_assets
         'tensorboard.plugins.projector': [
             'tf_projector_plugin/index.js',
             'tf_projector_plugin/projector_binary.html',
--- a/tensorboard/plugins/projector/projector_plugin.py
+++ b/tensorboard/plugins/projector/projector_plugin.py
@@ -19,8 +19,10 @@ from __future__ import division
 from __future__ import print_function
 
 import collections
+import functools
 import imghdr
 import math
+import mimetypes
 import os
 import threading
 
@@ -260,7 +262,12 @@ class ProjectorPlugin(base_plugin.TBPlugin):
         TENSOR_ROUTE: self._serve_tensor,
         METADATA_ROUTE: self._serve_metadata,
         BOOKMARKS_ROUTE: self._serve_bookmarks,
-        SPRITE_IMAGE_ROUTE: self._serve_sprite_image
+        SPRITE_IMAGE_ROUTE: self._serve_sprite_image,
+        '/index.js':
+            functools.partial(self._serve_file, 'tf_projector_plugin/index.js'),
+        '/projector_binary.html':
+            functools.partial(
+                self._serve_file, 'tf_projector_plugin/projector_binary.html'),
     }
     return self._handlers
 
@@ -298,7 +305,7 @@ class ProjectorPlugin(base_plugin.TBPlugin):
 
   def frontend_metadata(self):
     return super(ProjectorPlugin, self).frontend_metadata()._replace(
-        element_name='vz-projector-dashboard',
+        es_module_path="/index.js",
         disable_reload=True,
     )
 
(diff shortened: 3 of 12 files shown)


tensorboard/plugins/projector/tf_projector_plugin/index.js
```diff
@@ -0,0 +1,20 @@
+export function render() {
```
nfelt: I'm a bit confused at the role of this rendering logic.  Since this is fully controlled by the plugin, this should be running in an iframe, right?  But then it's also creating its own iframe and inserting the projector HTML there?

Is that the typical flow that we expect for plugins based on custom web components - our iframe holds plugin code that creates its own iframe?  It seems a little convoluted.

tensorboard/plugins/projector/projector_plugin.py
```diff
@@ -260,7 +262,12 @@ def get_plugin_apps(self):
         TENSOR_ROUTE: self._serve_tensor,
         METADATA_ROUTE: self._serve_metadata,
         BOOKMARKS_ROUTE: self._serve_bookmarks,
-        SPRITE_IMAGE_ROUTE: self._serve_sprite_image
+        SPRITE_IMAGE_ROUTE: self._serve_sprite_image,
+        '/index.js':
+            functools.partial(self._serve_file, 'tf_projector_plugin/index.js'),
```
wchargin: (Also, oof—the fact that this `functools.partial` works is not obvious,
though in hindsight it can be mostly inferred from the fact that bound
methods work at all… no action required, I suppose.)

tensorboard/plugins/projector/tf_projector_plugin/tf-projector-plugin.html
```diff
@@ -0,0 +1,17 @@
...
+  margin: 0;
+  width: 100%;
+  font-family: Roboto, sans-serif;
+}
+</style>
+
+<body>
+  <vz-projector-dashboard></vz-projector-dashboard>
+</body>
+
```
wchargin: nit: Superfluous newline.

tensorboard/plugins/projector/tf_projector_plugin/index.js
```diff
@@ -0,0 +1,20 @@
...
+body {
+  height: 100%;
+  margin: 0;
+  width: 100%;
+}`;
+  document.head.appendChild(style);
+
+  const iframe = document.createElement('iframe');
+  iframe.src = './projector_binary.html';
+  Object.assign(iframe.style, {
```
wchargin: If you like, you could just use

```css
html,
iframe,
body {
  border: 0;
  height: 100%;
  margin: 0;
  width: 100%;
}
```

on the stylesheet to slim this down a bit. But no action required if you
prefer it this way.

tensorboard/plugins/projector/projector_plugin.py
```diff
@@ -298,7 +305,7 @@ def is_active(self):
 
   def frontend_metadata(self):
     return super(ProjectorPlugin, self).frontend_metadata()._replace(
-        element_name='vz-projector-dashboard',
+        es_module_path="/index.js",
```
wchargin: nit: Single-quotes for file consistency?

tensorboard/pip_package/setup.py
```diff
@@ -66,6 +69,10 @@ def get_readme():
         'tensorboard.plugins.beholder': [
             'resources/*',
         ],
+        'tensorboard.plugins.projector': [
+            'tf_projector_plugin/index.js',
```
wchargin: Hmm; this is a little scary. Can we at least add a comment here that
this MUST be kept in sync with the srcs for the `projector_assets` web
library in Bazel? As is, it seems too easy to add a file dependency that
works fine in dev but is omitted in the built Pip package.

tensorboard/plugins/projector/tf_projector_plugin/BUILD
```diff
@@ -0,0 +1,36 @@
...
+load("//tensorboard/defs:web.bzl", "tf_web_library")
+load("//tensorboard/defs:vulcanize.bzl", "tensorboard_html_binary")
+
+licenses(["notice"])  # Apache 2.0
+
+tf_web_library(
+    name = "projector_assets",
+    srcs = [
+        ":projector_binary.html",
+        "index.js"
```
wchargin: nit: Missing trailing comma; can you run this file through `buildifier`?

tensorboard/plugins/projector/tf_projector_plugin/BUILD
```diff
@@ -0,0 +1,36 @@
...
+
+tf_web_library(
+    name = "tf_projector_plugin",
+    srcs = [
+        "tf-projector-plugin.html",
+    ],
+    path = "/tf-projector",
+    deps = [
+        "//tensorboard/plugins/projector/vz_projector",
+        "@com_google_fonts_roboto",
```
wchargin: To be clear, this bundles an extra copy of Roboto into the resulting
binary, right? (That’s an extra 644K raw and 492K compressed on a 3MB
Pip package.)

Also, if I’m reading this correctly, the Roboto font will still be
fetched from `/font-roboto/…`, which will resolve to the TensorBoard
root server, so we’re bundling the extra fonts and not actually using
them. :-(

Unfortunately, I don’t really see a quick fix for this with our current
build setup, given the whole “`fonts.bzl` generates a build file that
generates an html file with CSS that’s packaged into a `web_library`”
machinery. @nfelt and I discussed this briefly—there are a couple fixes
that we could take (such as making the generated CSS use relative paths
and serving the fonts from multiple paths with the same CSS), but I
can’t in good conscience ask for any of those in this PR. So… maybe open
an issue after this lands?

tensorboard/plugins/projector/tf_projector_plugin/tf-projector-plugin.html
```diff
@@ -0,0 +1,17 @@
+<link rel="import" href="../font-roboto/roboto.html">
+<link rel="import" href="../vz-projector/vz-projector-dashboard.html">
+
+<style>
```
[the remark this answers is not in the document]
stephanwlee: Yes, we are creating an iframe within an iframe so both needs to be styled separately.

tensorboard/plugins/projector/tf_projector_plugin/BUILD
stephanwlee: Ok good news and bad news. 

> this bundles an extra copy of Roboto into the resulting

Good news:
It seems like this line adds a style tag that has content in this file: https://github.com/tensorflow/tensorboard/blob/052a8ea45e90804544d0291200ddea7ff85099f2/third_party/fonts.bzl#L400-L434

I tinkered around by with the fonts.bzl and basically omitted the style tag (and prevented CSS fetching the font) to check the difference in the file size. The CSS basically adds x-kB which wasn't too bad. 

So-so news:
It seems like, regardless, paper-style has dependency on roboto and it gets loaded almost no matter what. 
https://github.com/tensorflow/tensorboard/blob/052a8ea45e90804544d0291200ddea7ff85099f2/third_party/polymer.bzl#L1199-L1205

roboto.html -> typography.html -> paper-style.html

Bad news:
Yes, the font fetching fails when you use path_prefix flag. We can easily fix things by using a relative path in the fonts.bzl but it gets quite tricky for the dynamic plugins. We may be able to alleviate this by creating a dynamic plugin variant of the roboto.html which feels not good at the moment.

tensorboard/plugins/projector/projector_plugin.py
```diff
@@ -471,6 +478,14 @@ def _append_plugin_asset_directories(self, run_path_pairs):
       assets_path_pair = (run, os.path.abspath(assets_dir))
       run_path_pairs.append(assets_path_pair)
 
+  @wrappers.Request.application
+  def _serve_file(self, file_path, request):
```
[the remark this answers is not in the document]
stephanwlee: I thought about that but could not think of a good place for this utility. I am currently leaning towards putting this somewhere in the base_plugin but I don't know how I feel about that. The major issue with putting this under http_util is the fact that it will need to take the asset_provider as an input and also become a bit stateful for memoization or up front gzipping. In my naive view, utils should try to remain very stateless. 

Anyways, since there is no good candidate for dynamic-plugification immediately, we have some time to do so and was going to take an attempt when we need that.

Btw, I did try just doing 302, from this endpoint, to /file/path but that changed the base URL of the iframe which was undesirable.

tensorboard/plugins/projector/tf_projector_plugin/index.js
stephanwlee: > Is that the typical flow that we expect for plugins based on custom web components

Yes but with a nuance: Yes you are using WebComponents that require HTML counterpart -- i.e., Polymer 1 and 2. Technically, we can still be using HTML import (which seems to work in my Chrome 75), I didn't want to rely on obsolete browser feature. Without using HTML import, a way to load HTML binary would be using an iframe which is the reason for this nested iframe.

tensorboard/plugins/projector/projector_plugin.py
stephanwlee: Ack. Btw, I copied https://github.com/tensorflow/tensorboard/blob/052a8ea45e90804544d0291200ddea7ff85099f2/tensorboard/plugins/core/core_plugin.py#L91-L105. 🙈

wchargin: Okay, fair enough—:see_no_evil: indeed. :-)

tensorboard/plugins/projector/tf_projector_plugin/BUILD
wchargin: Right: that was my analysis, too, so I think that we’re on the same page.

tensorboard/pip_package/build_pip_package.sh
```diff
@@ -158,6 +158,8 @@ smoke() {
   curl -fs "http://localhost:$(cat port)/data/plugin/projector/runs" >projector_runs.json
   # logdir does not contain any checkpoints and thus an empty runs.
   grep '\[\]' projector_runs.json
+  curl -fs "http://localhost:$(cat port)/data/plugin/projector/projector_binary.html" >projector_binary.html
```
wchargin: Great addition; thanks!

PR #2363, opened 2019-06-19
build: promote Pip wheels to Bazel target

wchargin:
Summary:
Running `bazel build //tensorboard/pip_package` now actually builds the
Pip package. Other targets can depend on the built wheels, which are
provided in a tarball. Smoke tests have been separated to a new
`:test_pip_package` target.

Test Plan:
Check that everything builds:

```
$ bazel build //tensorboard/pip_package 2>/dev/null
$ echo $?
0
$ tar tzf bazel-bin/tensorboard/pip_package/pip_packages.tar.gz
tensorboard-1.15.0a0-py2-none-any.whl
tensorboard-1.15.0a0-py3-none-any.whl
$ bazel run //tensorboard/pip_package:test_pip_package >/dev/null 2>&1
$ echo $?
0
$ tmpdir="$(mktemp -d)"
$ bazel run //tensorboard/pip_package:build_pip_package -- "${tmpdir}" \
> >/dev/null 2>/dev/null
$ ls -1 "${tmpdir}"
tensorboard-1.15.0a0-py2-none-any.whl
tensorboard-1.15.0a0-py3-none-any.whl
```

Then:

  - Add an `assert False` to the `pb` function in `scalar/summary.py`,
    and verify that `:test_pip_package` fails with default options but
    passes with `--tf-version ""`.
  - Add an `assert False` to `tensorboard/main.py` and verify that the
    test fails even with `--tf-version ""`.
  - Add `def foo(): yield from None` to `tensorboard/main.py`, and
    verify that, when run in a Python 3 virtualenv, the test fails by
    default but passes with `--default-python-only`.

wchargin-branch: bazel-pip-package


tensorboard/pip_package/BUILD
```diff
@@ -5,9 +5,36 @@ package(default_visibility = ["//visibility:private"])
...
-# pip install -U /tmp/tensorboard/*py2*.pip
+genrule(
+    name = "pip_package",
+    # Pip packages are provided in a tarball because we can't statically
+    # know the file names: they must contain the TensorBoard version,
+    # which is only defined in a Python module. Dependents of this
+    # target can extract the tarball and glob its contents.
+    outs = ["pip_packages.tar.gz"],
+    cmd = "\n".join([
+        "tmpdir=\"$$(mktemp -d)\" &&",
```
nfelt: It's hard to read this code embedded in the genrule definition like this.

What if we just made deterministic_tar_gz a data dep for build_pip_package, so that if the argument (currently named OUTPUT_DIR) ends in `.tar.gz` it emits the pip packages as an archive in the given file instead?

If we did that and also ensured that pip package building doesn't emit extraneous output, then it seems like we could make the entire genrule here just a one-liner: `$(location :build_pip_package) $@`.

tensorboard/pip_package/test_pip_package.sh
```diff
@@ -0,0 +1,217 @@
+#!/bin/sh
+# Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
nfelt: 2019

wchargin: Sure. (I went with 2015 because it’s a file split, but 2019 is fine.)

tensorboard/pip_package/BUILD
wchargin: Okay; that’s totally fine with me if we just make it always emit a
`.tar.gz`, and stomachable if it cases on the filename. Will change
tomorrow.

tensorboard/pip_package/build_pip_package.sh
```diff
@@ -12,9 +12,31 @@
...
-  RUNFILES="$(CDPATH= cd -- "$0.runfiles" && pwd)"
+
+set -eu
+
+usage() {
+  cat <<EOF
+usage: build_pip_package OUTPUT_DIR
+
+Build TensorBoard Pip packages and store the resulting wheel files
+into OUTPUT (a \`.tar.gz\` path or directory).
```
nfelt: nit: I'd say ```a directory or a \`.tar.gz\` path``` to make it clear that that there is no need for the directory name to end in `.tar.gz` since it's otherwise strictly speaking ambiguous

tensorboard/pip_package/build_pip_package.sh
```diff
@@ -94,140 +83,39 @@ cp -LR "${RUNFILES}/org_tensorflow_serving_api/tensorflow_serving" tensorboard/_
...
-    smoke 2 "${tf_version}"
-    smoke 3 "${tf_version}"
-    ;;
-  "1")
-    # Empty string indicates to use the default "python".
-    smoke "" "${tf_version}"
-    ;;
-  "0")
-    printf "\nSkipping smoke test\n\n"
+cd "${original_wd}"  # Bazel gives "${output}" as a relative path >_>
```
nfelt: Can we just resolve `output` to an absolute path while we're still in the original working directory, rather than having to `cd` back to it later?

tensorboard/pip_package/build_pip_package.sh
```diff
@@ -94,140 +83,39 @@ cp -LR "${RUNFILES}/org_tensorflow_serving_api/tensorflow_serving" tensorboard/_
...
+  *.tar.gz)
+    mkdir -p "$(dirname "${output}")"
+    "${RUNFILES}/org_tensorflow_tensorboard/tensorboard/pip_package/deterministic_tar_gz" \
+        "${output}" "${workdir}"/dist/*.whl
     ;;
   *)
-    echo >&2 'fatal: unknown smoke value:' "${smoke}"
-    exit 1
+    if ! [ -d "${output}" ]; then
+      printf >&2 'fatal: no such output directory: %s\n' "${output}"
```
nfelt: Seems more straightforward to just `mkdir -p "${output}"` here if it doesn't exist.  I don't see any particular reason why it should require the directory to already exist, and this makes it more consistent with the `.tar.gz` behavior?

tensorboard/pip_package/build_pip_package.sh
wchargin: Agreed, absolutely. (I usually try to do this, but missed this one.)

tensorboard/pip_package/build_pip_package.sh
wchargin: I don’t know a portable way to do that. GNU `readlink -e`/`-f` works
perfectly on Linux, but on macOS `readlink`(1) is basically just
`readlink`(2). `realpath`(1) is also not POSIX. ([Here is a handy list
of portable utilities][posix]—it’s not large, especially when you look
at (e.g.) `xargs` and see that there’s no portable `-0` option!)

[posix]: http://pubs.opengroup.org/onlinepubs/9699919799/idx/utilities.html

I mean, there’s always

```shell
path="$(python -c 'print(__import__("os").path.realpath(__import__("sys").argv[1]))' "${path}")"
```

which has the downside of being a bit hideous (and also not correct
because command substitutions strip whitespace); just `cd`ing back is
significantly simpler imho.

tensorboard/pip_package/build_pip_package.sh
wchargin: Requiring the output directory to exist is more consistent with other applications
that write files to specified directories—

```
$ tar czf - a.c | tar xzf - -C ./nonexistent/
tar: nonexistent: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
$ cp a.c ./nonexistent/
cp: cannot create regular file './nonexistent/': Not a directory
$ mv a.c ./nonexistent/
mv: cannot move 'a.c' to './nonexistent/': Not a directory
$ ln -s a.c ./nonexistent/
ln: target './nonexistent/' is not a directory: No such file or directory
$ gcc a.c -o ./nonexistent/
/usr/bin/ld: cannot open output file ./nonexistent/: Is a directory
collect2: error: ld returned 1 exit status
$ gcc a.c -o ./nonexistent/a.c
/usr/bin/ld: cannot open output file ./nonexistent/a.c: No such file or directory
collect2: error: ld returned 1 exit status
$ printf 'shell redirections\n' >./nonexistent/out
bash: ./nonexistent/out: No such file or directory
```

—so I’d prefer to keep it as is unless you feel strongly. (The explicit
error is just to have a nicer message.)

tensorboard/pip_package/build_pip_package.sh
wchargin: (If there _is_ a portable way to do this, I would love to know!)

nfelt: What about

```sh
case "${output}" in 
  /*) ;;
  *) output="${PWD}/${output}" ;;
esac
```

PR #2365, opened 2019-06-19
ci: add trailing whitespace test

wchargin:
Summary:
Prevents regressions of nice fixes like #2364.

Test Plan:
Run the script, and note that it passes with no output. Then remove
`README.md` from the exceptions list and add `nonexistent.txt`, and
observe that the script fails with:

```
Superfluous trailing whitespace:
README.md:10:For in-depth information on the Graph Visualizer, see this tutorial: $
README.md:18:through setting up and using TensorBoard. There's an associated $
README.md:36:For more details, see $
README.md:69:For this, you need $

Stale exceptions (no whitespace problems; prune exceptions list):
nonexistent.txt
```

wchargin-branch: whitespace-hygiene-test

### DIFF
--- a/tensorboard/tools/whitespace_hygiene_test.py
+++ b/tensorboard/tools/whitespace_hygiene_test.py
@@ -112,10 +112,8 @@ def git_grep(pattern):
     getattr(sys.stderr, "buffer", sys.stderr).write(stderr)  # Python 2 compat
     sys.exit(1)
   result = []
-  while stdout:
-    (filename_raw, _unused_nul, rest) = stdout.partition(b"\0")
-    (line_number_raw, _unused_nul, rest) = rest.partition(b"\0")
-    (line_raw, _unused_lf, stdout) = rest.partition(b"\n")
+  for line in stdout.splitlines():  # assumes no newline characters in filenames
+    (filename_raw, line_number_raw, line_raw) = line.split(b"\0", 2)
     match = Match(
         filename=filename_raw.decode("utf-8", errors="replace"),
         line_number=int(line_number_raw),
--- a/tensorboard/tools/whitespace_hygiene_test.py
+++ b/tensorboard/tools/whitespace_hygiene_test.py
@@ -1,0 +1,135 @@
+#!/usr/bin/python
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+"""Check for superfluous whitespace at ends of lines.
+
+Keeps diffs clean and persnickety developers happy.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import os
+import subprocess
+import sys
+
+
+# Remove files from this list as whitespace errors are fixed.
+exceptions = frozenset([
+    "README.md",
+    "docs/r1/overview.md",
+    "tensorboard/components/tf_color_scale/palettes.ts",
+    "tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-dashboard.html",
+    "tensorboard/plugins/debugger/tf_debugger_dashboard/tf-debugger-resizer.html",
+    "tensorboard/plugins/graph/tf_graph_common/edge.ts",
+    "tensorboard/plugins/hparams/api.d.ts",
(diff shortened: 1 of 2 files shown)


tensorboard/tools/whitespace_hygiene_test.py
```diff
@@ -0,0 +1,135 @@
...
+  """
+  cmd = ["git", "grep", "-Izn", "--", pattern]
+  p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+  (stdout, stderr) = p.communicate()
+  if stderr:
+    getattr(sys.stderr, "buffer", sys.stderr).write(stderr)  # Python 2 compat
+    sys.exit(1)
+  result = []
+  while stdout:
+    (filename_raw, _unused_nul, rest) = stdout.partition(b"\0")
```
nfelt: Since realistically we're not going to have newlines in filenames committed to our git repo, it seems simpler and more obvious to just do `for line in stdout.splitlines(): filename_raw, line_number_raw, line_number = line.split(b"\0", 2)`.

tensorboard/tools/whitespace_hygiene_test.py
```diff
@@ -0,0 +1,135 @@
...
+        line_number=int(line_number_raw),
+        line=line_raw.decode("utf-8", errors="replace"),
+    )
+    result.append(match)
+  return result
+
+
+def chdir_to_repo_root():
+  toplevel = subprocess.check_output(["git", "rev-parse", "--show-toplevel"])
+  toplevel = toplevel[:-1]  # trim trailing LF
```
nfelt: `os.chdir(toplevel.rstrip())`?

tensorboard/tools/whitespace_hygiene_test.py
```diff
@@ -0,0 +1,135 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import os
+import subprocess
+import sys
+
+
+# Remove files from this list as whitespace errors are fixed.
```
nfelt: Are any of these ones that should actually be longstanding exceptions?  If not, why not just commit the whitespace fixes to these now and avoid ever checking in the exceptions list support and stale exceptions detection?

tensorboard/tools/whitespace_hygiene_test.py
wchargin: This is again for whitespace-correctness—nothing wrong with

```
$ git clone -q gh:tensorflow/tensorboard "tensorboard "
$ cd "tensorboard "
$ pwd | xxd
00000000: 2f74 6d70 2f74 6d70 2e49 426d 7057 5446  /tmp/tmp.IBmpWTF
00000010: 7573 712f 7465 6e73 6f72 626f 6172 6420  usq/tensorboard 
00000020: 0a                                       .
$ python -c '
> import os
> import subprocess
> toplevel = subprocess.check_output(["git", "rev-parse", "--show-toplevel"])
> os.chdir(toplevel.rstrip())
> '
  File "<string>", line 5, in <module>
OSError: [Errno 2] No such file or directory: '/tmp/tmp.IBmpWTFusq/tensorboard'
```

It really costs us nothing here to do the right thing, and I’d love to
avoid proliferating even more programs that behave incorrectly on
filenames that have whitespace.

tensorboard/tools/whitespace_hygiene_test.py
wchargin: > Are any of these ones that should actually be longstanding exceptions?

Nope. (The only potential genuine exception that I could think of would
be Markdown hard-breaks, which should rarely be used anyway.)

> why not just commit the whitespace fixes to these now and avoid ever
> checking in the exceptions list support

Two benefits:

  - we avoid a megacommit that kills all the blame (existing such
    commits, like e1adbef8 and friends, are perennially frustrating in
    everyday TensorBoard development); and
  - in the case that we _do_ need a genuine exception (say, some strange
    data format for a plugin’s test data), the path of least resistance
    will be to mark it as an exception rather than just disabling the
    test because there’s an urgent deadline.

tensorboard/tools/whitespace_hygiene_test.py
wchargin: IMHO, the correct version is sufficiently simple as to be
unobjectionable. But in the interest of compromise, I can switch this
over to the incorrect-but-fewer-lines version. (I’d rather cede this
point, which is under our control, than the other, which is under the
control of whoever clones the repository.)

tensorboard/tools/whitespace_hygiene_test.py
nfelt: Running your script shows that there are only 88 affected individual lines.  I don't think an 88-line delta is a megacommit and it's certainly nothing like a 11,000 line one.  Blame impact would be localized only to affected lines.

(Though, even if we were talking about reformatting en masse (e.g. buildifiering all our build files) I think it's better to just get it over with.  Doing it one touched line at a time will never actually fix everything so you never achieve the uniformity that's the desired goal anyway.  Doing it file-by-file means that you still clobber the blame but now there's hundreds of such commits instead of one.)

The second benefit is entirely speculative.

wchargin: Good point: I was scared off by the file count and didn’t look at the
delta. That does make it much less bad. I’m convinced; will send out a
fix.

> The second benefit is entirely speculative.

Certainly.

PR #2370, opened 2019-06-20
Polymer 2: pr-curve: fix the basic functionalities

stephanwlee:
Problem 1: the DOM structure
Problem 2: tf-pr-curve-steps-selector broke because its usage of
querySelector. It uses an internal state, `runsToStepIndex`, to record
run to stepIndex (note the distinction from just step) map. It was
manually setting the value to the paper-sliders and listened to the
changed event via listeners. Something in Polymer 2 broke. Instead,
now, we use data-binding to update the value of the paper-slider
Polymer way.

stepIndex vs. step: when there are tons of data, due to our reservoir
sampling, there can be only a subset of the steps. Say we have steps
[0, 4, 10, 1000] for a run foo. When the steps are rendered, it has to
translated to indices since the paper-slider does not supporting an enum.
`runToStepIndex` stores slider index which gets translated to real
`step` via `runToAvailableTimeEntries`.

### DIFF
--- a/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-steps-selector.html
+++ b/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-steps-selector.html
@@ -141,9 +141,7 @@ limitations under the License.
       if (isNaN(val)) {
         delete newRunToStepIndex[run];
       } else {
-        Object.assign(newRunToStepIndex, {
-          [run]: event.target.immediateValue,
-        });
+        newRunToStepIndex[run] = event.target.immediateValue;
       }
       this._runToStepIndex = newRunToStepIndex;
     },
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-steps-selector.html
```diff
@@ -136,56 +134,29 @@
...
-        if (!_.isNumber(runToStepIndex[run])) {
+        if (!_.isNumber(this._runToStepIndex[run])) {
           // This run had previously lacked a slider. Initially set the slider
           // for the run to point to the last step.
-          newRunToStepIndex[run] = entries.length - 1;
-          runAdded = true;
-        } else {
-          // The step for this run did not change.
-          newRunToStepIndex[run] = runToStepIndex[run];
+          this.set(`_runToStepIndex.${run}`, entries.length - 1);
```
wchargin: Does this work with runs names like `foo.baz.bar`? This can be common
for TensorBoard runs that encode hyperparameter values in the run name,
like `train/lr=0.01;dropout=0.2`.

(If so, how does it work?)

stephanwlee: Thanks a lot for catching that. Yeah that wouldn't work and I now want to do a quick audit on where we use `this.set`.
Addressed it by using `Object.assign`.

stephanwlee:
> Can you elaborate? I don’t see any {{ … }} two-way binding added in
this change, or in the step selector file at all.

Two week old me made a mistake and it doesn't make any sense. I changed the commit message to say "we now use data-binding" instead.

PR #2371, opened 2019-06-20
Confirm that the connection to tensorboard works or change to localhost

miguelmorin:

The TensorBoard log told me to go to `http://laptop-name.corp.domain.tld:6006` and that connection failed, but `http://localhost:6006` worked.

I tested the connection to the host with `socket()`, and if it is closed, the host becomes `localhost`.

* Screenshots of UI changes

No UI changes.

I ran `export PYTHONPATH="$PYTHONPATH:$HOME/code/tensorflow:$HOME/code:tensorboard"; python3 tensorboard/program.py`, which threw this unrelated error (and both repos are in sync with upstream):

```
  File "/Users/mmorin/code/tensorboard/tensorboard/data_compat.py", line 24, in <module>
    from tensorboard.compat.proto import summary_pb2
ImportError: cannot import name 'summary_pb2'
```

None.

### DIFF
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -728,7 +728,7 @@ class WerkzeugServer(serving.ThreadedWSGIServer, TensorBoardServer):
 
     def get_url(self):
         if self._auto_wildcard:
-            display_host = socket.gethostname()
+            display_host = socket.getfqdn()
         else:
             host = self._host
             display_host = (
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -162,6 +162,7 @@ class TensorBoard(object):
                 raise ValueError("Duplicate subcommand name: %r" % name)
             self.subcommands[name] = subcommand
         self.flags = None
+        self.display_host = None  # Will be set by get_url() below
 
     def configure(self, argv=("",), **kwargs):
         """Configures TensorBoard behavior via flags.
@@ -727,17 +728,25 @@ class WerkzeugServer(serving.ThreadedWSGIServer, TensorBoardServer):
             logger.error("HTTP serving error", exc_info=exc_info)
 
     def get_url(self):
-        if self._auto_wildcard:
-            display_host = socket.getfqdn()
-        else:
-            host = self._host
-            display_host = (
-                "[%s]" % host
-                if ":" in host and not host.startswith("[")
-                else host
-            )
+        if not self.display_host:
+            if self._auto_wildcard:
+                self.display_host = socket.getfqdn()
+
+                # Confirm that the connection is open, otherwise change to `localhost`
+                try:
+                    socket.create_connection((self.display_host, self.server_port), timeout = 1)
+                except socket.error as e:
+                    self.display_host = "localhost"
+
+            else:
+                host = self._host
+                self.display_host = (
+                    "[%s]" % host
+                    if ":" in host and not host.startswith("[")
+                    else host
+                )
         return "http://%s:%d%s/" % (
-            display_host,
+            self.display_host,
             self.server_port,
             self._flags.path_prefix.rstrip("/"),
         )
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -731,7 +731,6 @@ class WerkzeugServer(serving.ThreadedWSGIServer, TensorBoardServer):
         if not self._url:
             if self._auto_wildcard:
                 display_host = socket.getfqdn()
-
                 # Confirm that the connection is open, otherwise change to `localhost`
                 try:
                     socket.create_connection(
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -597,7 +597,7 @@ class WerkzeugServer(serving.ThreadedWSGIServer, TensorBoardServer):
             host = "localhost"
 
         self._host = host
-        self.display_host = None  # Will be set by get_url() below
+        self._url = None  # Will be set by get_url() below
 
         self._fix_werkzeug_logging()
         try:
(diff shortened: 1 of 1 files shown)


tensorboard/program.py
```diff
@@ -526,6 +526,14 @@ def get_url(self):
       host = self._flags.host
       display_host = (
           '[%s]' % host if ':' in host and not host.startswith('[') else host)
+
+    # Confirm that the connection is open, otherwise change to `localhost`
+    socket.setdefaulttimeout(1)
+    try:
+        socket.socket().connect((display_host, self.server_port))
+    except socket.timeout as e:
```
nfelt: Let's catch `socket.error` generally so that if there is some other error attempting the connection (not a timeout), we fall back to localhost in that case too.  For example, it looks like if the DNS name doesn't resolve at all, this line fails with `socket.gaierror` rather than `socket.timeout`:

```python
>>> socket.socket().connect(("foobartensorboard.com", 6006))
  File "/usr/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
socket.gaierror: [Errno -2] Name or service not known
```

tensorboard/program.py
```diff
@@ -526,6 +526,14 @@ def get_url(self):
       host = self._flags.host
       display_host = (
           '[%s]' % host if ':' in host and not host.startswith('[') else host)
+
+    # Confirm that the connection is open, otherwise change to `localhost`
+    socket.setdefaulttimeout(1)
+    try:
```
nfelt: Now that creating the URL involves an actual network request, let's avoid doing that on each call to `get_url()` and just cache the constructed URL in a property and return that if it's called another time.

tensorboard/program.py
```diff
@@ -526,6 +526,14 @@ def get_url(self):
...
+        socket.socket().connect((display_host, self.server_port))
```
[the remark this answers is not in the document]
miguelmorin: I agree and addressed this in 88e25, 6a5fc8, and 6c796 respectively.

tensorboard/program.py
```diff
@@ -526,6 +526,14 @@ def get_url(self):
       host = self._flags.host
       display_host = (
           '[%s]' % host if ':' in host and not host.startswith('[') else host)
+
+    # Confirm that the connection is open, otherwise change to `localhost`
+    socket.setdefaulttimeout(1)
```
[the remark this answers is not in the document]
miguelmorin: Well spotted, fixed in 1ae50.

tensorboard/program.py
miguelmorin: Very well spotted, fixed in 71307.

tensorboard/program.py
miguelmorin: Yes, I proposed a solution in bf05d. Another solution could be to move the code into `__init__()` of `WerkzeugServer` and change `get_url()` to only the last two lines of the code. Also, why do we have the property `_auto_wildcard` if it's simply `not host`?

tensorboard/program.py
```diff
@@ -520,21 +521,25 @@ def handle_error(self, request, client_address):
       logger.error('HTTP serving error', exc_info=exc_info)
 
   def get_url(self):
-    if self._auto_wildcard:
-      display_host = socket.gethostname()
-    else:
-      host = self._flags.host
-      display_host = (
+    if not self.display_host:
```
nfelt: Let's make the variable `self._url` instead and just store the full URL.

It would probably read better if we factor out something like `get_url_host()` e.g.

```python
def get_url_host(self):
  if not self._auto_wildcard:
    return self._flags.host
  host = socket.getfqdn()
  try:
    socket.create_connection((host, self.server_port), timeout=1)
    return host
  except socket.error as e:
    return 'localhost'

def get_url(self):
  if not self._url:
    host = self.get_url_host()
    display_host = (
        '[%s]' % host if ':' in host and not host.startswith('[') else host)
    self._url = 'http://%s:%d%s/' % (self.display_host, self.server_port,
        self._flags.path_prefix.rstrip('/'))
  return self._url
```

tensorboard/program.py
```diff
@@ -520,21 +521,25 @@ def handle_error(self, request, client_address):
...
+
+      if self._auto_wildcard:
+        display_host = socket.getfqdn()
```
nfelt: Just "host" is a better name here since this is still a real hostname; the name "display_host" was meant to be only for the host after adding brackets around an IPv6 address so it renders correctly when printed out in the URL.  See also suggestion above for a `get_url_host()` helper.

miguelmorin:
@nfelt Thanks for your patience too, I was overwhelmed for a while. Given the move of tensorboard and the conflicts with the base branch, does this PR still make sense, or shall I open a new one from the new codebase?

nfelt:
@miguelmorin Thanks for following up - I might be misunderstanding but I don't think anything actually moved?  There was a behavior change in https://github.com/tensorflow/tensorboard/pull/2589 though that as mentioned in my last comment makes it so that now by default the URL would use  `localhost` anyway in all cases (unless passing `--bind_all` or an explicit hostname).

This PR would still affect the default value shown when using `--bind_all` but it's more of an edge case now.  If you're still interested in landing this and want to update the PR I think we're still open to it, but also fine if you'd prefer to drop it.

miguelmorin:
I misnamed the change of default behavior as a "move". Yes, I still want to update the PR for this edge case. What is simpler for you: that I start a new branch from the latest code or that I rebase master onto this branch?

Iin #2589 you mention "wildcard dual-binding behavior": is that when TensorBoard "defaults to serving on the entire local network"?

nfelt:
> What is simpler for you: that I start a new branch from the latest code or that I rebase master onto this branch?

@miguelmorin the easiest option is probably if you can rebase your local changes onto latest master and then force-push that branch to `minguelmorin:localhost` to update this PR.  Or if you prefer you can close this and open a new PR, either way.

> Iin #2589 you mention "wildcard dual-binding behavior": is that when TensorBoard "defaults to serving on the entire local network"?

Yes, this was the default behavior prior to 2.0.

miguelmorin:
Hi @nfelt, I've finally made time to finish this. The rebasing was straightforward and I force-pushed the branch. Your requested changes are still marked as unsolved and those commits are now gone. How would you like to proceed?

tensorboard/program.py
```diff
@@ -597,6 +597,7 @@ def __init__(self, wsgi_app, flags):
             host = "localhost"
 
         self._host = host
+        self.display_host = None  # Will be set by get_url() below
```
nfelt: Let's just cache the entire URL for this, so `self._url`.

That way we can skip even the last formatting step in `get_url()` if we already have generated the URL.

tensorboard/program.py
```diff
@@ -727,17 +728,27 @@ def handle_error(self, request, client_address):
...
-            host = self._host
-            display_host = (
-                "[%s]" % host
-                if ":" in host and not host.startswith("[")
-                else host
-            )
+        if not self.display_host:
+            if self._auto_wildcard:
+                self.display_host = socket.getfqdn()
+
```
nfelt: nit: can we leave out the blank lines inside the function, since it's not that long?

If we want separation for clarity I'd recommend just factoring out a private helper function as described in the previous comment.

tensorboard/program.py
miguelmorin: Yes, I cached the entire URL in 3f99276.

tensorboard/program.py
miguelmorin: Do you mean just removing line 734, which I did in f3bbb87? If not, I don't see you mean by "as described in the previous comment".

miguelmorin:
Sorry, I missed that second part of the 18 July comment. My changes already seem outdated, do you want me pull the master branch and do a manual rebase of my changes?

nfelt:
I don't see any conflicts with the base branch?  I can go ahead and merge now unless there were further changes you were going to make.

miguelmorin:
No, I don't see conflicts with the base branch. I see `tensorboard/program.py Outdated` above the code in your comment under commit 2b6067a.

nfelt:
Outdated just means the comment is on code that isn't at the most recent commit touching that file; it's fine in this case.  Will merge.

PR #2373, opened 2019-06-21
Support undefined number of data points in the mesh.

podlipensky:
Adds support of undefined (None) number of vertices/faces/colors in the mesh.

How to test:
- Run updated unit tests.
- Modify mesh_demo.py to provide different number of points at each run:
```
# Add extra vertex (1,2,3) to the mesh.
vertices_2 = np.concatenate((vertices, [[1, 2, 3]]))  
vertices_2 = np.expand_dims(vertices_2, 0)
...
for i in range(_MAX_STEPS):
    ver = vertices if i % 2 == 0 else vertices_2
    summary = sess.run(meshes_summary, feed_dict={
        vertices_tensor: ver,
        faces_tensor: faces,
        colors_tensor: colors,
        step: i,
    })
```

Run demo app and then launch tensorboard - observe mesh being rendered.

Also this PR changes the way we load data - from now on, every step is considered as unique data point and should be requested separately from the server (similar to image plugin). Please note that each mesh/step will produce 1 to 3 requests based on what data was passed into the summary (i.e. if there are no colors specified, then it will request only vertices and faces).

This PR depends on some TF-Graphics changes (I'll update it with new SHA once changes merged).


tensorboard/plugins/mesh/summary.py
```diff
@@ -157,19 +175,23 @@ def pb(name,
...
 
   summaries = []
-  tensors = [(vertices, plugin_data_pb2.MeshPluginData.VERTEX, tf.float32),
-             (faces, plugin_data_pb2.MeshPluginData.FACE, tf.int32),
-             (colors, plugin_data_pb2.MeshPluginData.COLOR, tf.uint8)]
-  for tensor, content_type, data_type in tensors:
+  tensors = [(vertices, plugin_data_pb2.MeshPluginData.VERTEX),
+             (faces, plugin_data_pb2.MeshPluginData.FACE),
+             (colors, plugin_data_pb2.MeshPluginData.COLOR)]
+  content_types = [tf.float32, tf.int32, tf.uint8]
```
stephanwlee: This is data type.

tensorboard/plugins/mesh/summary.py
```diff
@@ -157,19 +175,23 @@ def pb(name,
   json_config = _get_json_config(config_dict)
 
   summaries = []
-  tensors = [(vertices, plugin_data_pb2.MeshPluginData.VERTEX, tf.float32),
-             (faces, plugin_data_pb2.MeshPluginData.FACE, tf.int32),
-             (colors, plugin_data_pb2.MeshPluginData.COLOR, tf.uint8)]
-  for tensor, content_type, data_type in tensors:
+  tensors = [(vertices, plugin_data_pb2.MeshPluginData.VERTEX),
```
stephanwlee: Since this is used in various places now, I think using a namedtuple would make this more readable/manageable.

tensorboard/plugins/mesh/mesh_plugin.py
```diff
@@ -258,8 +269,9 @@ def _serve_mesh_metadata(self, request):
 
     # We convert the tensor data to text.
     response = [
-        self._get_tensor_metadata(tensor, meta.content_type, meta.shape,
-                                  meta.json_config)
+        self._get_tensor_metadata(
+          tensor, meta.content_type, meta.components, meta.shape,
```
stephanwlee: You will need to update doc: https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/mesh/http_api.md

tensorboard/plugins/mesh/mesh_plugin.py
```diff
@@ -216,7 +226,8 @@ def _serve_mesh_data(self, request):
     Returns:
       werkzeug.Response either float32 or int32 data in binary format.
     """
-    tensor_events = self._collect_tensor_events(request)
+    timestamp = float(request.args.get('timestamp', 0.))
```
stephanwlee: Also document this (put the unit of this timestamp: is it milliseconds? UTC?)

tensorboard/plugins/mesh/mesh_plugin_test.py
```diff
@@ -74,7 +78,7 @@ def setUp(self):
                                                 mesh_color.faces.shape)
     mesh_color_colors = tf.compat.v1.placeholder(tf.uint8,
                                                  mesh_color.colors.shape)
-    self.data = [point_cloud, mesh_no_color, mesh_color]
+    self.data = [point_cloud, mesh_no_color, mesh_no_color_upd,  mesh_color]
```
stephanwlee: Remove extra space after `mesh_no_color_upd`.

tensorboard/plugins/mesh/mesh_plugin_test.py
```diff
@@ -61,10 +64,11 @@ def setUp(self):
                                                     point_cloud.vertices.shape)
 
     mesh_no_color = test_utils.get_random_mesh(2000, add_faces=True)
+    mesh_no_color_upd = test_utils.get_random_mesh(2500, add_faces=True)
```
stephanwlee: `mesh_no_color_upd`: what does `upd` mean? Shortening variable for non-obvious case impairs readability.

tensorboard/plugins/mesh/mesh_plugin_test.py
```diff
@@ -150,34 +158,36 @@ def testTagsRoute(self):
     for name in self.names:
       self.assertIn(name, tags[self.runs[0]])
 
-  def testDataRoute(self):
-    """Tests that the /data route returns correct data for meshes."""
+  def validate_data_response(
+    self, run, tag, sample, content_type, dtype, ground_truth_data,
```
stephanwlee: +2 whitespace, I believe.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
```diff
@@ -167,11 +162,34 @@ Polymer({
...
+      })
+      .catch((error) => {
+        if (!error || !error.code || error.code != vz_mesh.ErrorCodes.CANCELLED) {
+          error = error || 'Response processing failed.';
+          throw new Error(error);
+        }
+      });;
     }
+
+    return dataPromise.then(function () {
```
stephanwlee: prefer to use an arrow function and remove the bind.
```ts
return dataPromise.then(() => {
   ...
});
```

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
```diff
@@ -167,11 +162,34 @@ Polymer({
...
+        currentStep, this.run, this.tag, this.sample, this._stepIndex)
+      .then(function (meshData) {
+        currentStep.mesh = meshData[0];
+      })
+      .catch((error) => {
+        if (!error || !error.code || error.code != vz_mesh.ErrorCodes.CANCELLED) {
+          error = error || 'Response processing failed.';
+          throw new Error(error);
+        }
+      });;
```
stephanwlee: Extraneous semi-colon.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
stephanwlee: Remove the `return`. If you want to keep it, please update the JSDoc.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
```diff
@@ -167,11 +162,34 @@ Polymer({
...
     }
+
+    return dataPromise.then(function () {
+      this._meshViewer.updateScene(currentStep, this);
+      if (!this._cameraPositionInitialized) {
+        this._meshViewer.resetView();
+        this._cameraPositionInitialized = true;
+      }
+      this.set('_isMeshLoading', false);
```
stephanwlee: IMO, this is quite confusing. So AFAICT, this change first fetches the `steps` and lazy loads the mesh of a `step` depending on what is currently active. There are two separate request/loading involved and `_updateScene` can get triggered multiple times when user uses the step slider to change the `currentStep`. In my opinion, there should be a two separate loading state potentially with a helpful text for users. Two loading states especially makes sense since user will move the slider and sees no UI change until the mesh is fetched. 

Separate refactor idea: I think `updateScene` should not have any fetching logic but simply update meshViewer when `currentStep.mesh` is present. Another observer can listen to the `currentStep` that fetches the mesh if it does not exist.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
```diff
@@ -167,11 +162,34 @@ Polymer({
    * @private
    */
   _updateScene: function(currentStep) {
-    this._meshViewer.updateScene(currentStep, this);
-    if (!this._cameraPositionInitialized) {
-      this._meshViewer.resetView();
-      this._cameraPositionInitialized = true;
+    let dataPromise = Promise.resolve();
```
stephanwlee: AFAICT, `currentStep` can change quite rapidly: setting slider aside, `reload` sets `steps` and `stepIndex` in two steps causing `currentStep` to change twice in rapid succession. Can you make sure we are not fetching the same mesh few times?

stephanwlee:
+reviewer: wchargin for Python readability.

tensorboard/plugins/mesh/summary.py
podlipensky: Good point

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
podlipensky: How can I guarantee order in  each observers are called and that data is fetched before `updateScene`?

tensorboard/plugins/mesh/summary.py
```diff
@@ -79,6 +84,16 @@ def _get_json_config(config_dict):
   return json_config
 
 
+def _get_components_bitmap(tensors):
+  """Creates bitmap for all existing components of the summary."""
```
wchargin: Please document the argument type, especially because the name is kind
of misleading: it’s called `tensors`, but it’s actually a list of pairs
that includes tensors, and you don’t actually care about the tensors.

tensorboard/plugins/mesh/test_utils.py
```diff
@@ -94,4 +94,4 @@ def deserialize_array_buffer_response(byte_content, data_type):
     Flat numpy array with the data.
   """
   return np.frombuffer(byte_content, dtype=data_type)
-  
\ No newline at end of file
+
```
wchargin: The proper fix here is to delete the empty line; please either do that
or just revert this change, because the file is otherwise unrelated.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
podlipensky: Also, loading indicator will be displayed when mesh is loading (as well as metadata), so I don't see why should we have two separate states. Especially considering that they will never overlap.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
podlipensky: Mmm, interesting behavior. But it is going to happen only once, when page is loaded, right? Should we complicate the logic for such small optimization?

stephanwlee: > But it is going to happen only once, when page is loaded, right?

It looks like this method gets triggered every time `currentStep` changes and fetch happens if no mesh is available for that step. `currentStep` can change when user uses the slider to change the `stepIndex`.

podlipensky: When you say `currentStep` can change when user uses the slider to change the `stepIndex`. Do you mean that `currentStep` object can be replaced?

tensorboard/plugins/mesh/plugin_data.proto
```diff
@@ -12,7 +12,7 @@ message MeshPluginData {
     COLOR = 3;
   }
 
-  // Version `0` is the only supported version.
+  // Version `1` is the only supported version.
```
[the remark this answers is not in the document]
podlipensky: Not really, the JS won't make any requests to the server if `components=0`

tensorboard/plugins/mesh/plugin_data.proto
```diff
@@ -26,4 +26,7 @@ message MeshPluginData {
 
   // Shape of underlying data. Cache it here for performance reasons.
   repeated int32 shape = 6;
+
+  // Bitmap of all existing components in addition to this one.
+  int32 components = 7;
```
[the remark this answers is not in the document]
podlipensky: Yes, uint is more appropriate type

tensorboard/plugins/mesh/plugin_data.proto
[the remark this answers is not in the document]
wchargin: Okay: then `parse_plugin_metadata` should be updated to translate from
the old format to the new format, if at all possible. Is it the case
that a version-0 summary should be interpreted as if its `components`
field were set to the union of all the content types? If so, then
consider something like:

```python
def get_components_bitmask(content_types):
  """<snip>"""
  result = 0
  for content_type in content_types:
    if content_type == plugin_data_pb2.MeshPluginData.ContentType.UNDEFINED:
      raise ValueError('Cannot include UNDEFINED content type in mask')
    result |= (1 << content_type)
  return result


_DEFAULT_COMPONENTS_MASK = get_components_bitmask([
    plugin_data_pb2.MeshPluginData.ContentType.VERTEX,
    plugin_data_pb2.MeshPluginData.ContentType.FACE,
    plugin_data_pb2.MeshPluginData.ContentType.COLOR,
])


def parse_plugin_metadata(content):
  """<snip>"""
  if not isinstance(content, bytes):
    raise TypeError('Content type must be bytes.')
  result = plugin_data_pb2.MeshPluginData.FromString(content)
  if not 0 <= result.version <= get_current_version():
    raise ValueError('Unknown metadata version: %s. The latest version known '
                     'to this build of TensorBoard is %s; perhaps a newer '
                     'build is available?' %
                     (result.version, get_current_version()))
  if result.version < 1:
    result.components = _DEFAULT_COMPONENTS_MASK
  result.version = get_current_version()
  return result
```

(This is the reason that the `parse_plugin_metadata` functions and the
`version` field exist in the first place: to provide a translation layer
for old data rather than just choking on it. We just announced the mesh
plugin as part of TensorBoard 1.14; we can’t immediately turn around and
drop data compatibility for the summaries that people have now written.)

Another way to solve this problem would be to change the `components`
field to a `missing_components` field representing a bitmask of the
elements that are _not_ included in the summary, such that the zero
value is useful. This would also have the benefit of saving 5 bytes per
summary. Your summaries would be backward-compatible for free, and you
wouldn’t need to bump the version. The two downsides that I see are that
(a) it’s slightly surprising, and (b) it makes it harder to add new
content types down the line.

tensorboard/plugins/mesh/plugin_data.proto
```diff
@@ -26,4 +26,7 @@ message MeshPluginData {
 
   // Shape of underlying data. Cache it here for performance reasons.
   repeated int32 shape = 6;
+
+  // Bitmap of all existing components in addition to this one.
```
wchargin: What does “in addition to this one” mean? In addition to this component?
Which component?

tensorboard/plugins/mesh/summary.py
```diff
@@ -79,6 +84,16 @@ def _get_json_config(config_dict):
...
+  components = 0
+  for tensor, content_type in tensors:
+    if tensor is None:
+      continue
+    components = components | (1 << content_type)
```
wchargin: Is `ContentType.UNDEFINED` a valid member of the mask? If so, what does
it represent/when might it be in the mask/what would the plugin do in
response? If not, any consideration for using `1 << (content_type - 1)`
instead, in part to make it obvious that `UNDEFINED` is not maskable?
(Just a suggestion; if you’ve considered this and disregarded it, no
problem, but I’d be interested to hear why.)

tensorboard/plugins/mesh/mesh_plugin_test.py
```diff
@@ -209,7 +220,9 @@ def testsEventsAlwaysSortedByWallTime(self):
       "PluginRunToTagToContent",
       return_value={"bar": {"foo": "".encode("utf-8")}},
   )
-  def testMetadataComputedOnce(self, run_to_tag_mock):
+  @mock.patch.object(metadata, "get_current_version", return_value=0)
```
wchargin: Please don’t patch the `get_current_version` return value here. IIUC,
fixing `parse_plugin_metadata` to properly handle old data should
obviate this line for free. If for some reason that is insufficient, the
right fix wsould be to change the mocked return value on line 210 from
`"".encode("utf-8")` (which should really just be written `b""`, but no
matter) to a properly encoded proto with the appropriate contents.

tensorboard/plugins/mesh/plugin_data.proto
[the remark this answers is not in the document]
podlipensky: I think interpreting 0 as "include all" makes more sense, since it is not possible to have 0 value after current PR. `UNDEFINED` is not being currently used.

tensorboard/plugins/mesh/plugin_data.proto
podlipensky: Yes, it means "current component"

tensorboard/plugins/mesh/summary.py
podlipensky: I'd just remove it in followup PR since it is not being used

tensorboard/plugins/mesh/metadata.py
```diff
@@ -18,15 +18,35 @@
...
 # `MeshPluginData` proto.
-_PROTO_VERSION = 1
+_PROTO_VERSION = 0
+
+
+def get_components_bitmask(tensors):
+  """Creates bitmask for all existing components of the summary.
+
+  Args:
+    tensors: list of MeshTensor tuples, representing all components
```
wchargin: Taking the whole `MeshTensor` when all you need is the `content_type`
seems unnecessary. The `tensor.data is None` effectively turns this
function from “reduction operation” to “reduction operation plus
filter”, which feels superfluous. If you change this to accept the
content types only, then the client in `summary.py` can simply write

```python
components = metadata.get_components_bitmask(
    tensor.content_type for tensor in tensors
    if tensor.data is not None
)
```

and the client in `parse_plugin_metadata` no longer has to construct
“fake” mesh tensors with empty dicts as `data` just to use this function
because the responsibilities have been more cleanly separated.

tensorboard/plugins/mesh/metadata_test.py
```diff
@@ -26,6 +26,13 @@
 from tensorboard.util import test_util
 
 
+class FakeResult(object):
```
wchargin: Why do you need a `FakeResult` class instead of just constructing the
actual metadata? You can change

```python
result = FakeResult(version=metadata.get_current_version(), components=0)
```

to

```python
result = plugin_data_pb2.MeshPluginData(
    version=metadata.get_current_version(),
    components=0,
)
```

and drop the class entirely.

tensorboard/plugins/mesh/metadata_test.py
```diff
@@ -93,6 +101,16 @@ def test_metadata_format(self):
     with six.assertRaisesRegex(self, TypeError, r'Content type must be bytes.'):
       metadata.parse_plugin_metadata(123)
 
+  def test_default_components(self):
+    """Tests that defult components are added when necessary."""
+    self._create_metadata()
+    result = FakeResult(metadata.get_current_version(), 0)
+    with patch.object(
+        plugin_data_pb2.MeshPluginData, 'FromString',
```
wchargin: No need to mock here at all—`parse_plugin_metadata` is a pure function;
you can just pass it the appropriate input:

```python
result = plugin_data_pb2.MeshPluginData(
    version=metadata.get_current_version(),
    components=0,
)
metadata.parse_plugin_metadata(result.SerializeToString())
```

tensorboard/plugins/mesh/metadata_test.py
```diff
@@ -75,9 +82,10 @@ def test_parse_plugin_metadata(self):
 
   def test_metadata_version(self):
     """Tests that only the latest version of metadata is supported."""
-    self._create_metadata()
-    # Change the version.
     with patch.object(metadata, 'get_current_version', return_value=100):
+      self._create_metadata()
+    # Change the version.
+    with patch.object(metadata, 'get_current_version', return_value=1):
```
wchargin: Why are we mocking this at all, much less to a return value that’s
larger than any known plugin version? Shouldn’t this cause the check on line 121 of `parse_plugin_metadata` to fail?

```python
  if not 0 <= result.version <= get_current_version():
    raise ValueError(...)
```

tensorboard/plugins/mesh/metadata_test.py
```diff
@@ -69,27 +82,40 @@ def test_parse_plugin_metadata(self):
...
+    self.assertEqual(self.components, parsed_metadata.components)
 
   def test_metadata_version(self):
     """Tests that only the latest version of metadata is supported."""
-    self._create_metadata()
-    # Change the version.
     with patch.object(metadata, 'get_current_version', return_value=100):
+      self._create_metadata()
+    # Change the version.
+    with patch.object(metadata, 'get_current_version', return_value=1):
```
wchargin: Given that this PR doesn’t actually bump the metadata version, we can
probably revert the changes to this test, yes?

tensorboard/plugins/mesh/metadata.py
```diff
@@ -18,15 +18,35 @@
 from __future__ import division
 from __future__ import print_function
 
+from collections import namedtuple
 from tensorboard.compat.proto import summary_pb2
 from tensorboard.plugins.mesh import plugin_data_pb2
 
 
+MeshTensor = namedtuple('MeshTensor', 'data content_type data_type')
```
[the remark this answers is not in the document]
podlipensky: It is direct association between `content_type` and `data_type` (i.e. TF type). I think it is better if mapping is done in single container and propagated everywhere as a whole.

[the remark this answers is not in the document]
podlipensky: The reason why we pass empty dict there is to simply be able reuse the same `get_components_bitmask` method. This method will only include components which has non-None data. In other cases data will be a Tensor.

tensorboard/plugins/mesh/plugin_data.proto
```diff
@@ -27,6 +27,6 @@ message MeshPluginData {
   // Shape of underlying data. Cache it here for performance reasons.
   repeated int32 shape = 6;
 
-  // Bitmask of all existing components in addition to this one.
+  // Bitmask of all existing components in addition current component.
```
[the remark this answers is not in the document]
podlipensky: It is all components, including current one. How would you phrase it then?

tensorboard/plugins/mesh/metadata.py
podlipensky: good idea

tensorboard/plugins/mesh/metadata_test.py
podlipensky: done

tensorboard/plugins/mesh/metadata_test.py
podlipensky: But it changed the logic of version validation, so I think this test is still valid.

podlipensky:
FYI, updated SHA for TF Graphics-hosted files, so you can test the whole thing if needed. Please let me know if there is anything else is needed to merge the change.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
```diff
@@ -163,15 +159,54 @@ Polymer({
...
+
+  _debouncedFetchMesh() {
+    this.debounce('fetchMesh', () => this._maybeFetchMesh(), 100);
+  },
+
+  async _maybeFetchMesh() {
+    const currentStep = this._currentStep;
+    if (currentStep.mesh || currentStep.meshFetching) return;
+    currentStep.meshFetching = true;
+    this._isMeshLoading = true;
```
stephanwlee: Just letting you know: do note that this can lead to incorrect UI state. 
Test case:
1. go to the network tab in Chrome and enable throttling to emulate slow internet
2. move the step slider with fervor

Multiple requests should be in flight at the same time and `this._isMeshLoading` will be `false` the moment one of them resolve the first even while others are in-flight.

tensorboard/plugins/mesh/summary.py
```diff
@@ -157,27 +165,35 @@ def pb(name,
...
-             (faces, plugin_data_pb2.MeshPluginData.FACE, tf.int32),
-             (colors, plugin_data_pb2.MeshPluginData.COLOR, tf.uint8)]
-  for tensor, content_type, data_type in tensors:
-    if tensor is None:
-      continue
-    tensor_shape = tensor.shape
-    tensor = tf.compat.v1.make_tensor_proto(tensor, dtype=data_type)
+  tensors = [
+      metadata.MeshTensor(
+        vertices, plugin_data_pb2.MeshPluginData.VERTEX, tf.float32),
```
stephanwlee: nit: +2 space

tensorboard/plugins/mesh/summary.py
```diff
@@ -157,27 +165,35 @@ def pb(name,
...
+      metadata.MeshTensor(
+        colors, plugin_data_pb2.MeshPluginData.COLOR, tf.uint8)
+  ]
+  tensors = [tensor for tensor in tensors if tensor.data is not None]
+  components = metadata.get_components_bitmask(tensors)
+  for tensor in tensors:
+    shape = tensor.data.shape
+    shape = [dim if dim is not None else -1 for dim in shape]
+    tensor_proto = tf.compat.v1.make_tensor_proto(
+      tensor.data, dtype=tensor.data_type)
```
stephanwlee: nit: +2 space

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-loader.js
podlipensky: Thanks, created ticket.

tensorboard/plugins/mesh/mesh_plugin.py
```diff
@@ -194,8 +200,12 @@ def _collect_tensor_events(self, request):
       meta = self._instance_tag_to_metadata[instance_tag]
       tensor_events += [(meta, tensor) for tensor in tensors]
 
-    # Make sure tensors sorted by timestamp in ascending order.
-    tensor_events = sorted(
+    if timestamp is not None:
+      tensor_events = [
+        event for event in tensor_events if event[1].wall_time == timestamp]
```
wchargin: Why are we filtering for timestamp equality here? The conventional way
to select a single step is to filter by step, not timestamp. Especially
given that mesh data is emitted as up to three separate summaries, it
seems plausible that the different components for a single step might
have been written out at different times.

tensorboard/plugins/mesh/mesh_plugin.py
```diff
@@ -194,8 +200,12 @@ def _collect_tensor_events(self, request):
...
+    else:
+      # Make sure tensors sorted by timestamp in ascending order.
+      tensor_events = sorted(
```
wchargin: Likewise, we usually sort by step (the step slider should be monotonic),
unless something special is going on here?

tensorboard/plugins/mesh/metadata.py
```diff
@@ -18,17 +18,36 @@
 from __future__ import division
 from __future__ import print_function
 
+import collections
 from tensorboard.compat.proto import summary_pb2
 from tensorboard.plugins.mesh import plugin_data_pb2
 
 
+MeshTensor = collections.namedtuple(
+  'MeshTensor', ('data', 'content_type', 'data_type'))
```
wchargin: nit: +2 spaces of indentation

tensorboard/plugins/mesh/http_api.md
```diff
@@ -49,23 +49,26 @@ Here is an example:
             "data_shape": [1, 17192, 3],
             "step": 0,
             "content_type": 2,
-            "wall_time": 1556678491.836787
+            "wall_time": 1556678491.836787,
+            "components": 14
```
wchargin: Please document this—a short sentence would suffice, but right now it
looks like this means that the summary has 14 components, which is not
correct.

tensorboard/plugins/mesh/plugin_data.proto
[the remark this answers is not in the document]
wchargin: This is fine; I understand now.

tensorboard/plugins/mesh/mesh_plugin.py
podlipensky: I thought the timestamp applies to summaries at the same step, no? Otherwise how do you organize summaries in steps?

wchargin: Not sure what you mean: `wall_time` and `step` are [two separate fields
on a proto that’s written to disk][1]; we don’t backfill steps at
runtime by grouping timestamps.

We use `step` as a key rather than `wall_time` because (among other
things) it’s an integer rather than an arbitrary floating-point number.

[1]: https://github.com/tensorflow/tensorboard/blob/master/tensorboard/compat/proto/event.proto#L14-L19

tensorboard/plugins/mesh/summary.py
```diff
@@ -122,7 +122,9 @@ def op(name, vertices, faces=None, colors=None, display_name=None,
         colors, plugin_data_pb2.MeshPluginData.COLOR, tf.uint8)
   ]
   tensors = [tensor for tensor in tensors if tensor.data is not None]
-  components = metadata.get_components_bitmask(tensors)
+
+  components = metadata.get_components_bitmask([
+    tensor.content_type for tensor in tensors])
```
wchargin: nit: +2 spaces indentation for continuations, please.

tensorboard/plugins/mesh/summary.py
```diff
@@ -173,7 +175,8 @@ def pb(name,
           colors, plugin_data_pb2.MeshPluginData.COLOR, tf.uint8)
   ]
   tensors = [tensor for tensor in tensors if tensor.data is not None]
-  components = metadata.get_components_bitmask(tensors)
+  components = metadata.get_components_bitmask([
+    tensor.content_type for tensor in tensors])
```
wchargin: nit: +2 spaces indentation for continuations, please.

PR #2382, opened 2019-06-24
cleanup: express explicit dependency on threejs

stephanwlee:
tf_mesh_dashboard did not explicitly define dependency on threejs and
implicitly relying on other plugins loading the threejs before its
resource is loaded on the Vulcanized HTML binary. Now that projector,
another user of the threejs, is now dynamically loaded, this pattern
becomes a lot more brittle.

WANT_LGTM=any
This unblocks internal sync.


nfelt:
cc @podlipensky

PR #2384, opened 2019-06-25
add "Rescale all plots" button to global actions

linzwatt:
rescales all the scalar graphs on click

as models train and data is logged, I have to constantly click the rescale button for each plot, usually 10+ plots for training and validation metrics. This is annoying and takes too long, the plots should be able to rescale together with one click.

Added a button to the global actions panel top right, on click it looks for every rescale button and clicks it
Changes were made only to `components/tf_tensorboard/tf-tensorboard.html`

* Screenshots of UI changes
[https://imgur.com/a/dQIN3Xo](https://imgur.com/a/dQIN3Xo)

Added the change, built using bazel and tf-nightly, ran tb, it worked well

Somebody with more knowledge of the repo could probably use a more elegant technique for activating the rescale of each plot, I am not sure where the reference to each plots object/datastructure lives, to call it's rescale function

### DIFF
--- a/tensorboard/components/tf_tensorboard/tf-tensorboard.html
+++ b/tensorboard/components/tf_tensorboard/tf-tensorboard.html
@@ -118,6 +118,12 @@ limitations under the License.
             </template>
           </template>
           <div class="global-actions">
+              <paper-icon-button
+              id="rescale-all-button"
+              icon="settings-overscan"
+              on-tap="rescale_all"
+              title="Rescale all plots"
+            ></paper-icon-button>
             <paper-icon-button
               id="reload-button"
               class$="[[_getDataRefreshingClass(_refreshing)]]"
@@ -1001,6 +1007,11 @@ limitations under the License.
           if (dashboard && dashboard.reload) dashboard.reload();
         });
         this._lastReloadTime = new Date().toString();
+      },
+
+      rescale_all() {
+        let rescale_buttons = document.querySelectorAll('paper-icon-button[icon="settings-overscan"]');
+        rescale_buttons.forEach((button) => { button.click() });
       },
 
       _reloadData() {


linzwatt:
I signed the cla, sorry for not doing this before submitting

linzwatt:
Is this something that can be merged in?
if not, how do I build an install-able pip package from my fork of the tb repo?

linzwatt:
Thanks for updating me, I'll wait for the Polymer update, and update this PR after that

wchargin:
Okay! Thank you for your patience. The Polymer 2 upgrade (#2392) has
been merged, so you should be good to merge those in and make any
changes necessary due to the Polymer change. In particular, I expect
that the call to `querySelectorAll` will need to change.

We’re also wondering whether a button is the best UX for this. It seems
like the usual case is that you’d want this to happen automatically on
every refresh, except in cases where you’ve manually adjusted the chart
domain to zoom in on some portion. Are you up for making those two
changes? Wiring it up to the reload cycle shouldn’t be too tricky, I
think, by hooking into the `reload` method; omitting charts that have
been manually adjusted might require adding a bit of logic to
`vz_line_chart2/line-chart.ts`, or maybe you’ll find a different way.

linzwatt:
Sorry I didn't see your most recent comment before that last commit.

I think you're right, a rescale on every refresh makes sense, for all plots that haven't been adjusted manually, and once you hit the fit domain button on an adjusted plot, it will rescale with the other ones on next refresh

I'll look into making the changes, thanks for your advice

carlogrisetti:
I was thinking about making a PR just like this myself, but then found there was this one already. I am in the same condition of having to costantly click around on all the graphs (sometimes 20+) to adjust the domain to fit all the data, and this would be very nice.
Both the "manual" button (and I agree with @wchargin: it's not the right graphic for this, I would have used the same graphic for the single graphs, maybe placed on the rightmost part of the "Filter Tags" searchbar, with a different tooltip), and the automatic domain adjustment in each refresh are good solutions to me, with the latter being the most desirable one.

@linzwatt are you still able to work on this one and push this to the end? If not I would like to take a shot at this, also using your code as a baseline if necessary.

Thanks!

carlogrisetti:
Oh, I almost forgot... I think the "manual" button is always useful, since I often find myself adding and removing different runs in the comparison, and they most often have different run duration (both in steps and relative time), and they surely have a different extension in the Y-axis.
Adding and removing runs would result in graphs being suboptimally sized, and without a "manual" button (like the one initially proposed in this PR) the only solution would be to force a data refresh, which is not the best approach IMHO.

linzwatt:
@carlogrisetti feel free to take over this PR! I'm not very good at web stuff, especially in this framework with TS, and I don't really have time to learn at the moment unfortunately. Keen to see TB become a bit more user friendly when you have it loaded up with 10 or 20 plots!

PR #2388, opened 2019-06-26
core: error msg for dynamic plugin in Colab

stephanwlee:
Colab is iterating on mechanism to use iframe within an output frame but
it is not yet ready for consumption. As a result, all usage of iframes
do not work properly in Colab and we would like to communicate that to
users.

image

After merging this PR, we will update the issue #1913 accordingly.

### DIFF
--- a/tensorboard/components/tf_tensorboard/tf-tensorboard.html
+++ b/tensorboard/components/tf_tensorboard/tf-tensorboard.html
@@ -365,6 +365,20 @@ limitations under the License.
         }
       }
     </style>
+    <template id="colabNotSupported" preserve-content>
+      <div style="max-width: 540px; margin: 80px auto 0 auto;">
+        <h3>Dynamic plugin isn’t supported in Colab yet.</h3>
+        <p>
+          Please see
+          <a
+            href="https://github.com/tensorflow/tensorboard/issues/1913"
+            rel="noopener"
+            target="_blank"
+          >GitHub issue #1913</a>
+          for more information.
+        </p>
+      </div>
+    </template>
   </template>
   <script src="autoReloadBehavior.js"></script>
   <script>
@@ -581,7 +595,12 @@ limitations under the License.
         _refreshing: {
           type: Boolean,
           value: false,
-        }
+        },
+        _inColab: {
+          type: Boolean,
+          value: () => !!(window.TENSORBOARD_ENV || {}).IN_COLAB,
+          readOnly: true,
+        },
       },
       observers: [
         ('_updateSelectedDashboardFromActive(' +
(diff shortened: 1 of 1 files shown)


tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -788,6 +789,36 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
         this.set('_isReloadDisabled', dashboard.disableReload);
       },
 
+      _renderPluginIFrame(container, selectedDashboard, loadingMechanism) {
```
wchargin: caps: `Iframe`, per the [standard camel case definition][1]

tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -365,6 +365,20 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
         }
       }
     </style>
+    <template id="colabNotSupported" preserve-content>
+      <div style="max-width: 540px; margin: 80px auto 0 auto;">
+        <h3>Dynamic plugin isn’t supported in Colab yet.</h3>
```
wchargin: caps: `Iframe`, per the [standard camel case definition][1]

tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -365,6 +365,20 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
...
+        <p>
+          Please see
+          <a
+            href="https://github.com/tensorflow/tensorboard/issues/1913"
```
wchargin: Do you plan to broaden this issue from the profile plugin to “plugins
that use iframes” or similar? I’d be confused if I happened upon this
issue from the projector plugin.

tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -788,6 +789,36 @@ <h3>There’s no dashboard by the name of “<tt>[[_selectedDashboard]]</tt>.”
         this.set('_isReloadDisabled', dashboard.disableReload);
       },
 
+      _renderPluginIFrame(container, selectedDashboard, loadingMechanism) {
+        if (this._inColab) {
+          const templateContent = this.$.colabNotSupported.content;
+          const node = document.importNode(templateContent, true);
+          this.scopeSubtree(node);
+          container.appendChild(node);
+          return;
```
wchargin: In this code path, we don’t set `x.id = 'dashboard';` for any `x`; just
want to confirm that that’s okay and doesn’t cause anything to crash
(e.g., reloading)?

stephanwlee: Added the id. Btw, the `id = 'dashboard'` is not effectively useful since we are not rendering an instance of a plugin and this conditional will never be truthy: https://github.com/tensorflow/tensorboard/blob/ff05d1af8d118d0add237766980d31163a94c40b/tensorboard/components/tf_tensorboard/tf-tensorboard.html#L1000-L1001.

Regardless, I added the `id` so future readers don't have to remember/understand every flow.

tensorboard/components/tf_tensorboard/tf-tensorboard.html
stephanwlee: I have a feeling that this is a glitch with GitHub or something. Did you mean to leave this comment on this line?

wchargin: Hmm; nope. Odd.

tensorboard/components/tf_tensorboard/tf-tensorboard.html
wchargin: Got it; thanks!

PR #2390, opened 2019-06-27
core: add CSP header for Colab output frames

wchargin:
Summary:
This is required to use the new `google.colab.kernel.proxyPort` Colab
feature (see Google-internal <http://b/130310433>).

Test Plan:
The included tests verify that the mechanism is implemented as intended.
To verify that the mechanism actually works in Colab, create a Colab
notebook with the following cells:

```python
import werkzeug

@werkzeug.Request.application
def app(request):
  frame_ancestors = request.args.get("frame_ancestors")
  response = werkzeug.Response("Hello, %s!\n" % (frame_ancestors,))
  response.headers["Content-Type"] = "text/html"
  if frame_ancestors is not None:
    response.headers["Content-Security-Policy"] = (
        "frame-ancestors %s" % frame_ancestors
    )
  return response
```

```python
import threading
import werkzeug.serving

if "server" in locals():
  server.shutdown()
server = werkzeug.serving.ThreadedWSGIServer("localhost", 2345, app)
threading.Thread(target=server.serve_forever).start()
```

```
!curl -i localhost:2345?frame_ancestors=foo
```

```javascript
%%javascript
google.colab.kernel.proxyPort(2345).then((base) => {
  const ancestors = "https://*.googleusercontent.com https://*.google.com";
  const url = new URL(
      "?frame_ancestors=" + encodeURIComponent(ancestors),
      base,
  );
  const iframe = document.createElement("iframe");
  iframe.src = url.toString();
  document.body.appendChild(iframe);
});
```

Run the notebook and verify that the final output frame renders
properly:

![Screenshot of output frame with intended “Hello” message][1]

Note that when changing the `iframe.src` to just `base`, the iframe
instead renders a “sad page”, and that a console error indicates that
the culprit is `X-Frame-Options: sameorigin`.

[1]: https://user-images.githubusercontent.com/4317806/60227895-cf530280-9845-11e9-93f0-cc5159b88e31.png

wchargin-branch: colab-csp

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -24,6 +24,7 @@ from __future__ import print_function
 
 import atexit
 import collections
+import functools
 import json
 import os
 import re
@@ -36,6 +37,7 @@ import time
 import six
 from six.moves.urllib import parse as urlparse  # pylint: disable=wrong-import-order
 
+import werkzeug
 from werkzeug import wrappers
 
 from tensorboard.backend import http_util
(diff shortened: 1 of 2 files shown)


tensorboard/backend/application.py
```diff
@@ -344,13 +373,17 @@ class are WSGI applications.
     parsed_url = urlparse.urlparse(request.path)
     clean_path = _clean_path(parsed_url.path, self._path_prefix)
 
+    @functools.wraps(start_response)
+    def new_start_response(status, headers):
```
nfelt: Was there a reason not to do this in `http_util.Respond()`?  That's where the other automatically populated headers are set: https://github.com/tensorflow/tensorboard/blob/63ab35a5a8e90a1e94d13b349a7ea56db8c748f4/tensorboard/backend/http_util.py#L148-L158

wchargin: Stephan and I talked about this briefly and thought that it was
reasonable to do in middleware. IMHO, we should consider moving most of
`Respond`’s functionality into middleware—our plugin API centers around
WSGI apps, which aren’t tied to Werkzeug or its `Request.application`
wrapper at all, so it’s strange to have this additional integration
point. What if plugin developers prefer Flask?

The alternative would be to make `http_util` a public module or move
`Request` into `plugin_util`.

nfelt: That's a fair point, but I guess it seems like then we should have a TODO/issue open to move that logic into the middleware, so there's a rationale available for why otherwise similar code is going in two different places.

PR #2394, opened 2019-06-27
build: add `:extract_pip_package` for convenience

wchargin:
Summary:
For ad hoc testing, one frequently wants to install a Pip package after
buliding it. A new script extracts the wheels and prints the path to the
enclosing directory (for `xdg-open` to drag-and-drop into Colab) and to
the wheels themselves (to copy-and-paste as a `pip install` argument).

Test Plan:
Run the new binary:

```
$ bazel run //tensorboard/pip_package:extract_pip_package 2>/dev/null
/tmp/tmp.caSTM9Sa7J
/tmp/tmp.caSTM9Sa7J/tensorboard-1.15.0a0-py2-none-any.whl
/tmp/tmp.caSTM9Sa7J/tensorboard-1.15.0a0-py3-none-any.whl
```

wchargin-branch: extract-pip-package

### DIFF
--- a/tensorboard/pip_package/extract_pip_package.sh
+++ b/tensorboard/pip_package/extract_pip_package.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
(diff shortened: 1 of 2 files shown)


tensorboard/pip_package/extract_pip_package.sh
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+tmpdir="$(mktemp -d)"
+tar xzf "$0.runfiles/org_tensorflow_tensorboard/tensorboard/pip_package/pip_packages.tar.gz" \
+    -C "${tmpdir}"
+find "${tmpdir}" | sort
```
nfelt: nit: `-type f`?  Printing the containing directory doesn't seem that useful.

wchargin: I originally had `-type f`, but Colab has a drag-and-drop file upload
interface, so my preferred workflow is: run `:extract_pip_package`,
triple-click the directory name, run `xdg-open <Paste>`, then drag the
appropriate wheel onto the notebook. (Clicking the “Upload” button and
pasting the path in directly requires using a modal dialog, which I’d
rather not do.)

PR #2396, opened 2019-06-27
test: unbreak all the tests

stephanwlee:
Applied the change in [1] to all tests and standalone apps.

We need to make another minor changes on hparams to account
for uninitialized properties in computed callback (in Polymer 2, it
no longer guarantees that all props are initialized before invoking
the callback). 

After this change `bazel test tensorboard/....` passes again.

Integration plan: squash all commits except changes to vz_projector (6f429e6d) then rebase.

[1]: https://github.com/tensorflow/tensorboard/blob/213212101309875489cbcab82dcbd183721c8bfc/tensorboard/components/tensorboard.html#L25-L36


tensorboard/components/tf_tensorboard/test/tensorboardTests.html
```diff
@@ -19,7 +19,7 @@
 <html>
 <head>
   <link rel="import" href="../../polymer/polymer.html">
-  <script src="../../webcomponentsjs/webcomponents-lite.js"></script>
+  <script jscomp-nocmpile src="../../webcomponentsjs/webcomponents-lite.js"></script>
```
wchargin: typo: `nocmpile` → `nocompile`; did automated tests not catch this?

tensorboard/plugins/projector/vz_projector/standalone.html
```diff
@@ -51,6 +48,9 @@
   </style>
 </head>
 <body>
+  <script>
+    <!-- HaCK: GA requires at least one script tag on document.  -->
```
wchargin: nit: is this capitalization intended?

stephanwlee: Nope but it looks cool (and hackerish!). Fixed it.

tensorboard/components/tf_tensorboard/test/tensorboardTests.html
stephanwlee: Nope :(. Fixed it manually for now.

stephanwlee:
> We already have a //tensorboard/components/tf_imports:polymer layer of
indirection. Does this not afford sufficient flexibility to fix clients
in one spot? I imagined that we could do something like

I took a different approach for now. Do note that, unlike other libs, we were really using tf_imports:polymer as a layer of indirection. It was merely there to address typing issue
for TypeScript files. When others imported `./polymer/polymer.html`, they were referring
to this specific file: https://github.com/tensorflow/tensorboard/blob/373eb09e4c5d2b3cc2493f0949dc4be6b6a45e81/third_party/polymer.bzl#L33

PR #2397, opened 2019-06-27
Revert "core: add CSP header for Colab output frames (#2390)"

wchargin:
Summary:
This commit broke Jupyter notebook integration: a content security
policy that whitelists Colab also blocks access by any Jupyter
notebooks.

I don’t see a simple fix for this other than permitting all frame
ancestors: Jupyter origins are arbitrary, and we can’t use `'self'`
because Jupyter and TensorBoard are not in fact served from the same
origin.

This reverts commit ff05d1af8d118d0add237766980d31163a94c40b.

Test Plan:
Build the Pip package, then install it into a virtualenv with Jupyter.
From a Jupyter notebook, run `%tensorboard --logdir logs`. Before this
commit, that rendered an empty cell and printed a console error,

> Refused to display 'http://localhost:6006/' in a frame because an
> ancestor violates the following Content Security Policy directive:
> "frame-ancestors https://*.googleusercontent.com
> https://*.google.com".

As of this commit, this renders a working TensorBoard.

wchargin-branch: revert-colab-csp

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -24,7 +24,6 @@ from __future__ import print_function
 
 import atexit
 import collections
-import functools
 import json
 import os
 import re
@@ -37,7 +36,6 @@ import time
 import six
 from six.moves.urllib import parse as urlparse  # pylint: disable=wrong-import-order
 
-import werkzeug
 from werkzeug import wrappers
 
 from tensorboard.backend import http_util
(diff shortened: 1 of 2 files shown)


wchargin:
Yeah. The secure way to do it probably involves some sort of nonce
mechanism provided as a command-line flag and required as a query
parameter. But getting that right may not be trivial (both because of
security and because of the flag parsing).

nfelt:
Could we put the behavior you added behind a flag, and in the tensorboard.manager API, detect if we're running in Colab and pass that flag only in that case?  We could potentially do something similar for Jupyter so that only TB running in that notebook context has the permissive CSP.

wchargin:
> Could we put the behavior you added behind a flag, and in the
> tensorboard.manager API, detect if we're running in Colab and pass
> that flag only in that case?

A simpler solution is to just add a `frame-ancestors *` CSP to all
responses. Note that this is _equivalent_ to the current behavior: the
CSP added in #2390 is a restriction, not a relaxation, and is only
necessary because the new Colab `proxyPort` middleware splices in an
even more restrictive header, which we want to override.

That said—yes, we could implement your proposal as an additional
security feature. It’s not entirely trivial. We’ve tried to avoid
splicing new args into the user’s input, partially because then we have
to make sure that the args that we emit are parsed correctly, but I
think that inserting them at the beginning of the command line should be
safe.

The other reason that we’ve avoided changing the command line is that it
was a goal that `notebook.display(port=6006)` should work even if you
started TensorBoard other than via `%tensorboard`. Now, users who
manually start TensorBoard will need to specify this flag, too. Maybe
that’s acceptable.

And, finally, this won’t automatically work for Jupyter kernels behind a
lab-internal Nginx proxy or university router. Maybe that’s acceptable,
too.

As far as the API goes, I’d be fine with

```
--frame_ancestors self  # default; sets CSP "'self'" (notes single quotes)
--frame_ancestors colab  # sets CSP with *.guc and *.google.com
--frame_ancestors custom=http://localhost:8888  # any policy; works for Jupyter
```

nfelt:
Okay.  So what exactly is the reasoning not to just add `frame-ancestors *` in that case?  Given that we've in general kept TensorBoard read-only, it doesn't seem like there's the security concern around embedding e.g. an iframe that coaxes a user to click a button.

PR #2406, opened 2019-07-02
build: removed bin/genfile prefix in declare_file

stephanwlee:
Commit 4ba5f51b6d3beb0746ce21d355cef29d918d021d changed ctx.new_file to
ctx.actions.declare_file which is the new way of declaring a file except
the new API behaves a bit differently than the old one.

new_file, as the first argument, took the root directory [1] and
previous to the commit, we passed the bazel genfile path. However,
declare_file no longer take the root path [2] (it generates a new root
path based on whether the rule is genrule or not [3]) and appended the
first argument to the root it generates.

This made TensorBoard have paths like below:
- bazel-out/k8-fastbuild/bin/t/c/tf_backend/bazel-out/k8-fastbuild/bin/backend.js
- bazel-out/k8-fastbuild/bin/t/c/bazel-out/k8-fastbuild/bin/index.pbtxt
- bazel-out/k8-fastbuild/bin/t/c/tf_backend/bazel-out/k8-fastbuild/bin/tf_backend-tsc.json

With the change, above look like below:
- bazel-out/k8-fastbuild/bin/t/c/tf_backend/backend.js
- bazel-out/k8-fastbuild/bin/t/c/index.pbtxt
- bazel-out/k8-fastbuild/bin/t/c/tf_backend/tf_backend-tsc.json

[1]:
https://github.com/bazelbuild/bazel/blob/2a116372382aedbf8a39477e0ea1123903ab5479/src/main/java/com/google/devtools/build/lib/analysis/skylark/SkylarkRuleContext.java#L759-L763
[2]: https://github.com/bazelbuild/bazel/blob/1ace20a6bb7587ae427d5917ffe13fa71d8885e9/src/main/java/com/google/devtools/build/lib/analysis/skylark/SkylarkActionFactory.java#L133
[3]: https://github.com/bazelbuild/bazel/blob/2a116372382aedbf8a39477e0ea1123903ab5479/src/main/java/com/google/devtools/build/lib/analysis/RuleContext.java#L621-L623

Looked for every `declare_file` instance and made sure we did not make the mistake.

Related: there were confusions around the difference between bazel-genfiles vs. bazel-bin.
As of 0.25, the Bazel team had merged two concept as there is no tangible difference and
only adds confusion.
https://groups.google.com/forum/#!msg/bazel-dev/5iq7n1soenA/xGM_M1NOCAAJ

### DIFF
--- a/tensorboard/defs/vulcanize.bzl
+++ b/tensorboard/defs/vulcanize.bzl
@@ -14,8 +14,6 @@
 
 """Rule for building the HTML binary using Closure Compiler."""
 
-load("@bazel_skylib//lib:paths.bzl", "paths")
-
 load("@io_bazel_rules_closure//closure:defs.bzl", "closure_js_aspect")
 load("@io_bazel_rules_closure//closure/private:defs.bzl", "collect_js", "unfurl", "long_path")
 
@@ -64,8 +62,7 @@ def _tensorboard_html_binary(ctx):
   manifest_srcs = [struct(path=ctx.outputs.html.path,
                           longpath=long_path(ctx, ctx.outputs.html),
                           webpath=ctx.attr.output_path)]
-  manifest = ctx.actions.declare_file(paths.join(
-      ctx.configuration.bin_dir.path, "%s.pbtxt" % ctx.label.name))
+  manifest = ctx.actions.declare_file("%s.pbtxt" % ctx.label.name)
   ctx.actions.write(
       output=manifest,
       content=struct(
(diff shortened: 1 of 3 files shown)


stephanwlee:
Yup, you are right. I was playing around with the rest to see if we were getting the same problem with the `ctx.configuration.bin_dir.path` and forgot to commit the changes. 

Added those changes back and confirmed that the app still works as expected.

wchargin:
thanks for the patch! (and sorry for introducing this problem :-) )

stephanwlee:
No problem! I've made tons of worse problems in the past and, in the PR, your change did look innocuous

Issue #1956, opened 2019-03-05
%tensorboard magic: host other than localhost


wookayin:
For bug reports, please include the following:

 - TensorBoard version (from pip package, also printed out when running `tensorboard`) : 1.13.0
 - OS Platform and version (e.g., Linux Ubuntu 16.04) : Linux, but applies to ALL distributions
 - Python version (e.g. 2.7, 3.5) : 3.6
 
The recently introduced `%tensorboard` magic is really cool! But it has some certain limitations: the iframe host is set to be `localhost` only. In many cases where one is running juypter/tensorboard in a remote server, this won't be simply working. Of course, the iframe will say `connection refused`.

See this line 
https://github.com/tensorflow/tensorboard/blob/master/tensorboard/notebook.py#L360 :  the hostname is hard-coded.

An easy fix to this would be: make host be configurable (as a parameter) and have the `--host` parameter, for example, from the `%tensorboard` magic propagated. What do you think about this change? It would be an easy fix, so once we reach to an agreement I could submit a patch.

wchargin:
Yes, the host is fixed. We discussed this as part of the original
design, and didn’t assign it a high priority because the initial goal
was to have solid support for TensorBoard in Colab—in Colab, a service
worker proxy forwards requests to `https://localhost:*/` to the actual
backing VM, hence this design.

TensorBoard already accepts a `--host` command-line argument, which
controls the interfaces to which the server will attempt to bind.
(Namely: it has _different_ semantics from what you seem to want.)
Furthermore, reading an argument value to determine how to render
TensorBoard in the UI is not actually an easy patch: the arguments
provided by the user are merely lexed by the notebook extension; only
the underlying tensorboard(1) binary actually parses them. Probably the
right thing to do _if_ you wanted to add a new flag would be to thread
the information through the TensorBoard info file, but I don’t think
that adding a new flag is the way to go.

If we wanted to add support for this, my initial thoughts would be as
follows. We should change `tensorboard.notebook.display` to accept
kwargs `host` and `protocol` that control the origin of the iframe,
defaulting to `"localhost"` and `"http"`, respectively. To facilitate
the simpler `%tensorboard` magic interface, we could let these
parameters have default values taken from module globals set by a new
function `tensorboard.notebook.configure`. Then a typical session might
look like:

```python
%load_ext tensorboard.notebook

from tensorboard import notebook
notebook.configure(host="wookayin-tensorboard.appspot.com")

[...]

Issue #260, opened 2017-07-24
SSL support


anfeng:
Currently, tensorboard is a simple HTTP server. Any plan to support SSL?

chihuahua:
TensorBoard actually [uses werkzeug to start a server that follows the WSGI spec](https://github.com/tensorflow/tensorboard/blob/530310dd26f8f14df7a49b6d8f286c3d63627e0d/tensorboard/main.py#L171). werkzeug accepts an [ssl_context](http://werkzeug.pocoo.org/docs/0.12/serving/).

It seems like a TensorBoard server could be served off of HTTPS. We could expose flags for that if there are compelling use cases. Thank you!

wchargin:
@jart do you think that this is something that an nginx proxy should handle, or should we consider incorporating it into core?

anfeng:
I see at least 2 reasons to support SSL:

1. TensorBoard could contain sensitive data (ex. Tensor or Image).
2. TensorBoard may be exposed as IFrame within some HTTPS page (ex. notebook). 

As @chihuahua mentioned, a quick fix could be based on WSGI spec.

wchargin:
@anfeng As far as I understand, both of those use cases would be handled by running TensorBoard behind an nginx proxy, correct? (I'm imagining that TensorBoard serves only to localhost, and nginx exposes a proxy to the LAN or WAN.)

I'm not saying that we shouldn't support SSL, but it's worth asking whether this is something best left to a different layer of the stack. nginx usually proves to be quite good at things like this.

anfeng:
@wchargin I am not running nginx in my environment. Temporarily, one could have SSL terminated at proxy layer. That will be considered not ideal for my security guys.

chihuahua:
@anfeng, based on #267, it seems like you or your org would actually benefit from support for SSL?

I think @wchargin has a point. Apps that follow WSGI are often deployed on nginx and/or passenger (which also handles say load balancing and managing the thread pool). Or in environments like App Engine. SSL is then readily handled by other parts of the stack.

At the same time, the major python frameworks (flask, django, etc) offer SSL support. They also try to function as self-standing servers (that say manage multiple threads), even though other parts of stack can handle that for them (for instance, App Engine may chose to start multiple instances of the entire app). Hence, I think supporting SSL within TensorBoard may suffice for a large number of cases (especially if the user base is relatively small, ie a little company, research team, or college course).

jart:
[...]

PR #2407, opened 2019-07-03
Allow Jupyter+TensorBoard to work w/ hostname other than localhost

cliffwoolley:
Fixes https://github.com/tensorflow/tensorboard/issues/1956 to allow the combination of Jupyter+TensorBoard to work when the browser is running on a system other than TensorBoard's localhost .

Rather than `localhost` being hardcoded, we use a short JavaScript snippet in the IFrame source to have the client determine for itself the hostname that should be contacted.  This allows the solution also to work from inside a container, where we might not know the host's outward-facing IP address.

Test Plan:
Install Docker and `nvidia-docker`, then run

```sh
tmpdir="$(mktemp -d)" \
&& bazel run //tensorboard/pip_package:extract_pip_package "${tmpdir}" \
&& nvidia-docker run -it --rm -v "${tmpdir}":/tensorboard \
    -p 8888:8888 -p 6006:6006 nvcr.io/nvidia/tensorflow:19.03-py3
```

to get a Docker shell. Inside the container, run

```sh
pip install -qU /tensorboard/*py3* \
&& jupyter notebook --allow-root --ip 0.0.0.0
```

and navigate to the provided notebook URL via a hostname other than
`localhost` (e.g., your machine’s `hostname`). Create a new IPython
notebook and execute the cell

```
%load_ext tensorboard
%tensorboard --logdir /tmp/whatever
```

and verify that the output contains a working TensorBoard iframe whose
`src` attribute is under the main frame’s origin (e.g., your machine’s
`hostname`), not `localhost`.


tensorboard/notebook.py
```diff
@@ -395,11 +395,15 @@ def _display_colab(port, height, display_handle):
...
 def _display_ipython(port, height, display_handle):
   import IPython.display
-  iframe = IPython.display.IFrame(
-      src="http://localhost:%d" % port,
-      height=height,
-      width="100%",
-  )
+  shell = """
+  <iframe
+    src="javascript:window.location='http://' + parent.location.hostname + ':%PORT%/';"
```
wchargin: This use of a JavaScript URI for the `src` attribute is fairly
surprising/confusing, and also generates a flash of bad content as the
JS expression evaluates to a string. Can we instead explicitly set the
location, using a shell like the following?

```html
<iframe id="tensorboard-frame" …></iframe>
<script>
(function() {
  const frame = document.getElementById("tensorboard-frame");
  const url = new URL("/", window.location);
  url.port = %PORT%;
  frame.contentWindow.location.replace(url);
})();
</script>
```

We also need to preserve the protocol, which may be `https` rather than
`http`; the above shell does that.

If you could verify that something like this works for your use case,
that’d be great.

cliffwoolley: I didn't preserve the protocol because I didn't think the TB port could handle that, but I suppose it's true enough that someone could have enabled this.
I did notice the "flash of bad content" issue you mentioned; the reason I didn't do it quite like you said is that it seemed possible that we could have multiple iframes of the same ID, and I was in a bit too much of a rush at the time to generate unique iframe IDs for each.  Thoughts on that?

cliffwoolley: (Regarding protocol, I note that the original implementation always specified `http://localhost/...`, regardless of whether the enclosing Jupyter notebook would have had been loaded via http or https etc., so I followed the assumption of the original code in that aspect.)

wchargin: Ah, you’re right about the IDs, given that Jupyter output cells aren’t themselves framed (like Colab output cells are).

My first thought was to try something like the following—

```html
<script>
(function() {
  const frame = document.createElement("iframe");
  const url = new URL("/", window.location);
  url.port = %PORT%;
  frame.src = url;
  const me = document.currentScript;
  me.parentNode.insertBefore(frame, me);
})();
</script>
```

—but Jupyter rehydrates scripts by appending them to `document.head`, so
`document.currentScript` actually refers to the wrong element here, and
this doesn’t work.

Perhaps we could fall back to a simple nonce-based solution? Something
like:

```python
import cgi
import json
import random

frame_id = "tensorboard-frame-{:08x}".format(random.randrange(0, 1 << 64))
shell = """
<iframe id="%HTML_ID%" width="100%" height="%HEIGHT%"></iframe>
<script>
(function() {
  const frame = document.getElementById(%JSON_ID%);
  const url = new URL("/", window.location);
  url.port = %PORT%;
  frame.src = url;
})()
</script>
"""
replacements = [
    ("%HTML_ID%", cgi.escape(frame_id, quote=True)),
    ("%JSON_ID%", json.dumps(frame_id)),
    ("%PORT%", "%d" % port),
    ("%HEIGHT%", "%d" % height),
]
for (k, v) in replacements:
  shell = shell.replace(k, v)
IPython.display.HTML(shell)
```

cliffwoolley: That seems reasonable; I'll give it a try.

What about the protocol preservation question?

wchargin: Preserving the protocol wasn’t super important in the old code because
the hostname was hard-coded, too. If preserving the protocol is
difficult for some reason, then it’s not a huge deal to assume HTTP; if
it’s basically free (as with the above shell), then let’s do that.

cliffwoolley: OK.  As long as we're in agreement that the "https" that would be supported here would be the Werkzeug "rudimentary" variety discussed in https://github.com/tensorflow/tensorboard/issues/260 , not a case with nginx or similar, unless somebody works out a way to give TensorBoard one `--host` (on a given port) and nginx terminates SSL through to the same port on a different IP address (i.e., the one visible to the client).

cliffwoolley: > you’re right about the IDs, given that Jupyter output cells aren’t themselves framed (like Colab output cells are).

A maybe simpler alternative I just realized - there's nothing stopping us from having nested IFrames inline here.  I'm trying out that approach now.

cliffwoolley: That didn't pan out - even though the nested javascript works, the nested iframe doesn't work inline without more gymnastics than what I was trying to eliminate in the first place.  :)  Following your suggestion and updating the PR.

cliffwoolley:
Hi @wchargin ; anything further you'd like me to change for this one?
Thanks!

tensorboard/notebook.py
```diff
@@ -395,11 +395,31 @@ def _display_colab(port, height, display_handle):
...
-  iframe = IPython.display.IFrame(
-      src="http://localhost:%d" % port,
-      height=height,
-      width="100%",
-  )
+  import cgi
+  import json
+  import random
+
+  frame_id = "tensorboard-frame-{:08x}".format(random.randrange(0, 1 << 64))
```
wchargin: nit: `random.randrange(0, 1 << 64)` can be `random.getrandbits(64)`,
which is a bit clearer.

wchargin:
@googlebot I consent.

wchargin:
I’ve gone ahead and fixed the nits for you, since they were all trivial.
Will merge once CI passes. (Re-tested manually; everything seems fine.)

wchargin:
@cliffwoolley: Thanks again for the contribution!

cliffwoolley:
Thanks for taking care of those last nits!

PR #2417, opened 2019-07-11
text: remove legacy `PluginAsset` support

### LINKED ISSUE #2416: Drop support for legacy text data format by TF 2.0
TensorBoard’s text plugin primarily reads data from events files, like
most other plugins. But it also supports a legacy format that was used
prior to TensorFlow v1.3.0, in which text data is stored in a
`tensors.json` file inside the plugin assets directory.

The text summary TensorFlow op was changed to use the new form in
TensorFlow commit [1e8d1fe6c3], and the dead code path was removed
entirely in [f53579308d].

TensorFlow v1.3.0 was released on 2017-08-16, almost two years ago. We
should drop compatibility with this legacy format as part of the 2.0
release. In the unlikely event that people have critical old-style text
data that needs to be migrated, we can provide a simple data migration
tool.

[1e8d1fe6c3]: https://github.com/tensorflow/tensorflow/commit/1e8d1fe6c3c1221f16c05dd1883f893f291c94d2
[f53579308d]: https://github.com/tensorflow/tensorflow/commit/f53579308d996bbf069eba4ffab68fcc9f6e69f3

wchargin:
Summary:
This is a breaking change for TensorBoard 2.0, which will impact text
summaries generated with version of TensorFlow prior to v1.3.0, which
was released on 2017-08-16. Text summaries generated with TensorFlow
v1.3.0 and later are still supported. Resolves #2416; see that issue for
details.

Test Plan:
Install TensorFlow v1.2.0, then run the following script to generate
test data:

```python
import tensorflow as tf

summ_op = tf.summary.text("name", tf.constant("Alice"))
with tf.Session() as sess:
  summ = sess.run(summ_op)
  w = tf.summary.FileWriter("old_logs", sess.graph)
  w.add_summary(summ)
  w.close()
```

(Including the session graph is necessary for the plugin assets to be
written to disk.)

Then, point TensorBoard at a directory containing bot this `old_logs`
directory and the standard `text_demo` output. Note that it displays
only the new data, whereas prior to this commit it displayed all data.

wchargin-branch: remove-text-plugin-assets


wchargin:
We may want to hold off on merging this until TF 2.0 is right around the
corner, to be sure that we don’t publish a minor or patch release with
this backward-incompatible change.

tensorboard/plugins/text/text_plugin.py
```diff
@@ -231,106 +212,25 @@ def is_active(self):
...
-
-  def _fetch_run_to_series_from_multiplexer(self):
-    # TensorBoard is obtaining summaries related to the text plugin based on
-    # SummaryMetadata stored within Value protos.
-    mapping = self._multiplexer.PluginRunToTagToContent(
-        metadata.PLUGIN_NAME)
+    mapping = self._multiplexer.PluginRunToTagToContent(metadata.PLUGIN_NAME)
     return {
-        run: list(tag_to_content.keys())
+        run: list(six.iterkeys(tag_to_content))
```
nfelt: No reason to use `iterkeys()` only to pass it to `list()` right?  Just `list(tag_to_content)`

wchargin: Done; thanks.

PR #2419, opened 2019-07-12
core: update polymer2 externs

stephanwlee:
### Nature of the change

With the Polymer upgrade, the extern files for JSCompiler is supplied by the Polymer project. We would like to upgrade the type information for better Polymer2 compatibility.

#### About the change in Vulcanize.java
For other externs, in vulcanize/BUILD, we were passing them as files to Vulcanize.java. However, due to the internal restrictions, we can only pass them as a tf_web_library which Vulcanize.java takes as a webfile.pbtxt. Now that externs exist in the webfile.pbtxt, we need to account for that when loading all externs.

Merge plan: squash into polymer2 branch.

### DIFF
--- a/tensorboard/components/tf_imports/BUILD
+++ b/tensorboard/components/tf_imports/BUILD
@@ -17,6 +17,7 @@ tf_web_library(
     name = "polymer",
     srcs = [
         "polymer.html",
+        "polymer-externs.html",
         "@org_definitelytyped//:polymer.d.ts"
     ],
     path = "/tf-imports",
@@ -27,19 +28,6 @@ tf_web_library(
         "@org_polymer",
         "@org_polymer_web_animations_js",
         "@org_polymer_webcomponentsjs",
-    ],
-)
-
-tf_web_library(
-    name = "polymer-externs",
-    srcs = [
-        "polymer-externs.html",
-    ],
-    path = "/polymer",
-    visibility = ["//visibility:public"],
-    exports = ["@org_polymer"],
-    deps = [
-        "@org_polymer",
     ],
 )
 
--- a/tensorboard/components/tf_imports/BUILD
+++ b/tensorboard/components/tf_imports/BUILD
@@ -17,17 +17,28 @@ tf_web_library(
     name = "polymer",
     srcs = [
         "polymer.html",
-        "polymer-externs.html",
         "@org_definitelytyped//:polymer.d.ts"
     ],
     path = "/tf-imports",
     visibility = ["//visibility:public"],
     exports = ["@org_polymer"],
     deps = [
+        ":polymer_externs",
         ":webcomponentsjs",
         "@org_polymer",
         "@org_polymer_web_animations_js",
         "@org_polymer_webcomponentsjs",
+    ],
+)
+
+tf_web_library(
+    name = "polymer_externs",
+    srcs = [
+        "polymer-externs.html",
+    ],
+    path = "/tf-imports",
+    deps = [
+        "@org_polymer_externs",
     ],
 )
 
--- a/tensorboard/components/tf_imports/BUILD
+++ b/tensorboard/components/tf_imports/BUILD
@@ -31,6 +31,19 @@ tf_web_library(
 )
 
 tf_web_library(
+    name = "polymer-externs",
+    srcs = [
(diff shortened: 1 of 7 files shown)


tensorboard/components/tf_imports/polymer-externs.html
```diff
@@ -0,0 +1,4 @@
+<script src="externs/closure-types.js"></script>
+<script src="externs/polymer-externs.js"></script>
+<script src="externs/polymer-internal-shared-types.js"></script>
```
wchargin: This file doesn’t have an `@externs`. Do we need to do anything special
to handle it correctly during vulcanization?

tensorboard/components/tf_imports/BUILD
```diff
@@ -30,6 +30,19 @@ tf_web_library(
     ],
 )
 
+tf_web_library(
+    name = "polymer-externs",
```
wchargin: Can we stick to snake\_case for target names where possible, just for
consistency?

tensorboard/components/tf_imports/BUILD
```diff
@@ -30,6 +30,19 @@ tf_web_library(
...
+    srcs = [
+        "polymer-externs.html",
+    ],
+    path = "/polymer",
+    visibility = ["//visibility:public"],
+    exports = ["@org_polymer"],
```
wchargin: Naïve question: If we omit this “exports”, what happens? (I don’t really
know what `exports` does.)

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -149,7 +149,18 @@ public static void main(String[] args) throws IOException {
       }
       Webfiles manifest = loadWebfilesPbtxt(Paths.get(args[i]));
       for (WebfilesSource src : manifest.getSrcList()) {
-        webfiles.put(Webpath.get(src.getWebpath()), Paths.get(src.getPath()));
+        String webpath = src.getWebpath();
+        Path srcPath = Paths.get(src.getPath());
+
+        webfiles.put(Webpath.get(webpath), srcPath);
+
+        if (webpath.startsWith("/polymer/externs")) {
```
wchargin: Check `"/polymer/externs/"` (with trailing slash) to prevent spliced-in
paths like `"/polymer/externs-foo"`?

tensorboard/components/tf_imports/polymer-externs.html
stephanwlee: There seems to be a discrepancy with the v3 source which internal polymer is synced with. As per your offline feedback, I decided to use the `jscomp-externs` attribute to detect an extern instead of looking for `@externs` in the code.

tensorboard/components/tf_imports/BUILD
stephanwlee: No longer needed but to answer question: it re-exports all the symbols in the `@org_polymer` and make it accessible when someone depends on this target.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -321,7 +321,14 @@ private static Node visitScript(Node node) throws IOException {
       script = INLINE_SOURCE_MAP_PATTERN.matcher(script).replaceAll("");
     }
     boolean wantsMinify = getAttrTransitive(node, "jscomp-minify").isPresent();
-    if (node.attr("src").endsWith(".min.js")
+
+    if (node.hasAttr("jscomp-externs")) {
+      Path filePath = getWebfile(path);
+      SourceFile sourceFile = SourceFile.fromCode(path.toString(), script);
+      externs.add(sourceFile);
```
wchargin: Is duplication an issue? If so, `externs` can be a `LinkedHashSet`.

stephanwlee: Added `LinkedHashMap` to dedup based on the file name. As I alluded to in the commit message, not all extern files have webpath and, hence, I decided to dedup with the file path.

stephanwlee:
Noticed I added some suppression in the latest commit.

PR #2429, opened 2019-07-15
vulcanize: factor out `removeNode` helper

wchargin:
Test Plan:
Built `//tensorboard/components:index.html` before and after this
change, and verified that the two copies are bit-for-bit equivalent.

wchargin-branch: remove-node

### DIFF
--- a/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
+++ b/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
@@ -277,10 +277,10 @@ public final class Vulcanize {
         if (licenseComment == null) {
           licenseComment = node;
         } else {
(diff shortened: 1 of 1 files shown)


tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -405,6 +405,10 @@ private static Node replaceNode(Node oldNode, Node newNode) {
     return newNode;
   }
 
+  private static Node removeNode(Node node) {
+    return replaceNode(node, new TextNode("", node.baseUri()));
```
stephanwlee: I assume this will be changed after merging #2430?

wchargin: Yep.

PR #2434, opened 2019-07-17
Add median stats to What-If Tool regression performance stats table

jameswex:

Users asked for median error stats in addition to current mean error stats

Added calculation of, and display of, median error, median absolute error, and median squared error to the regression performance table.

* Screenshots of UI changes

medians

Ran regression demo and viewed performance stats in total and broken down into slices. Verified correct median numbers and that sorting by those new columns worked.

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -3507,7 +3507,7 @@ limitations under the License.
             return b.medianError - a.medianError;
           } else if (this.selectedFeatureSort == 'Median absolute error') {
             return b.medianAbsError - a.medianAbsError;
-          } else {
+          } else if (this.selectedFeatureSort == 'Median squared error') {
             return b.medianSquaredError - a.medianSquaredError;
           }
         });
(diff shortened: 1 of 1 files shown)


jameswex:
@tolga-b please review

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3491,20 +3503,40 @@ <h2>Show similarity to selected datapoint</h2>
...
-        // calculate mean error, mean abs error, and mean squared error.
+        // calculate error statistics.
         function mean(data) {
           const sum = data.reduce((sum, value) => {
             return sum + value;
           }, 0);
 
           return sum / data.length;
         }
+        function median(sortedData) {
```
tolga-b: It may be a style thing, but if we sorted arrays within the median function, we could get rid of some repetition in the inputs below (l3549-l3551).

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3491,20 +3503,40 @@ <h2>Show similarity to selected datapoint</h2>
...
+          if (sortedData.length == 0) {
```
tolga-b: mean does not have this check, is it necessary there as well?

jameswex: mean will return NaN already due to its use of reduce with an initial value of 0. Since median uses indexing instead of a reduce fn, it needs the check.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
jameswex: done

PR #2442, opened 2019-07-18
hparams: allow setting trial ID

### LINKED ISSUE #2440: HParams Session Group Name.
When setting hyper parameter tuning, there is no way to pass a session group name.  following the folder name convention would be helpful to distinguish them.
HParams Table View

wchargin:
Summary:
Resolves #2440. See #1998 for discussion.

Test Plan:
The hparams demo still does not specify trial IDs (intentionally, as
this is the usual path). But apply the following patch—

```diff
diff --git a/tensorboard/plugins/hparams/hparams_demo.py b/tensorboard/plugins/hparams/hparams_demo.py
index ac4e762b..38b2b122 100644
--- a/tensorboard/plugins/hparams/hparams_demo.py
+++ b/tensorboard/plugins/hparams/hparams_demo.py
@@ -160,7 +160,7 @@ def model_fn(hparams, seed):
   return model
 
 
-def run(data, base_logdir, session_id, hparams):
+def run(data, base_logdir, session_id, trial_id, hparams):
   """Run a training/validation session.
 
   Flags must have been parsed for this function to behave.
@@ -179,7 +179,7 @@ def run(data, base_logdir, session_id, hparams):
       update_freq=flags.FLAGS.summary_freq,
       profile_batch=0,  # workaround for issue #2084
   )
-  hparams_callback = hp.KerasCallback(logdir, hparams)
+  hparams_callback = hp.KerasCallback(logdir, hparams, trial_id=trial_id)
   ((x_train, y_train), (x_test, y_test)) = data
   result = model.fit(
       x=x_train,
@@ -235,6 +235,7 @@ def run_all(logdir, verbose=False):
           data=data,
           base_logdir=logdir,
           session_id=session_id,
+          trial_id="trial-%d" % group_index,
           hparams=hparams,
       )
 
```

—and then run `//tensorboard/plugins/hparams:hparams_demo`, and observe
that the HParams dashboard renders a “Trial ID” column with the
specified IDs:

![Screenshot of new version of HParams dashboard][1]

[1]: https://user-images.githubusercontent.com/4317806/61491024-1fb01280-a963-11e9-8a47-35e0a01f3691.png

wchargin-branch: hparams-trial-id

### DIFF
--- a/tensorboard/plugins/hparams/summary_v2.py
+++ b/tensorboard/plugins/hparams/summary_v2.py
@@ -206,7 +206,7 @@ def _normalize_hparams(hparams):
 
 def _derive_session_group_name(trial_id, hparams):
   if trial_id is not None:
-    if not isinstance(trial_id, str):
+    if not isinstance(trial_id, six.string_types):
       raise TypeError("`trial_id` should be a `str`, but got: %r" % (trial_id,))
     return trial_id
   # Use `json.dumps` rather than `str` to ensure invariance under string
--- a/tensorboard/plugins/hparams/summary_v2.py
+++ b/tensorboard/plugins/hparams/summary_v2.py
@@ -36,7 +36,7 @@ from tensorboard.plugins.hparams import metadata
 from tensorboard.plugins.hparams import plugin_data_pb2
 
 
-def hparams(hparams, start_time_secs=None):
+def hparams(hparams, trial_id=None, start_time_secs=None):
   # NOTE: Keep docs in sync with `hparams_pb` below.
   """Write hyperparameter values for a single trial.
 
@@ -46,6 +46,8 @@ def hparams(hparams, start_time_secs=None):
       experiment, or the `HParam` objects themselves. Values should be
       Python `bool`, `int`, `float`, or `string` values, depending on
       the type of the hyperparameter.
+    trial_id: An optional `str` ID for the set of hyperparameter values
+      used in this trial. Defaults to a hash of the hyperparameters.
     start_time_secs: The time that this trial started training, as
       seconds since epoch. Defaults to the current time.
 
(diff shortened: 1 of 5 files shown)


tensorboard/plugins/hparams/summary_v2.py
```diff
@@ -199,7 +204,11 @@ def _normalize_hparams(hparams):
   return result
 
 
-def _derive_session_group_name(hparams):
+def _derive_session_group_name(trial_id, hparams):
+  if trial_id is not None:
+    if not isinstance(trial_id, str):
```
stephanwlee: Is this the prefered way? https://six.readthedocs.io/#six.string_types

tensorboard/plugins/hparams/summary_v2.py
```diff
@@ -55,12 +57,13 @@ def hparams(hparams, start_time_secs=None):
...
   pb = hparams_pb(
       hparams=hparams,
+      trial_id=trial_id,
       start_time_secs=start_time_secs,
   )
   return _write_summary("hparams", pb)
 
 
-def hparams_pb(hparams, start_time_secs=None):
+def hparams_pb(hparams, trial_id=None, start_time_secs=None):
```
stephanwlee: Probably no action required: if we shipped this API as part of 1.14, and if a user don't use the kwarg, this change will break them, right?

wchargin: Correct; you’re supposed to use use the kwarg.

(But we wouldn’t ship this as part of 1.14 anyway, because it’s a new
feature.)

tensorboard/plugins/hparams/summary_v2.py
wchargin: Yes; that’s probably better. Might as well fully support Python 2.

tensorboard/plugins/hparams/summary_v2.py
wchargin: Maybe we can change these to be keyword-only arguments once we drop
Python 2 support, and likewise with `TBContext` and friends.

PR #2444, opened 2019-07-18
cleanup: apply buildifier

stephanwlee:
Motivation: upcoming LSC will make changes to BUILD files and to keep
the process sane, we will be using buildifier. We would like to keep the
git history meaningful and want to keep the format changes in a separate
commit.

Did `find . -name "BUILD" | xargs buildifier`.


wchargin:
Also (no change required, but just so you know):

> `find . -name "BUILD" | xargs buildifier`

This isn’t safe in general, because `xargs` splits on whitespace and
file paths may contain whitespace. You can use

```sh
find . -name "BUILD" -print0 | xargs -0 buildifier
```

to use NUL delimiters (file paths cannot contain NUL characters), or you
can just use

```sh
find . -name "BUILD" -exec buildifier {} +
```

where `{}` will be automatically replaced with the file paths. (See
[`find`(1) spec][find] for details.) The version with `find` is more
portable, as `xargs -0` isn’t in POSIX(!).

[find]: https://pubs.opengroup.org/onlinepubs/009695399/utilities/find.html

wchargin:
Or, if you’d prefer to just get the LSC in (not sure what its timeline
is), we can punt on the enforcement and just approve this for now?

PR #2445, opened 2019-07-18
fix: update polymer dependency to legacy targets

stephanwlee:
For the google sync, TensorBoard have the legacy
tensorboard_webcomponent_library targets. Stale dependencies aside, now
in TensorBoard, we depend on tf_imports:polymer instead of the polymer
directly. This change reflects that.

### DIFF
--- a/tensorboard/components/tf_imports/BUILD
+++ b/tensorboard/components/tf_imports/BUILD
@@ -1,5 +1,6 @@
 package(default_visibility = ["//tensorboard:internal"])
 
+load("//tensorboard/defs:defs.bzl", "tensorboard_webcomponent_library")
 load("//tensorboard/defs:hacks.bzl", "tensorboard_typescript_bundle")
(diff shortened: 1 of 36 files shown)


wchargin:
Will review today.

tensorboard/components/tf_imports/BUILD
```diff
@@ -31,6 +32,19 @@ tf_web_library(
...
 )
 
+tensorboard_webcomponent_library(
+    name = "polymer_lib",
+    srcs = [":polymer"],
+    visibility = ["//visibility:public"],
+    destdir = "tf-imports",
+    deps = [
+        ":polymer_externs_lib",
+        "//third_party/javascript/polymer/v1/web-animations-js:other-js",
```
wchargin: Assuming intentional that this is v1 while others are v2.

stephanwlee: There isn't v2 version of web-animations-js :(

PR #2452, opened 2019-07-22
build: fix JSComp removing methods issue

### LINKED ISSUE #2447: Polymer 2.7 upgrade: JSComp mangles certain function names
https://github.com/tensorflow/tensorboard/pull/2392 caused the following problem in the debugger plugin: 
method _renderLatest not defined

This happens during stepping through the graph. The rows in the tensor summary table that are supposed to appear don't appear correctly.

@stephanwlee suggested looking into the JSComp options and find a way to prevent this mangling:
https://github.com/tensorflow/tensorboard/blob/master/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java#L446

### LINKED ISSUE #2451: HParams table always displays “No metrics are enabled”
…even when there actually are metrics:

![Screenshot of the HParams dashboard complaining that “No metrics are
enabled” but simultaneously showing those very metrics][img]

This occurs on HEAD, but not 1.14. Suspect Polymer 2 changes.

The [guard for the `dom-if`][if] is `[[!_haveMetrics(…)]]`, and indeed
[the `_haveMetrics` method appears to be defined correctly][hm]. But the
JS console has warnings, “method `_haveMetrics` not defined”.
Intriguingly, after selecting the element in the Chrome dev tools,
`$0._haveMetrics` is `undefined`, but `$0._haveMetricsAndSessionGroups`
is not, even though the two methods are siblings and look structurally
similar.

Maybe the Closure compiler is removing one of them but not the other,
but it’s not clear to me why that might be happening in this case.

[hm]: https://github.com/tensorflow/tensorboard/blob/223073b9fe085d092a407bb3466c9f82733cfe0d/tensorboard/plugins/hparams/tf_hparams_session_group_details/tf-hparams-session-group-details.html#L215-L219
[if]: https://github.com/tensorflow/tensorboard/blob/223073b9fe085d092a407bb3466c9f82733cfe0d/tensorboard/plugins/hparams/tf_hparams_session_group_details/tf-hparams-session-group-details.html#L41-L46
[img]: https://user-images.githubusercontent.com/4317806/61655420-62c7f980-ac73-11e9-8082-00106f5c8f32.png

stephanwlee:
There were cases when JSComp removed Polymer bound methods (invisible to
JSComp) away from compiled binary (e.g., inlined into another method).
This change prevents such optimization.

Fixes #2451. Fixes #2447.

### DIFF
--- a/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
+++ b/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
@@ -448,6 +448,9 @@ public final class Vulcanize {
     options.setRemoveUnusedPrototypeProperties(false);
     options.setRemoveUnusedPrototypePropertiesInExterns(false);
     options.setRemoveUnusedClassProperties(false);
+    // Prevent Polymer bound method (invisible to JSComp) to be over-optimized.
+    options.setInlineFunctions(CompilerOptions.Reach.NONE);
+    options.setDevirtualizeMethods(false);
 
     // Dependency management.
     options.setClosurePass(true);


caisq:
Confirmed it fixes #2447. Thanks, @stephanwlee !

stephanwlee:
I was curious about #2447 which I have not carefully looked into. It also had the same problem of inline function + hoisting the function (obfuscated) up out of the Polymer spec into the local module scope. This change does address it.

Also confirmed that after the change, cycling through all plugins, I no longer see the Polymer runtime warning about unbound methods.

PR #2461, opened 2019-07-25
What-If Tool: Add ability to sort PD plots by interestingness

jameswex:

Long-standing feature request to allow the tool to sort all features by how interesting their partial dependence plots are, to help users explore that space in a more directed manner.

Add new backend capability to sort features by how interesting their PD plots are (for a given selected example, or globally for all examples, depending on user selection). This is done by measuring the Y-axis movement in each PD plot for each feature and sorting them by this measure.
Add new button to run this sort, which runs in the background and can take a while. Therefore added a spinner on the button to show it is calculating
Add path in all implementations (TensorBoard, jupyter, and colab) to handle this sort request and return the response
Add UI to indicate features for which the PD plot is completely flat. These features are grayed-out and at the bottom of the sorted list
Added unit tests for new python code
Updated all demos to be able to handle sorting of eligible features in JS for demo purposes

* Screenshots of UI changes

During sorting processing:
pdthinking

After sorting:
aftersort

Ran demos and used feature. Verified correct sorting.
Ran TensorBoard, colab notebooks, and jupyter notebooks and used feature to verify it works through all code paths.
Ran unit tests of new py code


jameswex:
@tolga-b please review

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -5666,6 +5694,51 @@ <h2>Show similarity to selected datapoint</h2>
...
+          const setEligibleFields = result => {
+            this.set('partialDepPlotEligibleFeatures', result.value);
+          };
+          const sortErrorCallback = () => {
+            this.isSortingEligibleFeatures = false;
+          }
+          this.makeAsyncRequest_(url,
+            setEligibleFields, null, 'plot ordering', sortErrorCallback.bind(this));
+        } else {
+          urlParams['features'] = this.partialDepPlotEligibleFeatures;
```
tolga-b: nit: Not sure which is faster, but we could pass this to TensorBoard version as well so you don't have to call python get_eligible_features again. Not necessary.

tensorboard/plugins/interactive_inference/interactive_inference_plugin.py
```diff
@@ -322,9 +323,51 @@ def _eligible_features_from_example_handler(self, request):
...
+      request: A request for sorted features.
+
+    Returns:
+      A sorted list with a JSON object for each feature.
+      Numeric features are represented as
+      {name: observedMin: observedMax: interestingness:}.
+      Categorical features are repesented as
+      {name: samples:[] interestingness:}.
+    """
+    try:
```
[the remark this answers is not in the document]
jameswex: As discussed, they only exist for a given feature on JS side if the feature has been expanded, so there is no way to access all plots of JS without having python compute all of them.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
jameswex: that part isn't the bottleneck of this operation, the generating of all PD plots for all features is. so i don't think its necessary.

tolga-b: sounds good.

tensorboard/plugins/interactive_inference/interactive_inference_plugin.py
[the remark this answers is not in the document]
tolga-b: sounds good!

jameswex:
@davidsoergel can you please review? Thanks.

davidsoergel:
yep, will do today

tensorboard/plugins/interactive_inference/utils/inference_utils.py
```diff
@@ -650,6 +652,58 @@ def get_eligible_features(examples, num_mutants):
...
+          if is_numeric:
+            # For numeric features, interestingness is the total Y distance
+            # traveled across the line chart.
+            measure = 0
+            for i in range(len(series) - 1):
+              measure += abs(series[i]['scalar'] - series[i + 1]['scalar'])
+          else:
+            # For categorical features, interestingness is the difference
+            # between the min and max Y values in the chart. This is
+            # identical to total Y distance, as the entries in this chart
```
davidsoergel: Why make a distinction if the result is identical?  It looks like the above implementation for the numeric case would work for categorical too.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/demo/tf-interactive-inference-demo.html
```diff
@@ -340,131 +340,169 @@
...
               this.$.dash.makeChartForFeature(
                 isNum ? 'numeric' : 'categorical',
                 e.detail.feature_name,
-                [[{'1': results}]]
+                data
               );
             };
             setTimeout(method, 50);
           });
+          this.$.dash.addEventListener('sort-eligible-features', async (e) => {
```
davidsoergel: Can all this duplicated code be factored out?

tensorboard/plugins/interactive_inference/utils/inference_utils.py
```diff
@@ -650,6 +652,58 @@ def get_eligible_features(examples, num_mutants):
     features_list.append(v)
   return features_list
 
+def sort_eligible_features(features_list, chart_data):
+  """Returns a sorted list of JSON objects for each feature in the examples.
```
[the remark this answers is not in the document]

tensorboard/plugins/interactive_inference/utils/inference_utils.py
```diff
@@ -650,6 +652,58 @@ def get_eligible_features(examples, num_mutants):
...
+            for i in range(len(series)):
+              val = series[i]['scalar']
+              if val < min_y:
+                min_y = val
+              if val > max_y:
+                max_y = val
+            measure = max_y - min_y
+          if measure > max_measure:
+            max_measure = measure
+    feature['interestingness'] = max_measure
```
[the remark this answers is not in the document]
jameswex: updated to make a deep copy and then mutate and return that.

tensorboard/plugins/interactive_inference/utils/inference_utils.py
jameswex: reworded this comment as it was misleading, we do want the separate logic here so as not to depend on the categorical charts being sorted when this method is called. the charts are sorted before being visualized but don't want to have this logic depend on them being sorted here in py where they are created.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/demo/tf-interactive-inference-demo.html
jameswex: agreed. i would like to factor out all the duplicated code between demos, not just for this new feature, but didn't want to bog down this review with that whole effort.

tensorboard/plugins/interactive_inference/utils/inference_utils.py
davidsoergel: yep gotcha

PR #2462, opened 2019-07-25
What-If Tool: Fix error message overflow issue

jameswex:

Long WIT error messages caused text overflow, hiding some controls.

Added correct CSS rules to avoid overflow issue and to ellipsis the status string. The full error string is already shown in a toast as well.

* Screenshots of UI changes

Fixed:
overflow

Ran demo app, added fake very long error message, verified correct visual behavior (and incorrect behavior before fix).

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -159,6 +159,7 @@ limitations under the License.
 
       .settings-button {
         margin-top: 4px;
+        min-width: 40px;
       }
 
       .button {
@@ -400,6 +401,7 @@ limitations under the License.
       .datapoint-right-controls-holder {
         display: flex;
         flex-direction: row-reverse;
+        overflow: hidden;
       }
 
       .tf-category-pane {
@@ -736,6 +738,7 @@ limitations under the License.
         color: #5F6368;
         white-space: nowrap;
         overflow: hidden;
+        text-overflow: ellipsis;
         padding-top: 16px;
       }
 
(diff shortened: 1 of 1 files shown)


jameswex:
@tolga-b please review

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -400,6 +401,7 @@
       .datapoint-right-controls-holder {
         display: flex;
         flex-direction: row-reverse;
+        overflow: hidden;
```
stephanwlee: I think it is more correct to put `flex-shrink: 0;`.

jameswex: I did try that but it doesn't work. With this container having flex-shrink: 0, the child div with lots of text will still cause this container to grow in width beyond the screen, leading to the issue.

jameswex: moved to flex-shrink: 0 for the two buttons as you suggested and removed the min-width setting for the button.

still need the overflow: hidden on the right-controls-holder to avoid overflow.

stephanwlee: Yup, understand that now.

stephanwlee:
Btw, I think the text/buttons should be vertically aligned :)

jameswex:
> Btw, I think the text/buttons should be vertically aligned :)

adjusted margin so that status text matches alignment for left-side tab text and right side buttons. Thanks.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -742,7 +742,7 @@
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
-        padding-top: 16px;
```
stephanwlee: Oh, I think the more robust solution is to use `align-items: center` on flex container and remove all these padding/margins.

Issue #2458, opened 2019-07-25
Error running extract_pip_package


linzwatt:
Hi, I can build tensorboard using `bazel build //tensorboard` and I can run the compiled tb with my local changes. But I cannot get a pip package out, in order to deploy it for the rest of our team to use.

From some other issues/stackoverflow answers you build the pip package with
`bazel run //tensorboard/pip_package/extract_pip_package`

When I run this I get the error below

I have attached the output of `diagnose_tensorboard.py` here as well.

Thanks,

### Diagnostics

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version 393931f9685bd7e0f3898d7dcdf28819fef54c43

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=7, micro=3, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): posix.uname_result(sysname='Linux', nodename='ip-172-31-31-96', release='4.4.0-1087-aws', version='#98-Ubuntu SMP Wed Jun 26 05:50:53 UTC 2019', machine='x86_64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: True
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
INFO: installed: tb-nightly==1.15.0a20190724
diagnose_tensorboard.py:197: DeprecationWarning: The 'warn' function is deprecated, use 'warning' instead
  logging.warn("no installation among: %s", sorted(family))
WARNING: no installation among: ['tensorflow', 'tensorflow-gpu', 'tf-nightly', 'tf-nightly-2.0-preview', 'tf-nightly-gpu', 'tf-nightly-gpu-2.0-preview']
WARNING: no installation among: ['tensorflow-estimator', 'tensorflow-estimator-2.0-preview', 'tf-estimator-nightly']

--- check: tensorboard_python_version
INFO: tensorboard.version.VERSION: '1.15.0a0'

--- check: tensorflow_python_version
Traceback (most recent call last):
  File "diagnose_tensorboard.py", line 419, in main
    suggestions.extend(check())
  File "diagnose_tensorboard.py", line 77, in wrapper
    result = fn()
  File "diagnose_tensorboard.py", line 236, in tensorflow_python_version
    import tensorflow as tf
ModuleNotFoundError: No module named 'tensorflow'

--- check: tensorboard_binary_path
[...]

PR #2463, opened 2019-07-25
build_pip_package: print build log on failure

wchargin:
Summary:
We don’t use `set -x` indiscriminately because build tool output appears
in Bazel logs, and that would be a lot of spam. But we should print
output on failures, to clarify the cause of issues like #2458.

The original version of #2363 had this functionality in the BUILD rule
itself, but @nfelt found that the multiple layers of escaping made this
a bit hard to read (fair enough), so we now include the relevant
functionality in `build_pip_package.sh`.

Recommend ignoring whitespace for reading this patch.

Test Plan:
Building `//tensorboard/pip_package` on a machine with no `virtualenv`
binary now prints:

```
ERROR: /home/wchargin/tensorboard/tensorboard/pip_package/BUILD:8:1: Executing genrule //tensorboard/pip_package:pip_package failed (Exit 127) bash failed: error executing command /bin/bash -c ... (remaining 1 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
++ uname
+ [ Linux = Darwin ]
+ sedi=sed -i
+ command -v virtualenv
Target //tensorboard/pip_package:pip_package failed to build
Use --verbose_failures to see the command lines of failed build steps.
FAILED: Build did NOT complete successfully
```

which is pretty clear that the problem is that `virtualenv` is not
installed. After installing `virtualenv`, the build completes
successfully with no output, as desired.

wchargin-branch: pip-package-error-handling

### DIFF
--- a/tensorboard/pip_package/build_pip_package.sh
+++ b/tensorboard/pip_package/build_pip_package.sh
@@ -13,8 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eu
-
 usage() {
   cat <<EOF
 usage: build_pip_package OUTPUT_DIR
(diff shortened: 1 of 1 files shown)


tensorboard/pip_package/build_pip_package.sh
```diff
@@ -29,93 +27,113 @@ Arguments:
...
+    rm -r "${workdir}"
+  }
+  trap cleanup EXIT
+
+  log_file="${workdir}/log"
+  suppress_log() {
+    rm "${workdir}/log" || true
+  }
+
+  build 3>&1 4>&2 1>"${log_file}" 2>&1
```
nfelt: It seems like we could just do redirection on 1 and 2 and skip the slightly more unusual use of 3 and 4 if we were willing to just say that there isn't any way to write to the original stdout and stderr from within `build` - we don't seem to be using `3` at all, and the one use of `4` (to emit the no-such-directory error) seems like we could just write that to the redirected stderr and print a more verbose log message.  (In which case we don't need to implement `suppress_log` either, and we could change the `if ! [ -d "${output}" ]` check to just be an unadorned `[ -d "${output}" ]` like we do earlier for `RUNFILES`.)

wchargin: Sure thing: removed the redirections in favor of a noisier log. And indeed it’s really not that bad:

```
+ case "${output}" in
+ '[' -d /tmp/nope ']'
+ printf 'fatal: no such output directory: %s\n' /tmp/nope
fatal: no such output directory: /tmp/nope
+ return 1
```

I’ve kept the explicit `! [ -d "${output}" ]` because this is a usage
error and imho deserves a nicer message (and it’s not hurting
readability or anything).

Issue #2003, opened 2019-03-12
BUILD: revamp frontend building


stephanwlee:
# Current state
Most of our build is done in our Vulcanization process today.
Before bringing up what a Vulcanization entails, it is important to note that dependencies of a module is defined in HTML via `script` and `import` tags. Vulcanization talks the dependency graph and: 
- first builds TypeScript files (along with necessary d.ts files) that outputs .js and a .d.ts file
- compile all JSs (generated included) using JSCompiler into one bundle
- depending on options or "environment", it inlines the bundle into the script tag
In all of these processes, it makes sure a dependency is properly provided via Bazel deps.

One interesting artifact of Vulcanization is the concept of `webPath`. Instead of using relative or absolute path to a module, modules, in its `tf_web_library`, define a `webPath` which indicates a path of the module in the bundle artifact. For instance, say there is a module in *a/b/c/d/foo.js* which defines webPath of */tf-foo/foo*. When other modules use the foo, it will define dependency as `<import href="/tf-foo/foo" />` (or a relative path from its webPath). This loosely is related to concept of `webfiles` of Bazel/rules_closure, though, stephanwlee lacks profound knowledge in it yet. 

The design choice of `webPath` and Vulcanization impact how we author JS/TS. The natural way of defining dependency in JavaScript is via CommonJS or ES module (modern browsers have [native ES module support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import)) but we have to use `namespace` and `window` object (e.g., `window.tf.graph.common`) because we define dependency via HTML. This is not to say concept of Vulcanization is incompatible with es module completely but it requires non-trivial changes.

# Future state
In the future, we want our solution to:
- use Bazel: we don't want part of our app to be on Bazel and others on another build system
- have ability to easily test FE (karma integration at the very least is required)
- have less customization of build system if possible (i.e., deprecate `Vulcanization.java`)
- other framework (e.g., React) friendly 

With above requirements in mind, this section only is attempting capture one point in the vast solution space. 
  
1. Use Bazel/rules_typescript to build TypeScript
This is the canonical way of building TypeScript with Bazel.

2. Utilize Bazel/rules_nodejs as much as possible
rules_nodejs allows us to pull dependencies from NPM and use [node ecosystem to build and test](https://github.com/dropbox/rules_node).

4. Move off of JSCompiler and use Webpack, Rollup, or None
[...]

PR #2472, opened 2019-07-26
core: introduce rules_nodejs and angular shim

stephanwlee:
TensorBoard has been using custom Vulcanization to build the frontend
but we have plans to use better officially Bazel-team supported rule:
rules_nodejs. This adds the WORKSPACE dependency to the repo.

Also, to prove the correctness of the integration, this change also
Angular CLI generated (heavily trimmed down and modified) code with its
build rule.


tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
stephanwlee: Do note that, by default, ts_library and necessary rule use "//:tsconfig.json" as the default label for tsconfig field. https://github.com/bazelbuild/rules_nodejs/blob/432f88c8acc47b95ae6f368d3c5b7fe3e65baef7/packages/typescript/src/internal/build_defs.bzl#L424-L425

yarn.lock
```diff
@@ -0,0 +1,3323 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
nfelt: How and when does this file get updated / how do we keep it in sync?

package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "tensorboard",
+  "version": "1.14.0",
```
nfelt: Presumably we'll want some way to keep this number, at a minimum, in sync with our pip package version?  Maybe we can add a presubmit check that verifies that this number always matches version.py?

package.json
```diff
@@ -0,0 +1,49 @@
...
+    "ml",
+    "visualization"
+  ],
+  "author": "TensorFlow authors",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/tensorflow/tensorboard/issues"
+  },
+  "homepage": "https://github.com/tensorflow/tensorboard#readme",
+  "devDependencies": {
```
nfelt: So does this essentially replace js.bzl now?  This becomes the source of truth for JS dependency versioning?

WORKSPACE
```diff
@@ -16,7 +16,7 @@ http_archive(
 load("@bazel_skylib//lib:versions.bzl", "versions")
 # Keep this version in sync with the BAZEL environment variable defined
 # in our .travis.yml config.
-versions.check(minimum_bazel_version = "0.22.0")
+versions.check(minimum_bazel_version = "0.26.1")
```
nfelt: Just to clarify, is 0.26.1 the minimum for any reasonable version of rules_nodejs?  Or if we went back a few versions on that, could we retain compatibility across more than one bazel version?

BUILD
```diff
@@ -0,0 +1,15 @@
+load("@npm_bazel_typescript//:index.bzl", "ts_config")
+
+licenses(["notice"])  # Apache 2.0
+
+exports_files(["tsconfig.json"])
+
+
+ts_config(
```
nfelt: Since this level of the Bazel package hierarchy doesn't exist internally, do we know how we're going to sync this and the other files?  Or do we not use any of these internally?

yarn.lock
wchargin: I can’t speak to `rules_nodejs`, but how it works in the outside world
is that `yarn.lock` changes only when you add, remove, or upgrade
dependencies. You do so by running `yarn add` (to add or upgrade) or
`yarn remove`, which updates `package.json` and `yarn.lock` as well as
actually downloading the package to `node_modules` if required.

caisq:
Thanks for working on this, @stephanwlee . 

cc'ing myself for notifications.

stephanwlee:
Okay, now, I am confident that everything will work as intended. I tried to transfer the new addition internally and I did get everything to build correctly! cl/260512283
image

yarn.lock
stephanwlee: Yup, one is expected to call `yarn add` and `yarn remove` to add and remove a dependency, respectively.

package.json
stephanwlee: We can do that or we can make this version preposterous like 0. This number is mostly useful only when we are publishing a npm package which we are not (note, `private: "true"`). Which would you prefer?

package.json
stephanwlee: Yes! We would not need js.bzl/polymer.bzl and other bzl variants. A bazel_nodejs rule will fetch the dependency correctly for us.
https://github.com/bazelbuild/rules_nodejs#using-bazel-managed-dependencies

WORKSPACE
stephanwlee: > Just to clarify, is 0.26.1 the minimum for any reasonable version of rules_nodejs?

0.26.0 is the minimum version required by rules_nodejs.

BUILD
stephanwlee: > Or do we not use any of these internally?

This :)

stephanwlee:
Here is a small doc on how to manage dependency in the new world. https://github.com/tensorflow/tensorboard/issues/2003#issuecomment-516091708 

I will update it and keep it up-to-date.

package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "tensorboard",
+  "version": "0.0.0",
```
nfelt: Can we perhaps set this to `"0.0.0-unused"` just to be extra clear?

wchargin:
:tada:

stephanwlee:
:lipstick: ?

package.json
```diff
@@ -0,0 +1,49 @@
...
+  "bugs": {
+    "url": "https://github.com/tensorflow/tensorboard/issues"
+  },
+  "homepage": "https://github.com/tensorflow/tensorboard#readme",
+  "devDependencies": {
+    "@angular/bazel": "^8.1.2",
+    "@angular/cli": "^8.1.2",
+    "@angular/compiler": "^8.1.2",
+    "@angular/compiler-cli": "^8.1.2",
+    "@bazel/bazel": "^0.23.2",
```
wchargin: This PR bumps the Bazel version in `WORKSPACE` from 0.22.0 to 0.26.1,
but adds a dependency in `package.json` at version 0.23.2. Is there any
reason for the discrepancy?

Cf. <https://github.com/tensorflow/tensorboard/issues/2534#issuecomment-520073054>

Issue #2437, opened 2019-07-17
Debugger double stamps when loaded with event files without scalars


stephanwlee:
Repro step: 
1. `tensorboard --logdir foobar_idontexist --debugger_port 6007`
2. open tensorboard
3. it should open the debugger plugin because it is the only active plugin
4. notice that in the DOM, the scaffold for scalar plugin contains tf_debugger_dashboard.

This is not working as intended and, thus, is a bug.

Note that this is _not_ reproducible for images dashboard (and potentially others).

caisq:
@stephanwlee It turns out: even if tensorboard points to a non-empty logdir, the double stamping still happens if I navigate directly to the "host:port/#debugger" URL. It doesn't happen, if I navigate to the "host:port/#scalar" first and *then* navigate to "host:port/#debugger"

stephanwlee:
I had some time to look into the issue and it is non-trivial to fix it.

Fact 1: after the dashboard registration change, most plugins are registered/make-known-to-tb from the plugins listing. There are few plugins that use the old mechanism and they are "debugger", "profile", "beholder", "whatif", "hparams". 

Fact 2: when plugins_listing request resolves successfully, it can changes `_dashboardData` and the order of the plugin shifts around (debugger gets pushed back while scalars appear at the start)

Fact 3: because we manually stamp/manipulate the DOM, even if Polymer changes dashboardDatum on dom-repeat, it is largely invisible to us.

Problem: at tensorboard start up, the debugger is stamped right away if it is the `activePlugin` as the first dashboard (dom-wise). After the plugins_listing resolve, the scalars become the first plugin and Polymer manipulates the DOM to insert items at the end then update the existing DOM nodes instead of prepending new items and appending new items (this happens because there is no notion of key to dom-repeat). In Polymer v1, the solution would be to use the ArrayUpdateHelper (which facilitates correct usage of Polymer array API ) but Polymer v2/v3 seem to have broken the support for splice events. 

There is no easy way forward but we do have plans to do redo the tf-tensorboard that will address this.

caisq:
@stephanwlee Any ETA to the tf-tensorboard revision? If it's long (after the next release), I'll take some patching measures in the debugger pugin.

stephanwlee:
> Any ETA to the tf-tensorboard revision 

I don't think we want to fix tf-tensorboard with Polymer. We can discuss offline.

wchargin:
[...]

PR #2475, opened 2019-07-27
[debugger plugin] Temporary solution to avoid double stamping

caisq:
  * This is a temporary solution to #2437 . It prevents double stamping of tf-debugger-dashboard in DOM, when the plugin is enabled through the command line flag `--debugger_port`.

  * Remove the tf_tensorboard.registerDashboard() call in tf-debugger-dashboard. 
  * In `debugger_plugin_loader.py`, make sure that there is always one plugin returned. This avoids the problem in which the debugger plugin is missing from the dropdown menu when the `--debugger_port` flag is not provided (Thanks to @nfelt and @stephanwlee for the idea).
  * Remove check of grpc import now that grpc is a pip dependency of tensorboard.
  * Simplify the logic of opening the initial dialog and disabled dialog in tf-debugger-dashboard

* Screenshots of UI changes
  * N/A 

  * Manually verified that there is no longer double stamping of tf-debugger-dashboard when the plugin is enabled.
  * Manually verified that the scalar dashboard works correctly with or without the --debugger_port flag.

  * As @stephanwlee suggested, there is no easy and thorough solution to this problem. But the plugin loading system in tf-tensorboard will be redone, according to plan.
  * As @wchargin pointed out, we could implement additional "Not support plugin" type frontend support. But that seems to complex to justify the benefit considering the tf-tensorboard redo that will happen. 
  * As discussed in a design doc, we are considering enabling --debugger_port by default (e.g., default to a port such as 6007. When that happens, this will become a non-issue.

### DIFF
--- a/tensorboard/plugins/debugger/debugger_plugin_loader.py
+++ b/tensorboard/plugins/debugger/debugger_plugin_loader.py
@@ -23,11 +23,29 @@ import sys
 import six
 
 from tensorboard.plugins import base_plugin
-from tensorboard.plugins.debugger import debugger_plugin as debugger_plugin_lib
-from tensorboard.plugins.debugger import interactive_debugger_plugin as interactive_debugger_plugin_lib
+from tensorboard.plugins.debugger import constants
 from tensorboard.util import tb_logging
 
 logger = tb_logging.get_logger()
+
+
+class InactiveDebuggerPlugin(base_plugin.TBPlugin):
+  """A placeholder debugger plugin used when no grpc port is specified."""
+
+  plugin_name = constants.DEBUGGER_PLUGIN_NAME
+
+  def __init__(self):
+    pass
+
+  def is_active(self):
+    return False
+
+  def get_plugin_apps(self):
+    return {}
+
+  def frontend_metadata(self):
+    return super(InactiveDebuggerPlugin, self).frontend_metadata()._replace(
+        element_name='tf-debugger-dashboard')
 
 
 class DebuggerPluginLoader(base_plugin.TBLoader):
@@ -105,6 +123,10 @@ the interactive Debugger Dashboard. This flag is mutually exclusive with
             '  pip install tensorflow')
 
     if flags.debugger_data_server_grpc_port > 0:
+      # pylint: disable=line-too-long,g-import-not-at-top
+      from tensorboard.plugins.debugger import debugger_plugin as debugger_plugin_lib
+      # pylint: enable=line-too-long,g-import-not-at-top
+
       # debugger_data_server_grpc opens the non-interactive Debugger Plugin,
       # which appears as health pills in the Graph Plugin.
       noninteractive_plugin = debugger_plugin_lib.DebuggerPlugin(context)
@@ -112,7 +134,11 @@ the interactive Debugger Dashboard. This flag is mutually exclusive with
                    flags.debugger_data_server_grpc_port)
       noninteractive_plugin.listen(flags.debugger_data_server_grpc_port)
       return noninteractive_plugin
-    else:
+    elif flags.debugger_port > 0:
+      # pylint: disable=line-too-long,g-import-not-at-top
+      from tensorboard.plugins.debugger import interactive_debugger_plugin as interactive_debugger_plugin_lib
+      # pylint: enable=line-too-long,g-import-not-at-top
+
       # If debugger_data_server_grpc_port flag is not specified, we always
       # instantiate the interactive Debugger Plugin. If debugger_port
       # is not specified (i.e., defaults to -1), the frontend will display a
@@ -126,3 +152,5 @@ the interactive Debugger Dashboard. This flag is mutually exclusive with
                     flags.debugger_data_server_grpc_port)
         interactive_plugin.listen(flags.debugger_port)
       return interactive_plugin
+    else:
+      return InactiveDebuggerPlugin()
--- a/tensorboard/plugins/debugger/debugger_plugin_loader.py
+++ b/tensorboard/plugins/debugger/debugger_plugin_loader.py
@@ -21,7 +21,9 @@ from __future__ import print_function
 import sys
 
 import six
+from werkzeug import wrappers
 
+from tensorboard.backend import http_util
 from tensorboard.plugins import base_plugin
 from tensorboard.plugins.debugger import constants
 from tensorboard.util import tb_logging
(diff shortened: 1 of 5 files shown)


stephanwlee:
This will cause regression in user experience. If the double stamping bug hampers your productivity, can you, instead of merging this PR, cherrypick it in your local branch for development only?

caisq:
@stephanwlee Thanks for the comment. Yes, it's true that with this change, users will not see the "debugger" option when the "--debugger_port" is not specified. But that concern is ameliorated by the following facts (some of which is mentioned in the PR description)
1. The result will be better than getting double-stamping in the debugger plugin (when it's activated), and hence have many features of the plugin broken. 
2. The fact that this issue will go away with the rework on the frontend plugin system, as you pointed out.
3. The fact that I'm contemplating activating the --debugger_port by default (as detailed in one of the design docs. I can point it to you offline.)

In sum, this is a worthy tradeoff. This is not to just unblock my development workflow, but also to fix something that's currently broken for users.

caisq:
@stephanwlee Forgot to add: the effect of this change is limited to the debugger plugin. I.e., it doesn't affect other plugins.

stephanwlee:
> the effect of this change is limited to the debugger plugin. I.e., it doesn't affect other plugins.

Yes, was aware of that. 

> The fact that this issue will go away with the rework on the frontend plugin system, as you pointed out.

Given that this will happen relatively soon (before end of August), is it imperative to incorporate this change?

caisq:
> Given that this will happen relatively soon (before end of August), is it imperative to incorporate this change?

My inclination is yes. It'll at least get rid of a step from the developer workflow. Note that I may not be the only person working on the plugin. @davidsoergel is working on it too.

caisq:
@stephanwlee I've revised the PR according to the suggestion you told me (due to @nfelt). 

I manually verified that 
1. There is no more double stamping
2. When the `--debugger_port` is not provided, "Debugger" still shows up in the dropdown menu.

The description of the PR is updated accordingly.

PTAL.

tensorboard/plugins/debugger/debugger_plugin_loader.py
```diff
@@ -116,17 +113,26 @@ def load(self, context):
             '\n\nTo use the debugger plugin, you need to have '
             'gRPC installed:\n  pip install grpcio')
       six.reraise(e_type, e_value, e_traceback)
```
nfelt: Now that the import check isn't guarded by the flag, won't this always fail when `tensorflow` isn't present?  Or `grpcio` in principle, though that's now a `tensorboard` pip package dependency so we could probably skip that check these days.

For that reason I'd expect that this PR would now fail the no-TF startup test on Travis.

caisq: Good point. I moved these import checks under an `if` condition.

tensorboard/plugins/debugger/tf_debugger_dashboard/tf-debugger-dashboard.html
```diff
@@ -858,11 +858,11 @@
           // TODO(cais): If the TDP backend is not activated, display message
           // informing the user of that.
           this.$.initialDialog.openDialog(response.host, response.port);
-        }).catch(error => {
-          this.$.initialDialog.openDisabledDialog();
         }).then(() => {
           // Initiate long-polling.
           this.long_poll();
+        }).catch(error => {
```
stephanwlee: Honestly, this change confused me and I still think it is confusing. Please correct me but it looks like this change is making use of the fact that `openDialog` now throws an exception when either host or port is empty. The flow of thought jumps needlessly across modules and I am not sure if this is good.

Also, we should try to have a principle of having business logic higher up in the component. The InitialDialog (I think the name should be changed too or there should be a separate class that is for disabled state) should be only about presentation of the error state and whether something is valid/invalid should live in this component. If you care more about the separation of concern, I think, then, the InitialDialog should be able to detect the error case and show the disabled dialog itself (which I can see making things more complex).

tensorboard/plugins/debugger/debugger_plugin_loader.py
```diff
@@ -23,6 +23,8 @@
 import six
 
 from tensorboard.plugins import base_plugin
+from tensorboard.plugins.debugger import debugger_plugin as debugger_plugin_lib
+from tensorboard.plugins.debugger import interactive_debugger_plugin as interactive_debugger_plugin_lib
```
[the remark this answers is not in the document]
stephanwlee: Also, would you be able to test this in a virtualenv without TensorFlow? (I think you need to pip install werkzeug numpy absl-py manually)

tensorboard/plugins/debugger/debugger_plugin_loader.py
```diff
@@ -91,42 +125,46 @@ def load(self, context):
...
+
+      # If debugger_data_server_grpc_port flag is not specified, we always
+      # instantiate the interactive Debugger Plugin. If debugger_port
+      # is not specified (i.e., defaults to -1), the frontend will display a
+      # message indicating that the plugin is not active. It'll also display
+      # a command snippet to illustrate how to activate the interactive Debugger
+      # Plugin.
+      interactive_plugin = (
+          interactive_debugger_plugin_lib.InteractiveDebuggerPlugin(context))
+      if flags.debugger_port > 0:
```
stephanwlee: seems like this if statement is not needed because of L151.

wchargin:
Looks like @stephanwlee and @nfelt are on this one, so removing myself.
Please let me know if there’s anything specific that you’d like me to
look at.

tensorboard/plugins/debugger/debugger_plugin_loader.py
```diff
@@ -91,42 +125,46 @@ def load(self, context):
...
+      # Verify that the required Python packages are installed.
+      try:
+        # pylint: disable=g-import-not-at-top,unused-import
+        import tensorflow
+      except ImportError:
+        raise ImportError(
+            'To use the debugger plugin, you need to have TensorFlow installed:\n'
+            '  pip install tensorflow')
+
+    if flags.debugger_data_server_grpc_port > 0:
```
stephanwlee: Question about the behavioral change: it seems like debugger_data_server_grpc_port takes precedence over debugger_port and noninteractive_plugin is chosen when both are passed. Is that desired in terms of UX? Also, did we explicitly mention that in the README or flags help? I don't think it is that obvious just reading the flag help.

tensorboard/plugins/debugger/debugger_plugin_loader.py
caisq: Good catch. Fixed.

tensorboard/plugins/debugger/debugger_plugin_loader.py
caisq: The help message for the flag `--debugger_data_server_grpc_port` says it's mutually exclusive with `--debugger_port`:
https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/debugger/debugger_plugin_loader.py#L53

The logic that checks the mutual exclusion is at:
https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/debugger/debugger_plugin_loader.py#L80

Issue #2466, opened 2019-07-26
Use Prettier


wchargin:
@stephanwlee and I want to run [Prettier][prettier] on TensorBoard.

Prettier is a code formatter for JavaScript, TypeScript, HTML, and CSS
(among others). Its approach is different than that of most formatting
linters, like ESLint: instead of identifying “bad patterns” and
suggesting peephole fixes, it simply parses the entire file into a
concrete syntax tree and re-emits the tree with a pretty-printing
algorithm. This “uncompromising” nature makes it a pleasure to work
with: it’s pretty much entirely insensitive to however the input was
already formatted, and will give you entirely predictable results.

(It’s like Go’s `gofmt`, or Python’s [`black`][black].)

Anecdotally, I noticed that my process of writing code changed fairly
significantly when using Prettier. I can write hideous-looking one-line
monsters, then simply save the file and have it automatically formatted
correctly. (And if it doesn’t reformat on save then I know that I have a
syntax error!) Unlike with ESLint and friends, I can _actually_
completely forget about the formatting.

## Why?

Most importantly, you don’t have to think about formatting while
writing, reviewing, or reading code.

Consistent style also makes diffs smaller, and improves greppability
(e.g., you don’t have to account for two quote forms).

### Style

Prettier’s code style uses a bit more whitespace than our existing code
style. Running Prettier on our codebase increases the total LOC in JS,
TS, HTML, and CSS files by 18%.

Prettier also indents TypeScript namespaces, while we historically have
not, so this adds ~1 level of indentation to all TypeScript files.

This is all totally fine, imho; just wanted to mention it.

### One-shot versus per-file incremental

The diff is fairly large:

```
$ git ls-files -z '*.js' '*.ts' '*.html' '*.css' | xargs -0 wc -l | tail -n 1
  87771 total
$ prettier --write 'tensorboard/**/*.'{js,ts,html,css} >/dev/null 2>/dev/null
$ git ls-files -z '*.js' '*.ts' '*.html' '*.css' | xargs -0 wc -l | tail -n 1
 103474 total
$ git diff --shortstat
 367 files changed, 70988 insertions(+), 55286 deletions(-)
```

Essentially every file is modified, as should be expected. Some tools
will choke when trying to show a diff for a commit this large (e.g.,
[vim-fugitive]). There’s a valid argument that we shouldn’t design
around those broken tools. But Prettier also supports an incremental
per-file option:

```
[...]

PR #2476, opened 2019-07-29
Allow plugins to declare wildcard routes.

davidsoergel:
Currently, plugins must explicitly enumerate the routes that they serve via `get_plugin_apps()`.  This is inconvenient in the case of plugins that provide a set of related routes, such as a directory full of static files, whose names need not be known in advance.  This is the case, for instance, for dynamic plugins using Angular or React, where the build process produces a set of assets (including HTML, JS, CSS, source maps, etc.).

This PR allows a plugin to declare a wildcard route ending in `/*`, e.g. `{'/foo/*': self._foo_wildcard_handler}`.  Any request beginning with the specified path prefix will be passed to the given handler.

Note the handler may do whatever it likes with these requests; they do not necessarily correspond to filesystem paths (as in the static assets use case).  Thus this change also allows for RESTful apis, where object IDs and such are encoded as path elements.  Any security concerns (e.g., regarding paths containing `/../`), are the responsibility of the handler.


tensorboard/backend/application_test.py
```diff
@@ -347,7 +347,7 @@ def _test(self, route, should_be_okay):
             plugin_name='foo',
             is_active_value=True,
             routes_mapping={route: lambda environ, start_response: None}),
-    ]
+      ]
```
wchargin: Please revert (inconsistent indentation, and unrelated change).

tensorboard/backend/application_test.py
```diff
@@ -535,6 +559,30 @@ def testNameToPluginMapping(self):
...
+    self._test_route('', False)
+
+  def testSlashlessRoute(self):
+    self._test_route('runaway', False)
+
+  def testGoodWildcardRoute(self):
+    self._test_route('/data/plugin/bar/wildcard/ok', True)
+  
+  def testBadWildcardRoute(self):
+    self._test_route('/data/plugin/bar/wildcard/bogus', False)
```
wchargin: If I understand correctly, this test is a bit misleading: it suggests
that `…/wildcard/bogus` and `…/wildcard` are handled in the same way,
but they’re not—the former is handled by the plugin, while the latter
never resolves to a plugin app at all. There’s really nothing “bad”
about the `…/wildcard/bogus` route; the plugin handler just chooses to
reject it with the same status code.

What’s the purpose of having the `…/bogus` test case in addition to the
`…/ok` test case, given that they go through exactly the same codepaths
of the code under test?

tensorboard/backend/application.py
```diff
@@ -328,7 +336,8 @@ def __call__(self, environ, start_response):  # pylint: disable=invalid-name
     """Central entry point for the TensorBoard application.
 
     This method handles routing to sub-applications. It does simple routing
-    using regular expression matching.
+    using strict string matching.  Regular expressions are not supported.
+    Wildcard routes such as `/foo/*` are supported as a special case.
```
wchargin: As written, this implementation supports “`wildcard*`” routes even when
the wildcard is not its own path component, and so you could write a
wildcard route matching “`/foo1`”, “`/foo2`”, etc. Is this intentional,
or can we fail fast when a plugin declares a route like `/foo*`?

Separately, any thoughts as to disallowing asterisks in the middle of
routes, like `/foo/*/*` or `/bar/**/baz`, which will behave differently
than the author would expect?

These semantics should be documented on `TBPlugin.get_plugin_apps`,
whatever they end up being.

tensorboard/backend/application.py
```diff
@@ -264,6 +264,14 @@ def __init__(self, plugins, path_prefix=''):
           path = (self._path_prefix + DATA_PREFIX + PLUGIN_PREFIX + '/' +
                   plugin.plugin_name + route)
         self.data_applications[path] = app
+    
+    # Wildcard routes will be checked in the given order, so we sort them
+    # longest to shortest so that a more specific route will take precedence
+    # over a more general one (e.g., a catchall route `/*` should come last).
+    self.prefix_routes = sorted(
+      [x for x in self.data_applications.keys() if x.endswith('*')],
```
wchargin: This `.keys()` is superfluous.

tensorboard/backend/application.py
```diff
@@ -264,6 +264,14 @@ def __init__(self, plugins, path_prefix=''):
           path = (self._path_prefix + DATA_PREFIX + PLUGIN_PREFIX + '/' +
                   plugin.plugin_name + route)
         self.data_applications[path] = app
+    
+    # Wildcard routes will be checked in the given order, so we sort them
+    # longest to shortest so that a more specific route will take precedence
+    # over a more general one (e.g., a catchall route `/*` should come last).
```
wchargin: This seems important; want to test it?

tensorboard/backend/application_test.py
```diff
@@ -506,20 +509,41 @@ def setUp(self):
               FakePlugin,
               plugin_name='bar',
               is_active_value=True,
-              routes_mapping={'/bar_route': self._bar_handler},
+              routes_mapping={
+                '/bar_route': self._bar_handler,
+                '/wildcard/*': self._wildcard_handler
+                },
```
wchargin: nit: Please indent the closing delimiter of containers to the same level as
the line on which the container was opened:

```python
foo(
    routes_mapping={
        # ...
    },
    some_list=[
        # ...
    ],
)
```

tensorboard/backend/application_test.py
```diff
@@ -506,20 +509,41 @@ def setUp(self):
...
 
   def _bar_handler(self):
     pass
 
-  def _construction_callback(self, context):
-    """Called when a plugin is constructed."""
-    self.context = context
+  @wrappers.Request.application
+  def _wildcard_handler(self, request):
+    if(request.path == '/data/plugin/bar/wildcard/ok'):
```
wchargin: nit: Drop parentheses in Python.

tensorboard/backend/application.py
nfelt: +1 to forbidding asterisks anywhere else in the route, and to restricting them to follow a trailing `/` right now unless we have a good reason to allow generic wildcards.

Also, should we be creating a new `self.exact_routes` set for the non-wildcard routes that we can iterate over when we're doing the exact matching?  That way we won't match a request that comes in for a literal `/wildcard/*` path via the exact matching branch, which seems like a semantic error even if it's the same result in practice.

tensorboard/backend/application.py
davidsoergel: +1, added a test--thanks for the reminder!

tensorboard/backend/application.py
davidsoergel: Yep agreed throughout.  Refactored, replacing data_applications entirely with exact_routes and prefix_routes.  Added validation that there is at most one * (at the end only, and requiring a slash), and tests for that.

tensorboard/backend/application_test.py
davidsoergel: Done

tensorboard/backend/application_test.py
davidsoergel: This was really intended more as documentation than as a test per se (i.e., to show that the plugin may yet reject a request matching a wildcard).  Tried to clarify with naming and comments.

tensorboard/backend/application.py
davidsoergel: obsoleted.

tensorboard/backend/application_test.py
davidsoergel: done

tensorboard/backend/application_test.py
davidsoergel: done

tensorboard/backend/application.py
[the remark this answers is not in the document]
davidsoergel: obsoleted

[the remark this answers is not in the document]
davidsoergel: obsoleted

tensorboard/backend/application_test.py
```diff
@@ -506,20 +509,41 @@ def setUp(self):
               FakePlugin,
               plugin_name='bar',
               is_active_value=True,
-              routes_mapping={'/bar_route': self._bar_handler},
+              routes_mapping={
+                '/bar_route': self._bar_handler,
+                '/wildcard/*': self._wildcard_handler
```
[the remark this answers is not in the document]
davidsoergel: done

tensorboard/backend/application.py
```diff
@@ -263,7 +264,25 @@ def __init__(self, plugins, path_prefix=''):
         else:
           path = (self._path_prefix + DATA_PREFIX + PLUGIN_PREFIX + '/' +
                   plugin.plugin_name + route)
-        self.data_applications[path] = app
+
+        if path.endswith('/*'):
+          # Note we remove the '*' but leave the slash in place.
+          path = path[:-1]
+          if '*' in path:
+            raise ValueError('Invalid route: only trailing wildcards are supported (i.e., `/.../*`)')
```
wchargin: nit: Please wrap at 80:

```python
            raise ValueError(
                'Invalid route: only trailing wildcards are supported '
                '(i.e., `/.../*`)'
            )
```

wchargin: Please include the text and plugin name in the error message. (Note that
in the `if`-case you’ll have to re-append the asterisk, or equivalent.)

tensorboard/backend/application.py
```diff
@@ -345,9 +365,14 @@ class are WSGI applications.
     clean_path = _clean_path(parsed_url.path, self._path_prefix)
 
     # pylint: disable=too-many-function-args
-    if clean_path in self.data_applications:
-      return self.data_applications[clean_path](environ, start_response)
+    if clean_path in self.exact_routes:
+      return self.exact_routes[clean_path](environ, start_response)
     else:
+      for path_prefix in self.prefix_routes:
```
wchargin: When I read this block, I look up to see how `self.prefix_routes` was
initialized: it’s initialized to `{}`, an unordered `dict`, so it’s not
clear that this `for`-loop’s iteration order is in fact well defined
because the attribute is later converted to an `OrderedDict`.

Maybe consider refactoring to populate a `raw_prefix_routes` local
variable, or at least add a comment at the original initialization of
`self.prefix_routes` to indicate that its type changes at runtime and
what the contract is?

tensorboard/backend/application.py
```diff
@@ -345,9 +365,14 @@ class are WSGI applications.
...
+        if clean_path.startswith(path_prefix):
+          return self.prefix_routes[path_prefix](environ, start_response)
+        print(clean_path, ' does not start with ', path_prefix)
```
wchargin: If we keep this around, it should be a logging statement, using a format
string so that it does not format the arguments when the message
wouldn’t be printed anyway.

tensorboard/backend/application_test.py
```diff
@@ -535,6 +573,57 @@ def testNameToPluginMapping(self):
...
+    # A wildcard route requires a slash before the '*'.
+    # Lacking one, no route is matched.
+    self._test_route('/data/plugin/bar/wildcard', 404)
+
+  def testEmptyWildcardRouteWithSlash(self):
+    # A wildcard route requires a slash before the '*'.  Here we include the
+    # slash, so we might expect the route to match.
+    #
+    # However: Trailing slashes are automatically removed from incoming requests
+    # in _clean_path().  Consequently, this request does not match the wildcard
```
wchargin: Hmm, interesting. Thanks for including this test case. Is this the
behavior that we actually want? It seems surprising that `/foo/*` does
not match `/foo/`: in glob contexts, `*` always matches the empty
string, so as a plugin author I would not expect this behavior.

The clean-path behavior of always identifying `/foo/` and `/foo` as
user-provided paths is certainly nice. I can see a case for making
pattern `/foo/*` match both paths `/foo/` and `/foo`, and then
disallowing plugins from having conflicting exact and wildcard routes
(e.g., separate `/foo/*` and `/foo/` handlers). How do you think that we
should address this?

tensorboard/backend/application_test.py
wchargin: Much clearer now; thanks!

tensorboard/backend/application_test.py
nfelt: I don't really like the behavior that clean_path has today, FWIW.  Conflating the request that has a trailing slash and the one without one in general is a bit dangerous because it encourages users to treat the URLs interchangeably, but the semantics of relative paths are different across the two URLs at which the content is served.

If we can establish it's safe, I'd be happy to remove that conversion logic or at least change it into an actual redirect so we aren't serving content under both URLs.  Basically, for every `/foo` handler we could create an implicit `/foo/` redirect handler.  We could either then block the plugin having both `/foo` and `/foo/` handlers (to be consistent), or potentially even block all handlers ending in `/`.

wchargin: Oof, I was assuming that we already did have this redirect handler, à la
GitHub Pages, because in that case clean-path helps mitigate exactly the
confusion that you describe. Agreed, then.

tensorboard/backend/application.py
davidsoergel: Oops thanks.  I'm learning how dependent I've been on autoformatting... :P

davidsoergel: done

tensorboard/backend/application.py
davidsoergel: done via unordered_prefix_routes.

tensorboard/backend/application.py
davidsoergel: Sorry, I just overlooked this debugging cruft.

tensorboard/backend/application_test.py
davidsoergel: Yep, makes sense, but can we punt fixing `/foo` vs. `/foo/` more generally to another PR?  If we leave this test in place, then it will fail when the behavior changes, so we won't forget to revisit it.

wchargin: Okay—it’s not optimal, but if this change unblocks Angular/etc. then I’m
willing to defer.

tensorboard/backend/application.py
```diff
@@ -265,7 +266,32 @@ def __init__(self, plugins, path_prefix=''):
...
+                             'trailing wildcards are supported '
+                             '(i.e., `/.../*`)' %
+                             (plugin.plugin_name, path))
+          self.exact_routes[path] = app
+
+    # Wildcard routes will be checked in the given order, so we sort them
+    # longest to shortest so that a more specific route will take precedence
+    # over a more general one (e.g., a catchall route `/*` should come last).
+    self.prefix_routes = collections.OrderedDict(
+        sorted(six.iteritems(unordered_prefix_routes),
```
wchargin: nit: [Google Python style says][1] that when “using a hanging indent of
4 spaces […] there should be nothing after the open parenthesis or
bracket on the first line”.

[1]: https://google.github.io/styleguide/pyguide.html#34-indentation

```python
    self.prefix_routes = collections.OrderedDict(
        sorted(
            six.iteritems(unordered_prefix_routes),
            key=lambda x: len(x[0]),
            reverse=True))
```

tensorboard/backend/application_test.py
```diff
@@ -543,27 +555,53 @@ def setUp(self):
...
+
+  def _bar_handler(self):
+    pass
+
+  @wrappers.Request.application
+  def _wildcard_handler(self, request):
+    if request.path == '/data/plugin/bar/wildcard/ok':
+      return wrappers.Response(response='hello world', status=200)
+    elif request.path == '/data/plugin/bar/wildcard/':
+      # this route cannot actually be hit; see testEmptyWildCardRouteWithSlash.
```
wchargin: nit: caps (`testEmptyWildcardRouteWithSlash`)

tensorboard/backend/application.py
```diff
@@ -265,7 +266,32 @@ def __init__(self, plugins, path_prefix=''):
...
           path = (self._path_prefix + DATA_PREFIX + PLUGIN_PREFIX + '/' +
                   plugin.plugin_name + route)
-        self.data_applications[path] = app
+
+        if path.endswith('/*'):
+          # Note we remove the '*' but leave the slash in place.
+          path = path[:-1]
+          if '*' in path:
+            # note we re-add the removed * in the format string
+            raise ValueError('Plugin %r handles invalid route %r*:  Only '
```
wchargin: This will render as `invalid route '/foo/*/'*`, but the trailing
asterisk should be _inside_ the string.

wchargin: nit: Parallel errors in this file (and generally throughout TensorBoard)
use one space after a period.

tensorboard/backend/application_test.py
```diff
@@ -543,27 +555,53 @@ def setUp(self):
...
-    pass
+    self.server = werkzeug_test.Client(self.app, wrappers.BaseResponse)
 
   def _construction_callback(self, context):
     """Called when a plugin is constructed."""
     self.context = context
 
+  def _test_route(self, route, expected_status_code):
+    response = self.server.get(route)
+    self.assertEquals(response.status_code, expected_status_code)
```
wchargin: The `assertEquals` method is deprecated; prefer `assertEqual`.

tensorboard/backend/application.py
```diff
@@ -265,7 +266,32 @@ def __init__(self, plugins, path_prefix=''):
...
+                             'trailing wildcards are supported '
```
wchargin: Nit, and I’m not requesting a change or blocking you on this, but please
note that some Google style guides explicitly discourage this kind of
horizontal alignment, because it means that changing the first line
forces you to reindent all the other lines for consistency. e.g.:
<https://google.github.io/styleguide/jsguide.html#formatting-horizontal-alignment>

Over time, as large-scale changes happen, necessarily either the blame
becomes polluted or the indentation rots; e.g.:

https://github.com/tensorflow/tensorboard/blob/757628a814410574718623b0a881ae10855754c8/tensorboard/plugins/pr_curve/pr_curve_demo.py#L45-L46

That file changed from `tf.flags` to `tf.compat.v1.flags` to `flags`,
and the alignment was only correct in the original form.

(If we adopt Black, per #2466, then this will be enforced. :-) )

tensorboard/backend/application.py
davidsoergel: done

davidsoergel: done

tensorboard/backend/application.py
davidsoergel: Fair, but I just copied this from your https://github.com/tensorflow/tensorboard/blob/master/tensorboard/backend/application.py#L260 ;)

Anyway +100 to autoformatting.  I appreciate your rigor, but let's let automation handle this stuff :)  Many thanks for already doing the Prettier side of that!

tensorboard/backend/application.py
davidsoergel: done

tensorboard/backend/application_test.py
davidsoergel: done

tensorboard/backend/application_test.py
davidsoergel: done

tensorboard/backend/application_test.py
davidsoergel: OK.  I think it's topically different enough to merit a separate PR in any case.  Tracking at #2502.

PR #2498, opened 2019-08-05
[tensor-widget] add dtype utility and testing infrastructure

caisq:
  * Continue developing tensor-widget, in the new rules_nodejs build infrastructure.

  * Create ts_library targets for tensor_widget
  * Create a jasmine_node_test target
    * Add a small library (dtype-utils.ts) that gets tested by the new jasmine_node_test
  * Replace the old tf_web_library target with a rollup-based tf_web_library
    * Removed no-longer-used tensor-widget.html

* Screenshots of UI changes
  * Not applicable.

  * Ran all bazel tests.


package.json
```diff
@@ -33,7 +33,10 @@
     "@bazel/bazel": "^0.23.2",
     "@bazel/hide-bazel-files": "^0.34.0",
     "@bazel/ibazel": "^0.10.3",
+    "@bazel/jasmine": "^0.32.2",
+    "@bazel/karma": "^0.32.2",
```
stephanwlee: remove me

tensorboard/components/tensor_widget/dtype-utils-test.ts
```diff
@@ -0,0 +1,58 @@
...
+limitations under the License.
+==============================================================================*/
+
+// const {isIntegerDType, isFloatDType} = require('./dtype-utils');
+
+import {isIntegerDType, isFloatDType} from './dtype-utils';
+
+describe('isIntegerDType', () => {
+  it('unsigned ints', () => {
+    expect(isIntegerDType('uint8')).toEqual(true);
```
stephanwlee: please prefer to use chai with its fluent API. `expect(foo).to.be.true`.

tensorboard/components/tensor_widget/BUILD
```diff
@@ -1,18 +1,56 @@
...
+        ":tensor_widget_binary",
+    ]
+)
+
+rollup_bundle(
+    name = "tensor_widget_binary",
+    entry_point = ":tensor-widget.ts",
+    deps = [
+        ":tensor_widget_lib",
+    ]
```
stephanwlee: trailing comma here and above please.

package.json
caisq: Done.

tensorboard/components/tensor_widget/BUILD
caisq: Done.

tensorboard/components/tensor_widget/dtype-utils-test.ts
caisq: Done.

tensorboard/components/tensor_widget/BUILD
```diff
@@ -1,18 +1,58 @@
...
+ts_library(
+    name = "test_lib",
+    testonly = True,
+    srcs = [
+        "dtype-utils-test.ts",
+    ],
+    deps = [
+        ":tensor_widget_lib",
+        "@npm//@types/chai",
+        "@npm//@types/jasmine",
```
stephanwlee: maybe not but you _may_ be able to get rid of this: https://github.com/tensorflow/tensorboard/blob/master/tsconfig-test.json#L4

tensorboard/components/tensor_widget/dtype-utils-test.ts
```diff
@@ -0,0 +1,58 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+import {expect} from 'chai';
+
+import {isIntegerDType, isFloatDType} from './dtype-utils';
+
+describe('isIntegerDType', () => {
+  it('unsigned ints', () => {
```
stephanwlee: Please write a behavior here. We are using BDD style of testing but this seems more like a spec. 
In other words, it should read like English prose: "it returns whatever for foobar baz" instead of "it unsigned ints" doesn't make much sense.

caisq: Done.

tensorboard/components/tensor_widget/BUILD
caisq: I verified that I can't get rid of it. If I removed it, errors like the following will occur:

> tensorboard/components/tensor_widget/dtype-utils-test.ts:20:1 - error TS2593: Cannot find name 'describe'. Do you need to install type definitions for a test runner? Try `npm i @types/jest` or `npm i @types/mocha` and then add `jest` or `mocha` to the types field in your tsconfig.
> 
> 20 describe('isIntegerDType', () => {
>

caisq:
@stephanwlee 

> o minimize the friction, I would defer merging this until prettier change goes in. Does that sound okay?

Following offline discussion, I applied prettier using @wchargin's config file on the two new .ts files added in this PR. There are some changes in one of them and no change in the other one. PTAL.

tensorboard/components/tensor_widget/BUILD
stephanwlee: Alright. Thanks for trying. I wonder why we need the tsconfig-test then :thinking:

tensorboard/components/tensor_widget/dtype-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+==============================================================================*/
+
+/**
+ * Determine if a data type is an integer type.
+ * @param dtype
+ * @return Whether the dtype is an integer type.
+ */
+export function isIntegerDType(dtype: string): boolean {
+  return (
+    dtype.match(/^int[0-9]+$/) != null || dtype.match(/^uint[0-9]+$/) != null
```
stephanwlee: prefer to use triple equals. I believe `match` doesn't return `undefined` so these `!=` can be `!==`, right?

tensorboard/components/tensor_widget/dtype-utils-test.ts
```diff
@@ -0,0 +1,58 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+import {expect} from 'chai';
+
+import {isIntegerDType, isFloatDType} from './dtype-utils';
+
+describe('isIntegerDType', () => {
+  it('returns true for unsigned ints', () => {
```
stephanwlee: thanks for making the changes. 

Just wondering (I know the answer is 'yes', reading the code), but should this function return true for `int0` or `int2131232`?

caisq: Yes, it will. But if there is actually a dtype called `int0`, it should be reasonably categorized as an integer. What other dtype category may make more sense? Same for `int2131232`.

tensorboard/components/tensor_widget/dtype-utils.ts
caisq: Done.

tensorboard/components/tensor_widget/dtype-utils-test.ts
stephanwlee: Adding one assertion around it will make me understand the behavior better :)

caisq: Following offline discussion, I'm adding a few more assertions, but I'm skipping unrealistic cases such as 'int0' for now.

Issue #2324, opened 2019-06-06
version 1.14 tensorboard has the tab “MESH”,but keeps loading with no mesh displayed Ask Question


ufgtb24:
I use tensorboard of version 1.14 which has a new plugin--"mesh", to display points cloud. But nothing can be displayed.

The dashboard maintains blank, but the title of the summary is displayed. I have tried to add as few as possible points into the summary file to exclude the probablity of overload. I also tried to write another summary file with only a scalar inside, then put it into the same directory, and it is  displayed successfully.

## Environment information (required)
Ubuntu 19.04
pip install tf-nightly 

### Diagnostics

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version e093841ffaea564cb2410e0b430bd0c552ada208

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=6, micro=8, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): posix.uname_result(sysname='Linux', nodename='YU', release='5.0.0-15-generic', version='#16-Ubuntu SMP Mon May 6 17:41:33 UTC 2019', machine='x86_64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: True
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
INFO: installed: tb-nightly==1.14.0a20190605
INFO: installed: tf-nightly==1.14.1.dev20190605
INFO: installed: tf-estimator-nightly==1.14.0.dev2019060601

--- check: tensorboard_python_version
INFO: tensorboard.version.VERSION: '1.14.0a20190605'

--- check: tensorflow_python_version
INFO: tensorflow.__version__: '1.14.1-dev20190605'
INFO: tensorflow.__git_version__: 'v1.12.1-3374-g9eb67b17bf'

--- check: tensorboard_binary_path
INFO: which tensorboard: b'/home/yu/anaconda3/envs/display/bin/tensorboard\n'

--- check: readable_fqdn
INFO: socket.getfqdn(): 'YU'

--- check: stat_tensorboardinfo
INFO: directory: /tmp/.tensorboard-info
INFO: os.stat(...): os.stat_result(st_mode=16895, st_ino=4198242, st_dev=2049, st_nlink=2, st_uid=1000, st_gid=1000, st_size=4096, st_atime=1559826037, st_mtime=1559826087, st_ctime=1559826087)
INFO: mode: 0o40777

[...]

PR #2499, opened 2019-08-05
Update mesh-viewer.js reference.

podlipensky:
Fixes issue https://github.com/tensorflow/tensorboard/issues/2324

### DIFF
--- a/third_party/js.bzl
+++ b/third_party/js.bzl
@@ -505,11 +505,11 @@ def tensorboard_js_workspace():
       licenses = ["notice"],  # MIT
       sha256_urls = {
           "76ebbb763969cad7f66fadf24d97a8beec6b6e9c64da568139ad739a1c46ba14": [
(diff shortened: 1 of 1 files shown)


wchargin:
What’s the intended effect of this change? When I build this commit and
point TensorBoard at a directory generated by

```python
import tensorflow as tf
from tensorboard.plugins.mesh import summary as mesh_summary

points = tf.random.normal([1, 20, 3], 0, 1)

meshes_summary = mesh_summary.op("new_points", vertices=points)

writer = tf.summary.FileWriter("logs")

with tf.Session() as sess:
  summaries = sess.run(meshes_summary)
  writer.add_summary(summaries)
```

the point clouds still show up almost imperceptibly faintly. It looks
like each point spans only 1–2 pixels.

wchargin:
Ah—I see now that this is intended to solve the same problem as #2480
rather than the subsequent “faint point clouds” issue. I’ll close in
favor of that one, as it properly updates both URLs for each dependency.

PR #2500, opened 2019-08-05
Move example plugin to plugins_examples

davidsoergel:
This moves the example plugin from `tensorboard/plugins` to the new `tensorboard/plugins_examples`.  The intent is to clarify that the example plugin is not part of a standard TensorBoard installation.  The `plugins_examples` directory also provides a venue for additional examples, e.g. using React or Angular.

The name `plugins_examples` is a little awkward, but has the advantage that it shows up next to `plugins` in directory listings (unlike `example_plugins` or even `plugin_examples`).


davidsoergel:
I was expecting the siblings to be named `example_react`, `example_angular`, etc.  How about `example_basic` for this one? (Or if you prefer, `example_js`, `example_vanilla`, etc.)

wchargin:
Sure, `basic` works for me. (The `example_` prefix seems superfluous,
but if you prefer it, I won’t object.)

nfelt:
The naming `plugins_examples` seems kind of awkward to me - it seems sort of inconsistent to have the prefix `plugins` both generate actual child directories and generate quasi-related sibling directories (and makes tab-completing past this level in the tree more annoying).

Could we perhaps either just name it `examples` (as in `tensorboard/examples/plugins/example_basic`) and leave open the possibility of putting other example code there, or make a real subdirectory `tensorboard/plugins/examples/example_basic`?

davidsoergel:
I do think the plugins should be named with the `example_` prefix, because those names show up in various places on their own (i.e. without the directory structure context).

`tensorboard/examples/plugins/example_basic` works for me; will update the PR shortly if there's no objection from @wchargin or others?

wchargin:
> `tensorboard/examples/plugins/example_basic` works for me; will update
> the PR shortly if there's no objection from @wchargin or others?

Sounds good to me.

tensorboard/examples/plugins/example_basic/smoke_test.sh
```diff
@@ -66,7 +66,8 @@ curl -fs "http://localhost:${port}/data/plugin/example_basic/index.js" >index.js
 diff -u example-plugin/tensorboard_plugin_example/static/index.js index.js
 curl -fs "http://localhost:${port}/data/plugins_listing" >plugins_listing
 cat plugins_listing; printf '\n'
-grep -qP '"example_basic":(?:(?!"enabled").)*+"enabled": *true' plugins_listing
+
+grep -q '"example_basic": {[^}]*"enabled": true' plugins_listing
```
wchargin: This JSON response contains nested fields, like

```json
{
    "example_basic": {
        "loading_mechanism": {
            "type": "IFRAME",
            "module_path": "/data/plugin/example/static/index.js"
        },
        "enabled": true
    }
}
```

and there is no guarantee that the `"enabled": true` needle comes before
the closing brace for the `loading_mechanism`. In particular, we do not
pass `sort_keys=True` to `json.dumps` in `http_util.Respond`.

Presuming that the issue is that you use [a newer version of macOS,
which broke `grep -P`][1], please feel free to replace this with the
equivalent `perl` in a separate commit, but this PR should just be a
move. As a workaround, you can install GNU `grep` or downgrade to a BSD
`grep` with `-P`.

[1]: https://stackoverflow.com/q/16658333

davidsoergel: Ah I see, thanks.  It so happens to come out in a passing order for me (`loading_mechanism` after `enabled`, with nothing else nested anyway), which is how I was able to overlook this :)  Reverted.

FWIW the only reason I was concerned about this at all was that the test failed in CI and I wanted to debug it locally.  The real reason for the CI failure was just a timeout, nothing to do with the grep obviously.  Still we may need to address that too if flakiness is an issue.

Issue #1718, opened 2018-12-22
Reduce usage of tf.compat.v1 APIs


stephanwlee:
TensorBoard have migrated to TF 2.0 APIs but there are some usages of V1 APIs due to not-so-compatible changes. Since we are interested in moving off of `tf.compat.v1`, we will either have to find a replacement or introduce some changes in behavior.

Do note that this list exclude v1 API usages in tests (they are mainly written in graph style and is tracked in #1705) and TB v1 summary.

- [ ] `tf.compat.v1.app` -> `absl.app`
- [ ] `tf.compat.v1.flags` -> `absl.flags`
- [ ] `tf.compat.v1.gfile.GFile` -> `io.gfile.GFile`
- [ ] `tf.compat.v1.gfile.Open` -> `io.gfile.GFile`
- [ ] `tf.compat.v1.image.resize_images` -> `tf.image.resize`
- [ ] `tf.compat.v1.image.resize_nearest_neighbor` -> `tf.image.resize`
- [ ] `tf.compat.v1.MetaGraphDef` -> TB proto
- [ ] `tf.compat.v1.RunMetadata` -> TB proto
- [ ] `tf.compat.v1.Session` -> remove the deadcode
- [ ] `tf.compat.v1.SessionLog` -> TB proto
- [ ] `tf.compat.v1.train.summary_iterator` -> `tf.train.summary_iterator`
- [ ] `tf.contrib.ffmpeg` -> Have to find the replacement

`tf.compat.v1.python_io.tf_record_iterator`, `tf.compat.v1.pywrap_tensorflow.EventsWriter`, `tf.compat.v1.pywrap_tensorflow.NewCheckpointReader`, and `tf.compat.v1.pywrap_tensorflow.PyRecordReader_New` are tracked in #1711

nfelt:
Edited to note that I think `tf.compat.v1.train.summary_iterator` should also be tracked in #1711.  `tf.train.summary_iterator` should go away in TF 2.0, and we should replace it with whatever we resolve #1711 with.

amanp592:
@stephanwlee @nfelt  can I work on the remaining ones, the two mentioned 
tf.compat.v1.Session -> Use eager mode
tf.contrib.ffmpeg -> Have to find the replacement ??

stephanwlee:
I recently learned about https://github.com/tensorflow/io which now includes ffmpeg.

nfelt:
[...]

PR #2507, opened 2019-08-06
ci: discontinue TensorFlow 1.x environment

wchargin:
Summary:
Now that all the 1.x-only tests have been ported or removed (#1705), we
have no need to run tests against TF 1.x. We’re at TensorBoard 2.3 at
this point, and we don’t offer TF 1.x compatibility for future releases.

Test Plan:
Verify that this Travis build creates only two jobs instead of three.

wchargin-branch: tf2x-only


nfelt:
Hmm, one other thought: I'm realizing that turning off this CI effectively renders any TF 1.x-only tests (i.e. tests using `run_v1_only`) dead code.  That's not really ideal; we should preferably clean up the tests before disabling them entirely or we risk just having them hang around forever.

Cleaning up could either mean converting to `tf.compat.v1.*` symbols so they can still run against TF 2.0, if the tests are still useful (see also #1718) or it could mean deleting them entirely.

wchargin:
Okay; that’s reasonable. Cleaning up these tests seems feasible, but
isn’t high priority for me right now, so (per discussion) we’ll leave
this PR alone for now.

wchargin:
Update: TensorFlow and TensorBoard 2.0.0 are out, and nearly all of our
tests work on TensorFlow 2.x. The stragglers are Beholder and the
experimental DB mode (used in the core plugin and scalars only). For
details, see #1705 and comments therein.

stephanwlee:
Is this PR still relevant? If not, let's close it.

wchargin:
Now that Beholder’s gone, we actually have no more `run_v1_only` tests,
so I can resurrect this and kill the TF 1.x dependency (which is no
longer a nightly as of #2769, but still exists).

wchargin:
Yep, that sounds good to me.

PR #2514, opened 2019-08-07
Add TF2.0 migration guide

MarkDaoust:
* Fix Alpha -> Beta in TF2 docs toc.
* Add migration guide: [Colab preview](http://colab.research.google.com/github/MarkDaoust/tensorboard/blob/migrate/docs/r2/migrate.ipynb)


docs/_book.yaml
```diff
@@ -25,8 +25,10 @@ upper_tabs:
     #   skip_translation: true
     #   contents:
     #   - include: /tensorboard/api_docs/python/_toc.yaml
-    - name: TF 2.0 Alpha
+    - name: TF 2.0 Beta
```
nfelt: We should also update the button label on the subsite landing page here:
https://github.com/tensorflow/tensorboard/blob/abd11d6aac6203d2fd3fac11f70192ea0d141a70/docs/_index.yaml#L28

docs/_book.yaml
```diff
@@ -25,8 +25,10 @@ upper_tabs:
...
       contents:
+      - title: "Migrate to TF 2.0"
+        path: /tensorboard/r2/migrate
```
nfelt: Maybe let's name the file `migrate_tf2` so that when we move it up to the top level (as the TF 2.0 docs become the default) it will make more sense as a standalone name?

docs/r2/migrate.ipynb
```diff
@@ -0,0 +1,404 @@
...
+      "outputs": []
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "colab_type": "text",
+        "id": "HEYuO5NFwDK9"
+      },
+      "source": [
+        "# Migrate to TF 2.0\n",
```
nfelt: It seems clearer if this title says something about `tf.summary` or TensorBoard (though it's really pretty much just about `tf.summary`), since otherwise it's not that clear when viewed e.g. in Colab that it's not a generic TF 2.0 migration guide?

Maybe "Migrating tf.summary usage to TF 2.0"?

docs/r2/migrate.ipynb
```diff
@@ -0,0 +1,404 @@
...
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "colab_type": "text",
+        "id": "56V5oun18ZdZ"
+      },
+      "source": [
+        "> Note: This doc is for people who are already familiar with TensroFlow 1.X TensorBoard and who want to migrate large TensorFlow code bases from TensroFlow 1.X to 2.0. If you're new to TensorBoard, see the [get started](get_started.ipynb) doc instead. If you are using `tf.keras` there may be no action you need to take to upgrade to TenorFlow 2.0. \n",
```
nfelt: For consistency, do we have a site-wide standard of formatting `1.x` uppercase or lowercase?  I used lowercase in the prose below so all else being equal I'd stick with that for up here.  Also applies to the section headers `In TF 1.x` and `In TF 2.X` below which should probably match.

nfelt: Couple typos:

TensroFlow 1.X (2 places)
TenorFlow 2.0 (end of line)

docs/r2/migrate.ipynb
```diff
@@ -0,0 +1,404 @@
...
+        "\n",
+        "TensorFlow 2.0 includes significant changes to the tf.summary API used to write summary data for visualization in TensorBoard."
```
nfelt: Maybe let's move this line down to just after the pip install + import tf code boxes?  I feel like people will often skim over those parts so this might help keep this preamble from getting lost.

docs/r2/migrate.ipynb
```diff
@@ -0,0 +1,404 @@
...
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "colab_type": "text",
+        "id": "rh8R2g5FWbsQ"
+      },
+      "source": [
+        "### In TF 2.X\n",
+        "\n",
+        "A new integrated approach is used. Individual `tf.summary` ops write their data immediately when executed.  Using the API from your model code should still look familiar, but it's now friendly to eager execution while remaining graph-mode compatible.  Integrating both halves of the API means the `summary.FileWriter` is now part of the TensorFlow execution context and gets accessed directly by `tf.summary` ops, so configuring writers is the main part that looks different."
```
nfelt: Maybe tweak the initial wording here a little, like

>    The two halves are tightly integrated, and now individual `tf.summary` ops write their data immediately when executed.

docs/r2/migrate.ipynb
```diff
@@ -0,0 +1,404 @@
...
+      },
+      "source": [
+        "writer = tf.summary.create_file_writer(\"/tmp/mylogs/eager\")\n",
+        "\n",
+        "with writer.as_default():\n",
+        "  for step in range(100):\n",
+        "    # other model code would go here\n",
+        "    tf.summary.scalar(\"my_metric\", 0.5, step=step)\n",
+        "    \n",
+        "writer.flush()"
```
nfelt: Can we keep this flush() call indented inside the loop?  That way it's consistent with the other examples.

docs/r2/migrate.ipynb
```diff
@@ -0,0 +1,404 @@
...
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "colab_type": "code",
+        "id": "OyQgeqZhVRNB",
+        "colab": {}
+      },
+      "source": [
+        "with tf.compat.v1.Graph().as_default() as g:\n",
```
nfelt: It's a little weird to me to use the face that `g` exists as a local name outside the `with` block (I guess this is python's behavior, it was new to me though).  Maybe more obvious as 

```python
g = tf.compat.v1.Graph()
with g.as_default():
  ...
with tf.compat.v1.Session(graph=g) as sess:
  ...
```

docs/r2/migrate.ipynb
```diff
@@ -0,0 +1,404 @@
...
+        "colab_type": "text",
+        "id": "1GUZRWSkW3ZC"
+      },
+      "source": [
+        "## Additional tips\n",
+        "\n",
+        "In addition to the critical areas above, some auxiliary aspects have also changed:\n",
+        "\n",
+        "*  Conditional recording (like \"log every 100 steps\") has a new look\n",
+        "    -   To control ops and associated code, wrap them in a regular if statement (which works in eager mode and in `[@tf.function` via autograph](https://www.tensorflow.org/alpha/guide/autograph)) or a `tf.cond`\n",
```
nfelt: Backticks around `@tf.function` are not quite right, the first one should be inside the link bracket otherwise the link is broken

nfelt:
@MarkDaoust just wanted to check on the status of this? It'd be great to get this content up on the site.  I can take care of resolving any of the comments above if you'd like, or if you're too pressed for time one of us could take over the PR.

docs/r2/migrate.ipynb
MarkDaoust: Yes. that does seem to surprise everyone the first time.

MarkDaoust:
We'll see what presubmits say, but this should be all set now.

> https://www.tensorflow.org/beta/guide/effective_tf2#tfmetrics_aggregates_data_and_tfsummary_logs_them

Yeah I should give that one a once over before the site switch.

MarkDaoust:
Okay, I think this is ready to merge.

nfelt:
Done, thank you!

PR #2516, opened 2019-08-07
Use `rel="noopener"` for `target="_blank"` links

wchargin:
Summary:
Links with `target="_blank"` open in a new tab, but by default that new
tab can control the page that opened it via the `window.opener` object.
Using `rel="noopener"` closes this security hole.

Test Plan:
No automated test for now; we tend to catch these in code review, so
this is a backfill. Manual verification:

```
$ git grep -o noopener '*.html' | wc -l
20
$ git grep -o _blank '*.html' | wc -l
20
```

wchargin-branch: rel-noopener

### DIFF
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-projections-panel.html
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-projections-panel.html
@@ -387,7 +387,11 @@ limitations under the License.
           </p>
           <p>
             <iron-icon icon="book" class="book-icon"></iron-icon>
-            <a target="_blank" href="http://distill.pub/2016/misread-tsne/">
+            <a
+              target="_blank"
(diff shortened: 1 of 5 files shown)


tensorboard/plugins/projector/vz_projector/vz-projector-projections-panel.html
```diff
@@ -387,7 +387,11 @@
           </p>
           <p>
             <iron-icon icon="book" class="book-icon"></iron-icon>
-            <a target="_blank" href="http://distill.pub/2016/misread-tsne/">
+            <a
+              target="_blank"
+              href="http://distill.pub/2016/misread-tsne/"
```
stephanwlee: [Note to self] http -> https.

PR #2517, opened 2019-08-07
polymer2: fixed plugin dropdown UI issue

### LINKED ISSUE #2446: Inactive dashboards menu is unstyled
On HEAD (192ab46d), but not in today’s nightly (1.15.0a20190718):

Screenshot of inactive dashboards menu

Note the white (not orange) background and the lowercase selected plugin.

Probably related to Polymer 2 changes?

stephanwlee:
Now that we have real ShadowDom, the input within the paper-input has to
be styled through mulitple layers of webcomponents, correctly. Normally,
we would be using ShadyCSS (CSS mixin did not make it to the spec) and a
mixin API exposed by appropriate WebComponents to achieve that but it is
not viable in this case. Specifically, paper-dropdown-menu allows
styling of paper-input using a mixin but in order to style input within
the paper-input, we need to use another mixin. ShadyCSS explicitly says
the nested mixin is not supported.

Instead of transforming with CSS, we decided to fix the issue by
preprocessing the tabName in JavaScript.

See:
https://github.com/webcomponents/polyfills/tree/master/packages/shadycss#known-issues

Fixes #2446.

### DIFF
--- a/tensorboard/components/tf_tensorboard/tf-tensorboard.html
+++ b/tensorboard/components/tf_tensorboard/tf-tensorboard.html
@@ -321,13 +321,11 @@ limitations under the License.
           -webkit-font-smoothing: antialiased;
           font-size: 14px;
           font-weight: 500;
-          text-transform: uppercase;
         }
         --paper-input-container-label: {
           -webkit-font-smoothing: antialiased;
           font-size: 14px;
           font-weight: 500;
-          text-transform: uppercase;
         }
       }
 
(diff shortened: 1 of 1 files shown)


stephanwlee:
I totally forgot what the plugin dropdown looked like until you kindly pointed it out offline. With the new change, it looks like below:
image

I grepped for all instance of `--paper-menu-` and there were none after this change.

PR #558, opened 2017-09-22
Log dashboard selection if Analytics is enabled

chihuahua:
Added a tf-logging component that contains a logEvent function. This function takes an Analytics event and logs it if Analytics is enabled (as detected by whether the global 'ga' object is defined). Used that method to log dashboard selection.

The vz-projector-app is for now unaffected because that app is a standalone app not used within TensorBoard.

Fixes #524.

Test Plan:
1. Start TensorBoard externally. Note the requests in the Network tab - none are triggered by the `logEvent` method. Also `console.log`ging shows that the `logEvent` method is not called.
2. Start TensorBoard within Google. Note that requests are issued, and `console.log`ing reveals that `logEvent` is called. I think we'll have to wait a few days for the events to show up in the Analytics UI. We can iterate then if we see things that are awry.

PR #2519, opened 2019-08-07
Example dynamic plugin using Angular

davidsoergel:
This provides the Angular 'hello world' app as a TensorBoard plugin.

Everything under `.../frontend` was generated via `ng new frontend`, with a few small exceptions shown in https://github.com/tensorflow/tensorboard/commit/16dab62f3b0d3a0896bec39dc3c0c97013bc201d.

The other interesting bit is the use of an environment variable to provide the path to the built Angular app during development.  See `dev.sh` and `plugin.py`.

The rest is boilerplate, informed by prior work by William and Stephan on the plain-JS and React examples.

TODO in future PR: Bazel build.
TODO in future PR: reduce duplication of README material?
TODO in future PR: investigate hot reload when the Angular app is rebuilt.


tensorboard/examples/plugins/example_angular/README.md
```diff
@@ -0,0 +1,105 @@
+# Developing a TensorBoard plugin
```
stephanwlee: not sure it makes sense to duplicate the readme like this. Can we just move the one in example_basic to `examples/plugins/README.md` and add angular specific lines here?

tensorboard/examples/plugins/example_angular/build.sh
```diff
@@ -0,0 +1,33 @@
...
+echo
+python setup.py bdist_wheel --python-tag py2
+python setup.py bdist_wheel --python-tag py3
+
+echo
+echo "-------------------------------------------------------------------------"
+echo "Installing plugin..."
+echo
+# optional
+pip install dist/tensorboard_plugin_angular-0.0.1-py3-none-any.whl -U
```
stephanwlee: but I use python2! :) 
just echo the appropriate statement (also it becomes brittle since version change can cause this line to fail)?

tensorboard/examples/plugins/example_angular/tensorboard_plugin_angular/frontend/.gitignore
```diff
@@ -0,0 +1,46 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
+
+# compiled output
+/dist
+/tmp
+/out-tsc
+# Only exists if Bazel was run
+/bazel-out
```
stephanwlee: few questions: is this file strictly required? can we merge it with example_angular/.gitignore? Can some of these be included to .gitignore?

davidsoergel:
It's mostly angular CLI, as the description says.

Licensing: good point.  Angular as a whole is MIT-licensed; it didn't occur to me that the license would apply to code generated via `ng new`, but you're right, perhaps it does.  (`ng new` doesn't include any licensing information in the generated code, though, and I can't find any discussion online of this issue).

An alternative approach that I think would solve all of your first four bullets would be to provide only the diff (i.e., instructions, or a script, for using the CLI and then applying our small tweaks to the result to make it work).

tensorboard/examples/plugins/example_angular/README.md
davidsoergel: Yeah that's what I had in mind re the TODO in the description.  Sure, I can roll that change in here.

stephanwlee:
Ah, my eyes often miss the most important information in the PR description. Apologies.

I will would like, for easier review sake, see files that are hand written (on top of your commit you posted on the description; while we are on it, can you clean up the commit history please? it is strictly for review sake).

tensorboard/examples/plugins/example_angular/.gitignore
```diff
@@ -0,0 +1,7 @@
+node_modules
+.DS_Store
```
wchargin: Files specific to particular environments not generated by the project
itself, such as `.DS_Store`, should be in a user-level
`.gitignore_global` file rather than the project’s `.gitignore`.

See: <https://github.com/tensorflow/tensorboard/pull/558#issuecomment-332020237>

davidsoergel:
NP.  Before I address the other comments (which may be obviated), WDYT of the approach of providing only a script that calls `ng new` and applies the diff, rather than checking in generated code?

wchargin:
> a script that calls `ng new` and applies the diff

This is better in terms of code size, but is brittle; the maintainers of
`ng new` are totally within their rights to change the internal
structure of the output at any time. Again, we’d need tests; see my
comment above.

davidsoergel:
The motivation for an Angular example is that making it work required some tweaks (above and beyond the vanilla JS example + `ng new`) that new plugin developers shouldn't have to figure out from scratch.  Angular is a bit privileged because we expect to use it for our own plugins.  I thought providing a React example would be a good idea too, but don't feel strongly about that.

As for brittleness, the version of `ng new` would be fixed by package.json and/or WORKSPACE.  Updating the Angular version might indeed break things.  I agree that we should add a test like `smoke_test.sh` from `example_basic`; I figured that would go in an upcoming PR together with the Bazel build (sorry I forgot to call that out in the description).

tensorboard/examples/plugins/example_angular/.gitignore
davidsoergel: Addressed in #2649

tensorboard/examples/plugins/example_angular/build.sh
davidsoergel: Done in #2649

tensorboard/examples/plugins/example_angular/tensorboard_plugin_angular/frontend/.gitignore
davidsoergel: Done in #2649

davidsoergel:
Abandoning in favor of #2649.

PR #2522, opened 2019-08-08
Make scalars plugin tests mostly TF 2.x–compatible

wchargin:
Summary:
These tests were disabled for two reasons: they used `tf.placeholder`,
and they used `tf.contrib.summary` for DB mode tests. This patch revises
the implementation of the non–DB mode tests to avoid using placeholders,
and enables those tests in all versions of TensorFlow. The DB mode tests
still cannot be run in TF 2.x because `create_db_writer` is not exported
under any public APIs.

Test Plan:
Tests pass in current `tf-nightly` and `tf-nightly-2.0-preview`.

wchargin-branch: scalars-plugin-tests-tf2x

### DIFF
--- a/tensorboard/plugins/scalar/scalars_plugin_test.py
+++ b/tensorboard/plugins/scalar/scalars_plugin_test.py
@@ -37,7 +37,9 @@ from tensorboard.plugins.scalar import summary
 from tensorboard.util import test_util
 
 
-@test_util.run_v1_only('Requires contrib for db writer or uses op.Placeholder')
+tf.compat.v1.enable_eager_execution()
+
+
 class ScalarsPluginTest(tf.test.TestCase):
 
   _STEPS = 9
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/scalar/scalars_plugin_test.py
```diff
@@ -115,29 +116,28 @@ def testRoutesProvided(self):
...
-    summ = tf.compat.v1.summary.merge_all()
-
     subdir = os.path.join(self.logdir, run_name)
     with test_util.FileWriterCache.get(subdir) as writer:
-      writer.add_graph(sess.graph)
       for step in xrange(self._STEPS):
-        feed_dict = {placeholder: [1 + step, 2 + step, 3 + step]}
-        s = sess.run(summ, feed_dict=feed_dict)
-        writer.add_summary(s, global_step=step)
+        placeholder = [1 + step, 2 + step, 3 + step]
```
nfelt: `placeholder` doesn't really make sense as a name for this since it is no longer an actual placeholder

nfelt:
Also, this fixes most of the scalars_plugin_test.py entry for #1705 so we should probably update that when this goes in (we can perhaps break out a separate line for the tests that use the db_writer as a straggler cleanup item).

tensorboard/plugins/scalar/scalars_plugin_test.py
wchargin: Changed to `data`.

PR #2523, opened 2019-08-08
ui: fixed graph causing dropdown-menu bug

### LINKED ISSUE #2403: Polymer 2: hparams weird when you visit graph plugin before
If you visit graph plugin before hparams plugin is stamped, you get below UI.

image

Do notice the large chevron down on the paper-dropdown. 
This problem is unique in Polymer2.

stephanwlee:
repro case: https://next.plnkr.co/edit/4MCqtk5HAo3oTOKa

Unclear exactly why it behaves as it does. We can posit that it is a bug
with how shadyCSS handles the mixin and applies one.

While fixing the bug, vetted the CSS rules and removed unnecessary ones
(e.g., ---paper-font-subhead).

Fixes #2403.

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.html
+++ b/tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.html
@@ -46,6 +46,8 @@ limitations under the License.
           padding: 0;
           color: gray;
         }
+        --iron-icon-width: 15px;
+        --iron-icon-height: 15px;
         --primary-text-color: gray;
         --paper-item-min-height: 30px;
       }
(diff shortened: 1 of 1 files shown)


stephanwlee:
I think I roughly understand the behavior. The moment you have a mixin, say, that adds style through a mixin, it permanently adds a style tag to all new instances to take the variable. 

Variable specification in `.one`
image

iron-icon style changed to take an input from `--paper-dropdown-menu-icon_-_width` and `--paper-dropdown-menu-icon_-_height` (do note that the default is `width: var(--iron-icon-width, 24px);`
image

iron-icon without this mixin
image

Problem: `.two` does not define a variable/mixin like that.

PR #2529, opened 2019-08-08
data: add partial generic data provider base class

wchargin:
Summary:
This commit begins an initiative to enable TensorBoard to consume data
from multiple sources using a single API. Like TensorBoard’s existing
experimental “DB mode”, this opens the door to addressing fundamental
performance limitations. Unlike DB mode, this is not specific to any
particular database implementation.

In the short term, we’ll provide one implementation of these APIs via a
thin layer on top of the existing event multiplexer system. We will
migrate plugins to read data using the new APIs instead of from the
multiplexer directly. At that point, we can begin looking into providing
alternate, more efficient implementations: e.g., an SQLite-backed store
whose state persists across TensorBoard invocations.

For now, we include only the API calls for listing and reading scalar
data. We’ll expand the API surface as we proceed.

Test Plan:
Unit tests included (there’s not much to test).

wchargin-branch: data-provider

### DIFF
--- a/tensorboard/data/provider.py
+++ b/tensorboard/data/provider.py
@@ -26,18 +26,20 @@ import six
 
 @six.add_metaclass(abc.ABCMeta)
 class DataProvider(object):
+  """Interface for reading TensorBoard scalar, tensor, and blob data.
+
+  These APIs are under development and subject to change. For instance,
+  providers may be asked to implement more filtering mechanisms, such as
+  downsampling strategies or domain restriction by step or wall time.
+  """
+
   @abc.abstractmethod
-  def list_scalars(
-      self,
-      experiment_id,
-      owner_plugin,
-      run_tag_filter=None,
-  ):
+  def list_scalars(self, experiment_id, plugin_name, run_tag_filter=None):
     """List metadata about scalar time series.
 
     Args:
       experiment_id: ID of enclosing experiment.
-      owner_plugin: String name of the TensorBoard plugin that created
+      plugin_name: String name of the TensorBoard plugin that created
         the data to be queried. Required.
       run_tag_filter: Optional `RunTagFilter` value. If omitted, all
         runs and tags will be included.
@@ -47,34 +49,27 @@ class DataProvider(object):
     `run_tag_filter`.
 
     Returns:
-      A nested map `d` such that `d[run][tag]` is a `ScalarMetadata`
+      A nested map `d` such that `d[run][tag]` is a `ScalarTimeSeries`
       value.
     """
     pass
 
   @abc.abstractmethod
   def read_scalars(
-      self,
-      experiment_id,
-      owner_plugin,
-      downsample_to=None,
-      run_tag_filter=None,
-      step_filter=None,
+      self, experiment_id, plugin_name, downsample=None, run_tag_filter=None
   ):
     """Read values from scalar time series.
 
     Args:
       experiment_id: ID of enclosing experiment.
-      owner_plugin: String name of the TensorBoard plugin that created
+      plugin_name: String name of the TensorBoard plugin that created
         the data to be queried. Required.
-      downsample_to: Integer number of steps to which to downsample the
+      downsample: Integer number of steps to which to downsample the
         results (e.g., `1000`). Required.
       run_tag_filter: Optional `RunTagFilter` value. If provided, a time
         series will only be included in the result if its run and tag
         both pass this filter. If `None`, all time series will be
         included.
-      step_filter: Optional `StepFilter` value. If `None`, the entire
-        range of steps may be included.
 
     The result will only contain keys for run-tag combinations that
     actually exist, which may not include all entries in the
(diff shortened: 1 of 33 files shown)


tensorboard/data/provider.py
```diff
@@ -0,0 +1,223 @@
...
+_ScalarDatum = collections.namedtuple(
+    "ScalarDatum", ("step", "wall_time", "value")
+)
+class ScalarDatum(_ScalarDatum):
+  """A single datum in a scalar time series for a run and tag.
+
+  Attributes:
+    step: The global step at which this datum occurred; an integer. This
+      is a unique key among data of this time series.
+    wall_time: The real-world time at which this datum occurred, as
```
nfelt: Since we have the opportunity to newly define this field, do we want to stick with float, vs datetime or integer microseconds?  No particularly strong preference on my part but I thought it was worth asking.

tensorboard/data/provider.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+class StepFilter(object):
+  """Filters data in a time series by step."""
+
+  def __init__(self, lower_bound, upper_bound):
+    """Construct a `StepFilter`.
+
+    Negative values for `lower_bound` or `upper_bound` indicate indices
+    from the end of the array, with the same semantics as Python slices.
```
nfelt: It's somewhat confusing to me that here we use Python slicing semantics for what `-1` means here but then for the internal service break this out into a separate field.  If we're fine doing that here, can we just use int64/sint64 on the wire?  The event proto stores steps as `int64` anyway, so if we are only handling nonnegative steps, we only have 63 bits to fit in anyway.

tensorboard/data/provider.py
```diff
@@ -0,0 +1,223 @@
...
+    return (
+        self._resolve_step(max_step, self._lower_bound),
+        self._resolve_step(max_step, self._upper_bound),
+    )
+
+  def _resolve_step(self, max_step, step):
+    if step >= 0:
+      return step
+    else:
+      return max(0, max_step - ~step)
```
nfelt: I had to look up `~` in python because it's so uncommonly used.  I get that it's elegant, but can we avoid making twos-complement a prerequisite to understanding this logic and just do `max_step + (step + 1)` or `max_step - (abs(step) - 1)`?  Or at least add a comment like `# invert step from -1 ... -∞ to 0 ... ∞`

tensorboard/data/provider.py
```diff
@@ -0,0 +1,223 @@
...
+
+  def test(self, run, tag):
+    if self._runs is not None and run not in self._runs:
+      return False
+    if self._tags is not None and tag not in self._tags:
+      return False
+    return True
+
+
+class StepFilter(object):
```
nfelt: Design question, but would we ever want a filter on wall_time or relative time as well?  This could come up if imagine implementing efficient zooming, and supported all 3 x-axis options (not just step).

Doesn't have to be addressed here but thought I'd mention it since maybe we want to leave room for this in the API somewhere, e.g. by making this something more like TimeAxisFilter?

tensorboard/data/provider.py
```diff
@@ -0,0 +1,223 @@
...
+  def list_blob_sequences(self):
+    """Not yet specified."""
+    pass
+
+  def read_blob_sequences(self):
+    """Not yet specified."""
+    pass
+
+
+_ScalarMetadata = collections.namedtuple(
```
nfelt: Thinking about this a little more, maybe we should avoid namedtuple here.  The main reason is that we don't really need the tuple behavior, and having it will encourage uses that will make it hard to add new fields later, specifically if people do:

`max_step, wall_time, metadata = scalar_metadata`

This is expounded on in a bit more length at https://news.ycombinator.com/item?id=15132670.

If we just use `@property` and write out the class, we can still make it read-only so all we're really losing is `__slots__` which we could add by hand (even if it's a bit uglier then).  We don't get a nice repr and equality behavior, but we're already foregoing that for RunTagFilter and StepFilter and that seems fine, right?  And it would be nice if all the objects in the API used a consistent pattern.

tensorboard/data/provider.py
```diff
@@ -0,0 +1,223 @@
...
+
+  @property
+  def runs(self):
+    return self._runs
+
+  @property
+  def tags(self):
+    return self._tags
+
+  def test(self, run, tag):
```
nfelt: Optional, but you might consider making these separate utilities for implementations of the provider API (and perhaps initially private) rather than methods on the classes.  They don't actually use any private fields, and that keeps the API surface more minimal.

tensorboard/data/provider.py
wchargin: Sure, makes sense to me. This also avoids the common string formatting
errors (`"My datum: %s" % my_datum` is an error with `namedtuple`s).

All I really want is a simple data type. I’ll go ahead and write out the
whole thing.

tensorboard/data/provider.py
```diff
@@ -0,0 +1,223 @@
...
+_ScalarMetadata = collections.namedtuple(
+    "ScalarMetadata",
+    ("max_step", "corresponding_wall_time", "summary_metadata"),
+)
+class ScalarMetadata(_ScalarMetadata):
+  """Metadata about a scalar time series for a particular run and tag.
+
+  Attributes:
+    max_step: The largest step for this scalar time series; an integer.
+    corresponding_wall_time: The wall time of the datum with the largest
```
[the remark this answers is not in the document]
wchargin: Yes. I’ll post more details later.

tensorboard/data/provider.py
```diff
@@ -0,0 +1,223 @@
...
+      step in this time series, as `float` seconds since epoch.
+    summary_metadata: A `summary_pb2.SummaryMetadata` value.
```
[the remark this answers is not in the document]
wchargin: I suspected that you might hold that opinion. I think that I agree. Will
change.

tensorboard/data/provider.py
wchargin: I’m somewhat disinclined to use `datetime.datetime` because of the whole
“naïve vs. aware” distinction with poor time zone support in the
standard library, and issues like #2017. See #2054, where we switched
_from_ `datetime` to bare timestamps.

Timestamps represented as “f64 seconds” have full `1e-6` precision
through year 2100, so the raw precision isn’t a big concern. I could be
convinced to use integer microseconds, but afaik that’d be a new pattern
in TensorBoard, so I’m weakly inclined to keep this as is.

tensorboard/data/provider.py
wchargin: That’s probably a good idea. Will do.

tensorboard/data/provider.py
wchargin: Yes. I was actually planning to rename this to `TimeAxisFilter`
specifically, so I’m glad that you agree. I’ll post more details later.

tensorboard/data/provider.py
wchargin: Yes. (See internal threads.)

tensorboard/data/provider.py
wchargin: Fascinating. I thought that this _was_ the natural way to think about
slice indices: the sign bit determines the base (start/end), cleanly
separating the space into two equal-sized(!) parts with complementation
as the bijection. It’s not actually about the numeric values; it’s about
the bits. But apparently this perspective is not shared, so I’ll happily
add a comment.

tensorboard/data/provider.py
```diff
@@ -0,0 +1,223 @@
...
+      value.
+    """
+    pass
+
+  @abc.abstractmethod
+  def read_scalars(
+      self,
+      experiment_id,
+      owner_plugin,
+      downsample_to=None,
```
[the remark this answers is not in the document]
nfelt: Yeah I meant changing it to take a `SamplingStrategy` type, not just a name change.  I like the idea of just making it `downsample` though, and generalizing to a more complex value later.

tensorboard/data/provider.py
nfelt: That's a reasonable argument to stick with double/float64, sounds good to me.

tensorboard/data/provider.py
nfelt: Hah, yes, I did concede the elegance of the bijective approach :)  Comment sounds good.

PR #2530, opened 2019-08-08
data: add multiplexer provider implementation

wchargin:
Summary:
This commit implements the new generic data provider APIs on top of the
existing `plugin_event_multiplexer.EventMultiplexer` store.

Test Plan:
Unit tests included.

wchargin-branch: multiplexer-provider

### DIFF
--- a/tensorboard/backend/event_processing/data_provider.py
+++ b/tensorboard/backend/event_processing/data_provider.py
@@ -75,7 +75,10 @@ class MultiplexerDataProvider(provider.DataProvider):
   def read_scalars(
       self, experiment_id, plugin_name, downsample=None, run_tag_filter=None
   ):
-    del downsample  # ignored for now (see note below)
+    # TODO(@wchargin): Downsampling not implemented, as the multiplexer
+    # is already downsampled. We could downsample on top of the existing
+    # sampling, which would be nice for testing.
+    del downsample  # ignored for now
     index = self.list_scalars(
         experiment_id, plugin_name, run_tag_filter=run_tag_filter
     )
@@ -85,9 +88,6 @@ class MultiplexerDataProvider(provider.DataProvider):
       result[run] = result_for_run
       for (tag, metadata) in six.iteritems(tags_for_run):
         events = self._multiplexer.Tensors(run, tag)
-        # TODO(@wchargin): Downsampling not implemented, as the
-        # multiplexer is already downsampled. We could downsample on top
-        # of the existing sampling, which would be nice for testing.
         result_for_run[tag] = [self._convert_scalar_event(e) for e in events]
     return result
 
--- a/tensorboard/backend/event_processing/data_provider.py
+++ b/tensorboard/backend/event_processing/data_provider.py
@@ -1,0 +1,101 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Bridge from event multiplexer storage to generic data APIs."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import six
+
+from tensorboard.data import provider
+from tensorboard.util import tensor_util
+
+
+class MultiplexerDataProvider(provider.DataProvider):
+  def __init__(self, multiplexer):
+    """Trivial initializer.
+
+    Args:
+      multiplexer: A `plugin_event_multiplexer.EventMultiplexer` (note:
+        not a boring old `event_multiplexer.EventMultiplexer`).
+    """
+    self._multiplexer = multiplexer
+
+  def list_scalars(self, experiment_id, owner_plugin, run_tag_filter=None):
+    del experiment_id  # ignored for now
+    run_tag_content = self._multiplexer.PluginRunToTagToContent(owner_plugin)
+    result = {}
+    if run_tag_filter is None:
+      run_tag_filter = provider.RunTagFilter(runs=None, tags=None)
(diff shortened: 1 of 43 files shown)


tensorboard/backend/event_processing/data_provider.py
```diff
@@ -0,0 +1,101 @@
...
+    index = self._multiplexer.PluginRunToTagToContent(owner_plugin)
+    result = {}
+    if step_filter is None:
+      step_filter = provider.StepFilter(lower_bound=0, upper_bound=-1)
+    if (
+        run_tag_filter is None
+        or run_tag_filter.runs is None
+        or run_tag_filter.tags is None
+    ):
+      raise ValueError("Must provide both run and tag filters.")
```
nfelt: This diverges from the docs on `read_scalars()` right?  Since it requires explicit runs and tags, whereas the docs say

```
 run_tag_filter: Optional `RunTagFilter` value. If provided, a time
        series will only be included in the result if its run and tag
        both pass this filter. If `None`, all time series will be
        included.
```

tensorboard/backend/event_processing/data_provider.py
```diff
@@ -0,0 +1,101 @@
...
+    for run in run_tag_filter.runs:
+      result_for_run = {}
+      for tag in run_tag_filter.tags:
+        if tag not in index.get(run, {}):
+          continue
+        result[run] = result_for_run
+        all_events = self._multiplexer.Tensors(run, tag)
+        max_step = max(all_events, key=lambda event: event.step).step
```
nfelt: It might be worth special casing that just so that we don't need to iterate over the events an extra unnecessary time to get the max step if the upper_bound was -1 (or two extra unnecessary times in the case of the default step filter).

tensorboard/backend/event_processing/data_provider_test.py
```diff
@@ -0,0 +1,162 @@
...
+        owner_plugin=scalar_metadata.PLUGIN_NAME,
+        run_tag_filter=run_tag_filter,
+        step_filter=step_filter,
+        downsample_to=None,  # not yet implemented
+    )
+
+    self.assertItemsEqual(result.keys(), ["polynomials", "waves"])
+    self.assertItemsEqual(result["polynomials"].keys(), ["square", "cube"])
+    self.assertItemsEqual(result["waves"].keys(), ["square", "sine"])
+    sample_series = result["polynomials"]["square"]
```
nfelt: Perhaps we could load data from the underlying multiplexer and verify that this data matches the relevant subset?  Then we wouldn't have to rely only on the spot-checking (but wouldn't have to tediously hardcode the expected summary data output).

tensorboard/backend/event_processing/data_provider_test.py
```diff
@@ -0,0 +1,162 @@
...
+        owner_plugin=scalar_metadata.PLUGIN_NAME,
+        run_tag_filter=base_provider.RunTagFilter(["un"], ["likely"]),
+    )
+    self.assertEqual(result, {})
+
+  def test_read_scalars(self):
+    provider = self.create_provider()
+
+    run_tag_filter = base_provider.RunTagFilter(
+        runs=["waves", "polynomials"], tags=["sine", "square", "cube"]
```
nfelt: Maybe add a test for ignoring runs/tags that aren't valid?

tensorboard/backend/event_processing/data_provider_test.py
```diff
@@ -0,0 +1,162 @@
...
+    self.assertIsInstance(sample, base_provider.ScalarMetadata)
+    self.assertEqual(sample.max_step, 18)
+    # nothing to test for wall time, as it can't be mocked out
+    self.assertIsInstance(sample.summary_metadata, summary_pb2.SummaryMetadata)
+    self.assertEqual(
+        sample.summary_metadata.plugin_data.plugin_name,
+        scalar_metadata.PLUGIN_NAME,
+    )
+
+  def test_list_scalars_filters(self):
```
nfelt: Maybe we should also include a check here for a filter that specifies both runs and tags?

tensorboard/backend/event_processing/data_provider_test.py
wchargin: Done.

tensorboard/backend/event_processing/data_provider_test.py
wchargin: This does ignore _combinations_ of runs and tags that aren’t valid (no
tag `cube` in run `waves`). Added a completely unused tag and run, too.

tensorboard/backend/event_processing/data_provider_test.py
wchargin: Done.

tensorboard/backend/event_processing/data_provider.py
```diff
@@ -34,63 +34,61 @@ def __init__(self, multiplexer):
...
-      experiment_id,
-      owner_plugin,
-      downsample_to=None,
-      run_tag_filter=None,
-      step_filter=None,
+      self, experiment_id, plugin_name, downsample=None, run_tag_filter=None
   ):
-    del experiment_id  # ignored for now
-    index = self._multiplexer.PluginRunToTagToContent(owner_plugin)
+    del downsample  # ignored for now (see note below)
```
nfelt: Optional, but you could just move the TODO down there to here?

tensorboard/backend/event_processing/data_provider_test.py
```diff
@@ -103,45 +113,48 @@ def test_list_scalars_filters(self):
...
+    for run in result:
+      for tag in result[run]:
+        tensor_events = multiplexer.Tensors(run, tag)
+        self.assertLen(result[run][tag], len(tensor_events))
+        for (datum, event) in zip(result[run][tag], tensor_events):
+          self.assertEqual(datum.step, event.step)
+          self.assertEqual(datum.wall_time, event.wall_time)
+          self.assertEqual(
+              datum.value,
+              struct.unpack("<f", event.tensor_proto.tensor_content)[0],
```
nfelt: Optional, but IMO overall simpler if we always parse the tensors using `tensor_util.make_ndarray()` (or at least `np.frombuffer().item()`) even though it adds a dependency.

tensorboard/backend/event_processing/data_provider.py
wchargin: Sure; done.

tensorboard/backend/event_processing/data_provider_test.py
wchargin: Yes, I really should. :-) Done.

PR #2531, opened 2019-08-08
data: add `--generic_data` flag and `TBContext` field

wchargin:
Summary:
This patch adds a `--generic_data` flag that defaults to `auto`; in the
`auto` state, each plugin determines whether its usage of the generic
data APIs is stable enough. The flag can also be set to `true` for
testing and dogfooding, or to `false` as an escape hatch in case some
bugs make it into the plugins deemed stable. Setting the flag to `false`
also skips constructing the data provider object itself.

wchargin-branch: generic-data-flag


tensorboard/plugins/core/core_plugin.py
```diff
@@ -482,6 +482,17 @@ def define_flags(self, parser):
 it receives new data at least once per hour)\
 ''')
 
+    parser.add_argument(
+        '--generic_data',
+        type=lambda s: frozenset(x for x in s.split(',') if x),
+        default=frozenset(),
+        help='''\
+[experimental] Use generic data provider infrastructure for plugins in
```
nfelt: Is there any real need to be able to activate this plugin-by-plugin?  It seems like it will make it more onerous to test (especially as more plugins are converted, since we'll have to remember which ones they are and pass that at the command line) versus having the flag just switch all plugins that support the data provider API to use it.

tensorboard/plugins/core/core_plugin.py
```diff
@@ -482,6 +482,17 @@ def define_flags(self, parser):
...
+this comma-separated list. Currently only has any effect with `--logdir`
+data sources.
+''')
+
+
```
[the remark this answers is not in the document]
wchargin: Thanks; will fix.

tensorboard/plugins/core/core_plugin.py
wchargin: I originally had just `--generic_data=true` as you suggest. But given
that we explicitly plan to stage this as “scalars first, then tensors,
then blob sequences”, I think that it would be nice to be able to flip
defaults individually. Once we’re reasonably confident that the scalars
functionality is working as intended, we can enable that by default (for
aggressive dogfooding) while we start work on the images plugin.

It seems to me like this is a lesson that we can learn from the existing
DB mode, which only has a coarse flag and which is still off by default,
and for which the images plugin just doesn’t work.

I considered having individual `--generic_scalars=true` flags, which we
could operate like Bazel’s `--experimental_*` flags, and which solves
the problem of remembering the whole list, but it seemed a bit involved
to convert all the main plugins to define their own flags in loaders.
I suppose that we could just hack them all into `core_plugin`, and that
wouldn’t be so bad. Thoughts?

nfelt: For the record, in discussion we agreed on using a tri-state `false/auto/true`, defaulting to `auto`, in which `true` forces all plugins with support on, `false` forces them all off (as a failsafe), and `auto` lets each plugin choose whether it's on or off by default.  This way as support within a plugin matures (e.g. for scalars), we can flip it on by default without loosing the ability to both force-disable it and also enable the more experimental plugins for testing.

wchargin: Yep. Done.

tensorboard/plugins/core/core_plugin.py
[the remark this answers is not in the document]
wchargin: Done.

tensorboard/plugins/base_plugin.py
```diff
@@ -172,7 +172,8 @@ def __init__(
           handle this function returns must be closed. It is assumed that you
           will pass this file handle to zipfile.ZipFile. This zip file should
           also have been created by the tensorboard_zip_file build rule.
-      data_provider: Instance of `tensorboard.data.provider.DataProvider`.
+      data_provider: Instance of `tensorboard.data.provider.DataProvider`. May
+        be `None` if `flags.generic_data` is set to `"off"`.
```
nfelt: s/"off"/"false"/ I believe?

wchargin: Good catch; thanks. (Originally had `on`/`auto`/`off` but changed it to
`true`/`auto`/`false` per your comment.)

nfelt: FWIW the comment was just illustrative, on/auto/off is fine if you prefer that (and I guess it avoids sounding like stringified booleans).

wchargin: No preference; I’ll keep it as is.

PR #1534, opened 2018-10-20
Initial tb.summary.v2.scalar() implementation

nfelt:
Based on: #1533 

The new op implementation uses the tf.contrib.summary API.  It also drops the "scalar_summary" suffix on the tag, and drops display_name in favor of an explicit tag parameter (to be implemented since it requires some changes with tf.contrib.summary).  The new pb implementation the same as before except it also drops the "scalar_summary" suffix and display_name to match.

I reworked the tests so there are now 4 separate test cases (v1 op, v1 pb, v2 op, v2 pb) with the same set of test methods and assertions, rather than a "correspondence test" between op and pb versions.  The tests no longer check for display_name - it would have been a pain since the v2 ones don't have that parameter, and we shouldn't be changing the v1 code any more anyway.

Issue #59, opened 2017-06-16
Re-enable explicit control of summary tag


decentralion:
Migrated from https://github.com/tensorflow/tensorflow/issues/6150

The basic issue is:
It used to be possible to precisely control the display name (or tag) of summary data in TensorBoard, because users had direct control over the tag field.

However, this meant that summary code ignored tf name scopes when generating summaries and tags. This was frustrating because it meant the following code was broken:

```
def layer(input):
  W = get_weights_variable(somehow)
  b = get_bias_variable(somehow)
  activation = tf.nn.relu(input * W + b)
  s = tf.summary.histogram("activation_histogram", activation)
  return activation

input = tf.placeholder(...)
layer1 = layer(input)
layer2 = layer(layer1)
...
merged = tf.summary.merge_all()
....
```

This code would be broken because the tag "activation_histogram" would be used twice, creating a conflict.

To fix this, the summary module (which was added in tf1.0) removes explicit control of the tag from the user. Instead, the first argument to the summary is the "name", and the generated node_name is used as the summary tag. As discussed in the linked issue, however, this creates its own set of frustrations and headaches.

I think a robust solution would be to add a display_name argument to summaries, which is used to create a SummaryMetadata that TensorBoard parses. Then, TensorBoard can use the display_name as the default name in tb, and fall back to the tag in case of conflicts.

I'm not sure how to implement this in a way that generalizes to the plugin system, though. I don't want to fix it just for individual plugins. We also need to continue to support the case where users manually generate the summary proto with an explicit tag.

@jart @wchargin  @chihuahua thoughts?

dandelionmane:
Also see:
"Allow dynamic summary names in new summary interface"
https://github.com/tensorflow/tensorflow/issues/6603

wchargin:
I'd like to increase the priority on this as we encourage summary ops to be implemented as high-level TensorFlow code instead of as core TensorFlow ops. Any op that needs to process its input at all will (should) create a `name_scope` for this computation, but then we can't provide a summary op of the same name.

In the meantime, we can use the following workaround: a plugin scopes its computation and creates a summary op named `__internal_image_summary` or something (within the name scope). Then, when the plugin backend looks at any tag, it strips a suffix `/__internal_image_summary`. (Graph below.) But this is certainly not something that we want to actually require plugin authors to do, and it would be great to not have it in the graph at all, so it'd be nice to have this issue resolved before we publish these new APIs.
[...]

PR #2532, opened 2019-08-08
data: add scalars plugin integration

wchargin:
Summary:
This patch teaches the scalars plugin about the new generic data APIs.
These are guarded behind a feature flag (`--generic_data=true`), which
is intended to have no user-visible effect.

Test Plan:
Existing unit tests expanded, reusing the test cases from the existing
multiplexer tests. For manual testing, add logging to verify when the
new APIs are being used:

```diff
diff --git a/tensorboard/plugins/scalar/scalars_plugin.py b/tensorboard/plugins/scalar/scalars_plugin.py
index 41c3db3c..cd086f57 100644
--- a/tensorboard/plugins/scalar/scalars_plugin.py
+++ b/tensorboard/plugins/scalar/scalars_plugin.py
@@ -150,6 +150,9 @@ class ScalarsPlugin(base_plugin.TBPlugin):
   def scalars_impl(self, tag, run, experiment, output_format):
     """Result of the form `(body, mime_type)`."""
     if self._data_provider:
+      from tensorboard.util import tb_logging
+      tb_logging.get_logger().warn("Data provider: %r", self._data_provider)
+
       # Downsample reads to 1000 scalars per time series, which is the
       # default size guidance for scalars under the multiplexer loading
       # logic.
```

Then, run `//tensorboard` pointing at a logdir with scalar data, with
all three values for `--generic_data` (`false`, `auto`, `true`). Verify
that TensorBoard works correctly in all cases, and that the logging
indicates that it is using the new APIs only with `--generic_data=true`.

wchargin-branch: generic-scalars


tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -90,6 +95,22 @@ def frontend_metadata(self):
...
+          experiment_id=None,  # experiment support not yet implemented
+          owner_plugin=metadata.PLUGIN_NAME,
+      )
+      result = {run: {} for run in mapping}
+      for (run, tag_to_content) in six.iteritems(mapping):
+        for (tag, metadatum) in six.iteritems(tag_to_content):
+          sm = metadatum.summary_metadata
+          descr = plugin_util.markdown_to_safe_html(sm.summary_description)
+          result[run][tag] = {
+              'displayName': sm.display_name,
```
nfelt: For new code, it seems better to just use the tag going forward versus maintaining support for the display_name, since it's not in the new summary APIs (because it was not really the right solution to the problem it was introduced to solve).

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -129,7 +150,19 @@ def index_impl(self):
 
   def scalars_impl(self, tag, run, experiment, output_format):
     """Result of the form `(body, mime_type)`."""
-    if self._db_connection_provider:
+    if self._data_provider:
+      all_scalars = self._data_provider.read_scalars(
+          experiment_id=None,  # experiment support not yet implemented
+          owner_plugin=metadata.PLUGIN_NAME,
+          downsample_to=1000,
```
nfelt: Perhaps factor out the constant so we can explain where it comes from?

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -129,7 +150,19 @@ def index_impl(self):
...
+          run_tag_filter=provider.RunTagFilter(runs=[run], tags=[tag]),
+          step_filter=None,  # also the default
+      )
+      scalars = all_scalars.get(run, {}).get(tag, None)
+      if scalars is None:
+        raise KeyError('No scalar data for run=%r, tag=%r' % (run, tag))
```
nfelt: KeyError seems like a bit of a weird exception to continue using as a holdover.  Can we maybe user a better one in the new code and just just pass a tuple of that and KeyError to assertRaises() in the tests?

tensorboard/plugins/scalar/scalars_plugin_test.py
```diff
@@ -108,15 +110,32 @@ def generate_run_to_db(self, experiment_name, run_name):
...
-    """Tests that the plugin offers the correct routes."""
-    self.set_up_with_runs([self._RUN_WITH_SCALARS])
-    routes = self.plugin.get_plugin_apps()
-    self.assertIsInstance(routes['/scalars'], collections.Callable)
-    self.assertIsInstance(routes['/tags'], collections.Callable)
-
-  def generate_run(self, run_name):
-    subdir = os.path.join(self.logdir, run_name)
+  def with_runs(run_names):
+    """Run a test with a bare multiplexer and with a `data_provider`."""
```
nfelt: Maybe add some note about how the plugin will be passed as the first argument to the decorated function?

tensorboard/plugins/scalar/scalars_plugin_test.py
```diff
@@ -108,15 +110,32 @@ def generate_run_to_db(self, experiment_name, run_name):
...
+    def decorator(fn):
+      @functools.wraps(fn)
+      def wrapper(self, *args, **kwargs):
+        (logdir, multiplexer) = self.load_runs(run_names)
+        with self.subTest('bare multiplexer'):
```
nfelt: FWIW, was there a reason you prefer this approach over defining a mixin class with a set of common test methods, and then separate test cases that define different objects under test?  That's the approach I've seen more often (and done more often myself).

tensorboard/plugins/scalar/scalars_plugin.py
wchargin: Sure.

tensorboard/plugins/scalar/scalars_plugin.py
wchargin: Using `KeyError` seems fine with me (“Mapping key not found” is its
docstring), but I suppose that `ValueError` could also be appropriate.

No strong opinion given that the HTTP endpoint _should_ project it down
to a 400 Bad Request anyway (though it currently just 500s).

tensorboard/plugins/scalar/scalars_plugin_test.py
wchargin: Yes; thanks.

tensorboard/plugins/scalar/scalars_plugin_test.py
wchargin: It’s easier for me to reason about how decorators compose than about how
method resolution order works. With subclassing, you also have you think
about where any state should live. No strong preference.

tensorboard/plugins/scalar/scalars_plugin.py
nfelt: It seems fine to me to drop it.  It's in `tf.summary.write` only because we're re-using SummaryMetadata out of convenience.  None of the wrapper APIs allow passing a display_name and just because `write()` accepts one semi-indirectly doesn't mean it actually has to be honored.

I left in `description` since it seemed overall more of a correctly motivated feature (though perhaps still of limited use) and it somewhat softens the blow of removing display_name since then you can always just put anything meant to be human-oriented into the description.

wchargin: Out of curiosity, is there a reference describing why `display_name` is
not considered the right solution to the problem that it was introduced
to solve? Discussions on #59 and the linked TensorFlow issue don’t seem
to raise such a concern.

wchargin: If we do drop `display_name`, then I’d like to drop it across all of
TensorBoard for consistency, and preferably by the 2.0 release.
Otherwise, it’ll be a functional change from `--generic_data=false` to
`--generic_data=true`, and I’d like to minimize those.

(I’ll send a PR out for this.)

tensorboard/plugins/scalar/scalars_plugin.py
wchargin: Changed to `ValueError`.

tensorboard/plugins/scalar/scalars_plugin.py
nfelt: Thanks for pointing out that the deprecation of `display_name` was never really documented anywhere in one place (but rather spread across bits of #1534, https://github.com/tensorflow/tensorflow/commit/92937fb439e102f31c26bd9234e03552499f6ba4, and internal documentation).

I've addressed that by wrapping up #59 with a long-overdue explanation: https://github.com/tensorflow/tensorboard/issues/59#issuecomment-521100225

PR #2539, opened 2019-08-12
infra: fix tomservo and nightly build

stephanwlee:
With the symlink_node_modules (default is True), the macro created a
symlink from $PWD/node_modules into bazel internal directory
(e.g., wherever bazel-tensorboard/external/npm/node_modules leads to).
Because we rm -rf the source directory, the node_modules folder gets removed,
creating a broken symlink. This change makes sure npm install does not create
such symlink but instead creates every folder under the bazel internal
directory.


caisq:
cc myself. Thanks, @stephanwlee

Issue #2492, opened 2019-08-03
Embedding Projector: Add zoom to embedding on [key]+click


beasteers:
The embedding projector can be difficult to navigate and in order to zoom into the sprite of an image, you need to get it centered just right and then you zoom in to find out that you were off by a few degrees. There should be a way to click on a point (maybe while holding `option` or something, idk) and have it zoom and center the sprite that you clicked on.

Also, it would be nice if there was a panel that showed the embedding sprite blown up larger along with sprites for `x` of the nearest embeddings (I'm picturing just a single column of images).

At the moment, it's really hard to navigate image embeddings and it's hardly being helpful. I really want to be able to use it because it is such an amazing tool, but at the moment it's wasting more time than it's worth.

beasteers:
Another nice feature would be to add a slider that increases the size of the sprites in the point cloud.

wchargin:
Hi @beasteers—thanks for the feature requests! These all sound sensible
to me. They’re not going to be at the top of our priority list for a
while, but if anyone would like to contribute a PR then we’d be happy to
take a look. :-)

PR #2543, opened 2019-08-13
added images to nearest neighbors list

beasteers:

This is in reference to something in #2492. I'm working with image data (audio spectrograms, tbe) and I'm trying to visualize them with the embedding projector, but due to the transparency in the sprite textures and them just being small, it makes it difficult to actually make sense of and visualize neighboring points. So I wanted to add some view that allows us to see the images larger, more organized, without the shaders.

I added the sprite images over each item in the nearest neighbor list (that pops up when you select a point).
I also added a checkbox to toggle this feature which will only be shown if a sprite image file is available.
It uses css background-image/position to load the image, and uses the singleImageDim to calculate and give the div the proper aspect ratio.

Sometimes it takes a second to show the images. I would have thought it would be cached from when it was loaded for THREE.js, but it's only a second so it's not too bad. 

One solution could be lazy loading the divs (see [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)), which I considered doing. It's not yet supported for Internet Explorer 8, (which like, why haven't ppl downloaded chrome or firefox or something, anything else.. 😝) so I figured I'd hold off before tying in the polyfill and all that.

I also fixed a bug where the nearest neighbors list wasn't scrolling (that was a problem before I started) - I just added flex params to the containers that needed it.

* Screenshots of UI changes

image

Create a tensorboard run where you send your embeddings to the projector. My `projector_config.pbtxt` looks like this:
```
# projector_config.pbtxt
embeddings {
  tensor_name: "nat_embedding:0"
  metadata_path: "meta.txt"
  sprite {
    image_path: "sprite.png"
    single_image_dim: 256
    single_image_dim: 256
  }
}
```

Open the embedding projector and click on a point. This should open up the nearest neighbors panel, where you should see your sprites visible 🌈

 I still find it difficult to browse the images without zooming out far because they are all vertical and you can only look at a couple at a time, so having some way to expanding to multiple columns would be helpful. The two ways that I could see it working would be:
    - have the panel be expandable (drag the edge)
    - add a modal popup that shows some 6/8/12 column grid which could give us 20-50 on a page depending on how you like your zoom.

There are a few more things that I think we could do:
 - `metadata-info` is a very similar panel, but I haven't duplicated the code to have it handle both. Seeing as they are very similar, I think a good move could be to combine `updateNeighborsList` and `metadataEditorContext` using more general class names, because they are almost the same interface.
 - add an option to use a metadata column to provide "full scale" image links in the case where the sprites are too small. The big concern with this is that we'd definitely need to add lazy loading for this and potentially also decrease the default number of nearest neighbors.

### DIFF
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.html
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.html
@@ -227,10 +227,11 @@ limitations under the License.
         flex-direction: column;
       }
 
-      .results, .nn, .nn-list {
+      .results,
+      .nn,
+      .nn-list {
         flex: 1 0 100px;
       }
-
     </style>
     <div class="container">
       <div class="buttons">
@@ -293,17 +294,25 @@ limitations under the License.
             </div>
           </div>
           <div class="neighbor-image-controls">
-            <paper-checkbox id="neighbor-images-checkbox" class="no-images-available" checked="{{showNeighborImages}}">
-             show images
-             <paper-icon-button icon="help" class="help-icon"></paper-icon-button>
-             <paper-tooltip
-               position="bottom"
-               animation-delay="0"
-               fit-to-visible-bounds
-             >
-               If image data is available, show the images of the nearest neighbors.
-             </paper-tooltip>
-           </paper-checkbox>
+            <paper-checkbox
+              id="neighbor-images-checkbox"
+              class="no-images-available"
+              checked="{{showNeighborImages}}"
+            >
+              show images
+              <paper-icon-button
+                icon="help"
+                class="help-icon"
+              ></paper-icon-button>
+              <paper-tooltip
+                position="bottom"
+                animation-delay="0"
+                fit-to-visible-bounds
+              >
+                If image data is available, show the images of the nearest
+                neighbors.
+              </paper-tooltip>
+            </paper-checkbox>
           </div>
         </div>
         <p>Nearest points in the original space:</p>
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.html
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.html
@@ -127,7 +127,6 @@ limitations under the License.
       .nn-list .neighbor-image,
       .metadata-list .metadata-image {
         width: 100%;
-        display: None;
       }
 
       .nn-list.nn-img-show .neighbor-image,
@@ -160,10 +159,6 @@ limitations under the License.
       .neighbor-image-controls {
         display: flex;
         padding: 0.8em 0.1em;
-      }
-
-      .no-images-available {
-        display: None;
       }
 
       .options a {
@@ -294,25 +289,25 @@ limitations under the License.
             </div>
           </div>
           <div class="neighbor-image-controls">
-            <paper-checkbox
-              id="neighbor-images-checkbox"
-              class="no-images-available"
-              checked="{{showNeighborImages}}"
-            >
-              show images
-              <paper-icon-button
-                icon="help"
-                class="help-icon"
-              ></paper-icon-button>
-              <paper-tooltip
-                position="bottom"
-                animation-delay="0"
-                fit-to-visible-bounds
-              >
-                If image data is available, show the images of the nearest
-                neighbors.
-              </paper-tooltip>
-            </paper-checkbox>
+            <template is="dom-if" if="[[spriteImagesAvailable]]">
+                <paper-checkbox
+                  id="neighbor-images-checkbox"
+                  checked="{{showNeighborImages}}"
+                >
+                  show images
+                  <paper-icon-button
+                    icon="help"
+                    class="help-icon"
+                  ></paper-icon-button>
+                  <paper-tooltip
+                    position="bottom"
+                    animation-delay="0"
+                    fit-to-visible-bounds
+                  >
+                    Show the images of the nearest neighbors.
+                  </paper-tooltip>
+                </paper-checkbox>
+            </template>
           </div>
         </div>
         <p>Nearest points in the original space:</p>
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.html
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.html
@@ -290,23 +290,23 @@ limitations under the License.
           </div>
           <div class="neighbor-image-controls">
             <template is="dom-if" if="[[spriteImagesAvailable]]">
-                <paper-checkbox
-                  id="neighbor-images-checkbox"
-                  checked="{{showNeighborImages}}"
+              <paper-checkbox
+                id="neighbor-images-checkbox"
+                checked="{{showNeighborImages}}"
+              >
+                show images
+                <paper-icon-button
+                  icon="help"
+                  class="help-icon"
+                ></paper-icon-button>
+                <paper-tooltip
+                  position="bottom"
+                  animation-delay="0"
+                  fit-to-visible-bounds
                 >
-                  show images
-                  <paper-icon-button
-                    icon="help"
-                    class="help-icon"
-                  ></paper-icon-button>
-                  <paper-tooltip
-                    position="bottom"
-                    animation-delay="0"
-                    fit-to-visible-bounds
-                  >
-                    Show the images of the nearest neighbors.
-                  </paper-tooltip>
-                </paper-checkbox>
+                  Show the images of the nearest neighbors.
+                </paper-tooltip>
+              </paper-checkbox>
             </template>
           </div>
         </div>
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.html
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.html
@@ -124,6 +124,17 @@ limitations under the License.
         border-left: 1px solid rgba(0, 0, 0, 0.15);
       }
 
+      .nn-list .neighbor-image,
+      .metadata-list .metadata-image {
+        width: 100%;
+        display: None;
+      }
+
+      .nn-list.nn-img-show .neighbor-image,
+      .metadata-list.nn-img-show .metadata-image {
+        display: block;
+      }
+
       .nn-list .neighbor-link:hover,
       .metadata-list .metadata-link:hover {
         cursor: pointer;
@@ -146,6 +157,15 @@ limitations under the License.
         float: right;
       }
 
+      .neighbor-image-controls {
+        display: flex;
+        padding: 0.8em 0.1em;
+      }
+
+      .no-images-available {
+        display: None;
+      }
+
       .options a {
         color: #727272;
         font-size: 13px;
@@ -158,7 +178,7 @@ limitations under the License.
       }
 
       .neighbors {
-        margin-bottom: 30px;
+        margin-bottom: 15px;
       }
 
       .neighbors-options {
@@ -206,6 +226,11 @@ limitations under the License.
         display: flex;
         flex-direction: column;
       }
+
+      .results, .nn, .nn-list {
+        flex: 1 0 100px;
+      }
+
     </style>
     <div class="container">
       <div class="buttons">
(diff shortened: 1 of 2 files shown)


beasteers:
@googlebot I signed it!

wchargin:
Thanks for the PR! Could you please [add the email used for the Git
commits (a gmail.com address) to your GitHub account][1]?

You can run `yarn fix-lint` to fix the lint errors and make the build
happy. (Run just `yarn` first if you haven’t done that yet.)

[1]: https://help.github.com/en/articles/adding-an-email-address-to-your-github-account

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -25,9 +25,18 @@ namespace vz_projector {
       metadataColumn: String,
       numNN: {type: Number, value: 100},
       updateNumNN: Object,
+      spriteMetadata: Object,
+      showNeighborImages: Boolean,
     },
   });
 
+  export type spriteMetadata = {
```
wchargin: nit: Type names are capitalized, like `SpriteMetadata`.

wchargin: Can we keep this type unexported if possible?

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -25,9 +25,18 @@ namespace vz_projector {
       metadataColumn: String,
       numNN: {type: Number, value: 100},
       updateNumNN: Object,
+      spriteMetadata: Object,
```
wchargin: Should this property be `spriteMeta`?

(Or should the field be `spriteMetadata`?)

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -282,12 +327,35 @@ namespace vz_projector {
...
       for (let i = 0; i < neighbors.length; i++) {
         const neighbor = neighbors[i];
 
         const neighborElement = document.createElement('div');
         neighborElement.className = 'neighbor';
 
+        if (spriteImagePath) {
+          var neighborElementImage = document.createElement('div');
+          neighborElementImage.className = 'neighbor-image';
+          neighborElementImage.style.backgroundImage = `url(${spriteImagePath})`;
```
wchargin: Can we escape this URL with `CSS.escape(spriteImagePath)`?

At this point, might as well precompute the whole thing outside the loop:

```javascript
const backgroundImage = `url(${CSS.escape(this.spriteMeta.imagePath)})`;
```

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
beasteers: oops that was a copy/paste oversight

beasteers:
I just got a chance to run through and make the requested changes. Let me know if there's anything else. 

I couldn't get `spriteImagesAvailable` to be both computed and to have an observer (makes sense that it wouldn't, but,) so I'm just assigning it in `metadataChanged` for now.

>If I understand correctly, there’s only one image file (a sprite sheet),
right? so this would need some image processing (probably on the server)
to extract the correct sprite and serve it as a standalone image file?

Yeah I just figured if someone wanted to serve up arbitrary static images (or spin up a flask server 😉) we could use a metadata column as urls to populate those images. It wouldn't have to be anything that tensorboard backend would have to handle. 

My use case would be to expand the temporal context of the frames, but it's not too crucial and is probably pretty niche. The other reason is that the sprite sheet has a maximum size, so I guess we'll see what happens when I play around with larger test sets !!

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.html
```diff
@@ -267,6 +288,27 @@
               <a class="euclidean" href="javascript:void(0);">EUCLIDEAN</a>
             </div>
           </div>
+          <div class="neighbor-image-controls">
+            <template is="dom-if" if="[[spriteImagesAvailable]]">
+              <paper-checkbox
+                id="neighbor-images-checkbox"
```
wchargin: This ID looks unused; can we remove it?

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -332,6 +413,9 @@ namespace vz_projector {
           barElement.appendChild(tickElement);
         }
 
+        if (this.spriteImagesAvailable) {
+          neighborElement.appendChild(neighborElementImage);
```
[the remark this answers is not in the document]
beasteers: I was just trying to keep with the organization of the render code, but yep that works

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -282,12 +339,36 @@ namespace vz_projector {
...
+          neighborElementImage.style.backgroundImage = backgroundImage;
+          neighborElementImage.style.paddingBottom = paddingBottom;
+          neighborElementImage.style.backgroundSize = backgroundSize;
+          const [row, col] = [
+            Math.floor(neighbor.index / nCols),
+            neighbor.index % nCols,
+          ];
+          neighborElementImage.style.backgroundPosition = `${(col /
+            (nCols - 1)) *
+            100}%
```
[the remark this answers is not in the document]
beasteers: yeah that weird thing was done by "prettier" .....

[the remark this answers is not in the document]
wchargin: Yep, sorry. :-) It’s mostly very good, but it has a hard time with
template strings because it can’t add any whitespace in the middle of
them (because they’re actually strings), and it doesn’t do any AST
transformations (`` `foobar` `` to `` `foo` + `bar` ``, e.g.) so
sometimes with template strings specifically you have to help it out to
get the result to actually be pretty.

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -282,12 +339,36 @@ namespace vz_projector {
...
+        var backgroundImage = `url(${CSS.escape(spriteImagePath)})`;
+      }
+
       for (let i = 0; i < neighbors.length; i++) {
         const neighbor = neighbors[i];
 
         const neighborElement = document.createElement('div');
         neighborElement.className = 'neighbor';
 
+        if (this.spriteImagesAvailable) {
```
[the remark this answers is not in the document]
beasteers: yep. oversight. my bad.

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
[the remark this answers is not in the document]
beasteers: god prettier is so annoying..

beasteers:
>but the modal stays visible with a message, “TypeError: Cannot read property 'vector' of undefined”.

That was happening because I was calling findNeighbors in the observer function, but I guess that was getting called before neighbors were initialized (or something) so `vector` was not defined.

I removed the added findNeighbors call. Instead, I'm caching the last neighbors value, and reusing it if `updateNeighborsList()` is called without an argument. If neither have values, then the function will just pop out early.

beasteers:
Question: is there a reason why the neighbors list is being rendered with vanilla javascript and not polymer elements? I realize there's a bit of logic needed beforehand so I can see how it might be tricky, but I feel like the element definitions could be offloaded.

This is my first time with polymer/typescript, I'm more React/Redux folk.

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -282,6 +365,10 @@ namespace vz_projector {
       this.searchBox.message = '';
       const minDist = neighbors.length > 0 ? neighbors[0].dist : 0;
 
+      if (this.spriteImagesAvailable && this.showNeighborImages) {
+        var image_renderer = this.spriteImageRenderer();
```
wchargin: nit: Rename to `imageRenderer` (camel case).

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -269,7 +319,40 @@ namespace vz_projector {
...
 
-    private updateNeighborsList(neighbors: knn.NearestEntry[]) {
+    private spriteImageRenderer() {
+      const spriteImagePath = this.spriteMeta.imagePath;
+      const {aspectRatio, nCols} = this.spriteMeta;
+      const paddingBottom = 100 / aspectRatio + '%';
+      const backgroundSize = `${nCols * 100}% ${nCols * 100}%`;
+      const backgroundImage = `url(${CSS.escape(spriteImagePath)})`;
+
+      return (neighbor: knn.NearestEntry): HTMLElement => {
```
wchargin: Nice; this is clean.

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
wchargin: Okay—this is still hoisted, but it’s just one variable and much clearer,
so I’ll allow it. :+1:

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -127,6 +169,14 @@ namespace vz_projector {
       this.enableResetFilterButton(false);
     }
 
+    _updateNeighborsList() {
+      this.updateNeighborsList();
+    }
+
+    checkSpriteImagesAvailable() {
```
wchargin: Unused; remove?

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -127,6 +169,14 @@ namespace vz_projector {
       this.enableResetFilterButton(false);
     }
 
+    _updateNeighborsList() {
```
wchargin: Can this be inlined?

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
beasteers: too much time in python recently lol

beasteers: hahaha, yeah I'm not sure how to avoid that, other than putting the conditional inside the renderer, but that would make things a lot less clear so,,

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
beasteers: Wait I'm not totally sure what you mean inlined.

This?
```javascript
      showNeighborImages: {
        type: Boolean,
        value: true,
        observer: () => {
          this.updateNeighborsList();
        },
      },
```
this?
```javascript
    _updateNeighborsList() { this.updateNeighborsList(); }
```

Doing this doesn't work because it's not bound to `this` anymore:
```javascript
      showNeighborImages: {
        type: Boolean,
        value: true,
        observer: 'updateNeighborsList',
      },
```

wchargin: Ah, I see what’s happening. By “inlined” I meant “remove the function
`_updateNeighborsList` and replace all calls with `updateNeighborsList`,
which should ‘obviously’ be equivalent”. But the reason that this
doesn’t work is that Polymer invokes observers with the value of the
property, which for `showNeighborImages` is a boolean—so, when the
checkbox is clicked, Polymer calls `_updateNeighborsList(true)`, which
then calls `updateNeighborsList()`. After inlining, Polymer will instead
pass that argument to `updateNeighborsList` directly, but then it’s
interpreted as the value of `neighbors?: knn.NearestEntry[]` instead of
being ignored, which leads to errors.

The implementation is fine, but the above is confusing enough that the
function should have a comment and probably a rename—something like
`_refreshNeighborsList`, with a comment indicating that it specifically
drops all arguments before calling `updateNeighborsList`.

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
wchargin: (looks like this didn’t make it into the patch?)

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
wchargin: Yep, no problem.

wchargin:
Merged! Thanks very much—this is a great feature. :tada:

PR #2556, opened 2019-08-15
data: add `list_runs` API call and implementation

wchargin:
Summary:
The `list_scalars` and `read_scalars` API calls aren’t sufficient for a
fully end-to-end demo with a custom data provider implementation,
because the frontend gets the list of runs to display from `/data/runs`.
This commit adds a corresponding API call and wires it up to the client.

Test Plan:
Unit tests included. To manually verify that the new APIs are being
called, patch the `MultiplexerDataProvider` to include some bogus value:

```diff
diff --git a/tensorboard/backend/event_processing/data_provider.py b/tensorboard/backend/event_processing/data_provider.py
index ef602320..20e43800 100644
--- a/tensorboard/backend/event_processing/data_provider.py
+++ b/tensorboard/backend/event_processing/data_provider.py
@@ -55,7 +55,7 @@ class MultiplexerDataProvider(provider.DataProvider):

   def list_runs(self, experiment_id):
     del experiment_id  # ignored for now
-    return [
+    return [provider.Run("wat", "wot", 0.123)] + [
         provider.Run(
             run_id=run,  # use names as IDs
             run_name=run,
```

…then verify that the frontend displays the (real) runs in the same
order with and without `--generic_data=true`.

wchargin-branch: data-list-runs

### DIFF
--- a/tensorboard/data/provider.py
+++ b/tensorboard/data/provider.py
@@ -151,6 +151,9 @@ class Run(object):
       return False
     return True
 
+  def __hash__(self):
+    return hash((self._run_id, self._run_name, self._start_time))
+
   def __repr__(self):
     return "Run(%s)" % ", ".join((
         "run_id=%r" % (self._run_id,),
--- a/tensorboard/data/provider.py
+++ b/tensorboard/data/provider.py
@@ -37,18 +37,11 @@ class DataProvider(object):
   def list_runs(self, experiment_id):
     """List all runs within an experiment.
 
-    The returned sequence should be sorted by run start time (i.e., the
-    time of the first event in each run), so that as new runs are
-    created under a running TensorBoard the list of runs is append-only,
-    and thus each run's time series color stays constant over time. In
-    case two runs have the same start time, their relative order should
-    be tie-broken deterministically (e.g., by run name).
-
     Args:
       experiment_id: ID of enclosing experiment.
 
     Returns:
-      A sequence of distinct run names, sorted as described above.
+      A collection of `Run` values.
     """
     pass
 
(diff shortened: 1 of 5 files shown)


wchargin:
cc @bileschi

tensorboard/data/provider.py
```diff
@@ -33,6 +33,25 @@ class DataProvider(object):
...
+    created under a running TensorBoard the list of runs is append-only,
+    and thus each run's time series color stays constant over time. In
+    case two runs have the same start time, their relative order should
+    be tie-broken deterministically (e.g., by run name).
+
+    Args:
+      experiment_id: ID of enclosing experiment.
+
+    Returns:
+      A sequence of distinct run names, sorted as described above.
```
stephanwlee: No AI required: of course, not related to your code but I definitely want FE and BE to communicate via ids, not names in the future. WDYT? For local FS based implementation, the name can be the id 🤷‍♂ .

tensorboard/data/provider.py
```diff
@@ -33,6 +33,25 @@ class DataProvider(object):
...
+  @abc.abstractmethod
+  def list_runs(self, experiment_id):
+    """List all runs within an experiment.
+
+    The returned sequence should be sorted by run start time (i.e., the
+    time of the first event in each run), so that as new runs are
+    created under a running TensorBoard the list of runs is append-only,
+    and thus each run's time series color stays constant over time. In
+    case two runs have the same start time, their relative order should
+    be tie-broken deterministically (e.g., by run name).
```
stephanwlee: Is exact tie-breaking logic dictated by each implementation? If so, what is the benefit?

tensorboard/data/provider.py
wchargin: Good question; I also wondered that while writing these docs.

Definitely agree that communication should primarily be ID-based.

This API call probably should return the run names, because the frontend
needs to display them, but it could return `(id, name)` structs. Perhaps
the real place to ask this question is on `RunTagFilter`’s definition,
which is currently specified in terms of run and tag names, and should
probably be changed to run and tag IDs.

To properly make this change, we’d need to plumb the IDs through to the
frontend, so that requests to `/data/plugin/scalars/tags` can specify
the run IDs as query parameters. Just for the purpose of unblocking
internal work, I think that I’d prefer to keep these as run names for
now, though I could be convinced otherwise. These APIs are clearly and
explicitly marked experimental, so I will happily make breaking changes
to them when the time comes.

tensorboard/data/provider.py
stephanwlee: One small benefit I can think of is consistency of UX: whichever TensorBoard you use, it will always show the list of runs in the same order. Any discrepancy can be jarring. I think the frontend should do the sorting as the issue I mentioned is purely presentational though.

wchargin: > I think the frontend should do the sorting

Yeah, I agree; this bothered me a bit, too.

How about this: this API call can return structs with ID, name, and
start time, which is sufficient information for the actual behavior that
we want in the long term, and in the short term can simply be consumed
by sorting on the Python backend and dropping the run IDs?

wchargin:
@stephanwlee: Updated per your comments to make these a little more
future-friendly; PTAL?

PR #2557, opened 2019-08-16
Make audio-related functionality and tests TF 2.x–compatible

wchargin:
Summary:
These tests used `tf.contrib.ffmpeg.encode_audio`, but they can actually
just use `tf.audio.encode_wav`, which has existed since TensorFlow 1.14.
This also makes the encoder itself TF 2.x–compatible.

Makes progress toward #1705.

Test Plan:
Tests pass in both TF 1.x and TF 2.x, and the `run_v1_only` decorators
have been removed:

```
$ git grep run_v1_only '*encoder*_test.py' '*audio*_test.py' | wc -l
0
```

wchargin-branch: audio-tf2x


wchargin:

PR #2558, opened 2019-08-16
Make `plugin_event_accumulator_test` TF 2.x–compatible

wchargin:
Summary:
Depends on #2557 for compatibility of underlying audio functionality.
Other than that, this test just needs to execute a bunch of things in
graph mode. Parts could be converted to use more idiomatic TF 2.x style;
this is a minimal fix.

Makes progress toward #1705.

Test Plan:
Tests pass in both TF 1.x and TF 2.x, and the test file no longer uses
`run_v1_only`.

wchargin-branch: pea-tests-tf2x


wchargin:

PR #2560, opened 2019-08-16
Make mesh plugin tests TF 2.x–compatible

wchargin:
Summary:
These just needed to drop down to graph mode in a few places. The
metadata test actually already worked in TF 2.x! :-)

Makes progress toward #1705.

Test Plan:
Tests pass in both TF 1.x and TF 2.x, and the mesh plugin no longer uses
`run_v1_only`:

```
$ git grep run_v1_only '*mesh*' | wc -l
0
```

wchargin-branch: mesh-tests-tf2x

### DIFF
--- a/tensorboard/plugins/mesh/mesh_plugin_test.py
+++ b/tensorboard/plugins/mesh/mesh_plugin_test.py
@@ -44,7 +44,6 @@ except ImportError:
   import mock  # pylint: disable=g-import-not-at-top,unused-import
 
 
-@tensorboard_test_util.run_v1_only('requires tf.Session')
 class MeshPluginTest(tf.test.TestCase):
   """Tests for mesh plugin server."""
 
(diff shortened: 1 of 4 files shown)


wchargin:

wchargin:
cc @podlipensky

wchargin:
> Even in TF2 eager world, I think these tests are definitely adding
> values!

Compared to no tests, these tests absolutely add value—without dropping
down into graph mode, we couldn’t generate any data to actually tes the
mesh plugin.

If we also have tests that generate data via TF 2.x APIs, then imho the
value of these tests is significantly reduced. The `data_compat` layer
should ideally take care of any incompatibilities between the two format
versions, so the 1.x tests would only serve as extra integration tests
for `data_compat` (which is a very simple and easily tested module).

I’d be happy to remove 1.x-specific code tests we have 2.x equivalents,
but if others object then I wouldn’t push for this.

PR #2563, opened 2019-08-16
hparams: remove remaining dependency on TensorFlow

### LINKED ISSUE #2562: hparams plugin still depends on Tensorflow
From `tensorboard/plugins/hparams/metrics.py`, it still uses the make_ndarray from tensorflow. But this function actually is provided by the `tensorboard/util/tensor_util.py` already.

wchargin:
Summary:
The hparams plugin backend depended on TensorFlow but only used the
`make_ndarray` function, for which we already have a TF-free port in
`tensor_util.py`. This patch switches to that implementation, removing
the TensorFlow dependency and associated loading complications.

Fixes #2562.

Test Plan:
Check that no transitive dependency imports TensorFlow:

```
$ bazel query '
>     labels(
>         srcs,
>         deps(//tensorboard/plugins/hparams:hparams_plugin)
>             - //tensorboard/compat
>     )
> ' --null |
> sed -z -e '/^@/d' -e '/_pb2.py$/d' -e 's!^//!!' -e 's!:!/!g' |
> xargs -0 grep -F -e 'import tensorflow' -e 'from tensorflow'
```

(This query should print no output.)

Then, run `bazel run //tensorboard/pip_package:extract_pip_package`,
install the resulting wheel into a new virtualenv by itself, launch
TensorBoard with hparams data, and note that the plugin works correctly.

wchargin-branch: hparams-notf


tensorboard/plugins/hparams/tf_hparams_dashboard/tf-hparams-dashboard.html
```diff
@@ -55,41 +54,13 @@
...
-            this.$['hparams-main'].reload();
-          }
+          this.$['hparams-main'].reload();
         },
       });
-
-      // TODO(#2338): Remove this, and set up a "no TF" message properly.
-      // Keep this in sync with `frontend_metadata` in
-      // `hparams_plugin.py`.
-      tf_tensorboard.registerDashboard({
```
stephanwlee: :heart_eyes_cat:

wchargin:
> Nice! We'll need to update this internally as well now that we're not using the loader.

<http://cl/263837536>

PR #2564, opened 2019-08-16
data: fix and test `__repr__` methods

wchargin:
Summary:
Hubris…

Test Plan:
Tests pass now. The `ScalarTimeSeries` test would have failed on the
previous code, and the `ScalarDatum` test would have similarly failed
after removing the first return statement in its implementation.

wchargin-branch: data-repr


wchargin:
:kiwi_fruit: :pray: :clap:

PR #2566, opened 2019-08-16
build: remove `//tensorboard:tf_contrib_ffmpeg`

wchargin:
Summary:
As of #2557, we’ve migrated all users off of `tf.contrib.ffmpeg`, so
this BUILD target can be removed.

Test Plan:
Running `git grep contrib.ffmpeg` yields no results, and the only
mentions of `ffmpeg` are in `beholder/video_writing.py` (which shells
out to `ffmpeg(1)` as a subprocess).

wchargin-branch: remove-ffmpeg

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -312,13 +312,6 @@ py_library(
     visibility = ["//visibility:public"],
 )
 
-py_library(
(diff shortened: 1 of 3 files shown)


nfelt:
For posterity, this addresses an item from #1718 (where we include contrib as well as compat v1).

PR #2568, opened 2019-08-16
graph: fixed the node inspector view

stephanwlee:
Source of bug:
1: typo: getSvgDefineableElement -> getSvgDefinableElement
2: wrong switch statement

fixed view:
image


tensorboard/plugins/graph/tf_graph_common/tf-graph-icon.ts
```diff
@@ -74,7 +74,7 @@ namespace tf.graph.icon {
       },
     },
 
-    getSvgDefinableElement(): HTMLElement {
+    getSvgDefineableElement(): HTMLElement {
```
wchargin: Can we fix the call site instead? *Definable* is the correct spelling.

stephanwlee: :astonished:  Definable....? wow

Issue #2573, opened 2019-08-19
Restructure the WSGI app for easier direct use


nfelt:
This issue tracks refactoring the main TensorBoard WSGI app to make it easier to use the app as a standalone WSGI app (rather than only via the `tensorboard` CLI).

The essential stream of work is the "core" stream below, but it leaves behind a lot of cleanup that we should at least partially address as well.

CORE
- [ ] Migrate callers of TensorBoardWSGIApp users to TensorBoardWSGI + explicit `Reload()`
- [ ] Reorganize the logic in standard_tb_wsgi() so that first half creates the multiplexer + launches reloading, second half creates the TBContext and instantiates plugins
- [ ] Factor out second half of standard_tb_wsgi() as a reincarnated TensorBoardWSGIApp entry point that accepts data_provider (and multiplexer as the legacy option)
- [ ] Encapsulate standard_tb_wsgi() multiplexer creation into new setup_multiplexer(flags, plugins) helper (can pass in plugins to allow loaders to customize the size_guidance for the multiplexer)
- [ ] Move multiplexer-related bits into a separate file - setup_multiplexer() along with all the associated stuff:
  - `SIZE_GUIDANCE, tensor_size_guidance_from_flags, parse_event_files_spec, start_reloading_multiplexer, create_sqlite_connection_provider, _get_connect_params, _get_event_file_active_filter`
- [ ] Refactor program.py to call launch_multiplexer() manually, and get rid of standard_tensorboard_wsgi()

CLEANUP: WSGI
- [ ] Create dispatch middleware for mapping WSGI apps to paths
  - some example implementations: [werkzeug](https://github.com/pallets/werkzeug/blob/0.15.0/src/werkzeug/middleware/dispatcher.py), [cheroot](https://github.com/cherrypy/cheroot/blob/9e3cbe0c01555dbd362cdf05d8e824fe8de136b1/cheroot/wsgi.py#L370-L424)
- [ ] Change path_prefix to be application of dispatch middleware to TensorBoardWSGIApp, rather than built-in feature
- [ ] Make each plugin itself a WSGI app, so that their backends can be tested entirely independent of the main TB app, and remove calls to TensorBoardWSGI
  - could do this by implementing __call__ in BasePlugin using dispatch middleware plus get_plugin_apps()?
[...]

PR #2575, opened 2019-08-19
Refactor away TensorBoardWSGIApp in its current form

nfelt:
This removes TensorBoardWSGIApp, a thin wrapper over TensorBoardWSGI that launches the reloading thread.  Instead, we move that logic up into `standard_tensorboard_wsgi()` which is the only non-test caller of TensorBoardWSGIApp.

For tests, we update these to use TensorBoardWSGI directly, and add an explicit `multiplexer.Reload()` call where necessary to maintain the previous behavior.

The only usage I know of outside the TB repo is in the TF Lite plugin which I have a PR out to fix in https://github.com/lintian06/tensorflow-lite-plugin/pull/1, but I'm not blocking on that to land.

This addresses core task 1 in https://github.com/tensorflow/tensorboard/issues/2573, and frees up the name TensorBoardWSGIApp to reuse for a new API to come.

### DIFF
--- a/tensorboard/backend/application_test.py
+++ b/tensorboard/backend/application_test.py
@@ -81,10 +81,10 @@ class FakePlugin(base_plugin.TBPlugin):
   """A plugin with no functionality."""
 
   def __init__(self,
-               context,
-               plugin_name,
-               is_active_value,
-               routes_mapping,
+               context=None,
+               plugin_name='foo',
+               is_active_value=True,
+               routes_mapping={},
                element_name_value=None,
                es_module_path_value=None,
                construction_callback=None):
@@ -138,19 +138,14 @@ class FakePlugin(base_plugin.TBPlugin):
 class ApplicationTest(tb_test.TestCase):
   def setUp(self):
     plugins = [
+        FakePlugin(plugin_name='foo'),
         FakePlugin(
-            None, plugin_name='foo', is_active_value=True, routes_mapping={}),
-        FakePlugin(
-            None,
             plugin_name='bar',
             is_active_value=False,
-            routes_mapping={},
             element_name_value='tf-bar-dashboard',
         ),
         FakePlugin(
-            None,
             plugin_name='baz',
-            is_active_value=True,
             routes_mapping={
                 '/esmodule': lambda req: None,
             },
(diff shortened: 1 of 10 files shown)


tensorboard/backend/application_test.py
```diff
@@ -298,89 +288,73 @@ def testPluginsListing(self):
...
-    self._test('/runs', True)
+    application.TensorBoardWSGI([self._make_plugin('/runs')])
 
   def testWildcardRoute(self):
-    self._test('/foo/*', True)
+    application.TensorBoardWSGI([self._make_plugin('/foo/*')])
 
   def testNonPathComponentWildcardRoute(self):
-    self._test('/foo*', False)
+    with six.assertRaisesRegex(self, ValueError, r'invalid route'):
```
wchargin: Thanks; this is much easier to read.

tensorboard/plugins/projector/projector_plugin_test.py
```diff
@@ -273,9 +273,7 @@ def _SetupWSGIApp(self):
     context = base_plugin.TBContext(
         logdir=self.log_dir, multiplexer=multiplexer)
     self.plugin = projector_plugin.ProjectorPlugin(context)
-    wsgi_app = application.TensorBoardWSGIApp(
-        self.log_dir, [self.plugin], multiplexer, reload_interval=0,
```
wchargin: This had `reload_interval=0`, so if I understand correctly we would
usually need to `AddRunsFromDirectory` and `Reload` manually here to
preserve existing behavior. Is it just that the projector plugin doesn’t
actually query data from the multiplexer except for `PluginAssets`, and
the plugin assets data path is not covered by these tests?

nfelt: Yes, I think that's basically the case. Specifically the tests here just rely on the fallback behavior that checks the root logdir directly (in the case of no event files at all):
https://github.com/tensorflow/tensorboard/blob/15a63688e6e711f80aa62811bcb3761d02be8b3b/tensorboard/plugins/projector/projector_plugin.py#L331-L334

Thanks for pointing this out though since it reminded me I was going to file a bug based on what I determined after looking at this part of the mesh plugin test.

PR #2576, opened 2019-08-19
Reincarnate TensorBoardWSGIApp as new DataProvider-friendly entry point

nfelt:
This reincarnates `TensorBoardWSGIApp()`, which is removed in https://github.com/tensorflow/tensorboard/pull/2575, as a new API for constructing the TB WSGI app using a specific DataProvider implementation (with a deprecated argument for passing in a multiplexer instead, the current behavior that we intend to migrate away from).

To do this, we rearrange a lot of the logic in `standard_tensorboard_wsgi()` so that all three possible paths (regular logdir loading, db import, and db read-only) emit a single multiplexer instance.  To that end, I've changed `DbImportMultiplexer` to implement the read surface (by returning empty data) and introduced a shim `_DbModeMultiplexer` for the db read-only mode just as a conduit for parameters.

This addresses core tasks 2 and 3 in https://github.com/tensorflow/tensorboard/issues/2573, with the first four commits corresponding to task 2 and the next 2 to task 3.

cc @bileschi

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -107,31 +107,20 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
   :type plugin_loaders: list[base_plugin.TBLoader]
   :rtype: TensorBoardWSGI
   """
-  event_file_active_filter = _get_event_file_active_filter(flags)
-  multiplexer = event_multiplexer.EventMultiplexer(
-      size_guidance=DEFAULT_SIZE_GUIDANCE,
-      tensor_size_guidance=tensor_size_guidance_from_flags(flags),
-      purge_orphaned_data=flags.purge_orphaned_data,
-      max_reload_threads=flags.max_reload_threads,
-      event_file_active_filter=event_file_active_filter)
-  if flags.generic_data == 'false':
-    data_provider = None
-  else:
-    data_provider = event_data_provider.MultiplexerDataProvider(multiplexer)
-  loading_multiplexer = multiplexer
+  data_provider = None
+  multiplexer = None
   reload_interval = flags.reload_interval
-  db_uri = flags.db
-  db_connection_provider = None
-  # For DB import mode, create a DB file if we weren't given one.
-  if flags.db_import and not flags.db:
-    tmpdir = tempfile.mkdtemp(prefix='tbimport')
-    atexit.register(shutil.rmtree, tmpdir)
-    db_uri = 'sqlite:%s/tmp.sqlite' % tmpdir
   if flags.db_import:
     # DB import mode.
-    logger.info('Importing logdir into DB at %s', db_uri)
+    db_uri = flags.db
+    # Create a temporary DB file if we weren't given one.
+    if not db_uri:
+      tmpdir = tempfile.mkdtemp(prefix='tbimport')
+      atexit.register(shutil.rmtree, tmpdir)
+      db_uri = 'sqlite:%s/tmp.sqlite' % tmpdir
     db_connection_provider = create_sqlite_connection_provider(db_uri)
-    loading_multiplexer = db_import_multiplexer.DbImportMultiplexer(
+    logger.info('Importing logdir into DB at %s', db_uri)
+    multiplexer = db_import_multiplexer.DbImportMultiplexer(
         db_uri=db_uri,
         db_connection_provider=db_connection_provider,
         purge_orphaned_data=flags.purge_orphaned_data,
@@ -139,7 +128,28 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
   elif flags.db:
     # DB read-only mode, never load event logs.
     reload_interval = -1
-    db_connection_provider = create_sqlite_connection_provider(db_uri)
+    db_connection_provider = create_sqlite_connection_provider(flags.db)
+    multiplexer = _DbModeMultiplexer(flags.db, db_connection_provider)
+  else:
+    # Regular logdir loading mode.
+    multiplexer = event_multiplexer.EventMultiplexer(
+        size_guidance=DEFAULT_SIZE_GUIDANCE,
+        tensor_size_guidance=tensor_size_guidance_from_flags(flags),
+        purge_orphaned_data=flags.purge_orphaned_data,
+        max_reload_threads=flags.max_reload_threads,
+        event_file_active_filter=_get_event_file_active_filter(flags))
+    if flags.generic_data != 'false':
+      data_provider = event_data_provider.MultiplexerDataProvider(multiplexer)
+
+  if reload_interval >= 0:
+    # We either reload the multiplexer once when TensorBoard starts up, or we
+    # continuously reload the multiplexer.
+    path_to_run = parse_event_files_spec(flags.logdir)
+    start_reloading_multiplexer(
+        multiplexer, path_to_run, reload_interval, flags.reload_task)
+
+  db_uri = getattr(multiplexer, 'db_uri', None)
+  db_connection_provider = getattr(multiplexer, 'db_connection_provider', None)
   plugin_name_to_instance = {}
   context = base_plugin.TBContext(
       data_provider=data_provider,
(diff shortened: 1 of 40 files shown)


tensorboard/backend/application.py
```diff
@@ -531,3 +562,58 @@ def _get_event_file_active_filter(flags):
...
+  Args:
+    plugin: A TBPlugin subclass, or a TBLoader instance or subclass.
+
+  Returns:
+    A TBLoader for the given plugin.
+  """
+  if isinstance(plugin, base_plugin.TBLoader):
+    return plugin
+  if issubclass(plugin, base_plugin.TBLoader):
+    return plugin()
```
wchargin: This is a nice QOL change; I know that it’s bitten us in the past. Want
to update `tensorboard/default.py` to use this form for internal
consistency?

tensorboard/backend/application.py
```diff
@@ -531,3 +562,58 @@ def _get_event_file_active_filter(flags):
...
+  if issubclass(plugin, base_plugin.TBPlugin):
+    return base_plugin.BasicLoader(plugin)
+  raise ValueError("Not a TBLoader or TBPlugin subclass: %s" % plugin)
```
wchargin: Should this be `TypeError`? Docs for `ValueError` say, “Inappropriate
argument value (of correct type).”

wchargin: Consider `"…" % (plugin,)` to prevent explosions if `plugin` is a tuple.

tensorboard/backend/application.py
```diff
@@ -107,64 +106,96 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
...
+    # We either reload the multiplexer once when TensorBoard starts up, or we
+    # continuously reload the multiplexer.
+    path_to_run = parse_event_files_spec(flags.logdir)
+    start_reloading_multiplexer(
+        multiplexer, path_to_run, reload_interval, flags.reload_task)
+  return TensorBoardWSGIApp(
+      flags, plugin_loaders, data_provider, assets_zip_provider, multiplexer)
+
+
+def TensorBoardWSGIApp(
```
wchargin: Should we take this opportunity to give this function a better name? Its
casing suggests that it’s a type, but it’s a function; and  it’s not
clear from the name how this differs from `class TensorBoardWSGI`.

Perhaps something like `build_tensorboard_wsgi` connotes its purpose?

tensorboard/backend/application.py
```diff
@@ -107,64 +106,96 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
...
+
+  Args:
+    flags: An argparse.Namespace containing TensorBoard CLI flags.
+    plugins: A list of plugins, which can be provided as TBPlugin subclasses
+        or TBLoader instances or subclasses.
+    assets_zip_provider: See TBContext documentation for more information.
+    data_provider: Instance of `tensorboard.data.provider.DataProvider`. May
+        be `None` if `flags.generic_data` is set to `"false"` in which case
+        `deprecated_multiplexer` must be passed instead.
+    deprecated_multiplexer: Optional instance of EventMultiplexer to use
```
wchargin: Worth noting that this is a `plugin_event_multiplexer.EventMultiplexer`?

tensorboard/backend/application.py
```diff
@@ -531,3 +562,58 @@ def _get_event_file_active_filter(flags):
...
+  def __init__(self, db_uri, db_connection_provider):
+    """Constructor for `_DbModeMultiplexer`.
+
+    Args:
+      db_uri: A URI to the database file in use.
+      db_connection_provider: Provider function for creating a DB connection.
+    """
+    logger.info('_DbModeMultiplexer initializing for %s', db_uri)
+    super(_DbModeMultiplexer, self).__init__()
+    self.db_uri = db_uri
```
wchargin: ~~Why does this serve as a conduit for DB-related parameters? IIUC nothing
uses this curently—the URI and connection provider are always pulled off
the context, not the multiplexer.~~

Oof, I see: there’s a `getattr`, which is why my `grep`s turned up
nothing. What do you think of either explicitly mentioning that these
are read in `TensorBoardWSGIApp`, or changing that logic above to be
something like

```python
db_uri = None
db_connection_provider = None
if isinstance(deprecated_multiplexer, _DbModeMultiplexer):
  db_uri = deprecated_multiplexer.db_uri
  db_connection_provider = deprecated_multiplexer.db_connection_provider
  deprecated_multiplexer = None  # don't pass no-op muxer into `TBContext`
```

to also resolve my gripe about including a non-functional multiplexer in
the context at all?

tensorboard/backend/application.py
```diff
@@ -107,64 +106,96 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
...
+        db_uri=db_uri,
         db_connection_provider=db_connection_provider,
         purge_orphaned_data=flags.purge_orphaned_data,
         max_reload_threads=flags.max_reload_threads)
   elif flags.db:
     # DB read-only mode, never load event logs.
     reload_interval = -1
-    db_connection_provider = create_sqlite_connection_provider(db_uri)
+    db_connection_provider = create_sqlite_connection_provider(flags.db)
+    multiplexer = _DbModeMultiplexer(flags.db, db_connection_provider)
```
wchargin: It’s a little funny for a non-functional multiplexer to exist in the
`TBContext`: currently plugins case on DB mode by saying, “if I have a
DB connection provider, use DB mode, else use multiplexer mode”, but
they could just as well say, “if I have a multiplexer, use multiplexer
mode, else use DB mode”—but that will no longer work. Concern?

(See also comment on `_DbModeMultiplexer.__init__` below.)

nfelt: The problem is basically that I don't trust plugins to all defensively check whether multiplexer is not None.  Some of them do this, but inconsistently.  For example, the projector plugin checks that `self.multiplexer` is not None in only 2 out of 3 uses.  The profile plugin doesn't check at all.

So rather than trying to go through all the existing plugins to ensure they can handle a non-existent multiplexer, I figured I'd ensure that there is always a dummy placeholder multiplexer that just returns empty data instead.

I realize this isn't that clean, but our near-term goal is to refactor away all the multiplexer-calling code anyway, so I was reluctant to spend much time improving it now.

tensorboard/backend/application.py
nfelt: My plan (in #2573, one of the WSGI cleanup items) was to convert this into the actual plugin class and remove TensorBoardWSGI.  This is blocked on updating the plugin tests so they don't need it as a shim, which I was going to do as a cleanup task to avoid blocking this PR.

So that's why I left it named in class style.

tensorboard/backend/application.py
nfelt: See above response about the non-functional multiplexer.

I can change the getattr to two different `isinstance` checks (one per db multiplexer type) if you'd prefer that.  I was not trying to be terribly principled since this code is all going away soon, but I'm happy to make that change.

tensorboard/backend/application.py
nfelt: Sure, but I'll also need to update `program.py` to support this form since that's actually where we do the expansion in the current code, and I'd rather not do that in this PR.

I can either leave this functionality in this path only, or I can back out the commit adding it and do the change to both places and `default.py` in a separate PR.  Any preference?

Down the road of course there should be just one path, but I was going to do more refactoring to pull out the contents of `standard_tensorboard_wsgi()` so that it can be invoked from `program.py` and then passed into `TensorBoardWSGIApp()` from there in a clean way.

wchargin: Ah, didn’t realize that this was an _additional_ path. Up to you.

tensorboard/backend/application.py
wchargin: Got it; that this is all slated to be removed soon changes the equation.
This is fine with me, then.

tensorboard/backend/application.py
wchargin: SGTM.

tensorboard/backend/application.py
nfelt: Done.

tensorboard/backend/application.py
nfelt: Replaced getattrs with isinstance.

tensorboard/backend/application.py
nfelt: I've reverted this change for this PR and will send a followup for this separately.

tensorboard/backend/application.py
nfelt: Done in the followup.

nfelt: Done in the followup.

PR #378, opened 2017-08-17
Restore event_processing APIs

jart:
This is blocking sync. Many teams depend on those APIs internally.

PR #348, opened 2017-08-12
Support new-style scalar summaries

wchargin:
Summary:
This commit adds support for new-style scalar summaries to the plugin
frontend and backend.

Here is a screenshot of the resulting dashboard:
Screenshot

This change is fully compatible with the old data format.

Removing the scalars from the event accumulator required more changes
than usual to test code, because test code often used scalars as a
"dummy event type" (understandably, when it was privileged). I switched
all of those occurrences to use tensors.

With respect to summary display names and titles, the scalar dashboard
is different from others: as it aggregates across runs, it needs to be
prepared for variation in these values. The screenshot above shows how
this is handled; see comments in `tf-scalar-dashboard.html` for a more
precise description. This was made by applying the following change to
the demo script:
```diff
diff --git a/tensorboard/plugins/scalar/scalars_demo.py b/tensorboard/plugins/scalar
/scalars_demo.py
index 142b68f..f9c0d16 100644
--- a/tensorboard/plugins/scalar/scalars_demo.py
+++ b/tensorboard/plugins/scalar/scalars_demo.py
@@ -66,9 +66,9 @@ def run(logdir, run_name,
     temperature = tf.Variable(tf.constant(initial_temperature),
                               name='temperature')
     summary.op('current', temperature,
+               display_name='Temperature &%s' % run_name,
-               display_name='Temperature',
                description='The temperature of the object under '
+               '<%s> simulation, in Kelvins.' % run_name[:20])
-               'simulation, in Kelvins.')

     # Compute how much the object's temperature differs from that of its
     # environment, and track this, too: likewise, as
```

Test Plan:
Run `bazel run //tensorboard/plugins/scalar:scalars_demo` to get data
using new-style scalar summaries. Also grab some old-style scalar
summaries, perhaps by running the version of that target that appeared
before this commit (and changing the logdir to avoid writing into the
same directory). Launch a TensorBoard with a logdir enclosing all of
this data, and verify that it all renders correctly.

wchargin-branch: new-style-scalar-frontend

PR #345, opened 2017-08-11
Support new-style audio summaries

wchargin:
Summary:
This commit adds support for new-style audio summaries to the plugin
frontend and backend.

Here is a screenshot of the resulting dashboard:
Screenshot

This change is fully compatible with the old data format.

Test Plan:
Run `bazel run //tensorboard/plugins/audio:audio_demo` to get data
using new-style audio summaries. Also grab some old-style audio
summaries, perhaps by running the version of that target that appeared
before this commit (and changing the logdir to avoid writing into the
same directory). Launch a TensorBoard with a logdir enclosing all of
this data, and verify that it all renders correctly.

When running `bazel test //tensorboard/...`, be aware that this diff
contains a particularly large number of uses of `tf.compat.as_bytes` and
friends. Thus, it is wise to run this test command in both py2 and py3
virtualenvs.

wchargin-branch: new-style-audio-frontend

PR #2577, opened 2019-08-19
Remove unused vestigial plugin EventMultiplexer APIs

nfelt:
I'm not sure why the plugin EventMultiplexer still has `Scalars()` and `Audio()` methods, but they are unused, and they would never have worked because they invoke the corresponding methods on the plugin EventAccumulator, which doesn't even have those methods.


wchargin:
I’m not sure, either. These were removed in #348 (scalars) and #345
(audio), but #378 seems to have resurrected them both. ¯\\\_(ツ)\_/¯

PR #2578, opened 2019-08-19
Allow use of mesh_demo_v2 with PLY files lacking vertex colors

nfelt:
This makes it somewhat easier to run the demo, since at least in my quick pass over the internet, a decent number of PLY files don't contain vertex colors (like the top result for "teapot ply" on google).

It seems reasonable to support since the color parameter to the summary op itself is optional.

I only added support in the v2 demo since the v1 demo uses graph mode and making the color optional there would be a little more work (and for my purposes, I just want at least one script I can easily use to generate sample summary data).

PTAL @podlipensky

### DIFF
--- a/tensorboard/plugins/mesh/demo_utils.py
+++ b/tensorboard/plugins/mesh/demo_utils.py
@@ -26,12 +26,13 @@ def _parse_vertex(vertex_row):
   """Parses a line in a PLY file which encodes a vertex coordinates.
 
   Args:
-    vertex_row: string with vertex coordinates and color.
+    vertex_row: string with vertex coordinates and optional color.
 
   Returns:
     2-tuple containing a length-3 array of vertex coordinates (as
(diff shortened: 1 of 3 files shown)


wchargin:
For what it’s worth, we do have a sample PLY mesh with color data in the
repository for testing purposes:
https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/mesh/test_data/icosphere.ply

(The mesh data is a little wonky (dupe verts) because I screwed it up
when creating the file, and I’ve been meaning to replace it with a
non-borked version that also shows orientation, but haven’t gotten
around to it. It works just fine, though.)

nfelt:
Thanks for pointing that out, I totally missed that :man_facepalming:  It still seems useful to me to allow non-colored meshes here, but I am happy to change the demo so that the default behavior uses that in-repo file if @podlipensky thinks that's reasonable.

PR #2580, opened 2019-08-19
data: plumb experiment ID through runs and scalars

wchargin:
Summary:
This implements a rudimentary experiment selection mechanism for the
runs selector and scalars dashboard only, for the purposes of testing
data provider implementations. The experiment ID is stored as a query
parameter, which isn’t perfect because it precludes caching the big
TensorBoard HTML blob across experiments. In the long term, we can
consider refactoring to serve the big HTML blob from a static path
that’s fetched by the page.

The TensorBoard codebase still has some fragments of a previous
experimental data selector, which was partially removed in #2290, and so
parts of this diff look like changes but are functionally additions.

Test Plan:
Instrument the multiplexer data provider to log experiment IDs:

```diff
diff --git a/tensorboard/backend/event_processing/data_provider.py b/tensorboard/backend/event_processing/data_provider.py
index ef602320..4c72d096 100644
--- a/tensorboard/backend/event_processing/data_provider.py
+++ b/tensorboard/backend/event_processing/data_provider.py
@@ -54,7 +54,7 @@ class MultiplexerDataProvider(provider.DataProvider):
       return None

   def list_runs(self, experiment_id):
-    del experiment_id  # ignored for now
+    logger.warn("Listing runs for experiment %r", experiment_id)
     return [
         provider.Run(
             run_id=run,  # use names as IDs
@@ -65,7 +65,7 @@ class MultiplexerDataProvider(provider.DataProvider):
     ]

   def list_scalars(self, experiment_id, plugin_name, run_tag_filter=None):
-    del experiment_id  # ignored for now
+    logger.warn("Listing scalars for experiment %r", experiment_id)
     run_tag_content = self._multiplexer.PluginRunToTagToContent(plugin_name)
     result = {}
     if run_tag_filter is None:
@@ -96,6 +96,7 @@ class MultiplexerDataProvider(provider.DataProvider):
   def read_scalars(
       self, experiment_id, plugin_name, downsample=None, run_tag_filter=None
   ):
+    logger.warn("Reading scalars for experiment %r", experiment_id)
     # TODO(@wchargin): Downsampling not implemented, as the multiplexer
     # is already downsampled. We could downsample on top of the existing
     # sampling, which would be nice for testing.
```

Then launch TensorBoard with `--generic_data=true` and navigate to
<http://localhost:6006/?experiment=foo>; verify that all three varieties
of server logs are exclusively for the correct experiment ID.

Then, remove the query parameter, and ensure that TensorBoard still
works normally (with experiment ID the empty string).

wchargin-branch: data-experiment

### DIFF
--- a/tensorboard/components/tf_backend/router.ts
+++ b/tensorboard/components/tf_backend/router.ts
@@ -22,7 +22,7 @@ namespace tf_backend {
       params?: URLSearchParams
     ) => string;
     pluginsListing: () => string;
-    runs: () => string;
+    runs: (experiment?: string) => string;
     runsForExperiment: (id: tf_backend.ExperimentId) => string;
   }
 
@@ -53,7 +53,11 @@ namespace tf_backend {
         );
       },
       pluginsListing: () => createDataPath(dataDir, '/plugins_listing'),
-      runs: () => createDataPath(dataDir, '/runs'),
+      runs: (experiment?: string) => {
+        const searchParams = new URLSearchParams();
+        searchParams.set('experiment', experiment || '');
+        return createDataPath(dataDir, '/runs', searchParams);
+      },
       runsForExperiment: (id) => {
         return createDataPath(
           dataDir,
--- a/tensorboard/components/tf_backend/router.ts
+++ b/tensorboard/components/tf_backend/router.ts
@@ -76,6 +76,13 @@ namespace tf_backend {
   }
 
   /**
+   * @return {string} the experiment ID for the currently loaded page
+   */
+  export function getExperimentId() {
+    return new URLSearchParams(window.location.search).get('experiment') || '';
+  }
+
+  /**
    * Set the global router, to be returned by future calls to `getRouter`.
    * You may wish to invoke this if you are running a demo server with a
    * custom path prefix, or if you have customized the TensorBoard backend
--- a/tensorboard/components/tf_backend/runsStore.ts
+++ b/tensorboard/components/tf_backend/runsStore.ts
@@ -17,19 +17,13 @@ namespace tf_backend {
     private _runs: string[] = [];
 
     load() {
-      const url = getRouter().runs(this._getExperimentId());
+      const url = tf_backend.getRouter().runs(tf_backend.getExperimentId());
       return this.requestManager.request(url).then((newRuns) => {
         if (!_.isEqual(this._runs, newRuns)) {
           this._runs = newRuns;
           this.emitChange();
         }
       });
-    }
-
-    _getExperimentId() {
-      return (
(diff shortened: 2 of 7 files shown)


tensorboard/components/tf_backend/router.ts
```diff
@@ -22,7 +22,7 @@ namespace tf_backend {
       params?: URLSearchParams
     ) => string;
     pluginsListing: () => string;
-    runs: () => string;
+    runs: (experiment?: string) => string;
     runsForExperiment: (id: tf_backend.ExperimentId) => string;
```
wchargin: This machinery is unrelated; I’ll remove it in a future change.

I didn’t want to reuse it because it assumes some things about both
inputs (e.g., `type ExperimentId = number`) and outputs (the output of
the `/experiment_runs` route is not just a list of string names) that
may not be what we want.

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -233,7 +234,7 @@ def scalars_route(self, request):
     # TODO: return HTTP status code for malformed requests
     tag = request.args.get('tag')
     run = request.args.get('run')
-    experiment = request.args.get('experiment')
+    experiment = request.args.get('experiment', '')
```
nfelt: Any particular reason to have this default to empty instead of None?

tensorboard/components/tf_backend/runsStore.ts
```diff
@@ -26,6 +26,12 @@ namespace tf_backend {
       });
     }
 
+    _getExperimentId() {
+      return (
+        new URLSearchParams(window.location.search).get('experiment') || ''
```
nfelt: Could we expose this somewhere so that the scalars code can re-use it rather than replicating it?  It could potentially be here although experimentsStore would perhaps be the best fit, even though this isn't strictly speaking the same as the old experiment data being served by that API.

That way it's easier to refactor later if we migrate away from a query parameter to e.g. embedding the experiment ID in the path.

tensorboard/plugins/scalar/scalars_plugin.py
wchargin: The experiment ID is required for any calls to the data provider API
(even though our current _implementation_ doesn’t strictly validate
this—after this PR, maybe it should), so it would have to be converted
from empty string to `None` at some point. It seemed cleanest to me to
reduce the space of input types as early as possible. No strong opinion.

tensorboard/components/tf_backend/runsStore.ts
wchargin: Exposed `tf_backend.getExperimentId()`. It’s still a bit of a kludge,
but it’s one kludge instead of two. :-)

PR #2581, opened 2019-08-20
Update wit-widget npm package

jameswex:

Update package to fix dependency version issue.

Need to rely on >= rules for dependencies, so this package works with latest JupyterLab notebooks.

Verified new version of npm package works in JupyterLab 1.0+, which old version did not.

### DIFF
--- a/tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/package.json
+++ b/tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/package.json
@@ -27,8 +27,8 @@
     "webpack": "^3.5.5"
   },
   "dependencies": {
-    "@jupyter-widgets/base": ">=1.0.0",
-    "@jupyter-widgets/jupyterlab-manager": ">=0.38.1"
+    "@jupyter-widgets/base": "^1.0.0",
+    "@jupyter-widgets/jupyterlab-manager": ">=0.38.1 <2.0.0"
   },
   "jupyterlab": {
     "extension": "lib/labplugin"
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/package.json
```diff
@@ -27,8 +27,8 @@
     "webpack": "^3.5.5"
   },
   "dependencies": {
-    "@jupyter-widgets/base": "^1.0.0",
-    "@jupyter-widgets/jupyterlab-manager": "^0.38.1"
+    "@jupyter-widgets/base": ">=1.0.0",
```
stephanwlee: The line before your change should roughly achieve the same effect. Is there a reason why you prefer this? 
i.e., ^1.0.0 would match any patch & minor release like 1.9.1. What you have here would make 15.3.1 also match but that is generally unfavorable since major bump often come with breaking changes and you want to make sure you are not impacted.

jameswex: What would be the right change for jupyterlab-manager in that case? I want it to work with 0.38 but also 1.x.

stephanwlee: I suppose, if that's what you want, you can do `>=0.38.1 <2.0.0`.

jameswex: thanks, done

PR #2583, opened 2019-08-20
[Debugger Plugin] Fix Python unit tests for TF2

caisq:
  * Re. https://github.com/tensorflow/tensorboard/issues/1705
  * Make sure all debugger plugin's python tests pass with TF2

  * In session_debug_test.py and interactive_debugger_plugin_test.py: use tf.compat.v1 and disable_v2_behavior(), as tfdbg (V1) is indeed wedded to the Session::Run() paradigm. 
  * In debug_graphs_helper_test.py: make conditional assertions to fit v2-specific behaviors

### DIFF
--- a/tensorboard/plugins/debugger/debug_graphs_helper_test.py
+++ b/tensorboard/plugins/debugger/debug_graphs_helper_test.py
@@ -37,6 +37,7 @@ from __future__ import print_function
 
 import tensorflow as tf
 from tensorflow.python import debug as tf_debug
+from tensorflow.python import tf2 as tensorflow_python_tf2
 from tensorflow.python.debug.lib import grpc_debug_test_server
 
 from tensorboard.compat.proto import config_pb2
@@ -108,15 +109,18 @@ class ExtractGatedGrpcDebugOpsTest(tf.test.TestCase):
       gated_debug_ops = [
           (item[0], item[2], item[3]) for item in gated_debug_ops]
 
-      # TODO(#1705): TF 2.0 breaks below.
       self.assertIn(('a', 0, 'DebugIdentity'), gated_debug_ops)
-      self.assertIn(('a/read', 0, 'DebugIdentity'), gated_debug_ops)
       self.assertIn(('b', 0, 'DebugIdentity'), gated_debug_ops)
-      self.assertIn(('b/read', 0, 'DebugIdentity'), gated_debug_ops)
       self.assertIn(('c', 0, 'DebugIdentity'), gated_debug_ops)
-      self.assertIn(('c/read', 0, 'DebugIdentity'), gated_debug_ops)
       self.assertIn(('d', 0, 'DebugIdentity'), gated_debug_ops)
-      self.assertIn(('d/read', 0, 'DebugIdentity'), gated_debug_ops)
+
+      if not tensorflow_python_tf2.enabled():
+        # Variable reading ops are different in TF1 and TF2.
+        self.assertIn(('a/read', 0, 'DebugIdentity'), gated_debug_ops)
+        self.assertIn(('b/read', 0, 'DebugIdentity'), gated_debug_ops)
+        self.assertIn(('c/read', 0, 'DebugIdentity'), gated_debug_ops)
+        self.assertIn(('d/read', 0, 'DebugIdentity'), gated_debug_ops)
+
       self.assertIn(('x', 0, 'DebugIdentity'), gated_debug_ops)
       self.assertIn(('y', 0, 'DebugIdentity'), gated_debug_ops)
       self.assertIn(('z', 0, 'DebugIdentity'), gated_debug_ops)
@@ -174,9 +178,14 @@ class BaseExpandedNodeNameTest(tf.test.TestCase):
       self.assertEqual(
           'bar/b/read',
           graph_wrapper.maybe_base_expanded_node_name('bar/b/read'))
-      # TODO(#1705): TF 2.0 tf.add creates nested nodes.
-      self.assertEqual(
-          'baz/c', graph_wrapper.maybe_base_expanded_node_name('baz/c'))
+
+      if tensorflow_python_tf2.enabled():
+        # NOTE(#1705): TF 2.0 tf.add creates nested nodes.
+        self.assertEqual(
+            'baz/c/(c)', graph_wrapper.maybe_base_expanded_node_name('baz/c'))
+      else:
+        self.assertEqual(
+            'baz/c', graph_wrapper.maybe_base_expanded_node_name('baz/c'))
 
 
 if __name__ == '__main__':
--- a/tensorboard/plugins/debugger/debug_graphs_helper_test.py
+++ b/tensorboard/plugins/debugger/debug_graphs_helper_test.py
@@ -86,7 +86,6 @@ class ExtractGatedGrpcDebugOpsTest(tf.test.TestCase):
                          debug_urls=self.debug_server_url)
     return z, run_options
 
-  @test_util.run_v1_only('Ops differ. Similar to [1].')
   def testExtractGatedGrpcTensorsFoundGatedGrpcOps(self):
     with tf.compat.v1.Session() as sess:
       z, run_options = self._createTestGraphAndRunOptions(sess, gated_grpc=True)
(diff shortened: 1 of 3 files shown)


tensorboard/plugins/debugger/debug_graphs_helper_test.py
```diff
@@ -37,6 +37,7 @@
 
 import tensorflow as tf
 from tensorflow.python import debug as tf_debug
+from tensorflow.python import tf2 as tensorflow_python_tf2
```
wchargin: Could you please add

```python
# See discussion on issue #1996 for private module import justification.
from tensorflow.python import tf2 as tensorflow_python_tf2
```

(as in `test_util.py`)?

tensorboard/plugins/debugger/debug_graphs_helper_test.py
```diff
@@ -108,15 +108,18 @@ def testExtractGatedGrpcTensorsFoundGatedGrpcOps(self):
...
       self.assertIn(('a', 0, 'DebugIdentity'), gated_debug_ops)
-      self.assertIn(('a/read', 0, 'DebugIdentity'), gated_debug_ops)
       self.assertIn(('b', 0, 'DebugIdentity'), gated_debug_ops)
-      self.assertIn(('b/read', 0, 'DebugIdentity'), gated_debug_ops)
       self.assertIn(('c', 0, 'DebugIdentity'), gated_debug_ops)
-      self.assertIn(('c/read', 0, 'DebugIdentity'), gated_debug_ops)
       self.assertIn(('d', 0, 'DebugIdentity'), gated_debug_ops)
-      self.assertIn(('d/read', 0, 'DebugIdentity'), gated_debug_ops)
+
+      if not tensorflow_python_tf2.enabled():
```
wchargin: We’ll shortly stop running tests on TF 1.x, so these will become dead
code. Can we invert the guard and test against the TF 2.x names instead?

tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
```diff
@@ -33,7 +33,7 @@
 import numpy as np
 import portpicker  # pylint: disable=import-error
 from six.moves import urllib  # pylint: disable=wrong-import-order
-import tensorflow as tf  # pylint: disable=wrong-import-order
+import tensorflow  # pylint: disable=wrong-import-order
```
wchargin: Can this be `import tensorflow.compat.v1 as tf` rather than leaving a
lurking (unused) `tensorflow` identifier?

tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
```diff
@@ -44,11 +44,13 @@
 from tensorboard.plugins.debugger import interactive_debugger_plugin
 from tensorboard.util import test_util
 
+tf = tensorflow.compat.v1
+tf.disable_v2_behavior()
```
wchargin: Interesting: simply adding `tf.compat.v1.disable_v2_behavior()` without
the rest of the patch doesn’t suffice to fix the timeouts on my end, so
some of the API differences must(?) be causing these unbounded
timeouts—do we know which ones?

In the short term, this patch is great because it lets us continue to
run the tests, but ideally we’d like to get rid of `tf.compat.v1`
wherever possible, too (#1718).

tensorboard/plugins/debugger/debug_graphs_helper_test.py
caisq: I removed this chunk of assertions.

tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
caisq: I'm worried that this will lead to copybara issues, as I've run into in tensorflow (see CL/264270116). The identifier `tensorflow` is not unused. It's used on line 47.

tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
caisq: I added a comment here. The Debugger Plugin is currently tied to v1 behavior (`tf.Session`s), which is the reason why it is necessary to pin V1 behavior here. When tfdbg V2 comes along, this will of course no longer be necessary.

tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
wchargin: We use `import tensorflow.compat.v1 as tf` extensively in TensorBoard
(you can find existing uses with `git grep`), and it doesn’t cause any
Copybara issues.

> The identifier `tensorflow` is not unused. It's used on line 47.

Of course, but if you change this to `import tensorflow.compat.v1 as tf`
then line 47 goes away, and there are no other uses of `tensorflow`. The
identifier is “lurking” in that `tensorflow` and `tf` are both in-scope
identifiers but have different values, which is confusing.

tensorboard/plugins/debugger/debug_graphs_helper_test.py
wchargin: Okay, that works, if you’re satisfied with the coverage.

tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
wchargin: Right, that makes sense. But if session and node behavior were the only
salient changes then we would expect adding `disable_v2_behavior()` to
fix the tests, right? And yet if we make only that change then the
`testMultipleSessionRunsTensorValueFullHistory` test case hangs at
various points—sometimes in the `listen` call in `setUp`, sometimes in
the first call to `_serverGet` in the test body. I was wondering why
that is, and why the remaining changes, which just change `tf.foo` to
`tf.compat.v1.foo`, _do_ fix the tests.

(This isn’t necessarily a blocking question, but I would be interested
to learn how this PR actually works.)

wchargin:
Re-reviewed.

tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
caisq: OK. I didn't know `import tensorflow.compat.v1 as tf` was used extensively in TensorBoard. Done.

tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
caisq: I confirmed what you said. It seems that making only the change of
1. Using `import tensorflow.compat.v1 as tf` 
2. Using `tf.disable_v2_behavior()`

leads to hanging tests. This looks like some unexpected behavior in how `tf.compat.v1.compat.v1....` behaves. 

As you said, this doesn't have to be blocking. I can make a TODO item for myself to investigate it and potentially file  bug for tensorflow while merging this PR. LMK WYT.

wchargin: Using `tf.compat.v1.compat.v1` will simply not work, in either TF 1.x or
TF 2.x. But making _only_ the `disable_v2_behavior()` change seems like
it _should_ work, because the rest of the code already uses
`tf.compat.v1.Session()` explicitly.

This is good to merge; feel free to investigate iff you feel like it.

caisq: `disable_v2_behavior()` exists only in `tf` from TF1 or `tf.compat.v1` from TF2. So to use it, we'd have to do `import tf.compat.v1 as tf` regardless (TF1 tests are passing, so they are out of concern). But if you use that and don't change anything below, you'll just end up using `tf.compat.v1.compat.v1.*`, which I think is what causes the hanging. 

I'll merge the PR. Thanks for the review.

wchargin: Let me be concrete: the patch that I’m talking about is simply

```diff
diff --git a/tensorboard/plugins/debugger/interactive_debugger_plugin_test.py b/tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
index 1b127456..653af030 100644
--- a/tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
+++ b/tensorboard/plugins/debugger/interactive_debugger_plugin_test.py
@@ -45,6 +45,9 @@ from tensorboard.plugins.debugger import interactive_debugger_plugin
 from tensorboard.util import test_util
 
 
+tf.compat.v1.disable_v2_behavior()
+
+
 _SERVER_URL_PREFIX = '/data/plugin/debugger/'
 
 
```

against `master` prior to this PR (i.e., 2b96c2a1).

This properly disables V2 behavior, does not use `compat.v1.compat.v1`,
and uses `tf.compat.v1.Summary`, but still hangs.

wchargin: And `tf.compat.v1.compat.v1` is a hard error, not a loop.

caisq: Acknowledged. Thanks for the concrete info.

wchargin:
Thanks for the fix!

PR #628, opened 2017-10-05
Improve card header word breaks

nfelt:
Insert <wbr> tags into the tag names used in card headers, so that in the case of long tag names, they will break more gracefully at word boundaries.

Before:

image

After:

image

Note that incidentally this also seems to address a bug where the long unbroken tag name pushed the containing div out too far, and bumped the "show description" circled-i icon out of view, making it inaccessible.  The bug is not 100% fixed because in the case of a tag name which has an unbreakable component that exceeds the card width (e.g. "someVeryLongUnbrokenRunNameWithNoHyphensSlashesOrUnderscores/loss/scalar_summary") we would still have the problem.  But that seems much less likely with this change in place.

PR #2586, opened 2019-08-21
Create `tf-wbr-string` for safe word-breaking

wchargin:
Summary:
Prior to this commit, `tf-runs-selector` used `innerHTML` to inset word
breaks into text, without explicit attention to sanitization. It
happened to “accidentally” to a pretty good job of preventing XSS
because it inserts `<wbr>` after each any equals sign, making it
impossible(?) to set any attribute value without closing the tag. But
some lesser injection vectors, like remote CSS fetching, are still
possible with a bit of thought.

This commit introduces a new component, `tf-wbr-string`, that performs
the word-breaking safely with Polymer APIs.

Test Plan:
Run TensorBoard, and verify that the data location at the bottom of the
runs selector is still rendered properly. Inspect the `tf-wbr-string` in
the dev tools, and note that `wbr`s appear in the appropriate places.
Then, select the `tf-wbr-string` element and use `$0.value = "…"` to
make sure that it breaks correctly in a variety of cases (empty string,
consecutive delimiters, leading delimiters, trailing delimiters).

Note that `git grep inner tensorboard/components/tf_runs_selector` no
longer returns any results.

wchargin-branch: tf-wbr-string

### DIFF
--- a/tensorboard/components/tf_runs_selector/tf-wbr-string.html
+++ b/tensorboard/components/tf_runs_selector/tf-wbr-string.html
@@ -1,0 +1,75 @@
+<!--
+@license
+Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<link rel="import" href="../tf-imports/polymer.html" />
+
+<!--
+  tf-wbr-string safely renders a string, with <wbr> elements inserted
+  around select delimiters.
+-->
+<dom-module id="tf-wbr-string">
+  <template>
+    <!--
+      This ugly formatting is required to prevent spaces from slipping
+      into the HTML.
+    -->
+    <template is="dom-repeat" items="[[_parts]]" as="part"
+      >[[part.contents]]<template is="dom-if" if="[[!part.last]]"
+        ><wbr /></template
+    ></template>
+  </template>
+  <script>
+    Polymer({
+      is: 'tf-wbr-string',
+      properties: {
+        /** The value to render with word breaks. */
+        value: String,
+        /**
+         * Insert a word-break after each match of this regex.
+         *
+         * This `RegExp` object will not be mutated.
+         */
+        delimiterPattern: {
+          type: Object /* RegExp */,
+          value: /([\/=\-_,])/,
+        },
+        _parts: {
+          type: Array /* {contents: string, last: boolean}[] */,
+          computed: '_computeParts(value, delimiterPattern)',
+        },
+      },
+      _computeParts(value, delimiterPattern) {
+        const result = [];
+        if (value == null) {
+          value = '';
+        }
+        while (true) {
+          const idx = value.search(delimiterPattern);
+          if (idx === -1) {
+            result.push({contents: value, last: true});
+            break;
+          } else {
+            const part = value.slice(0, idx + 1);
+            value = value.slice(idx + 1);
+            result.push({contents: part, last: false});
+          }
+        }
+        return result;
+      },
+    });
+  </script>
+</dom-module>
--- a/tensorboard/components/tf_runs_selector/tf-wbr-string.html
+++ b/tensorboard/components/tf_runs_selector/tf-wbr-string.html
@@ -38,7 +38,7 @@ limitations under the License.
         /** The value to render with word breaks. */
         value: String,
         _parts: {
-          type: Array /* {contents: string, last: boolean}[] */,
+          type: Array /* string[] */,
           computed: '_computeParts(value)',
         },
       },
--- a/tensorboard/components/tf_runs_selector/tf-wbr-string.html
+++ b/tensorboard/components/tf_runs_selector/tf-wbr-string.html
@@ -19,7 +19,7 @@ limitations under the License.
 
 <!--
   tf-wbr-string safely renders a string, with <wbr> elements inserted
-  around select delimiters.
+  around select delimiters (see `const delimiterPattern` in this file).
 -->
 <dom-module id="tf-wbr-string">
   <template>
--- a/tensorboard/components/tf_runs_selector/tf-wbr-string.html
+++ b/tensorboard/components/tf_runs_selector/tf-wbr-string.html
@@ -28,9 +28,8 @@ limitations under the License.
       into the HTML.
     -->
     <template is="dom-repeat" items="[[_parts]]" as="part"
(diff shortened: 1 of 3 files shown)


tensorboard/components/tf_runs_selector/tf-wbr-string.html
```diff
@@ -0,0 +1,75 @@
...
+<link rel="import" href="../tf-imports/polymer.html" />
+
+<!--
+  tf-wbr-string safely renders a string, with <wbr> elements inserted
+  around select delimiters.
+-->
+<dom-module id="tf-wbr-string">
+  <template>
+    <!--
+      This ugly formatting is required to prevent spaces from slipping
```
nfelt: Optional but IMO it can be nicer to achieve this by inserting inline comments between the tags, like

```html
<template is="dom-repeat" items="[[_parts]]" as="part"><!--
  -->[[part.contents]]<!--
  --><template is="dom-if" if="[[!part.last]]"><!--
    --><wbr /><!--
  --></template><!--
--></template>
```

Then at least you can preserve the hierarchical shape, though obviously it's a bit more verbose and cumbersome to write.

tensorboard/components/tf_runs_selector/tf-wbr-string.html
```diff
@@ -0,0 +1,75 @@
...
+        if (value == null) {
+          value = '';
+        }
+        while (true) {
+          const idx = value.search(delimiterPattern);
+          if (idx === -1) {
+            result.push({contents: value, last: true});
+            break;
+          } else {
+            const part = value.slice(0, idx + 1);
```
nfelt: Isn't `idx + 1` only correct if the delimiter match is exactly 1 character?  If that's all we want to support it should be documented as such, and then in that case maybe we should construct the regex ourselves to control for this.

tensorboard/components/tf_runs_selector/tf-wbr-string.html
```diff
@@ -0,0 +1,75 @@
...
+      },
+      _computeParts(value, delimiterPattern) {
+        const result = [];
+        if (value == null) {
+          value = '';
+        }
+        while (true) {
+          const idx = value.search(delimiterPattern);
+          if (idx === -1) {
+            result.push({contents: value, last: true});
```
nfelt: If value was empty (i.e. the last delimiter matched is at the end of the string), then we'll be pushing an extra `<wbr>` onto the end, which seems unnecessary.

Seems like a fix would just be to do `while (value)` instead?

tensorboard/components/tf_runs_selector/tf-wbr-string.html
wchargin: Agreed, but Prettier rewrites this to

```html
    <template is="dom-repeat" items="[[_parts]]" as="part"
      ><!--
      -->[[part.contents]]<!--
      --><template is="dom-if" if="[[!part.last]]"
        ><!--
        --><wbr /><!--
      --></template
      ><!--
    --></template>
```

<sub>Daily reminder that JSX does not have semantic whitespace.</sub>

tensorboard/components/tf_runs_selector/tf-wbr-string.html
wchargin: Either seems fine with me (they should have the same effect), but sure.

tensorboard/components/tf_runs_selector/tf-wbr-string.html
wchargin: Yes; good point. I’ll make the pattern private.

wchargin:
> https://github.com/tensorflow/tensorboard/pull/628#issuecomment-335279840

Truth be told, I did not remember writing that comment while writing
this PR, but each one of those thoughts did go through my head once
again. :sigh:

tensorboard/components/tf_runs_selector/tf-wbr-string.html
```diff
@@ -0,0 +1,75 @@
...
+        /** The value to render with word breaks. */
+        value: String,
+        /**
+         * Insert a word-break after each match of this regex.
+         *
+         * This `RegExp` object will not be mutated.
+         */
+        delimiterPattern: {
+          type: Object /* RegExp */,
+          value: /([\/=\-_,])/,
```
wchargin: I’ve also taken this opportunity to reformat the `/[\/=\-_,]/` pattern
for clarity: it does not match a backslash, it does match a hyphen, and
it does not match a caret (ASCII value 0x5E between backslash 0x5C and
underscore 0x5F). This is of course _unambiguous_ in the original form,
but I don’t think that it was as _obvious_ as with `/[/=_,-]/`.

tensorboard/components/tf_runs_selector/tf-wbr-string.html
wchargin: Using `while (value)` doesn’t actually prevent the last-added part from
having `last: false` on an input like `"a/"`, and thus doesn’t prevent
the extra `<wbr>`. Given that the unnecessary `<wbr>`s do no harm, and
this code is clearly proving too complicated for my tiny brain, I’ve
gone the other way and just included the `<wbr>` after every part. Much
simpler. Everybody’s a winner.

tensorboard/components/tf_runs_selector/tf-wbr-string.html
```diff
@@ -0,0 +1,65 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<link rel="import" href="../tf-imports/polymer.html" />
+
+<!--
+  tf-wbr-string safely renders a string, with <wbr> elements inserted
+  around select delimiters.
```
stephanwlee: am I reading it correctly that wbr will be added at the end? Also, since delimiter is no longer configurable, can we just write out all the delimiter this would break? e.g., this will be "noop" for input like `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` right?

wchargin: > am I reading it correctly that wbr will be added at the end?

Yes. (<https://github.com/tensorflow/tensorboard/pull/2586#discussion_r316451820>)

> Also, since delimiter is no longer configurable, can we just write out
> all the delimiter this would break?

Not sure what you mean—are you asking that we specify the delimiter
pattern in this doc comment? Sure, if you like.

> e.g., this will be "noop" for input like
> `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` right?

Yes (as the previous implementation was).

tensorboard/components/tf_runs_selector/tf-wbr-string.html
```diff
@@ -0,0 +1,65 @@
...
+      _computeParts(value) {
+        const result = [];
+        const delimiterPattern = /[/=_,-]/;
+        if (value == null) {
+          value = '';
+        }
+        while (true) {
+          const idx = value.search(delimiterPattern);
+          if (idx === -1) {
+            result.push(value);
```
stephanwlee: Stating obvious: the `value` can be an empty string and you are pushing an empty string into the `result` which can cause multiple `<wbr><wbr>` which probably has zero visual impact.

wchargin: Yep!

Issue #2387, opened 2019-06-26
Listen only localhost by default


KOLANICH:
By default tensorbord listens all interfaces, which may be a security issue. IMHO by default it should listen only on localhost, and if one uses it on a server, he usually configures the stuff, so it is not an issue for him.

nfelt:
Thanks for the feedback, and yes, our current thinking is that we'll cut over to listening on localhost only by default for our 2.0 release, though there are a few details to work out around making sure it's easy for users who want the current behavior to still get it.  Stay tuned.

nfelt:
Update: as mentioned, we plan on changing the default to `localhost` for our 2.0 release, since this is the more conservative and safe behavior.

Users who specifically want to expose TensorBoard (e.g. in a lab context to let other users access the instance) can either pass `--host=<interface>` explicitly to bind to the desired interface, or configure some type of proxied access (for example, ssh port forwarding like `ssh -L 6006:localhost:6006 <my-tensorboard-server>`).

Strictly speaking the default behavior is actually to attempt to bind to `::` and enable IPv6-mapped IPv4 for the socket, so that the same socket can serve both IPv4 and IPv6 traffic, which useful in cases where a hostname for the server (including sometimes `localhost` itself) has both IPv4 and IPv6 addresses associated with it, since this prevents spurious connection-denied errors from having chosen the wrong protocol.  See https://github.com/tensorflow/tensorboard/pull/1449 for the details.

Since the existing behavior can't be achieved by specifying either `0.0.0.0` or `::` explicitly, I propose that we accept an additional special host value, `--host=*`, which behave as though `::` was passed but attempt to make the socket dual-bind to IPv4 as well.

This change should mostly resolve the following issues:
- https://github.com/tensorflow/tensorboard/issues/1713 (bug still exists, but won't be triggered except on passing `--host=*`, and can be resolved by passing `--host=0.0.0.0` instead)
- https://github.com/tensorflow/tensorboard/issues/89 (localhost will work)
- PR https://github.com/tensorflow/tensorboard/pull/2371 (with defaulting to localhost, a potentially-not-reachable domain name will only be emitted when explicitly passing `--host={0.0.0.0,::,*}` and can be resolved by passing a domain name for an interface known to be valid; this PR could still be useful but the incidence of the issue should be much lower)
[...]

Issue #1320, opened 2018-07-27
Fails silently if port is occupied


yaroslavvb:
TensorBoard 1.9.0 fails silently if port is already occupied

```
tensorboard --logdir /tmp/hi --port 6006
TensorBoard 1.9.0 at http://mac.lan:6006 (Press CTRL+C to quit)
```
Downgrading to version 1.8 fixes this

```
base) ÷18-07-27 09:34:02|ttys012|mac|~\÷ tensorboard --logdir /tmp/hi --port 6006
E0727 09:34:20.147497 MainThread main.py:197] TensorBoard attempted to bind to port 6006, but it was already in use
TensorBoard attempted to bind to port 6006, but it was already in use
```

nfelt:
Ok, I finally got to the bottom of this issue, which I will send a PR to fix.  For posterity:

Th IPv4/v6 handling logic first attempts to bind to an empty host, and then assumes any error was due to a lack of IPv4, and would then fall back to an explicit IPv6 wildcard (`::`).  There are a number of problems with this approach:

1) Passing in a blank host is unreliable.  Python's socket module says the following:

    > For IPv4 addresses, two special forms are accepted instead of a host address: the empty string represents INADDR_ANY, and the string '<broadcast>' represents INADDR_BROADCAST.

    This is true of socket.bind(), which Werkzeug used to call directly, resulting in TensorBoard binding to the IPv4 wildcard address.  However, in the past year https://github.com/pallets/werkzeug/pull/1150 changed this logic to pass the host and port through a call to getaddrinfo() first, and `getaddrinfo('')` exhibits inconsistent behavior depending on the platform:

    - On Debian, the empty string raises an error (which leads Werkzeug to do what it used to, pass the empty string to bind, and results in a wildcard bind as expected)
    - On macOS, the empty string makes getaddrinfo() return only the loopback/local IPs
    - On Windows, the empty string makes getaddrinfo() return [all registered addresses](https://docs.microsoft.com/en-us/windows/desktop/api/ws2tcpip/nf-ws2tcpip-getaddrinfo#remarks)

2) Errors from the empty-host bind weren't necessarily IPv4 failures - sometimes they were "address in use" errors, which the handling logic expects later, but not until after trying the fallback bind().

[...]

PR #2589, opened 2019-08-22
core: bind to localhost rather than all interfaces

wchargin:
Summary:
Prior to this change, TensorBoard would default to serving on the entire
local network; now, TensorBoard serves to the local machine only, and
the flag `--bind_all` can be used to dual-bind to IPv4 and IPv6 on the
entire local network (the previous default). See #2387 and comments
therein for details.

Test Plan:
On my Debian machine, running with `strace -e trace=%network`:

  - running with no `--host` flag:
      - can connect to loopback on IPv4 only
      - cannot connect over LAN
      - `strace` shows binding on `AF_INET`
      - a notice about `--bind_all` is printed to stderr
  - running with `--host=localhost`:
      - same behavior as with no `--host` flag, but no notice is printed
  - running with `--host='::1'`:
      - can connect to loopback on IPv6 only
      - cannot connect over LAN
      - `strace` shows binding on `AF_INET6`
  - running with `--host=0.0.0.0`:
      - can connect to loopback on IPv4 only
      - **can** connect over LAN
      - `strace` shows binding on `AF_INET`
  - running with `--host='::0'`:
      - can connect on both IPv4 and IPv6
      - **can** connect over LAN
      - `strace` shows binding on `AF_INET6`
  - running with `--bind_all`:
      - can connect on both IPv4 and IPv6
      - **can** connect over LAN
      - `strace` shows binding on `AF_INET6` with an additional syscall
        to `setsockopt(3, SOL_IPV6, IPV6_V6ONLY, [0], 4)` to facilitate
        the dual-binding, which is not present in any other tested case

In all cases, the printed serving URL (“TensorBoard x.y.z running at…”)
bears the exact `--host` flag, or my full hostname if `--bind_all` was
given, or `localhost` if neither was given. In all cases, the URL is a
clickable link in my `gnome-terminal`.

Note that on my system dual binding to `::0` works without an explicit
syscall—i.e., `IPV6_V6ONLY` defaults to `0`—but this is not portable.

Connection testing was performed via

```shell
for ipv in 4 6; do
    if curl -sfL -"${ipv}" localhost:6006/data/logdir >/dev/null; then
        printf 'v%d OK\n' "${ipv}"
    else
        printf 'v%d FAIL\n' "${ipv}"
    fi
done
```

in all cases.

wchargin-branch: localhost-only

### DIFF
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -219,7 +219,6 @@ class TensorBoard(object):
     try:
       server = self._make_server()
       server.print_serving_message()
-      sys.stderr.flush()
       self._register_info(server)
       server.serve_forever()
       return 0
@@ -334,6 +333,7 @@ class TensorBoardServer(object):
         'TensorBoard %s at %s (Press CTRL+C to quit)\n'
         % (version.VERSION, self.get_url())
     )
+    sys.stderr.flush()
 
 
 class TensorBoardServerException(Exception):
@@ -540,6 +540,7 @@ class WerkzeugServer(serving.ThreadedWSGIServer, TensorBoardServer):
           'Serving TensorBoard on localhost; to expose to the network, '
           'use a proxy or pass --bind_all\n'
       )
+      sys.stderr.flush()
     super(WerkzeugServer, self).print_serving_message()
 
   def _fix_werkzeug_logging(self):
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -425,7 +425,7 @@ class WerkzeugServer(serving.ThreadedWSGIServer, TensorBoardServer):
     # Without an explicit host, we default to serving on all interfaces,
     # and will attempt to serve both IPv4 and IPv6 traffic through one
     # socket.
-    self._auto_wildcard = not host
+    self._auto_wildcard = host == '*'
     if self._auto_wildcard:
       host = self._get_wildcard_address(port)
 
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -218,8 +218,7 @@ class TensorBoard(object):
       return 0
     try:
       server = self._make_server()
-      sys.stderr.write('TensorBoard %s at %s (Press CTRL+C to quit)\n' %
-                       (version.VERSION, server.get_url()))
+      server.print_serving_message()
       sys.stderr.flush()
       self._register_info(server)
       server.serve_forever()
@@ -326,6 +325,16 @@ class TensorBoardServer(object):
     """Returns a URL at which this server should be reachable."""
     raise NotImplementedError()
 
+  def print_serving_message(self):
+    """Prints a user-friendly message prior to server start.
+
+    This will be called just before `serve_forever`.
+    """
+    sys.stderr.write(
+        'TensorBoard %s at %s (Press CTRL+C to quit)\n'
+        % (version.VERSION, self.get_url())
+    )
+
 
 class TensorBoardServerException(Exception):
   """Exception raised by TensorBoardServer for user-friendly errors.
@@ -413,12 +422,15 @@ class WerkzeugServer(serving.ThreadedWSGIServer, TensorBoardServer):
     host = flags.host
     port = flags.port
 
-    # Without an explicit host, we default to serving on all interfaces,
-    # and will attempt to serve both IPv4 and IPv6 traffic through one
-    # socket.
-    self._auto_wildcard = host == '*'
+    self._auto_wildcard = flags.bind_all
     if self._auto_wildcard:
+      # Serve on all interfaces, and attempt to serve both IPv4 and IPv6
+      # traffic through one socket.
       host = self._get_wildcard_address(port)
+    elif host is None:
+      host = 'localhost'
+
+    self._host = host
 
     self._fix_werkzeug_logging()
     try:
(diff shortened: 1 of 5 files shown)


wchargin:
I’ve attempted to implement this to the spec in #2387. I’ve read through
<https://github.com/tensorflow/tensorboard/issues/1320#issuecomment-423364207>,
but I can claim no expertise here, so @nfelt’s review and testing are
warmly welcomed.

wchargin:
From offline discussion: we don’t much like that the `--host='*'` form
uses a literal asterisk, because if one writes `--host *` without quotes
then glob expansion will produce an argument list that is not what was
intended. (Technically, `--host=*` unquoted could do the same, but this
less unlikely.)

We chose “`*`” in the first place because it’s one of very few strings
that can’t be a valid address. For instance, choosing `--host=all` as
the magic incantation would be confusing on machines with literal
hostname `all`. An option that is neither a POSIX shell special
character nor a valid address would be `--host=.`, but this is not
terribly clear, as `.` more typically refers to the current directory
and there’s nothing to suggest that it should mean “bind publicly to all
addresses”.

We propose instead the following:

  - If no flags are specified, use `::1`, an IPv6 loopback address (like
    IPv4’s `127.0.0.1`), as the default host.
  - If a new flag `--public` is given, change the default host to `::`,
    the IPv6 wildcard address (like IPv4’s `0.0.0.0`). It shall be an
    error to use both `--public` and `--host`.
  - Always attempt to dual-bind an IPv6 socket as IPv4. This is
    implemented by using `setsockopt` to clear the `IPV6_V6ONLY` flag,
    so will have no effect if the host is explicitly IPv4: e.g., if the
    user passes `--host 0.0.0.0` or `--host localhost`. We see no harm
    in dual-binding to IPv4 even when the user explicitly gives us an
    IPv6 host like `::1`.

This way, the stock answer to user questions of the form “why can’t my
collaborators see my TensorBoards?” can be “just add `--public`”, and
it’s clear from the command line at least that `--public` is _some_ sort
of ACL expansion.

nfelt:
So... after mulling this over, the flag `--public` still doesn't sit quite right with me since it's not clearly related to the host flag, and the implications of "public" are a little ambiguous (it's not necessarily actually public in a meaningful sense).  And conversely, any use of `--host` that isn't with localhost might also result in a "public" tensorboard, so the absence of `--public` isn't any guarantee that your instance is inaccessible.

Would using `all` as a sentinel value really be that bad?  It seems very unlikely to be an entire hostname in practice, since even if the local part of the hostname was `all`, there would typically be further parts to actually form a FQDN of any practical use.

Also, I think we probably want to special-case `localhost` to ensure it gets dual-mapped (and maybe even `127.0.0.1` as well) because it's not necessarily IPv4-only.  In fact, what I recall from testing this is that at least on my machine and presumably many others, `localhost` has DNS mappings for both IPv4 and IPv6, and which one actually gets used by your browser is browser-dependent.  In particular, if you don't dual-map it becomes possible in some cases to run two different servers on localhost on the same port, and two tabs pointed at that address might actually end up connecting over different IP flavors to get different backends.

wchargin:
Okay, I’ll hold off for now. A few quick notes:

> Would using `all` as a sentinel value really be that bad? It seems
> very unlikely to be an entire hostname in practice, since even if the
> local part of the hostname was `all`, there would typically be further
> parts to actually form a FQDN of any practical use.

It doesn’t have to be the entire hostname; with current TensorBoard,
invoking

```
tensorboard --logdir /tmp/x1 --port 6006 --host "$(hostname -s)"
```

(where my `hostname -s` is a single label) exposes the instance to LAN.
Special-casing `all` would change that behavior.

Not to say that we _can’t_ bite the bullet and do that anyway.

> it becomes possible in some cases to run two different servers on
> localhost on the same port, and two tabs pointed at that address might
> actually end up connecting over different IP flavors to get different
> backends

Yes; I ran into this while writing `manager.py`. A simple example:

```shell
$ tensorboard --logdir /tmp/x1 --port 6006 --host 127.0.0.1
TensorBoard 1.15.0a20190822 at http://127.0.0.1:6006/ (Press CTRL+C to quit)
^Z
[1]+  Stopped                 tensorboard --logdir /tmp/x1 --port 6006 --host 127.0.0.1
$ tensorboard --logdir /tmp/x2 --port 6006 --host ::1
TensorBoard 1.15.0a20190822 at http://[::1]:6006/ (Press CTRL+C to quit)
^Z
[2]+  Stopped                 tensorboard --logdir /tmp/x2 --port 6006 --host ::1
$ bg %1 %2
[1]- tensorboard --logdir /tmp/x1 --port 6006 --host 127.0.0.1 &
[2]+ tensorboard --logdir /tmp/x2 --port 6006 --host ::1 &
$ curl -4 localhost:6006/data/logdir; printf '\n'
{"logdir": "/tmp/x1"}
$ curl -6 localhost:6006/data/logdir; printf '\n'
{"logdir": "/tmp/x2"}
```

When given an IPv4 loopback address (possibly including `localhost`), we
can’t actually force it to be dual-bound, right? On my testing on
current Debian, binding one socket to `127.0.0.1` and another to `::1`
on the same port seems to always succeed and allow version-specific
connections to both (and in particular binding to `::1` immediately
resets the `IPV6_V6ONLY` socket option to `1`), so if the user specifies
those two hosts I don’t see how any dual-binding configuration helps.

wchargin:
> We chose “`*`” in the first place because it’s one of very few strings
> that can’t be a valid address

FWIW, `--host=...` would be another unambiguous specifier, which may be
more or less confusing than `--host=.` but is still not great.

nfelt:
Okay, what about the following path forward:

- add a `--bind-all` flag separate from `--host` to activate today's wildcard dual-binding behavior
- make it an error to pass `--bind-all` and `--host`
- default `--host` to None
- if `--host` is None (the default), behave like passing `--host localhost`, except also print a message to stderr just before the normal [launch message](https://github.com/tensorflow/tensorboard/blob/d5ea5a3f14681c5685ed7deb5cbf4008049073dd/tensorboard/program.py#L230) that says:

`Serving TensorBoard on localhost - to expose to the network use a proxy or pass --bind-all`

wchargin:
That plan sounds good to me. Done; PTAL.

tensorboard/program.py
```diff
@@ -413,12 +422,15 @@ def __init__(self, wsgi_app, flags):
...
-    self._auto_wildcard = not host
+    self._auto_wildcard = flags.bind_all
     if self._auto_wildcard:
+      # Serve on all interfaces, and attempt to serve both IPv4 and IPv6
+      # traffic through one socket.
       host = self._get_wildcard_address(port)
+    elif host is None:
+      host = 'localhost'
+
+    self._host = host
```
nfelt: Is `self._host` actually used anywhere?  Later down you read `self.host`, which is the property set by werkzeug:
https://github.com/pallets/werkzeug/blob/bd60d52ba14f32a38caffc674fb17c9090ef70ce/src/werkzeug/serving.py#L728

Not sure if those are intentionally different or a typo.

wchargin: Good catch; this was an oversight. I’ll conservatively use `self._host`.

PR #2593, opened 2019-08-23
Look for `tensorflow_core._api` in summary component package

wchargin:
Summary:
The internal API for the base `tf.summary` module has changed from
`tensorflow._api.*` to `tensorflow_core._api.*` as of 2019-08-23
nightlies. This is a quick fix to update our component package
accordingly, though it does not prevent future similar regressions.

Test Plan:
Running

```python
import importlib
print(hasattr(importlib.import_module("tensorflow._api.v2"), "summary"))
print(hasattr(importlib.import_module("tensorflow_core._api.v2"), "summary"))
```

prints `True` `False` on yesterday’s nightly and `False` `True` on
today’s. The `//tensorboard/examples/plugins/example_basic:smoke_test`
target fails before this commit and passes after it.

wchargin-branch: tf-core-summary

### DIFF
--- a/tensorboard/summary/_tf/summary/__init__.py
+++ b/tensorboard/summary/_tf/summary/__init__.py
@@ -128,6 +128,10 @@ def reexport_tf_summary():
   packages = [
       'tensorflow',
       'tensorflow.compat.v2',
+      'tensorflow_core._api.v2',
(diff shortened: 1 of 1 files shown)


wchargin:
This unblocks our build.

PR #2594, opened 2019-08-24
[tensor-widget] Add detailed-value tooltip and single-element selection

caisq:
  * Continue developing TensorWidget.

  * Add hover callback for value divs, which shows the detailed (many decimal points, e.g., for float-type tensors).
  * Add `selection.ts` and the TensorElementSelection to encapsulate the logic of how multi-element selection works and how it's updated during moving.
  * Add unit tests for `selection.ts` in `selection-test.ts`.
  * Make initial use of the TensorElementSelection class in tensor-widget-impl.ts. Achieve
    * single-element selection
    * scrolling based on the selection and keyboard arrow keys
   * Address a TODO item: override mouse wheel events only when a selection exists.

* Screenshots of UI changes
  * single-element selection within a 4D tensor: 
image
  * Detailed value tooltip: 
image
  * It's hard to make a screenshot of the selection and keyboard based scrolling because that would require an animation.

### DIFF
--- a/tensorboard/components/tensor_widget/selection.ts
+++ b/tensorboard/components/tensor_widget/selection.ts
@@ -14,7 +14,7 @@ limitations under the License.
 ==============================================================================*/
 
 import {numElements} from './shape-utils';
-import {Shape, TensorViewSlicingSpec, TensorView} from './types';
+import {MoveDirection, Shape, TensorViewSlicingSpec} from './types';
 
 /**
  * The possible status of a selected cell.
@@ -24,16 +24,6 @@ export interface CellSelectionStatus {
   bottomEdge?: boolean;
   leftEdge?: boolean;
   rightEdge?: boolean;
-}
-
-/**
- * Possible directions of selection movement.
- */
-export enum MoveDirection {
-  UP = 1,
-  DOWN,
-  LEFT,
-  RIGHT,
 }
 
 /**
--- a/tensorboard/components/tensor_widget/selection.ts
+++ b/tensorboard/components/tensor_widget/selection.ts
@@ -14,7 +14,7 @@ limitations under the License.
 ==============================================================================*/
 
 import {numElements} from './shape-utils';
-import {Shape, TensorViewSlicingSpec} from './types';
+import {Shape, TensorViewSlicingSpec, TensorView} from './types';
 
 /**
  * The possible status of a selected cell.
@@ -29,16 +29,12 @@ export interface CellSelectionStatus {
 /**
  * Possible directions of selection movement.
  */
-export enum SelectionMoveDirection {
+export enum MoveDirection {
   UP = 1,
   DOWN,
   LEFT,
   RIGHT,
 }
-
-export type OnSlicingSpecChangeCallback = (
-  slicingSpec: TensorViewSlicingSpec
-) => void;
 
 /**
  * The selection state within a n-dimensional tensor.
@@ -52,7 +48,6 @@ export type OnSlicingSpecChangeCallback = (
  *   slicing spec ought to be updated to accommodate it.
  */
 export class TensorElementSelection {
-  private slicingSpec: TensorViewSlicingSpec;
   private sliceDims: number[] = [];
   private sliceIndices: number[] = [];
   private viewDims: number[];
@@ -82,17 +77,13 @@ export class TensorElementSelection {
     rowStart?: number,
     colStart?: number,
     rowCount?: number,
-    colCount?: number,
-    private readonly onSlicingSpecChange: OnSlicingSpecChangeCallback | null = null
+    colCount?: number
   ) {
     if (numElements(this.shape) === 0) {
       throw new Error(
         `TensorElementSelection doesn't support tensor with zero elements.`
       );
     }
-
-    // NOTE: Make a copy to avoid mutate the input one.
-    this.slicingSpec = JSON.parse(JSON.stringify(slicingSpec));
 
     for (let i = 0; i < slicingSpec.slicingDimsAndIndices.length; ++i) {
       this.sliceDims.push(slicingSpec.slicingDimsAndIndices[i].dim);
@@ -208,10 +199,6 @@ export class TensorElementSelection {
     return status;
   }
 
-  public setSlicingSpec(slicingSpec: TensorViewSlicingSpec) {
-    this.slicingSpec = JSON.parse(JSON.stringify(slicingSpec));
-  }
-
   /**
    * Move the selection.
    *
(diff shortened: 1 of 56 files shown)


tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,340 @@
...
+    private readonly shape: Shape,
+    readonly slicingSpec: TensorViewSlicingSpec,
+    rowStart?: number,
+    colStart?: number,
+    rowCount?: number,
+    colCount?: number
+  ) {
+    if (size(this.shape) === 0) {
+      throw new Error(
+        `TensorElementSelection doesn't support tensor with zero elements.`
```
bileschi: Doesn't `(size(this.shape)) === 0)` imply that the tensor is a scalar, not that the tensor has zero elements?

caisq: `size(this.shape) === 0` doesn't mean this is a scalar. It means you have a shape such as [0, 10], or [3, 0, 5]. A scalar has a `size()` of 1.

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,340 @@
...
+        this.slicingSpec.viewingDims != null &&
+        this.slicingSpec.viewingDims[0] != null &&
+        this.rowStart < this.shape[this.slicingSpec.viewingDims[0]] - 1
+      ) {
+        this.rowStart++;
+        if (
+          this.slicingSpec.verticalRange != null &&
+          this.rowStart >= this.slicingSpec.verticalRange[1]
+        ) {
+          const increment = 1;
```
bileschi: Why is this pattern (with increment pulled out as a constant) only used for DOWN?

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,340 @@
...
+   * Compute whether a given set of indices falls into the selection.
+   *
+   * ... and if so, whether the set of indices belongs to any of the four
+   * edges of the selected region.
+   *
+   * @param indices
+   * @return Cell selection status, if the set of indices falls into the
+   *   selection. `null` otherwise.
+   */
+  public getElementStatus(indices: number[]): CellSelectionStatus[] | null {
```
bileschi: This function is somewhat long and involves keeping track of several numbers.  Is it possible to break it into testable helper functions?

wchargin:
Can we use standard BMP arrows instead of the supplemental planes? It
took me a few minutes for me to find this bit of the code—I started with
`git grep 2194` and `git grep ↔`, neither of which turned up the
relevant code because “↔” (U+2194) is not “🡘” (U+1F858). This seems
fairly likely to fix the rendering on macOS, too.

caisq:
@bileschi Thanks for catching the issue about the arrows. As @wchargin suggested, I'm using standard BMP arrows now.

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,340 @@
...
+   * @param colStart The starting column of selection, within the indices
+   *   framework of the original tensor (i.e., *not* with respect to the
+   *   slicing spec.)
+   * @param rowCount How many rows are selected.
+   * @param colCount How many columns are selected.
+   */
+  constructor(
+    private readonly shape: Shape,
+    readonly slicingSpec: TensorViewSlicingSpec,
+    rowStart?: number,
```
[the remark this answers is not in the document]
caisq: The name of the args should make it clear. Also note this is an internal API, which alleviates the concern.

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,339 @@
...
+      throw new Error(`Only selections in 1D and 2D are supported.`);
+    }
+
+    this.rowStart = rowStart == null ? 0 : rowStart;
+    this.colStart = colStart == null ? 0 : colStart;
+    this.rowCount = rowCount == null ? 1 : rowCount;
+    this.colCount = colCount == null ? 1 : colCount;
+  }
+
+  private rank(): number {
```
davidsoergel: `shape` is readonly, so there's no need for this to be a function.  Instead declare `private final rank` and set it in the constructor.

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,339 @@
...
+            `undetermined slicing index at dimension ${i}`
+        );
+      }
+      this.sliceIndices.push(index);
+    }
+
+    // Sanity check the size of the the slicing dimensions.
+    if (this.rank() > 0 && this.sliceDims.length >= this.rank()) {
+      throw new Error(
+        `Expected sliceDims to have a length less than rank ${this.rank}, ` +
```
davidsoergel: `${this.rank}` should have been `${this.rank()}`, except my suggestion below obviates that.

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,339 @@
...
+limitations under the License.
+==============================================================================*/
+
+import {numElements} from './shape-utils';
+import {Shape, TensorViewSlicingSpec} from './types';
+
+/**
+ * The possible status of a selected cell.
+ */
+export enum CellSelectionStatus {
```
davidsoergel: A given cell can have any of these properties.  Instead of `CellSelectionStatus[]`, I think it would be more natural to represent that as as interface with five booleans.  If you want to get really precise about the edge states being valid only if the cell is selected, you could have something like `type CellSelectionStatus = null | {leftEdge?: boolean, rightEdge?: boolean ...}`.  Then `null` would mean 'not selected' and `{}` would mean 'selected but not on an edge'.

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,339 @@
...
+        `Expected indices to have a rank of ${this.rank}, ` +
+          `but got ${indices.length} ([${indices}])`
+      );
+    }
+
+    // First, make sure that the indices belongs to a selected slice.
+    for (let i = 0; i < indices.length; ++i) {
+      if (this.sliceDims.indexOf(i) !== -1) {
+        if (indices[i] !== this.sliceIndices[this.sliceDims.indexOf(i)]) {
+          return null;
```
davidsoergel: Related to the typing suggestion above: currently `null` and `[]` have the same meaning (cell is not selected), and in fact this function might return either one; this seems confusing.

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,339 @@
...
+
+    let status: CellSelectionStatus[] | null = null;
+
+    const rowEnd = this.rowStart + this.rowCount;
+    const colEnd = this.colStart + this.colCount;
+
+    // Second, check the viewing dims.
+    if (this.viewDims.length === 0) {
+      if (indices.length === 0) {
+        if (status == null) {
```
davidsoergel: Currently always true.  With the current typing, you could initialize `status` to `[]` above.  With my suggested change, this block would become `status = {leftEdge: true, rightEdge: true, ...};`.

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,339 @@
...
+    ) {
+      // No-op for moving left or right in a 1D tensor.
+      return null;
+    }
+
+    if (
+      this.slicingSpec.verticalRange === null ||
+      this.slicingSpec.verticalRange[1] === null
+    ) {
+      throw new Error(`Failed to move odue to undetermined vertical range.`);
```
davidsoergel: typo

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    if (direction === SelectionMoveDirection.UP) {
+      if (this.rowStart > 0) {
+        this.rowStart--;
+        if (
+          this.slicingSpec.verticalRange != null &&
+          this.rowStart < this.slicingSpec.verticalRange[0]
+        ) {
+          this.slicingSpec.verticalRange[0]--;
```
davidsoergel: This mutates the provided `slicingSpec`, since the constructor did not make a defensive copy.  In any case this function seems weirdly entangled with the caller (which needs to know when the slicingSpec changed, etc.)  Maybe it makes sense to construct this object with a callback for reporting a slicingSpec change?

tensorboard/components/tensor_widget/selection.ts
caisq: Great point. I've adopted your "precise" option.

tensorboard/components/tensor_widget/selection.ts
caisq: This function itself is thoroughly tested. At about 70 lines, the length of the function is okay IMO.

tensorboard/components/tensor_widget/selection.ts
caisq: Done (see above).

tensorboard/components/tensor_widget/selection.ts
caisq: Done (see above).

tensorboard/components/tensor_widget/selection.ts
caisq: Fixed.

tensorboard/components/tensor_widget/selection.ts
caisq: Done.

tensorboard/components/tensor_widget/selection.ts
caisq: Done.

tensorboard/components/tensor_widget/selection.ts
caisq: Making a defensive copy of  the slicing spec in the constructor now. Good catch. Thanks.

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,347 @@
...
+          (newSlicingSpec.horizontalRange as [number, number])[1]++;
+        }
+      }
+    }
+    // Moving the selection causes the selection size to collapse to 1x1.
+    this.rowCount = 1;
+    this.colCount = 1;
+
+    if (newSlicingSpec !== null) {
+      this.slicingSpec = newSlicingSpec;
```
davidsoergel: I'm still concerned about this: TensorWidgetImpl holds one slicingSpec, and TensorElementSelection holds another.  These stay in sync only if TensorWidgetImpl is the only caller of move(), and if it always updates itself with the return value.  `TensorWidgetImpl.selection` is protected, so that helps; making it private would alleviate this a bit more.  But I think it would be more robust to pass an `updateSlicingSpec()` callback to the constructor, and then call it here.

Furthermore TensorElementSelection should not retain the slicingSpec reference from the constructor at all-- if the parent TensorWidgetImpl updated its version for any reason other than move(), the one here would still point to the old one.  Maybe you always throw out the selection in this case?  Anyway it feels entangled/brittle.

tensorboard/components/tensor_widget/tensor-widget-impl.ts
```diff
@@ -521,16 +623,152 @@ export class TensorWidgetImpl implements TensorWidget {
...
+          if (status.rightEdge) {
+            valueDiv.classList.add('tensor-widget-value-div-selection-right');
+          }
+        }
+      }
+    }
+  }
+
+  /**
+   * Calcualte the set of indices that a value div currently maps to.
```
davidsoergel: typo

tensorboard/components/tensor_widget/tensor-widget-impl.ts
```diff
@@ -521,16 +623,152 @@ export class TensorWidgetImpl implements TensorWidget {
...
+   * @param row Row index of the value div, 0-based. This is with respect
+   *   to the 2D array of value divs that the widget currently possess, not
+   *   with respect to the indices of the underlying tensor. Same below.
+   * @param col Column indices of the value div, 0-based.
+   * @return The set of indices of the underlying tensor.
+   */
+  private calculateIndices(row: number, col: number): number[] {
```
davidsoergel: How about `viewRow` (vs. `tensorRow`), and similarly `calculateTensorIndices()`, to make that distinction as clear as possible

tensorboard/components/tensor_widget/tensor-widget-impl.ts
```diff
@@ -521,16 +623,152 @@ export class TensorWidgetImpl implements TensorWidget {
...
+          );
+        }
+        indices.push(index);
+      } else if (i === this.slicingSpec.viewingDims[0]) {
+        if (
+          this.slicingSpec.verticalRange === null ||
+          this.slicingSpec.verticalRange[0] === null
+        ) {
+          throw new Error(
+            'Failed to calculate indices to do undertermined vertical range.'
```
davidsoergel: typo, and I can't quite parse the error message (here and below)

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -225,26 +221,25 @@ export class TensorElementSelection {
...
-   *   slicing spec (i.e., if the new single element selection is within
-   *   the current slicing spec), `null`. Else, a new `TensorViewSlicingSpec`
-   *   object that can accommodate the new single-element selection.
    */
-  public move(direction: SelectionMoveDirection): TensorViewSlicingSpec | null {
-    // let viewRangeChanged = false;
-    let newSlicingSpec: TensorViewSlicingSpec | null = null;
+  public move(direction: SelectionMoveDirection): void {
+    let viewRangeChanged = false;
+    // let newSlicingSpec: TensorViewSlicingSpec | null = null;
```
davidsoergel: Per offline discussion: reinstate this, as a copy of a slicingSpec passed to move(), and remove this.slicingSpec entirely.  Then returning newSlicingSpec is fine after all; no need for the callback either.  Thanks and sorry again for going in circles :)

caisq: @davidsoergel I made a more thorough clean up: `move()` doesn't really return a slicingSpec anymore. It returns a MoveDirection enum to tell the caller which direction the slicingSpec should move. It's up to the caller to actually update the slicingSpec. The caller doesn't really have to update it.

There gets rid of the unclear ownership, removes the need to copy things and makes testing easier.

davidsoergel: Sweet

tensorboard/components/tensor_widget/selection.ts
```diff
@@ -0,0 +1,301 @@
...
+   * @param current slicing spec.
+   * @return The direction in which the vertical or horizontal viewing range
+   *   of the slicing spec should change. This is just an advisory.
+   *   It is up to the caller to actually update the slicing spec.
+   */
+  public move(
+    direction: MoveDirection,
+    slicingSpec: TensorViewSlicingSpec
+  ): MoveDirection | null {
+    let moveDirection: MoveDirection | null = null;
```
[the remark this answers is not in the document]
caisq: Renamed it to `slicingMoveDirection`

Issue #2595, opened 2019-08-25
Multiple regressions in display of charts in scalars tab.


iansf:
## Environment information (required)

Please run `diagnose_tensorboard.py` (link below) in the same
environment from which you normally run TensorFlow/TensorBoard, and
paste the output here:

Run on Google's internal infrastructure using latest prebuilt tensorboard binaries.

For browser-related issues, please additionally specify:

  - Browser type and version: Chrome 76.0.3809.100 (Official Build) (64-bit)
  - Screenshot, if it’s a visual issue:

![image](https://user-images.githubusercontent.com/12631059/63644379-f76fae00-c69c-11e9-9796-826c0c5c1c26.png)

## Issue description

About a week ago, multiple regressions appeared in the scalars tab. One such regression is shown in the attached screenshot. This occurs intermittently when refreshing with the tensorboard refresh button.

Another regression is that, when refreshing, charts often fail to automatically expand their x dimension to accommodate new data.

Another regression is that closing and reopening a particular panel of charts used to refresh the charts. Now the charts remain cached across panel closes. This removed the easiest way to fix situations where multiple charts had been zoomed or panned, or were not updating correctly. Instead, the only way I have found to reset the charts (besides a full page reload) is by entering a random value into the Runs regex field such that no runs are selected, and then clearing out that value. This is much less convenient than just closing and reopening the problematic panel.

iansf:
The tiny charts issue is not resolved in the Google prebuilt binary as of today:
![image](https://user-images.githubusercontent.com/12631059/64300106-7a5ff680-cf30-11e9-9858-4fc6da41aac5.png)

Perhaps the difference between my setup and what you test with is that my experiments have a large number of scalars (in the hundreds)? As you can see in both screenshots, I also am increasing the maximum number of plots a single pane shows above the default of 12.

All of these issues started occurring for me at the same time about a week before I submitted this bug, which is why I claim that they are multiple recent regressions. I have been running internal builds of tensorboard with large numbers of scalars for a couple of years at this point. I was not hitting these particular issues until that point. Regardless of when the regressions occurred, however, they are all quite annoying bugs.

stephanwlee:
Hi @iansf, sorry for the late reply.

> Run on Google's internal infrastructure using latest prebuilt tensorboard binaries.
> The tiny charts issue is not resolved in the Google prebuilt binary as of today:

[...]

PR #2605, opened 2019-08-28
scalar: fix tiny chart

stephanwlee:
This change updates `active` prop to a template of a
tf-category-paginated-view when the category pane is collapsed/expanded.
It also updates the `active` when a component is no longer mounted onto
the DOM.

The change does not contain any redraw logic but it magically works
because the DataLoaderBehavior, when active changes to active, fetches
data point for ones that are not in cache and then triggers rerender.
This flow of logic is indirect and thus can be made to be more direct
but that can cause too many "redraw" call which can be quite expensive
(scales poorly with # of runs).

Related to #2595.

### DIFF
--- a/tensorboard/components/tf_paginated_view/tf-dom-repeat.html
+++ b/tensorboard/components/tf_paginated_view/tf-dom-repeat.html
@@ -48,6 +48,15 @@ limitations under the License.
           value: 'item',
         },
 
+        /**
+         * Whether all stamped items are active or not.
+         * @protected
+         */
+        _contentActive: {
+          type: Boolean,
+          value: true,
+        },
+
         _domBootstrapped: {
           type: Boolean,
           value: false,
@@ -97,6 +106,7 @@ limitations under the License.
       observers: [
         '_bootstrapDom(_itemsRendered, isAttached)',
         '_updateDom(_renderedItems.*, _domBootstrapped)',
+        '_updateActive(_contentActive)',
         '_trimCache(_cacheSize)',
       ],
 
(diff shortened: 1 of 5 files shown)


tensorboard/components/tf_paginated_view/tf-dom-repeat.html
```diff
@@ -166,6 +176,14 @@
         this._domBootstrapped = true;
       },
 
+      _updateActive() {
+        if (!this._domBootstrapped) return;
+
+        Array.from(this._renderedTemplateInst.values()).forEach((inst) => {
+          inst.notifyPath('active', this._contentActive, true /* fromAbove */);
```
wchargin: I’m trying to find out what `fromAbove` does, but I couldn’t find any
actual API docs for `notifyPath`, and what I presume is the definition
only takes `path` and `props`:

https://github.com/Polymer/polymer/blob/54da98b18b3b613c9fcc400bd8f8590a1ae6a303/lib/mixins/property-effects.html#L275

I also don’t see any 3-argument `notifyPath` invocations in the
implementation of `dom-repeat`.

Where is this defined or documented? How can I learn more about it?

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-dashboard.html
```diff
@@ -156,7 +156,7 @@ <h3>No scalar data was found.</h3>
             >
               <template>
                 <tf-scalar-card
-                  active="[[active]"
+                  active="[[active]]"
```
wchargin: …was this just like `active="true"` always?

tensorboard/components/tf_paginated_view/tf-dom-repeat.html
wchargin: Per discussion: looks like this is a Polymer 1.x-ism.

PR #2607, opened 2019-08-29
Add ability to set custom distance function for counterfactuals

tolga-b:
WIT did not support counterfactual computation for image and text features as it is not straightforward to define the similarity of two images or two sentences. This PR adds the ability to pass a custom python function for similarity computation such that WIT can query that function to compute counterfactuals. This allows the flexibility for users to define and experiment with their custom similary metrics and opens the option to load even a separate deep learning model for similarity purposes.
In addition, smile detection demo is updated to use tfjs MobileNet model embeddings for example distances, allowing counterfactuals using image features (previously this demo only used tabular data for counterfactuals while smile detection model only used the image.)
 
WitConfig now has set_custom_distance_fn to pass a python function.
If a custom distance function is not set, WIT behaves the same as before (L1/L2 distance).
Boilerplate code to request distances from js side and call a js callback function from the python side when the distances are ready.

* Screenshots of UI changes
If custom distance function is set, L1 L2 distance selection is greyed out and a third option "User-specified" appears. The same is true for show similarity to selected datapoint option.
Counterfactuals:
cf_mobilenet
Show similarity to selected datapoint pane:
pane_cf
Facets scatterplot w.r.t MobileNet embedding distance:
custom_distance

Verified jupyter notebook and colab notebook usage by creating a Smile detection demo with custom distance method that uses the difference in average image color for distance between two images.
Ran all browser demos, including the new smile detection demo that uses custom distance function.

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -3253,6 +3253,13 @@ limitations under the License.
     const inferenceSquaredErrorStr = ' Inference squared error';
     const inferenceScoreStr = ' Inference score';
     const exampleIdKey = ' Datapoint ID';
+    // requestDistanceWithCallback can be invoked with the callbacks in this
+    // enum. Please add a function to this list if you want it to be invoked
+    // at the end of custom distance computation.
+    const distanceCallbacks = {
+      finalizeClosestCounterfactual: 0,
+      finalizeAddDistanceMetric: 1,
+    };
 
     // Special feature names and feature name prefixes and suffixes for parsing
     // inference and attribution information from TFMA-generated input.
@@ -3262,6 +3269,7 @@ limitations under the License.
     const featureForAttributionPrefix = 'features__';
     const attributionSparseKeySuffix = '_keys';
     const attributionSparseValueSuffix = '_values';
+
 
     function deleteElement(elt) {
       if (elt && elt.parentElement) {
@@ -3825,7 +3833,7 @@ limitations under the License.
           if (this.local && this.customDistanceFunctionSet) {
             this.requestDistanceWithCallback(
               selected,
-              'finalizeAddDistanceMetric',
+              distanceCallbacks.finalizeAddDistanceMetric,
               params
             );
             return;
@@ -3847,6 +3855,17 @@ limitations under the License.
               params.callbackParams
             );
           }
+        },
+
+        invokeCustomDistanceCallback: function(callbackObj){
+          switch (callbackObj.funId) {
+            case distanceCallbacks.finalizeClosestCounterfactual:
+              this.finalizeClosestCounterfactual(callbackObj.exInd, callbackObj.distances, callbackObj.params);
+              break;
+            case distanceCallbacks.finalizeAddDistanceMetric:
+              this.finalizeAddDistanceMetric(callbackObj.exInd, callbackObj.distances, callbackObj.params);
+              break;
+          };
         },
 
         finalizeAddDistanceMetric: function(exInd, distances, callbackParams) {
@@ -3926,7 +3945,7 @@ limitations under the License.
           if (this.local && this.customDistanceFunctionSet) {
             this.requestDistanceWithCallback(
               selected,
-              'finalizeClosestCounterfactual',
+              distanceCallbacks.finalizeClosestCounterfactual,
               {callbackParams: {}, distanceParams: {}}
             );
             return;
@@ -3969,6 +3988,8 @@ limitations under the License.
 
         // Call backend for distance computation, backend calls callback function
         // with computed distances and parameters
+        // TODO (tolgab) If TensorBoard supports custom distance functions
+        // add ability to call into their backend as well (non-local mode).
         requestDistanceWithCallback: function(exInd, callbackFun, params) {
           const urlParams = {
             index: exInd,
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -1418,6 +1418,7 @@ limitations under the License.
                               Show nearest counterfactual datapoint
                             </paper-toggle-button>
                           </div>
+                          <template is="dom-if" if="[[customDistanceFunctionSet]]">
                           <paper-radio-group
                             selected="{{nearestCounterfactualDist}}"
                           >
@@ -1428,6 +1429,10 @@ limitations under the License.
                               >L2</paper-radio-button
                             >
                           </paper-radio-group>
+                          </template>
+                          <template is="dom-if" if="[[!customDistanceFunctionSet]]">
+                            Using custom distance function.
+                          </template>
                           <paper-dropdown-menu
                             label="Model:"
                             no-label-float
@@ -3819,7 +3824,8 @@ limitations under the License.
 
         findClosestCounterfactual_: function() {
           const selected = this.selected[0];
-          // Custom distance function can only be used when local
+          // Custom distance function can only be used when local.
+          // If using custom distance function, request distances and return.
           if (this.local && this.customDistanceFunctionSet) {
             this.requestDistanceWithCallback(
               selected,
@@ -3829,7 +3835,6 @@ limitations under the License.
             return;
           }
 
-          // If not using custom distance function compute counterfactuals here.
           const modelInferenceValueStr = this.strWithModelName_(
             inferenceValueStr,
             this.nearestCounterfactualModelIndex
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -1312,8 +1312,9 @@ limitations under the License.
       </paper-input>
       <div class="radiolabel">Distance type</div>
       <paper-radio-group class="dist-switch" selected="{{facetDistSwitch}}">
-        <paper-radio-button class="dist-radio" name="L1">L1</paper-radio-button>
-        <paper-radio-button class="dist-radio" name="L2">L2</paper-radio-button>
+        <paper-radio-button class="dist-radio" name="L1" disabled$=[[customDistanceFunctionSet]]>L1</paper-radio-button>
+        <paper-radio-button class="dist-radio" name="L2" disabled$=[[customDistanceFunctionSet]]>L2</paper-radio-button>
+        <paper-radio-button class="dist-radio" name="Custom" hidden$=[[!customDistanceFunctionSet]]>Custom</paper-radio-button>
       </paper-radio-group>
       <paper-dropdown-menu
         label="Apply to datapoints visualization"
@@ -1418,27 +1419,19 @@ limitations under the License.
                               Show nearest counterfactual datapoint
                             </paper-toggle-button>
                           </div>
-                          <template
-                            is="dom-if"
-                            if="[[!customDistanceFunctionSet]]"
+                          <paper-radio-group
+                            selected="{{nearestCounterfactualDist}}"
                           >
-                            <paper-radio-group
-                              selected="{{nearestCounterfactualDist}}"
+                            <paper-radio-button name="L1" disabled$="[[customDistanceFunctionSet]]"
+                              >L1</paper-radio-button
                             >
-                              <paper-radio-button name="L1"
-                                >L1</paper-radio-button
-                              >
-                              <paper-radio-button name="L2"
-                                >L2</paper-radio-button
-                              >
-                            </paper-radio-group>
-                          </template>
-                          <template
-                            is="dom-if"
-                            if="[[customDistanceFunctionSet]]"
-                          >
-                            Using custom distance function.
-                          </template>
+                            <paper-radio-button name="L2" disabled$="[[customDistanceFunctionSet]]"
+                              >L2</paper-radio-button
+                            >
+                            <paper-radio-button name="Custom" hidden$=[[!customDistanceFunctionSet]]
+                            >Custom</paper-radio-button
+                            >
+                          </paper-radio-group>
                           <paper-dropdown-menu
                             label="Model:"
                             no-label-float
@@ -1477,7 +1470,8 @@ limitations under the License.
                             <div>
                               Compares the selected datapoint with its nearest
                               neighbor from a different classification using L1
-                              or L2 distance.
+                              or L2 distance. If a custom distance function is set,
+                              it uses that function instead.
                             </div>
                           </paper-dialog>
                         </div>
@@ -3375,6 +3369,7 @@ limitations under the License.
           customDistanceFunctionSet: {
             type: Boolean,
             value: false,
+            observer: 'customDistanceFunctionSetChanged_',
           },
           // Feature for true label.
           selectedLabelFeature: {
@@ -3753,6 +3748,15 @@ limitations under the License.
 
         showDistanceClicked_: function() {
           this.$.distancedialog.toggle();
+        },
+        customDistanceFunctionSetChanged_: function() {
+          if (this.customDistanceFunctionSet){
+            this.nearestCounterfactualDist = 'Custom';
+            this.facetDistSwitch = 'Custom';
+          } else {
+            this.nearestCounterfactualDist = 'L1';
+            this.facetDistSwitch = 'L1'
+          }
         },
 
         requestAddDistanceMetric: function() {
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -1338,7 +1338,7 @@ limitations under the License.
         <paper-button
           dialog-confirm
           autofocus
-          on-tap="addDistanceMetric"
+          on-tap="requestAddDistanceMetric"
           class="main-button"
           >Apply</paper-button
         >
@@ -3783,7 +3783,7 @@ limitations under the License.
         },
 
         finalizeAddDistanceMetric: function(exInd, distances, callbackParams) {
-          const facetDistFeatureName = callbackParams.facetDistFeatureName;
+          const facetDistFeatureName = callbackParams.featureName;
           this.facetDistFeatureNames.push(facetDistFeatureName);
           for (let i = 0; i < this.visdata.length; i++) {
             this.visdata[i][facetDistFeatureName] = distances[i];
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -3746,7 +3746,7 @@ limitations under the License.
           let closest = -1;
           const useL2Distance = this.facetDistSwitch == 'L2';
           for (let i = 0; i < this.visdata.length; i++) {
-            let dist = this.getDist(
+            let dist = this.getDistL1L2(
               this.visdata[selected],
               this.visdata[i],
               useL2Distance
@@ -3784,6 +3784,7 @@ limitations under the License.
           }
         },
 
+
         findClosestCounterfactual_: function() {
           const selected = this.selected[0];
           const modelInferenceValueStr = this.strWithModelName_(
@@ -3801,11 +3802,9 @@ limitations under the License.
             ) {
               continue;
             }
-            let dist = this.getDist(
-              this.visdata[selected],
-              this.visdata[i],
-              this.nearestCounterfactualDist == 'L2'
-            );
+            // TODO(tolgab) benchmark this backend call, may need to pull it
+            // out of for loop for custom distance metric
+            let dist = this.getDist(selected, i, this.nearestCounterfactualDist == 'L2');
             if (dist < closestDist) {
               closestDist = dist;
               closest = i;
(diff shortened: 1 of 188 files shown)


tolga-b:
@jameswex I will do one more pass on this to check if everything works correctly (demos etc.). addDistance with custom distance function will have it's own PR. You can start checking it now or after that pass, thank you!

mpushkarna:
@tolga-b cc:@jameswex 
For a complete UI implementation, when users set a custom distance function:
(a) If users can only use the custom distance metric, radio buttons for L1 or L2 distances should be disabled _or_ 
(b) Users should be able to pick from L1, L2 and custom distance in a dropdown.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3784,8 +3784,52 @@ <h2>Show similarity to selected datapoint</h2>
...
+          if (this.local && this.customDistanceFunctionSet) {
+            this.requestDistanceWithCallback(
+              selected,
+              'computeClosestCounterfactual',
+              {callbackParams: {}, distanceParams: {}}
+            );
+            return;
+          }
+
+          // If not using custom distance function compute counterfactuals here.
```
jameswex: dont see the computation that the comment refers to

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
```diff
@@ -74,8 +77,11 @@ def __init__(self, config_builder):
       if 'compare_adjust_example' in config else None)
     if 'custom_predict_fn' in copied_config:
       del copied_config['custom_predict_fn']
+      copied_config['uses_custom_distance_fn'] = True
```
jameswex: why is this set here?

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
```diff
@@ -111,6 +117,12 @@ def set_examples(self, examples):
     self.examples = [json_format.MessageToJson(ex) for ex in examples]
     self.updated_example_indices = set(range(len(examples)))
 
+  def compute_custom_distance_impl(self, index, params=None):
+    exs_for_distance = [
+        self.json_to_proto(example) for example in self.examples]
```
jameswex: should we convert to proto if the model accepts json and not tf examples?

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
```diff
@@ -97,8 +106,15 @@ def infer_mutants(wit_id, details):
         wit.labelVocab = inferences.label_vocab;
         wit.inferences = inferences.inferences;
         wit.attributions = {{indices: wit.inferences.indices,
-                            attributions: inferences.attributions}}
+                            attributions: inferences.attributions}};
       }};
+
+      window.distanceCallback = callback_dict => {{
```
jameswex: use camelCase in the js code

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
```diff
@@ -256,6 +277,13 @@ def duplicate_example(self, index):
     self.updated_example_indices.add(len(self.examples) - 1)
     self._generate_sprite()
 
+  def compute_custom_distance(self, index, callback_fn, params):
+    distances = base.WitWidgetBase.compute_custom_distance_impl(self, index,
+                                                                params['distanceParams'])
+    callback_dict = {'distances':distances, 'exInd':index, 'callback_fn':callback_fn, 'params':params['callbackParams']}
```
jameswex: line overflow?

tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/wit.py
```diff
@@ -119,6 +121,22 @@ def _delete_example(self, change):
...
+    params = info['params']
+    callback_fn = info['callback']
+    try:
+      distances = base.WitWidgetBase.compute_custom_distance_impl(self, index,
+                                                       params['distanceParams'])
+      self.custom_distance_dict = {'distances':distances,
+                                   'exInd':index,
+                                   'callback_fn':callback_fn,
+                                   'params':params['callbackParams']}
+    except Exception as e:
```
jameswex: jupyter has error catching and colab version doesn't. do we have this issue with other methods?

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
```diff
@@ -499,6 +499,27 @@ def set_compare_custom_predict_fn(self, predict_fn):
       self.set_compare_model_name('2')
     return self
 
+  def set_custom_distance_fn(self, distance_fn):
+    """Sets a custom function for distance computation.
+
+    WIT can directly use a custom function for all distance computations within
+    the tool. In this case, the provided function should accept a query example
+    proto and a list of example protos to compute the distance against and
```
jameswex: do we have the ability for the model to take provided json and not examples? i forget. if we do, we would want to do that here as well.

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
```diff
@@ -499,6 +499,27 @@ def set_compare_custom_predict_fn(self, predict_fn):
...
+
+    Returns:
+      self, in order to enabled method chaining.
+    """
+    if distance_fn is None:
+      self.delete('custom_distance_fn')
+    else:
+      self.store('custom_distance_fn', distance_fn)
+    return self
+
```
jameswex: can you update the smiling demo to use some sort of simple distance measure from the image pixels? or a pre-trained tf.js model that gives image embeddings :)

jameswex:
> @tolga-b cc:@jameswex
> For a complete UI implementation, when users set a custom distance function:
> (a) If users can only use the custom distance metric, radio buttons for L1 or L2 distances should be disabled _or_
> (b) Users should be able to pick from L1, L2 and custom distance in a dropdown.

I think if users set their own custom distance measure that we can always use that and remove the controls for l1 and l2, as in these cases the auto-computed measure doesn't make sense (like we currently see in our text and image examples).

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
tolga-b: would you like me to set it where I set custom_predict_fn?
Setting custom_predict_fn should normally be enough, but we have to remove it from copied_config since it cannot be passed to js side so I added this extra key to convey that information.

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
tolga-b: did we add the ability to support json for custom_predict_fn? I assumed that since WIT always uses tf.example internally keep the api of custom_predict_fn and custom_distance_fn consistent and add json support for both of them at the same time. I can also see why it is useful other way around too though (since user only deals with json for json models).

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
jameswex: you're setting uses_custom_distance_fn if 'custom_predict_fn' in copied_config, but use_custom_distance_fn is unrelated to if the model has a custom predict function.

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
jameswex: we probably don't have that ability yet, so no need to do anything now. 

i think we should do a PR at some point soon to allow custom_predict_fns to be able to handle json (if the examples are provided as json to begin with). but you're right, its separate work.

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
tolga-b: ups you are correct, thats a good catch. Should test notebook demos without custom_predict_fn as well.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
tolga-b: I meant the whole block, could do something nicer in future to avoid code duplication. Moved the comment up.

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
tolga-b: yes, good idea :)

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
tolga-b: done.

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
tolga-b: done.

tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/wit.py
tolga-b: added error catching to colab as well.

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
tolga-b: I think this is resolved with your other comment, we will add json ability in a future PR.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3360,6 +3371,11 @@ <h2>Show similarity to selected datapoint</h2>
             value: '',
             observer: 'breakdownFeatureSelected_',
           },
+          // True if an example has been updated.
```
jameswex: comment wrong

jameswex:
what is left to get this fully working with appropriate UI? if nothing, then can you also update the smiling demo and a notebook demo with custom distance so it can be tested in this PR? Thanks.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -1418,16 +1418,27 @@ <h2>Show similarity to selected datapoint</h2>
                               Show nearest counterfactual datapoint
                             </paper-toggle-button>
                           </div>
-                          <paper-radio-group
-                            selected="{{nearestCounterfactualDist}}"
+                          <template
+                            is="dom-if"
+                            if="[[customDistanceFunctionSet]]"
```
jameswex: is this if statement backwards?

tolga-b: the commits are not complete yet. I will let you know when it is done.

tensorboard/plugins/interactive_inference/witwidget/notebook/visualization.py
tolga-b: I am deferring this in the interest of time.

jameswex:
@stephanwlee I have reviewed this code. would you or another TB team member also like to before it is merged?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3784,8 +3853,52 @@ <h2>Show similarity to selected datapoint</h2>
           }
         },
 
+        computeClosestCounterfactual: function(exInd, distances) {
```
stephanwlee: readability nit: prefix `compute` means something in Polymer and it often means a function that is side-effect-free. Can you please rename it?

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
```diff
@@ -60,6 +60,9 @@ def __init__(self, config_builder):
     self.compare_custom_predict_fn = (
       config.get('compare_custom_predict_fn')
       if 'compare_custom_predict_fn' in config else None)
+    self.custom_distance_fn = (
+      config.get('custom_distance_fn')
```
stephanwlee: sorry, did not read the code yet but if this is dict, can't you just say `config.get('custom_distance_fn', None)`?

third_party/js.bzl
```diff
@@ -17,500 +17,509 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
...
+            ],
+        },
+    )
+
+    filegroup_external(
+        name = "org_tensorflow_tfjs_mobilenet",
+        licenses = ["notice"],  # Apache 2.0
+        sha256_urls = {
+            "aca32ec8a778d17acb623bf495c6e60ec38312b761b9a8c88af4c841c25eba70": [
+                "https://cdn.jsdelivr.net/npm/@tensorflow-models/mobilenet@1.0.0/dist/mobilenet.js",
```
stephanwlee: Please use the tensorflow.org mirror in addition to this.

tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/lib/wit.js
```diff
@@ -118,14 +123,19 @@ var WITView = widgets.DOMWidgetView.extend({
...
     this.inferMutantsCounter = 0;
     this.view_.addEventListener('infer-mutants', (e) => {
       e.detail['infer_mutants_counter'] = this.inferMutantsCounter++;
       this.model.set('infer_mutants', e.detail);
       this.mutantFeature = e.detail.feature_name;
       this.touch();
     });
+    this.computeDistanceCounter = 0;
+    this.view_.addEventListener('compute-custom-distance', (e) => {
+      e.detail['compute_distance_counter'] = this.computeDistanceCounter++;
```
stephanwlee: this feels quite odd and not clean. Would it be possible to use data binding instead?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3784,8 +3853,52 @@ <h2>Show similarity to selected datapoint</h2>
...
+        },
+
         findClosestCounterfactual_: function() {
           const selected = this.selected[0];
+          // Custom distance function can only be used when local.
+          // If using custom distance function, request distances and return.
+          if (this.local && this.customDistanceFunctionSet) {
+            this.requestDistanceWithCallback(
+              selected,
+              'computeClosestCounterfactual',
```
stephanwlee: I am super surprised that you cannot pass the reference to the function. Can you not tack on function to the Event object?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
tolga-b: Didn't know, thanks. Renamed it.

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
tolga-b: good catch, I just repeated the pattern from other keys. I now changed all other instances of it as well.

tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/lib/wit.js
tolga-b: The counter is there as it is in the case of infer-mutants above. We use data binding for what is passed ('compute_custom_distance' is set to e.detail). However, we only pass the example id back to the python side instead of the full example since examples are already in sync between the backend and frontend. In this case, if someone requests the same example again after updating some features, the change on bound dict may not trigger since it is the same dict. We force trigger it by incrementing the counter.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
tolga-b: This function sends request to the backend to compute distances on the python side. It is called by multiple functions from js side so the python side needs to know which callback function to call on the js side after the computation ends. Flow is js(request distance for x) -> python(compute distance) -> js(finalize x with distances). Is there a way to pass a js function reference back to python?

third_party/js.bzl
tolga-b: do you know what that mirror url would be? This is an npm package here https://www.npmjs.com/package/@tensorflow-models/mobilenet/v/1.0.0

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3854,6 +3967,17 @@ <h2>Show similarity to selected datapoint</h2>
...
 
+        // Call backend for distance computation, backend calls callback function
+        // with computed distances and parameters
+        requestDistanceWithCallback: function(exInd, callbackFun, params) {
+          const urlParams = {
+            index: exInd,
+            callback: callbackFun,
+            params: params,
+          };
+          this.fire('compute-custom-distance', urlParams);
```
stephanwlee: I find it weird to see notebook specific code in the main dashboard code. What is the expected behavior outside of the notebook?

tolga-b: requestDistanceWithCallback would not be invoked in non-local instances (local=demos and notebook). In case we are in non-local mode, WIT defaults to it's previous behavior of computing counterfactuals completely on the js side with L1 and L2 distance between examples. This is slightly similar in terms of behavior to custom_predict_fn where it is only supported in notebook mode.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3817,8 +3886,52 @@ <h2>Show similarity to selected datapoint</h2>
...
+        },
+
         findClosestCounterfactual_: function() {
           const selected = this.selected[0];
+          // Custom distance function can only be used when local.
+          // If using custom distance function, request distances and return.
+          if (this.local && this.customDistanceFunctionSet) {
+            this.requestDistanceWithCallback(
+              selected,
+              'finalizeClosestCounterfactual',
```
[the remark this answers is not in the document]
tolga-b: Thank you Stephan.
`http://mirror.tensorflow.org/cdn.jsdelivr.net/npm/@tensorflow-models/mobilenet@1.0.0/dist/mobilenet.js`
does not seem to open, is this normal?
"This XML file does not appear to have any style information associated with it. The document tree is shown below."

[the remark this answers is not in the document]
stephanwlee: When you are ready to submit, we will update the mirror to contain that file.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
tolga-b: Added TODO note here so that if there is a support in TensorBoard mode to provide custom distance functions for counterfactuals then we should update this function to reflect that.

third_party/js.bzl
tolga-b: done.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
tolga-b: Added TODO note here so that if there is a support in TensorBoard mode to provide custom distance functions for counterfactuals then we should update this function to reflect that.

tolga-b:
Thank you, @stephanwlee I addressed the review comments. Could you please check it out?

third_party/js.bzl
```diff
@@ -494,8 +494,19 @@ def tensorboard_js_workspace():
       name = "org_tensorflow_tfjs",
       licenses = ["notice"],  # Apache 2.0
       sha256_urls = {
-          "fccd26db2da462ec48e2d90fbdff1ee9a9d740f2c7efbd9789ba46eb98ecd1ae": [
-              "https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@0.11.5/dist/tf.min.js",
+          "3870fc1d4eaefce7370b5bce3992e4dadcad8e3a5999b034302e38b2094873ea": [
+              "https://cdn.jsdelivr.net/npm/@tensorflow/tfjs@1.0.1/dist/tf.min.js",
```
stephanwlee: Please mirror this link using tensorflow.org too.

tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/lib/wit.js
stephanwlee: Sorry but I must confess I am very lost. Even if this logic is correct, I do think this has readability problem.

AFAICT, the event is fired from JavaScript side. When firing the event, the object does not contain the property `compute_distance_counter`. 1. how is it being used by owner of the event object? 2. where does the Python come into play?

tolga-b: The dashboard launches this event from js side by passing some parameters (example index, callback ind etc.).  "compute_custom_distance" is sync variable between js and python sides. This synchronization is managed by jupyter widget framework. When we call this.model.set('compute_custom_distance', e.detail), ideally jupyter should recognize the change in "compute_custom_distance" and trigger distance computation. James and I found out previously (for infer-mutants) that if someone edits an example then fires this event again from the dashboard, since e.detail does not contain the contents of the example but only it's index, jupyter thinks nothing has changed and model.set does not trigger distance computation. James solved it by adding a counter so that we make sure the serialized dictionary always has one field that changes between calls and jupyter triggers the sync.
Since this is a jupyter specific issue, we did not include this in the event but append it in wit.js related to jupyter.

third_party/js.bzl
tolga-b: done.

tolga-b:
Thank you Stephan, I removed the side effect in wit.js event object per your suggestion!

tensorboard/plugins/interactive_inference/witwidget/notebook/jupyter/js/lib/wit.js
```diff
@@ -118,19 +123,32 @@ var WITView = widgets.DOMWidgetView.extend({
...
-
     this.inferMutantsCounter = 0;
     this.view_.addEventListener('infer-mutants', (e) => {
-      e.detail['infer_mutants_counter'] = this.inferMutantsCounter++;
-      this.model.set('infer_mutants', e.detail);
+      this.model.set(
+        'infer_mutants',
+        Object.assign({}, e.detail, {
+          infer_mutants_counter_for_busting_cache: this.inferMutantsCounter++,
+        })
```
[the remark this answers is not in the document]
tolga-b: Good to know, thank you Stephan!

Issue #719, opened 2017-11-10
Protobuf breaks bazel fetch


abbradar:
OS: NixOS 18.03pre119292.cfafd6f5a8
TensorBoard version: 0.1.8
Bazel: 0.4.5

```
$ bazel fetch tensorboard:tensorboard
ERROR: /run/user/1000/.cache/bazel/_bazel_abbradar/84fb9207e5a69b2add72d42fc5eb546e/external/protobuf/BUILD:623:1: no such target '//external:python_headers': target 'python_headers' not declared in package 'external' defined by /home/abbradar/nixpkgs/tensorboard/WORKSPACE and referenced by '@protobuf//:python/google/protobuf/pyext/_message.so'.
ERROR: /run/user/1000/.cache/bazel/_bazel_abbradar/84fb9207e5a69b2add72d42fc5eb546e/external/protobuf/BUILD:609:1: no such target '//external:python_headers': target 'python_headers' not declared in package 'external' defined by /home/abbradar/nixpkgs/tensorboard/WORKSPACE and referenced by '@protobuf//:python/google/protobuf/internal/_api_implementation.so'.
ERROR: Evaluation of query "deps(tensorboard:tensorboard)" failed: errors were encountered while computing transitive closure.
```

jart:
Does installing the Python headers package work?

abbradar:
No, I have Python headers installed (but they are installed in an unconventional place, should still be picked up). How does Bazel detect is there are needed headers installed?

jart:
Protobuf might want us to have something like this in our `third_party/workspace.bzl`:

```py
  native.bind(
      name = "python_headers",
      actual = str(Label("//util/python:python_headers")),
  )
```

Which is most likely super painful to set up. So I would ask:

1. Do all the `bazel build` commands you want to run, work? Please note that building fetches automatically.
2. Is copying the headers to the standard path a viable workaround?

abbradar:
Unfortunately, (1) is not an option because we do our builds in a hermetized environment without a network connection. We do this in two stages -- first `fetch` with checksumming and then `build`.

(2) is not an option -- builds are done in a standartized environment (Nix build manager).

jart:
That's a legitimate use case.

[...]

PR #2609, opened 2019-08-29
Upgrade protobuf to fix compilation warnings

wchargin:
Summary:
We upgrade TensorFlow, which defines our protobuf dependency, past
508f76b1d99 (actual dep upgrade), f4e64ca17ce (build fix), and
dad4197fcf0 (workspace fix).

Test Plan:
Run `bazel build @com_google_protobuf//:protoc_lib` using Bazel 0.29.0,
which now completes silently, but previously would print a bunch of
compiler warnings in `descriptor.cc`.

wchargin-branch: protobuf-fix-warnings

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -82,11 +82,11 @@ install_bazel_dependencies()
 
 http_archive(
     name = "org_tensorflow",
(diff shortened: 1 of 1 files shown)


wchargin:
Can reproduce CI failure, which occurs only when using `bazel fetch`
explicitly. Seems to be #719. Investigating.

PR #2610, opened 2019-08-30
build: make ng source sync-able

stephanwlee:
In this change:
- made tsconfig more strict to roughly reflect internal one
- TypeScript code change to conform to the stricer rule
- created "expected_*" target to denote internal compat dependency graph
- defined missing bazel deps
- the type hacks are annotated for easier migration
- `export` annotation for JSCompiler compatibility.

Corresponding change is cl/266471999.
I have tried `bazel test tensorboard/components/tf_ng_tensorboard/...` both internally and externally after the change.

### DIFF
--- a/tensorboard/components/tf_ng_tensorboard/angular/BUILD
+++ b/tensorboard/components/tf_ng_tensorboard/angular/BUILD
@@ -1,0 +1,62 @@
+package(default_visibility = ["//tensorboard:internal"])
+
+load("//tensorboard/defs:defs.bzl", "tf_ts_library")
+
+licenses(["notice"])  # Apache 2.0
+
+# This is a dummy rule used as a @angular/common/http dependency.
+# This is not a replacement for @angular/common dependency.
+tf_ts_library(
+    name = "expect_angular_common_http",
+    srcs = [],
+)
+
+# This is a dummy rule used as a @angular/common/testing dependency.
+# This is not a replacement for @angular/common dependency.
+tf_ts_library(
+    name = "expect_angular_common_http_testing",
+    srcs = [],
+)
+
+# This is a dummy rule used as a @angular/core/testing dependency.
+# This is not a replacement for @angular/core dependency.
+tf_ts_library(
+    name = "expect_angular_core_testing",
+    srcs = [],
+    deps = [
+        "@npm//@angular/core",
+    ],
+)
+
+# This is a dummy rule used as a @angular/material/tabs dependency.
+tf_ts_library(
+    name = "expect_angular_material_tabs",
+    srcs = [],
+    deps = [
+        "@npm//@angular/material",
+    ],
+)
+
+# This is a dummy rule used as a @angular/material/toolbar dependency.
+tf_ts_library(
+    name = "expect_angular_material_toolbar",
+    srcs = [],
+    deps = [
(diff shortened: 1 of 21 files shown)


tensorboard/components/tf_ng_tensorboard/app.component.ts
```diff
@@ -17,7 +17,7 @@ import {Store} from '@ngrx/store';
 import {State} from './core/core.reducers';
 import {coreLoaded} from './core/core.actions';
 
-import * as _typeHackRxjs from 'rxjs';
+/** @typehack */ import * as _typeHackRxjs from 'rxjs';
```
caisq: I'm not familiar with this annotation. Can you explain, mainly for my edification?

tensorboard/components/tf_ng_tensorboard/angular/BUILD
```diff
@@ -0,0 +1,62 @@
...
+tf_ts_library(
+    name = "expect_angular_common_http",
+    srcs = [],
+)
+
+# This is a dummy rule used as a @angular/common/testing dependency.
+# This is not a replacement for @angular/common dependency.
+tf_ts_library(
+    name = "expect_angular_common_http_testing",
+    srcs = [],
```
caisq: Might be a dumb question: should  this declare a dependency on something, like an @npm package?

tsconfig.json
```diff
@@ -6,6 +6,9 @@
     "noResolve": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
+    "strictNullChecks": true,
+    "noImplicitAny": true,
```
caisq: nit: might be good to sort this alphabetically.

tensorboard/components/tf_ng_tensorboard/app.component.ts
stephanwlee: These hacks are not required and, in fact, it gets in the way of syncing. For more detail, I can talk to you offline.

tensorboard/components/tf_ng_tensorboard/angular/BUILD
stephanwlee: I worry that this statement will be used as a replacement for correct dependency. `@angular/common` should be separately depended on when you are using, for instance, `import {CommonModule} from '@angular/common'`. I can be convinced otherwise.

tensorboard/components/tf_ng_tensorboard/app.component.ts
caisq: The `_typeHackRxjs` isn't used AFAICT. Can it be simplified as

```
/** @typehack */ import 'rxjs';
```
?

stephanwlee: Nope, I confess I do not understand the nature of the TSC error but it does not like that.

caisq: OK. Got it. Can you document this for other developers so they'll know what's going on here?

stephanwlee: It was documented in some places in a GitHub issue that I filed. I am not sure if there is a value in repeating this link to GitHub over and over.

PR #2611, opened 2019-08-31
Properly handle bad requests to plugin data endpoints

wchargin:
Summary:
Most\* plugin endpoints for serving actual data have historically tried
to access data from the multiplexer with no validation or error
handling, and as a result would crash and burn (with 500s) given a run
or tag that doesn’t exist. This actually happens fairly commonly: if you
start TensorBoard and interact with the web UI, kill the server while
leaving the browser tab open, and start a new TensorBoard on the same
port but a different log directory, then the old TensorBoard’s refresh
loop will request data from the runs and tags that were previously
available.

These are now properly handled as 400s, not 500s, and they don’t spam
the command line anymore, either. We resolve some TODOs from 2015!

\* Weighted by frequency of use. :-) By raw counts, more plugins behaved
well than ill!

Test Plan:
Visit the following endpoints, and verify that each one responds with a
500 Internal Server Error before this patch, but a 400 Bad Request (with
human-readable message) after this patch:

  - <http://localhost:6006/data/plugin/audio/audio>
  - <http://localhost:6006/data/plugin/audio/individualAudio>
  - <http://localhost:6006/data/plugin/example_basic/greetings?run=&tag=>
  - <http://localhost:6006/data/plugin/images/images>
  - <http://localhost:6006/data/plugin/images/individualImage>
  - <http://localhost:6006/data/plugin/mesh/data?content_type=>
  - <http://localhost:6006/data/plugin/scalars/scalars>

wchargin-branch: keyerror-400

### DIFF
--- a/tensorboard/examples/plugins/example_basic/tensorboard_plugin_example/plugin.py
+++ b/tensorboard/examples/plugins/example_basic/tensorboard_plugin_example/plugin.py
@@ -74,12 +74,21 @@ class ExamplePlugin(base_plugin.TBPlugin):
 
   @wrappers.Request.application
   def _serve_greetings(self, request):
-    run = request.args["run"]
-    tag = request.args["tag"]
-    data = [
-        np.asscalar(tensor_util.make_ndarray(event.tensor_proto))
-            .decode("utf-8")
-        for event in self._multiplexer.Tensors(run, tag)
-    ]
+    run = request.args.get("run")
+    tag = request.args.get("tag")
+    if run is None or tag is None:
+      return werkzeug.Response(
+          "Must specify run and tag",
+          content_type="text/plain",
+          code=400,
+      )
+    try:
+      data = [
+          np.asscalar(tensor_util.make_ndarray(event.tensor_proto))
+              .decode("utf-8")
+          for event in self._multiplexer.Tensors(run, tag)
+      ]
+    except KeyError:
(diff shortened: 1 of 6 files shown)


tensorboard/examples/plugins/example_basic/tensorboard_plugin_example/plugin.py
```diff
@@ -74,12 +74,21 @@ def _serve_tags(self, request):
...
+          code=400,
+      )
+    try:
+      data = [
+          np.asscalar(tensor_util.make_ndarray(event.tensor_proto))
+              .decode("utf-8")
+          for event in self._multiplexer.Tensors(run, tag)
+      ]
+    except KeyError:
+      raise werkzeug.exceptions.BadRequest("Invalid run or tag")
```
stephanwlee: just wondering, HTTP 400 seems to be the BAD_REQUEST. Why do we opt out of using `werkzeug.exceptions.BadRequest` in line 80?

wchargin: …I can’t remember why I might have done that. Will change them both to
use the exception formulation. Thanks for the catch.

(Prior to this patch, lines 77 and 78 actually were already responding
with 400 Bad Request on failure (`request.args` is a magic container
whose key error type is `exceptions.BadRequest`), but I thought that
they should get a nicer message. But they should use the same form.)

PR #2616, opened 2019-09-06
ng: render inactive plugins

stephanwlee:
Inactive plugins are supposed to show up in the dropdown list.
This change renders the inactive plugins correctly.

Screenshot:
image

### DIFF
--- a/tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
+++ b/tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
@@ -13,29 +13,40 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {Component} from '@angular/core';
+import {MatTabChangeEvent} from '@angular/material/tabs';
+import {MatSelectChange} from '@angular/material/select';
 import {Store, select, createSelector} from '@ngrx/store';
 import {combineLatest, of} from 'rxjs';
 
 import {State, getActivePlugin, getPlugins} from '../../core/core.reducers';
 import {changePlugin} from '../../core/core.actions';
 
+import {PluginMetadata, PluginId} from '../../types/api';
+
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
+
+interface UiPluginMetadata extends PluginMetadata {
+  id: PluginId;
+}
 
 const selectPlugins = createSelector(
   getPlugins,
-  (listing) =>
+  (listing): UiPluginMetadata[] =>
     Object.keys(listing).map((key) =>
       Object.assign({}, {id: key}, listing[key])
     )
 );
 
+const selectDisabledPlugins = createSelector(
+  selectPlugins,
+  (plugins): UiPluginMetadata[] => plugins.filter((plugin) => !plugin.enabled)
+);
+
 const selectActivePluginIndex = createSelector(
-  getPlugins,
+  selectPlugins,
   getActivePlugin,
   (plugins, activePlugin) => {
-    return Object.keys(plugins).findIndex(
-      (pluginId) => pluginId === activePlugin
-    );
+    return plugins.findIndex((plugin) => plugin.id === activePlugin);
   }
 );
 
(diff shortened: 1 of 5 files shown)


tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
```diff
@@ -45,15 +56,23 @@ const selectActivePluginIndex = createSelector(
...
+    select(selectActivePluginIndex)
+  );
+  readonly plugins$ = this.store.pipe(select(selectPlugins));
+  readonly disabledPlugins$ = this.store.pipe(select(selectDisabledPlugins));
 
-  constructor(private store: Store<State>) {}
+  constructor(private readonly store: Store<State>) {}
 
-  onPluginSelectionChanged(index: number) {
+  onPluginSelectionChanged({index}: MatTabChangeEvent) {
```
wchargin: q: What change in the surrounding code forces this to change? It looks
like `onPluginSelectionChanged` is called from the same places in the
same ways before and after this change.

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
```diff
@@ -13,29 +13,40 @@ See the License for the specific language governing permissions and
...
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
+interface UiPluginMetadata extends PluginMetadata {
+  id: PluginId;
+}
+
 const selectPlugins = createSelector(
   getPlugins,
-  (listing) =>
+  (listing): UiPluginMetadata[] =>
```
wchargin: Do I understand correctly that these are strongly typed, even though the
types are inferred, so this is `Selector<BigState, UiPluginMetadata[]>`
or something?

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
```diff
@@ -45,15 +56,23 @@ const selectActivePluginIndex = createSelector(
   styleUrls: ['./header.component.css'],
 })
 export class HeaderComponent {
-  activePluginIndex$ = this.store.pipe(select(selectActivePluginIndex));
-  plugins$ = this.store.pipe(select(selectPlugins));
+  readonly activePlugin$ = this.store.pipe(select(getActivePlugin));
```
wchargin: q: Why do we need to `select` over `getActivePlugin` here? It’s already
a selector, right?

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
```diff
@@ -13,29 +13,40 @@ See the License for the specific language governing permissions and
...
+import {PluginMetadata, PluginId} from '../../types/api';
+
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
+interface UiPluginMetadata extends PluginMetadata {
+  id: PluginId;
+}
+
 const selectPlugins = createSelector(
   getPlugins,
```
wchargin: q: I see that `getPlugins` and `selectPlugins` are different things but
are both “selectors”. Is there some naming convention here, such that if
I were more familiar I could easily say “oh, `getFoo` does X while
`selectFoo` does Y” without having to know the details?

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
stephanwlee: Yup, that's correct. `BigState` -> `PluginMetadata[]` -> `UiPluginMetadata[]`.

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
stephanwlee: Below is my understanding of things:
- Selector: maps State into smaller state (you can create intermediate UI-only state with multiple data)
- `select`: Create a function that returns an observable that invokes the `Selector` underneath.

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
stephanwlee: Ah, that's because I changed the listener from `selectedIndexChange` to `selectedTabChange`. They had functional difference which escapes me for now.

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
stephanwlee: I mis-remembered the code somewhere and took this as a convention. I changed everything to read `get` instead. Thanks for pointing this out!

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
wchargin: Okay, so `createSelector` is just function composition with memoization,
like (simplified)

```js
function createSelector(dep1, dep2, dep3, fn) {
  const memo = new Map();
  return function selector(store) {
    if (!memo.has(store)) {
      const val1 = dep1(store);
      const val2 = dep2(store);
      const val3 = dep3(store);
      const result = fn(val1, val2, val3);
      memo.set(store, result);
    }
    return memo.get(store);
  };
}
```

and `select` creates a function that takes… an `Observable` of `State`
or of `Store<State>`, where `Store<State>` is an `Observable<State>`
itself? and then (maybe after extracting an inner state) passes this
through an observable-aware `map` (`fmap`) to project out the substate
via the selector, resulting in an `Observable<Substate>`?

It’ll take a while for this to solidify—I don’t yet see how the
primitives fit together—and I’m not having the easiest time reading the
RxJS and ngrx sources, but it’d be helpful to know if this is
approximately on the right track.

PR #2618, opened 2019-09-06
data: make `ScalarTimeSeries` and `ScalarDatum` value types

wchargin:
Test Plan:
Unit tests included.

wchargin-branch: data-value-types

### DIFF
--- a/tensorboard/data/provider_test.py
+++ b/tensorboard/data/provider_test.py
@@ -63,6 +63,24 @@ class ScalarTimeSeriesTest(tb_test.TestCase):
     self.assertIn(repr(x.description), repr_)
     self.assertIn(repr(x.display_name), repr_)
 
+  def test_eq(self):
+    x1 = provider.ScalarTimeSeries(77, 1234.5, b"\x12", "one", "two")
+    x2 = provider.ScalarTimeSeries(77, 1234.5, b"\x12", "one", "two")
+    x3 = provider.ScalarTimeSeries(66, 4321.0, b"\x7F", "hmm", "hum")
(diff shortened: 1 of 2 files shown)


tensorboard/data/provider_test.py
```diff
@@ -63,6 +63,24 @@ def test_repr(self):
...
+
+  def test_hash(self):
+    x1 = provider.ScalarTimeSeries(77, 1234.5, b"\x12", "one", "two")
+    x2 = provider.ScalarTimeSeries(77, 1234.5, b"\x12", "one", "two")
+    x3 = provider.ScalarTimeSeries(66, 4321.0, b"\x7F", "hmm", "hum")
+    self.assertEqual(hash(x1), hash(x2))
+    # The next check is technically not required by the `__hash__`
+    # contract, but _should_ pass; failure on this assertion would at
+    # least warrant some scrutiny.
+    self.assertNotEqual(hash(x1), hash(x3))
```
nfelt: In practice I think this is fine, but just to be clear, couldn't this have a false positive in theory?  Since we're basically never guaranteed that any two calls to `hash()` are going to output different things.

wchargin: Yes (hence comment), but it seems more likely that the custom `__hash__`
implementation would be broken to accidentally collapse hashes than that
the built-in tuple `hash` would happen to collide in 64-bit space.

wchargin: I’m not too attached to this, though; if you think that we’d be better
off removing these checks and their friends in `base_plugin_test.py`,
that’s fine with me.

PR #2623, opened 2019-09-09
[debugger] Add bool & string support; Utilize TensorWidget for visualization of 3D+ tensors

caisq:
  * Continue developing TensorWidget and use it in the existing debugger plugin

  * Add support for displaying boolean tensors
  * Add support for displaying string tensors
  * Add demos for boolean and string tensors
    * Note that in demo.ts, we have to work around the limitation that TensorFlow.js currently doesn't support tf.slice() for string tensors.
  * Add dom-module css for tensor-widget
  * Use the css dom-module in the debugger plugin
  * Reroute 3D+ tensor values in the debugger plugin to TensorWidget.

* Screenshots of UI changes
- Screenshot for boolean and string tensors in TensorWidget
image

- Screenshot of TensorWidget in the Debugger Plugin, with both float and string type tensors:
image


tensorboard/components/tensor_widget/slicing-control.ts
```diff
@@ -363,6 +363,14 @@ export class SlicingControl {
       dropdown.style.top = `${top}px`;
       dropdown.style.left = `${left}px`;
       dropdown.style.display = 'block';
+
+      // Check the actual position of the dropdown menu and make sure
```
davidsoergel: Seems hacky-- why is this necessary?  Maybe comment further about why the dropdown is initially in the wrong place?

tensorboard/components/tensor_widget/string-utils.ts
```diff
@@ -82,3 +82,33 @@ export function numericValueToString(
...
 }
+
+/**
+ * Format a boolean value as a human-readiable string.
+ * @param value
+ * @param shortForm Returns full-length string (i.e., 'true' or 'false') if and
+ *   only if this is `true`.
+ * @returns Formatted string.
+ */
+export function booleanValueToString(
```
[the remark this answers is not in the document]
bileschi: +1

tensorboard/components/tensor_widget/slicing-control.ts
caisq: The `top` and `left` styles of the dropdown menu are set relative to those of the dim control. Things all work well in the demo page. But for some reason the dropdown menu is offset by a few hundred pixels along every dimension in the Debugger Plugin. I don't fully understand the cause of this yet. My guess is it has to do with some interaction with the CSS of the Debugger Plugin itself. I added a TODO item to investigate that.

But the approach I'm using here is lightweight and solves the problem for now. It makes sure the dropdown menu is in the right position for both the demo page and the Debugger Plugin.

tensorboard/components/tensor_widget/tensor-widget-impl.ts
```diff
@@ -526,18 +543,28 @@ export class TensorWidgetImpl implements TensorWidget {
...
         valueDiv.addEventListener('mouseenter', () => {
-          const detailedValue = valueDiv.getAttribute(DETAILED_VALUE_ATTR_KEY);
-          if (!detailedValue) {
+          let detailedValue:
+            | number
+            | boolean
+            | string
+            | null = valueDiv.getAttribute(DETAILED_VALUE_ATTR_KEY);
+          const valueClass = this.getValueClass();
+          if (valueClass !== 'string' && !detailedValue) {
```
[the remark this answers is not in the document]
davidsoergel: OK, but now it's even more confusing because of the `JSON.parse()`, which IIUC will convert `"true"` and `"false"` to the boolean values `true` and `false`, thereby actually manifesting the issue I was worried about in the first place.  Similarly a detailedValue of the number `0` will now do the wrong thing, I think.

Aha: you already returned if `!valueAttr`, which I think is all you need anyway.  Can you delete this conditional return entirely?

tensorboard/components/tensor_widget/tensor-widget-impl.ts
```diff
@@ -526,18 +543,35 @@ export class TensorWidgetImpl implements TensorWidget {
...
+          if (!valueAttr) {
             return;
           }
+          let detailedValue: number | boolean | string = JSON.parse(valueAttr);
+          const valueClass = this.getValueClass();
+          if (valueClass !== 'string' && !detailedValue) {
+            return;
+          }
+
+          if (valueClass === 'boolean') {
```
davidsoergel: Pretty sure this one is now wrong too.  If `detailedValue === true`, then `detailedValue !== '1'`, and so it gets converted to `false`.  Can this be deleted, or can you instead use it to check that `detailedValue` is already  boolean?

tensorboard/components/tensor_widget/tensor-widget-impl.ts
```diff
@@ -526,18 +543,35 @@ export class TensorWidgetImpl implements TensorWidget {
...
+          // tensor element. It is detailed in the sense that it has as many
+          // decimal points as supported by JavaScript's string representation
+          // of numbers, in the case of float dtypes. The only exceptions
+          // are very long string elements, which we still truncate in order
+          // to avoid overtaxing the DOM.
+          const valueAttr = valueDiv.getAttribute(DETAILED_VALUE_ATTR_KEY);
+          if (!valueAttr) {
             return;
           }
+          let detailedValue: number | boolean | string = JSON.parse(valueAttr);
```
davidsoergel: Alternate option that would resolve the below comments differently: don't `JSON.parse`, and instead just propagate `detailedValue` as a string only, regardless of the dtype.  It's just for the tooltip anyway, right?

caisq: Adopting the suggestion of just storing the tooltip text in the attribute, without JSON stringify or parse. Thanks.

tensorboard/components/tensor_widget/tensor-widget-impl.ts
caisq: Done.

Issue #2624, opened 2019-09-09
Re-enable mesh test 


stephanwlee:
In the latest TensorFlow, `@tf.function` broke due to missing symbols in gast.

TensorFlow describes its dependency on [gast with `>=0.2.0`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/setup.py#L57) and the new version 0.3.x removed symbols like `Str`, `Num`, etc... This breaks the autograph and below it will even break a simple function like below:
```py
import tensorflow as tf
@tf.function
def foo():
   return 1 + 2

foo()
```

(if you increase the verbosity, it will tell you that it is missing the symbol). 

Currently, the `//tensorboard/plugins/mesh:summary_v2_test` is the only one that uses `@tf.function` in TensorBoard and is impacted. 

We can re-enable the test when (1) tf pins the version of gast or (2) TF autograph uses the correct symbols from gast.

nfelt:
I'm confused by this issue - the mesh summary_v2 tests aren't the only one using @tf.function, several others do as well:

https://github.com/tensorflow/tensorboard/blob/1b623e7c914558b0e521d946fccee2b8ebdb53b9/tensorboard/plugins/scalar/summary_test.py#L192
https://github.com/tensorflow/tensorboard/blob/1b623e7c914558b0e521d946fccee2b8ebdb53b9/tensorboard/plugins/histogram/summary_test.py#L218
https://github.com/tensorflow/tensorboard/blob/1b623e7c914558b0e521d946fccee2b8ebdb53b9/tensorboard/plugins/text/summary_test.py#L209

As far as I know the mesh summary test was basically copied from those, so it's surprising to me that it would break and not those ones.  What exactly was the failure mode?

stephanwlee:
You are correct and autograph in those files indeed fails. Do note that even if autograph fails, it seems to run in some kind of fixture of graph mode / eager mode.

The type of error we are getting is unique to the mesh plugin and it is because of the conditional.
```
    OperatorNotAllowedInGraphError: using a `tf.Tensor` as a Python `bool` is not allowed: AutoGraph did not convert this function. Try decorating it directly with @tf.function.     
```
https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/mesh/summary_v2.py#L125

I tried to debug what kind of code is actually executed under such failure (because pure eager works) but I could not figure it out in an hour that I used to investigate.

wchargin:
[...]

PR #2625, opened 2019-09-09
build: disable mesh summary v2 test

stephanwlee:
This is a temporary disable until #2624 can be resolved.


wchargin:
Perhaps another approach would be to add `pip install gast==0.2.2` to
our CI, but given that the fix has landed and should be in tomorrow’s
nightly this is fine with me.

PR #2626, opened 2019-09-09
core: remove unused `mode` on `/data/environment`

wchargin:
Summary:
The `data_location` field will soon be made pluggable on the data
provider, which will muddy the semantics of the `mode` field. Given that
it’s unused, we can just remove it.

(The semantics were already a bit wonky when both `--logdir` and `--db`
were specified, as `data_location` would point to the logdir but `mode`
would be `db`. Removing it entirely fixes this problem.)

Test Plan:
There are no other uses of the removed `getMode` method:

    $ git grep -ho 'getMode\w*' | sort -u
    getModelClass
    getModelHeaderClass
    getModelName
    getModelName_

The frontend still seems to work fine in both logdir and DB modes.

wchargin-branch: remove-environ-mode

### DIFF
--- a/tensorboard/components/tensor_widget/BUILD
+++ b/tensorboard/components/tensor_widget/BUILD
@@ -31,6 +31,7 @@ tf_ts_library(
     srcs = [
         "dtype-utils.ts",
         "health-pill-types.ts",
+        "selection.ts",
         "shape-utils.ts",
         "slicing-control.ts",
         "string-utils.ts",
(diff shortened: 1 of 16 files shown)


stephanwlee:
Btw, while it is not used, the original (unspoken) intention was to use this for metrics in far far future :)

wchargin:
Interesting! I didn’t know that.

wchargin:
Postponing CI until master is green (wouldn’t pass, anyway).

PR #2631, opened 2019-09-11
errors: add unified error types

wchargin:
Summary:
We’d like to clarify the error semantics in the data provider APIs, but
there’s not currently a good way for errors at the data provider layer
to propagate through the stack: every plugin handler would need to
handle the various exception types appropriately, which would be verbose
and error-prone. To remedy this, we add a family of unified error types
that may be raised from anywhere inside a TensorBoard HTTP handler, and
which will be caught and converted to HTTP responses by TensorBoard
itself. For errors not in a whitelist, we defer to the default “Internal
Server Error” code path, to avoid accidentally leaking sensitive data.

The error types themselves are selected from Google’s standard errors;
we can add more as appropriate.

Test Plan:

```
$ curl -i 'localhost:6006/data/plugin/scalars/scalars?run=foo&tag=bar'
HTTP/1.1 404 NOT FOUND
Content-Length: 50
Expires: 0
Cache-Control: no-cache, must-revalidate
Content-Type: text/plain; charset=utf-8
Server: Werkzeug/0.15.6 Python/3.6.8
Date: Tue, 10 Sep 2019 17:58:30 GMT

Not found: No scalar data for run='foo', tag='bar'
```

wchargin-branch: unified-errors

### DIFF
--- a/tensorboard/__init__.py
+++ b/tensorboard/__init__.py
@@ -19,7 +19,6 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from tensorboard import errors  # public export
 from tensorboard import lazy
 from tensorboard import version
 
@@ -65,6 +64,12 @@ from tensorboard import version
 #
 # See <https://github.com/tensorflow/tensorboard/issues/1989> for
 # additional discussion.
+
+
+@lazy.lazy_load('tensorboard.errors')
+def errors():
+  import importlib  # pylint: disable=g-import-not-at-top
+  return importlib.import_module('tensorboard.errors')
 
 
 @lazy.lazy_load('tensorboard.notebook')
--- a/tensorboard/__init__.py
+++ b/tensorboard/__init__.py
@@ -66,6 +66,12 @@ from tensorboard import version
 # additional discussion.
 
 
+@lazy.lazy_load('tensorboard.errors')
+def errors():
+  import importlib  # pylint: disable=g-import-not-at-top
+  return importlib.import_module('tensorboard.errors')
+
+
 @lazy.lazy_load('tensorboard.notebook')
 def notebook():
   import importlib  # pylint: disable=g-import-not-at-top
--- a/tensorboard/__init__.py
+++ b/tensorboard/__init__.py
@@ -19,6 +19,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+from tensorboard import errors  # public export
 from tensorboard import lazy
 from tensorboard import version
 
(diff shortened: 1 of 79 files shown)


tensorboard/__init__.py
```diff
@@ -66,6 +66,12 @@
 # additional discussion.
 
 
+@lazy.lazy_load('tensorboard.errors')
```
nfelt: Any reason to make this lazy?  It seems fine to just do `from tensorboard import errors` here right, since that has no dependencies?

tensorboard/backend/application.py
```diff
@@ -282,14 +284,14 @@ def __init__(self, plugins, path_prefix=''):
...
-          unordered_prefix_routes[path] = app
+          unordered_prefix_routes[path] = _handling_errors(app)
         else:
           if '*' in path:
             raise ValueError('Plugin %r handles invalid route %r: Only '
                              'trailing wildcards are supported '
                              '(i.e., `/.../*`)' %
                              (plugin.plugin_name, path))
-          self.exact_routes[path] = app
+          self.exact_routes[path] = _handling_errors(app)
```
nfelt: Any particular reason to wrap each route rather than just wrapping `__call__` itself below?  Then it could also be applied in just one place.

wchargin: Good point. I put it here to avoid wrapping _within_ `__call__` on every
request, but you’re absolutely right that we can just wrap `__call__`
itself.

tensorboard/__init__.py
wchargin: Aside from just following the existing pattern, I figured that this was
slightly safer in case we introduced a dependency on `tf.errors` down
the line (even via `tb.compat`), though it’s true that the notf smoke
tests added in this commit would catch that. Can change to a direct
import if you prefer.

nfelt: Yeah I think I'd prefer a direct import, just because the lazy loading is a bit harder to reason about and I'd rather not end up having an entire lazy-loaded API namespace just out of an overabundance of caution.

FWIW, assuming we do a hard import here then introducing a non-lazy TF dependency should be caught by the summary dep test as well:
https://github.com/tensorflow/tensorboard/blob/690f4f3c4d102a53f2c4ed1e1854b735ee0c560f/tensorboard/summary/summary_test.py#L110

tensorboard/backend/application.py
nfelt: Yeah I assume we can just decorate `__call__` itself?  Or we can explicitly wrap the middleware (like above) around the current call implementation, and then cache that and invoke it from `__call__` which amounts to essentially the same thing.

tensorboard/__init__.py
wchargin: Cool, will do.

tensorboard/backend/application.py
nfelt: LGTM

Issue #2614, opened 2019-09-05
UI: run selector toggle all doesn't toggle _on_ on first click


stephanwlee:
When there are [more than 40 runs](https://github.com/tensorflow/tensorboard/blob/master/tensorboard/components/tf_dashboard_common/tf-multi-checkbox.ts#L56), the run selector does not enable runs by default. Unlike my expectation, when no runs are selected, if you click on `Toggle All Runs`, runs do not get enabled. Only after the second click on the button, it toggles all runs.

stephanwlee:
Closed by #2633.

PR #2633, opened 2019-09-11
ui: fix run selector toggle all with many items

psybuzz:
The run selector has a toggle all runs button.  Before, toggleAll()'s intent is to toggle everything off when the user hasn't interacted with the list and the list is short (<40, everything selected by default).  This logic was broken in [1], which renamed "maxRunsToEnableByDefault" to "maxNamesToEnableByDefault", except for one callsite (causing issue #2614).

This PR fixes the issue and simplifies toggleAll() to rely on "this.outSelected", which already represents the computed selected state.

[1] https://github.com/tensorflow/tensorboard/commit/aacd5cb7ab1076679aa4115191aa3f00e4575f0d


psybuzz:
Thanks for the review!  Adding Stephan fyi, and landing!

PR #2639, opened 2019-09-11
Revert "build: disable mesh summary v2 test (#2625)"

### LINKED ISSUE #2624: Re-enable mesh test

stephanwlee:
This reverts commit 69848209eaccbdcf2985aab754d1eb3ec2eafc81.

Fixes #2624.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 dist: trusty
 language: python
 python:
-  - "2.7"
-  - "3.6"
+  - '2.7'
+  - '3.6'
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


wchargin:
@googlebot I consent.

PR #2640, opened 2019-09-11
proto: sync to TensorFlow v1.15.0-rc0

wchargin:
Summary:
These are synced to TensorFlow tag `v1.15.0-rc0`, which resolves to
a0163a0a727c9c7af5ac976debd5e28d42275b8c.

There are no new protos or new `import` dependencies, so the existing
build rules and tests suffice.

Test Plan:
Running `bazel test //tensorboard/compat/proto:proto_test` now passes in
a virtualenv with `tensorflow==1.15.0rc0` installed. This is the same as
current TF 1.x nightlies, which are frozen, but _not_ the same as
current TF 2.x nightlies, due to changes in HEAD.

wchargin-branch: proto-sync-v1.15.0-rc0

### DIFF
--- a/tensorboard/compat/proto/config.proto
+++ b/tensorboard/compat/proto/config.proto
@@ -7,7 +7,7 @@ option java_outer_classname = "ConfigProtos";
 option java_multiple_files = true;
 option java_package = "org.tensorflow.framework";
 
-// add go_package externally with copybara
+option go_package = "github.com/tensorflow/tensorflow/tensorflow/go/core/protobuf";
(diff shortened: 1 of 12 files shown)


wchargin:
This probably needs a gast pin in our CI; I’ll add that.

wchargin:
(Changed base branch: precedent is to merge to master and then
cherry-pick onto the release branch. This is a safe change from a CI
perspective, as `master` and `1.15` are currently identical. Also, this
test does not actually run in CI. :-) )

PR #2644, opened 2019-09-11
TensorBoard 1.15.0

wchargin:
Cherrypicks:

* 73db3480 Revert "text: remove legacy `PluginAsset` support (#2417)"
* 690f4f3c proto: sync to TensorFlow v1.15.0-rc0 (#2640)

Note that the CI change is slightly different from usual (it pins a
transitive dependency), and the revert is not present in master.


wchargin:
I’d like to block release on #2646 if we can get a fix in soon.

PR #2647, opened 2019-09-12
Add counterfactual analysis for regression models (What-If Tool)

grovina:

Add the ability to show nearest counterfactual example in regression models.

The feature and its use are analogous to the existing counterfactual analysis available in non-regression models.

There is one fundamental difference in the way we compare inference classes. In binary classification, the counterfactual examples are the ones belonging to the alternative class. In regression, we will define them as the ones whose inferred values are farther from the original example than a given delta, which will be defined by the user.

For example, if delta is set to 2 and the inferred value of the selected example is 7, then we'll be looking for most similar example with an inferred value outside of the (5, 9) interval.

* Screenshots of UI changes

For regression only:
Screenshot 2019-09-11 at 20 51 07

1. Open binary classification problem and check that nothing's changed. (multi_demo)
1. Open regression classification problem (age_demo) and wait for data to load, then:
   * select an example
   * enable "Show nearest counterfactual datapoint"
   * play around with Delta and distance
   * repeat for some other datapoints

Plotting _Distance to datapoint vs Inference value_ also helps to confirm that the counterfactual found is reasonable.

Some other ways of defining the nearest counterfactual example:
1. Define separate upper and lower deltas for the factual interval.
1. Within a given distance in the feature space, find the datapoint with highest difference in inference value.
1. Finding the datapoint with the highest ratio between difference in inference value and distance in features space.

These alternative ways could be added in the future.

cc @jameswex @tolga-b

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -1197,6 +1197,9 @@ limitations under the License.
       .counterfactual-delta label {
         padding-top: 10px;
       }
+      .counterfactual-delta paper-slider {
+        height: 40px;
+      }
       .datapoint-button {
         color: #202124;
         background: #fde293;
@@ -3818,24 +3821,25 @@ limitations under the License.
         isSameInferenceClass_: function(val1, val2) {
           return this.isRegression_(this.modelType)
             ? Math.abs(val1 - val2) < this.minCounterfactualValueDist
-            : val1 == val2;
+            : val1 === val2;
         },
 
-        adjustMaxCounterfactualValueDist_: function(selected, valueStr) {
+        adjustMaxCounterfactualValueDist_: function(selected, valueName) {
           this.maxCounterfactualValueDist = Math.max(
-            this.distanceStats_[valueStr].max -
-              this.visdata[selected][valueStr],
-            this.visdata[selected][valueStr] - this.distanceStats_[valueStr].min
+            this.distanceStats_[valueName].max -
+              this.visdata[selected][valueName],
+            this.visdata[selected][valueName] -
+              this.distanceStats_[valueName].min
           );
         },
 
         adjustMinCounterfactualValueDist_: function() {
-          const valueStr = this.strWithModelName_(
+          const valueName = this.strWithModelName_(
             inferenceValueStr,
             this.nearestCounterfactualModelIndex
           );
           this.minCounterfactualValueDist = this.distanceStats_[
-            valueStr
+            valueName
           ].stdDev;
         },
 
@@ -3856,7 +3860,7 @@ limitations under the License.
           for (let i = 0; i < this.visdata.length; i++) {
             // Skip the selected example itself and examples with the same inference class.
             if (
-              i == selected ||
+              i === selected ||
               this.isSameInferenceClass_(
                 this.visdata[selected][modelInferenceValueStr],
                 this.visdata[i][modelInferenceValueStr]
@@ -6149,14 +6153,14 @@ limitations under the License.
             const feature = featureStats.name;
             this.distanceStats_[feature] = {};
             if (featureStats.numStats) {
-              // For numeric features, store standard deviation, min and max.
+              // Numeric features:
               this.distanceStats_[feature] = {
                 stdDev: featureStats.numStats.stdDev,
                 min: featureStats.numStats.min,
                 max: featureStats.numStats.max,
               };
             } else {
-              // For categorical features, calculate and store  the probability
+              // Categorical features: calculate and store the probability
               // that any two feature values across all examples are the same.
               let probSameValue = 0;
               const buckets =
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -1187,12 +1187,15 @@ limitations under the License.
       }
       .counterfactual-toggle {
         margin: 4px 4px 4px 6px;
+        padding-top: 4px;
         --paper-toggle-button-checked-bar-color: #81c995;
       }
+      .counterfactual-delta {
+        display: flex;
+        margin-right: 6px;
+      }
       .counterfactual-delta label {
-        font-size: 12px;
-        color: #3c4043;
-        margin-left: 10px;
+        padding-top: 10px;
       }
       .datapoint-button {
         color: #202124;
@@ -1433,12 +1436,12 @@ limitations under the License.
                             title="Minimum distance in inferred value to consider counterfactual"
                             class="counterfactual-delta"
                           >
-                            <label>Delta</label>
                             <paper-slider
                               pin
                               value="{{minCounterfactualValueDist}}"
                               max="{{maxCounterfactualValueDist}}"
                             ></paper-slider>
+                            <label>Delta</label>
                           </div>
                         </template>
                         <paper-dropdown-menu
@@ -1463,7 +1466,7 @@ limitations under the License.
                         </paper-dropdown-menu>
                         <paper-icon-button
                           icon="info-outline"
-                          class="info-icon cf-info-icon no-padding"
+                          class="info-icon no-padding"
                           on-tap="openDialog"
                         >
                         </paper-icon-button>
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -1425,18 +1425,17 @@ limitations under the License.
                         <paper-radio-group
                           selected="{{nearestCounterfactualDist}}"
                         >
-                          <paper-radio-button name="L1"
-                            >L1</paper-radio-button
-                          >
-                          <paper-radio-button name="L2"
-                            >L2</paper-radio-button
-                          >
+                          <paper-radio-button name="L1">L1</paper-radio-button>
+                          <paper-radio-button name="L2">L2</paper-radio-button>
                         </paper-radio-group>
                         <template is="dom-if" if="[[isRegression_(modelType)]]">
-                          <div title="Minimum distance in inferred value to consider counterfactual"
-                            class="counterfactual-delta">
+                          <div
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -1407,70 +1412,89 @@ <h2>Show similarity to selected datapoint</h2>
...
+                            title="Minimum distance in inferred value to consider counterfactual"
+                            class="counterfactual-delta"
                           >
-                            <paper-listbox
-                              class="dropdown-content"
-                              selected="{{nearestCounterfactualModelIndex}}"
+                            <label>Delta</label>
+                            <paper-slider
+                              pin
+                              value="{{minCounterfactualValueDist}}"
```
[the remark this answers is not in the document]
grovina: Sure. I've set it to 1 standard deviation, so that it has some statistical meaning and always falls within the valid range.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -1407,70 +1412,89 @@ <h2>Show similarity to selected datapoint</h2>
...
+                        <template is="dom-if" if="[[isRegression_(modelType)]]">
+                          <div
+                            title="Minimum distance in inferred value to consider counterfactual"
+                            class="counterfactual-delta"
                           >
-                            <paper-listbox
-                              class="dropdown-content"
-                              selected="{{nearestCounterfactualModelIndex}}"
+                            <label>Delta</label>
+                            <paper-slider
```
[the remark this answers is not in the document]
grovina: Screenshot 2019-09-13 at 17 11 32

jameswex:
@stephanwlee please review.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3784,20 +3815,52 @@ <h2>Show similarity to selected datapoint</h2>
...
+        isSameInferenceClass_: function(val1, val2) {
+          return this.isRegression_(this.modelType)
+            ? Math.abs(val1 - val2) < this.minCounterfactualValueDist
+            : val1 == val2;
+        },
+
+        adjustMaxCounterfactualValueDist_: function(selected, valueStr) {
+          this.maxCounterfactualValueDist = Math.max(
+            this.distanceStats_[valueStr].max -
+              this.visdata[selected][valueStr],
```
stephanwlee: Do not know about the nature of `distanceStats` and `visData`: can all of these values be negative? Can `visData` be lower than both max and min of the `distanceStats`? Can this be doing Math.max on, for example, `Math.max(-2, -3)`?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3784,20 +3815,52 @@ <h2>Show similarity to selected datapoint</h2>
...
+            this.visdata[selected][valueStr] - this.distanceStats_[valueStr].min
+          );
+        },
+
+        adjustMinCounterfactualValueDist_: function() {
+          const valueStr = this.strWithModelName_(
```
stephanwlee: `valueStr` -> `modelName` seems more appropriate.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3784,20 +3815,52 @@ <h2>Show similarity to selected datapoint</h2>
...
           let closestDist = Number.POSITIVE_INFINITY;
           let closest = -1;
           for (let i = 0; i < this.visdata.length; i++) {
-            // Skip examples with the same inference class as the selected
-            // examples.
+            // Skip the selected example itself and examples with the same inference class.
             if (
-              this.visdata[selected][modelInferenceValueStr] ==
-              this.visdata[i][modelInferenceValueStr]
+              i == selected ||
```
stephanwlee: nit: prefer to use triple equals.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -6083,9 +6149,12 @@ <h2>Show similarity to selected datapoint</h2>
             const feature = featureStats.name;
             this.distanceStats_[feature] = {};
             if (featureStats.numStats) {
-              // For numeric features, store standard deviation.
-              this.distanceStats_[feature].stdDev =
-                featureStats.numStats.stdDev;
+              // For numeric features, store standard deviation, min and max.
```
stephanwlee: Not sure if this comment is adding any information that is not present in the code below. Consider removing it.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3784,20 +3815,52 @@ <h2>Show similarity to selected datapoint</h2>
...
+            this.visdata[selected][valueStr] - this.distanceStats_[valueStr].min
+          );
+        },
+
+        adjustMinCounterfactualValueDist_: function() {
+          const valueStr = this.strWithModelName_(
+            inferenceValueStr,
+            this.nearestCounterfactualModelIndex
+          );
+          this.minCounterfactualValueDist = this.distanceStats_[
```
stephanwlee: Can this be a computed property of `modelType`, `distanceStats_` and `nearestCounterfactualModelIndex`? or does it need to only update when we are looking for the closest counter factual?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3784,20 +3815,52 @@ <h2>Show similarity to selected datapoint</h2>
...
+          this.maxCounterfactualValueDist = Math.max(
+            this.distanceStats_[valueStr].max -
+              this.visdata[selected][valueStr],
+            this.visdata[selected][valueStr] - this.distanceStats_[valueStr].min
+          );
+        },
+
+        adjustMinCounterfactualValueDist_: function() {
+          const valueStr = this.strWithModelName_(
+            inferenceValueStr,
```
stephanwlee: Where does this symbol `inferenceValueStr` come from? 

Ah, it is a global variable. Do action required but all module level constant should be ALL_CAPS :(

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
grovina: `minCounterfactualValueDist` is actually controlled by the user. Here we're just defining its initial value.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
grovina: I dropped the redundant part, but do think it's important to say that this block handles numeric features, as opposed to `For categorical features...` in the following block.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
grovina: It's the name of the key holding the value, which is basically the string `' Inference value'` suffixed with the model name... I've changed it to `valueName` to make it clearer.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
grovina: Should I send a separate PR converting them to ALL_CAPS?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
grovina: `distanceStats_` is based on the `visData` data, so that:
`distanceStats_[valueStr].min` <= `visdata[selected][valueStr]` <= `distanceStats_[valueStr].max`

This `Math.max` op then receives only non-negative values.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
jameswex: we can do a separate PR changing all our model level constants to CAPS. no need in this review.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -1407,70 +1418,91 @@ <h2>Show similarity to selected datapoint</h2>
...
+                            title="Minimum distance in inferred value to consider counterfactual"
+                            class="counterfactual-delta"
                           >
-                            <paper-listbox
-                              class="dropdown-content"
-                              selected="{{nearestCounterfactualModelIndex}}"
+                            <paper-slider
+                              pin
+                              value="{{minCounterfactualValueDist}}"
+                              max="{{maxCounterfactualValueDist}}"
```
stephanwlee: Do you require the two way binding for the max value? That seems odd.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
stephanwlee: Ok, and I presume the initial value cannot be set in the property declaration?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
grovina: No need indeed. Fixed.

PR #2649, opened 2019-09-12
Initialize graph_v2 plugin with Angular boilerplate.

davidsoergel:
Here we begin development of graph_v2, a substantial rewrite of the Graph Plugin.

This first PR is basically the same as #2519.  It contains:
* Standard TensorBoard Dynamic Plugin structure, e.g. plugin.py.
* An Angular 'hello world' app generated via `ng new frontend` (everything under `.../frontend`; the bulk of the PR).
* A few small tweaks needed to make the Angular app work as a TB plugin (see https://github.com/tensorflow/tensorboard/commit/16dab62f3b0d3a0896bec39dc3c0c97013bc201d from the other PR).
* Build and dev scripts, with a mechanism for 'watch', i.e. hot rebuilding the plugin frontend without restarting TensorBoard.

Here I addressed the outstanding comments from #2519, with the exception of:
* Figuring out the licensing story for the generated Angular boilerplate (still researching)
* Running Prettier.  This is coming in a following PR, but I thought it good recordkeeping to start from the generated app as is.

WANT_LGTM=stephanwlee, wchargin
optional/FYI=caisq, psybuzz, lakovar


davidsoergel:
Legal confirms that the Angular license does not apply to code that Angular generates via `ng new my-app`, so we can just treat that like our original code, covered here by the Apache 2 license.  That license is already present at the root of the TensorBoard tree.

I should still add the Apache 2 license header to each source file, but I'd like to do that in a followup PR, together with the Prettier pass.

davidsoergel:
I certainly plan to get the Bazel build working, but I figured that's separable to another PR, just in the interests of keeping each PR individually manageable.  Do you think the Bazel build needs to be part of this PR?

Even once Bazel works, I want to retain the npm build option for the sake of `npm run watch`--at least until/unless I'm convinced that iBazel can do the same thing reliably. That also seems separable.

stephanwlee:
For one, Bazel rules nodejs prefers to use yarn instead of npm. I wish not to keep around multiple package manager within a same codebase. Also, when working on the Bazel integation, you will find yourself not doing the way you are doing. For instance, you would not commit the .gitignore changes. Please consider doing it in this PR.

davidsoergel:
OK, it looks like the Bazel version will be different enough that I'll likely abandon this PR and start a new one for cleanliness.  Please neglect this one, at least for now.

PR #2654, opened 2019-09-13
Travis CI: Do not hardcore trusty because it is EOL

cclauss:
* Motivation for features / changes
Travis CI has switch to [Xenial as the default distro](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment) because [Trusty reached the end of its five-year LTS window in April 2019](https://ubuntu.com/blog/ubuntu-14-04-trusty-tahr).

* Technical description of changes

Remove the hardcoding of the Trusty distro.

* Screenshots of UI changes

* Detailed steps to verify changes work correctly (as executed by you)

Execute test steps locally and on a branch in Travis CI.

* Alternate designs / implementations considered

Also see #2292

PR #2655, opened 2019-09-13
ci: upgrade flake8 and yamllint

cclauss:
Like #2654 but without the distro change from Trusty to Xenial in order to make it easier to review these improvements separately from the complexities of changing the distro.

Both flake8 and yamllint have progressively added new tests and some new flake8 tests have been found to have a positive effect on runtime safety.

1. Upgrade flake8 and yamllint to current versions.
2. Add more flake8 tests
3. Fix two instances of __use ==/!= to compare str, bytes, and int literals__ because identity is not the same thing as equality in Python. These instances will raise __SyntaxWarnings__ on Python 3.8 so best to fix them now. https://docs.python.org/3.8/whatsnew/3.8.html#porting-to-python-3-8

* Screenshots of UI changes

Execute test steps locally and on a branch in Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,8 +50,8 @@ install:
   # to versions other than the most recent(!).
   - pip install -U pip
   # Lint check deps.
-  - pip install flake8==3.5.0
-  - pip install yamllint==1.5.0
+  - pip install flake8==3.7.8
+  - pip install yamllint==1.17.0
   # TensorBoard deps.
   - pip install futures==3.1.1
   - pip install grpcio==1.6.3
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -79,7 +79,7 @@ before_script:
   - elapsed "before_script"
   # Do a fail-fast check for Python syntax errors or undefined names.
   # Use the comment '# noqa: <error code>' to suppress.
-  - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+  - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
wchargin: Also, a link to <http://flake8.pycqa.org/en/3.7.8/user/error-codes.html>
in a comment would be much appreciated here.

wchargin:
Thanks again! :-)

PR #2657, opened 2019-09-13
build: make tensor_widget build with JSComp

stephanwlee:
The name, tensor_widget_binary, is unique to rollup_bundle; could not
figure out how to disable such global export in the bundler yet. As
such, we need to explicitly export a module to a global variable.

The externs file was introduced so the exported names are not mangle and
can remain stable.

Internal POC can be found here: cl/268757810

### DIFF
--- a/tensorboard/components/tensor_widget/externs.js
+++ b/tensorboard/components/tensor_widget/externs.js
@@ -1,0 +1,26 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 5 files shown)


tensorboard/components/tensor_widget/externs.js
```diff
@@ -0,0 +1,26 @@
...
+ * @externs Externs for TensorWidget for interoperability.
+ */
+
+/** @type {!Object} */
+var tensor_widget = {};
+
+/** @type {!Function} */
+tensor_widget.tensorWidget;
+
+/** @type {!number} */
```
caisq: I believe this is a string (e.g., "0.1.2")

tensorboard/components/tensor_widget/tensor-widget-interop.ts
```diff
@@ -0,0 +1,4 @@
+import * as tensorWidget from './tensor-widget';
```
caisq: Add License header?

PR #2663, opened 2019-09-17
What-If Tool Attribution display redesign

jameswex:

Updates to What-If Tool to work towards matching new attribution mode mocks.

Moved attribution min/max calculation and legend creation into WIT component, from example viewer
Display attribution legend in example viewer header in WIT component when attributions available
Added sort dropdown when attributions available to sort features by attributions or default (alphabetical).
Add attribution value tooltip (through title setting) to value pills when attributions available
Use white text in value pills when attribution value causes background to be very dark
Improve setting of ticks in the attribution legend

* Screenshots of UI changes

attributions

Run normal WIT with no attributions and see no visual changes.
Run WIT with attributions (all positive, all negative, and mix of both) and see correct display of attributions

### DIFF
--- a/tensorboard/components/vz_example_viewer/vz-example-viewer.ts
+++ b/tensorboard/components/vz_example_viewer/vz-example-viewer.ts
@@ -174,7 +174,7 @@ namespace vz_example_viewer {
         observer: 'updateCompareMode',
       },
       // The order to sort features. Can be 'attribution',
-      // 'revserse-attribution', or 'alphabetical'.
+      // 'reverse-attribution', or 'alphabetical'.
       sortOrder: {
         type: String,
         value: 'attribution',
--- a/tensorboard/components/vz_example_viewer/vz-example-viewer.ts
+++ b/tensorboard/components/vz_example_viewer/vz-example-viewer.ts
@@ -173,6 +173,8 @@ namespace vz_example_viewer {
         computed: 'getSeqFeatures(compareExample)',
         observer: 'updateCompareMode',
       },
+      // The order to sort features. Can be 'attribution',
+      // 'revserse-attribution', or 'alphabetical'.
       sortOrder: {
         type: String,
         value: 'attribution',
@@ -427,6 +429,8 @@ namespace vz_example_viewer {
       requestAnimationFrame(() => this._haveSaliencyImpl());
     },
 
+    // Determines if text should be light or dark due to the saliency-generated
+    // background of the text box.
     _useLightColor(saliency) {
       const percentile = (saliency - this.minSal) / (this.maxSal - this.minSal);
       if (this.minSal < 0 && this.maxSal > 0) {
@@ -462,6 +466,8 @@ namespace vz_example_viewer {
         if (!val) {
           continue;
         }
+        // Set background color, tooltip, and text color, which are all based
+        // on saliency score.
         const colorFn = Array.isArray(val)
           ? (d: {}, i: number) => this.getColorForSaliency(val[i])
           : () => this.getColorForSaliency(val);
--- a/tensorboard/components/vz_example_viewer/vz-example-viewer.ts
+++ b/tensorboard/components/vz_example_viewer/vz-example-viewer.ts
@@ -181,6 +181,10 @@ namespace vz_example_viewer {
         computed: 'getSeqFeatures(compareExample)',
         observer: 'updateCompareMode',
       },
+      sortOrder: {
+        type: String,
+        value: 'attribution'
+      },
       compareMode: Boolean,
       compareImageInfo: {type: Object, value: {}},
       compareTitle: String,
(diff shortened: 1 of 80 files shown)


jameswex:
@tolga-b please review

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -6227,20 +6296,160 @@ <h2>Show similarity to selected datapoint</h2>
...
           this.setAttributions();
+          this.createAttributionLegend();
           this.refreshDive_();
         },
 
+        // Create and display the attribution legend.
+        createAttributionLegend: function() {
+          this.attributionColorScale = d3
+            .scaleLinear()
+            .domain([this.minAttribution, 0, this.maxAttribution])
```
tolga-b: does this create an issue when minAttribution is positive or maxAttribution is positive?

tensorboard/components/vz_example_viewer/vz-example-viewer.ts
```diff
@@ -381,7 +387,7 @@ namespace vz_example_viewer {
             } else {
               const diff = saliencyTotals[b.name] - saliencyTotals[a.name];
               if (diff != 0) {
```
tolga-b: can diff be larger than 1, causing images to be below some features?
what happens when diff == 0 the function doesn't seem to return anything.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -1593,6 +1599,34 @@ <h2>Show similarity to selected datapoint</h2>
...
+                                >
+                                <paper-item name="attribution"
+                                  >Attributions</paper-item
+                                >
+                                <paper-item name="reverse-attribution"
+                                  >Reverse attributions</paper-item
+                                >
+                              </paper-listbox>
+                            </paper-dropdown-menu>
+                            <svg id="attributionlegend"></svg>
```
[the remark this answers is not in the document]
jameswex: its just an ID to lookup in the code, so not necessary. But can make it capital for clarity.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -5889,8 +5946,19 @@ <h2>Show similarity to selected datapoint</h2>
...
+            attributions != null &&
+            Object.keys(attributions.attributions).length != 0 &&
+            attributions.attributions[0] != null
+          );
+        },
+
         newAttributions_: function(attributions) {
-          if (Object.keys(attributions).length == 0) {
+          if (
+            Object.keys(attributions.attributions).length == 0 &&
```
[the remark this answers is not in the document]
jameswex: done

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -5918,20 +5987,160 @@ <h2>Show similarity to selected datapoint</h2>
                 const attribSparseKeyKey =
                   attributionPrefix + keys[j] + attributionSparseKeySuffix;
                 datapoint[attribSparseKeyKey] = attribs[keys[j]][0];
-                const attribSparseValueKey =
+                attribKey =
```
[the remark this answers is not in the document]
jameswex: The value key is used down below now to find the global max and min attributions, so in both the case of normal or sparse attributions, we need to be able to key into the attribution value for bounds checking. Since the sparse key isn't needed for that part, didn't need to change its name

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
jameswex: good point. fixed

tensorboard/components/vz_example_viewer/vz-example-viewer.ts
```diff
@@ -181,15 +173,21 @@ namespace vz_example_viewer {
         computed: 'getSeqFeatures(compareExample)',
         observer: 'updateCompareMode',
       },
+      // The order to sort features. Can be 'attribution',
+      // 'revserse-attribution', or 'alphabetical'.
```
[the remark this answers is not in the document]
jameswex: done

tensorboard/components/vz_example_viewer/vz-example-viewer.ts
jameswex: sort function just needs to return pos value for two items that are equal position, negative value if it should be before the second item, and positive value if it should be after the second item. the magnitude of the number doesn't actually matter, just the sign. so this should do the right thing and keep images at top and beyond that, put features in sort order correctly.

tolga-b: nevermind, it actually returns string sort if diff==0 :)

PR #2664, opened 2019-09-18
core: split magic `--logdir_spec` out from verbatim `--logdir`

### LINKED ISSUE #1220: --logdir unexpectedly interprets commas and colons
Consider Stack Overflow for getting support using TensorBoard - they have a larger community with better searchability:

https://stackoverflow.com/questions/tagged/tensorboard

For bug reports, please include the following:

 - TensorBoard version (from pip package, also printed out when running `tensorboard`) 1.8.0
 - TensorFlow version if different from TensorBoard
 - OS Platform and version (e.g., Linux Ubuntu 16.04) Win10 64bit
 - Python version (e.g. 2.7, 3.5) 3.6.4
 - For browser-related issues:
   - Browser type and version (e.g. Chrome 64.0.3282.140)
   - Screenshot if it's a visual issue
 
Please describe the bug as clearly as possible, and if possible provide a minimal example (code, data, and/or command line) to reproduce the issue.  Thanks!
How can I pass a logdir with comma in it? It's not work even I put the path in the quotes.

### LINKED ISSUE #2615: Remove special treatment for commas and colons in logdir specifiers
Generally, the argument to `--logdir` is interpreted simply as a path.
But two characters have special meanings: commas allow specifying
multiple directories, and colons allow specifying custom names for runs.
Together, then:

    tensorboard --logdir model1:/my/logs/model1,model2:/my/logs/model2

This is problematic when people try to specify a log directory that
actually contains commas or colons in its name, which is fairly common
for specifying hyper parameters: e.g., `./logs/lr_1E-03,conv=2,fc=2`. It
also complicates the parsing logic on Windows: `--logdir C:\logs` needs
to be a bare name, not a run named `C` (while `--logdir foo:C:\logs`
_does_ need to be named).

It’s fairly common to get issues from users who can’t understand why
TensorBoard isn’t loading the data that they’re telling it to (e.g.,
#1220), so this is a real usability issue.

Furthermore, the behavior is mostly superfluous. TensorBoard already
loads runs recursively under the logdir, so it often suffices to just
point TensorBoard at a common ancestor directory (`/my/logs` in the
above example). When one wishes to include _some_ runs from a very large
parent directory but not all, one can simply create a symlink tree,
which can be packaged up into [a convenient shell function][1].

The only genuine value-add that this provides, near as I can tell, is
when one wishes to load a proper subset of runs in a directory _on a
remote filesystem_ that cannot also be locally mounted, because in that
case a symlink tree will not suffice.

It would be great to remove this in TF 2.0.

[1]: https://github.com/tensorflow/tensorboard/issues/179#issuecomment-518729885

wchargin:
Summary:
This introduces a new flag, `--logdir_spec`, whose behavior is the same
as the old behavior of `--logdir`. The `--logdir` flag now specifies a
single log directory, without special treatment for commas and colons,
though it still has the `expanduser` call for compatibility with users
who use the `--logdir=~/...` form rather than `--logdir ~/...`.

I audited all uses of `flags.logdir` and `context.logdir`, and, other
than those changed in this commit, all really should be using `logdir`
rather than `logdir_spec` (i.e., they do not properly parse the logdir
spec, and so were broken before this commit).

Fixes #2615, fixes #1220, and fixes lots of user confusion.

Test Plan:
Run with the following args:

    --logdir ~/data/  # should work as usual
    --logdir=~/data/  # should work as usual
    --logdir ~/data/scalars_demo,~/data/audio_demo  # should be empty
    --logdir ~/data/mnist/lr_1E-03,conv=1,fc=2  # should work now!
    --logdir_spec ~/data/  # should work, but without projector, et al.
    --logdir_spec ~/data/scalars_demo,~/data/audio_demo  # should work

wchargin-branch: logdir-spec

### DIFF
--- a/tensorboard/plugins/base_plugin.py
+++ b/tensorboard/plugins/base_plugin.py
@@ -221,6 +221,7 @@ class TBContext(object):
       db_uri=None,
       flags=None,
       logdir=None,
+      logdir_spec=None,
       multiplexer=None,
       plugin_name_to_instance=None,
       window_title=None):
@@ -245,12 +246,12 @@ class TBContext(object):
           function is cheap. The returned connection must only be used by a
           single thread. Things like connection pooling are considered
           implementation details of the provider.
-      db_uri: The string db URI TensorBoard was started with. If this is set,
-          the logdir should be None.
+      db_uri: The string db URI TensorBoard was started with.
       flags: An object of the runtime flags provided to TensorBoard to their
           values.
-      logdir: The string logging directory TensorBoard was started with. If this
-          is set, the db_uri should be None.
+      logdir: The string logging directory TensorBoard was started with.
+      logdir_spec: The value of `--logdir_spec` passed to TensorBoard (see
+          flag-level documentation).
       multiplexer: An EventMultiplexer with underlying TB data. Plugins should
           copy this data over to the database when the db fields are set.
       plugin_name_to_instance: A mapping between plugin name to instance.
@@ -267,6 +268,7 @@ class TBContext(object):
     self.db_uri = db_uri
     self.flags = flags
     self.logdir = logdir
+    self.logdir_spec = logdir_spec
     self.multiplexer = multiplexer
     self.plugin_name_to_instance = plugin_name_to_instance
     self.window_title = window_title
--- a/tensorboard/plugins/base_plugin.py
+++ b/tensorboard/plugins/base_plugin.py
@@ -221,7 +221,6 @@ class TBContext(object):
       db_uri=None,
       flags=None,
       logdir=None,
-      logdir_spec=None,
       multiplexer=None,
       plugin_name_to_instance=None,
       window_title=None):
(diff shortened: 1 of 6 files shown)


tensorboard/plugins/base_plugin.py
```diff
@@ -245,12 +246,12 @@ def __init__(
...
           implementation details of the provider.
-      db_uri: The string db URI TensorBoard was started with. If this is set,
-          the logdir should be None.
+      db_uri: The string db URI TensorBoard was started with.
       flags: An object of the runtime flags provided to TensorBoard to their
           values.
-      logdir: The string logging directory TensorBoard was started with. If this
-          is set, the db_uri should be None.
+      logdir: The string logging directory TensorBoard was started with.
+      logdir_spec: The value of `--logdir_spec` passed to TensorBoard (see
```
nfelt: Let's just have the data location handler consult `context.flags.logdir_spec` directly, even though it's uglier, to avoid presenting logdir_spec as something that plugin implementers can depend on.  (If we did want them to use it we'd probably want to pass it to them in parsed form, not the colon and comma delimited form.)

Issue #2471, opened 2019-07-26
Preparing embeddings for projector with tensorflow2


asitplus-pteufl:
Preparing embeddings for projector with tensorflow2.

tensorflow1 code would look something like that:

    embeddings = tf.compat.v1.Variable(latent_data, name='embeddings')
    CHECKPOINT_FILE = TENSORBOARD_DIR + '/model.ckpt'
    # Write summaries for tensorboard
    with tf.compat.v1.Session() as sess:
        saver = tf.compat.v1.train.Saver([embeddings])
        sess.run(embeddings.initializer)
        saver.save(sess, CHECKPOINT_FILE)
        config = projector.ProjectorConfig()
        embedding = config.embeddings.add()
        embedding.tensor_name = embeddings.name
        embedding.metadata_path = TENSORBOARD_METADATA_FILE

    projector.visualize_embeddings(tf.summary.FileWriter(TENSORBOARD_DIR), config)

when using eager mode in tensorflow2 this should (?) look somehow like this:

    embeddings = tf.Variable(latent_data, name='embeddings')
    CHECKPOINT_FILE = TENSORBOARD_DIR + '/model.ckpt'
    ckpt = tf.train.Checkpoint(embeddings=embeddings)
    ckpt.save(CHECKPOINT_FILE)

    config = projector.ProjectorConfig()
    embedding = config.embeddings.add()
    embedding.tensor_name = embeddings.name
    embedding.metadata_path = TENSORBOARD_METADATA_FILE

    writer = tf.summary.create_file_writer(TENSORBOARD_DIR)
    projector.visualize_embeddings(writer, config)

however, there are 2 issues:

- the `writer` created with `tf.summary.create_file_writer` does not have the function `get_logdir()` required by `projector.visualize_embeddings`, a simple workaround is to patch the `visualize_embeddings` function to take the logdir as parameter.
- the checkpoint format has changed, when reading the checkpoint with load_checkpoint (which seems to be the tensorboard way of loading the file), the variable names change. e.g. `embeddings` changes to something like `embeddings/.ATTRIBUTES/VARIABLE_VALUE` (also there are additional variables in the map extracted by `get_variable_to_shape_map()`but they are empty anyways).

the second issue was solved with the following quick-and-dirty workaround (and logdir is now a parameter of `visualize_embeddings()`)

    embeddings = tf.Variable(latent_data, name='embeddings')
[...]

PR #2665, opened 2019-09-18
projector: visualize_embeddings API change

stephanwlee:
Previously, it used to take an FileWriter as an input so it can infer
logdir to write embedding to. Because summary v2 API deprecated an
ability to get the logdir, we now instead take the logdir as string.

To work with the older code, we make sure we can also take the v1
summary writer.

Addresses #2471, aside from supporting the new checkpoint naming

### DIFF
--- a/tensorboard/plugins/projector/projector_api_test.py
+++ b/tensorboard/plugins/projector/projector_api_test.py
@@ -21,6 +21,7 @@ from __future__ import print_function
 import os
 import shutil
 
+import six
 import tensorflow as tf
 
 from google.protobuf import text_format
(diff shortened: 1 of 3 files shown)


tensorboard/plugins/projector/projector_api_test.py
```diff
@@ -21,37 +21,66 @@
...
+    logdir = self.get_temp_dir()
+    self.addCleanup(shutil.rmtree, logdir)
+
+    config = create_dummy_config()
+    projector.visualize_embeddings(logdir, config)
+
+    # Read the configurations from disk and make sure it matches the original.
+    with tf.io.gfile.GFile(os.path.join(logdir, 'projector_config.pbtxt')) as f:
+      config2 = projector.ProjectorConfig()
+      # config2.ParseFromString(f.read())
```
wchargin: Delete?

tensorboard/plugins/projector/projector_api_test.py
```diff
@@ -21,37 +21,66 @@
...
-  def testVisualizeEmbeddings(self):
-    # Create a dummy configuration.
-    config = projector.ProjectorConfig()
-    config.model_checkpoint_path = 'test'
-    emb1 = config.embeddings.add()
-    emb1.tensor_name = 'tensor1'
-    emb1.metadata_path = 'metadata1'
+  def test_visualize_embeddings_with_logdir(self):
+    logdir = self.get_temp_dir()
+    self.addCleanup(shutil.rmtree, logdir)
```
wchargin: No need: `get_temp_dir()` is cleaned up automatically, and every `rm -r`
that we can remove from the codebase makes me sleep a little easier. :-)

tensorboard/plugins/projector/projector_api_test.py
```diff
@@ -21,37 +21,66 @@
...
 from google.protobuf import text_format
 
 from tensorboard.plugins import projector
 from tensorboard.util import test_util
 
-tf.compat.v1.disable_v2_behavior()
-
+def create_dummy_config():
+  return projector.ProjectorConfig(
+      model_checkpoint_path = 'test',
```
wchargin: nit: No spaces around `kwarg=value`s in Python.

tensorboard/plugins/projector/projector_api_test.py
stephanwlee: Did not know! Thanks!

tensorboard/plugins/projector/projector_api_test.py
stephanwlee: I got lazy and expected formatter to kick in :( I am spoiled!

PR #2666, opened 2019-09-18
cleanup: set style using CSS

stephanwlee:
Setting the style manually generally is not recommended as it can
contain JavaScript invoking CSS.

### DIFF
--- a/tensorboard/components/tf_line_chart_data_loader/tf-line-chart-data-loader.html
+++ b/tensorboard/components/tf_line_chart_data_loader/tf-line-chart-data-loader.html
@@ -31,6 +31,7 @@ limitations under the License.
     <div id="chart-and-spinner-container">
       <vz-line-chart2
         id="chart"
+        data-loading$="[[dataLoading]]"
         color-scale="[[colorScale]]"
         default-x-range="[[defaultXRange]]"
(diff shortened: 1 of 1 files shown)


stephanwlee:
> Is this attack vector relevant in modern browsers?

Not that I know of but polymer-resin (https://github.com/Polymer/polymer-resin) hates it. As usual, I won't say I understand why/how it decides to hate it but I was able to witness the hatred thru an experiment. It reads something like  
```
Failed to sanitize attribute of <vz-line-chart2>: <vz-line-chart2 style="opacity: 0.3;">
```

wchargin:
Cool (I inferred as much from the branch name :-) ). If you do find a
proof-of-concept snippet for this vector, I’d love to see it.

psybuzz:
Nice!  I have a style question: do we prefer "data-" attributes instead of classnames?

stephanwlee:
> do we prefer "data-" attributes instead of classnames

There is no policy but reflection to attribute is too darn easy for boolean ones. It is a lot more cumbersome to do something like https://www.npmjs.com/package/classnames. Also, not sure if this is relevant, but especially because styles are scoped per shadow DOM, such attributes do not tend to bite you in bad ways.

PR #2667, opened 2019-09-18
infra: upgrade threejs to r108

stephanwlee:
This updates version of threejs from r104 to r108.


psybuzz:
I checked the THREE.js migration guide, and none of the breaking changes seem to affect us.
LGTM

PR #2669, opened 2019-09-18
Fix PD plots in py3

jameswex:
PD plots didn't work at all in notebook mode with py3 kernels.

Ensure string feature values used in PD plot generation are sent to JS as strings for json serialization, not as bytes, to avoid failing serialization.
Correctly use the bytes when running in py3 as that is what tf.Example expects for bytes_list, whereas in py2 it functions on strings.
Make use of six.bytes_type and encode and decode functions to correctly handle this.

Ran notebooks in py2 and py3 with this change and verified that PD plots are shown correctly (with correct values as well) for both numeric and categorical features.

### DIFF
--- a/tensorboard/plugins/interactive_inference/utils/inference_utils.py
+++ b/tensorboard/plugins/interactive_inference/utils/inference_utils.py
@@ -125,7 +125,8 @@ class OriginalFeatureList(object):
   def __init__(self, feature_name, original_value, feature_type):
     """Inits OriginalFeatureList."""
     self.feature_name = feature_name
-    self.original_value = [ensure_not_binary(v) for v in original_value]
+    self.original_value = [
+      ensure_not_binary(value) for value in original_value]
     self.feature_type = feature_type
 
     # Derived attributes.
--- a/tensorboard/plugins/interactive_inference/utils/inference_utils.py
+++ b/tensorboard/plugins/interactive_inference/utils/inference_utils.py
@@ -26,7 +26,6 @@ import numpy as np
 import tensorflow as tf
 from google.protobuf import json_format
 from six import binary_type, string_types, integer_types
-from six import ensure_binary, ensure_str
 from six import iteritems
 from six.moves import zip  # pylint: disable=redefined-builtin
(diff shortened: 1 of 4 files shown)


jameswex:
@tolga-b please review

tensorboard/plugins/interactive_inference/utils/inference_utils.py
```diff
@@ -125,7 +125,9 @@ class OriginalFeatureList(object):
   def __init__(self, feature_name, original_value, feature_type):
     """Inits OriginalFeatureList."""
     self.feature_name = feature_name
-    self.original_value = original_value
+    self.original_value = [
```
tolga-b: I think it's better to use ensure_text from six for consistency with https://github.com/tensorflow/tensorboard/pull/2311.

jameswex: actually can't as tensorboard/tensorflow depend on six 1.10.0 which doesn't include the ensure_* functions. our separate witwidget code seems to pull in six 1.12.0 due to transitive dependencies that is why #2311 works for us. so i reverted back to the first implementation and added an explicit dependency in witwidget setup.py on six 1.12.0 to improve the pip package setup dependencies.

tolga-b: I see, sounds good then.

tensorboard/plugins/interactive_inference/utils/inference_utils.py
```diff
@@ -125,7 +125,7 @@ class OriginalFeatureList(object):
   def __init__(self, feature_name, original_value, feature_type):
     """Inits OriginalFeatureList."""
     self.feature_name = feature_name
-    self.original_value = original_value
+    self.original_value = [ensure_not_binary(v) for v in original_value]
```
stephanwlee: v -> human readable name please :)

jameswex: done, thx

PR #2678, opened 2019-09-19
Aggregate numeric values when slicing (What-If Tool)

grovina:

When slicing by a numeric feature, aggregate values in intervals instead of creating a slice for each different value.

The current implementation creates one slice for each different value of the feature. This might result in a huge number of slices that become unmanageable (imagine a slicing by features like "age" or "salary", which can have hundreds of unique values).
To work around this issue, we aggregate the values into buckets (like a histogram), then show aggregated statistics and performance results for each bucket.
Nothing changes for categorical features.

* Screenshots of UI changes

Screenshot from 2019-09-18 17-54-25

For both _multi_ and _age_ demos, wait for data to load and then:
1. Open the Performance tab
1. Slice by different features and check that:
   * Categorical features behave just like before (compare with demos in prod at https://pair-code.github.io/what-if-tool)
   * Numeric features with up to 10 unique values also remain the same
   * Numeric features with more than 10 unique values are aggregated in buckets, and results are coherent with the correct totals.

I opted for the equally-sized intervals (as opposed to the equally-populated ones, for example), mainly because they should be simpler to understand and because one of comparative columns is precisely the number of examples in each slice. The number of slices is controlled by a variable that could become a setting if needed.
As for notations, I also considered other a simple `a — b`, but the mathematical intervals are particularly precise and seem to be accessible enough.
Finally, generating the bucket edges beforehand was important to make the values look ok, since we would oftentimes get things like _[17.43333333333, 20]_ printed otherwise. I tried to simply calculate them ad hoc, but having consistent and simple display turned out to be more complex than expected.

cc @jameswex and @tolga-b

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -3263,6 +3263,11 @@ limitations under the License.
             type: String,
           },
           maxInferenceEntriesPerRun: Number,
+          // Number of buckets When aggregating numeric features.
+          numBuckets: {
+            type: Number,
+            value: 10,
+          },
 
           // Inferences from servo.
           inferences: {
@@ -6097,22 +6102,25 @@ limitations under the License.
          * Whether values of this feature should be aggregated in bucket slices.
          */
         shouldBeAggregated_: function(feature) {
-          const maxUniqueValueSlices = 10; // TODO: Move this to a variable.
           return (
             this.distanceStats_[feature] &&
             this.distanceStats_[feature].isNumeric &&
-            this.stats[feature].uniqueCount > maxUniqueValueSlices
+            // No point in aggregating if less unique values than buckets.
+            this.stats[feature].uniqueCount > this.numBuckets
           );
         },
 
+        /**
+         * Calculate edges between buckets for aggregating numeric features.
+         * We do this beforehand to round numbers and avoid ugly interval labels.
+         */
         calculateFeatureBucketEdges_: function() {
-          const numBuckets = 10; // TODO: Move this to a variable.
           this.featureBucketEdges_ = {};
           for (const feature of Object.keys(this.distanceStats_)) {
             if (this.shouldBeAggregated_(feature)) {
               const min = this.stats[feature].numberMin;
               const max = this.stats[feature].numberMax;
-              const len = (max - min) / numBuckets;
+              const len = (max - min) / this.numBuckets;
               const stdDev = this.distanceStats_[feature].stdDev;
               function round(val) {
                 // Round to a precision near the magnitude of standard deviation.
@@ -6121,7 +6129,7 @@ limitations under the License.
               }
               const bucketEdges = [];
               bucketEdges.push(min);
-              for (let i = 1; i < numBuckets; i++) {
+              for (let i = 1; i < this.numBuckets; i++) {
                 bucketEdges.push(round(min + i * len));
               }
               bucketEdges.push(max);
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -3213,6 +3213,10 @@ limitations under the License.
       }
     }
 
+    function deepClone(obj) {
+      return JSON.parse(JSON.stringify(obj));
+    }
+
     (function() {
       const PLUGIN_NAME = 'whatif';
 
@@ -4156,17 +4160,10 @@ limitations under the License.
               );
               const key = this.createCombinedValueString_(key1, key2);
               if (!(key in thresholdsMap)) {
-                // Deep copy thresholds for each facet
-                const modelThresholds = [];
-                for (let i = 0; i < this.overallThresholds.length; i++) {
-                  modelThresholds.push({
-                    threshold: this.overallThresholds[i].threshold,
-                  });
-                }
                 thresholdsMap[key] = {
                   value: key1,
                   value2: key2,
-                  threshold: modelThresholds,
+                  threshold: deepClone(this.overallThresholds),
                   opened: false,
                 };
               }
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -4123,6 +4123,17 @@ limitations under the License.
         },
 
         /**
+         * Get the key to which this example belongs when the dataset is sliced by the given feature(s).
+         */
+        getSliceKey_: function(example, feature1, feature2) {
+          // TODO: Aggregate numeric keys into buckets.
+          return this.createCombinedValueString_(
+            example[feature1],
+            example[feature2]
+          );
+        },
+
+        /**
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3288,6 +3292,11 @@ <h2>Show similarity to selected datapoint</h2>
             type: String,
           },
           maxInferenceEntriesPerRun: Number,
+          // Number of buckets when aggregating numeric features.
+          numBuckets: {
+            type: Number,
+            value: 10,
+          },
```
jameswex: can you add this as a numeric paper-input element below the slice by dropdown (when a numeric feature is selected) with a range of 2-10 and default value of 2? can use a <template is="dom-if...> wrapping the element so it only appears in that case. same thing with another dropdown under the secondary slice if that is numeric too. then ensure the buckets are recalculated whenever numPrimaryBuckets or numSecondaryBuckets changes.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -3676,6 +3685,7 @@ <h2>Show similarity to selected datapoint</h2>
             type: Boolean,
             value: false,
           },
+          featureBucketEdges_: Object,
```
jameswex: add short comment on what it is

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -4165,77 +4201,30 @@ <h2>Show similarity to selected datapoint</h2>
           // back to custom (default).
           this.resetOptimizationSelected_();
 
-          const feature1 = this.selectedBreakdownFeature;
-          if (feature1 == '') {
-            this.selectedSecondBreakdownFeature = '';
-          }
```
jameswex: probably still want this logic here, which clears out the secondary slice selection when the primary slice selection is cleared. otherwise if you have both slices set, then clear the primary, the view goes back to showing only overall stats, but then when you select a slice again, the secondary slice is already selected as whatever it was last and you don't just see the primary slice you selected.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -4165,77 +4201,30 @@ <h2>Show similarity to selected datapoint</h2>
...
-                    thresholds.push(thresh);
-                    const mapKey = this.createCombinedValueString_(
-                      feature1Value,
-                      feature2Value
-                    );
-                    thresholdsMap[mapKey] = thresh;
-                  }
-                }
+          this.visdata.forEach(
+            function(item) {
```
jameswex: use arrow notation to simplify this, remove need for .bind(this) :
this.visdata.forEach(item => { blah blah.... return blah;});

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -4155,6 +4165,32 @@ <h2>Show similarity to selected datapoint</h2>
...
+         * Get the key to which this example belongs when the dataset is sliced by the given feature(s).
+         */
+        getSliceKey_: function(example, feature1, feature2) {
+          const bucketEdges = this.featureBucketEdges_;
+          function maybeAggregate(feature) {
+            const edges = bucketEdges[feature];
+            if (edges) {
+              for (let i = 1; i < edges.length; i++) {
+                if (
+                  example[feature] < edges[i] ||
```
jameswex: how does this handle when an example is missing the feature value for the selected feature?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -4165,77 +4201,30 @@ <h2>Show similarity to selected datapoint</h2>
...
+                  threshold: deepClone(this.overallThresholds),
+                  opened: false,
+                };
               }
-            }
-          }
+            }.bind(this)
+          );
+          const thresholds = Object.values(thresholdsMap);
+          this.set('selectedFeatureSort', 'Alphabetical');
```
jameswex: why reset the sort order when slices change? i can imagine wanting to keep viewing slices by something like false positive rate as i change slicing criteria.

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -6136,6 +6146,46 @@ <h2>Show similarity to selected datapoint</h2>
...
+            // No point in aggregating if less unique values than buckets.
+            this.stats[feature].uniqueCount > this.numBuckets
+          );
+        },
+
+        /**
+         * Calculate edges between buckets for aggregating numeric features.
+         * We do this beforehand to round numbers and avoid ugly interval labels.
+         */
+        calculateFeatureBucketEdges_: function() {
```
jameswex: nice :)

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
grovina: :+1:
Screenshot from 2019-09-19 16-36-07

jameswex:
Thanks. Only issue is that the you don't necessary end up with equally-sized intervals when slicing by a numeric feature. Slice by age with 4 buckets and some buckets are width 10 and some width 20 (and the initial one is width 13 due to the min age of 17. Can the logic be adjusted so they all end up of width 10 or 20? Or we would need to document this in the tool to avoid confusion

grovina:
> Thanks. Only issue is that the you don't necessary end up with equally-sized intervals when slicing by a numeric feature. Slice by age with 4 buckets and some buckets are width 10 and some width 20 (and the initial one is width 13 due to the min age of 17. Can the logic be adjusted so they all end up of width 10 or 20? Or we would need to document this in the tool to avoid confusion

Hey James, this is purely due to rounding the edges of the buckets. Increasing precision by 1 order of magnitude should solve it:
Screenshot from 2019-09-19 17-52-52

Note that the 3rd bucket (`[49, 64)`) has width 15, and the others 16: that's pretty much as good as it gets for reasonably rounded numbers. What that rounding does is to make sure these oscillations are smaller than the order of magnitude of the std. deviation (actually 1 order below it after that last commit).

jameswex:
@nfelt please review or assign for review

jameswex:
@stephanwlee please review or assign for review, thx

nfelt:
@jameswex is there a reason this needs an additional review?  I think your approval is sufficient to merge this since it only affects WIT code.

jameswex:
> @jameswex is there a reason this needs an additional review? I think your approval is sufficient to merge this since it only affects WIT code.

Historically we've given TB a chance to review our code before we commit to the repo, but I'm ok merging this with just my approval if you are.

nfelt:
Fair enough - I appreciate the notice!  But I think going forward it's fine to merge after your review for PRs that only affect the WIT plugin codebase.  Of course if there are bigger structural changes or cases where you want specific reviewer expertise you're still welcome to request review, but don't feel like you need to be blocked on a TB core team reviewer for changes like this one.

PR #2681, opened 2019-09-19
Move angular build rules to a common location

davidsoergel:
Per offline discussion, this makes it clearer that Angular plugins can share some BUILD rules currently used only in tf_ng_tensorboard.


davidsoergel:
Abandoning as this is obviated by #2896.

PR #2682, opened 2019-09-19
Fix threshold optimizations in WIT

jameswex:

Polymer2 update broke the updating of threshold sliders in WIT when certain optimization buttons are pressed.

Simplified threshold setting data-binding logic. No longer is there a map and a list to the same threshold information that is data-bound to paper-sliders. Now there is a simple list and the map is still used to quickly find items in the list by slice, but the map contains an index to the list and not the same object as is contained in the list. This fixes the issue where the updates to the slider bound values doesn't always cause the physical slider to update its setting.

Ran through demos apps, sliced performance statistics by a feature, and verified that all threshold optimization buttons now correctly update the sliders, whereas before the sliders would stay stuck in one position even though other information would correctly update.


jameswex:
@tolga-b please review

PR #2686, opened 2019-09-20
program_test: bind to all interfaces

wchargin:
Summary:
These tests are currently broken on IPv6-only systems. While it’s valid
to open a _socket_ that binds to `("localhost", 0)` even if the system
only supports IPv6, it’s not valid to start a _Werkzeug server_ with
host `localhost`, because Werkzeug detects `localhost` as an IPv4
address and will open an `AF_INET` socket, causing catastrophe.

This restores the test behavior prior to #2589, which IMHO isn’t great,
but unbreaks these tests on IPv6-only systems.

Test Plan:
This test still works on my dual-IPv4/IPv6 machine, and now also works
on an IPv6-only machine where previously it failed with `EAFNOSUPPORT`.

wchargin-branch: program-test-bind-all

### DIFF
--- a/tensorboard/program_test.py
+++ b/tensorboard/program_test.py
@@ -73,7 +73,8 @@ class WerkzeugServerTest(tb_test.TestCase):
 
   def make_flags(self, **kwargs):
     flags = argparse.Namespace()
-    kwargs.setdefault('bind_all', False)
+    kwargs.setdefault('host', None)
+    kwargs.setdefault('bind_all', kwargs['host'] is None)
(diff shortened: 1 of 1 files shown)


wchargin:
WANT_LGTM=any

psybuzz:
Using host=None by default LGTM

wchargin:
Yes; as far as I can tell, the now-default options will fail on an
IPv6-only machine. I don’t have any local IPv6-only machines to confirm,
but I opened #2691 for tracking.

PR #2688, opened 2019-09-21
ui: add expand arrow to category paginated view

psybuzz:
The UI could make it clearer when sections in a category-paginated-view are expanded/collapsed at first glance, and indicate that expansion is an affordance.

  - Adds an iron-icon to the right of the section heading count that is '>' when collapsed, 'v' when opened
  -  Truncates long section names to prevent overflow
  -  Gives section name text a tooltip when hovered
  -  Only show count if there are multiple items

* Screenshots of UI changes
image
image

### DIFF
--- a/tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
+++ b/tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
@@ -14,6 +14,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
+<link rel="import" href="../iron-icon/iron-icon.html" />
 <link rel="import" href="../iron-collapse/iron-collapse.html" />
 <link rel="import" href="../paper-button/paper-button.html" />
 <link rel="import" href="../paper-input/paper-input.html" />
@@ -53,7 +54,7 @@ limitations under the License.
             </template>
             <template is="dom-if" if="[[!_isCompositeSearch(category)]]">
               <span class="light">Tags matching /</span>
-              <span class="category-name">[[category.name]]</span>
+              <span class="category-name" title=[[category.name]]>[[category.name]]</span>
               <span class="light">/</span>
               <template is="dom-if" if="[[_isUniversalSearchQuery]]">
                 <span> (all tags)</span>
@@ -64,10 +65,13 @@ limitations under the License.
             </template>
           </template>
           <template is="dom-if" if="[[!_isSearchResults]]">
-            <span class="category-name">[[category.name]]</span>
+            <span class="category-name" title=[[category.name]]>[[category.name]]</span>
           </template>
         </span>
-        <span class="count"><span>[[_count]]</span></span>
+        <span class="count">
+          <span>[[_count]]</span>
+          <iron-icon icon="expand-more" class="expand-arrow"></iron-icon>
+        </span>
       </button>
       <!-- TODO(stephanwlee): investigate further. For some reason,
         transitionend that the iron-collapse relies on sometimes does not
@@ -154,6 +158,9 @@ limitations under the License.
         line-height: 1;
         box-shadow: 0 1px 5px rgba(0, 0, 0, 0.2);
         padding: 10px 15px;
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
       }
 
       .heading::-moz-focus-inner {
@@ -165,9 +172,16 @@ limitations under the License.
         border-bottom-right-radius: 0 !important;
       }
 
+      [open-button] .expand-arrow {
+        transform: rotateZ(180deg);
+      }
+
       .name {
         display: inline-flex;
-        float: left;
+        padding: 2px 0;
+        white-space: pre;
+        overflow: hidden;
+        text-overflow: ellipsis;
       }
 
       .light {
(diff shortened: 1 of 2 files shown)


stephanwlee:
Two things off the bat: when collapsed, please use the up-chevron instead of right-chevron. It aligns better with material stuff (you may argue that the tree view has this but that's for the tree!).

Also consider using just CSS transform instead of switching the icon type. I don't trust Polymer to be efficient for some reason.

psybuzz:
> when collapsed, please use the up-chevron instead of right-chevron. It aligns better with material stuff (you may argue that the tree view has this but that's for the tree!).

Okay.  Did you mean down-chevron when collapsed?  Looking at some Material examples, it seems they have collapsed=pointing-down, expaded=pointing-up
https://material.angular.io/components/expansion/overview
https://material.io/components/lists/#types
https://material.io/archive/guidelines/components/expansion-panels.html#



> Also consider using just CSS transform instead of switching the icon type. I don't trust Polymer to be efficient for some reason.

Me too.  Done.

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
```diff
@@ -154,6 +164,9 @@
         line-height: 1;
         box-shadow: 0 1px 5px rgba(0, 0, 0, 0.2);
         padding: 10px 15px;
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
       }
```
stephanwlee: there are so many counter examples but prefer to sort the CSS a-z (as per style guide but it is hard to enforce that)

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
```diff
@@ -165,9 +178,16 @@
...
+        transform: rotateZ(180deg);
+      }
+
       .name {
         display: inline-flex;
-        float: left;
+        padding: 2px 0;
+        white-space: pre;
+        overflow: hidden;
+        text-overflow: ellipsis;
```
stephanwlee: would you be able to provide screenshot with this change?

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
```diff
@@ -165,9 +178,16 @@
...
 
+      [open-button] .expand-arrow {
+        transform: rotateZ(180deg);
+      }
+
       .name {
         display: inline-flex;
-        float: left;
+        padding: 2px 0;
+        white-space: pre;
```
stephanwlee: forgot: how does it differ from `nowrap`?

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
```diff
@@ -282,6 +300,10 @@
           type: Number,
           computed: '_computeCount(category.items.*)',
         },
+        _hasMultiple: {
+          type: Number,
```
stephanwlee: `type: Boolean`

psybuzz: Done.

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
psybuzz: Sequences of white space are preserved.  E.g. if I type "a    b" (many spaces in the middle) in the search field, I prefer seeing what I typed, instead of "a b" (collapsed to 1 space).

Since .category-name actually holds the text, I think this one is unnecessary.

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
psybuzz: I've updated the screenshot in the description to have a case with a long, truncated title name.

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
psybuzz: Done, sorted this rule.

I would really prefer updating our style guide, a-z makes it much harder to read, in my opinion.  "align" and "justify" are very related to "display", and seeing them sorted a-z makes it harder to reason about layout.

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
```diff
@@ -53,7 +54,7 @@
             </template>
             <template is="dom-if" if="[[!_isCompositeSearch(category)]]">
               <span class="light">Tags matching /</span>
-              <span class="category-name">[[category.name]]</span>
+              <span class="category-name" title=[[category.name]]>[[category.name]]</span>
```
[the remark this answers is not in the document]
psybuzz: Done.

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
stephanwlee: For posterity, this the styleguide I am referring to: https://google.github.io/styleguide/htmlcssguide.html#Declaration_Order

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
```diff
@@ -176,13 +193,17 @@
...
       }
 
       .count {
-        float: right;
-        margin-right: 5px;
+        margin: 0 5px;
         font-size: 12px;
         color: var(--paper-grey-500);
+        display: flex;
+        align-items: center;
```
stephanwlee: you may want to add flex: 0; (also consider using a-z :) )

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
psybuzz: Thanks for the link.  I realize that we should follow a guide, but since the codebase is already unsorted, and we have no enforcement atm, I would prefer to defer style-guide-alignment to later.

tensorboard/components/tf_paginated_view/tf-category-paginated-view.html
psybuzz: Done as flex: none.

Deferring a-z to another day.

PR #2693, opened 2019-09-24
ng: data loader

stephanwlee:
tf-tensorboard is responsible for loading data for runs, experiments and
trigger reloading. Without this feature, the runStore will be empty and
will not make the tf-runs-selector operable.

This change implements this behavior in Angular in flux way.

### DIFF
--- a/tensorboard/components/tf_ng_tensorboard/core/core.effects.ts
+++ b/tensorboard/components/tf_ng_tensorboard/core/core.effects.ts
@@ -13,16 +13,27 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {Injectable} from '@angular/core';
+import {Action, Store} from '@ngrx/store';
 import {Actions, ofType, createEffect} from '@ngrx/effects';
-import {of} from 'rxjs';
-import {map, flatMap, catchError} from 'rxjs/operators';
-
+import {Observable, of, zip} from 'rxjs';
+import {
+  map,
+  mergeMap,
+  catchError,
+  withLatestFrom,
+  filter,
+  tap,
+} from 'rxjs/operators';
 import {CoreService} from './core.service';
 import {
   coreLoaded,
+  reload,
+  pluginsListingRequested,
   pluginsListingLoaded,
   pluginsListingFailed,
 } from './core.actions';
+import {State, getPluginsListLoaded} from './core.reducers';
+import {LoadState} from '../types/api';
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 /** @typehack */ import * as _typeHackNgrx from '@ngrx/store/src/models';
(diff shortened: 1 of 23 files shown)


stephanwlee:
+r psybuzz for exposure to Angular stuff :)

stephanwlee:
In the hindsight, I feel super bad about the large diff :( Sorry!

tensorboard/components/tf_ng_tensorboard/core/core.reducers.ts
```diff
@@ -29,30 +33,85 @@ import * as actions from './core.actions';
...
+  }),
+  on(actions.toggleReload, (state: CoreState) => {
+    return {
+      ...state,
+      reloadEnabled: !state.reloadEnabled,
+    };
+  }),
+  on(actions.changeReloadPeriod, (state: CoreState, {periodInMs}) => {
+    const nextReloadPeriod =
+      periodInMs > 0 ? periodInMs : state.reloadPeriodInMs;
```
psybuzz: Can we treat period = 0 as "never reload"?  (if it's not controversial)

tensorboard/components/tf_ng_tensorboard/reloader/reloader.component.ts
```diff
@@ -0,0 +1,72 @@
...
+  State,
+  getReloadEnabled,
+  getReloadPeriodInMs,
+} from '../core/core.reducers';
+import {reload} from '../core/core.actions';
+
+/** @typehack */ import * as _typeHackRxjs from 'rxjs';
+
+@Component({
+  selector: 'reloader',
```
psybuzz: (i might be totally wrong) but if there's no plan to use the actual DOM Element, should this be a ReloaderService provided to the main AppComponent?

also, can we drop the template: ''?

stephanwlee: I really hope to make our codebase more like Redux instead of buying into Angular's service-based architecture. In other words, I do not want to see the store injected into any service if possible. In React world, DOM-less functional component is a pattern that I seldom employed. It was basically for enabling/disabling a feature in declarative manner.

Dunno why Angular insisted that we have it but it complains without the template.

tensorboard/components/tf_ng_tensorboard/core/core.reducers.ts
stephanwlee: When user toggles enable/disable reload, I hope it to remember this value, so no.

tensorboard/components/tf_ng_tensorboard/core/core.effects.ts
```diff
@@ -36,22 +47,27 @@ export class CoreEffects {
...
-        this.coreService
-          .fetchPluginsListing()
-          .pipe(
-            map(
-              (plugins) => pluginsListingLoaded({plugins}),
-              catchError(() => of(pluginsListingFailed()))
-            )
-          )
-      )
+      ofType(coreLoaded, reload),
```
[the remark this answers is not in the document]
stephanwlee: Here, I am simply trying to mimic the behavior of tf-tensorboard. Btw, pluginsListing *do* change and environments list technically can change.

tensorboard/components/tf_ng_tensorboard/core/core.reducers.ts
psybuzz: That makes more sense

tensorboard/components/tf_ng_tensorboard/core/core.effects.ts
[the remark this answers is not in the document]
psybuzz: I see!  Let's be sure to use them in a followup.  Otherwise, if there's no plan to use this code, maybe we can remove for now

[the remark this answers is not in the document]
stephanwlee: > Otherwise, if there's no plan to use this code, maybe we can remove for now

What do you mean?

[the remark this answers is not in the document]
psybuzz: Ah, I now see why fetching pluginsListing and environment make sense.
Per offline discussion: 
- if a user stops the running TB process and restarts it with a different logdir, the environment can change 
- since we have the notion of active/inactive plugins, the list needs to be refreshed (although ideally we might not need this notion of active/inactive)

tensorboard/components/tf_ng_tensorboard/reloader/reloader.component.ts
psybuzz: Thanks for explaining this!  While it isn't interfering with the app, I guess these functional components in DOM are okay, as long as they stay at the root level.

PR #2694, opened 2019-09-24
Fix Fairness optimization dialogs (What-If Tool)

grovina:
All Fairness optimization info icons were opening the first one (Custom threshold) when clicked.
Also updated description for Equal Opportunity optimisation, which was inaccurate.

Made function `openDialog` open the next sibling to the clicked element.
Also checked that all `paper-dialog` DOMs were positioned where they should, so that this function would work for all cases -- the only two cases I moved would actually work as they were, but we'd rather have positioned in the same way as the others (they were children; we want siblings).

* Screenshots of UI changes
Only actual change is the Equal Opportunity description.
Screenshot 2019-09-24 at 13 54 51

Open each dialog and confirm it's the one supposed to open:
   - Nearest counterfactual
   - Similarity to selected datapoint
   - Partial Dependence Plots
   - Ground truth
   - Cost ratio
   - Fairness (6 of them)
   - Classification Performance Table
   - Regression Performance Table
   - Exploring Classification Performance
   - ROC curve
   - PR curve

N/A

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -2021,25 +2021,24 @@ limitations under the License.
                               The feature that your model is trying to predict.
                               <span class="dialog-link" on-tap="openDialog"
                                 >More.
-                                <paper-dialog
-                                  class="dialog-text"
-                                  horizontal-align="auto"
-                                  vertical-align="auto"
-                                >
-                                  <div class="dialog-title">
-                                    Ground Truth Feature
-                                  </div>
-                                  <div>
-                                    If the datapoints contain a feature that
-                                    represents the ground truth for what the
-                                    model is attempting to predict, then
-                                    selecting that feature here allows the tool
-                                    to investigate the performance of the model
-                                    by comparing the model's results to the
-                                    ground truth feature.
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -2302,10 +2301,9 @@ <h2>Show similarity to selected datapoint</h2>
                                 slices achieve equal opportunity.
                               </div>
                               <div>
-                                Equal opportunity means that among those
-                                predicted as positive classifications, there is
-                                a similar percentage of correct predictions in
-                                each slice.
+                                Equal opportunity means that a similar fraction
+                                of the real positives is recalled for each slice
+                                (i.e., <i>sensitivity</i> is the same).
```
jameswex: I think this definition is confusing in its use of "real positives" and "recalled" as we want this to be clear even to novices. Let's keep it as close to the initial wording as possible.

Perhaps "Equal opportunity means that among those datapoints with the positive ground truth label, there is a similar percentage of correct predictions in each slice".

grovina: Sounds good!

grovina:
For completeness: the Settings dialog is not concerned here, because it uses a separate function (`settingsClicked_`).

PR #2696, opened 2019-09-26
Various fixes from latest changes

jameswex:

Bashing bugs before witwidget 1.4 launch.

Fixed datapoint editor panel slider to work in all environments (demos were broken) - by explicitly using the correct dom element and also ensuring some content flex-wraps.

Added checks for when extra outputs is null so it doesn't break WIT when that happens.

Allow fallback in PD plot checking when a string can't be decoded as text (for images).

Fixed boolean setting of if custom distance function set, for correct enabling/disabling of L1/L2 radio buttons.

Ran all demos and verified all features work as expected (including PD plots)
Ran notebook demos that show attributions and extra outputs, along with custom distance setting.

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -6255,7 +6255,10 @@ limitations under the License.
             modelNum < extraOutputs.extra.length;
             modelNum++
           ) {
-            if ('attributions' in extraOutputs.extra[modelNum]) {
+            if (
+              extraOutputs.extra[modelNum] != null &&
+              'attributions' in extraOutputs.extra[modelNum]
+            ) {
               attributions.push(extraOutputs.extra[modelNum].attributions);
             }
           }
@@ -6275,6 +6278,9 @@ limitations under the License.
               modelNum < extraOutputs.extra.length;
               modelNum++
             ) {
+              if (extraOutputs.extra[modelNum] == null) {
+                continue;
+              }
               const keys = Object.keys(extraOutputs.extra[modelNum]);
               for (let j = 0; j < keys.length; j++) {
                 const key = keys[j];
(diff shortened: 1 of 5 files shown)


tensorboard/plugins/interactive_inference/utils/inference_utils.py
```diff
@@ -230,7 +230,10 @@ def __init__(self, inference_address, model_name, model_type, model_version,
 
 def ensure_not_binary(value):
   """Return non-binary version of value."""
-  return value.decode() if isinstance(value, binary_type) else value
+  try:
+    return value.decode() if isinstance(value, binary_type) else value
```
stephanwlee: Can we update a test with this check?

tensorboard/plugins/interactive_inference/utils/inference_utils.py
```diff
@@ -230,7 +230,10 @@ def __init__(self, inference_address, model_name, model_type, model_version,
...
+  except UnicodeDecodeError:
```
tolga-b: when does this happen? can you put a short comment here?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -6255,7 +6255,10 @@ <h2>Show similarity to selected datapoint</h2>
             modelNum < extraOutputs.extra.length;
             modelNum++
           ) {
-            if ('attributions' in extraOutputs.extra[modelNum]) {
+            if (
+              extraOutputs.extra[modelNum] != null &&
```
tolga-b: I don't remember the exact code path for this right now. Can there be a case where one model returns extraOutputs and other doesn't? In that case do we still create the extraOutputs.extra array of the correct size (2) and fill one with null?

tensorboard/plugins/interactive_inference/utils/inference_utils.py

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
jameswex: yes that is the behavior

PR #2698, opened 2019-09-26
Add ability to adjust attributions.

jameswex:

For CAIP models that return attributions, if users need an adjust_example fn, they would also need an adjust_attribution fn to convert attribution values back to a map of the pre-adjusted features, so WIT can map the attributions to the features it knows about.

Add adjust_attribution function in same manner as adjust_example and adjust_prediction. Call it if attributions are returned.

* Screenshots of UI changes

N/A

Ran against real model with attributions and adjusted examples to verify that this optional function is called and can correctly display attributions for adjusted examples.


jameswex:
> Looks good to me. One general question is would adjust_attribution need access to the example itself as it is more paired with adjust_example?

It doesn't seem like we'll need that for CAIP adjust_attribution because it only needs to correct the attributions in reverse of what adjust_example does, which is more just for adjusting feature types (like categorical to one-hot).

PR #2700, opened 2019-09-26
ng: invoke dashboard reload

stephanwlee:
tf-tensorboard invokes reload method on the plugins when plugins listing
are reloaded.

This type of component/code naturally won't be there in Ngrx based application,
but it had to be introduced for Polymer-interop.


tensorboard/components/tf_ng_tensorboard/header/containers/header.component.test.ts
```diff
@@ -25,12 +25,13 @@ import {provideMockStore, MockStore} from '@ngrx/store/testing';
...
-import {State, CoreState} from '../../core/core.reducers';
+import {State} from '../../core/core.reducers';
 import {
   createPluginMetadata,
   createState,
   createCoreState,
 } from '../../core/testing';
+import {PluginId} from '../../types/api';
 
 /** @typehack */ import * as _typeHackStore from '@ngrx/store';
```
davidsoergel: what's this?  Why not `import {Foo, Bar} from '@ngrx/store';`?

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.test.ts
```diff
@@ -145,4 +150,24 @@ describe('header.component', () => {
     expect(dispatch).toHaveBeenCalledTimes(1);
     expect(dispatch).toHaveBeenCalledWith(changePlugin({plugin: 'bar'}));
   });
+
+  it('does not keep observer after getting the most recent value', async () => {
```
davidsoergel: This test pattern seems to mix presentation-level concerns (e.g., `By.css(...)` and `click()`) with data model/connector concerns, i.e. ngrx/rxjs stuff.

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.test.ts
```diff
@@ -145,4 +150,24 @@ describe('header.component', () => {
...
+    const fixture = TestBed.createComponent(HeaderComponent);
+    fixture.detectChanges();
+
+    const [, debugEl] = fixture.debugElement.queryAll(By.css('.mat-tab-label'));
+    debugEl.nativeElement.click();
+    fixture.detectChanges();
+    await fixture.whenStable();
+    expect(dispatch).toHaveBeenCalledTimes(1);
+
+    // If an observer is still listening, this should cause the store to dispatch.
```
davidsoergel: Sorry, I don't follow.  Which observer were you worried might still be listening, but in fact should not be?  Also, `setActivePlugin()` calls `setState()` but not `dispatch()`, so I'm not sure this is actually testing what you meant.

davidsoergel: Oh, is this about the `take(1)` at header.component.ts:70?  If so refactoring it per my comment there may obviate the need for the test.

tensorboard/components/tf_ng_tensorboard/plugins/plugins.component.ts
```diff
@@ -39,6 +45,13 @@ const activePlugin = createSelector(
   }
 );
 
+const lastLoadedTimeInMs = createSelector(
```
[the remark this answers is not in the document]
stephanwlee: I don't agree with this.

1. There are selectors that are store aware in the respective reducers. These selectors are unique to the frontend and do perform some transformations for presentational logic which should reside in the view.
2. Unlike Redux, Ngrx with Angular has modular reducer wrapped around in `NgModule` that can be even lazy loaded. `store` is a generic name that does not mean anything and word `core` here is simply to mimic that of the `CorePlugin` on the backend.

tensorboard/components/tf_ng_tensorboard/plugins/plugins.component.ts
```diff
@@ -57,8 +74,28 @@ export class PluginsComponent implements OnInit {
     // We manually create plugin DOM (with custom tagName and script inside
     // an iframe) when the `activePlugin` changes.
     this.activePlugin$
-      .pipe(filter(Boolean))
-      .subscribe((plugin) => this.renderPlugin(plugin as UiPluginMetadata));
+      .pipe(
+        filter(Boolean),
```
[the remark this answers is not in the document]
stephanwlee: You are correct, but as a view, I don't want to make such assumptions about the how store initialization works. We just know that it is null-able and we are not interested in that case.

tensorboard/components/tf_ng_tensorboard/plugins/plugins.component.ts
```diff
@@ -57,8 +74,28 @@ export class PluginsComponent implements OnInit {
...
+        distinctUntilChanged(
+          (prev: UiPluginMetadata, curr: UiPluginMetadata) =>
+            prev.id === curr.id
+        )
+      )
+      .subscribe((plugin: UiPluginMetadata) => this.renderPlugin(plugin));
```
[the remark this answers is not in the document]
stephanwlee: I am aware except I was not relying on Angular to do the plugin rendering for me since they are polymer components.

Of course, you can argue that a thin `component` that implements `OnChange` is simpler than using observable. Is that the fair representation of your comment here?

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.test.ts
stephanwlee: Sorry, what are you suggesting here? Are you suggesting that I should stub the callback of the click?

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.test.ts
stephanwlee: That still makes TSC under bazel complain. :man_shrugging:

stephanwlee:
> For one thing, if the plugin list is updated, why does that mean that the Polymer dashboards should have reload() called on them? Is it because their PluginMetadata might have changed--which Angular would just handle, but here we have to kick the Polymer components to update? If that's it, which PluginMetadata changes are expected to actually happen (enabled, maybe?).

Kinda. This is copying the behavior of Polymer-based tf-tensorboard and the logic is as follows:
1. Each Polymer based dashboard is responsible for their own data (caching too) and they implement a `reload` interface
2. The contract between TensorBoard and the dashboards is when user wants the data to be reloaded, TensorBoard will invoke the reload method every X-seconds. 
3. Because dashboards care about which runs/experiments are available in the system (they are parameters to the API calls they make), tf-tensorboard first fetches list of runs, list of plugins, and some other environmental info (the last one is not too important).
4. Only after the request for all three are finished, we invoke reload method on the dashboard.

tensorboard/components/tf_ng_tensorboard/plugins/plugins.component.ts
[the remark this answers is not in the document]
davidsoergel: I'm confident that we can factor things in a way that makes sense re. what belongs in the store vs. what is presentation logic, without leaky abstractions.  It's not super urgent, certainly not for this PR, but let's discuss offline reasonably soon.  It'd be good to establish consistent patterns/conventions to use as more things get angularized.

tensorboard/components/tf_ng_tensorboard/plugins/plugins.component.ts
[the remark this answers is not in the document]
davidsoergel: My point is not about the store initialization, except to the extent that it's the reason for making activePlugin$ nullable at all.  As this stands, if either the plugin list or the active plugin ID became null at some point, renderPlugin() would not be called and so the previously active plugin would remain visible (hence out of sync with the state).  IIUC this is not a concern because, once the plugin list and the active id are set to a real value, they should never revert to null.  Right?

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.test.ts
davidsoergel: Huh ok

tensorboard/components/tf_ng_tensorboard/plugins/plugins.component.ts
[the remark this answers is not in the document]
davidsoergel: Yep that's it :)

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.test.ts
davidsoergel: Similar to suggestions elsewhere re separating the store from the view.  You could have separate tests for 1) a click fires the appropriate action, 2) upon receiving the action, the store updates appropriately, and 3) there are no memory leaks in the form of observers that should have been unsubscribed.

Per the other comment I'm still unclear on what this is meant to test.  The title suggests #3, in which case I'd expect you can test that more directly, without involving a click.

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
```diff
@@ -42,35 +39,25 @@ const getDisabledPlugins = createSelector(
...
+  template: `
+    <app-header-component
+      [activePlugins]="plugins$ | async"
+      [disabledPlugins]="disabledPlugins$ | async"
+      [selectedPlugin]="activePlugin$ | async"
+      (onPluginSelectionChanged)="onPluginSelectionChange($event)"
+    ></app-header-component>
+  `,
 })
 export class HeaderComponent {
```
davidsoergel: This can now be `HeaderContainer`, and the filename can be `header-container.component.ts` or something.

Also, I'm not sure that we should separate things into `./components` and `./containers` subdirectories.  That's nbd here--just saying that this is part of settling naming (etc.) conventions at some point.

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.test.ts
davidsoergel: looks like the test was obsoleted in the refactoring.

tensorboard/components/tf_ng_tensorboard/header/containers/header.component.ts
stephanwlee: Will take an action but I want us to talk about it later too.

Also, about the directory structure: let's fix that later holistically.

PR #2701, opened 2019-09-27
Update WIT attribution colors

jameswex:

Updating WIT attribution color scales per mocks

Changed colors and to a 5 color scale for better blending along the gradient.
Removed the duplication of the colors in vz-example-viewer.
Updated at what value the feature value text changes to white due to the background color for attribution.

* Screenshots of UI changes

attrcolors

Ran attribution WIT notebook with attributions with different domains ( all negative, all positive, mix of both) and ensured color and text color looked correct and matched mocks.

### DIFF
--- a/tensorboard/components/vz_example_viewer/vz-example-viewer.ts
+++ b/tensorboard/components/vz_example_viewer/vz-example-viewer.ts
@@ -83,10 +83,7 @@ namespace vz_example_viewer {
   const FLOAT_FEATURE_NAME = 'float';
   const BASE_64_IMAGE_ENCODING_PREFIX = 'base64,';
   const CHANGE_CALLBACK_TIMER_DELAY_MS = 1000;
-  // Colors for the saliency color scale.
-  const posSaliencyColor = '#007B83';
-  const negSaliencyColor = '#FF847C';
-  const neutralSaliencyColor = '#fff';
+  const noAttributionColor = '#fff';
   const defaultTextColor = '#3c4043';
   const lightTextColor = '#fff';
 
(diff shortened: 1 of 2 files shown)


tensorboard/components/vz_example_viewer/vz-example-viewer.ts
```diff
@@ -434,18 +431,18 @@ namespace vz_example_viewer {
     _useLightColor(saliency) {
       const percentile = (saliency - this.minSal) / (this.maxSal - this.minSal);
       if (this.minSal < 0 && this.maxSal > 0) {
-        return percentile < 0.1 || percentile > 0.8;
+        return percentile < 0.3 || percentile > 0.7;
```
tolga-b: are these numbers from Mahima?

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -6403,13 +6405,17 @@ <h2>Show similarity to selected datapoint</h2>
           let attributionColorRange = [];
           if (this.minAttribution < 0) {
             attributionDomain.push(this.minAttribution);
```
tolga-b: This looks good. However, since we have so many color levels now, I wonder if it is better to pass an object with keys as the color strings and values as the domain values above if this is the only place they are being used.

tensorboard/components/vz_example_viewer/vz-example-viewer.ts
jameswex: yeah from her slides

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
jameswex: makes sense as a future improvement

PR #747, opened 2017-11-17
Make TensorBoard work in raw sources mode

jart:
With this change we withdraw our support for the ECMA `import` keyword hitherto
mandated by certain tools. In doing so, TensorBoard compiles in seconds, rather
than taking a minute.

The ES6 `import` standard only defines / and ./ prefixed paths. All other paths
that exist are undefined. Since most tooling and early-adopter codebases rely
on the undefined behavior, the standard itself is mostly syntactic sugar (that
isn't even sweet.) The true problem is having a world where Closure, Node, and
web server namespaces coexist.

The ES6 `import` keyword makes JavaScript incompatible with web browsers. The
support added in Chrome 61 is not sufficient, as it only allows the `export`
keyword on functions and not variables. This is a point of view much different
from early-adopters, which raises concerns that the standard is unlikely to
gain a meaningful consensus.

The ES6 `import` keyword makes TypeScript incompatible with web browsers. This
keyword is not supported by the TypeScript compiler. All it can do is punt the
problem to rollup. However we can't use rollup since it doesn't support Closure
namespaces.

The only tool that supports all three of Web, Node, and Closure namespaces is
the Closure Compiler. Without it, ES6 TypeScript or JavaScript can't run in any
web browser.

But since the Closure Compiler was designed to create production builds for
google.com, it isn't very fast. Therefore it's unreasonable that developers on
a small project like TensorBoard should be expected to wait on this tool
whenever making one line changes in a development iteration cycle.

Since the Closure Tools were originally designed to be compatible with web
browsers in the absense of tooling, this change seeks to bring TensorBoard back
into accordance with those principles.

PR #2703, opened 2019-09-27
plugin: message channels for dynamic plugins

psybuzz:

This iterates on the experimental code in plugin_util/ for communication with IFrames.  This PR does NOT address the public API.

Using `window.postMessage` and `window.addEventListener('message', ...)` works, but
Having multiple concurrent IPCs (e.g. a frame cannot be a host and a guest, a host cannot have multiple guests) required today's `idPrefix` system.  Random-number-generated idPrefixes are prepended to message ids to distinguish which frame it comes from, to resolve the appropriate `responseWait`.

This PR uses [MessageChannels](https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel), instead of random-numbers, to establish a private channel between frames.

The IPC now requires an `isReply: true` to be present on reply messages.  This allows a frame to distinguish a case where (it receives a foreign message with id=X) vs (it receives a reply for its own previous message with id=X).

The flow looks like
- Guest frame sends a normal hostWindow.postMessage to bootstrap, transferring ownership of one of it's MessagePorts
- Host frame establishes an IPC on the received port (no need to worry about stray a postMessage from an extension!)
- Guest can sendMessage(), listen(), unlisten() as usual.  Host can broadcast() on all established ports.

WANT_LGTM=all


psybuzz:
WANT_LGTM=Stephan
FYI=David

I'm still working on the tests.  Planning to cover the isReply cases and stray postMessage, soon.

tensorboard/components/plugin_util/message.ts
```diff
@@ -40,21 +46,15 @@ interface PromiseResolver {
...
-  constructor() {
-    window.addEventListener('message', this.onMessage.bind(this));
-
-    // TODO(tensorboard-team): remove this by using MessageChannel.
-    const randomArray = new Uint8Array(16);
-    window.crypto.getRandomValues(randomArray);
-    this.idPrefix = Array.from(randomArray)
-      .map((int: number) => int.toString(16))
-      .join('');
+  constructor(port) {
```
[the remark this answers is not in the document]
psybuzz: Done

tensorboard/components/plugin_util/message.ts
```diff
@@ -100,22 +99,22 @@ export abstract class IPC {
...
-    targetWindow: Window,
+  sendMessage(
     type: MessageType,
-    payload: PayloadType
+    payload: PayloadType,
   ): Promise<PayloadType> {
-    const id = `${this.idPrefix}_${this.id++}`;
-    const message: Message = {type, id, payload, error: null};
-    this.postMessage(targetWindow, JSON.stringify(message));
+    const id = `${this.id++}`;
```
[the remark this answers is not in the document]
psybuzz: Done

tensorboard/components/plugin_util/plugin-guest.ts
```diff
@@ -45,3 +51,7 @@ export const listen = _guestIPC.listen.bind(_guestIPC);
  * Unsubscribes a callback to a message.
  */
 export const unlisten = _guestIPC.unlisten.bind(_guestIPC);
+
+
+// TODO: Register API for authors.
```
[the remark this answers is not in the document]
psybuzz: Done

tensorboard/components/plugin_util/plugin-host.ts
```diff
@@ -12,46 +12,63 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-   * Sends a message to the frame specified.
-   * @return Promise that resolves with a payload from frame in response to the message.
+   * Sends a message to all dynamic plugins. Individual plugins decide whether
+   * or not to listen.
+   * @return Promise that resolves with a list of payloads from each plugin's
+   *         response (or null) to the message.
    *
    * @example
-   * const someList = await sendMessage('v1.some.type.guest.understands');
+   * const someList = await broadcast('some.type.guest.understands');
```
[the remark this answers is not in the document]
psybuzz: Done

psybuzz:
Added tests, cr address.  Ready for another look!

tensorboard/components/plugin_util/message.ts
```diff
@@ -40,21 +46,14 @@ interface PromiseResolver {
...
-    window.addEventListener('message', this.onMessage.bind(this));
-
-    // TODO(tensorboard-team): remove this by using MessageChannel.
-    const randomArray = new Uint8Array(16);
-    window.crypto.getRandomValues(randomArray);
-    this.idPrefix = Array.from(randomArray)
-      .map((int: number) => int.toString(16))
-      .join('');
+  constructor(private port: MessagePort) {
+    this.port = port;
```
stephanwlee: hmm isn't this extraneous?

tensorboard/components/plugin_util/message.ts
```diff
@@ -40,21 +46,14 @@ interface PromiseResolver {
...
+    port.addEventListener('message', this.onMessage.bind(this));
```
stephanwlee: `bind` is a bit ew unless it is required. consider using fancy arrow function?

tensorboard/components/plugin_util/message.ts
```diff
@@ -27,9 +32,10 @@ export type PayloadType =
 
 export interface Message {
   type: string;
-  id: string;
+  id: number;
   payload: PayloadType;
   error: string | null;
+  isReply: boolean;
```
stephanwlee: Is this for readability of the module? Or does it have functional differences?

tensorboard/components/plugin_util/plugin-guest.ts
```diff
@@ -12,19 +12,27 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-  }
+import {IPC, Message} from './message.js';
+
+/**
+ * This code is part of a public bundle provided to plugin authors,
+ * and runs within an IFrame to setup communication with TensorBoard's frame.
+ */
+if (!window.parent) {
+  throw Error(
+    'This library must be run from within a loaded TensorBoard dynamic plugin.'
```
stephanwlee: Thanks for adding the invariance. `...be run from within` part reads a bit funny to me. Can this be reworded as `The library must run within a TensorBoard iframe-based plugin.`. (dynamic-ness is irrelevant too)

tensorboard/components/plugin_util/plugin-host.ts
```diff
@@ -12,46 +12,87 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+): Promise<PayloadType[]> {
+  for (const ipc of portIPCs) {
+    if (!ipcToFrame.get(ipc).isConnected) {
+      portIPCs.delete(ipc);
+      ipcToFrame.delete(ipc);
+    }
   }
+
+  const ipcs = [...portIPCs];
+  const promises = ipcs.map((ipc) => ipc.sendMessage(type, payload));
```
stephanwlee: opinion: combine L55 and L56?

tensorboard/components/plugin_util/plugin-host.ts
```diff
@@ -12,46 +12,87 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    portIPC.listen(type, callback);
+  }
+});
+
+function _broadcast(
+  type: MessageType,
+  payload: PayloadType
+): Promise<PayloadType[]> {
+  for (const ipc of portIPCs) {
+    if (!ipcToFrame.get(ipc).isConnected) {
```
stephanwlee: I presume this cleanup code can removed with your TODO in L21?

tensorboard/components/plugin_util/plugin-host.ts
```diff
@@ -12,46 +12,87 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+const VERSION = 'experimental';
+const listeners = new Map<MessageType, MessageCallback>();
+
+// TODO(@psybuzz): replace this and the port cleanup logic in broadcast() with
+// a MutationObserver to notify us when iframes disconnect.
+const ipcToFrame = new Map<IPC, HTMLIFrameElement>();
+
+// The initial Window-level listener is needed to bootstrap only.
+// All further communication is done over MessagePorts.
+window.addEventListener('message', (event) => {
```
stephanwlee: consider creating a method like `onBootstrap` that you use here. I expect that method to get lengthier in the future if we have to support different versions of it.

tensorboard/components/plugin_util/test/plugin-test.ts
```diff
@@ -54,25 +55,31 @@ namespace tf_plugin.test {
...
         spec: 'guest (src) to host (dest)',
         beforeEachFunc: function() {
+          this.destWindow = window;
           this.destListen = pluginHost.listen;
           this.destUnlisten = pluginHost.unlisten;
-          this.srcSendMessage = (type, payload) => {
-            return this.guestWindow.test.sendMessage(type, payload);
+          this.destSendMessage = (type, payload) => {
+            return new Promise(async (resolve) => {
+              const results = await pluginHost.broadcast(type, payload);
```
stephanwlee: any thoughts on rewriting this as 
```ts
return pluginHost.broadcast(type, payload).then(([result]) => result);
```

tensorboard/components/plugin_util/test/plugin-test.ts
```diff
@@ -175,6 +179,45 @@ namespace tf_plugin.test {
...
+          });
+
+          const longTaskCb = this.sandbox.stub();
+          const longTaskPromise = this.srcSendMessage('longTask', 'hello').then(
+            longTaskCb
+          );
+
+          await this.srcSendMessage('foo');
+          await this.destSendMessage('bar');
+          expect(longTaskCb.callCount).to.equal(0);
```
stephanwlee: consider using the fluent alternative.
`expect(longTaskCb).to.not.have.been.called;`

tensorboard/components/plugin_util/test/plugin-test.ts
```diff
@@ -175,6 +179,45 @@ namespace tf_plugin.test {
...
+
+          resolveLongTask();
+          await longTaskPromise;
+          expect(longTaskCb.callCount).to.equal(1);
```
stephanwlee: consider using the fluent alternative.
`expect(longTaskCb).to.have.been.calledOnce;`

tensorboard/components/plugin_util/test/plugin-test.ts
```diff
@@ -175,6 +179,45 @@ namespace tf_plugin.test {
...
+
+        it('processes messages while waiting for a reponse', async function() {
+          let resolveLongTask = null;
+          this.destListen('longTask', () => {
+            return new Promise((resolve) => {
+              resolveLongTask = resolve;
+            });
+          });
+
+          const longTaskCb = this.sandbox.stub();
```
stephanwlee: suffix `Stub` (instead of Cb) may be more legible as it is clear that you are using sinon stub here.

tensorboard/components/plugin_util/test/plugin-test.ts
stephanwlee: consider adding assertion on the payload. `expect(longTaskCb).to.have.been.calledWith(...)`

tensorboard/components/plugin_util/message.ts
psybuzz: whoops, yes.  Done

tensorboard/components/plugin_util/message.ts
psybuzz: Functional difference, and the test I added covers it!

Basically, if a frame receives message `{id: 0, ...}`, it doesn't know whether it's the reply-message to something it previously sent, or whether it's a brand new incoming message from another frame.

In both cases, the origin is a foreign frame, but we need to distinguish that the former case should resolve the responseWait, while the latter should not.

tensorboard/components/plugin_util/plugin-host.ts
psybuzz: Yes!  I can add another comment...

tensorboard/components/plugin_util/message.ts
psybuzz: Done

tensorboard/components/plugin_util/plugin-guest.ts
psybuzz: Done

tensorboard/components/plugin_util/plugin-host.ts
psybuzz: Done

tensorboard/components/plugin_util/plugin-host.ts
psybuzz: Done

tensorboard/components/plugin_util/test/plugin-test.ts
psybuzz: Done, that's readable to me!

tensorboard/components/plugin_util/test/plugin-test.ts
psybuzz: Done

tensorboard/components/plugin_util/test/plugin-test.ts
psybuzz: Done x2

tensorboard/components/plugin_util/test/plugin-test.ts
psybuzz: Hmm, Done

psybuzz:
Comments addressed, ready for another look.

davidsoergel:
I really think we need more consensus on what plugin architecture we're aiming for before checking in code.  The two open docs (Stephan's and Erik's) are a good start, as is the new weekly meeting.  Let's use those venues, as well as perhaps the TB design review, to solidify our plans--and then return to this.

stephanwlee:
@davidsoergel would this _draft_ PR (#2705) what you are seeking? Just do note that this work is time sensitive.

psybuzz:
@davidsoergel as per offline discussion, I've moved plugin_util/ under tensorboard/components/experiments/, to indicate that this is current work is unstable.

I opted to not put it at the root level (tensorboard/experiments/) because there happens to be another build target "plugin_util" at the root level BUILD that has nothing todo with components.

Would this be fine with you?

tensorboard/components/experimental/plugin_util/plugin-host.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function _unlisten(type: MessageType) {
+  listeners.delete(type);
+  for (const ipc of portIPCs) {
+    ipc.unlisten(type);
+  }
+}
+
+export const broadcast = _broadcast;
```
davidsoergel: Is the purpose of this pattern (as opposed to just `export function broadcast(...) {...}`) to prevent monkey-patching?  Just curious if I should adopt this generally.

tensorboard/components/experimental/plugin_util/plugin-host.ts
```diff
@@ -0,0 +1,100 @@
...
+export const listen = _listen;
+export const unlisten = _unlisten;
+
+namespace tf_plugin {
```
davidsoergel: Namespaces are disallowed by go/tsstyle (publicly: https://github.com/google/gts/blob/83cb107010f7a956bb008291066304a9f26a0f34/tslint-rules.json#L43).  TensorBoard contains a fair number of them anyway, but let's not add more if possible.

tensorboard/components/experimental/plugin_util/plugin-host.ts
psybuzz: The underscored version is just there to allow us to export it via es6 (which is used/imported by the test) and export it under a namespace, at the same time.

tensorboard/components/experimental/plugin_util/plugin-host.ts
psybuzz: Good point.  @stephanwlee, could you weigh in on this?

There does seem to be alot of existing namespaces, ever since [this PR](https://github.com/tensorflow/tensorboard/pull/747/), but I think the Angular ng-tensorboard/ folder avoids this.

stephanwlee: 1. plugin-host (at least until the tf_ng_tensorboard work is finished) will be loaded inside Polymer context which uses the namespace
2. (in anticipation of inevitable question) in ideal world, we want to share base code between host and guest impl except it is super hard to make ts_library build rule work with tf_web_library. The latter is used in mainly for polymer based TensorBoard. For now, since the experimental API is only used in context of Polymer anyways,  I decided to just deprecate the [es]module style and use namespace everywhere as v0.

psybuzz: Thanks for the explanation.  It's unfortunate that Polymer code depends on these files.  I'll land as-is for now, but it seems we all agree that 'namespaces' are not ideal, and should be avoided if possible.

PR #2705, opened 2019-09-27
core: implement experimental API for runs

stephanwlee:
With tb.plugin.lib.run, you can now fetch list of runs and subscribe to
changes.

Do note that this change includes changes to the BUILD where we no
longer use modules for the library. We will revisit this in near future
to make this more usable for non-TB Polymer projects.

### DIFF
--- a/tensorboard/components/plugin_lib/BUILD
+++ b/tensorboard/components/plugin_lib/BUILD
@@ -1,0 +1,32 @@
+package(default_visibility = ["//tensorboard:internal"])
+
+load("//tensorboard/defs:web.bzl", "tf_web_library")
+
+licenses(["notice"])  # Apache 2.0
+
+# TODO(stephanwlee): figure out how this tf_web_library can be used to create
+# maybe a NPM package.
+tf_web_library(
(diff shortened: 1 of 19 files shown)


tensorboard/components/plugin_lib/BUILD
```diff
@@ -0,0 +1,32 @@
...
+load("//tensorboard/defs:web.bzl", "tf_web_library")
+
+licenses(["notice"])  # Apache 2.0
+
+# TODO(stephanwlee): figure out how this tf_web_library can be used to create
+# maybe a NPM package.
+tf_web_library(
+    name = "plugin_lib",
+    srcs = [
+        "runs.ts",
```
psybuzz: Since the host impls are named "runs-host-impl.ts", I'd prefer if "runs.ts" was named something like "runs-guest.ts".  Having both host/guest sources in plugin_lib/ would be less confusing if we could see at a glance, which files will be slurped into the guest library.
Or, wdyt about the pattern of separating out a plugin_lib/guest/ folder?

PR #2715, opened 2019-10-02
sec: enable Polymer strictTemplatePolicy

stephanwlee:
As per Polymer/polymer#5323, it roughly
disallows other instances of dom-module with the same id to be present
on the DOM. It roughly means that
- all Polymer components should have accompanying dom-module with a
template inside.
- there cannot be a duplicate definition of dom-module

This change includes also removes instances where our components violate the policy.

It requires #2714.


tensorboard/plugins/graph/tf_graph_loader/tf-graph-loader.ts
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 namespace tf.graph.loader {
   Polymer({
     is: 'tf-graph-loader',
-
+    _template: null,
```
nfelt: Can we add the same comment here, e.g. ` // strictTemplatePolicy requires a template (even a null one).`

tensorboard/plugins/graph/tf_graph_loader/tf-graph-dashboard-loader.html
```diff
@@ -28,6 +28,6 @@
 `selection` from the tf-graph-controls as an input as opposed to URL path of an
 data endpoint.
 -->
-<dom-module id="tf-graph-dashboard-loader"> </dom-module>
+<dom-module id="tf-graph-dashboard-loader"><template></template></dom-module>
```
nfelt: Can we add a comment about why we need the empty template?  Like  `<!-- strictTemplatePolicy requires a template (even a null one) -->`

tensorboard/components/security.html
```diff
@@ -0,0 +1,37 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<script>
+  window.Polymer = {strictTemplatePolicy: true};
+</script>
+<!-- These polymer iron/paper components violate the strictTemplatePolicy. -->
```
nfelt: Can you just briefly elaborate to explain why the lines below address the issue?  Aka "... violate the strictTemplatePolicy by not declaring a <dom-module> element or template.  We define them here ourselves so that strictTemplatePolicy will be satisfied."

tensorboard/components/ng_polymer_lib.html
```diff
@@ -14,6 +14,7 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
+<link rel="import" href="security.html" />
```
nfelt: Does this always have to be imported first?  If so, maybe worth a comment.

stephanwlee: Yes and done!

PR #2716, opened 2019-10-02
Easier WIT custom_predict_fn when using JSON inputs and other witwidget 1.4 fixes

jameswex:

If data is provided to WIT as JSON (not example protos) and the user defines a custom prediction function, the examples passed to that function should be in that JSON format, not in tf.Example format. This makes it much easier to use custom predict functions.

Also fix indexing error during fairness thresholding and fix log level setting.

Call the existing _json_from_tf_examples before passing examples to the custom predict function if they were provided to WIT as JSON originally.

Note: This is technically a breaking change to the API of the custom_predict_fn (when it is used in combination with JSON inputs), so we will need to announce it as such when the next witwidget version is released.

Fix setting of tf logging level to use correct API

Fix setting of featureValueThresholdsIndexMap to always re-index when the list changes.

Ran witwidget in colab with custom predict functions, for both examples with tf.Example input and with JSON input and saw correct behavior in both cases.

Ran notebooks with fairness analysis to verify correct threshold setting (matches witwidget 1.3 values).


jameswex:
@tolga-b this PR also fixes the WIT TF logging level set issue

PR #2719, opened 2019-10-03
vulc: combine script in compile=True and shasum

stephanwlee:
Motivation: we would like to introduce CSP and not trust any script tags
that does not match our shasum.

Change 1: introduce a logic to combine all script tags (in the order
that it was laid out) into one so we can just shasum 1 script tag.

Change 2: Vulcanization outputs another file that includes sha256s of
all script elements's content or content of resource `src` points at. Note
that the .scripts_sha256 file includes newline delimited list of shas without
keys (script names are not named).

Change 3: webfiles.zip now have a _shasum directory that can be used by
Python server.

Future work: change our Python static file serving to include the shasum
in the CSP for HTML files. We special case the _shasum directory in the
zip files to not serve that.

WANT_LGTM=any

### DIFF
--- a/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
+++ b/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
@@ -30,6 +30,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
+import com.google.common.hash.Hashing;
 import com.google.javascript.jscomp.BasicErrorManager;
 import com.google.javascript.jscomp.CheckLevel;
 import com.google.javascript.jscomp.CompilationLevel;
@@ -79,6 +80,7 @@ import org.jsoup.nodes.Node;
 import org.jsoup.nodes.TextNode;
 import org.jsoup.parser.Parser;
 import org.jsoup.parser.Tag;
+import org.jsoup.select.Elements;
 
 /** Simple one-off solution for TensorBoard vulcanization. */
 public final class Vulcanize {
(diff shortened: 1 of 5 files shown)


stephanwlee:
This change breaks the mesh plugin so I am going back to dev mode.

stephanwlee:
Confirmed the fix on mesh and found no regression on most dashboards including projector (which loads from its own vulcanized source)

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -707,6 +718,46 @@ private static boolean shouldIgnoreUri(String uri) {
     return ImmutableMultimap.copyOf(builder);
   }
 
+  // Combines all script tags into one and append it at the bottom of the
+  // document.
+  private static void combineScriptTags(Document document) {
+    Elements scripts = document.getElementsByTag("script");
+    final StringBuilder sources = new StringBuilder();
+    for (Element script : scripts) {
+      if (!script.attr("src").equals("")) continue;
```
wchargin: `if (!script.attr("src").isEmpty())`?

<https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#isEmpty-->

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -707,6 +718,46 @@ private static boolean shouldIgnoreUri(String uri) {
     return ImmutableMultimap.copyOf(builder);
   }
 
+  // Combines all script tags into one and append it at the bottom of the
+  // document.
+  private static void combineScriptTags(Document document) {
```
wchargin: nit: These are script elements, not script tags (the tag is just the
literal `<script>` part).

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -707,6 +718,46 @@ private static boolean shouldIgnoreUri(String uri) {
...
+    Elements scripts = document.getElementsByTag("script");
+    final StringBuilder sources = new StringBuilder();
+    for (Element script : scripts) {
+      if (!script.attr("src").equals("")) continue;
+      sources.append(script.html()).append("\n");
+      script.remove();
+    }
+    // TensorBoard HTMLs tend to be extremely malformed and it can have multiple
```
wchargin: If this is actually JSoup parsing creating the body elements, maybe
amend the comment?

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -707,6 +718,46 @@ private static boolean shouldIgnoreUri(String uri) {
...
+
+    Element scriptTag = new Element(Tag.valueOf("script"), "")
+        .appendChild(new DataNode(sources.toString(), ""));
+    lastBody.appendChild(scriptTag);
+  }
+
+  private static ArrayList<String> computeScriptShasum(Document document) {
+    ArrayList hashes = new ArrayList<String>();
+    for (Element script : document.getElementsByTag("script")) {
+      String hash = Hashing.sha256().hashString(script.html(), UTF_8).toString();
```
wchargin: Maybe add an assertion here that `src` is empty, because otherwise we
won’t actually set up the CSP properly, right?

wchargin:
By the way, did you check the profile plugin’s Google Chart demo?

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
stephanwlee: Thanks for pointing this out. I changed the logic a little now.

Since we have full filepath to all webfiles, we are actually able to get the source content of the script `src` so we now shasum that too.

stephanwlee:
> By the way, did you check the profile plugin’s Google Chart demo?

Thanks for the reminder. I just checked and we just need to lax the style-src to allow gstatic. Thanks again!

PR #2726, opened 2019-10-04
sec: Enable CSP and disallow random script tags

stephanwlee:
For our HTML-based resources, we now apply the CSP so we do not load
unknown image, styles, and scripts.

Design choice made:
- the CSP is applied at the http_util with one API to configure the hash
  part so no response can set the CSP to "unsafe" and make entire
  application unsafe. e.g., dynamic resource is served with "unsafe" and
  the unsafe script tag gets loaded which now has access to the parent
  frame and exfiltrate.

Future work:
- Turn `'strict-dynamic'` off
- Apply CSP to the projector_binary.html.
- CSP needs to work better with the dynamic plugin. One possibility is
  to modify our API so the plugins can define static resources which can
  be shasumed by TensorBoard.
- Above logic can introduce more complexity to the core_plugin. We
  should consider creating richer asset_provider API so it can mediate
  with serving.

tests:
- [x] caused CSP violation to see how browser react (on localhost, too)
- [x] loaded application with CSP
- [x] loaded projector plugin
- [x] set compile=False and confirmed the application working with 420
hashes.
- [x] google_charts loaded properly
- [x] firefox

### DIFF
--- a/tensorboard/backend/http_util.py
+++ b/tensorboard/backend/http_util.py
@@ -107,8 +107,9 @@ def Respond(request,
     content_encoding: Encoding if content is already encoded, e.g. 'gzip'.
     encoding: Input charset if content parameter has byte strings.
     csp_scripts_sha256s: List of base64 serialized sha256 of whitelisted script
-      elements for script-src of the Content-Security-Policy. It is only be used
-      when the content_type is text/html.
+      elements for script-src of the Content-Security-Policy. If it is None, the
+      HTML will disallow any script to execute. It is only be used when the
+      content_type is text/html.
 
   Returns:
     A werkzeug Response object (a WSGI application).
@@ -163,21 +164,29 @@ def Respond(request,
   else:
     headers.append(('Expires', '0'))
     headers.append(('Cache-Control', 'no-cache, must-revalidate'))
-  if mimetype == _HTML_MIMETYPE and csp_scripts_sha256s:
+  if mimetype == _HTML_MIMETYPE:
     # The quotes need to be single quotes per csp requirement.
-    script_srcs = ' '.join(
+    if csp_scripts_sha256s:
+      whitelist_hashes = ' '.join(
         ["'sha256-{}'".format(sha256) for sha256 in csp_scripts_sha256s])
+      # TODO(stephanwlee): remove `'strict dynamic'` when dynamic plugin
+      # resources can be hashed upfront.
+      script_srcs = 'strict-dynamic %s' % whitelist_hashes
+    else:
+      script_srcs = "'none'"
+
     csp_string = ';'.join([
-            "default-src 'self'",
-            # data uri used by favicon
-            "img-src 'self' data:",
-            # gstatic: used by google-chart
-            # inline styles: Polymer templates + d3 uses inline styles.
-            "style-src https://www.gstatic.com data: 'unsafe-inline'",
-            # TODO(stephanwlee): remove `'strict dynamic'` when dynamic plugin
-            # resources can be hashed upfront.
-            "script-src 'strict-dynamic' %s" % script_srcs,
-        ])
+        "default-src 'self'",
+        "base-uri 'self'",
+        "object-src 'none'",
+        # data uri used by favicon
+        "img-src 'self' data:",
+        # gstatic: used by google-chart
+        # inline styles: Polymer templates + d3 uses inline styles.
+        "style-src https://www.gstatic.com data: 'unsafe-inline'",
+        "script-src %s" % script_srcs,
+    ])
+
     headers.append(('Content-Security-Policy', csp_string))
 
   if request.method == 'HEAD':
--- a/tensorboard/backend/http_util.py
+++ b/tensorboard/backend/http_util.py
@@ -60,6 +60,8 @@ _JSON_MIMETYPES = set([
     'application/json+protobuf',
 ])
 
+# Do not support xhtml for now.
+_HTML_MIMETYPE = 'text/html'
 
 def Respond(request,
             content,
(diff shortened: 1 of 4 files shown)


tensorboard/backend/http_util.py
```diff
@@ -157,6 +163,22 @@ def Respond(request,
...
+            # data uri used by favicon
+            "img-src 'self' data:",
+            # gstatic: used by google-chart
+            # inline styles: Polymer templates + d3 uses inline styles.
+            "style-src https://www.gstatic.com data: 'unsafe-inline'",
+            # TODO(stephanwlee): remove `'strict dynamic'` when dynamic plugin
+            # resources can be hashed upfront.
+            "script-src 'strict-dynamic' %s" % script_srcs,
+        ])
+    headers.append(('Content-Security-Policy', csp_string))
```
nfelt: FWIW, I ran this through https://csp-evaluator.withgoogle.com/ and it reports a few other things (also mentioned internally):

- Missing base-uri allows the injection of base tags. They can be used to set the base URL for all relative (script) URLs to an attacker controlled domain. Can you set it to 'none' or 'self'?
- Can you restrict object-src to 'none'?

tensorboard/plugins/core/core_plugin.py
```diff
@@ -88,14 +91,47 @@ def get_plugin_apps(self):
...
+
+  def get_resource_apps(self):
+    apps = {}
+    if not self._assets_zip_provider:
+      return apps
+
+    with self._assets_zip_provider() as fp:
+      with zipfile.ZipFile(fp) as zip_:
+        # First collect the shasums that will later be used. Since hashes are
+        # in the hex, we convert it to base64 for CSP usage.
```
nfelt: Hmm, maybe we should just emit the hashes as base64 in the first place since this is the primary use case?  It's a bit less common than the hex format but would make the logic here in CorePlugin simpler.

Also, if we do it in Vulcanize we can use the nice typesafe Guava APIs and just replace

```java
String hash = Hashing.sha256().hashString(sourceContent, UTF_8).toString();
```

with

```java
// import com.google.common.io.BaseEncoding;
String hash = BaseEncoding.base64().encode(
     Hashing.sha256().hashString(sourceContent, UTF_8).asBytes());
```

Instead of having to pass strings to `codecs`.

tensorboard/plugins/core/core_plugin.py
```diff
@@ -41,6 +42,8 @@
 # for more details.
 DEFAULT_PORT = 6006
 
+SHASUM_DIR = '_shasums'
+SHASUM_FILE_SUFFIX = '.scripts_sha256'
```
nfelt: Not actually used, right now.

tensorboard/plugins/core/core_plugin.py
```diff
@@ -88,14 +91,47 @@ def get_plugin_apps(self):
...
+    return apps
+
+  def get_resource_apps(self):
+    apps = {}
+    if not self._assets_zip_provider:
+      return apps
+
+    with self._assets_zip_provider() as fp:
+      with zipfile.ZipFile(fp) as zip_:
+        # First collect the shasums that will later be used. Since hashes are
```
nfelt: Optional, but might be a little more straightforward to just loop over the HTML files we want to serve, and for each one, derive the shasum file name and read it then?  Then it's more like the shasums are a specific extra step we do for each HTML file, rather than a bunch of work we do up front no matter what.

tensorboard/plugins/core/core_plugin.py
```diff
@@ -88,14 +91,47 @@ def get_plugin_apps(self):
...
+        # in the hex, we convert it to base64 for CSP usage.
+        # TODO(stephanwlee): since index.html creates dynamic plugin and imports
+        # their entry point (es_module_path), the CSP should include shasum
+        # of those, too.
+        shasums = {}
+        for path in zip_.namelist():
+          if not path.startswith(SHASUM_DIR):
+            continue
+          html_file_name = os.path.splitext(os.path.basename(path))[0]
```
nfelt: Any reason we're calling `basename()` here?  It seems like then we have a risk of colliding if two files in different directories have the same basename?

(Same applies to where we call it down below to compute `filename`.)

tensorboard/backend/http_util.py
stephanwlee: We do make use of the `base` so I changed it to an explicit `self` (it used to be implicit `self` by `default-src`). Done!

tensorboard/plugins/core/core_plugin.py
```diff
@@ -102,20 +101,21 @@ def get_resource_apps(self):
...
-        # in the hex, we convert it to base64 for CSP usage.
+
+        # First collect the shasums that will later be used.
         # TODO(stephanwlee): since index.html creates dynamic plugin and imports
         # their entry point (es_module_path), the CSP should include shasum
         # of those, too.
         shasums = {}
         for path in zip_.namelist():
-          if not path.startswith(SHASUM_DIR):
+          if os.path.splitext(path)[1] != '.html':
```
nfelt: I think we can combine the for loops now, right?  And avoid having a `shasums` dict at all.

```python
for path in zip_.namelist():
  # Do not serve the shasum data as static files.
  if path.startswith(SHASUM_DIR):
    continue
  gzipped_asset_bytes = _gzip(zip_.read(path))
  if os.path.splitext(path)[1] == '.html':
    checksum_path = os.path.join(SHASUM_DIR, path + SHASUM_FILE_SUFFIX)
    if checksum_path in zip_.namelist():
      shasums = zip_.read(checksum_path).splitlines(False)
    else:
      shasums = None
    wsgi_app = functools.partial(self._serve_html, shasums, gzipped_asset_bytes)
  else:
    wsgi_app = functools.partial(self._serve_asset, path, gzipped_asset_bytes)
  apps['/' + path] = wsgi_app

stephanwlee: done!

PR #1117, opened 2018-04-10
Extend the profile plugin to take per host profile traces

qiuminxu:
* Profile plugin is extended to take per host traces
* Added a host dropdown menu in profile dashboard
* Moved the _request_manager from each tool to tf_profile_dashboard, so that individual tool can be directly used by internal tools
* Add per host test in profile_plugin_test

PR #2733, opened 2019-10-06
application: extract path prefix logic to middleware

wchargin:
Summary:
This way, we can reason about and test it in isolation as we add more
middleware to the application.

Note that when using `--path_prefix`, loading the main TensorBoard page
without a trailing slash on the URL has long been broken (#1117). This
commit does not fix that, but changes the exact failure mode: rather
than 404ing, we now load a broken TensorBoard (relative URLs don’t
resolve). A follow-up commit will fix this properly.

Test Plan:
Verify that TensorBoard still works fully, with `--path_prefix` and
without it.

wchargin-branch: path-prefix-middleware

### DIFF
--- a/tensorboard/backend/path_prefix.py
+++ b/tensorboard/backend/path_prefix.py
@@ -1,0 +1,65 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Internal path prefix support for TensorBoard.
+
+Using a path prefix of `/foo/bar` enables TensorBoard to serve from
+`http://localhost:6006/foo/bar/` rather than `http://localhost:6006/`.
+See the `--path_prefix` flag docs for more details.
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from tensorboard import errors
+
+
+class PathPrefixMiddleware(object):
+  """WSGI middleware for path prefixes.
+
+  All requests to this middleware must begin with the specified path
+  prefix (otherwise, a 404 will be returned immediately). Requests will
+  be forwarded to the underlying application with the path prefix
+  stripped and appended to `SCRIPT_NAME` (see the WSGI spec, PEP 3333,
+  for details).
+  """
+
+  def __init__(self, application, path_prefix):
+    """Initializes this middleware.
+
+    Args:
+      application: The WSGI application to wrap (see PEP 3333).
+      path_prefix: A string path prefix to be stripped from incoming
+        requests. If empty, this middleware is a no-op. If non-empty,
+        the path prefix must start with a slash and not end with one
+        (e.g., "/tensorboard").
+    """
+    if path_prefix.endswith("/"):
+      raise ValueError("Path prefix must not end with slash: %r" % path_prefix)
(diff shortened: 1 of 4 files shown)


tensorboard/backend/path_prefix_test.py
```diff
@@ -0,0 +1,112 @@
...
+      response = server.get("/foo/bar")
+      self._assert_ok(server.get("/foo/bar"), path="/foo/bar", script="")
+
+  def test_nonempty_path_prefix(self):
+    app = path_prefix.PathPrefixMiddleware(self._echo_app, "/pfx")
+    server = werkzeug.test.Client(app, werkzeug.BaseResponse)
+
+    with self.subTest("at root"):
+      response = server.get("/pfx")
+      self._assert_ok(response, path="", script="/pfx")
```
stephanwlee: no action required; I simply dunno: how does our server respond when the path is "" vs. "/"?

wchargin: Good question. The answer is that Werkzeug normalizes that out, so our
app (which uses Werkzeug for routing) only sees `request.path == "/"`:

```python
# main.py
import werkzeug

def bare_app(environ, start_response):
  start_response("200 OK", [("Content-Type", "text/plain")])
  return [repr(environ.get("PATH_INFO")).encode("utf-8")]

@werkzeug.Request.application
def werkzeug_app(request):
  return werkzeug.Response(repr(request.path), content_type="text/plain")

start_response = lambda status, headers: None
request_empty = {"REQUEST_METHOD": "GET", "PATH_INFO": ""}
request_root = {"REQUEST_METHOD": "GET", "PATH_INFO": "/"}

print("With base WSGI:")
print(list(bare_app(request_empty, start_response)))
print(list(bare_app(request_root, start_response)))

print("With Werkzeug:")
print(list(werkzeug_app(request_empty, start_response)))
print(list(werkzeug_app(request_root, start_response)))
```

```
$ python main.py
With base WSGI:
[b"''"]
[b"'/'"]
With Werkzeug:
[b"'/'"]
[b"'/'"]
```

tensorboard/backend/path_prefix.py
```diff
@@ -0,0 +1,65 @@
...
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from tensorboard import errors
+
+
+class PathPrefixMiddleware(object):
```
nfelt: So I had been thinking we could generalize this as taking a dict mapping prefixes to WSGI apps.  This unblocks using this for the plugin routing as well as described in https://github.com/tensorflow/tensorboard/issues/2573.

Feel free to leave this PR as-is, but just FYI, I think it will make sense to fold this into a more generic middleware sooner or later.

tensorboard/backend/path_prefix.py
```diff
@@ -0,0 +1,65 @@
...
+
+  def __init__(self, application, path_prefix):
+    """Initializes this middleware.
+
+    Args:
+      application: The WSGI application to wrap (see PEP 3333).
+      path_prefix: A string path prefix to be stripped from incoming
+        requests. If empty, this middleware is a no-op. If non-empty,
+        the path prefix must start with a slash and not end with one
+        (e.g., "/tensorboard").
```
nfelt: IMO the semantics are more obvious if we require that the path prefix end in a slash, since then we don't necessarily even need to support the redirect, and there's no confusion about what the canonical URL will be.

WDYT?  I think we could fairly consider that not to be a breaking change given that using the prefix without the trailing slash is broken today due to the lack of redirect.

tensorboard/backend/path_prefix.py
wchargin: Makes sense to me.

tensorboard/backend/path_prefix.py
wchargin: I’m fine with making that change at the user level (i.e., a requirement
on the flag value). As an implementation detail, I find it useful to
represent path components like this with a leading slash and no trailing
slash, because then they concatenate properly. This is also what
Werkzeug does for the `.script_root` property (they don’t expose
`SCRIPT_NAME` correctly), and is the convention that `application.py`
uses for constants like `DATA_PREFIX = "/data"`.

Given that this PR doesn’t actually make any user-facing changes modulo
existing undefined behavior, I’m comfortable deferring the tightening of
the flag domain. Let me know if you object.

PR #2736, opened 2019-10-06
frontend: use experiment ID from path, not query

wchargin:
Summary:
This commit changes the frontend experiment ID plumbing from using the
query string to using a path component: experiment URLs are now of the
form `http://localhost:6006/experiment/123/`. This supersedes the
short-term changes from #2580, #2613, and #2627.

Test Plan:
Apply the below patch to the multiplexer data provider, then launch
TensorBoard with `--generic_data=true` and verify that the correct
experiment ID is printed for each call, at `/` and at `/experiment/123`.
Verify also that no request has an `experiment` query parameter. Patch:

```diff
diff --git a/tensorboard/backend/event_processing/data_provider.py b/tensorboard/backend/event_processing/data_provider.py
index 0b354aa8..1c00b930 100644
--- a/tensorboard/backend/event_processing/data_provider.py
+++ b/tensorboard/backend/event_processing/data_provider.py
@@ -59,2 +59,3 @@ class MultiplexerDataProvider(provider.DataProvider):
   def data_location(self, experiment_id):
+    logger.warn("data_location(%r)", experiment_id)
     del experiment_id   # ignored
@@ -63,2 +64,3 @@ class MultiplexerDataProvider(provider.DataProvider):
   def list_runs(self, experiment_id):
+    logger.warn("list_runs(%r)", experiment_id)
     del experiment_id  # ignored for now
@@ -74,2 +76,3 @@ class MultiplexerDataProvider(provider.DataProvider):
   def list_scalars(self, experiment_id, plugin_name, run_tag_filter=None):
+    logger.warn("list_scalars(%r, ...)", experiment_id)
     del experiment_id  # ignored for now
@@ -105,2 +108,3 @@ class MultiplexerDataProvider(provider.DataProvider):
   ):
+    logger.warn("read_scalars(%r, ...)", experiment_id)
     # TODO(@wchargin): Downsampling not implemented, as the multiplexer
```

wchargin-branch: eid-frontend


wchargin:
I’m not sure that I fully understand the question. The frontend already
assumes that it’s served from `/`, because otherwise the relative URLs
will resolve incorrectly. After this change, we still support loading
from `/` (with empty experiment ID) and additionally support loading
from `/experiment/EID/`. Does this answer your question?

If you’re asking about caching of the large HTML blob of the document
itself: It’s true that this won’t have a shared cache across two
experiments, but that was also true for the query parameter version. In
either case, we can improve the caching by loading the bundle (or parts
of it) dynamically on the frontend.

> When it increasingly is aware of the experiment, I think it is
> inevitable that we will introduce `experiment` param back to the
> router.

I agree that this is plausible, especially because you can’t just route
to `../456/` to change the experiment ID, because when you’re at `/`
instead of `/experiment/123/` this would do the wrong thing.

PR #2737, opened 2019-10-06
backend: fix `--path_prefix` validation

wchargin:
Summary:
Specifying a path prefix without leading slash has always been
disallowed by the docs, but in TensorBoard 2.0.0 this just hits an
inscrutable internal error when starting the server:

```
ValueError: invalid literal for int() with base 10: '6006my_path_prefix'
```

As of #2733, this is at least a correctly reported error, but it still
has a stack trace. This commit makes it a normal, fast-failing usage
error.

The logic to strip trailing slashes was also duplicated, so `/foo` and
`/foo/` and `/foo//` were all the same but differed from `/foo///`. This
commit removes one copy of the logic and makes the remaining copy strip
all trailing slashes instead of just one.

Test Plan:
Run with `--path_prefix /foo`, and note that TensorBoard exits with a
single-line error. Run with `/foo`, `/foo/`, `/foo//`, and `/foo///`,
and note that all four now have the same behavior.

wchargin-branch: path-prefix-validation


tensorboard/plugins/core/core_plugin.py
```diff
@@ -568,8 +568,10 @@ def fix_flags(self, flags):
     elif flags.host is not None and flags.bind_all:
       raise FlagsError('Must not specify both --host and --bind_all.')
 
-    if flags.path_prefix.endswith('/'):
-      flags.path_prefix = flags.path_prefix[:-1]
+    flags.path_prefix = flags.path_prefix.rstrip('/')
+    if flags.path_prefix and not flags.path_prefix.startswith('/'):
+      raise FlagsError(
+          'Path prefix must start with slash, but got: %r.' % flags.path_prefix)
```
stephanwlee: unless the test is already present, consider adding a test on this.

wchargin: Thanks for the suggestion. Done.

PR #2744, opened 2019-10-08
Allow set_example to accept JSON

jameswex:

If examples are provided to WitConfigBuilder as JSON, then the set_examples method on WitWidget for changing examples should also accept JSON. Currently set_examples only accepts Example protos.

Move conversion of JSON to Example proto into WitWidget so the same code can be used in constructor and override example setting.

Ran notebooks with Example protos and JSON dicts and JSON lists as inputs, and in all cases ensured correct parsing of examples and correct overriding of examples through set_examples.

### DIFF
--- a/tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
+++ b/tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
@@ -22,6 +22,7 @@ from IPython import display
 from google.protobuf import json_format
 from numbers import Number
 from six import ensure_str
+from six import integer_types
 from tensorboard.plugins.interactive_inference.utils import inference_utils
 
 # Constants used in mutant inference generation.
@@ -84,10 +85,10 @@ class WitWidgetBase(object):
     if 'compare_adjust_attribution' in copied_config:
       del copied_config['compare_adjust_attribution']
 
+    self.config = copied_config
     self.set_examples(config['examples'])
     del copied_config['examples']
 
-    self.config = copied_config
 
     # If using AI Platform for prediction, set the correct custom prediction
     # functions.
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
```diff
@@ -122,7 +123,11 @@ def set_examples(self, examples):
     builder during construction. This method can change which examples WIT
     displays.
     """
-    self.examples = [json_format.MessageToJson(ex) for ex in examples]
+    if self.config.get('uses_json_input'):
+      tf_examples = self._json_to_tf_examples(examples)
```
tolga-b: is there any reason we use a separate variable tf_examples for casting to tf_examples?

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
```diff
@@ -318,6 +323,37 @@ def _json_from_tf_examples(self, tf_examples):
...
     return json_exs
 
+  def _json_to_tf_examples(self, examples):
+    def add_single_feature(feat, value, ex):
+      if isinstance(value, integer_types):
+        ex.features.feature[feat].int64_list.value.append(value)
+      elif isinstance(value, Number):
+        ex.features.feature[feat].float_list.value.append(value)
+      else:
+        ex.features.feature[feat].bytes_list.value.append(value.encode('utf-8'))
```
tolga-b: just double checking if we need py2 py3 protection for value.encode here.

jameswex: verified this works in both py2 and py3

PR #2746, opened 2019-10-08
vulcanization: improvements

stephanwlee:
**vulc: graceful handling of absolute URL**
Recently with #2719, we introduced script tag hashing with the
Vulcanization. This did not correctly address the case when a script tag
refers to a remote resource or a case when webpath could not be resolved
to a webfiles passed to Vulcanization.

tested on more complex set up where we use `vulcanize-noinline` and an absolute URL.

**vulc: break scripts when there are src in between**
Order of script declarations cannot be changed for correctness sake.
Previously, we were combining all script tags into one while leaving the
ones with `src` in the same place. This caused the correctness issue so
we now combine scripts up to occurrences of src-full script tag.

WANT_LGTM=any
merge plan=rebase

### DIFF
--- a/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
+++ b/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
@@ -50,8 +50,9 @@ import io.bazel.rules.closure.Webpath;
 import io.bazel.rules.closure.webfiles.BuildInfo.Webfiles;
 import io.bazel.rules.closure.webfiles.BuildInfo.WebfilesSource;
 import java.io.ByteArrayInputStream;
-import java.io.IOException;
+import java.io.FileNotFoundException;
 import java.io.InputStream;
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -747,11 +746,26 @@ private static void combineScriptElements(Document document) {
...
+                  + " refers to a remote resource. Please add it to CSP manually. Detail: "
+                  + script.outerHtml());
+          continue;
+        } else if (!webfiles.containsKey(webpathSrc)) {
+          System.err.println(
+              "WARNING: source content of "
+                  + webpathSrc
+                  + " is unknown to the Vulcanization. Please add it to CSP manually. Detail: "
+                  + script.outerHtml());
+          continue;
```
stephanwlee: to reviewer: I can hard fail in this situation; there should never be a case when a webpath does not resolve to a webfiles. Practically, `tf_web_library` and other closure tooling would catch such case even before/after Vulcanization though.

nfelt: IMO maybe a bit better to hard fail, yeah.

I wonder if there's a way we can make the other case (remote resource) also more obviously enforceable?  I worry about people not seeing these warnings.  For example, maybe we could make that check also hard-fail unless there's a special attribute on the `<script>` tag like `manual-csp` or something?

stephanwlee:
> I worry about people not seeing these warnings. For example, maybe we could make that check also hard-fail unless there's a special attribute on the <script> tag like manual-csp or something?

I, too, am worry about it and that can be a good follow up item. I thought failing CSP would be a good enough but I can see why that would be hard to catch.

PR #2750, opened 2019-10-09
cleanup: import werkzeug test

stephanwlee:
werkzeug internally do not have `test` module.

This unblocks the sync.

### DIFF
--- a/tensorboard/backend/path_prefix_test.py
+++ b/tensorboard/backend/path_prefix_test.py
@@ -21,6 +21,7 @@ from __future__ import print_function
 import json
 
 import werkzeug
+from werkzeug import test as werkzeug_test
 
 from tensorboard import errors
 from tensorboard import test as tb_test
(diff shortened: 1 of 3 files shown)


tensorboard/backend/path_prefix_test.py
```diff
@@ -62,7 +63,7 @@ def test_bad_path_prefix_with_trailing_slash(self):
 
   def test_empty_path_prefix(self):
     app = path_prefix.PathPrefixMiddleware(self._echo_app, "")
-    server = werkzeug.test.Client(app, werkzeug.BaseResponse)
+    server = werkzeug_test.Client(app, werkzeug.BaseResponse)
```
wchargin: Is `from werkzeug import wrappers; wrappers.BaseReponse` not needed?

stephanwlee: BaseResponse is aliased correctly. We may want to change this when we upgrade to werkzeug 1.0.

wchargin: Interesting; thanks for investigating.

PR #2751, opened 2019-10-09
Expose graph plugin name

davidsoergel:
This allows referring to the graph plugin name from elsewhere, as we already do with other plugins.

    from tensorboard.plugins.graph import metadata as graph_metadata
    ...
    if(foobar == graph_metadata.PLUGIN_NAME): ...


tensorboard/plugins/graph/metadata.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+# This name is used as the plugin prefix route and to identify this plugin
+# generally.
+# Note however that different 'plugin names' are used in the context of
+# graph Summaries.
+# See `graphs_plugin.py` for details.
+PLUGIN_NAME = 'graphs'
```
wchargin: nit: missing newline character at EOF

davidsoergel: oops thanks!

PR #2755, opened 2019-10-10
ROC curves for multi-class (What-If Tool)

grovina:

Show ROC curves for multi-class models.

For each class, we can plot a ROC curve considering the class in question as the positive one and all the others as negatives (what is called a *binarized* version of the problem).
To achieve this, we iterate through all examples, populating an object with the following structure:
```> model > slice > label > threshold > classification stats```
Where the classification data contain the number and rates of true and false positives and negatives. This is done for every value of the threshold (every 1% between 0 to 100%), for every class label, for each feature slice (either each value for categorical features or each interval for numeric features), for every model.
The overall case (not sliced by any feature) is trivially treated as a single slice.

When displaying the ROC curves, we end up with the same case as the ordinary ROC curves from binary classification problems, once for each slice.

* Screenshots of UI changes

The overall case:
Screenshot from 2019-10-10 14-34-36

The sliced feature case:
Screenshot from 2019-10-10 14-34-08

Open performance tab in the iris demo, and check that:
   - the correct number of ROC curves appear (same number as number of classes in the problem)
   - the values and shape of the ROC curve are coherent
   - the curve respects the data

Repeat for sliced features, and also for different numbers of buckets

We could think of averaged ROC curves (like https://scikit-learn.org/stable/auto_examples/model_selection/plot_roc.html). I personally think they are a bit harder to interpret.

Another possibility is merging all curves into a single plot. Although more compact, it wouldn't be as clear that each curve refers to a distinct version of the problem, and it could also get a bit too crowded with many classes or multiple models.

### DIFF
--- a/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
+++ b/tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
@@ -2319,7 +2319,7 @@ limitations under the License.
                           <paper-radio-button
                             name="demoparity"
                             class="optimization-radio"
-                            disabled$="[[!shouldShowFeaturePrCharts_(selectedLabelFeature, selectedBreakdownFeature, inferences)]]"
+                            disabled$="[[!shouldShowFeaturePerfCharts_(selectedLabelFeature, selectedBreakdownFeature, inferences)]]"
                             >Demographic parity
                             <paper-icon-button
                               icon="info-outline"
@@ -2348,7 +2348,7 @@ limitations under the License.
                           <paper-radio-button
                             name="equalopp"
                             class="optimization-radio"
-                            disabled$="[[!shouldShowFeaturePrCharts_(selectedLabelFeature, selectedBreakdownFeature, inferences)]]"
+                            disabled$="[[!shouldShowFeaturePerfCharts_(selectedLabelFeature, selectedBreakdownFeature, inferences)]]"
                             >Equal opportunity
                             <paper-icon-button
                               icon="info-outline"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by grovina** [comment]
cc @jameswex and @tolga-b

File: tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html [hunk 1]
```diff
@@ -4913,6 +5074,53 @@ <h2>Show similarity to selected datapoint</h2>
...
+                      this.getRocChartLabelId(
+                        labelInd,
+                        isSliced ? sliceInd : ''
+                      )
+                  ),
+                  plotStats,
+                  plotThresholds,
+                  regenInferenceStats,
+                  true
+                );
```
jameswex: should really do ROC and PR curves for each class just like with binary

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -4854,7 +5015,7 @@ <h2>Show similarity to selected datapoint</h2>
           this.featureValueThresholds = [];
           this.featureValueThresholds = this.sortFeatureValues(tempArray);
 
-          // ROC curves should only exist for the binary case
+          // ROC curves for the binary case
```
jameswex: update comment since also does PR curves

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -5864,6 +6072,14 @@ <h2>Show similarity to selected datapoint</h2>
           return 'rocchart' + index;
         },
 
+        getRocChartLabelId: function(label, index) {
+          return this.getRocChartId(index) + '-' + label;
+        },
+
+        getLabelVocab: function(index) {
```
jameswex: call this getLabel or getLabelForIndex for clarity

jameswex:
thanks @grovina , just the one comment about padding now

jameswex:
also make sure to run lint

grovina:
Adjusted some CSS and fixed lint. Here screenshots for:

* the multi-class case:
Screenshot from 2019-10-11 17-38-12

* the binary case:
Screenshot from 2019-10-11 17-37-53

PR #2756, opened 2019-10-10
csp: make it globally configurable

stephanwlee:
This is a temporary measure to unblock a high priority item. In the
ideal world, we would want some kind of global CSP configuration that
would also know about all static contents in the application.

### DIFF
--- a/tensorboard/backend/http_util_test.py
+++ b/tensorboard/backend/http_util_test.py
@@ -36,6 +36,7 @@ class RespondTest(tb_test.TestCase):
   def setUp(self):
     # http_util is stateful. Unset the state.
     http_util.DO_NOT_USE_CSP_SCRIPT_DOMAINS_WHITELIST = []
+    http_util.DO_NOT_USE_CSP_SCRIPT_HASHES_STRICT_DYNAMIC = True
 
   def testHelloWorld(self):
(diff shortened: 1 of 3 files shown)


tensorboard/backend/http_util_test.py
```diff
@@ -205,6 +206,18 @@ def testCsp_noHash(self):
     )
     self.assertEqual(r.headers.get('Content-Security-Policy'), expected_csp)
 
+  def testCsp_disableStrictDynamic(self):
+    http_util.DO_NOT_USE_CSP_SCRIPT_HASHES_STRICT_DYNAMIC = False
```
nfelt: Prefer mock.object.patch() for this since it can be auto-restored after the test.  E.g. you can just use it as as decorator on the test method.

Note we need a conditional import for it:
```
try:
  # python version >= 3.3
  from unittest import mock  # pylint: disable=g-import-not-at-top
except ImportError:
  import mock  # pylint: disable=g-import-not-at-top,unused-import
```

stephanwlee: It was unclear how I would mock out module global like the ones I declared so I converted them to methods. Please advise on how I can better improve this code :)

PR #2768, opened 2019-10-13
cleared numpy depreciation warnings

LordGhostX:
fixed numpy compatibility issues with tensorflow, no longer displays depreciation messages when importing tensorflow on systems with np version > 1.14.5

Tested with numpy versions 1.14.5, 1.16.5, 1.17.0

* Screenshots of UI changes

### DIFF
--- a/tensorboard/compat/tensorflow_stub/dtypes.py
+++ b/tensorboard/compat/tensorflow_stub/dtypes.py
@@ -538,16 +538,16 @@ _STRING_TO_TF["double_ref"] = float64_ref
 # quantized types.
 # TODO(mrry,keveman): Investigate Numpy type registration to replace this
 # hard-coding of names.
-_np_qint8 = np.dtype([("qint8", np.int8)])
-_np_quint8 = np.dtype([("quint8", np.uint8)])
-_np_qint16 = np.dtype([("qint16", np.int16)])
-_np_quint16 = np.dtype([("quint16", np.uint16)])
-_np_qint32 = np.dtype([("qint32", np.int32)])
+_np_qint8 = np.dtype([("qint8", "i1")])
+_np_quint8 = np.dtype([("quint8", "i1")])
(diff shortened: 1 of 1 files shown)


wchargin:
These seem to be enforcing little-endian byte ordering, but my
understanding\* is that numpy defaults to the native byte ordering,
which makes sense for performance. So I suspect that these definitions
aren’t correct on big-endian systems, and I’m especially hesitant given
that this change would introduce an inconsistency with [TensorFlow’s
dtypes module][tf]. As the primary purpose of this module is to be
compatible with TensorFlow, it might be best to integrate these changes
upstream first.

Can you elaborate about how you’re running into these deprecation
errors? I don’t see them:

```
$ virtualenv -q -p python3.6 ./ve
$ . ./ve/bin/activate
(ve) $ pip install -q numpy==1.17.0 tensorboard==2.0.0
(ve) $ python
Python 3.6.8 (default, Jan  3 2019, 03:42:36) 
[GCC 8.2.0] on linux
>>> from tensorboard.compat import tf
>>> tf.dtypes.quint16.as_numpy_dtype
dtype([('quint16', '<u2')])
>>> type(_)
<class 'numpy.dtype'>
>>> # no errors or warnings
```

\* See numpy `arraytypes.c.src`, [lines 4230 and 4261][src-1] and below
at [lines 4341 and 4350][src-2].

[src-1]: https://github.com/numpy/numpy/blob/e2c75d94ee1ea4dd0eb0ab0a394f984c968d9237/numpy/core/src/multiarray/arraytypes.c.src#L4229-L4261
[src-2]: https://github.com/numpy/numpy/blob/e2c75d94ee1ea4dd0eb0ab0a394f984c968d9237/numpy/core/src/multiarray/arraytypes.c.src#L4341-L4350
[tf]: https://github.com/tensorflow/tensorflow/blob/1d2ae0c8ca0e2899ce03b93a0ce77c42a5ca6309/tensorflow/python/framework/dtypes.py#L520-L524

PR #2773, opened 2019-10-14
Consume java_import_external from bazel_tools instead of rules_closure

Yannic:
The version in rules_closure is being removed here: https://github.com/bazelbuild/rules_closure/pull/239


Yannic:
//cc @davido

davido:
@Yannic Thanks!

PR #2774, opened 2019-10-14
core: allow frame_ancestors configuration from CLI

stephanwlee:
Now, with --frame_ancestors, one can change the frame-ancestors
Content-Security-Policy to one of four configurations. The
configurations are:

- none: does not allow TensorBoard to be iframed
- colab: CSP for colab only
- notebook: CSP for notebook environments like Jupyter
- unsafe: allow any sites from embedding TensorBoard

The default for now is `unsafe` since that is current behavior as of
TB 2.0.

Do note that notebook one is the same as unsafe for now because in
context of, at least, Jupyter, localhost:[port] document loads
TensorBoard in an iframe that is running at localhost:6006. Because for
CSP keyword like "self" has to match both host-uri and the port, it does
not work for us. Moreover, it is, depending on Jupyter configuration,
naturally to run one on "ip" which would make the address not
guess-able. Until we have a better solution, we will have the "unsafe"
behavior for the notebook configuration.

Inspirations from #2397.

### DIFF
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -151,6 +151,7 @@ def start(args_string):
     else:
       handle.update(IPython.display.Pretty(message))
 
+  args_string='%s %s' % (_get_context_default_args(), args_string)
   parsed_args = shlex.split(args_string, comments=True, posix=True)
   start_result = manager.start(parsed_args)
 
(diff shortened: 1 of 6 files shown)


tensorboard/notebook.py
```diff
@@ -151,6 +151,7 @@ def print_or_update(message):
     else:
       handle.update(IPython.display.Pretty(message))
 
+  args_string='%s %s' % (_get_context_default_args(), args_string)
```
wchargin: I’m slightly hesitant to add this because (a) it breaks the property
that “`%tensorboard ARGS` is the same as running `tensorboard ARGS` at a
shell”, which greatly simplifies the mental model, and (b) it makes it
impossible to reuse existing TensorBoard instances started outside of
the notebook mode (and for a non-obvious reason: see point (a)).
Similarly, `notebook.display(port=1234)` won’t be fixed by this change,
a further inconsistency.

I’ll think about this a bit and see if I can convince myself that it’s
fine. :-)

stephanwlee: Ah, I totally forgot class of people who would be doing `notebook.display`. I will, too, have to think about this.

When I originally wrote this, I conceived it as context aware defaults but I can see both (a) and (b) tricky. I will sleep on it on more night :)

PR #2775, opened 2019-10-14
csp: fix bugs and properly treat projector

stephanwlee:
With enabling CSP on TensorBoard, we have broken projector in few ways.
Because we enforce CSP for all text/html response, we need to have
proper sha256s for all scripts including ones for the projector. Several
fixes are:

1. strict-dynamic -> 'strict-dynamic'
2. 'frame-src' is now set for `iframe src="..."`
3. script hashes are piped in for projector
4. `unsafe-eval` because of numericjs [1].

[1]:
https://github.com/sloisel/numeric/blob/656fa1254be540f428710738ca9c1539625777f1/src/numeric.js#L576-L585

Confession: tested projector before applying stricter rules on _all_ `text/html` before :\. Tested on projector now.

### DIFF
--- a/tensorboard/backend/http_util.py
+++ b/tensorboard/backend/http_util.py
@@ -44,6 +44,8 @@ _CSP_IMG_DOMAINS_WHITELIST = []
 _CSP_SCRIPT_DOMAINS_WHITELIST = []
 _CSP_SCRIPT_HASHES_STRICT_DYNAMIC = True
 _CSP_SCRIPT_SELF = False
+# numericjs (via projector) uses unsafe-eval :(.
+_CSP_SCRIPT_UNSAFE_EVAL = True
 _CSP_STYLE_DOMAINS_WHITELIST = []
 
 _EXTRACT_MIMETYPE_PATTERN = re.compile(r'^[^;\s]*')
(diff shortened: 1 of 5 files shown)


tensorboard/backend/http_util.py
```diff
@@ -44,6 +44,8 @@
 _CSP_SCRIPT_DOMAINS_WHITELIST = []
 _CSP_SCRIPT_HASHES_STRICT_DYNAMIC = True
 _CSP_SCRIPT_SELF = False
+# numericjs (via projector) uses unsafe-eval :(.
+_CSP_SCRIPT_UNSAFE_EVAL = True
```
wchargin: Does this need to _default_ to true, or can the projector plugin set it
to true for its HTML only so that we don’t also allow it for the rest of
TensorBoard?

stephanwlee: No, but ideally, we want nearly all of these CSP fields to be configurable so I think it makes a lot more sense to think of the new API holistically. 

I am thinking of expanding asset_provider that plugin authors can implement as part of the plugin API but I am not yet sure how CSP part should be configured; do we take shasum as input or does TensorBoard core parse the document to shasum the script tags? 

Since this PR is not as time critical as I originally thought, I will wait for your reply.

tensorboard/backend/http_util.py
```diff
@@ -188,22 +190,27 @@ def Respond(request,
...
-        "default-src 'none'",
-        "base-uri 'self'",
-        "connect-src 'self'",
+        "default-src 'self'",
         'font-src %s' % _create_csp_string(
             "'self'",
             *_CSP_FONT_DOMAINS_WHITELIST
         ),
+        # Dynamic plugins are rendered inside an iframe.
+        "frame-src 'self'",
```
davidsoergel: not needed anymore?  (Or, does frame-src not follow the default?)

davidsoergel: Gotcha, thanks!

PR #2776, opened 2019-10-14
util: utility for getting resource file

stephanwlee:
TensorBoard may want to use a project like google/subpar to create a
self-contained python binary and we need to make sure resources inside
the container to be accessible. This utility creates an abstraction
point for such instances.


stephanwlee:
Actually, I think I know a better way. I will close this :)

MarkAlive:
hi @stephanwlee, could you explain what the "better way" is? Thx!

PR #2785, opened 2019-10-16
Fix sliced intersection (What-If Tool)

grovina:

This bug:
Screenshot 2019-10-16 at 21 53 29

The feature to update `featureValueThresholdsIndexMap` was considering
only the first sliced feature, when it should be using the combined
string.

I also tweaked the function that combines these two values to do strict
comparisons for null and empty strings, because it was considering the
value `0 == ''`, ignoring these values.

* Screenshots of UI changes

Screenshot 2019-10-16 at 21 53 09

Slice by two features in the Performance tab and checking that we don't have those `NaN`s.

N/A


grovina:
cc @jameswex

PR #2787, opened 2019-10-17
Add ml service name and version for CAIP

jameswex:

WitWidget cloud users need a way to override the cloud service and service version to send requests to.

Added parameters for service name and version to cloud setup functions, defaulting those parameters to the correct default server.

Ran against default server models and a non-default server model.


jameswex:
To fix #2787

tolga-b:
Hi James, to fix link points back to this issue, did you mean to put something else?
Thanks

PR #2789, opened 2019-10-17
plugin: load plugin bootstrap from backend

stephanwlee:
background: to load the plugin module, we created an iframe whose origin
was ambiguous (to us, TB developer). It turns out, the origin of the
src-less iframe are "about:blank" even with the `<base>` and correct
`document.domain` and, as a result, service worker did not work
correctly inside. We now create the iframe content from our backend and
serve it from our origin.

changes:
1. new `/plugin.html` endpoint that requires a plugin name that matches
  ones registered on the backend.
2. no more 'strict-dynamic' as we no longer create unknown script! We
  need to trust scripts from `self` though.


tensorboard/backend/application.py
```diff
@@ -337,6 +342,55 @@ def _create_wsgi_app(self):
...
+
+    if not plugins:
+      return http_util.Respond(
+          request, 'Plugin not found', 'text/plain', code=404)
+
+    if len(plugins) > 1:
+      # Technically is not possible as plugin names are unique and is checked
+      # by the check on __init__.
+      return http_util.Respond(
+          request, 'Invariant error', 'text/plain', code=400)
```
wchargin: Should be code=500? You could also just raise an `AssertionError` here,
which will properly communicate a 500 as well as logging this exception,
which we probably want.

tensorboard/backend/application.py
```diff
@@ -337,6 +342,55 @@ def _create_wsgi_app(self):
...
+    if not module_url:
+      return http_util.Respond(
+          request, 'Plugin is not module loadable', 'text/plain', code=400)
+
+    # Prevent a malicious plugin from using an absolute path that would make this trusted
+    # script load an arbitrary script from a remote source.
+    module_path = urlparse.urlparse(module_url).path
+    script_content = 'import(".{module}").then((m) => void m.render());'.format(
+        module = module_path)
+    digest = hashlib.sha256(script_content.encode('utf8')).digest()
```
wchargin: The proper name of the encoding is `utf-8`, with hyphen. Python has an
alias, but still better to promote the correct name, as not all contexts
do: e.g., `charset=utf8` in a `Content-Type` header can be a security
vulnerability.

tensorboard/backend/application.py
```diff
@@ -337,6 +342,55 @@ def _create_wsgi_app(self):
...
+      return http_util.Respond(
+          request, 'Invariant error', 'text/plain', code=400)
+
+    plugin = plugins[0]
+    module_url = plugin.frontend_metadata().es_module_path
+    if not module_url:
+      return http_util.Respond(
+          request, 'Plugin is not module loadable', 'text/plain', code=400)
+
+    # Prevent a malicious plugin from using an absolute path that would make this trusted
```
wchargin: This is fine and good, but I’d be interested to know what attack it’s
preventing. Plugins can already put arbitrary contents into their
script, so they could just fetch the remote script in the Python code
and serve it inline. What would they gain by being able to directly load
from another domain?

Also, just a note: this doesn’t prevent directory traversal (`..`)
attacks.

tensorboard/backend/application.py
```diff
@@ -337,6 +342,55 @@ def _create_wsgi_app(self):
...
+    # script load an arbitrary script from a remote source.
+    module_path = urlparse.urlparse(module_url).path
+    script_content = 'import(".{module}").then((m) => void m.render());'.format(
+        module = module_path)
```
wchargin: nit: Python style is `module=module_path` (no space). Likewise below.

tensorboard/backend/application.py
```diff
@@ -337,6 +342,55 @@ def _create_wsgi_app(self):
...
+  def _serve_plugin_entry(self, request):
+    """Serves a HTML for iframed plugin entry point.
+
+    Args:
+      request: The werkzeug.Request object.
+
+    Returns:
+      A werkzeug.Response object.
+    """
+    name = request.args.get('name')
```
wchargin: Curious whether you considered `/data/plugin_entry/PLUGIN` or similar
rather than a query parameter? No strong preference.

tensorboard/backend/application.py
```diff
@@ -337,6 +342,55 @@ def _create_wsgi_app(self):
...
+    # Prevent a malicious plugin from using an absolute path that would make this trusted
+    # script load an arbitrary script from a remote source.
+    module_path = urlparse.urlparse(module_url).path
+    script_content = 'import(".{module}").then((m) => void m.render());'.format(
+        module = module_path)
+    digest = hashlib.sha256(script_content.encode('utf8')).digest()
+    script_sha = base64.b64encode(digest).decode("ascii")
+
+    html = textwrap.dedent("""
+      <head><base href="data/plugin/{name}/" /></head>
```
wchargin: Add `<!DOCTYPE html>`?

tensorboard/backend/application.py
```diff
@@ -337,6 +342,55 @@ def _create_wsgi_app(self):
...
+    plugin = plugins[0]
+    module_url = plugin.frontend_metadata().es_module_path
+    if not module_url:
+      return http_util.Respond(
+          request, 'Plugin is not module loadable', 'text/plain', code=400)
+
+    # Prevent a malicious plugin from using an absolute path that would make this trusted
+    # script load an arbitrary script from a remote source.
+    module_path = urlparse.urlparse(module_url).path
+    script_content = 'import(".{module}").then((m) => void m.render());'.format(
```
wchargin: This should be JSON-encoded for security, right?

```python
>>> from urllib import parse
>>> module_url = '/index.js"); alert("hmm"); //'
>>> module_path = parse.urlparse(module_url).path
>>> print('import(".{module}").then(...);'.format(module=module_path))
import("./index.js"); alert("hmm"); //").then(...);
```

Consider instead:

```python
module_json = json.dumps("." + module_path)
script_content = 'import({}).then(...);'.format(module_json)
```

tensorboard/backend/application.py
```diff
@@ -262,6 +266,7 @@ def __init__(self, plugins, path_prefix=''):
         # obviate the need for the frontend to determine which plugins are
         # active.
         DATA_PREFIX + PLUGINS_LISTING_ROUTE: self._serve_plugins_listing,
+        PLUGIN_ENTRY_ROUTE: self._serve_plugin_entry,
```
wchargin: Just noticed: this should be under `/data/`, like all our other routes.

tensorboard/backend/application.py
stephanwlee: As discussed, since we can change this transparently, I will keep it as is. If we have better support in the future for url parameters like `/foo/bar/:baz`, we will definitely consider doing that.

tensorboard/backend/application.py
```diff
@@ -337,6 +342,56 @@ def _create_wsgi_app(self):
...
+        plugin for plugin in self._plugins if plugin.plugin_name == name]
+
+    if not plugins:
+      return http_util.Respond(
+          request, 'Plugin not found', 'text/plain', code=404)
+
+    if len(plugins) > 1:
+      # Technically is not possible as plugin names are unique and is checked
+      # by the check on __init__.
+      raise AssertionError('Plugin invariant error')
```
wchargin: Suggest including the bad `plugins` list here, which will include the
Python type names of the problematic plugins. Assertions should
generally be accompanied by _some_ debugging information.

tensorboard/backend/application.py
```diff
@@ -337,6 +342,56 @@ def _create_wsgi_app(self):
...
+
+    plugin = plugins[0]
+    module_path = plugin.frontend_metadata().es_module_path
+    if not module_path:
+      return http_util.Respond(
+          request, 'Plugin is not module loadable', 'text/plain', code=400)
+
+    # non-self origin is blocked by CSP but this is a good invariant checking.
+    if urlparse.urlparse(module_path).netloc:
+      raise AssertionError('Expected es_module_path to be non-absolute path')
```
wchargin: This should be a `ValueError` (or some other normal error type), not an
`AssertionError`: I can easily construct a plugin that hits this error,
so it can only be a contract violation, not an invariant violation.

tensorboard/backend/application.py
wchargin: SGTM; thanks!

tensorboard/backend/application.py
```diff
@@ -337,6 +342,56 @@ def _create_wsgi_app(self):
...
+
+    Returns:
+      A werkzeug.Response object.
+    """
+    name = request.args.get('name')
+    plugins = [
+        plugin for plugin in self._plugins if plugin.plugin_name == name]
+
+    if not plugins:
+      return http_util.Respond(
```
wchargin: Optional, but can just `raise errors.NotFound(name)` here.

PR #2791, opened 2019-10-17
Fix GFile reading CRLF bug

### LINKED ISSUE #2777: tf.GFile read file with more characters at the end of the last line
## Environment information
python=3.7
### Diagnostics

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version 4725c70c7ed724e2d1b9ba5618d7c30b957ee8a4

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=7, micro=3, releaselevel='final', serial=0)
INFO: os.name: nt
INFO: os.uname(): N/A
INFO: sys.getwindowsversion(): sys.getwindowsversion(major=10, minor=0, build=18362, platform=2, service_pack='')

--- check: package_management
INFO: has conda-meta: True
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
INFO: installed: tensorboard==2.0.0
WARNING: no installation among: ['tensorflow', 'tensorflow-gpu', 'tf-nightly', 'tf-nightly-2.0-preview', 'tf-nightly-gpu', 'tf-nightly-gpu-2.0-preview']
WARNING: no installation among: ['tensorflow-estimator', 'tensorflow-estimator-2.0-preview', 'tf-estimator-nightly']

--- check: tensorboard_python_version
  File "tensorboard.py", line 420, in main
    suggestions.extend(check())
  File "tensorboard.py", line 77, in wrapper
    result = fn()
  File "tensorboard.py", line 231, in tensorboard_python_version
    from tensorboard import version
ImportError: cannot import name 'version' from 'tensorboard' (C:\Users\13664\Desktop\test\tensorboard.py)

--- check: tensorflow_python_version
  File "tensorboard.py", line 420, in main
    suggestions.extend(check())
  File "tensorboard.py", line 77, in wrapper
    result = fn()
  File "tensorboard.py", line 237, in tensorflow_python_version
    import tensorflow as tf
ModuleNotFoundError: No module named 'tensorflow'

--- check: tensorboard_binary_path
INFO: which tensorboard: b'F:\\Programs\\anaconda\\envs\\pytorch-nightly\\Scripts\\tensorboard.exe\r\nF:\\Programs\\anaconda\\envs\\pytorch\\Scripts\\tensorboard.exe\r\nF:\\Programs\\anaconda\\envs\\tensorflow\\Scripts\\tensorboard.exe\r\n'

--- check: readable_fqdn
INFO: socket.getfqdn(): 'DESKTOP-SBR4LHH.tsinghua.edu.cn'

--- check: stat_tensorboardinfo
INFO: directory: C:\Users\13664\AppData\Local\Temp\.tensorboard-info
INFO: os.stat(...): os.stat_result(st_mode=16895, st_ino=23362423067258806, st_dev=2856197527, st_nlink=1, st_uid=0, st_gid=0, st_size=4096, st_atime=1571050441, st_mtime=1571050441, st_ctime=1570974589)
INFO: mode: 0o40777

--- check: source_trees_without_genfiles
INFO: tensorboard_roots (1): ['F:\\Programs\\anaconda\\envs\\pytorch-nightly\\lib\\site-packages']; bad_roots (0): []

--- check: full_pip_freeze
INFO: pip freeze --all:
absl-py==0.8.1

av==6.2.0

backcall==0.1.0

certifi==2019.6.16

cffi==1.12.3

colorama==0.4.1

cycler==0.10.0

decorator==4.4.0

grpcio==1.24.1

ipykernel==5.1.2

ipython==7.8.0

ipython-genutils==0.2.0

jedi==0.15.1

jupyter-client==5.3.4

jupyter-core==4.6.0

kiwisolver==1.1.0

Markdown==3.1.1

matplotlib==3.1.1

numpy==1.17.1

olefile==0.46

parso==0.5.1

pickleshare==0.7.5

Pillow==6.2.0

pip==19.2.3

prompt-toolkit==2.0.10

protobuf==3.10.0

pycparser==2.19

Pygments==2.4.2

pyparsing==2.4.2

python-dateutil==2.8.0

pywin32==224

pyzmq==18.1.0

setuptools==41.4.0

six==1.12.0

tensorboard==2.0.0

torch==1.3.0.dev20191010

torchvision==0.5.0.dev20191013

tornado==6.0.3

traitlets==4.3.3

wcwidth==0.1.7

Werkzeug==0.16.0

wheel==0.33.6

wincertstore==0.2


``````

</details>

## Issue description

I try to read  `tensors.txt` using tf.io.GFile, 
The file:
[tensors.txt](https://github.com/tensorflow/tensorboard/files/3726978/tensors.txt)
It has 100 lines and each line has 784 characters

here is the code:
```
from tensorboard.compat import tf

with tf.io.gfile.GFile("tensors.txt", 'r') as f:
    tensor = []
    for i,line in enumerate(f):
        line = line.rstrip('\n')
        if line:
            a=list(map(float, line.split('\t')))
            try:
                assert len(a)==784
            except:
                print(i)
            tensor.append(a)
```
Output:
```
99
833
```
The length of the last line should have been 784, but it has 833 characters and in fact, there are a lot more of zeros at the end of the last line

## Reproduce
code:
```
from tensorboard.compat import tf

with tf.io.gfile.GFile("tensors.txt", 'r') as f:
    tensor = []
    for i,line in enumerate(f):
        line = line.rstrip('\n')
        if line:
            a=list(map(float, line.split('\t')))
            try:
                assert len(a)==784
            except:
                print(i)
            tensor.append(a)
```
file:
[tensors.txt](https://github.com/tensorflow/tensorboard/files/3726978/tensors.txt)

### PR DESCRIPTION
**Pull request by davidsoergel**
Fixes #2777.

The issue was that our stub GFile implementation confused byte offsets in a file on disk with character offsets in the data read from that file.

These two kinds of offsets become desynchronized from each other when a) there are multibyte characters, or b) there are CRLFs (`\r\n`), which Python automatically translates to `\n` on read.

`testReadLines` is modified here to exercise this issue, and indeed it would now fail in the absence of the given fix.

### CODE REVIEW

**Comment by davidsoergel** [comment]
@orionr FYI

File: tensorboard/compat/tensorflow_stub/io/gfile.py [hunk 1]
```diff
@@ -101,21 +101,26 @@ def read(self, filename, binary_mode=False, size=None, offset=None):
             binary_mode: bool, read as binary if True, otherwise text
             size: int, number of bytes or characters to read, otherwise
                 read all the contents of the file from the offset
-            offset: int, offset into file to read from, otherwise read
+            offset: int, offset into file to read from, in bytes; otherwise read
```
nfelt: Unfortunately I don't think it works to use a byte offset if we're not in binary mode.

For example when we pass an encoding to `io.open()` what we get is a `TextIOBase` in which `seek()` and `tell()` operate in terms of characters rather than bytes:
https://docs.python.org/3/library/io.html#io.TextIOBase.tell

For consistency, I think if we're in binary mode we want size and offset to represent bytes, and if we're in text mode we want them both to represent characters.  Otherwise we might end up doing text reads that return partial values of multibyte characters, which basically defeats the point of having a text mode.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -386,8 +397,8 @@ def __init__(self, filename, mode):
         self.fs_supports_append = hasattr(self.fs, 'append')
         self.buff_chunk_size = _DEFAULT_BLOCK_SIZE
         self.buff = None
-        self.buff_offset = 0
-        self.offset = 0
+        self.buff_char_offset = 0
```
nfelt: Similar to the above, I think we want the buff offset to be measured in either bytes or chars depending on whether we're in binary mode, rather than always chars.

orionr:
cc @sanekmelnikov, @natalialunova and @lanpa

tensorboard/compat/tensorflow_stub/io/gfile.py
davidsoergel: Completely agree.  This is actually what the code was already doing, but the variable names and docstrings were wrong/confusing.

I've clarified those, particularly emphasizing that a) we don't care what `f.tell()` returns, so long as we can pass it back to `f.seek()`, and b) similarly at the `FileSystem.read(...)` level, the continuation tokens should be opaque.

tensorboard/compat/tensorflow_stub/io/gfile.py
davidsoergel: done

davidsoergel:
Oh one more thing: In this commit, I left a TODO(orionr) regarding how this bug may persist in the S3 case. :)  Fixing that one requires a bit more logic (i.e., keeping a carryover for incomplete characters).  I don't expect to address that further--at least, until it becomes a real issue for someone.  Would you all like to take a look (@orionr, @sanekmelnikov, @natalialunova, @lanpa)?  

Once this PR is in, I'll likely file a new bug to track that case.

orionr:
Sounds good. Thank you, @davidsoergel!

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -227,30 +233,41 @@ def join(self, path, *paths):
...
+            `read(...) ' in order to continue from the last read position.
         """
         s3 = boto3.resource("s3")
         bucket, path = self.bucket_and_path(filename)
         args = {}
         endpoint = 0
+        # For the S3 case, we use continuation tokens of the form
+        # {byte_offset: number}
+        offset = None
+        if(continue_from is not None):
```
nfelt: nit: space after if, also unnecessary parens

davidsoergel: Done

PR #2793, opened 2019-10-17
program: add subcommand CLI support

wchargin:
Summary:
This commit adds a subcommand interface to support invocations in the
manner of `apt-get install` or `git commit`. All existing invocations
remain valid, and are now also available under the `tensorboard serve`
subcommand. Other subcommands may be injected programmatically. For
instance, we could migrate the existing `tensorboard --inspect` to
`tensorboard inspect`, which makes more sense semantically (as it does
something quite different from starting a web server).

To support both `tensorboard --flags` and `tensorboard serve --flags`,
we require some hacks first to get off the ground at all in Python 2,
and then to have sane error messages without massive flag duplication in
all Python versions. Even recent Python versions don’t have a notion of
a _default_ subcommand, so we need to manually inspect `argv` to
determine what kind of argument parser to construct.

Test Plan:
Unit tests added, and the smoke test for the dynamic plugin (plus the
Pip package test script) serve as integration tests for existing
behavior. It’s important to test this in both Python 2 and Python 3. For
manual testing:

  - Check that the output of `tensorboard --help` still includes all the
    normal flags, and now has a list of subcommands (one item long).
  - Check that the output of `tensorboard serve --help` is just like the
    output of `tensorboard --help`, but without the list of subcommands.
  - Patch a `demo` subcommand into `main.py`, and check that the output
    of `tensorboard demo --help` is just that subcommand’s help (without
    flags to `serve`) and that `tensorboard --help` shows the `demo`
    subcommand.

Finally, note that changing the implementation of the monkey patch to
just `yield; return` causes both `argparse_util_test` and `program_test`
to fail on Python 2 (but not Python 3).

wchargin-branch: subcommands

### DIFF
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -35,7 +35,6 @@ import argparse
 import atexit
 from collections import defaultdict
 import errno
-import gettext
 import inspect
 import logging
 import os
@@ -59,6 +58,7 @@ from tensorboard.backend import application
 from tensorboard.backend.event_processing import event_file_inspector as efi
 from tensorboard.plugins import base_plugin
 from tensorboard.plugins.core import core_plugin
+from tensorboard.util import argparse_util
 from tensorboard.util import tb_logging
 
 
@@ -206,10 +206,8 @@ class TensorBoard(object):
 
     arg0 = argv[0] if argv else ''
 
-    argparse_monkey_patch = _ArgparseMonkeyPatch()
-    argparse_monkey_patch.install()
-    flags = base_parser.parse_args(argv[1:])  # Strip binary name from argv.
-    argparse_monkey_patch.uninstall()
+    with argparse_util.allow_missing_subcommand():
+      flags = base_parser.parse_args(argv[1:])  # Strip binary name from argv.
 
     self.cache_key = manager.cache_key(
         working_directory=os.getcwd(),
@@ -260,14 +258,15 @@ class TensorBoard(object):
     return runner(self.flags) or 0
 
   def _run_serve_subcommand(self, flags):
+    # TODO(#2801): Make `--version` a flag on only the base parser, not `serve`.
+    if flags.version_tb:
+      print(version.VERSION)
+      return 0
     if flags.inspect:
       # TODO(@wchargin): Convert `inspect` to a normal subcommand?
       logger.info('Not bringing up TensorBoard, but inspecting event files.')
       event_file = os.path.expanduser(flags.event_file)
       efi.inspect(flags.logdir, event_file, flags.tag)
-      return 0
-    if flags.version_tb:
-      print(version.VERSION)
       return 0
     try:
       server = self._make_server()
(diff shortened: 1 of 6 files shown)


wchargin:
> All existing invocations remain valid, and are now also available
> under the `tensorboard serve` subcommand.

Another idea: we could call this `start`, by analogy with `npm start`
and friends. Open to input.

tensorboard/program.py
```diff
@@ -556,4 +664,41 @@ def _fix_werkzeug_logging(self):
     logging.getLogger('werkzeug').setLevel(logging.NOTSET)
 
 
+class _ArgparseMonkeyPatch(object):
```
nfelt: Optional, but maybe nicer to put this in an argparse_util.py or something since it's otherwise not really relevant to program.py.

tensorboard/program.py
```diff
@@ -556,4 +664,41 @@ def _fix_werkzeug_logging(self):
...
+  """Make Python 2.7 behave like Python 3 w.r.t. default subcommands.
+
+  The behavior of argparse was changed in Python 3.3. When a parser
```
nfelt: Might as well add a reference to https://bugs.python.org/issue16308 and https://github.com/python/cpython/commit/f97c59aaba2d93e48cbc6d25f7ff9f9c87f8d0b2 for the curious?

tensorboard/util/argparse_util.py
```diff
@@ -0,0 +1,63 @@
...
+    if "message" not in kwargs and len(args) < 2:
+      return real_error(*args, **kwargs)
+    message = kwargs["message"] if "message" in kwargs else args[1]
+    if message == ignored_message:
+      return None
+    else:
+      return real_error(*args, **kwargs)
+
+  argparse.ArgumentParser.error = error
+  yield
```
nfelt: We want a try-finally around the yield, right?  As shown in https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

tensorboard/util/argparse_util_test.py
```diff
@@ -0,0 +1,56 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import argparse
+
+from tensorboard import test as tb_test
+from tensorboard.util import argparse_util
+
+
+class AllowMissingSubcommandTest(tb_test.TestCase):
```
nfelt: Awesome, thanks for adding these tests too!

tensorboard/util/argparse_util.py
wchargin: Agh, yes, of course. For some reason I thought that `contextlib` handled
that for us, but of course it can’t, because `_handling_errors` and
friends need to be able to respond to errors. Thanks; fixed.

tensorboard/util/argparse_util_test.py
wchargin: Of course.

PR #2798, opened 2019-10-18
colab: use proxyPort for dynamic plugin

stephanwlee:
proxyPort is a relatively new API that lets colab proxy a server from a
port to url given by the method. It installs a serviceworker when
invoked so a request is proxied to the correct kernel.

This API makes:
1. Colab integration simpler
2. allows an iframe to be nestable and, hence, enable dynamic plugin

With this change, we now removed the colab not supported page.

### DIFF
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -337,19 +337,30 @@ def _display_colab(port, height, display_handle):
   redirect them, we change the document base URI, which transparently
   affects all requests (XHRs and resources alike).
   """
-  from google.colab.output import eval_js
-  url = eval_js("google.colab.kernel.proxyPort(%d)" % port)
-
   import IPython.display
-  iframe = IPython.display.IFrame(
-      src=url,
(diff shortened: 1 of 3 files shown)


tensorboard/notebook.py
```diff
@@ -337,63 +337,19 @@ def _display_colab(port, height, display_handle):
   redirect them, we change the document base URI, which transparently
   affects all requests (XHRs and resources alike).
   """
+  from google.colab.output import eval_js
+  url = eval_js('google.colab.kernel.proxyPort(%d)' % port)
```
wchargin: nit: This file uses double quotes.

stephanwlee:
> If this works for both newly launched
TensorBoards and reused TensorBoards (if display_handle:’s two cases),
then it looks good to me.

In my testing, it looked fine but we, during offline discussion, identified potential
user experience regression from using `proxyPort`. Will hold off merging.

PR #2800, opened 2019-10-18
proto: enable compiling gRPC services

wchargin:
Summary:
This commit teaches `tb_proto_library` a new `has_services` flag. When
set, the proto library will be compiled with gRPC service support, and
an additional `*_py_pb2_grpc` `py_library` target will be created.

The `py_proto_library` macro provided by the protobuf build rules no
longer suits our purpose, as it can only create one `*_py_pb2` build
target that includes both the `*_pb2` and `*_pb2_grpc` Python modules.
This is inconsistent with TensorFlow and some Google internal rules. We
now directly use `proto_gen` rather than using `py_proto_library`. This
is fine: that macro is marked as unstable because its interface may
change, and our goal is precisely to change the interface.

Test Plan:
Run

```
bazel query --output=build \
    'filter(".*protos_all.*", //tensorboard/...:all)'
```

and note that the output is identical before and after this change, so
existing protos will be unchanged.

Then, add a simple service to `tensorboard/plugins/hparams/api.proto`:

```proto
service TestService {
  rpc ListSesssionGroups(ListSessionGroupsRequest)
      returns (ListSessionGroupsResponse);
}
```

Build the Pip package and install it into a new virtualenv. Try to
import `tensorboard.plugins.hparams.api_pb2_grpc`, but note that
this fails. Then, add `has_services = True` to the hparams `BUILD` file,
and rebuild and reinstall the Pip package. The import should still fail.
Finally, add a `:protos_all_py_pb2_grpc` dep to `:hparams_plugin`, and
try once more. The import should now work, and the imported module
should have a `TestServiceStub` attribute.

wchargin-branch: proto-grpc

### DIFF
--- a/tensorboard/defs/protos.bzl
+++ b/tensorboard/defs/protos.bzl
@@ -14,14 +14,20 @@
 
 load("@com_google_protobuf//:protobuf.bzl", "py_proto_library")
 
-def tb_proto_library(name, srcs=None, visibility=None, testonly=None):
-  py_proto_library(
-      name = name + "_py_pb2",
-      srcs = srcs,
-      srcs_version = "PY2AND3",
-      deps = ["@com_google_protobuf//:protobuf_python"],
-      protoc = "@com_google_protobuf//:protoc",
-      visibility = visibility,
-      default_runtime = "@com_google_protobuf//:protobuf_python",
-      testonly = testonly,
-  )
+def tb_proto_library(
+        name,
+        srcs = None,
+        visibility = None,
+        testonly = None,
+        has_services = False):
+    py_proto_library(
+        name = name + "_py_pb2",
+        srcs = srcs,
+        srcs_version = "PY2AND3",
+        use_grpc_plugin = has_services,
+        deps = ["@com_google_protobuf//:protobuf_python"],
+        protoc = "@com_google_protobuf//:protoc",
+        visibility = visibility,
+        default_runtime = "@com_google_protobuf//:protobuf_python",
+        testonly = testonly,
+    )
--- a/tensorboard/defs/protos.bzl
+++ b/tensorboard/defs/protos.bzl
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


davidsoergel:
Can you confirm that this change works internally as well?  It seems to me that the external py_proto_library knows about use_grpc_plugin, but I'm not clear that the internal one does.  (Note TF redefines the py_proto_library rule, so copying that usage may not work as expected).  If this works as is, great!  But if not, maybe there is also a copybara thing needed too, etc. (?)

davidsoergel:
Oh nevermind, protos.bzl is only for external use anyway.  LGTM

wchargin:
Right; this will have an accompanying internal change.

wchargin:
I will restructure this to more cleanly separate the `_py_pb2` targets
from the `_py_pb2_grpc` targets. Currently, they’re in separate Python
modules but one build rule, which is not what TensorFlow does and is
also not what is used internally. TensorFlow hacks around this by
[committing the generated Python code directly to the repository][1],
but I think that can do better than that.

[1]: https://github.com/tensorflow/tensorflow/blob/f9ccd02fb44b1272798058cb80cb31e0d12a1881/tensorflow/python/debug/lib/debug_service_pb2_grpc.py#L15

wchargin:
> Can you confirm that this change works internally as well?

Thanks for this prompt—you’re correct that this `protos.bzl` is only
used in open-source TensorBoard, but the rules were still not as close
as I’d have liked them to be. Fixed; thanks! See <http://cl/275570880>.

wchargin:
Also cc @caisq, who has some experience with gRPC in open source. If
anything here seems obviously wrong to you (presumably there’s a reason
that TensorFlow doesn’t due this, but it might just be “historical”),
I’d love to hear it.

caisq:
@wchargin The debugger plugin checks in generated Python code because at the time that was done (mid 2017), there was no py grpc genrule that worked both internally and externally yet.

wchargin:
Great; thanks, all.

PR #2809, opened 2019-10-21
proto: support build deps

wchargin:
Summary:
By happenstance, none of our existing proto definitions need build-level
dependencies (dependencies on proto files from other build targets). But
this is no fundamental restriction. We teach `tb_proto_library` a `deps`
argument, pointing to other `tb_proto_library` macro invocations.

Test Plan:
Unit tests added. After building the Pip package and installing it into
a new virtualenv, we can still use `tensorboard.compat.proto` modules.

wchargin-branch: proto-deps

### DIFF
--- a/tensorboard/defs/BUILD
+++ b/tensorboard/defs/BUILD
@@ -1,5 +1,7 @@
 package(default_visibility = ["//tensorboard:internal"])
 
+load("//tensorboard/defs:protos.bzl", "tb_proto_library")
+
 licenses(["notice"])  # Apache 2.0
 
 filegroup(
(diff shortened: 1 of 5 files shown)


wchargin:
> I kinda wish the test_*.proto appeared inside tensorboard/defs/test
> directory

No such directory yet exists, and given that the tests are for a file in
the same package I think that it’s reasonable to keep them as is. For
instance, `//tensorboard:program` has `//tensorboard:program_test`, not
`//tensorboard/test:program_test`.

It’s `testonly = True`, so no risk of people accidentally depending on
them in real code without hitting a build error.

PR #2810, opened 2019-10-21
util: add `grpc_util` with backoff helper

wchargin:
Summary:
This commit adds `tensorboard.util.grpc_util`, which provides methods
for consumers of gRPC services.

Original implementation due to @nfelt.

Test Plan:
Unit tests included. They require that the `grpcio` Pip distribution
(which provides the `grpc` Python package) be installed in the enclosing
virtualenv. This is already a dependency of `tensorboard`, so on typical
setups no additional action should be required.

Co-authored-by: Nick Felt <nfelt@users.noreply.github.com>
wchargin-branch: grpc-util

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -303,6 +303,14 @@ py_library(
 )
 
 py_library(
+    name = "expect_grpc_installed",
+    # This is a dummy rule used as a grpc dependency in open-source.
+    # We expect grpc to already be installed on the system, e.g. via
+    # `pip install grpc`
+    visibility = ["//visibility:public"],
+)
+
+py_library(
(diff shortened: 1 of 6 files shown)


tensorboard/BUILD
```diff
@@ -302,6 +302,14 @@ py_library(
     visibility = ["//visibility:public"],
 )
 
+py_library(
+    name = "expect_grpc_installed",
+    # This is a dummy rule used as a grpc dependency in open-source.
+    # We expect grpc to already be installed on the system, e.g. via
+    # `pip install grpc`
```
wchargin: s/grpc/grpcio/

PR #2811, opened 2019-10-21
core: fix double requests in data loader

psybuzz:
   - Loading the scalars dashboard results in 2x network requests against data/plugin/scalars/scalars.  It turns out that our [data-loader](https://github.com/tensorflow/tensorboard/blob/418fb939fbf2505f31096eb233896733b4641354/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts#L133) fires on both leading and trailing, so 2 calls to loadData() within 100ms will trigger 2 invocations.

   - Also, the simultaneous request limit is not needed, since modern browsers have their own internal connection limit scheme.  (although we need to keep the parameter for BaseStore, which assumes we can only make 1 request at a time).

  - Removes the leading call (trailing is true by default)
  - Increases simultaneous request limit
  - Removes the Polymer async/cancelAsync

WANT_LGTM=stephan

### DIFF
--- a/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
+++ b/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
@@ -90,11 +90,11 @@ namespace tf_dashboard_common {
       },
 
       /*
-       * A set of data that has been loaded the data already. This exists to
-       * prevent fetching same data again.
+       * A set of data that is inflight or has been loaded already. This exists
+       * to prevent fetching same data again.
        * Invoking `reload` or a change in `loadKey` clears the cache.
        */
-      _loadedData: {
+      _inflightOrLoadedData: {
         type: Object,
         value: () => new Set(),
       },
@@ -112,7 +112,7 @@ namespace tf_dashboard_common {
     },
 
     reload() {
-      this._loadedData.clear();
+      this._inflightOrLoadedData.clear();
       this._loadData();
     },
 
@@ -120,7 +120,7 @@ namespace tf_dashboard_common {
       // https://github.com/tensorflow/tensorboard/issues/1499
       // Cannot use the observer to observe `loadKey` changes directly.
       if (this._canceller) this._canceller.cancelAll();
-      if (this._loadedData) this._loadedData.clear();
+      if (this._inflightOrLoadedData) this._inflightOrLoadedData.clear();
       if (this.isAttached) this._loadData();
     },
 
@@ -131,7 +131,7 @@ namespace tf_dashboard_common {
     created() {
       this._loadData = _.throttle(this._loadDataImpl, 100, {
         leading: true,
-        trailing: false,
+        trailing: true,
       });
     },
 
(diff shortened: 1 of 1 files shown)


tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -146,40 +142,39 @@ namespace tf_dashboard_common {
       }
     },
 
-    _loadDataImpl() {
+    async _loadDataImpl() {
       if (!this.active) return;
-      this.cancelAsync(this._loadDataAsync);
-      this._loadDataAsync = this.async(() => {
```
stephanwlee: you may want to keep this async around as it ties better with Polymer lifecycle.

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -146,40 +142,39 @@ namespace tf_dashboard_common {
...
+      this._setDataLoading(true);
+
+      const update = this._canceller.cancellable(async (result) => {
         const promises = this.dataToLoad
           .filter((datum) => {
             const name = this.getDataLoadName(datum);
             return !this._loadedData.has(name);
           })
-          .map((datum) => {
+          .map(async (datum) => {
```
stephanwlee: I personally prefer not using async here because it is harder to understand which promise this is referring to.

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -129,15 +129,11 @@ namespace tf_dashboard_common {
     },
 
     created() {
-      this._loadData = _.debounce(this._loadDataImpl, 100, {
-        leading: true,
```
stephanwlee: I do not want to wait 100ms for the first XHR to get fired.

wchargin:
In general it’s nice to separate changes like “fix double requests on
first load” and “raise simultaneous request limit” into separate PRs, so
that if we need to (e.g.) revert the request limit change due to some
Firefox compatibility issue we don’t also have to revert the bug fix.

psybuzz:
Split out request-limit changes into #2812 

I'm thinking, maybe [_.throttle + leading] is close to the desired behavior?
- The initial call does not wait
- A flood of loadData() calls does not starve the UI
- Loaded data comes from the most recently finished request (or a request 100ms before it)
- LoadDataCallback still fires at most once within 100ms

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -129,9 +129,9 @@ namespace tf_dashboard_common {
     },
 
     created() {
-      this._loadData = _.debounce(this._loadDataImpl, 100, {
+      this._loadData = _.throttle(this._loadDataImpl, 100, {
         leading: true,
-        trailing: true,
+        trailing: false,
```
stephanwlee: :( you really want to invoke the method on trailing as well. If `dataToLoad` as changed, you want to fetch the ones that are new.

stephanwlee: You want this to be true, too. When `dataToLoad` changes while `loadDataImpl` is doing its work, you want to invoke it again at the end so we fetch the new data we need to load. FWIW, I think `throttle` may make more sense than `debounce` so I like the change above :)

psybuzz:
@stephanwlee ready for another look

PR #2813, opened 2019-10-22
ng: make Polymer pagination configurable from Ng

stephanwlee:
This implements the last settings page setting for Polymer based
TensorBoard -- by changing the pagination setting, you can now change
how many pages ofg categories will be rendered on the Polymer side.

There still are few works to do which will be done in a follow up
changes. They are:
1. make settings page look prettier
2. bootstrap ngrx store state from tf_storage values


tensorboard/components/tf_ng_tensorboard/core/core.effects.ts
```diff
@@ -53,9 +54,9 @@ export class CoreEffects {
       tap(() => this.store.dispatch(pluginsListingRequested())),
       mergeMap(() => {
         return zip(
-          this.coreService.fetchPluginsListing(),
-          this.coreService.fetchRuns(),
-          this.coreService.fetchEnvironments()
+          this.fetchPluginsListing(),
+          this.interop.fetchRuns(),
+          this.interop.fetchEnvironments()
```
psybuzz: When we eventually get rid of the interop, will all these be defined here in core effects?

If so, maybe it makes sense to move them here directly, and instead expose `this.interop.tfStorage`, `this.interop.tfBackend`, `this.interop.paginatedViewStore`?

tensorboard/components/tf_ng_tensorboard/polymer_interop/polymer_interop.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    return from(this.tfBackend.runsStore.refresh());
+  }
+
+  fetchEnvironments() {
+    return from(this.tfBackend.environmentStore.refresh());
+  }
+
+  getPaginationLimit(): number {
+    return this.tfPaginatedViewStorage.getLimit();
+  }
```
psybuzz: Angular consumers should never `getPaginationLimit()`, they should only read from ngrx state.  And I think `setPaginationLimit()` is only needed because we need Angular `changePageSize` to notify Polymer consumers.  Can we

- Drop getPaginationLimit()
- Rename setPaginationLimit() to `notifyPaginationLimitChanged()` or something?

tensorboard/components/tf_ng_tensorboard/polymer_interop/reloader/reloader.component.ts
```diff
@@ -21,13 +21,13 @@ import {
...
-} from '../core/core.reducers';
-import {reload} from '../core/core.actions';
+} from '../../core/core.reducers';
+import {reload} from '../../core/core.actions';
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
 @Component({
-  selector: 'reloader',
+  selector: 'reloader-interop',
```
psybuzz: Could you please explain why is this renamed to "*-interop"?  I thought any file under "-interop" sounds like it should be safe to remove one day, once we get rid of Polymer.  That applies to storage-interop, but not reloader-interop.

tensorboard/components/tf_ng_tensorboard/polymer_interop/polymer_interop.service.ts
stephanwlee: I will drop it for now (because it is unused). I will soon want to use it to bootstrap the initial state for it.

tensorboard/components/tf_ng_tensorboard/core/core.effects.ts
stephanwlee: I am very hesitant to expose `tfStorage`, `tfBackend` and `paginatedViewStore` directly to any other part of the application other than what truly needs to know. I've made some change regardless. PTAL

tensorboard/components/tf_ng_tensorboard/polymer_interop/reloader/reloader.component.ts
stephanwlee: You are right. We would have this component without Polymer. I reverted the change back. Sorry!

tensorboard/components/tf_ng_tensorboard/polymer_interop/polymer_interop.service.ts
```diff
@@ -0,0 +1,44 @@
...
+==============================================================================*/
+import {Injectable} from '@angular/core';
+
+import {from} from 'rxjs';
+
+/** @typehack */ import * as _typeHackRxjs from 'rxjs';
+
+@Injectable()
+export class PolymerInteropService {
+  private tfStorage = document.createElement('tf-storage') as any;
```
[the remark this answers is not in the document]
stephanwlee: ACK. That feels like a micro-optimization that is not needed at the moment.

tensorboard/components/tf_ng_tensorboard/settings/dialog.component.test.ts
```diff
@@ -166,4 +170,35 @@ describe('settings dialog test', () => {
...
+
+    it('does not dispatch when input is invalid', fakeAsync(async () => {
+      const fixture = TestBed.createComponent(SettingsDialogComponent);
+      fixture.detectChanges();
+
+      const reloadPeriod = fixture.debugElement.query(By.css('.reload-period'));
+      reloadPeriod.nativeElement.value = 0;
+      reloadPeriod.nativeElement.dispatchEvent(new Event('input'));
+
+      tick(1e10);
```
psybuzz: Wow!  Perhaps just `flush()` from '@angular/core/testing', to skip to the end, instead of using a large, magic number?

and if that works, on L186 also?

tensorboard/components/tf_ng_tensorboard/core/core.effects.ts
psybuzz: Sounds reasonable!

PR #2815, opened 2019-10-22
ng: break direct dependency on material

stephanwlee:
This change is to unblock our internal sync which requires more granular
dependency on material components.


psybuzz:
LGTM

PR #2816, opened 2019-10-22
Decode the bytes

fhuanming:
Try to decode the "bytes" type to "str". It was comparing the 'bytes' with 'str', which will not get the expected results.

Decode the 'bytes' type to 'str' type.

* Screenshots of UI changes
No UI change.

Tested it under colab.

No.

### DIFF
--- a/tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
+++ b/tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
@@ -462,7 +462,7 @@ class WitWidgetBase(object):
     def handle_selection(selected):
       def extract_values(feat):
         if feat.HasField('bytes_list'):
-          return feat.bytes_list.value
+          return [v.decode('utf-8') for v in feat.bytes_list.value]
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
```diff
@@ -462,7 +462,7 @@ def create_selection_callback(self, examples, max_examples):
     def handle_selection(selected):
       def extract_values(feat):
         if feat.HasField('bytes_list'):
-          return feat.bytes_list.value
+          return [v.decode('utf-8') for v in feat.bytes_list.value]
```
jameswex: just want to verify, does this fix work in both py2 and py3

fhuanming: Hi, just tested on python2 colab. It also works with python2.

fhuanming:
@googlebot I signed it!

Best regards

--
Huanming Fang

PR #2817, opened 2019-10-23
core: avoid extra network request upon run selection

psybuzz:

Changing the selected runs should not cause new network requests, when the data to load is already loaded.  This was not the case in dashboards with category items that show multiple runs (e.g. scalars, pr curves).

To avoid unnecessary network requests, this PR makes some CategoryPaginatedView items keyed by tag, instead of the default tag+runs.  When runs change and tag remains constant, the new scheme lets dom-repeat efficiently reuse a template instance.

### DIFF
--- a/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
+++ b/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
@@ -116,6 +116,7 @@ limitations under the License.
             <tf-category-paginated-view
               category="[[category]]"
               initial-opened="[[_shouldOpen(index)]]"
+              get-category-item-key="[[_getCategoryItemKey]]"
             >
               <template>
                 <tf-pr-curve-card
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
```diff
@@ -195,6 +196,12 @@ <h3>No precision–recall curve data was found.</h3>
           computed:
             '_makeCategories(_runToTagInfo, _selectedRuns, _tagFilter, _categoriesDomReady)',
         },
+        // Items show multiple runs, so exclude runs from category item keys for
+        // efficient template reuse.
+        _getCategoryItemKey: {
+          type: Function,
+          value: () => (item) => item.tag,
```
stephanwlee: I think this should just call https://github.com/tensorflow/tensorboard/blob/46e6f8c2e2d98ca2154af63d91d697137b1f966a/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-dashboard.html#L294-L299.

stephanwlee: Ooops! `categoryKey` != `categoryItemKey`.

psybuzz: Actually, I think _getCategoryKey is independent of _getCategoryItemKey, so the current PR is still correct.

After checking, the PR Curves dashboard still makes an extra network request.  I'll investigate that in a followup.

PR #2818, opened 2019-10-23
ng: add dependency on material icon and input

stephanwlee:
Internal build has more fine-grained angular dependency targets. This change
address all the missing "shadow" deps.

Example of the sync: cl/276189775


wchargin:
@googlebot I consent.

PR #2824, opened 2019-10-23
build: allow configuring of extra routes based on experiment id

psybuzz:
`core_plugin.py` explicitly registers the route '/' to serve the generated '/index.html' file. This PR replaces the explicit registration with a mechanism for anyone creating a `TensorBoardWSGIApp` to pass in `extra_routes`, letting them configure paths like this.  Doing so makes it easier to serve different assets, depending on whether an `/experiment/foo` is part of the path. 

- Removes explicit '/' --> '/index.html' mapping from core_plugin.py to TensorBoardWSGIApp() callers
- Introduces configurable extra_routes with experimentId-based condition

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -160,8 +160,16 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
       path_to_run = parse_event_files_spec(flags.logdir_spec)
     start_reloading_multiplexer(
         multiplexer, path_to_run, reload_interval, flags.reload_task)
+
+  extra_routes = {
+    '/': {
+        'with_experiment': '/index.html',
+        'without_experiment': '/index.html',
+    },
+  }
   return TensorBoardWSGIApp(
-      flags, plugin_loaders, data_provider, assets_zip_provider, multiplexer)
+      flags, plugin_loaders, data_provider, assets_zip_provider, multiplexer,
+      extra_routes)
 
 
 def _handling_errors(wsgi_app):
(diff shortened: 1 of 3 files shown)


tensorboard/backend/application.py
```diff
@@ -199,6 +208,7 @@ def TensorBoardWSGIApp(
     deprecated_multiplexer: Optional `plugin_event_multiplexer.EventMultiplexer`
         to use for any plugins not yet enabled for the DataProvider API.
         Required if the data_provider argument is not passed.
+    extra_routes: See TBContext documentation for more information.
```
nfelt: Let's put the main docs here rather than TBContext, since it's only on TBContext as a hack to shoehorn it into CorePlugin.

Also, let's maybe name it `experimental_extra_routes` and/or add a warning in the docstring to indicate that it's not a stable API.

tensorboard/plugins/core/core_plugin.py
```diff
@@ -135,6 +142,16 @@ def _send_404_without_logging(self, request):
   def _redirect_to_index(self, unused_request):
     return utils.redirect('/')
 
+  @wrappers.Request.application
+  def _serve_if_experiment(self, apps, path_experiment, path_no_experiment, request):
+    """Conditionally serve paths based on whether there is an experiment id."""
+    has_experiment = plugin_util.experiment_id(request.environ)
```
nfelt: Let's call this `experiment_id_from_path` or something that indicates it's an experiment ID rather than a boolean, even if we're only using it in a boolean context right now.

psybuzz:
Thanks for the review.  Closing for now.

PR #2825, opened 2019-10-23
scalar: fix the timing issues

stephanwlee:
Repro case
----------
1. load scalars
2. simulate slower network connection (on your favorite browser or
proxy)
3. toggle on a run
4. toggle on another run
5. (depending on the dataset) you should see nothing and line chart
defaults to 0-1 scale.
6. if you double click on the line-chart, the line is shown!

Explanation
-----------
Setting: `dataToLoad` was [a, b] and "changes" to [a, b].

Before:
  1. requests for [a, b] goes out
  2. before the requests resolve, second `loadDataImpl` gets invoked and
  cancels Promises for (1).
  3. requests from (1) resolve but because Promise was cancelled, it
  does not invoke the `onLoadFinish` and domain is not reset

After:
We now invoke `onLoadFinish` if we ever invoke `loadDataCallback` even
once.
  1. requests for [a, b] goes out
  2. before the requests resolve, second `loadDataImpl` gets invoked but
  it fetches no data and Promise.all resolves on next tick. Because no
  data is fetched, it does not call the `onLoadFinish`.
  3. requests from (1) resolve and it invokes `onLoadFinish`.

Remaining "issue"
-----------------
Setting: `dataToLoad` was [a, b] and "changes" to [a, c]. Request to
fetch `c` ends before ones for `a` and `b`.
Problem: we set the domain based on `c` only but the fix will cause
major regression with regarding auto update where any data fetches will
cause the chart to rescale even if user overrode the domains.

Confirmed the fix by running the integration test 100 times on otherwise failing test.

### DIFF
--- a/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
+++ b/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
@@ -106,6 +106,11 @@ namespace tf_dashboard_common {
         type: Object,
         value: () => new Map<CacheKey, LoadState>(),
       },
+
+      _canceller: {
+        type: Object,
+        value: () => new tf_backend.Canceller(),
+      },
     },
 
     observers: ['_dataToLoadChanged(isAttached, dataToLoad.*)'],
@@ -122,6 +127,8 @@ namespace tf_dashboard_common {
     reset() {
       // https://github.com/tensorflow/tensorboard/issues/1499
       // Cannot use the observer to observe `loadKey` changes directly.
+      this.cancelAsync(this._loadDataAsync);
+      if (this._canceller) this._canceller.cancelAll();
       if (this._dataLoadState) this._dataLoadState.clear();
       if (this.isAttached) this._loadData();
     },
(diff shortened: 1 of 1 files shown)


tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -112,15 +115,14 @@ namespace tf_dashboard_common {
...
     reload() {
-      this._inflightOrLoadedData.clear();
+      this._dataLoadState.clear();
       this._loadData();
     },
 
     reset() {
       // https://github.com/tensorflow/tensorboard/issues/1499
       // Cannot use the observer to observe `loadKey` changes directly.
-      if (this._canceller) this._canceller.cancelAll();
```
psybuzz: Hm, if someone calls `reset()` while a datum is still LOADING, can the old request resolve later and trigger `loadDataCallback`, and possibly unwanted `onLoadFinish`?

Maybe we can keep this call in `reset()`, but remove the one in `loadDataImpl()` (as you've done).

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -149,43 +164,62 @@ namespace tf_dashboard_common {
...
-            if (result.cancelled) return;
-            this.onLoadFinish();
-          })
-        );
-      });
+
+          return Promise.all(promises).then(
+            this._canceller.cancellable((result) => {
+              // It was resetted. Do not notify of the data load.
+              if (!result.cancelled) {
```
psybuzz: veeery small nit: I prefer early returns, to prevent deeper indentation

stephanwlee: Thanks for the review. Since our CI takes super long and we need this to unblock sync, I will merge as is and we can address this in future time if it truly bothers you.

PR #2830, opened 2019-10-24
ng: load material icon sources

stephanwlee:
mat-icon allows many options for loading the assets. For TensorBoard, we
would like to use, instead of a CDN, the embeddable option and serve
either icon-font or SVG from our server.

This change adds dependency on google/material icon repository and
create a build rule similar to that of the internal rule to create a SVG
bundle out of many SVG icons.

image

### DIFF
--- a/tensorboard/components/tf_ng_tensorboard/app.module.ts
+++ b/tensorboard/components/tf_ng_tensorboard/app.module.ts
@@ -12,9 +12,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-import {BrowserModule} from '@angular/platform-browser';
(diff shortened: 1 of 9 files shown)


tensorboard/components/tf_ng_tensorboard/app.module.ts
```diff
@@ -48,4 +52,15 @@ import {ReloaderModule} from './reloader/reloader.module';
...
+export class AppModule {
+  constructor(
+    private domSanitizer: DomSanitizer,
+    iconRegistry: MatIconRegistry
+  ) {
+    iconRegistry.addSvgIconSet(this.getIconUrl());
+  }
+
+  private getIconUrl() {
+    return this.domSanitizer.bypassSecurityTrustResourceUrl(SVG_PATH);
```
psybuzz: Ah, so we need to explicitly bypass any URL to make it a SafeResourceURL that Angular trusts while sanitizing

PR #2832, opened 2019-10-24
data load: prevent JSE when canceling async

stephanwlee:
Polymer.prototype.async without a wait time defaults to next microtask.
Polymer implemented custom microtask scheduling which is not impervious
to some edge case.

Polymer has some internal counter for the microtasks and the `async`
call returns (contrived for illustration) `currentMicroTaskCounter + i`.
When canelling the async, it checks whether the handle is in between
`currentMicroTaskCounter` and `currentMicroTaskCounter` + [numAsyncs]
and it expects callback to exist when it is in between. The problem
happened when a handle is cancelled twice (our `reset` method gets
called more than once in a microtask for hparams) as it cannot find a
callback for the handle.

reference: https://github.com/Polymer/polymer/blob/dc20feca22ed6915eb67a6df2a8475ceba295de7/lib/utils/async.js#L198-L204

simpler repro:
```ts
const handle = this.async(() => {});
this.cancelAsync(handle);
this.cancelAsync(handle); // throws the error.
```

### DIFF
--- a/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
+++ b/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
@@ -132,8 +132,10 @@ namespace tf_dashboard_common {
     reset() {
       // https://github.com/tensorflow/tensorboard/issues/1499
       // Cannot use the observer to observe `loadKey` changes directly.
-      this.cancelAsync(this._loadDataAsync);
-      this._loadDataAsync = null;
+      if (this._loadDataAsync != null) {
+        this.cancelAsync(this._loadDataAsync);
+        this._loadDataAsync = null;
+      }
       if (this._canceller) this._canceller.cancelAll();
       if (this._dataLoadState) this._dataLoadState.clear();
       if (this.isAttached) this._loadData();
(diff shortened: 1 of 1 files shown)


tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -216,6 +223,8 @@ namespace tf_dashboard_common {
                 // Read-only property have a special setter.
                 this._setDataLoading(false);
               }
+
+              this._loadDataAsync = null;
```
psybuzz: This should go in the `!result.cancelled` section?

Or if we guard every `cancelAsync()` call, we could drop this.

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -128,6 +133,7 @@ namespace tf_dashboard_common {
       // https://github.com/tensorflow/tensorboard/issues/1499
       // Cannot use the observer to observe `loadKey` changes directly.
       this.cancelAsync(this._loadDataAsync);
+      this._loadDataAsync = null;
```
psybuzz: By setting to `null`, we assume that `null` makes Polymer internally use its `Polymer.Async.timeOut.cancel`, which does not throw.

Instead of relying on internal Polymer behavior, can we guard every `cancelAsync()` call?  e.g.

```
if (this._loadDataAsync)
  this.cancelAsync(this._loadDataAsync);
```

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
stephanwlee: Why? After the `async` is done, it should set the handle to null.

wchargin:
@googlebot I consent.

PR #2843, opened 2019-10-26
uploader: include TensorBoard version in RPCs

wchargin:
Summary:
This will enable servers to detect when clients are using an old version
and send instructions to show the user how to upgrade, especially in the
case of known bugs.

We could implement this with gRPC’s client-side interceptor API, but
that API is loudly disclaimed as experimental, so I’ve gone with the
manual plumbing for now.

Test Plan:
Ran the new uploader with an existing server and verified that it still
works (extra metadata is safely ignored). Verified that a server can
be modified to read this metadata for all three flows (upload, delete,
export). Tested in both Python 2 and Python 3.

wchargin-branch: uploader-version

### DIFF
--- a/tensorboard/uploader/BUILD
+++ b/tensorboard/uploader/BUILD
@@ -18,6 +18,7 @@ py_library(
     deps = [
         ":util",
         "//tensorboard/uploader/proto:protos_all_py_pb2",
+        "//tensorboard/util:grpc_util",
     ],
 )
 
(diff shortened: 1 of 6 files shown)


wchargin:
Yes; we may well want to add such an endpoint. One nice thing about this
formulation is that it handles the case where an uploader runs for a
long period of time during which the server version changes (and the
gRPC channel transparently reconnects). Some more discussion internally
at <http://b/142802157>.

PR #2868, opened 2019-10-30
Fix Colab notebooks to be valid json and to enable the docs generation workflow.

manivaradarajan:
Docs generation is broken for two reasons.

One file used the Colab-only magic without a try/except wrapper.
Another was missing a comma and was invalid json.

I've tested that all files changed in this PR now load in Colab correctly.


manivaradarajan:
I've checked that all files load in Colab successfully.

Issue #2634, opened 2019-09-11
TensorBoard: gracefully handle deleted event files


makslevental:
as far as i can tell this is exactly the same as this issue https://github.com/tensorflow/tensorflow/issues/3267

if i delete files from the logdir while tensorboard is running i get things like

```
E0911 11:27:19.441699 139989077399296 plugin_event_multiplexer.py:226] Unable to reload accumulator 'srresnet_voc_2x': [Errno 2] No such file or directory: b'/home/maksim/data/tensorboard/srresnet_voc_2x/events.out.tfevents.1568215513.maksim-desktop.105092.0'
E0911 11:27:24.447706 139989077399296 plugin_event_multiplexer.py:226] Unable to reload accumulator 'srresnet_voc_2x': [Errno 2] No such file or directory: b'/home/maksim/data/tensorboard/srresnet_voc_2x/events.out.tfevents.1568215513.maksim-desktop.105092.0'
E0911 11:27:29.453577 139989077399296 plugin_event_multiplexer.py:226] Unable to reload accumulator 'srresnet_voc_2x': [Errno 2] No such file or directory: b'/home/maksim/data/tensorboard/srresnet_voc_2x/events.out.tfevents.1568215513.maksim-desktop.105092.0'
E0911 11:27:34.459517 139989077399296 plugin_event_multiplexer.py:226] Unable to reload accumulator 'srresnet_voc_2x': [Errno 2] No such file or directory: b'/home/maksim/data/tensorboard/srresnet_voc_2x/events.out.tfevents.1568215513.maksim-desktop.105092.0'
```

i'm using `tb-nightly==1.15.0a20190911` through pytorch.

i'm not sure when reaping is supposed to happen e.g. as in https://github.com/tensorflow/tensorflow/issues/3267#issuecomment-292911229
or how to manually force

```
WARNING:tensorflow:Deleting accumulator 'run1/test'
WARNING:tensorflow:Deleting accumulator 'run1'
WARNING:tensorflow:Deleting accumulator 'run2/test'
WARNING:tensorflow:Deleting accumulator 'run2'
```

stephanwlee:
Hi @makslevental, perhaps I am missing few things here but what exactly is an issue? 

AFAICT, it is gracefully handling the deleted event files by printing _warning_ about the deleted runs and it prints something like below.
```
[...]

PR #2872, opened 2019-10-31
fix(event_multiplexer): handling the deleting error when AddRun or Reloader

thincal:
#2634 

- Reloader.Worker should handle the newly introduced `FileDeletedError` exception
- AddRunsFromDirectory should handle the file/dir deleted exception else the application.Reloader main process would be crashed


thincal:
@makslevental, could you help check the change ? I will add the unit test after your general checking.

thincal:
@googlebot I signed it!

thincal:
@googlebot I fixed

thincal:
@stephanwlee could help check this change about the "delete run" support ?

PR #2873, opened 2019-10-31
Fix wrong docstring: plugin tag content is bytes

davidsoergel:
Plugin content is bytes, not string as the docstring asserted.  See https://github.com/tensorflow/tensorboard/blob/79a5a194769fc245ab68cb70e5cfdc470bda88b5/tensorboard/compat/proto/summary.proto#L45


tensorboard/backend/event_processing/event_accumulator.py
```diff
@@ -294,8 +294,9 @@ def PluginTagToContent(self, plugin_name):
       KeyError: if the plugin name is not found.
 
     Returns:
-      A dict mapping tags to plugin-specific content (which are always strings).
-      Those strings are often serialized protos.
+      A dict mapping tags to plugin-specific content (which are always bytes,
```
nfelt: I think we can word this a bit more strongly - this documentation is quite stale.  E.g.

`A dict mapping tag names to bytestrings of plugin-specific content, by convention binary serialized protos.`


Also, let's make the same change to plugin_event_accumulator.py, which is the version that TensorBoard actually uses:
https://github.com/tensorflow/tensorboard/blob/52591c28ca2f8790e09b0d003a6a49efcd176501/tensorboard/backend/event_processing/plugin_event_accumulator.py#L244-L246

davidsoergel: Done, thanks!

PR #2876, opened 2019-11-01
Fix minor test failure re plugin_lib/test

davidsoergel:
The failing test was not detected by either internal or external CI, for reasons partially understood.

### DIFF
--- a/tensorboard/components/experimental/plugin_lib/BUILD
+++ b/tensorboard/components/experimental/plugin_lib/BUILD
@@ -1,6 +1,6 @@
-package(default_visibility = ["//tensorboard:internal"])
-
 load("//tensorboard/defs:defs.bzl", "tf_ts_library")
+
+package(default_visibility = ["//tensorboard:internal"])
 
 licenses(["notice"])  # Apache 2.0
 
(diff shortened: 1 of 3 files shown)


davidsoergel:
Yeah, that change *corrects* a buildifier complaint-- I don't remember why I ran it, but anyway I'm comfortable fixing that here.

psybuzz:
Curious about the description, what are the "reasons partially understood" ?  I was under the impression Travis CI is supposed to `bazel test tensorboard/...`, which should run this test?

davidsoergel:
OK, I think I should correct that to "not understood" ;)  I don't know why Travis doesn't run the test (or alternatively, why it passes).  Internally the test itself is actually a no-op, but Blaze tries to build the target anyway, and that fails without this fix.  So this PR fixes the internal build, but does not address the Travis question.

wchargin:
> > Consider removing build change, because it is irrelevant, unless you
> > want to add buildifier and enforce it :)
>
> Yeah, that change corrects a buildifier complaint-- I don't remember
> why I ran it, but anyway I'm comfortable fixing that here.

+1 to what Stephan said—if you want to change our CI to make Buildifier
errors cause build failures, then please feel free to do so, but until
we have such a check the changes will backslide and just make it harder
to see what’s actually going on in the diff (never mind polluting the
blame). This 14-line diff across 2 files is actually a 4-line change in
1 file, but that’s hard to see.

PR #2878, opened 2019-11-02
uploader: add `ServerInfo` protos and logic

wchargin:
Summary:
This commit adds an RPC definition by which the uploader can connect to
the frontend web server at the start of an upload session. This resolves
a number of outstanding issues:

  - The frontend can tell the uploader which backend server to connect
    to, rather than requiring a hard-coded endpoint in the uploader.
  - The frontend can tell the uploader how to generate experiment URLs,
    rather than requiring the backend server to provide this information
    (which it can’t, really, in general).
  - The frontend can check whether the uploader client is recent enough
    and instruct the end user to update if it’s not.
  - The frontend can warn the user about transient issues in case the
    service is down, degraded, under maintenance, etc.

An endpoint `https://tensorboard.dev/api/uploader` on the server will
provide this information.

Test Plan:
Unit tests suffice.

wchargin-branch: uploader-serverinfo-protos

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -399,6 +399,14 @@ py_library(
     visibility = ["//visibility:public"],
 )
 
+py_library(
+    name = "expect_requests_installed",
+    # This is a dummy rule used as a requests dependency in open-source.
+    # We expect requests to already be installed on the system, e.g., via
+    # `pip install requests`.
+    visibility = ["//visibility:public"],
+)
+
 filegroup(
     name = "tf_web_library_default_typings",
     srcs = [
--- a/tensorboard/uploader/proto/server_info.proto
+++ b/tensorboard/uploader/proto/server_info.proto
@@ -1,0 +1,59 @@
+syntax = "proto3";
+
+package tensorboard.service;
+
+// Request sent by uploader clients at the start of an upload session. Used to
+// determine whether the client is recent enough to communicate with the
+// server, and to receive any metadata needed for the upload session.
+message ServerInfoRequest {
+  // Client-side TensorBoard version, per `tensorboard.version.VERSION`.
+  string version = 1;
+}
+
+message ServerInfoResponse {
+  // Primary bottom-line: is the server compatible with the client, and is
+  // there anything that the end user should be aware of?
+  Compatibility compatibility = 1;
+  // Identifier for a gRPC server providing the `TensorBoardExporterService` and
+  // `TensorBoardWriterService` services (under the `tensorboard.service` proto
+  // package).
+  ApiServer api_server = 2;
+  // How to generate URLs to experiment pages.
+  ExperimentUrlFormat url_format = 3;
+}
+
+enum CompatibilityVerdict {
+  VERDICT_UNKNOWN = 0;
+  // All is well. The client may proceed.
+  VERDICT_OK = 1;
+  // The client may proceed, but should heed the accompanying message. This
+  // may be the case if the user is on a version of TensorBoard that will
+  // soon be unsupported, or if the server is experiencing transient issues.
+  VERDICT_WARN = 2;
+  // The client should cease further communication with the server and abort
+  // operation.
+  VERDICT_ERROR = 3;
+}
+
+message Compatibility {
+  CompatibilityVerdict verdict = 1;
+  // Human-readable message to display. When non-empty, will be displayed in
+  // all cases, even when the client may proceed.
+  string details = 2;
+}
+
+message ApiServer {
+  // Address of a gRPC server. For example: "dns:///api.tensorboard.dev:443".
+  string endpoint = 1;
+}
+
+message ExperimentUrlFormat {
+  // Template string for experiment URLs. All occurrences of the value of the
+  // `placeholder` field in this template string should be replaced with an
+  // experiment ID. For example, if `placeholder` is "{{EXPERIMENT_ID}}"`, then
+  // `template` might be "https://tensorboard.dev/experiment/{{EXPERIMENT_ID}}".
+  string template = 1;
+  // Placeholder string that should be replaced with an actual experiment ID.
+  // (See docs for `template` field.)
+  string placeholder = 2;
+}
--- a/tensorboard/uploader/proto/server_info.proto
+++ b/tensorboard/uploader/proto/server_info.proto
@@ -31,7 +31,7 @@ enum CompatibilityVerdict {
   // soon be unsupported, or if the server is experiencing transient issues.
   VERDICT_WARN = 2;
   // The client should cease further communication with the server and abort
-  // operation.
+  // operation after printing the accompanying `details` message.
   VERDICT_ERROR = 3;
 }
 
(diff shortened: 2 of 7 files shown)


wchargin:
Googlers: see <http://cl/278027294> for server implementation.

tensorboard/uploader/proto/server_info.proto
```diff
@@ -0,0 +1,59 @@
...
+  string details = 2;
+}
+
+message ApiServer {
+  // Address of a gRPC server. For example: "dns:///api.tensorboard.dev:443".
+  string endpoint = 1;
+}
+
+message ExperimentUrlFormat {
+  // Template string for experiment URLs. All occurrences of the value of the
```
nfelt: Do we want these to be absolute or relative?

Relative has the advantage that it allows a single frontend to be exposed across multiple domain names and clients would generate the appropriate URLs.  Of course, we could I suppose extract this from the incoming `/api/uploader` request and do the substitution server-side; I could see an argument for either approach.

Another possibility would be adding a template placeholder for the base URL, which if included is basically the same as a relative URL, but could be omitted in the actual template to allow the frontend to return an absolute URL if it wanted to avoid ambiguity (e.g. if wanted to do that on prod).

tensorboard/uploader/server_info.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+  try:
+    return server_info_pb2.ServerInfoResponse.FromString(response.content)
+  except message.DecodeError as e:
+    raise CommunicationError(
+        "Corrupt response from backend (%s): %r" % (e, response.content)
+    )
+
+
+def create_server_info(frontend_origin, api_endpoint):
```
nfelt: If this is only used in tests, I think it's preferable to keep it in test code or utilities.

tensorboard/uploader/server_info_test.py
```diff
@@ -0,0 +1,152 @@
...
+  """Gets family and nodename for a loopback address."""
+  s = socket
+  infos = s.getaddrinfo(None, 0, s.AF_UNSPEC, s.SOCK_STREAM, 0, s.AI_ADDRCONFIG)
+  (family, _, _, _, address) = infos[0]
+  nodename = address[0]
+  return (family, nodename)
+
+
+class _Ipv6CompatibleWsgiServer(simple_server.WSGIServer):
+  address_family = _localhost()[0]
```
nfelt: It's a wee bit ugly IMO to have this resolved at class definition time.  Optional, but we could instead pass the address_family into `__init__` explicitly and set it on the instance then (prior to calling superclass init which is what constructs the socket).  We'd have to explicitly construct the server rather than using make_server() but it's only 3 lines anyway.

https://github.com/python/cpython/blob/2.7/Lib/wsgiref/simple_server.py#L147

tensorboard/BUILD
```diff
@@ -399,6 +399,14 @@ py_library(
     visibility = ["//visibility:public"],
 )
 
+py_library(
+    name = "expect_requests_installed",
```
nfelt: We should take an explicit dep on this in our `setup.py` now that it's a direct dependency, right?

tensorboard/uploader/server_info_test.py
wchargin: OK: overriding `address_family` directly on the class like this is
common (e.g., [this Stack Overflow question][1]), but I’ll grant that
it’s weird for the right-hand side to not be a compile-time constant
expression.

For some reason, `TCPServer` (the ancestor class that actually consumes
`address_family`) [explicitly says not to override its initializer][2],
so even though I can’t directly see what might go wrong if we should
ignore this directive, I’ve gone ahead and just created the class
dynamically. Done.

[1]: https://stackoverflow.com/q/10085996
[2]: https://github.com/python/cpython/blob/17269090940aa20f6079a6b9f27ae319f8cdae14/Lib/socketserver.py#L446

tensorboard/uploader/proto/server_info.proto
```diff
@@ -0,0 +1,59 @@
...
+  // `placeholder` field in this template string should be replaced with an
+  // experiment ID. For example, if `placeholder` is "{{EXPERIMENT_ID}}", then
+  // `template` may be "https://tensorboard.dev/experiment/{{EXPERIMENT_ID}}/".
+  string template = 1;
+  // Placeholder string that should be replaced with an actual experiment ID.
+  // (See docs for `template` field.)
+  string placeholder = 2;
```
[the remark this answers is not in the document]
wchargin: Done.

tensorboard/uploader/proto/server_info.proto
```diff
@@ -0,0 +1,59 @@
...
+
+message Compatibility {
+  CompatibilityVerdict verdict = 1;
+  // Human-readable message to display. When non-empty, will be displayed in
+  // all cases, even when the client may proceed.
+  string details = 2;
+}
+
+message ApiServer {
+  // Address of a gRPC server. For example: "dns:///api.tensorboard.dev:443".
```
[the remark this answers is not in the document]
wchargin: Thanks for the reference! Looks like those docs say that `dns` is the
default scheme, which is consistent for all our observations except for
some Google-specific internal tools. That seems fine: changed to call
this a “gRPC server URI” and link to those docs, and changed the example
to omit the `dns:///` prefix (like the actual server response will).

tensorboard/BUILD
wchargin: Yes—thanks!

tensorboard/uploader/proto/server_info.proto
nfelt: SGTM, thanks for the explanation.

tensorboard/uploader/server_info_test.py
nfelt: I'm about 95% sure that the directive `May be extended, do not override` is a poorly worded instruction that subclasses that define `__init__()` should be sure to call `super().__init__()` and thus "extend" `__init__()`, as opposed to if they didn't call `super()` which would be "overriding".

Evidence for this would be that otherwise "extended" makes no sense, and also that plenty of subclasses of `TCPServer` exist that define their own `__init__()` including the werkzeug dev server we currently use, as well as some examples in stdlib like https://github.com/python/cpython/blob/484edbf9bf1a9e6bae0fcb10a0c165b89ea79295/Lib/logging/config.py#L869-L884

Manually constructing the type on the fly is... new to me :P But if you'd rather not touch `__init__` then it will do.

tensorboard/uploader/server_info.py
nfelt: Ok, I didn't check follow-on changes for usage.

If the goal is to allow locally overriding the API endpoint though, it seems better to me to just have `--api_endpoint` override whatever's returned from the frontend's ServerInfo endpoint, rather using it to construct a new ServerInfo without hitting the frontend at all.  As-is, it's semantically not really right to always use `/experiment/` as the URL pattern here, since that should be determined by the frontend.

Or concretely, if I were running local storzy but against the hosted-tensorboard-dev Spanner, I might want to be able to run `tensorboard dev --origin=https://hosted-tensorboard-dev.appspot.com --api_endpoint=localhost:10000` and I would argue that should use the experiment pattern and other metadata fetched from the origin still.

If it would seem weird if specifying a local `--api_endpoint` but not specifying `--origin` for it to default to pinging `tensorboard.dev`, we could change the default for `--origin` so that it only defaults to `tensorboard.dev` when `--api_endpoint` is not passed, and otherwise it defaults to empty and then only in that case do we locally construct the `ServerInfo` (and maybe in that case we should just use a dummy experiment URL pattern).

tensorboard/uploader/server_info_test.py
wchargin: Ah—not “This class may be extended; do not override its initializer” but
rather ”You may extend this initializer, as long as you do not override
(= fail to invoke) its superclass implementation.” Yeah, that at least
makes some sense. I had also noted that `TCPServer`’s base class has the
same directive, and of course `TCPServer` defines an initializer, so
this is more evidence for your idea.

Either approach is fine with me, then, so I’ll keep it as is.

tensorboard/uploader/server_info.py
wchargin: Agreed; I also wasn’t fully content with the flags formulation in #2879.
Your approach sounds fine to me: I’ll revise that PR accordingly.

PR #2884, opened 2019-11-04
docs: fix annotations related to TBLoader

psybuzz:
One step towards improving plugin authoring is to understand and document the public interfaces.  Authors can write a class that extends base_plugin's `TBLoader`, but its `load()` method was not documented.

Adds a base class method `TBLoader.load()` and updates some related docstrings.

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -190,8 +190,7 @@ def TensorBoardWSGIApp(
 
   Args:
     flags: An argparse.Namespace containing TensorBoard CLI flags.
-    plugins: A list of plugins, which can be provided as TBPlugin subclasses
-        or TBLoader instances or subclasses.
+    plugins: A list of plugin loader instances.
     assets_zip_provider: See TBContext documentation for more information.
     data_provider: Instance of `tensorboard.data.provider.DataProvider`. May
         be `None` if `flags.generic_data` is set to `"false"` in which case
@@ -202,6 +201,8 @@ def TensorBoardWSGIApp(
 
   Returns:
     A WSGI application that implements the TensorBoard backend.
+
+  :type plugins: list[base_plugin.TBLoader]
   """
   db_uri = None
   db_connection_provider = None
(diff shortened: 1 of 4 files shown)


tensorboard/backend/application.py
```diff
@@ -190,8 +190,7 @@ def TensorBoardWSGIApp(
 
   Args:
     flags: An argparse.Namespace containing TensorBoard CLI flags.
-    plugins: A list of plugins, which can be provided as TBPlugin subclasses
-        or TBLoader instances or subclasses.
+    plugins: A list of plugin loader instances.
```
stephanwlee: Technically, old comment was correct. The plugin can be either TBPlugin or TBLoader.

psybuzz: That's what I thought at first, but if I'm reading the code correctly:
If we did pass in a TBPlugin instance, things would break.  This method calls `make_plugin_loader()` on all the plugins passed in, which will throw a TypeError, since it only accepts TBLoader instances, or TBPlugin/TBLoader subclasses.

psybuzz:
Thanks for the review!

Per offline discussion: it looks like there's an opportunity to clean things up.  In the future, we could look into removing a redundant call to make_plugin_loader() in application.py, and only call it in program.py.

tensorboard/backend/application.py
nfelt: It's correct that this can't be a TBPlugin instance, but it can be any of the three things the docstring originally listed (TBPlugin subclass, TBLoader subclass, or TBLoader instance) so unless I'm missing something, I don't think we actually wanted to change the wording?

psybuzz: Sorry, I probably should have waited before merging too soon.

The docstring originally lists those 3, but the only type being passed in is `TBLoader` instances.  `TensorBoardWSGIApp` only has one caller, which provides a list that has already gone through `make_plugin_loader`.

`application.py` is not supposed to be publicly exposed, so if we assume that we are the only ones using TBWSGIApp, updating the comment is just aligning it to the existing narrow case.

PR #2893, opened 2019-11-05
Encode audio data only when summaries are written

hongjunChoi:
Currently for audio summary, encoding logic is executed every step regardless whether tf.summary.record_if() is satisfied or not. This causes performance overhead. As so, pass in callable that returns tensor to tf.summary.write() instead of directly passing tensor.


nfelt:

PR #2894, opened 2019-11-05
Execute histogram bucketing logic only when summary is written for tf.summary.histogram()

hongjunChoi:
Currently for histogram summary, bucketing logic is executed every step regardless whether tf.summary.record_if() is satisfied or not.. As so, pass in callable that returns tensor to tf.summary.write() instead of directly passing tensor.


nfelt:

PR #2895, opened 2019-11-05
Encode image data only when summary is written for tf.summary.image()

hongjunChoi:
Pass callable to tf.summary.write() `tensor` parameter instead of tensor itself. This makes sure that input data encoding logic is only executed when summaries are written.


hongjunChoi:
Failing travis seems to be caused by out-of-sync tf nightly.

nfelt:

PR #2896, opened 2019-11-05
Refactor /webapp per Angular structure discussion

davidsoergel:
This refactors the new `/webapp` directory, structured as agreed per internal doc and discussion.  Aside from file moves and associated import updates, some changes are:

* Consolidate build rules into `tensorboard/webapp/BUILD` (including two from `/tensorboard/BUILD`).
* Reorder rules in `tensorboard/webapp/BUILD` according to their dependency structure.  The main app component is wrapped in 6 layers of rules- all for good reason, but I was too easily confused when they were out of order.
* Moved `webapp/core/core.service.ts` to `webapp/webapp_data_source`, with its own `TBServerDataSource` module.

```
tensorboard/webapp
   |-- settings
   |   |-- settings.module.ts
   |   |-- settings_button.component.ts
   |   |-- dialog.component.ts
   |   |-- BUILD
   |   |-- dialog.component.test.ts
   |   |-- dialog.component.css
   |-- app.container.html
   |-- webapp_data_source
   |   |-- tb_server.data_source.ts
   |   |-- tb_server.data_source.module.ts
   |   |-- BUILD
   |-- reloader
   |   |-- reloader.module.ts
   |   |-- BUILD
   |   |-- reloader.component.test.ts
   |   |-- reloader.component.ts
   |-- types
   |   |-- api.ts
   |   |-- BUILD
   |-- index.html
   |-- core
   |   |-- testing
   |   |   |-- index.ts
   |   |   |-- BUILD
   |   |-- actions
   |   |   |-- core.actions.ts
   |   |   |-- index.ts
   |   |-- core.module.ts
   |   |-- BUILD
   |   |-- effects
   |   |   |-- core.effects.ts
   |   |   |-- core.effects.test.ts
   |   |   |-- index.ts
   |   |-- store
   |   |   |-- core.reducers.test.ts
   |   |   |-- core.selectors.ts
   |   |   |-- core.reducers.ts
   |   |   |-- core.types.ts
   |   |   |-- index.ts
   |-- main.ts
   |-- plugins
   |   |-- plugins.component.html
   |   |-- plugins.container.ts
   |   |-- plugins.container.test.ts
   |   |-- plugins.component.css
   |   |-- plugins.module.ts
   |   |-- BUILD
   |-- app.container.css
   |-- main.prod.ts
   |-- config.prod.ts
   |-- angular
   |   |-- BUILD
   |-- app.module.ts
   |-- testing
   |   |-- initialize_testbed.ts
   |   |-- BUILD
   |   |-- rxjs_shims.js
   |-- config.dev.ts
   |-- BUILD
   |-- app.container.ts
   |-- reducer_config.ts
   |-- header
   |   |-- header.test.ts
   |   |-- types.ts
   |   |-- header.component.html
   |   |-- header.component.css
   |   |-- header.component.ts
   |   |-- header.module.ts
   |   |-- header.container.ts
   |   |-- BUILD
   |-- bootstrap.ts
```


davidsoergel:
Oh I forgot to add separate `BUILD` files for the `./core/*` subdirectories.  Gotta run now but will do soon.

tensorboard/webapp/core/effects/core.effects.ts
```diff
@@ -24,19 +24,20 @@ import {
...
-} from './core.actions';
-import {State, getPluginsListLoaded} from './core.reducers';
-import {LoadState} from '../types/api';
+} from '../actions';
+import {getPluginsListLoaded} from '../store';
+import {LoadState} from '../../types/api';
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 /** @typehack */ import * as _typeHackNgrx from '@ngrx/store/src/models';
+import {State} from '../store/core.types';
```
stephanwlee: Prefer to keep non-hack related imports above the hack. Here and other files.

tensorboard/webapp/webapp_data_source/BUILD
```diff
@@ -0,0 +1,18 @@
+package(default_visibility = ["//tensorboard:internal"])
+
+load("@npm_angular_bazel//:index.bzl", "ng_module")
+load("@npm_bazel_jasmine//:index.bzl", "jasmine_node_test")
```
stephanwlee: extraneous load.

tensorboard/webapp/webapp_data_source/BUILD
```diff
@@ -0,0 +1,18 @@
...
+ng_module(
+    name = "webapp_data_source",
+    srcs = [
+        "tb_server.data_source.module.ts",
+        "tb_server.data_source.ts",
+    ],
+    deps = [
+        "//tensorboard/webapp/angular:expect_angular_common_http",
+        "//tensorboard/webapp/types",
+        "@npm//@angular/common",
```
stephanwlee: I forgot which package provides `@angular/core`. Is this the one? Is `@angular/common' actually being used? We need better linter/checker (gts!) :(

tensorboard/webapp/core/store/core.types.ts
```diff
@@ -0,0 +1,45 @@
...
+  state: DataLoadState;
+  // Time since epoch.
+  lastLoadedTimeInMs: number | null;
+}
+
+export interface CoreState {
+  activePlugin: PluginId | null;
+  plugins: PluginsListing;
+  pluginsListLoaded: LoadState;
+  reloadPeriodInMs: number;
```
caisq: nit: Prefer alphabetical sort of keys?

tensorboard/webapp/plugins/plugins.container.ts
```diff
@@ -54,10 +49,10 @@ const lastLoadedTimeInMs = createSelector(
 
 @Component({
   selector: 'plugins',
-  templateUrl: './plugins.component.html',
-  styleUrls: ['./plugins.component.css'],
+  template: '<div #plugins class="plugins"></div>',
+  styles: ['.plugins { height: 100%; }', 'iframe { border: 0; }'],
```
caisq: Might be naive question: Why inlining the CSS instead of importing a .css file like before?

tensorboard/webapp/webapp_data_source/tb_server.data_source.module.ts
```diff
@@ -0,0 +1,23 @@
...
+==============================================================================*/
+import {NgModule} from '@angular/core';
+import {HttpClientModule} from '@angular/common/http';
+import {TBServerDataSource} from './tb_server.data_source';
+
+@NgModule({
+  imports: [HttpClientModule],
+  providers: [TBServerDataSource],
+})
+export class TBServerDataSourceModule {}
```
caisq: nit: This is not in a complete state, right? If so, add TODO items to remind us what needs to be added?

tensorboard/webapp/core/store/core.selectors.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const getPluginsListLoaded = createSelector(
+  selectCoreState,
+  (state: CoreState): LoadState => state.pluginsListLoaded
+);
+
+export const getActivePlugin = createSelector(
+  selectCoreState,
+  (state: CoreState): PluginId | null => {
+    return state.activePlugin;
```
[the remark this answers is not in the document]
stephanwlee: Let us minimize the mostly move PR to moves.

tensorboard/webapp/plugins/plugins.container.ts
stephanwlee: If I may guess, I presume that is because we would like to avoid unnecessary template extraction and building step of Angular by just providing the content of CSS and HTML directly here. Not sure if there is any build time performance gain when using AOT but it should help with JIT.

tensorboard/webapp/webapp_data_source/tb_server.data_source.module.ts
stephanwlee: This feels like more less complete. What needs to be done?

caisq: OK. I see. The business logic is in the `class TBServerDataSource` in `tb_server.data_source.ts`. 

Resolve this question.

tensorboard/webapp/plugins/plugins.container.ts
davidsoergel: I was just following our rule of thumb that inlining is easier when there are only a few lines of code.  Additional files carry the burden of maintaining BUILD, directory listing bloat, etc.  That's worth it only when they have enough content that it'd be awkward to inline.

tensorboard/webapp/webapp_data_source/BUILD
davidsoergel: Done here; I'm afraid there is more of this throughout (especially unused deps).

tensorboard/webapp/webapp_data_source/BUILD
davidsoergel: `common` is used for `HttpClient`.  I guess it was picking up `core` transitively but I went ahead and included it explicitly for completeness.

tensorboard/webapp/core/effects/core.effects.ts
davidsoergel: Done, thanks.  Looking forward to having this stuff automated...

tensorboard/webapp/core/store/core.selectors.ts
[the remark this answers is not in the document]
davidsoergel: +1 to both comments-- let's such cleanups in later PRs.

tensorboard/webapp/core/store/core.types.ts
davidsoergel: +1 but punted to later PR

davidsoergel:
Oh, to clarify, Taze is about BUILD rules.  For import statements, I let VSCode insert them.  In other projects we've had an autoformatter to further clean up and sort the imports; that would be great here too, but not today.

tensorboard/webapp/core/store/core.selectors.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import {createSelector, createFeatureSelector} from '@ngrx/store';
+import {PluginId, PluginsListing} from '../../types/api';
+
+// HACK: These imports are for type inference.
+// https://github.com/bazelbuild/rules_nodejs/issues/1013
+/** @typehack */ import * as _typeHackSelector from '@ngrx/store/src/selector';
+/** @typehack */ import * as _typeHackStore from '@ngrx/store/store';
+
+import {CoreState, State, CORE_FEATURE_KEY, LoadState} from './core.types';
```
stephanwlee: nit: import statement below the hacks.

davidsoergel: Oops, the way I searched for these before was incomplete.  Fixed this and two more :)

tensorboard/webapp/header/header.container.ts
```diff
@@ -15,17 +15,14 @@ limitations under the License.
...
+import {PluginId} from '../types/api';
+import {UiPluginMetadata} from './types';
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
-export interface UiPluginMetadata extends PluginMetadata {
-  id: PluginId;
-}
-
 const getUiPlugins = createSelector(
```
psybuzz: For my understanding, just to confirm: Containers can define their own Selectors in *container.ts instead of from core/store/selectors.ts or <feature>/store/selectors.ts?  so long as the logic is specific to the Container?

tensorboard/webapp/plugins/BUILD
```diff
@@ -8,15 +8,12 @@ licenses(["notice"])  # Apache 2.0
...
+        "plugins.container.ts",
         "plugins.module.ts",
     ],
-    assets = [
-        "plugins.component.css",
-        "plugins.component.html",
-    ],
     deps = [
         "//tensorboard/webapp/core",
+        "//tensorboard/webapp/core/store",
```
psybuzz: Many places are depending on both /core and /core/store.  Is the latter necessary here?  Or can we drop the former, if it's not using /core/actions, etc?

tensorboard/webapp/header/header.container.ts
davidsoergel: Correct.  In this context, 'selector' is just a synonym for 'memoized observable'.  Critically, it does not reach inside the NgRx state, because it obtains data from another Selector imported from `.../store` (`getPlugins`, in this case).

tensorboard/webapp/plugins/BUILD
davidsoergel: The two deps provide different things: `core/store` provides the State, Reducer, and Selectors, while `core` provides only the NgModule (`CoreModule`).  It may be that some of these deps are unused, but I'd like to punt that cleanup to a later PR.

tensorboard/webapp/core/effects/core.effects.test.ts
```diff
@@ -65,7 +65,7 @@ describe('core.effects', () => {
     store = TestBed.get(Store);
     dispatchSpy = spyOn(store, 'dispatch');
 
-    const coreService = TestBed.get(CoreService);
+    const coreService = TestBed.get(TBServerDataSource);
```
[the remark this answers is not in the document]
davidsoergel: done

tensorboard/webapp/plugins/BUILD
psybuzz: Makes sense, thanks!

tensorboard/webapp/header/header.container.ts
psybuzz: Thanks for the explanation.  In the design doc, I didn't realize that selectors can be defined in both places.  It seems very reasonable though, since these selectors are deriving information that is close to where it is used.

PR #2899, opened 2019-11-06
Defer image/audio/histogram v2 summary preprocessing using LazyTensorCreator

nfelt:
Continuation of PRs #2893, #2894, #2895 by @hongjunChoi .

This takes advantage of a recent change in `tf.summary.write()` as of https://github.com/tensorflow/tensorflow/commit/661f0ac0015184e970f9a3f04e845a9166ebfa7a that allows passing a callable as the `tensor` argument instead of an actual `tf.Tensor`.  With this functionality, `write()` will call the callable to get the tensor only if the tensor is going to be used (i.e. the summary writer exists and the summary recording condition is met).  This avoids overhead in executing preprocessing steps when they are not needed.  Among V2 summaries, only image, audio, and histogram do any real preprocessing, so those are the only ones changed in this PR.

In order to maintain backwards compatibility with the previous `write()` API, this introduces a shim layer called `LazyTensorCreator` that uses `tf.register_tensor_conversion_function()` to ensure that the callable we pass in can be automatically converted to a `tf.Tensor` in many contexts (and in particular, in the context of the call to `tf.identity()` which is used inside `write()`).  That way a version skew (e.g. TensorBoard 2.1 but TensorFlow 2.0) won't break the summary API integration; while technically we don't guarantee that a skew in this direction will work, this is a likely enough situation and a confusing enough error that it's worth a bit of work to avoid.

cc @alextp FYI

### DIFF
--- a/tensorboard/plugins/histogram/summary_v2.py
+++ b/tensorboard/plugins/histogram/summary_v2.py
@@ -29,11 +29,14 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import functools
+
 import numpy as np
 
 from tensorboard.compat import tf2 as tf
 from tensorboard.compat.proto import summary_pb2
 from tensorboard.plugins.histogram import metadata
+from tensorboard.util import lazy_tensor_creator
 from tensorboard.util import tensor_util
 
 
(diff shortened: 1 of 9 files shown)


tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -29,11 +29,14 @@
 from __future__ import division
 from __future__ import print_function
 
+import functools
```
wchargin: Unused?

tensorboard/util/lazy_tensor_creator_test.py
```diff
@@ -0,0 +1,100 @@
...
+      return tf.constant(1, dtype=tf.int32)
+    with self.assertRaisesRegex(RuntimeError, "ref tensor"):
+      # Call conversion routine manually since this isn't actually
+      # exposed as an argument to tf.convert_to_tensor.
+      lazy_tensor_creator._lazy_tensor_creator_converter(
+          lazy_tensor, as_ref=True)
+
+  def test_reentrant_callable_does_not_deadlock(self):
+    @lazy_tensor_creator.LazyTensorCreator
+    def lazy_tensor(nested_call=False):
```
wchargin: Is this `nested_call=False` vestigial?

tensorboard/plugins/histogram/summary_v2.py
nfelt: Removed, thanks.

tensorboard/util/lazy_tensor_creator_test.py
nfelt: Yep, removed.  Thanks for the catch.

PR #2903, opened 2019-11-07
uploader: add simple `list` subcommand

wchargin:
Summary:
The new `tensorboard dev list` command prints links to your experiments.
This is implemented by repurposing the `StreamExperiments` export RPC,
which only includes experiment IDs. We can expand this to additionally
show useful metadata: experiment creation time and last-modified time;
total number of scalars; counts of runs, tags, or time series; and
selected run and tag names could all be useful to include.

Test Plan:
Ran `tensorboard dev list` on an account with 12 experiments and an
account with no experiments, starting from both logged-in and logged-out
states. Verified that the printed experiment links resolve correctly.
Verified that the normal export flow still works.

wchargin-branch: uploader-list

### DIFF
--- a/tensorboard/uploader/exporter.py
+++ b/tensorboard/uploader/exporter.py
@@ -126,13 +126,7 @@ class TensorBoardExporter(object):
 
   def _request_experiment_ids(self, read_time):
     """Yields all of the calling user's experiment IDs, as strings."""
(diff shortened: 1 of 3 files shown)


wchargin:
> We can expand this to additionally show useful metadata: experiment
> creation time and last-modified time; total number of scalars; counts
> of runs, tags, or time series; and selected run and tag names could
> all be useful to include.

Done in #2906 and #2941.

PR #2906, opened 2019-11-08
uploader: add metadata to experiment listing RPC

wchargin:
Summary:
This extends the `StreamExperiments` RPC such that the client can
specify a set of additional metadata fields that the server should
include, like “creation time” or “number of scalar points”.

The format is both forward- and backward-compatible. Servers are
expected to send responses with both `experiment_ids` and `experiments`
until we drop support for clients that do not support `experiments`, at
which point they need only send `experiments`.

Test Plan:
Unit test added to simulate the future behavior of servers.

wchargin-branch: streamexperiments-metadata


wchargin:
Leaving this as a draft; I’ll write the server-side implementation and
make sure that this checks out before requesting review. Pre-emptive
drive-by comments are always welcome.

wchargin:
Googlers, see server-side implementation in <http://cl/281200267>.

wchargin:
Closing and re-opening to trigger CI…

PR #2909, opened 2019-11-08
What-If Tool attribution display improvements

jameswex:

Visual updates per user feedback and testing with real-world models.

Allow displaying of attributions from multiple models. Updated attribution values for use in Facets Dive to be from both models if two models are used, with appropriate model name suffixes. Add dropdown to select which model the attributions are from that are used in the datapoint editor display.

Changed the coloring scale for attributions to allow for more color variation near 0, as in real-world usage there is a lot of values bunched up near zero versus at the extremes. Updated legend generation code to handle this non-linear color gradient on the linear scale.

When attributions are provided (and not in counterfactual mode), use the 2nd column in the datapoint editor to show the actual attribution values next to the feature values, per feedback. Required allowing this "compare mode column" to be used in a way where it doesn't highlight differences.

Updated styling on dropdowns to be consistent throughout tool.

* Screenshots of UI changes
One regression model:
attrtable

Attributions with multi models:
attr-twomodel

Ran attribution and non-attribution demos and tried all visual options in them. Also tried two models with attributions to test this case, although not submitting this update as we don't want to include attributions in the multi model demo permanently.


tensorboard/components/vz_example_viewer/vz-example-viewer.ts
```diff
@@ -1359,6 +1389,13 @@ namespace vz_example_viewer {
      * json.
      */
     createCompareExamplesFromJson: function(json: string) {
+      // If setting the compare example back to empty, and there is saliency
+      // information, then set the compare info to the saliency so that it
+      // is displayed instead of nothing.
+      if (!json || (!Object.keys(json).length && this.saliencyJson)) {
```
tolga-b: I think these two if statements could be simplified to make them easier to read.
(e.g. if !json then you always set json = this.saliencyJson which will enter the if below if !this.saliencyJson is true. )

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
```diff
@@ -6656,48 +6700,56 @@ <h2>Show similarity to selected datapoint</h2>
...
-                : [datapoint[attribKey]];
-              for (
-                let attribIdx = 0;
-                attribIdx < newAttributions.length;
-                attribIdx++
-              ) {
-                if (newAttributions[attribIdx] > this.maxAttribution) {
-                  this.maxAttribution = newAttributions[attribIdx];
+            for (
+              let modelNum = 0;
```
tolga-b: In the case model 2 has attributions and model 1 does not, WIT may falsely show model 2's attributions as model 1's. We can hide the model indicator whenever the attribution length is 1 for now and add a TODO to handle this case correctly with the indicator in future.

tensorboard/components/vz_example_viewer/vz-example-viewer.ts
jameswex: simplified

tensorboard/plugins/interactive_inference/tf_interactive_inference_dashboard/tf-interactive-inference-dashboard.html
jameswex: implemented this by only appending model index to these new datapoint values when both models have attributions, through new method attributionStrWithModelName_ and correct hiding of the attribution model selector. added TODO for further work.

PR #2910, opened 2019-11-08
Remove unused deps blocking sync

davidsoergel:
Two copypasta dependencies broke our internal sync.

Also, internally we enforce strict deps, so I added a bunch that were transitive before. 

There are probably more unused deps that we can clean up later, but that are not blocking as far as I know.

### DIFF
--- a/tensorboard/webapp/settings/BUILD
+++ b/tensorboard/webapp/settings/BUILD
@@ -20,6 +20,8 @@ ng_module(
         "//tensorboard/webapp/angular:expect_angular_material_icon",
         "//tensorboard/webapp/angular:expect_angular_material_input",
         "//tensorboard/webapp/core",
+        "//tensorboard/webapp/core/store",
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/settings/BUILD
```diff
@@ -20,6 +20,8 @@ ng_module(
         "//tensorboard/webapp/angular:expect_angular_material_icon",
         "//tensorboard/webapp/angular:expect_angular_material_input",
         "//tensorboard/webapp/core",
+        "//tensorboard/webapp/core/store",
+        "//tensorboard/webapp/core/actions",
```
stephanwlee: some buildifier on these files would be nice (I used it before so it is kosher here; this file requires alphasorting).

davidsoergel: I'd really like to get buildifier working externally as a matter of course (i.e., even in our CI).  Let's punt this until then.

PR #2914, opened 2019-11-09
Relax minimum version of requests to allow 2.21.0

nfelt:
Confirmed that `server_info_test` still passes with `requests==2.21.0`.


wchargin:
For posterity: The latest version of `requests` that supports Python 3.4
is 2.21.0. While we don’t officially support Python 3.4, we also don’t
need any of the functionality introduced specifically in 2.22.0, so it
doesn’t hurt us to relax the dependency bound.

PR #2916, opened 2019-11-11
WitWidget small fixes for upcoming 1.5.0 version

jameswex:

Updates needed after testing for 1.5.0 launch.

Update method to sanitize feature names for use in CSS rules to also sanitize whitespace. This was incorrectly missing and would lead to incorrect styling of feature values for features that included whitespace in their names.

Update basic notebook usage example to be compatible with tf 2.0 (some namespace changes).

Add missing new pip package dependency on oauth2client for Cloud AI Platform usage.

Clone relevant info from pd plot changed input event for debouncing to avoid issue where event.target changes by the time the debounced callback happens.

Ran WitWidget in all environments with many model and data types and found these issues and verified they are fixed with these changes.


jameswex:
> Thanks, do these changes work in TF1.x too?

yes they are backwards compatible

PR #2920, opened 2019-11-12
core: security validator middleware

stephanwlee:
http_util enforces good practice in term of security to prevent content
sniffing and the content security policy. However, this is not
enforceable for other plugins.

With this middleware, we would like to tell developer to enhance the
security to protect users. When we go 3.0, we may convert the warns into
hard failures.

### DIFF
--- a/tensorboard/backend/security_validator.py
+++ b/tensorboard/backend/security_validator.py
@@ -36,11 +36,16 @@ _HTML_MIME_TYPE = "text/html"
 # Matches CSP rule like "strict-src foobar baz".
 _CSP_PATTERN = re.compile(r"^([a-zA-Z0-9-]+) (.*)")
 _CSP_DEFAULT_SRC = "default-src"
-# Whitelist of allowed CSP rules.
+# Whitelist of allowed CSP violations.
 _CSP_IGNORE = {
+    # Allow TensorBoard to be iframed.
     "frame-ancestors": ["*"],
+    # Polymer-based code uses unsafe-inline.
     "style-src": ["'unsafe-inline'", "data:"],
+    # Used in canvas
     "img-src": ["blob:", "data:"],
+    # Used by numericjs.
+    # TODO(stephanwlee): remove it eventually.
     "script-src": ["'unsafe-eval'"],
 }
 
--- a/tensorboard/backend/security_validator.py
+++ b/tensorboard/backend/security_validator.py
@@ -1,0 +1,156 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Validates responses and their security features"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import re
+
+from werkzeug import wrappers
+from werkzeug.datastructures import Headers
+from werkzeug import http
+
+from tensorboard.util import tb_logging
+
+Policy = collections.namedtuple("Poicy", ["name", "rules"])
+
+logger = tb_logging.get_logger()
+
+_HTML_MIME_TYPE = "text/html"
+_CSP_PATTERN = re.compile(r"^([a-z]+-[a-z]+)(.*)")
+_CSP_DEFAULT_SRC = "default-src"
+# Whitelist of allowed CSP rules.
+_CSP_IGNORE = {
+    "frame-ancestors": "*",
+    "style-src": ["'unsafe-inline'", "data:"],
+    "img-src": ["blob:", "data:"],
+    "script-src": ["'unsafe-eval'"],
+}
+
+
+class SecurityValidatorMiddleware(object):
+  """WSGI middleware validating security on response.
+
+  It validates:
+  - responses have Content-Type
+  - responses have X-Content-Type-Options: nosniff
+  - text/html responses have CSP header.
+
+  Instances of this class are WSGI applications (see PEP 3333).
+  """
+
+  def __init__(self, application):
+    """Initializes an `SecurityValidatorMiddleware`.
+
+    Args:
+      application: The WSGI application to wrap (see PEP 3333).
+    """
+    self._application = application
+
+  def __call__(self, environ, start_response):
+
+    def start_resposne_proxy(status, headers, exc_info=None):
+      self._validate_headers(headers)
+      return start_response(status, headers, exc_info)
+
+    return self._application(environ, start_resposne_proxy)
+
+  def _validate_headers(self, headers_list):
+    headers = Headers(headers_list)
+    self._validate_content_type(headers)
+    self._validate_x_content_type_options(headers)
+    self._validate_csps(headers)
+
+  def _validate_content_type(self, headers):
+    if headers.get("Content-Type"):
+      return
+
+    self._maybe_raise_value_error("Content-Type is required on a Response")
+
+  def _validate_x_content_type_options(self, headers):
+    # Both none and empty string are
+    option = headers.get("X-Content-Type-Options")
+    if option == "nosniff":
+      return
+
+    self._maybe_raise_value_error(
+        'X-Content-Type-Options is required to be "nosniff"'
+    )
+
+  def _validate_csps(self, headers):
+    mime_type, _ = http.parse_options_header(headers.get("Content-Type"))
+    if mime_type != _HTML_MIME_TYPE:
+      return
+
+    csp_texts = headers.get_all("Content-Security-Policy")
+    policies = []
+
+    for csp_text in csp_texts:
+      policies += self._parse_csp_text(csp_text)
+
+    self._validate_csp_policies(policies)
+
+  def _validate_csp_policies(self, policies):
+    has_default_src = False
+    violations = []
+
+    for name, rules in policies:
+      has_default_src = has_default_src or name == _CSP_DEFAULT_SRC
+
+      for rule in rules:
+        if rule in _CSP_IGNORE.get(name, []):
+          # There are cases where certain values are legitimate.
+          continue
+
+        if (
+            rule == "'self'" or rule == "'none'" or rule.startswith("https:")
+            or rule.startswith("'sha256-")
+        ):
+          continue
+
+        msg = "Illegal Content-Security-Policy for {name}: {rule}".format(
+            name=name, rule=rule
+        )
+        violations.append(msg)
+
+    if not has_default_src:
+      violations.append("Requires default-src for Content-Security-Policy")
+
+    if violations:
+      self._maybe_raise_value_error("\n".join(violations))
+
+  def _parse_csp_text(self, csp_text):
+    csp_srcs = csp_text.split(";")
+    policies = []
+    for csp_src in csp_srcs:
+      match = _CSP_PATTERN.match(csp_src)
+      if not match or len(match.groups()) < 2:
+        continue
+      name, rules = _CSP_PATTERN.match(csp_src).groups()
+      policies.append(
+          Policy(
+              name=name,
+              rules=[rule.strip() for rule in rules.split(" ") if rule]
+          )
+      )
+
+    return policies
+
+  def _maybe_raise_value_error(self, error_msg):
+    logger.warn("In 3.0, this warning will become an error:\n%s" % error_msg)
+    # TODO(3.x): raise a value error.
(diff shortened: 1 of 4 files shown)


tensorboard/backend/security_validator.py
```diff
@@ -0,0 +1,156 @@
...
+import collections
+import re
+
+from werkzeug import wrappers
+from werkzeug.datastructures import Headers
+from werkzeug import http
+
+from tensorboard.util import tb_logging
+
+Policy = collections.namedtuple("Poicy", ["name", "rules"])
```
nfelt: Overall if we're writing something general-purpose here (for parsing any CSP that a plugin emits), it would be nice to use the terminology from the spec: https://www.w3.org/TR/CSP/#framework-directives

So in particular the policy would actually be the entire value of the CSP header.  I think this object would actually be what the spec calls a Directive, which has a directive name (fine as is) and then a directive value.  The directive value is always a set of strings and it looks like this specifically a "source list" for almost all directives, so we could perhaps just call it `source_list`?

tensorboard/backend/security_validator_test.py
```diff
@@ -0,0 +1,193 @@
...
+
+logger = tb_logging.get_logger()
+
+_WARN_PREFIX = "In 3.0, this warning will become an error:\n"
+
+
+def create_headers(
+    content_type="application/json",
+    x_content_type_options="nosniff",
+    content_secrurity_policy=""
```
nfelt: typo: security

tensorboard/backend/security_validator_test.py
```diff
@@ -0,0 +1,193 @@
...
+      }
+  )
+
+
+class SecurityValidatorMiddlewareTest(tb_test.TestCase):
+  """Tests for `SecurityValidatorMiddleware`."""
+
+  def setUp(self):
+    super(SecurityValidatorMiddlewareTest, self).setUp()
+    self.get_header = lambda req: {}
```
nfelt: Does this need to be a function?  It looks like it never actually uses the `req` argument so it could just be a property on the test case that we set.

Also optional but IMO a bit nicer if we can avoid the dependency on manipulating test case state indirectly to change the response.  What if we just passed in the CSP policy to return as a request header?  We'd implement the app request handler to just reflect that value back out, and then `make_request_and_maybe_assert_warn()` would just pass the headers to the call to `server.get()`.

tensorboard/backend/security_validator.py
```diff
@@ -0,0 +1,156 @@
...
+      policies.append(
+          Policy(
+              name=name,
+              rules=[rule.strip() for rule in rules.split(" ") if rule]
+          )
+      )
+
+    return policies
+
+  def _maybe_raise_value_error(self, error_msg):
```
nfelt: Could be a module-level function since it doesn't depend on self?

tensorboard/backend/security_validator_test.py
```diff
@@ -0,0 +1,193 @@
...
+    mock.patch.stopall()
+
+  def make_request_and_maybe_assert_warn(
+      self, headers, expected_warn_substr=None
+  ):
+    self.get_header = lambda req: headers
+    self.server.get("")
+
+    if expected_warn_substr is None:
+      self.mock_warn.assert_not_called()
```
nfelt: If we're going to assert this on the same mock we probably should reset the mock at the top of this function, so that we don't have an ordering dependency between calls to this helper (e.g. right now wouldn't the test fail if somebody re-ordered the assertions so that the failing assertion was first in the test method?).

That said IMO it would be better to just use a fresh mock each time for isolation (versus resetting the mock over and over), and to not patch the logger (since the `mock_warn.assert_not_called()` would fail if we were to start emitting an unrelated warning from some other dependency) but instead to just mock the call to `_maybe_raise_value_error()` since we already have that separated out.  So we could do something like:

```python
with mock.patch.object(security_validator, _maybe_raise_value_error) as mock_error:
  self.server.get("")
mock_error.assert_...
```

tensorboard/backend/security_validator_test.py
```diff
@@ -0,0 +1,193 @@
...
+    )
+
+    headers = create_headers(
+        content_type="text/html; charset=UTF-8",
+        content_secrurity_policy=(
+            "script-src * 'sha256-foo';"
+            "style-src http://google.com"
+        )
+    )
+    headers.add(
```
nfelt: If the point here is to test multiple CSP headers, IMO would be better to have a separate test method for this, both since it's testing a slightly different code path and since this test method is already quite long.

tensorboard/backend/security_validator.py
```diff
@@ -0,0 +1,156 @@
...
+  def _validate_csps(self, headers):
+    mime_type, _ = http.parse_options_header(headers.get("Content-Type"))
+    if mime_type != _HTML_MIME_TYPE:
+      return
+
+    csp_texts = headers.get_all("Content-Security-Policy")
+    policies = []
+
+    for csp_text in csp_texts:
+      policies += self._parse_csp_text(csp_text)
```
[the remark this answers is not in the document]
stephanwlee: Wouldn't that create a list of list?

tensorboard/backend/security_validator.py
```diff
@@ -0,0 +1,156 @@
...
+    for name, rules in policies:
+      has_default_src = has_default_src or name == _CSP_DEFAULT_SRC
+
+      for rule in rules:
+        if rule in _CSP_IGNORE.get(name, []):
+          # There are cases where certain values are legitimate.
+          continue
+
+        if (
+            rule == "'self'" or rule == "'none'" or rule.startswith("https:")
```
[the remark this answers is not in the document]
stephanwlee: > Will that be enforced somewhere? Just requiring https doesn't hurt but doesn't really do all that much.

I want to start with just https. Would you be happier if we start with allowing gstatic and certain google domains first then ease up the whitelist upon request?

tensorboard/backend/security_validator_test.py
stephanwlee: > What if we just passed in the CSP policy to return as a request header

I tried it and it was surprisingly hard because (1) content-type is auto inferred so reflection requires logic and (2) duplicative header name is omitted on `EnvironHeaders` making it more difficult to test the case. 

I took a stab at improving it.

tensorboard/backend/security_validator.py
```diff
@@ -0,0 +1,156 @@
...
+from tensorboard.util import tb_logging
+
+Policy = collections.namedtuple("Poicy", ["name", "rules"])
+
+logger = tb_logging.get_logger()
+
+_HTML_MIME_TYPE = "text/html"
+_CSP_PATTERN = re.compile(r"^([a-z]+-[a-z]+)(.*)")
+_CSP_DEFAULT_SRC = "default-src"
+# Whitelist of allowed CSP rules.
```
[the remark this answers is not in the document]
stephanwlee: > In particular it should be enough to understand what happens if: 1) directive is not one of the keys here, 2) directive is a key, but the set of given values a) is empty...

I think this question stemmed from my poor wording. I now call it "whitelist of csp *violation*". If the directive is non-violation, middleware will be happy,

tensorboard/backend/security_validator.py
```diff
@@ -27,29 +27,44 @@
...
-Policy = collections.namedtuple("Poicy", ["name", "rules"])
+# Loosely follow vocabulary from https://www.w3.org/TR/CSP/#framework-directives.
+Directive = collections.namedtuple("Directive", ["name", "value"])
 
 logger = tb_logging.get_logger()
 
 _HTML_MIME_TYPE = "text/html"
-_CSP_PATTERN = re.compile(r"^([a-z]+-[a-z]+)(.*)")
+# Matches CSP rule like "strict-src foobar baz".
+_CSP_PATTERN = re.compile(r"^([a-zA-Z0-9-]+) (.*)")
```
nfelt: Strictly speaking the space following the directive name shouldn't be required if the value is empty (for directives with no value), so this is also a bit of a deviation since it always requires the space.

tensorboard/backend/security_validator.py
```diff
@@ -101,56 +115,81 @@ def _validate_csps(self, headers):
...
-      name, rules = _CSP_PATTERN.match(csp_src).groups()
-      policies.append(
-          Policy(
-              name=name,
-              rules=[rule.strip() for rule in rules.split(" ") if rule]
-          )
-      )
 
-    return policies
+      name, values = _CSP_PATTERN.match(csp_src).groups()
```
nfelt: Can this just use `match` instead of re-invoking `match()`?

tensorboard/backend/security_validator.py
```diff
@@ -101,56 +115,81 @@ def _validate_csps(self, headers):
...
-  def _maybe_raise_value_error(self, error_msg):
-    logger.warn("In 3.0, this warning will become an error:\n%s" % error_msg)
-    # TODO(3.x): raise a value error.
+      # Step 2.6
+      for value in values.split(" "):
+        if not value:
+          continue
+
+        # Step 2.7
+        directive = Directive(name=name, value=value)
```
nfelt: This isn't exactly the step as described in the spec - there should be a single directive emitted with the name `name` for which the value of the directive is a set of tokens.  So basically just `policies.append(Directive(name=name, value=values.split()))`.

Also a nit but I think it should be `split()` not `split(" ")` to split on chunks of whitespace rather than only on spaces (and then we don't have to filter out empty strings in the case of consecutive spaces either).

tensorboard/backend/security_validator.py
```diff
@@ -101,56 +115,81 @@ def _validate_csps(self, headers):
...
+        # Step 2.8
+        policies.append(directive)
+
+    return policies
```
nfelt: Optional, but IMO better to call this just "policy" or "directives" if we want a plural, since it conceptually represents a single policy.

tensorboard/backend/security_validator_test.py
```diff
@@ -40,48 +41,51 @@
...
-    )
-    app = security_validator.SecurityValidatorMiddleware(app)
-    self.server = werkzeug_test.Client(app, werkzeug.BaseResponse)
-    self.mock_warn = mock.patch.object(logger, "warn").start()
-
-  def tearDown(self):
-    super(SecurityValidatorMiddlewareTest, self).tearDown()
-    mock.patch.stopall()
-
+  @mock.patch.object(logger, "warn")
```
nfelt: I still think if you're going to patch logger itself it would be best to scope it down as tightly as possible to just the call to `server.get("")` (by using a context manager).  Otherwise if anything in this method ends up emitting a warning, like the construction of `SecurityValidatorMiddleware`, it would fail the test.

tensorboard/backend/security_validator_test.py
nfelt: Sure, LGTM as is.

tensorboard/backend/security_validator.py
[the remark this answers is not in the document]
nfelt: Ah that's true, my bad.

tensorboard/backend/security_validator.py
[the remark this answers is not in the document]
nfelt: So I'm ok if we have the default policy for local / OSS TensorBoard allow any HTTPS site, but it seems a bit broad for TensorBoard.dev?  Restricting that to just the known domains feels a little bit safer.

tensorboard/backend/security_validator.py
stephanwlee: I instead decided to use `split`. PTAL.

stephanwlee:
(Since I can't easily find the comment) about the enforcing stricter rule on allowed domain on top of `https:`, while I agree that is better, here are my two concerns:
1. does not have to be done in this change because this change will emit only warn anyways, and
2. I do not yet know how it can be configured: I don't want users to pass the configuration is flags but I also don't want to trust plugin authors blindly. We can forgo configurability and just hardcode a reasonable list here, too.

tensorboard/backend/security_validator.py
```diff
@@ -125,29 +122,32 @@ def _validate_csp_policies(self, policies):
 
     for directive in policies:
       name = directive.name
-      value = directive.value
-      has_default_src = has_default_src or name == _CSP_DEFAULT_SRC
+      for directive in directive.value:
```
nfelt: Sorry to ask for another change, but this should really still be called `value` or perhaps `source` (since it's a "source list") rather than `directive` since in the spec `directive` refers to the name and value together (basically the entire thing between semicolons) rather than only to the values like `'self'`.

nfelt:
Re the domains, it's fine to punt for now, sure.  I don't really mind if by default we allow any domain that's HTTPS, but mostly I was hoping there would be a configuration option that lets us lock down the allowed list (i.e. be more restrictive than the default) for specific TensorBoard instances that we want to be more secure.

IMO it would be okay to make that a flag, but we could I suppose also use an environment variable or some other more indirect mechanism if you wanted it to be less user-visible.

tensorboard/backend/security_validator.py
stephanwlee: Changed it to `value` since future me would find `for source in directive.value` odd. 

Thanks for the feedback :D

stephanwlee:
> IMO it would be okay to make that a flag, but we could I suppose also use an environment variable or some other more indirect mechanism if you wanted it to be less user-visible.

Noted. I can do a follow up that configures whitelist via env variable.

PR #2925, opened 2019-11-12
Make --reload_multifile default value falsy None instead of False

nfelt:
This changes the default value from `False` to `None`, which doesn't change the semantics or interpretation (since we check it with `if flags.reload_multifile` so both False and falsy-None will fail that check), but does mean that other code can distinguish whether we're truly using the default or whether `--reload_multifile=false` was explicitly passed on the command line.  This means we can effectively change the flag from opt-in to opt-out in some contexts by setting the flag value to `True` if it's still `None` after parsing.


nfelt:
Yes, the plan would be that after validating this a bit more we'll enable this by default, and then eventually we'd probably remove the flag entirely and just use the new logic unconditionally.

Issue #2926, opened 2019-11-12
Scalars tutorial is missing a TF import


GalOshri:
Hello! I think the [Scalars tutorial](https://www.tensorflow.org/tensorboard/scalars_and_keras) is missing a TF import.

Repro steps:
1. Go to the [Scalars tutorial](https://colab.research.google.com/github/tensorflow/tensorboard/blob/master/docs/scalars_and_keras.ipynb) in Colab
2. Run the first two code cells

![image](https://user-images.githubusercontent.com/2601900/68717740-681bbd80-055c-11ea-9ac2-6662e9319fc5.png)

This is solved by adding `import tensorflow as tf`. 

Additional stylistic issue: `%tensorflow_version 2.x` appears twice. I believe this will break outside of Colab (as the second one is not in a `try` `except`.

psybuzz:
Thanks for the report.  This is really unfortunate, it looks like the tutorials "Scalars and metrics", "Image data", and "TensorBoard in Notebooks" all lack this import, giving the same error when running the notebook.

ymodak:
Closing this issue since the associated PR has been merged and the issue is fixed. Thanks!

PR #2927, opened 2019-11-13
docs: fix notebooks with broken imports and images

psybuzz:
See #2926

This adds some missing "import tensorflow as tf" statements required to make a few of the notebooks work.  The "TensorBoard in Notebooks" notebook had relative image links, which work on the [docs site](https://www.tensorflow.org/tensorboard/tensorboard_in_notebooks), but fail if you inspect the notebook on its own.  This makes it use absolute URLs, similar to the rest of our notebooks, so that they show up locally.

Manually uploaded the updated notebooks to Colab, Ran all cells, and checked that the output looked normal.
https://colab.sandbox.google.com/drive/1L3oNRFsGoaTt-hQvd4A0Mq5fxoYZlESc#scrollTo=6pck56gKReON
https://colab.sandbox.google.com/drive/16v6X5MupV_0qLhzIJDZ-HP4DO1VKH-N8#scrollTo=k6CV7dy-oJZu
https://colab.sandbox.google.com/drive/1KzmS190G-X8rWDadvaRkhJ8iJ1cUzE0z


docs/image_summaries.ipynb
```diff
@@ -154,6 +154,7 @@
         "import matplotlib.pyplot as plt\n",
         "import numpy as np\n",
         "import sklearn.metrics\n",
+        "import tensorflow as tf\n",
```
nfelt: Maybe put this above `from tensorflow import keras` just to group the related imports?

Same goes for the next one as well.

psybuzz: Done for both

psybuzz:
> Do we want to get rid of the non-try-catch-protected redundant uses of `%tensorflow_version` as well, while we're at it?

I found 3 sites that were redundant.  Done.

PR #2933, opened 2019-11-14
build: create JavaScript file from Vulc

stephanwlee:
This change adds an ability to create JavaScript file from
Vulcanization. This is in response to code debt we accrued while adding
Content Security Policy. We decided to keep CSP simple by, instead of
computing shasum on individual script block, serving JavaScript from our
server and have "script-src 'self'".

This change does not yet remove an ability to use shasum but it will be
quickly deprecated afterwards.

Tested the change internally by making changes directly:
cl/280296740

### DIFF
--- a/tensorboard/defs/vulcanize.bzl
+++ b/tensorboard/defs/vulcanize.bzl
@@ -28,9 +28,6 @@ def _tensorboard_html_binary(ctx):
   point at). The hashes are delimited by newline.
   """
 
-  if ctx.attr.extract_script and not ctx.attr.script_path:
-    fail("script_path is required when extract_script is True.")
-
   deps = unfurl(ctx.attr.deps, provider="webfiles")
   manifests = depset(order="postorder")
   files = depset()
@@ -63,10 +60,9 @@ def _tensorboard_html_binary(ctx):
       arguments=([ctx.attr.compilation_level,
                   "true" if ctx.attr.compile else "false",
                   "true" if ctx.attr.testonly else "false",
-                  "true" if ctx.attr.extract_script else "false",
                   ctx.attr.input_path,
                   ctx.attr.output_path,
-                  ctx.attr.script_path,
+                  ctx.attr.js_path,
                   ctx.outputs.html.path,
                   ctx.outputs.js.path,
                   ctx.outputs.shasum.path,
@@ -136,10 +132,9 @@ tensorboard_html_binary = rule(
         "compilation_level": attr.string(default="ADVANCED"),
         "input_path": attr.string(mandatory=True),
         "output_path": attr.string(mandatory=True),
-        # Required when extract_script is True.
-        "script_path": attr.string(),
+        # If specified, it extracts scripts into {name}.js and inserts <script src="{js_path}">.
+        "js_path": attr.string(),
         "compile": attr.bool(),
-        "extract_script": attr.bool(),
         "data": attr.label_list(allow_files=True),
         "deps": attr.label_list(aspects=[closure_js_aspect], mandatory=True),
         "external_assets": attr.string_dict(default={"/_/runfiles": "."}),
--- a/tensorboard/defs/vulcanize.bzl
+++ b/tensorboard/defs/vulcanize.bzl
@@ -27,6 +27,10 @@ def _tensorboard_html_binary(ctx):
   script elements (sources inside element and content of JavaScript src they
   point at). The hashes are delimited by newline.
   """
+
+  if ctx.attr.extract_script and not ctx.attr.script_path:
+    fail("script_path is required when extract_script is True.")
+
   deps = unfurl(ctx.attr.deps, provider="webfiles")
   manifests = depset(order="postorder")
   files = depset()
@@ -54,14 +58,17 @@ def _tensorboard_html_binary(ctx):
           ignore_regexs_file_set,
       ]).to_list(),
       tools=jslibs,
-      outputs=[ctx.outputs.html, ctx.outputs.shasum],
+      outputs=[ctx.outputs.html, ctx.outputs.js, ctx.outputs.shasum],
       executable=ctx.executable._Vulcanize,
       arguments=([ctx.attr.compilation_level,
                   "true" if ctx.attr.compile else "false",
                   "true" if ctx.attr.testonly else "false",
+                  "true" if ctx.attr.extract_script else "false",
                   ctx.attr.input_path,
                   ctx.attr.output_path,
+                  ctx.attr.script_path,
                   ctx.outputs.html.path,
+                  ctx.outputs.js.path,
                   ctx.outputs.shasum.path,
                   ignore_regexs_file_path] +
                  [f.path for f in jslibs.to_list()] +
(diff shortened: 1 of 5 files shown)


tensorboard/defs/vulcanize.bzl
```diff
@@ -27,6 +27,10 @@ def _tensorboard_html_binary(ctx):
   script elements (sources inside element and content of JavaScript src they
   point at). The hashes are delimited by newline.
   """
+
+  if ctx.attr.extract_script and not ctx.attr.script_path:
+    fail("script_path is required when extract_script is True.")
```
wchargin: Conversely, it doesn’t make sense for `script_path` to be provided
unless `extract_script` is set, right? Maybe consider just letting
`extract_script` or `output_script` be the output path rather than
requiring two attributes to be in sync:

```python
tensorboard_html_binary(
    name = "index",
    compile = True,
    input_path = "/tensorboard.html",
    output_path = "/index.html",
    output_script = "/index.js",
    deps = [":tensorboard"],
)
```

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -793,22 +802,73 @@ private static void combineScriptElements(Document document) {
...
+        // script element that remains are the ones with src that is absolute or annotated with
+        // `jscomp-ignore`. They must resolve from the root because those srcs are rootified.
+        Webpath webpathSrc = Webpath.get(src);
+        Webpath webpath = Webpath.get("/").resolve(webpathSrc).normalize();
+        if (isAbsolutePath(webpathSrc)) {
+          if (script.hasAttr("defer") || script.hasAttr("async")) {
+            continue;
+          }
+          throw new IllegalArgumentException(
+              "Script refers to a remote resource ("
```
wchargin: Great idea to include this check, and nice message.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -720,25 +730,18 @@ private static boolean shouldIgnoreUri(String uri) {
...
-  //   <script>A,B</script>
-  //   <script src="srcful1"></script>
-  //   <script src="srcful2"></script>
-  //   <script>C,D</script>
-  //   <script src="srcful3"></script>
-  //   <script>E</script>
+  /**
+   * Combine content of script tags into a group. To guarantee the correctness, it only groups
+   * content of `src`-less scripts between `src`-full scripts. The last combination gets inserted at
+   * the end of the document. e.g., <script>A</script> <script>B</script> <script
```
wchargin: Javadoc is interpreted as HTML, so these literal `<script>` tags will
get clobbered (sanitized away, hopefully). You can use `{@code ...}`
markup to write them safely:

```java
  /**
   * Combine content of script tags into a group. To guarantee the correctness, it only groups
   * content of `src`-less scripts between `src`-full scripts. The last combination gets inserted at
   * the end of the document. e.g., {@code <script>A</script> <script>B</script> <script
   * src="srcful1"></script> <script src="srcful2"></script> <script>C</script> <script>D</script>
   * <script src="srcful3"></script> <script>E</script>} is compiled as {@code <script>A,B</script>
   * <script src="srcful1"></script> <script src="srcful2"></script> <script>C,D</script> <script
   * src="srcful3"></script> <script>E</script>}
   *
```

You can also see that the lexical highlighting here and in your editor
makes it much easier to see where the `<script>` tags stop and start.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -388,11 +394,12 @@ private static Node inlineScript(Node node) throws IOException {
       String code = new String(Files.readAllBytes(getWebfile(href)), UTF_8);
       code = code.replace("</script>", "</JAVA_SCRIIIIPT/>");
       code = INLINE_SOURCE_MAP_PATTERN.matcher(code).replaceAll("");
-      result = replaceNode(
-          node,
-          new Element(Tag.valueOf("script"), node.baseUri(), node.attributes())
-              .appendChild(new DataNode(code, node.baseUri()))
-              .removeAttr("src"));
+      result =
```
wchargin: Perhaps I’m misreading this, but it looks like this diff hunk (and the
ones above, and some below) are no-ops; mind reverting from this commit?
I’d be happy to approve them separately, but it’s hard to tell what
changes are important here.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -793,22 +802,73 @@ private static void combineScriptElements(Document document) {
...
+                  + " correctness of execution, please make sure it is async-able or defer-able:"
+                  + script.outerHtml());
+        } else if (!webfiles.containsKey(webpath)) {
+          throw new FileNotFoundException(
+              "Expected webfiles for " + webpath + " to exist. Related: " + script.outerHtml());
+        }
+        sourceContent = new String(Files.readAllBytes(webfiles.get(webpath)), UTF_8);
+      }
+
+      sourcesBuilder.append(sourceContent).append("\n");
```
wchargin: Should we include a comment here indicating the provenance of the file,
for debugging purposes? Or are the `~~~WEBPATH~~~` (or whatever)
comments included here already?

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -793,22 +802,73 @@ private static void combineScriptElements(Document document) {
...
+
+    return sourcesBuilder.toString();
+  }
+
+  private static String extractAndTransformJavaScript(Document document, Webpath scriptPath)
+      throws FileNotFoundException, IOException, IllegalArgumentException {
+    String scriptContent = extractScriptContent(document);
+
+    Element lastBody = Iterables.getLast(document.getElementsByTag("body"));
+    Element scriptTag = new Element(Tag.valueOf("script"), "");
```
wchargin: Half-nit, but calling an element a tag is a bit confusing, especially
when the rest of the code clearly and correctly distinguishes them;
consider `scriptElement` or just `script`?

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
stephanwlee: Sorry, I used a formatter (internal; should have applied it to diff only) and it made more changes than I intended. Reverted changes.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
stephanwlee: Thinking out loud here.

This is combining already inlined scripts so they do not have `webpath` or name associated with them. We can put indices but I am not sure how useful that is. Also, while compiling, we've already put bunch of `sourceURL=` for souremap. What kind of delimitation would help us debug issues?

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
wchargin: Ah, sorry, didn’t realize that there actually was precedent in this
file. Either’s fine then; thanks.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
wchargin: Okay, source map URLs should be sufficient. Thanks for considering!

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
wchargin: Thanks! Looks like there are a few more remaining (lines 247, 519, 820).

Issue #2935, opened 2019-11-14
Polymer: disable data binding on style attribute


stephanwlee:
For security enhanced TensorBoard, we use [polymer-resin](https://github.com/Polymer/polymer-resin). It fails to correctly sanitize data-bound style attribute and completely omits the style.

Very simple regex found few 7 instances of the violations:
```
tensorboard/plugins/text/tf_text_dashboard/tf-text-loader.html
39:      style="border-color: [[_runColor]]"

tensorboard/plugins/graph/tf_graph_debugger_data_card/tf-graph-debugger-data-card.html
246:            style="background:[[item.background_color]]"

tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.html
600:                    style="fill:[[item.color]]"

tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-steps-selector.html
31:            style="background:[[_computeColorForRun(run)]];"

tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-card.html
75:            style="background: [[_computeRunColor(run)]];"

tensorboard/plugins/projector/vz_projector/vz-projector-legend.html
57:        <div class="shape" style="background-color: [[item.color]];"></div>
59:        <div class="info" style="color: [[item.color]];">[[item.count]]</div>
```

PR #2936, opened 2019-11-14
polymer: (partial) remove data binding on style

stephanwlee:
More details here: https://github.com/tensorflow/tensorboard/issues/2935

Enforcing resin causes several styles in TensorBoard to misbehave
because it fails to correctly sanitize content of the `style` attribute.
We use different techniques (such as CSS variable and imperative
JavaScript) to workaround the issue.

This change only addresses the issues on scalars and custom scalars
dashboard.

### DIFF
--- a/tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.html
+++ b/tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.html
@@ -195,10 +195,7 @@ limitations under the License.
         <iron-collapse opened="[[_matchesListOpened]]">
           <div id="matches-list">
             <template is="dom-repeat" items="[[_seriesNames]]" as="seriesName">
-              <div
-                class="match-list-entry"
-                style="color: [[_determineColor(_colorScale, seriesName)]];"
-              >
+              <div class="match-list-entry">
                 <span class="match-entry-symbol">
                   [[_determineSymbol(_nameToDataSeries, seriesName)]]
                 </span>
@@ -453,6 +450,9 @@ limitations under the License.
         // Clear the data series when the tag filter changes.
         '_refreshDataSeries(_tagFilter)',
       ],
+      listeners: {
+        'dom-change': '_matchListEntryColorUpdated',
+      },
       reload() {
         this.$.loader.reload();
       },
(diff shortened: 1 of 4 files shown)


stephanwlee:
@psybuzz ping :)

tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.html
```diff
@@ -453,6 +450,9 @@
         // Clear the data series when the tag filter changes.
         '_refreshDataSeries(_tagFilter)',
       ],
+      listeners: {
+        'dom-change': '_matchListEntryColorUpdated',
```
psybuzz: It's unfortunate that we would have to
- Rely on the fact that <dom-repeat> is on the path of a dom-change event (I can't find in Polymer docs that this is intended)
- QuerySelect the list items

This is all because we cannot dynamically build styles, so this approach is to manipulate elements directly.  I'd like to propose a still-hacky change: Can we

- store the seriesName on the match-list-entry itself
- have an observer listen to `_seriesName` changes, then select the list entries and update color using the prop/attribute on the element?
Then at least we aren't tied to dom-repeat itemForElement or dom-change.

stephanwlee: No, we cannot. You cannot guarantee the presence of the DOM when the observer is called. The presence of DOM is predicated on two dom-ifs and one dom-repeat.

psybuzz: I see.  I'll have to remember about 'dom-change' for the future :/

This works, but is hacky, let's migrate to Angular as soon as we can.

stephanwlee: Just so we are on the same page, the `dom-change` is documented here: 
https://polymer-library.polymer-project.org/2.0/api/elements/Polymer.DomRepeat#event-dom-change

We rely on `path` because inside shadow DOM, `target` becomes very weird and is not useful for our use case. Also see, https://polymer-library.polymer-project.org/2.0/api/elements/Polymer.DomIf#event-dom-change

PR #2941, opened 2019-11-15
uploader: include metadata in `list` output

wchargin:
Summary:
This commit teaches the uploader to display experiment metadata included
in `StreamExperiments` responses by supported servers. For servers
without this support, the change is a backward-compatible no-op.

The format is intentionally undocumented and not under any compatibility
guarantees, but is designed to be easily parseable for ad hoc usage. For
instance, this simple one-liner finds experiments with lots of points so
that the user can delete them:

```
tensorboard dev list |
    awk '$1 == "Id" { id = $2 } $1 == "Scalars" && $2 > 1000 { print id }'
```

Test Plan:
Running against current prod, which does not yet support the new RPCs,
the behavior is unchanged:

```
$ bazel run //tensorboard -- dev list
https://tensorboard.dev/experiment/IAVF94GPSWWBTvonQe4kgQ/
https://tensorboard.dev/experiment/LiQNYkOHRSGEWj42xtgtjA/
<snip>
Total: 12 experiment(s)
```

Running against a local server with support for the new RPCs, we see
lots of additional data (tested on both Linux and Windows):

```
$ bazel run //tensorboard -- dev --origin http://localhost:8080 --grpc_creds_type ssl_dev list
http://localhost:8080/experiment/WtPawgPIQXi2SZ1fQszOFA/
	Id         WtPawgPIQXi2SZ1fQszOFA
	Created    2019-11-25 10:30:18 (23 seconds ago)
	Updated    2019-11-25 10:30:39 (just now)
	Scalars    18814
	Runs       21
	Tags       7
http://localhost:8080/experiment/jD7Qc7l6S8Wy5gWKYTAHOA/
	Id         jD7Qc7l6S8Wy5gWKYTAHOA
	Created    2019-11-13 18:32:06
	Updated    2019-11-13 18:32:06
	Scalars    0
	Runs       0
	Tags       0
http://localhost:8080/experiment/do8uvvEOSNWOUEANmQIprQ/
	Id         do8uvvEOSNWOUEANmQIprQ
	Created    2019-11-13 18:15:25
	Updated    2019-11-13 18:15:37
	Scalars    3208
	Runs       8
	Tags       4
<snip>
Total: 9 experiment(s)
```

Also tested that the `tensorboard dev export` service still works
against both old and new servers.

wchargin-branch: uploader-list-metadata


tensorboard/uploader/uploader_main.py
```diff
@@ -538,6 +562,23 @@ def _die(message):
...
+def _format_time(timestamp_pb):
+  # Add and subtract a day for <https://bugs.python.org/issue29097>,
+  # which breaks early datetime conversions on Windows for small
+  # timestamps.
+  dt = datetime.datetime.fromtimestamp(timestamp_pb.seconds + 86400)
+  dt = dt - datetime.timedelta(seconds=86400)
+
+  ago = datetime.datetime.now().replace(microsecond=0) - dt
+  if ago < datetime.timedelta(seconds=5):
+    return "just now"
```
nfelt: How would you feel about providing the exact timestamp in all cases and maybe just add the human-friendly interpretations of recent times in parentheses after the timestamp or something?  Maybe just my pet peeve, but as a user I'm often a little annoyed when there's a humanized timestamp with no way to access the exact value (e.g. via the title text or something).

wchargin: Totally agree; uncharacteristic of me to omit that. Will change.

GalOshri:
LGTM with the change mentioned above regarding "Created" and "Updated" formats. Consistency would be useful, and the second example ("23:42:10 ago") is somewhat confusing. 

Since we are not including a message on "you are using X% of your quota", let's make sure we have a bug to track that for the future?

GalOshri:
Also, thank you for adding this! This is super helpful.

wchargin:
Moved time formatting logic into a utility module so that it can be
tested. See updated PR description for revised output form:

```
	Created    2019-11-25 10:30:18 (23 seconds ago)
	Updated    2019-11-25 10:30:39 (just now)
```

Tested on both Linux and Windows against both prod and a dev API server
that supports the new functionality.

wchargin:
> Since we are not including a message on "you are using X% of your
> quota",

Correct; the backend doesn’t send down that information, and it would be
restrictive to hard-code a value (doesn’t give much flexibility to
change it later without a new uploader version).

> let's make sure we have a bug to track that for the future?

Opened #2969.

PR #2944, opened 2019-11-19
backend: document behavior on empty experiment IDs

wchargin:
Summary:
Empty experiment IDs are handled just like non-empty ones. This is an
edge case that was being handled as intended, but should have a test to
clarify that this is in fact the desired behavior. This commit adds such
a test.

Test Plan:
Newly added test passes.

wchargin-branch: backend-empty-eid-test

### DIFF
--- a/tensorboard/backend/experiment_id_test.py
+++ b/tensorboard/backend/experiment_id_test.py
@@ -76,6 +76,10 @@ class ExperimentIdMiddlewareTest(tb_test.TestCase):
     response = self.server.get("/experiment/123/x/y")
     self._assert_ok(response, eid="123", path="/x/y", script="/experiment/123")
 
+  def test_with_empty_experiment_empty_path(self):
+    response = self.server.get("/experiment/")
+    self._assert_ok(response, eid="", path="", script="/experiment/")
+
   def test_with_empty_experiment_root_path(self):
     response = self.server.get("/experiment//")
     self._assert_ok(response, eid="", path="/", script="/experiment/")
--- a/tensorboard/backend/experiment_id_test.py
+++ b/tensorboard/backend/experiment_id_test.py
@@ -76,6 +76,14 @@ class ExperimentIdMiddlewareTest(tb_test.TestCase):
     response = self.server.get("/experiment/123/x/y")
     self._assert_ok(response, eid="123", path="/x/y", script="/experiment/123")
 
+  def test_with_empty_experiment_root_path(self):
+    response = self.server.get("/experiment//")
+    self._assert_ok(response, eid="", path="/", script="/experiment/")
+
+  def test_with_empty_experiment_sub_path(self):
+    response = self.server.get("/experiment//x/y")
+    self._assert_ok(response, eid="", path="/x/y", script="/experiment/")
+
 
 if __name__ == "__main__":
   tb_test.main()


tensorboard/backend/experiment_id_test.py
```diff
@@ -76,6 +76,14 @@ def test_with_experiment_sub_path(self):
     response = self.server.get("/experiment/123/x/y")
     self._assert_ok(response, eid="123", path="/x/y", script="/experiment/123")
 
+  def test_with_empty_experiment_root_path(self):
+    response = self.server.get("/experiment//")
+    self._assert_ok(response, eid="", path="/", script="/experiment/")
```
stephanwlee: I quickly glossed over PEP-333 to see if trailing slash is strongly discouraged and it was not. Where can I find out more about good practice around SCRIPT_NAME and PATH_NAME?

tensorboard/backend/experiment_id_test.py
```diff
@@ -76,6 +76,14 @@ def test_with_experiment_sub_path(self):
     response = self.server.get("/experiment/123/x/y")
     self._assert_ok(response, eid="123", path="/x/y", script="/experiment/123")
 
+  def test_with_empty_experiment_root_path(self):
```
stephanwlee: Could you also document the `self.server.get("/experiment/")` case?

tensorboard/backend/experiment_id_test.py
wchargin: I have no authoritative source, only some conventions that seem to point
in this direction. For instance, Werkzeug’s [`request.script_root`][1]
and [`request.path`][2] both normalize to be of `/leading-slash` form,
so it’s clearer if the underlying environment is already in that form.
And `/leading-slash` paths compose: `"/a" + "/b" == "/a/b"`, whereas for
example `"/a/" + "/b/" == "/a//b/"` is not what was wanted.

I’ll grant that the particular case of `eid=""` is awkward because there
is an actual empty path component, so the first point about Werkzeug
doesn’t apply. For consistency, I still think that this behavior makes
sense, but I could be convinced that a different behavior is ketter.

[1]: https://werkzeug.palletsprojects.com/en/0.15.x/wrappers/#werkzeug.wrappers.BaseRequest.script_root
[2]: https://werkzeug.palletsprojects.com/en/0.15.x/wrappers/#werkzeug.wrappers.BaseRequest.path

tensorboard/backend/experiment_id_test.py
wchargin: Good point; yes. Done.

wchargin:
The travis-ci.com build failed for unrelated reasons and is stuck in
limbo, but the travis-ci.org build passed, so I’m merging this.

tensorboard/backend/experiment_id_test.py
wchargin: Okay, I did happen across some authoritative sources: the standard
library’s `wsgiref.validate` requires that `SCRIPT_NAME` and `PATH_INFO`
[start with a slash when non-empty][val], which combined with the
standard [URL reconstruction algorithm][url] implies the relevant
semantics around trailing slashes. This is further codified by the
semantics of [`wsgiref.util.shift_path_info`][spi]. (Again, this is all
a bit messier when considering empty path components, which the standard
library’s manipulators [normalize out entirely][nrm].)

[nrm]: https://github.com/python/cpython/blob/86d9933cc627c4232f9702ca0766713714ebbc53/Lib/wsgiref/util.py#L101
[spi]: https://docs.python.org/3.8/library/wsgiref.html#wsgiref.util.shift_path_info
[url]: https://docs.python.org/3.8/library/wsgiref.html#wsgiref.util.request_uri
[val]: https://github.com/python/cpython/blob/86d9933cc627c4232f9702ca0766713714ebbc53/Lib/wsgiref/validate.py#L39

stephanwlee: Insightful. Thanks.

As [written](https://github.com/python/cpython/blob/86d9933cc627c4232f9702ca0766713714ebbc53/Lib/wsgiref/validate.py#L39), it sounds possible to keep `PATH_INFO` empty while having trailing slash for `SCRIPT_NAME`. In my cursory reading, I cannot sense that it preferring a slash on `PATH_INFO` over a trailing slash on `SCRIPT_NAME`.

PR #2945, opened 2019-11-19
cleanup: set targeted dom-change listener

stephanwlee:
Instead of component-level listener for dom-change, we now put the
listener on the specific template (dom-if or dom-repeat) that we care
about.

### DIFF
--- a/tensorboard/components/tf_dashboard_common/tf-filterable-checkbox-list.html
+++ b/tensorboard/components/tf_dashboard_common/tf-filterable-checkbox-list.html
@@ -58,7 +58,11 @@ Properties out:
           No matches
         </div>
       </template>
-      <template is="dom-repeat" items="[[_itemsMatchingRegex]]">
+      <template
+        is="dom-repeat"
+        items="[[_itemsMatchingRegex]]"
+        on-dom-change="_synchronizeColors"
+      >
         <div class="item">
           <paper-checkbox
             checked$="[[_isChecked(item, selectionState.*)]]"
--- a/tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.html
+++ b/tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.html
@@ -194,7 +194,13 @@ limitations under the License.
       <template is="dom-if" if="[[_seriesNames.length]]">
         <iron-collapse opened="[[_matchesListOpened]]">
           <div id="matches-list">
-            <template is="dom-repeat" items="[[_seriesNames]]" as="seriesName">
+            <template
+              is="dom-repeat"
+              items="[[_seriesNames]]"
+              as="seriesName"
+              id="match-list-repeat"
+              on-dom-change="_matchListEntryColorUpdated"
+            >
               <div class="match-list-entry">
                 <span class="match-entry-symbol">
                   [[_determineSymbol(_nameToDataSeries, seriesName)]]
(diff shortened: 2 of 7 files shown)


tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.html
```diff
@@ -719,12 +722,14 @@
         );
       },
       _matchListEntryColorUpdated(event) {
```
psybuzz: We can now drop `event` param

tensorboard/components/tf_dashboard_common/tf-filterable-checkbox-list.html
```diff
@@ -58,7 +58,11 @@
           No matches
         </div>
       </template>
-      <template is="dom-repeat" items="[[_itemsMatchingRegex]]">
+      <template
+        is="dom-repeat"
+        items="[[_itemsMatchingRegex]]"
+        on-dom-change="_synchronizeColors"
```
psybuzz: Is this deep / Does this trigger when users check/uncheck a checkbox?
If so, maybe consider an observer: `_synchronizeColors(_itemsMatchingRegex)`

tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.html
```diff
@@ -719,12 +722,14 @@
...
-        const [maybeDomRepeat] = event.path;
-        if (maybeDomRepeat.nodeName !== 'DOM-REPEAT') return;
+        const domRepeat = this.$$('#match-list-repeat');
```
psybuzz: Curious why the change from `event.path` to this selector.  Is .path, .target not reliable?

stephanwlee: We only needed `event.path` because `dom-if` also emits the same event and we wanted to discern the events. Now that we have more localized listener, we simply don't need one.

tensorboard/components/tf_dashboard_common/tf-filterable-checkbox-list.html
stephanwlee: I do not follow. Why is that required?

psybuzz: It seems it may be unnecessary to call `_synchronizeColors` on every single dom change.  It depends on the items, but not on whether a checkbox is checked.  If a user checks a checkbox and triggers dom to change, then we don't actually need to call it.

tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.html
psybuzz: The `event.path` works before and after, just curious whether we gain something by switching to a querySelector.  Either way is fine with me, though

tensorboard/components/tf_dashboard_common/tf-filterable-checkbox-list.html
stephanwlee: > unnecessary to call _synchronizeColors on every single dom change.

this is the part I am not getting. `dom-change` is a feature of dom-repeat and dom-if. Are you saying paper-checkbox uses that and thus will fire bunch of events?

psybuzz: > dom-change is a special event of dom-repeat and dom-if

Ahh, things make sense then, I was wrong!
For my own understanding: so before, when a component (not dom-repeat/if) has a `listeners: {'dom-change': ...}`, that means it is listening specifically for events dispatched by dom-repeat/if.

Issue #2938, opened 2019-11-14
'probabilities' predict_output_tensor is confusing


htappen:
I noticed in https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/interactive_inference/witwidget/notebook/base.py#L510
that it expects a few hard-coded tensor names for output. A few issues here:

1) Could WIT just default to grabbing the first tensor in case predict_output_tensor isn't set? That will match many Keras models.
2) In the case there are multiples, could we have a smarter guess? Keras Sequential will use "output_1", or "dense_1" in many cases. Not sure what estimators will do.
3) Finally, if it's set incorrectly, could we have a better error message? Right now it's a black box called "probabilities"

We may want to make this a required param, too.

psybuzz:
+jameswex, would you be the best person to assign this WIT related issue to?

jameswex:
Thanks @psybuzz and @htappen . I'll take this and work with Henry offline to figure out the best solution

PR #2946, opened 2019-11-19
Improve WIT parsing of model results

jameswex:

If a model returns multiple predictions in a dictionary, make the parsing of this dictionary more fool-proof, improve error statements, and simplify parsing of binary classification results.

For both AI platform models and estimators, if the model returns a dictionary of results and the dictionary only has one key, use that result automatically instead of requiring the user to specify the key name of that only result.

Add more-detailed error string in the case of not being able to parse prediction results correctly.

Use repr(error) instead of str(error) to get better displayed errors in WIT from the backend.

If a classification model only returns a single class score, auto-convert that to be the class score for class 1 and set the class score for class 0 correctly. This matches up with how many binary classification models return results and simplifies code necessary to use them. It was done in a 100% backwards compatible way with what users were already doing to get around this issue.

* Screenshots of UI changes

N/A

Tested not needing to call set_predict_output_tensor anymore for models that only return a single result in a dictionary.

Tested exception case in notebooks to see better errors displayed.

Tested binary classifiers with adjustment code still work as before, and also they now work without needing the adjust_prediction override as well.

### DIFF
--- a/tensorboard/plugins/interactive_inference/utils/inference_utils.py
+++ b/tensorboard/plugins/interactive_inference/utils/inference_utils.py
@@ -832,7 +832,7 @@ def run_inference(examples, serving_bundle):
         if len(returned_keys) == 1:
           key_to_use = returned_keys[0]
         # Use default keys if necessary.
-        elif serving_bundle.model_type == == 'classification':
+        elif serving_bundle.model_type == 'classification':
           key_to_use = 'probabilities'
         else:
           key_to_use = 'predictions'
--- a/tensorboard/plugins/interactive_inference/utils/inference_utils.py
+++ b/tensorboard/plugins/interactive_inference/utils/inference_utils.py
@@ -820,22 +820,22 @@ def run_inference(examples, serving_bundle):
         tf.io.parse_example([ex.SerializeToString() for ex in examples],
         serving_bundle.feature_spec)).batch(batch_size))
 
-    if serving_bundle.use_predict:
-      preds_key = serving_bundle.predict_output_tensor
-    elif serving_bundle.model_type == 'regression':
-      preds_key = 'predictions'
-    else:
-      preds_key = 'probabilities'
+    # Use the specified key if one is provided.
+    key_to_use = (serving_bundle.predict_output_tensor
+        if serving_bundle.use_predict else None)
 
     values = []
     for pred in preds:
-      # If the prediction dictionary contains only one key, use that key.
-      returned_keys = list(pred.keys())
-      if len(returned_keys) == 1:
-        key_to_use = returned_keys[0]
-      else:
-        key_to_use = preds_key
-
+      if key_to_use is None:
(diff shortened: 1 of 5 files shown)


jameswex:
This fixed #2938

tensorboard/plugins/interactive_inference/utils/inference_utils.py
```diff
@@ -829,7 +829,18 @@ def run_inference(examples, serving_bundle):
 
     values = []
     for pred in preds:
-      values.append(pred[preds_key])
+      # If the prediction dictionary contains only one key, use that key.
+      returned_keys = list(pred.keys())
+      if len(returned_keys) == 1:
+        key_to_use = returned_keys[0]
```
tolga-b: This will pick whatever the model is outputting even in the case where predict_output_tensor is set. If the model is set up incorrectly and is not returning the tensor user wants, this may mislead the user into thinking they are actually using the tensor they specified in predict_output_tensor.

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
```diff
@@ -535,16 +527,46 @@ def chunks(l, n):
       for pred in response['predictions']:
         # If the prediction contains a key to fetch the prediction, use it.
         if isinstance(pred, dict):
+          # If the dictionary only contains one key, use it.
+          results_keys = list(pred.keys())
+          if len(results_keys) == 1:
+            results_key = results_keys[0]
+          else:
+            results_key = self.config.get('predict_output_tensor')
```
tolga-b: Same behavior as above.

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
```diff
@@ -275,7 +275,7 @@ def infer(self):
           inferences=json.dumps(inferences)))
     except Exception as e:
       output.eval_js("""backendError({error})""".format(
-          error=json.dumps({'msg': str(e)})))
+          error=json.dumps({'msg': repr(e)})))
```
tolga-b: Could you comment on why we changed from str to repr?

tensorboard/plugins/interactive_inference/utils/inference_utils.py
jameswex: Fixed

tensorboard/plugins/interactive_inference/witwidget/notebook/base.py
jameswex: Fixed

tensorboard/plugins/interactive_inference/witwidget/notebook/colab/wit.py
jameswex: repr(error) returns a string that includes the error type, like "KeyError: ...", whereas str(e) will just return the detailed text, without the context of what type of error occurred.

PR #2949, opened 2019-11-19
uploader: add cookie support to frontend handshake

wchargin:
Summary:
This enables communicating with frontends that require authentication.

Test Plan:
Unit tests included. As an integration test, visit the landing page for
an auth-walled frontend and capture your cookie via browser dev tools.
Securely store the cookie via `(umask 277 && xsel -ob >/tmp/cookie)`,
then pass it to the uploader:

```
bazel run //tensorboard -- \
    dev --origin http://localhost:9090 --cookie_file /tmp/cookie list
```

Also, make sure that the normal non-cookie flow still works when
connecting to a local server that does support the handshake but does
not require authentication:

```
bazel run //tensorboard -- dev --origin http://localhost:8080 list
```

wchargin-branch: uploader-cookie

### DIFF
--- a/tensorboard/uploader/server_info.py
+++ b/tensorboard/uploader/server_info.py
@@ -35,13 +35,15 @@ def _server_info_request():
   return request
 
 
-def fetch_server_info(origin):
+def fetch_server_info(origin, cookie=None):
   """Fetches server info from a remote server.
 
   Args:
     origin: The server with which to communicate. Should be a string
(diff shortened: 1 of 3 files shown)


wchargin:
Dropping this for now. We’re not sure that we want to encourage
exfiltrating cookies. It’s possible to do securely (see `umask` in
commit message), but easy to mess up if you’re not careful. We’ll hold
off and see if we can’t find a different solution.

PR #2947, opened 2019-11-19
ci: pin to non-broken nightlies

wchargin:
Summary:
Due to <https://github.com/tensorflow/tensorflow/issues/34427>.

This unblocks our CI until the upstream problem is identified and fixed.

Test Plan:
Fingers crossed.

wchargin-branch: pin-20191118-nightly

PR #2950, opened 2019-11-19
build: deprecate shasum in Vulcanization

stephanwlee:
As part of our security hardening and CSP support, we decided to
introduce script checksum as part of the Vulcanization that gets used at
the serving time using http_util.py. We understand that this imposes
a bit too stringent requirements for all and we decided to relax the CSP
comformance a little by trusting JavaScript from our server
(effectively, default-src 'self').

Note that we do not yet remove shasum related code from http_util
because it is being used by the Plugin HTML endpoint in application.py.
For that, we may want to use nonce or separate JavaScript file approach.

### DIFF
--- a/tensorboard/plugins/projector/projector_plugin.py
+++ b/tensorboard/plugins/projector/projector_plugin.py
@@ -269,8 +269,12 @@ class ProjectorPlugin(base_plugin.TBPlugin):
                 os.path.join('tf_projector_plugin', 'index.js')),
         '/projector_binary.html':
             functools.partial(
-                self._serve_html,
+                self._serve_file,
                 os.path.join('tf_projector_plugin', 'projector_binary.html')),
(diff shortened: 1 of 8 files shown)


wchargin:
Cancelled build to unblock #2947, without which this build would fail
anyway.

tensorboard/plugins/projector/projector_plugin.py
```diff
@@ -269,8 +269,12 @@ def get_plugin_apps(self):
                 os.path.join('tf_projector_plugin', 'index.js')),
         '/projector_binary.html':
             functools.partial(
-                self._serve_html,
+                self._serve_file,
                 os.path.join('tf_projector_plugin', 'projector_binary.html')),
+      '/projector_binary.js':
```
nfelt: nit: indentation off by 2?

PR #2952, opened 2019-11-20
ui: add slot for injecting header items

psybuzz:
Add the ability for embedders to inject items in the header, next to the 'Reload', 'Settings' buttons.

  - Adds a new slot for extra header items, to the left of the 'Reload' button, centered vertically
  - Drive by fixes the slot name for 'injected-overview'.  Slots define insertion map by the 'name' attribute, not 'id'
  - Re-enable the tf-tensorboard test


wchargin:
Cancelled build to unblock #2947, without which this build would fail
anyway.

tensorboard/components/tf_tensorboard/test/tensorboardTests.ts
```diff
@@ -23,11 +27,14 @@ namespace tf_tensorboard {
...
-        chai.assert.isNotNull(injected);
+        let overview = tensorboard.querySelector('#custom-overview');
+        expect(overview.assignedSlot.name).to.equal('injected-overview');
+
+        let headerItems = tensorboard.querySelector('#custom-header-items');
+        expect(headerItems.assignedSlot.name).to.equal('injected-header-items');
       });
 
-      it('reloads the active dashboard on request', (done) => {
+      xit('reloads the active dashboard on request', (done) => {
```
stephanwlee: prefer to accompany skipped tests with a reason.

psybuzz: I've added a comment

stephanwlee: xdescribe, too. if you think they are stale, maybe just omit.

tensorboard/components/tf_tensorboard/test/tensorboardTests.ts
```diff
@@ -23,11 +27,32 @@ namespace tf_tensorboard {
...
-        chai.assert.isNotNull(injected);
+        const overview = tensorboard.querySelector('#custom-overview');
+        checkSlottedUnderAncestor(overview, '#content-pane');
+
+        const headerItem1 = tensorboard.querySelector('#custom-header-item1');
+        const headerItem2 = tensorboard.querySelector('#custom-header-item2');
+        checkSlottedUnderAncestor(headerItem1, '.header');
+        checkSlottedUnderAncestor(headerItem2, '.header');
+
+        function checkSlottedUnderAncestor(
```
stephanwlee: several higher level comments:
1. hoist this up function since it is general utility.
2. `Polymer.dom($0).deepContains($1)` may want to look into this.
3. consider moving new additions in tf_util to testing.ts or tf_util/shadow_dom.ts since we probably want some common utilities for ShadowDom that are missing from `Polymer.dom`.

tensorboard/components/tf_tensorboard/test/tensorboardTests.ts
psybuzz: Done - added to xdescribe

tensorboard/components/tf_tensorboard/test/tensorboardTests.ts
psybuzz: 1. Done
2. Thanks, I didn't know Polymer offered this!
3. I'll keep in mind for the future.  Polymer's `deepContains()` addresses the need here, but I imagine we might want our own Shadow utils in the future

tensorboard/components/tf_tensorboard/test/tensorboardTests.ts
```diff
@@ -40,7 +62,9 @@ namespace tf_tensorboard {
         });
       });
 
-      describe('top right global icons', function() {
+      // TODO(psybuzz): Restore or remove these tests. This folder's tests seems
+      // to have not been running since the GitHub repo history was tracked.
```
stephanwlee: nit: instead of describing the history, for TODOs at least, describe precondition that can resolve this TODO.

psybuzz: Rewritten to note that the test was not running, and at least 1 reason for their failure.

PR #2953, opened 2019-11-21
ci: enable GitHub actions for tests

wchargin:
Summary:
This commit adds a GitHub Actions workflow definition for our main CI
build and tests. This new workflow will replace our use of Travis CI.
GitHub Actions has much better latency, on the order of seconds instead
of Travis’s minutes (usually) or hours (recently); better concurrency;
comparable incremental build times; and notably faster (30%) clean build
times. Travis is also turning down travis-ci.org in favor of
travis-ci.com, which requires overly broad repository permissions, so we
couldn’t continue with Travis even if we wanted to.

Build artifacts are cached with [Bazel remote caching] talking to a GCS
bucket, `gs://tensorboard-build-cache`. This bucket is publicly
readable, so PRs from untrusted forks may freely read from it. Only
trusted commits will be authorized to write to the build cache. Commits
to master are always trusted, so most of the time, the cache should be
fairly fresh.

[Bazel remote caching]: https://docs.bazel.build/versions/master/remote-caching.html

To proceed, we should merge this commit, make sure that its master build
works, then make it required and make Travis not required. After a few
days of soaking, disable Travis entirely. If something goes wrong, we
can just roll back by making Travis required again.

This does not yet have daily scheduled runs.

Test Plan:
Note that this commit triggers a GitHub Actions workflow that succeeds.

wchargin-branch: gh-actions


wchargin:
Build appears to not be configured on this organization repository (I’ve
alerted the powers that be), but consult a run on my personal fork here:
<https://github.com/wchargin/tensorboard/commit/365ca58317e3cbca522717a4fd99c0350556abfc/checks?check_suite_id=321679584>

.github/workflows/test.yml
```diff
@@ -0,0 +1,148 @@
...
+# Helpful YAML parser to clarify YAML syntax:
+# https://yaml-online-parser.appspot.com/
+
+name: CI
+
+on:
+  push: {}
+  pull_request: {}
+  schedule:
+    # 16:00 UTC is noon in Pacific standard time (UTC-8).
```
nfelt: These are the automatic builds on master, right?  Would it be worth doing these earlier in the day so that they're less likely to collide with real PR builds (and get us a fresh cache if we get caching working)?

Looking at recent `tf-nightly`  they all seem to be published between 2am and 3am pacific, so running this job at 5am should be fine (and would pick up our `tb-nightly` then too for consistency, not that it should matter).

tensorboard/tools/diagnose_tensorboard.py
```diff
@@ -244,6 +256,29 @@ def tensorboard_binary_path():
   logging.info("which tensorboard: %r", which("tensorboard"))
 
 
+@check
+def addrinfos():
+  """Gets family and nodename for a loopback address."""
+  print("On `socket`: %r" % (dir(socket),))
```
nfelt: Printing 4KB of everything on `socket` seems maybe like too much to me - it certainly could be useful for debugging and shouldn't contain anything sensitive, but I could see it causing more hesitation on the part of users since it's such a large blob.  Also, it's not collapsed in the email version of the issue so it does increase the amount we have to scroll through every time.

IMO a bit better just to stick to reading the targeted attributes and expanding that list over time as need be (and if we really need the full dump of `socket` from a user, we can ask for it).

.github/workflows/test.yml
wchargin: Sure, that sounds good to me. Will change.

tensorboard/tools/diagnose_tensorboard.py
wchargin: Fine with me. I’ll just remove it for now—the `sgetattrs` are printing
what we care about—and, like you say, we can add more later. (I didn’t
realize that the email notifications didn’t render the `<details>`, but
I guess that makes sense.)

wchargin:
Done. Force-pushed because these commits will be rebased. Diff:
<https://github.com/tensorflow/tensorboard/compare/365ca58317e3cbca522717a4fd99c0350556abfc..78b6b99f01a8beee988bbf80a798ffead094cf34>

.github/workflows/test.yml
wchargin: Apparently my comment was wrong anyway because I am incapable of
performing arithmetic. This was 08:00 PT and is now 05:00 PT/08:00 ET.

wchargin:
Status update, for those following along:

  - GitHub Actions is now enabled for our organization, so this PR
    actually works now!
  - GitHub fixed their Python 2 support, so we can test in Python 2 now,
    but we can also drop Python 2 support because it’s 2020.
  - Pre- and post-build latency is still significantly better than
    Travis, by minutes in each direction. For instance, failures of the
    license check or Bazel mirror URLs check will be surfaced within
    **10 seconds** of pushing the commit to GitHub, which is pretty hard
    to beat.
  - Actual build times slower than on Travis, because we don’t cache
    build targets, so each environment has to spend 26 minutes
    rebuilding all targets. The [open issue on the GitHub-provided cache
    action][cache-issue] has received some enthusiasm from GitHub staff,
    which is great to see. It doesn’t look like any concrete plans have
    been made or implementation started, though.
  - Consequently, even though the GitHub pipelines are much easier to
    configure (e.g., better granularity; simpler, more inspectable build
    setup) and have much better latency, the end-to-end throughput is
    still about 38 minutes for GitHub actions compared to 17 minutes for
    Travis, assuming full parallelism in each case. This doesn’t seem
    like an acceptable regression.
  - If I have some spare time, I’ll look into how easy it would be to
    set up managed Bazel remote caching ourselves using a GCS bucket
    that we own, and we can consider whether this is something that we
    want to take on. It looks like [Bazel has some built-in support for
    GCS caching][gcs-cache], so we likely wouldn’t need to maintain a
    service of test workers on GCP, but we would need to create a
    bucket, set up a service account, think at least a little bit about
    permissions and PRs running from forks, and whether we’re
    comfortable accepting any resulting maintenance burdens. IMHO, this
    is still worht considering even if it requires a bit of ops work on
    our part because the system is so much cleaner and all signs suggest
    that debugging Heisenbugs will be much easier. If `actions/cache` or
    similar gains native support for hermetic-style caches, there will
    be a very strong case.

[cache-issue]: https://github.com/actions/cache/issues/109
[gcs-cache]: https://docs.bazel.build/versions/master/remote-caching.html#google-cloud-storage

nfelt:
Would it make sense to enable Actions for the lint checks so that we can get the advantage of low-latency there, and just hold off on migrating the tests?  It would mean needing to look in two places for failures but surfacing them much faster, which seems nice especially now that we have much more stringent linting.

wchargin:
Sure, that makes sense to me. I’ll hook that up.

wchargin:
Another benefit of that is that we only need to run lint checks once
rather than for all builds.

wchargin:
Pardon the delay. We’re ready for another look.

wchargin:
To note, this takes the conservative path on the two axes that we
discussed: we always re-run tests in CI (no caching), and we don’t
populate the cache on pull requests (even if they’re trusted). 
I did use this PR to seed the cache, but I’ve now disabled that behavior
in the workflow file.

We can easily adjust either of these later. Starting with no test
caching is easier because we can turn it on without having to rmrf the
bucket (which I would want to do if we were turning it off).

ci/bazelrc_github
```diff
@@ -0,0 +1,34 @@
+# Limit resources since GitHub Actions VMs have 2 cores and 7 GB RAM.
```
nfelt, line 1: Slightly nicer if you don't mind I think to instead fork `ci/bazelrc_travis` from our current file, and then edit it in place for GitHub Actions.  This way 1) we see the diff and 2) we associate the history with the file we intend to keep long-term.

ci/bazelrc_github
```diff
@@ -0,0 +1,34 @@
...
+# it for build too, this causes a rebuild at test time, and if we set it for
+# both we hit https://github.com/bazelbuild/bazel/issues/8237.
+#
+# See also:
+# https://github.com/bazelbuild/bazel/issues/7095 (protobuf PATH sensitivity)
+# https://github.com/bazelbuild/bazel/issues/7026 (future of action_env)
+build --action_env=PATH
+
+# Log more information to help with debugging, and disable curses output which
+# just adds more clutter to the log. (Travis spoofs an interactive terminal.)
```
nfelt, line 22: nit: Travis mention here is stale

ci/bazelrc_github
```diff
@@ -0,0 +1,34 @@
...
+common --curses=no
+build --verbose_failures
+build --worker_verbose
+test --test_output=errors
+test --test_verbose_timeout_warnings
+
+# Enable remote caching. This cache is publicly readable, but only writable on
+# trusted commits (e.g., commits to master).
```
nfelt, line 30: Maybe worth a comment about why we aren't caching test results (i.e. because our tests aren't really hermetic).

.github/workflows/ci.yml
```diff
@@ -20,11 +17,87 @@ on:
...
+          printf 'Using writable cache\n'
+          creds_file=/tmp/service_account_creds.json
+          printf '%s\n' "${CREDS}" >"${creds_file}"
+          printf '%s\n' >>~/.bazelrc \
+            "common --google_credentials=${creds_file}" \
+            "common --remote_upload_local_results=true" \
+            ;
+      - name: 'Install Python dependencies'
+        run: |
+          python -m pip install -U pip
```
nfelt, line 71: We had a few comments in the Travis version for these lines; worth preserving?
https://github.com/tensorflow/tensorboard/blob/e34809c4e49aedd62aba616a96ab255a251d2816/.travis.yml#L53-L56

Also the `numpy` uninstall was there for a possibly Travis-specific issue.  Do we know if GH Actions has the same problem?  Maybe worth omitting it unless we still need it.

nfelt (approved):
Very exciting!  Thanks for persevering in this quest :)

ci/bazelrc_github
wchargin: Ah, good idea. Thanks.

ci/bazelrc_github
wchargin: Yeah—I had updated most of the references, but left this one and the
`numpy` business because I didn’t want to bother debugging whether they
were still necessary. But I can try removing them, sure.

ci/bazelrc_github
wchargin: Done.

.github/workflows/ci.yml
wchargin: Removed (see other comment). Let’s see. From experience, I would expect
GitHub’s defaults to be saner than Travis’s.

wchargin:
Very emblematic: the GitHub build passed half an hour ago, and the
Travis has not even started queueing. It’s holding onto life as long as
it can…

ci/bazelrc_travis
```diff
@@ -0,0 +1,35 @@
+# Limit resources since Travis Trusty GCE VMs have 2 cores and 7.5 GB RAM.
```
wchargin, line 1: (this shows as a 0-delta rename with `git diff -C`)

wchargin:
Confirmed that this uses writable cache on master:

![Screenshot of the actions workflow for “ci: enable GitHub actions for
tests (#2953)” on the `master` branch, showing the “Configure build
cache write credentials” step having written “Using writable cache”][ss]

[ss]: https://user-images.githubusercontent.com/4317806/101224143-b063bd00-3642-11eb-9802-adc0fb9a52b8.png

PR #2956, opened 2019-11-21
refactor: middlewares configurable by specific app

stephanwlee:
Rationale: application.py did three things: (1) implement application
logic like /data/plugins_listing, (2) configured application with
plugins, and (3) configured middlewares.

Requirement: we would like to be able to apply different middlewares in
different order to configure how 404 gets handled and apply certain
application specific logic.

Changes:
Use `application.TensorBoardWSGI` as the base app that supplies feature
(1) and (2). `application.standard_tensorboard_wsgi` will have a default
middleware configurations appropriate for local TensorBoard.

Plans:
I plan on applying experiment_id middleware which is not used in local TensorBoard
in appropriate applications. I also plan on omitting path_prefix to the hosted version.

Tests:
- Ran application
- Ran application with path_prefix.

WANT_LGTM=all

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -161,11 +161,17 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
       path_to_run = parse_event_files_spec(flags.logdir_spec)
     start_reloading_multiplexer(
         multiplexer, path_to_run, reload_interval, flags.reload_task)
-  return TensorBoardWSGIApp(
+
+  app = TensorBoardWSGIApp(
       flags, plugin_loaders, data_provider, assets_zip_provider, multiplexer)
+  app = empty_path_redirect.EmptyPathRedirectMiddleware(app)
+  app = path_prefix.PathPrefixMiddleware(app, flags.path_prefix)
+  app = security_validator.SecurityValidatorMiddleware(app)
+  app = handling_errors(app)
+  return app
 
 
-def _handling_errors(wsgi_app):
+def handling_errors(wsgi_app):
   def wrapper(*args):
     (environ, start_response) = (args[-2], args[-1])
     try:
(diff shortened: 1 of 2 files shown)


tensorboard/backend/application.py
```diff
@@ -161,11 +161,17 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
       path_to_run = parse_event_files_spec(flags.logdir_spec)
     start_reloading_multiplexer(
         multiplexer, path_to_run, reload_interval, flags.reload_task)
-  return TensorBoardWSGIApp(
+
+  app = TensorBoardWSGIApp(
       flags, plugin_loaders, data_provider, assets_zip_provider, multiplexer)
+  app = empty_path_redirect.EmptyPathRedirectMiddleware(app)
```
wchargin: Why is `experiment_id.ExperimentIdMiddleware` removed?

tensorboard/backend/application_test.py
```diff
@@ -317,25 +318,29 @@ def testPluginEntryBadModulePath(self):
...
 
 
-class ApplicationBaseUrlTest(tb_test.TestCase):
+class StandardTensorBoardWSGIAppUrlTest(tb_test.TestCase):
   path_prefix = '/test'
   def setUp(self):
     plugins = [
-        FakePlugin(plugin_name='foo'),
-        FakePlugin(
+        FakePluginLoader(plugin_name='foo'),
```
wchargin: I don’t follow; why are these changes required?

stephanwlee: Answered offline but for posterity:

Previously, it was using TensorBoardWSGI which does not call `make_plugin_loader` as opposed to now, it uses `standard_tensorboard_wsgi` which calls `make_plugin_loader` through TensorBoardWSGIApp. `make_plugin_loader` requires plugins to be a class (not instance) of TBPlugin or an instance of TBPluginLoader.

stephanwlee:
Thanks for suggesting the alternative.

```py
def _create_wsgi_app(self):
  ...
  app = empty_path_redirect.EmptyPathRedirectMiddleware(app)
  for middleware in self._experiment_id_middlewares:
    app = middleware(app)
  app = path_prefix.PathPrefixMiddleware(app, flags.path_prefix)
  ...
```
I did consider this option but I thought this was problematic for few reasons. Yes, we would like some global handler for errors and global validation of the security for all apps but (1) I want to be able to configure them programmatically and (2) I may want to swap and use other implementation of them. 
 Similarly different apps have different requirements that, for instance, I may want to omit `PathPrefixMiddleware` or add more in strategic places (before PathPrefix but after ExperimentId).

I am curious to learn @nfelt's opinion on this.

PR #2966, opened 2019-11-22
csp: allow frame-src to be configured

psybuzz:
Apps that embed TensorBoard backend can now configure `frame-src`, to allow the frontend to render iframes from contexts other than self.

Adds _CSP_FRAME_DOMAINS_WHITELIST.


psybuzz:
Adds _CSP_FRAME_DOMAINS_WHITELIST, so that apps that embed TensorBoard backend can configure frame-src.  This would allow the frontend to render iframes from contexts other than self.

PR #2975, opened 2019-11-27
Add empty `__init__.py`s where missing

wchargin:
Summary:
Bazel generates these automatically, but having them in the source tree
is convenient for REPLs and tools like `pydoc`.

Generated with:

```
find tensorboard -type f -name '*.py' |
    xargs -n 1 -d '\n' dirname |
    sort -u |
    xargs -n 1 -d '\n' sh -c 'cp -n tensorboard/util/__init__.py "$1/"' cmd
```

Test Plan:
Running `pydoc tensorboard.data.provider` after this commit successfully
shows useful docs; before this commit, it fails with, “no Python
documentation found for 'tensorboard.data.provider'”.

wchargin-branch: empty-init-py

### DIFF
--- a/tensorboard/data/__init__.py
+++ b/tensorboard/data/__init__.py
@@ -1,0 +1,14 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


wchargin:
Yeah… At least I think they don’t make it into the final Pip package,
because we never depend on the `__init__.py`s at the BUILD level. So
we’re not shipping kilobytes of redundant licenses to end users, just
ourselves. :-)

nfelt:
FWIW Bazel's behavior of auto-creating the `__init__.py` is in theory going to be deprecated and removed eventually: https://github.com/bazelbuild/bazel/issues/7386

wchargin:
Good to know; thanks. Perhaps as that time draws near we can write a
test that in the extracted `//tensorboard/pip_package`, every directory
with a `*.py` file also has an `__init__.py` file.

PR #2976, opened 2019-11-27
Angular: Fix NgRx Action console logging in dev mode

davidsoergel:
Console logging for NgRx actions was not working because of file loading order issues.  Basically (I think) our `const metaReducers = ...` was not actually set at the time that the AppModule @NgModule decorator was evaluated.

While I was at it, I swapped in ngrx-store-logger in place of our manual logger, for nice color coding and such.


davidsoergel:
Sorry for the crufty commit history-- I think it's not worth cleaning up in this case since the actual changes are tiny and easy to see, but I'll try to arrange my branches more cleanly in the future :)

tensorboard/webapp/reducer_config.ts
```diff
@@ -12,32 +12,32 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-
-    return result;
-  };
+export function logger(reducer: ActionReducer<State>): ActionReducer<State> {
+  // We must call isDevMode() within this function, not at file load time,
+  // because of load ordering issues vs. the config/bootstrap setup.
+
+  // TODO(stephanwlee, davidsoergel): Why is this called twice?
+  console.log(
+    isDevMode() ? 'DEV mode: Logger enabled' : 'PROD mode: Logger disabled'
```
psybuzz: nit: in Prod, I think we don't gain much by seeing 'Logger disabled'.

tensorboard/webapp/reducer_config.ts
```diff
@@ -12,32 +12,32 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-    console.log('next state', result);
-    console.groupEnd();
-
-    return result;
-  };
+export function logger(reducer: ActionReducer<State>): ActionReducer<State> {
+  // We must call isDevMode() within this function, not at file load time,
+  // because of load ordering issues vs. the config/bootstrap setup.
+
+  // TODO(stephanwlee, davidsoergel): Why is this called twice?
```
psybuzz: I'm curious as well.  It seems ngrx calls metareducers once when creating Reducers, and again in 'updateReducers', called when it actually adds features/reducers to the store.

Maybe we can just do this console.log outside of this function, in the module itself?

tensorboard/webapp/reducer_config.ts
```diff
@@ -12,32 +12,32 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+  console.log(
+    isDevMode() ? 'DEV mode: Logger enabled' : 'PROD mode: Logger disabled'
+  );
+  return isDevMode() ? storeLogger()(reducer) : reducer;
 }
 
-export const metaReducers: MetaReducer<any>[] =
-  config.env === 'dev' ? [logger] : [];
+// TODO(stephanwlee, davidsoergel): consider injecting metareducers via
+// https://ngrx.io/guide/store/recipes/injecting#injecting-meta-reducers
```
psybuzz: Both ways sound reasonable, are there benefits to this approach?

tensorboard/webapp/reducer_config.ts
caisq: +1.

Also, I think "Logger" is pretty vague. Can we be more specific in the log message? I.e,

```typescript
if (isDevMod()) {
  console.log('DEV mode: NgRx Store Logger enabled');
}
```

caisq: nit: `isDevMod()` is called twice here. Can we just call it once and use the return value twice?

stephanwlee:
This is fixed by recent changes to the dev environment and what not. Thanks a lot for PR, David!

PR #2979, opened 2019-11-27
data: add tensor support to data provider interface

wchargin:
Summary:
This commit specifies the `list_tensors` and `read_tensors` methods on
the data provider interface. These methods are optional for now (i.e.,
not decorated with `abc.abstractmethod`) for compatibility reasons, but
we’ll make them required soon.

Test Plan:
Unit tests included.

wchargin-branch: data-tensors-interface

### DIFF
--- a/tensorboard/data/provider.py
+++ b/tensorboard/data/provider.py
@@ -22,6 +22,7 @@ import abc
 import collections
 
 import six
+import numpy as np
 
 
 @six.add_metaclass(abc.ABCMeta)
@@ -120,12 +121,56 @@ class DataProvider(object):
     """
     pass
 
-  def list_tensors(self):
-    """Not yet specified."""
+  def list_tensors(self, experiment_id, plugin_name, run_tag_filter=None):
+    """List metadata about tensor time series.
+
+    Args:
+      experiment_id: ID of enclosing experiment.
+      plugin_name: String name of the TensorBoard plugin that created
+        the data to be queried. Required.
+      run_tag_filter: Optional `RunTagFilter` value. If omitted, all
+        runs and tags will be included.
+
+    The result will only contain keys for run-tag combinations that
+    actually exist, which may not include all entries in the
+    `run_tag_filter`.
+
+    Returns:
+      A nested map `d` such that `d[run][tag]` is a `TensorTimeSeries`
+      value.
+
+    Raises:
+      tensorboard.errors.PublicError: See `DataProvider` class docstring.
+    """
     pass
 
-  def read_tensors(self):
-    """Not yet specified."""
+  def read_tensors(
+      self, experiment_id, plugin_name, downsample=None, run_tag_filter=None
+  ):
+    """Read values from tensor time series.
+
+    Args:
+      experiment_id: ID of enclosing experiment.
+      plugin_name: String name of the TensorBoard plugin that created
+        the data to be queried. Required.
+      downsample: Integer number of steps to which to downsample the
+        results (e.g., `1000`). Required.
+      run_tag_filter: Optional `RunTagFilter` value. If provided, a time
+        series will only be included in the result if its run and tag
+        both pass this filter. If `None`, all time series will be
+        included.
+
+    The result will only contain keys for run-tag combinations that
+    actually exist, which may not include all entries in the
+    `run_tag_filter`.
+
+    Returns:
+      A nested map `d` such that `d[run][tag]` is a list of
+      `TensorDatum` values sorted by step.
+
+    Raises:
+      tensorboard.errors.PublicError: See `DataProvider` class docstring.
+    """
     pass
 
   def list_blob_sequences(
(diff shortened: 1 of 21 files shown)


tensorboard/data/provider.py
```diff
@@ -120,12 +121,56 @@ def read_scalars(
...
-  def read_tensors(self):
-    """Not yet specified."""
+  def read_tensors(
+      self, experiment_id, plugin_name, downsample=None, run_tag_filter=None
+  ):
+    """Read values from tensor time series.
+
+    Args:
+      experiment_id: ID of enclosing experiment.
+      plugin_name: String name of the TensorBoard plugin that created
```
stephanwlee: No action required at all but asking for a short discussion.

It feels weird that a data is tied to a plugin. I think we should call this `summary_name`. Of course, I think there should be a equivalent changes on the summary API. What do you think?

tensorboard/data/provider.py
```diff
@@ -392,6 +437,110 @@ def __repr__(self):
...
+
+class TensorDatum(object):
+  """A single datum in a tensor time series for a run and tag.
+
+  Attributes:
+    step: The global step at which this datum occurred; an integer. This
+      is a unique key among data of this time series.
+    wall_time: The real-world time at which this datum occurred, as
+      `float` seconds since epoch.
+    numpy: The `numpy.ndarray` value with the tensor contents of this
```
stephanwlee: Asking because I lack knowledge: is a Tensor always ndarray?

tensorboard/data/provider.py
wchargin: Heh. Yes, agreed; it’s not clear whether “plugin name” is a better
ontology than something more like “data type” or “MIME type”, where
there is potentially a many-to-many relationship between producers
(plugins) and consumers (dashboards) of a given data type.

As you observe, there’s not really a good place to fit this into the
current summary APIs, so I opted to be internally consistent. I can
certainly imagine a world in which we intentionally coöpt the semantics
of `plugin_name` to mean something more like “MIME type”.

I suspect that `summary_name` could be a bit confusing; I’d be inclined
to say that the “name” of my summary is `loss` or `accuracy`, not
`scalars` or `images`.

Would be happy to discuss more, and I think that @davidsoergel also has
some thoughts here.

tensorboard/data/provider.py
wchargin: That’s up to us! :-)

Every “normal” tensor (I’m excluding, e.g., resource tensors) *can* be
represented as an ndarray, via `.numpy()` at eager execution time—

```python
>>> import tensorflow as tf
>>> tf.constant(17).numpy()
17
>>> type(tf.constant(17).numpy())
<class 'numpy.int32'>
>>> tf.constant([42.0, 43.0]).numpy()
array([42., 43.], dtype=float32)
>>> tf.constant(["hello", "world"]).numpy()
array([b'hello', b'world'], dtype=object)
```

—or from proto via `tf.make_ndarray` or our own notf equivalent in
`util/tensor_util.py`. So the choice that I was considering was whether
to return an `ndarray` here or a raw `TensorProto`. I opted to use an
`ndarray` because (a) I think that that’s what most clients will find
much more convenient to use, and (b) it reduces API-level coupling to
TensorFlow’s protos, which TensorFlow appreciates.

The downside is decoding performance for plugins that really want to
operate on the serialized form without a roundtrip. If this becomes a
real use case, we can imagine adding a flag to return the raw protos.

stephanwlee: > ... I opted to use an ndarray because (a) I think that that’s what most clients will find
much more convenient to use, and (b) it reduces API-level coupling to
TensorFlow’s protos, which TensorFlow appreciates.

Yup, the upside was clear just reading the code.

I actually did not know `np.ndarray(1).shape` would be `()`. Given that a single value can be correctly represented by ndarray with the right shape and right dtype, I now see the benefit of using ndarray here.

wchargin: > I actually did not know `np.ndarray(1).shape` would be `()`.

Yeah, I had to double-check, myself. It’s extra non-obvious because
numpy uses the same `repr` for the rank-0 ndarray as for its item,
unlike TensorFlow’s eager tensors, which are quite clear:

```python
>>> tf.constant(17)
<tf.Tensor: shape=(), dtype=int32, numpy=17>
```

> Given that a single value can be correctly represented by ndarray with
> the right shape and right dtype, I now see the benefit of using
> ndarray here.

Cool; thanks for your input! Glad that we agree.

tensorboard/data/provider.py
nfelt: FWIW William and I talked about this a while back as well.  I still think ultimately we want something more like data type or MIME type, but we don't really have a coherent plan for this yet, so the thought was that we'd stick to consistency with the existing pattern of using plugin names until we have a better strategy.

PR #2980, opened 2019-11-27
data: add tensor support to multiplexer provider

wchargin:
Summary:
This commit implements the new `list_tensors` and `read_tensors` methods
for the data provider implementation backed by the event multiplexer.

Test Plan:
Unit tests included.

wchargin-branch: data-tensors-mux

### DIFF
--- a/tensorboard/backend/event_processing/data_provider.py
+++ b/tensorboard/backend/event_processing/data_provider.py
@@ -57,7 +57,7 @@ class MultiplexerDataProvider(provider.DataProvider):
       return None
 
   def data_location(self, experiment_id):
-    del experiment_id   # ignored
+    del experiment_id  # ignored
     return str(self._logdir)
 
   def list_runs(self, experiment_id):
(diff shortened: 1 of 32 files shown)


tensorboard/backend/event_processing/data_provider.py
```diff
@@ -72,8 +72,69 @@ def list_runs(self, experiment_id):
...
+    def convert_scalar_event(event):
+      return provider.ScalarDatum(
+          step=event.step,
+          wall_time=event.wall_time,
+          value=tensor_util.make_ndarray(event.tensor_proto).item(),
+      )
+
+    return self._read(convert_scalar_event, index)
+
+  def list_tensors(self, experiment_id, plugin_name, run_tag_filter=None):
```
stephanwlee: I generally don't think copy/paste is a bad thing but do we expect these implementations (list_tensors and read_tensors) to differ much beyond `TensorTimeSeries` and `convert_tensor_event`?

tensorboard/backend/event_processing/data_provider_test.py
```diff
@@ -211,6 +218,64 @@ def test_read_scalars_but_not_rank_0(self):
...
+    result = provider.read_tensors(
+        experiment_id="unused",
+        plugin_name=image_metadata.PLUGIN_NAME,
+        run_tag_filter=run_tag_filter,
+        downsample=None,  # not yet implemented
+    )
+
+    self.assertItemsEqual(result.keys(), ["pictures"])
+    self.assertItemsEqual(result["pictures"].keys(), ["purple", "green"])
+    for run in result:
```
stephanwlee: optional: It may be easier to assert by forming an expected map of `TensorTimeSeries` and just do `assertEqual` since we implemented `eq`.

tensorboard/backend/event_processing/data_provider.py
wchargin: I expect `list_tensors` and `read_tensors` to remain similar to their
scalar counterparts, though the scalar counterparts could gain
additional features, like “list scalar time series with tag name
`accuracy` whose moving average is above 0.999”, that don’t make sense
for general tensor time series. (This is why we chose to separate scalar
time series out in the first place, even though the basic functionality
is subsumed by tensor time series.)

In such a future, I’d be happy to inline `self._read` and `self._list`
with appropriate changes. If you’d prefer that these be inlined today,
that’s also fine with me. I’m not exactly sure what you’re asking, so
let me know if you’d like any changes?

tensorboard/backend/event_processing/data_provider_test.py
wchargin: Indeed, but I wanted to use `np.testing.assert_equal` for better error
messages than “`giant_numpy_array_1 != giant_numpy_array_2`”.

tensorboard/backend/event_processing/data_provider.py
```diff
@@ -72,8 +72,69 @@ def list_runs(self, experiment_id):
...
+    )
+
+  def read_tensors(
+      self, experiment_id, plugin_name, downsample=None, run_tag_filter=None
+  ):
+    # TODO(@wchargin): Downsampling not implemented, as the multiplexer
+    # is already downsampled. We could downsample on top of the existing
+    # sampling, which would be nice for testing.
+    del downsample  # ignored for now
+    index = self.list_tensors(
```
nfelt: I guess I had no objection in the original PR, but isn't it kind of inefficient to implement the read operation in terms of listing?  The list operation already iterates over all the tensors once in order to determine max step and wall time, but then the read operation doesn't return that metadata at all (it's just discarded) so the computation is wasted, and we have to iterate over the tensors again a second time to get their actual values.

Performance wise it's probably not the end of the world, but it seems suboptimal.  (The same argument we had for the TB.dev backend might suggest also that we don't really need to return max step and wall time all the time; if we made those optional then listing could be made more efficient and the reuse wouldn't result in as much redundant iteration, although it's still a little duplicative even then.)

tensorboard/backend/event_processing/data_provider.py
```diff
@@ -72,8 +72,69 @@ def list_runs(self, experiment_id):
...
+  ):
+    # TODO(@wchargin): Downsampling not implemented, as the multiplexer
+    # is already downsampled. We could downsample on top of the existing
+    # sampling, which would be nice for testing.
+    del downsample  # ignored for now
+    index = self.list_scalars(
+        experiment_id, plugin_name, run_tag_filter=run_tag_filter
+    )
+
+    def convert_scalar_event(event):
```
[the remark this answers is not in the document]
wchargin: Fair enough; done in #2994.

tensorboard/backend/event_processing/data_provider.py
wchargin: Yes, agreed. I went with this because it was convenient and because, as
you say, it’s not the end of the world (due to downsampling, these
queries are all bounded). For a data provider implementation where this
actually required an extra RPC, I of course would make a different call.

I’d be happy to accept a change that streamlined this, but it’s not high
enough on my priorities for me to do so myself, unless you feel strongly
about it.

PR #2840, opened 2019-10-25
Align stub GFile with current TF FileIO API

davidsoergel:
Our stub GFile implementation had drifted a bit out of sync with the official TensorFlow implementation.  This realigns our version with TF.  The process was roughly:

* Fork the TF GFile tests, retaining only those that test functionality that our GFile stub implements at all.
* Observe that many of the tests failed, due to minor API differences (perhaps divergence over time), such as throwing different flavors of exceptions.
* Update our stub GFile as needed to make the tests pass.
* Minor code formatting (just a few newlines and such), but not a full Black run.

PR #2982, opened 2019-11-28
Implements Google Cloud Storage for tf.io.gfile.GFile

lanpa:
This PR implements GCS for tensorboard-compat so that torch.utils.tensorboard can use GCS buckets.
ref issue: https://github.com/pytorch/pytorch/issues/24468

A porting of S3FileSystem called GCSFileSystem is implemented as well as the corresponding tests.
However, I didn't find a similar package like `moto` so I use my own GCS credential for testing. Besides the unittest, I also write tensorboard event to GCS with torch.utils.tensorboard, and the file downloaded can be rendered correctly.
To run the test, the credential file in  https://github.com/tensorflow/tensorboard/compare/master...lanpa:gcs-writer?expand=1#diff-2ffeb20dada1725639d2f50455f20a90R46 needs to be changed to a valid one.
[TODO] Modify the unit tests so that it can pass in CI without a credential file.

* Screenshots of UI changes
N/A

`bazel test //tensorboard/summary/writer:event_file_writer_gcs_test --test_output=all --cache_test_results=no --test_timeout 180`
`bazel test //tensorboard/compat/tensorflow_stub:gfile_gcs_test --test_output=all --cache_test_results=no --test_timeout 180`

cc @orionr @nfelt


davidsoergel:
Thanks for this very nice contribution!

It would be good if we can get #2840 in before this one, because of the updated tests and API alignment there.  That PR is on the back burner but I'll discuss with @nfelt how best to proceed.

davidsoergel:
OK, #2840 is in.  Can you merge, and also skim that PR to see if there are any changes that should be propagated to this one?  Thanks!

lanpa:
@davidsoergel I have rebased the PR on 78c919738f2. The failed test is expected because it needs a valid credential. Is there any workaround for a mock up test?

tarrade:
What is the status of this PR @davidsoergel ? This is a very cool contribution that will allow us to use GCS to store Tensorboard logs without the need to install Tensorflow for example when using Pytorch.

drubinstein:
Any updates or any way I can help push this PR forward? Tested it locally and it helped me get past [an issue with fsspec/gcsfs](https://github.com/fsspec/gcsfs/issues/379) which prevents logs from being logged to tensorboard in some scenarios e.g. when the [Tensorboard Logger from Torch Lightning](https://pytorch-lightning.readthedocs.io/en/stable/_modules/pytorch_lightning/loggers/tensorboard.html#TensorBoardLogger) is called.

martindurant:
>  I didn't find a similar package like moto so I use my own GCS credential for testing

gcsfs uses the docker image fsouza/fake-gcs-server for testing, see https://github.com/fsspec/gcsfs/blob/main/gcsfs/tests/conftest.py

PR #2986, opened 2019-12-02
Use `tensorboard.errors` in histograms and distributions

wchargin:
Summary:
We now raise `errors.NotFoundError` rather than manually propagating an
error message and status code.

The response code was previously 400 Bad Request, but 404 Not Found is
more appropriate, and is consistent with the scalars plugin.

Test Plan:
Requesting `/data/plugin/histograms/histograms?run=foo&tag=bar` yields
an error with the same message as before (but now with a “Not found:”
prefix), and the histograms and distributions dashboards both work.

wchargin-branch: histograms-notfound

### DIFF
--- a/tensorboard/plugins/distribution/distributions_plugin.py
+++ b/tensorboard/plugins/distribution/distributions_plugin.py
@@ -74,7 +74,11 @@ class DistributionsPlugin(base_plugin.TBPlugin):
     )
 
   def distributions_impl(self, tag, run):
-    """Result of the form `(body, mime_type)`, or `ValueError`."""
+    """Result of the form `(body, mime_type)`.
+
+    Raises:
+      tensorboard.errors.PublicError: On invalid request.
+    """
     (histograms, mime_type) = self._histograms_plugin.histograms_impl(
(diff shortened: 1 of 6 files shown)


tensorboard/plugins/distribution/distributions_plugin.py
```diff
@@ -74,7 +74,11 @@ def frontend_metadata(self):
     )
 
   def distributions_impl(self, tag, run):
-    """Result of the form `(body, mime_type)`, or `ValueError`."""
+    """Result of the form `(body, mime_type)`.
+
+    Raises:
+      tensorboard.errors.PublicError: On invalid request.
```
stephanwlee: Not important: the error message will say histogram. Can that be confusing? FWIW, I am fine as is.

wchargin: Correct, and good point. It might be somewhat confusing, but it could
also be confusing the other way (there _aren’t_ any separate tags for
distributions). It’s also not a regression, and `http_api.md` has
likewise used the same wording, so it’s fine with me, too.

PR #2996, opened 2019-12-04
Expose visibility for build rules needed for WIT dynamic plugin

jameswex:

In order for WIT to move to a dynamic plugin, some build rules need to be made visibility public so WIT build can use them.

Building current draft of WIT dynamic plugin in its own repo and needed access to these rules.

### DIFF
--- a/tensorboard/components/vz_line_chart2/BUILD
+++ b/tensorboard/components/vz_line_chart2/BUILD
@@ -1,4 +1,4 @@
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = ["//tensorboard:internal"])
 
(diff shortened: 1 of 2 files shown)


tensorboard/components/vz_line_chart2/BUILD
```diff
@@ -1,4 +1,4 @@
-package(default_visibility = ["//tensorboard:internal"])
+package(default_visibility = ["//visibility:public"])
```
stephanwlee: Please, instead, make specific targets public.

jameswex: done

PR #2997, opened 2019-12-04
data: add docs about data classes

wchargin:
Summary:
We’ve specified the methods for listing and reading data from the three
data classes, but haven’t actually described the high-level semantics of
those classes themselves. This commit expands the `DataProvider` docs to
include such a description.

wchargin-branch: data-classes-docs


tensorboard/data/provider.py
```diff
@@ -33,6 +33,54 @@ class DataProvider(object):
   providers may be asked to implement more filtering mechanisms, such as
   downsampling strategies or domain restriction by step or wall time.
 
+  The data provider interface specifies three *data classes*: scalars,
+  tensors, and blob sequences. All data is stored in *time series* for
+  one of these data classes. A time series is identified by run name and
+  tag name (each a non-empty text string), as well as an experiment ID
```
davidsoergel: This is an outstanding confusion, at least for me.  Graphs, at least, may be stored at the run level, in which case the tag is `None`.  See also the final commit on #3001.  Happy to chat offline-- probably faster to get unconfused that way :)

wchargin: The tag must never be `None`. Indeed, run-level data, including graphs
and hyperparameters, does not have an obviously canonical embedding into
this model. The way that the `hparams` plugin does it is to use [fixed
tag names that are not `None`][hp], rather than fixed tag names that are
`None`.

Permitting `None` would allow _one_ extra value, at the cost of
complicating the typing significantly (especially in Python(…), but also
in database layers). Requiring plugins to use their own namespaces, like

```
__run_graph__
__conceptual_graph__/USER_PROVIDED_TAG
__profile_graph__/USER_PROVIDED_TAG
```

doesn’t introduce extra typing complexity and also has better semantics.

[hp]: https://github.com/tensorflow/tensorboard/blob/5a5023f96059caabd048f2687604b06d5ee8fa8f/tensorboard/plugins/hparams/metadata.py#L30-L32

wchargin: And yes, I certainlyagree that there is opportunity for confusion when
this is not specified, hence this PR. :-)

davidsoergel: This is all much clearer now, thanks!

PR #3001, opened 2019-12-05
Fix graph plugin with data provider, and its test

davidsoergel:
This fixes the bug in #2991 found by @wchargin.

More importantly, it fixes the test that should have caught that bug in the first place (which has been broken since #1853): expected and actual data were being compared with `assertItemsEqual` instead of `assertEqual`, and so differences were not detected.


davidsoergel:
Gah sorry something is still broken.  Investigating.

davidsoergel:
Ok, this works now.  (Pretty sure the console error is a red herring, btw: that happens regardless, even with generic_data=false)

PR #3009, opened 2019-12-09
images: remove unused width and height on frontend

wchargin:
Summary:
The images plugin backend sends down images’ widths and heights, but the
frontend does not actually need or use this data. (A comment claiming
otherwise was in error.) This commit removes that behavior.

Test Plan:
The dashboard behavior appears unchanged, with both fixed-width and
actual-size display behaviors at both the per-image and dashboard-global
levels. All references to `width` or `height` in modified files are now
irrelevant. No network responses now contain `width` or `height`.

wchargin-branch: images-no-dims


wchargin:
> One thing to consider: often on the frontend, it is better to know the
> dimensions upfront without fetching the data for tiling or
> pre-allocating box for the layout. As you indicated, we do not make
> use of that right now but we may want it.

Thanks: I appreciate you calling this out, as I was indeed considering
dropping it, but you’re probably right that we should keep it around.
I think that I’ll go ahead and merge this, as the change is still valid
and useful, but will retain the actual width and height data for now.

PR #3017, opened 2019-12-10
ng: material icon svg

stephanwlee:
Unlike the default material-icon implementation, TensorBoard, which
tries to optimize to not make remote requests, bundles the SVG icons
with the pip package. To do so, we download the SVG from GitHub then use
a build rule to bundle the ones we use into our app. This change creates
the infrastructure and make use of the cog icon.

This is a manual revert of #2855.
Internal counter-part: cl/277216449


tensorboard/tools/mat_bundle_icon_svg.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+#!/bin/bash
```
davidsoergel: duplicated line

stephanwlee:
> This seems a bit intricate, with the shell script and all. I trust that it works, but is there a convenient way to test it?

Which part requires testing? If you are asking how I manually tested it, I ran the ng_tensorboard and made sure the cog icon shows up on the header.

davidsoergel:
Well, there are aspects of this that could become broken in the future, right, if the build rules get munged, or the dependency no longer provides the desired icon, or the naming scheme in the shell script changes or whatever.  Currently, IIUC, CI would not catch that.  I'm not saying testing this is absolutely essential--it's a fairly minor concern--but more wondering out loud whether we have test infrastructure that could validate this sort of thing in principle.  A Karma test would work, right?

stephanwlee:
Thoughts:
1. mat-icon complains out loud (console.error) when a component refers to an icon that does not exist
2. remediation: screenshot tests would catch badly broken one.
3. karma test _may not_ catch it. Unless it is critical for behavior, you may use stub implementation of NgModule dependency.
4. internally, there is something called jswire that would be better suited for testing if it is super critical.
5. without going through entire dependency graph and parsing the angular template, it is impossible for the build rule to know if it is complete.

tensorboard/tools/mat_bundle_icon_svg.sh
```diff
@@ -0,0 +1,33 @@
...
+# ids.
+# Usage: mat_bundle_icon_svg.sh [output.svg] [in_1.svg] [in_2.svg] ...
+
+OUTPUT_PATH=$1
+shift
+
+PREAMBLE='<svg><defs>'
+POSTAMBLE='</defs></svg>'
+
+echo $PREAMBLE > $OUTPUT_PATH
```
wchargin: Please quote all variable expansions.

tensorboard/tools/mat_bundle_icon_svg.sh
```diff
@@ -0,0 +1,33 @@
...
+for file in "$@"; do
+  svg_id="$(basename $file | sed 's|\.svg$||g')"
+  cat $file | sed "s|<svg|<svg id=\"$svg_id\"|g" >> $OUTPUT_PATH
```
wchargin: You can avoid re-opening the file *n* times by putting the whole thing
inside a redirection:

```shell
{
    echo "${PREAMBLE}"
    for file; do
        ...
    done
    echo "${POSTAMBLE}"
} >"${OUTPUT_PATH}"
```

wchargin: This is a “useless use of `cat`”: you can write `<"${file}" sed EXPR`
(which avoids spawning extra processes), or just `sed EXPR "${file}"`
because `sed` takes files as arguments

(In general, it’s better to pass files as arguments rather than streams
when you can, because some programs can execute more efficiently that
way, because they can seek across files. Doesn’t matter for `sed`,
though.)

tensorboard/defs/defs.bzl
```diff
@@ -68,3 +68,14 @@ def tf_ng_web_test_suite(runtime_deps = [], bootstrap = [], deps = [], **kwargs)
         ],
         **kwargs
     )
+
+def tf_svg_bundle(name, srcs, output_name):
```
wchargin: Optional: Maybe rename `output_name` to `out` for consistency with
native rules and `tensorboard_typescript_bundle`? It’s also shorter. :-)

tensorboard/tools/mat_bundle_icon_svg.sh
stephanwlee: I don't feel too good about the `sed ... 2>&0` part. Is it correct?

stephanwlee:
David et al., I created a MatIconTestingModule that checks whether the `svgIcon` input is one of the whitelisted one. There is no guarantee that a test author will use the TestingModule but this was my attempt at mitigating your concern. PTAL.

tensorboard/tools/mat_bundle_icon_svg.sh
wchargin: The code looks good without `2>&0`, but I, too, find that a bit
confusing. The effect of `2>&0` is to let you _read_ from stderr as if
it were stdin; i.e., it’s exactly equivalent to `2<&0`. This isn’t
usually useful, because most programs don’t try to read from stderr.

It should work without any redirections: `sed` will read from `"$file"`
and print to stdout, and stdout is redirected to `"${OUTPUT_PATH}"` in
the outer `{ … }` group.

PR #3020, opened 2019-12-11
ng: fixed the TS conformance issue

stephanwlee:
1. Cannot set `src` without TrustedURL. Fixed it by tricking the
conformance checker for now.
2. `SimpleChanges` is an object whose keys are names of the input or prop.
Prevent variable mangling by accessing via square brackets.

### DIFF
--- a/tensorboard/webapp/plugins/plugins.component.ts
+++ b/tensorboard/webapp/plugins/plugins.component.ts
@@ -58,10 +58,10 @@ export class PluginsComponent implements OnChanges {
   private readonly pluginInstances = new Map<string, HTMLElement>();
 
   ngOnChanges(change: SimpleChanges): void {
-    if (change.activePlugin && this.activePlugin) {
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/plugins/plugins.component.ts
```diff
@@ -58,10 +58,10 @@ export class PluginsComponent implements OnChanges {
   private readonly pluginInstances = new Map<string, HTMLElement>();
 
   ngOnChanges(change: SimpleChanges): void {
-    if (change.activePlugin && this.activePlugin) {
+    if (change['activePlugin'] && this.activePlugin) {
```
caisq: This is arcane. How are we supposed to know it beforehand? Is there a way to enforce this in OSS?

stephanwlee: AFAIK, no :(. I kinda hope gts includes more conformance checks but there was no clear evidence that it is the case.

PR #3022, opened 2019-12-11
pylint: remove `g-import-not-at-top` suppressions

wchargin:
Summary:
This is a Google-specific lint that we kind of just cargo cult around
and don’t actually lint against in CI, which makes it less than useful.

Generated with:

```
git ls-files -z '*.py' |
    xargs -0 sed -i \
        -e 's/g-import-not-at-top,//' \
        -e 's/,g-import-not-at-top//' \
        -e 's/ *# pylint:.*=g-import-not-at-top$//' \
        ;
```

Test Plan:
Our standard lint still passes:

```
$ flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
0
```

wchargin-branch: unsuppress-nontop-imports

### DIFF
--- a/tensorboard/plugins/debugger/debugger_plugin_loader.py
+++ b/tensorboard/plugins/debugger/debugger_plugin_loader.py
@@ -128,7 +128,7 @@ the interactive Debugger Dashboard. This flag is mutually exclusive with
     if flags.debugger_data_server_grpc_port > 0 or flags.debugger_port > 0:
       # Verify that the required Python packages are installed.
       try:
-        # pylint: disable=g-import-not-at-top,unused-import
+        # pylint: disable=unused-import
         import tensorflow
       except ImportError:
         raise ImportError(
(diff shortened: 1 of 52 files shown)


tensorboard/plugins/debugger/debugger_plugin_loader.py
```diff
@@ -128,17 +128,17 @@ def load(self, context):
...
+        # pylint: disable=unused-import
         import tensorflow
       except ImportError:
         raise ImportError(
             'To use the debugger plugin, you need to have TensorFlow installed:\n'
             '  pip install tensorflow')
 
     if flags.debugger_data_server_grpc_port > 0:
-      # pylint: disable=line-too-long,g-import-not-at-top
+      # pylint: disable=line-too-long
```
nfelt: Optional since orthogonal to this PR but conceptually makes sense to kill these as of #3021, right?

tensorboard/plugins/interactive_inference/interactive_inference_plugin_loader.py
```diff
@@ -37,10 +37,10 @@ def load(self, context):
       A InteractiveInferencePlugin instance or None if it couldn't be loaded.
     """
     try:
-      # pylint: disable=g-import-not-at-top,unused-import
+      # pylint: disable=unused-import
       import tensorflow
     except ImportError:
       return
-    # pylint: disable=line-too-long,g-import-not-at-top
+    # pylint: disable=line-too-long
```
nfelt: Also kill this one?

tensorboard/plugins/debugger/debugger_plugin_loader.py
wchargin: Yeah—I didn’t want to bother automatically matching those in #3021
because they’re not directly on the import lines, which makes it
non-trivial, whereas `g-import-not-at-top` can be removed in all cases.
But I’m happy to remove those; almost all our `line-too-long`s are
useless in one way or another.

wchargin: Done in #3025.

PR #3024, opened 2019-12-12
python: fix invalid escape sequences

wchargin:
Summary:
Python has historically been lax about invalid escapes: `"m\n\o"` is a
four-character string `m<Newline><Backslash>o`. As of Python 3.8, this
is a syntax warning. These can all be fixed by explicitly escaping the
backslashes or just using raw strings.

Test Plan:
All of our Python files now parse without warning on Python 3.8:

```
~/git/cpython/python - <<EOF
import ast
import glob
for f in glob.glob("tensorboard/**/*.py", recursive=True):
    ast.parse(open(f).read(), filename=f)
EOF
```

wchargin-branch: fix-py38-escapes

### DIFF
--- a/tensorboard/backend/application_test.py
+++ b/tensorboard/backend/application_test.py
@@ -591,7 +591,7 @@ class ParseEventFilesSpecTest(tb_test.TestCase):
     self.assertPlatformSpecificLogdirParsing(
         posixpath, '/lol/cat', {'/lol/cat': None})
     self.assertPlatformSpecificLogdirParsing(
-        ntpath, 'C:\\lol\cat', {'C:\\lol\cat': None})
+        ntpath, r'C:\lol\cat', {r'C:\lol\cat': None})
 
   def testRunName(self):
     self.assertPlatformSpecificLogdirParsing(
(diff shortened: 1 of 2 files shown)


tensorboard/backend/application_test.py
```diff
@@ -623,10 +623,10 @@ def testExpandsUserForMultipleDirectories(self):
       self.assertPlatformSpecificLogdirParsing(
           posixpath, 'a:~/lol,b:~/cat',
           {'/usr/eliza/lol': 'a', '/usr/eliza/cat': 'b'})
-      os.environ['HOME'] = 'C:\\Users\eliza'
+      os.environ['HOME'] = r'C:\Users\eliza'
```
stephanwlee: \e is highlighted funky. Is that just GitHub not understanding r'foo'?

wchargin: Yeah, amusingly, that’s GitHub doubly broken: `\e` is an escape
character in lots of languages (for `\x1b`), but not Python, and of
course this is a raw string anyway, so the highlight on `\U` is also
wrong. ¯\\\_(ツ)\_/¯

PR #3031, opened 2019-12-14
Add vscode to gitignore

psybuzz:
Simply adds the folder to .gitignore

Ran `git status` before and after, saw that the .vscode folder was ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 *.egg-info/
 *.swp
 node_modules
+.vscode/


wchargin:
To be a bit more explicit, ignores should be added at the level of the
entity that creates them. All TensorBoard developers will generate
`bazel-*` directories and `*.pyc` bytecode caches, so those are in the
project `.gitignore`. But various people use IntelliJ or Emacs or Vim or
VS Code, so `.idea` and `~*` and `*.swp` and `.vscode` get ignored at
the user level.

stephanwlee:
I was going to cite William but I guess there is no need :)
I agree with the principle here.

psybuzz:
Thanks, I didn't know global gitignore was a thing!  I'll close this

Issue #2924, opened 2019-11-12
Style guide update


psybuzz:
Consider renaming angular template filenames to end with ".ng.html"
Consider using snake case names for all other files, e.g. foo_module, foo_component.

PR #3037, opened 2019-12-16
Rename Angular files to match internal style guide

psybuzz:
#2924

For all frontend files under `tensorboard/webapp/`, this PR changes the filenames in 2 ways
  - Angular templates `foo.html` > `foo.ng.html`
  - convert to snake case `foo.component.ts` > `foo_component.ts`

I've verified that building the Angular webapp has no errors and renders correctly in the browser.  Also checked that `bazel test //tensorboard/webapp/...` tests pass.

Googlers, see b/144351474


wchargin:
Thank you! If it’s not too much trouble to run a test sync before
merging, that’d be much appreciated.

psybuzz:
Thanks for the review, I've made sure that an internal sync would work.  Merging.

PR #3050, opened 2019-12-18
Minor fixes to migration script

psybuzz:
Not sure why GitHub is showing this as 2 commits, when the 1st one should already be in tensorboard/migrate.

Changes
- Fallback in case process.stdout.columns is not defined (e.g. running in ndb)
- getDomModules() used to be able to return an array with 'undefined' entries

### DIFF
--- a/tensorboard/tools/migration/src/polymer-html.ts
+++ b/tensorboard/tools/migration/src/polymer-html.ts
@@ -50,6 +50,7 @@ function findDomModuleStatement(
   return node.statements.find((statement) => {
     return (
       ts.isClassDeclaration(statement) &&
+      statement.decorators &&
       statement.decorators.find((decorator) => {
         if (!ts.isCallExpression(decorator.expression)) return false;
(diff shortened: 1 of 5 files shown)


tensorboard/tools/migration/src/polymer-html.ts
```diff
@@ -40,7 +40,7 @@ function getDomModules(
 
     const template = parse5.serialize((templateNode as any).content).trim();
     return {id, template};
-  });
+  }).filter(x => !!x);
```
stephanwlee: Oh, I have a fix for this in another branch but LGTM

wchargin:
> Not sure why GitHub is showing this as 2 commits, when the 1st one
> should already be in tensorboard/migrate.

It appears that you cherry-picked a81e1f03a1b7 (which is in the repo
already) to form 5934b8fd0ba3 (which is not).

You can just rebase on top of `origin/migrate` to fix this.

PR #3054, opened 2019-12-19
Add Buildifier for linting BUILD files

### LINKED ISSUE #2963: Use Buildifier, enforcedly
We have a soft policy for BUILD file style: “try to use [`buildifier`],
but if you forget then don’t introduce spurious changes to fix unrelated
files”. This is not enforced, so our BUILD files are often “nearly
compliant”. We should enforce it.

Like #2466 and #2962, but easier because there’s only one option and
we’re already probably pretty close to parity, so we can just do it.

wchargin:
Fixes #2963. Three commits, to be rebased. The actual reformatting delta
is small (+78/−85), so I’ve lumped it into one commit.

As mentioned in the last commit, we don’t yet check `*.bzl` files
because that would force a global indentation change (~4000 lines
changed). I’m open to doing that if we want, not least because we also
use 4-space indentation for normal Python now.

Recommend reviewing the first commit with `git show -C` to find copies.

This will surely get Googlebot to complain about the PR author; I’ll
manually approve the CLAs right before we’re ready to merge it.


wchargin:
@googlebot: CLAs are fine because I wrote or generated all of these
commits. The author is intentionally tensorboard-gardener@google.com.

wchargin:
@stephanwlee: Oops, my mistake. I had that as an old name for what ended
up being called `download_executable.sh`. Sent #3071.

PR #3060, opened 2019-12-19
Add download links to HParams plugin

### LINKED ISSUE #2210: HParams Table Download as csv/markdown/latex
For the new awesome HParam feature, it would be great to download the generated tables in some format that's helpful for including in papers or documentation. I've integrated the feature into my company's internally facing tensorboard and besides all the awesome feedback this question keeps popping up.

I might have some bandwidth to contribute to this if you think it's valuable. Thanks!

eisenjulian:
Fixes #2210.

As suggested in #2210, the tables computed in the HParams dashboard are very useful and are currently hard to extract, either by inspecting the DOM or calling the API. This provides a more clear UI to dump the tables and use directly in downstream tasks.

Added another endpoint to the HParams backend to support downloading the table with session groups in different format

* Screenshots of UI changes
Screenshot 2019-12-19 at 22 48 43

Create mock HParams data by using the script at

> bazel run //tensorboard/plugins/hparams:hparams_minimal_demo

And then verifying the download links at 
> bazel run //tensorboard -- --logdir /tmp/hparams_minimal_demo

New tests are run with
> bazel test //tensorboard/plugins/hparams:download_data_test

The buttons could have been inside the table but I found hard to find the right location without disrupting the layout. The current location on the sidebar is consistent with the way in which the data is fetched, no matter the rendering chosen

### DIFF
--- a/tensorboard/plugins/hparams/download_data.py
+++ b/tensorboard/plugins/hparams/download_data.py
@@ -99,7 +99,7 @@ class Handler(object):
                 elif isinstance(value, float):
                     if math.isnan(value):
                         return r"$\mathrm{NaN}$"
-                    if not math.isfinite(value):
+                    if value in (float('inf'), float('-inf')):
                         return r"$%s\infty$" % ("-" if value < 0 else "+")
                     scientific = "%.3g" % value
                     if "e" in scientific:
--- a/tensorboard/plugins/hparams/download_data.py
+++ b/tensorboard/plugins/hparams/download_data.py
@@ -18,11 +18,12 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import csv
+import math
+
+import six
 
 from tensorboard.plugins.hparams import error
-from six import StringIO
-import math
-import csv
 
 
 class OutputFormat(object):
@@ -36,7 +37,14 @@ class OutputFormat(object):
 class Handler(object):
     """Handles a DownloadData request."""
 
-    def __init__(self, context, experiment, session_groups, response_format):
+    def __init__(
+        self,
+        context,
+        experiment,
+        session_groups,
+        response_format,
+        columns_visibility,
+    ):
         """Constructor.
 
         Args:
@@ -44,11 +52,13 @@ class Handler(object):
           experiment: Experiment proto.
           session_groups: ListSessionGroupsResponse proto.
           response_format: A string in the OutputFormat enum.
+          columns_visibility: A list of boolean values to filter columns.
         """
         self._context = context
         self._experiment = experiment
         self._session_groups = session_groups
         self._response_format = response_format
+        self._columns_visibility = columns_visibility
 
     def run(self):
         """Handles the request specified on construction.
@@ -60,6 +70,7 @@ class Handler(object):
         experiment = self._experiment
         session_groups = self._session_groups
         response_format = self._response_format
+        visibility = self._columns_visibility
 
         header = []
         for hparam_info in experiment.hparam_infos:
@@ -67,6 +78,11 @@ class Handler(object):
 
         for metric_info in experiment.metric_infos:
             header.append(metric_info.display_name or metric_info.name.tag)
+
+        def _filter_columns(row):
+            return [value for value, visible in zip(row, visibility) if visible]
+
+        header = _filter_columns(header)
 
         rows = []
 
@@ -80,13 +96,21 @@ class Handler(object):
             # hyperparameter values can be optional in a session group
             return ""
 
+        def _get_metric_id(metric):
+            return metric.group + "." + metric.tag
+
         for group in session_groups.session_groups:
             row = []
             for hparam_info in experiment.hparam_infos:
                 row.append(_get_value(group.hparams[hparam_info.name]))
+            metric_values = {}
             for metric_value in group.metric_values:
-                row.append(metric_value.value)
-            rows.append(row)
+                metric_id = _get_metric_id(metric_value.name)
+                metric_values[metric_id] = metric_value.value
+            for metric_info in experiment.metric_infos:
+                metric_id = _get_metric_id(metric_info.name)
+                row.append(metric_values.get(metric_id))
+            rows.append(_filter_columns(row))
 
         if response_format == OutputFormat.JSON:
             mime_type = "application/json"
@@ -94,7 +118,9 @@ class Handler(object):
         elif response_format == OutputFormat.LATEX:
 
             def latex_format(value):
-                if isinstance(value, int):
+                if value is None:
+                    return "-"
+                elif isinstance(value, int):
                     return "$%d$" % value
                 elif isinstance(value, float):
                     if math.isnan(value):
@@ -124,7 +150,7 @@ class Handler(object):
             bottom_part = "\\hline\n\\end{tabular}\n\\end{table}\n"
             body = top_part + header_part + middle_part + bottom_part
         elif response_format == OutputFormat.CSV:
-            string_io = StringIO()
+            string_io = six.StringIO()
             writer = csv.writer(string_io)
             writer.writerow(header)
             writer.writerows(rows)
--- a/tensorboard/plugins/hparams/download_data.py
+++ b/tensorboard/plugins/hparams/download_data.py
@@ -99,7 +99,7 @@ class Handler(object):
                 elif isinstance(value, float):
                     if math.isnan(value):
                         return r"$\mathrm{NaN}$"
-                    if value in (float('inf'), float('-inf')):
+                    if value in (float("inf"), float("-inf")):
                         return r"$%s\infty$" % ("-" if value < 0 else "+")
                     scientific = "%.3g" % value
                     if "e" in scientific:
--- a/tensorboard/plugins/hparams/download_data.py
+++ b/tensorboard/plugins/hparams/download_data.py
@@ -1,0 +1,137 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Classes and functions for handling the DownloadData API call."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+
+from tensorboard.plugins.hparams import error
+from six import StringIO
+import math
+import csv
+
+
+class OutputFormat(object):
+    """An enum used to list the valid output formats for API calls."""
+
+    JSON = "json"
+    CSV = "csv"
+    LATEX = "latex"
+
+
+class Handler(object):
+    """Handles a DownloadData request."""
+
+    def __init__(self, context, experiment, session_groups, response_format):
+        """Constructor.
+
+        Args:
+          context: A backend_context.Context instance.
+          experiment: Experiment proto.
+          session_groups: ListSessionGroupsResponse proto.
+          response_format: A string in the OutputFormat enum.
+        """
+        self._context = context
+        self._experiment = experiment
+        self._session_groups = session_groups
+        self._response_format = response_format
+
+    def run(self):
+        """Handles the request specified on construction.
+
+        Returns:
+          A response body.
+          A mime type (string) for the response.
+        """
+        experiment = self._experiment
+        session_groups = self._session_groups
+        response_format = self._response_format
+
+        header = []
+        for hparam_info in experiment.hparam_infos:
+            header.append(hparam_info.display_name or hparam_info.name)
+
+        for metric_info in experiment.metric_infos:
+            header.append(metric_info.display_name or metric_info.name.tag)
+
+        rows = []
+
+        def _get_value(value):
+            if value.HasField("number_value"):
+                return value.number_value
+            if value.HasField("string_value"):
+                return value.string_value
+            if value.HasField("bool_value"):
+                return value.bool_value
+            # hyperparameter values can be optional in a session group
+            return ""
+
+        for group in session_groups.session_groups:
+            row = []
+            for hparam_info in experiment.hparam_infos:
+                row.append(_get_value(group.hparams[hparam_info.name]))
+            for metric_value in group.metric_values:
+                row.append(metric_value.value)
+            rows.append(row)
+
+        if response_format == OutputFormat.JSON:
+            mime_type = "application/json"
+            body = dict(header=header, rows=rows)
+        elif response_format == OutputFormat.LATEX:
+
+            def latex_format(value):
+                if isinstance(value, int):
+                    return "$%d$" % value
+                elif isinstance(value, float):
+                    if math.isnan(value):
+                        return r"$\mathrm{NaN}$"
+                    if not math.isfinite(value):
+                        return r"$%s\infty$" % ("-" if value < 0 else "+")
+                    scientific = "%.3g" % value
+                    if "e" in scientific:
+                        coefficient, exponent = scientific.split("e")
+                        return "$%s\\cdot 10^{%d}$" % (
+                            coefficient,
+                            int(exponent),
+                        )
+                    return "$%s$" % scientific
+                return value.replace("_", "\\_").replace("%", "\\%")
+
+            mime_type = "application/x-latex"
+            top_part = "\\begin{table}[tbp]\n\\begin{tabular}{%s}\n" % (
+                "l" * len(header)
+            )
+            header_part = (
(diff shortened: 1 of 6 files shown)


tensorboard/plugins/hparams/hparams_plugin.py
```diff
@@ -90,6 +102,107 @@ def is_active(self):
...
+        if response_format == OutputFormat.JSON:
+            mime_type = "application/json"
+            body = dict(header=header, rows=rows)
+        elif response_format == OutputFormat.LATEX:
+
+            def latex_format(value):
+                if isinstance(value, int):
+                    return str(value)
+                elif isinstance(value, float):
+                    return "%.3g" % value
```
wchargin: Numbers should be emitted in math mode, so something like `"$%.3g$"` is
a closer format string, but I’m not sure that that’s the exponent
formatting that we want. What do you think of something like

```python
import math
import re

_RE_EXPONENT = re.compile(r"e(?:\+|(-))0*([0-9]*)")


def latex_format(value):
    ...
    if isinstance(value, float):
        if math.isnan(value):
            return r"$\mathrm{NaN}$"
        if not math.isfinite(value):
            return r"$%s\infty$" % ("-" if value < 0 else "+")
        scientific = "%.3g" % value
        mathmode = _RE_EXPONENT.sub(r" \\cdot 10^{\1\2}", scientific)
        return "$%s$" % mathmode
```

or similar?

Sample output:

```tex
$1 \cdot 10^{-50}$
$1 \cdot 10^{-5}$
$0.01$
$0$
$100$
$1 \cdot 10^{5}$
$1 \cdot 10^{50}$
```

tensorboard/plugins/hparams/hparams_plugin.py
```diff
@@ -90,6 +102,107 @@ def is_active(self):
...
+    def download_data_impl(self, experiment, session_groups, response_format):
+        """Provides a response for downloading data for an HParams table.
+
+        Args:
+          experiment: Experiment proto.
+          session_groups: ListSessionGroupsResponse proto.
+          response_format: A string in the OutputFormat enum.
+
+        Raises:
+          ValueError: If the scalars plugin is not registered.
```
wchargin: Stale docs?

tensorboard/plugins/hparams/hparams_plugin.py
```diff
@@ -90,6 +102,107 @@ def is_active(self):
...
+        rows = []
+
+        def _get_value(value):
+            if value.HasField("number_value"):
+                return value.number_value
+            if value.HasField("string_value"):
+                return value.string_value
+            if value.HasField("bool_value"):
+                return value.bool_value
+            return None
```
wchargin: Is this case meant to be reachable? If so, how—maybe comment? it’s not
obvious to me—and is `None` the right result? E.g., `latex_format(None)`
will raise a `TypeError`. If not, should we raise an `AssertionError` to
fail fast in case of a logic error?

tensorboard/plugins/hparams/hparams_plugin.py
```diff
@@ -90,6 +102,107 @@ def is_active(self):
...
+
+            def latex_format(value):
+                if isinstance(value, int):
+                    return str(value)
+                elif isinstance(value, float):
+                    return "%.3g" % value
+                return value.replace("_", "\\_").replace("%", "\\%")
+
+            mime_type = "application/x-latex"
+            top_part = "\\begin{table}[]\n\\begin{tabular}{%s}\n" % (
```
wchargin: Maybe add the default `[tbp]` layout specifier to avoid a `pdflatex`
warning?

tensorboard/plugins/hparams/hparams_plugin.py
```diff
@@ -90,6 +102,107 @@ def is_active(self):
...
+                "l" * len(header)
+            )
+            header_part = (
+                " & ".join(map(latex_format, header)) + " \\\\ \\hline\n"
+            )
+            middle_part = "".join(
+                [" & ".join(map(latex_format, row)) + " \\\\\n" for row in rows]
```
wchargin: Can drop the list comprehension to use a generator comprehension
directly, which avoids allocating the intermediate list:

```python
            middle_part = "".join(
                " & ".join(map(latex_format, row)) + " \\\\\n" for row in rows
            )
```

tensorboard/plugins/hparams/tf_hparams_backend/tf-hparams-backend.html
```diff
@@ -48,6 +48,17 @@
           return this._sendRequest('experiment', experimentRequest);
         }
 
+        getDowloadURL(format, listSessionGroupsRequest) {
```
wchargin: sp.: *dow**n**load*, and [please capitalize as `Url`][1].

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
```diff
@@ -201,35 +201,65 @@
...
+                    id="jsonLink"
+                    download="hparams_table.json"
+                    href="[[_jsonUrl()]]"
+                    >JSON</a
+                  >
+                  <a
+                    id="latexLink"
+                    download="hparams_table.tex"
+                    href="[[_latexUrl()]]"
+                    >LATEX</a
```
wchargin: Please spell as `LaTeX`.

eisenjulian:
Thanks for all the feedback, I updated the PR with all the suggestions.

tensorboard/plugins/hparams/download_data_test.py
```diff
@@ -0,0 +1,295 @@
...
+                "string_hparam",
+                "bool_hparam",
+                "optional_string_hparam",
+                "current_temp",
+                "delta_temp",
+                "optional_metric",
+            ],
+            "rows": [
+                [270.0, 150.0, "a string", True, "", 10.0, 15.0, 33.0],
+                [280.0, 100.0, "AAAAA", False, "", 51.0, 44.5],
```
wchargin: Thanks for adding these test cases! They’re clear and helpful.

It strikes me as a bit odd that optional metrics are silently omitted
from the output rather than having `null` entries or the like. If I had
two or more optional metrics, or optional metrics not at the end of the
list of metrics, how would I know which of the metrics were present?

For instance, if I had hparams `["x"]` and metrics `["y1", "y2", "y3"]`
all optional, how would we represent a row that had only `y1: 1.0` set
vs. a row that had only `y2: 1.0` set? (Same question goes for the other
output formats, too.)

tensorboard/plugins/hparams/download_data.py
```diff
@@ -0,0 +1,137 @@
...
+# limitations under the License.
+# ==============================================================================
+"""Classes and functions for handling the DownloadData API call."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+
+from tensorboard.plugins.hparams import error
```
wchargin: nit: Please [group and order imports per the style guide][1], and [only
import modules][2] (`import six`, not `from six import StringIO`).

[1]: https://google.github.io/styleguide/pyguide.html#313-imports-formatting
[2]: https://google.github.io/styleguide/pyguide.html#22-imports

wchargin:
@googlebot I consent.

sumanthratna:
It would be great if this plugin could also download an SVG of the parallel coordinates view and SVGs of the scatterplots. 

Downloading scatterplots shouldn't be too difficult, as scalars have this functionality (I believe). I can't speak on the feasibility of the parallel coordinates view (the SVG of the graph seems to be missing some styles so copy-pasting the SVG element doesn't work).

sumanthratna:
**IMPORTANT NOTE:** 

I just tested this feature with 1105 runs and the CSV only contained 100 entries, instead of all of them. Updating the paging so that all runs appear in the Table View doesn't work either. The same issue occurs with the JSON download and LaTeX download.

wchargin:
> I just tested this feature with 1105 runs and the CSV only contained
> 100 entries, instead of all of them. Updating the paging so that all
> runs appear in the Table View doesn't work either.

Confirmed—thanks for catching this, @sumanthratna!

It looks like the cause of the error is on the client, not the server:
the download URL’s `ListSessionGroupsRequest` has `"sliceSize":100`. The
download URLs are computed attributes (`href="[[_latexUrl()]]"`), but
don’t specify that they depend on the page size, which is why the
download URL is not being updated when you change the number of session
groups in the table pagination settings.

One way to fix this is to declare the appropriate dependencies, but it
seems likely that when people want to download data they’ll always want
to download all the data. Perhaps consider always setting `startIndex`
to `0` and `sliceSize` to `-1`, and modifying `list_session_groups.py`
such that when `slice_size` is `-1` it returns all the data (which it’s
already computed, anyway)?

tensorboard/plugins/hparams/download_data_test.py
eisenjulian: Good point, this slipped my mind. I updated the code to explicitly manage this cases.

wchargin: I don’t see any new commits in the PR—forgot to push?

eisenjulian:
Haven't pushed yet, I am fixing one more issue: It's not only the changes
to the slice size that don't get synced with the download data, other
checkboxes in the query panel also don't alter the link that gets computed
only once, because of the issue you described. I was planning on saving the
request params in the state every time the panel is changed and the table
updated. That said, do you still think we should always download all the
data or should we respect the paging config numbers?

>

wchargin:
I can see reasonable arguments both ways: you probably don’t want to
limit to an arbitrary page size, but it would be reasonable to expect
that filters on (e.g.) metric ranges or running states be reflected in
the download, and you could say that it’s weird to respect some of the
filters but not all of them. No strong preference here. If you don’t
care about this particular issue and want to do whatever’s easiest (and
at least somewhat reasonable), given that users can always set the page
size to 9999 manually, that’s totally fine with me. :-)

eisenjulian:
I have updated the branch fixing the issues. Regarding the empty hparams scenario, the download buttons are not rendered, as it happens with the rest of the selectors. If there are no session groups, everything should work but the tables would be empty.

eisenjulian:
Quick question, I saw some analytics code (which of course would work only within a specific deployment of tensorboard if opted into). Is there a way to track the usage of this download links? Do you think it is a good idea?

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
```diff
@@ -741,6 +799,7 @@
           // If query configuration is not valid don't send a request.
           return;
         }
+        this.set('_sessionGroupsRequest', request);
```
wchargin: OK—it’s a bit awkward to put this here, but I agree that it’s about the
best that can be done given the current code organization, and it
wouldn’t be fair to ask you to fix the existing design as part of this
change, so this is fine with me.

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
```diff
@@ -201,35 +201,65 @@
...
-              >
-              </paper-input>
+            <div class="section download">
+              <template is="dom-if" if="[[_sessionGroupsRequest]]">
+                Download data as
+                <span>
+                  <a
+                    id="csvLink"
+                    download="hparams_table.csv"
+                    href="[[_csvUrl(_sessionGroupsRequest, _hparams, _metrics)]]"
```
wchargin: These deps need to be on `_hparams.*` and `_metrics.*` (as in the
observer to `_updateConfiguration` above) for the events to listen
properly. As written, I can get the frontend into states where the table
view columns don’t match the columns in the downloaded table.

But simpler, I think, is to depend directly on `configuration` and pull
the visibility off of `configuration.columnsVisibility`. That way, you
don’t need to recompute it, and you’re depending on what you actually
need rather than just one of its dependents.

This patch works for me:

```diff
diff --git a/tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html b/tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
index 102adaee..55589ecd 100644
--- a/tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
+++ b/tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
@@ -241,19 +241,19 @@ TODO(erez): Add aggregation controls for repeated sessions.
                   <a
                     id="csvLink"
                     download="hparams_table.csv"
-                    href="[[_csvUrl(_sessionGroupsRequest, _hparams, _metrics)]]"
+                    href="[[_csvUrl(_sessionGroupsRequest, configuration)]]"
                     >CSV</a
                   >
                   <a
                     id="jsonLink"
                     download="hparams_table.json"
-                    href="[[_jsonUrl(_sessionGroupsRequest, _hparams, _metrics)]]"
+                    href="[[_jsonUrl(_sessionGroupsRequest, configuration)]]"
                     >JSON</a
                   >
                   <a
                     id="latexLink"
                     download="hparams_table.tex"
-                    href="[[_latexUrl(_sessionGroupsRequest, _hparams, _metrics)]]"
+                    href="[[_latexUrl(_sessionGroupsRequest, configuration)]]"
                     >LaTeX</a
                   >
                 </span>
@@ -589,17 +589,17 @@ TODO(erez): Add aggregation controls for repeated sessions.
         this._queryServer();
       },
 
-      _csvUrl(request, _, __) {
-        return this._downloadDataUrl(request, 'csv');
+      _csvUrl(request, configuration) {
+        return this._downloadDataUrl(request, configuration, 'csv');
       },
-      _jsonUrl(request, _, __) {
-        return this._downloadDataUrl(request, 'json');
+      _jsonUrl(request, configuration) {
+        return this._downloadDataUrl(request, configuration, 'json');
       },
-      _latexUrl(request, _, __) {
-        return this._downloadDataUrl(request, 'latex');
+      _latexUrl(request, configuration) {
+        return this._downloadDataUrl(request, configuration, 'latex');
       },
-      _downloadDataUrl(request, format) {
-        const visibility = this._computeColumnsVisibility();
+      _downloadDataUrl(request, configuration, format) {
+        const visibility = configuration.columnsVisibility;
         return this.backend.getDownloadUrl(format, request, visibility);
       },
 
```

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
eisenjulian: Yeah, I didn't love the solution, but it was the best I could think of.

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
eisenjulian: Thanks for the suggestion, much better :)

JimAva:
Is the CSV download feature for HParams available?  I'm trying to to extract the HParams and the Metrics results for all runs.  I'm able to extract the HParams. I can see the metric (loss) results in Table View but have not been able to extract them based on the code suggestion in #3091.

wchargin:
Hi @JimAva! This is in current nightlies, but has not yet been in a
stable release. You can see this by clicking on the merge commit:

![Screenshot of the “wchargin merged commit c6e181c” pull request timeline event][ss-1]

…then looking at the list of enclosing tags and branches:

![Screenshot of the commit page, with commit branches highlighted][ss-2]

This shows that this commit is currently only in `master`, and is not
reachable from any tags.

If you prefer a command line, `git describe --contains c6e181c53934`
asks which tags contain this commit; currently, that command fails,
because no tags do.

To get access to this feature via `tb-nightly`, first **uninstall** all
copies of `tensorboard` and `tb-nightly`, then `pip install tb-nightly`.
(The explicit uninstall is necessary to prevent Pip from corrupting your
installation.)

[ss-1]: https://user-images.githubusercontent.com/4317806/74002499-1baa0900-4924-11ea-9a7a-77c1a11b8ee1.png
[ss-2]: https://user-images.githubusercontent.com/4317806/74002546-472cf380-4924-11ea-8430-08ef977bd085.png

JimAva:
Thank you  @wchargin - how about the tensorboardcolab module?  Does that need to be removed and installed as well?

wchargin:
The `tensorboardcolab` module is not maintained by us (and I don’t know
what it does). You don’t need to install it. TensorBoard works in Colab
natively via the `%tensorboard` magic, which you can read about in [our
documentation][1].

[1]: https://www.tensorflow.org/tensorboard/tensorboard_in_notebooks

JimAva:
Works great!!! Thank you @wchargin

wchargin:
Glad to hear it! You’re quite welcome.

JimAva:
@wchargin - I've noticed that the download only exports 100 rows.  I have several jobs that create 100s of HParams rows.  Is this a known limitation?  I'm using the following command in m Colab notebook.  Am I missing any special parameters?

%tensorboard --logdir {logs_v2_Qmm2}

Here is the list of the runs:
!ls -ht $logs_v2_Qmm3/hparam_tuning | wc -l # count of files
result: >> **188**

wchargin:
@JimAva: The download link shows as many entries as are on the page,
which defaults to 100. Can you try changing the “Max # of session
groups” parameter in the bottom-left, which defaults to 100?

Screenshot of pagination selector

PR #3061, opened 2019-12-19
Migrate tf_backend

psybuzz:
Checked that `bazel build tensorboard/components/tf_backend` successfully builds.


psybuzz:
I've removed .prettierignore and package-lock.json.

tensorboard/components/tf_backend/tf-backend.ts
```diff
@@ -0,0 +1,28 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+export * from "../tf-imports/polymer.html";
+export * from "../tf-imports/lodash.html";
+export * from "../vz-sorting/vz-sorting.html";
+export * from "requestManager";
+export * from "urlPathHelpers";
+export * from "router";
```
stephanwlee: please fix this file, too.

psybuzz: Sorry, I forgot about that.  I've renamed it to `index.ts`, removed the non-backend exports, and made the module names relative paths.

tensorboard/components/tf_backend/index.ts
```diff
@@ -0,0 +1,25 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the 'License');
```
stephanwlee: In the license, the quotes should be double quotes. There are few rare instances when that is not true and you probably copied from the wrong file :(

psybuzz: Done

PR #3064, opened 2019-12-20
[DebuggerV2] Move constituent views of the debugger-v2 plugin to views/

caisq:
- Currently, there's only one such view (`Inactive`), but more will be added in the near future.


caisq:
Didn't realize you already noticed that issue, @wchargin . But anyway, I'll merge this PR, which will fix the issue.

wchargin:
In the future, let’s try not to squash totally unrelated changes into
the same PR. Unbreaking the build isn’t related to moving the debugger
plugin views.

PR #3066, opened 2019-12-21
[migration] Migrate vz_line_chart

stephanwlee:
This change converts HTML based vz_line_chart* to TypeScript version.


tensorboard/components/vz_line_chart/dragZoomInteraction.ts
```diff
@@ -1,227 +1,217 @@
...
+import * as Plottable from 'plottable';
+
+export class DragZoomLayer extends Plottable.Components.SelectionBoxLayer {
+  private _dragInteraction: Plottable.Interactions.Drag;
+  private _doubleClickInteraction: Plottable.Interactions.Click;
+  private easeFn: (t: number) => number = d3.easeCubicInOut;
+  private _animationTime = 750;
+  private onStart: () => void = () => void {};
+  private onEnd: () => void = () => void {};
+  private unzoomMethod: () => void;
```
psybuzz: Doesn't unzoomMethod also need to be initialized?

tensorboard/components/vz_line_chart2/line-chart-exporter.ts
```diff
@@ -12,136 +12,126 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+      node.nodeType == Node.ELEMENT_NODE &&
+      (nodeName == NodeName.DIV || nodeName == NodeName.SVG)
+    ) {
+      newNode = document.createElement(NodeName.GROUP);
+      const style = window.getComputedStyle(node as Element);
+      const left = parseInt(style.left || '', 10);
+      const top = parseInt(style.top || '', 10);
+      if (left || top) {
+        const clipId = this.createUniqueId('clip');
+        if (newNode instanceof Element) {
```
psybuzz: Aren't we guaranteed `newNode` is always an Element?

tensorboard/components/vz_line_chart2/line-chart-exporter.ts
```diff
@@ -12,136 +12,126 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+        const clipPath = document.createElementNS('svg', 'clipPath');
+        clipPath.id = clipId;
+        clipPath.appendChild(rect);
+        newNode.appendChild(clipPath);
       }
+    } else {
+      newNode = node.cloneNode();
+    }
+
+    if (newNode) {
```
psybuzz: Aren't we guaranteed newNode is valid by here?

tensorboard/components/vz_line_chart2/panZoomDragLayer.ts
```diff
@@ -12,147 +12,180 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-          return 'panning';
-        case State.DRAG_ZOOMING:
-          return 'drag-zooming';
-        case State.NONE:
-        default:
-          return '';
+    </style>
+  </custom-style>
+`;
+template.setAttribute('style', 'display: none;');
```
psybuzz: The display none is not needed

tensorboard/components/vz_line_chart2/vz-line-chart2.ts
```diff
@@ -1,454 +1,393 @@
...
-  const valueFormatter = vz_chart_helpers.multiscaleFormatter(
-    vz_chart_helpers.Y_TOOLTIP_FORMATTER_PRECISION
-  );
-  const formatValueOrNaN = (x) => (isNaN(x) ? 'NaN' : valueFormatter(x));
-
-  export const DEFAULT_TOOLTIP_COLUMNS = [
-    {
-      title: 'Name',
-      evaluate: (d: vz_chart_helpers.Point) => d.dataset.metadata().name,
+import {PolymerElement, html} from '@polymer/polymer';
```
psybuzz: license

tensorboard/components/vz_line_chart2/vz-line-chart2.ts
```diff
@@ -1,454 +1,393 @@
...
+    evaluate: (d: vz_chart_helpers.Point) =>
+      vz_chart_helpers.relativeFormatter(
+        vz_chart_helpers.relativeAccessor(d.datum, -1, d.dataset)
+      ),
+  },
+];
+
+type NumberFormatter = (num: number) => string;
+
+type EventAttachable = Node | Window;
```
psybuzz: Maybe just `EventTarget`, instead of creating a new type?

tensorboard/components/vz_line_chart/dragZoomInteraction.ts
stephanwlee: unzoomMethod is initialized in the constructor.

psybuzz: Ahh sorry, I didn't properly read!

tensorboard/components/vz_line_chart2/line-chart-exporter.ts
stephanwlee: `newNode` is typed as `Node | null` (L50). That's because (1) `node` in L49 is `Node` and (2) L77 clones `Node`, not `Element`.

tensorboard/components/vz_line_chart2/line-chart-exporter.ts
stephanwlee: Right, I think I meant to write `!` which should achieve the same effect (otherwise, it is nullable)

tensorboard/components/vz_line_chart2/line-chart-exporter.ts
psybuzz: I agree `newNode` is typed correctly at declaration on L50, but here on L62, we know that it's been set to an element because it passed L56's createElement('G').  So this `if` condition should return true, no matter what.  Would it make sense to cast it to Element inside this if branch?

stephanwlee: This is one way to do type coercion, too, though.

psybuzz: In my opinion, we should not mix type annotations with runtime logic in this way.  Before mixing, I have a guarantee that I can remove unnecessary code, without worrying that I might break the type checker.  Now, we have to be extra careful when rewriting any code we come across, to make distinguish what is actually there for logic vs. type coercion only.

tensorboard/components/vz_line_chart2/vz-line-chart2.ts
stephanwlee: Didn't know about this. Thanks!

PR #3070, opened 2019-12-23
Upgrade Travis to Xenial and floating tf-nightly

caisq:
  * Unblock TensorBoard Travis CI from using the latest (floating) tf-nightly

  * Update to .travis.yml to use Ubuntu 16.04 (xenial)
  * Update to .travis.yml to float the tf-nightly version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: xenial
 language: python
 python:
   - "3.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ env:
     - BUILDIFIER_SHA256SUM=4c985c883eafdde9c0e8cf3c8595b8bfdf32e77571c369bf8ddae83b042028d6
   matrix:
     - TF_VERSION_ID=tensorflow==1.15.0rc3
-    - TF_VERSION_ID=tf-nightly==2.1.0.dev20191222
+    - TF_VERSION_ID=tf-nightly
     - TF_VERSION_ID=  # Do not install TensorFlow in this case
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ env:
     - BUILDIFIER_SHA256SUM=4c985c883eafdde9c0e8cf3c8595b8bfdf32e77571c369bf8ddae83b042028d6
   matrix:
     - TF_VERSION_ID=tensorflow==1.15.0rc3
-    - TF_VERSION_ID=tf-nightly==2.1.0.dev20191118
+    - TF_VERSION_ID=tf-nightly==2.1.0.dev2019122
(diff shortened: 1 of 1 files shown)


wchargin:
We usually keep this floating, but have pinned it (in #2947) due to
low-level errors in TensorFlow. I can still reproduce the SIGABRT error
on today’s nightly (20191222), and this is one of the cases where Travis
caching creates Heisenbugs, so it’s plausible that we’ll have to revert
this PR soon.

caisq:
@wchargin Thanks for the background info. I can see the exact double-free and size-corruption errors in the test logs of this PR: https://api.travis-ci.org/v3/job/628578238/log.txt

I've pinged the owner of the tensorflow issue again. This needs to be fixed soon, or it'll block the debugger work and the next release.

caisq:
@mihaimaruseac Trying out `dist: xenial` (Ubuntu 16.04). See https://github.com/tensorflow/tensorboard/pull/3070/commits/4aa30351fa9f6f091d16847287663cc3aa846837

caisq:
For motivation of the upgrade to xenial, see https://github.com/tensorflow/tensorflow/issues/34427#issuecomment-568549093

caisq:
@mihaimaruseac It worked! Thanks!!

wchargin:
(Also, the platform-specific segfaults aren’t exactly encouraging… it’s
good that this unblocks us in the short term, but it doesn’t sound like
the underlying problem has really been fixed.)

caisq:
@wchargin I'm not sure I agree with https://github.com/tensorflow/tensorboard/pull/3070#issuecomment-568566827 , considering the fact that the official System Requirements doc states that Ubuntu 16.04+ is required:

https://www.tensorflow.org/install/pip

cc @mihaimaruseac for context and any opinions on whether that related tensorflow issue should be closed or not.

wchargin:
Hmm, fair. I guess it could be a legitimate fix for an ABI change, then.

mihaimaruseac:
We switched to being maylinux2010 compliant and that required a modern toolchain which is not provided by Ubuntu14.

Furthermore, Ubuntu14 eol'ed this summer, so we can no longer support it as no patches would come to it and the toolchains.

mihaimaruseac:
If Ubuntu16 fixes this, it could be that the change that exposed the heisenbug causing the segfaults to show up is  the commit cherry-picked by https://github.com/tensorflow/tensorflow/pull/34764 Unfortunately, we cannot roll that back as it would break all manylinux2010 builds.

caisq:
@mihaimaruseac @wchargin  Thanks for the info. Given the documented system requirements of tensorflow, I think it's better to run tensorboard's CI on xenial. So I'm merging the PR now, there is another PR (#3051) of which the CI won't pass without this.

@nickfelt If you have further comments, please let me know and I'll be happy to follow up.

PR #3075, opened 2019-12-24
[DebuggerV2] Add basic ng impl of actions, store, data_source & new Alerts view

caisq:
  * Add HTTP data source implementation in data_soruce/tfdbg2_data_source.ts. Currently the data source doesn't exist in the backend, but will be added soon (#3051)
  * Add actions in actions/. Currently there is only a small number of actions, related to the loading of the debugger v2 plugin, loading of the debugger runs from the data source.
  * Add effects in effects/. Currently there is only one effect: dispatch actions on `debuggerLoaded`.
  * Add types, store and redudcer in store/.
  * Add a new view in view/alerts/. The new `tf-debugger-v2-alerts` element (`AlertsContainer`) is rendered conditionally in lieu of the existing `InactiveContainer` when there is at least 1 debugger run available via the data source.
  * Add unit tests:
    * For reducer: debugger_reducer_test.ts
    * For the overall DebuggerContainer: debugger_container_test.ts
    * To make //tensorboard/webapp:karma_test_chromium pass, we need to decouple the states of the PluginModule and those of the DebuggerModule. This is achieve by creating `TestingDebuggerModule` in testing/ and using it in /tensorboard/webapp.

The code pattern is largely based on the work of @stephanwlee

### DIFF
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/BUILD
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/BUILD
@@ -1,6 +1,7 @@
 package(default_visibility = ["//tensorboard:internal"])
 
 load("@npm_angular_bazel//:index.bzl", "ng_module")
+load("//tensorboard/defs:defs.bzl", "tf_ng_web_test_suite", "tf_ts_library")
 
 licenses(["notice"])  # Apache 2.0
 
@@ -19,9 +20,49 @@ ng_module(
         "debugger_component.ng.html",
     ],
     deps = [
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/actions",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/alerts",
         "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive",
+        "@npm//@angular/common",
         "@npm//@angular/core",
+        "@npm//@ngrx/effects",
         "@npm//@ngrx/store",
         "@npm//rxjs",
+    ],
+)
+
+tf_ts_library(
+    name = "test_lib",
+    testonly = True,
+    srcs = [
+        "debugger_container_test.ts",
+    ],
+    tsconfig = "//:tsconfig-test",
+    deps = [
+        ":debugger_v2",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/actions",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store:types",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/testing",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/alerts",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive",
+        "//tensorboard/webapp/angular:expect_angular_core_testing",
+        "@npm//@angular/common",
+        "@npm//@angular/compiler",
+        "@npm//@angular/core",
+        "@npm//@angular/platform-browser",
+        "@npm//@ngrx/store",
+        "@npm//@types/jasmine",
+    ],
+)
+
+tf_ng_web_test_suite(
+    name = "karma_test",
+    deps = [
+        ":test_lib",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store:debugger_store_test_lib",
     ],
 )
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/BUILD
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/BUILD
@@ -63,6 +63,7 @@ tf_ng_web_test_suite(
     name = "karma_test",
     deps = [
         ":test_lib",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects:debugger_effects_test_lib",
         "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store:debugger_store_test_lib",
     ],
 )
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
@@ -29,7 +29,7 @@ export abstract class Tfdbg2DataSource {
 
 @Injectable()
 export class Tfdbg2HttpServerDataSource implements Tfdbg2DataSource {
-  private readonly httpPathPrefix = '/data/plugin/debugger-v2';
+  private readonly httpPathPrefix = 'data/plugin/debugger-v2';
 
   constructor(private http: HttpClient) {}
 
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
@@ -13,15 +13,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {HttpClient} from '@angular/common/http';
-import {Injectable, InjectionToken} from '@angular/core';
+import {Injectable} from '@angular/core';
 import {Observable} from 'rxjs';
 import {DebuggerRunListing} from '../store/debugger_types';
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
-
-export const TFDBG2_DATA_SOURCE = new InjectionToken<Tfdbg2DataSource>(
-  'tfdbg2 data source'
-);
 
 export abstract class Tfdbg2DataSource {
   abstract fetchRuns(): Observable<DebuggerRunListing>;
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
@@ -1,0 +1,43 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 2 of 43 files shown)


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container.ts
```diff
@@ -12,20 +12,46 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+      [runIds]="runsIds$ | async"
+    ></debugger-component>
   `,
 })
-export class DebuggerContainer {
+export class DebuggerContainer implements OnInit {
+  readonly runs$ = this.store.pipe(
+    select(
+      createSelector(
+        getDebuggerRuns,
```
stephanwlee: Drive by comment: when there are no projection, you can just use `getDebuggerRuns` without `createSelector`

caisq: Thanks for the driveby. :) Done.

stephanwlee: Marry Christmas Eve 🎅

caisq: Happy Holidays! ❄️

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/BUILD
```diff
@@ -19,9 +20,49 @@ ng_module(
...
+        "@npm//@ngrx/store",
+        "@npm//@types/jasmine",
+    ],
+)
+
+tf_ng_web_test_suite(
+    name = "karma_test",
+    deps = [
+        ":test_lib",
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store:debugger_store_test_lib",
```
stephanwlee: Is this a dependency of the test lib up there?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container.ts
```diff
@@ -12,20 +12,39 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+        (runs): string[] => Object.keys(runs)
+      )
+    )
+  );
+
   constructor(private readonly store: Store<State>) {}
+
+  ngOnInit(): void {
+    this.store.dispatch(debuggerLoaded());
```
stephanwlee: Optional but consider this. There are rare cases when you want to use this. We should avoid that certain view is initialized if possible since that can lead to proliferation of bad patterns. Since view is a reflection of state in the store, effects and reducers should be able to update correctly without this. 

For instance, whatever you do here can probably be replaced with PluginChanged action, right?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Injectable()
+export class Tfdbg2HttpServerDataSource implements Tfdbg2DataSource {
+  private readonly httpPathPrefix = '/data/plugin/debugger-v2';
+
+  constructor(private http: HttpClient) {}
+
+  fetchRuns() {
+    // TODO(cais): Once the backend uses an DataProvider that unifies tfdbg and
+    // non-tfdbg plugins, switch to using `tf_backend.runStore.refresh()`.
```
stephanwlee: Is it referring to fact that the ngrx store does not contain the runs right now? Would it be possible to add it there if so? Maybe runs here is different than that in general sense?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/BUILD
```diff
@@ -0,0 +1,25 @@
...
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store",
+        "@npm//@angular/common",
+        "@npm//@angular/core",
+        "@npm//@ngrx/effects",
+        "@npm//@ngrx/store",
+        "@npm//rxjs",
+    ],
+)
+
+# TODO(cais): After TestBed.inject() is figured out in OSS, add
```
stephanwlee: Can you explain here in GitHub comments what this would mean? You seem to have written a good test with TestBed.get which is more or less equal to inject

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Requires to be exported for JSCompiler. JSCompiler, otherwise,
+   * think it is unused property and deadcode eliminate away.
+   */
+  /** @export */
+  readonly loadPluginsListing$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(debuggerLoaded),
+      withLatestFrom(this.store.select(getDebuggerRunsLoaded)),
+      filter(([, {state}]) => state !== DataLoadState.LOADING),
```
stephanwlee: Just to confirm, it is your intention to re fetch when it is loaded?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -0,0 +1,52 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+export const DEBUGGER_FEATURE_KEY = 'debugger';
+
+// TODO(cais): Maybe deduplicate this with tensorboard/webapp.
+export enum DataLoadState {
```
stephanwlee: You should certainly dedup ;) also this code, as discussed in offline, should move to webapp so that would be quite natural.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container.ts
stephanwlee: > Also, why do you say these are rare cases? The DebuggerContainer being loaded

It is rare because this information is easily attainable from store's perspective and view really should not fire an action with intention of telling the store. 

> which is at a higher level than the individual plugin (DebuggerV2 specifically), should know something about the internal workings of the plugin.

I am not sure if I agree with this. One of the point of building on the same framework and same instance of the store is better "integration". Being part of the greater system is not a wrong thing.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/BUILD
stephanwlee: Ah, I misread it. I really expected there to be only one test target for all karma based testing then misread this.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
stephanwlee: Right, the intention in that case was to re-fetch even when the data is already loaded because it is mutable on the backend.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
stephanwlee: > tensorboard/webapp/types/api.ts

I don't think the LoadState belongs in there. I would love it if it gets moved to tensorboard/webapp/types/data.ts or something to that effect.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
caisq: Acknowledged. The intention here is the same. The debugger data (the existence of a run) may also change in the backend.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
caisq: Done. I created a new file in the same directory (tensorboard/webapp/types/data.ts) and moved the `DataLoadState` and `LoadState` interfaces to that file. Changes are made accordingly in tensorboard/webapp and in tensorboard/plugins/debugger_v2.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container.ts
caisq: I added a TODO item here to explore consolidate this effect into the integrated webapp.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
```diff
@@ -0,0 +1,64 @@
...
+    it('set runsLoad to failed on request failure', () => {
+      const state = createDebuggerState({
+        runsLoaded: {state: DataLoadState.LOADING, lastLoadedTimeInMs: null},
+      });
+      const nextState = reducers(state, actions.debuggerRunsRequestFailed());
+      expect(nextState.runsLoaded.state).toEqual(DataLoadState.FAILED);
+      expect(nextState.runsLoaded.lastLoadedTimeInMs).toBeNull();
+    });
+
+    it('sets runsLoaded and runs on successful runs loading', () => {
```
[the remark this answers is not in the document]
caisq: Done. Thanks for the suggestion.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export interface DebuggerState {
+  // Names of the runs that are available.
+  runs: DebuggerRunListing;
+  runsLoaded: LoadState;
+}
+
+export interface State {
+  [DEBUGGER_FEATURE_KEY]?: DebuggerState;
```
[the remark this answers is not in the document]
caisq: I agree. Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors.ts
```diff
@@ -0,0 +1,44 @@
...
+/** @typehack */ import * as _typeHackStore from '@ngrx/store/store';
+
+const selectDebuggerState = createFeatureSelector<State, DebuggerState>(
+  DEBUGGER_FEATURE_KEY
+);
+
+export const getDebuggerRuns = createSelector(
+  selectDebuggerState,
+  (state: DebuggerState): DebuggerRunListing => {
+    return state.runs;
```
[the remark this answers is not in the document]
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container_test.ts
```diff
@@ -0,0 +1,132 @@
...
+    const fixture = TestBed.createComponent(DebuggerContainer);
+    fixture.detectChanges();
+    expect(dispatchSpy).toHaveBeenCalledWith(debuggerLoaded());
+  });
+
+  it('updates the UI to hide inactive component when store has 1 run', () => {
+    const fixture = TestBed.createComponent(DebuggerContainer);
+    fixture.detectChanges();
+
+    store.setState(
```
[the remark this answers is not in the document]
caisq: I see your point. I think this pattern (namely setting store state) is more aligned with the testing principle of "test behavior, not internal implementation". For one thing, it's more resistant against changes in how the details of the selectors change. So I prefer to keep the current pattern. (Also, as you may know, the component tests in tensorboard/webapp are based on this pattern, instead of the `overrideSelector()` pattern.

caisq:
Gentle ping :)

PR #3076, opened 2019-12-24
[ng] Expose tf_globals as Polymer component for interop

stephanwlee:
We need to be able to turn off the fake hashing in the storage module when using angular based TensorBoard.

### DIFF
--- a/tensorboard/components/tf_globals/globals-polymer.ts
+++ b/tensorboard/components/tf_globals/globals-polymer.ts
@@ -1,4 +1,4 @@
-/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


tensorboard/components/tf_globals/globals-polymer.ts
```diff
@@ -0,0 +1,23 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
caisq: 2020

tensorboard/components/tf_globals/globals-polymer.ts
```diff
@@ -0,0 +1,23 @@
...
+==============================================================================*/
+namespace tf_globals {
+  // HACK: this Polymer component allows stores to be accessible from
+  // tf-ng-tensorboard by exposing otherwise mangled smybols.
+  Polymer({
+    is: 'tf-globals',
+    _template: null, // strictTemplatePolicy requires a template (even a null one).
+    tf_globals: tf_globals,
+  });
+} // namespace tf_globals
```
caisq: style nit: put two spaces before `//`

tensorboard/components/tf_globals/globals-polymer.ts
stephanwlee: but... but... commit was made in 2019! Will fix it.

tensorboard/components/tf_globals/globals-polymer.ts
stephanwlee: hmm... prettier changes it back to single space. Will not fix.

PR #3079, opened 2019-12-26
Add list of cleanup commits for `blame.ignoreRevsFile`

wchargin:
Summary:
A new file lists hashes of commits that should be ignored by typical
invocations of `git blame`. To use this functionality, you must set the
`blame.ignoreRevsFile` Git config variable to point to this file. This
functionality requires Git 2.23 or newer.

Test Plan:
Before running `git config blame.ignoreRevsFile .git-blame-ignore-revs`,
asking for `git blame` on (say) `core_plugin.py` blamed all lines to a
single commit. After running that `git config` command, it blames most
lines to useful, older commits. (Some lines still blame to the reformat
commit, but they tend to be “uninteresting” blank lines or lines with
only punctuation.)

wchargin-branch: git-ignorerevsfile

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,86 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


wchargin:
Good idea. Done.

PR #3080, opened 2019-12-27
Fix broken What-If Tool path string

jameswex:

Fix string broken by previous commit.

String cleanup change https://github.com/tensorflow/tensorboard/commit/ff5d2b42c4184505b4417aadba25c499b05c1e8c#diff-14e865ec253ff0abe583324f8028144b accidentally replaced a '>' character with an 'r' character.


wchargin:
Sorry—thanks for the fix!

wchargin:
Cross-referencing #3049 (culprit) for posterity.

PR #3082, opened 2019-12-30
[JS] upgrade dependencies

stephanwlee:
In order to receive necessary security and bug fixes, we should seldom
upgrade our dependencies. This updates our shrinkwrapped dependencies
using `yarn upgrade`.


stephanwlee:
> If there are any specific security fixes that we’re trying to pull in,
we should also update package.json, right?

Yes and no. 6b555c5 updates our dependency on handlebar, but it is not our immediate dependency (all the changes in that commit are our transitive dependencies). Likewise, even without our immediate dependencies version bumps, our transitive dependencies can change and we seldom want to update them. 

I made changes to `package.json` to update few of our immediate dependencies (e.g., dependency on bazel is left out; we do not want to depend on 2.0 just yet).

PR #3086, opened 2019-12-30
Revert "[JS] upgrade dependencies (#3082)"

stephanwlee:
This reverts commit c127fa418c90357951b6d0e76cc62abe5b3fea48.

It breaks ng_index.js due to `__values` function missing in a closure.
Source of real breakage is not identified.


wchargin:
For posterity, subsumed by fix-forward in #3087.

PR #3087, opened 2019-12-30
Revert zone.js upgrade and pin to 0.9

stephanwlee:
zone.js 0.10.2 makes the Angular TensorBoard build miss `__values`
symbol. Pin to the last good version for now.

This change also correctly generate the yarn.lock based on the
TypeScript requirement defined in package.json.

Additional information: the changelog for zone.js can be found [here](https://github.com/angular/angular/blob/master/packages/zone.js/CHANGELOG.md).


stephanwlee:
Yes, it is a fix forward since I timely found the culprit for the build error.

PR #3093, opened 2020-01-02
Updates images_demo for TF v2 compatiblity.

bileschi:
`bazel run tensorboard/plugins/image:images_demo` failed in TF v2+

modifies calls locations where API changed between v1 + v2

```
bazel run tensorboard/plugins/image:images_demo
tensorboard --logdir /tmp/images_demo
```

May be nice to produce a new version without compat calls (that wouldn't work with tf v1)

### DIFF
--- a/tensorboard/plugins/image/images_demo.py
+++ b/tensorboard/plugins/image/images_demo.py
@@ -53,6 +53,7 @@ IMAGE_CREDIT = textwrap.dedent(
 (IMAGE_WIDTH, IMAGE_HEIGHT) = (640, 480)
 _IMAGE_DATA = None
 
+tf.compat.v1.disable_eager_execution()
 
 def image_data(verbose=False):
     """Get the raw encoded image data, downloading it if necessary."""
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/image/images_demo.py
```diff
@@ -53,6 +53,7 @@
 (IMAGE_WIDTH, IMAGE_HEIGHT) = (640, 480)
 _IMAGE_DATA = None
 
+tf.compat.v1.disable_eager_execution()
```
stephanwlee: `tf.compat.v1.disable_v2_behavior` might be more correct to call.

tensorboard/plugins/image/images_demo.py
```diff
@@ -85,13 +86,13 @@ def convolve(image, pixel_filter, channels=3, name=None):
     Returns:
       A 3D `float32` `Tensor` of the same shape as the input.
     """
-    with tf.name_scope(name, "convolve"):
+    with tf.name_scope(name or "convolve"):
```
stephanwlee: wouldn't it be more correct to use `tf.compat.v1.name_scope`?

While at it, would it be possible to do `import tensorflowl.compat.v1 as tf` instead?

bileschi: Updated (requires removing existing compat.v1 calls)

tensorboard/plugins/image/images_demo.py
bileschi: done

PR #3094, opened 2020-01-02
ci: remove Python 2.7 from `.travis.yml`

wchargin:
Summary:
A new year, a new dawn. The <https://pythonclock.org/> has expired, and
TensorFlow has dropped Python 2 support:
<https://groups.google.com/a/tensorflow.org/d/msg/developers/ifEAGK3aPls/jlEuWYObBgAJ>

As noted in the above announcement, TensorFlow 2.1.0 will be released
with Python 2.7 support, but we’ve already released the corresponding
TensorBoard 2.1.0, so we no longer need to test with Python 2 at head.

Test Plan:
That CI passes quickly suffices.

wchargin-branch: travis-nopy2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ dist: xenial
 language: python
 python:
   - "3.6"
-  - "2.7"
 
 branches:
   only:


caisq:
I'm approving this PR based on the following considerations: 
- It'll ease the load on tensorboard's CI on Travis, which is currently pretty slow and puts a hindrance on developer speed.
- Officially, python2 has reached its EOL.

There might be other constraints that require tensorboard to be python2-compatible for a certain period of time (e.g., 1 year) from now. If the constraint emerges, we can revisit this.

wchargin:
Agreed, and note also that the `2.1` release branch still tests against
Python 2.7, so any patch releases will run against 2.7 as well by
default, too. This is already stronger than what TensorFlow promises in
their announcement. Of course, as you note, we can always adjust in
either direction.

PR #3095, opened 2020-01-02
[migration] computed decorator should not be used on complex properties

psybuzz:
According to [polymer docs](https://github.com/Polymer/polymer-decorators#computedtargets-string), complex properties that require wildcards or special props like 'reflectToAttribute' cannot use the @computed decorator.

### DIFF
--- a/tensorboard/tools/migration/src/polymerelementify.ts
+++ b/tensorboard/tools/migration/src/polymerelementify.ts
@@ -322,8 +322,15 @@ function getPropsAstNodes(props: ts.PropertyAssignment) {
           ? valueInitializer.initializer
           : undefined;
       }
+      const allowedPropertiesForComputedDecorator = new Set([
+        'computed',
+        'type',
+      ]);
       const canExtractComputed =
         propProperties &&
+        propProperties.every((prop: ts.PropertyAssignment) => {
+          return allowedPropertiesForComputedDecorator.has(propName(prop));
+        }) &&
         (propProperties.find((prop: ts.PropertyAssignment) => {
           return (
             propName(prop) === 'computed' &&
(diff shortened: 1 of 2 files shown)


tensorboard/tools/migration/src/polymerelementify.ts
```diff
@@ -322,16 +322,20 @@ function getPropsAstNodes(props: ts.PropertyAssignment) {
           ? valueInitializer.initializer
           : undefined;
       }
+      const allowedPropertiesForComputedDecorator = ['type', 'computed'];
```
stephanwlee: seems like `Set` is more semantically meaningful here.

tensorboard/tools/migration/src/polymerelementify.ts
```diff
@@ -322,16 +322,20 @@ function getPropsAstNodes(props: ts.PropertyAssignment) {
...
+        }) &&
         (propProperties.find((prop: ts.PropertyAssignment) => {
           return (
             propName(prop) === 'computed' &&
             ts.isStringLiteral(prop.initializer) &&
             // "computed" decorator is incompatible with the preperty accessor
             // or "*" for fancy diff-binding.
             parsePolymerStringDeclaration(prop.initializer.text).args.every(
-              (arg) => !arg.includes('.')
+              (arg) => !arg.includes('.') && !arg.includes('*')
```
stephanwlee: Presence of `.` should be enough check. `*` often supersedes a `.`. Which case is this addressing?

tensorboard/tools/migration/src/polymerelementify.ts
psybuzz: Done

tensorboard/tools/migration/src/polymerelementify.ts
psybuzz: That's true!  Done

PR #3097, opened 2020-01-02
[migration] migrate components/experimental

psybuzz:


tensorboard/components/experimental/plugin_util/BUILD
```diff
@@ -1,39 +1,18 @@
...
     srcs = [
-        "plugin-host.html",
+        "index.ts",
+        "plugin-host-ipc.ts",
         "runs-host-impl.ts",
     ],
-    path = "/tf-plugin-util",
     deps = [
-        ":host_internals",
+        "//tensorboard/components/experimental/plugin_lib:message",
```
stephanwlee: WOOHOOO! Finally! Btw, I think we may want to move this to `plugin_util` instead.

psybuzz: Done!  Really happy to not have to keep 2 files in sync :D

PR #3102, opened 2020-01-03
ci: move linting to GitHub actions

wchargin:
Summary:
GitHub Actions is a new first-party CI service offered by GitHub. It
requires no extra permissions. Its concurrency limits are appealing,
at 20 workflows per repo (1 workflow ≈ 1 commit) and concurrent jobs
ranging from 20 (free tier) to 180 (enterprise tier), with the option to
run on your own servers if this isn’t enough.

This commit adds a workflow definition to replace only the lint phase of
our Travis workflow, to take advantage of the improved latency: order of
seconds from pushing a commit to seeing actionable failure messages.
A full workflow for building and running tests is also possible, but is
too slow for comfort because it runs entirely without cache, pending
some proposed improvements to the caching support for GitHub Actions.

Test Plan:
Note that this commit triggers a GitHub Actions workflow that succeeds,
and that all Travis builds are a bit faster due to not running lint.

wchargin-branch: gh-actions-lint-only

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,22 +24,15 @@ env:
   BUILDIFIER_SHA256SUM: '4c985c883eafdde9c0e8cf3c8595b8bfdf32e77571c369bf8ddae83b042028d6'
 
 jobs:
-  lint-flake8:
+  lint-python:
     runs-on: ubuntu-16.04
-    strategy:
-      fail-fast: false
-      matrix:
-        # flake8 should run on each Python version that we target,
-        # because the errors and warnings can differ due to language
-        # changes, and we want to catch them all.
-        python_version: ['3.6']
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-python@v1
         with:
-          python-version: ${{ matrix.python_version }}
+          python-version: '3.6'
           architecture: 'x64'
-      - name: 'Install flake8'
+      - name: 'Install black and flake8'
         run: |
           python -m pip install -U pip
           pip install -r ./tensorboard/pip_package/requirements_dev.txt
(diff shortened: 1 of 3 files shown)


wchargin:
Total latency for all lints is 55 seconds. :-)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          python -m pip install -U pip
+          pip install -r ./tensorboard/pip_package/requirements_dev.txt
+      - run: pip freeze --all
+      - name: 'Lint Python code for errors with flake8'
+        # See: http://flake8.pycqa.org/en/3.7.8/user/error-codes.html
+        # Use the comment '# noqa: <error code>' to suppress.
+        run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+
+  lint-black-and-yaml:
```
nfelt: It's slightly odd to me to combine these together since they're pretty much linting disjoint sets of files that aren't particularly related.  I guess it's because they both involve a pip install, but unlike flake8 they don't need to run for every python version we target?

In practice since we are only running flake8 against 3.6 anyway it seems maybe simpler to just have a `lint-python-and-yaml` option, or to break it down to be even more granular and have `lint-black` and `lint-yaml` (or `lint-docs`) separately.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+          sudo mv ~/buildifier /usr/local/bin/buildifier
+      - name: 'Lint BUILD files'
+        run: git ls-files -z '*BUILD' | xargs -0 buildifier --mode=check
+
+  check-misc:
+    runs-on: ubuntu-16.04
+    steps:
+      - uses: actions/checkout@v1
+      - run: ./tensorboard/tools/do_not_submit_test.sh
+      - run: ./tensorboard/tools/docs_list_format_test.sh
```
nfelt: Optional, but could maybe make sense to put this in a `lint-docs` rule along with the yamllint check so that the pattern of checks that fails is more organized along conceptual lines rather than the mechanism of linter used.

If we did that it could also make sense to run the URL mirror test as part of the `lint-build` check.

WDYT?

wchargin:
Sure—as you surmised, the intent of these groupings was to minimize the
number of VMs and Pip installs (within reason), but grouping them more
semantically is fine with me. Done on both counts.

PR #3103, opened 2020-01-03
[DebuggerV2] Implement the backend execution/digests route

caisq:
  * Add `serve_execution_digests()` application route to DebuggerV2Plugin. It serves the HTTP requests to `/execution/digests`.
  * The logic of this new method is implemented using the `read_blob_sequences()` and `read_blob()`  methods of the `LocalDebuggerV2DataProvider`. As designed, the tags carry semantics specific to this route, as will be the case for other routes.
  * Note that there is some slight revision of the original design: we no longer surface the idea of "books" and "pages" in the DataProvider API. These are internal implementation details of data backend that should be hidden at this API level for simplicity.
  * `DebuggerV2EventMultiplexer` now has a new method `ExecutionDigests` to fulfill the data requests.

Also in this PR:
* The `/runs` route now response with a JSON object instead of an JSON array, with `startTimeMs` and `tensorFlowVersion` information. This matches the FE type specs.
* Flesh out the `DebuggerV2Plugin.is_active()` method.

### DIFF
--- a/tensorboard/plugins/debugger_v2/debug_data_provider.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_provider.py
@@ -155,40 +155,17 @@ class LocalDebuggerV2DataProvider(provider.DataProvider):
             raise ValueError(
                 "run_tag_filter.runs is expected to be specified, but is not."
             )
-        if len(run_tag_filter.runs) != 1:
-            raise ValueError(
-                "run_tag_filter.runs is expected to have length 1, "
-                "but instead has length %d" % len(run_tag_filter.tags)
-            )
-        if not run_tag_filter:
-            raise ValueError(
-                "run_tag_filter is expected to be specified, but is not."
-            )
-        if not run_tag_filter.tags:
-            raise ValueError(
-                "run_tag_filter.tags is expected to be specified, but is not."
-            )
-        if len(run_tag_filter.tags) != 1:
-            raise ValueError(
-                "run_tag_filter.tags is expected to have length 1, "
-                "but instead has length %d" % len(run_tag_filter.tags)
-            )
 
-        run = next(iter(run_tag_filter.runs))
-        tag = next(iter(run_tag_filter.tags))
-        # TODO(cais): Can we do a loop here instead and relax the len == 1
-        # requirement?
-
-        if run in self._multiplexer.Runs():
-            if tag.startswith(EXECUTION_DIGESTS_BLOB_TAG_PREFIX):
-                blob_ref = provider.BlobReference(blob_key="%s.%s" % (tag, run))
-                output = {run: {tag: [blob_ref]}}
-                return output
-
-            else:
-                raise ValueError("Unrecognized tag: %s" % tag)
-        else:
-            return {}
+        output = dict()
+        for run in run_tag_filter.runs:
+            if run in self._multiplexer.Runs():
+                output[run] = dict()
+            for tag in run_tag_filter.tags:
+                if tag.startswith(EXECUTION_DIGESTS_BLOB_TAG_PREFIX):
+                    output[run][tag] = [
+                        provider.BlobReference(blob_key="%s.%s" % (tag, run))
+                    ]
+        return output
 
     def read_blob(self, blob_key):
         if blob_key.startswith(EXECUTION_DIGESTS_BLOB_TAG_PREFIX):
--- a/tensorboard/plugins/debugger_v2/debug_data_provider.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_provider.py
@@ -174,4 +174,4 @@ class LocalDebuggerV2DataProvider(provider.DataProvider):
                 self._multiplexer.ExecutionDigests(run, begin, end)
             )
         else:
-            raise ValueError("Unrecognized blob_key: %s" % key)
+            raise ValueError("Unrecognized blob_key: %s" % blob_key)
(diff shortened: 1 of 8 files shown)


tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -95,11 +148,30 @@ def list_blob_sequences(
...
+                        provider.BlobReference(blob_key="%s.%s" % (tag, run))
+                    ]
+        return output
 
     def read_blob(self, blob_key):
-        del blob_key  # Unused currently.
-        # TODO(cais): Implement this.
-        raise NotImplementedError()
+        if blob_key.startswith(EXECUTION_DIGESTS_BLOB_TAG_PREFIX):
+            run, begin, end = _parse_execution_digest_blob_key(blob_key)
```
wchargin: Just to make sure we’re on the same page: real data providers won’t be
able to dynamically generate blobs for particular plugins, so if you
want to be able to read from blobs identified by `(run, begin, end)`
then time series by these names will need to be created at upload time,
so `begin` and `end` will (probably) need to have fixed alignment and
fixed difference. For instance, perhaps `begin` is always a multiple of
10 and `end` is always `begin + 10`. We’re on the same page here, right?

tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -95,11 +148,30 @@ def list_blob_sequences(
     def read_blob_sequences(
         self, experiment_id, plugin_name, downsample=None, run_tag_filter=None
     ):
-        del experiment_id, plugin_name, downsample, run_tag_filter  # Unused.
-        # TODO(cais): Implement this.
-        raise NotImplementedError()
+        del experiment_id, downsample  # Unused.
+        if plugin_name != PLUGIN_NAME:
+            raise ValueError("Unsupported plugin_name: %s" % plugin_name)
+        if not run_tag_filter.runs:
```
wchargin: Should be `if run_tag_filter.runs is None`, lest this also fail when
`runs` is the empty list.

tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -95,11 +148,30 @@ def list_blob_sequences(
...
+            raise ValueError(
+                "run_tag_filter.runs is expected to be specified, but is not."
+            )
+
+        output = dict()
+        for run in run_tag_filter.runs:
+            if run in self._multiplexer.Runs():
+                output[run] = dict()
+            for tag in run_tag_filter.tags:
```
wchargin: A couple questions:

  - If `run not in self._multiplexer.Runs()`, then won’t `output[run]`
    below be a `KeyError`? Or is it impossible to hit that for some
    reason that I’m not seeing?

  - As written, `run_tag_filter.runs` needs to be specified, and
    `run_tag_filter.tags` doesn’t. This feels kind of backward to me.
    When `runs` isn’t specified, everything in `_multiplexer.Runs()`
    should match (see [usage of `_test_run_tag`][trt] in the multiplexer
    provider). But if `rtf.tags` is `None` then you’d have to synthesize
    a list of `(run, begin, end)` keys, and there doesn’t seem to be an
    obvious way to do that. As written, the code will just `TypeError`
    when trying to iterate `for tag in None`.

[trt]: https://github.com/tensorflow/tensorboard/blob/b98ae59ba9d63bd28ac5064723b54bf599868e15/tensorboard/backend/event_processing/data_provider.py#L232-L234

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -90,16 +90,28 @@ def setUp(self):
         wsgi_app = application.TensorBoardWSGI([self.plugin])
         self.server = werkzeug_test.Client(wsgi_app, wrappers.BaseResponse)
 
+    def _getExactlyOneRun(self):
```
wchargin: LGTM for tests.

tensorboard/plugins/debugger_v2/debug_data_provider.py
wchargin: > In general, are implementations of DataProvider allowed to have
> internal logic beyond, say, requesting a whole object from GCS and
> passing it along to the caller without processing? Is the answer is
> yes, then begin and end can be arbitrary valid integer values, because
> the DataProvider impl and take care of making multiple requests to the
> backend and slicing/dicing the results to fulfill the given begin and
> end values.

They’re not expected to, no. The point of the data provider abstraction
is to decouple the storage layer from the plugin space, turning an
*m*-times-*n* problem into an *m*-plus-*n* problem. We don’t want to
write and maintain extra code for the debugger plugin for each storage
layer (local TensorBoard’s from-disk reading, TensorBoard.dev, any
internal TensorBoard hostd instances…). This is why I’ve been calling
out all the cases where the plugin is implicitly depending on undefined
or special-cased behavior from this data provider, because when we put
in a real data provider it _won’t_ have any debugger-specific code.

> The `DebuggerV2Plugin` just needs to do the work of making multiple
> `read_blob()` calls + slicing/dicing the results according to begin
> and end.

Yes, and as @stephanwlee pointed out, you’ll want to be careful about
the total number of `read_blob` RPCs needed to service one request here.

caisq:
> > The DebuggerV2Plugin just needs to do the work of making multiple
> > read_blob() calls + slicing/dicing the results according to begin
> > and end.
> > 
> 
> Yes, and as @stephanwlee pointed out, you’ll want to be careful about
> the total number of read_blob RPCs needed to service one request here.

OK. We're on the same page. Note that the current code of DebuggerV2Plugin doesn't have the awareness that only certain begin/end values are valid yet. This is better left to the (hopefully near) future, when it becomes clearer what the backing storage for DebuggerV2 data in HTB will be and what the page size is etc.

PR #3105, opened 2020-01-04
[migration] migrate plugins/audio

psybuzz:


tensorboard/plugins/audio/tf_audio_dashboard/tf-audio-dashboard.ts
```diff
@@ -0,0 +1,188 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
stephanwlee: Although git did not think so, this is effectively a move from tf-audio-dashboard.html and we should keep the license year there. Change this to 2016.

tensorboard/plugins/audio/tf_audio_dashboard/tf-audio-dashboard.ts
```diff
@@ -0,0 +1,188 @@
...
+      (audio: TfAudioLoader) => {
+        audio.reload();
+      }
+    );
+  }
+  _shouldOpen(index: number) {
+    return index <= 2;
+  }
+  @computed('_runToTagInfo', '_selectedRuns', '_tagFilter')
+  get _categories() {
```
stephanwlee: can you do me a favor and specify the return type for this? This method is a bit more involved and I cannot picture what the return type should be.

tensorboard/plugins/audio/tf_audio_dashboard/tf-audio-loader.ts
```diff
@@ -0,0 +1,223 @@
...
+import {customElement, property, computed, observe} from '@polymer/decorators';
+import '@polymer/paper-slider';
+
+import * as tf_backend from '../../../components/tf_backend';
+import {formatDate} from '../../../components/tf_card_heading/util';
+import {runsColorScale} from '../../../components/tf_color_scale/colorScale';
+import '../../../components/tf_dashboard_common/tensorboard-color';
+import '../../../components/tf_markdown_view/tf-markdown-view';
+import {AudioBackendMetadataResponse, AudioBackendTagInfo} from './types';
+'use strict';
```
stephanwlee: Remove me. this is extraneous now.

tensorboard/plugins/audio/tf_audio_dashboard/tf-audio-dashboard.ts
psybuzz: Done

tensorboard/plugins/audio/tf_audio_dashboard/tf-audio-dashboard.ts
psybuzz: Done.  Yeah, I should have annotated this when migrating Images, too.

tensorboard/plugins/audio/tf_audio_dashboard/tf-audio-loader.ts
psybuzz: Done

Issue #2084, opened 2019-04-03
Keras callback creating .profile-empty file blocks loading data


wchargin:
Repro steps:

 1. Create a virtualenv with `tf-nightly-2.0-preview==2.0.0.dev20190402`
    and open two terminals in this environment.

 2. In one terminal, run the following simple Python script (but
    continue to the next step while this script is still running):

    ```py
    from __future__ import absolute_import
    from __future__ import division
    from __future__ import print_function

    import tensorflow as tf


    DATASET = tf.keras.datasets.mnist
    INPUT_SHAPE = (28, 28)
    OUTPUT_CLASSES = 10


    def model_fn():
      model = tf.keras.models.Sequential([
          tf.keras.layers.Input(INPUT_SHAPE),
          tf.keras.layers.Flatten(),
          tf.keras.layers.Dense(128, activation="relu"),
          tf.keras.layers.BatchNormalization(),
          tf.keras.layers.Dense(256, activation="relu"),
          tf.keras.layers.Dropout(0.2),
          tf.keras.layers.Dense(OUTPUT_CLASSES, activation="softmax"),
      ])
      model.compile(
          loss="sparse_categorical_crossentropy",
          optimizer="adagrad",
          metrics=["accuracy"],
      )
      return model


    def main():
      model = model_fn()
      ((x_train, y_train), (x_test, y_test)) = DATASET.load_data()
      model.fit(
          x=x_train,
          y=y_train,
          validation_data=(x_test, y_test),
          callbacks=[tf.keras.callbacks.TensorBoard()],
          epochs=5,
      )


    if __name__ == "__main__":
      main()
    ```

 3. Wait for (say) epoch 2/5 to finish training. Then, in the other
    terminal, launch `tensorboard --logdir ./logs`.

 4. Open TensorBoard and observe that both training and validation runs
    appear with two epochs’ worth of data:

    ![Screenshot just after launching TensorBoard](https://user-images.githubusercontent.com/4317806/55447830-409e5480-557a-11e9-9524-90cf31242ccc.png)

 5. As training continues, refresh TensorBoard and/or reload the page.
    Observe that validation data continues to appear, but training data
    has stalled—even after well after the training has completed, the
    plot is incomplete:

    ![Screenshot of bad state](https://user-images.githubusercontent.com/4317806/55447840-485df900-557a-11e9-9fae-d0b5756d6c3b.png)

 6. Kill the TensorBoard process and restart it. Note that the data
    appears as desired:

[...]

PR #3108, opened 2020-01-05
[backend] Split IsTensorFlowEventsFile into two separate functions

caisq:
  * Fix https://github.com/tensorflow/tensorboard/issues/2084
  * The approach is suggested by @wchargin and @nfelt in https://github.com/tensorflow/tensorboard/issues/2084#issuecomment-570685175

  * Split the logic of `IsTensorFlowEventsFile()` into two separate functions
    1. An unchanged `IsTensorFlowEventsFile()` function, which only checks the existence of the 'tfevents' substring in the path string.
    2. A new `IsSummaryEventsFile()` function, which returns `True` if and only if  `IsTensorFlowEventsFile()` returns `True` for the input path name *and* the path name does not end in the special suffix `.profile-empty`.
  * This prevents the `EventAccumulator` implementation from picking up the empty events.tfevents.*.profile-empty files, which under the single-event file mode, causes TensorBoard backend to stop reading the latest summaries in the other (i.e., main, non-profiler-generated) events.* file. The *.profile-empty file was designed to make the TensorBoard backend recognize the subfolder created by the Profile plugin as a valid sub logdir, event when it contains no other events files.

  * Added new unit tests
  * Manually verified that #2084 is resolved by running `bazel run -c opt tensorboard -- --logdir /path/to/logdir` using the reproduction code in #2084. 
    * Screenshot: 
image

### DIFF
--- a/tensorboard/backend/event_processing/event_accumulator.py
+++ b/tensorboard/backend/event_processing/event_accumulator.py
@@ -820,13 +820,13 @@ def _GeneratorFromPath(path):
     """Create an event generator for file or directory at given path string."""
     if not path:
         raise ValueError("path must be a valid string")
-    if io_wrapper.IsTensorFlowEventsFile(path):
+    if io_wrapper.IsSummaryEventsFile(path):
         return event_file_loader.EventFileLoader(path)
     else:
         return directory_watcher.DirectoryWatcher(
             path,
             event_file_loader.EventFileLoader,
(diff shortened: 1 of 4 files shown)


caisq:
Thanks, @wchargin . I'll wait for @nfelt 's comment.

caisq:
cc @qiuminxu for visibility re. the Profiler plugin.

Issue #893, opened 2018-01-19
Allow ad-hoc colors changing


martinpopel:
As discussed in #288 and #449, it is difficult to choose one color palette that suits everyone in all situations.
If there are many runs and I select just some of them (with a regex or toggling individual runs), I quite often end up with e.g., **red, red, orange, red** and I cannot distinguish the runs especially if the curves are crossing each other.
I suggest to add a drop-down color picker (e.g. using http://jscolor.com/) to choose a color for each run.
I plan to use this especially before posting TensorBoard screenshots to my colleagues.

nfelt:
This could be useful.  Note that jscolor itself though we couldn't use since it's GPL.

martinpopel:
OK, jscolor is not needed (it was just an example). Choosing e.g. from 16 pre-defined colors would be enough for my needs.

jart:
Our team probably can't do something like this now. I suspect configuring color might be more appropriate if done in code/config rather than UI.

martinpopel:
I understand there may be higher priority features to implement. Just note that configuring colors in code/config would not solve this issue.
The point is that I don't know in advance which runs I will want to compare. I want to change the colors ad hoc *from the GUI*.
Also in case of multiple users connecting to the same TensorBoard server, I think the colors should be per-user configurable, i.e. client-side.

It is a question whether to solve the run<-->color mapping (semi)permanently in the browser. If possible this may be useful.
Another issue with TensorBoard is that when I delete a run (e.g. to save disk space), all colors change, so I have to redo all screenshots again to have consistent colors.

netheril96:
A way to change color in config/code/command line is better than nothing, although I'd favor a Web UI too.

jart:
Is the root problem that colors are not predictable and might change as you add additional runs? I'm not sure how color selection works at the moment. Although one possible thing to consider is using `hash(run_name) % #colors` to make color selection somewhat more deterministic.

Could that solve your problem?

martinpopel:
The color of a given run changes only if I delete any older runs. This is bothersome, but it is not the issue I reported.

My issue is:
I have hundreds of runs in TensorBoard and I select various subsets using regex filters or checkboxes.
Each subset has at most eight runs, so it would be possible to use distinct color for each run, but quite often this is not the case.

Maybe there could be a button "change colors" which would automatically select colors so the currently selected runs have distinct colors - this would solve my issue.
[...]

PR #3109, opened 2020-01-06
[WIP] Ad-hoc run color changing

RuRo:

This is a proof of concept/work in progress PR for the #893 feature request. Currently, this just adds a "Reassign Colors" button under the "Toggle All Runs" button, which selects the colours for the runs randomly from the current pallet (instead of sequentially, which is the current default).

## Images
<details>

Before clicking the "Reassign Colors" button:
image

After:
image

</details>

## ToDo/Considerations

- This is a really bare-bones implementation, just to get things started and hopefully motivate other people to also contribute to this feature.
- The original feature request asked for a colour picker interface, which I might add later (I don't have a lot of free time, so any contributions/suggestions are welcome).
- There already exists a colour pallet system builtin into `colorScale`, so if a colour picker interface ever gets implemented, it would probably be pretty easy to also allow users to switch colour pallets.
- Currently, the color of the _plots_ gets updated on the next data reload and not immediately after pressing the "Reassign Colors" button. We should probably trigger an update in the `colorScale.reassignColors` function, but I didn't have time to figure out how to do that yet.
- It is completely untested apart from the "Scalars" tab, so there -may be- probably is some bugs.

### DIFF
--- a/tensorboard/components/tf_color_scale/colorScale.ts
+++ b/tensorboard/components/tf_color_scale/colorScale.ts
@@ -55,6 +55,15 @@ namespace tf_color_scale {
       }
       return this.identifiers.get(s) as string;
     }
+
+    public reassignColors(): this {
+      const old_identifiers = this.identifiers;
+      this.identifiers = d3.map();
+      old_identifiers.each((s, i) => {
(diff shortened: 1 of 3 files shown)


RuRo:
@googlebot I signed it!

PR #3122, opened 2020-01-09
backend: fix locking for `_plugin_to_tag_to_content`

wchargin:
Summary:
The `plugin_event_accumulator` tries to maintain fine-grained locks for
each plugin, but didn’t properly lock its table of locks, and didn’t
lock around accesses to the table of plugins. This commit replaces the
fine-grained locking with a simple coarse-grained lock.

Because all `dict`s involved are keyed only on Python strings, for which
`__hash__`, `__eq__`, and `__del__` are implemented in C, this should be
a no-op in CPython implementations. But correct concurrency is
[explicitly part of Google style][style], and without this patch some
subsequent changes would need to make this problem worse, so we might as
well just fix it Once and For All.

[style]: https://google.github.io/styleguide/pyguide.html#218-threading

Test Plan:
The primary safety concern due to this change is potential for liveness
failure, but this change basically contracts a number of locks into one,
so any lock waiting cycle that exists now would also have existed
previously.

wchargin-branch: mux-lock-locks

### DIFF
--- a/tensorboard/backend/event_processing/plugin_event_accumulator.py
+++ b/tensorboard/backend/event_processing/plugin_event_accumulator.py
@@ -156,6 +156,13 @@ class EventAccumulator(object):
         # first event encountered per tag, so we must store that first instance of
         # content for each tag.
         self._plugin_to_tag_to_content = collections.defaultdict(dict)
+        # Locks the dict `_plugin_to_tag_to_content` (but not its individual
+        # values), as well as the `_plugin_tag_locks` dict.
+        self._plugin_tag_metalock = threading.Lock()
+        # `_plugin_tag_locks[p]` locks the dict `_plugin_to_tag_to_content[p]`.
+        #
+        # Liveness condition: while `_plugin_tag_locks[p]` is held,
+        # `_plugin_tag_metalock` must not be waited-for.
         self._plugin_tag_locks = collections.defaultdict(threading.Lock)
 
         self.path = path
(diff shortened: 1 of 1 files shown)


tensorboard/backend/event_processing/plugin_event_accumulator.py
```diff
@@ -156,6 +156,13 @@ def __init__(
         # first event encountered per tag, so we must store that first instance of
         # content for each tag.
         self._plugin_to_tag_to_content = collections.defaultdict(dict)
+        # Locks the dict `_plugin_to_tag_to_content` (but not its individual
+        # values), as well as the `_plugin_tag_locks` dict.
```
nfelt: Is it really worth even keeping `_plugin_tag_locks` rather than just switching entirely to the coarse-grained lock?

The only thing we would lose is the ability to do the dict snapshotting of `plugin_to_tag_to_content[plugin_name]` for distinct plugins in a concurrent fashion, but given that this operation doesn't do any I/O I'm struggling to convince myself that concurrency is any benefit here, instead of just forcing those snapshots to be entirely serialized.  One can construct hypotheticals in which we have e.g. a "high-priority" plugin that might want to interrupt a "low-priority" plugin's expensive dict copy but that seems far over the line into premature optimization territory.

wchargin: Sure, that’s more than fine with me if we’re happy to do that. Any time
I _don’t_ have to think about the fact that CPython and C extensions can
throw an exception to any thread between arbitrary bytecode statements,
I’m happier for it.

nfelt:
`Properly lock the [singular] lock` now

wchargin:
> `Properly lock the [singular] lock` now

Darn, you’re right. That’s a much less fun title. :-)

PR #3123, opened 2020-01-09
[DebuggerV2] Implement backend routes: source_files

caisq:
  * Route source_files/list lists all files
  * Route source_files/source_file provides access to the content
    of a given source file.
  * In `debug_data_multiplexer`, now call `self._reader.update()` on a separate thread instead of on the main thread. This prevents blocking the TensorBoad backend for a long time when the size of the debugger data is large. This asynchronous reading behavior is overridden by a mock to a synchronous behavior during unit tests.

  * Unit tests added

### DIFF
--- a/tensorboard/plugins/debugger_v2/debug_data_provider.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_provider.py
@@ -86,8 +86,7 @@ def _parse_source_file_list_blob_key(blob_key):
     Returns:
       - run ID
     """
-    return blob_key[blob_key.index(".") + 1:]
-
+    return blob_key[blob_key.index(".") + 1 :]
 
 
 def source_file_list_run_tag_filter(run):
--- a/tensorboard/plugins/debugger_v2/debug_data_provider.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_provider.py
@@ -123,7 +123,7 @@ def _parse_source_file_blob_key(blob_key):
 
     Args:
       blob_key: The BLOB key to parse. By contract, it should have the format:
-       `${SOURCE_FILE_LIST_BLOB_TAG}_${index}.${run_id}`
+       `${SOURCE_FILE_BLOB_TAG_PREFIX}_${index}.${run_id}`
 
     Returns:
       - run ID, as a str.
--- a/tensorboard/plugins/debugger_v2/debug_data_provider.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_provider.py
@@ -37,6 +37,7 @@ PLUGIN_NAME = "debugger-v2"
 
 EXECUTION_DIGESTS_BLOB_TAG_PREFIX = "execution_digests"
 SOURCE_FILE_LIST_BLOB_TAG = "source_file_list"
+SOURCE_FILE_BLOB_TAG_PREFIX = "source_file"
 
 
 def execution_digest_run_tag_filter(run, begin, end):
@@ -76,6 +77,18 @@ def _parse_execution_digest_blob_key(blob_key):
     return run, begin, end
 
 
+def source_file_list_run_tag_filter(run):
+    """Create a RunTagFilter for listing source files.
+
+    Args:
+      run: tfdbg2 run name.
+
+    Returns:
+      `RunTagFilter` for listing the source files in the tfdbg2 run.
+    """
+    return provider.RunTagFilter(runs=[run], tags=[SOURCE_FILE_LIST_BLOB_TAG])
+
+
 def _parse_source_file_list_blob_key(blob_key):
     """Parse the BLOB key for source file list.
 
@@ -89,16 +102,37 @@ def _parse_source_file_list_blob_key(blob_key):
     return blob_key[blob_key.index(".") + 1 :]
 
 
-def source_file_list_run_tag_filter(run):
+def source_file_run_tag_filter(run, index):
     """Create a RunTagFilter for listing source files.
 
     Args:
       run: tfdbg2 run name.
+      index: The index for the source file of which the content is to be
+        accessed.
 
     Returns:
-      `RunTagFilter` for listing the sourc eilfes in the tfdbg2 run.
+      `RunTagFilter` for accessing the content of the source file.
     """
-    return provider.RunTagFilter(runs=[run], tags=[SOURCE_FILE_LIST_BLOB_TAG])
+    return provider.RunTagFilter(
+        runs=[run], tags=["%s_%d" % (source_file, index)],
+    )
+
+
+def _parse_source_file_blob_key(blob_key):
+    """Parse the BLOB key for accessing the content of a source file.
+
+    Args:
+      blob_key: The BLOB key to parse. By contract, it should have the format:
+       `${SOURCE_FILE_LIST_BLOB_TAG}_${index}.${run_id}`
+
+    Returns:
+      - run ID, as a str.
+      - File index, as an int.
+    """
+    key_body, run = blob_key.split(".", 1)
+    key_body = key_body[len(EXECUTION_DIGESTS_BLOB_TAG_PREFIX) :]
+    index = int(key_body.split("_")[1])
+    return run, index
 
 
 class LocalDebuggerV2DataProvider(provider.DataProvider):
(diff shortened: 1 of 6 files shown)


tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -89,6 +96,14 @@ def setUp(self):
         self.plugin = debugger_v2_plugin.DebuggerV2Plugin(context)
         wsgi_app = application.TensorBoardWSGI([self.plugin])
         self.server = werkzeug_test.Client(wsgi_app, wrappers.BaseResponse)
+        # The multiplexer reads data asynchrnously on a separate thread, so
+        # as not to block the main thread of the TensorBoard backend. During
+        # unit test, we disable the asynchronous behavior, so that we can
+        # load the debugger data synchronously on the main thread, which leads
+        # to determinisic behavior.
+        tf.compat.v1.test.mock.patch.object(
+            debug_data_multiplexer, "run_in_background", lambda target: target()
```
wchargin: This mocked-out `run_in_background` has pretty different semantics from
the real one: it should probably return `None` and not propagate errors
to the caller (either printing an error or failing the test would be
fine with me). Otherwise, the code under test could accidentally depend
on synchronous execution.

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -89,6 +96,14 @@ def setUp(self):
         self.plugin = debugger_v2_plugin.DebuggerV2Plugin(context)
         wsgi_app = application.TensorBoardWSGI([self.plugin])
         self.server = werkzeug_test.Client(wsgi_app, wrappers.BaseResponse)
+        # The multiplexer reads data asynchrnously on a separate thread, so
```
wchargin: It’s fine with me to enforce deterministic execution for testing, but
this also has the effect that we’re not testing the code paths the occur
when the reader hasn’t finished updating by the time that the other
methods are invoked, which will certainly occur in real code. Do you
think that it makes sense to add some tests where `run_in_background` is
stubbed out as a no-op rather than a synchronous execution?

(also, sp.: s/`asynchrnously`/`asynchronously`/)

tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -75,6 +77,63 @@ def _parse_execution_digest_blob_key(blob_key):
...
+        runs=[run], tags=["%s_%d" % (SOURCE_FILE_BLOB_TAG_PREFIX, index)],
+    )
+
+
+def _parse_source_file_blob_key(blob_key):
+    """Parse the BLOB key for accessing the content of a source file.
+
+    Args:
+      blob_key: The BLOB key to parse. By contract, it should have the format:
+       `${SOURCE_FILE_LIST_BLOB_TAG}_${index}.${run_id}`
```
wchargin: Should be `SOURCE_FILE_BLOB_TAG_PREFIX`?

tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -75,6 +77,63 @@ def _parse_execution_digest_blob_key(blob_key):
...
+    """
+    return provider.RunTagFilter(runs=[run], tags=[SOURCE_FILE_LIST_BLOB_TAG])
+
+
+def _parse_source_file_list_blob_key(blob_key):
+    """Parse the BLOB key for source file list.
+
+    Args:
+      blob_key: The BLOB key to parse. By contract, it should have the format:
+       `${SOURCE_FILE_LIST_BLOB_TAG}.${run_id}`
```
wchargin: I don’t understand this: above you set tag `SOURCE_FILE_LIST_BLOB_TAG`
with no `.${run_id}` suffix. Where is the run ID expected to come from?

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -27,9 +29,14 @@
 from tensorboard.backend import application
 from tensorboard.plugins import base_plugin
 from tensorboard.plugins.debugger_v2 import debugger_v2_plugin
+from tensorboard.plugins.debugger_v2 import debug_data_multiplexer
```
wchargin: nit: alphabetize import (`debug_*` < `debugger_*`)

tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
```diff
@@ -110,3 +116,72 @@ def serve_execution_digests(self, request):
...
+            return http_util.Respond(
+                request,
+                self._data_provider.read_blob(
+                    blob_sequences[run][tag][0].blob_key
+                ),
+                "application/json",
+            )
+        except IndexError as e:
+            return http_util.Respond(
+                request, {"error": str(e)}, "application/json", code=400,
```
wchargin: This is a bit of a confusing message—the “list index out of range”
doesn’t mean that the user-provided index was out of bounds but rather
that the `[0]` on line 180 above failed, right? The raw Python error
message is an implementation detail; do you think that it would be more
appropriate to raise “no source file %d for run %s” as a 404?

tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
```diff
@@ -110,3 +116,72 @@ def serve_execution_digests(self, request):
...
+            self._data_provider.read_blob(blob_sequences[run][tag][0].blob_key),
+            "application/json",
+        )
+
+    @wrappers.Request.application
+    def serve_source_file(self, request):
+        """Serves the content of a given source file.
+
+        The source file is referred to by the index in the list of all source files
+        invovled in the execution of the debugged program, which is available via the
```
wchargin: sp.: s/`invovled`/`involved`/

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -89,6 +96,14 @@ def setUp(self):
...
         wsgi_app = application.TensorBoardWSGI([self.plugin])
         self.server = werkzeug_test.Client(wsgi_app, wrappers.BaseResponse)
+        # The multiplexer reads data asynchrnously on a separate thread, so
+        # as not to block the main thread of the TensorBoard backend. During
+        # unit test, we disable the asynchronous behavior, so that we can
+        # load the debugger data synchronously on the main thread, which leads
+        # to determinisic behavior.
+        tf.compat.v1.test.mock.patch.object(
+            debug_data_multiplexer, "run_in_background", lambda target: target()
+        ).start()
```
wchargin: This patcher is never stopped—`self.addCleanup`?

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -278,6 +293,97 @@ def testServeExecutionDigests400ResponseIfRunParamIsNotSpecified(self):
...
+            _ROUTE_PREFIX + "/source_files/file?run=%s&index=%d" % (run, index)
+        )
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(
+            "application/json", response.headers.get("content-type")
+        )
+        data = json.loads(response.get_data())
+        self.assertEqual(data["host_name"], _HOST_NAME)
+        self.assertEqual(data["file_path"], _CURRENT_FILE_FULL_PATH)
+        with open(__file__, "rt") as f:
```
wchargin: nit: Text-mode is the default, and I’ve never seen it specified
explicitly before (I had to look it up to see what it did). If this
doesn’t do anything, maybe we should drop it, for readability?

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
caisq: I agree with the point that the logic in `run_in_background` ought to be tested. To address this comment, I added a new test file `debug_data_multiplexer_test.py`. The file currently has only a simple assertion, aided by some mocks. This is because the real `run_in_background` method currently just calls `target` once. When it gets more complex, I'll add more mock-based tests in the new file. I added a TODO item to remind myself of that.

wchargin: Sounds reasonable; thanks!

tensorboard/plugins/debugger_v2/debug_data_provider.py
wchargin: Ah, now I understand. Thanks for the explanation.

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
wchargin: No, the TensorBoard CI never complained about out-of-order imports (at
least, not any time in the past year or so). Our `flake8` checks are
minimal, only looking for obvious problems like syntax errors or unused
or undefined variables. Thanks for fixing. :-)

PR #3124, opened 2020-01-09
data: add `list_plugins` for `is_active` support

wchargin:
Summary:
The `is_active` method of TensorBoard plugins hasn’t worked well with
data providers so far, because it doesn’t take an `experiment_id`
argument, and adding one would break backward compatibility. Instead, we
can pull `is_active` handling into the core application logic, where
both the data provider and experiment ID are still available.

A new data provider method `list_plugins` takes an experiment ID and
returns names of plugins with relevant data. A new (experimental) plugin
API lets plugins specify whose data they want to read, defaulting to
just their own. The `/data/plugins_listing` logic now joins the list of
plugins with data (from the provider) with each plugin’s list of data
that it cares about to determine which plugins are active. As a fallback
for plugins whose data is not yet understood by the data provider, we
also check `is_active` for plugins that don’t pass the first test.

The new data provider method is optional (for now). If not implemented,
this patch has no effect.

As a side benefit, this can speed up the plugins listing route by a fair
amount: we only make one targeted query against the multiplexer rather
than querying all its `Tensors` once per plugin, and we don’t need to
invoke `is_active` at all for plugins that are known to be active by
this first check.

Test Plan:
Change the `is_active` methods for the scalars, graphs, and histograms
plugins to return `False` rather than `True` when a data provider is
set. Then, run TensorBoard in generic data mode pointing at (a) the
scalars demo, (b) the histograms demo, and (c) the images demo. Note
that in case (a) only the scalars and graphs dashboards are active, in
case (b) only the histograms and distributions dashboards are active
(this demo does not write graphs), and in case (c) only the images and
graphs dashboards are active. Note that this behavior is the same across
generic and non-generic data modes. Note that the behavior prior to this
patch is restored if the implementation of `list_plugins` in the
multiplexer-backed data provider is removed, so this is a no-op until
data providers choose to implement this method.

wchargin-branch: data-list-active-plugins


wchargin:
Verified that this has no visible change to TensorBoard.dev.

tensorboard/backend/application.py
```diff
@@ -451,14 +456,33 @@ def _serve_plugins_listing(self, request):
           A werkzeug.Response object.
         """
         response = collections.OrderedDict()
+        eid = plugin_util.experiment_id(request.environ)
+        plugins_with_data = (
+            self._data_provider.list_active_plugins(eid)
+            if self._data_provider is not None
+            else None
```
nfelt: Simpler below if this just defaults to the empty set?  Then we don't need a branch.

tensorboard/backend/application.py
```diff
@@ -451,14 +456,33 @@ def _serve_plugins_listing(self, request):
...
                 type(plugin) is core_plugin.CorePlugin
             ):  # pylint: disable=unidiomatic-typecheck
                 # This plugin's existence is a backend implementation detail.
                 continue
             start = time.time()
-            is_active = plugin.is_active()
+
+            if plugins_with_data is not None:
+                is_active = (
+                    any(
```
nfelt: Optional, but IMO more obvious if we just check for a non-empty set intersection?

tensorboard/backend/application_test.py
```diff
@@ -297,6 +321,46 @@ def testPluginsListing(self):
...
+
+        parsed_object = self._get_json("/data/plugins_listing")
+        actives = {k: v["enabled"] for (k, v) in parsed_object.items()}
+        self.assertEqual(
+            actives,
+            {
+                "foo": True,  # directly has data
+                "bar": False,  # has data, but does not depend on itself
+                "baz": False,  # no data, and no dependencies
+                "quux": True,  # no data, but depends on "baz"
```
nfelt: it's the dependency on "bar" that makes this active, right?  not "baz"

tensorboard/backend/event_processing/plugin_event_accumulator.py
```diff
@@ -278,6 +278,16 @@ def PluginTagToContent(self, plugin_name):
                     # before setting `released_metalock`.
                     pass
 
+    def ActivePlugins(self):
+        """Return a list of plugins with summary data.
```
nfelt: Any reason not to return a frozenset instead?  That way it's the same return type as on EventMultiplexer.

tensorboard/data/provider.py
```diff
@@ -52,6 +52,25 @@ def data_location(self, experiment_id):
         """
         return ""
 
+    def list_active_plugins(self, experiment_id):
```
nfelt: Maybe just `list_plugins()`?  Given that this is scoped to the experiment already, I'm not sure that "active" as a qualifier really means much in the data provider context?

tensorboard/data/provider.py
```diff
@@ -52,6 +52,25 @@ def data_location(self, experiment_id):
...
+        """List all plugins that own data in a given experiment.
+
+        This should be the set of all plugin names `p` such that calling
+        `list_scalars`, `list_tensors`, and `list_blob_sequences` for
+        the given `experiment_id` and plugin name `p` gives a non-empty
```
nfelt: typo in "gives a non-empty results"

tensorboard/plugins/base_plugin.py
```diff
@@ -97,6 +97,16 @@ def frontend_metadata(self):
         """
         return FrontendMetadata()
 
+    def relevant_summary_plugins(self):
```
nfelt: Naming nit: maybe something more like `data_plugin_names()`?

"Relevant" seems a little unnecessary IMO in that this is already clearly plugin-specific and the fact that it's plural already implies that it's not necessarily just this plugin's own name.

I'd maybe avoid "summary" because in the long run we don't necessarily consider everything exposed via the data provider to be "summary" data per se.  E.g. profile data might get exposed via the data provider API at some point, or projector data, and while those are maybe sort of still conceptually connected to the `tf.summary` API, they're never materialized as a literal `tf.Summary` proto so it seems a little odd to label them as such.  (I also just don't love the term "summary" in general since it's so disconnected from the normal English meaning of the word, so to the extent we can get away with other terminology I tend to favor that.)

nfelt: Design question: would it be worth exposing this instead on TBLoader?

It seems reasonable to assume that in general we expect this to be a fixed list / static property of the plugin, rather than something dynamically determined.  This would allow us to defer loading of the plugin until we encounter an experiment that actually needs it, which could help the backend start up faster and use less memory in a world where there are many dynamic plugins and they're more heavyweight.  We probably don't get much advantage from it right now (where the summary-based plugins tend to be the lightweight ones anyway), and there are some downsides as well (if the plugin takes a while to load or has an import-time error, this impact occurs later on, which might be bad), but I wonder if it might be more future-proof that way.

WDYT?

tensorboard/backend/application.py
nfelt: Out of laziness I probably would have had list_plugins() default to frozenset(), but it's true that it is possibly useful down the road to be able to distinguish the two cases so I'm fine leaving it as you have it here.

tensorboard/plugins/base_plugin.py
nfelt: Yeah, I agree it makes sense to defer worrying about this for the points you made.  I suspect if we really wanted to do proper dynamic plugin loading like that we'd maybe want to revisit the TBLoader/TBPlugin breakdown more holistically anyway.  But useful to hear your thoughts about this, thanks.

PR #3126, opened 2020-01-09
data: use `six.string_types` for `RunTagFilter` type checking

wchargin:
Summary:
We don’t generally support Python 2 anymore, but this is required in the
short term for an internal fix. (This fixes an error “expected `None` or
collection of strings” when the list has a `unicode` value.)

wchargin-branch: rtf-six-strings

### DIFF
--- a/tensorboard/data/provider.py
+++ b/tensorboard/data/provider.py
@@ -784,7 +784,7 @@ class RunTagFilter(object):
     def _parse_optional_string_set(self, name, value):
         if value is None:
             return None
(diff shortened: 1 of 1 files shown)


wchargin:
Thanks!

In Python 2, `"\u2026"` is a six-character `str` equal to `"\\u2026"`.
In particular, it is a bytestring. On the other hand, `u"\u2026"` is an
ellipsis, and is a `unicode` value: `not isinstance(u"\u2026", str)`.

PR #3127, opened 2020-01-09
TB versions after 2.1 should work with TF 2.1

psybuzz:
TB 2.1.0 works great with TF 2.1.0.  However, tip-of-tree TB doesn't work with TF 2.1.0, since ToT TB depends on TF's DebugDataReader, which was introduced after TF 2.1.0.  Running this combination produces:

```
AttributeError: module 'tensorflow.python.debug.lib.debug_events_reader' has no attribute 'DebugDataReader'
```

@caisq, do we plan to support, say, TB 2.2 or ToT TB working with TF 2.1?
If not, we can close this.

### DIFF
--- a/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
@@ -115,6 +115,10 @@ class DebuggerV2EventMultiplexer(object):
                 self._reader.update()
                 # TODO(cais): Start off a reading thread here, instead of being
                 # called only once here.
+            except AttributeError as error:
+                # Gracefully fail for users with TensorFlow 2.1.0 and older,
+                # which may have debug_events_reader without DebugDataReader.
+                return {}
             except ValueError as error:
                 # When no DebugEvent file set is found in the logdir, a
                 # `ValueError` is thrown.
--- a/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
@@ -117,8 +117,9 @@ class DebuggerV2EventMultiplexer(object):
                 # called only once here.
             except AttributeError as error:
                 # Gracefully fail for users without the required API changes to
-                # debug_events_reader introduced in TF 2.1.0.dev20200103.
-                # This should be safe to remove when TF 2.2 is released.
+                # debug_events_reader.DebugDataReader introduced in
+                # TF 2.1.0.dev20200103. This should be safe to remove when
+                # TF 2.2 is released.
                 return {}
             except ValueError as error:
                 # When no DebugEvent file set is found in the logdir, a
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -115,6 +115,10 @@ def Runs(self):
                 self._reader.update()
                 # TODO(cais): Start off a reading thread here, instead of being
                 # called only once here.
+            except AttributeError as error:
```
caisq: Instead of this, what do you think about checking

```py
if tensorflow.__version__ <= "2.1.0":
  return {}
```

psybuzz: I'd still like to investigate why the tb-nightly packager failed, and why it was using tensorflow 2.1.0.dev20200102 when it ran on Jan 9.  Then, it would be nice if we gracefully failed when running ToT TB with TF 2.1.0.dev20200102.

So it would make sense to do a check like:
`if parse_version(tf_version) < parse_version("2.1.0.dev20200103"):`

I don't fully understand why, but tf-nightlies are still marked with major/minor "2.1.0", so this check would fail.  That is, "2.1.0" was released before "2.1.0.dev20200103", but parse_version thinks that the former is greater than the latter.

Would it suffice to keep the AttributeError check for now, and have a more descriptive comment that it should be safe to remove when TF 2.2 is released and stable?

caisq: The change that started around 20200103 does look strange and needs investigation for full understanding. I'll approve the `AttributeError` approach now to unblock builds.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -115,6 +115,11 @@ def Runs(self):
...
+                # Gracefully fail for users without the required API changes to
+                # debug_events_reader introduced in TF 2.1.0.dev20200103.
```
caisq: nit: debug_events_reader.DebugDataReader to be precise.

psybuzz: Done.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
wchargin: > I don't fully understand why, but tf-nightlies are still marked with
> major/minor "2.1.0", so this check would fail. That is, "2.1.0" was
> released before "2.1.0.dev20200103", but parse_version thinks that the
> former is greater than the latter.

TensorFlow 2.1.0 was released on 2020-01-08, after the 2020-01-03
nightly.

The behavior of `parse_version` is correct: `.devWHATEVER` is a
[pre-release specifier, as listed in PEP 440][pep-440].

TensorFlow’s 2.1.0 version bump was just merged into master a few hours
ago, so we should expect tomorrow’s nightly to be `2.2.0.dev20200111`:
<https://github.com/tensorflow/tensorflow/commit/f6404f4f24e756196fd710c27243e10f625ac2e8>

[pep-440]: https://www.python.org/dev/peps/pep-0440/#pre-releases

PR #3128, opened 2020-01-10
windows: explicitly register correct MIME types

### LINKED ISSUE #3120: Hard-code some MIME types to work around broken Windows setups
Context:
<https://github.com/tensorflow/tensorboard/issues/2771#issuecomment-569810749>

TL;DR: We use `mimetypes.guess_type` from the Python standard library to
guess MIME types, but on Windows any application can corrupt the backing
registry, and it appears that some popular programs (maybe Visual
Studio) do just that. This breaks TensorBoard for users because we send
`text/plain` for scripts, blocking execution.

Proposed fix: Either (a) at TensorBoard `main` time, patch the
`mimetypes` registry to hard-code fixes for `.js` and any other
known-bad extensions (should be possible with `mimetypes.add_type`, no
monkey patching required), or (b) find all places that call `guess_type`
and extract a common helper function that handles the known-bad cases
itself and delegates to `mimetypes` for the rest. Or something to that
effect.

wchargin:
Summary:
The standard library’s `mimetypes.guess_type` can give incorrect results
on Windows, because arbitrary programs can influence its behavior by
writing to the registry. This can be worked around by reinstating the
correct MIME types with `mimetypes.add_type`, as in this patch. Fixes
#3120.

Test Plan:
To check that this patch works at all on Windows, edited the registry to
add a bad content type:

```
reg add HKCR\.js /v "Content Type" /d application/x-testing-123
```

…then verified that before this change, `mimetypes.guess_type("foo.js")`
gives the wrong value and TensorBoard does not render properly (strict
MIME type checking error in Chrome, as expected), whereas after this
change `mimetypes.guess_type("foo.js")` gives the correct value and
TensorBoard returns the correct content type in its JavaScript
responses. TensorBoard still does not render on Windows due to unrelated
errors in the debugger plugin.

To check that this set of patches is sufficient, `git grep guess_type`
shows call sites only in `core_plugin` and `projector_plugin`. The
former takes file names from `webfiles.zip`, which has only `html`,
`js`, and `woff2` files (plus a `LICENSE`). The latter has a fixed set
of inputs with extensions `html` and `js`.

wchargin-branch: windows-explicit-mime

### DIFF
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -382,7 +382,7 @@ class TensorBoard(object):
 
         old_signal_handler = signal.signal(signal_number, handler)
 
-    def _fix_mime_types(self, target=mimetypes):
+    def _fix_mime_types(self):
         """Fix incorrect entries in the `mimetypes` registry.
 
         On Windows, the Python standard library's `mimetypes` reads in
@@ -395,11 +395,6 @@ class TensorBoard(object):
         This method hard-codes the correct mappings for certain MIME
         types that are known to be either used by TensorBoard or
         problematic in general.
-
-        Args:
-          target: The registry to which to add types, either the
-            `mimetypes` module (default) or a `mimetypes.MimeTypes`
-            value.
         """
         # Known to be problematic when Visual Studio is installed:
         # <https://github.com/tensorflow/tensorboard/issues/3120>
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -37,6 +37,7 @@ from collections import defaultdict
 import errno
 import inspect
 import logging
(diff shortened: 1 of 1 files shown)


wchargin:
Can confirm that after patching in #3129, TensorBoard actually works on
Windows even with a bad MIME type in the registry (as described in the
test plan).

rohangpatil:
This is happening in mac as well. This is with Tensorboard 2.2.0
Screen Shot 2020-04-13 at 12 53 19 PM

aviasd:
Happening on AWS Sagemaker as well..
Tensorboard 2.2.1

PR #3129, opened 2020-01-10
plugins_listing: handle failures in plugin `is_active`

wchargin:
Summary:
Prior to this change, any error thrown by a plugin’s `is_active`
implementation would cause the entire response to 500, which renders
TensorBoard completely inoperable. Now, such failures mark the offending
plugin as inactive and continue with the rest of the response.

Test Plan:
Unit tests added. With this patch installed onto a Windows machine,
TensorBoard now works instead of crashing trying to load the debugger
plugin.

wchargin-branch: robust-is-active

### DIFF
--- a/tensorboard/backend/application_test.py
+++ b/tensorboard/backend/application_test.py
@@ -297,6 +297,20 @@ class ApplicationTest(tb_test.TestCase):
             },
         )
 
+    def testPluginsListingRobustToIsActiveFailures(self):
+        real_is_active = FakePlugin.is_active
+
+        def fake_is_active(self):
+            if self.plugin_name == "foo":
+                raise RuntimeError("this plugin is actually radioactive")
+            else:
+                return real_is_active(self)
+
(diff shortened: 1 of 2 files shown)


tensorboard/backend/application_test.py
```diff
@@ -297,6 +297,20 @@ def testPluginsListing(self):
             },
         )
 
+    def testPluginsListingRobustToIsActiveFailures(self):
+        real_is_active = FakePlugin.is_active
+
+        def fake_is_active(self):
+            if self.plugin_name == "foo":
+                raise RuntimeError("this plugin is actually radioactive")
```
psybuzz: XD

wchargin:
> It does make me wonder though, whether we should treat basically any
> method we call on authored plugins as possibly throwing an error.

Yes, that principle makes sense. In practice, `is_active` is going to be
more likely to throw than `frontend_metadata` or `get_plugin_apps`,
which are expected to just return ~constants.

PR #3130, opened 2020-01-10
data: fix `is_active`s in generic data mode

wchargin:
Summary:
Now that the TensorBoard backend core marks plugins as enabled if the
data provider has corresponding data (#3124), plugins can mark
themselves inactive when they don’t have data, rather than being enabled
unconditionally in generic data mode.

Test Plan:
Run TensorBoard against the scalars, images, and histograms demos
(individually), and note that it behaves the same way with generic data
mode enabled or disabled. (Note that the histogram demo has no graphs.)

wchargin-branch: data-isactive-false-by-default


wchargin:
Googlers, note that this change is safe for TensorBoard.dev because it
implements the new data provider method as of <http://cl/289179596>.
I’ve manually verified this with a test sync.

PR #3132, opened 2020-01-11
data: add data class protos and compat layer

wchargin:
Summary:
Summary metadata protos may now declare their generic data class, one of
scalar, tensor, or blob sequence. This affords ingestion of summaries on
disk into storage layers that implement the `DataProvider` API. A new
compatibility layer transparently adds this metadata to existing
first-party summaries (currently only the dashboards with data provider
support: scalars, histograms, and graphs). The multiplexer-backed data
provider now respects these data classes, removing all special-casing
code for graphs.

Test Plan:
Unit tests included. As an end-to-end test, all existing dashboards
continue to work with generic data mode either enabled or disabled.

wchargin-branch: dataclass-compat


tensorboard/compat/proto/summary.proto
```diff
@@ -53,8 +53,32 @@ message SummaryMetadata {
 
   // Longform readable description of the summary sequence. Markdown supported.
   string summary_description = 3;
+
+  // Class of data stored in this time series. Required for compatibility with
```
wchargin: Note: This file is meant to be kept in sync with TensorFlow’s copy, so
it’s not great to be introducing manual changes here. Once this file is
reviewed and approved (but not necessarily merged), I’ll make a
corresponding change to TensorFlow and get that submitted. My
understanding is that we can be fairly comfortable making this change
given that TensorFlow no longer exposes the protos from its Python API
and to avoid guaranteeing any API compatibility. Please correct me if
anything here sounds off.

wchargin:
This PR is a bit larger than I usually aim for (half of it is tests, but
still). I didn’t see an obvious way to split it up, but if it’s too much
to comfortably review, let me know and I’ll see what I can do.

tensorboard/backend/event_processing/data_provider.py
```diff
@@ -204,99 +219,101 @@ def list_blob_sequences(
...
-            )
-            time_series.append(datum)
+        result = {}
+        for (run, tags_for_run) in six.iteritems(index):
+            result_for_run = {}
+            result[run] = result_for_run
+            for (tag, metadata) in six.iteritems(tags_for_run):
+                events = self._multiplexer.Tensors(run, tag)
+                data = []
+                for event in events:
```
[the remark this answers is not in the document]
wchargin: That works out quite nicely, actually. Done.

tensorboard/backend/event_processing/data_provider.py
```diff
@@ -204,99 +219,101 @@ def list_blob_sequences(
...
-            raise errors.NotFoundError()
-
-        # TODO(davidsoergel): consider internal structure of non-graphdef blobs.
-        # In particular, note we ignore the requested index, since it's always 0.
-        return serialized_graph
+        summary_metadata = self._multiplexer.SummaryMetadata(run, tag)
+        if summary_metadata.data_class != summary_pb2.DATA_CLASS_BLOB_SEQUENCE:
+            raise errors.NotFoundError(blob_key)
+        tensor_events = self._multiplexer.Tensors(run, tag)
+        matching_steps = [e for e in tensor_events if e.step == step]
```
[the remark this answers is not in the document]
wchargin: Sure, take-first is fine with me. Done, and deduped above.

tensorboard/dataclass_compat.py
```diff
@@ -0,0 +1,96 @@
...
+    transformer = {
+        histograms_metadata.PLUGIN_NAME: _migrate_histogram_value,
+        scalars_metadata.PLUGIN_NAME: _migrate_scalar_value,
+    }.get(value.metadata.plugin_data.plugin_name, lambda v: (v,))
+    return transformer(value)
+
+
+def _migrate_scalar_value(value):
+    new_value = summary_pb2.Summary.Value()
+    new_value.CopyFrom(value)
```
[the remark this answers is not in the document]
wchargin: Okay, I’ve made `dataclass_compat` mutate when possible (which is
currently always). The code path where we need to copy can be tested by
applying this patch:

```diff
diff --git a/tensorboard/dataclass_compat.py b/tensorboard/dataclass_compat.py
index eb57af0a..51d2063d 100644
--- a/tensorboard/dataclass_compat.py
+++ b/tensorboard/dataclass_compat.py
@@ -95,4 +95,6 @@ def _migrate_value(value):
 def _migrate_scalar_value(value):
-    value.metadata.data_class = summary_pb2.DATA_CLASS_SCALAR
-    return (value,)
+    new_value = summary_pb2.Summary.Value()
+    new_value.CopyFrom(value)
+    new_value.metadata.data_class = summary_pb2.DATA_CLASS_SCALAR
+    return (new_value,)
```

…and checking that all tests still pass. (They do.)

For what it’s worth, I looked into making `data_compat` mutate, too, and
concluded that it would be safe to do so for all the current call sites,
but that the savings were less obvious because `data_compat` creates
structurally different events rather than just changing a single field.
So I’ve held off on that for now.

tensorboard/backend/event_processing/data_provider.py
```diff
@@ -265,29 +265,14 @@ def read_blob_sequences(
...
-                            wall_time=event.wall_time,
-                            step=event.step,
-                            values=values,
-                        )
+                    if event.step in data_by_step:
+                        continue
+                    data_by_step[event.step] = _convert_blob_sequence_event(
+                        experiment_id, plugin_name, run, tag, event
                     )
+                data = [datum for (step, datum) in sorted(data_by_step.items())]
```
nfelt: Just to confirm, this is new now that the results will be sorted by step?  I think that should be fine (and it's consistent with other DataProvider impls) but it is different from before, right?

wchargin: The spec requires this (because it needs to be done at some level and
can usually be done efficiently at the database layer); it’s true that
the earlier draft didn’t ensure that, but that was an oversight.

tensorboard/compat/proto/summary.proto
wchargin: Changes merged upstream into TensorFlow, so this is now ready:
<https://github.com/tensorflow/tensorflow/commit/ac6248d02be521dddddc1be1e2cd70e75d7e0c4b>

Issue #3134, opened 2020-01-11
Directory traversal vulnerability in example raw scalars plugin


wchargin:
The “example raw scalars” sample plugin serves static content from an
arbitrary user-specified path, and is vulnerable to basic [directory
traversal][dt] attacks.

If the plugin is installed to `/tmp/workdir/example_raw_scalars` and the
server is at `localhost:6006`, then one can read `/tmp/test.txt` from:

```
http://localhost:6006/data/plugin/example_raw_scalars/static/%2e%2e/%2e%2e/%2e%2e/%2e%2e/%2e%2e/%2e%2e/tmp/test.txt
```

(The `%2e` sequences are needed to prevent some layer of the stack,
either WSGI or Werkzeug, from normalizing the path, but this is a super
common variation—the first one listed on the OWASP page.)

This plugin’s README explicitly says to copy the entire directory as a
starting point and then just change the frontend to create your desired
visualization, so security vulnerabilities in the backend are especially
important.

(Found while auditing MIME type inference for #3128.)

[dt]: https://www.owasp.org/index.php/Path_Traversal

PR #3137, opened 2020-01-12
Update example plugin to check static asset paths

psybuzz:
#3134 

Followed a recommendation in [1] to prevent path traversal by checking whether the requested absolute path starts with the static/ dir's absolute path (+ separator) as prefix.

[1] https://security.openstack.org/guidelines/dg_using-file-paths.html#correct

Installed the example plugin and did before/after check that URLs returned 404
```
http://localhost:6006/data/plugin/example_raw_scalars/static/..%2f..%2f..%2f..%2ftest.txt
http://localhost:6006/data/plugin/example_raw_scalars/static/%2e%2e%2ftest.txt
```
Also added a unit test.


tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py
```diff
@@ -35,6 +35,7 @@
 from tensorboard.plugins import base_plugin
 from tensorboard.util import tensor_util
 from tensorboard.plugins.scalar import metadata
+from util import can_serve_from_static
```
wchargin: This doesn’t actually work for me: `util` is not the fully qualified
name of a module reachable from `sys.path`, and this is specified as an
absolute import, not a relative import. When creating a new Python 3.7
virtualenv, installing `tf-nightly` into it, running `setup.py develop`,
and then launching TensorBoard, I see:

```
  File ".../tensorboard/tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py", line 38, in <module>
    from util import can_serve_from_static
ModuleNotFoundError: No module named 'util'
```

The reason that the corresponding import works in the test is that when
the test is run as a main module, its working directory is inserted into
`sys.path`.

Can you elaborate on how you installed and ran the example plugin in
your testing? I’m not sure what environment this could have worked in;
this module has `from __future__ import absolute_import` (as it should),
so the semantics should be the same even in Python 2.

In any case, [relative imports are forbidden by Google style][1],
precisely so that we don’t have to think about these sorts of things. So
this import, if it exists at all, needs to be:

```python
from tensorboard_plugin_example_raw_scalars import util
```

(note furthermore that we only import modules, not their exports).

[1]: https://google.github.io/styleguide/pyguide.html#224-decision

tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/util.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def can_serve_from_static(path):
+    """Returns whether a filepath can be served from the static folder.
+
+    Args:
+        path: A string file path.
+    """
+    abs_path = os.path.abspath(path)
+    container_dir_prefix = os.path.abspath("./static") + os.sep
```
wchargin: As written, doesn’t this still let an attacker glean information about
the host directory setup? For example, the following test fails:

```python
# Should be false, but is true in the Bazel test runner!
self.assertFalse(
    can_serve_from_static(
        "static/../../org_tensorflow_tensorboard/static/index.js"
    )
)                                                                     
```

…because in the Bazel test environment the working directory is
`org_tensorflow_tensorboard`. You can use further `../EXPECTED_DIR`
components to determine the full path. This shouldn’t be possible.

Logically, we should be able to determine whether a path is safe without
making any syscalls, and doing so will likely be more clearly correct.
(Does your existing code do the right thing on Windows? You’ve used both
`os.path.join` and hard-coded forward slashes, plus HTTP paths always
use `/` as a path component separator, regardless of platform.)

Instead of using `os.path.abspath`, consider using `posixpath.normpath`,
which is Python’s equivalent of `require("path").posix.normalize` as
opposed to `require("path").resolve`.

tensorboard/examples/plugins/example_raw_scalars/BUILD
```diff
@@ -29,3 +29,12 @@ sh_test(
         "manual",  # https://github.com/tensorflow/tensorboard/issues/2987
     ],
 )
+
+py_test(
+    name = "unit_test",
+    size = "small",
+    srcs = ["test.py"],
```
wchargin: Need to declare a dependency on `util.py` so that changes to `util.py`
actually cause this test to be re-run.

tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/util.py
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Utilities for the plugin backend."""
+
+import os
+
+
+def can_serve_from_static(path):
```
wchargin: Acknowledged that this is in its own module for testing purposes, but
this has two notable drawbacks: (a) it’s harder to follow the plugin
code than if this were inlined or at least a module-private function;
and (b) the test-time imports are hard to reason about, as [discussed
earlier][rc]. It’s not clear to me what the best solution is; do you see
a way to better address these two concerns?

[rc]: https://github.com/tensorflow/tensorboard/pull/3137#discussion_r365884435

tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/util.py
psybuzz: Thanks for the suggestion, it's now using `posixpath.normpath`!

> Does your existing code do the right thing on Windows?

I still need to check this on a Windows machine.  My thinking is that we could use `os.path.join(os.path.dirname(__file__), os.path.normpath(...))`, which should always join the result of 2 OS-specific URLs.

psybuzz:
I tried out the raw scalars example yesterday with and without this patch on Windows 10 running Python 3.6.8, and both cases properly serve assets, including nested assets (/static/another/main.css) to the browser.

Before this PR, the path given to `open()` had mixed slashes, and contents were still properly read.  I haven't found anything in Python docs (from a brief search) saying that this is officially supported behavior, but the latest patch using normpath before the join should never produce mixed slashes.

tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py
```diff
@@ -84,10 +85,18 @@ def _serve_static_file(self, request):
...
         """
         static_path_part = request.path[len(_PLUGIN_DIRECTORY_PATH_PART) :]
-        res_path = os.path.join(os.path.dirname(__file__), static_path_part)
+        posix_norm_path = posixpath.normpath(static_path_part)
+        if not posix_norm_path.startswith("static/"):
+            return http_util.Respond(
+                request, "Not found", "text/plain", code=404
+            )
 
+        norm_path = os.path.normpath(static_path_part)
```
wchargin: If I read this correctly, the vulnerability still exists on Windows with
a path like `static/%2e%2e%5c%2e%2e%5ctest.txt`. You’re checking for
safety of the POSIX-parsed path but then using the NT path resolver to
actually determine the path, so the thing that you’ve tested for safety
isn’t the thing that’s actually used:

```
>>> url = urllib.parse.unquote("static/%2e%2e%5c%2e%2e%5ctest.txt")
>>> print(posixpath.normpath(url))  # note: starts with "static/"
static/..\..\test.txt
>>> print(ntpath.join(r"C:\path\to\dir", ntpath.normpath(url)))
C:\path\to\dir\..\test.txt
```

tensorboard/examples/plugins/example_raw_scalars/test.py
```diff
@@ -0,0 +1,59 @@
...
+import unittest
+
+from werkzeug import test
+from werkzeug import wrappers
+
+from tensorboard.plugins import base_plugin
+from tensorboard_plugin_example_raw_scalars import plugin
+
+example_plugin = plugin.ExampleRawScalarsPlugin(base_plugin.TBContext())
+serve_static_file = example_plugin._serve_static_file
```
wchargin: Avoid executing code at module import time. Failures at module import
time are basically impossible to recover from. This should run inside
the relevant test method, either explicitly (inlined or as a function
call) or implicitly (via `setUp`).

tensorboard/examples/plugins/example_raw_scalars/test.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def is_path_safe(path):
+    """Returns the result depending on the plugin's static file handler."""
+    client = test.Client(serve_static_file, wrappers.BaseResponse)
+    response = client.get(plugin._PLUGIN_DIRECTORY_PATH_PART + path)
+    return response.status_code == 200
+
+
+class URLSafetyTest(unittest.TestCase):
```
wchargin: nit: `UrlSafetyTest` ([camel case][1])

tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py
psybuzz: Yes, it was broken!  I realize now that normpath("a\\b\\c") doesn't convert to a posixpath "a/b/c".

I think what we need is to:
convert to posix path (no mixed slashes now) -->
posixpath.join with the current directory (closer to the path that `open` will call) -->
posixpath.normpath (any dot-dot-slash's should be at the beginning) -->
check that it starts with the right prefix.

That way, the result of the check can be the same thing we pass to `open`.

One thing I'm wondering is if `replace(ntpath.sep, posixpath.sep)` is the best way to ensure a path is a posixpath.

tensorboard/examples/plugins/example_raw_scalars/test.py
psybuzz: Done.

tensorboard/examples/plugins/example_raw_scalars/test.py
psybuzz: Done.

tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py
```diff
@@ -84,12 +86,23 @@ def _serve_static_file(self, request):
...
-        res_path = os.path.join(os.path.dirname(__file__), static_path_part)
 
-        with open(res_path, "rb") as read_file:
-            mimetype = mimetypes.guess_type(res_path)[0]
+        posix_path = static_path_part.replace(ntpath.sep, posixpath.sep)
+        required_prefix = posixpath.join(posixpath.dirname(__file__), "static/")
+        full_path = posixpath.join(posixpath.dirname(__file__), posix_path)
+        norm_path = posixpath.normpath(full_path)
+
+        if not norm_path.startswith(required_prefix):
```
wchargin: Okay, I think that this is safe, but I don’t see how it reads the
correct file. On Windows, `__file__` is like `C:\Users\someone\...`, but
the `posixpath.dirname` of that value is the empty string, because it
contains no POSIX path separators. So `norm_path` ends up being just the
NT-to-POSIX-converted input path without any directory prefix at all,
which behaves correctly under the safety check but won’t actually be
read from the filesystem correctly (because it will be a relative path,
and the current working directory will not be `dirname(__file__)`).

Do you think that it would make more sense to use `os.path.normpath`,
given that we’re trying to sanitize access to the platform-specific file
system? Something like:

```python
resource_name = os.path.normpath(static_path_part)
if not resource_name.startswith("static" + os.path.sep):
  return ...
resource_path = os.path.join(os.path.dirname(__file__), resource_name)
with open(resource_path, "rb") as infile:
  contents = infile.read()
...
```

I had hoped that `posixpath.normpath` could make this behavior more
portable, but I guess given a URL with encoded backslashes there isn’t
really a clearly portable way to handle it (that I can see—?).

psybuzz: You're correct, it didn't work on Windows, but it was safe.

I haven't come up with a better solution, so I've gone with your suggestion.  It looks safe, and the only strange thing is that behavior may differ between platforms if the user want to serve a file with a name that includes backslashes.  I've added a test case for this, and that behavior seems acceptable to me.

Once again, I'll update this thread when I've tested that it works on a Windows machine.

psybuzz: Can confirm this plugin works on Windows 10, Python 3.6.8

tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py
```diff
@@ -24,6 +24,8 @@
 import json
 import mimetypes
 import os
+import ntpath
+import posixpath
```
wchargin: Unused; remove?

tensorboard/examples/plugins/example_raw_scalars/test.py
```diff
@@ -0,0 +1,88 @@
...
+    response = client.get(plugin._PLUGIN_DIRECTORY_PATH_PART + path)
+    return response.status_code == 200
+
+
+class UrlSafetyTest(unittest.TestCase):
+    def test_path_traversal(self):
+        """Properly check whether a URL can be served from the static folder."""
+        with mock.patch(
+            "builtins.open", mock.mock_open(read_data="data")
+        ) as mock_file:
```
wchargin: You never use `mock_file`, so you don’t need the `as` clause.

tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py
psybuzz: Done

tensorboard/examples/plugins/example_raw_scalars/test.py
psybuzz: Done

PR #3141, opened 2020-01-13
Update summary_v2.py

boronhub:
Saw lack of documentation

Added usage example for tf.summary.audio()

* Screenshots of UI changes
None

Run local tests

None

### DIFF
--- a/tensorboard/plugins/audio/summary_v2.py
+++ b/tensorboard/plugins/audio/summary_v2.py
@@ -42,7 +42,13 @@ def audio(
     description=None,
 ):
     """Write an audio summary.
-
+    
+    Usage Example:
+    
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/audio/summary_v2.py
```diff
@@ -43,6 +43,12 @@ def audio(
 ):
     """Write an audio summary.
 
+    Usage Example:
+
+    >>> audio = tf.reshape(tf.linspace(0.0, 50.0, 4 * 8 * 2), (4, 8, 2))
+    >>> tf.summary.audio('k488', audio, 2)
+    <tf.Tensor: shape=(), dtype=bool, numpy=...>
```
WilliamHYZhang: Could we go a bit deeper into whatever `numpy` is supposed to be instead of going straight to the `...`?

PR #3142, opened 2020-01-14
Support display tensorboard in JupyterHub

qzchenwl:

To display tensorboard in JupyterHub.

When in JupyterHub, path_prefix should be `$JUPYTERHUB_SERVICE_PREFIX/proxy/absolute/$PORT/` so that `jupyter-server-proxy` can redirect requests for us.

* Screenshots of UI changes

### DIFF
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -23,6 +23,7 @@ from __future__ import print_function
 import datetime
 import errno
 import json
+import os
 import random
 import shlex
 import sys
@@ -47,6 +48,7 @@ from tensorboard import manager
 # details).
 _CONTEXT_COLAB = "_CONTEXT_COLAB"
 _CONTEXT_IPYTHON = "_CONTEXT_IPYTHON"
+_CONTEXT_JUPYTERHUB = "_CONTEXT_JUPYTERHUB"
 _CONTEXT_NONE = "_CONTEXT_NONE"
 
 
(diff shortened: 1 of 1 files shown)


qzchenwl:
@googlebot I signed it!

kopwei:
This PR looks great. I've tested it with 2.4 branch and it works as expected. It is exactly what we need for displaying Tensorboard in Jupyterhub with K8s

wchargin (commented):
Hi @qzchenwl! Thanks for the PR. A few concerns:

  - Adding special casing in our code for JupyterHub is a large
    maintenance burden for us. Most obviously, it means that we need to
    know what it is an how to test changes on it. Our build, test, and
    deploy processes don’t have any baggage related to Kubernetes or
    Docker or The Cloud, and changing that would be expensive.

  - The existing system passes args from `%tensorboard ARGS...` directly
    down to the TensorBoard binary. In particular, we do not attempt to
    parse, inspect, or modify the arguments, and this is by design.
    Correctly parsing flags would require discovering and executing
    arbitrary plugin code in the host IPython environment, would
    introduce an interdependency between the host `%tensorboard` and the
    target binary, etc.

Are there other ways that you could achieve your end goal that fare
better on these two constraints?

jhgoebbert:
This is great PR. At the Juelich Supercomputing Centre, we also give our users access to the HPC clusters through a JupyterHub/K8 cloud solution that enables supercomputing in the browser.  
From our point of view, it would obviously be an essential complement to Tensorboard.

DennisZhao168:
May I know when this feature will be released? As we are using JupyterHub and want to integrate the TensorBoard in the notebook which is spawned by JupyterHub.

ekungurov:
Hi all, 
Any updates on this issue?

PR #3143, opened 2020-01-14
Disable tests based on tf.contrib if tf.contrib is not available

manivaradarajan:

This is in preparation for the impending removal of tf.contrib from TensorFlow.
See b/147155091.

Support for db mode is dependent on tf.contrib, so db tests are disabled in this PR.
This has been done in two different ways to avoid changing the code in an unduly large manner.

In scalars_plugin_test, an annotation was added to disable the db tests.
In core_plugin_test, code dependent on the db was flagged off depending on whether tf.contrib was found to exist. 

Ran tests against two separate virtual environments with different versions of tensorflow 
(tf-nightly on 2020-01-14 and tensorflow 1.15) to ensure all tests still pass

Ensured that in the tensorflow 1.15 environment tf.contrib is present.


tensorboard/plugins/core/core_plugin_test.py
```diff
@@ -162,6 +170,8 @@ def testDataPaths_disableAllCaching(self):
             self.assertEqual(200, response.status_code, msg=path)
             self.assertEqual("0", response.headers.get("Expires"), msg=path)
 
+    @unittest.skipIf(not ENABLE_DB_TESTS,
```
wchargin: nit: Prefer `@unittest.skipUnless(ENABLE_DB_TESTS, …)` for clarity.

tensorboard/plugins/core/core_plugin_test.py
```diff
@@ -432,7 +461,20 @@ def _generate_test_data(self, run_name, experiment_name):
             else:
                 writer.add_graph(graph=None, graph_def=graph_def)
 
-        # Write data for the run to the database.
+        if ENABLE_DB_TESTS:
+            self._generate_test_data_db(run_name, experiment_name)
+
+        with tf.compat.v1.Session() as sess:
+            sess.run(tf.compat.v1.global_variables_initializer())
```
wchargin: Move this inside the `if HAS_TF_CONTRIB` block? It doesn’t do anything
otherwise, which is a bit confusing.

Once you’ve done that, you can just move this whole block in:

```python
        if ENABLE_DB_TESTS:
            self._generate_test_data_db(run_name, experiment_name)
            with tf.compat.v1.Session() as sess:
                sess.run(tf.compat.v1.global_variables_initializer())
                sess.run(tf.contrib.summary.summary_writer_initializer_op())
                sess.run(tf.contrib.summary.all_summary_ops())
```

…and then we don’t need `HAS_TF_CONTRIB` at all (can inline it into the
definition of `ENABLE_DB_TESTS`).

tensorboard/plugins/core/core_plugin_test.py
manivaradarajan: Done, thanks for the suggestion.

tensorboard/plugins/core/core_plugin_test.py
manivaradarajan: Done.

PR #3146, opened 2020-01-17
Rewrite core_plugin_test.py to use DB import mode instead of tf.contrib DB writer

nfelt:
This removes the `tf.contrib.summary.create_db_writer()` dependency from core_plugin_test.py, which partially addresses #1718 and #1705, and reverts the applicable part of #3143 (a temporary fix).

I've preserved most of the test coverage of the underlying DB-reading code by changing the test data population strategy to instead just use the DbImportMultiplexer logic that already exists within TensorBoard.  That logic doesn't set `started_time` so we lose some coverage in that the test that specifically tests that behavior is now logdir-only, because it didn't seem worth the trouble to get it to keep working.

I also refactored the rest of the tests a little bit to remove cruft and reorganize the test cases a little more logically.  Ultimately a lot of this should be cleaned up further as part of #2573.  (And ultimately we should either drop the DB reading logic entirely or encapsulate it inside a DataProvider, but that's a battle for another day.)

### DIFF
--- a/tensorboard/plugins/core/core_plugin_test.py
+++ b/tensorboard/plugins/core/core_plugin_test.py
@@ -350,7 +350,6 @@ class CorePluginLogdirModeTest(CorePluginTestBase, tf.test.TestCase):
 class CorePluginDbImportModeTest(CorePluginTestBase, tf.test.TestCase):
     def create_multiplexer(self):
         db_path = os.path.join(self.get_temp_dir(), "db.db")
-        print(self.id(), db_path)  # DO NOT SUBMIT
         db_uri = "sqlite:%s" % db_path
         db_connection_provider = application.create_sqlite_connection_provider(
             db_uri
--- a/tensorboard/plugins/core/core_plugin_test.py
+++ b/tensorboard/plugins/core/core_plugin_test.py
@@ -131,6 +131,64 @@ class CorePluginFlagsTest(tf.test.TestCase):
         self.assertIn(repr("noslash"), msg)
 
 
+class CorePluginNoDataTest(tf.test.TestCase):
+
+    def setUp(self):
+        super(CorePluginNoDataTest, self).setUp()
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/core/core_plugin_test.py
```diff
@@ -166,111 +174,112 @@ def testIndex_returnsActualHtml(self):
...
+        parsed_object = self._get_json(self.server, "/data/environment")
+        self.assertEqual(parsed_object["data_location"], self.get_temp_dir())
+
+    def testLogdir(self):
+        """Test the format of the data/logdir endpoint."""
+        parsed_object = self._get_json(self.server, "/data/logdir")
+        self.assertEqual(parsed_object, {"logdir": self.get_temp_dir()})
+
+
+class CorePluginDbModeTest(tf.test.TestCase):
```
stephanwlee: Before merge, would you kindly remind me how CorePluginDbMode differs from CorePluginDbImportMode?

nfelt: It's basically the branch here:

DB import mode: https://github.com/tensorflow/tensorboard/blob/94faac5eee1aeefdba1fdd86dcd0344dca72bf2d/tensorboard/backend/application.py#L131
DB read-only mode:
https://github.com/tensorflow/tensorboard/blob/94faac5eee1aeefdba1fdd86dcd0344dca72bf2d/tensorboard/backend/application.py#L148

Both modes exercise the SQL-based read path within the plugins, which is what we actually want to test.  The difference is that DB read-only mode expects you to pass an existing SQLite DB file with the `--db` flag (and the old test was using the `create_db_writer()` code to create such a file), whereas the DB import mode takes a logdir and automatically imports the logdir event files into a temporary DB (without using `create_db_writer()` but instead implementing mostly equivalent behavior).

PR #3150, opened 2020-01-18
[DebuggerV2] Some minor refactoring and cleanup in backend

caisq:
- In debugger_v2_plugin.py: add helper method _error_response()
  to shorten code.
- As suggested by wchargin@, avoid catching ValueError.

### DIFF
--- a/tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
+++ b/tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
@@ -26,13 +26,14 @@ from tensorboard.plugins.debugger_v2 import debug_data_provider
 from tensorboard.backend import http_util
 
 
-def _missing_run_error_response(request):
+def _error_response(request, error_message):
     return http_util.Respond(
-        request,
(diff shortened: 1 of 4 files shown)


tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
```diff
@@ -142,10 +141,8 @@ def serve_execution_data(self, request):
                 ),
                 "application/json",
             )
-        except (IndexError, ValueError) as e:
-            return http_util.Respond(
-                request, {"error": str(e)}, "application/json", code=400,
-            )
+        except IndexError as e:
```
wchargin: This change appears to still be implicitly propagating an `IndexError`
out of `read_blob`. It’s fine to catch the index errors resulting from
subscripting `blob_sequences[run][tag][0]`, but it’s not going to work
to expect `read_blob` to throw `IndexError` or `ValueError` or anything
other than `tensorboard.errors.PublicError`.

caisq:
Thanks for the review, @wchargin

PR #3157, opened 2020-01-22
plugins: add experimental URL reading API for iframed plugins

psybuzz:
Clients may wish to construct a link to TensorBoard that automatically opens a specific dashboard with a specific view.  This could be achieved with `window.parent.location`, but is discouraged since it is fragile (assumes 1 layer of nested iframe and modifying the URL has wide consequences).

This provides iframed plugins a safer approach to reading the URL through the plugin library.

Adds `getURLPluginData()` to the plugin library.

### DIFF
--- a/tensorboard/components/experimental/plugin_lib/BUILD
+++ b/tensorboard/components/experimental/plugin_lib/BUILD
@@ -1,4 +1,4 @@
-load("//tensorboard/defs:defs.bzl", "tf_js_binary", "tf_ts_library")
+load("//tensorboard/defs:defs.bzl", "tf_ts_library")
 
 package(default_visibility = ["//tensorboard:internal"])
 
--- a/tensorboard/components/experimental/plugin_lib/BUILD
+++ b/tensorboard/components/experimental/plugin_lib/BUILD
@@ -1,4 +1,4 @@
-load("//tensorboard/defs:defs.bzl", "tf_ts_library", "tf_js_binary")
+load("//tensorboard/defs:defs.bzl", "tf_js_binary", "tf_ts_library")
 
 package(default_visibility = ["//tensorboard:internal"])
 
@@ -9,10 +9,10 @@ licenses(["notice"])  # Apache 2.0
 tf_ts_library(
     name = "plugin_lib",
     srcs = [
+        "core.ts",
         "index.ts",
         "message.ts",
         "plugin-guest.ts",
-        "core.ts",
         "runs.ts",
     ],
 )
--- a/tensorboard/components/experimental/plugin_lib/BUILD
+++ b/tensorboard/components/experimental/plugin_lib/BUILD
@@ -16,12 +16,3 @@ tf_ts_library(
         "runs.ts",
     ],
 )
-
-tf_js_binary(
-    name = "plugin_lib_rollup",
-    compile = 1,
-    entry_point = "index.ts",
-    deps = [
-        ":plugin_lib",
-    ],
-)
--- a/tensorboard/components/experimental/plugin_lib/BUILD
+++ b/tensorboard/components/experimental/plugin_lib/BUILD
@@ -1,4 +1,4 @@
-load("//tensorboard/defs:defs.bzl", "tf_ts_library")
+load("//tensorboard/defs:defs.bzl", "tf_ts_library", "tf_js_binary")
 
 package(default_visibility = ["//tensorboard:internal"])
 
@@ -12,6 +12,16 @@ tf_ts_library(
         "index.ts",
         "message.ts",
         "plugin-guest.ts",
+        "core.ts",
         "runs.ts",
+    ],
+)
+
+tf_js_binary(
+    name = "plugin_lib_rollup",
+    compile = 1,
+    entry_point = "index.ts",
+    deps = [
+        ":plugin_lib",
     ],
 )
--- a/tensorboard/components/experimental/plugin_lib/test/test.ts
+++ b/tensorboard/components/experimental/plugin_lib/test/test.ts
@@ -119,6 +119,7 @@ describe('plugin lib integration', () => {
           'sample_plugin',
           'tagFilter=loss',
           'p.sample_plugin.foo=bar',
+          'p.sample_plugin.foo=bar_from_duplicate',
           'smoothing=0.5',
           'p.sample_plugin.foo2=bar2',
           'p.sample_plugin2.foo=bar',
@@ -128,7 +129,7 @@ describe('plugin lib integration', () => {
 
         const data = await this.lib.core.getURLPluginData();
         expect(data).to.deep.equal({
-          foo: 'bar',
+          foo: 'bar_from_duplicate',
           foo2: 'bar2',
         });
       });
--- a/tensorboard/components/experimental/plugin_lib/test/test.ts
+++ b/tensorboard/components/experimental/plugin_lib/test/test.ts
@@ -99,11 +99,12 @@ describe('plugin lib integration', () => {
        * host's URL has been updated.
        */
       it('returns URL data', async function() {
-        tf_globals.setFakeHash(
-          'sample_plugin' +
-            '&p.sample_plugin.foo=bar' +
-            '&p.sample_plugin.foo2=bar2'
-        );
+        const hash = [
+          'sample_plugin',
+          'p.sample_plugin.foo=bar',
+          'p.sample_plugin.foo2=bar2',
+        ].join('&');
+        tf_globals.setFakeHash(hash);
         window.dispatchEvent(new Event('hashchange'));
 
         const data = await this.lib.core.getURLPluginData();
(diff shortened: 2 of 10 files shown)


tensorboard/components/experimental/plugin_util/plugin-host-ipc.ts
```diff
@@ -13,11 +13,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 namespace tb_plugin.host {
+  export type PluginHostMessageContext = {
+    pluginName: string | null;
```
stephanwlee: leave a comment on when this can be a `null`.

tensorboard/components/experimental/plugin_util/plugin-host-ipc.ts
```diff
@@ -42,10 +51,33 @@ namespace tb_plugin.host {
...
+    ipc: lib.DO_NOT_USE_INTERNAL.IPC
+  ): lib.DO_NOT_USE_INTERNAL.MessageCallback {
+    return (payload) => {
+      const context: PluginHostMessageContext = {pluginName: null};
+
+      // 'name' is a keyword set in the iframe URL by tf-tensorboard's
+      // `_renderPluginIframe`.
+      const frame = ipcToFrame.get(ipc);
+      if (frame) {
+        const pluginName = new URL(frame.src).searchParams.get('name');
```
stephanwlee: I believe this is modifiable by the plugin. I think there should be some global map of instance of an iframe to plugin metadata. https://github.com/tensorflow/tensorboard/blob/893775b8e0d4236ad6ee2082018aae672fcdbe48/tensorboard/components/tf_tensorboard/tf-tensorboard.html#L864-L871

We can of course do this later as it would be a larger surgery.

tensorboard/components/tf_storage/storage.ts
```diff
@@ -211,6 +218,50 @@ namespace tf_storage {
     return {get, set, getInitializer, getObserver, disposeBinding};
   }
 
+  export function initializeUrlDict() {
```
stephanwlee: `transformLegacyStorage`, `convertLegacyStorage`, or `migrateLegacyScheme`?

tensorboard/components/experimental/plugin_lib/test/test.ts
```diff
@@ -89,4 +89,16 @@ describe('plugin lib integration', () => {
       });
     });
   });
+
+  describe('lib.core', () => {
+    describe('#getURLPluginData', () => {
+      it('returns URL data', async function() {
+        tf_globals.setFakeHash('sample_plugin&p.sample_plugin.foo=bar');
```
stephanwlee: How does this test work? Is this setting the hash on the parent frame and are we testing both lib and the host?

Also, please add more complex cases :) Ideas:
- weird values
- p.wrong_plugin.foo
- p.sample_plugin.foo.bar

tensorboard/components/experimental/plugin_util/core-host-impl.ts
```diff
@@ -0,0 +1,29 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+/**
+ * Implements core plugin APIs.
+ */
+tb_plugin.host.listen('experimental.GetURLPluginData', (context) => {
+  const prefix = `p.${context.pluginName}.`;
```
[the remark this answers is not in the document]
psybuzz: Good point, I didn't consider that.  @wchargin wdyt about disallowing periods?
I've not found any plugins that use periods, but technically it would be a breaking change.
Alternatively, we could use some character like `~` or `*`, and just communicate that in the docs.

https://github.com/tensorflow/tensorboard/pull/3160

tensorboard/components/experimental/plugin_util/plugin-host-ipc.ts
psybuzz: Done

tensorboard/components/experimental/plugin_util/plugin-host-ipc.ts
psybuzz: Done in this PR

tensorboard/components/tf_storage/storage.ts
psybuzz: Done as `migrateLegacyURLScheme`

tensorboard/components/experimental/plugin_util/core-host-impl.ts
```diff
@@ -0,0 +1,32 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+/**
+ * Implements core plugin APIs.
+ */
+tb_plugin.host.listen('experimental.GetURLPluginData', (context) => {
+  if (!context) {
```
stephanwlee: I am really forgetful. Would it be possible to leave a comment on when this can happen?

tensorboard/components/experimental/plugin_lib/test/test.ts
```diff
@@ -89,4 +91,67 @@ describe('plugin lib integration', () => {
...
+        const data = await this.lib.core.getURLPluginData();
+        expect(data).to.deep.equal({
+          foo: 'bar',
+          foo2: 'bar2',
+        });
+      });
+
+      it('handles non alphanumeric data', async function() {
+        tf_globals.setFakeHash(
+          'sample_plugin' +
```
stephanwlee: idea:

```ts
[
  'sample_plugin',
  'p.sample_plugin.foo=bar%20baz',
  ...
].join('&')
```
may read better.

tensorboard/components/experimental/plugin_lib/test/test.ts
```diff
@@ -89,4 +91,67 @@ describe('plugin lib integration', () => {
...
     });
   });
+
+  describe('lib.core', () => {
+    describe('#getURLPluginData', () => {
+      /**
+       * These tests use tf_globals' fake hash to make tf_storage think that the
+       * host's URL has been updated.
+       */
+      it('returns URL data', async function() {
```
stephanwlee: let's write one more spec. 
```ts
tf_globals.setFakeHash('sample_plugin&p.sample_plugin&p.sample_plugin.');
```

(user handcrafted the hash but forgot to add something after the period)

tensorboard/components/experimental/plugin_util/core-host-impl.ts
stephanwlee: Never mind. The comment on `context` already covers this.

tensorboard/components/experimental/plugin_lib/test/test.ts
psybuzz: Done.

tensorboard/components/experimental/plugin_lib/test/test.ts
psybuzz: Sounds good, done.

tensorboard/components/experimental/plugin_lib/test/test.ts
```diff
@@ -89,4 +91,84 @@ describe('plugin lib integration', () => {
...
+        const hash = [
+          'sample_plugin',
+          'p.sample_plugin.foo=bar',
```
stephanwlee: Consider adding a duplicate entry, `'p.sample_plugin.foo=qaz',`

tensorboard/components/experimental/plugin_lib/BUILD
```diff
@@ -1,4 +1,4 @@
-load("//tensorboard/defs:defs.bzl", "tf_ts_library")
+load("//tensorboard/defs:defs.bzl", "tf_js_binary", "tf_ts_library")
```
stephanwlee: `tf_js_binary` is extraneous.

tensorboard/components/experimental/plugin_lib/test/test.ts
psybuzz: Done.

tensorboard/components/experimental/plugin_lib/BUILD
psybuzz: Done.

tensorboard/components/experimental/plugin_util/core-host-impl.ts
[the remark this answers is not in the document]
wchargin: Disallowing periods seems reasonable to me. It prevents putting (some
forms of) version numbers in plugin names, but that seems fine. We could
use `~`, but I think that that would be confusing (it also frequently
breaks links in PDFs written by TeX novices). We can’t use `*` without
encoding it, because it is not an unreserved character per [RFC 3986,
section 2.3][rfc]. So given that we want to permit `[A-Za-z0-9_-]`,
reserving `.` seems like the best option.

[rfc]: https://tools.ietf.org/html/rfc3986#section-2.3

PR #3159, opened 2020-01-22
Enable strict Typescript option

psybuzz:
Stricter TS options allow TensorBoard to capture more errors.

Enables TS's 'strict' compiler option.

Ran Bazel build on tensorboard with+without the `-c dbg` option and observed no errors after fixing plugin_lib.


psybuzz:
cc @mprobst

mprobst:
Thanks for the quick turnaround!

PR #3160, opened 2020-01-22
Disallow period from plugin names

psybuzz:

A proposed change to the URL scheme for plugins would involve describing a plugin-specific URL value like so:
`https://<tensorboard>/#my_plugin&p.my_plugin.key=value`

In this scheme, the prefix `p.` is special, and a period is used to separate a plugin name from a key name.  For the scheme to work, periods need to be disallowed from plugin names to prevent ambiguity.  E.g. is `&p.my_plugin.foo.key` referring to a plugin `my_plugin` with a key `foo.key`, OR a plugin `my_plugin.foo` with a key `key`?

Plugin names are already restricted, so alternatively, we could use some other character like `~` or `*` to distinguish a plugin name from key name in the URL.  e.g.
`https://<tensorboard>/#my_plugin&p.my_plugin~key=value`
`https://<tensorboard>/#my_plugin&p.my_plugin*key=value`

### DIFF
--- a/ADDING_A_PLUGIN.md
+++ b/ADDING_A_PLUGIN.md
@@ -54,7 +54,7 @@ You can start building the backend by subclassing `TBPlugin` in [`base_plugin.py
 
 ```python
 class MyPlugin(base_plugin.TBPlugin):
-  plugin_name = "My_Awesome_Plugin"
+  plugin_name = "my_awesome_plugin"
 
   def __init__(self, context): # ...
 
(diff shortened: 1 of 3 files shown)


ADDING_A_PLUGIN.md
```diff
@@ -54,7 +54,7 @@ You can start building the backend by subclassing `TBPlugin` in [`base_plugin.py
 
 ```python
 class MyPlugin(base_plugin.TBPlugin):
-  plugin_name = "My Awesome Plugin"
+  plugin_name = "My_Awesome_Plugin"
```
wchargin: OK: Plugin names are also conventionally lowercase (as they’re internal
keys that appear in URLs, and their capitalization is not user-facing),
but it’s fine with me if we change that in a separate PR.

psybuzz: Done in this PR.  Thanks for the review!

PR #3161, opened 2020-01-23
data: rename `latest_max_index` to `max_length`

wchargin:
Summary:
The metadata for a blob sequence has historically included the number of
blob values in the most recent blob sequence datum (`latest_max_index`).
The intent of this field was to enable the audio and image dashboards to
determine the total number of “samples” per summary. But the value that
is actually needed here is the _maximum_ length of any datum, not the
length of just the latest datum.

This commit renames `latest_max_index` to `max_length`, keeping aliases
for both the initializer and the accessor property while we migrate
clients. All clients in this repository are migrated in this commit, but
some Google-internal clients need to be handled separately.

Test Plan:
Unit tests show that both the old and new invocations still work. A test
sync shows that relevant Google-internal targets are still fine.

wchargin-branch: data-max-length

### DIFF
--- a/tensorboard/backend/event_processing/data_provider.py
+++ b/tensorboard/backend/event_processing/data_provider.py
@@ -235,17 +235,19 @@ class MultiplexerDataProvider(provider.DataProvider):
                 result[run] = result_for_run
                 max_step = None
                 max_wall_time = None
-                latest_max_index = -1
+                max_length = None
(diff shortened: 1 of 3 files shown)


wchargin:
Google-internal change (to be submitted after this is merged):
<http://cl/291069814>

PR #3162, opened 2020-01-23
data: remove `latest_max_index`

wchargin:
Summary:
This was replaced with `max_length`, which has the semantics that we
actually want.

Test Plan:
Running `git grep latest_max_index` gives no results.

wchargin-branch: data-remove-latest-max-index


wchargin:
Google-internal compatibility CL is <http://cl/291473156>.

PR #3165, opened 2020-01-23
Break plugin controls away from header

stephanwlee:
In order to make the header re-composable, we want to make the
components inside header more modular and reusable. This change breaks
the list of plugins into its own component.

It visually results in the same UI.


tensorboard/webapp/header/header_module.ts
```diff
@@ -20,14 +20,20 @@ import {MatTabsModule} from '@angular/material/tabs';
...
+
 @NgModule({
-  declarations: [HeaderComponent, HeaderContainer],
-  exports: [HeaderComponent, HeaderContainer],
+  declarations: [
+    HeaderComponent,
+    PluginControlComponent,
+    PluginControlContainer,
+  ],
+  exports: [HeaderComponent],
```
psybuzz: If the purpose of this is to allow embedders to customize the header, and we expect embedders to usually include the plugin tabs, should we also export `PluginControlContainer`?

stephanwlee: Excellent point. Will change for sure.

PR #3172, opened 2020-01-24
Fix buildifier lint errors and improve linting on CI:

stephanwlee:
Internal CI became a bit too strict about the lint errors and started to hard failing.
This change is our attempt to (1) have less lint errors on BUILD files and (2) unblock
the sync.

To reviewer: I am not too confident about the bash script I ended up writing. Please do advise on better ways of failing on lint error (with proper error message).


stephanwlee:
Confirmed that the new lint script properly report the error:
```
 Lint BUILD files0s
##[error]Process completed with exit code 123.
Run git ls-files -z '*BUILD' | xargs -0 ./tensorboard/tools/build_lint.sh
tensorboard/plugins/core/BUILD:6: load: Loaded symbol "ng_module" is unused. Please remove it.
To disable the warning, add '@unused' in a comment. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#load)
##[error]Process completed with exit code 123.
```

tensorboard/tools/build_lint.sh
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+ret_val=0
+
+for build_path in "$@"
+do
+  # We do not care about few lint warns just yet.
+  buildifier --mode=check --lint warn "$build_path" 2>&1 | grep -v -E "(load-on-top:|native-py:|native-java:|constant-glob:)" 1>&2
```
wchargin: By passing each file separately to `buildifier` and `grep`, you’re
spawning 190 times as many processes as you would otherwise, and forcing
`buildifier` to reinitialize each time. This makes the script take 3.6
seconds instead of 0.136 seconds on my machine.

From the code inside the loop, it looks we could probably use [the
`--warnings` argument to `buildifier`][warn] to process these all at
once. (We wouldn’t get the fancy return code, but I think that that’s
okay.) What do you think?

[warn]: https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md#linter

tensorboard/tools/build_lint.sh
```diff
@@ -0,0 +1,28 @@
...
+  if [ "$?" -eq 0 ]; then
+    ret_val=$((1 + $ret_val))
+  fi
+done
+
+exit $ret_val
```
wchargin: Having a “count of failures” exit code is fairly common and certainly
reasonable, but when you do this you need to cap it at some value
because return codes wrap around after 256, so having exactly 256 errors
would be misclassified as passing. (Often I see this capped at 100 or
99; it’s good form to not return higher than 127, as those error codes
are also used by the shell to indicate signals.)

tensorboard/tools/build_lint.sh
stephanwlee: The warnings flag is perfect and I really think it is useful. 

> We wouldn’t get the fancy return code,

There are problems with the code I wrote as you mentioned below and buildifier also does have pretty cool return code so I am happy to use vanilla buildifier :)

```
Return codes used by buildifier:

  0: success, everything went well
  1: syntax errors in input
  2: usage errors: invoked incorrectly
  3: unexpected runtime errors: file I/O problems or internal bugs
  4: check mode failed (reformat is needed)
```

tensorboard/tools/build_lint.sh
stephanwlee: Yeah, I did not think about that! Thanks for catching it. Unfortunately, it is moot now :)

PR #3176, opened 2020-01-27
[ng] expose tb_server_data_source as interface

stephanwlee:
We can provide an alternative implementation of the DataSource by using
Angular DI but the provided implementations require some common APIs so
they do not get out of sync. By implementing an interface, we can
guarantee some correctness at the compilation time.

### DIFF
--- a/tensorboard/webapp/webapp_data_source/tb_server_data_source.ts
+++ b/tensorboard/webapp/webapp_data_source/tb_server_data_source.ts
@@ -19,12 +19,19 @@ import {from} from 'rxjs';
 
 import {PluginsListing} from '../types/api';
 
+import {
+  TBServerDataSourceInterface,
+  TfBackendComponent,
+} from './tb_server_data_source_types';
+
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
 @Injectable()
-export class TBServerDataSource {
+export class TBServerDataSource implements TBServerDataSourceInterface {
   // TODO(soergel): implements WebappDataSource
-  private tfBackend = (document.createElement('tf-backend') as any).tf_backend;
+  private tfBackend = (document.createElement(
+    'tf-backend'
+  ) as TfBackendComponent).tf_backend;
 
   constructor(private http: HttpClient) {}
 
--- a/tensorboard/webapp/webapp_data_source/BUILD
+++ b/tensorboard/webapp/webapp_data_source/BUILD
@@ -7,6 +7,7 @@ ng_module(
     srcs = [
         "tb_server_data_source.ts",
         "tb_server_data_source_module.ts",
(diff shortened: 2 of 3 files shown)


tensorboard/webapp/webapp_data_source/data_api_route_converter_test.ts
```diff
@@ -0,0 +1,45 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+import {TEST_ONLY, DataApiRouteConverter} from './data_api_route_converter';
+
+describe('data_api_route_converter test', () => {
+  describe('getUrl', () => {
+    // let hrefSpy = jasmine.Spy;
```
psybuzz: nit: drop

tensorboard/webapp/webapp_data_source/tb_server_data_source.ts
```diff
@@ -14,29 +14,37 @@ limitations under the License.
...
 
+import {DataApiRouteConverter} from './data_api_route_converter';
+import {DataSource} from './data_source';
+
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
 @Injectable()
-export class TBServerDataSource {
+export class TBServerDataSource implements DataSource {
   // TODO(soergel): implements WebappDataSource
```
psybuzz: Can we now drop this TODO, or explain what more we plan to address later?

tensorboard/webapp/webapp_data_source/data_api_route_converter.ts
```diff
@@ -0,0 +1,50 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {Injectable} from '@angular/core';
+
+export interface ApiRouteContext {
```
psybuzz: Do we have plans for `ApiRouteContext` in OSS?  If not, is there some reason that TB variants must rely on this being in OSS?

tensorboard/webapp/webapp_data_source/tb_server_data_source.ts
stephanwlee: Honestly... idk what it means and I rather not touch the part of the code that is not related. Feel free to file an issue :)

stephanwlee:
You are correct. I decided to provide an alternative implementation of the tb_server_data_source instead. Thanks for the idea!

PR #3178, opened 2020-01-27
Logo to home page link

ericdnielsen:
We wanted the site logo/brand to link to the homepage in Tensorboard.dev.  It seemed like the simplest way was to unconditionally add this link here in OSS.  Its effectively a no-op here -- the URL fragment is preserved across the link.

Tested both in OSS and .dev deployments.

Only touched the one file, I see there's also webapp/header/header_component.ng.html, but couldn't figure out how that's used, or even if its used.  If its a WIP that's not enabled yet, let me know how to test and I'll update that too.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -36,11 +36,11 @@ py_repositories()
 
 http_archive(
     name = "io_bazel_rules_closure",
-    sha256 = "9b99615f73aa574a2947226c6034a6f7c319e1e42905abc4dc30ddbbb16f4a31",
-    strip_prefix = "rules_closure-4a79cc6e6eea5e272fe615db7c98beb8cf8e7eb5",
+    sha256 = "6a900831c1eb8dbfc9d6879b5820fd614d4ea1db180eb5ff8aedcb75ee747c1f",
+    strip_prefix = "rules_closure-db4683a2a1836ac8e265804ca5fa31852395185b",
     urls = [
-        "http://mirror.tensorflow.org/github.com/bazelbuild/rules_closure/archive/4a79cc6e6eea5e272fe615db7c98beb8cf8e7eb5.tar.gz",
-        "https://github.com/bazelbuild/rules_closure/archive/4a79cc6e6eea5e272fe615db7c98beb8cf8e7eb5.tar.gz",  # 2019-09-16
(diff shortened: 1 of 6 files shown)


WORKSPACE
```diff
@@ -36,11 +36,11 @@ py_repositories()
 
 http_archive(
     name = "io_bazel_rules_closure",
-    sha256 = "9b99615f73aa574a2947226c6034a6f7c319e1e42905abc4dc30ddbbb16f4a31",
-    strip_prefix = "rules_closure-4a79cc6e6eea5e272fe615db7c98beb8cf8e7eb5",
+    sha256 = "6a900831c1eb8dbfc9d6879b5820fd614d4ea1db180eb5ff8aedcb75ee747c1f",
```
stephanwlee: Are these changes required? If not, please drop them.

tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -61,7 +61,7 @@ <h2>Settings</h2>
     <paper-header-panel>
       <paper-toolbar id="toolbar" slot="header" class="header">
         <div id="toolbar-content" slot="top">
-          <div class="toolbar-title">[[brand]]</div>
+          <div class="toolbar-title"><a href="/">[[brand]]</a></div>
```
stephanwlee: This will not work AFAIK. Can you please test this with path_prefix flag?

WORKSPACE
ericdnielsen: Sorry, I hate Git.  It only showed the one file in the PR, else I would have removed this before.

tensorboard/components/tf_tensorboard/tf-tensorboard.html
ericdnielsen: Confirmed bad behavior with path_prefix. 

Looks like next simplest option is to add a parameter to the tf-tensorboard component, and pass in the link target.  Only render the link if the link target is non-null/empty. -- I don't see an option that is likely to do the right thing in the presence of path_prefix across the mutliple deployment types.

ericdnielsen:
Planning to abandon this PR and come back with another one that approach it differently.

stephanwlee:
I believe this approach is abandoned and we now have `homePath` on `tf-tensorboard` instead. Closing.

PR #3181, opened 2020-01-28
test_util: don’t warn when raw byte strings passed

wchargin:
Summary:
Some arguments may be passed as TensorFlow protos, TensorBoard protos,
or raw byte strings. We want to warn against the TensorFlow protos, but
were also warning against the raw byte strings. This change fixes all
existing warnings, so we bump the log level to `error` to discourage
regressions.

Test Plan:
After changing the warnings in `test_util.py` to `AssertionError`s, all
tests still pass where 18 of them used to fail.

wchargin-branch: testutil-permit-bytes-protos

### DIFF
--- a/tensorboard/plugins/custom_scalar/custom_scalars_plugin_test.py
+++ b/tensorboard/plugins/custom_scalar/custom_scalars_plugin_test.py
@@ -120,10 +120,15 @@ class CustomScalarsPluginTest(tf.test.TestCase):
         with test_util.FileWriterCache.get(
             os.path.join(self.logdir, "foo")
         ) as writer:
-            writer.add_summary(summary.pb(self.foo_layout))
+            writer.add_summary(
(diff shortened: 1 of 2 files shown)


wchargin:
> This change fixes all existing warnings, so we bump the log level to
> `error` to discourage regressions.

We can also make these hard errors if you think that that’d be more
appropriate.

wchargin:
> LGTM, but just wondering, is it better to use six.string_type?

Presuming that you mean `six.string_types`? (Sorry, don’t mean to “well,
actually” you, but this stuff is confusing enough as it is.) No, I don’t
think so: that would make it so that `str` is allowed but not `bytes` in
Python 3, when in reality it’s valid to pass a byte string to these
functions but not a text string.

stephanwlee:
Got it. Thanks for the clarification!

PR #3182, opened 2020-01-28
Dynamically created plugins from a component registry for extensibility

psybuzz:
(I accidentally pressed ctrl+Enter too soon, this is meant to be a draft for discussion)

In webapp/ (ng tensorboard), Angular plugins must be manually added to `plugin_component.ng.html`, a configuration template that determines which ng Component is rendered when the user selects a certain plugin via the tabs/dropdown.

Notably,
- Configuring TB requires knowledge of an extra file (`plugin_component.ng.html`)
- TB variants typically supplant their own custom `app_module.ts`.  To get a different list of plugins, variants need to fork logic in the PluginModule or PluginComponent.  This breaks the abstraction, forcing other modules to care about which `activePlugin` is selected, for the sole purpose of showing non-builtin plugin dashboards.

This PR proposes an alternate world where plugins are dynamically created.  Instead of PluginModule depending on a list of plugins in `plugin_component.ng.html`, plugins are declared in `app_module.ts`, and PluginModule uses Angular's resolveComponentFactory and createComponent to instantiate them.

Advantages
- App configuration is more centralized in one place (`app_module`).  Updating a single file is simpler.
- TB variants always depend on TB to inject their plugins.  Having 1 way to inject plugins is simpler to maintain than 2, as there is nothing to keep in sync.
- [Not needed yet] TB can call functions on plugins without assuming a fixed list.  E.g. we could add a call to `plugin.reload()` over all registered plugins, if needed.

Disadvantages
- Compilers cannot know ahead of time which Components will be rendered.
- Safety is harder to guarantee, unless all possible Components that can be created are analyzed.

**EDIT**

This PR is upgraded from Draft status.  With this change, TB variants can register their own plugin-component bindings without updating the hardcoded list in `plugin_component.ng.html`.  NgModules can register via `PluginRegistryModule.forPlugin(...)`.

### DIFF
--- a/tensorboard/webapp/plugins/plugin_registry_module.ts
+++ b/tensorboard/webapp/plugins/plugin_registry_module.ts
@@ -1,0 +1,76 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {
+  NgModule,
+  Component,
+  Type,
+  ModuleWithProviders,
+  Inject,
+  Optional,
+} from '@angular/core';
+import {PluginConfig, PLUGIN_CONFIG_TOKEN} from './plugin_registry_types';
+
+const pluginNameToComponent = new Map<string, Type<Component>>();
+
+@NgModule({})
+export class PluginRegistryModule {
+  constructor(
+    @Optional() @Inject(PLUGIN_CONFIG_TOKEN) configs: PluginConfig[]
+  ) {
+    if (!configs) {
+      return;
+    }
+    const configSet = new Set(configs);
+    console.assert(
+      configSet.size === configs.length,
+      'Cannot register the same plugin multiple times.'
+    );
+
+    for (const config of configs) {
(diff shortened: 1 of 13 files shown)


tensorboard/webapp/plugins/plugins_component.ts
```diff
@@ -79,7 +90,6 @@ export class PluginsComponent implements OnChanges {
 
     const pluginElement = this.createPlugin(plugin);
     if (pluginElement) {
-      pluginElement.id = plugin.id;
```
psybuzz: Now that components are created via `createComponent`, Angular started to complain due to this line, because it didn't expect the id to be modified.  These `id`s were only used for testing purposes, so I updated the tests.

tensorboard/webapp/plugins/plugins_component.ts
```diff
@@ -44,9 +47,17 @@ import {
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class PluginsComponent implements OnChanges {
+  constructor(
+    private componentFactoryResolver: ComponentFactoryResolver,
```
stephanwlee: nit: `readonly` here and below.

tensorboard/webapp/plugins/plugin_registry_module.ts
```diff
@@ -0,0 +1,78 @@
...
+  constructor(
+    @Optional() @Inject(PLUGIN_CONFIG_TOKEN) configs: PluginConfig[]
+  ) {
+    if (!configs) {
+      return;
+    }
+    const configSet = new Set(configs);
+    console.assert(
+      configSet.size === configs.length,
+      'Cannot register the same plugin multiple times.'
```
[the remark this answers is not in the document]
psybuzz: Ahh good catch, I've updated it to be a set of names, where string references suffice for comparison.

tensorboard/webapp/plugins/plugins_component.ts
psybuzz: Done.

PR #3183, opened 2020-01-28
images: add data provider support

wchargin:
Summary:
This patch teaches the images plugin about the generic data APIs and
newfound blob sequence support, guarded behind the `--generic_data=true`
feature flag.

Test Plan:
Setting `--generic_data` to either `false` or `true` has no observable
effect, even if TensorBoard is patched to never pass a multiplexer to
plugins’ `TBContext` (to prevent accidentally reads along the old path).
Unit tests expanded to cover both the old and new paths.

wchargin-branch: generic-images

### DIFF
--- a/tensorboard/plugins/image/images_plugin.py
+++ b/tensorboard/plugins/image/images_plugin.py
@@ -25,8 +25,10 @@ import six
 from six.moves import urllib
 from werkzeug import wrappers
 
+from tensorboard import errors
 from tensorboard import plugin_util
 from tensorboard.backend import http_util
+from tensorboard.data import provider
 from tensorboard.plugins import base_plugin
 from tensorboard.plugins.image import metadata
 from tensorboard.compat import tf
@@ -67,6 +69,10 @@ class ImagesPlugin(base_plugin.TBPlugin):
         """
         self._multiplexer = context.multiplexer
         self._db_connection_provider = context.db_connection_provider
+        if context.flags and context.flags.generic_data == "true":
+            self._data_provider = context.data_provider
+        else:
+            self._data_provider = None
 
     def get_plugin_apps(self):
         return {
@@ -78,6 +84,9 @@ class ImagesPlugin(base_plugin.TBPlugin):
     def is_active(self):
         """The images plugin is active iff any run has at least one relevant
         tag."""
+        if self._data_provider:
+            return False  # `list_plugins` as called by TB core suffices
+
         if self._db_connection_provider:
             # The plugin is active if one relevant tag can be found in the database.
             db = self._db_connection_provider()
@@ -100,7 +109,24 @@ class ImagesPlugin(base_plugin.TBPlugin):
     def frontend_metadata(self):
         return base_plugin.FrontendMetadata(element_name="tf-image-dashboard")
 
-    def _index_impl(self):
+    def _index_impl(self, experiment):
+        if self._data_provider:
+            mapping = self._data_provider.list_blob_sequences(
+                experiment_id=experiment, plugin_name=metadata.PLUGIN_NAME,
+            )
+            result = {run: {} for run in mapping}
+            for (run, tag_to_content) in six.iteritems(mapping):
+                for (tag, metadatum) in six.iteritems(tag_to_content):
+                    description = plugin_util.markdown_to_safe_html(
+                        metadatum.description
+                    )
+                    result[run][tag] = {
+                        "displayName": metadatum.display_name,
+                        "description": description,
+                        "samples": metadatum.max_length - 2,
+                    }
+            return result
+
         if self._db_connection_provider:
             db = self._db_connection_provider()
             cursor = db.execute(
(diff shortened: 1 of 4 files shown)


tensorboard/plugins/image/images_plugin.py
```diff
@@ -353,12 +412,16 @@ def _get_individual_image(self, run, tag, index, sample):
...
         """Serves an individual image."""
-        run = request.args.get("run")
-        tag = request.args.get("tag")
-        index = int(request.args.get("index", "0"))
-        sample = int(request.args.get("sample", "0"))
         try:
-            data = self._get_individual_image(run, tag, index, sample)
+            if self._data_provider:
+                blob_key = request.args["blob_key"]
+                data = self._data_provider.read_blob(blob_key)
```
psybuzz: If `read_blob` can raise NotFoundError, should we add that to the `except`?

tensorboard/plugins/image/images_plugin.py
```diff
@@ -353,12 +412,16 @@ def _get_individual_image(self, run, tag, index, sample):
...
+            else:
+                run = request.args.get("run")
+                tag = request.args.get("tag")
+                index = int(request.args.get("index", "0"))
+                sample = int(request.args.get("sample", "0"))
+                data = self._get_individual_image(run, tag, index, sample)
```
psybuzz: `_get_individual_image()` internally uses either DB connection or Multiplexer, but do we also need to add a DataProvider path there?

tensorboard/plugins/image/images_plugin.py
```diff
@@ -100,7 +109,24 @@ def is_active(self):
...
+            result = {run: {} for run in mapping}
+            for (run, tag_to_content) in six.iteritems(mapping):
+                for (tag, metadatum) in six.iteritems(tag_to_content):
+                    description = plugin_util.markdown_to_safe_html(
+                        metadatum.description
+                    )
+                    result[run][tag] = {
+                        "displayName": metadatum.display_name,
+                        "description": description,
+                        "samples": metadatum.max_length - 2,
```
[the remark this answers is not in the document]
wchargin: Yes—thanks! Done.

tensorboard/plugins/image/images_plugin.py
wchargin: No need to; `errors.PublicError`s (including `NotFoundError`) can
propagate out of plugin code and are [handled by TensorBoard core][1],
so this will gracefully turn into a human-readable 404.

[1]: https://github.com/tensorflow/tensorboard/blob/f14006003b93d299d9713395ee4a4fef824b546e/tensorboard/backend/application.py#L181-L195

tensorboard/plugins/image/images_plugin.py
```diff
@@ -205,6 +234,33 @@ def _image_response_for_run(self, run, tag, sample):
...
+            SAMPLE_COUNT = 10
+            all_images = self._data_provider.read_blob_sequences(
+                experiment_id=experiment,
+                plugin_name=metadata.PLUGIN_NAME,
+                downsample=SAMPLE_COUNT,
+                run_tag_filter=provider.RunTagFilter(runs=[run], tags=[tag]),
+            )
+            images = all_images.get(run, {}).get(tag, None)
+            if images is None:
+                raise errors.NotFoundError(
```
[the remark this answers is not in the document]
wchargin: Sure—I don’t think that it’s strictly necessary because `NotFoundError`
propagates cleanly (see other reply), but it can’t hurt.

tensorboard/plugins/image/images_plugin.py
wchargin: Yeah, this API is slightly wonky because the query parameters are
different in the two paths: the multiplexer/DB paths take run, tag,
index, and sample, which are interpreted by this plugin; but the generic
data path takes an opaque blob key, which is interpreted only by the
data provider. Python doesn’t have a nice way to express union types, so
I didn’t see a clean way for `_get_individaul_image` to handle this. But
for clarity I’ve gone ahead and factored this out into a method. Does
this help?

tensorboard/plugins/image/images_plugin.py
psybuzz: Ah, I didn't know that!  That's convenient

tensorboard/plugins/image/images_plugin.py
psybuzz: Sorry, I didn't realize earlier that the data provider path doesn't use `_get_individual_image()`.  The rename to `_get_legacy_individual_image()` looks clearer to me.

For `_get_generic_data_individual_image`, feel free to inline it as before, if you prefer.

wchargin:
Thanks for the reviews!

> For `_get_generic_data_individual_image`, feel free to inline it as
> before, if you prefer.

No strong preference, but I think that the function ([while short][1])
is at the right level of abstraction, and is also helpful to point in
the assertion error message, so I’ll keep it around.

[1]: https://www.martinfowler.com/bliki/FunctionLength.html

PR #3185, opened 2020-01-28
Change EventFileLoader to use tf_record_iterator when possible

nfelt:
This change addresses the most critical part of https://github.com/tensorflow/tensorboard/issues/1711 by relaxing our dependency on PyRecordReader, which was never part of the public TF API and is slated to be removed imminently (cc @av8ramit).  Now when it's missing, we instead use `tf.compat.v1.io.tf_record_iterator` which was recently updated in https://github.com/tensorflow/tensorflow/commit/54f1de4fcf81ed7e25247d84f9dadc761b5a1bb7 to no longer abandon the read offset after reaching EOF (which resolves the issue described in #1711 that prevented us from using it, since we couldn't use it to poll the end of a file for newly appended data).

The implementation here adopts the iterator style throughout, and wraps our legacy PyRecordReader usage inside an iterator API so it can be interchanged.

As part of this, I rewrote event_file_loader_test.py to modernize the structure and improve coverage, including making it actually exercise the truncated record path (the previous approach did not actually trigger that branch in the code), and adding coverage for TimestampedEventFileLoader.

Test plan: confirmed that the new test passes:
1. with no code-under-test changes on TF 2.0, py2 and py3
2. with code-under-test changes on TF 2.0, py2 and py3
3. with no code-under-test changes on TF nightly py3
4. with code-under-test changes on TF nightly py3
5. with code-under-test changes on TF nightly py3, PyRecordReader fallback disabled

The new test does not pass in case 5 against TF 2.0 rather than nightly, but instead fails because the tests for the dynamic event file correctly report that the tf_record_iterator implementation does not implement the expected preservation of the read offset.

### DIFF
--- a/tensorboard/backend/event_processing/event_file_loader_test.py
+++ b/tensorboard/backend/event_processing/event_file_loader_test.py
@@ -22,7 +22,6 @@ from __future__ import print_function
 import abc
 import io
 import os
-import tempfile
 
 import six
 import tensorflow as tf
--- a/tensorboard/backend/event_processing/event_file_loader_test.py
+++ b/tensorboard/backend/event_processing/event_file_loader_test.py
@@ -19,94 +19,136 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import abc
+import io
 import os
 import tempfile
 
+import six
 import tensorflow as tf
 
 
 from tensorboard.backend.event_processing import event_file_loader
+from tensorboard.compat.proto import event_pb2
+from tensorboard.summary.writer import record_writer
 
 
-class EventFileLoaderTest(tf.test.TestCase):
-    # A record containing a simple event.
-    RECORD = (
-        b'\x18\x00\x00\x00\x00\x00\x00\x00\xa3\x7fK"\t\x00\x00\xc0%\xddu'
-        b"\xd5A\x1a\rbrain.Event:1\xec\xf32\x8d"
-    )
+FILENAME = "test.events"
 
-    def _WriteToFile(self, filename, data):
-        with open(filename, "ab") as f:
-            f.write(data)
 
-    def _LoaderForTestFile(self, filename):
-        return event_file_loader.EventFileLoader(
-            os.path.join(self.get_temp_dir(), filename)
-        )
+@six.add_metaclass(abc.ABCMeta)
+class EventFileLoaderTestBase(object):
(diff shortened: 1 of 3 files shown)


tensorboard/backend/event_processing/event_file_loader_test.py
```diff
@@ -19,94 +19,136 @@
...
+        loader = self._make_loader()
+        self.assertEventWallTimes(loader.Load(), [1.0, 2.0, 3.0])
+
+    def testLoad_dynamicEventFile(self):
+        self._append_record(_make_event(wall_time=1.0))
+        loader = self._make_loader()
+        self.assertEventWallTimes(loader.Load(), [1.0])
+        self._append_record(_make_event(wall_time=2.0))
+        self.assertEventWallTimes(loader.Load(), [2.0])
+        self.assertEmpty(list(loader.Load()))
```
wchargin: Huh, I’d always used `assertFalse` for this. [TIL!][absl]

[absl]: https://github.com/abseil/abseil-py/blob/06edd9c20592cec39178b94240b5e86f32e19768/absl/testing/absltest.py#L820-L821

tensorboard/backend/event_processing/event_file_loader_test.py
```diff
@@ -19,94 +19,136 @@
...
+        with open(filepath, "ab", buffering=0) as f:
+            # Record missing the last byte should result in data loss error
+            # that we log and swallow.
+            f.write(record[:-1])
+            self.assertEventWallTimes(loader.Load(), [1.0, 2.0])
+            # Retrying against the incomplete record has no effect.
+            self.assertEmpty(list(loader.Load()))
+            # Retrying after completing the record should return the new event.
+            f.write(record[-1:])
+            self.assertEventWallTimes(loader.Load(), [3.0])
```
wchargin: This is great; thanks!

tensorboard/backend/event_processing/event_file_loader_test.py
```diff
@@ -19,94 +19,136 @@
 from __future__ import division
 from __future__ import print_function
 
+import abc
+import io
 import os
 import tempfile
```
wchargin: `tempfile` unneeded now?

nfelt: Done, thanks.

PR #3187, opened 2020-01-28
[DebuggerV2] Display detailed execution data and stack traces

caisq:
  * Continue developing the frontend of Debugger V2 plugin

  * Pull data from the /execution/data route and render the detailed top-level execution data in
    the newly-added ExecutionDataComponent
  * Pull stack-trace data from the /stack_frames/stack_frames route and render the stack frames
    (host name, file path, lineno and function) name in the newly-added StackTraceComponent
    * Note that SourceCodeComponent is still to be implemented in future PRs.
  * Add effects and reducers to support the two aforementioned actions

* Screenshots of UI changes
  * CURT_HEALTH mode: 
image
  * CONCISE_HEALTH mode: image
  * SHAPE mode: 
image

  * Newly-added unit tests.
  * Manually run the ng_tensorboard frontend with the new Debugger V2 features against real tfdbg2-format dumps (see screenshots above)


caisq:
cc @baileydesign

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/execution_data/execution_data_container.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  readonly focusedExecutionIndex$ = this.store.pipe(
+    select(getFocusedExecutionIndex)
+  );
+
+  readonly focusedExecutionData$ = this.store.pipe(
+    select(getFocusedExecutionData)
+  );
+
+  readonly tensorDebugModeName$ = this.store.pipe(
```
stephanwlee: Wouldn't it be better to just move enum values around instead of enum name? If you are going to ever show enum as string in the UI, I would rather want template to have enum to displayName conversion (for i18n, too).

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
```diff
@@ -507,4 +512,65 @@ describe('Debugger reducers', () => {
     const nextState = reducers(state, actions.executionScrollRight());
     expect(nextState.executions.scrollBeginIndex).toEqual(1450);
   });
+
+  it(`Updates states on executionDigestFocused`, () => {
```
stephanwlee: Could you add a spec that tests `scrollBeginIndex` + `displayIndex` where `scrollBeginIndex > 0`?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -46,12 +46,15 @@ const initialState: DebuggerState = {
       lastLoadedTimeInMs: null,
     },
     scrollBeginIndex: 0,
+    focusIndex: null,
```
stephanwlee: FYI, `executions` is getting quite fat. It might benefit from creating a sub-reducer.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
```diff
@@ -507,4 +512,65 @@ describe('Debugger reducers', () => {
...
+    const state = createDebuggerState({
+      activeRunId: '__default_debugger_run__',
```
stephanwlee: Is this override used in any meaningful way? If not, do not override it. Instead, override what can impact the test like `executions.scrollBeginIndex`.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
```diff
@@ -507,4 +512,65 @@ describe('Debugger reducers', () => {
...
+
+      const nextState = reducers(
+        state,
+        actions.executionDataLoaded({
+          begin: 0,
+          end: numObjects,
+          executions: executionDataObjects,
+        })
+      );
+      expect(Object.keys(nextState.executions.executionData).length).toEqual(
```
stephanwlee: where possible, use `toBe` instead of `toEqual`.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
```diff
@@ -507,4 +512,65 @@ describe('Debugger reducers', () => {
...
+            input_tensor_ids: [i * 10],
+            output_tensor_ids: [i * 10 + 1],
+          })
+        );
+      }
+
+      const nextState = reducers(
+        state,
+        actions.executionDataLoaded({
+          begin: 0,
```
stephanwlee: add a spec that has `begin != 0`

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
```diff
@@ -507,4 +512,65 @@ describe('Debugger reducers', () => {
...
+    const state = createDebuggerState({
+      activeRunId: '__default_debugger_run__',
+    });
+    const stackFrames: StackFramesById = {
+      aaa: createTestStackFrame(),
+      bbb: createTestStackFrame(),
+      ccc: createTestStackFrame(),
+    };
+    const nextState = reducers(state, actions.stackFramesLoaded({stackFrames}));
+    expect(nextState.stackFrames).toEqual(stackFrames);
```
stephanwlee: Do not use the same reference as L568 as per b/148791050

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors.ts
```diff
@@ -111,3 +114,72 @@ export const getVisibleExecutionDigests = createSelector(
...
+  (state: DebuggerState): Execution | null => {
+    const {focusIndex, executionData} = state.executions;
+    if (focusIndex === null || executionData[focusIndex] === undefined) {
+      return null;
+    }
+    return executionData[focusIndex];
+  }
+);
+
+export const getFocusedExecutionStackFrames = createSelector(
```
stephanwlee: I find the behavior at L180 quite interesting. Can you leave a comment on the contract (i.e., if any stackframe data is missing, this will return `null` for the whole thing)? Also, what is an expected behavior in that case?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors.ts
```diff
@@ -13,14 +13,17 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-import {createSelector, createFeatureSelector} from '@ngrx/store';
+import {createSelector, createFeatureSelector, select} from '@ngrx/store';
```
stephanwlee: I do not see corresponding test changes. Especially since there are non-trivial logic in the selector, I will strongly recommend writing the tests.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -19,6 +19,18 @@ export {DataLoadState, LoadState} from '../../../../webapp/types/data';
 
 export const DEBUGGER_FEATURE_KEY = 'debugger';
 
+export enum TensorDebugMode {
+  UNSPECIFIED = 0,
```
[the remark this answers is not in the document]
stephanwlee: Lovely! Thanks!

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -83,8 +129,14 @@ export interface Executions {
   // Beginning index of the current scrolling position.
   scrollBeginIndex: number;
 
+  // Index of focusing.
```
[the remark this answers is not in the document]
stephanwlee: > Yes, null means nothing is focused.

Please transfer this to the comment :) Some people use `-1` to convey that and we want to be clear here.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/execution_data/execution_data_component.css
```diff
@@ -0,0 +1,63 @@
...
+  text-align: left;
+}
+
+.execution-data-field {
+  white-space: nowrap;
+}
+
+.execution-data-key {
+  display: inline-block;
+  max-width: 120px;
```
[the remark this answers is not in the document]
stephanwlee: I assume it is because descendent can take up width larger than 120px? If so, would you want to `overflow: hidden` them (and put ellipsis on the text)?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/stack_trace/stack_trace_container.ts
```diff
@@ -0,0 +1,63 @@
...
+})
+export class StackTraceContainer {
+  readonly stackFramesForDisplay$ = this.store.pipe(
+    select(
+      createSelector(
+        getFocusedExecutionStackFrames,
+        (stackFrames) => {
+          if (stackFrames === null) {
+            return null;
+          } else {
```
stephanwlee: style nit: no need for else block

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/stack_trace/stack_trace_component.css
```diff
@@ -0,0 +1,66 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+.stack-frame-array {
+  height: 340px;
+  max-height: 340px;
```
stephanwlee: :) remove

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/execution_data/execution_data_container.ts
caisq: Good suggestion. I do want the "debug mode" field in the UI to show the text name, which is much easier to understand than the numerical enum value. I made the change so that this property returns a numerical value and the to-string conversion is done in the html template.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
caisq: Good suggestion. Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors_test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+import {
+  getFocusedExecutionData,
+  getFocusedExecutionIndex,
+} from './debugger_selectors';
+import {DataLoadState, DEBUGGER_FEATURE_KEY} from './debugger_types';
+import {createDebuggerState, createTestExecutionData} from '../testing';
+
+describe('debugger selectors', () => {
+  describe('getFocusedExecutionIndex', () => {
```
stephanwlee: FYI: tests on `getFocusedExecutionStackFrames` are missing.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container_test.ts
```diff
@@ -288,4 +301,494 @@ describe('Debugger Container', () => {
...
+        )
+      );
+      fixture.detectChanges();
+
+      const opTypeElement = fixture.debugElement.query(By.css('.op-type'));
+      expect(opTypeElement.nativeElement.innerText).toEqual('Inverse');
+      const inputTensorsElement = fixture.debugElement.query(
+        By.css('.input-tensors')
+      );
+      expect(inputTensorsElement.nativeElement.innerText).toEqual('1');
```
stephanwlee: hmm, `input-tensors` renders `input_tensor_ids` but I see no state about that above. What am I missing? (is it hidden in `createDebuggerState`?)

BTW, instead of very integration style test for entire debugger_container, you may want to have one for stack_trace vs. timeline vs. others. Those components are complex by themselves, too.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/execution_data/execution_data_container.ts
```diff
@@ -0,0 +1,99 @@
...
+  readonly tensorDebugMode$ = this.store.pipe(
+    select(
+      createSelector(
+        getFocusedExecutionData,
+        (execution: Execution | null) => {
+          if (execution === null) {
+            return TensorDebugMode.UNSPECIFIED;
+          } else {
+            return execution.tensor_debug_mode;
+            // return TensorDebugMode[execution.tensor_debug_mode];
```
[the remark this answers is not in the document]
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -197,6 +214,27 @@ export class DebuggerEffects {
...
+      ),
+      filter(([props, runId, focusIndex]) => {
+        return props.numExecutions > 0 && runId !== null && focusIndex === null;
+      }),
+      map(() => {
+        // Automatically focus on the first execution event when any
+        // execution events exist.
+        // TODO(cais): This should auto-focus on the first alert even
+        // when any alerts exist.
+        return executionDigestFocused({displayIndex: 0});
```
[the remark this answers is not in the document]
caisq: I appreciate your review. However, in this particular case, I think the existing code is a good implementation AFAICS. I could use reducer to set `focusIndex` to 0 if a) `numExecutions` is > 0 and b) `focusIndex` is null. However, that doesn't automatically trigger the loading of the detailed execution data of the first execution event from the data source. This effect achieves that with reasonable amount of complexity IMO. The underlying reason that necessitates this is the fact that the digests and detailed data objects (for execution events) are served via two different HTTP routes. The downside is slightly complex flux cycles like this, but the separation of routes makes it possible to load only a limited amount of data that the frontend needs. Also the concern of complexity here is ameliorated by the fact that we have tests covering this effect and the related effects. 

LMK WYT.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors_test.ts
```diff
@@ -0,0 +1,234 @@
...
+            scrollBeginIndex: 0,
+            executionDigests: {},
+            executionData: {},
+          },
+        })
+      );
+      expect(getFocusedExecutionStackFrames(state)).toBe(null);
+    });
+
+    it('returns correct stack frames when there is no focus', () => {
```
stephanwlee: one more spec: can we add one that has partial stack frames?

caisq:
@stephanwlee I've pulled your PR against my branch into this PR. I like the modularity and clarity of the new pattern in debugger_effects.ts. Thanks for taking time to showing it to me. I've also updated the tests in debugger_effects_test.ts accordingly. As we discussed in the other PR, I'm using a testing pattern that asserts on the external behavior of the effects, primarily in terms of actions dispatched and the fetches made via the data source. PTAL.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors_test.ts
```diff
@@ -0,0 +1,269 @@
...
+          },
+        })
+      );
+      expect(getFocusedExecutionStackFrames(state)).toEqual([
+        stackFrame1,
+        stackFrame3,
+      ]);
+    });
+
+    it('returns correct stack frames when subset of frames is missing', () => {
```
stephanwlee: `returns correct stack frames when subset of frames is missing` -> `returns null when subset of frames is missing`

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
```diff
@@ -205,283 +230,433 @@ describe('Debugger effects', () => {
...
       };
-      const fetchExecutionDigests = spyOn(
-        TestBed.get(Tfdbg2HttpServerDataSource),
-        'fetchExecutionDigests'
-      )
-        .withArgs('__default_debugger_run__', 0, 2)
-        .and.returnValue(of(executionDigestForTest));
+    }
+
+    it('loads execution digests, data & stack trace loading if numExecutions>0', () => {
```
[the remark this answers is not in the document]
caisq: Ack. There is a balance between keeping the body of the `it` small and minimizing the amount of indirection. I chose to err on the side of the former in this case.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
```diff
@@ -205,283 +230,433 @@ describe('Debugger effects', () => {
...
-                    pageLoadedSizes: {0: 2},
-                  },
-                  pageSize: 2,
-                  displayCount: 2,
-                  scrollBeginIndex: 2,
-                  executionDigests: {},
-                },
-              })
-            )
+          const dataAlreadyExists = false;
```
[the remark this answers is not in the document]
caisq: Oops. This is my oversight. Sorry. Fixed. Thanks for catching it.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
```diff
@@ -205,283 +230,433 @@ describe('Debugger effects', () => {
...
+          store.overrideSelector(getExecutionPageSize, pageSize);
+          let pageLoadedSizes: {[pageIndex: number]: number} = {};
+          pageLoadedSizes = {
+            0: 20,
+            1: 20,
+            2: 20,
+          };
+          if (dataAlreadyExists) {
+            pageLoadedSizes[3] = 5;
+          }
```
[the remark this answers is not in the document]
caisq: 1. Done.
2. I prefer to keep the current style. The `for-of` part currently states clearly what condition is being tested (e.g., whether data already exists or not). If the `for-of` iterates over numbers instead of booleans, it'll be harder to follow what the tests are about. I agree the logic inside test code shouldn't be overly complex. But I don't think I've cross the line here yet.

PR #3193, opened 2020-01-29
base_plugin: make `is_ng_component` default idiomatic

wchargin:
Summary:
It’s customary\* to use `None` as the default value for an argument,
because otherwise clients cannot specify to “use the default value”.
This patch makes `is_ng_component` consistent with all sibling kwargs.

\* In some dialects of Python, including TensorFlow and TensorBoard.
See, e.g.:
<https://unhashable.com/always-use-none-for-default-args-in-python/>

Test Plan:
Unit tests suffice.

wchargin-branch: idiomatic-ng-default


wchargin:
Yep, no problem.

PR #3195, opened 2020-01-29
[DebuggerV2] Backend: Implement basic features of /alerts route

caisq:
### Technical description of changes
- The /alerts route is based on the recently-added InfNanMonitor
  in tfdbg2's tensorflow part.
- The route serves information about:
  - How many inf/nan events (InfNanAlert) exist
  - Number limit for each type of alerts (currently there is only
    one type: InfNanAlert)
  - The details of each alert, in the specified begin/end range.
- Modify the dummy TF2 program (`_generate_tfdbg_v2_data()`) in
  debugger_v2_plugin_test.py to intentionally generate `inf`s and `nan`s.

### Detailed steps to verify changes work correctly (as executed by you)
- Python unit tests added.

### DIFF
--- a/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
@@ -30,6 +30,12 @@ from tensorboard import errors
 # the same logdir, replace this magic string with actual run names.
 DEFAULT_DEBUGGER_RUN_NAME = "__default_debugger_run__"
 
+# Default number of alerts per monitor type.
+# Limiting the number of alerts is based on the consideration that usually
+# only the first few alerting events are the most critical and the subsequent
+# ones are either repetitions of the earlier ones or caused by the earlier ones.
+DEFAULT_PER_TYPE_ALERT_LIMIT = 1000
+
 
 def run_in_background(target):
     """Run a target task in the background.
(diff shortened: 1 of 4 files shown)


tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -49,6 +55,28 @@ def run_in_background(target):
     thread.start()
 
 
+def _alert_to_json(alert):
+    # TODO(cais): Replace this with Alert.to_json() when supported by the
+    # backend.
+    global debug_events_monitors
```
wchargin: Can we directly import `debug_events_monitors` here instead of relying
on it having been implicitly set by a previous call to `Runs()`?

caisq: Done. This will be replaced by a `to_json()` call on the Alert object anyway. So it's just temporary.

BTW, I sync'ed with @davidsoergel offline. He's okay with me merging the PR as is with only @wchargin's approval.

caisq: Thanks for the review, @wchargin!

wchargin: SGTM; thanks.

PR #3201, opened 2020-01-31
pip_package: only build Python 3 wheel

wchargin:
Summary:
We no longer support Python 2, and so should not build packages for it,
because they’re not guaranteed to work at all. We likewise should run
the smoke test in Python 3 by default.

Test Plan:
Running `bazel run //tensorboard/pip_package:extract_pip_package` still
succeeds, now extracting only `tensorboard-2.2.0a0-py3-none-any.whl`.

wchargin-branch: pip-py3-only


wchargin:
Ah; thanks!

PR #3204, opened 2020-01-31
Import `Callable` from `collections.abc`

### LINKED ISSUE #3202: Importing ABC directly from collections was removed in Python 3.9
# Issue description

Importing ABC directly from collections was removed in Python 3.9 use collections.abc instead. Below are instances

```
python
Python 3.8.0 (default, Nov  6 2019, 21:49:08) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
>>> import collections
>>> collections.Callable
<stdin>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
<class 'collections.abc.Callable'>
```

```
rg "collections.*(Awaitable|Coroutine|AsyncIterable|AsyncIterator|AsyncGenerator|Hashable|Iterable|Iterator|Generator|Reversible|Sized|Container|Callable|Collection|MutableSet|Mapping|MutableMapping|MappingView|KeysView|ItemsView|ValuesView|Sequence|MutableSequence|ByteString)\b" 
tensorboard/summary/summary_test.py
43:                getattr(self.module, plugin), collections.Callable
49:                getattr(self.module, "%s_pb" % plugin), collections.Callable

tensorboard/plugins/scalar/scalars_plugin_test.py
188:        self.assertIsInstance(routes["/scalars"], collections.Callable)
189:        self.assertIsInstance(routes["/tags"], collections.Callable)

tensorboard/plugins/histogram/histograms_plugin_test.py
159:        self.assertIsInstance(routes["/histograms"], collections.Callable)
160:        self.assertIsInstance(routes["/tags"], collections.Callable)

tensorboard/plugins/image/images_plugin_test.py
170:        self.assertIsInstance(self.routes["/images"], collections.Callable)
172:            self.routes["/individualImage"], collections.Callable
174:        self.assertIsInstance(self.routes["/tags"], collections.Callable)

tensorboard/plugins/mesh/mesh_plugin_test.py
173:        self.assertIsInstance(self.routes["/tags"], collections.Callable)
174:        self.assertIsInstance(self.routes["/meshes"], collections.Callable)
175:        self.assertIsInstance(self.routes["/data"], collections.Callable)

tensorboard/plugins/debugger/events_writer_manager_test.py
40:        self.assertIsInstance(apps["/health_pills"], collections.Callable)

tensorboard/plugins/pr_curve/pr_curves_plugin_test.py
124:        self.assertIsInstance(routes["/tags"], collections.Callable)
125:        self.assertIsInstance(routes["/pr_curves"], collections.Callable)
127:            routes["/available_time_entries"], collections.Callable

tensorboard/plugins/audio/audio_plugin_test.py
146:        self.assertIsInstance(routes["/audio"], collections.Callable)
147:        self.assertIsInstance(routes["/individualAudio"], collections.Callable)
148:        self.assertIsInstance(routes["/tags"], collections.Callable)

tensorboard/plugins/distribution/distributions_plugin_test.py
118:        self.assertIsInstance(routes["/distributions"], collections.Callable)
119:        self.assertIsInstance(routes["/tags"], collections.Callable)

tensorboard/plugins/graph/graphs_plugin_test.py
138:        self.assertIsInstance(routes["/graph"], collections.Callable)
139:        self.assertIsInstance(routes["/run_metadata"], collections.Callable)
140:        self.assertIsInstance(routes["/info"], collections.Callable)

tensorboard/plugins/debugger/debugger_plugin_test.py
40:        self.assertIsInstance(apps["/health_pills"], collections.Callable)

tensorboard/plugins/core/core_plugin_test.py
163:        self.assertIsInstance(routes["/data/logdir"], collections.Callable)
164:        self.assertIsInstance(routes["/data/runs"], collections.Callable)

tensorboard/plugins/text/text_plugin_test.py
55:        self.assertIsInstance(routes["/tags"], collections.Callable)
56:        self.assertIsInstance(routes["/text"], collections.Callable)

```

wchargin:
Summary:
In Python 3.9, `collections.Callable` is slated to be removed in favor
of `collections.abc.Callable` (and likewise with the other ABCs, but
this is the only one that we use).

Fixes #3202.

Test Plan:
Running `git grep -P 'collections\.(?!OrderedDict|Counter)[A-Z]'` no
longer finds any matches.

wchargin-branch: collections-abc-callable

### DIFF
--- a/tensorboard/plugins/audio/audio_plugin_test.py
+++ b/tensorboard/plugins/audio/audio_plugin_test.py
@@ -18,7 +18,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import collections
+import collections.abc
 import json
 import os
 import shutil
@@ -143,9 +143,11 @@ class AudioPluginTest(tf.test.TestCase):
     def testRoutesProvided(self):
         """Tests that the plugin offers the correct routes."""
         routes = self.plugin.get_plugin_apps()
-        self.assertIsInstance(routes["/audio"], collections.Callable)
-        self.assertIsInstance(routes["/individualAudio"], collections.Callable)
-        self.assertIsInstance(routes["/tags"], collections.Callable)
+        self.assertIsInstance(routes["/audio"], collections.abc.Callable)
+        self.assertIsInstance(
+            routes["/individualAudio"], collections.abc.Callable
+        )
+        self.assertIsInstance(routes["/tags"], collections.abc.Callable)
 
     def testOldStyleAudioRoute(self):
         """Tests that the /audio routes returns correct old-style data."""
--- a/tensorboard/plugins/core/core_plugin_test.py
+++ b/tensorboard/plugins/core/core_plugin_test.py
@@ -18,7 +18,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import collections
+import collections.abc
 import contextlib
 import json
 import os
@@ -160,8 +160,8 @@ class CorePluginNoDataTest(tf.test.TestCase):
     def testRoutesProvided(self):
         """Tests that the plugin offers the correct routes."""
         routes = self.plugin.get_plugin_apps()
-        self.assertIsInstance(routes["/data/logdir"], collections.Callable)
-        self.assertIsInstance(routes["/data/runs"], collections.Callable)
+        self.assertIsInstance(routes["/data/logdir"], collections.abc.Callable)
+        self.assertIsInstance(routes["/data/runs"], collections.abc.Callable)
 
     def testIndex_returnsActualHtml(self):
         """Test the format of the root / endpoint."""
--- a/tensorboard/plugins/debugger/debugger_plugin_test.py
+++ b/tensorboard/plugins/debugger/debugger_plugin_test.py
@@ -18,7 +18,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import collections
+import collections.abc
 import json
 
 import tensorflow as tf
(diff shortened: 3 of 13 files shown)


tensorboard/plugins/audio/audio_plugin_test.py
```diff
@@ -18,7 +18,7 @@
 from __future__ import division
 from __future__ import print_function
 
-import collections
+import collections.abc
```
caisq: Is it necessary to make the change in this line? If you just `import collections`, the instances of `collections.abc.Callable` below will still work right? 
If the purpose is to limit the scope and import only what's used, consider `from collections import abc`.

This comment applies to other files touch in this PR as well.

wchargin: Yes, the change is necessary: `collections.abc` is a submodule, and must
be imported to be available on the `collections` object:

```python
>>> import collections
>>> collections.abc.Callable
  File "/usr/lib/python3.7/collections/__init__.py", line 55, in __getattr__
    raise AttributeError(f'module {__name__!r} has no attribute {name!r}')
AttributeError: module 'collections' has no attribute 'abc'
```

(Note that `os.path` is a special case: the `os` module itself
explicitly [installs an `os.path` module][1] because it’s dynamically
chosen from `posixpath` or `ntpath` at runtime.)

I considered `from collections import abc`, but decided that it was
confusing given that `abc` is also a standard library module that we use
fairly frequently.

[1]: https://github.com/python/cpython/blob/58a4054760bffbb20aff90290dd0f3554f7bea42/Lib/os.py#L93

caisq: @wchargin Hmm. Okay, I tested manually in my local ipython and the following worked:

```
In [1]: import collections                                                                                                                                                                             

In [2]: collections.abc                                                                                                                                                                                
Out[2]: <module 'collections.abc' from '/usr/local/google/home/cais/venv_tf_nightly/lib/python3.7/collections/abc.py'>

In [3]: collections.abc.Callable                                                                                                                                                                       
Out[3]: collections.abc.Callable
```

But I guess the tensorboard CI infra may be using a different version of Python.

wchargin: No, you’re using IPython, and the IPython infrastructure has already
imported `collections.abc` at some point, which causes the `abc`
attribute to be assigned onto the `collections` module object. That is,
importing a new submodule has a side effect of mutating the parent
module object.

It’s not a Python version thing. You should be able to reproduce this in
a standard Python shell or Python script.

PR #3216, opened 2020-02-05
Add plugin to redirect users to new profile plugin

wchargin:
Summary:
We add a new `profile_redirect` plugin, called “Profile (moved)” in the
UI, which teaches users how to install the dynamic plugin:

![Screenshot of the new plugin][ss]

The instructions currently pull in Test PyPI sources; we’ll swap these
out for real PyPI once the package is published.

[ss]: https://user-images.githubusercontent.com/4317806/73811360-229f1300-478e-11ea-9341-f38d59170e3b.png

Test Plan:
Create a new virtualenv, install `tf-nightly`, uninstall `tb-nightly`,
and run `//tensorboard/pip_package:extract_pip_package` and install the
resulting wheel. Launch TensorBoard and note that the “Profile (moved)”
plugin appears in the inactive plugins list. Click the “Copy to
clipboard” button and execute the copied `pip install` command from the
test virtualenv. Relaunch TensorBoard, and note that the dynamic profile
plugin now appears and the “Profile (moved)” plugin is gone entirely.

Unit tests included for the conditional loading behavior. Changing the
attempted import (in `ProfileRedirectPluginLoader.load`) to a module
that always can be imported (like `sys`) causes one of the tests to
fail; changing it to one that can never be imported (like `zzz`) causes
the other to fail.

wchargin-branch: profile-redirect

### DIFF
--- a/tensorboard/plugins/profile_redirect/profile_redirect_plugin_test.py
+++ b/tensorboard/plugins/profile_redirect/profile_redirect_plugin_test.py
@@ -35,6 +35,9 @@ class ProfileRedirectPluginLoaderTest(tb_test.TestCase):
 
     def test_loads_when_no_dynamic_plugin(self):
         with contextlib.ExitStack() as stack:
+            stack.enter_context(mock.patch.dict(sys.modules))
+            sys.modules.pop(_DYNAMIC_PLUGIN_MODULE, None)
+
             real_import = __import__
 
             def fake_import(name, *args, **kwargs):
@@ -43,12 +46,12 @@ class ProfileRedirectPluginLoaderTest(tb_test.TestCase):
                 else:
                     return real_import(name, *args, **kwargs)
 
+            stack.enter_context(mock.patch("builtins.__import__", fake_import))
+
             plugin_class = profile_redirect_plugin._ProfileRedirectPlugin
             plugin_init = stack.enter_context(
                 mock.patch.object(plugin_class, "__init__", return_value=None)
             )
-            stack.enter_context(mock.patch("builtins.__import__", fake_import))
-            stack.enter_context(mock.patch.dict(sys.modules))
 
             loader = profile_redirect_plugin.ProfileRedirectPluginLoader()
             context = base_plugin.TBContext()
@@ -73,12 +76,12 @@ class ProfileRedirectPluginLoaderTest(tb_test.TestCase):
                 else:
                     return real_import(name, *args, **kwargs)
 
+            stack.enter_context(mock.patch("builtins.__import__", fake_import))
+
             plugin_class = profile_redirect_plugin._ProfileRedirectPlugin
             plugin_init = stack.enter_context(
                 mock.patch.object(plugin_class, "__init__", return_value=None)
             )
-            stack.enter_context(mock.patch("builtins.__import__", fake_import))
-            sys.modules.pop(_DYNAMIC_PLUGIN_MODULE, None)
 
             loader = profile_redirect_plugin.ProfileRedirectPluginLoader()
             context = base_plugin.TBContext()
--- a/tensorboard/plugins/profile_redirect/profile_redirect_plugin_test.py
+++ b/tensorboard/plugins/profile_redirect/profile_redirect_plugin_test.py
@@ -1,0 +1,91 @@
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Tests for `profile_redirect_plugin`."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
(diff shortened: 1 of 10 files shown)


wchargin:
Reviewers: @qiuminxu for anything profile-specific (wording, etc.);
@stephanwlee for general frontend.

tensorboard/plugins/profile_redirect/profile_redirect_plugin_test.py
```diff
@@ -0,0 +1,94 @@
...
+                    sys.modules.setdefault(
+                        _DYNAMIC_PLUGIN_MODULE, arbitrary_module
+                    )
+                    return arbitrary_module
+                else:
+                    return real_import(name, *args, **kwargs)
+
+            stack.enter_context(mock.patch("builtins.__import__", fake_import))
+
+            plugin_class = profile_redirect_plugin._ProfileRedirectPlugin
```
stephanwlee: General code convention question: is it okay to reach/probe into a private variable? How important is it that we assert the `__init__` as long as we assert the right `result` from the loader?

tensorboard/plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-redirect-dashboard.html
```diff
@@ -0,0 +1,110 @@
...
+      >
+      <div id="copyContainer">
+        <span id="copiedMessage"></span>
+        <paper-button raised on-tap="_copyInstallCommand"
+          >Copy to clipboard</paper-button
+        >
+      </div>
+    </div>
+
+    <style include="dashboard-style"></style>
```
stephanwlee: Omit this unless you are making use of it. Else, I believe you need to import https://github.com/tensorflow/tensorboard/blob/master/tensorboard/components/tf_dashboard_common/dashboard-style.html in rel="import".

tensorboard/plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-redirect-dashboard.html
```diff
@@ -0,0 +1,110 @@
...
+            try {
+              await navigator.clipboard.writeText(textarea.textContent);
+            } catch (error) {
+              // Fallback approach.
+              if (!document.execCommand('copy')) {
+                return Promise.reject();
+              }
+            }
+          })();
+          tryCopy
```
stephanwlee: FYI: since we are only "officially supporting" evergreen browsers, you may use `async/await` and make this code read better (wait... you are already using async/await :)).

```ts
async _copyInstallCommand() {
    const tryCopy = async () => {
      const textarea = this.$.commandTextarea;
      textarea.select();
      try {
        await navigator.clipboard.writeText(textarea.textContent);
      } catch (error) {
        // Fallback approach.
        if (!document.execCommand('copy')) {
          return Promise.reject();
        }
      }
    };

    try {
      await tryCopy();
      this.$.copiedMessage.innerText = 'Copied.';
    }  catch (e) {
      this.$.copiedMessage.innerText = 'Failed to copy to clipboard..';
    }
  }
```

tensorboard/plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-redirect-dashboard.html
```diff
@@ -0,0 +1,110 @@
...
+              '    tensorboard_plugin_profile',
+            ].join('\n'),
+          },
+        },
+        _copyInstallCommand() {
+          const tryCopy = (async () => {
+            const textarea = this.$.commandTextarea;
+            textarea.select();
+            try {
+              await navigator.clipboard.writeText(textarea.textContent);
```
stephanwlee: naive question: wouldn't it be better to just use `this. _installCommand` instead of reading the DOM (even if it is readonly)? Is it because we already select the `textNode` in the textarea for the `execCommand` case? If answer is, 'DOM should be SOT', then would it be nicer to inline content of `_installCommand` onto the template instead of having the property?

tensorboard/plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-redirect-dashboard.html
```diff
@@ -0,0 +1,110 @@
...
+            } catch (error) {
+              // Fallback approach.
+              if (!document.execCommand('copy')) {
+                return Promise.reject();
```
stephanwlee: I believe you can throw an error here instead (since this is inside an async block; it effectively is the same but functionally might differ--promise that returns rejected promise vs. promise that rejects).

tensorboard/plugins/profile_redirect/profile_redirect_plugin_test.py
stephanwlee: No need for remove.

tensorboard/plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-redirect-dashboard.html
```diff
@@ -0,0 +1,108 @@
...
+  </template>
+
+  <script>
+    (function() {
+      Polymer({
+        is: 'tf-profile-redirect-dashboard',
+        properties: {
+          _installCommand: {
+            type: String,
+            readonly: true,
```
stephanwlee: readonly -> readOnly

wchargin: Good catch; thanks! Done.

qiuminxu:
Tested works great, thanks William for the friendlier solution!

wchargin:
Thanks for the review, @qiuminxu, and thanks for testing!

PR #3218, opened 2020-02-05
profile: defer to dynamic plugin when available

wchargin:
Summary:
When the dynamic profile plugin is installed, the static profile plugin
will no longer be loaded. This enables the dynamic profile plugin to use
the same plugin name as the static plugin without worrying about
conflicts, and also enables a clean migration path so that we don’t need
to remove the static plugin at the exact same time that we release the
dynamic plugin.

Test Plan:
Create a new virtualenv, install `tf-nightly`, uninstall `tb-nightly`,
and run `//tensorboard/pip_package:extract_pip_package` and install the
resulting wheel. Launch TensorBoard and note that the static profile
plugin works as before. Install the dynamic profile plugin via the
instructions on the `profile_redirect` plugin and relaunch TensorBoard.
Note that the dynamic profile plugin has replaced both the static plugin
and the redirect notice. In the virtualenv contents, manually change all
instances of `experiment_profile` to `profile` in the plugin code and
resources, and relaunch TensorBoard:

```
find "$VIRTUAL_ENV"/lib/python?.?/site-packages/tensorboard_plugin_profile \
    -type f -execdir sed -i s/experiment_profile/profile/g {} +
```

Note that the dynamic profile plugin is now called just “profile”, and
still works: prior to this commit, this would have raised a “Duplicate
plugins for name profile” startup error.

wchargin-branch: profile-defer-dynamic

### DIFF
--- a/tensorboard/plugins/profile/profile_plugin_loader.py
+++ b/tensorboard/plugins/profile/profile_plugin_loader.py
@@ -50,6 +50,15 @@ through tpu profiler analysis grpc. The grpc channel is not secured.\
           A ProfilePlugin instance or None if it couldn't be loaded.
         """
         try:
+            import tensorboard_plugin_profile
+
+            # If the dynamic plugin is available, don't load this plugin
+            # at all.
+            return None
+        except ImportError:
+            pass
+
+        try:
             # pylint: disable=unused-import
             import tensorflow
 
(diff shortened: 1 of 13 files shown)


tensorboard/plugins/profile/profile_plugin_loader.py
```diff
@@ -49,6 +49,15 @@ def load(self, context):
...
         """
+        try:
+            import tensorboard_plugin_profile
+
+            # If the dynamic plugin is available, don't load this plugin
+            # at all.
+            return None
+        except ImportError:
+            pass
+
```
qiuminxu: It seems we will need line:61-69 some where in profiler dynamic plugin.
We cannot have dynamic plugin directly depends on tensorflow in pip dependency, as that will cause circular dependency.

wchargin: Yes, agreed.

PR #3221, opened 2020-02-06
docs: publish “Generic data ingestion” design doc

wchargin:
Summary:
This design doc was written while generalizing the support for data
providers beyond just scalar data. It describes an approach for reading
tensor and blob sequence data from disk that is extensible along the
plugin axis (any plugin can write compatible data) and the storage layer
axis (any backend can ingest this data with the same logic).

wchargin-branch: docs-design-generic-data-ingestion

### DIFF
--- a/docs/design/generic_data_ingestion.md
+++ b/docs/design/generic_data_ingestion.md
@@ -124,8 +124,8 @@ message SummaryMetadata {  // (extend)
 ```
 
 The value of `data_class` on a tag’s summary metadata, whether explicitly set or
-implicitly zero, establishes constraints on the tensor values for all Values of
-the same time series:
+implicitly zero, establishes constraints on the tensor values for all `Value`s
+of the same time series:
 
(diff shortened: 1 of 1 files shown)


docs/design/generic_data_ingestion.md
```diff
@@ -0,0 +1,325 @@
...
+Then, extend the existing `SummaryMetadata` to include a field of this type:
+
+```proto
+message SummaryMetadata {  // (extend)
+    DataClass data_class = 4;
+}
+```
+
+The value of `data_class` on a tag’s summary metadata, whether explicitly set or
+implicitly zero, establishes constraints on the tensor values for all Values of
```
bileschi: nit strange capital V in Values

wchargin: Thanks! This is correct, actually—it’s referring to the `Value` protos
in the `Summary.value` repeated field. But it should have code
formatting to make that clear: done.

PR #3222, opened 2020-02-06
Surface HTML import interop for plugin_lib

stephanwlee:
plugin_lib was initially designed to be used with npm package and a
bundler. However, because we have a need to use it under traditional
Polymer code that uses HTML import, we need to devise a way for the
lib to work with tf_web_library.

Changed our test to use the new interop layer.

Also tested whether WIT builds with the library (https://github.com/stephanwlee/what-if-tool/commit/5bc972f544b0e23d1ac488b1c43a2c7c668eba7e). I was not able to check for correctness as wit does not build correct pip package.

### DIFF
--- a/tensorboard/components/experimental/plugin_lib/polymer-interop.ts
+++ b/tensorboard/components/experimental/plugin_lib/polymer-interop.ts
@@ -1,6 +1,4 @@
-<!--
-@license
-Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


tensorboard/components/experimental/plugin_lib/polymer-interop.ts
```diff
@@ -0,0 +1,29 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import * as publicApi from './index';
+
+namespace tb_plugin_lib.experimental {
+  export const core = publicApi.core;
+  export const runs = publicApi.runs;
+}
```
psybuzz: Would we want to carefully select which APIs to expose, or should we just wholesale export:
```
namespace tb_plugin_lib {
    experimental: publicApi
}
```
?

PR #3226, opened 2020-02-07
Add the WriteBlob bidirectional streaming RPC to the write service.

davidsoergel:


tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +121,82 @@ message WriteScalarResponse {
...
+// A single chunk of the blob write stream.
+message WriteBlobRequest {
+  WriteBlobFirstRequestData write_blob_first_request_data = 1;
+  bytes data = 2 [ctype = CORD];
+  // The position in the blob where this chunk begins.
+  // This must equal the sum of the sizes of the chunks sent so far.
+  int64 offset = 3;
+  // CRC32C of current data buffer. While optional it is recommended that
+  // clients include the crc32c for every data buffer, to protect against data
+  // corruption. Note that for multi-shot writes, specifying the crc32c for
```
wchargin: We have the flexibility to make these required if you want, even if
they’re not required in the APIs from which these are inspired. What do
you think?

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +121,82 @@ message WriteScalarResponse {
   // This is empty on purpose.
 }
 
+// The blob writing procedure works as follows:
```
wchargin: Appreciate the super clear protocol docs. Thanks.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +121,82 @@ message WriteScalarResponse {
...
+//    and the subsequent ones must not.
+//  * The last of these requests must set finalize_object=True, and the prior
+//    ones must not.
+//  * For a blob <= 4MB, a single chunk suffices, so the single request is both
+//    the first and the last.
+//  * The caller receives a sequence of WriteBlobResponses in return.  Each
+//    WriteBlobResponse corresponds to one WriteBlobRequest in order, reporting
+//    the state of the write operation.
+//  * The caller may choose to verify each response before sending the next
+//    request, or it may continue sending requests optimistically.
```
wchargin: What does it mean to verify the response? If (e.g.) the response
contains a `size` other than the size that the client expects to have
seen, what can the client do?

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +121,82 @@ message WriteScalarResponse {
...
+  fixed32 crc32c = 4;
+  // Indicates that this is the last chunk of the stream.
+  bool finalize_object = 5;
+  // CRC32C of the entire blob. Optional but recommended, to protect against
+  // data corruption. This should be set only when finalize_object=True.
+  fixed32 final_crc32 = 6;
+}
+
+message WriteBlobResponse {
+  enum ObjectState {
```
wchargin: I don’t understand what this `ObjectState` means or how the client will
use it. If the client sends the final chunk of a blob and the server
responds `UNFINALIZED` because the blob is still being asynchronously
written to the backing store, there’s no way that the client will be
able to see `CURRENT`, right, because there are no more request–response
pairs to be sent?

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +121,82 @@ message WriteScalarResponse {
...
+// Note that the WriteBlobRequest does not mirror the nested structure of
+// WriteScalarRequest, because we only ever send one blob at a time.
+
+// Metadata included with only the first chunk of the blob write stream.
+message WriteBlobFirstRequestData {
+  string experiment_id = 1;
+  string run = 2;
+  string tag = 3;
+  int64 step = 4;
+  int64 index = 5;
```
wchargin: Does this mean that it’s generally possible to append to the blob
sequence at a particular step? I had assumed that the client would be
required to specify the sequence length atomically at initial write
time, because conceptually the whole sequence is one unit. (Just as it
would be weird to show the user a torn state where we’ve received only
the first 4 bytes of their scalar, it seems slightly strange to show a
torn state where we’ve received only the first blob in a sequence.
Recall that sequences aren’t homogeneous—for instance, images are
represented as `[width, height, ...pngs]`.)

Relatedly, when we want to support resumable uploads, how will we know
whether there remain blobs that need to be appended to any sequence?

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +121,82 @@ message WriteScalarResponse {
...
+//  * The caller receives a sequence of WriteBlobResponses in return.  Each
+//    WriteBlobResponse corresponds to one WriteBlobRequest in order, reporting
+//    the state of the write operation.
+//  * The caller may choose to verify each response before sending the next
+//    request, or it may continue sending requests optimistically.
+//
+// Note that the WriteBlobRequest does not mirror the nested structure of
+// WriteScalarRequest, because we only ever send one blob at a time.
+
+// Metadata included with only the first chunk of the blob write stream.
```
wchargin: Potentially no action required: it’s also true that metadata is only
required on the first blob sequence in a time series. Up to you whether
this should be noted/clarified.

tensorboard/uploader/proto/write_service.proto
davidsoergel: I don't feel strongly either way, but sure, may as well.

tensorboard/uploader/proto/write_service.proto
davidsoergel: Right, it can check that the object_state and size are as expected, and raise an error if they're not, i.e. to avoid continuing to send data after the stream is already borked.  Perhaps we can rely on the RPC itself throwing errors in this case.  This is entangled with outstanding issues re whether the bidi-stream can be async (see below).

tensorboard/uploader/proto/write_service.proto
davidsoergel: Thanks!  Though I think I will need to update them in light of the comment threads. :)

tensorboard/uploader/proto/write_service.proto
davidsoergel: Refactoring in the works, pending offline discussion.

tensorboard/uploader/proto/write_service.proto
davidsoergel: The server shouldn't send any response at all until the backing write is complete.  That's how things should work whether the server is using sync or async RPCs to backing storage.  In the sync case it's clear: the response to the client is placed on the stream after the backing write succeeds.  In the async case, the response to the client is placed on the stream in the backing RPC callback-- again, after the write succeeds.

A key point here is that the client RPC, as a bidirectional stream, has a sort-of-async quality: the client may choose to ignore the response stream while continuing to send requests, but then still look at the responses later.  At least, that's what I *thought* would happen, but now it appears that's [not directly supported yet](https://github.com/grpc/grpc/issues/18376).  The default behavior is basically synchronous after all (i.e., the client receives each response before sending the next request).  There is a (kinda messy) [workaround](https://stackoverflow.com/questions/55029342/handling-async-streaming-request-in-grpc-python) though.

So I think my (strawman) plan is:
 * Phase 1: sync calls everywhere, for simplicity.
 * Phase 2: async calls from the server to storage.  This is not too hard, and means that the client will still block but the server won't.
 * Phase 3: Apply the async workaround above to the client.

The latter two are optimizations re resource usage and bandwidth, so I think they can be postponed for a short time at least.

tensorboard/uploader/proto/write_service.proto
davidsoergel: Refactoring in the works, pending offline discussion.

tensorboard/uploader/proto/write_service.proto
davidsoergel: Updated substantially re BlobSequences and resumable uploads.

tensorboard/uploader/proto/write_service.proto
davidsoergel: I left the fields in the response because they are _sometimes_ useful, but the docstring now discourages validating every chunk, since that reduces throughput with little benefit.

tensorboard/uploader/proto/write_service.proto
davidsoergel: I think the updated design re CreateBlobSequence vs WriteBlob fully clarifies this.

tensorboard/uploader/proto/write_service.proto
davidsoergel: I think the updated design re CreateBlobSequence vs WriteBlob fully clarifies this.  Basically I now allow writing the blobs in any order, and in a resumable manner.

Re the corresponding internal change, I didn't add an object_state to the BlobSequence table, because it's easy enough to select all of the blobs and check their object_states.  Indeed that will happen naturally on the read path anyway, so we can handle the incomplete-upload scenario appropriately there.

tensorboard/uploader/proto/write_service.proto
davidsoergel: Done (here, just as a matter of documentation).

tensorboard/uploader/proto/write_service.proto
davidsoergel: Discussed offline; sounds like the above plan is ok.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +124,125 @@ message WriteScalarResponse {
...
+//    Since this is a second (or later) attempt regarding an existing sequence,
+//    a new sequence is not created.  Instead the existing sequence ID is
+//    returned.
+//  * The client attempts to upload each blob in the sequence.  These uploads
+//    resume or may be skipped entirely, as needed, according to the below
+//    procedure.
+
+// The blob writing procedure is as follows.  It is intrinsically resumable.
+//  * The caller instantiates a WriteBlob RPC, which is a bidirectional stream.
+//  * The first of these requests must provide write_blob_first_request_data,
```
nfelt: I'm not sure I see the benefit of providing the identifying info for the blob to write on only the first request in the stream.  It's a very small amount of overhead relative to the blob size itself, and I think the client side logic is simpler if it has to maintain less state about which chunk it's on.  In general it seems cleaner if the first chunk sent doesn't look any different from an arbitrary chunk _n_ (except for the final chunk).

The server side already has to keep track of the state, so there's no change there (except it should validate that it's the same blob being written and fail the request if not, but this is nice since it's a cheap check and helps catch client errors and avoids ambiguity).

Making the "key" to WriteBlobRequest mandatory is also just nice structurally since then even in the future if we pass around those request objects (or conceivably want to do something fancy like multiplex several blob writes into a single streaming RPC), we never have to worry about losing track of which blob we're talking about.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +124,125 @@ message WriteScalarResponse {
...
+//    and the subsequent ones must not.
+//  * The first request should send no actual data; it acts as a request for the
```
nfelt: IMO mandating that the first request be empty feels like it unduly favors an uncommon case (resuming an partial upload where the client is unsure what progress it made) over the far more common case (where the entire blob is <4MB and we could otherwise upload the entire thing in a single request).

If we make that part optional, the client could still use that to check progress in cases where it knows it's resuming an upload or if it knows that it's using a blob sequence that already exists, but in the common case where it's just churning through a time series and for each step creating 1 blob sequence and 1 blob, I think that would basically get us a free reduction from 3 to 2 round trips per step.

One other consideration too is that if we had CreateBlobSequence/GetBlobSequence return the WriteBlobRequest properties (for the blob if it already exists) then the client could still do the check against sending redundant data, but as part of the first RPC rather than needing a second intermediary RPC.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +124,125 @@ message WriteScalarResponse {
...
+//    object so far, i.e. taking into account any previously uploaded data.
+//  * The client proceeds to upload from the specified offset (optionally
+//    validating the crc32c up to that point first).
+//  * The caller sends a number of additional WriteBlobRequests, providing the
+//    blob data in an ordered sequence of chunks.
+//  * The caller receives a sequence of WriteBlobResponses in return.  Each
+//    WriteBlobResponse corresponds to one WriteBlobRequest in order, reporting
+//    the state of the write operation.
+//  * In general, the caller should not wait to verify that each response
+//    reports the expected status, size, and crc32c before sending the next
```
nfelt: FWIW, verifying the status and size is cheap, it's only the crc32c verification that is expensive.  If we don't expect the clients to use that field, is it worth setting it on the server?  I forget if we get it for free from GCS.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +124,125 @@ message WriteScalarResponse {
...
+//  * The last of these requests must set finalize_object=True, and the prior
+//    ones must not.
+//  * For a blob <= 4MB, a single data chunk suffices, so there are two
+//    requests: one to learn the status, and one to upload the data.
+//
+// Note that the WriteBlobRequest does not mirror the nested structure of
+// WriteScalarRequest, because we only ever send one blob at a time.
+
+message CreateBlobSequenceRequest {
+  string experiment_id = 1;
```
nfelt: Nit: could we get a comment for each field?  Can be a terse 1-liner but still worth having IMO per https://aip.dev/192

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +124,125 @@ message WriteScalarResponse {
...
+}
+
+message CreateBlobSequenceResponse {
+  string sequence_id = 1;
+}
+
+// Metadata included with only the first chunk of the blob write stream.
+message WriteBlobFirstRequestData {
+  string sequence_id = 1;
+  int64 index = 2;
```
nfelt: I assume the server validates that this must be between 0 and `final_sequence_length - 1`?

nfelt: I assume the server validates that this must be between 0 and `final_sequence_length - 1`?

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +124,125 @@ message WriteScalarResponse {
...
+// A single chunk of the blob write stream.
+message WriteBlobRequest {
+  WriteBlobFirstRequestData write_blob_first_request_data = 1;
+  // The bytes in this chunk.  For the first chunk only, this value should be
+  // empty.  This allows determining from the response how much of the blob has
+  // already been uploaded, if any, without sending redundant data.
+  bytes data = 2 [ctype = CORD];
+  // The position in the blob where this chunk begins.
+  // This must equal the sum of the sizes of the chunks sent so far (except on
+  // the first request).
```
nfelt: Clearer I think to state that this field is ignored if `data` is empty.  (That should be true regardless of whether we preserve the special first request semantics, and is also less ambiguous about whether the validation is based on the actual arrival sequence of the request at the server, or on the contents of the rest of the request message.)

tensorboard/uploader/proto/write_service.proto
davidsoergel: Agreed that the requirement to send no data on the first request was too strong.  It's fine either way, so we can leave the tradeoff to the client (e.g., it may choose a blob size threshold below which it sends data optimistically, and above which it checks status first, etc.).

I did previously have the `GetOrCreateBlobSequenceResponse` report `repeated WriteBlobResponse blob_statuses;`, but removed it on the theory that it added complexity, and that the extra backend queries needed to populate that would not be worth it most of the time.  Of course the empty-request plan recapitulated those costs, only much worse.  Oops.

Anyway, in the updated plan we do neither check, so the common case is as cheap as possible.  Additional costs are borne only when resuming uploads.  (We can always reinstate `GetOrCreateBlobSequenceResponse.blob_statuses` later, so let's not block on that now).

tensorboard/uploader/proto/write_service.proto
davidsoergel: Done.

tensorboard/uploader/proto/write_service.proto
davidsoergel: +1, thanks

tensorboard/uploader/proto/write_service.proto
```diff
@@ -190,6 +196,152 @@ message WriteTensorResponse {
...
+  .tensorboard.SummaryMetadata metadata = 6;
+}
+
+message GetOrCreateBlobSequenceResponse {
+  // A unique ID for the the requested blob sequence.
+  string sequence_id = 1;
+}
+
+// A single chunk of the blob write stream.
+message WriteBlobRequest {
```
davidsoergel: We could add a field here for the total file size.  That way we could avoid sending a partial file when the whole thing would exceed quota anyway.

It'd be optional, and typically would produce a quota failure (if any) on the first request-- so long as the user is uploading only one blob at a time.  To handle this in the case of parallel write streams, we'd have to pre-reserve quota for the incoming blobs.  Maybe not worth the complexity, and the server-side implementation is postponable in any case.

For now, any opinions on whether a file size field is worth adding right away?

tensorboard/uploader/proto/write_service.proto
```diff
@@ -119,6 +124,125 @@ message WriteScalarResponse {
...
+//    RPC as described below, providing the blob sequence ID and sequence index
+//    for each blob.
+//  * Multiple WriteBlob streams may operate in parallel, with no constraint
+//    on their ordering.
+
+// Blob sequence uploads are resumable, as follows:
+//  * The client starts the upload process with a new CreateBlobSequence RPC.
+//  * The blob sequences are uniquely keyed by experiment, run, tag, and step.
+//    Since this is a second (or later) attempt regarding an existing sequence,
+//    a new sequence is not created.  Instead the existing sequence ID is
```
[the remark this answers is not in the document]
nfelt: Works for me.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -175,77 +218,99 @@ message WriteScalarResponse {
...
+//  * The streams need not operate in 'ping-pong' style-- i.e. the client may
+//    continue to send requests, regardless of whether responses to prior
+//    requests have arrived yet.
+//
+//  * The first request sent over the stream may optionally contain no actual
+//    data. In this case it acts only as a request for the current status of
+//    the blob, in case it has already been partially or fully uploaded.
+//  * The first WriteBlobResponse reports the state, size, and crc32c of the
+//    object so far, i.e. taking into account any previously uploaded data.
+//    This response may imply that any provided data was ignored as redundant.
```
nfelt: I'm not quite sure what this sentence means.  Isn't it an error to provide data at an offset that doesn't correspond to the end of the blob?  So providing redundant data would always be an error, even if it's on the first request.

tensorboard/uploader/proto/write_service.proto
nfelt: If the field isn't always set on responses, for correctness we should have an explicit indicator to mark it as unset rather than leaving it at the default value of 0, since 0 is a valid CRC output: https://stackoverflow.com/a/49804094/1179226

The most idiomatic way to do that would be using the UInt32Value wrapper type (the well-known wrappers have special handling in JSON interop) and then we can leave it unset if no CRC was provided by the server:
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto

That said, up to you but I'd personally be inclined to leave it out unless we're confident that the server will populate it and that we think it's valuable for the client to check it (even in a deferred fashion).  As mentioned in your comment lower in the file, we have robust protection already from the full-file and per-chunk CRCs.

And actually, thinking about it just a bit further, I'm not sure that maintaining a running CRC provides any additional value over the per-chunk CRCs, because if a per-chunk CRC fails to detect corruption in that chunk, I think necessarily any further CRC calculations are not going to make a difference.  See the discussion in https://stackoverflow.com/a/23050676/1179226 which explains that for strings `a` and `b` you can calculate `crc(a + b)` using only `crc(a)`, `crc(b)` and `len(b)`.  So IIUC then if there's a chunk `b'` such that `crc(b') = crc(b)` and `len(b') = len(b)`, then we already know that `crc(a + b') = crc_combine(crc(a), crc(b'), len(b')) = crc_combine(crc(a), crc(b), len(b)) = crc(a + b)`, so the running CRC would also be identical.

I'm definitely not a CRC expert so maybe this analysis is flawed or there are ways to avoid that issue, but given that it seems like even if we did calculate it's possibly doing a lot of extra work for no extra protection, my inclination would be to omit it for now and we can always add it later.

tensorboard/uploader/proto/write_service.proto
nfelt: I'd defer for now since it adds complexity and hitting the quota limit is intended to be a rare edge case.  If we change our minds, a strength of protobuf is that it's very easy to add fields later (and we basically just have to do a PR and a sync and then we can use it internally).

tensorboard/uploader/proto/write_service.proto
davidsoergel: Ah, I see how that was unclear.  I think the current refactoring cleans it all up.

tensorboard/uploader/proto/write_service.proto
davidsoergel: Further clarified RPC structure and docs around this.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -287,15 +289,47 @@ message GetOrCreateBlobSequenceResponse {
...
+  }
+
+  // State of the object (still appending vs. complete).
+  BlobState object_state = 1;
+  // Size of the object in bytes.  This is the sum of the chunk sizes
+  // received from the stream so far.  In the response to the final chunk,
+  // this size should equal the total size of the blob.
+  int64 size = 2;
+  // crc32c of the blob data read so far, i.e. over the byte range [0, size).
+  fixed32 cumulative_crc32c = 3;
```
nfelt: Optional but IMO at this point maybe better to just call this `crc32c` (no cumulative) for consistency with `size` which is just `size` rather than `cumulative_size`.  I think there's only really one obvious interpretation of what the CRC and size refer to in the new structure, which is whatever amount of the blob currently exists on the server (since that's the whole point of this RPC).

tensorboard/uploader/proto/write_service.proto
```diff
@@ -287,15 +289,47 @@ message GetOrCreateBlobSequenceResponse {
...
+    UNFINALIZED = 1;
+
+    // Object is finalized.
+    CURRENT = 2;
+  }
+
+  // State of the object (still appending vs. complete).
+  BlobState object_state = 1;
+  // Size of the object in bytes.  This is the sum of the chunk sizes
+  // received from the stream so far.  In the response to the final chunk,
```
nfelt: This description is a bit stale - in the new context we don't necessarily have a stream or chunks, since we might send this request before even initiating a `WriteBlob` stream, and this also isn't ever in response to the final chunk.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -287,15 +289,47 @@ message GetOrCreateBlobSequenceResponse {
...
 
+message GetBlobMetadataRequest {
+  // The ID of the BlobSequence of which this blob is a member.
+  string blob_sequence_id = 1;
+  // The position of this Blob within the BlobSequence
+  int64 index = 2;
+}
+
+message GetBlobMetadataResponse {
+  enum BlobState {
```
nfelt: Only noticed now, but let's just make this a top-level symbol (not inside any single response), which is recommended by the proto style guide anyway: https://developers.google.com/protocol-buffers/docs/style#enums

That way we can reuse it across messages easily if we need to (though I'm not sure we do right now).

Even though it's small, might be worth putting it in a separate resource-message .proto file (`blob.proto`) so that it can be shared across services as well, e.g. hypothetically if we want to have the export service export the blob state, since as we discovered with Experiment it's a bit of a pain to relocate later.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -314,7 +348,7 @@ message WriteBlobRequest {
 }
 
 message WriteBlobResponse {
-  enum ObjectState {
+  enum BlobState {
```
nfelt: Do we actually need to report the BlobState in WriteBlobResponse any more?  I guess just to confirm that the `finalize_object` instruction was accepted?  IMO just getting a Status OK response to a message that contained the finalization instruction should probably be good enough, but I guess we could retain as a sanity check.

tensorboard/uploader/proto/write_service.proto
davidsoergel: Haha I had just reverted the top-level enum thing :)  Happy to reinstate that, especially in another file.

tensorboard/uploader/proto/write_service.proto
davidsoergel: Done

tensorboard/uploader/proto/write_service.proto
davidsoergel: Good catch

tensorboard/uploader/proto/write_service.proto
davidsoergel: Yeah, I figured they're a super cheap sanity check, so why not.

PR #3230, opened 2020-02-08
Convert BaseResponse to wrappers.BaseResponse

### LINKED ISSUE #3229: werkzeug 1.0.0 removed symbols
https://pypi.org/project/Werkzeug/1.0.0/ was release 2020-02-06 and it breaks our tests.

There are about 5 instances where we use `werkzeug.BaseResponse` (all in tests) but the API is now deprecated (`werkzeug.wrappers.BaseResponse` may be what we want).

Do note that this should impact CI, too, very soon (https://github.com/tensorflow/tensorboard/blob/38a326bd55b17d92bc60baf323b4e09c4a3f5f6f/tensorboard/pip_package/requirements.txt#L30)

stephanwlee:
Although the werkzeug changelog did not say so, it seem to have removed
certain symbols from `werkzeug`. We now use
`werkzeug.wrappers.BaseResponse` instead.

Context:
https://werkzeug.palletsprojects.com/en/master/changes/#version-1-0-0

Fixes #3229.

### DIFF
--- a/tensorboard/backend/empty_path_redirect_test.py
+++ b/tensorboard/backend/empty_path_redirect_test.py
@@ -38,7 +38,9 @@ class EmptyPathRedirectMiddlewareTest(tb_test.TestCase):
         app = empty_path_redirect.EmptyPathRedirectMiddleware(app)
         app = self._lax_strip_foo_middleware(app)
         self.app = app
-        self.server = werkzeug_test.Client(self.app, werkzeug.BaseResponse)
+        self.server = werkzeug_test.Client(
+            self.app, werkzeug.wrappers.BaseResponse
+        )
 
     def _lax_strip_foo_middleware(self, app):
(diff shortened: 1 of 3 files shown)


tensorboard/backend/empty_path_redirect_test.py
```diff
@@ -38,7 +38,9 @@ def setUp(self):
         app = empty_path_redirect.EmptyPathRedirectMiddleware(app)
         app = self._lax_strip_foo_middleware(app)
         self.app = app
-        self.server = werkzeug_test.Client(self.app, werkzeug.BaseResponse)
+        self.server = werkzeug_test.Client(
+            self.app, werkzeug.wrappers.BaseResponse
```
wchargin: OK: Confirmed that you don’t need to separately import `wrappers` and
that it’s available as `werkzeug.wrappers` by default. LGTM.

PR #3232, opened 2020-02-10
[infra] Upgrade FE dependencies

stephanwlee:
Notable changes:
- Angular 8 -> Angular 9
- zonejs upgrade
- ngrx @ 9
- TypeScript @ 3.8.3
- Karma @ 5
- Angular Ivy compiler (ngcc)
- Updated rules_nodejs to make it work with Ivy compiler
- use node 11 in travis

We used [1] as inspiration to our changes.

Other notable changes:
- window.require is frozen object after require.js upgrade

[1]:
https://github.com/bazelbuild/rules_nodejs/tree/master/examples/angular
cc: @caisq


stephanwlee:
The change is ready to be reviewed. Will update the TF cache when I get LGTM.

.travis.yml
```diff
@@ -58,6 +61,7 @@ install:
       -r tensorboard/pip_package/requirements.txt \
       -r tensorboard/pip_package/requirements_dev.txt \
       ;
+  - nvm use v11
```
nfelt, line 64: Maybe a comment about keeping this in sync with the `node_js` stanza above?  Assuming they should be linked?

package.json
```diff
@@ -28,43 +28,59 @@
...
-    "@angular/cli": "^8.3.21",
-    "@angular/compiler": "^8.2.14",
-    "@angular/compiler-cli": "^8.2.14",
-    "@bazel/bazel": "0.26.1",
-    "@bazel/hide-bazel-files": "^0.34.0",
+    "@angular/bazel": "^9.0.0",
+    "@angular/cli": "^9.0.1",
+    "@angular/compiler": "^9.0.0",
+    "@angular/compiler-cli": "^9.0.0",
+    "@bazel/bazel": "^0.23.2",
```
nfelt, line 35: This actually downgrades our `@bazel/bazel` from `0.26.1` to just `0.23.2` (per the impact on yarn.lock).  Was that actually necessary?

If not, it seems better if we can to use 2.1.0 to match our Bazel version, since it looks like this package is released approximately on the overall Bazel release schedule:
https://www.npmjs.com/package/@bazel/bazel/v/2.1.0?activeTab=versions

tensorboard/components/tensor_widget/BUILD
```diff
@@ -100,7 +100,6 @@ tf_ts_library(
 
 tf_ts_devserver(
     name = "dev_server",
-    index_html = "demo/demo.html",
```
nfelt, line 103: I noticed that removing this was required when I was trying to do this upgrade too. Not blocking but cc @caisq in case this affects the devserver and it needs to be retooled at all to work again after this change.

tensorboard/webapp/BUILD
```diff
@@ -55,10 +55,10 @@ ng_module(
 # compatibility, Vulcanization, and configuration (e.g. prod vs. dev).
```
nfelt, line 55: Comment here is stale now that it's not switching prod vs dev.

Is it actually fine to be using prod mode all the time, or is this just temporary until we can get it working again?  If the latter, maybe we should include a TODO?

.bazelrc
```diff
@@ -1,5 +1,3 @@
-# For compatibility with Bazel 0.27 through 0.29. See:
```
nfelt, line 1: A bit confusing to remove these options as part of this PR - I've factored them out into https://github.com/tensorflow/tensorboard/pull/3535

package.json
```diff
@@ -28,43 +28,59 @@
...
     "@bazel/ibazel": "^0.10.3",
-    "@bazel/jasmine": "^0.32.2",
-    "@bazel/karma": "^0.36.1",
-    "@bazel/typescript": "^0.34.0",
+    "@bazel/jasmine": "^1.2.4",
```
nfelt, line 37: These packages are all provided by rules_nodejs and in the past we had them on the same version as our WORKSPACE `rules_nodejs` dep (which this PR sets to 1.6.0).

Would it be possible to set these all to `^1.6.0` so that we're matching the two types of dependencies?

.bazelrc
stephanwlee: Thanks for doing that!

PR #3227, opened 2020-02-07
Proto changes to support experiment name & description.

bileschi:
* Motivation for features / changes
Support for experiment name & description in TensorBoard.dev

* Technical description of changes
Extends proto service adding:
SetExperimentName and SetExperimentDescription
Extends uploader to read those fields during export

* Screenshots of UI changes
OSS product users should see no chnage.

* Detailed steps to verify changes work correctly (as executed by you)
bazel test tensorboard:all in a virtual env

PR #3234, opened 2020-02-11
Proto changes to support experiment name & description.

bileschi:
*  Motivation for features / changes
Support for experiment name & description in TensorBoard.dev

*  Technical description of changes
Extends proto service adding:
UpdateExperimentMetadata
OSS product users should see no chnage.

Detailed steps to verify changes work correctly (as executed by you)

$ bazel test //tensorboard/... --test_tag_filters=-webtest --build_tests_only

(this is a continuation of a previous messed up PR #3227 )


bileschi:
Investigating the failure of //tensorboard/pip_package:test_pip_package in Travis

bileschi:
Updated to remove duplicate definition of Experiment.  Moved single source of truth to within export_service.proto

tensorboard/uploader/proto/experiment.proto
```diff
@@ -0,0 +1,43 @@
+syntax = "proto3";
+
+package tensorboard.service;
+
+import "google/protobuf/timestamp.proto";
+
+// Metadata about an experiment.
```
nfelt: It's good form to indicate (once we support updating some of this metadata) which fields are output-only and cannot be directly set by the client.  (In this case, all of them except name and description.)

See e.g. https://aip.dev/203#output-only (I don't think we necessarily need to use the programmatic specifier, but having the documentation in a comment is still valuable)

wchargin:
Delegating to others while I’m out; see my comments on the previous
#3227 and the internal <http://cl/293665881> for my perspective.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -2,6 +2,10 @@ syntax = "proto3";
 
 package tensorboard.service;
 
+// TODO(bileschi): write_service.proto imports export_service.proto for the
```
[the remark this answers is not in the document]
nfelt: Ok, reading comments from @wchargin on #3227 it's true that removing the Experiment from export_service.proto will require internal python changes at sync time, but I'm not sure it's worth trying to avoid that anyway.  If we did, we'd have to do a much more involved series of steps (create a copy of the proto under a different temporary name, then update the python code to use the copy, then move the actual proto message to the new location, then update the python code to use that, and finally remove the temporary proto).

But at any rate, I was wrong to say it's "very straightforward" so ok to defer (but then let's revert the addition of `experiment.proto` in this PR until we do that change).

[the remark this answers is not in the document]
bileschi: I had thought experiment.proto has been reverted.  My bad.  Removed.

caisq:
cc myself

tensorboard/uploader/proto/write_service.proto
```diff
@@ -23,13 +27,19 @@ service TensorBoardWriterService {
   // Request that the calling user and all their data be permanently deleted.
   // Used for testing purposes.
   rpc DeleteOwnUser(DeleteOwnUserRequest) returns (DeleteOwnUserResponse) {}
+  // Request to mutate metadata associated with an experiment.
+  rpc UpdateExperimentMetadata(UpdateExperimentMetadataRequest)
```
nfelt: To echo William's comments internally, and for consistency with usual google API style, can we call this just `UpdateExperiment` (and ditto for the request and response messages) since it's acting on the `Experiment` resource directly?

Also, nit but would be nice to put it up between CreateExperiment and DeleteExperiment so that the CRUD operations fro Experiment are all adjacent to each other.

bileschi: Done, thanks for the well motivated comment!

PR #3238, opened 2020-02-11
Fix scalar demo

ericdnielsen:
Fix the Tf/tf2 issue in the scalars demo.


ericdnielsen:
Demo runs and generates data now.  However its appears that the local tensorboard doesn't recognize the data. (Nothing rendered when pointing to that log_dir).  So I suspect there's still some hidden tf/tf2 issues.

I've grabbed a non-demo logdir for my own testing that I was trying to use the demo for, so this is no longer on my critical path.

bileschi (approved):
LGTM

bileschi:
This bit us again.  Going to merge this CL.

PR #3239, opened 2020-02-12
Upgrade min Bazel required and upgrade CI to use 2.1

stephanwlee:
Bazel is now out of the beta mode (version 0.x). We want to depend on
1+ for better long term support in the future.


stephanwlee:
This is the reason: 
image

49m is too tight for our comfort and I wasn't sure if we should pursue this route. It just has to miraculously pass only once for cache to prolong our life with Travis but that's... not cool.

nfelt:
💥 💥 💥

PR #3242, opened 2020-02-12
Move What-If Tool to a dynamic plugin

jameswex:
Remove WIT code from TensorBoard and move it to a dynamic plugin.

Removed all WIT code, which now exists in github.com/pair-code/what-if-tool
Added TB pip package dependency on new tensorboard-plugin-wit dynamic plugin package
Remove tf-serving vendoring from pip-package building as it was only used by WIT code

* Screenshots of UI changes
No visual change

Built and ran TensorBoard locally with this change, with tensorboard-plugin-wit package installed, and used What-If Tool, including parsing of plugin-specific URL parameters.

N/A


tensorboard/pip_package/requirements.txt
```diff
@@ -27,6 +27,7 @@ protobuf >= 3.6.0
 requests >= 2.21.0, < 3
 setuptools >= 41.0.0
 six >= 1.10.0
+tensorboard-plugin-wit >= 1.6.0
```
stephanwlee: @nfelt @wchargin FYI

PR #3245, opened 2020-02-13
Add experient_metadata to DataProvider

caisq:
  * Lay foundation for displaying experiment metadata (incl. timestamp for creation, name and description) in tensorboard.dev

  * Add method `experiment_metadata()` to `DataProvider` base class, with a default return value of `None`.
  * Add data class for experiment metadata: `ExperimentMetadata`.

### DIFF
--- a/tensorboard/data/provider.py
+++ b/tensorboard/data/provider.py
@@ -102,6 +102,22 @@ class DataProvider(object):
         """
         return ""
 
+    def experiment_metadata(self, experiment_id):
+        """Retrive metadata of a given experiment.
+
+        The metadata may include fields such as name and description
+        of the experiment, as well as a timestamp for the experiment.
+
+        Args:
+          experiment_id:  ID of the experiment in question.
+
+        Returns:
+          If the metadata does not exist, `None`.
+          Otherwise, an `ExperimentMetadata` object containing metadata about
+            the experiment.
+        """
+        return None
+
     def list_plugins(self, experiment_id):
         """List all plugins that own data in a given experiment.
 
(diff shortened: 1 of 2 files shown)


caisq:
cc @bileschi

tensorboard/data/provider.py
```diff
@@ -307,6 +323,35 @@ def read_blob(self, blob_key):
         pass
 
 
+class ExperimentMetadata(object):
```
nfelt: Just curious, is the thought that we'll have some custom code to serialize this into the JSON XHR response?  Or should this class itself define basically a way to convert it into a dict that can then be converted to JSON?

caisq: Good question. The thinking is that

1. Defining a data class like this spells out what the experiment metadata is expected to look like.
2. It can potentially have methods like `to_json()` to support convenient conversion to a serializable format.
3. `Run` is also a class in the same file. So this seems nice and consistent.

wchargin:
Should `experiment_metadata` subsume `data_location`? Both of them get
metadata about the experiment, just different pieces of metadata. It
seems strange to have two separate methods, especially given that
they’re each called exactly once, one after the other, in the handler
for `/data/environment`.

Can we clean this up by (1) adding a `data_location` field to the
`ExperimentMetadata` struct, (2) fixing up existing implementations to
populate it, (3) changing `CorePlugin` to read from it, and (4) removing
the `data_location` method? This wouldn’t even be a breaking change if
we include a default-`""` value in step (2), but we’re entirely
justified to make breaking changes to these APIs, and I think that it’s
important to leave them better than we found them.

caisq:
@wchargin You have more knowledge and experience on this topic than I do. But I'm not entirely convinced that the current division between `experiment_metadata()` and `data_location()` is that strange. `experiment_metadata()` describes something about the experiment, which is intrinsic to the experiment. `data_location()` is about where the experiment's data lives. The data location is external to the experiment per se and may change with time (migrating somehow), without the metadata changing. 

Will be happy to discuss this further.

PR #3248, opened 2020-02-13
Show warning if trace viewer is used with WebComponents v0

psybuzz:

Warn users of tf-nightly who use the trace viewer in the existing Profiler plugin that it is broken.

Within the Trace viewer iframe, detect whether WebComponentsV0 are available.  Show the warning if it is not.

* Screenshots of UI changes

image

Tested on both Chrome M80 and M79, checked that the warning only shows in the newer version.  Note that, when origin trial is available, the warning should not appear.

### DIFF
--- a/tensorboard/components/tf_trace_viewer/tf-trace-viewer.html
+++ b/tensorboard/components/tf_trace_viewer/tf-trace-viewer.html
@@ -15,6 +15,75 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
+<!-- Before any Trace Viewer scripts run, show warning if needed. -->
+<style>
+  .no-webcomponents-container.hidden {
+    display: none;
+  }
+
+  .no-webcomponents-container {
+    display: flex;
+    justify-content: center;
+    margin-top: 80px;
+    font-family: Roboto, sans-serif;
+  }
+
+  .no-webcomponents-container .title {
+    font-size: 22px;
+    font-weight: bold;
+  }
+
+  .no-webcomponents-container .contents {
+    max-width: 600px;
+  }
+</style>
+
+<div class="no-webcomponents-container hidden">
+  <div class="contents">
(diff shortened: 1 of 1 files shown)


psybuzz:
stephanwlee: Please take a look
qiumin, galoshri: FYI, feel free to comment

tensorboard/components/tf_trace_viewer/tf-trace-viewer.html
```diff
@@ -15,6 +15,75 @@
...
+    );
+    containerEl.classList.remove('hidden');
+
+    const currentURL = new URL(window.location.href);
+    const relativeTraceURL = currentURL.searchParams.get('trace_data_url');
+    const traceURL = new URL(relativeTraceURL, window.location.href).toString();
+
+    const workaroundEl = document.createElement('p');
+    workaroundEl.textContent =
+      `If you are running a Chromium-based browser, ` +
```
stephanwlee: feels like this can be part of the DOM in L52. Can we just do that?

tensorboard/components/tf_trace_viewer/tf-trace-viewer.html
```diff
@@ -15,6 +15,75 @@
 limitations under the License.
 -->
 
+<!-- Before any Trace Viewer scripts run, show warning if needed. -->
```
stephanwlee: Please note that this is meant to work without Polymer and hence cannot be part of Polymer component.

tensorboard/components/tf_trace_viewer/tf-trace-viewer.html
psybuzz: Done.

tensorboard/components/tf_trace_viewer/tf-trace-viewer.html
psybuzz: Sounds good to call out.  I've updated the comment.

PR #3249, opened 2020-02-14
Allow tensorboard title in header to be click target with configurable url

bmd3k:
Allow instances of tensorboard (tensorboard.dev, for example) to specify that the title in the header should be a click target to some URL.

tf-tensorboard component now contains a homeHref property. If homeHref is set, title is rendered as anchor element with href attribute set appropriately. If unset, title is just a div and not a click target.

Wrote tests.
Modified tensorboard.html to include home-href attribute and verified title is clickable target to the configured href.

### DIFF
--- a/tensorboard/components/tf_tensorboard/tf-tensorboard.html
+++ b/tensorboard/components/tf_tensorboard/tf-tensorboard.html
@@ -61,11 +61,16 @@ limitations under the License.
     <paper-header-panel>
       <paper-toolbar id="toolbar" slot="header" class="header">
         <div id="toolbar-content" slot="top">
-          <template is="dom-if" if="[[!homeHref]]">
+          <template is="dom-if" if="[[!_homePath]]">
             <div class="toolbar-title">[[brand]]</div>
           </template>
-          <template is="dom-if" if="[[homeHref]]">
-            <a href="[[homeHref]]" class="toolbar-title">[[brand]]</a>
+          <template is="dom-if" if="[[_homePath]]">
+            <a
+              href="[[_homePath]]"
+              rel="noopener noreferrer"
+              class="toolbar-title"
+              >[[brand]]</a
+            >
           </template>
           <template is="dom-if" if="[[_activeDashboardsNotLoaded]]">
             <span class="toolbar-message">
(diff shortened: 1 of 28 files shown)


tensorboard/components/tf_tensorboard/tf-tensorboard.html
```diff
@@ -61,7 +61,12 @@ <h2>Settings</h2>
     <paper-header-panel>
       <paper-toolbar id="toolbar" slot="header" class="header">
         <div id="toolbar-content" slot="top">
-          <div class="toolbar-title">[[brand]]</div>
+          <template is="dom-if" if="[[!homeHref]]">
+            <div class="toolbar-title">[[brand]]</div>
+          </template>
+          <template is="dom-if" if="[[homeHref]]">
+            <a href="[[homeHref]]" class="toolbar-title">[[brand]]</a>
```
stephanwlee: Especially given we do not add `rel="nooopener"`, we should sanitize the input to make sure it is relative. WDYT?

With the change, we probably should change `homeHref` to `homePath`.

bmd3k: I think only the admin will be able to set this value so I don't think we need to worry about malicious URLs.

I also don't think we need to restrict this value to relative paths.

Do you have an example scenario that concerns you?

stephanwlee: Technically, there are more ways to abuse it.

Although impractical:
- MITM (since TB is often hosted without HTTPS)
- Chrome Extension (admit that you can already can pwn so many websites when you are on extension)
- Any JavaScript running on page
  - `document.querySelector('tf-tensorboard').homeUrl = 'javascript:alert(...)'.`
  - note that you do not have resin on OSS
  - note that CSP header can be stripped out by random proxy

bmd3k: Thanks for helping me take care of this with your commits. If the sanitization ends up being too restrictive in the future we can always reconsider. In the meantime it should suit our needs.

bmd3k:
> It feels weird to LGTM my changes but are you comfortable as is? Would you like to make changes before I take a final look?

I'm comfortable as is.

PR #3258, opened 2020-02-18
profile: replace static plugin with dynamic plugin

wchargin:
Summary:
A working version of the dynamic profile plugin is live on stable PyPI!
It’s already a notable improvement over the static profile plugin,
because it includes a web components polyfill so that the trace viewer
actually works. Thus, we can now safely retire the legacy static plugin
and the trace viewer component (only used by the profile plugin).

Test Plan:
Launch TensorBoard in a fresh virtualenv with only `tf-nightly`,
pointing to a directory with profile data. Note that the “Profile”
plugin shows installation instructions rather than the old static
plugin. Copy and paste the installation instructions into a shell and
relaunch TensorBoard. Note that the dynamic plugin dashboard is now
available and active by default, and the installation instructions have
disappeared.

wchargin-branch: profile-dynamic-only


wchargin:
Reviewers: @caisq for general TensorBoard; @qiuminxu as profile team
representative in case there’s anything wrong with the approach here.

Google sync is currently broken for unrelated reasons, so I haven’t
tested it.

wchargin:
Okay, I’m open to holding off for a few days. I would prefer to merge
this earlier rather than later: if there are problems on nightly, we
want to know about those before a full release (that’s what nightly is
for, after all; it’s expected that there may be some churn). Deferring
to “just before” the 2.2 release puts TensorBoard at risk, as this is a
large change.

When is the ETA for `tensorboard-plugin-profile` to be released on
stable PyPI?

nfelt:
+1 to merging this - we should merge well before the release is due so that we can catch problems without the added deadline pressure.  The `tb-nightly` package is meant to be unstable and to contain changes that are not fully finished, so if the `tensorboard-plugin-profile` package is still in progress that's okay.

qiuminxu:
Sure, let me catch up with the status of dynamic plugin. We should have a stable version by the end of this week. Does that timeline sound Okay?

wchargin:
Yep, that sounds good. I’d be happy to merge this on Friday or even next
Monday if we really need to.

wchargin:
Updated to remove the underlying trace viewer component, too; end-to-end
tested with `tensorboard-plugin-profile==0.9.32`, which works great.

wchargin:
(Googlers: A test sync now works, as of <http://cl/296300225>.)

qiuminxu:
Hi William, can you check if this release works? Maybe also update the profile moved page to point to pip install -U tensorboard-plugin-profile, instead of from pypi test. Once this works, we can safely remove the static plugin.

https://pypi.org/project/tensorboard-plugin-profile/#history

Try tensorboard --logdir=gs://brain-demo for testing.

wchargin:
Looks good to me!

![Screenshot of the trace viewer in the dynamic profile plugin][ss]

[ss]: https://user-images.githubusercontent.com/4317806/75005974-13be8e80-5425-11ea-9145-b508757d6dfc.png

wchargin:
Updated to `pip install -U tensorboard_plugin_profile` in #3275.

adampl:
Hi, could someone reveal where is the source code of this plugin? The page on PyPI refers to a nonexistent file in this repo.

qiuminxu:
Please check out this repo: https://github.com/tensorflow/profiler

PR #3261, opened 2020-02-19
ng: make Polymer pagination configurable from Ng

stephanwlee:
This implements the pagination setting for Polymer based TensorBoard.
When changing the pagination setting, you can now change how many pages
of categories will be rendered on the Polymer side. Do note that
Angular based components will simply use selector to read the state from
the store instead.

TODO after the PR:
- stylize the settings dialog.

### DIFF
--- a/tensorboard/webapp/app_container.ng.html
+++ b/tensorboard/webapp/app_container.ng.html
@@ -19,3 +19,4 @@ limitations under the License.
 <plugins class="plugins"></plugins>
 <reloader></reloader>
 <hash-storage></hash-storage>
+<settings-polymer-interop></settings-polymer-interop>
--- a/tensorboard/webapp/reducer_config.ts
+++ b/tensorboard/webapp/reducer_config.ts
@@ -36,8 +36,7 @@ export function logger(reducer: ActionReducer<any>): ActionReducer<any> {
   };
 }
 
-export const metaReducers: MetaReducer<any>[] =
-  config.env === 'dev' ? [logger] : [];
+export const metaReducers: MetaReducer<any>[] = [logger];
 
 export const ROOT_REDUCERS = new InjectionToken<
   ActionReducerMap<State, Action>
--- a/tensorboard/webapp/reducer_config.ts
+++ b/tensorboard/webapp/reducer_config.ts
@@ -36,7 +36,8 @@ export function logger(reducer: ActionReducer<any>): ActionReducer<any> {
   };
 }
 
-export const metaReducers: MetaReducer<any>[] = [logger];
+export const metaReducers: MetaReducer<any>[] =
+  config.env === 'dev' ? [logger] : [];
 
 export const ROOT_REDUCERS = new InjectionToken<
   ActionReducerMap<State, Action>
--- a/tensorboard/webapp/settings/dialog_component.ts
+++ b/tensorboard/webapp/settings/dialog_component.ts
@@ -13,14 +13,28 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {Component, OnInit, OnDestroy} from '@angular/core';
-import {FormControl, Validators} from '@angular/forms';
+import {
+  FormControl,
+  Validators,
+  AbstractControl,
+  ValidatorFn,
+} from '@angular/forms';
 import {Store, select, createSelector} from '@ngrx/store';
 
 import {Subject} from 'rxjs';
 import {takeUntil, debounceTime, filter} from 'rxjs/operators';
 
-import {getReloadEnabled, getReloadPeriodInMs, State} from '../core/store';
-import {toggleReloadEnabled, changeReloadPeriod} from '../core/actions';
+import {
+  getReloadEnabled,
+  getReloadPeriodInMs,
+  State,
+  getPageSize,
+} from '../core/store';
+import {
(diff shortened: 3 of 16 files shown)


stephanwlee:
Do note that this is a duplicate of #2813. Because of folder restructure refactors, the change became nearly un-salvageable and history got quite messy.

tensorboard/webapp/settings/polymer_interop_test.ts
```diff
@@ -0,0 +1,84 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
psybuzz: nit: 2020

tensorboard/webapp/settings/polymer_interop_test.ts
```diff
@@ -0,0 +1,84 @@
...
+  let store: MockStore<State>;
+  let setLimitCalls: number[];
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      providers: [provideMockStore()],
+      declarations: [SettingsPolymerInteropContainer],
+    }).compileComponents();
+
+    store = TestBed.get(Store);
```
psybuzz: Prefer `TestBed.inject` for future Angular 9 compat?

tensorboard/webapp/settings/dialog_component.ts
```diff
@@ -59,18 +81,45 @@ const getReloadPeriodInSec = createSelector(
...
+          type="number"
+          placeholder="Pagination Limit"
+          [formControl]="paginationControl"
+        />
+      </mat-form-field>
+      <mat-error
+        *ngIf="
+          paginationControl.hasError('min') ||
+          paginationControl.hasError('required') ||
+          paginationControl.hasError('integer')
```
psybuzz: Does `paginationControl.invalid` suffice?

tensorboard/webapp/reducer_config.ts
```diff
@@ -36,8 +36,7 @@ export function logger(reducer: ActionReducer<any>): ActionReducer<any> {
   };
 }
 
-export const metaReducers: MetaReducer<any>[] =
-  config.env === 'dev' ? [logger] : [];
```
psybuzz: Seems unrelated to this PR, is making the logger enabled in prod a deliberate change?

tensorboard/webapp/settings/polymer_interop_container.ts
```diff
@@ -0,0 +1,69 @@
...
+  ngOnInit() {
+    this.getPageSize$
+      .pipe(
+        takeUntil(this.ngUnsubscribe),
+        distinctUntilChanged()
+      )
+      .subscribe((pageSize) => {
+        this.paginatedViewStore.setLimit(pageSize);
+      });
+  }
```
psybuzz: When would the `settings-polymer-interop` ever be destroyed?  Its lifetime should be the same as the entire app.  Can we drop ngOnInit/Destroy?

tensorboard/webapp/app_container.ng.html
```diff
@@ -19,3 +19,4 @@
 <plugins class="plugins"></plugins>
 <reloader></reloader>
 <hash-storage></hash-storage>
+<settings-polymer-interop></settings-polymer-interop>
```
psybuzz: Is there some scenario when TensorBoard should not have `settings-polymer-interop`?  My understanding was that elements made available in this HTML template allow for customization, e.g. custom header.

If hash-storage and settings-polymer-interop are actually required functionality, it would be nice not to have to update each variant.  Wdyt about making some `<core-features></core-features>` that internally appends required fields hash-storage (which already lives in core/), settings-polymer-interop?  (not necessarily in this CL)

tensorboard/webapp/settings/polymer_interop_container.ts
stephanwlee: It depends on the user. As a component, you should not make decisions and omit cleanup based on users.

tensorboard/webapp/app_container.ng.html
stephanwlee: I don't feel too strongly about it but I am inclined _not to_ create `core-features`. If there is one component that you do not want, you will end up needing to fork it and copy-pasta anyways.

tensorboard/webapp/reducer_config.ts
stephanwlee: Oops! Reverted!

tensorboard/webapp/settings/dialog_component.ts
stephanwlee: Yes, it does! Used.

tensorboard/webapp/settings/polymer_interop_container.ts
psybuzz: Makes sense.  (now I will try to remember to do this myself)

tensorboard/webapp/app_container.ng.html
psybuzz: I'm fine with landing as-is for now, since there are only a handful of elements in this template.

I still prefer fewer points of customization, since adding support for more (when there is a use case) is always in our control.

If this template had 50 elements that customized some TB feature, we must remember to maintain configurations of all variants.  That requires relying on memory, rather than process.

PR #3262, opened 2020-02-19
support meta labels shorter than the embedding (ie unlabelled embeddi…

hfiller:
  Tensorboard currently doesn't support partial metadata files (not all embeddings have labels) but this can happen a lot when working with subwords (eg. 
  Label "unlabelled" changes as such, with embedding number for differentiation.

### DIFF
--- a/tensorboard/plugins/projector/vz_projector/data.ts
+++ b/tensorboard/plugins/projector/vz_projector/data.ts
@@ -527,8 +527,8 @@ namespace vz_projector {
           .slice(0, TSNE_SAMPLE_SIZE)
           .map((index) =>
             this.points[index].metadata[superviseColumn] !== undefined
-              ? this.points[index].metadata[superviseColumn].toString()
-              : `unknown #${index}`
+              ? String(this.points[index].metadata[superviseColumn])
+              : `Unknown #${index}`
           );
       }
       if (superviseInput != null) {
--- a/tensorboard/plugins/projector/vz_projector/data.ts
+++ b/tensorboard/plugins/projector/vz_projector/data.ts
@@ -523,21 +523,13 @@ namespace vz_projector {
 
     setSupervision(superviseColumn: string, superviseInput?: string) {
       if (superviseColumn != null) {
-        let sampledIndices = this.shuffledDataIndices.slice(
-          0,
-          TSNE_SAMPLE_SIZE
-        );
-        let labels = new Array(sampledIndices.length);
-        sampledIndices.forEach((index, i) => {
-          let label = this.points[index].metadata[superviseColumn];
-          if (!label || !label.toString()) {
-            // if rows don't match up, backfill unknown values.
-            label = 'unknown #' + index.toString();
-            this.points[index].metadata[superviseColumn] = label;
-          }
-          labels[i] = label.toString();
-        });
-        this.superviseLabels = labels;
+        this.superviseLabels = this.shuffledDataIndices
+          .slice(0, TSNE_SAMPLE_SIZE)
+          .map((index) =>
+            this.points[index].metadata[superviseColumn] !== undefined
+              ? this.points[index].metadata[superviseColumn].toString()
+              : `unknown #${index}`
+          );
       }
       if (superviseInput != null) {
         this.superviseInput = superviseInput;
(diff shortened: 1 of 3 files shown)


tensorboard/plugins/projector/vz_projector/data.ts
```diff
@@ -528,12 +528,15 @@ namespace vz_projector {
...
-            ].toString())
-        );
+        sampledIndices.forEach((index, i) => {
+          let label = this.points[index].metadata[superviseColumn];
+          if (!label || !label.toString()) {
+            // if rows don't match up, backfill unknown values.
+            label = 'unknown #' + index.toString();
+            this.points[index].metadata[superviseColumn] = label;
+          }
+          labels[i] = label.toString();
```
stephanwlee: I am a bit unfamiliar with this code. Can you tell us why you need to cast it to string? What is a type of `label`? Here and 533.

tensorboard/plugins/projector/vz_projector/data.ts
```diff
@@ -528,12 +528,15 @@ namespace vz_projector {
...
-            (labels[i] = this.points[index].metadata[
-              superviseColumn
-            ].toString())
-        );
+        sampledIndices.forEach((index, i) => {
+          let label = this.points[index].metadata[superviseColumn];
+          if (!label || !label.toString()) {
+            // if rows don't match up, backfill unknown values.
+            label = 'unknown #' + index.toString();
+            this.points[index].metadata[superviseColumn] = label;
```
stephanwlee: We generally, for right reasons, avoid side effects like this. Do you need to mutate the object instance attached to this or can you make a copy then assign later?

tensorboard/plugins/projector/vz_projector/data.ts
```diff
@@ -528,12 +528,15 @@ namespace vz_projector {
...
-          (index, i) =>
-            (labels[i] = this.points[index].metadata[
-              superviseColumn
-            ].toString())
-        );
+        sampledIndices.forEach((index, i) => {
+          let label = this.points[index].metadata[superviseColumn];
+          if (!label || !label.toString()) {
+            // if rows don't match up, backfill unknown values.
+            label = 'unknown #' + index.toString();
```
stephanwlee: code style: template strings are cool. Use it.

tensorboard/plugins/projector/vz_projector/data.ts
```diff
@@ -528,12 +528,15 @@ namespace vz_projector {
           TSNE_SAMPLE_SIZE
         );
         let labels = new Array(sampledIndices.length);
-        sampledIndices.forEach(
-          (index, i) =>
-            (labels[i] = this.points[index].metadata[
-              superviseColumn
-            ].toString())
-        );
+        sampledIndices.forEach((index, i) => {
```
stephanwlee: Seems like this `forEach` is actually doing `map`. Can we change it?

tensorboard/plugins/projector/vz_projector/data.ts
hfiller: label is a [string|number], this casts it to a string.

tensorboard/plugins/projector/vz_projector/data.ts
hfiller: I re-wrote this section not to manipulate state, I am instead handling meta-data labels being null at the edges.

tensorboard/plugins/projector/vz_projector/data.ts
hfiller: This whole section was weird, re-wrote

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -311,8 +311,12 @@ namespace vz_projector {
     }
 
     private getLabelFromIndex(pointIndex: number): string {
-      const point = this.projector.dataSet.points[pointIndex];
-      return point.metadata[this.selectedMetadataField].toString();
+      const metadata = this.projector.dataSet.points[pointIndex].metadata[
+        this.selectedMetadataField
+      ];
+      return metadata !== undefined
+        ? metadata.toString()
```
stephanwlee: Kinda my preference but I think `String(metadata)` reads cleaner.

tensorboard/plugins/projector/vz_projector/vz-projector-inspector-panel.ts
```diff
@@ -311,8 +311,12 @@ namespace vz_projector {
...
+        : `unknown #${pointIndex}`;
```
stephanwlee: Stupid question, is there a reason why this is lower case (as opposed to one in PlotAdapter)? Is this good for UX?

Issue #86, opened 2017-06-16
Embedding Projector needs docs and demo code


decentralion:
Migrated from https://github.com/tensorflow/tensorflow/issues/6322

The demo should follow conventions from the other demo code, and show off all the features. It should also have a README.md that explains the demo code and how to use the frontend to explore the embedding projector that gets setup by the demo. (see histograms plugin for an example)

nfelt:
Just to catalog what is available:
- https://www.tensorflow.org/guide/embedding (old guide, doesn't specifically cover the TensorBoard `visualize_embeddings()` API)
- 2017 TF dev summit example code uses the projector API: https://github.com/martinwicke/tf-dev-summit-tensorboard-tutorial/blob/8c12cc0eac88db2239cbe453c9e828367021a515/mnist.py#L117-L124
- Crowdsourced info on StackOverflow: https://stackoverflow.com/questions/40849116/how-to-use-tensorboard-embedding-projector
- This third-party guide: http://www.pinchofintelligence.com/simple-introduction-to-tensorboard-embedding-visualisation/

pindinagesh:
@decentralion 

Could you please move this to closed status as the [PR](https://github.com/tensorflow/tensorboard/pull/3263) already merged. Thanks

pindinagesh:
@decentralion 

Closing this issue.  feel free to reopen if this still persists, and also refer this. Thanks

nfelt:
The PR added a demo, but we still need some actual documentation, so let's leave this open.

PR #3263, opened 2020-02-20
Add demo for projector plugin

hfiller:
A basic example of how to use the **projector plugin**. This follows the [word embeddings](https://www.tensorflow.org/tutorials/text/word_embeddings) tutorial, and demonstrating how to use the plugin while running tensorboard.

Addresses missing demo part of #86.

## Commit Message

Create a demo for using the Embedding Projector Plugin locally

This example demonstrates how to use the projector using up-to-date standards with the embedding projector.


tensorboard/examples/embedding_projector.py
```diff
@@ -0,0 +1,110 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+import tensorflow as tf
+from tensorboard.plugins import projector
+import tensorflow_datasets as tfds
+
+# This example extends on the word embeddings tutorial found
```
nfelt: Conventionally we actually call these things "demos" and put them under the plugin directory, e.g. tensorboard/plugins/scalar/scalars_demo.py - I think that's probably best for consistency.

So this would go at tensorboard/plugins/projector/projector_demo.py.

Also, for the output directory, other demos use `/tmp/<plugin-name>_demo` so in this case that would be `/tmp/projector_demo` for consistency - that way it's easy to lump all the demo data together into a single directory that showcases multiple plugins.

tensorboard/examples/embedding_projector.py
```diff
@@ -0,0 +1,110 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+import tensorflow as tf
```
nfelt: Does this require a particular version of TF (e.g. 1.x vs 2.x)?  Might be worth specifying

tensorboard/examples/embedding_projector.py
```diff
@@ -0,0 +1,110 @@
...
+from tensorboard.plugins import projector
+import tensorflow_datasets as tfds
+
+# This example extends on the word embeddings tutorial found
+# here: https://www.tensorflow.org/tutorials/text/word_embeddings)
+# and is intended to demonstrate the use of the embedding projector.
+
+LOG_DIR = "/tmp/logs/model"  # Tensorboard log dir
+META_DATA_FNAME = "meta.tsv"  # Labels will be stored here
```
nfelt: nit: METADATA usually as one word?

tensorboard/examples/embedding_projector.py
hfiller: Yeah, we use keras, which was implemented in 2.0

tensorboard/examples/embedding_projector.py
```diff
@@ -0,0 +1,110 @@
+# Copyright 2016 The TensorFlow Authors. All Rights Reserved.
```
manivaradarajan: Update to Copyright 2020?

PR #3265, opened 2020-02-20
scalars: promote generic data to stable

wchargin:
Summary:
At this point, the generic data APIs are a fully featured replacement
for the multiplexer APIs, including `is_active` support. We now enable
them by default. They can still be disabled with `--generic_data false`.

Test Plan:
Patch the `MultiplexerDataProvider` to return hard-coded descriptions so
that the generic and non-generic flows are distinguishable, then note
that `tensorboard` uses the generic data APIs with `--generic_data`
either set to `true` or unspecified, and not when set to `false`.

wchargin-branch: data-stabilize-scalars

### DIFF
--- a/tensorboard/plugins/scalar/scalars_plugin.py
+++ b/tensorboard/plugins/scalar/scalars_plugin.py
@@ -60,7 +60,7 @@ class ScalarsPlugin(base_plugin.TBPlugin):
         """
         self._multiplexer = context.multiplexer
         self._db_connection_provider = context.db_connection_provider
-        if context.flags and context.flags.generic_data == "true":
+        if context.flags and context.flags.generic_data != "false":
             self._data_provider = context.data_provider
         else:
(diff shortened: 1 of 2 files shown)


wchargin:
Release plan (comment if you object):

  - TensorBoard 2.2: Release as is (scalars on by default, all others
    off by default).
  - TensorBoard 2.3: Remove the legacy path for scalars, and turn
    everything else on by default.
  - TensorBoard 2.4: Remove all the other legacy paths.

Removing the legacy paths is nice because it enables us to clean up the
tests, as well as reducing the complexity of the actual code. We should
probably go ahead and remove DB mode at some point in here, too.

wchargin:
> TensorBoard 2.3: Remove the legacy path for scalars, and turn
> everything else on by default.

Should qualify that “everything else” here excludes graphs, which are
not yet at feature parity.

PR #3266, opened 2020-02-20
text: add data provider support

wchargin:
Summary:
This patch teaches the text plugin about the generic data APIs using
text tensors. This has the implication that text tensors are required to
be not too large in size, which should be fine: the complete works of
Shakespeare fit in 5.5 MB, and text summaries should be expected to be
far smaller than that.

Test Plan:
Change `TensorBoardWSGIApp` to set `multiplexer=None` on the context,
and note that `--generic_data true` still has a working text plugin,
while `--generic_data false` is completely broken, so everything is
wired up through the new APIs.

wchargin-branch: generic-text


tensorboard/plugins/text/text_plugin.py
```diff
@@ -230,33 +236,59 @@ def is_active(self):
...
     def tags_route(self, request):
-        index = self.index_impl()
+        experiment = plugin_util.experiment_id(request.environ)
+        index = self.index_impl(experiment)
         return http_util.Respond(request, index, "application/json")
 
-    def text_impl(self, run, tag):
+    def text_impl(self, run, tag, experiment):
+        if self._data_provider:
+            # Downsample reads to 10 text samples per time series, which
```
nfelt: I wonder if we should make this higher?  Personally 10 has always seemed pretty low for text samples (which are generally going to not get as big as images or audio).  The multiplexer DataProvider would still of course max out at 10 samples but then at least other ones wouldn't be subject to the same constraint.

Also, hardcoding this to 10 means that the `--samples_per_plugin` flag would no longer provide an escape hatch once the DataProvider path is on by default.

In fairness, I think this is a specific instance of a more general issue which is that random sampling to a fixed _k_ steps is kind of a weird approach to use in general for the data that isn't being rendered as the entire time series at once (like scalars or histograms/distributions).  The limits we have today are more about avoiding too much memory usage in the multiplexer than really about what makes sense in terms of user experience.

But it seems like in the short term it would at least be better to make this respect `--samples_per_plugin` rather than hardcoding the default value.

tensorboard/plugins/text/text_plugin.py
```diff
@@ -189,13 +190,12 @@ def text_array_to_html(text_arr):
     return warning + make_table(html_arr)
 
 
-def process_string_tensor_event(event):
+def process_event(wall_time, step, string_ndarray):
     """Convert a TensorEvent into a JSON-compatible response."""
```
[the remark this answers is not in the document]
wchargin: Yes; thanks. Will do.

tensorboard/plugins/text/text_plugin.py
wchargin: Hmm; good point. I’m happy to increase it. I hadn’t considered the
(non-)effect of the `--samples_per_plugin` flag on the generic data
path. I’ll see if I can come up with a patch for that tomorrow, as it
makes #3265 technically a minor regression.

tensorboard/plugins/text/text_plugin.py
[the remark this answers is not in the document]
wchargin: Done.

tensorboard/plugins/text/text_plugin.py
wchargin: Went ahead and upped to 100 (why not). Will make the change to respect
`--samples_per_plugin` separately, as it touches all the plugins.

wchargin: Wait, no, it’s not technically a regression, because the multiplexer
data provider ignores `downsample` entirely, so `samples_per_plugin` in
fact still behaves how it always did. :-)

I’ll fix it the right way, anyway.

PR #3268, opened 2020-02-20
Refactor Uploader to use Generic Data, and upload blobs.

davidsoergel:
Note blobs will not actually be uploaded because of this PR alone; that must be enabled on the server side via the allowed_plugins mechanism.


tensorboard/uploader/uploader.py
```diff
@@ -224,52 +236,77 @@ def send_requests(self, run_to_events):
...
-                metadata.plugin_data.plugin_name == scalar_metadata.PLUGIN_NAME
-            ):
-                self._scalar_request_sender.add_event(
-                    run_name, event, value, metadata
-                )
-            # TODO(nielsene): add Tensor plugin cases here
-            # TODO(soergel): add Graphs blob case here
+        for (run_name, event) in self._run_events(run_to_events):
+            if event.graph_def or event.meta_graph_def:
+                self._send_graph(run_name, event)
```
wchargin: This should be using `dataclass_compat.migrate_event` rather than poking
into the event directly. The whole point of `dataclass_compat` is to
enable the upload flow to work in the same way as the local multiplexer
flow. In particular, that will set the correct tag name here, which is
required for the frontend to request the correct graph.

davidsoergel: Ah, right, thanks for unconfusing me.  Doing that required rejiggering things to be entirely generic-data-flavored, which was a) necessary anyway and b) surprisingly straightforward :)

tensorboard/uploader/uploader.py
```diff
@@ -445,19 +485,146 @@ def _create_point(self, tag_proto, event, value):
...
+        """Sends the current blob sequence fully, and clears it to make way for the next.
+        """
+        if self._value:
+            blob_sequence_id = self._get_or_create_blob_sequence()
+            print("Flushing to blob sequence id: ", blob_sequence_id)
+
+            # TODO(soergel): should we really unpack the tensor here, or ship
+            # it wholesale and unpack server side, or something else?
+            blobs = tensor_util.make_ndarray(self._value.tensor)
+            for seq_index, blob in np.ndenumerate(blobs):
```
wchargin: Blob sequences are always rank-1 string tensors, so:

  - the `ndenumerate`/`seq_index[0]` can just be `enumerate`/`idx`; and
  - if you like, you can avoid `make_ndarray`ing by directly
    dereferencing `string_val` on the `TensorProto` (unlike tensors with
    fixed-size dtypes, string tensors can only be represented with
    `string_val`, not `tensor_content`), which should save potentially a
    lot of memcpying, though the change is a wee bit sketchy due to
    poking into the proto internals;
  - you should probably validate this constraint (anything *transformed*
    by `dataclass_compat` is guaranteed to satisfy it, but users could
    write bad data directly into their logdir) before projecting down
    into a lower dimension.

tensorboard/uploader/uploader.py
```diff
@@ -445,19 +485,146 @@ def _create_point(self, tag_proto, event, value):
...
+                response = grpc_util.call_with_retries(
+                    self._api.GetOrCreateBlobSequence, request
+                )
+                blob_sequence_id = response.blob_sequence_id
+            except grpc.RpcError as e:
+                if e.code() == grpc.StatusCode.NOT_FOUND:
+                    raise ExperimentNotFoundError()
+                logger.error("Upload call failed with error %s", e)
+                # TODO(soergel): clean up
+                raise e
```
wchargin: Just `raise`—this preserves the original stacktrace.

tensorboard/uploader/uploader.py
```diff
@@ -224,52 +235,81 @@ def send_requests(self, run_to_events):
...
+        # Note that this join in principle has deletion anomalies: if the input
+        # stream contains runs with no events, we'll lose that information. This
+        # is not a problem: we would need to prune such data from the request
+        # anyway.
         for (run_name, events) in six.iteritems(run_to_events):
             for event in events:
-                for value in event.summary.value:
-                    yield (run_name, event, value)
+                v2_event = data_compat.migrate_event(event)
+                dataclass_events = dataclass_compat.migrate_event(v2_event)
```
wchargin: Note that this *is* a behavior change for existing flows. Once this is
released, metadata for scalar time series will have the `data_class`
field set where previously it did not. This is fine, but tightens that
data compatibility constraint, so we should release it intentionally.

tensorboard/uploader/uploader.py
```diff
@@ -445,19 +485,146 @@ def _create_point(self, tag_proto, event, value):
...
+
+        self._new_request()
+
+    def _get_or_create_blob_sequence(self):
+        request = write_service_pb2.GetOrCreateBlobSequenceRequest(
+            experiment_id=self._experiment_id,
+            run=self._run_name,
+            tag=self._value.tag,
+            step=self._event.step,
+            final_sequence_length=1,
```
wchargin: Final sequence length should not always be 1, right? cf. `_tensor_size`
in `event_processing/data_provider.py`.

tensorboard/uploader/uploader.py
```diff
@@ -445,19 +485,146 @@ def _create_point(self, tag_proto, event, value):
...
+        # refactoring here.
+        for offset in range(0, len(blob), BLOB_CHUNK_SIZE):
+            chunk = blob[offset : offset + BLOB_CHUNK_SIZE]
+            finalize_object = offset + BLOB_CHUNK_SIZE >= len(blob)
+            request = write_service_pb2.WriteBlobRequest(
+                blob_sequence_id=blob_sequence_id,
+                index=seq_index,
+                data=chunk,
+                offset=offset,
+                crc32c=None,
```
wchargin: Are these TODOs, DO NOT SUBMITs, or nothings?

tensorboard/uploader/uploader.py
davidsoergel: I agree that here we set data_class where previously we didn't.  I'm not sure I follow your concern, but I think it is that server-side we have to make sure things work whether the field is set or not.  Indeed, in the associated internal changes, we do fall back to scalars when data_class is empty, but that is worth doublechecking in tests and reviews.

tensorboard/uploader/uploader.py
```diff
@@ -445,19 +485,146 @@ def _create_point(self, tag_proto, event, value):
...
+        self._value = value
+        self._metadata = metadata
+        self.flush()
+
+    def flush(self):
+        """Sends the current blob sequence fully, and clears it to make way for the next.
+        """
+        if self._value:
+            blob_sequence_id = self._get_or_create_blob_sequence()
+            print("Flushing to blob sequence id: ", blob_sequence_id)
```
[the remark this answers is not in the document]
davidsoergel: Done

tensorboard/uploader/uploader.py
davidsoergel: Added TODO.

tensorboard/uploader/uploader.py
davidsoergel: Fixed

tensorboard/uploader/uploader.py
davidsoergel: Done

tensorboard/uploader/uploader.py
```diff
@@ -224,52 +235,81 @@ def send_requests(self, run_to_events):
...
+
+        # The metadata for a time series is memorized on the first event.
+        # If later events arrive with a mismatching plugin_name, they are
+        # ignored with a warning.
+        metadata = self._tag_metadata.get(time_series_key)
+        if metadata is None:
+            metadata = value.metadata
+            self._tag_metadata[time_series_key] = metadata
+
+        # TODO(soergel): Allow enabling/disabling upload per plugin
```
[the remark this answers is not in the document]
davidsoergel: Yep, definitely.  Done.

tensorboard/uploader/uploader.py
```diff
@@ -224,52 +235,81 @@ def send_requests(self, run_to_events):
...
+    def _run_events(self, run_to_events):
+        """Helper generator to create a single stream of work items.
+
+        The events are passed through the `data_compat` and `dataclass_compat`
+        layers before being emitted, so downstream consumers may process them
+        uniformly.  Note that `dataclass_compat` may emit multiple variants of
+        the same event, for backwards compatibility.  Thus this stream should
+        be filtered to obtain the desired version of each event.  Here, this
+        happens below in `_summary_values(...)`, which ignores any event that
+        does not have a `summary` field.
```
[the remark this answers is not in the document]
davidsoergel: Done.

tensorboard/uploader/uploader.py
```diff
@@ -224,52 +235,81 @@ def send_requests(self, run_to_events):
...
         for (run_name, events) in six.iteritems(run_to_events):
             for event in events:
-                for value in event.summary.value:
-                    yield (run_name, event, value)
+                v2_event = data_compat.migrate_event(event)
+                dataclass_events = dataclass_compat.migrate_event(v2_event)
+                for dataclass_event in dataclass_events:
+                    yield (run_name, dataclass_event)
+
+    def _summary_values(self, run_name, event):
```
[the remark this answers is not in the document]
davidsoergel: Done.

wchargin:
Just checking: this PR is not marked for re-review. Is that correct?

tensorboard/uploader/uploader.py
wchargin: The concern/observation was that this is the first place that a proto
with `data_class` set is actually written to any persistent store, which
reduces our flexibility to change the structure or interpretation (but
not to zero, because we’re still only writing it to TensorBoard.dev, not
users’ local disks). No action required.

tensorboard/uploader/uploader_test.py
```diff
@@ -228,6 +236,138 @@ def scalar_event(tag, value):
...
+        uploader.create_experiment()
+
+        def graph_event(tag, value):
+            return event_pb2.Event(graph_def=value)
+
+        mock_logdir_loader = mock.create_autospec(logdir_loader.LogdirLoader)
+        mock_logdir_loader.get_run_events.side_effect = [
+            {
+                "run 1": [
+                    graph_event("1.1", bytes(950)),
```
wchargin: Please mark `srcs_version = "PY3"` for this test; in Python 2, this
evaluates to `"950"` rather than `b"\x00" * 950`.

tensorboard/uploader/uploader_test.py
```diff
@@ -228,6 +236,138 @@ def scalar_event(tag, value):
...
+            mock_client,
+            "/logs/foo",
+            rpc_rate_limiter=mock_rate_limiter,
+            blob_rpc_rate_limiter=mock_blob_rate_limiter,
+            allowed_plugins=["scalars", "graphs"],
+        )
+        uploader.create_experiment()
+
+        def graph_event(tag, value):
+            return event_pb2.Event(graph_def=value)
```
wchargin: These `tag`s are unused; drop?

(In fact, all your call sites are the same, so you could optionally drop
both args if you want.)

tensorboard/uploader/uploader_test.py
```diff
@@ -228,6 +236,138 @@ def scalar_event(tag, value):
...
+    def test_start_uploading_graphs(self):
+        mock_client = _create_mock_client()
+        mock_rate_limiter = mock.create_autospec(util.RateLimiter)
+        mock_blob_rate_limiter = mock.create_autospec(util.RateLimiter)
+        uploader = _create_uploader(
+            mock_client,
+            "/logs/foo",
+            rpc_rate_limiter=mock_rate_limiter,
+            blob_rpc_rate_limiter=mock_blob_rate_limiter,
+            allowed_plugins=["scalars", "graphs"],
```
wchargin: Since you’re relying on `graphs` being the actual plugin name, not just
an arbitrary string, could you please use `graphs_metadata.PLUGIN_NAME`?

tensorboard/uploader/uploader_test.py
```diff
@@ -228,6 +236,138 @@ def scalar_event(tag, value):
...
+            {"run 1": [graph_event("1.2", bytes(950))],},
+            AbortUploadError,
+        ]
+
+        mock_client.WriteBlob.side_effect = [
+            [write_service_pb2.WriteBlobResponse()],
+            test_util.grpc_error(grpc.StatusCode.INTERNAL, "nope"),
+        ]
+
+        # This demonstrates that the ALREADY_EXISTS error is handled
```
wchargin: Fix comment? This comment applies to the test case below, but not to
this test case.

tensorboard/plugins/scalar/metadata.py
```diff
@@ -41,6 +41,7 @@ def create_summary_metadata(display_name, description):
     metadata = summary_pb2.SummaryMetadata(
         display_name=display_name,
         summary_description=description,
+        data_class=summary_pb2.DATA_CLASS_SCALAR,
```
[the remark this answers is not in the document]
davidsoergel: Huh, weird.  Maybe part of an early experiment from before I understood the `dataclass_compat` situation.  Reverted.

tensorboard/uploader/uploader_test.py
davidsoergel: Done.  I left some preexisting instances of `"scalars"` where it doesn't actually matter.  I think `"foobar"` would be clearer that those cases allow arbitrary strings, but this PR isn't the place for that.

tensorboard/uploader/uploader_test.py
davidsoergel: One better: there's no need for a function at all.

tensorboard/uploader/uploader_test.py
davidsoergel: Done

tensorboard/uploader/uploader_test.py
davidsoergel: Oops, copypasta, fixed

tensorboard/uploader/uploader.py
```diff
@@ -541,19 +604,182 @@ def _create_point(self, tag_proto, event, value):
...
+
+    def _send_blob(self, blob_sequence_id, seq_index, blob):
+        # TODO(soergel): retry and resume logic
+
+        request_iterator = self._write_blob_request_iterator(
+            blob_sequence_id, seq_index, blob
+        )
+        upload_start_time = time.time()
+        count = 0
+        print("Uploading blob", end="", flush=True)
```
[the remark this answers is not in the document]
davidsoergel: Removed for now, but see b/150723481 (and please discuss priority with the team in my absence)

tensorboard/uploader/uploader.py
```diff
@@ -541,19 +604,182 @@ def _create_point(self, tag_proto, event, value):
...
+        self._value = value
+        # TODO(soergel): should we really unpack the tensor here, or ship
+        # it wholesale and unpack server side, or something else?
+        # TODO(soergel): can we extract the proto fields directly instead?
+        self._blobs = tensor_util.make_ndarray(self._value.tensor)
+        if np.ndim(self._blobs) != 1:
+            raise ValueError(
+                "A blob sequence must be represented as a rank-1 Tensor"
+            )
+        self._metadata = metadata
```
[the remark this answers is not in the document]
wchargin: AFAICT, `_blobs` isn’t initialized in either the initializer proper or
its `self._new_request()` call?

(Initializing in `_new_request` solves the confusing `AttributeError`
problem, of course, though I still think it’s a bit more readable to
explicitly initialize in the initializer, because our brains are trained
to scan that method for fields.)

Issue #3260, opened 2020-02-19
io_wrapper_test failing at head


wchargin:
Failing with 2020-02-19 TF nightly on Python 3, current master:

```
$ python -c 'print(__import__("tensorflow").__version__)'
2.2.0-dev20200218
$ python --version
Python 3.7.5rc1
$ git status && git rev-parse --verify HEAD
HEAD detached at origin/master
nothing to commit, working tree clean
7ea2af7ae0f1539dbc1041a2f73a1aeb74803b2c
$ bazel test //tensorboard/backend/event_processing:io_wrapper_test --test_output=errors
INFO: Analyzed target //tensorboard/backend/event_processing:io_wrapper_test (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
FAIL: //tensorboard/backend/event_processing:io_wrapper_test (see /HOMEDIR/.cache/bazel/_bazel_wchargin/52a95bbdd50941251730eb33b7476a66/execroot/org_tensorflow_tensorboard/bazel-out/k8-fastbuild/testlogs/tensorboard/backend/event_processing/io_wrapper_test/test.log)
INFO: From Testing //tensorboard/backend/event_processing:io_wrapper_test:
==================== Test output for //tensorboard/backend/event_processing:io_wrapper_test:
WARNING:root:Limited tf.compat.v2.summary API due to missing TensorBoard installation.
WARNING:root:Limited tf.compat.v2.summary API due to missing TensorBoard installation.
WARNING:root:Limited tf.compat.v2.summary API due to missing TensorBoard installation.
WARNING:root:Limited tf.summary API due to missing TensorBoard installation.
Running tests under Python 3.7.5: /HOMEDIR/virtualenv/tf-nightly-20200219-py3.7/bin/python
[ RUN      ] IoWrapperTest.testGetLogdirSubdirectories
I0219 18:54:26.423715 140524062205760 io_wrapper.py:227] GetLogdirSubdirectories: Starting to list directories via walking.
[       OK ] IoWrapperTest.testGetLogdirSubdirectories
[ RUN      ] IoWrapperTest.testIsCloudPathCnsIsTrue
[       OK ] IoWrapperTest.testIsCloudPathCnsIsTrue
[ RUN      ] IoWrapperTest.testIsCloudPathFileIsFalse
[       OK ] IoWrapperTest.testIsCloudPathFileIsFalse
[ RUN      ] IoWrapperTest.testIsCloudPathGcsIsTrue
[...]

PR #3270, opened 2020-02-20
Disable io_wrapper glob test

psybuzz:
According to https://github.com/tensorflow/tensorboard/issues/3260, `testListRecursivelyViaGlobbingForPathWithGlobCharacters` is failing due to a breaking change in tf-nightly's `tf.io.gfile.glob()` implementation.  While that is being resolved, this disables the test to avoid interference with normal TB development.

Renames the test to `skipTestListRecursivelyViaGlobbingForPathWithGlobCharacters`

Ran bazel test on the io_wrapper_target and saw that it passes when it failed before.

### DIFF
--- a/tensorboard/backend/event_processing/io_wrapper_test.py
+++ b/tensorboard/backend/event_processing/io_wrapper_test.py
@@ -147,7 +147,10 @@ class IoWrapperTest(tf.test.TestCase):
             expected, io_wrapper.ListRecursivelyViaGlobbing(temp_dir)
         )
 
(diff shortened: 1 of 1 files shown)


wchargin:
Sorry, I read the comment in the diff but not the code change closely
enough—for future reference, it’s nice to use `@unittest.skip()` or
`self.skipTest("...")` rather than just renaming the method, because
then it shows up as a “skipped” test in the report rather than just
being an unused method. But no big deal/no action required given that we
expect this to be re-enabled promptly.

psybuzz:
Ah, thanks, I'll make a note for future me!

PR #3271, opened 2020-02-20
data: expose downsampling preferences to plugins

wchargin:
Summary:
We add a `sampling_hints` attribute to the `TBContext` magic container,
which is populated with the parsed form of the `--samples_per_plugin`
flag. Existing plugins’ generic data modes are updated to read from this
map instead of using hard-coded thresholds.

Test Plan:
This change is not actually observable as is, because the multiplexer
data provider ignores its downsampling argument. But after patching in a
change to make the data provider respect the downsampling argument, this
change has the effect that increasing the `--samples_per_plugin` over
the default (e.g., `images=20`) now properly increases the number of
samples shown in generic data mode, whereas previously it had no effect.

wchargin-branch: data-downsampling-flag

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -97,17 +97,20 @@ _VALID_PLUGIN_RE = re.compile(r"^[A-Za-z0-9_-]+$")
 logger = tb_logging.get_logger()
 
 
-def tensor_size_guidance_from_flags(flags):
-    """Apply user per-summary size guidance overrides."""
-
-    tensor_size_guidance = dict(DEFAULT_TENSOR_SIZE_GUIDANCE)
+def _parse_samples_per_plugin(flags):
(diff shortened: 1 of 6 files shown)


tensorboard/backend/application.py
```diff
@@ -238,6 +242,7 @@ def TensorBoardWSGIApp(
         multiplexer=deprecated_multiplexer,
         assets_zip_provider=assets_zip_provider,
         plugin_name_to_instance=plugin_name_to_instance,
+        sampling_hints=_parse_samples_per_plugin(flags),
```
wchargin: Parsing this twice felt easier than adding an extra parameter to this
method, but I’d be happy to change that if others object.

wchargin:
(Reassigning just to spread load.)

PR #3272, opened 2020-02-20
data: perform downsampling in multiplexer provider

wchargin:
Summary:
The `MultiplexerDataProvider` now respects its `downsample` parameter,
even though the backing `PluginEventMultiplexer` already performs its
own sampling. This serves two purposes:

  - It enforces that clients are always specifying the `downsample`
    argument, which is required.
  - It enables us to test plugins’ downsampling parameters to verify
    that they will behave correctly with other data providers.

Test Plan:
Unit tests included. Note that changing the `_DEFAULT_DOWNSAMPLING`
constant in (e.g.) the scalars plugin to a small number (like `5`) now
actually causes charts in the frontend to be downsampled.

wchargin-branch: data-mux-downsample

### DIFF
--- a/tensorboard/backend/event_processing/data_provider_test.py
+++ b/tensorboard/backend/event_processing/data_provider_test.py
@@ -249,7 +249,7 @@ class MultiplexerDataProviderTest(tf.test.TestCase):
             experiment_id="unused",
             plugin_name=scalar_metadata.PLUGIN_NAME,
             run_tag_filter=run_tag_filter,
-            downsample=None,  # not yet implemented
+            downsample=100,
         )
 
         self.assertItemsEqual(result.keys(), ["polynomials", "waves"])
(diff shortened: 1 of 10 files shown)


wchargin:
(Reassigning just to spread load.)

tensorboard/backend/event_processing/data_provider_test.py
```diff
@@ -346,6 +359,46 @@ def test_read_tensors(self):
...
+    def test_underlong_ok(self):
+        xs = list("abcdefg")
+        actual = data_provider._downsample(xs, k=10)
+        expected = list("abcdefg")
+        self.assertIsNot(actual, xs)
+        self.assertEqual(actual, expected)
+
+    def test_inorder(self):
+        xs = list(range(10000))
+        actual = data_provider._downsample(xs, k=100)
```
stephanwlee: No AI. Simply noting: there is non-zero chance where this passes even if it behaved more like `random.sample`. :\

wchargin: Yep, but that chance is `1 / (100!)`, which is somewhat small.

PR #3275, opened 2020-02-21
profile: remove Test PyPI index specification

wchargin:
Summary:
The dynamic profile plugin is now available on stable PyPI:
<https://pypi.org/project/tensorboard-plugin-profile/2.2.0a1/>

Styling updated because the number of lines of text changed.

Test Plan:
The instructions still work in a new virtualenv, pulling in `2.2.0a1`.

wchargin-branch: profile-stable-pypi


wchargin:
> I think this is also in #3258 ?

Hmm; I don’t think so? The last change to that PR was pushed 4 hours
before the dynamic profile package was released on stable PyPI, and
I don’t see this HTML file changed in the diff of #3258.

In any case, it’s supposed to be in this PR and not that one.

Issue #3279, opened 2020-02-21
Scalar cards not loading in Hparams, Custom Scalars.


psybuzz:
When enabling generic data by default for Scalars in https://github.com/tensorflow/tensorboard/commit/e3f0f05641a6a21a5c570e2ee0027e82990fa23c, clients who called `scalars_impl` and failed to pass an experiment ID broke.  `scalars_impl` only validates experiment IDs in the DataProvider case, so when Hparams and Custom Scalars passed `None`, it failed on the backend.

We should provide the correct experiment IDs from Hparams and Custom Scalars in these cases.

![image](https://user-images.githubusercontent.com/2322480/75074661-00f19b80-54b1-11ea-8aa9-61c80d3c2d08.png)

![image](https://user-images.githubusercontent.com/2322480/75074684-0d75f400-54b1-11ea-8f7a-1ac27d16dc26.png)

PR #3280, opened 2020-02-21
Always pass experiment IDs to scalars_impl

psybuzz:
Hparams and Custom Scalars use the Scalars plugin's implementation to
generate data on the frontend.  When --generic_data was enabled by
default, the DataProvider started validating experiment IDs, so
callers of scalars_impl who passed experiment `None` would break.

This fixes https://github.com/tensorflow/tensorboard/issues/3279
by making Hparams, Custom Scalars pass the correct experiment ID.


wchargin:
(Instead of “addresses #3279”, you can write “fixes #3279” to auto-link
the issues.)

PR #3284, opened 2020-02-22
data: require kwargs for time series initializers

wchargin:
Summary:
This enables us to more easily add and change fields on these objects
without worrying about positional parameters. (Such changes are still
breaking, but these APIs have no stability guarantee.)

Test Plan:
All open-source consumers are already compliant except for tests, which
are updated in this commit. Google-internal clients have already been
updated; a test sync shows that all tests pass.

wchargin-branch: time-series-kwargs

### DIFF
--- a/tensorboard/data/provider.py
+++ b/tensorboard/data/provider.py
@@ -421,7 +421,13 @@ class _TimeSeries(object):
     )
 
     def __init__(
-        self, max_step, max_wall_time, plugin_content, description, display_name
+        self,
+        *,
+        max_step,
+        max_wall_time,
+        plugin_content,
+        description,
+        display_name,
     ):
         self._max_step = max_step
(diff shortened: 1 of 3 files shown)


wchargin:
Do not merge until <http://cl/296507451> (Google-internal) is submitted.

tensorboard/data/provider.py
```diff
@@ -421,7 +421,13 @@ class _TimeSeries(object):
     )
 
     def __init__(
-        self, max_step, max_wall_time, plugin_content, description, display_name
+        self,
```
caisq: In the doc string of this class, please add that this class is also a super class of `TensorTimeSeries`.

wchargin: Thanks for the catch. That’s unrelated to this change, so I’ve done so
in #3290.

PR #3269, opened 2020-02-20
[DebuggerV2] Add click callback to alert type; highlight alerts in timeline

caisq:
* Motivation for features / changes
  * Continue developing the Debugger V2 plugin

* Technical description of changes
  * In the serving route `/alerts`, support filtering by alert type.
  * When an alert type (e.g., InfNanAlert) in the `AlertComponent`'s breakdown
    list is clicked, fetch the detailed alert data.
  * Use the alert data and the execution index associated with each alert to
    highlight the alert-generating executions in the `TimelineComponent`
  * When alert data is loaded from the data source, scroll onto the execution
    digest that corresponds to the first alert in the `TimelineComponent`.
  * Updated the effects diagram in `debugger_effects.ts`.

* Screenshot(s) of UI changes
  * image
  * image

* Detailed steps to verify changes work correctly (as executed by you)
  * Python unit test added for the new filter mode of the `/alerts` serving route.
  * TypeScript/Angular Unit tests added for the new selectors, reducers, effects and
    component/container rendering logic.
  * Manually verified the correct functioning of new functionalities by using
    actual tfdbg2-format debug data consistent of InfNanAlerts.

PR #3285, opened 2020-02-22
[DebuggerV2] Add support for alert-type filtering to /alerts route

caisq:
  * Forms a basis for the already open https://github.com/tensorflow/tensorboard/pull/3269
  * Allow the DebuggerV2 frontend to query alerts that belong to a specific alert type.

  * Modify the multiplexer, data provider and serving method to support the "alert_type" parameter in requests.
  * Not that there is a slight change to requests to the same route without filters: previously there "alert_type" field was not populated; now it's populated with a `None` (`null`) value.

  * Unit tests added.

### DIFF
--- a/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
@@ -242,13 +242,9 @@ class DebuggerV2EventMultiplexer(object):
                     "Filtering of alerts failed: alert type %s does not exist"
                     % alert_type_filter
                 )
-            end = self._checkBeginEndIndices(
-                begin, end, alerts_breakdown[alert_type_filter]
-            )
-            alert_list = alerts_by_type[alert_type_filter][begin:end]
-        else:
-            end = self._checkBeginEndIndices(begin, end, len(alerts))
-            alert_list = alerts[begin:end]
+            alerts = alerts_by_type[alert_type_filter]
+        end = self._checkBeginEndIndices(begin, end, alerts)
+        alerts = alerts[begin:end]
         return {
             "begin": begin,
             "end": end,
@@ -256,7 +252,7 @@ class DebuggerV2EventMultiplexer(object):
             "num_alerts": len(alerts),
             "alerts_breakdown": alerts_breakdown,
             "per_type_alert_limit": DEFAULT_PER_TYPE_ALERT_LIMIT,
-            "alerts": [_alert_to_json(alert) for alert in alert_list],
+            "alerts": [_alert_to_json(alert) for alert in alerts],
         }
 
     def ExecutionDigests(self, run, begin, end):
--- a/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
@@ -235,7 +235,7 @@ class DebuggerV2EventMultiplexer(object):
                 alert_type = "__MiscellaneousAlert__"
             alerts_breakdown[alert_type] = len(monitor_alerts)
             alerts_by_type[alert_type] = monitor_alerts
-
+        num_alerts = len(alerts)
         if alert_type_filter is not None:
             if alert_type_filter not in alerts_breakdown:
                 raise errors.InvalidArgumentError(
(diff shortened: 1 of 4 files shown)


tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -229,15 +234,29 @@ def Alerts(self, run, begin, end):
...
+
+        if alert_type_filter is not None:
+            if alert_type_filter not in alerts_breakdown:
+                raise errors.InvalidArgumentError(
+                    "Filtering of alerts failed: alert type %s does not exist"
+                    % alert_type_filter
+                )
+            end = self._checkBeginEndIndices(
+                begin, end, alerts_breakdown[alert_type_filter]
+            )
```
wchargin: Can you deduplicate this by checking against `alert_list` after the
`if`-statement?

tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -43,41 +43,51 @@
...
       `RunTagFilter` for the run and range of Alerts.
     """
-    # TODO(cais): Add filter for alert type.
-    return provider.RunTagFilter(
-        runs=[run], tags=["%s_%d_%d" % (ALERTS_BLOB_TAG_PREFIX, begin, end)],
-    )
+    tag = "%s_%d_%d" % (ALERTS_BLOB_TAG_PREFIX, begin, end)
+    if alert_type is not None:
+        tag += "_%s" % alert_type
+    return provider.RunTagFilter(runs=[run], tags=[tag])
```
wchargin: OK, but recall that if this really is intended to be substitutable with
a generic data provider then each possible value of `alert_type` will
need to have a separate blob stored, duplicating the data. If this is
fine with you, then proceed.

tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -43,41 +43,51 @@
...
 
 def _parse_alerts_blob_key(blob_key):
     """Parse the BLOB key for Alerts.
 
     Args:
       blob_key: The BLOB key to parse. By contract, it should have the format:
-       `${ALERTS_BLOB_TAG_PREFIX}_${begin}_${end}.${run_id}`
+       - `${ALERTS_BLOB_TAG_PREFIX}_${begin}_${end}.${run_id}` when there is no
+         alert type filter.
+      - `${ALERTS_BLOB_TAG_PREFIX}_${begin}_${end}_${alert_filter}.${run_id}`
```
wchargin: nit: Inconsistent indentation.

tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -43,41 +43,51 @@
...
     key_body = key_body[len(ALERTS_BLOB_TAG_PREFIX) :]
-    begin = int(key_body.split("_")[1])
-    end = int(key_body.split("_")[2])
-    return run, begin, end
+    key_items = key_body.split("_")
+    begin = int(key_items[1])
+    end = int(key_items[2])
+    alert_type = None
+    if len(key_items) > 3:
+        alert_type = key_body.split("_", 3)[-1]
```
wchargin: Just call `key_body.split("_", 3)` above and use `key_items[2]` here?

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Done.

tensorboard/plugins/debugger_v2/debug_data_provider.py
caisq: The data duplication is fine with me. When we work on the BLOB storage, we can find ways to minimize the amount of data duplication by only allowing querying detailed alerts data of a given type at a time.

tensorboard/plugins/debugger_v2/debug_data_provider.py
caisq: Done.

tensorboard/plugins/debugger_v2/debug_data_provider.py
caisq: Done. Sorry - I missed this comment in that last commit.

PR #3289, opened 2020-02-24
projector: factor metadata out from frontend plugin

wchargin:
Summary:
This refactoring enables the public API `:projector` to not depend on
the frontend code. While factoring out these constant, we rename a
couple of them to match conventions in TensorBoard: e.g., `PLUGIN_NAME`
is typically used to represent the plugin name.

This is a helpful step in migrating `base_plugin.py` to use PY3-only
syntax (namely, hard-requiring kwargs to `TBContext`’s initializer).

Test Plan:
After syncing into Google, there are no longer any valid PY2 binaries
that transitively depend on `//tensorboard/plugins:base_plugin`.

wchargin-branch: projector-metadata

### DIFF
--- a/tensorboard/plugins/projector/BUILD
+++ b/tensorboard/plugins/projector/BUILD
@@ -17,6 +17,7 @@ py_library(
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [
+        ":metadata",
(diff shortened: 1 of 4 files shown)


wchargin:
cc @hfiller

PR #3299, opened 2020-02-26
uploader: add `PluginControl` handshake protos

wchargin:
Summary:
This enables us to flag features on the server side, so that we can
stage releases, and so that we can roll back in a pinch without
requiring an expedited PyPI push and end user version upgrade.

Test Plan:
No new behavior in this change. The protos still build.

wchargin-branch: uploader-plugincontrol-protos

### DIFF
--- a/tensorboard/uploader/proto/server_info.proto
+++ b/tensorboard/uploader/proto/server_info.proto
@@ -20,6 +20,16 @@ message ServerInfoResponse {
   ApiServer api_server = 2;
   // How to generate URLs to experiment pages.
   ExperimentUrlFormat url_format = 3;
+  // For which plugins should we upload data? (Even if the uploader is
+  // structurally capable of uploading data from many plugins, we only actually
+  // upload data that can be currently displayed in TensorBoard.dev. Otherwise,
+  // users may be surprised to see that experiments that they uploaded a while
+  // ago and have since shared or published now have extra information that
+  // they didn't realize had been uploaded.)
+  //
+  // If this field is omitted, the behavior of which data to send is at
+  // the discretion of the client.
+  PluginControl plugin_control = 4;
 }
 
 enum CompatibilityVerdict {
@@ -58,4 +68,10 @@ message ExperimentUrlFormat {
   // Placeholder string that should be replaced with an actual experiment ID.
   // (See docs for `template` field.)
   string id_placeholder = 2;
+}
+
+message PluginControl {
+  // Only send data from plugins with these names. These are plugin names as
+  // stored in the the `SummaryMetadata.plugin_data.plugin_name` proto field.
+  repeated string allowed_plugins = 1;
 }
--- a/tensorboard/uploader/proto/server_info.proto
+++ b/tensorboard/uploader/proto/server_info.proto
@@ -27,8 +27,8 @@ message ServerInfoResponse {
   // ago and have since shared or published now have extra information that
   // they didn't realize had been uploaded.)
   //
-  // If this field is omitted, the behavior of which data to send is at
-  // the discretion of the client.
+  // If this field is omitted, the behavior of which data to send is at the
+  // discretion of the client.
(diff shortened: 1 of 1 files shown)


tensorboard/uploader/proto/server_info.proto
```diff
@@ -20,6 +20,16 @@ message ServerInfoResponse {
...
   // How to generate URLs to experiment pages.
   ExperimentUrlFormat url_format = 3;
+  // For which plugins should we upload data? (Even if the uploader is
+  // structurally capable of uploading data from many plugins, we only actually
+  // upload data that can be currently displayed in TensorBoard.dev. Otherwise,
+  // users may be surprised to see that experiments that they uploaded a while
+  // ago and have since shared or published now have extra information that
+  // they didn't realize had been uploaded.)
+  //
+  // If this field is omitted, the behavior of which data to send is at the
```
davidsoergel: Maybe reword this comment a bit: which data to send is still at the discretion of the client, within the constraints given by this field.

tensorboard/uploader/proto/server_info.proto
```diff
@@ -59,3 +69,9 @@ message ExperimentUrlFormat {
   // (See docs for `template` field.)
   string id_placeholder = 2;
 }
+
+message PluginControl {
```
davidsoergel: I like how generic this message is.  We could in principle extend it with rate limits per plugin, informational messages per plugin, whatever makes sense later.

A slightly different design might prepare us for that eventuality even better.  Instead of one `PluginControl` for all plugins, you could do `repeated PluginControl plugin_controls;` where for now `PluginControl` contains just `string plugin_name`.  Or do that in a nested manner with `AllPluginControls` where you have it, with `repeated SinglePluginControl` inside (mod naming).

(Just to consider; I'm OK with it either way)

tensorboard/uploader/proto/server_info.proto
wchargin: Good point. Done.

tensorboard/uploader/proto/server_info.proto
wchargin: > I like how generic this message is. We could in principle extend it
> with rate limits per plugin, informational messages per plugin,
> whatever makes sense later.

Yep, exactly. Rate limits were something that I had in mind, as was a
flag “`bool allow_all_plugins = 2;`” for general-purpose storage (e.g.,
if people run their own instances, they _might_ want that).

> A slightly different design might prepare us for that eventuality even
> better […]

A downside of `repeated PluginControl plugin_controls;` is that it
doesn’t admit distinguishing between an old server that doesn’t know
about `PluginControl`s and a server that is denying all uploads. If we
did that, then rolling back the server to a pre-`PluginControl` version
would be dangerous once we’d pushed a new uploader, so I like having the
wrapper message.

An `AllPluginControls` with repeated `SinglePluginControl`s does resolve
this, and indeed is more flexible along the plugin axis. But it’s a bit
less flexible on other axes: e.g., I could imagine wanting

```proto
message ServerInfoResponse {
  // ...
  PluginControl plugin_control = 4;
  repeated RateLimit rate_limit = 5;
}

message RateLimit {
  oneof what {
    // Rate-limit a specific plugin.
    string plugin_name = 1;
    // Rate-limit all data of a given data class.
    .tensorboard.DataClass data_class = 2;
  }
  // Send requests at most once every this often.
  Duration min_request_period = 3;
}
```

in which case the `SinglePluginControl`s aren’t obviously the right
place to put the rate limiting.

Given that both options seem reasonable and the worst-case costs aren’t
too high (we can either deprecate the old fields entirely, or use a
hybrid solution like denormalizing plugin names or using the plugin
control ordering as an index; neither is perfect, but they’re not
maintenance nightmares), I’m going to keep this as is, because this way
is simpler right now. Thank you for the suggestion!

PR #3300, opened 2020-02-26
uploader: only send data for allowed plugins

wchargin:
Summary:
When the server sends a `PluginControl` message in the initial
handshake, we now only send data for plugins specified therein.

Test Plan:
Unit tests included. As end-to-end tests:

  - When running against prod (which does not send `plugin_control`),
    experiment upload is unchanged. Info logs now include “Skipping time
    series […] with unsupported plugin name 'histograms'”, just once per
    time series.

  - When running against a local server that serves scalars-only
    explicitly—

    ```
    plugin_control {
      allowed_plugins: "scalars"
    }
    ```

    —the behavior is also the same.

  - When running against a local server that serves “no plugins allowed”
    explicitly—

    ```
    plugin_control {
    }
    ```

    —nothing is uploaded, and the logs include “Skipping…” messages for
    scalar time series.

wchargin-branch: uploader-plugincontrol


wchargin:
Server-side change (Google-internal): <http://cl/297460920>

wchargin:
(Made this patch a bunch shorter and cleaner by pulling out #3304, which
facilitates #3305 as well.)

tensorboard/uploader/uploader.py
```diff
@@ -295,23 +304,32 @@ def send_requests(self, run_to_events):
             # If later events arrive with a mismatching plugin_name, they are
             # ignored with a warning.
             metadata = self._tag_metadata.get(time_series_key)
+            first_in_time_series = False
```
davidsoergel: This mechanism of detecting 'first time' is a little entangled hard to follow.  How about maintaining a collection of `plugin_names_seen` and testing `if plugin_name in plugin_names_seen:`?

Related, the 'first time' gets reset for every `_upload_once()`.  Is that desirable, or should we maintain a singleton `plugin_names_seen` (or similar) that persists over cycles?  In this case you'd change the error message too, of course, so it's not time-series-specific but just appears once per plugin.

PR #3305, opened 2020-02-28
uploader: fix rate limiting for logdir polling

### LINKED ISSUE #3301: Rate limiting broken for uploader logdir polling
Repro:

```
bazel run //tensorboard -- dev upload --logdir /tmp/nope --verbosity 0
```

(where `/tmp/nope` is any path; a nonexistent or empty directory will
repro most quickly).

Expected behavior: We start an upload cycle and poll the logdir once
every 5 seconds.

Actual behavior (latest nightly): We poll the logdir with unbounded
frequency. Logs fly off the screen. I’m seeing about 6000 polls per
second on my workstation (SSD). This is bad for CPU, bad for disks, and
bad for your wallet if your logdir is on a remote filesystem.

`git bisect` says: 9077c10dfef3a3dcf00853f436340bbc17e68b71 is the first
bad commit. (This makes sense.)

We can’t release to PyPI stable in this state.

wchargin:
Summary:
Recent refactorings to the RPC rate limiting had the unintended side
effect of removing rate limiting on logdir polling. This meant that
after all data had been read from the logdir, the uploader would
continue to poll it aggressively (thousands of times per second on my
machine), which is bad for disks and expensive on network file systems.

This commit adds a separate rate limiter for the logdir polling. We
don’t reuse the RPC rate limiter because that would force us to tick
twice on the same timer every cycle.

Fixes #3301.

Test Plan:
Run `bazel run //tensorboard -- dev upload --logdir /nope --verbosity 0`
(with any logdir path, but it’s easier when it’s empty), and note that
the “Starting an upload cycle” logs now progress at 0.2 per second
rather than 6000 per second. Unit tests also included: if the `.tick()`
is removed, they fail (quickly, without hanging).

wchargin-branch: uploader-rate-limit-polling


tensorboard/uploader/uploader.py
```diff
@@ -38,6 +38,11 @@
 from tensorboard.util import tb_logging
 from tensorboard.util import tensor_util
 
+# Minimum length of a logdir polling cycle in seconds. Shorter cycles will
+# sleep to avoid spinning over the logdir, which isn't great for disks and can
+# be expensive for network file sytems.
```
davidsoergel: typo: sytems

wchargin: Fixed; thanks.

PR #3308, opened 2020-02-29
uploader: export experiment metadata

wchargin:
Summary:
When running `tensorboard dev export`, experiment names and descriptions
(set with the `update-metadata` subcommand or the `--name` and
`--description` flags to `upload`) as well as experiment creation and
modification times are now emitted to a new `metadata.json` file in the
experiment directory.

Test Plan:
Unit tests included. Also end-to-end tested against a live server,
exporting some experiments with name/description and some without.

wchargin-branch: uploader-export-metadata

### DIFF
--- a/tensorboard/uploader/exporter.py
+++ b/tensorboard/uploader/exporter.py
@@ -121,7 +121,6 @@ class TensorBoardExporter(object):
         """
         if read_time is None:
             read_time = time.time()
-<<<<<<< HEAD
         experiment_metadata_mask = experiment_pb2.ExperimentMask(
             create_time=True, update_time=True, name=True, description=True,
         )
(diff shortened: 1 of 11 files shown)


tensorboard/uploader/exporter.py
```diff
@@ -115,10 +118,40 @@ def export(self, read_time=None):
...
+                    "name": experiment.name,
+                    "description": experiment.description,
+                    "create_time": util.format_time_absolute(
+                        experiment.create_time
+                    ),
+                    "update_time": util.format_time_absolute(
+                        experiment.update_time
+                    ),
+                }
+            elif isinstance(experiment, str):
```
bileschi: Are there actually anymore of these old servers that stream experimen_id instead of experiment proto?

wchargin: Good question—no, there shouldn’t be. I was considering what to do about
this and thought that it was easier and a bit more conservative to keep
the code as is (this is moved out of `_request_experiment_ids`) than to
change `list_experiments` below and drop the old code path. But you’re
right that this has been deployed for a while, so it looks unlikely that
we’ll need to roll the server back. I can go ahead and make that change.

wchargin: Done as a pre-change in #3310, which indeed simplifies this diff by a
fair amount.

PR #3312, opened 2020-02-29
base_plugin: enforce kwarg-only initializers

wchargin:
Summary:
These initializers have always been documented as requiring keyword
arguments, so this is not a breaking change. We can now use Python 3’s
keyword-only argument syntax to actually enforce this.

The trailing comma removal is necessary to work around a parser bug in
cPython 3.5: <https://bugs.python.org/issue9232>

Test Plan:
All tests pass, and no Google-internal Python 2 binaries depend on this
target.

wchargin-branch: base-plugin-kwargs


wchargin:
Merging master into this branch; I expect that this will cause a fast CI
failure caught by #3320, which I’ll then fix.

PR #3313, opened 2020-02-29
[DebuggerV2] Add mat-slider to TimelineComponent

caisq:
  * Continue developing DebuggerV2 Plugin
  * Add `mat-slider` to TimelineComponent
  * Let it's `max` and `value` properties be bound to number of executions and current `scrollBeginIndex` for executions, respectively
  * Note that the drag action still doesn't work with the `mat-slider` yet. A TODO item reflects that.
  * Another TODO item, which may be related to the ongoing ng themeing task, is to set the color of the thumb and track of the `mat-slider`.
  * Replace two `button`s in TimelineComponent with `mat-button`s.

* Screenshots of UI changes
image

  * Unit tests added for the new parts in container, reducer and effects
  * Manual verification based on real tfdbg2-format dump data in logdir


caisq:
cc @baileydesign

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/timeline/timeline_container.ts
```diff
@@ -122,6 +125,18 @@ export class TimelineContainer {
     select(getExecutionScrollBeginIndex)
   );
 
+  readonly scrollBeginIndexUpperLimit$ = this.store.pipe(
+    select(
+      createSelector(
+        getNumExecutions,
+        getDisplayCount,
+        (numExecutions, displayCount) => {
+          return Math.max(0, numExecutions - displayCount);
```
stephanwlee: I think this invariant has to be enforced by the store (write time). Do we absolutely need one at the read time?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/timeline/timeline_component.css
```diff
@@ -58,8 +58,35 @@ limitations under the License.
   margin-bottom: 5px;
 }
 
+.timeline-slider {
+  display: inline-block;
+  height: 48px;
+  padding: 0px;
```
stephanwlee: 0px -> 0.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
```diff
@@ -870,6 +870,75 @@ describe('Debugger reducers', () => {
...
+          actions.executionScrollToIndex({index: scrollIndex})
+        );
+        expect(nextState.executions.scrollBeginIndex).toBe(
+          originalScrollBeginIndex
+        );
+      }
+    );
+  }
+
+  for (const scrollIndex of [-1, 0.5, 1, 2, 20]) {
```
stephanwlee: I don't think I am understanding the test by itself. I get -1 and 0.5 but why are 1 and 2 problematic? What about 20?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -410,6 +410,35 @@ const reducer = createReducer(
...
     }
   ),
+  on(
+    actions.executionScrollToIndex,
+    (state: DebuggerState, action: {index: number}): DebuggerState => {
+      if (action.index < 0 || !Number.isInteger(action.index)) {
+        console.error(
+          `Attempt to scroll to negative or non-integer execution index ` +
+            `(${action.index})`
+        );
```
stephanwlee: If this cannot happen for any reasons (with UI), just throw an invariant error instead

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container_test.ts
```diff
@@ -341,8 +342,93 @@ describe('Debugger Container', () => {
...
+            );
+            store.setState(createState(debuggerState));
+            fixture.detectChanges();
+
+            const sliders = fixture.debugElement.queryAll(
+              By.css('.timeline-slider')
+            );
+            expect(sliders.length).toBe(1);
+            const [slider] = sliders;
+            expect(slider.attributes['ng-reflect-min']).toBe('0');
```
stephanwlee: I think `ng-reflect` attributes are not reliable. Is there no other way to assert this? You can probably grab `input` and assert its property instead.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
```diff
@@ -708,6 +709,76 @@ describe('Debugger effects', () => {
...
+            end: 60 + pageSize,
+            num_digests: numExecutions,
+            execution_digests: [],
+          };
+          for (let i = 0; i < pageSize; ++i) {
+            executionDigestsResponse.execution_digests.push({
+              op_type: 'FooOp',
+              output_tensor_device_ids: ['d1'],
+            });
+          }
```
stephanwlee: `new Array(pageSize).fill({...})`?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
```diff
@@ -708,6 +709,76 @@ describe('Debugger effects', () => {
...
+          store.overrideSelector(getDisplayCount, displayCount);
+          store.overrideSelector(getExecutionPageSize, pageSize);
+          const pageLoadedSizes: {[pageIndex: number]: number} = {
+            0: 20,
+            1: 20,
+            2: 20,
+          };
+          if (dataAlreadyExists) {
+            pageLoadedSizes[3] = 10;
+          }
```
stephanwlee: instead of logic inside test, just parameterize pageLoadSize too. Otherwise, I will expect to have test on tests.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container_test.ts
caisq: Did not find the `input` property for `slider` or `slider.nativeElement`. I'm now using `aria-valuemin`, `aria-valuemax` and `aria-valuenow`.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
caisq: I made this just throw an `Error`. We don't have a dedicated class for invariant errors yet.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
caisq: Here, displayCount is 50 and there are only 20 execution digests (< 50). In this case, the scroll bar is hidden in the TimelineComponent and scrolling is disabled (i.e., the only valid scrolling begin index is 0). I added a comment here to clarify that.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/timeline/timeline_component.css
caisq: Done. Fixed another instance of this mistake as well.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/timeline/timeline_container.ts
caisq: There is no invariant here. `numExecutions` can be legitimately `>`, `=`, or `<` `displayCount`. This selector is just determining which of those is the case and help decide whether the slider should be shown or not.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container_test.ts
```diff
@@ -341,8 +342,93 @@ describe('Debugger Container', () => {
...
+        );
+        expect(sliders.length).toBe(0);
+      });
+    }
+
+    for (const numExecutions of [11, 12, 20]) {
+      const displayCount = 10;
+      for (const scrollBeginIndex of [0, 1, numExecutions - displayCount]) {
+        it(
+          `shows slider if # of executions ${numExecutions} > display count,` +
```
stephanwlee: need a whitespace at the end the spec read better.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/timeline/timeline_module.ts
```diff
@@ -13,16 +13,43 @@ See the License for the specific language governing permissions and
...
+  // dragging to work. But the dragging is still not working, needs more
+  // investigation. May be related to:
+  // https://github.com/angular/components/issues/7905
+  providers: [
+    {
+      provide: HAMMER_LOADER,
+      useValue: () => new Promise(() => {}),
+    },
+    {provide: HAMMER_GESTURE_CONFIG, useClass: GestureConfig},
+  ],
```
stephanwlee: If HAMMER_LOADER does not help, would it make sense to just omit it for now and wait til Angular 9? 

I strongly recommend just using http://go/mdn/HTML/Element/input/range until we are ready.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/debugger_container_test.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/timeline/timeline_module.ts
caisq: Done.

PR #3315, opened 2020-03-01
Added S3_ENDPOINT variable

### LINKED ISSUE #3014: Support environment variable S3_ENDPOINT in tensorboard without tensorflow
When using tensorboard with tensorflow installed, we are able to set the s3_endpoint using S3_ENDPOINT environment variable. It is particularly usefull when running a local s3 instance like minio.

But when using the tensorboard package (without tensorflow), it uses the boto3 client, and don't use the S3_ENDPOINT variable.

A simple modification would be to replace boto3.client/session call with the same call but with the parameter: endpoint_url=os.environ.get('S3_ENDPOINT', None)

What do you think ?

thealphacod3r:
Fixes #3014 

@stephanwlee please review this!

### DIFF
--- a/tensorboard/compat/tensorflow_stub/io/gfile.py
+++ b/tensorboard/compat/tensorflow_stub/io/gfile.py
@@ -220,6 +220,7 @@ class S3FileSystem(object):
     def __init__(self):
         if not boto3:
             raise ImportError("boto3 must be installed for S3 support.")
+        self.s3_endpoint = os.environ.get("S3_ENDPOINT", None)
 
     def bucket_and_path(self, url):
         """Split an S3-prefixed URL into bucket and path."""
@@ -233,7 +234,7 @@ class S3FileSystem(object):
 
     def exists(self, filename):
         """Determines whether a path exists or not."""
-        client = boto3.client("s3")
+        client = boto3.client("s3", endpoint_url = self.s3_endpoint)
         bucket, path = self.bucket_and_path(filename)
         r = client.list_objects(Bucket=bucket, Prefix=path, Delimiter="/")
         if r.get("Contents") or r.get("CommonPrefixes"):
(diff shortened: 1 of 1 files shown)


stephanwlee:
I've learned recently that TensorFlow generally prefer non-environment variable solutions. Any thoughts on this, @nfelt?

While this change looks reasonable, I think we actually should prefer the CLI flags to mirror AWS cli (https://docs.aws.amazon.com/cli/latest/reference/).

thealphacod3r:
@nfelt please review this!

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -220,6 +220,7 @@ class S3FileSystem(object):
     def __init__(self):
         if not boto3:
             raise ImportError("boto3 must be installed for S3 support.")
+        self.s3_endpoint = os.environ.get("S3_ENDPOINT", None)
```
nfelt: Please prefix with an underscore (`self._s3_endpoint`).

thealphacod3r:
@googlebot I signed it!

tensorboard/compat/tensorflow_stub/io/gfile.py
thealphacod3r: @nfelt i made the changes as you suggested please look into this, i ran black and reformatted the code but still the lint error exists!

thealphacod3r:
@googlebot I consent.

thealphacod3r:
@nfelt I recently changed my email address and now googlebot has marked my cla:no, i signed the cla with my new email address but still it's telling cla as no.

tensorboard/compat/tensorflow_stub/io/gfile.py
nfelt: Can you rerun black?  As long as you're running it correctly it should ensure the lint passes on CI.

It looks like the problem is the spaces around the equals sign in the `endpoint_url = self._s3_endpoint` line below.

thealphacod3r:
@nfelt should i close this and start with a new PR, it would be much easier to sort things then. Is it ok if i do so?

nfelt:
It shouldn't be too hard to fix this PR, but if you prefer opening a new one, you can also do that.

PR #3318, opened 2020-03-02
ci: test against Python 3.5 rather than 3.6

wchargin:
Summary:
We used to test on Python 2.7 and 3.6, which was sufficient to catch
most version-specific bugs in our support range. When we dropped support
for Python 2.7, we left CI at 3.6 only, but this means that we don’t
catch Python 3.5–specific failures. Running at Python 3.5 only should be
sufficient to mostly cover 3.5 through 3.7, especially given that the
TensorBoard team mostly uses Python 3.7 for daily development.

Thus, we also work around a bug in cPython 3.5, where parsing trailing
commas after `*args` is broken: <https://bugs.python.org/issue9232>. The
change to `pyproject.toml` prevents Black from re-introducing these
trailing commas.

We leave `py27` in the `pyproject.toml` to avoid a mass reformat
removing `u`-prefixes from Unicode string literals (which is fine, but
should be a separate change for `.git-blame-ignore-revs`).

Test Plan:
That CI passes suffices.

wchargin-branch: ci-py35

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-python@v1
         with:
-          python-version: '3.6'
+          python-version: '3.5'
           architecture: 'x64'
       - name: 'Install black and flake8'
         run: |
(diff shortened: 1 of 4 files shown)


wchargin:
Not yet ready for review because I expect this to fail, as mentioned in
the commit message.

wchargin:
Ah, yes: `AttributeError: assert_called_once`. How could I forget.

Hmm. I’ll think about this.

wchargin:
Python 3.5 is end of life in 2 weeks and 3 days. This doesn’t look like
it has much chance to be useful.

PR #3319, opened 2020-03-02
Silence deprecation warning spam about tf_record_iterator

nfelt:
This puts a nice 🩹 over the deprecation warning spam emitted to the console by our usage of `tf.compat.v1.io.tf_record_iterator` (which we now do for new TF versions, as of https://github.com/tensorflow/tensorboard/pull/3185).

Actually migrating to something not deprecated is still tracked in #1711.

### DIFF
--- a/tensorboard/backend/event_processing/event_file_loader.py
+++ b/tensorboard/backend/event_processing/event_file_loader.py
@@ -18,6 +18,8 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import contextlib
+
 from tensorboard.compat import tf
 from tensorboard.compat.proto import event_pb2
 from tensorboard.util import platform_util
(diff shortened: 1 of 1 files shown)


tensorboard/backend/event_processing/event_file_loader.py
```diff
@@ -27,6 +29,26 @@
...
+
+
+def _silence_deprecation_warnings():
+  """Context manager that best-effort silences TF deprecation warnings."""
+  try:
+    # Learn this one weird trick to make TF deprecation warnings go away.
+    from tensorflow.python.util import deprecation
+    return deprecation.silence()
+  except (ImportError, AttributeError):
+    return _NULL_CONTEXT
```
wchargin: You could avoid the `_NULL_CONTEXT` backport with something like this:

```python
@contextlib.contextmanager
def _silence_deprecation_warnings():
  try:
    # Learn this one weird trick to make TF deprecation warnings go away.
    from tensorflow.python.util import deprecation
    cm = deprecation.silence()
  except (ImportError, AttributeError):
    yield
  else:
    with cm:
      yield
```

(Or `cm = None; ...; if cm is not None: ...` if you dislike the `else`.)

If you prefer it as is, then nit: [`nullcontext`], not `null_context`.

[`nullcontext`]: https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext

wchargin:
Also, wow: my Noto doesn’t have U+1FA79. I guess there’s still some Tofu
left after all.

tensorboard/backend/event_processing/event_file_loader.py
nfelt: Fixed spelling.  IMO it's actually simpler and more readable just to use the null context manager and have an either-or rather than nested context managers?

Especially given that it's essentially no code to backport (I didn't even realize there was a stdlib one in any version until I checked on a whim).

wchargin: Okay; fine with me if you prefer this.

PR #3320, opened 2020-03-03
Use cPython 3.5–compatible function definitions

wchargin:
Summary:
In Python as of PEP 3102, function definitions may include keyword-only
arguments after a splat (`def foo(x, *, y)`). But cPython 3.5 has a
parser bug (<https://bugs.python.org/issue9232>) when trailing commas
are used with this syntax, as in `def foo(x, *, y,)`. This is fixed in
Python 3.6 and up, but not backported to 3.5, so for now we should make
sure to omit these trailing commas.

The change to `pyproject.toml` prevents Black from re-introducing these
trailing commas. We leave `py27` in the `pyproject.toml` to avoid a mass
reformat removing `u`-prefixes from Unicode string literals (which is
fine, but should be a separate change for Bazel `srcs_version`).

We don’t run our full test suite on Python 3.5 yet, due to unrelated
issues (primarily, lack of `assert_not_called` and that `json.loads`
does not accept `bytes`). For now, it suffices to run `flake8` on both
Python 3.5 and Python 3.7, because that at least parses all the code.

Test Plan:
That CI passes suffices.

wchargin-branch: py35-fundefs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ env:
   BUILDIFIER_SHA256SUM: '4c985c883eafdde9c0e8cf3c8595b8bfdf32e77571c369bf8ddae83b042028d6'
 
 jobs:
-  lint-flake8:
+  lint-python-flake8:
     runs-on: ubuntu-16.04
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -24,15 +24,22 @@ env:
   BUILDIFIER_SHA256SUM: '4c985c883eafdde9c0e8cf3c8595b8bfdf32e77571c369bf8ddae83b042028d6'
 
 jobs:
-  lint-python:
+  lint-flake8:
```
nfelt: nit: maybe call this `lint-python-flake8` just to be clear to the uninitiated that it's also linting python?  Since otherwise separating it out from `lint-python` is a little confusing.

wchargin: Sure; fine with me. Done.

PR #3326, opened 2020-03-04
Experimental plugins

bmd3k:

Allow an administrator of a Tensorboard to mark certain plugins as "experimental" and hidden to users by default. Users would then need to add query parameter "experimentalPlugin=<plugin name>" to the URL of their experiment to see the experimental plugin.

standard_tensorboard_wsgi and TensorBoardWSGIApp include an argument "experimental_plugins" that lists plugins, by name, that are in an experimental state.

There is logic in _serve_plugins_listing() that filters out plugins that are included in experimental_plugins but are not specified as a value for an experimentalPlugin query parameter.

There is logic in createRouter() that forward "experimentalPlugin" query parameter values from the original TensorBoard request to the call to the plugins_listing endpoint.

* Screenshots of UI changes

Using an example where "images" and "graphs" are marked as experimental.

No query parameters:
image

"experimentalPlugin=images" query parameter:
image

"experimentalPlugin=images&experimentalPlugin=graphs" query parameters:
image

Modify your favorite instance of Tensorboard to include experimental_plugins = ["images", "graphs"] in call to standard_tensorboard_wsgi() or TensorBoardWSGIApp().

Start server.

Open "http://localhost:6006/" without query parameters. Note that "Images" and "Graphs" do not appear as plugins.

Open "http://localhost:6006/?experimentalPlugin=images&experimentalPlugin=graphs". Note that "Images" and "Graphs" appear as plugins.

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -89,6 +89,8 @@ PLUGIN_PREFIX = "/plugin"
 PLUGINS_LISTING_ROUTE = "/plugins_listing"
 PLUGIN_ENTRY_ROUTE = "/plugin_entry.html"
 
+EXPERIMENTAL_PLUGINS_QUERY_PARAM = "expplugin"
+
 # Slashes in a plugin name could throw the router for a loop. An empty
 # name would be confusing, too. To be safe, let's restrict the valid
 # names as follows.
@@ -114,13 +116,17 @@ def _apply_tensor_size_guidance(sampling_hints):
     return tensor_size_guidance
 
 
-def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
+def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider, experimental_plugins=[]):
     """Construct a TensorBoardWSGIApp with standard plugins and multiplexer.
 
     Args:
       flags: An argparse.Namespace containing TensorBoard CLI flags.
       plugin_loaders: A list of TBLoader instances.
       assets_zip_provider: See TBContext documentation for more information.
+      experimental_plugins: A list of plugin names that are only provided
+        experimentally. The corresponding plugins will only be activated for
+        a user if the user has specified the plugin with the expplugin query
+        parameter in the URL.
 
     Returns:
       The new TensorBoard WSGI application.
@@ -177,10 +183,10 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
         start_reloading_multiplexer(
             multiplexer, path_to_run, reload_interval, flags.reload_task
         )
-    return TensorBoardWSGIApp(
-        flags, plugin_loaders, data_provider, assets_zip_provider, multiplexer
-    )
 
+    return TensorBoardWSGIApp(
+        flags, plugin_loaders, data_provider, assets_zip_provider, multiplexer, experimental_plugins
+    )
 
 def _handling_errors(wsgi_app):
     def wrapper(*args):
@@ -205,6 +211,7 @@ def TensorBoardWSGIApp(
     data_provider=None,
     assets_zip_provider=None,
     deprecated_multiplexer=None,
+    experimental_plugins=[],
 ):
     """Constructs a TensorBoard WSGI app from plugins and data providers.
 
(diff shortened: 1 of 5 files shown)


tensorboard/backend/application.py
```diff
@@ -116,7 +116,9 @@ def _apply_tensor_size_guidance(sampling_hints):
     return tensor_size_guidance
 
 
-def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider, experimental_plugins=[]):
+def standard_tensorboard_wsgi(
+    flags, plugin_loaders, assets_zip_provider, experimental_plugins=[]
```
wchargin: Use `experimental_plugins=None` instead and do the resolution inside the
function body, because:

  - default argument values are evaluated at function definition time,
    not invocation time, so this introduces a globally shared mutable
    list object—not what you want! (cf. [style guide][style])
  - using `None` as a default argument value in general is a [common
    Python pattern][none] enabling callers to specify “use default”
    without having to duplicate the actual value of the default

(Likewise below.)

[style]: https://google.github.io/styleguide/pyguide.html#2124-decision
[none]: https://unhashable.com/always-use-none-for-default-args-in-python/

tensorboard/backend/application.py
```diff
@@ -89,6 +89,8 @@
 PLUGINS_LISTING_ROUTE = "/plugins_listing"
 PLUGIN_ENTRY_ROUTE = "/plugin_entry.html"
 
+EXPERIMENTAL_PLUGINS_QUERY_PARAM = "expplugin"
```
wchargin: Could we call this `experimentalPlugin` instead? *exp* is ambiguous in
general, and especially so given that TensorBoard has a notion of
experiments that’s not what’s meant here.

tensorboard/backend/application.py
```diff
@@ -285,6 +316,7 @@ def __init__(self, plugins, path_prefix="", data_provider=None):
         self._plugins = plugins
         self._path_prefix = path_prefix
         self._data_provider = data_provider
+        self._experimental_plugins = experimental_plugins
```
wchargin: Since this is only used for membership tests, best to convert it to a
`frozenset` for faster queries and more idiomatic set operations (see
suggestion below).

tensorboard/backend/application.py
```diff
@@ -467,7 +499,18 @@ def _serve_plugins_listing(self, request):
...
-        for plugin in self._plugins:
+        plugins_to_consider = filter(
+            # Filter out experimental plugins that were not activated using the query param.
+            lambda plugin: (
+                plugin.plugin_name not in self._experimental_plugins
+            )
+            or (
+                plugin.plugin_name
+                in request.args.getlist(EXPERIMENTAL_PLUGINS_QUERY_PARAM)
+            ),
```
wchargin: Instead of `filter(lambda: ..., ...)`, use a list comprehension, per the
style guide:
<https://google.github.io/styleguide/pyguide.html#215-deprecated-language-features>

But probably simpler here to just conditionally `continue` at the top of
the `for`-loop? Certainly shorter:

```python
        plugins_to_skip = self._experimental_plugins - frozenset(
            request.args.getlist(EXPERIMENTAL_PLUGINS_QUERY_PARAM)
        )
        for plugin in self._plugins:
            if p.plugin_name in plugins_to_skip:
                continue
            ...
```

tensorboard/backend/application_test.py
```diff
@@ -373,6 +373,64 @@ def fake_is_active(self):
         self.assertEqual(parsed_object["foo"]["enabled"], False)
         self.assertEqual(parsed_object["baz"]["enabled"], True)
 
+    def testPluginsListingWithExperimentalPlugin(self):
```
wchargin: Nice tests; thanks!

tensorboard/components/tf_backend/router.ts
```diff
@@ -34,7 +36,10 @@ namespace tf_backend {
    *
    * @param dataDir {string=} The base prefix for data endpoints.
    */
-  export function createRouter(dataDir = 'data'): Router {
+  export function createRouter(
+    dataDir = 'data',
+    windowLocationUrl = new URL(window.location.href)
```
wchargin: This is a little scary; in the past, we’ve had issues with code like
this breaking `--path_prefix` handling. It’s fine here because you’re
only using it for the search params rather than the path, but I wonder
if we can make that clearer—something like

```typescript
  export function createRouter(
    dataDir = 'data',
    searchParams = new URLSearchParams(window.location.search)
  ): Router {
    // ...
  }
```

such that future changes to the function don’t accidentally rely on
`path`?

tensorboard/program.py
```diff
@@ -122,6 +123,7 @@ def __init__(
         assets_zip_provider=None,
         server_class=None,
         subcommands=None,
+        experimental_plugins=[],
```
wchargin: Hmm; it would be great if we could avoid exposing this on `program`;
unlike `backend.application`, `program` is a public API and I’m not sure
that we want to commit to supporting this forever.

Do you have any thoughts about the best way to do this? The internal
clients that need this don’t call through `program` (instead calling
`TensorBoardWSGIApp` directly), but I see that `tensorboard.main` does.

wchargin:
(For posterity, this is superseded by #3344.)

PR #3328, opened 2020-03-04
ng: reconfigure MetaReducer to use Injectable

stephanwlee:
Previously, even in the debug mode, the logger metaReducer had no
effect. When debugging the issue, it was found that the ngfactory which
is part of Angular's DI system prebuilds injectable for the META_REDUCER
and inlines an empty object [1]. Do note that this problem only manifests
when we try to configure metaReducer in a conditional. Also do note that
calling `isDevMode()` method is disallowed in AOT compiler, which is the
default compiler in Angular 9.

Tested the correct behavior under `bazel run -c dbg` vs. `bazel run`.

[1]: To reproduce, please build without this commit and inspect
bazel-bin/tensorboard/webapp/app_module.ngfactory.js. Look for, in very
dense code, USER_PROVIDED_META_REDUCERS and notice that it provides
an empty array.

### DIFF
--- a/tensorboard/webapp/app_module.ts
+++ b/tensorboard/webapp/app_module.ts
@@ -15,7 +15,7 @@ limitations under the License.
 import {BrowserModule} from '@angular/platform-browser';
 import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
 import {NgModule} from '@angular/core';
-import {StoreModule} from '@ngrx/store';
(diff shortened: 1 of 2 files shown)


stephanwlee:
Relates to #2976.

stephanwlee:
> Yikes, iiuc Angular's AOT compiler is currently thrown off by this conditional, and doesn't give any error? If so, that's extremely scary.

Yup, it does complain about function calls but not conditionals... :man_shrugging:

PR #3330, opened 2020-03-05
[DebuggerV2] Add types, selectors, reducers & effects for source code

caisq:
  * Continue developing DebuggerV2 plugin.

  * Add some supporting types for source code and stack frames
  * Add data source method for fetching list of all source files known to the debugger
  * Add selectors for the source-file-list-related state fields
  * Add actions for requesting and loading the source file list
  * Add effects for loading the source-file list on debugger view load

* Screenshots of UI changes
  * N/A since this is a pure type + ngrx change

  * Unit tests added

### DIFF
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
@@ -25,6 +25,8 @@ import {TBHttpClient} from '../../../../webapp/webapp_data_source/tb_http_client
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
+export type SourceFileListResponse = Array<[string, string]>;
+
 export interface StackFramesResponse {
   stack_frames: StackFrame[];
 }
@@ -77,6 +79,8 @@ export abstract class Tfdbg2DataSource {
     begin: number,
     end: number
   ): Observable<ExecutionDataResponse>;
+
+  abstract fetchSourceFileList(run: string): Observable<SourceFileListResponse>;
 
   abstract fetchStackFrames(
     run: string,
@@ -137,6 +141,17 @@ export class Tfdbg2HttpServerDataSource implements Tfdbg2DataSource {
           run,
           begin: String(begin),
           end: String(end),
+        },
+      }
+    );
+  }
+
+  fetchSourceFileList(run: string): Observable<SourceFileListResponse> {
+    return this.http.get<SourceFileListResponse>(
+      this.httpPathPrefix + '/source_files/list',
+      {
+        params: {
+          run,
         },
       }
     );
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
@@ -25,6 +25,8 @@ import {TBHttpClient} from '../../../../webapp/webapp_data_source/tb_http_client
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
+// The backend route for source-file list responds with an array
+// of 2-tuples: <host_name, file_path>.
 export type SourceFileListResponse = Array<[string, string]>;
 
 export interface StackFramesResponse {
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
@@ -622,9 +622,10 @@ export class DebuggerEffects {
         this.store.select(getActiveRunId),
         this.store.select(getSourceFileListLoaded)
       ),
-      filter(([, runId, fileIndexLoadState]) => {
+      filter(([, runId, sourceFileListLoadState]) => {
         return (
-          runId !== null && fileIndexLoadState.state !== DataLoadState.LOADING
+          runId !== null &&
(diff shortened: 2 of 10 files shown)


caisq:
cc @stephanwlee @baileydesign

caisq:
Gentle ping @psybuzz :) Please again note that this CL is orthogonal to the ongoing design discussion of the third-party library for actual code rendering.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -469,6 +478,37 @@ const reducer = createReducer(
...
+  ),
+  on(
+    actions.sourceFileListLoaded,
+    (state: DebuggerState, sourceFileList): DebuggerState => {
+      return {
+        ...state,
+        sourceCode: {
+          ...state.sourceCode,
+          sourceFileListLoaded: {
+            state: DataLoadState.LOADED,
```
psybuzz: nit: since L489 adds 
```
sourceFileListLoaded: {
            ...state.sourceCode.sourceFileListLoaded,
```
we should probably add the same thing here for safety?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -206,6 +206,26 @@ export interface RunState {
...
 }
 
+// A source-code file.
+export interface SourceFileSpec {
+  host_name: string;
+  file_path: string;
+}
+
+// A specific line of a source-code file, i.e., a stack frame.
+export interface SourceLineSpec extends SourceFileSpec {
```
psybuzz: nit: totally optional and up to you, but for naming, wdyt about `SourceLocation` ?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -469,6 +478,37 @@ const reducer = createReducer(
...
+            lastLoadedTimeInMs: Date.now(),
+          },
+          sourceFileList: sourceFileList.sourceFiles,
```
psybuzz: Based on the data access, we always pass a `run` when requesting sourceFiles.

To double check: is there any use case in the UX when a user wants to keep seeing stale source files?  or source files from a previously selected run?  (e.g. should we just merge the new + previous list of source files?)

I'd imagine if there's a run-selector UI control, a user might switch back and forth between 2 runs many times, and since we don't index the sourceFiles by run-id, we would need to re-fetch on each switch.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/data_source/tfdbg2_data_source.ts
```diff
@@ -25,6 +25,8 @@ import {TBHttpClient} from '../../../../webapp/webapp_data_source/tb_http_client
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
+export type SourceFileListResponse = Array<[string, string]>;
```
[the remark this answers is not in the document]
caisq: Added comment to clarify what the two strings are here.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -608,13 +613,44 @@ export class DebuggerEffects {
     );
   }
 
+  /**
+   * Load list of source files when debugger plugin is loaded.
+   */
+  private loadSourceFileList(prevStream$: Observable<void>) {
```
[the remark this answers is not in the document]
caisq: Acknowledged. Your understanding is correct.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -608,13 +613,44 @@ export class DebuggerEffects {
...
+    return prevStream$.pipe(
+      withLatestFrom(
+        this.store.select(getActiveRunId),
+        this.store.select(getSourceFileListLoaded)
+      ),
+      filter(([, runId, fileIndexLoadState]) => {
```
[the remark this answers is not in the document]
caisq: Agreed. I made revisions to make the naming more consistent.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
caisq: I prefer keeping `SourceLineSpec`. Reasons:

1. It's symmetric w.r.t. `SourceFileSpec` above.
2. It's specific about whether it's talking about a line or a range of lines (the former is the case currently), and whether it's talking about a single stack frame or a list of them (again the former is the case.)

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
caisq: Good question. The debugger v2 FE currently is implemented in a way that shows only one run at a time. This keeps the data structure simpler than what it would be if multiple runs are handled. This is sufficient IMO because when during TF debugging, you typically examine one run (i.e., one training job) at a time. There is less concern for comparing, say the loss curves from different hyperparameters from different runs, than in other TB dashboards such as scalars.

PR #3332, opened 2020-03-05
Buildifier format js.bzl for future edits

stephanwlee:
With autoformatter for JS and Python, we got used to auto-formatting
everything. Similarly, BUILD files can be formatted and validated using
buildifier but we currently do not apply it to .bzl.

However, recently, there were needs to make changes to js.bzl and, due
to its bad formatting, a single line change in js.bzl makes the overzealous
autoformatter change many lines which is quite bad for code review. This
change only auto-formats the js.bzl with buildifier for future edits.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
       - name: 'Lint BUILD files'
         # TODO(tensorboard-team): address all lint warnings and remove the exemption.
         run:
-          git ls-files -z '*BUILD' | xargs -0 buildifier --mode=check --lint=warn
+          git ls-files -z '*BUILD' third_party/js.bzl | xargs -0 buildifier --mode=check --lint=warn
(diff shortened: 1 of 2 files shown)


.git-blame-ignore-revs
```diff
@@ -84,3 +84,6 @@ ae32a4adcab8c3349c3bff21240078334fef1371  # black: reformat directory tensorboar
 
 # Small BUILD reformat (<https://github.com/tensorflow/tensorboard/pull/3054>)
 ea4e9bbc885784a283b8b79974132df7c6cdcc50  # Reformat all `*BUILD` files with `buildifier`
+
+# Reformat third_party/js.bzl (<https://github.com/tensorflow/tensorboard/pull/3332>)
+97b2535b28eae5318836ba1a427e0f81dcfbdaed  # Reformat third_party/js.bzl with `buildifier`
```
wchargin: This hash will change at rebase, so need to send a separate PR.

PR #3335, opened 2020-03-05
Fix resin compat bug with graph dashboard

psybuzz:
When running TensorBoard with Polymer Resin, the graph dashboard's
op details pane will have no data. Resin blocks the 'tf-graph-info'
from passing a 'node-name' property to the 'tf-node-info' component,
breaking the UI.

This changes the 'nodeName' property to 'graphNodeName', which Resin
does not treat as unsafe. Note that other properties like 'title'
are Resin-approved because of its whitelist of property contracts.
See https://github.com/Polymer/polymer-resin/blob/master/lib/contracts/contracts.js


psybuzz:
Test plan:
I've manually tested this using a variant of TensorBoard with and without Polymer Resin enabled, and saw that the Graph's details pane in the top right now has data.

I've also searched and found 0 other instances of "nodeName" as a Polymer property in the codebase.

PR #3336, opened 2020-03-05
ng: implement reload button and info icon

stephanwlee:
Polymer based TensorBoard renders info icon that brings user to our
documentation on GitHub and reload button for manual reload. This change
enables both in ng tensorboard.

**Screenshot**
image

### DIFF
--- a/tensorboard/webapp/header/header_component.ts
+++ b/tensorboard/webapp/header/header_component.ts
@@ -20,9 +20,18 @@ import {Component} from '@angular/core';
     <mat-toolbar color="primary">
       <span class="brand">TensorBoard</span>
       <plugin-selector class="plugins"></plugin-selector>
-      <span class="settings">
-        <settings-button></settings-button>
-      </span>
+      <app-header-reload></app-header-reload>
+      <settings-button></settings-button>
+      <a
+        class="readme"
+        mat-icon-button
+        href="https://github.com/tensorflow/tensorboard/blob/master/README.md"
+        rel="noopener noreferrer"
+        target="_blank"
+        aria-label="Help"
+      >
+        <mat-icon svgIcon="help_outline_24px"></mat-icon>
+      </a>
     </mat-toolbar>
   `,
   styles: [
@@ -36,7 +45,9 @@ import {Component} from '@angular/core';
       }
 
       .brand,
-      .settings {
+      .readme,
+      app-header-reload,
+      settings-button {
         flex: 0 0 auto;
       }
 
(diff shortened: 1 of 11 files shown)


tensorboard/webapp/header/header_component.ts
```diff
@@ -36,7 +44,9 @@ import {Component} from '@angular/core';
       }
 
       .brand,
-      .settings {
+      settings-button,
+      app-header-reload,
+      a {
```
bmd3k: It's potentially confusing to have selectors solely on common tag names.  I would prefer to introduce a class and refer to that instead. On the other hand, if this is a common pattern in the code base and you expect size of each of these components to be small, then potential for future confusion may be limited.

In this case I'm thinking about the "a" selector. but the component-tag selectors seems fine ("settings-button" and "app-header-reload") since you can be pretty assured they will be unique.

tensorboard/webapp/header/header_test.ts
```diff
@@ -157,4 +165,77 @@ describe('header test', () => {
...
+        state: DataLoadState.LOADING,
+        lastLoadedTimeInMs: null,
+      });
+      store.refreshState();
+      fixture.detectChanges();
+
+      const buttonAfter = fixture.debugElement.query(
+        By.css('app-header-reload button')
+      );
+      expect(buttonAfter.classes['loading']).toBe(true);
```
bmd3k: Is it also worth testing that it stops spinning?

tensorboard/webapp/header/reload_container.ts
```diff
@@ -0,0 +1,94 @@
...
+      (click)="triggerReload()"
+      [title]="getReloadTitle(lastLoadedTimeInMs$ | async | date: 'medium')"
+    >
+      <mat-icon svgIcon="refresh_24px"></mat-icon>
+    </button>
+  `,
+  styles: [
+    `
+      button,
+      mat-icon {
```
bmd3k: Same: selector on "button" and "mat-icon" are potentially confusing but potential is lowered if size of components remains small.

tensorboard/webapp/header/header_component.ts
stephanwlee: I am aware of the general principle when writing CSS, but I think that does not apply to Angular and Polymer. The CSS/SCSS is compiled to be scoped by a component so technically, the confusion should almost never happen unless you have more than one instance of a tag.

I can be convinced that we still need semantic names around tags for semantically meaningful selectors.

bmd3k: I'm aware that selectors are scoped. It is nonetheless potentially confusing. I'm thinking specifically about the case where someone modifies this component to add another anchor tag. (or modifies the reload_container to add a button or icon).

But if we can assume that these components will be limited in scope or rarely modified then likelihood of confusion decreases. Making changes here is up to you.

stephanwlee: Done.

Do note, however, that a given component should not be very large in general. When a component becomes unwieldy, one must consider breaking it down to smaller components for readability.

PR #3339, opened 2020-03-06
ci: tighten Pip installs for linting

wchargin:
Summary:
The `flake8`, `black`, and `yamllint` workflows currently install all
dev dependencies even though they only need their respective linters.
This pulls in a bunch of stuff, like `grpcio-testing`, `docker`, and
`boto3`, and takes about 30 seconds per job. It should be significantly
faster to only install the distributions that are actually required.

Test Plan:
Check that the “Install flake8” (etc.) workflow steps now take only a
handful of seconds rather than 25–30 seconds.

wchargin-branch: ci-tight-lint-installs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
       - name: 'Install flake8'
         run: |
           python -m pip install -U pip
-          pip install -r ./tensorboard/pip_package/requirements_dev.txt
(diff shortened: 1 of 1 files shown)


wchargin:
`flake8` is down to 5 seconds, `black` is down to 10 seconds, and
`yamllint` is down to 3 seconds.

PR #3343, opened 2020-03-09
[DebuggerV2] Support FULL_HEALTH debug mode; Display readable dtype string

caisq:
  * Continue developing DebuggerV2 plugin
  * Add "/dtypes_map" endpoint to serve a map from dtype enum to dtype name for the front end to use.
  * The front end requests the dtype map when loaded
  * Add FULL_HEALTH rendering logic to the ExecutionDataComponent
  * Under the FULL_HEALTH and SHAPE modes, where dtype is available, the front end uses the loaded dtype map to render a human readable dtype string (e.g., "float32" instead of the previous "1")
* Screenshots of UI changes
 * image
  * Unit tests added
  * Tested against two real tfdbg2 data logs under the FULL_HEALTH and SHAPE modes.


caisq:
cc @baileydesign

tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
```diff
@@ -277,3 +279,20 @@ def serve_stack_frames(self, request):
...
+        """Serves the map from dtype num value to dtype details.
+
+        Args:
+          request: HTTP request.
+
+        Returns:
+          Response to the request.
+        """
+        dtypes_map = {}
+        for dtype_enum_value, dtype_string in dtypes._TYPE_TO_STRING.items():
```
stephanwlee: this is using a private symbol of another module. Is this kosher?

tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
```diff
@@ -277,3 +279,20 @@ def serve_stack_frames(self, request):
             )
         except errors.NotFoundError as e:
             return _error_response(request, str(e))
+
+    @wrappers.Request.application
+    def serve_dtypes_map(self, request):
```
stephanwlee: did you think about the alternative? It is a bit weird for backend to regularly flush down the map like this. I can see two other designs:

1. transform other endpoints that flushes the dtype enum in the payload to the right value
2. frontend has this map; if we had jspb and used protobuf everywhere, this would have been very natural place to do the conversion, anyways.

caisq: Thanks for the suggestion! I think your option 1 makes sense. I can make changes in this PR, so that the endpoint serves debug tensor values are served as a JSON object such as 

```
{
  "tensor_id": 123,
  "dtype": "float32",
  "rank": 2,
  "size": 6
}
```
instead of a simple array such as `[123, 1, 2, 6]` from the end point.

This is going to be a pretty large change. But as you suggested, it's going to make the front-end code simpler and will get rid of the "/dtypes_map" endpoint. 

Let me know what you think.

stephanwlee: If you think that will bloat the response beyond reasonable bounds, do consider the second option, too. protos do not change anyways.

caisq: Option 2 will necessitate hard-coding of the dtypes in typescript (either by creating a new .ts file or using genrules).

I think option 1 you suggested is the most elegant
1. It doesn't at to the complexity of the build system
2. It simplifies the FE
3. The increase in response size is acceptible, because the size of the execution/data is protected against thanks to execution/digest anyway.

caisq: @stephanwlee After some further thinking, I decided that the current approach is reasonable and more appropriate compared to the two options you mentioned. The reason is as follows:

1. The human-readable string names of the dtypes (e.g., "float16") are not a part of the protobuf type definition it self (which is at https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/types.proto#L12). The string names of the proto types (e.g., "DT_FLOAT") is not the most human-friendly to show. Instead, the human-readable strings are defined at the Python level, at https://github.com/tensorflow/tensorboard/blob/master/tensorboard/compat/tensorflow_stub/dtypes.py#L480.

As such, it is natural to populate the map from Python. Doing so also leaves open the possibility of attaching additional properties of the types in the response in the future, e.g., including fields such as "is_floating", "is_integer", and "is_string", which are also not pieces of information directly available from the protobuf definition.

This is argument against option 2 above.

2. With regard to option 1 (encoding the semantics, e.g., "dtype" and "num_nans") in the response itself. This will cause the response format to deviate from the on-disk format for the execution data and therefore increase the complexity of the backend. Increase in the size of the response is another minor concern, as mentioned above. 

Based on these considerations, I'd like to stick with the current approach. In the meantime, I've slightly optimized the Python code in `debugger_v2_plugin.py`, so that the `dtypes_map` is populated once, when the plugin is created, instead of being created on the fly during every request.

PTAL.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -176,6 +180,28 @@ export class DebuggerEffects {
...
+        return dtypesMapLoadState.state !== DataLoadState.LOADING;
+      }),
+      tap(() => this.store.dispatch(dtypesMapRequested())),
+      mergeMap(() => {
+        return this.dataSource.fetchDtypesMap().pipe(
+          tap((dtypesMapResponse: DtypesMapResponse) => {
+            this.store.dispatch(dtypesMapLoaded(dtypesMapResponse));
+          }),
+          map(() => void null)
+          // TODO(cais): Add catchError() to pipe.
```
stephanwlee: What do you think should happen when the failure happens? Does entire UI break when this happens?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -19,6 +19,8 @@ export {DataLoadState, LoadState} from '../../../../webapp/types/data';
 
 export const DEBUGGER_FEATURE_KEY = 'debugger';
 
+export type DtypesMap = {[dtype_enum_value: string]: {name: string}};
```
stephanwlee: camelCase for `dtypeEnumValue`.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/execution_data/execution_data_container.ts
```diff
@@ -94,5 +102,47 @@ export class ExecutionDataContainer {
...
+                  ? tensorValue[2] // tensor_debug_mode: FULL_HEALTH
+                  : tensorValue[1] // tensor_debug_mode: SHAPE
+              );
+              dtypes.push(
+                dtypeEnum in dtypesMap
+                  ? dtypesMap[dtypeEnum].name
+                  : UNKNOWN_DTYPE_NAME
+              );
+            }
+          }
```
stephanwlee: consider using `map` instead since it is more semantically meaningful than `for-of` and `push`.

caisq: No longer relevant now we're using the genrule.

caisq:
@stephanwlee I forgot to mention another downside of this genrule-based approach. VSCode gets confused about the types. If you know of a way to fix that, I'd love to hear it.

stephanwlee:
Noted. There is no simple way since we do not want VSCode to see the guts of the bazel-out (generated file goes in there).

PR #3344, opened 2020-03-09
Experimental plugins specified using ExperimentalPluginLoader

bmd3k:

Allow an administrator of a Tensorboard to mark certain plugins as "experimental" and hidden to users by default. Users would then need to add query parameter "experimentalPlugin=" to the URL to see the experimental plugin.

Introduce simpler marker class ExperimentalPlugin which acts as an annotation on either TBPlugin or TBLoader that indicates a plugin is experimental. TensorBoardWSGIApp() uses isinstance() to determine if a plugin is experimental and passes the information as a list of names, "experimental_plugins", in its call to TensorBoardWSGI()

TensorBoardWSGI._serve_plugins_listing() filters out plugins that are included in experimental_plugins but are not specified in an experimentalPlugin query parameter.

There is logic in createRouter() that forward "experimentalPlugin" query parameter values from the original TensorBoard request to the call to the plugins_listing endpoint.

* Screenshots of UI changes

Using an example where "images" and "graphs" are marked as experimental.

No query parameters:
image

"experimentalPlugin=images" query parameter:
image

"experimentalPlugin=images&experimentalPlugin=graphs" query parameters:
image

Modify the list of plugins in default.py to use ExperimentalPlugin. Modify Images, Graphs,a nd Debugger plugins (a mix of TBPlugin and TBLoader):

class ExperimentalImagePlugin(
    images_plugin.ImagesPlugin,
    application.ExperimentalPlugin
):
    pass

class ExperimentalGraphsPlugin(
    graphs_plugin.GraphsPlugin,
    application.ExperimentalPlugin,
):
    pass

class ExperimentalDebuggerPluginLoader(
    debugger_plugin_loader.DebuggerPluginLoader,
    application.ExperimentalPlugin
):
    pass

Start server.

Open "http://localhost:6006/" without query parameters. Note that "Images" and "Graphs" do not appear as plugins.

Open "http://localhost:6006/?experimentalPlugin=images&experimentalPlugin=graphs&experimentalPlugin=debugger". Note that "Images", "Graphs", and "Debugger" appear as plugins.

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -256,7 +256,9 @@ def TensorBoardWSGIApp(
         if plugin is None:
             continue
         tbplugins.append(plugin)
-        if isinstance(loader, ExperimentalPluginLoader):
+        if isinstance(loader, ExperimentalPlugin) or isinstance(
+            plugin, ExperimentalPlugin
+        ):
             experimental_plugins.append(plugin.plugin_name)
         plugin_name_to_instance[plugin.plugin_name] = plugin
     return TensorBoardWSGI(
@@ -878,29 +880,32 @@ def make_plugin_loader(plugin_spec):
     raise TypeError("Not a TBLoader or TBPlugin subclass: %r" % (plugin_spec,))
 
 
-class ExperimentalPluginLoader(base_plugin.TBLoader):
-    """A wrapper around TBLoader to annotate a plugin as experimental.
+class ExperimentalPlugin(object):
+    """A marker class used to annotate a plugin as experimental.
 
-       Experimental plugins are hidden from users by default. The plugin will only
-       be enabled for a user if the user has specified the plugin with the
-       experimentalPlugin query parameter in the URL.
+    Experimental plugins are hidden from users by default. The plugin will only
+    be enabled for a user if the user has specified the plugin with the
+    experimentalPlugin query parameter in the URL.
 
-       ExperimentalPluginLoader is also a TBLoader. All calls are proxied to the
-       base_loader TBLoader passed at init.
+    The marker class can annotate either TBPlugin or TBLoader instances, whichever
+    is most convenient.
+
+    Typical usage is to create a new class that inherits from both an existing
+    TBPlugin/TBLoader class and this marker class. For example:
+
+
+    class ExperimentalGraphsPlugin(
+        graphs_plugin.GraphsPlugin,
+        application.ExperimentalPlugin,
+    ):
+        pass
+
+
+    class ExperimentalDebuggerPluginLoader(
+        debugger_plugin_loader.DebuggerPluginLoader,
+        application.ExperimentalPlugin
+    ):
+        pass
     """
 
-    def __init__(self, base_loader):
-        """Creates the loader, which proxies all calls to the underlying base_loader.
-
-        :param base_loader: :class:`TBLoader`
-        """
-        self.base_loader = base_loader
-
-    def define_flags(self, parser):
-        self.base_loader.define_flags(parser)
-
-    def fix_flags(self, flags):
-        self.base_loader.fix_flags(flags)
-
-    def load(self, context):
-        return self.base_loader.load(context)
+    pass
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -89,6 +89,8 @@ PLUGIN_PREFIX = "/plugin"
 PLUGINS_LISTING_ROUTE = "/plugins_listing"
 PLUGIN_ENTRY_ROUTE = "/plugin_entry.html"
 
+EXPERIMENTAL_PLUGINS_QUERY_PARAM = "expplugin"
+
 # Slashes in a plugin name could throw the router for a loop. An empty
 # name would be confusing, too. To be safe, let's restrict the valid
 # names as follows.
@@ -114,13 +116,17 @@ def _apply_tensor_size_guidance(sampling_hints):
     return tensor_size_guidance
 
 
-def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
+def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider, experimental_plugins=[]):
     """Construct a TensorBoardWSGIApp with standard plugins and multiplexer.
 
     Args:
       flags: An argparse.Namespace containing TensorBoard CLI flags.
       plugin_loaders: A list of TBLoader instances.
       assets_zip_provider: See TBContext documentation for more information.
+      experimental_plugins: A list of plugin names that are only provided
+        experimentally. The corresponding plugins will only be activated for
+        a user if the user has specified the plugin with the expplugin query
+        parameter in the URL.
 
     Returns:
       The new TensorBoard WSGI application.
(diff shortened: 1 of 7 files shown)


tensorboard/backend/application.py
```diff
@@ -849,3 +876,31 @@ def make_plugin_loader(plugin_spec):
         if issubclass(plugin_spec, base_plugin.TBPlugin):
             return base_plugin.BasicLoader(plugin_spec)
     raise TypeError("Not a TBLoader or TBPlugin subclass: %r" % (plugin_spec,))
+
+
+class ExperimentalPluginLoader(base_plugin.TBLoader):
```
bmd3k: Do you have a recommendation about where this class could live? I've put it here just to get the ball rolling but suspect it would be preferable to move elsewhere.

tensorboard/backend/application.py
```diff
@@ -849,3 +876,31 @@ def make_plugin_loader(plugin_spec):
...
+    """A wrapper around TBLoader to annotate a plugin as experimental.
+
+       Experimental plugins are hidden from users by default. The plugin will only
+       be enabled for a user if the user has specified the plugin with the
+       experimentalPlugin query parameter in the URL.
+
+       ExperimentalPluginLoader is also a TBLoader. All calls are proxied to the
+       base_loader TBLoader passed at init.
```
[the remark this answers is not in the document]
bmd3k: Now that ExperimentalPlugin is a marker class there is no longer a need to do this.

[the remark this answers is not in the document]
bmd3k: Decided to go with the marker class suggestion. I hadn't considered the drawback of multiple proxies. It's also cleaner not having to worry about the actual proxying code.

Marker approach also means I can use it to annotate both TBPlugin and TBLoader instances. It's honestly sometimes way more convenient and cleaner to use this on TBPlugin. I think this is just a natural side effect of the fact that the TensorBoard object accepts a list of mixed TBPlugin and TBLoader types. And it still keeps the decision to make a plugin experimental the responsibility of the administrator.

tensorboard/backend/application.py
```diff
@@ -849,3 +876,31 @@ def make_plugin_loader(plugin_spec):
...
+    """
+
+    def __init__(self, base_loader):
+        """Creates the loader, which proxies all calls to the underlying base_loader.
+
+        :param base_loader: :class:`TBLoader`
```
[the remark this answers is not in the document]
bmd3k: Now that ExperimentalPlugin is a marker class there is no longer anything to type.

tensorboard/backend/application.py
```diff
@@ -849,3 +876,31 @@ def make_plugin_loader(plugin_spec):
         if issubclass(plugin_spec, base_plugin.TBPlugin):
             return base_plugin.BasicLoader(plugin_spec)
     raise TypeError("Not a TBLoader or TBPlugin subclass: %r" % (plugin_spec,))
+
+
+class ExperimentalPluginLoader(base_plugin.TBLoader):
+    """A wrapper around TBLoader to annotate a plugin as experimental.
+
+       Experimental plugins are hidden from users by default. The plugin will only
```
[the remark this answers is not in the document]
bmd3k: Done.

tensorboard/backend/application.py
bmd3k: Great. Left it here.

tensorboard/backend/application.py
```diff
@@ -878,29 +880,32 @@ def make_plugin_loader(plugin_spec):
...
-       experimentalPlugin query parameter in the URL.
+    Experimental plugins are hidden from users by default. The plugin will only
+    be enabled for a user if the user has specified the plugin with the
+    experimentalPlugin query parameter in the URL.
 
-       ExperimentalPluginLoader is also a TBLoader. All calls are proxied to the
-       base_loader TBLoader passed at init.
-    """
+    The marker class can annotate either TBPlugin or TBLoader instances, whichever
+    is most convenient.
```
wchargin: Cool; this is a nice side benefit, mitigating your verbosity concern.

tensorboard/backend/application.py
```diff
@@ -285,6 +303,9 @@ def __init__(self, plugins, path_prefix="", data_provider=None):
         self._plugins = plugins
         self._path_prefix = path_prefix
         self._data_provider = data_provider
+        self._experimental_plugins = frozenset(
+            experimental_plugins if experimental_plugins is not None else []
```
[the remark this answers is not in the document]
bmd3k: Done.

tensorboard/backend/application.py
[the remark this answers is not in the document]
wchargin: > Marker approach also means I can use it to annotate both TBPlugin and
> TBLoader instances

Yep, that is a nice side benefit. (The distinction between plugins and
loaders at various phases of the stack is sometimes confusing, but I
don’t think that this makes it more confusing.)

tensorboard/backend/application.py
```diff
@@ -177,8 +179,9 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
         start_reloading_multiplexer(
             multiplexer, path_to_run, reload_interval, flags.reload_task
         )
+
     return TensorBoardWSGIApp(
-        flags, plugin_loaders, data_provider, assets_zip_provider, multiplexer
+        flags, plugin_loaders, data_provider, assets_zip_provider, multiplexer,
```
wchargin: sorry, forgot to mention earlier: can we revert the churn to these two
lines? (Black should be fine with it either way, so there’s no need to
pollute the diff and blame.)

tensorboard/backend/application.py
nfelt: FWIW I've been trying to reduce what's going on in application.py (cf. #2573) so if you don't mind I'd have a slight preference for a separate module in backend/.

tensorboard/backend/application.py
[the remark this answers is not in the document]
nfelt: LGTM, the marker class works nicely here :)

tensorboard/backend/application.py
```diff
@@ -849,3 +876,34 @@ def make_plugin_loader(plugin_spec):
         if issubclass(plugin_spec, base_plugin.TBPlugin):
             return base_plugin.BasicLoader(plugin_spec)
     raise TypeError("Not a TBLoader or TBPlugin subclass: %r" % (plugin_spec,))
+
+
+class ExperimentalPlugin(object):
```
nfelt: Probably worth mentioning on the docstring that this marker class is itself experimental (i.e. that the class itself and how it's interpreted may have breaking changes in the future, so there are no guarantees around using it).

bmd3k: Done.

tensorboard/backend/application.py
bmd3k: Done. It also means I can make it visible to tensorboard:internal, which I think we want for now.

tensorboard/backend/experimental_plugin.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
wchargin: nit: 2020 :-)

bmd3k: Done.

PR #3349, opened 2020-03-09
ng: refactor theming for reusability

stephanwlee:
Absolute dependencies on "tb-theme" allows us to provide different theme
files for different applications.

Related change: cl/299960760

### DIFF
--- a/tensorboard/webapp/BUILD
+++ b/tensorboard/webapp/BUILD
@@ -156,15 +156,11 @@ tf_sass_library(
     ],
 )
 
-tf_sass_library(
+# TODO(stephanwlee): remove the alias when all usages, internal, too,
+# //tensorboard/webapp:theme are removed.
+alias(
     name = "theme",
-    srcs = [
-        "_palette.scss",
-        "_theme.scss",
-    ],
-    deps = [
-        ":angular_material_theming",
-    ],
+    actual = "//tensorboard/webapp/theme",
 )
 
 tf_sass_binary(
--- a/tensorboard/webapp/theme/_tb_theme.scss
+++ b/tensorboard/webapp/theme/_tb_theme.scss
@@ -13,8 +13,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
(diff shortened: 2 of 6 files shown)


tensorboard/webapp/theme/_tb_theme.scss
```diff
@@ -28,17 +28,16 @@ $_warn: mat-palette($mat-red);
...
   (
-    // Overriding mat-light-theme-foreground variables.
     text: mat-color($mat-gray, 900),
     secondary-text: mat-color($mat-gray, 700),
     disabled-text: mat-color($mat-gray, 600),
-
     // TB specific variable.
-    border: #ebebeb,
-    link: mat-color($mat-blue, 700)
+      border: #ebebeb,
```
psybuzz: super nit: dedent?

tensorboard/webapp/BUILD
```diff
@@ -156,15 +156,9 @@ tf_sass_library(
...
+alias(
     name = "theme",
-    srcs = [
-        "_palette.scss",
-        "_theme.scss",
-    ],
-    deps = [
-        ":angular_material_theming",
-    ],
+    actual = "//tensorboard/webapp/theme",
```
psybuzz: I don't fully understand this alias, isn't it the `actual` field exactly the same?

stephanwlee: It is used by the internal rule. This change guarantees no breakage when syncing.

tensorboard/webapp/theme/_tb_theme.scss
stephanwlee: That's.... actually prettier.

stephanwlee: https://github.com/tensorflow/tensorboard/pull/3350 :(

tensorboard/webapp/BUILD
psybuzz: In that case, can we please add a comment, so that nobody accidentally removes this in OSS?

PR #3350, opened 2020-03-09
cleanup: lint scss formats using prettier

stephanwlee:
🧼We now have scss files in the repo. Lint and pretty format them!

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
     "postinstall": "ngc -p angular-metadata.tsconfig.json",
     "build": "bazel build //...",
     "test": "ibazel test //...",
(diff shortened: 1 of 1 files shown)


wchargin:
I support this change. For past such changes, we’ve prettified the files
first, then checked in the lint enforcement separately, such that tests
pass at each commit we can mark the first one as ignored for blame.

stephanwlee:
@wchargin the fix is in #3349 which is why I am not bothering to fix the lint failure in the checks. I will officially ask for review when the PR lands.

wchargin:
Ah, my oversight—sorry! :-)

stephanwlee:
Sorry for not making that clearer in the description 🙇

PR #3358, opened 2020-03-11
Revert #3355 (Rename blob states) to do things in a better order

davidsoergel:


davidsoergel:
Abandoning in favor of go/tbpr/3363, which does the git revert thing properly.

PR #3368, opened 2020-03-13
Added S3_ENDPOINT variable

### LINKED ISSUE #3014: Support environment variable S3_ENDPOINT in tensorboard without tensorflow

thealphacod3r:
Fixes #3014


thealphacod3r:
@nfelt now take a look at this please

PR #3369, opened 2020-03-13
ng: allow palette to be inlined via genrule.

stephanwlee:
We have different instances of the TensorBoard that need to be themed
differently (colors and definitions of the colors). Because include_path
of sass_binary is getting deprecated, we decided to use genrules to
conditionally supply correct colors when building the SASS.

Sync: requires cl/300794214. 
End state: cl/300395379

### DIFF
--- a/tensorboard/webapp/theme/_tb_theme.template.scss
+++ b/tensorboard/webapp/theme/_tb_theme.template.scss
@@ -13,20 +13,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-@import '../angular_material_theming';
-@import './tb_palette';
+@import 'tensorboard/webapp/angular_material_theming';
+
+/***********************************************************
+ * VARIABLE FILE IS INLINED WITH GENRULE. DO NOT IMPORT IT.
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/theme/_tb_theme.template.scss
```diff
@@ -13,20 +13,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-@import '../angular_material_theming';
-@import './tb_palette';
+@import 'tensorboard/webapp/angular_material_theming';
+
+// VARIABLE FILE IS INLINED WITH GENRULE. DO NOT IMPORT IT.
```
caisq: 1. Can you add a little more info about what "VARIABLE FILE" is to this comment (e.g., what the variables are for? palette? what the file name is, _variable.scss)? 

2. Consider make this comment more salient by adding flanking lines like `//////////////////////////////////` and exclamation points.

stephanwlee: Do note that the name of the file, _variable.scss, is not required. It is only useful to the genrule.

PR #3373, opened 2020-03-14
Add message types and rpc method to support exporting BLOBs

caisq:
  * Support tbdev uploader's exporting of BLOBs (e.g., GraphDefs)

  * Add the following message types to tensorboard/uploader/proto/blob.proto:
    * `Blob`, with fields blob_id and state (using the existing `BlobState` message type in the same file)
    * `BlobSequence`, which is composed of `Blob`s
  * In tensorboard/uploader/exporter_service.proto:
    * Add the following message type:  `BlobSequencePoints`, this is in parallel to the existing `ScalarPoints` and `TensorPoints` message types in the same file.
    * Change the existing `points` (scalars), `tensors`, and to-be-added `blob_sequences` into a `oneof`.
    * Add the following rpc method to `TensorBoardExporterService`: `StreamBlobData`.

Co-author: wchargin@

### DIFF
--- a/tensorboard/uploader/proto/blob.proto
+++ b/tensorboard/uploader/proto/blob.proto
@@ -12,3 +12,14 @@ enum BlobState {
   // Object is finalized.
   CURRENT = 2;
 }
+
+message Blob {
+  // For Blobs that haven't started uploading, this is an empty string.
+  string blob_id = 1;
+
+  BlobState state = 2;
+}
+
+message BlobSequence {
+  repeated Blob blob = 1;
+}
--- a/tensorboard/uploader/proto/blob.proto
+++ b/tensorboard/uploader/proto/blob.proto
@@ -14,12 +14,17 @@ enum BlobState {
 }
 
 message Blob {
-  // For Blobs that haven't started uploading, this is an empty string.
+  // A non-empty ID for the blob.
   string blob_id = 1;
-
   BlobState state = 2;
 }
 
+message BlobSequenceEntry {
+  // Optional. If absent, this represents a "hole" in the sequence:
+  // there is expected to be a blob here, but upload has not started.
+  Blob blob = 1;
+}
+
 message BlobSequence {
-  repeated Blob blob = 1;
+  repeated BlobSequenceEntry entries = 1;
 }
--- a/tensorboard/uploader/proto/export_service.proto
+++ b/tensorboard/uploader/proto/export_service.proto
@@ -16,8 +16,8 @@ service TensorBoardExporterService {
   // Stream scalars for all the runs and tags in an experiment.
   rpc StreamExperimentData(StreamExperimentDataRequest)
       returns (stream StreamExperimentDataResponse) {}
-  // Stream BLOB as chunks for a given blob_id.
-  rpc StreamBlobData(StreamBlobRequest) returns (stream StreamBlobResponse) {}
+  // Stream blob as chunks for a given blob_id.
+  rpc StreamBlobData(StreamBlobDataRequest) returns (stream StreamBlobDataResponse) {}
 }
 
 // Request to stream the experiment_id of all the experiments owned by the
@@ -150,16 +150,16 @@ message StreamExperimentDataResponse {
     repeated int64 steps = 1;
     // Timestamp of the creation of this point.
     repeated google.protobuf.Timestamp wall_times = 2;
-    // Value of the BLOB sequence at this step / timestamp.
+    // Value of the blob sequence at this step / timestamp.
     repeated BlobSequence values = 3;
   }
 }
 
-message StreamBlobRequest {
+message StreamBlobDataRequest {
   string blob_id = 1;
 }
 
-message StreamBlobResponse {
+message StreamBlobDataResponse {
   // The bytes in this chunk.
   bytes data = 1;
   // The position in the blob where this chunk begins.
@@ -170,7 +170,7 @@ message StreamBlobResponse {
   reserved 3;
   // Indicates that this is the last chunk of the stream.
   bool final_chunk = 4;
-  // CRC32C of the entire blob.  Required, to protect against data corruption.
-  // This should be set only when finalize_object=True.
+  // CRC32C of the entire blob. This should be set when final_chunk=True, to
+  // protect against data corruption.
   fixed32 final_crc32c = 5;
 }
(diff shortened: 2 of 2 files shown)


tensorboard/uploader/proto/blob.proto
```diff
@@ -12,3 +12,14 @@ enum BlobState {
   // Object is finalized.
   CURRENT = 2;
 }
+
+message Blob {
```
ericdnielsen: Blob/BlobSequence seem very minimal here, is the plan to add the other fields as you work through the implementation?

tensorboard/uploader/proto/export_service.proto
```diff
@@ -103,10 +106,18 @@ message StreamExperimentDataResponse {
   string run_name = 2;
   // The metadata of the tag `tag_name`.
   .tensorboard.SummaryMetadata tag_metadata = 3;
+
   // Data to store for the tag `tag_name.
-  ScalarPoints points = 4;
-  // Tensor data to store.
-  TensorPoints tensors = 5;
+  oneof data {
```
ericdnielsen: I'm surprised that no other code needs to change as a result of this nesting.  There are no tests that access the existing scalar points subfield?

tensorboard/uploader/proto/blob.proto
caisq: They are minimal, but they satisfy the need of the takeout task.

Other fields can be added to these message types in the future when the need arises.

tensorboard/uploader/proto/export_service.proto
caisq: I understand your point. `StreamExperimentDataResponse` has no usage in the OSS codebase, but tests of it do exists internally. Those tests will be run during sync.

tensorboard/uploader/proto/blob.proto
```diff
@@ -12,3 +12,14 @@ enum BlobState {
...
+  // For Blobs that haven't started uploading, this is an empty string.
+  string blob_id = 1;
```
wchargin: (Copied from internal discussion.) This doesn’t feel right to me. `Blob`
and `BlobSequence` are not just portions of responses from particular
export RPCs. They’re resources: <https://aip.dev/121>. Resources are
supposed to be well defined by themselves, and have a hierarchical
relationship. As written here, a `BlobSequence` could contain two copies
of `Blob(id="", state=UNFINALIZED)` to represent two different holes.
But those two `Blob` entries don’t really represent blobs—just the
opposite!—and they’re certainly not “the same thing”, despite being
structurally identical.

If we were designing this in a sane type system, surely we would write
something like:

```rust
struct BlobSequence {
  blobs: Vec<Option<Blob>>,
}
struct Blob {
  id: Uuid,  // not optional, not empty
  state: BlobState,
}
enum BlobState { Unfinalized, Current }
```

The simplest way to express that in the proto IDL is something like:

```proto
message BlobSequence {
  repeated BlobSequenceEntry entries = 1;
}
message BlobSequenceEntry {
  // Optional. If absent, this represents a "hole" in the sequence:
  // there is expected to be a blob here, but upload has not started.
  Blob blob = 1;
}
message Blob {
  string id = 1;
  BlobState state = 2;
}
```

This way, any `Blob` actually represents a blob, rather than just
coöpting the bit patterns of the `Blob` struct to encode something
totally different. Can we adopt this structure instead?

tensorboard/uploader/proto/export_service.proto
```diff
@@ -15,6 +16,8 @@ service TensorBoardExporterService {
   // Stream scalars for all the runs and tags in an experiment.
   rpc StreamExperimentData(StreamExperimentDataRequest)
       returns (stream StreamExperimentDataResponse) {}
+  // Stream BLOB as chunks for a given blob_id.
```
wchargin: Please downcase “blob” as in the rest of TensorBoard and Blobstore.
(I recognize that it’s sometimes taken to stand for “binary large
object”, but we don’t use it that way.)

tensorboard/uploader/proto/export_service.proto
```diff
@@ -15,6 +16,8 @@ service TensorBoardExporterService {
...
+  rpc StreamBlobData(StreamBlobRequest) returns (stream StreamBlobResponse) {}
```
wchargin: Request/response type names should match method name: add `Data`.

tensorboard/uploader/proto/export_service.proto
```diff
@@ -133,4 +144,33 @@ message StreamExperimentDataResponse {
...
+  // This must equal the sum of the sizes of the chunks sent so far.  Ignored
+  // if no data is provided.
+  int64 offset = 2;
+  // TODO(b/150443776): Add `crc32a = 3;` once the server can populate it.
+  reserved 3;
+  // Indicates that this is the last chunk of the stream.
+  bool final_chunk = 4;
+  // CRC32C of the entire blob.  Required, to protect against data corruption.
+  // This should be set only when finalize_object=True.
+  fixed32 final_crc32c = 5;
```
wchargin: This is a bit confusing; it can’t be both required and only set
sometimes. Given that in proto3 it’s not possible to distinguish between
a `fixed32` of value 0 and a missing `fixed32`, perhaps what’s actually
meant here is that when `final_chunk` is `true`, this always represents
a CRC-32 (even if its value happens to be `0`), and that otherwise the
value is not expected to be set and should be ignored?

(Also, presumably `finalize_object` is meant to be `final_chunk`?)

tensorboard/uploader/proto/export_service.proto
caisq: Done.

tensorboard/uploader/proto/export_service.proto
caisq: Done.

tensorboard/uploader/proto/export_service.proto
caisq: Ack. Thanks for the detailed explanation, @wchargin

tensorboard/uploader/proto/blob.proto
caisq: OK. The suggested way of dealing with holes n the blob sequence makes sense to me. Adopted with minor modifications (doc strings and ordering).

tensorboard/uploader/proto/export_service.proto
caisq: This is copied from an internal equivalent proto. I've fixed the doc-string issues you raised. The corresponding issues in the internal origin of this code will go away once this is sync'ed and the internal code is deduplicated in favor of this one.

tensorboard/uploader/proto/blob.proto
wchargin: Of course—thanks!

PR #3383, opened 2020-03-17
[uploader] Move BlobSequence to a nested location to avoid name clash

caisq:
  * Fix a proto name clash discovered during sync

  * `BlobSequence` is already defined in another .proto file, in the same namespace. To avoid the clash and the resultant test failure, move it to a nested location first. Later we can work on deduplication internally.


caisq:
@wchargin Sounds good.

PR #3386, opened 2020-03-17
hparams: write valid tensor summaries

wchargin:
Summary:
The summaries written by the hparams plugin only contain metadata, which
is conceptually fine but not valid at an implementation level: all
summary `Value`s must contain a `tensor`. Otherwise, the read paths get
into confusing states where `SummaryMetadata` can be present for a tag
but actually reading that tag’s `Tensors` raises `KeyError`. As of this
patch, hparams summaries include a rank-1 length-0 float tensor. We also
add a read-time `dataclass_compat` migration for legacy data.

Test Plan:
Calling `list_tensors` or `read_tensors` on a `MultiplexerDataProvider`
now properly returns hparams data. Before this patch, it returned
nothing as the summaries had no data class; if the null tensor backfill
is removed from this patch, it fails with a `KeyError`.

wchargin-branch: hparams-summary-tensors


tensorboard/plugins/hparams/summary.py
```diff
@@ -193,5 +193,14 @@ def _summary(tag, hparams_plugin_data):
...
-    summary.value.add(tag=tag, metadata=tf_metadata)
+    summary.value.add(
+        tag=tag, metadata=tf_metadata, tensor=_TF_NULL_TENSOR,
+    )
     return summary
+
+
+# Like `metadata.NULL_TENSOR`, but with the TensorFlow version of the
+# proto. Slight kludge needed to expose the `TensorProto` type.
+_TF_NULL_TENSOR = type(tf.compat.v1.Summary.Value().tensor).FromString(
```
nfelt: IMO a bit cleaner to do `type(tf.make_tensor_proto(0))` for this?

wchargin: Sure: done.

PR #3393, opened 2020-03-18
EngProd: Improve error message for fixing mismatched tf protos.

bileschi:
Improves an error message, giving better advice for how to fix the common issue wherein tf and tb proto common definitions have diverged.

Tested:
*  Deliberately caused a proto mismatch and witnessed the new error message.
*  Fixed the proto mismatched and saw the test passes.

### DIFF
--- a/tensorboard/compat/proto/proto_test.py
+++ b/tensorboard/compat/proto/proto_test.py
@@ -202,7 +202,7 @@ class ProtoMatchTest(tf.test.TestCase):
                 tb_string.splitlines(1),
                 tf_string.splitlines(1),
                 fromfile=tb_path,
-                tofile=tf_path
+                tofile=tf_path,
             )
             diff = "".join(diff)
 
--- a/tensorboard/compat/proto/proto_test.py
+++ b/tensorboard/compat/proto/proto_test.py
@@ -161,7 +161,7 @@ PROTO_REPLACEMENTS = [
 
 
 MATCH_FAIL_MESSAGE_TEMPLATE = """
-{} and {} did not match: {}
+{}
 
 NOTE!
 ====
@@ -172,7 +172,7 @@ time.
 The proper fix is:
 
 1. In your TensorFlow clone, check out the version of TensorFlow whose
-   protos you want to update (e.g., `git cehckout v2.2.0-rc0`)
+   protos you want to update (e.g., `git checkout v2.2.0-rc0`)
 2. In your tensorboard repo, run:
 
     ./tensorboard/compat/proto/update.sh PATH_TO_TENSORFLOW_REPO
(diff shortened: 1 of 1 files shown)


tensorboard/compat/proto/proto_test.py
```diff
@@ -160,7 +160,29 @@
...
+NOTE!
+====
+This is expected to happen when TensorFlow updates their proto definitions.
+We pin copies of the protos, but TensorFlow can freely update them at any
+time.
+
+The proper fix is:
+
+1. In your TensorFlow clone, check out the version of TensorFlow whose
+   protos you want to update (e.g., `git cehckout v2.2.0-rc0`)
```
wchargin: s/cehckout/checkout/

tensorboard/compat/proto/proto_test.py
```diff
@@ -185,7 +207,7 @@ def test_each_proto_matches_tensorflow(self):
             self.assertEquals(
                 diff,
                 "",
-                "{} and {} did not match:\n{}".format(tf_path, tb_path, diff),
+                MATCH_FAIL_MESSAGE_TEMPLATE.format(tf_path, tb_path, diff),
             )
```
wchargin: As written, this prints the diff twice (output below). It looks like the
intent of this format string was to display the “NOTE!” below the diff
because the note is more actionable. That makes sense to me. In that
case, can we change the call site to just

```python
            if diff:
                self.fail(MATCH_FAIL_MESSAGE_TEMPLATE.format(diff))
```

?

<details>
<summary>Current test failure output</summary>

```
FAIL: test_each_proto_matches_tensorflow (__main__.ProtoMatchTest)
test_each_proto_matches_tensorflow (__main__.ProtoMatchTest)
----------------------------------------------------------------------
  File "/RUNFILES_PATH/tensorboard/compat/proto/proto_test.py", line 210, in test_each_proto_matches_tensorflow
    MATCH_FAIL_MESSAGE_TEMPLATE.format(tf_path, tb_path, diff),
AssertionError: 
tensorflow.core.framework.allocation_description_pb2 and tensorboard.compat.proto.allocation_description_pb2 did not match: --- 
+++ 
@@ -43,7 +43,7 @@
   java_package: "org.tensorflow.framework"
   java_outer_classname: "AllocationDescriptionProtos"
   java_multiple_files: true
-  go_package: "github.com/tensorflow/tensorflow/tensorflow/go/core/framework"
+  go_package: "github.com/tensorflow/tensorflow/tensorflow/go/core/framework/allocation_description_go_proto"
   cc_enable_arenas: true
 }
 syntax: "proto3"


NOTE!
====
This is expected to happen when TensorFlow updates their proto definitions.
We pin copies of the protos, but TensorFlow can freely update them at any
time.

The proper fix is:

1. In your TensorFlow clone, check out the version of TensorFlow whose
   protos you want to update (e.g., `git cehckout v2.2.0-rc0`)
2. In your tensorboard repo, run:

    ./tensorboard/compat/proto/update.sh PATH_TO_TENSORFLOW_REPO

3. Review and commit any changes.
:
- --- 
- +++ 
- @@ -43,7 +43,7 @@
-    java_package: "org.tensorflow.framework"
-    java_outer_classname: "AllocationDescriptionProtos"
-    java_multiple_files: true
- -  go_package: "github.com/tensorflow/tensorflow/tensorflow/go/core/framework"
- +  go_package: "github.com/tensorflow/tensorflow/tensorflow/go/core/framework/allocation_description_go_proto"
-    cc_enable_arenas: true
-  }
-  syntax: "proto3"


----------------------------------------------------------------------
Ran 2 tests in 0.003s
```

</details>

tensorboard/compat/proto/proto_test.py
```diff
@@ -160,7 +160,29 @@
 ]
 
 
+MATCH_FAIL_MESSAGE_TEMPLATE = """
+{} and {} did not match: {}
```
wchargin: Need a newline between `match:` and `{}`, as in the original format
string, or the diff will appear wonky (the leading `---` will appear a
line early). 

But I think that it would be easier to read the error if we just let
`difflib` handle the filenames, by passing `fromfile=tf_path` and
`tofile=tb_path` kwargs to [`difflib.unified_diff`][ud]. Then the diff
headers will properly encode the paths, like this:

```diff
--- tensorflow.core.framework.allocation_description_pb2
+++ tensorboard.compat.proto.allocation_description_pb2
@@ -43,7 +43,7 @@
   java_package: "org.tensorflow.framework"
   java_outer_classname: "AllocationDescriptionProtos"
   java_multiple_files: true
-  go_package: "github.com/tensorflow/tensorflow/tensorflow/go/core/framework"
+  go_package: "github.com/tensorflow/tensorflow/tensorflow/go/core/framework/allocation_description_go_proto"
   cc_enable_arenas: true
 }
 syntax: "proto3"
```

[ud]: https://docs.python.org/3/library/difflib.html#difflib.unified_diff

bileschi: Done, I also updated the variable names for clarity, replacing actual / expected with tf / tb

PR #3347, opened 2020-03-09
bug: fix the tooltip not disappearing issue

stephanwlee:
Fixes #3282.

Repro case: when tooltip is shown, slowly move the mouse to an edge of a
chart. We expect the tooltip the disappear when the cursor is on the
edge of the chart.

Cause:
1. For Polymer 2 and its Shadow DOM compatibility, TensorBoard opted out of the
event delegation of Plottable. Plottable, by default, attaches a set of event
listeners on document.body and appropriate callbacks depending on the
circumstances. In Shadow DOM, the event re-targetting broke (harder to identify
`event.target`) so TensorBoard, instead, attaches event listeners on every
Plottable container, SVGs.

2. When mouse leaves (mouseout) the container, Plottable remaps the event as
mouse move and calculate whether the cursor is inside a component
(Interaction.prototype._isInsideComponent, specifically) to trigger appropriate
callback. The method, however is flawed since it returns, for a component that
is at for instance (0, 0) with size of (100, 100), true when pointer is at
(100, 100). It should only return true for [0, 100), instead.  As a result, the
mouseout event occurred at (100, 100) was treated as an event inside the
component but all the subsequent mouse movements are not captured since they
are events that occurred outside of the event target.  In vanilla Plottable,
this bug do not manifest since event delegation on the entire document will
eventually trigger mouse out when cursor is at, for instance, (101, 100).

Fix:
Overwrote the method, _isInsideComponent, with the correct implementation.

Issue #3282, opened 2020-02-22
plot hover popup remains if mouse is moved slowly out of the plot through the right side of a plot


santiontanon:
## Environment information (required)

Please run `diagnose_tensorboard.py` (link below) in the same
environment from which you normally run TensorFlow/TensorBoard, and
paste the output here:

### Diagnostics

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version d515ab103e2b1cfcea2b096187741a0eeb8822ef

--- check: general
INFO: sys.version_info: sys.version_info(major=2, minor=7, micro=17, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): ('Linux', '[BLINDED]', '5.2.17-1rodete3-amd64', '#1 SMP Debian 5.2.17-1rodete3 (2019-10-21 > 2018)', 'x86_64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: False
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/lib/python2.7/dist-packages/pip/__main__.py", line 16, in <module>
    from pip._internal import main as _main  # isort:skip # noqa
  File "/usr/lib/python2.7/dist-packages/pip/_internal/__init__.py", line 40, in <module>
    from pip._internal.cli.autocompletion import autocomplete
  File "/usr/lib/python2.7/dist-packages/pip/_internal/cli/autocompletion.py", line 8, in <module>
    from pip._internal.cli.main_parser import create_main_parser
  File "/usr/lib/python2.7/dist-packages/pip/_internal/cli/main_parser.py", line 12, in <module>
    from pip._internal.commands import (
  File "/usr/lib/python2.7/dist-packages/pip/_internal/commands/__init__.py", line 6, in <module>
    from pip._internal.commands.completion import CompletionCommand
  File "/usr/lib/python2.7/dist-packages/pip/_internal/commands/completion.py", line 6, in <module>
    from pip._internal.cli.base_command import Command
  File "/usr/lib/python2.7/dist-packages/pip/_internal/cli/base_command.py", line 18, in <module>
[...]

PR #3398, opened 2020-03-19
TensorBoard 2.2.0

bileschi:


RELEASE.md
```diff
@@ -1,3 +1,40 @@
+# Release 2.2.0
+
+The 2.2 minor series tracks TensorFlow 2.2.
+
+## Features
+
+- Profile plugin now should be pip installed.
```
wchargin: Could you please provide here the package name that users should
install, and maybe slightly expand the expository text to communicate
the benefits of this change? (The new version works in Chrome 80 and
Firefox, has better model insights, will be more actively maintained,
etc.)

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+- Debugger:
+  - Added support for alerts (#3285)
+  - Support FULL_HEALTH debug mode (#3343)
+  - Integration with ngrx storage (#3359)
```
wchargin: Are these actually user-facing features yet? My understanding was that
the new UI is only available in “ng-tensorboard” and so isn’t actually
included in the Pip package. (Also, we try to keep the release notes
more user-understandable than “Integration with ngrx storage”, which
doesn’t really communicate why users would want this.)

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+- Add S3_ENDPOINT variable (#3347)
```
wchargin: This should be #3368, not #3347, and we generally credit contributions
from third-party contributors as ` - thanks @their-github-username` (see
examples below in this file).

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+- Confirm that the connection to tensorboard works or change to localhost (#2371)
+- Update --reload_multifile_inactive_secs default to 24 hours (#3243)
+
+## TensorBoard.dev updates
+
+- Improvements to [TensorBoard.dev] support:
+  - New `tensorboard dev update-metadata` command allows for updating the name and
+    description of experiments (#3277)
```
wchargin: We’ve henceforth used the “TensorBoard.dev updates” section to describe
changes that occur at the same time as the release but are actually only
server-side changes and thus won’t appear in the Pip package. The new
`update-metadata` subcommand is actually in this release, so can go in
the features section above.

Likewise for the #3307 entry below.

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+  - Handshake protocol allows for more graceful control of which plugins are supported
+    for upload (#3299)
+  - Improved organization of artifacts downloaded during export (#3307)
+  - Adds `--plugins` option to uploader (#3377)
```
wchargin: The `--plugins` option doesn’t have the requisite server-side support to
be wired up yet, so I’m not sure that we want to advertise this?

E.g., `tensorboard dev upload --plugins scalars graphs --logdir foo`
currently completes successfully but does not actually upload any graphs
and does not print any diagnostics to the user. This seems confusing.

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+  - Support for exporting blob data (#3373)
```
wchargin: We don’t yet have blob export support (the PR referenced here was just
the protos; the client and server haven’t been written yet), so let’s
strike this. We probably also want to bill this as just “support for
the graphs plugin” when that’s all ready rather than breaking out each
aspect; export support is just part of shipping graphs.

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+
+## Features
+
+- Profile plugin now should be pip installed.
+- Debugger:
+  - Added support for alerts (#3285)
+  - Support FULL_HEALTH debug mode (#3343)
+  - Integration with ngrx storage (#3359)
+- Add S3_ENDPOINT variable (#3347)
+- Confirm that the connection to tensorboard works or change to localhost (#2371)
```
wchargin: Please thank external contributor.

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+    description of experiments (#3277)
+  - Handshake protocol allows for more graceful control of which plugins are supported
+    for upload (#3299)
+  - Improved organization of artifacts downloaded during export (#3307)
+  - Adds `--plugins` option to uploader (#3377)
+  - Support for exporting blob data (#3373)
+
+## Bug fixes
+
+- Fix for the tooltip not disappearing issue (#3347)
```
wchargin: Please clarify what “the tooltip not disappearing issue” is: something
like “#3282 - Tooltips now properly disappear when the mouse pointer is
slowly moved out of the chart body (PR #3347)”.

In general, we try to make the release notes self-contained and helpful
to readers, which often means providing a bit more context than is in
the pull request or issue title.

Note also the reference to both the bug and the fix PR; see previous
“Bug fixes” sections for examples.

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+- hparams: write valid tensor summaries (#3386)
```
wchargin: OK; this is an internal fix that was not yet user-symptomatic, but the
fix does change what data is written to disk, so it’s fine with me to
include it. If we do include it, please provide some context: “Internal
fix: HParams summary protos now properly include tensor values (PR
#3386)”.

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+  - Added support for alerts (#3285)
+  - Support FULL_HEALTH debug mode (#3343)
+  - Integration with ngrx storage (#3359)
+- Add S3_ENDPOINT variable (#3347)
+- Confirm that the connection to tensorboard works or change to localhost (#2371)
+- Update --reload_multifile_inactive_secs default to 24 hours (#3243)
+
+## TensorBoard.dev updates
+
+- Improvements to [TensorBoard.dev] support:
```
wchargin: Let’s drop this level of nesting. The “Improvements to TensorBoard.dev
support” introductory bullet is helpful when we have multiple relevant
entries under the “Features” heading, but under the “TensorBoard.dev
updates” heading it’s just superfluous.

RELEASE.md
```diff
@@ -1,3 +1,40 @@
...
+  - New `tensorboard dev update-metadata` command allows for updating the name and
+    description of experiments (#3277)
+  - Handshake protocol allows for more graceful control of which plugins are supported
+    for upload (#3299)
```
wchargin: This seems like an internal change, not a user-facing change. Do we
expect users to be excited about this change, or even understand what it
means? If not, can we strike it?

.travis.yml
```diff
@@ -19,7 +19,7 @@ env:
     - BAZEL_SHA256SUM=6c50e142a0a405d3d8598050d6c1b3920c8cdb82a7ffca6fc067cb474275148f
   matrix:
     - TF_VERSION_ID=tensorflow==1.15.0
-    - TF_VERSION_ID=tf-nightly
+    - TF_VERSION_ID=tensorflow==2.2.0rc0
```
wchargin: Let’s use `tensorflow==2.2.0rc1`, since that’s out?

RELEASE.md
bileschi: Clarified

RELEASE.md
bileschi: Removed.

RELEASE.md
bileschi: Done

RELEASE.md
bileschi: Done.

RELEASE.md
bileschi: Removed.

RELEASE.md
bileschi: Removed

PR #3377, opened 2020-03-16
Add --plugins option to uploader

bmd3k:
* Motivation for features / changes

Allow uploader users to specify the plugins for which data should be uploaded. It has two-fold purpose: 
(1) Allow users to specify "experimental" plugins that would otherwise not be uploaded.
(2) Allow users to list a subset of launched plugins, if they do not want to upload all launched plugin data.

* Technical description of changes

Add "--plugins" command line option for upload subcommand. The information is sent in the ServerInfoRequest (aka the handshake) and may impact the list of plugins returned in ServerInfoResponse.plugin_control.allowed_plugins.

Sample usage:
tensorboard dev upload --logdir <log dir> --plugins scalars graphs histograms

Outside of these specific changes:

It's expected that supported servers will evaluate the list of plugins and decide whether it is valid. If valid, the server will respond with the entire list of plugins or perhaps a sublist. If the list is invalid then it will respond with a CompatibilityVerdict of VERDICT_ERROR and a useful detail message to print to console.

If --plugins is not specified then the server is expected to respond with a default list of plugins to upload.

PR #3402, opened 2020-03-20
Add --plugins option to uploader

bmd3k:
Note: This functionality has previously been merged then reverted. This PR includes a fix for the bug that required the revert.

Original: https://github.com/tensorflow/tensorboard/pull/3377
Revert: https://github.com/tensorflow/tensorboard/pull/3400

In addition to previous testing I also ran the following commands successfully:

`bazel run //tensorboard -- dev upload --logdir <some log dir> --plugins scalars graphs`
Note this only uploads scalars since the current prod TensorBoard.dev ignores "--plugins"

`bazel run //tensorboard -- dev list`
Creates experiment with id OzqFG8zQSwCP9q5ZWV2JsA

`bazel run //tensorboard -- dev delete --experiment_id OzqFG8zQSwCP9q5ZWV2JsA`

Motivation for features / changes
Allow uploader users to specify the plugins for which data should be uploaded. It has two-fold purpose:
(1) Allow users to specify "experimental" plugins that would otherwise not be uploaded.
(2) Allow users to list a subset of launched plugins, if they do not want to upload all launched plugin data.

Technical description of changes
Add "--plugins" command line option for upload subcommand. The information is sent in the ServerInfoRequest (aka the handshake) and may impact the list of plugins returned in ServerInfoResponse.plugin_control.allowed_plugins.

Sample usage:
tensorboard dev upload --logdir --plugins scalars graphs histograms

Outside of these specific changes:

It's expected that supported servers will evaluate the list of plugins and decide whether it is valid. If valid, the server will respond with the entire list of plugins or perhaps a sublist. If the list is invalid then it will respond with a CompatibilityVerdict of VERDICT_ERROR and a useful detail message to print to console.

If --plugins is not specified then the server is expected to respond with a default list of plugins to upload.

### DIFF
--- a/tensorboard/uploader/uploader_main.py
+++ b/tensorboard/uploader/uploader_main.py
@@ -158,7 +158,6 @@ def _define_flags(parser):
         default=None,
         help="Experiment description. Markdown format.  Max 600 characters.",
     )
-
     upload.add_argument(
         "--plugins",
         type=str,
@@ -743,7 +742,7 @@ def _get_intent(flags):
 
 def _get_server_info(flags):
     origin = flags.origin or _DEFAULT_ORIGIN
-    plugins = flags.plugins if hasattr(flags, "plugins") else []
+    plugins = getattr(flags, "plugins", [])
 
     if flags.api_endpoint and not flags.origin:
         return server_info_lib.create_server_info(
--- a/tensorboard/uploader/uploader_main.py
+++ b/tensorboard/uploader/uploader_main.py
@@ -159,6 +159,16 @@ def _define_flags(parser):
         help="Experiment description. Markdown format.  Max 600 characters.",
     )
 
+    upload.add_argument(
+        "--plugins",
+        type=str,
+        nargs="*",
+        default=[],
+        help="List of plugins for which data should be uploaded. If "
+        "unspecified then data will be uploaded for all plugins supported by "
+        "the server.",
+    )
+
     update_metadata = subparsers.add_parser(
         "update-metadata",
         help="change the name, description, or other user "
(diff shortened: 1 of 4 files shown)


tensorboard/uploader/uploader_main.py
```diff
@@ -733,9 +743,13 @@ def _get_intent(flags):
 
 def _get_server_info(flags):
```
bmd3k: Changes to fix the issues in pr 3344 are all in this function.

Happy to volunteer to improve test coverage for some of the code in this file. Would it be appropriate to create an Issue for that?

tensorboard/uploader/uploader_main.py
```diff
@@ -159,6 +159,16 @@ def _define_flags(parser):
         help="Experiment description. Markdown format.  Max 600 characters.",
     )
 
+    upload.add_argument(
```
wchargin: nit: Can we remove the blank line before this paragraph? I originally
misread this as a new top-level argument due to the spacing.

tensorboard/uploader/uploader_main.py
```diff
@@ -733,9 +743,13 @@ def _get_intent(flags):
 
 def _get_server_info(flags):
     origin = flags.origin or _DEFAULT_ORIGIN
+    plugins = flags.plugins if hasattr(flags, "plugins") else []
```
wchargin: Should be `getattr(flags, "plugins", [])`.

(`hasattr` just tries to do a `getattr`, then throws away the result and
returns true or catches an `AttributeError` exception and returns false.
Thus, the way it’s currently written does two lookups instead of one.)

tensorboard/uploader/uploader_main.py
wchargin: Yes, feel free to open an issue that `uploader_main.py` needs tests.
This goes hand in hand with turning `uploader_main.py` into more of a
library so that we can have a proper Python API, like

```python
import tensorboard

exp = tensorboard.dev.upload(logdir)
print(exp.url)
model.fit()
...
tensorboard.dev.delete(exp.experiment_id)
```

or something. It’s not obvious to me what the best way to test it is, so
opening an issue sounds good. thanks!

tensorboard/uploader/uploader_main.py
bmd3k: Done.

tensorboard/uploader/uploader_main.py
bmd3k: Done.

tensorboard/uploader/uploader_main.py
bmd3k: Done. https://github.com/tensorflow/tensorboard/issues/3404

PR #3405, opened 2020-03-21
favicon: check in source image and inline at build time

wchargin:
Summary:
Our favicon has hitherto been defined only as four copies of the same
base64 data URI scattered about the repo. This commit creates an actual
file and inlines it into the various destinations at build time with a
very small shell script.

Test Plan:
Files are unchanged, modulo Prettier formatting:

```shell
prettify() { yarn -s prettier --stdin --stdin-filepath index.html; }
check() (
    set -eu
    bazel build "$1" >/dev/null 2>&1
    git show "HEAD~:$1" | prettify | shasum -a 256
    prettify <"bazel-bin/$1" | shasum -a 256
)
```

```
$ check tensorboard/components/tensorboard.html
e6139eb8dab40c5082b326dcaef810ccc9bd3f4ec5fe1903e18cc13e1eb80226  -
e6139eb8dab40c5082b326dcaef810ccc9bd3f4ec5fe1903e18cc13e1eb80226  -
$ check tensorboard/webapp/index.html
56088876962e466665d0c2a522b1bb323d771e7bd8890f5826627804957c0640  -
56088876962e466665d0c2a522b1bb323d771e7bd8890f5826627804957c0640  -
```

Also, `git grep iVBOR` shows that there are no more occurrences of the
base64 string in the repository.

wchargin-branch: favicon-build


tensorboard/logo/BUILD
```diff
@@ -0,0 +1,19 @@
...
+
+sh_binary(
+    name = "inline_favicon",
+    srcs = ["inline_favicon.sh"],
+    data = [
+        "favicon.png",
+    ],
+    visibility = [
+        "//tensorboard:internal",
+    ],
```
stephanwlee: redundant since we have the default visibility on L4.

wchargin: Removed; thanks.

PR #3410, opened 2020-03-21
uploader: add blob exporting

caisq:
  * Support exporting of blob data from tensorboard.dev uploader

  * In exporter.py, branch the handling logic for `StreamExperimentDataResponse` protos
    depending on whether
    * `points` is available (for scalars), or
    * `blob_sequences` is available (for blob sequences)
  * For the blob sequences, add method to download the blob via rpc in chunks.

  * Unit tests added, see exporter_test.py
  * Manual testing against a locally running backend (cross reference with CL/301719321)


tensorboard/uploader/exporter.py
```diff
@@ -180,19 +210,121 @@ def _request_scalar_data(self, experiment_id, read_time):
...
+            blob_file_paths.append(seq_blob_file_paths)
+        return json_object
+
+    def _download_blob(self, blob_id, experiment_id):
+        """Download the blob via rpc.
+
+        Args:
+          blob_id: Id of the blob.
+          experiment_id: Id of the experiment where the blob belongs is
+            from.
```
wchargin: nit: “experiment where the blob belongs is from” → “enclosing experiment”?

tensorboard/uploader/exporter.py
```diff
@@ -162,8 +180,20 @@ def export(self, read_time=None):
...
+    def _request_json_data(self, experiment_id, read_time):
+        """Given experiment id, generates JSON data and destination file name.
+
+        The JSON data describes the run, tag, metadata, in addition to
+          - Actual data in the case of scalars
+          - Pointer to binary files in the case of blob sequences.
+
+        Args:
+          experiment_id: The id of the experiment to request data for.
+          read_time: Timestamp to use snapshot the Spanner database for reads.
```
wchargin: Is this an `int` milliseconds since epoch, `float` seconds since epoch,
`datetime.datetime` naive value in local time, `datetime.datetime` naive
value in UTC, `datetime.datetime` aware value, `timestamp_pb2.Timestamp`
proto, …?

Also, as in the docs for the protos themselves, we generally try to
avoid specifying exactly which storage mechanism is used.

Can resolve both of these by just following the existing precedent for
`read_time` docs in this module.

tensorboard/uploader/exporter.py
```diff
@@ -162,8 +180,20 @@ def export(self, read_time=None):
...
                     raise
 
-    def _request_scalar_data(self, experiment_id, read_time):
-        """Yields JSON-serializable blocks of scalar data."""
+    def _request_json_data(self, experiment_id, read_time):
+        """Given experiment id, generates JSON data and destination file name.
+
+        The JSON data describes the run, tag, metadata, in addition to
+          - Actual data in the case of scalars
+          - Pointer to binary files in the case of blob sequences.
```
wchargin: Please document that this method also has the side effect of actually
writing those files (this was not my expectation just from reading the
docs).

tensorboard/uploader/exporter.py
```diff
@@ -50,9 +50,18 @@
 _FILENAME_METADATA = "metadata.json"
 # Output filename for scalar data within an experiment directory.
 _FILENAME_SCALARS = "scalars.json"
+# Output filename for blob sequences data within an experimental directly.
```
wchargin: nit: “experimental directly” → “experiment directory”

tensorboard/uploader/exporter.py
```diff
@@ -50,9 +50,18 @@
...
+# This file does not contain the actual content of the blobs. Instead,
+# it holds `blob_file_paths` that point to the binary files that contain
+# the blobs' contents, in addition other metadata such as run and tag names.
```
wchargin: nit: missing *to* after “in addition”

tensorboard/uploader/exporter_test.py
```diff
@@ -227,6 +229,328 @@ def outdir_files():
             },
         )
 
+    def test_e2e_success_case_with_finished_blob_sequence_data(self):
```
wchargin: Can we try to have just one end-to-end test case that exercises as much
behavior as possible? It looks like we should be able to get all happy
paths: scalars, finished blob sequences, unfinished blob sequences, blob
sequences with holes.

End-to-end tests like this are great for coverage, but require a ton of
code for setup and mock implementations, so it’s much easier to read
(and debug failures) if we can do that only once on a carefully designed
input that covers everything. Once the end-to-end test convinces us that
the main wiring is correct, the rest can be unit-tested. For example,
note how comparatively simple all the error case `test_*` methods are,
spanning about 20 lines instead of about 100–200.

tensorboard/uploader/exporter_test.py
```diff
@@ -227,6 +229,328 @@ def outdir_files():
...
+                    state=blob_pb2.BlobState.BLOB_STATE_CURRENT,
+                )
+                blob_sequence.entries.append(
+                    blob_pb2.BlobSequenceEntry(blob=blob)
+                )
+                response.blob_sequences.values.append(blob_sequence)
+            yield response
+
+        def stream_blob_data(request, **kwargs):
+            raise grpc.RpcError("Error for testing")
```
wchargin: Please use `test_util.grpc_error` to more faithfully match the actual
errors that we see in prod.

caisq:
Also updated the PR title and description.

tensorboard/uploader/exporter.py
```diff
@@ -180,19 +227,120 @@ def _request_scalar_data(self, experiment_id, read_time):
...
                 u"run": response.run_name,
                 u"tag": response.tag_name,
                 u"summary_metadata": metadata,
-                u"points": {
-                    u"steps": list(response.points.steps),
-                    u"wall_times": wall_times,
-                    u"values": list(response.points.values),
-                },
             }
+            filename = None
```
wchargin: Does this cause a `KeyError` in the caller if the server returns a
response with neither `points` nor `blob_sequences` set—say, because
we’ve added tensor storage and it returns `tensors` instead?

Should we perhaps instead warn in that case and not yield, since the
`json_data` is unfinished and there’s no filename to write, anyway?

tensorboard/uploader/exporter.py
```diff
@@ -180,19 +227,120 @@ def _request_scalar_data(self, experiment_id, read_time):
...
+        # TODO(cais): Deduplicate with internal method perhaps.
+        experiment_dir = _experiment_directory(self._outdir, experiment_id)
+        blobs_dir = os.path.join(experiment_dir, _BLOBS_DIR)
+        if not os.path.isdir(blobs_dir):
+            _mkdir_p(blobs_dir)
+        request = export_service_pb2.StreamBlobDataRequest(blob_id=blob_id)
+        blob_abspath = os.path.join(
+            blobs_dir, _BLOB_FILE_PREFIX + blob_id + _BLOB_FILE_SUFFIX
+        )
+        with open(blob_abspath, "wb") as f:
```
wchargin: Can this be `_open_excl(blobs_abspath, "wb")` (adding a `mode` argument
to `_open_excl` is fine), or is there a reason that we need to be able
to silently stomp this file?

tensorboard/uploader/exporter.py
```diff
@@ -50,9 +51,18 @@
...
+# This file does not contain the actual contents of the blobs. Instead,
+# it holds `blob_file_path`s that point to the binary files that contain
+# the blobs' contents, in addition to other metadata such as run and tag names.
+_FILENAME_BLOB_SEQUENCES = "blob_sequences.json"
 
 logger = tb_logging.get_logger()
 
+_BLOBS_DIR = "blobs"
+_BLOB_FILE_PREFIX = "blob_"
+_BLOB_FILE_SUFFIX = ".bin"
```
wchargin: nit: Please collocate these constants with their friends rather than
being sundered by `logger = tb_logging.get_logger()`. (Also, imho it’s
clearer to have `_FILENAME_BLOBS_DIR`, `_FILENAME_BLOB_PREFIX`, etc.,
but if you really want the mixed-and-matched names I can live with it.)

tensorboard/uploader/exporter_test.py
```diff
@@ -227,6 +229,214 @@ def outdir_files():
...
+            self.assertEqual(kwargs["metadata"], grpc_util.version_metadata())
+
+            tag = "__default_graph__"
+            for run in ("train", "test"):
+                response = export_service_pb2.StreamExperimentDataResponse()
+                response.run_name = run
+                response.tag_name = tag
+                display_name = "%s:%s" % (request.experiment_id, tag)
+                response.tag_metadata.CopyFrom(
+                    test_util.scalar_metadata(display_name)
```
wchargin: It seems a bit confusing to use the scalar plugin metadata here,
especially because it has the wrong data class. Can we at least use a
metadata value with `data_class=summary_pb2.DATA_CLASS_BLOB_SEQUENCE`?

tensorboard/uploader/exporter_test.py
```diff
@@ -227,6 +229,214 @@ def outdir_files():
...
+                        # A finished blob sequence.
+                        blob = blob_pb2.Blob(
+                            blob_id="%s_blob" % run,
+                            state=blob_pb2.BlobState.BLOB_STATE_CURRENT,
+                        )
+                        blob_sequence.entries.append(
+                            blob_pb2.BlobSequenceEntry(blob=blob)
+                        )
+                    elif run == "test":
+                        # An unfinihsed blob.
```
wchargin: sp.: *unfinihsed* → *unfinished*

tensorboard/uploader/exporter_test.py
```diff
@@ -227,6 +229,214 @@ def outdir_files():
...
+                        blob_sequence.entries.append(
+                            # `blob` unspecified: An unfinished blob.
+                            blob_pb2.BlobSequenceEntry()
```
wchargin: This isn’t an unfinished blob; it’s a hole. It’s good to test for this;
can we also have a test for an actual unfinished blob: i.e., a blob with
state `blob_pb2.BLOB_STATE_UNFINALIZED`?

tensorboard/uploader/exporter.py
```diff
@@ -180,19 +227,120 @@ def _request_scalar_data(self, experiment_id, read_time):
...
+        blobs_dir = os.path.join(experiment_dir, _BLOBS_DIR)
+        if not os.path.isdir(blobs_dir):
+            _mkdir_p(blobs_dir)
+        request = export_service_pb2.StreamBlobDataRequest(blob_id=blob_id)
+        blob_abspath = os.path.join(
+            blobs_dir, _BLOB_FILE_PREFIX + blob_id + _BLOB_FILE_SUFFIX
+        )
+        with open(blob_abspath, "wb") as f:
+            try:
+                for response in self._api.StreamBlobData(request):
```
wchargin: All API calls need to have `metadata=grpc_util.version_metadata`. Please
fix, and consider testing for this, as the other API calls are tested.

tensorboard/uploader/exporter_test.py
```diff
@@ -227,6 +229,214 @@ def outdir_files():
...
+                        )
+                    response.blob_sequences.values.append(blob_sequence)
+                yield response
+
+        mock_api_client.StreamExperiments = mock.Mock(wraps=stream_experiments)
+        mock_api_client.StreamExperimentData = mock.Mock(
+            wraps=stream_experiment_data
+        )
+        mock_api_client.StreamBlobData.side_effect = [
```
wchargin: It’s especially hard for me to follow the logic of this stub because
it’s partially defined here and then overwritten with a _different_ mock
later. Can we instead just add the appropriate `test_util.grpc_error`(s)
to this list of side effects, with maybe a brief comment noting that the
first two blobs are for the first experiment and the last two errors for
the second experiment?

tensorboard/uploader/exporter.py
```diff
@@ -143,15 +152,24 @@ def export(self, read_time=None):
...
-                    data = self._request_scalar_data(experiment_id, read_time)
-                    for block in data:
+                data = self._request_json_data(experiment_id, read_time)
+                for block, filename in data:
+                    if filename == _FILENAME_SCALARS:
+                        filepath = scalars_filepath
+                    elif filename == _FILENAME_BLOB_SEQUENCES:
+                        filepath = blob_sequences_filepath
+                    else:
+                        continue
```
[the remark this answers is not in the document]
wchargin: > Note that a test in expoerter_test.py has to be revised, as it previously violated the assumption that each StreamExperimentDataResponse contains something (scalars or blob sequences)

Hmm, not exactly? Previously the code was, I think, robust to responses
from future servers that might not have `points` set, such as the ones
that we’re currently introducing. But this patch makes the code instead
raise an internal `KeyError` when talking to newer servers, if I’m
reading it correctly. It seems to me like the test can stay unchanged
and the code needs to change, [as mentioned in another comment][1].

[1]: https://github.com/tensorflow/tensorboard/pull/3410/files/54bfd6922e3c08bde6921c77d9dcb1e1899c6618#diff-be28af56e60b5d47b458dce7d9a705c6

tensorboard/uploader/exporter_test.py
caisq: Acknowledging the readability issue. The difficulty is that I can't find a way to combine return-value side effects and error-raising side effects in the same place. I've added a comment here and above to make it clearer that this will be overridden again below for testing the error condition.

tensorboard/uploader/exporter_test.py
caisq: Done adding a test condition for `BLOB_STATE_UNFINALIZED`.

tensorboard/uploader/exporter_test.py
wchargin: Hmm; this test passes for me:

```python
import unittest
from unittest import mock


class Test(unittest.TestCase):
    def test(self):
        m = mock.Mock()
        m.side_effect = [
            "first result",
            ValueError("second exception"),
            iter(("third", "stream", "result")),
        ]
        self.assertEqual(m(), "first result")
        with self.assertRaisesRegex(ValueError, "second exception"):
            m()
        self.assertEqual(list(m()), ["third", "stream", "result"])


if __name__ == "__main__":
    unittest.main()
```

Maybe I’m misunderstanding, but the [docs on `side_effect`][1] suggested
to me that the values of the `side_effect` iterable can be exceptions or
values independently of each other.

[1]: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

caisq: Thanks for bearing with me. Sorry - there was a superfluous item in the `list` for `side_effect` before. Now that I removed it, it's working. Updated this test. Sorry for the confusion.

Also added an assertion for the call arg. PTAL.

caisq:
@wchargin ping. Let me know if you'd like to take another look at the PR.

wchargin:
Yep, I’m taking another look, thanks.

tensorboard/uploader/exporter.py
```diff
@@ -276,8 +425,8 @@ def _mkdir_p(path):
             raise
 
 
-def _open_excl(path):
-    """Like `open(path, "x")`, but Python 2-compatible."""
+def _open_excl(path, mode="w"):
+    """Like `open(path, "x" + mode)`, but Python 2-compatible."""
```
wchargin: nit: This is cute :-) but `xw` is not a valid mode; `x` and `xb` are, so
maybe `open(path, mode.replace("w", "x"))` does the job instead?

tensorboard/uploader/exporter_test.py
```diff
@@ -212,6 +226,13 @@ def outdir_files():
         self.assertEqual(points, {})
         self.assertEqual(datum, {})
 
+        # Check that one of the blob_sequences data file is empty, because there
+        # no blob sequences in this experiment.
```
wchargin: Great; thanks!

tensorboard/uploader/exporter_test.py
```diff
@@ -227,6 +248,214 @@ def outdir_files():
...
+        summary_metadata = summary_pb2.SummaryMetadata.FromString(
+            base64.b64decode(datum.pop("summary_metadata"))
+        )
+        self.assertEqual(summary_metadata, expected_summary_metadata)
+        points = datum.pop("points")
+        self.assertEqual(points["steps"], [0])
+        self.assertEqual(points["wall_times"], [1571084520.862939144])
+        # `None` represents the blob that experienced error during downloading
+        # and hence is missing.
+        self.assertEqual(points["blob_file_paths"], [[None, None]])
```
wchargin: In the other tests we check `self.assertEqual(datum, {})` after reading
each field to verify that we didn’t get anything unexpected; did we mean
to do the same thing here and below? (Otherwise, I don’t see the reason
for using `pop` instead of just subscripting.)

wchargin:
(Also, I ran a test sync for you, and it passed.)

tensorboard/uploader/exporter.py
```diff
@@ -147,23 +158,55 @@ def export(self, read_time=None):
...
+                            _open_excl(os.path.join(experiment_dir, filename))
+                        )
+                        for filename in (
+                            _FILENAME_SCALARS,
+                            _FILENAME_BLOB_SEQUENCES,
+                        )
+                    }
+                    os.mkdir(os.path.join(experiment_dir, _DIRNAME_BLOBS))
+                    for block, filename in data:
+                        if file_handles[filename] is None:
```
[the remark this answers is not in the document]
caisq: Done.

tensorboard/uploader/exporter.py
```diff
@@ -180,19 +223,125 @@ def _request_scalar_data(self, experiment_id, read_time):
...
+                filename = _FILENAME_SCALARS
+            elif response.HasField("blob_sequences"):
+                json_data[u"points"] = self._process_blob_sequence_points(
+                    response.blob_sequences, experiment_id
+                )
+                filename = _FILENAME_BLOB_SEQUENCES
+            if filename:
+                yield json_data, filename
+            else:
+                logger.warn(
```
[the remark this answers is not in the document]
caisq: Yes, there are deprecation warnings before. Changed it to `logger.warning()`.

tensorboard/uploader/exporter.py
```diff
@@ -180,19 +223,125 @@ def _request_scalar_data(self, experiment_id, read_time):
...
+                    "Skipping a response from experiment-data stream "
+                    "due to the lack of any valid data type (such as "
+                    "scalars and blob sequences): run=%s, tag=%s"
+                    % (response.run_name, response.tag_name)
```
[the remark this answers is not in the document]
caisq: I prefer to keep this as a warning. If someone gets this warning after we roll out tensor support, it means their uploader is out of date, so it is actionable in that sense. I added a sentence in this warning to indicate that.
This is also a somewhat rare occurrence, because if a user has uploaded tensor data before, they should most likely already installed a new version.

tensorboard/uploader/exporter.py
caisq: Done.

tensorboard/uploader/exporter_test.py
caisq: Good point. I revised the new test code in this PR to do `pop()` and empty-dict assertion.

tensorboard/uploader/exporter.py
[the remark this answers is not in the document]
wchargin: Okay, fine with me with the new sentence; thanks!

PR #3411, opened 2020-03-21
[DebuggerV2] Add shim for loadMonaco() method & supporting 3rd-party libraries

caisq:
  * Roll forward go/tbpr/3374 with fixes.
  * Same as go/tbpr/3374, with the following exception: the `tf_web_library` BUILD targets for requirejs, monaco-editor (core) and monaco-languages are moved to a dedicated BUILD file (tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/source_code/monaco/BUILD) to facilitate sync'ing.
  * copybara change CL: CL/302218915
  * Final state: CL/302222581. Build passes, after some manual reverts deal with JSCompiler artifacts
  * Alternative: use copybara rules to make the requirejs, monaco-editor and monaco-languages rules BUILD internally.
    * Con: Those transformed targets are *not* used internally. This is potentially confusing to future developers. It is also an unnecessary maintenance overhead. 
    * Con: While using `tensorboard_html_binary` to wrap around third_party libraries may work for requirejs, which involves a single .js file, it is more tedious for monaco-editors and monaco-languages, which involve multiple separate files (5 for monaco-editors and 3 for monaco-languages). This would involve a large number of confusing rules added to the copybara file. Those rules would be useless except for making the build pass (see the previous Con item). Some of these files are not .js or .css files (e.g., the .ttf file), further increasing the complexity.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "@types/chai": "^4.2.7",
     "@types/jasmine": "^3.5.0",
     "@types/node": "^13.1.2",
+    "@types/requirejs": "^2.1.31",
     "@types/sinon": "^7.5.1",
     "chai": "^4.2.0",
     "prettier": "1.18.2",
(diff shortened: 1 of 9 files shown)


caisq:
Thanks for the review, @stephanwlee . I added a comment to the internal BUILD file so that it is non-empty.

PR #3416, opened 2020-03-23
hparams: follow standard `is_active` pattern

wchargin:
Summary:
Until now, the hparams plugin has been active only when hparams data is
present *and* the scalars plugin is loaded and active. This cross-plugin
dependency is unique and doesn’t fit well into the data provider world,
so we remove it. The hparams plugin is now active iff hparams data is
available, via the standard `data_plugin_names` API.

Correspondingly, we loosen the error handling in parts of the plugin
that rely on the scalars plugin being active. As long as the scalars
plugin is included in the TensorBoard instance, this should not be a
problem: TensorBoard core loads all plugins before accepting requests to
any of them. If the hparams plugin is used in an installation with no
scalars plugin, the errors will now be 404s instead of 500s. This seems
acceptable, as running without a scalars plugin is not a common pattern.

Test Plan:
The hparams plugin is still marked as active in logdirs that contain
hparams data and inactive in logdirs that don’t.

wchargin-branch: hparams-data-provider-is-active

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -165,10 +165,9 @@ def standard_tensorboard_wsgi(flags, plugin_loaders, assets_zip_provider):
             max_reload_threads=flags.max_reload_threads,
             event_file_active_filter=_get_event_file_active_filter(flags),
         )
-        if flags.generic_data != "false":
-            data_provider = event_data_provider.MultiplexerDataProvider(
-                multiplexer, flags.logdir or flags.logdir_spec
-            )
+        data_provider = event_data_provider.MultiplexerDataProvider(
+            multiplexer, flags.logdir or flags.logdir_spec
+        )
 
     if reload_interval >= 0:
         # We either reload the multiplexer once when TensorBoard starts up, or we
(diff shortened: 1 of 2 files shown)


wchargin:
I thought about this a bit. The `data_plugin_names` are a disjunction,
so listing `scalars` would cause the hparams plugin to be shown even
when only scalar data is available (i.e., almost all experiments). If we
want to emulate the old behavior, we could change `data_plugin_names` to
permit specifying a more flexible boolean function (e.g., a tree of
`Conjunction`/`Disjunction` nodes)—it’s still experimental, exactly so
that we can make changes like this. But I think that the new behavior is
fine, and maybe even better: if I’ve logged hparams metadata for an
experiment and just don’t have any scalars yet, it seems reasonable to
me to still show that metadata in the dashboard. What do you think?

stephanwlee:
>  if I’ve logged hparams metadata for an
experiment and just don’t have any scalars yet, it seems reasonable to
me to still show that metadata in the dashboard. What do you think?

That's actually true. Other way around does not make any sense so showing
hparams plugin only when hparams data is present makes sense.

wchargin:
Cool; sounds good, then. Thanks for the reviews!

PR #3417, opened 2020-03-23
hparams: encapsulate direct multiplexer access

wchargin:
Summary:
The hparams plugin makes three kinds of queries against the multiplexer:
listing hparams metadata, listing scalars metadata, and reading scalar
points. This commit encapsulates each such operation in a method so that
we can migrate them to use data provider APIs. For now, the underlying
implementations are unchanged.

Test Plan:
Running `git grep '\.multiplexer' tensorboard/plugins/hparams/` shows
only matches in `backend_context.py` and tests. Existing tests pass, and
all three views of the hparams plugin, including embedded scalar charts
and the data download links, still work with and without `--generic_data
true` (though in all cases the plugin actually still reads directly from
the multiplexer).

wchargin-branch: hparams-encapsulate-mux


wchargin:
@googlebot wake up -_-

PR #3418, opened 2020-03-23
hparams: thread experiment ID through data access

wchargin:
Summary:
All data access in the hparams plugin is now performed via methods that
require an experiment ID. The ID is currently unused except for an
assertion. In a subsequent change, these methods will switch to using
a data provider instead of the multiplexer, at which point the
experiment ID will be required.

Test Plan:
Unit tests suffice for the internals. To check the wiring at the plugin
level, note that all three views of the hparams plugin, including
embedded scalar charts and the download links, still render properly.

wchargin-branch: hparams-thread-eid


tensorboard/plugins/hparams/backend_context.py
```diff
@@ -53,11 +53,11 @@ def __init__(self, tb_context, max_domain_discrete_len=10):
             Typically, only tests should specify a value for this parameter.
         """
         self._tb_context = tb_context
-        self._experiment_from_tag = None
+        self._experiment_from_tag = {}  # experiment ID to Experiment proto
```
nfelt: This name doesn't really make sense any more now that it's a dict, IMO.  Something like `self._cached_experiments` would be clearer.

For that matter though, this is now an ever-growing cache and I don't think it's actually needed (or at least not worth the risk of unbounded memory, even if in practice it probably wouldn't be a very fast leak).  Could we just remove it?

wchargin: True about the name, and fair enough that it can probably just be
removed. Done.

PR #3419, opened 2020-03-23
hparams: read from generic data APIs

wchargin:
Summary:
This commit changes the data access methods in `hparams.backend_context`
to read from the context’s data provider rather than directly from its
multiplexer. All data access in the hparams plugin already goes through
these methods.

Test Plan:
Ideally, existing tests would have sufficed, but the tests made heavy
use of mocking and so required manual intervention. The updated tests
pass, and an inspection of the various views of the hparams dashboard
(including embedded scalar charts and data download links) checks out,
even when TensorBoard core is patched to not provide a multiplexer on
the context.

wchargin-branch: hparams-generic-data


tensorboard/plugins/hparams/backend_context.py
```diff
@@ -136,10 +138,25 @@ def read_scalars(self, experiment_id, run, tag):
...
+            raise KeyError("No scalar data for run=%r, tag=%r" % (run, tag))
+        return [
+            event_accumulator.TensorEvent(
+                wall_time=e.wall_time,
+                step=e.step,
+                tensor_proto=tensor_util.make_tensor_proto(e.value),
+            )
+            for e in data
+        ]
         return self._deprecated_multiplexer.Tensors(run, tag)
```
nfelt: Remove?

tensorboard/plugins/hparams/backend_context.py
```diff
@@ -136,10 +138,25 @@ def read_scalars(self, experiment_id, run, tag):
...
+            run_tag_filter=provider.RunTagFilter([run], [tag]),
+            downsample=(self._tb_context.sampling_hints or {}).get(
+                scalar_metadata.PLUGIN_NAME, 1000
+            ),
         )
+        data = data_provider_output.get(run, {}).get(tag)
+        if data is None:
+            raise KeyError("No scalar data for run=%r, tag=%r" % (run, tag))
+        return [
+            event_accumulator.TensorEvent(
```
nfelt: Can we maybe file a cleanup github issue to remove the vestigial bits of the multiplexer API from the hparams plugin and/or from plugins generally after they're migrated?  Would be nice to at least leave a TODO here, since it's kind of weird to still have TensorEvent used for this.

tensorboard/plugins/hparams/backend_context_test.py
```diff
@@ -46,14 +49,20 @@ class BackendContextTest(tf.test.TestCase):
     maxDiff = None  # pylint: disable=invalid-name
 
     def setUp(self):
-        self._mock_tb_context = mock.create_autospec(base_plugin.TBContext)
+        self._mock_tb_context = base_plugin.TBContext()
         self._mock_multiplexer = mock.create_autospec(
```
nfelt: Ditto here re: having a TODO to clean up the test code here so that if it's going to use mocks it at least mocks the DataProvider APIs directly rather than mocking EventMultiplexer and then wrapping that in MultiplexerDataProvider.

tensorboard/plugins/hparams/list_session_groups_test.py
```diff
@@ -53,12 +56,24 @@ class ListSessionGroupsTest(tf.test.TestCase):
     maxDiff = None  # pylint: disable=invalid-name
 
     def setUp(self):
-        self._mock_tb_context = mock.create_autospec(base_plugin.TBContext)
+        self._mock_tb_context = base_plugin.TBContext()
         self._mock_multiplexer = mock.create_autospec(
```
nfelt: Ditto here re: TODO.

tensorboard/plugins/hparams/backend_context.py
wchargin: Yep, thanks!

tensorboard/plugins/hparams/backend_context.py
wchargin: Yes, definitely. I was planning to do so right after these changes
(switch these functions to directly return the data provider results and
simultaneously migrate callers), but wanted to get these in first to
unblock some internal work. Filed #3425.

For what it’s worth, this is not a performance bottleneck. The
post-processing takes on the order of 1–5 microseconds. The call to
`data_provider.read_scalars` is slow because it calls `list_scalars`
internally, and `list_scalars` itself iterates over all the tensors to
generate metadata that we just throw away. It currently takes on the
order of 10 milliseconds, so that’s not good. And the hparams usage of
`read_scalars` is slow, too, because we call it once for each of the
O(|metrics| × |sessions|) time series instead of passing a single
`RunTagFilter` that represents the cross product, so that’s not good,
either. The data-munging here is negligible in comparison.

(I will see if I can speed up `_list_scalars` before merging this PR.)

tensorboard/plugins/hparams/list_session_groups_test.py
wchargin: Will do.

tensorboard/plugins/hparams/backend_context_test.py
wchargin: Will do.

tensorboard/plugins/hparams/backend_context.py
nfelt: Yeah, I wasn't concerned at all about performance.  Just seemed like a vestigial weirdness to leave behind - thanks for filing the issue!

tensorboard/plugins/hparams/backend_context.py
```diff
@@ -136,10 +138,25 @@ def read_scalars(self, experiment_id, run, tag):
...
+                wall_time=e.wall_time,
+                step=e.step,
+                tensor_proto=tensor_util.make_tensor_proto(e.value),
+            )
+            for e in data
+        ]
         return self._deprecated_multiplexer.Tensors(run, tag)
 
     def _find_experiment_tag(self, experiment_id):
```
nfelt: Above in this PR `hparams_metadata()` will resolve now to an unfiltered `list_tensors()` call, but in this particular callsite within `_find_experiment_tag()` we only want a single tag, so IMO it'd be nice if we just call the data provider with an appropriate tag filter rather than postfiltering all tags.  (Or optionally I guess `hparams_metadata()` could learn a `tags` argument, if you prefer.)

wchargin:
The perf improvements from #3433 _do_ have similar effect in google3,
taking the average test run time (over 100 trials) to… 59.9 seconds!
That’s just a wee bit close to the timeout, so I’m setting bumping it to
`moderate` to avoid flakiness, while noting that doing less work (by
making only one `read_scalars` call) is the real solution.

tensorboard/plugins/hparams/backend_context.py
```diff
@@ -166,18 +169,18 @@ def _find_experiment_tag(self, experiment_id):
...
+        mapping = self.hparams_metadata(
+            experiment_id,
+            run_tag_filter=provider.RunTagFilter(
+                tags=[metadata.EXPERIMENT_TAG]
+            ),
+        )
+        if not mapping:
+            return None
+        # We expect only one experiment; pick arbitrarily.
+        tag_to_content = mapping[next(iter(mapping))]
```
nfelt: nit: `next(iter(mapping.values()))` seems slightly simpler than feeding the key back into the dict?  Same below

tensorboard/plugins/hparams/backend_context.py
```diff
@@ -166,18 +169,18 @@ def _find_experiment_tag(self, experiment_id):
...
-        return experiment
+        mapping = self.hparams_metadata(
+            experiment_id,
+            run_tag_filter=provider.RunTagFilter(
+                tags=[metadata.EXPERIMENT_TAG]
+            ),
+        )
+        if not mapping:
+            return None
+        # We expect only one experiment; pick arbitrarily.
```
nfelt: This momentarily confused me between the hparams notion of experiment and the experiment ID experiment.  Perhaps slightly clearer to say "expect only a single run/tag pair with HParams experiment plugin data"?

wchargin: Done, with slightly different wording because “HParams experiment plugin
data” took me more than one try to parse.

PR #3420, opened 2020-03-23
ng: create feature_flag store for enabling exp features

stephanwlee:
We do not currently have a solid design for our flagging (experimental)
system. Regardless of the details of the system, because the flagging
system can impact many things (e.g., UI, data source, etc...), we want
to persist the flagging state in the store.

This is laying that foundation with a QueryParams backed implementation.

### DIFF
--- a/tensorboard/webapp/app_module.ts
+++ b/tensorboard/webapp/app_module.ts
@@ -21,15 +21,13 @@ import {EffectsModule} from '@ngrx/effects';
 import {AppContainer} from './app_container';
 import {CoreModule} from './core/core_module';
 import {HashStorageModule} from './core/views/hash_storage_module';
-import {PluginsModule} from './plugins/plugins_module';
-import {SettingsModule} from './settings/settings_module';
-
-import {ROOT_REDUCERS, loggerMetaReducerFactory} from './reducer_config';
-
-import {HeaderModule} from './header/header_module';
-import {ReloaderModule} from './reloader/reloader_module';
-import {MatIconModule} from './mat_icon_module';
 import {FeatureFlagModule} from './feature_flag/feature_flag_module';
+import {HeaderModule} from './header/header_module';
+import {MatIconModule} from './mat_icon_module';
+import {PluginsModule} from './plugins/plugins_module';
+import {ROOT_REDUCERS, loggerMetaReducerFactory} from './reducer_config';
+import {ReloaderModule} from './reloader/reloader_module';
+import {SettingsModule} from './settings/settings_module';
 
 @NgModule({
   declarations: [AppContainer],
--- a/tensorboard/webapp/app_module.ts
+++ b/tensorboard/webapp/app_module.ts
@@ -29,6 +29,7 @@ import {ROOT_REDUCERS, loggerMetaReducerFactory} from './reducer_config';
 import {HeaderModule} from './header/header_module';
 import {ReloaderModule} from './reloader/reloader_module';
 import {MatIconModule} from './mat_icon_module';
+import {FeatureFlagModule} from './feature_flag/feature_flag_module';
 
 @NgModule({
   declarations: [AppContainer],
@@ -36,6 +37,7 @@ import {MatIconModule} from './mat_icon_module';
     BrowserModule,
     BrowserAnimationsModule,
     CoreModule,
+    FeatureFlagModule,
     HashStorageModule,
     HeaderModule,
     MatIconModule,
--- a/tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
+++ b/tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
@@ -1,0 +1,52 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+import {Injectable} from '@angular/core';
+import {createAction, Action} from '@ngrx/store';
+import {Actions, ofType, createEffect} from '@ngrx/effects';
+import {map} from 'rxjs/operators';
+
+import {TBFeatureFlagDataSource} from '../../webapp_data_source/tb_feature_flag_data_source_types';
+import {featuresLoaded} from '../actions/feature_flag_actions';
+
+/** @typehack */ import * as _typeHackRxjs from 'rxjs';
+/** @typehack */ import * as _typeHackNgrx from '@ngrx/store/src/models';
+/** @typehack */ import * as _typeHackNgrxEffects from '@ngrx/effects';
+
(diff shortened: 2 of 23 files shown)


tensorboard/webapp/app_module.ts
```diff
@@ -29,13 +29,15 @@ import {ROOT_REDUCERS, loggerMetaReducerFactory} from './reducer_config';
 import {HeaderModule} from './header/header_module';
 import {ReloaderModule} from './reloader/reloader_module';
 import {MatIconModule} from './mat_icon_module';
+import {FeatureFlagModule} from './feature_flag/feature_flag_module';
```
caisq: nit: Sort imports by paths

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {Injectable} from '@angular/core';
+
+export type FeatureValue = string | boolean | number | string[];
```
caisq: Can you add a doc string to describe how this is used? When should it be a string, a boolean, a number, or a string array? It would be helpful to give some examples. Without them, it's hard to understand it by just looking at the code.

tensorboard/webapp/feature_flag/store/feature_flag_reducers_test.ts
```diff
@@ -0,0 +1,37 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import * as actions from '../actions/feature_flag_actions';
+import {reducers} from './feature_flag_reducers';
+import {buildFeatureFlagState} from './testing';
+
+describe('feature_flag_reducers', () => {
+  describe('featuresLoaded', () => {
```
caisq: Does it make sense to add a test to cover the case in which the action payload contains featureKeys?

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+@Injectable()
+export abstract class TBFeatureFlagDataSource {
+  abstract getFeatures(): {
+    enabledExperimentalPlugins: string[];
+    [featureKey: string]: FeatureValue;
```
caisq: This seems very generic. Related to the comment above, how is this used? Is it currently being used by anything? If not, how will it be used in the future? Is there a concrete plan? Give some examples.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -0,0 +1,33 @@
...
+export class QueryParamsFeatureFlagDataSource extends TBFeatureFlagDataSource {
+  getFeatures() {
+    const params = util.getParams();
+    return {
+      enabledExperimentalPlugins: params.getAll('experimentalPlugin'),
+    };
+  }
+}
+
+export const TEST_ONLY = util;
```
caisq: This way `util` ends up with a confusing name `TEST_ONLY`. Can you do this instead?

export const TEST_ONLY = {util};

tensorboard/webapp/feature_flag/store/feature_flag_types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const FEAUTURE_FLAG_FEATURE_KEY = 'feature';
+
+export interface FeatureFlagState {
+  enabledExperimentalPlugins: string[];
+  [featureId: string]: FeatureValue;
+}
+
+export interface State {
+  [FEAUTURE_FLAG_FEATURE_KEY]?: FeatureFlagState;
```
caisq: Can this be made non-optional?

tensorboard/webapp/feature_flag/store/feature_flag_selectors_test.ts
```diff
@@ -0,0 +1,55 @@
...
+          enabledExperimentalPlugins: ['foo'],
+        })
+      );
+      const actual = selectors.getFeature(state, 'bar');
+
+      expect(actual).toBeNull();
+    });
+  });
+
+  describe('getEnabledExperimentalPlugins', () => {
```
caisq: Isn't this test redundant with `'returns a current feature'`?

tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
```diff
@@ -0,0 +1,52 @@
...
+import {map} from 'rxjs/operators';
+
+import {TBFeatureFlagDataSource} from '../../webapp_data_source/tb_feature_flag_data_source_types';
+import {featuresLoaded} from '../actions/feature_flag_actions';
+
+/** @typehack */ import * as _typeHackRxjs from 'rxjs';
+/** @typehack */ import * as _typeHackNgrx from '@ngrx/store/src/models';
+/** @typehack */ import * as _typeHackNgrxEffects from '@ngrx/effects';
+
+const effectsInitialized = createAction('[FEATURE FLAG] Effects Init');
```
caisq: This action looks a little peculiar. Who / what is supposed to fire this action (outside tests)?

tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
```diff
@@ -0,0 +1,52 @@
...
+    )
+  );
+
+  constructor(
+    private readonly actions$: Actions,
+    private readonly dataSource: TBFeatureFlagDataSource
+  ) {}
+
+  /** @export */
+  ngrxOnInitEffects(): Action {
```
caisq: Related to the question above: Who is supposed to call this method (apart from tests)?

stephanwlee: Not us. It is called by Ngrx when an effect is initialized.

tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
stephanwlee: Ngrx, on effects initialization, calls `ngrxOnInitEffects` gets this action's payload. It, then, puts the payload into the action stream so we can write effects that do stuff on the initialization.

tensorboard/webapp/app_module.ts
stephanwlee: Since it is not enforceable or autoformattable, I hope not to make this a rule.

caisq: OK. I"ve been trying to do this manually myself. But I agree it's better to find a way to automate it than to always do it manually.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
```diff
@@ -14,10 +14,24 @@ limitations under the License.
...
+ */
 export type FeatureValue = string | boolean | number | string[];
 
 @Injectable()
 export abstract class TBFeatureFlagDataSource {
+  /**
+   * Gets feature flags defined.
+   *
+   * The "feature" is very loosely defined so other applications can define more
+   * flags. It is upto the application to better type the flags and create necessary
```
caisq: grammar nit: "upto" --> "up to"

PR #3426, opened 2020-03-24
ng: reload now only reloads current dashboard

stephanwlee:
Previously, we were reloading all stamped dashboards which is not the
behavior of the tf-tensorboard.

Side-effect: because we were reloading scalars dashboard when it was in
the background, the charts could more frequently appear tiny when the
reload kicked in. This change removes that bug.

### DIFF
--- a/tensorboard/webapp/plugins/plugins_container_test.ts
+++ b/tensorboard/webapp/plugins/plugins_container_test.ts
@@ -19,19 +19,26 @@ import {provideMockStore, MockStore} from '@ngrx/store/testing';
 
 import {PluginsContainer} from './plugins_container';
 import {PluginsComponent} from './plugins_component';
+import {PluginRegistryModule} from './plugin_registry_module';
+import {ExtraDashboardModule} from './testing';
 
-import {PluginId, LoadingMechanismType} from '../types/api';
+import {
+  PluginId,
+  LoadingMechanismType,
+  CustomElementLoadingMechanism,
+  IframeLoadingMechanism,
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/plugins/plugins_container_test.ts
```diff
@@ -219,42 +215,56 @@ describe('plugins_component', () => {
...
+      fixture.detectChanges();
+      setActivePlugin('bar');
       fixture.detectChanges();
 
       const {nativeElement} = fixture.debugElement.query(By.css('.plugins'));
-      const [barElement] = nativeElement.children;
+      // Stamped 'bar' and 'foo'
+      expect(nativeElement.children.length).toBe(2);
+      const [barElement, fooElement] = nativeElement.children;
       const reloadSpy = jasmine.createSpy();
```
psybuzz: nit: can we name this `barReloadSpy` to match the foo version?

stephanwlee:
> 1. Can the original bug (small charts) still technically occur, if an inactive dashboard's reload() is still pending?

Correct

> 2. This behavior now matches the Polymer tf-tensorboard, but I'd like to question the ideal UX. Feel free to defer/not answer this part, if it is beyond the scope.
> Let's say a user opens Scalars at t = 0, opens Images, and autoreloads until t = 5000s, then returns to Scalars. For up to 30s, they'll experience 2 dashboards with 5000s difference in data freshness.

Beyond scope and the same behavior is what I am trying to achieve here. Your assessment is correct and it is out of the scope.

PR #3428, opened 2020-03-25
ng: disable reload button for disable_reload=True

stephanwlee:
Our plugin API allows plugin authors to disable reload. In that case, we
gray out the reload button.

WANT_LGTM=any

### DIFF
--- a/tensorboard/webapp/header/reload_container.ts
+++ b/tensorboard/webapp/header/reload_container.ts
@@ -13,14 +13,27 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {Component} from '@angular/core';
-import {Store} from '@ngrx/store';
+import {Store, createSelector} from '@ngrx/store';
 import {Observable} from 'rxjs';
 import {map} from 'rxjs/operators';
 
 import {State} from '../core/store/core_types';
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/header/reload_container.ts
```diff
@@ -13,15 +13,28 @@ See the License for the specific language governing permissions and
...
-import {getPluginsListLoaded} from '../core/store/core_selectors';
+import {
+  getPlugins,
+  getActivePlugin,
+  getPluginsListLoaded,
+} from '../core/store/core_selectors';
 import {manualReload} from '../core/actions';
 import {DataLoadState} from '../types/data';
 
+const isReloadDiabledByPlugin = createSelector(
```
psybuzz: nit: typo `isReloadDisabledByPlugin`

tensorboard/webapp/header/reload_container.ts
```diff
@@ -13,15 +13,28 @@ See the License for the specific language governing permissions and
...
+  getPlugins,
+  getActivePlugin,
+  (plugins, id) => {
+    if (!id || !plugins || !plugins[id]) return false;
```
psybuzz: I think `plugins` starts as an `{}` and is never non-truthy?  Is it safe to drop `!plugins`?

PR #3430, opened 2020-03-25
[DebuggerV2] Avoid loading source files that are loaded or loading

caisq:
  * Previously, a source-file content was always loaded from the data source even if it's been loaded already.
  * In this PR, we change the behavior so that source files that are already loaded or currently loading are not requested again under the `sourceLineFocused` action.
  * Unit test added.
  * Verified in a child PR that no repeated request occurs in the browser.

### DIFF
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
@@ -1150,7 +1150,7 @@ describe('Debugger effects', () => {
     });
 
     for (const loadState of [DataLoadState.LOADED, DataLoadState.LOADING]) {
-      it('skips loading a loaded or loeading source file', () => {
+      it(`skips loading when file is loaded or loading: ${loadState}`, () => {
         const runId = '__default_debugger_run__';
         const fileIndex = 2;
         const fileContentC: SourceFileResponse = {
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects_test.ts
```diff
@@ -1149,6 +1149,50 @@ describe('Debugger effects', () => {
       ]);
     });
 
+    for (const loadState of [DataLoadState.LOADED, DataLoadState.LOADING]) {
+      it('skips loading a loaded or loeading source file', () => {
```
psybuzz: Let's use a test spec title that varies depending on the loadstate.  That way, one can read the Karma test results and know which one failed. e.g. perhaps something like
```
for (const spec of [{loadState: DataLoadState.Loaded, name: 'loaded'}, ...]) {
  it(`skips loading a source file that is ${spec.loadState}`, () => {
    // ...
  });
}
```

caisq: Good point! Thanks for catching it.

caisq: Done.

caisq:
Thanks for the review, @psybuzz

PR #3433, opened 2020-03-26
data: optimize `read_scalars` by skipping scans

wchargin:
Summary:
Prior to this change, `read_scalars` (resp. `read_tensors`) delegated to
`list_scalars` (resp. `list_tensors`) to find the set of time series to
read. This is slower than it might sound, because `list_scalars` itself
needs to scan over all relevant `multiplexer.Tensors` to identify
`max_step` and `max_wall_time`, which are thrown away by `read_scalars`.
(That `list_scalars` needs this full scan at all is its own issue;
ideally, these would be memoized onto the event multiplexer.)

When a `RunTagFilter` specifying a single run and tag is given, we
optimize further by requesting individual `SummaryMetadata` rather than
paring down `AllSummaryMetadata`.

Resolves a comment of @nfelt on #2980:
<https://github.com/tensorflow/tensorboard/pull/2980#discussion_r353469128>

Test Plan:
When applied on top of #3419, `:list_session_groups_test` improves from
taking 11.1 seconds to taking 6.6 seconds on my machine. This doesn’t
seem to fully generalize; I see only ~13% speedups in a microbenchmark
that hammers `read_scalars` on a logdir with all the demo data, but the
improvement on that test is important.

wchargin-branch: data-read-without-list


wchargin:
Agreed. Unfortunately, it looks like the blob code doesn’t have any
tests, so I’d prefer to add those tests before changing the code. The
`graphs_plugin_test` has some coverage, but I’d be more comfortable if
the data provider itself were tested. Filed #3434.

tensorboard/backend/event_processing/data_provider.py
```diff
@@ -221,51 +227,57 @@ def _list(self, construct_time_series, index):
                 )
         return result
 
-    def _read(self, convert_event, index, downsample):
-        """Helper to read scalar or tensor data from the multiplexer.
+    def _read(
```
nfelt: Hmm, so in my original comment I mostly meant changing the blobs to also use the `_index()` helper.  I'm a little ambivalent about whether the consolidation to have them also use the `_read()` helper is worth it?  It complicates the event converters (since they now need all the extra parameters) and it forces `_read()` to take more parameters and become kind of monolithic, whereas I actually kind of liked how before the `read_*()` implementations could be decomposed into validation, generating the index, and then reading from it.

That said since you've already done the work to consolidate it, totally up to you, just my $0.02.

tensorboard/backend/event_processing/data_provider.py
```diff
@@ -394,7 +393,7 @@ def _decode_blob_key(key):
     return (experiment_id, plugin_name, run, tag, step, index)
 
 
-def _convert_scalar_event(event):
+def _convert_scalar_event(experiment_id, plugin_name, run, tag, event):
```
nfelt: Optional, but IMO a little nicer if we make `event` the first argument for all of these so that the other 4 arguments are more semantically positioned to be optional arguments, even if we don't bother actually making them optional.  Might also be nice to add a `del experiment_id, plugin_name, run, tag  # unused` just to make it clearer that these first two converters ignore all of those arguments.

tensorboard/backend/event_processing/data_provider.py
wchargin: Cool, I’m glad you agree. :-) I also don’t like the smörgåsbord of extra
arguments to… everything. Reverted to just the `_index` simplification.

> That said since you've already done the work to consolidate it,
> totally up to you, just my $0.02.

Not at all—my willingness to rewrite a PR depends only on the value of
the proposed change and not on how many times I’ve already rewritten it.

tensorboard/backend/event_processing/data_provider.py
wchargin: Removed arguments entirely.

tensorboard/backend/event_processing/data_provider.py
nfelt: LGTM :)

PR #3435, opened 2020-03-26
data: add tests for blob sequence handling

### LINKED ISSUE #3434: MultiplexerDataProvider blob code needs tests
The changes to `event_processing/data_provider.py` from #2991 should
have dedicated tests in `data_provider_test.py` so that we can feel
comfortable changing that code. (Cf. #3433.)

wchargin:
Summary:
Follow-up to #2991. Fixes #3434.

Test Plan:
Tests pass as written.

wchargin-branch: data-blob-sequence-tests


tensorboard/backend/event_processing/data_provider_test.py
```diff
@@ -91,6 +93,27 @@ def setUp(self):
...
+        with tf.summary.create_file_writer(logdir).as_default():
+            data = [
+                ("red", (221, 28, 38), "top-right"),
+                ("blue", (1, 91, 158), "bottom-left"),
+                ("yellow", (239, 220, 111), "bottom-right"),
+            ]
+            for (name, color, description) in data:
+                image_1x1 = tf.constant([[[color]]], dtype=tf.uint8)
+                for i in xrange(1, 11):
+                    k = 6 - abs(6 - i)  # 1, .., 6, .., 2
```
nfelt: Could you add a comment like "Use a non-monotonic sequence of sample sizes to test max_length calculation" or something?  Wasn't obvious to me why the formula for k had to be this way otherwise.

wchargin: Ah, yes, sure. (Yes, that’s the reason.) Done.

PR #3438, opened 2020-03-26
hparams: reduce scalar reads while building sessions

wchargin:
Summary:
Prior to this change, `list_session_groups` called `last_metric_eval`
once for each session-metric combination, so we ended up calling
`read_scalars` O(|sessions| × |metrics|) times when we could call it
just once with a compound `RunTagFilter`. The performance hit isn’t too
bad with the in-memory `MultiplexerDataProvider`, but with networked
data providers this dramatically reduces the number of RPCs sent. This
also reins in `:list_session_groups_test` to approximately restore its
performance prior to the generic data change (about 3.0s on my machine).

Test Plan:
Existing tests suffice. As a sanity check, the three views of the
dashboard still seem to work, including inline charts and data download
links.

wchargin-branch: hparams-yoro


tensorboard/plugins/hparams/backend_context.py
```diff
@@ -129,38 +129,32 @@ def scalars_metadata(self, experiment_id):
...
             plugin_name=scalar_metadata.PLUGIN_NAME,
-            run_tag_filter=provider.RunTagFilter([run], [tag]),
-            downsample=(self._tb_context.sampling_hints or {}).get(
-                scalar_metadata.PLUGIN_NAME, 1000
-            ),
+            run_tag_filter=run_tag_filter,
+            # TODO(#3436): We assume that downsampling always includes
+            # the most recent datum, which holds for all implementations
+            # of interest but is not yet required by the interface.
+            downsample=1,
```
wchargin: Note: I went ahead and made this change to unblock some internal work.

PR #3440, opened 2020-03-26
docs: fix broken links in `overview.md`

### LINKED ISSUE #2814: doc: r1 overview's link to custom scalar is not useful
https://github.com/tensorflow/tensorboard/blob/master/docs/r1/overview.md#can-i-customize-which-lines-appear-in-a-plot for "custom scalar plugin" links to https://github.com/tensorflow/tensorboard/blob/master/docs/r1/tensorboard/plugins/custom_scalar which is does not exist.

LexusH:
Fixed 2 broken links that were pointed out in #2814.

Fixes #2814.

### DIFF
--- a/docs/r1/overview.md
+++ b/docs/r1/overview.md
@@ -311,7 +311,7 @@ method.
 
 ### Can I customize which lines appear in a plot?
 
-Using the [custom scalars plugin](tensorboard/plugins/custom_scalar), you can
+Using the [custom scalars plugin](https://github.com/tensorflow/tensorboard/tree/master/tensorboard/plugins/custom_scalar), you can
 create scalar plots with lines for custom run-tag pairs. However, within the
 original scalars dashboard, each scalar plot corresponds to data for a specific
 tag and contains lines for each run that includes that tag.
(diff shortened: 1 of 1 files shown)


docs/r1/overview.md
```diff
@@ -311,15 +311,15 @@ method.
 
 ### Can I customize which lines appear in a plot?
 
-Using the [custom scalars plugin](tensorboard/plugins/custom_scalar), you can
+Using the [custom scalars plugin](https://github.com/tensorflow/tensorboard/tree/master/tensorboard/plugins/custom_scalar), you can
```
stephanwlee: Could you use a relative path? i.e., `../../../tensorboard/plugins/custom_scalar`. Here and below

LexusH: Sure! I'll work on fixing the other two broken links you pointed out too!

docs/r1/overview.md
```diff
@@ -311,15 +311,15 @@ method.
 
 ### Can I customize which lines appear in a plot?
 
-Using the [custom scalars plugin](tensorboard/plugins/custom_scalar), you can
+Using the [custom scalars plugin](/../../tree/master/tensorboard/plugins/custom_scalar), you can
```
stephanwlee: Remove tree/master and the leading "/"

docs/r1/overview.md
```diff
@@ -335,7 +335,7 @@ TensorBoard uses [reservoir
 sampling](https://en.wikipedia.org/wiki/Reservoir_sampling) to downsample your
 data so that it can be loaded into RAM. You can modify the number of elements it
 will keep per tag in
-[tensorboard/backend/application.py](tensorboard/backend/application.py).
+[tensorboard/backend/application.py](/tensorboard/backend/application.py).
```
stephanwlee: Please remove "/" and use relative path instead.

docs/r1/overview.md
wchargin: Thank you! For context, the reason that we like relative links is that
they enable editors to cross-link. For instance, in Vim, I can just
press `gf` (go-file) with my cursor on the relative link and it’ll open
in my editor.

PR #3448, opened 2020-03-30
Uploader: Add `total_blob_bytes` field to `Experiment` proto

caisq:
  * Towards fixing b/152749189: displaying the number of blob bytes that an experiment contains int the `tensorboard dev list` command output
  * Add a `total_blob_bytes` field to the `Experiment` proto, which is the respnose message type of `StreamExperiemnts`

### DIFF
--- a/tensorboard/uploader/proto/experiment.proto
+++ b/tensorboard/uploader/proto/experiment.proto
@@ -29,7 +29,7 @@ message Experiment {
   // The number of bytes used for storage of tensors in this experiment,
   // across all time series, including estimated overhead. Output-only.
   int64 total_tensor_bytes = 9;
-  // The number of bytes used for storage of the content of blobs in this
+  // The number of bytes used for storage of the contents of blobs in this
   // experiment, across all time series, including estimated overhead.
   // Output-only.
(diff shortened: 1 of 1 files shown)


tensorboard/uploader/proto/experiment.proto
```diff
@@ -29,6 +29,10 @@ message Experiment {
   // The number of bytes used for storage of tensors in this experiment,
   // across all time series, including estimated overhead. Output-only.
   int64 total_tensor_bytes = 9;
+  // The number of bytes used for storage of the contents of blobs in this
+  // experiment, across all time series, including estimated overhead.
+  // Output-only.
+  int64 total_blob_bytes = 10;
```
wchargin, line 35: Needs corresponding fieldmask update?

caisq: Looks like so. cc @ericdnielsen for confirmation. If confirmed, I can send a follow-up PR to add that.

ericdnielsen: I'm not too familiar with the fieldmaks.  But it sounds likely that its needed here.

PR #3450, opened 2020-03-30
Update yarn.lock

stephanwlee:
The change includes instruction on how to manage our dependencies.

5252a1f84 updates the yarn.lock but seems to not contain changes
that would be created if done `yarn add`. This change manually is
generated by running `yarn`. 

Note that it was not possible to reproduce 5252a1f84.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,8 @@ jobs:
       - run: yarn install --ignore-engines
         # You can run `yarn fix-lint` to fix all Prettier complaints.
       - run: yarn lint
+        # Make sure yarn.lock did not change with `yarn install`
+      - run: git diff --stat --exit-code -- yarn.lock
 
   lint-build:
     runs-on: ubuntu-16.04
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -76,9 +76,6 @@ $ yarn upgrade d3
 
 ```
 
-At the end, please make sure `yarn.lock` does not update by running `yarn` after
-making a necessary commit.
-
 ## Pro tips
 
 You may find the following optional tips useful for development.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -57,6 +57,28 @@ $ git cherry-pick bc4e7a6e5517daf918433a8f5983fc6bd239358f
 [pr-1334]: https://github.com/tensorflow/tensorboard/pull/1334
 [yarn]: https://yarnpkg.com/
 
+## Managing frontend dependencies
+
+TensorBoard manages frontend dependencies in two places: third_party/js.bzl and
+package.json. Polymer based components make use of the js.bzl whereas the new
+Angular modules make use of package.json through `yarn`. When adding or updating
+dependencies on package.json, please use `yarn` to manage the dependencies.
+
+For instance,
+```bash
+
+# To install new dependencies (-D for development dependencies; please refer to
+# yarn add -h for help):
+$ yarn add d3 -D @types/d3
+
+# To upgrade dependencies:
+$ yarn upgrade d3
+
+```
+
+At the end, please make sure `yarn.lock` does not update by running `yarn` after
+making a necessary commit.
+
 ## Pro tips
 
 You may find the following optional tips useful for development.
--- a/yarn.lock
+++ b/yarn.lock
@@ -493,11 +493,6 @@ JSONStream@^1.3.4:
     jsonparse "^1.2.0"
     through ">=2.2.7 <3"
 
-abbrev@1:
-  version "1.1.1"
-  resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.1.tgz#f8f2c887ad10bf67f634f005b6987fed3179aac8"
-  integrity sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==
-
 accepts@~1.3.4:
   version "1.3.7"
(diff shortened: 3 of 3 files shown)


DEVELOPMENT.md
```diff
@@ -57,6 +57,28 @@ $ git cherry-pick bc4e7a6e5517daf918433a8f5983fc6bd239358f
...
+# yarn add -h for help):
+$ yarn add d3 -D @types/d3
+
+# To upgrade dependencies:
+$ yarn upgrade d3
+
+```
+
+At the end, please make sure `yarn.lock` does not update by running `yarn` after
+making a necessary commit.
```
caisq: This sentence is a little hard to parse and understand.

`yarn.lock` will change as a result of adding new dependencies or explicitly upgrading dependencies. Its change should be committed, right? This sentence seems to imply that the change should never be committed.

caisq:
BTW, thanks for adding this guideline.

DEVELOPMENT.md
wchargin: Would it be possible to add a CI check for this? Since `lint-frontend`
already needs to run `yarn` to run Prettier, perhaps it could also check
that the Git status of `yarn.lock` is clean? Otherwise, this seems like
the kind of thing that people will frequently forget to do.

.github/workflows/ci.yml
```diff
@@ -92,6 +92,8 @@ jobs:
       - run: yarn install --ignore-engines
         # You can run `yarn fix-lint` to fix all Prettier complaints.
       - run: yarn lint
+        # Make sure yarn.lock did not change with `yarn install`
+      - run: git diff --stat --exit-code -- yarn.lock
```
wchargin: Perfect; thanks! Extra points for the extra safe `-- yarn.lock` form,
and thanks for verifying that it fails at head.

DEVELOPMENT.md
nfelt: Is it true that the `yarn.lock` diff only ever arises when package.json changes, or does it arise sometimes from e.g. a change in our dependencies, or a new release of a dependency that we don't have pinned to a specific version?

If the latter I'm a bit hesitant to make this a CI check as-is, because it seems like it would fail frequently at master with no changes from us, even though it doesn't really indicate a problem (versus just a chore we need to attend to).  Also it would make people inclined to update yarn.lock as part of an unrelated change which is bad for blame, etc.  It would seem better in that case to try to make the check conditional on the PR touching package.json or yarn.lock.

stephanwlee: Yes, it can change. 

I think the best solution to move with is to use `--frozen-lockfile` option on `yarn install` so it uses the lockfile as the source of truth (frozen-lockfile should be default unless you are meaning to update the dependencies; or at least that was the semantics on npm and its shrinkwrap). 

Unfortunately, the version of bazel_nodejs we are using does not support any arguments to yarn, but we can use yarnrc (forgot what it was exactly called) to add one. Want me to alter the PR into that direction instead?

stephanwlee:
The approach is generally flawed. Abandoned.

DEVELOPMENT.md
wchargin: At least in normal JS development, when the repository has an existing
lockfile and `package.json` has not changed, running `yarn` is
guaranteed not to update the lockfile.\* This must be the case, because
the first thing that you do when cloning a JS repo is to run `yarn`, and
if that dirtied the lockfile then the kind of problem addressed by this
PR would be commonplace. Removing `node_modules` and running `yarn` to
re-populate it is also a common operation (at least among people like me
who routinely poke into `node_modules` for debugging), and that’s also
safe.

\* Unless you’re using a newer version of Yarn to run `yarn` than you
did when you originally generated the lockfile, in which case Yarn may
add integrity checks and stuff as migrations, but that’s kind of a
one-time thing and not a routine occurrence.

We definitely should make sure that this is hermetic, and definitely
shouldn’t update `yarn.lock` in unrelated commits. I think that the only
way that this could not do what we want is if the Bazel/`rules_nodejs`
integrations have some behavior that we don’t want, but Bazel is all
about hermeticity, so I wouldn’t expect that to be the case.

(@stephanwlee can correct me if any of this understanding is wrong.)

Gating on `package.json`/`yarn.lock` changes is certainly fine with me
if it’s necessary for some reason, and also okay with me if it’s not
strictly necessary but keeps us more comfortable, though it is a bit of
extra code that requires a bit of care to get the right diff range.

stephanwlee: > At least in normal JS development, when the repository has an existing
lockfile and package.json has not changed, running yarn is
guaranteed not to update the lockfile.*

Not sure if that is entirely true. IIRC, that is correct on `npm` but not on `yarn`:
https://github.com/yarnpkg/yarn/issues/4147

wchargin: It looks like that issue is discussing lock files that don’t accurately
reflect `package.json`, though? By “`package.json` has not changed” I
meant “…since the last run of (the same version of) `yarn`”. Running
`npm install --save lodash` breaks that condition by changing
`package.json` without running `yarn`. Or am I misunderstanding?

(The issue reference is nice, though; maybe adding `--frozen-lockfile`
to the `yarn install` earlier in CI is all that we need?)

stephanwlee: I attached another GH issue that is more pertinent to our issue. PTAL.

wchargin: Interesting, thanks. I wasn’t aware of that. It seems to me that that
issue makes it *more* important to have this check in our CI. As far as
I can tell, that issue is solely about idempotency, not hermeticity.
That is, it sounds like Yarn eventually converges to a fixpoint of the
lockfile (specifically, after at most 2 steps) that depends only on
`package.json` and not on the actual non-hermetic contents of
dependencies. So having it in CI ensures that we’ve reached that
fixpoint, and doesn’t introduce flakiness because the correct output is
always determined by the repository contents. Is that right? (I mean,
I’m prepared to be wrong again! :-) )

(For what it’s worth, I haven’t personally run into either of these
issues on SourceCred or any personal projects.)

PR #3453, opened 2020-03-31
Provide upload limits in ServerInfo, and skip uploading blobs that are too large

davidsoergel:
Introduces a new `UploadLimits` submessage within `ServerInfoResponse`, to allow the server to advise the uploader client about limits that it should honor (e.g. things like data sizes, upload rates, and so on).

The only limit actually supported is `max_blob_size`.  The uploader uses this value to skip sending large blobs that the server would end up rejecting anyway. 

Context: b/152312275, b/152770109, and internal email threads.

### DIFF
--- a/tensorboard/uploader/proto/server_info.proto
+++ b/tensorboard/uploader/proto/server_info.proto
@@ -40,12 +40,16 @@ message ServerInfoResponse {
   // The client should only upload data for the plugins in the response even
   // if it is capable of uploading more data.
   PluginControl plugin_control = 4;
-  // The maximum allowed size for blob uploads.
-  // Note that the WriteBlob RPC endpoint also enforces a limit, which should be
-  // kept in sync with the one reported here.  Providing this limit in advance
-  // allows the uploader to print a better error message and to skip the upload
-  // if it is expected to fail anyway.
-  int64 max_blob_size = 5;
+  // Limits on the upload process that the client should honor.
+  // This may include limits on data size, chunk size, records per request,
+  // request rate, bandwidth, etc.  The server may enforce such limits, in which
+  // case the values reported here should be kept in sync.  Providing these
+  // limits in advance allows the client to avoid triggering server errors (both
+  // through more efficent operation and by earlier detection of real error
+  // conditions), and to print better error messages.
+  // If this field is not set, the client should choose reasonable default
+  // values to guide its behavior.
+  UploadLimits upload_limits = 5;
 }
 
 enum CompatibilityVerdict {
@@ -97,4 +101,10 @@ message PluginControl {
   // Plugins for which data should be uploaded. These are plugin names as
   // stored in the the `SummaryMetadata.plugin_data.plugin_name` proto field.
   repeated string allowed_plugins = 1;
+}
+
+message UploadLimits {
+  // The maximum allowed size for blob uploads.
+  // If this is 0, no blobs should be uploaded.
+  int64 max_blob_size = 1;
 }
--- a/tensorboard/uploader/proto/server_info.proto
+++ b/tensorboard/uploader/proto/server_info.proto
@@ -40,6 +40,12 @@ message ServerInfoResponse {
   // The client should only upload data for the plugins in the response even
   // if it is capable of uploading more data.
   PluginControl plugin_control = 4;
+  // The maximum allowed size for blob uploads.
+  // Note that the WriteBlob RPC endpoint also enforces a limit, which should be
+  // kept in sync with the one reported here.  Providing this limit in advance
+  // allows the uploader to print a better error message and to skip the upload
+  // if it is expected to fail anyway.
+  int64 max_blob_size = 5;
 }
 
 enum CompatibilityVerdict {
--- a/tensorboard/uploader/proto/server_info.proto
+++ b/tensorboard/uploader/proto/server_info.proto
@@ -105,6 +105,8 @@ message PluginControl {
 
 message UploadLimits {
   // The maximum allowed size for blob uploads.
-  // If this is 0, no blobs should be uploaded.
+  // If this is 0, it is allowed to upload blobs of size 0 (though note
+  // the server may enforce a *minimum* blob size, which is a separate issue).
+  // If this is negative, no blobs should be uploaded.
   int64 max_blob_size = 1;
 }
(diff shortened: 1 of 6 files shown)


tensorboard/uploader/server_info.py
```diff
@@ -30,6 +30,8 @@
 # Request timeout for communicating with remote server.
 _REQUEST_TIMEOUT_SECONDS = 10
 
+# Maximum blob size, if not specified by server_info
+_DEFAULT_MAX_BLOB_SIZE = 10 * (2 ** 20)   # 10 MB
```
wchargin: This is 10 MiB, not 10 MB.

(Likewise in comment below.)

tensorboard/uploader/proto/server_info.proto
```diff
@@ -40,6 +40,12 @@ message ServerInfoResponse {
   // The client should only upload data for the plugins in the response even
   // if it is capable of uploading more data.
   PluginControl plugin_control = 4;
+  // The maximum allowed size for blob uploads.
+  // Note that the WriteBlob RPC endpoint also enforces a limit, which should be
+  // kept in sync with the one reported here.  Providing this limit in advance
+  // allows the uploader to print a better error message and to skip the upload
+  // if it is expected to fail anyway.
+  int64 max_blob_size = 5;
```
wchargin: Based on usage, it looks like `max_blob_size = 0` means “pick a
reasonable default” rather than either “do not upload any blobs” or
“there is no limit; upload blobs of any size”, which would have been my
two guesses; please document?

wchargin: Internal API guidance proscribes fields of primitive type on a top-level
response proto. We’ve been following this best practice for this RPC so
far; can we keep doing so?

tensorboard/uploader/server_info.py
```diff
@@ -152,6 +154,32 @@ def allowed_plugins(server_info):
...
+    a better error message and to avoid attempting the upload if it is expected
+    to fail anyway.
+
+    Args:
+      server_info: A `server_info_pb2.ServerInfoResponse` message.
+
+    Returns:
+      A integer giving the maximum size allowed for blob uploads, in bytes.
+    """
+    if server_info.HasField("max_blob_size"):
```
wchargin: On the proto, `max_blob_size` is an `int64`, so this is equivalent to
just `if server_info.max_blob_size` (i.e., if-nonzero). IMHO this makes
the code a bit misleading, as it suggests that `max_blob_size` can be
“unset”. Can we either clarify that we’re testing the integer value of
the field or just `return si.max_blob_size or _DEFAULT`?

tensorboard/uploader/server_info.py
```diff
@@ -152,6 +154,32 @@ def allowed_plugins(server_info):
...
+        return server_info.plugin_control.max_blob_size
```
wchargin: This dereferences `max_blob_size` from `plugin_control`, but the proto
actually has it on the root response. Please add tests (as for sibling
function `allowed_plugins`) that would catch this.

tensorboard/uploader/uploader.py
```diff
@@ -720,6 +727,14 @@ def _get_or_create_blob_sequence(self):
     def _send_blob(self, blob_sequence_id, seq_index, blob):
         # TODO(soergel): retry and resume logic
 
+        if len(blob) > self._max_blob_size:
+            logger.warn(
```
wchargin: The `warn` method is deprecated; please use `warning`. (This should have
printed a `DeprecationWarning` in tests?)

tensorboard/uploader/uploader.py
```diff
@@ -720,6 +727,14 @@ def _get_or_create_blob_sequence(self):
     def _send_blob(self, blob_sequence_id, seq_index, blob):
```
wchargin: Could you please document the return value now that it’s non-`None`?
I can infer the semantics from clients but not from the function itself.

tensorboard/uploader/uploader_test.py
```diff
@@ -72,6 +72,7 @@ def _create_mock_client():
 
 _SCALARS_ONLY = frozenset((scalars_metadata.PLUGIN_NAME,))
 
+_TEN_MEGABYTES = 10 * (2 ** 20)
```
wchargin: This kind of `const TWELVE = 12;` definition is generally considered an
antipattern; if we want a constant here, can it describe the semantics
rather than the value? Also fine with me to just inline it, as it looks
like the particular value is not important (which is good, imho).

tensorboard/uploader/server_info.py
davidsoergel: Good point.  By adding the submessage, we do now have to test for the presence of `upload_limits`.  If it is present, then then`max_blob_size=0` will now be interpreted in the natural way (don't upload blobs).

tensorboard/uploader/server_info.py
davidsoergel: Whoopsiedoodle.  Fixed and added tests.

tensorboard/uploader/uploader.py
davidsoergel: Huh, no `DeprecationWarning` that I noticed on a quick skim.  Done, anyway.

tensorboard/uploader/uploader.py
davidsoergel: Done.

tensorboard/uploader/uploader_test.py
davidsoergel: Yeah I totally agree that's an antipattern.  I must have been quite addled at the time.  I inlined `12345` to emphasize that the value is arbitrary here.

tensorboard/uploader/uploader.py
wchargin: Interesting. I see it on Python 3.7.6:

> uploader.py:741: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead

Perhaps we have different environments somehow. Thanks, anyway.

tensorboard/uploader/uploader_test.py
wchargin: :-) thanks; sgtm!

tensorboard/uploader/proto/server_info.proto
```diff
@@ -92,3 +102,9 @@ message PluginControl {
   // stored in the the `SummaryMetadata.plugin_data.plugin_name` proto field.
   repeated string allowed_plugins = 1;
 }
+
+message UploadLimits {
+  // The maximum allowed size for blob uploads.
+  // If this is 0, no blobs should be uploaded.
+  int64 max_blob_size = 1;
+}
```
wchargin, line 110: Hmm, I approved thinking that lint would catch this, but it didn’t,
somehow: missing newline at end of file.

(I’ll look into why.)

PR #3455, opened 2020-03-31
docs: adopt `nbfmt.py` formatting, enforced in CI

wchargin:
Summary:
Notebook formatting is a frequent source of friction in pull requests,
due to spurious diffs and unintended changes. @lamberta’s `nbfmt.py` is
an opinionated autoformatter that should resolve these issues entirely.

Test Plan:
A test run of the CI change fails before the reformat and passes after:

  - <https://github.com/tensorflow/tensorboard/runs/549352035>
  - <https://github.com/tensorflow/tensorboard/runs/549365513>

Both the failure and passing cases have nice output.

wchargin-branch: docs-nbfmt


.github/workflows/ci.yml
```diff
@@ -82,7 +82,14 @@ jobs:
...
         run: yamllint -c docs/.yamllint docs docs/.yamllint
-      - run: ./tensorboard/tools/docs_list_format_test.sh
+      - name: 'Install nbfmt.py for checking notebook formatting'
+        run: |
+          nbfmt_version="e9d8ddf8c0f1160d0eae654f30c028d2b0287538"
+          wget "https://raw.githubusercontent.com/tensorflow/docs/${nbfmt_version}/tools/nbfmt.py"
+          pip install absl-py
+          chmod +x ./nbfmt.py
+      - name: 'Check Colab notebooks for formatting'
+        run: git ls-files -z '*.ipynb' | xargs -0 ./nbfmt.py --test
```
bileschi: like the preventative work here to avoid quoted fns

wchargin: Yep! I generally try to be funnynames-safe when it’s not too
inconvenient, and in most cases it’s easy enough.

PR #3456, opened 2020-03-31
Debugger V1: Disable flaky tests with BUILD tags

caisq:
  * Disable the intereactive_debugger_plugin_test that is showing flaky timeout in internal builds, to avoid blocking releases
  * Add the manual and notap tags.


caisq:
Closing in favor of copybara-based approach suggested by @nfelt

PR #3457, opened 2020-03-31
Include the size of the blob in the WriteBlobRequest

ericdnielsen:
Quota management assumes that the blob's final size is available before we allow the upload to begin.  The existing quota tools correctly allocate or reject based on a known final size, but the backend at present is always passing zero, with a comment that it should be added to the request.

### DIFF
--- a/tensorboard/uploader/proto/write_service.proto
+++ b/tensorboard/uploader/proto/write_service.proto
@@ -332,6 +332,9 @@ message WriteBlobRequest {
   // CRC32C of the entire blob.  Required, to protect against data corruption.
   // This should be set only when finalize_object=True.
   fixed32 final_crc32c = 7;
+  // Final size of the entire blob.  Required in the first request to allow
+  // quota allocaion and management.
+  int64 final_size = 8;
 }
 
 message WriteBlobResponse {
--- a/tensorboard/uploader/proto/write_service.proto
+++ b/tensorboard/uploader/proto/write_service.proto
@@ -332,8 +332,8 @@ message WriteBlobRequest {
   // CRC32C of the entire blob.  Required, to protect against data corruption.
   // This should be set only when finalize_object=True.
   fixed32 final_crc32c = 7;
-  // Final size of the entire blob.  Required in the first request to allow
-  // quota allocaion and management.
+  // Final size in bytes of the entire blob.  Required in the first request
+  // to allow quota allocaion and management.
   int64 final_size = 8;
 }
 
--- a/tensorboard/uploader/proto/write_service.proto
+++ b/tensorboard/uploader/proto/write_service.proto
@@ -332,9 +332,9 @@ message WriteBlobRequest {
   // CRC32C of the entire blob.  Required, to protect against data corruption.
   // This should be set only when finalize_object=True.
   fixed32 final_crc32c = 7;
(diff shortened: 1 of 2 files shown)


tensorboard/uploader/proto/write_service.proto
```diff
@@ -332,6 +332,9 @@ message WriteBlobRequest {
   // CRC32C of the entire blob.  Required, to protect against data corruption.
   // This should be set only when finalize_object=True.
   fixed32 final_crc32c = 7;
+  // Final size of the entire blob.  Required in the first request to allow
```
bileschi: Please specify the units (bytes)

tensorboard/uploader/proto/write_service.proto
```diff
@@ -332,6 +332,9 @@ message WriteBlobRequest {
   // CRC32C of the entire blob.  Required, to protect against data corruption.
   // This should be set only when finalize_object=True.
   fixed32 final_crc32c = 7;
+  // Final size in bytes of the entire blob.  Required in the first request
+  // to allow quota allocaion and management.
+  int64 final_size = 8;
```
wchargin: Old uploaders will appear to send `0` for this field. Are we prepared
for how the server will handle that (e.g., rejecting the requests
because the client tries to send a nonempty blob after “declaring” that
it would be of length 0)? As long as the only relevant uploaders are
nightly releases, breaking them is fine with me, just worth considering.

tensorboard/uploader/proto/write_service.proto
```diff
@@ -332,6 +332,9 @@ message WriteBlobRequest {
   // CRC32C of the entire blob.  Required, to protect against data corruption.
   // This should be set only when finalize_object=True.
   fixed32 final_crc32c = 7;
+  // Final size in bytes of the entire blob.  Required in the first request
+  // to allow quota allocaion and management.
```
wchargin: sp.: *allocaion* → *allocation*

tensorboard/uploader/uploader.py
```diff
@@ -763,6 +763,7 @@ def _write_blob_request_iterator(self, blob_sequence_id, seq_index, blob):
                 crc32c=None,
                 finalize_object=finalize_object,
                 final_crc32c=None,
+                final_size=len(blob),
```
wchargin: Please test this? (Should be easy to modify existing tests.)

tensorboard/uploader/proto/write_service.proto
ericdnielsen: renamed to blob_bytes.

No usages outside nightlies at present.  I'll discuss with David about enforcement issues.

tensorboard/uploader/proto/write_service.proto
ericdnielsen: done

tensorboard/uploader/uploader.py
ericdnielsen: I went pretty deep on investigating the mock call args, but it was getting pretty messy. Not seeing a simple approach.

davidsoergel:
Please hold off submitting this pending discussion on the corresponding internal CL.  Thanks!

ericdnielsen:
David and I synced on the internal CL.  This PR is good to go.

PR #3458, opened 2020-03-31
cleanup: lint and protect against skipped tests

stephanwlee:
As a developer, it is super easy to erroneously commit fdescribe or
fit. This lint protects against that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
       - run: yarn lint
         # Make sure no tests are skipped with "focused" tests.
       - run: |
-          ! git grep -E 'f(it|describe)\(' 'tensorboard/webapp/*_test.ts' 'tensorboard/plugins/*_test.ts'
+          ! git grep -E 'f(it|describe)\(' 'tensorboard/*_test.ts'
 
   lint-build:
     runs-on: ubuntu-16.04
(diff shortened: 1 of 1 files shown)


caisq:
driveby: Thanks for doing this, @stephanwlee

.github/workflows/ci.yml
```diff
@@ -92,6 +92,9 @@ jobs:
       - run: yarn install --ignore-engines
         # You can run `yarn fix-lint` to fix all Prettier complaints.
       - run: yarn lint
+        # Make sure no tests are skipped with "focused" tests.
+      - run: |
+          ! find tensorboard/webapp -type f -regex '.*_test.ts$' -exec grep -E 'f(it|describe)\(' {} +
```
wchargin, line 97: This is correct and works fine, but using `git grep` is clearer:

```
! git grep -E 'f(it|describe)\(' 'tensorboard/webapp/*_test.ts'
```

(It’s also often faster, for what it’s worth.)

caisq: @stephanwlee , if you go with @wchargin's suggestion, please consider including *_test.ts files under tensorboard/plugins as well.

.github/workflows/ci.yml
```diff
@@ -94,7 +94,7 @@ jobs:
       - run: yarn lint
         # Make sure no tests are skipped with "focused" tests.
       - run: |
-          ! find tensorboard/webapp -type f -regex '.*_test.ts$' -exec grep -E 'f(it|describe)\(' {} +
+          ! git grep -E 'f(it|describe)\(' 'tensorboard/webapp/*_test.ts' 'tensorboard/plugins/*_test.ts'
```
wchargin, line 97: Fine with me if this is what you want, though if we’re cherry-picking
two directories it’s a bit unclear to me why we don’t just run against
all of `*_test.ts`.

PR #3466, opened 2020-04-01
uploader: remove dev_creds.py

nfelt:
This removes the empty dev_creds.py file which we no longer need.


nfelt:
No changes to the PR, but requested a review just so it's marked as one you reviewed (like setting R= after already getting an LGTM).

PR #3441, opened 2020-03-28
Fix bad images and other minor text changes

anirudh161:
* Motivation for features / changes

Fixed missing TensorBoard iframes by saving notebook with outputs. Made some minor text changes

PR #3468, opened 2020-04-01
docs: update profiler tutorial

anirudh161:
* Motivation for features / changes:
1. Saved the notebooks with outputs enabled (and private outputs disabled) as a workaround to prevent broken iframes rendering on tensorflow.org
2. Made some minor text changes and added a link to the Dev Summit 2020 talk

### DIFF
--- a/docs/tensorboard_profiling_keras.ipynb
+++ b/docs/tensorboard_profiling_keras.ipynb
@@ -604,7 +604,7 @@
       "source": [
         "## Summary\n",
         "\n",
-        "Use the TensorFlow Profiler to profile and debug model training performance. Read the [Profiler guide](https://www.tensorflow.org/guide/profiler) and watchg the [Performance profiling in TF 2](https://www.youtube.com/watch?v=pXHAQIhhMhI) talk from the TensorFlow Dev Summit 2020 to learn more about the TensorFlow Profiler."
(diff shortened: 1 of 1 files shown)


wchargin:
For context, this supersedes #3441.

anirudh161:
That's weird. The outputs were there before I ran nbfmt.py on the notebook. Let me look into it and submit another commit

wchargin:
As I [mentioned on #3441][1], make sure that `private_outputs: false` is
set in the notebook to preserve the contents. If you have a notebook
with output cells, you can set that easily by running `nbfmt.py` on the
notebook and passing `--preserve_outputs`. This will write the option
into the notebook so that future `nbfmt.py` runs know not to remove
them, even when the flag is not passed.

It doesn’t look like `private_outputs: false` was set at any point in
this change, but it also doesn’t look like the outputs were saved. The
“Formatted notebook with nbfmt” commit has a delta of 111 insertions and
110 deletions (i.e., just sorted keys and fixed trailing newline), and
you can see from the diff that it didn’t actually remove any outputs
(they were all `outputs: []` already).

[1]: https://github.com/tensorflow/tensorboard/pull/3441#issuecomment-606732347

anirudh161:
I was so sure I saved the outputs then! Guess I didn't after all :) I added another commit with outputs saved

anirudh161:
That's a good point. I fixed the cache entries and added a new commit

wchargin:
Awesome; thanks! Tested; this works for me both before and after running
the notebook.

I’ll go ahead and merge this now.

PR #3470, opened 2020-04-02
Use python3 -m before creating the virtualenv

rithiksachdev:
We might an error because of using two python versions.


rithiksachdev:
Ubuntu version: 18.04.4
Hi @wchargin, I am really sorry for this late reply. I had created my virtualenv for development, a few days ago, and for some reason, my virtualenv was using pip from /usr/bin/python2.7.
Due to an accident last weekend, I lost all my data and had to create the virtualenv again. I faced this issue while trying for the first time and then once I created the virtualenv, I didn't face the issue again. I completely agree with what you are saying, the pip on our path should be the pip from the virtualenv, so, I would like to suggest whether we could use python3 -m before creating the virtualenv, if necessary. For example in the Requirements of [Fairness Indicator](https://github.com/tensorflow/tensorboard/blob/master/docs/fairness-indicators.md)

wchargin:
No worries! Glad that we’re on the same page. Yes, we can go ahead and
suggest creating the virtualenv with Python 3. I’ve sent out #3487 to
this effect.

PR #3472, opened 2020-04-02
[DebuggerV2] Add HTTP route /graph_execution/digests

caisq:
  * Continue developing the HTTP backend of DebuggerV2: the `/graph_execution/digests` serves small-size digest data objects for intra-graph execution events.
  * The added route `/graph_execution/digests` is a parallel to the existing route `/execution/digests`
  * The underlying methods are also parallels to the existing execution route:
    * `DebugDataMultiplexer.GraphExecutionDigests()` --> `DebugDataMultiplexer.ExecutionDigests()`
    * `DebuggerV2Plugin.serve_graph_execution_digests()` --> `DebuggerV2Plugin.serve_execution_digests()`
  * A follow-up PR will add the corresponding route for serving detailed data objects (i.e, `/graph_execution/data`)
  * Unit tests added.
  * Parameterize the existing `/execution/digests` path to accommodate both top-level and intra-graph executions
    * Pro: slightly less code
    * Con: More complex control flow logic in the code, especially considering that the `/graph_execution/digests` path will need to handle the `trace_id` parameter in the future, while the `/execution/digest` doesn't have that parameter.
    * Con: More confusing HTTP route pattern to understand

### DIFF
--- a/tensorboard/plugins/debugger_v2/debug_data_provider.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_provider.py
@@ -198,7 +198,7 @@ def graph_execution_digest_run_tag_filter(run, begin, end, trace_id=None):
     # execution and intra-graph execution is supported by DebugDataReader.
     if trace_id is not None:
         raise NotImplementedError(
-            "trace_id support for graph_execution_digest_run_tag_filter is "
+            "trace_id support for graph_execution_digest_run_tag_filter() is "
             "not implemented yet."
         )
     return provider.RunTagFilter(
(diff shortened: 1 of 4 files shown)


caisq:
Invited @wchargin to review this PR. @wchargin has reviewed similar PRs recently. Thanks!

tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -177,6 +179,64 @@ def _parse_execution_data_blob_key(blob_key):
     return run, begin, end
 
 
+def graph_execution_digest_run_tag_filter(run, begin, end, trace_id=None):
```
wchargin, line 182: Are graph execution digests not keyed by both run and tag? For normal
graph summaries, at least, users use `tf.summary.trace_export(name)`,
which writes a tag name.

caisq: In the design of DebuggerV2 backend, internally-generated tags are used to tag different types of data (e.g., see the `GRAPH_EXECUTION_DIGESTS_BLOB_TAG_PREFIX` and similar string constant tags for constructing special tags in this file.) Therefore user data is not queried by tags.

PR #3475, opened 2020-04-03
ng: bootstrap core store from hash

stephanwlee:
Noet that this introduces deeplink module that manages everything
related to deeplinking.

Deeplinking, whether we have a router or not, is a feature want to
support. Since most of the deeplink states are present in the hash, the
module is responsible for read/write with the hash directly.

hash_storage module should be part of the deep link module but since it
is already being used internally, we will have to make a careful changes
while not breaking any apps. The change will be made in a follow up.

### DIFF
--- a/tensorboard/webapp/core/core_module.ts
+++ b/tensorboard/webapp/core/core_module.ts
@@ -30,9 +30,9 @@ import {HashDeepLinker} from '../deeplink';
 
 @NgModule({
   imports: [
-    TBServerDataSourceModule,
-    StoreModule.forFeature(CORE_FEATURE_KEY, reducers, CORE_STORE_CONFIG_TOKEN),
     EffectsModule.forFeature([CoreEffects]),
+    StoreModule.forFeature(CORE_FEATURE_KEY, reducers, CORE_STORE_CONFIG_TOKEN),
+    TBServerDataSourceModule,
   ],
   providers: [
     {
--- a/tensorboard/webapp/core/core_module.ts
+++ b/tensorboard/webapp/core/core_module.ts
@@ -16,16 +16,30 @@ import {NgModule} from '@angular/core';
 import {StoreModule} from '@ngrx/store';
 import {EffectsModule} from '@ngrx/effects';
 
+import {TBServerDataSourceModule} from '../webapp_data_source/tb_server_data_source_module';
+
 import {reducers} from './store';
 import {CoreEffects} from './effects';
 import {CORE_FEATURE_KEY} from './store/core_types';
-import {TBServerDataSourceModule} from '../webapp_data_source/tb_server_data_source_module';
+import {
+  CORE_STORE_CONFIG_TOKEN,
+  getConfig,
+} from './store/core_initial_state_provider';
+import {DeepLinkModule} from '../deeplink/deeplink_module';
+import {HashDeepLinker} from '../deeplink';
 
 @NgModule({
   imports: [
     TBServerDataSourceModule,
-    StoreModule.forFeature(CORE_FEATURE_KEY, reducers),
+    StoreModule.forFeature(CORE_FEATURE_KEY, reducers, CORE_STORE_CONFIG_TOKEN),
     EffectsModule.forFeature([CoreEffects]),
+  ],
+  providers: [
+    {
+      provide: CORE_STORE_CONFIG_TOKEN,
+      deps: [HashDeepLinker],
+      useFactory: getConfig,
+    },
   ],
 })
 export class CoreModule {}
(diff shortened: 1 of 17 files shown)


tensorboard/webapp/deeplink/hash_test.ts
```diff
@@ -0,0 +1,94 @@
...
+    } as any);
+    createElement.and.callThrough();
+  });
+
+  it('sets the hash to plugin id by replacing on first load', () => {
+    store.overrideSelector(getActivePlugin, 'foo');
+    const fixture = TestBed.createComponent(HashStorageContainer);
+    fixture.detectChanges();
+
+    expect(setStringSpy).toHaveBeenCalledWith(jasmine.any(String), 'foo', {
```
caisq, line 64: Is there value in asserting that the key argument used is `'__tab__'` by using a `TEST_ONLY`? Same in the following parts in this file.

tensorboard/webapp/core/core_module.ts
```diff
@@ -16,16 +16,30 @@ import {NgModule} from '@angular/core';
...
+  getConfig,
+} from './store/core_initial_state_provider';
+import {DeepLinkModule} from '../deeplink/deeplink_module';
+import {HashDeepLinker} from '../deeplink';
 
 @NgModule({
   imports: [
     TBServerDataSourceModule,
-    StoreModule.forFeature(CORE_FEATURE_KEY, reducers),
+    StoreModule.forFeature(CORE_FEATURE_KEY, reducers, CORE_STORE_CONFIG_TOKEN),
```
caisq, line 34: nit: Sort args please.

tensorboard/webapp/core/store/core_reducers.ts
```diff
@@ -112,6 +115,6 @@ const reducer = createReducer(
   })
 );
 
-export function reducers(state: CoreState, action: Action) {
+export function reducers(state: CoreState | undefined, action: Action) {
```
caisq, line 118: To help me understand: why is this change necessary?

tensorboard/webapp/deeplink/hash.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    tfGlobals.tf_globals.setUseHash(true);
+  }
+
+  getString(key: string): string {
+    return this.tfStorage.tf_storage.getString(key);
+  }
+
+  setString(key: string, value: string, options?: SetStringOption): void {
+    return this.tfStorage.tf_storage.setString(key, value, options);
```
caisq, line 53: `return` can be omitted. Same for `setPluginId()` below.

tensorboard/webapp/deeplink/types.ts
```diff
@@ -0,0 +1,26 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+export interface SetStringOption {
+  defaultValue?: string;
+  useLocationReplace?: boolean;
```
caisq, line 18: I don't believe there is a doc string for this parameter in the tensorboard code base. Consider providing one here.

tensorboard/webapp/plugins/plugins_container.ts
```diff
@@ -30,7 +30,7 @@ const activePlugin = createSelector(
   getPlugins,
   getActivePlugin,
   (plugins, id): UiPluginMetadata | null => {
-    if (!id || !plugins) return null;
+    if (!id || !plugins[id]) return null;
```
caisq, line 33: This is not directly related to the main changes in this PR, but fixing a pre-existing bug, right?

tensorboard/webapp/core/store/core_reducers.ts
stephanwlee: Correct. When injected the state via third arguments in the `StoreModule.forFeature`, it looks like the `state` can be undefined.

tensorboard/webapp/plugins/plugins_container.ts
stephanwlee: It is related. After this change, the store now has `activePlugin` state but no the `plugins`. Previously, the `activePlugin` was available only after `plugins` were set. 

In other words, only after this change, this bug manifests.

tensorboard/webapp/core/core_module.ts
stephanwlee: Btw, that is generally not applicable. Order of dependencies changes order of execution so if you want to override a provider for customization, that has to appear at the end.

PR #3476, opened 2020-04-03
build: refactor ng_polymer_lib to be reused.

stephanwlee:
ng_polymer_lib and tensorboard.html should differ in only two things:
- only former has Angular related code bundled separately.
- only latter has entry point for tf-tensorboard
(`<body><tf-tensorboard>`).

This change refactors so that all the Polymer binary without the entry
point is shared between Polymer TensorBoard and Angular TensorBoard.

### DIFF
--- a/tensorboard/components/BUILD
+++ b/tensorboard/components/BUILD
@@ -6,11 +6,39 @@ load("//tensorboard/defs:vulcanize.bzl", "tensorboard_html_binary")
 licenses(["notice"])  # Apache 2.0
 
 tf_web_library(
+    name = "polymer_lib",
+    srcs = [
+        "polymer_lib.html",
+    ],
+    path = "/",
+    deps = [
+        ":analytics",
+        ":security",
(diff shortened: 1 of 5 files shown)


stephanwlee:
cc @caisq

tensorboard/components/BUILD
```diff
@@ -5,12 +5,40 @@ load("//tensorboard/defs:vulcanize.bzl", "tensorboard_html_binary")
...
+tf_web_library(
+    name = "polymer_lib",
+    srcs = [
+        "polymer_lib.html",
+    ],
+    path = "/",
+    deps = [
+        ":analytics",
+        ":security",
+        "//tensorboard/components/tf_backend",
```
psybuzz, line 17: :security, tf_backend, tf_storage, tf_tensorboard:registry were only present in the old `ng_polymer.html`.  Now that the new `polymer_lib.html` doesn't HTML-import them, can they be dropped from these BUILD deps?

PR #3479, opened 2020-04-04
hparams: recognize empty plugin content

wchargin:
Summary:
Addresses a deferred review comment from #3449. This shouldn’t matter in
practice, because `HParamsPluginData` values always supposed to have one
of the `oneof data` fields set, so even an experiment with no hparams,
no metrics, and no start time would have plugin content of `\x12\x00`.
But conceptually this code is trying to check for `None`, not “`None` or
empty”, so this saves future readers from having to convince themselves
that the two conditions are equivalent.

Test Plan:
Unit tests pass and a spot-check of the dashboard checks out.

wchargin-branch: hparams-empty-plugin-content


wchargin:
No problem! It was a nice relaxing commit for a Friday afternoon. :-)

PR #3480, opened 2020-04-05
uploader: add `--json` flag to the `list` command

caisq:
  * Fulfill feature request bug b/153232102
  * Add the `--json` flag to the `list` subcommand of `tensorboard dev`.
  * If the flag is used, the experiments will be printed as a JSON object mapping experiment URLs to experiment data (name, description, runs, tags, etc.)
* Screenshots of UI changes
  * image
  * Manually ran `tensorboard dev list --json` (see screenshot above)
  * Output a single big json array at the end:
    * Pro: may be easier to parse programmatically
    * Con: no streaming


bileschi:
Do we envision more formats being desired here? (csv?).  Did you consider whether this should be a more general `--output_format` flag?  Is there guidance or precedent we can follow?

tensorboard/uploader/uploader_main.py
```diff
@@ -332,6 +333,15 @@ class _ListIntent(_Intent):
         """
     )
 
+    def __init__(self, json=None):
+        """Constructor of _ListIntent.
+
+        Args:
+          json: If and ony if `True`, will print the list as a pretty-formatted
```
wchargin, line 340: sp.: *ony* → *only*

caisq:
@bileschi Good question. I opted for the current approach (i.e., `--json` instead of `--output_format=json`) for following reasons:

1. I don't think it's a good idea to show the data in the csv format on the screen in general. The experiment description can contain newline characters, leading to potentially messy and hard-to-read outputs (see https://stackoverflow.com/questions/566052/can-you-encode-cr-lf-in-into-csv-files). JSON handles this well, by using the `\n` escape character in its string values.
2. Given `--json` will be supported, I don't see too much extra value in supporting csv additionally.
3. Even if we do support csv in the future, it makes sense to output the results to a file directly, in which case a separate arg such as `--csv_file=...` makes sense.

tensorboard/uploader/uploader_main.py
```diff
@@ -363,8 +374,16 @@ def execute(self, server_info, channel):
...
+            if self.json:
+                experiments_json.append(
+                    collections.OrderedDict([("URL", url)] + data)
+                )
+            else:
+                print(url)
+                for (name, value) in data:
+                    print("\t%s %s" % (name.ljust(12), value))
+        if self.json:
+            print(json.dumps(experiments_json, indent=2))
```
wchargin, line 386: So the output here looks like this:

```json
[
  {
    "URL": "https://tensorboard.dev/experiment/abcdefghijklmnopqrstuv/",
    "Name": "[No Name]",
    "Description": "[No Description]",
    "Id": "abcdefghijklmnopqrstuv",
    "Created": "2019-10-25 07:22:30",
    "Updated": "2019-10-25 07:22:34",
    "Scalars": "3208",
    "Runs": "8",
    "Tags": "4"
  }
]
```

There are a few problems with this:

 1. The keys are the same as the human-readable form, and in particular
    are capitalized and (after #3464) may have whitespace.
 2. Numeric values like `runs` and `tags` are strings, not ints.
 3. Name and description have dummy values instead of proper `""`s.
 4. The output is slurped and not streamed.

On (1): After #3464, this will have an extra `"Binary object bytes"`
key. But this is pretty unidiomatic. JSON keys do not typically have
whitespace; you can’t use such a key in a `jq` dot-selector:

```
$ printf '{"one": 1, "two three": 23}' | jq '.one'
1
$ printf '{"one": 1, "two three": 23}' | jq '.two three'
jq: error: syntax error, unexpected IDENT, expecting $end (Unix shell quoting issues?) at <top-level>, line 1:
.two three     
jq: 1 compile error
```

On (2): The integers are being converted to string for human
presentation, but should be left as ints in the JSON output for
programmatic manipulation.

On (3): Similar to (2). The dummy values are just to prevent confusing
human-readable output with a blank space. There’s no ambiguity in using
JSON `""`.

On (4): I see that you considered a streaming response but didn’t like
that it was “less easy to parse programmatically”. Is it? You can
convert a streaming response to a buffered response by simply piping
through `jq -s`, but you can’t go the other way around without waiting
for the whole buffer. And as long as we make sure to print one record
per line, people using tools other than JQ should be able to deal with
it handily (`for line in file: json.loads(line)` works in Python, e.g.).

One way to resolve these could be to change `data` from a 2-column table
to a 4-column table with human readable name, JSON key, raw value, and
formatter used when displaying to humans: e.g.,

```
            data = [
                ("Name", "name", experiment.name, lambda x: x or "[No Name]"),
                ("Description", "description", experiment.description, lambda x: x or "[No Description]"),
                ("Id", "id", experiment.experiment_id, str),
                ("Created", "created", util.format_time(experiment.create_time), str),
                ("Updated", "updated", util.format_time(experiment.update_time), str),
                ("Scalars", "scalars", experiment.num_scalars, str),
                ("Runs", "runs", experiment.num_runs, str),
                ("Tags", "tags", experiment.num_tags, str),
            ]
```

and project out the appropriate fields in the JSON and non-JSON cases;
of course feel free to pick a different implementation if you prefer.

tensorboard/uploader/uploader_main.py
```diff
@@ -363,8 +374,16 @@ def execute(self, server_info, channel):
                 ("Runs", str(experiment.num_runs)),
                 ("Tags", str(experiment.num_tags)),
             ]
-            for (name, value) in data:
-                print("\t%s %s" % (name.ljust(12), value))
+            if self.json:
+                experiments_json.append(
+                    collections.OrderedDict([("URL", url)] + data)
```
wchargin, line 379: Could we just pass `sort_keys=True` instead of using an `OrderedDict`?
The determinism is nice, but I don’t know that we really need to enforce
the same order as with the human-readable output; sorting keys is the
standard “stable stringify”.

tensorboard/uploader/uploader_main.py
```diff
@@ -348,11 +358,12 @@ def execute(self, server_info, channel):
         )
         gen = exporter_lib.list_experiments(api_client, fieldmask=fieldmask)
         count = 0
+        if self.json:
```
davidsoergel, line 361: The code paths here are asymmetric and intertwined (as the need for multiple `if self.json:` shows).  This is in part because the json path collects all the entries first and then prints them at the end (as noted in the PR description), but the non-json version is streaming.

First let's make this print exactly the same json output, except in a streaming manner:

```
if self.json:
    print "["
for experiment in gen:
    ...
    if self.json:
        print(textwrap.indent(json.dumps(data, indent=2), prefix="  "))
    ...
if self.json:
    print "]"
```

Next, to be more persnickety and to make it easier to add other formats, you could factor out the format-specific parts by creating an abstract `Formatter` class with methods `prefix()`, `format_experiment(data)`, and `suffix()`.  Then implement those for text and json, respectively, and pass the appropriate Formatter in to `_ListIntent` (or at least, pick the right one in the constructor).

By making the Formatter methods return strings (as opposed to `print(...)`ing as a side effect) you can make the choice of destination (stdout vs file, etc.) an orthogonal concern.

tensorboard/uploader/flags_parser.py
```diff
@@ -177,6 +177,11 @@ def define_flags(parser):
         "list", help="list previously uploaded experiments"
     )
     list_parser.set_defaults(**{SUBCOMMAND_FLAG: SUBCOMMAND_KEY_LIST})
+    list_parser.add_argument(
+        "--json",
```
caisq, line 181: @bileschi  As pointed out by @wchargin, "prior art" varies. So I opted to err on the side of conciseness. Also see my earlier comments regarding this.

tensorboard/uploader/formatters.py
```diff
@@ -0,0 +1,99 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import json
+
+EXPERIMENT_METADATA_URL_JSON_KEY = "url"
+ExperimentMetadataField = collections.namedtuple(
+    "ExperimentMetadataField",
+    ("json_key", "readable_name", "value", "formatter"),
```
davidsoergel, line 27: I was confused by the name collision of "formatter".  Can this one be "field_formatter" or "to_readable" or similar?

Also, it still feels a bit intertwined to pass a lambda that is specifically used only in the ReadableFormatter case.  I think you can remove the field entirely and automate this.  In `ReadableFormatter.format_experiment`, you could:

```
if not metadata_field.value and not typeof(metadata_field.value) == 'int':
  readable_value = "[No %s]".format(metadata_field.readable_name)
else:
  readable_value = str(metadata_field.value)
```

The typecheck there is a little iffy (because, as written, `None` will trigger the substitution even for fields that should be int, and a future float-valued field containing `0.0` would trigger too).  I guess `ExperimentMetadataField` could carry a Type enum, but that seems overwrought.  I think it'd be fine to gloss over all that for now (maybe with a code comment about the limitations).

tensorboard/uploader/uploader_main.py
davidsoergel: Awesome, thanks!

tensorboard/uploader/formatters.py
wchargin, line 27: Adding a formatter interface sounds reasonable to me, but this doesn’t
really feel like the right structure. The `ExperimentMetadataField`
struct is tightly coupled to the two implementations of formatters. The
`readable_name` nad `formatter` fields are only relevant to the
`ReadableFormatter`, and the `json_key` field is only relevant to the
`JsonFormatter`. And as written, although the name `Experiment*` is
everywhere, nothing actually knows about experiments.

Why introduce a new type with a big list of vague “fields” when we
already have the experiment proto? The interface could just be

```python
class ExperimentFormatter(metaclass=abc.ABCMeta):
    def format(self, experiment):
        """Format an experiment.

        Args:
          experiment: An `experiment_pb2.Experiment` value.

        Returns:
          A string.
        """
        pass
```

The `ReadableExperimentFormatter` would encapsulate the readable field
names and how to format them, and the `JsonExperiment` would encapsulate
the JSON keys and how to order and indent them. The tests would be on
the same output that we would actually see from the binary rather than
just an arbitrary set of fields defined in the test case.

Then, you could remove the giant block of field definitions in
`uploader_main.py`, too: just `print(formatter.format(experiment))`.

tensorboard/uploader/formatters_test.py
```diff
@@ -0,0 +1,124 @@
...
+        self.assertEqual(lines[0], "{")
+        self.assertEqual(
+            lines[1], '  "url": "http://tensorboard.dev/deadbeef",'
+        )
+        self.assertEqual(lines[2], '  "name": "",')
+        self.assertEqual(lines[3], '  "description": "",')
+        self.assertEqual(lines[4], '  "runs": 8,')
+        self.assertEqual(lines[5], '  "tags": 12,')
+        self.assertEqual(lines[6], '  "binary_object_bytes": 2000')
+        self.assertEqual(lines[7], "}")
```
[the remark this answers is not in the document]
caisq: Done.

tensorboard/uploader/formatters.py
```diff
@@ -0,0 +1,99 @@
...
+        """Format a list of `ExperimentMetadataField`s as a representing string.
+
+        Args:
+          experiment_metadata: A list of `ExperimentMetadataField`s that
+            describes an experiment.
+
+        Returns:
+          A string that represents the `experiment_metadata`.
+        """
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
caisq: Done.

tensorboard/uploader/formatters.py
caisq: Done effecting the refactoring that you suggested. This revision has pushed a lot of the variables related to the formatting into `formatters.py`, which is indeed nicer than before. 

Also note that I changed the signature of the `format_experiment()` method slightly so that it takes two input arguments: `experiment` (a `experiment_pb2.Experiment` proto) and `experiment_url` (a string). Getting the latter from an `Experiment` proto requires the `server_info` module, which I think shouldn't be known to the formatter module, which is purely concerned with string formatting.

tensorboard/uploader/uploader_main.py
```diff
@@ -348,23 +358,16 @@ def execute(self, server_info, channel):
         )
         gen = exporter_lib.list_experiments(api_client, fieldmask=fieldmask)
         count = 0
+
+        if self.json:
+            formatter = formatters.JsonFormatterI()
```
wchargin, line 363: Presumably `JsonFormatterI` should be `JsonFormatter`?

This breaks when run:

```
    formatter = formatters.JsonFormatterI()
AttributeError: module 'tensorboard.uploader.formatters' has no attribute 'JsonFormatterI'
```

tensorboard/uploader/formatters.py
```diff
@@ -0,0 +1,100 @@
...
+            ("description", experiment.description),
+            ("id", experiment.experiment_id),
+            ("created", util.format_time(experiment.create_time)),
+            ("updated", util.format_time(experiment.update_time)),
+            ("runs", experiment.num_runs),
+            ("tags", experiment.num_tags),
+            ("scalars", experiment.num_scalars),
+        ]
+        return json.dumps(
+            collections.OrderedDict(data), indent=self._JSON_INDENT,
```
wchargin, line 99: Optional, but: if you’re concerned about ease of parsing the JSON output
without `jq`, you could consider using `indent=None` (the default), so
that the output is one JSON object per line. This makes it easy to parse
with a parser that can only take a complete JSON string, like Python’s
`json.loads` or JS’s `JSON.parse`, because the user can easily identify
the framing boundaries (just split by newline). Otherwise, they have to
identify the actual object boundaries themselves, which is slightly less
trivial.

Up to you; just mentioning because you noted this concern in the
original PR description.

tensorboard/uploader/formatters.py
```diff
@@ -0,0 +1,100 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import abc
+import collections
+import json
+
+import six
```
wchargin, line 25: No need to support Python 2 in new `srcs_version = "PY3"` modules.

tensorboard/uploader/formatters.py
```diff
@@ -0,0 +1,100 @@
...
+        super(JsonFormatter, self).__init__()
+
+    def format_experiment(self, experiment, experiment_url):
+        data = [
+            ("url", experiment_url),
+            ("name", experiment.name),
+            ("description", experiment.description),
+            ("id", experiment.experiment_id),
+            ("created", util.format_time(experiment.create_time)),
+            ("updated", util.format_time(experiment.update_time)),
```
wchargin, line 93: It probably makes more sense for to use `format_time_absolute` for the
JSON formatter? Since the point is to have a simple, parseable format,
outputs like `"just now"` probably aren’t what’s wanted.

tensorboard/uploader/uploader_main.py
caisq: Hmmm. I don't know how this got in. I'll be more careful with my keyboard setup. Thanks for catching this.

tensorboard/uploader/formatters.py
caisq: Done. There are a few other modules in tensorboard/uploader that are still using `six.add_metaclass()`. Those can be corrected later.

tensorboard/uploader/formatters.py
caisq: Done. Unit test is updated accordingly.

tensorboard/uploader/formatters.py
caisq: I did a test with `tensorboard dev list --json | jq -s` under this PR. `jq` seems to be able to handle the `indent=2` just fine. So I'll leave it as is.

wchargin: Right, yes, `jq` handles it fine, hence “[…] ease of parsing the JSON
output without `jq`.” This is fine with me.

tensorboard/uploader/formatters.py
```diff
@@ -0,0 +1,99 @@
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
wchargin, line 1: nit: 2019 → 2020 (and test)

caisq: Done here and in `formatters_test.py`.

tensorboard/uploader/formatters.py
caisq: Ack. I think the slight difficulty of parsing with other tools should be acceptable.

PR #3482, opened 2020-04-06
[DebuggerV2] Add HTTP route /graph_execution/data

caisq:
  * Continue developing the HTTP backend of DebuggerV2: the `/graph_execution/data` serves full-size data objects for intra-graph execution events.
  * This is a followed up #3472 that added the `/graph_execution/digests` route.
  * The added route `/graph_execution/digests` is a parallel to the existing route `/execution/data`
  * The underlying methods are also parallels to the existing execution route:
    * `DebugDataMultiplexer.GraphExecutionData()` --> `DebugDataMultiplexer.ExecutionData()`
    * `DebuggerV2Plugin.serve_graph_execution_data()` --> `DebuggerV2Plugin.serve_execution_data)`
  * Unit tests added.
  * Parameterize the existing `/execution/data` path to accommodate both top-level and intra-graph executions
    * Pro: slightly less code
    * Con: More complex control flow logic in the code, especially considering that the `/graph_execution/data` path will need to handle the `trace_id` parameter in the future, while the `/execution/data` doesn't have that parameter.
    * Con: More confusing HTTP route pattern to understand


caisq:
The test failures are pre-existing and unrelated to this PR.

caisq:
@wchargin Ack re. go link.

caisq:
Thanks for the review, @wchargin !

PR #3483, opened 2020-04-06
Disable auto reload if page is not visible

bmd3k:
Reduce the load on TensorBoard servers by disabling auto reload if the page is not considered visible.

Do not auto reload if document is not visible according to the Page
Visibility API. When page again becomes visible perform an auto reload
if one has been missed.

Run a local TensorBoard.
Open localhost:6006.
Ensure option "Reload data every 30s." is selected.
Open detached network tab in chrome developer tools.
Watch for several minutes and observe that many network calls are made every 30s.
Open different tab in same window but with chrome developer tools still visible.
Watch for several minutes and observe that no network calls are made.
Switch back to TensorBoard tab and watch as network calls are immediately made and UI reloads graphs.
Continue to watch network tab as it again follows a 30s period for reload network calls.

### DIFF
--- a/tensorboard/components/tf_tensorboard/autoReloadBehavior.ts
+++ b/tensorboard/components/tf_tensorboard/autoReloadBehavior.ts
@@ -39,13 +39,28 @@ namespace tf_tensorboard {
       _autoReloadId: {
         type: Number,
       },
+      // Tracks whethere an auto reload was missed because the document was not visible.
+      _missedAutoReload: {
+        type: Boolean,
+        value: false,
+      },
       autoReloadIntervalSecs: {
         type: Number,
         value: 30,
       },
     },
+    attached: function() {
+      document.addEventListener(
+        'visibilitychange',
+        this._handleVisibilityChange.bind(this)
+      );
+    },
     detached: function() {
       window.clearTimeout(this._autoReloadId);
+      document.removeEventListener(
+        'visibilitychange',
+        this._handleVisibilityChange.bind(this)
+      );
     },
     _autoReloadObserver: function(autoReload) {
       window.localStorage.setItem(AUTORELOAD_LOCALSTORAGE_KEY, autoReload);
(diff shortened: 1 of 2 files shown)


tensorboard/components/tf_tensorboard/autoReloadBehavior.ts
```diff
@@ -59,14 +74,34 @@ namespace tf_tensorboard {
...
+        this._missedAutoReload = true;
       }
-      this.reload();
       this._autoReloadId = window.setTimeout(
         () => this._doAutoReload(),
         this.autoReloadIntervalSecs * 1000
       );
     },
+    _doReload: function() {
+      if (this.reload == null) {
```
stephanwlee, line 88: Prefer triple equals.

tensorboard/components/tf_tensorboard/autoReloadBehavior.ts
```diff
@@ -39,13 +39,28 @@ namespace tf_tensorboard {
...
+      document.addEventListener(
+        'visibilitychange',
+        this._handleVisibilityChange.bind(this)
+      );
+    },
     detached: function() {
       window.clearTimeout(this._autoReloadId);
+      document.removeEventListener(
+        'visibilitychange',
+        this._handleVisibilityChange.bind(this)
```
stephanwlee, line 62: I believe `bind` call creates a new reference and this is not removing the event listener attached at L55.

bmd3k: Ya, yuck. I looked at how some other components handle this and picked what looks like the least-bad option.

bmd3k:
> > FWIW, would perhaps be good to update the `Reload data every 30s.` settings panel text to reflect the fact that this is now more complex behavior, e.g. "Reload data every 30s while page is visible" or something?
> > cc @GalOshri
> 
> I am following up with Gal on it. Will send a followup PR if he thinks the message should be updated. My personal opinion is that most people would understand the new behavior without the explanation .

I followed up with Gal. He doesn't think it's necessary to change the existing text.

tensorboard/components/tf_tensorboard/autoReloadBehavior.ts
stephanwlee: ACK

PR #3485, opened 2020-04-06
build: include `__init__.py` in `//tensorboard`

### LINKED ISSUE #3484: `bazel run //tensorboard` broken with What-If Tool 1.6.0post3
This only affects TensorBoard as run via Bazel.

With `tensorboard-plugin-wit==1.6.0post3` (pushed this morning),
TensorBoard fails to start while loading the What-If Tool plugin:

```
  File "VIRTUAL_ENV/tensorboard_plugin_wit/wit_plugin_loader.py", line 51, in load
    version = pkg_resources.parse_version(tensorboard.__version__)
AttributeError: module 'tensorboard' has no attribute '__version__'
terminate called without an active exception
Fatal Python error: Aborted
```

Why? The What-If Tool dereferences `tensorboard.__version__`, which is
set in TensorBoard’s `__init__.py`, but that `__init__.py` is not
actually included in the `//tensorboard` binary. It’s in the Pip package
only as a direct dep of `//tensorboard/pip_package:build_pip_package`.

cc @jameswex

wchargin:
Summary:
Needed for `tensorboard.__version__` to resolve, as used by (e.g.) the
What-If Tool. Fixes #3484.

Test Plan:
Running `bazel run //tensorboard -- --logdir /tmp/logs` in a virtualenv
with `tensorboard-plugin-wit==1.6.0.post3` installed now works, instead
of failing with:

```
  File "/VIRTUAL_ENV/lib/python3.7/site-packages/tensorboard_plugin_wit/wit_plugin_loader.py", line 51, in load
    version = pkg_resources.parse_version(tensorboard.__version__)
AttributeError: module 'tensorboard' has no attribute '__version__'
terminate called without an active exception
Fatal Python error: Aborted
```

wchargin-branch: build-include-init

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -25,6 +25,7 @@ py_binary(
     deps = [
         ":default",
         ":dynamic_plugins",
+        ":lib",
         ":program",
         "//tensorboard:expect_tensorflow_installed",
         "//tensorboard/plugins:base_plugin",


wchargin:
You’re right that it’s not strictly necessary, but I think that it’s
good to keep in the spirit of strict deps. That is, it’s something that
the Pip package semantically needs to provide, and is only incidentally
provided by `//tensorboard`.

Does this sound reasonable to you, or would you prefer that I remove it?

stephanwlee:
Ah, strict deps. I forgot that building pip package uses the init. Current change LGTM.

PR #3486, opened 2020-04-06
application: on plugin load failure, log and continue

wchargin:
Summary:
Prior to this change, any exception raised from `loader.load()` would
crash TensorBoard (see, e.g., #3484). As of this patch, we log an error,
drop the plugin, and continue loading.

Test Plan:
Run `bazel run //tensorboard -- --logdir /tmp/logs` in a virtualenv with
`tensorboard-plugin-wit==1.6.0.post3` but *without* the patch in #3485
(i.e., run `git revert e59e9c4f376a`). Note that TensorBoard now prints:

```
E0406 11:00:39.494363 140089282381632 application.py:260] Failed to load plugin WhatIfToolPluginLoader.load; ignoring it.
  File ".../tensorboard/backend/application.py", line 255, in TensorBoardWSGIApp
    plugin = loader.load(context)
  File "/VIRTUAL_ENV/lib/python3.7/site-packages/tensorboard_plugin_wit/wit_plugin_loader.py", line 51, in load
    version = pkg_resources.parse_version(tensorboard.__version__)
AttributeError: module 'tensorboard' has no attribute '__version__'
Serving TensorBoard on localhost; to expose to the network, use a proxy or pass --bind_all
TensorBoard 2.3.0a0 at http://localhost:6006/ (Press CTRL+C to quit)
```

wchargin-branch: application-load-failure-robustness


wchargin:
The `PY2AND3` → `PY3` change is safe, because there are no
Google-internal Python 2 binaries that depend on this target.

tensorboard/backend/application.py
```diff
@@ -251,7 +251,15 @@ def TensorBoardWSGIApp(
     experimental_plugins = []
     for plugin_spec in plugins:
         loader = make_plugin_loader(plugin_spec)
-        plugin = loader.load(context)
+        try:
+            plugin = loader.load(context)
+        except Exception:
+            logger.error(
+                "Failed to invoke plugin loader %s",
```
stephanwlee, line 258: Is this supposed to be user visible error string? Not sure how much user cares about "invoking plugin loader". I think it should say something like "Failed to load plugin %s and ignoring it".

wchargin: Yes, it’s user-facing, and you’re right, it should be improved. Took
your wording; thanks.

wchargin: (It’s slightly awkward because the actual plugin name is given by the
result of `load`, so all we have is the qualname of the loader itself,
but hopefully “WhatIfToolPluginLoader.load” is clear enough.)

PR #3487, opened 2020-04-06
docs: specify Python 3 for virtualenv creation

wchargin:
Summary:
TensorBoard only supports Python 3, but on some systems `virtualenv`
defaults to Python 2, so we update the dev instructions to explicitly
specify which Python to use.

wchargin-branch: docs-virtualenv-py3


wchargin:
context: #3470 (cc @rithiksachdev)

PR #3490, opened 2020-04-06
util: add `timing.log_latency`

wchargin:
Summary:
This patch adds a decorator and context manager for latency profiling.
Any file can just add `@timing.log_latency` to a function definition or
introduce a `with timing.log_latency("some_name_here"):` context
anywhere. The output structure reveals the nesting of the call graph.

The implementation is thread-safe, and prints the current thread’s name
and ID so that the logs can easily be disentangled. (Basically, just
find a line that you’re interested in, note the thread ID, and pass the
log through `grep THREAD_ID` to get a coherent picture.)

I’ve added this as a global build dep on `//tensorboard` so that it’s
easy to patch timing in for debugging without worrying about build deps.
Any `log_latency` uses that are actually committed should be accompanied
by a strict local dep.

This patch combines previous independent drafts by @nfelt and @wchargin.

Test Plan:
Simple API usage tests included. As a manual test, adding latency
decoration to a few places in the text plugin’s Markdown rendering
pipeline generates output like this:

```
I0417 13:31:01.586319 140068322420480 timing.py:118] Thread-2[7f64329a2700] ENTER markdown_to_safe_html
I0417 13:31:01.586349 140068322420480 timing.py:118] Thread-2[7f64329a2700]   ENTER Markdown.convert
I0417 13:31:01.586455 140068322420480 timing.py:118] Thread-2[7f64329a2700]   LEAVE Markdown.convert - 0.000105s elapsed
I0417 13:31:01.586492 140068322420480 timing.py:118] Thread-2[7f64329a2700]   ENTER bleach.clean
I0417 13:31:01.586783 140068322420480 timing.py:118] Thread-2[7f64329a2700]   LEAVE bleach.clean - 0.000288s elapsed
I0417 13:31:01.586838 140068322420480 timing.py:118] Thread-2[7f64329a2700] LEAVE markdown_to_safe_html - 0.000520s elapsed
```

This is only emitted when `--verbosity 0` is passed, not by default.

Co-authored-by: Nick Felt <nfelt@users.noreply.github.com>
wchargin-branch: timing-log-latency


tensorboard/util/timing.py
```diff
@@ -0,0 +1,99 @@
...
+logger = tb_logging.get_logger()
+
+
+def log_latency(name_or_function, log_level=None):
+    """Log latency in a function or region.
+
+    This can act as a decorator: if passed a function, it returns a
+    decorated version of the function. Otherwise, the argument is
+    interpreted as a region name, and the result is either a context
+    manager or a decorator at the caller's discretion. Thus, the
```
wchargin, line 33: I am open to opinions or suggestions about this design. I considered
making the signaling more in-band by requiring the decorator to be
invoked as `@log_latency()` rather than `@log_latency`, but this has the
downside that if you accidentally invoke it as `@log_latency` then it
does the wrong thing in a confusing way.

One potential alternative would have the following semantics:

  - `log_latency()` returns a decorator;
  - `log_latency(name)` returns a context manager/decorator hybrid when
    `isinstance(name, str)`;
  - `log_latency(x)` raises a `TypeError` when `x` is not a string,
    catching the `@log_latency def foo(): pass` usage error.

wchargin: Okay, I’ve convinced myself: I’ve switched it to the API described in
this comment, and also added tests.

tensorboard/util/timing.py
```diff
@@ -0,0 +1,133 @@
...
+            "log_latency() cannot be used as a context manager unless "
+            "a region name is specified"
+        )
+
+    def __exit__(self, *exc_info):
+        pass
+
+    def __call__(self, func):
+        qualname = getattr(func, "__qualname__", func)
+        return _log_latency(qualname, self._log_level)(func)
```
nfelt, line 133: FWIW, this was a head-scratcher until I found this part of the contextlib docs:

> contextmanager() uses ContextDecorator so the context managers it creates can be used as decorators as well as in with statements. When used as a decorator, a new generator instance is implicitly created on each function call (this allows the otherwise “one-shot” context managers created by contextmanager() to meet the requirement that context managers support multiple invocations in order to be used as decorators).

This was non-obvious to me (especially the nuance about how it manages to recreate the context manager via internal machinery) and would be helpful to at least have a comment - or maybe just write out the decorator by hand to make it obvious that the context manager is being regenerated on each function call.  The fact that these context managers are also decorators feels mostly like a convenience when actually using them as decorators directly; we don't need that convenience (since we're using it in the context of our *own* decorator) so we could just be explicit about wrapping the decorated function in a `with` block.

Also, the fact that this is defined in an object that is *also* a context manager, although I see why it's there (to guard against incorrect usage of `with log_latency()`, feels extra confusing.  Reintroducing the `@log_latency` form of the decorator might make sense just so that we can make the first parameter to `log_latency()` mandatory and avoid this class of error entirely, even if it makes the contract slightly more subtle?

Or alternatively, we could perhaps at least combine this context manager with the overall `log_latency()` and always go through this context-manager-decorator rather than taking the shortcut of returning `_log_latency(...)` as a context-manager-decorator sometimes and other times returning this object.  E.g. the following:

```python
class log_latency(object):
    def __init__(self, region_name=None, log_level=None, reusable=True):
        self._context_manager = _log_latency(region_name, log_level)

    def __enter__(self):
        if self._region_name is None:
            raise TypeError(...)
        self._context_manager.__enter__()
        return self

    def __exit__(self, *exc):
        return self._context_manager.__exit__(*exc)

    def __call__(self, func):
        qualname = getattr(func, "__qualname__", func)
        return _log_latency(qualname, self._log_level)(func)
```

wchargin:
Thanks for the review! I’ll update this when I get a chance.

tensorboard/util/timing.py
```diff
@@ -0,0 +1,133 @@
...
+        raise TypeError(
+            "log_latency() cannot be used as a context manager unless "
+            "a region name is specified"
+        )
+
+    def __exit__(self, *exc_info):
+        pass
+
+    def __call__(self, func):
+        qualname = getattr(func, "__qualname__", func)
```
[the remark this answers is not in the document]
wchargin: Yeah, I was trying to avoid the overhead—changed to `None`-sentinel
check, which should be fast enough.

tensorboard/util/timing.py
wchargin: > Reintroducing the `@log_latency` form of the decorator might make
> sense just so that we can make the first parameter to `log_latency()`
> mandatory and avoid this class of error entirely, even if it makes the
> contract slightly more subtle?

Okay, done. The analogous error that we no longer have a nice message
for is `with log_latency:`, which seems less likely (more obviously
wrong at call site) and would even more hackery to fix (something like
`log_latency = _LogLatency()` or a metaclass). This is fine enough.

wchargin: Oh, meant to note—

> The fact that these context managers are also decorators feels mostly
> like a convenience when actually using them as decorators directly; we
> don't need that convenience (since we're using it in the context of
> our _own_ decorator) so we could just be explicit about wrapping the
> decorated function in a `with` block.

—in the revised version I’m still using this convenience, because it’s
simpler than doing the `functools.wraps`/`*args, **kwargs` dance
ourselves, but I’ve spelled it out a bit more to make it clearer. Agreed
that the `contextlib` magic is a bit magic.

tensorboard/util/timing.py
```diff
@@ -0,0 +1,118 @@
...
+        log_level = logging.INFO
+
+    if isinstance(region_name_or_function_to_decorate, str):
+        region_name = region_name_or_function_to_decorate
+        return _log_latency(region_name, log_level)
+    else:
+        function_to_decorate = region_name_or_function_to_decorate
+        qualname = getattr(function_to_decorate, "__qualname__", None)
+        if qualname is None:
+            qualname = str(function_to_decorate)
```
wchargin, line 76: It’s not super easy to get a handle to a function without `qualname` to
test this case, but with at least current versions of numpy the
following suffices as a manual test:

```python
>>> import logging
>>> import numpy as np
>>> from tensorboard.util import timing
>>>
>>> np.__version__
'1.18.2'
>>> np.sin.__qualname__
AttributeError: 'numpy.ufunc' object has no attribute '__qualname__'
>>> sin = timing.log_latency(np.sin, log_level=logging.WARNING)
>>> sin([1, 2, 3])
MainThread[7f81fc8d7740] ENTER <ufunc 'sin'>
MainThread[7f81fc8d7740] LEAVE <ufunc 'sin'> - 0.000302s elapsed
array([0.84147098, 0.90929743, 0.14112001])
```

(Don’t want to commit this in case [the upstream bug][1] is ever fixed.)

[1]: https://github.com/numpy/numpy/issues/4952

tensorboard/util/timing.py
nfelt: Thanks! This version is vastly easier for me to wrap my head around :)

(I guess `with log_latency:` will just raise an `AttributeError: __enter__` - but if Python 3.7 feels that's a good enough way to indicate "you tried to use something as a context manager and it wasn't one" then I think this is good enough for us.)

PR #3492, opened 2020-04-07
Support comma-separated --plugins lists for "dev" subcommand

bmd3k:

Address feedback for the new `--plugins` flag to support comma-separated lists of plugins instead of space-separated lists.

User can still specify plugins like this:
```
tensorboard dev upload --logdir <your logdir> --plugins=scalars
```

User can no longer specify multiple plugins like this:
```
tensorboard dev upload --logdir <your logdir> --plugins scalars graphs histograms
```

User must now instead specify mutiple plugins like this:
```
tensorboard dev upload --logdir <your logdir> --plugins=scalars,graphs,histograms
tensorboard dev upload --logdir <your logdir> --plugins scalars,graphs,histograms
```

Modify `--plugins` add_argument call by removing `nargs="*"` and replacing with lambda to split input by `","`.

Run the command with the various versions of the `--plugins` flag. Added tests.

### DIFF
--- a/tensorboard/uploader/flags_parser.py
+++ b/tensorboard/uploader/flags_parser.py
@@ -1,0 +1,203 @@
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Flags parser for TensorBoard.dev uploader."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from absl.flags import argparse_flags
+
+SUBCOMMAND_FLAG = "_uploader__subcommand"
+SUBCOMMAND_KEY_UPLOAD = "UPLOAD"
+SUBCOMMAND_KEY_DELETE = "DELETE"
+SUBCOMMAND_KEY_LIST = "LIST"
+SUBCOMMAND_KEY_EXPORT = "EXPORT"
+SUBCOMMAND_KEY_UPDATE_METADATA = "UPDATEMETADATA"
+SUBCOMMAND_KEY_AUTH = "AUTH"
+AUTH_SUBCOMMAND_FLAG = "_uploader__subcommand_auth"
+AUTH_SUBCOMMAND_KEY_REVOKE = "REVOKE"
+
+DEFAULT_ORIGIN = "https://tensorboard.dev"
+
+
+def parse_flags(argv):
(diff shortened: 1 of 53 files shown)


tensorboard/uploader/flags_parser.py
```diff
@@ -201,3 +203,11 @@ def define_flags(parser):
     auth_revoke.set_defaults(
         **{AUTH_SUBCOMMAND_FLAG: AUTH_SUBCOMMAND_KEY_REVOKE}
     )
+
+
+def fix_flags(flags):
+    if hasattr(flags, "plugins"):
+        expanded_plugins = []
+        for plugin in flags.plugins:
+            expanded_plugins.extend(str.split(plugin, ","))
```
wchargin, line 212: It seems most confusing to support *both* comma-separated and tokenized
flag values. If we want to pick comma-separated flag values, can we drop
the `nargs="*"`?

If we’re worried about compatibility, we can always push a 2.2.1 release
with this cherry-pick before launching any features that depend on this.
Presumably we’d need to do that anyway: if we want comma-separated to be
the documented and encouraged form, then it’ll need to be published
before we use it in examples.

tensorboard/uploader/flags_parser.py
```diff
@@ -201,3 +203,11 @@ def define_flags(parser):
     auth_revoke.set_defaults(
         **{AUTH_SUBCOMMAND_FLAG: AUTH_SUBCOMMAND_KEY_REVOKE}
     )
+
+
+def fix_flags(flags):
```
wchargin, line 208: Instead of adding a `fix_flags` mechanism, can we just use the `type`
kwarg to `parser.add_argument` to control the parsing ourselves? Like:

```python
    upload.add_argument(
        "--plugins",
        type=lambda s: s.split(","),
        default=[],
        help="...",
    )
```

and then you’re just done, I think?

If the parsing is nontrivial and needs to be done in multiple places,
you can add a helper function, like `_parse_samples_per_plugin` from
`backend/application.py`. But I’m not sure that adding a whole new axis
to the subcommand layer is necessary.

bmd3k: Yes, this is trivial if all we want is comma-separated lists.

wchargin: Even if we wanted to support both comma-separated and tokenized lists,
it would be easy to accomplish with a simple custom action:

```python
import argparse


class ExtendFromAction(argparse.Action):
    def __call__(self, parser, namespace, values, option_string=None):
        items = getattr(namespace, self.dest, None) or []
        items.extend(x for xs in values for x in xs)
        setattr(namespace, self.dest, items)


parser = argparse.ArgumentParser()
parser.add_argument(
    "--plugins", nargs="*", action=ExtendFromAction, type=lambda x: x.split(",")
)

print(parser.parse_args(["--plugins", "a,b", "c,d", "--plugins", "e,f"]))
# prints: Namespace(plugins=['a', 'b', 'c', 'd', 'e', 'f'])
```

Or, we could just use `nargs="*", type=lambda x: x.split(",")` and
flatten with a helper function in the application code.

The `fix_flags` used by the plugin system is mainly used to signal
errors due to _interactions_ of flags: “if inspect-mode is set but not
exactly one of a logdir or event file is given, then die”. As long as
all we’re doing is controlling the parsing of a _single_ flag, I’d
prefer to do so with `type=` or `action=` or similar rather than adding
a new mechanism.

PR #3494, opened 2020-04-07
ng: make sure we do not push 'null' onto hash

stephanwlee:
In the Angular version of TensorBoard, we keep the activePlugin state in
the Ngrx Store. In it, when activePlugin was never set, we use `null` to
denote the emptiness. However, this null state was incompatible with how
hash_storage works.

tf_storage coerces the input, which wrongly coerced the null state
(Angular build tools did not catch the wrong typing) in the activePlugin
to "null" and set the hash as "#null".

This change fixes the issue by converting `null` to '' which tf_storage
understands to be the empty state.

### DIFF
--- a/tensorboard/webapp/core/views/hash_storage_test.ts
+++ b/tensorboard/webapp/core/views/hash_storage_test.ts
@@ -70,6 +70,17 @@ describe('hash storage test', () => {
     fixture.detectChanges();
     expect(setPluginIdSpy).toHaveBeenCalledWith('foo', {
       useLocationReplace: true,
+      defaultValue: '',
+    });
+  });
+
+  it('sets the hash to empty string when activePlugin is not set', () => {
+    store.overrideSelector(getActivePlugin, null);
+    const fixture = TestBed.createComponent(HashStorageContainer);
+    fixture.detectChanges();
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/core/views/hash_storage_test.ts
```diff
@@ -70,6 +70,17 @@ describe('hash storage test', () => {
     fixture.detectChanges();
     expect(setPluginIdSpy).toHaveBeenCalledWith('foo', {
       useLocationReplace: true,
+      defaultValue: '',
+    });
+  });
+
+  it('sets the hash to empty string when activePlugin is not set', () => {
```
caisq, line 77: You might to add a test for `getActivePlugin()` returning empty string. Not sure if that's a possible scenario.

stephanwlee: That is not a possible scenario. plugin id is a required field (although neither python nor JS part checks it) and expected to be non-empty.

stephanwlee: In the hindsight, it is technically possible to have the empty string when the store bootstraps from the empty hash (`/#`). Added the case.

PR #3495, opened 2020-04-08
uploader: use fixed (UTC) timezone in formatter unit tests

caisq:
  * @stephanwlee brought to my attention a test failure that happens under certain test configs in `//tensorboard/uploader:formatters_test`
  * This failure has to do with different timezone settings that are apparently used in different test environments
  * Add a helper method that sets system timezone to "UTC" and use the helper method from the unit tests.

### DIFF
--- a/tensorboard/uploader/formatters_test.py
+++ b/tensorboard/uploader/formatters_test.py
@@ -161,6 +161,43 @@ class TensorBoardExporterTest(tb_test.TestCase):
         ]
         self.assertEqual(output.split("\n"), expected_lines)
 
+    def testJsonFormatterWithNonUtcTimezone(self):
+        experiment = experiment_pb2.Experiment(
+            experiment_id="deadbeef",
+            # NOTE(cais): `name` and `description` are missing here.
+            num_runs=2,
+            num_tags=4,
+            num_scalars=60,
+            total_blob_bytes=1234,
+        )
+        util.set_timestamp(experiment.create_time, 981173106)
+        util.set_timestamp(experiment.update_time, 1015218367)
+        experiment_url = "http://tensorboard.dev/deadbeef"
+        formatter = formatters.JsonFormatter()
+        output = self._format(
+            formatter,
+            experiment,
+            experiment_url,
+            timezone="America/Los_Angeles",
+        )
+        expected_lines = [
+            "{",
+            '  "url": "http://tensorboard.dev/deadbeef",',
+            '  "name": "",',
+            '  "description": "",',
+            '  "id": "deadbeef",',
+            # NOTE(cais): Here we assert that the JsonFormat output is not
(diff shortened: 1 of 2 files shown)


tensorboard/uploader/formatters_test.py
```diff
@@ -27,6 +36,15 @@
 
 
 class TensorBoardExporterTest(tb_test.TestCase):
+    def _run(self, formatter, experiment, experiment_url):
+        """Test helper that ensures formatting is done with fixed timezone."""
+        try:
+            with mock.patch.dict(os.environ, {"TZ": "UTC"}):
+                time.tzset()
+                return formatter.format_experiment(experiment, experiment_url)
```
stephanwlee, line 44: `run` and passing formatter do not semantically correspond. Please name the utility something like `format_experiment` instead?

caisq: Point taken. I revised it to `_format`. I don't want to be too verbose.

tensorboard/uploader/formatters_test.py
```diff
@@ -19,6 +19,15 @@
...
 from __future__ import print_function
 
+import os
+import time
+
+try:
+    # python version >= 3.3
+    from unittest import mock
+except ImportError:
+    import mock  # pylint: disable=unused-import
```
wchargin, line 29: All our unit tests run exclusively in Python 3. Please just import
`unittest.mock`. (We’d like to remove the `mock` backport. It has lots
of problems.)

tensorboard/uploader/formatters_test.py
```diff
@@ -97,7 +115,7 @@ def testJsonFormatterWithEmptyNameAndDescription(self):
         util.set_timestamp(experiment.update_time, 1015218367)
         experiment_url = "http://tensorboard.dev/deadbeef"
         formatter = formatters.JsonFormatter()
-        output = formatter.format_experiment(experiment, experiment_url)
+        output = self._format(formatter, experiment, experiment_url)
```
wchargin, line 118: Can we test that the JSON formatter *is* timezone-independent even
though the readable formatter is not?

tensorboard/uploader/formatters_test.py
caisq: Done for this file. There are a few instances of this importing pattern in other parts of the tensorboard/uploader code. That can be fixed in a separate PR.

tensorboard/uploader/formatters_test.py
caisq: Done.

tensorboard/uploader/formatters_test.py
wchargin: Yep, of course—no imminent need to update existing code, just would
prefer to avoid introducing it in new code.

PR #3497, opened 2020-04-08
Prefilter graphs in the uploader

davidsoergel:
There is no point in uploading GraphDef data that won't be displayed anyway.  In particular, the TensorBoard frontend filters out node attributes larger that 1024 bytes, since it has no good way to present those.  So we may as well filter those out before upload to TensorBoard.dev, so as not to waste bandwidth, storage, and read-time processing.

### DIFF
--- a/tensorboard/backend/process_graph.py
+++ b/tensorboard/backend/process_graph.py
@@ -47,7 +47,8 @@ def prepare_graph_for_ui(
       ValueError: If `limit_attr_size` is defined, but <= 0.
     """
     # TODO(@davidsoergel): detect whether a graph has been filtered already
-    # (to a limit_attr_size <= what is requested here).
+    # (to a limit_attr_size <= what is requested here).  If it is already
+    # filtered, return immediately.
 
     # Check input for validity.
     if limit_attr_size is not None:
--- a/tensorboard/backend/process_graph.py
+++ b/tensorboard/backend/process_graph.py
@@ -46,6 +46,9 @@ def prepare_graph_for_ui(
       ValueError: If `large_attrs_key is None` while `limit_attr_size != None`.
       ValueError: If `limit_attr_size` is defined, but <= 0.
     """
+    # TODO(@davidsoergel): detect whether a graph has been filtered already
+    # (to a limit_attr_size <= what is requested here).
+
     # Check input for validity.
     if limit_attr_size is not None:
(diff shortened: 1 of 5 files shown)


tensorboard/dataclass_compat.py
```diff
@@ -37,7 +40,7 @@
 from tensorboard.util import tensor_util
 
 
-def migrate_event(event):
+def migrate_event(event, experimental_filter_graph=False):
```
caisq, line 43: Please add doc string for the new kwarg.

tensorboard/backend/process_graph.py
```diff
@@ -46,6 +46,9 @@ def prepare_graph_for_ui(
       ValueError: If `large_attrs_key is None` while `limit_attr_size != None`.
       ValueError: If `limit_attr_size` is defined, but <= 0.
     """
+    # TODO(@davidsoergel): detect whether a graph has been filtered already
```
caisq, line 49: To make this TODO item clearer, I believe you can say in addition: "if it is already filtered, return immediately".

tensorboard/dataclass_compat_test.py
```diff
@@ -212,6 +215,49 @@ def test_graph_def(self):
 
         self.assertProtoEquals(graph_def, new_graph_def)
 
+    def test_graph_def_experimental_filter_graph(self):
+        # Create a `GraphDef` and write it to disk as an event.
+        logdir = self.get_temp_dir()
```
caisq, line 220: Question about testing strategy: is it really necessary to create a logdir and a writer? If your intention here is to simply test `self._migrate_event()` (and ultimately `dataclass_compat.migrate_event()`), you can just construct a `Event`, pass it to that method and check the result, right?

tensorboard/backend/process_graph.py
davidsoergel: Done.

tensorboard/dataclass_compat.py
davidsoergel: Oops, done.

tensorboard/dataclass_compat_test.py
davidsoergel: Oh good point.  I had pasted that from the test above but here the logdir part is superfluous.  (There too, maybe, but that's out of scope).  Thanks!

PR #3498, opened 2020-04-09
Disable auto reload if page is not visible in ng_tensorboard

bmd3k:
Note that this is the ng_tensorboard version of #3483.

Reduce the load on TensorBoard servers by disabling auto reload if the page is not considered visible.

Do not auto reload if document is not visible according to the Page
Visibility API. When page again becomes visible perform an auto reload
if one has been missed.

Run a local TensorBoard.
Open localhost:6006.
Ensure option "Reload data" is selected and choose a 15s "Reload Period".
Open detached network tab in chrome developer tools.
Watch for several minutes and observe that many network calls are made every 15s.
Open different tab in same window but with chrome developer tools still visible.
Watch for several minutes and observe that no network calls are made.
Switch back to TensorBoard tab and watch as network calls are immediately made and UI reloads graphs.
Continue to watch network tab as it again follows a 15s period for reload network calls.

### DIFF
--- a/tensorboard/webapp/reloader/reloader_component.ts
+++ b/tensorboard/webapp/reloader/reloader_component.ts
@@ -12,7 +12,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-import {Component, ChangeDetectionStrategy} from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+import {Component, ChangeDetectionStrategy, Inject} from '@angular/core';
 import {Store, select} from '@ngrx/store';
 import {combineLatest} from 'rxjs';
 import {distinctUntilChanged} from 'rxjs/operators';
@@ -21,16 +22,6 @@ import {getReloadEnabled, getReloadPeriodInMs, State} from '../core/store';
 import {reload} from '../core/actions';
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
-
-const documentUtil = {
-  /**
-   * Wraps Page Visibility API call to determine if document is visible.
-   * Can be overriden for testing purposes.
-   */
-  getVisibilityState(): string {
-    return document.visibilityState;
-  },
-};
 
 @Component({
   selector: 'reloader',
@@ -46,10 +37,13 @@ export class ReloaderComponent {
   private reloadTimerId: ReturnType<typeof setTimeout> | null = null;
   private missedAutoReload: boolean = false;
 
-  constructor(private store: Store<State>) {}
+  constructor(
+    private store: Store<State>,
+    @Inject(DOCUMENT) private document: Document
+  ) {}
 
   ngOnInit() {
-    window.addEventListener('visibilitychange', this.onVisibilityChange);
+    this.document.addEventListener('visibilitychange', this.onVisibilityChange);
     combineLatest(
       this.reloadEnabled$.pipe(distinctUntilChanged()),
       this.reloadPeriodInMs$.pipe(distinctUntilChanged())
@@ -62,10 +56,7 @@ export class ReloaderComponent {
   }
 
   private onVisibilityChangeImpl() {
-    if (
-      documentUtil.getVisibilityState() === 'visible' &&
-      this.missedAutoReload
-    ) {
+    if (this.document.visibilityState === 'visible' && this.missedAutoReload) {
       this.missedAutoReload = false;
       this.store.dispatch(reload());
     }
@@ -73,7 +64,7 @@ export class ReloaderComponent {
 
   private load(reloadPeriodInMs: number) {
     this.reloadTimerId = setTimeout(() => {
-      if (documentUtil.getVisibilityState() === 'visible') {
+      if (this.document.visibilityState === 'visible') {
         this.store.dispatch(reload());
       } else {
         this.missedAutoReload = true;
@@ -91,8 +82,9 @@ export class ReloaderComponent {
 
   ngOnDestroy() {
     this.cancelLoad();
-    window.removeEventListener('visibilitychange', this.onVisibilityChange);
+    this.document.removeEventListener(
+      'visibilitychange',
+      this.onVisibilityChange
+    );
   }
 }
-
-export const TEST_ONLY = {documentUtil};
--- a/tensorboard/webapp/reloader/reloader_component.ts
+++ b/tensorboard/webapp/reloader/reloader_component.ts
@@ -22,21 +22,34 @@ import {reload} from '../core/actions';
 
 /** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
+const documentUtil = {
+  /**
+   * Wraps Page Visibility API call to determine if document is visible.
+   * Can be overriden for testing purposes.
+   */
+  getVisibilityState(): string {
+    return document.visibilityState;
+  },
+};
+
 @Component({
   selector: 'reloader',
   template: '',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ReloaderComponent {
+  private readonly onVisibilityChange = this.onVisibilityChangeImpl.bind(this);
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/reloader/reloader_component.ts
```diff
@@ -22,21 +22,34 @@ import {reload} from '../core/actions';
...
   private readonly reloadPeriodInMs$ = this.store.pipe(
     select(getReloadPeriodInMs)
   );
   private reloadTimerId: ReturnType<typeof setTimeout> | null = null;
+  private missedAutoReload: boolean = false;
 
   constructor(private store: Store<State>) {}
 
   ngOnInit() {
+    window.addEventListener('visibilitychange', this.onVisibilityChange);
```
stephanwlee, line 52: rxjs has utility called `fromEvent` and you can write this in more Rxjs way.

Pseudo code
```ts
fromEvent(window. 'visibilitychange').pipe(
  filter(() => {
    return documentUtil.getVisibilityState() === 'visible';
  }),
  tap(() => {
    this.store.dispatch(reload()),
  }),
);
```

Of course, since we do not really want hard dependency on DOM, you may also want to DI the `Document` helper or https://angular.io/api/common/DOCUMENT.

tensorboard/webapp/reloader/reloader_component.ts
```diff
@@ -48,9 +61,23 @@ export class ReloaderComponent {
...
+      this.missedAutoReload = false;
+      this.store.dispatch(reload());
+    }
+  }
+
   private load(reloadPeriodInMs: number) {
     this.reloadTimerId = setTimeout(() => {
-      this.store.dispatch(reload());
+      if (documentUtil.getVisibilityState() === 'visible') {
+        this.store.dispatch(reload());
```
stephanwlee, line 77: Can this make the Reloader reload twice in almost succession? i.e., the other reload is still scheduled to be fired and can dispatch reload event right after this.

tensorboard/webapp/reloader/reloader_component.ts
bmd3k: I went with the approach of injecting the Document.  Since some of the test framework relies on the injected Document it isn't totally straightforward to fake in the tests and is a bit brittle.  Take a look at `fakeDocument` and let me know what you think.

tensorboard/webapp/reloader/reloader_component.ts
bmd3k: Yes. I didn't think it would be a big deal but I can easily be convinced otherwise.

tensorboard/webapp/reloader/reloader_component.ts
```diff
@@ -28,15 +29,21 @@ import {reload} from '../core/actions';
...
   private readonly reloadPeriodInMs$ = this.store.pipe(
     select(getReloadPeriodInMs)
   );
   private reloadTimerId: ReturnType<typeof setTimeout> | null = null;
+  private missedAutoReload: boolean = false;
 
-  constructor(private store: Store<State>) {}
+  constructor(
+    private store: Store<State>,
+    @Inject(DOCUMENT) private document: Document
```
stephanwlee, line 42: `private readonly`

tensorboard/webapp/reloader/reloader_component_test.ts
```diff
@@ -161,4 +182,158 @@ describe('reloader_component', () => {
...
+    fixture.detectChanges();
+
+    tick(5);
+    expect(dispatchSpy).toHaveBeenCalledTimes(1);
+
+    simulateVisibilityChange(false);
+    tick(5);
+    expect(dispatchSpy).toHaveBeenCalledTimes(1);
+
+    fixture.destroy();
```
stephanwlee, line 205: do this in `afterEach`?

tensorboard/webapp/reloader/reloader_component_test.ts
```diff
@@ -47,6 +62,11 @@ describe('reloader_component', () => {
     dispatchSpy = spyOn(store, 'dispatch');
   });
 
+  function simulateVisibilityChange(visible: boolean) {
+    fakeDocument.visibilityState = visible ? 'visible' : 'hidden';
```
stephanwlee, line 66: do note that you are modifying the object over and over and it can lead to flaky test depending on random seed. Better way would be for us to create a new object on each `beforeEach`. 

```ts
function createFakeDocument() {
  return {...}
}

let document: DOCUMENT;

beforeEach(async () => {
      providers: [
        {
          provide: DOCUMENT,
          useFactory: createFakeDocument
        },
        ...
  );
  document = TestBed.get(Document);
});

```

tensorboard/webapp/reloader/reloader_component_test.ts
```diff
@@ -161,4 +182,158 @@ describe('reloader_component', () => {
...
+    expect(dispatchSpy).toHaveBeenCalledTimes(0);
+
+    // Dispatch a reload when next visible.
+    simulateVisibilityChange(true);
+    expect(dispatchSpy).toHaveBeenCalledTimes(1);
+
+    fixture.destroy();
+  }));
+
+  it('reloads when document becomes visible if missed reload, regardless of how long not visible', fakeAsync(() => {
```
stephanwlee, line 232: Is this longer than 80?

tensorboard/webapp/reloader/reloader_component_test.ts
bmd3k: `fixture` is scoped to the test function. Is it worth instead scoping it more broadly?

tensorboard/webapp/reloader/reloader_component_test.ts
bmd3k: Yes. But this seems to be how the formatter wants it (both the prettier vscode extension and prettier in `yarn fix-lint`.)  We could work around it by shortening or breaking up the string into multiple lines. Is it worth it?

tensorboard/webapp/reloader/reloader_component_test.ts
bmd3k: Done. But didn't type fakeDocument as Document since it would require additional hoops to fake visibilityState.

tensorboard/webapp/reloader/reloader_component.ts
bmd3k: Done.

tensorboard/webapp/reloader/reloader_component_test.ts
stephanwlee: feel free to hoist it if you find it tedious, too. Else, ignore this :)

tensorboard/webapp/reloader/reloader_component_test.ts
```diff
@@ -44,6 +67,7 @@ describe('reloader_component', () => {
       declarations: [ReloaderComponent],
     }).compileComponents();
     store = TestBed.get(Store);
+    fakeDocument = TestBed.get(DOCUMENT);
```
stephanwlee, line 70: just wondering, what happens if you change this to `TestBed.get(DOCUMENT) as typeof document`?
I want to avoid `any` where possible.

tensorboard/webapp/reloader/reloader_component_test.ts
bmd3k: I attempted to do this. I agree it's tedious. Unfortunately fakeAsync() will complain that "1 timer(s) still in the queue." if fixture isn't destroyed before the end of the test (because the destroy clears a Timeout). I could not find any documentation explaining how to make this work with fixture.destroy() in afterEach().

tensorboard/webapp/reloader/reloader_component_test.ts
bmd3k: `typeof document` doesn't seem to help.  I can (and have) made `fakeDocument: Document` but it requires some slight grossness to override the visibilityState property. That's ok - I think the grossness is outweighed by the better typing.

stephanwlee: Acknowledged.

PR #3503, opened 2020-04-10
Detect and skip corrupt GraphDefs

davidsoergel:
Since #3497, we parse `GraphDef`s in `dataclass_compat.py` during upload.  If a graph is corrupt, that parsing fails.  Here we catch the resulting exception, issue a warning, and continue (omitting the graph).

This also updates tests to use valid GraphDefs where appropriate, as opposed to `bytes(1024)`, which apparently produces inconsistent results with different proto parsers (e.g., OSS vs. Google internal).

### DIFF
--- a/tensorboard/dataclass_compat.py
+++ b/tensorboard/dataclass_compat.py
@@ -26,7 +26,6 @@ from __future__ import division
 from __future__ import print_function
 
 
-
 from google.protobuf import message
 from tensorboard.backend import process_graph
 from tensorboard.compat.proto import event_pb2
--- a/tensorboard/dataclass_compat.py
+++ b/tensorboard/dataclass_compat.py
@@ -26,6 +26,8 @@ from __future__ import division
 from __future__ import print_function
 
 
+
+from google.protobuf import message
 from tensorboard.backend import process_graph
 from tensorboard.compat.proto import event_pb2
 from tensorboard.compat.proto import graph_pb2
@@ -38,6 +40,9 @@ from tensorboard.plugins.image import metadata as images_metadata
 from tensorboard.plugins.scalar import metadata as scalars_metadata
 from tensorboard.plugins.text import metadata as text_metadata
 from tensorboard.util import tensor_util
+from tensorboard.util import tb_logging
+
+logger = tb_logging.get_logger()
 
 
 def migrate_event(event, experimental_filter_graph=False):
@@ -72,11 +77,15 @@ def _migrate_graph_event(old_event, experimental_filter_graph=False):
 
     # TODO(@davidsoergel): Move this stopgap to a more appropriate place.
     if experimental_filter_graph:
-        graph_def = graph_pb2.GraphDef().FromString(graph_bytes)
-        # Use the default filter parameters:
-        # limit_attr_size=1024, large_attrs_key="_too_large_attrs"
-        process_graph.prepare_graph_for_ui(graph_def)
-        graph_bytes = graph_def.SerializeToString()
+        try:
+            graph_def = graph_pb2.GraphDef().FromString(graph_bytes)
+            # Use the default filter parameters:
+            # limit_attr_size=1024, large_attrs_key="_too_large_attrs"
+            process_graph.prepare_graph_for_ui(graph_def)
+            graph_bytes = graph_def.SerializeToString()
+        except message.DecodeError:
+            logger.warning("Could not parse GraphDef.  Skipping.")
+            return (old_event,)
 
     value.tensor.CopyFrom(tensor_util.make_tensor_proto([graph_bytes]))
     value.metadata.plugin_data.plugin_name = graphs_metadata.PLUGIN_NAME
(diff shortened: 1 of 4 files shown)


tensorboard/dataclass_compat_test.py
```diff
@@ -253,6 +253,23 @@ def test_graph_def_experimental_filter_graph(self):
...
+        # others don't (silently producing an empty proto, I guess).
+        # Thus `old_event.graph_def = bytes(1024)` is an unreliable example.
+        old_event.graph_def = b"bogus"
+
+        new_events = self._migrate_event(
+            old_event, experimental_filter_graph=True
+        )
+        # _migrate_event emits both the original event and the migrated event,
+        # but here there is no migrated event becasue the graph was unparseable.
+        self.assertLen(new_events, 1)
```
caisq, line 271: Also assert that `new_events[0]` is equal to `old_event`.

tensorboard/dataclass_compat.py
```diff
@@ -72,11 +76,15 @@ def _migrate_graph_event(old_event, experimental_filter_graph=False):
...
-        process_graph.prepare_graph_for_ui(graph_def)
-        graph_bytes = graph_def.SerializeToString()
+        try:
+            graph_def = graph_pb2.GraphDef().FromString(graph_bytes)
+            # Use the default filter parameters:
+            # limit_attr_size=1024, large_attrs_key="_too_large_attrs"
+            process_graph.prepare_graph_for_ui(graph_def)
+            graph_bytes = graph_def.SerializeToString()
+        except message.DecodeError:
+            logger.warning("Could not parse GraphDef.  Skipping.")
```
caisq, line 86: Would it be nicer to print a more information, including how big the GraphDef is, to facilitate user debugging in case necessary?

tensorboard/uploader/uploader_test.py
```diff
@@ -342,7 +356,7 @@ def test_upload_server_error(self):
 
         # Of course a real Event stream will never produce the same Event twice,
         # but is this test context it's fine to reuse this one.
-        graph_event = event_pb2.Event(graph_def=bytes(950))
+        graph_event = event_pb2.Event(graph_def=_create_example_graph(950))
```
caisq, line 359: 1. Just so that I understand: some of the changes like these are not strictly necessary to fix the test, but are just so that we are using valid GraphDefs consistently right?

2. Previously the graph_def has an exact size of 950. But now it's >950. Do you think there is the risk of breaking something?

tensorboard/dataclass_compat.py
```diff
@@ -72,11 +76,15 @@ def _migrate_graph_event(old_event, experimental_filter_graph=False):
...
-        # Use the default filter parameters:
-        # limit_attr_size=1024, large_attrs_key="_too_large_attrs"
-        process_graph.prepare_graph_for_ui(graph_def)
-        graph_bytes = graph_def.SerializeToString()
+        try:
+            graph_def = graph_pb2.GraphDef().FromString(graph_bytes)
+            # Use the default filter parameters:
+            # limit_attr_size=1024, large_attrs_key="_too_large_attrs"
+            process_graph.prepare_graph_for_ui(graph_def)
+            graph_bytes = graph_def.SerializeToString()
```
caisq, line 84: You can move these two lines outside the try scope, IIUC.

tensorboard/uploader/uploader_test.py
```diff
@@ -50,6 +51,19 @@
 from tensorboard.util import test_util as tb_test_util
 
 
+def _create_example_graph(test_attr_size):
```
caisq, line 54: Rename this to _create_example_graph_bytes to make it clear that it'll return `bytes`, not a `GraphDef`.

caisq, line 54: The arg name `test_attr_size` is potentially misleading. How about calling it `large_test_attr_size` to reflect the fact that it controls only one of the two attrs that is meant to be bigger?

tensorboard/dataclass_compat_test.py
davidsoergel: Done.

tensorboard/dataclass_compat.py
davidsoergel: Done.

tensorboard/dataclass_compat.py
davidsoergel: I added the size, but at this point I don't think there's anything else we can say about the graph itself, since it's not parseable.  I guess we have the Event metadata, but I don't think the step number (always 0 for graphs, in practice) is relevant here.

tensorboard/uploader/uploader_test.py
davidsoergel: 1.) Correct.  2.) Also correct, and I was worried about this too because we test e.g. how many 100-byte chunks are needed to transmit the entire blob.  As it happens, the actual size is still less than 1000, so the number of chunks is still 10 and the tests work as is.  (If this had been an issue, I would have just reduced the attr size to 925 or something to compensate).

tensorboard/uploader/uploader_test.py
davidsoergel: Done.

davidsoergel: Done as `large_attr_size`

PR #3506, opened 2020-04-12
[DebuggerV2] Add basic store support and component for graph execution

caisq:
  * Continue developing DebuggerV2 plugin, specifically the part that focuses on intra-graph execution.
  * debugger_types: add `graphExecutions` field to `DebuggerState`
  * tfdbg2_data_source.ts: add `fetchGraphExecutionDigests()` and `fetchGraphExecutionData()`
  * debugger_actions.ts: add action for requesting # of graph executions and action that signifies the successful loading of the # of graph executions
  * debugger_reducers.ts: add reducers for the said actions and states
  * views/graph_executions folder: add component and container for visualizing graph executions. Currently only the number of graph executions is shown. Follow-up PRs will add a virtual scroll element that shows the graph execution digests.
  * Update the CSS and layout in the top-section of DebuggerV2 Plugin to accommodate the new `GraphExecutionContainer`.
* Screenshots of UI changes
  * image
  * Unit tests added
  * Running the DebuggerV2 plugin against a logdir with real tfdbg2 dump data


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -165,14 +202,20 @@ export interface Alerts {
   focusType: AlertType | null;
 }
 
-export interface Executions {
-  // Load state for the total number of top-level executions.
+/**
+ * Base interface shared between top-level and intra-graph executions.
+ *
+ * Supports paged lazy loading of digess (i.e., concise data objects
```
stephanwlee, line 208: s/digess/digest 

`Supports paged lazy loading of digest` doesn't read nicely properly in my head. It may be just me.

caisq: Fixed the typo and tweaked the wording here for clarification.

caisq:
@stephanwlee about the question of why separate RPCs for single numbers. Yes, there are three two such rpcs currently:
- `/executions/digests` for the number of top-level (eager) executions
- `/graph_execution/digests` for the number of intra-graph executions. 

Theoretically, they can be grouped into a single rpc call. But the argument for not combining them include:
1. It's nice to have a conceptual separation between the two. They do use different tags at the DataProvider level.
2. There won't be more of them. It's not like there is a third or fourth type of execution. So there won't be too much concern about a proliferation of such things in the future.

PR #3508, opened 2020-04-13
Catch RuntimeWarning when parsing GraphDef protos.

davidsoergel:


davidsoergel:
Added mock-based tests.

PR #3509, opened 2020-04-13
Uploader: Add test for completely empty GraphDef blob. 

davidsoergel:


tensorboard/dataclass_compat_test.py
```diff
@@ -271,6 +271,27 @@ def test_graph_def_experimental_filter_graph_corrupt(self):
...
+        # _migrate_event emits both the original event and the migrated event.
+        # The migrated event contains a default empty GraphDef.
+        self.assertLen(new_events, 2)
+        self.assertProtoEquals(new_events[0], old_event)
+
+        new_event = new_events[1]
+        tensor = tensor_util.make_ndarray(new_event.summary.value[0].tensor)
+        new_graph_def_bytes = tensor[0]
+        new_graph_def = graph_pb2.GraphDef.FromString(new_graph_def_bytes)
+        self.assertProtoEquals(new_graph_def, graph_pb2.GraphDef())
```
caisq, line 293: Can you add a comment here to explain what these lines are doing and how they are related to the lines above?

davidsoergel: Simplified the test (the last few lines were quite superfluous) and added comments.

bileschi (approved):

PR #3510, opened 2020-04-14
uploader: inline graph filtering from `dataclass_compat`

wchargin:
Summary:
We initially used `dataclass_compat` to perform filtering of large
graphs as a stopgap mechanism. This commit moves that filtering into the
uploader, which is the only surface in which it’s actually used. As a
result, `dataclass_compat` no longer takes extra arguments and so can be
moved into `EventFileLoader` in a future change.

Test Plan:
Unit tests added to the uploader for the small graph, large graph, and
corrupt graph cases.

wchargin-branch: uploader-graph-filtering

### DIFF
--- a/tensorboard/uploader/uploader.py
+++ b/tensorboard/uploader/uploader.py
@@ -28,6 +28,7 @@ import six
 from google.protobuf import message
 from tensorboard.compat.proto import graph_pb2
 from tensorboard.compat.proto import summary_pb2
+from tensorboard.compat.proto import types_pb2
 from tensorboard.uploader.proto import write_service_pb2
 from tensorboard.uploader.proto import experiment_pb2
 from tensorboard.uploader import logdir_loader
@@ -847,8 +848,14 @@ def _filter_graph_defs(events):
             ):
                 continue
             if v.tag == graphs_metadata.RUN_GRAPH_NAME:
-                data = v.tensor.string_val
-                data[:] = map(_filtered_graph_bytes, data)
+                data = list(v.tensor.string_val)
+                filtered_data = [_filtered_graph_bytes(x) for x in data]
+                filtered_data = [x for x in filtered_data if x is not None]
+                if filtered_data != data:
+                    new_tensor = tensor_util.make_tensor_proto(
+                        filtered_data, dtype=types_pb2.DT_STRING
+                    )
+                    v.tensor.CopyFrom(new_tensor)
         yield e
 
 
@@ -863,7 +870,7 @@ def _filtered_graph_bytes(graph_bytes):
         logger.warning(
             "Could not parse GraphDef of size %d.", len(graph_bytes),
         )
-        return graph_bytes
+        return None
     # Use the default filter parameters:
     # limit_attr_size=1024, large_attrs_key="_too_large_attrs"
     process_graph.prepare_graph_for_ui(graph_def)
--- a/tensorboard/uploader/uploader.py
+++ b/tensorboard/uploader/uploader.py
@@ -25,6 +25,8 @@ import time
 import grpc
 import six
 
+from google.protobuf import message
+from tensorboard.compat.proto import graph_pb2
 from tensorboard.compat.proto import summary_pb2
 from tensorboard.uploader.proto import write_service_pb2
 from tensorboard.uploader.proto import experiment_pb2
@@ -32,9 +34,11 @@ from tensorboard.uploader import logdir_loader
 from tensorboard.uploader import util
 from tensorboard import data_compat
 from tensorboard import dataclass_compat
+from tensorboard.backend import process_graph
 from tensorboard.backend.event_processing import directory_loader
 from tensorboard.backend.event_processing import event_file_loader
 from tensorboard.backend.event_processing import io_wrapper
+from tensorboard.plugins.graph import metadata as graphs_metadata
 from tensorboard.plugins.scalar import metadata as scalar_metadata
 from tensorboard.util import grpc_util
 from tensorboard.util import tb_logging
@@ -425,12 +429,11 @@ class _BatchedRequestSender(object):
         for (run_name, events) in six.iteritems(run_to_events):
             for event in events:
                 v2_event = data_compat.migrate_event(event)
-                dataclass_events = dataclass_compat.migrate_event(
-                    v2_event, experimental_filter_graph=True
-                )
-                for dataclass_event in dataclass_events:
-                    if dataclass_event.summary:
-                        for value in dataclass_event.summary.value:
+                events = dataclass_compat.migrate_event(v2_event)
+                events = _filter_graph_defs(events)
+                for event in events:
+                    if event.summary:
+                        for value in event.summary.value:
                             yield (run_name, event, value)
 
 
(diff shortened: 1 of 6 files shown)


tensorboard/uploader/uploader.py
```diff
@@ -425,12 +429,11 @@ def _run_values(self, run_to_events):
         for (run_name, events) in six.iteritems(run_to_events):
             for event in events:
                 v2_event = data_compat.migrate_event(event)
-                dataclass_events = dataclass_compat.migrate_event(
-                    v2_event, experimental_filter_graph=True
-                )
-                for dataclass_event in dataclass_events:
-                    if dataclass_event.summary:
-                        for value in dataclass_event.summary.value:
+                events = dataclass_compat.migrate_event(v2_event)
```
davidsoergel, line 432: Please don't shadow `events` (and `event` below)-- that is confusing and error-prone.  `dataclass_events` was fine imho.

tensorboard/uploader/uploader.py
```diff
@@ -833,3 +836,35 @@ def _varint_cost(n):
...
+
+def _filter_graph_defs(events):
+    for e in events:
+        for v in e.summary.value:
+            if (
+                v.metadata.plugin_data.plugin_name
+                != graphs_metadata.PLUGIN_NAME
+            ):
+                continue
+            if v.tag == graphs_metadata.RUN_GRAPH_NAME:
```
davidsoergel, line 849: This seems unnecessarily restrictive.  In practice, for now, we have only run-level graphs anyway.  But IIRC we expect to allow `"__op_graph__/foo"` and `"__conceptual_graph__/bar"` also, and those should also be filtered.  NBD for now but maybe add a comment or TODO to address this later.

(I know this is a straight refactor of code that already ignored those cases, but still)

tensorboard/uploader/uploader_test.py
```diff
@@ -355,6 +356,65 @@ def test_upload_skip_large_blob(self):
         self.assertEqual(0, mock_rate_limiter.tick.call_count)
         self.assertEqual(1, mock_blob_rate_limiter.tick.call_count)
 
+    def test_filter_graphs(self):
+        # Three graphs: one short, one long, one corrupt.
+        bytes_0 = _create_example_graph_bytes(123)
+        bytes_1 = _create_example_graph_bytes(9999)
+        bytes_2 = b"\x0a\x7fbogus"  # invalid (truncated) proto
```
davidsoergel, line 363: Please add a comment about what these values mean and why they are necessary for the test.  `"bogus"` alone was sufficient to trigger `DecodeError`.  Is this the case that sometimes produces `RuntimeWarning`?

tensorboard/uploader/uploader.py
```diff
@@ -833,3 +836,35 @@ def _varint_cost(n):
...
+        graph_def = graph_pb2.GraphDef().FromString(graph_bytes)
+    # The reason for the RuntimeWarning catch here is b/27494216, whereby
+    # some proto parsers incorrectly raise that instead of DecodeError
+    # on certain kinds of malformed input. Triggering this seems to require
+    # a combination of mysterious circumstances.
+    except (message.DecodeError, RuntimeWarning):
+        logger.warning(
+            "Could not parse GraphDef of size %d.", len(graph_bytes),
+        )
+        return graph_bytes
```
davidsoergel, line 866: Previously this effectively returned `None`, leading to the event being skipped entirely.  I think that's correct: the philosophy here is that we only bother uploading a graph that we will later be able to render.  Matching that will require adjusting the logic in `_filter_graph_defs` above, because uploading an empty blob is different from dropping the event.

tensorboard/uploader/uploader_test.py
```diff
@@ -355,6 +356,65 @@ def test_upload_skip_large_blob(self):
...
+            actual_blobs.append(b"".join(r.data for r in requests))
+
+        actual_graph_defs = []
+        for blob in actual_blobs:
+            try:
+                actual_graph_defs.append(graph_pb2.GraphDef.FromString(blob))
+            except message.DecodeError:
+                actual_graph_defs.append(None)
+
+        with self.subTest("small graphs should pass through unchanged"):
```
davidsoergel, line 401: "graphs with small attr values..."

tensorboard/uploader/uploader_test.py
```diff
@@ -355,6 +356,65 @@ def test_upload_skip_large_blob(self):
...
+            expected_graph_def_0 = graph_pb2.GraphDef.FromString(bytes_0)
+            self.assertEqual(actual_graph_defs[0], expected_graph_def_0)
+
+        with self.subTest("large graphs should be filtered"):
```
davidsoergel, line 405: "large attr values should be filtered out"

tensorboard/uploader/uploader_test.py
```diff
@@ -355,6 +356,65 @@ def test_upload_skip_large_blob(self):
...
+            expected_graph_def_1 = graph_pb2.GraphDef.FromString(bytes_1)
+            del expected_graph_def_1.node[1].attr["large"]
+            expected_graph_def_1.node[1].attr["_too_large_attrs"].list.s.append(
+                b"large"
+            )
+            requests = list(mock_client.WriteBlob.call_args[0][0])
+            self.assertEqual(actual_graph_defs[1], expected_graph_def_1)
+
+        with self.subTest("corrupt graphs should be passed through unchanged"):
```
davidsoergel, line 414: "...should be skipped", per previous

tensorboard/uploader/uploader_test.py
```diff
@@ -355,6 +356,65 @@ def test_upload_skip_large_blob(self):
         self.assertEqual(0, mock_rate_limiter.tick.call_count)
         self.assertEqual(1, mock_blob_rate_limiter.tick.call_count)
 
+    def test_filter_graphs(self):
```
davidsoergel, line 359: Thanks for cleaning up and simplifying these tests!  However, I am a little skeptical of this design because the tests are no longer independent.  Can you factor out some common setup but keep the tests separate (while retaining most of the simplicity)?

Also, why did you remove the explicit tests for `RuntimeWarning` vs. `DecodeError` that @caisq had just requested in #3508?

Finally FYI #3509 is outstanding.  NP, I'll merge it, assuming you submit first.  The behavior of the empty graph case may actually change, depending on how you address the 'corrupt graph' case above.

tensorboard/uploader/uploader.py
nfelt: Just for what it's worth, all of this goes away in the subsequent PR anyway (https://github.com/tensorflow/tensorboard/pull/3511/files#diff-64ad30888691c0bf32cae63247f4ca5c).

davidsoergel: I remain opposed to any shadowing whatsoever, but don't need to argue this case since it's going away in #3511

(Eep, the former bug does suck, though; thanks for the fix).

tensorboard/uploader/uploader.py
davidsoergel: Oh right, I forgot about the different plugin names.  So this condition is protecting us from data that is associated with the graph plugin, but under a different tag.  There is currently no such thing; if something arises in the future, we can't predict whether it should be filtered or not-- so now I agree the condition is correct.

tensorboard/uploader/uploader_test.py
davidsoergel: That's all good; I'm just saying that you can't expect a reader to recognize and decode the proto wire format on sight.  A comment would solve that:
```
# Declare a proto field of length 127 bytes (`\x7f`), but provide only 5 bytes.
```
Similarly explaining the \x0a would be nice.

tensorboard/uploader/uploader_test.py
davidsoergel: OK for now for expediency.  For the record, my concern is that the setup code is not generic; it includes setup specific to the tests (e.g., the values and order of the bytes examples).  If I wanted to add another subtest--perhaps the one from #3509--I'd have to change the setup code here.  I can see this both ways (i.e., I could also add more tests that use the same examples, without changing the setup).  The fact that currently parts of the setup correspond 1:1 with specific tests makes it look more entangled than it necessarily is.

tensorboard/uploader/uploader.py
```diff
@@ -833,3 +837,41 @@ def _varint_cost(n):
...
+def _filtered_graph_bytes(graph_bytes):
+    try:
+        graph_def = graph_pb2.GraphDef().FromString(graph_bytes)
+    # The reason for the RuntimeWarning catch here is b/27494216, whereby
+    # some proto parsers incorrectly raise that instead of DecodeError
+    # on certain kinds of malformed input. Triggering this seems to require
+    # a combination of mysterious circumstances.
+    except (message.DecodeError, RuntimeWarning):
+        logger.warning(
+            "Could not parse GraphDef of size %d.", len(graph_bytes),
```
davidsoergel, line 871: Reinstate "Skipping" or similar in this warning, to clarify that the failure to parse means we don't upload.

tensorboard/uploader/uploader_test.py
wchargin: Okay, added an explicit comment per your suggestion.

tensorboard/uploader/uploader.py
wchargin: Meant to; thanks. Done.

wchargin:
Test sync passes.

PR #3514, opened 2020-04-14
audio: add generic data support

wchargin:
Summary:
This patch teaches the audio plugin to support generic data providers.

This requires a few more `list_blob_sequences` queries than we’d like,
due to the current structure of the frontend. Some have been avoided by
pushing the content type of the waveform into the query string. A bit of
care is required to ensure that a malicious request cannot turn this
into an XSS hole; we do so by simply whitelisting a (singleton) set of
audio content types, which means that the safety properties are
unchanged by this patch.

Test Plan:
The audio dashboard now works even if the multiplexer is removed. The
audio dashboard doesn’t have any “extras” (data download links, etc.),
so just checking the basic functionality suffices. Unit tests updated.

wchargin-branch: audio-generic


tensorboard/plugins/audio/audio_plugin.py
```diff
@@ -57,18 +66,12 @@ def get_plugin_apps(self):
...
 
     def is_active(self):
-        """The audio plugin is active iff any run has at least one relevant
-        tag."""
-        if not self._multiplexer:
-            return False
-        return bool(
-            self._multiplexer.PluginRunToTagToContent(metadata.PLUGIN_NAME)
-        )
+        return False
```
nfelt, line 69: maybe add the ``# `list_plugins` as called by TB core suffices`` comment here as well?

tensorboard/plugins/audio/audio_plugin.py
```diff
@@ -178,78 +148,73 @@ def _audio_response_for_run(self, tensor_events, run, tag, sample):
...
-        except (KeyError, IndexError):
-            return http_util.Respond(
-                request,
-                "Invalid run, tag, index, or sample",
-                "text/plain",
-                code=400,
+        experiment = plugin_util.experiment_id(request.environ)
+        mime_type = request.args["content_type"]
+        if mime_type not in _ALLOWED_MIME_TYPES:
+            raise errors.InvalidArgumentError(
```
nfelt, line 209: Thanks for catching this possible issue!  Can we add a test that exercises this to ensure that we refuse to return audio as, say, `text/javascript`?

I think the Werkzeug client `get()` API takes a `content_type` parameter so this should be straightforward.

In the longer term it would be nice if we didn't have to rely on the client behavior for this. Perhaps we should add a per-blob `content_type` onto the blob data model?  We would have to go around adding it to the various data providers and being sure to set it correctly on upload, but it seems worth doing to properly handle cases like this.

tensorboard/plugins/audio/audio_plugin.py
wchargin: Done.

tensorboard/plugins/audio/audio_plugin.py
wchargin: Sure: done, I think, though I didn’t understand what you meant about the
Werkzeug client API taking a `content_type` parameter. (Isn’t that for
the content type of a POST request body?)

> Perhaps we should add a per-blob `content_type` onto the blob data
> model?

Perhaps. Presumably the source of truth would be a new field on the
`SummaryMetadata` proto (`string content_type = 5;`), required to be
homogeneous within a time series across all steps and indices, and
defaulting to `application/octet-stream` if empty for backward
compatibility? I’ll keep it in mind.

nfelt: Ah, I was confused and misread this and somehow thought we were getting the content type from a header that the browser was populating (even though yes, that isn't a thing for GET requests) rather than the query parameter.  Never mind :P 

Re: per-blob content types, yeah, I guess it could have a field in SummaryMetadata, although it's perhaps a little odd to have a field that is returned per-value stored in the per-tag SummaryMetadata (even if for now we don't anticipate allowing it to vary per-value).

Issue #3516, opened 2020-04-14
Eliminate transitive JS dependencies on old versions of `minimist`


nfelt:
Old versions of the `miminist` CLI parsing JS library have a security hole that allows polluting the Object prototype: https://www.npmjs.com/advisories/1179  The fix is to update to version 1.2.3 or later (0.2.1 is also patched for 0.x but it would be best to just update to 1.2.3+).

The only way that we depend on old versions of `minimist` is via a wrapper library `optimist` that was abandoned at version 0.6.1 back in 2014, so realistically what this means is upgrading our dependencies to avoid a dependency on `optimist` at all.

We have two dependency paths that lead via `optimist` to the old `minimist`:

![minimist_deps](https://user-images.githubusercontent.com/710113/79187997-b61a4300-7dd2-11ea-8d63-43addd42a6d4.png)

Note that since the direct dependencies involved are both JS test dependencies, I don't think there is any actual security implication to the old `minimist` since the vulnerability involves being able to control the CLI arguments passed to the parser, and so at best this could be controlled during test execution time to "attack" the javascript being run within a test.  That said, GitHub is still reporting the issue, and it would be good to clean this up just so it stops cropping up.

The handlebars dependency path is easy to fix since a compatible recent release no longer has the `optimist` dep, so we can just update it: https://github.com/handlebars-lang/handlebars.js/issues/1658

The karma dependency path is more annoying.  The recently released karma 5.0.0 package removes the bad dep thanks to https://github.com/karma-runner/karma/pull/3451.  However, our dependency on `@bazel/karma@^0.36.1` depends on `karma@^4.0.0` meaning we need to update rules_nodejs (which provides `@bazel/karma`) in order to safely update karma.

Updating rules_nodejs has proven to be a pain.  It would look like 0.41.0 would suffice since it includes https://github.com/bazelbuild/rules_nodejs/pull/1346 which relaxes the karma dep to just `>=4.0.0`, but when I tried that I hit various issues getting a combination of deps that would actually work, including a version of `@angular/bazel` (which we have at `^8.2.14`) that played nicely with the respective version of `rules_nodejs` - instead I hit some variation of https://github.com/bazelbuild/rules_nodejs/issues/1337.  Fixing that requires an update to `@angular/bazel` that in turns requires a more recent `rules_nodejs`, and then we start hitting other incompatible changes (for example https://github.com/bazelbuild/rules_nodejs/wiki#migrating-off-internalrollup among others).

[...]

PR #3517, opened 2020-04-14
Upgrade handlebars 4.7.3 -> 4.7.6 to avoid optimist dep

nfelt:
This updates our yarn.lock file to bump our `handlebars` dep from 4.7.3 to 4.7.6 (the latest available release), which I did just by deleting the old entry and rerunning yarn install.  This addresses the easy part of #3516 by picking up https://github.com/wycats/handlebars.js/issues/1658 to prune a transitive dependency via `optimist` on an old version of `minimist`.


nfelt:
> Note that the minor version of Handlebars 4.6 and 4.7 doesn't have a matching minor equivalent internally. I'm not sure how strict we need to be with having the exact same version internally vs. external, but lgtm

Good point, but we're already pretty skewed (4.0.14 internally) so I figure a small amount more here is unlikely to cause issues.

PR #3519, opened 2020-04-14
uploader_test: check logical equality of protos

wchargin:
Summary:
Fixes an oversight in #3507: we can’t assert that the raw bytes are what
was expected because the code under test does a proto serialization
roundtrip, which is permitted to permute keys.

Test Plan:
Tests still pass; a test sync shows that internal tests pass now, too.

wchargin-branch: uploader-test-proto-equal

### DIFF
--- a/tensorboard/uploader/uploader_test.py
+++ b/tensorboard/uploader/uploader_test.py
@@ -315,7 +315,7 @@ class TensorboardUploaderTest(tf.test.TestCase):
             requests = list(call[0][0])
             data = b"".join(r.data for r in requests)
             actual_graph_def = graph_pb2.GraphDef.FromString(data)
-            self.assertEqual(actual_graph_def, expected_graph_def)
+            self.assertProtoEquals(expected_graph_def, actual_graph_def)
             self.assertEqual(
                 set(r.blob_sequence_id for r in requests), {"blob%d" % i},
             )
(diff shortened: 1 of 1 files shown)


wchargin:
> Tests still pass, and this should fix an internal sync error.

Confirmed: <http://cl/306474605> (passes on 100/100 runs)

tensorboard/uploader/uploader_test.py
```diff
@@ -311,7 +314,8 @@ def test_start_uploading_graphs(self):
         for (i, call) in enumerate(mock_client.WriteBlob.call_args_list):
             requests = list(call[0][0])
             data = b"".join(r.data for r in requests)
-            self.assertEqual(data, graph_event.graph_def)
+            actual_graph_def = graph_pb2.GraphDef.FromString(data)
+            self.assertEqual(actual_graph_def, expected_graph_def)
```
nfelt, line 318: might as well use `assertProtoEquals()` for the nicer diff?

wchargin: Sure; done.

wchargin: I can never remember the `assertProtoEquals` argument order; thankfully,
a web search turns up familiar faces…

![Screenshot of DuckDuckGo search for `assertprotoequals`: the top
result is a TensorBoard commit!][ss]

[ss]: https://user-images.githubusercontent.com/4317806/79261133-3d53cf00-7e44-11ea-930e-71e81956f3b6.png

PR #3524, opened 2020-04-16
improve rerender of vz_line_chart

stephanwlee:
We identified few several expensive operations going on underneath Plottable.
- When we modified dataset() attached to a plot, Plottable recalculated
  domain which in turn triggered redraw of the plot.
- Plot redraw is scheduled but repeated `clearTimeout` and `setTimeout`
  take considerable time which was the large part of the dataset update
- Changing a dataset caused us to updateSpecialDataset N times (N =
  number of runs)

We addressed the issue by:
- `commit` API after making all the changes
- remember which runs/dataset were changed
- update datasets only once per commit

Empirically, the render time:
- toggling run on the selector went down from 1740ms to 314ms
- triggering the log scale went down from 315ms to 264ms
  - from ~670ms to ~620ms when measured from mouse up

Do note that this change does not improve time for other charting
operation like zoom, smoothing changes, and etc...

Test plan: rendered the temperature demo and manually clicked around UI elements and compared the result against TensorBoard 2.1.1.

### DIFF
--- a/tensorboard/components/tf_line_chart_data_loader/tf-line-chart-data-loader.html
+++ b/tensorboard/components/tf_line_chart_data_loader/tf-line-chart-data-loader.html
@@ -219,6 +219,10 @@ limitations under the License.
           this.$.chart.setSeriesMetadata(name, metadata);
         },
 
+        commitChanges() {
+          this.$.chart.commitChanges();
+        },
+
         redraw() {
           cancelAnimationFrame(this._redrawRaf);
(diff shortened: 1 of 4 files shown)


wchargin:
Very nice. On the hparams demo’s `batch_accuracy` tag (50 time series),
I’m seeing a toggle-run improvement of 4.5 seconds → ~0.3 seconds. The
new version is fast enough that it’s hard for me to time it accurately
with my stopwatch.

PR #3526, opened 2020-04-17
[DebuggerV2] Use more efficient read in /execution/data and /graph_execution/data routes

caisq:
  * Fix a performance issue in the HTTP route /graph_execution/data
  * The issue has to do with the fact that `DebugDataMultiplexer.GraphExecutionData()` uses the inefficient approach of reading all the `GraphExecutionTrace`s first and then slice the array. A much more efficient approach (analogous to the `DebugDataMultiplexer.ExecutionData()` for top-level executions) is to use the range reading support of `DebugDatataReader.graph_execution_traces()` that is now available.
  * Similarly, use the range reading support of `DebugDatataReader.execution()` in the `/execution/data` route.
  * See the description above. Use range reading support of `DebugDatataReader.execution()` and `DebugDatataReader.graph_execution_trace()`.
  * Also, remove an extraneous element in the json response (`num_digests`). It is not in the design spec and was included due to oversight on my part.
  * No new unit tests are required to verify the correctness of the new implementation. The existing ones suffice. 
  * Manual testing in the web GUI to confirm much more efficient loading.


caisq:
@wchargin Thanks for the review and comment. 

I've updated this PR. The underlying methods of `DebugDataReader` now support range reading with the `begin` and `end` kwargs in its `execution()` and `graph_execution_trace()` methods. I'm using them in this PR now. This avoids the N + 1 request problem and makes the implementation consistent between (eager) execution and intra-graph execution.

PTAL.

PR #3529, opened 2020-04-17
text: batch HTML sanitization

wchargin:
Summary:
By design, we only expose an API for converting Markdown to *safe* HTML.
But clients who call this method many times end up performing HTML
sanitization many times, which is expensive: about an order of magnitude
more expensive than Markdown conversion itself. This patch introduces a
new API that still only emits safe HTML but enables clients to combine
multiple input documents with only one round of sanitization.

Test Plan:
The `/data/plugin/text/text` route sees 40–60% speedup: on my machine,

  - from 0.38 ± 0.04 seconds to 0.211 ± 0.005 seconds on the “higher
    order tensors” text demo downsampled to 10 steps; and
  - from 5.3 ± 0.9 seconds to 2.1 ± 0.2 seconds on a Google-internal
    dataset with 32 Markdown cells per step downsampled to 100 steps.

wchargin-branch: text-batch-bleach

### DIFF
--- a/tensorboard/plugin_util.py
+++ b/tensorboard/plugin_util.py
@@ -113,7 +113,6 @@ def markdowns_to_safe_html(markdown_strings, combine):
     """
     unsafe_htmls = []
     total_null_bytes = 0
-    import contextlib
 
     for source in markdown_strings:
         # Convert to utf-8 whenever we have a binary input.
--- a/tensorboard/plugin_util.py
+++ b/tensorboard/plugin_util.py
@@ -85,25 +85,60 @@ def markdown_to_safe_html(markdown_string):
     Returns:
       A string containing safe HTML.
     """
-    warning = ""
-    # Convert to utf-8 whenever we have a binary input.
-    if isinstance(markdown_string, six.binary_type):
-        markdown_string_decoded = markdown_string.decode("utf-8")
-        # Remove null bytes and warn if there were any, since it probably means
-        # we were given a bad encoding.
-        markdown_string = markdown_string_decoded.replace(u"\x00", u"")
-        num_null_bytes = len(markdown_string_decoded) - len(markdown_string)
-        if num_null_bytes:
-            warning = (
-                "<!-- WARNING: discarded %d null bytes in markdown string "
-                "after UTF-8 decoding -->\n"
-            ) % num_null_bytes
+    return markdowns_to_safe_html([markdown_string], lambda xs: xs[0])
 
-    string_html = _MARKDOWN_STORE.markdown.convert(markdown_string)
-    string_sanitized = bleach.clean(
-        string_html, tags=_ALLOWED_TAGS, attributes=_ALLOWED_ATTRIBUTES
+
+def markdowns_to_safe_html(markdown_strings, combine):
+    """Convert multiple Markdown documents to one safe HTML document.
+
+    One could also achieve this by calling `markdown_to_safe_html`
+    multiple times and combining the results. Compared to that approach,
+    this function may be faster, because HTML sanitization (which can be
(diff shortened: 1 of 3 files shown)


tensorboard/plugin_util.py
```diff
@@ -85,25 +85,60 @@ def markdown_to_safe_html(markdown_string):
...
+def markdowns_to_safe_html(markdown_strings, combine):
+    """Convert multiple Markdown documents to one safe HTML document.
+
+    One could also achieve this by calling `markdown_to_safe_html`
+    multiple times and combining the results. Compared to that approach,
+    this function may be faster, because HTML sanitization (which can be
+    expensive) is performed only once rather than once per input. It may
+    also be less precise: if one of the input documents has unsafe HTML
+    that is sanitized away, that sanitization might affect other
+    documents, even if those documents are safe.
```
wchargin, line 100: I can’t actually find a way to trigger this. What I mean is that if you
could find input strings `x`, `y`, and `z` such that

  - `markdown.markdown(x) == '<a target="*chuckles* '`
  - `markdown.markdown(y) == "I'm in danger"`
  - `markdown.markdown(z) == '"></a>'`

then `markdowns_to_safe_html([x, y, z], "".join)` would lose the content
of `y` even though it was itself safe and didn’t “deserve” to get
bleached. But I can’t find any strings `x` that would create open tags
like that (though admittedly I didn’t try very hard).

If `x == "<div>hmm"` then `markdown.markdown(x) == x`, but the same is
not true if `x == "<a>hmm`, and `<div>` is not on our Bleach whitelist.

Given that the output is *always* safe and is unreasonably behaved (but
still safe) only when the input is unsafe, this seems reasonable to me.

nfelt: Happy to oblige.

```
---
 tensorboard/plugin_util_test.py | 7 +++++++
 1 file changed, 7 insertions(+)

diff --git a/tensorboard/plugin_util_test.py b/tensorboard/plugin_util_test.py
index c0fb6798..e8658772 100644
--- a/tensorboard/plugin_util_test.py
+++ b/tensorboard/plugin_util_test.py
@@ -157,6 +157,13 @@ class MarkdownsToSafeHTMLTest(tb_test.TestCase):
         expected = "&lt;script&gt;alert('unsafe!')&lt;/script&gt;<p>safe</p>"
         self.assertEqual(actual, expected)

+    def test_sanitization_can_have_collateral_damage(self):
+        inputs = ["<table title=\"chuckles", "I'm in danger", "<table>\">"]
+        combine = lambda xs: "".join(xs)
+        actual = plugin_util.markdowns_to_safe_html(inputs, combine)
+        expected = "<table></table>"
+        self.assertEqual(actual, expected)
+

 class ExperimentIdTest(tb_test.TestCase):
     """Tests for `plugin_util.experiment_id`."""
--
```

.. but yes, I agree this isn't a concern in practice.

tensorboard/plugin_util.py
```diff
@@ -85,25 +85,60 @@ def markdown_to_safe_html(markdown_string):
...
+        strings of the same shape as `markdown_strings` and combines
+        them into a single unsafe HTML string, which will be sanitized
+        and returned.
+
+    Returns:
+      A string containing safe HTML.
+    """
+    unsafe_htmls = []
+    total_null_bytes = 0
+    import contextlib
```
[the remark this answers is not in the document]
wchargin: Oops, yes, thanks. (There was a `nullcontext` in here for testing…)

tensorboard/plugin_util.py
wchargin: Wow, perfect. Thank you! Added test.

PR #3530, opened 2020-04-19
[DebuggerV2] Start DebugDataMultiplexer reading at plugin creation

caisq:
  * Fix b/154132088: if a debugger run exists in the logdir, `DebugDataMultiplexer` should start loading the data as soon as the DebuggerV2 plugin backend is created. This will shorten the time a user of the DebuggerV2 needs to wait until the data is fully loaded after opening the plugin in the frontend. The current behavior is that run and data loading starts only when the first request is make to the `/runs` route.
  * This should not be a concern for the performance of other TensorBoard plugins, because the tfdbg2 data (tfdbg_events.* files) are created *only* if the generating TensorFlow program calls the `tf.debugging.experimental.enable_dump_debug_info()` explicitly.
  * `DebugDataMultiplexer` creates a `DebugDataReader` in its constructor, instead of when handling the first `Runs()` call.
  * After the `DebugDataReader` is created, start updating the reader in a periodic fashion in a background thread. The sleep interval between updates is 30 seconds.
  * The object used for `sleep()` between `_reader.update()` calls is an interruptable (wakeable) custom class called `Timer` in debug_data_multiplexer.py. The sleeps will be interrupted (waken) on when data-access methods of `DebugDataMultiplexer` is called.
  * New unit tests are added for the new behavior added in this PR.
  * Manual verification by running DebuggerV2 plugin in ng_tensorboard with real logdirs with tfdbg2 data.


tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -36,9 +37,41 @@
 # ones are either repetitions of the earlier ones or caused by the earlier ones.
 DEFAULT_PER_TYPE_ALERT_LIMIT = 1000
 
+# Default interval between successive calls to `DebugDataReader.update()``.
+DEFAULT_RELOAD_INTERVAL_SEC = 60
 
-def run_in_background(target):
-    """Run a target task in the background.
+
+class Timer(object):
```
nfelt, line 44: This is a very thin abstraction over `threading.Event` and IMHO I think it'd be simpler to just use the underlying APIs directly (it's also slightly confusing as named, since `threading.Timer` exists but is a different thing entirely).

All this abstraction really provides is tying together the calls to `clear()` and `wait()`, but `sleep()` is only called in a single place so inlining the two calls would be easy.  See also comments below.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -36,9 +37,41 @@
...
+
+    def wake(self):
+        """Interrupt all sleep() calls that are ongoing.
+
+        This is a no-op if no sleep() is ongoing.
+        """
+        self._event.set()
+
+
+_timer = Timer(DEFAULT_RELOAD_INTERVAL_SEC)
```
nfelt, line 70: Do we really want a global Timer instance here?  In general we should only have global objects when we really need them, and looking at the code that uses it, it seems like more logically there should be one for each instance of a DebugDataMultiplexer (e.g. as a property).

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -36,9 +37,41 @@
...
+# Default interval between successive calls to `DebugDataReader.update()``.
+DEFAULT_RELOAD_INTERVAL_SEC = 60
 
-def run_in_background(target):
-    """Run a target task in the background.
+
+class Timer(object):
+    """A timer that supports interruptible sleep()."""
+
+    def __init__(self, interval_sec):
```
nfelt, line 47: It seems kind of arbitrary to take the interval as a constructor parameter when it would be simpler and more flexible to just pass it into `sleep()` - this would matter I suppose if we had many different calls to `sleep()` but this is in fact called in just a single place anyway, so we could reduce the length through which this constant gets plumbed by just passing it directly there.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -48,10 +81,14 @@ def run_in_background(target):
...
-    # TODO(cais): Implement repetition with sleeping periods in between.
     # TODO(cais): Add more unit tests in debug_data_multiplexer_test.py when the
     # the behavior gets more complex.
-    thread = threading.Thread(target=target)
+    def _run_repeatedly():
+        while True:
+            target()
+            _timer.sleep()
+
+    thread = threading.Thread(target=_run_repeatedly)
```
nfelt, line 91: Existing code, but shouldn't this be a daemon thread?  https://docs.python.org/3.6/library/threading.html#threading.Thread.daemon

Otherwise if this thread isn't terminated properly at program exit it could prevent TensorBoard from exiting.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -36,9 +37,41 @@
...
+
+        This is a no-op if no sleep() is ongoing.
+        """
+        self._event.set()
+
+
+_timer = Timer(DEFAULT_RELOAD_INTERVAL_SEC)
+
+
+def run_repeatedly_in_background(target):
```
nfelt, line 73: FWIW, I think a somewhat nicer API for this function would be:

- take the interval at which to repeatedly call the target as a parameter to pass into `sleep()`, rather than making it a hidden global config that this function happens to use
- create the `threading.Event` as a local variable within this function and have `_run_repeatedly()` close over it
- return the `Event` as the return value to the caller, to be used by the caller to signal when `target` should run immediately

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -98,6 +135,57 @@ def __init__(self, logdir):
         """
         self._logdir = logdir
         self._reader = None
+        self._reader_lock = threading.Lock()
+        self._createOrReloadReader()
+
+    def _createOrReloadReader(self):
```
nfelt, line 141: It'd be helpful to break this method down since it's doing several different things and that makes it hard to reliably follow the details of the threading-related parts.  In particular it doesn't seem like it's necessary to obtain the reader lock if the reader already exists, because then you can just directly call `wake()/set()` to trigger an update.

FWIW, I'm also not entirely sure why it makes sense to trigger this logic from `Runs()` as well as from `__init__()`.  I guess because there's a chance the reader was not created due to a transient issue (the "no DebugEvent file set is found in the logdir") so we want to retry periodically?  But since this logic tightly couples both the reader creation and the immediate-update-triggering logic, it's not clear to me if `Runs()` also needs to trigger an immediate update of the reader (since the logic in `Runs()` doesn't actually use the reader directly, and it would be an asynchronous update anyway), or if that just sort of happens as an unnecessary side effect.

More comments about the expected lifecycle would help here, as would breaking this into more granular pieces that are explicitly called elsewhere in the class when needed.  For example:

```
def __init__(self):
    ...
    self._reader = None
    self._reader_lock = threading.Lock()
    self._monitors = []
    self._reload_needed_event = None
    self._tryCreateReader()
    self._tryReloadReader()

def _tryCreateReader(self):
    # Attempt to create the reader, using double-checked locking.
    if not self._reader:
        with self._reader_lock:
            if not self._reader:
                self._reader = self._createReader()
    # Reader creation may fail due to transient errors; abort and try later.
    if not self._reader:
        return
    self._monitors = self._createMonitors(self._reader)
    self._reload_needed_event = run_repeatedly_in_background(
        self._reader.update, DEFAULT_RELOAD_INTERVAL_SEC)

def _tryReloadReader(self)
    if self._reader:
        self._reload_needed_event.set()

def _createReader(self):
    ...

def _createMonitors(self, reader):
    ...
```

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -98,6 +135,57 @@ def __init__(self, logdir):
...
+        instead, the existing reader's `update()` method will be unblocked on a
+        separate thread immediately.
+
+        If a reader has not been created, create it and start periodic calls to
+        `update()` on a separate thread.
+        """
+        with self._reader_lock:
+            if not self._reader:
+                try:
+                    from tensorflow.python.debug.lib import debug_events_reader
```
nfelt, line 154: FWIW, I realize this is existing code, but would be nice if we can avoid doing conditional imports like this and instead use the plugin loader functionality to gate the loading of the entire plugin so that if we can't import the necessary deps we don't end up failing later on like this.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -98,6 +135,57 @@ def __init__(self, logdir):
...
+                    )
+                    self._monitors = [
+                        debug_events_monitors.InfNanMonitor(
+                            self._reader, limit=DEFAULT_PER_TYPE_ALERT_LIMIT
+                        )
+                    ]
+                    # NOTE(cais): Currently each logdir is enforced to have only one
+                    # DebugEvent file set. So we add hard-coded default run name.
+                    run_repeatedly_in_background(self._reader.update)
+                    # TODO(cais): Start off a reading thread here, instead of being
```
nfelt, line 170: This looks like it was handled?

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -98,6 +135,57 @@ def __init__(self, logdir):
...
+                    # called only once here.
+                except ImportError:
```
nfelt, line 172: I realize this is existing code, but especially when swallowing exceptions,  it's best practice to scope the `except` blocks at tightly as possible to avoid catching exceptions that we did not intend to catch.  E.g. catching a blanket `AttributeError` or `ValueError` could easily catch all kinds of exceptions other than the ones intended.

I'd strongly recommend scoping the errors to the minimum statement required (e.g. ImportError and AttributeError to _only_ the import statements, ValueError to _only_ the `DebugDataReader` construction) and would also recommend changing `DebugDataReader` so it raises a custom exception rather than a generic ValueError for a case like this where it's expected behavior that calling code must handle.

tensorboard/plugins/debugger_v2/debug_data_multiplexer_test.py
```diff
@@ -34,11 +37,45 @@ def start(self):
 
 
 class DebuggerV2PluginTest(tf.test.TestCase):
-    def testRunInBackground(self):
-        mock_target = tf.compat.v1.test.mock.Mock()
-        with tf.compat.v1.test.mock.patch("threading.Thread", MockThread):
-            debug_data_multiplexer.run_in_background(mock_target)
-            mock_target.assert_called_once()
+    def testRunInBackgroundRepeatedly(self):
```
nfelt, line 40: nit, but would be better to have the test method name match the actual function (so testRunRepeatedlyInBackground)

Also, existing code, but a bit odd that this test class is named DebuggerV2PluginTest when this is in `debug_data_multiplexer_test.py`.

tensorboard/plugins/debugger_v2/debug_data_multiplexer_test.py
```diff
@@ -34,11 +37,45 @@ def start(self):
...
+
+        with mock.patch("threading.Thread", MockThread), mock.patch.object(
+            debug_data_multiplexer._timer, "sleep", mock_sleep
+        ):
+            with self.assertRaises(StopIteration):
+                debug_data_multiplexer.run_repeatedly_in_background(mock_target)
+            self.assertEqual(mock_target.call_count, 10)
+
+
+class TimerTest(tf.test.TestCase):
```
nfelt, line 57: FWIW, see other comment, but this is basically just testing the `threading.Event` API so I'm not sure how much value it provides.

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -124,11 +125,13 @@ def setUp(self):
         # unit test, we disable the asynchronous behavior, so that we can
         # load the debugger data synchronously on the main thread and get
         # determinisic behavior in the tests.
-        def run_in_background_mock(target):
+        def run_repeatedly_in_background_mock(target):
```
nfelt, line 128: This doesn't seem like a very accurate mock (or really a "fake" in testing terms) for the new code, since the real code now calls `target` more than once, but this only calls it a single time?

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -151,6 +154,33 @@ def testPluginIsActiveWithDataExists(self):
         _generate_tfdbg_v2_data(self.logdir)
         self.assertTrue(self.plugin.is_active())
 
+    def testConcurrentCallsToPluginIsActiveWhenNotActive(self):
+        results = []
+
+        def query_is_active():
+            results.append(self.plugin.is_active())
```
nfelt, line 161: It's best not to reply on the built in python data structures to be thread safe (even if some of their operations happen to be for cpython).  For this we could just use collections.deque which is documented to be thread safe: https://docs.python.org/2/library/collections.html#collections.deque

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -172,6 +202,51 @@ def testServeRunsWithExistingRuns(self):
...
+        for thread in threads:
+            thread.join()
+
+        self.assertLen(responses, 4)
+        for response in responses:
+            self.assertEqual(200, response.status_code)
+            self.assertEqual(
+                "application/json", response.headers.get("content-type")
+            )
+            self.assertEqual(json.loads(response.get_data()), dict())
```
nfelt, line 223: Could just do `self.assertEmpty`

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: OK. Done removing this Timer class and its associated unit tests. Thanks for the suggestion.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: The revised `run_repeatedly_in_background()` now takes an argument for the sleeping interval (`interval_sec`).

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Good point. Removed this module-level object and added a `_reload_needed_event` member variable to the `DebugDataMultiplexer` class as suggested.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Your suggestion here makes sense. Adopted. Unit tests are updated accordingly. Thanks.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Done.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Done breaking out these new methods:
- `_tryCreateReader()`
- `_reloadReader()`

I didn't break out additional new methods. The `self._monitor` is closedly related to the `self._reader`. So it's clearer to create the two in the same method, which is not overly long right now.

Re. the question about why calling `_tryCreateReader()` in the constructor and in `Runs()` I added comments in both places to clarify that. Briefly, the method is called in the constructor so the backend can start loading data as soon as the plugin backend is created, without waiting for the first request to come in from the FE. It's called in `Runs()` to cover the case in which the tfdbg2 data in the logdir is created after the tensorboard backend starts. I also added a few other comments to clarify the lifecycle of the reader.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: I'd like to defer that to a later PR, ideally after tf 2.2 is released so we can remove one of the `except` branches below. Added a TODO item here.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Removed this obsolete TODO item.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Done tightening the try-except blocks in this method.

tensorboard/plugins/debugger_v2/debug_data_multiplexer_test.py
caisq: Renamed the test class and test method as a part of rewriting these tests for the new signature of `run_repeatedly_in_background()`.

tensorboard/plugins/debugger_v2/debug_data_multiplexer_test.py
caisq: These tests are removed together with the `Timer` class.

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
caisq: As the comment above this line mentions, the purpose of this mock/fake is simply to avoid multi-threaded behavior in this test and ensure determinism in test results. The multi-threaded aspect of the `run_repeatedly_in_background()` is covered separately in debug_data_multiplexer_test.py.

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
caisq: Done here an in other similar places in this file.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -80,16 +51,24 @@ def run_repeatedly_in_background(target):
...
     """
-    # TODO(cais): Add more unit tests in debug_data_multiplexer_test.py when the
-    # the behavior gets more complex.
+    event = threading.Event()
+
     def _run_repeatedly():
         while True:
             target()
-            _timer.sleep()
+            event.wait(interval_sec)
```
nfelt, line 66: Don't we still need `event.clear()` here in addition to `wait()`?

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -136,56 +115,73 @@ def __init__(self, logdir):
...
-        self._createOrReloadReader()
+        self._reload_needed_event = None
+        # Create the reader for the tfdbg2 data in the lodir as soon as
+        # the backend of the debugger-v2 plugin is created, so it doesn't need
+        # to wait for the first request from the FE to start loading data.
+        self._tryCreateLoader()
 
-    def _createOrReloadReader(self):
+    def _tryCreateLoader(self):
         """Creates or reloads reader for tfdbg2 data in the logdir.
```
nfelt, line 125: "reloads" part is stale

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -136,56 +115,73 @@ def __init__(self, logdir):
...
-            else:
-                _timer.wake()
+                    return
+
+                self._monitors = [
+                    debug_events_monitors.InfNanMonitor(
+                        self._reader, limit=DEFAULT_PER_TYPE_ALERT_LIMIT
+                    )
+                ]
+                # NOTE(cais): Currently each logdir is enforced to have only one
```
nfelt, line 170: I'm not sure I understand this comment - where is the hard-coded default run name?

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -136,56 +115,73 @@ def __init__(self, logdir):
...
         self._reader_lock = threading.Lock()
-        self._createOrReloadReader()
+        self._reload_needed_event = None
+        # Create the reader for the tfdbg2 data in the lodir as soon as
+        # the backend of the debugger-v2 plugin is created, so it doesn't need
+        # to wait for the first request from the FE to start loading data.
+        self._tryCreateLoader()
 
-    def _createOrReloadReader(self):
+    def _tryCreateLoader(self):
```
nfelt, line 124: It seems clearer and more consistent to call this `_tryCreateReader`?  That's the term used in other places, and in the implementation, since it creates a `DebugDataReader` stored in `self._reader`.

tensorboard/plugins/debugger_v2/debug_data_multiplexer_test.py
```diff
@@ -29,53 +30,55 @@
...
-        self.assertGreaterEqual(time.time() - t0, 0.1)
+                debug_data_multiplexer.run_repeatedly_in_background(
+                    mock_target, 0.2
+                )
+            # Check calls to the target.
+            self.assertEqual(mock_target.call_count, 3)
+            # Check calls to mocked Event.set().
+            self.assertEqual(mock_event.timeouts, [0.2, 0.2, 0.2])
+
+    def testRunIsStoppedByReturnedEventSetMethodCall(self):
```
nfelt, line 76: I don't see an actual assertion in this test - how would it fail?  Since `run_repeatedly_in_background` just starts a background thread I think this test doesn't really inspect anything about what happens in that thread, so I'm not sure what it is really testing.

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
nfelt: Fair enough, ok.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Done.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: This comment is out of place. It's also redundant with the doc string of `FirstEventTimestamp` and `DEFAULT_DEBUGGER_RUN_NAME` in this same file. Removed it.

tensorboard/plugins/debugger_v2/debug_data_multiplexer_test.py
caisq: Yeah this test was incorrect. It was able to terminate without any error only because the thread has `daemon=True`. I updated this test to make it effective and clearer. This also removed two "mock" classes from this file and made it more concise.

PR #3532, opened 2020-04-20
Refactor tests for _ScalarBatchedRequestSender

bmd3k:

Identify tests in uploader_test.py that specifically target logic in _ScalarBatchedRequestSender and rewrite them to only use _ScalarBatchedRequestSender and none of the other layers in uploader.py.

This is a useful exercise for the work on _TensorBatchedRequestSender.  I was having a difficult time using _ScalarBatchedRequestSender tests to establish a pattern for testing _TensorBatchedRequestSender.

### DIFF
--- a/tensorboard/uploader/uploader_test.py
+++ b/tensorboard/uploader/uploader_test.py
@@ -165,6 +165,18 @@ def _create_request_sender(
     )
 
 
+def _create_scalar_request_sender(
+    experiment_id=None, api=None,
+):
+    if api is _USE_DEFAULT:
+        api = _create_mock_client()
+    return uploader_lib._ScalarBatchedRequestSender(
+        experiment_id=experiment_id,
+        api=api,
+        rpc_rate_limiter=util.RateLimiter(0),
+    )
+
+
 class TensorboardUploaderTest(tf.test.TestCase):
     def test_create_experiment(self):
         logdir = "/logs/foo"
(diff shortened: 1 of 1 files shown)


tensorboard/uploader/uploader_test.py
```diff
@@ -564,42 +576,6 @@ def test_upload_swallows_rpc_failure(self):
...
-            writer.add_test_summary("foo")
-        mock_client = _create_mock_client()
-        uploader = _create_uploader(mock_client, logdir)
-        uploader.create_experiment()
-        error = test_util.grpc_error(grpc.StatusCode.NOT_FOUND, "nope")
-        mock_client.WriteScalar.side_effect = error
-        with self.assertRaises(uploader_lib.ExperimentNotFoundError):
-            uploader._upload_once()
-
-    def test_upload_preserves_wall_time(self):
```
bmd3k, line 579: I decided to just plain delete this test. I think it's already covered by test_wall_time_precision().

tensorboard/uploader/uploader_test.py
```diff
@@ -838,9 +779,11 @@ def test_remembers_first_metadata_in_scalar_time_series(self):
         tag_counts = {tag.name: len(tag.points) for tag in run_proto.tags}
         self.assertEqual(tag_counts, {"loss": 2})
 
-    def test_v1_summary_single_value(self):
+    def test_expands_multiple_values_in_event(self):
```
bmd3k, line 782: This is one place where the diff gets confusing.

* test_v1_summary_single_value() has been migrated to ScalarBatchedRequestSenderTest as test_v1_summary().
* test_v1_summary_multiple_value() is kept here but renamed to test_expands_multiple_values_in_event(), which I think is its real value.

tensorboard/uploader/uploader_test.py
bileschi: +1

PR #3533, opened 2020-04-21
ng: expose core initialState as public

stephanwlee:
Internal application would like to use the intialState for other
purposes. Expose it as a public API.

### DIFF
--- a/tensorboard/webapp/core/store/core_initial_state_provider.ts
+++ b/tensorboard/webapp/core/store/core_initial_state_provider.ts
@@ -18,6 +18,7 @@ import {StoreConfig} from '@ngrx/store';
 import {CoreState} from './core_types';
 import {HashDeepLinker} from '../../deeplink';
 import {DataLoadState} from '../../types/data';
+import {initialState} from './core_types';
 
 /** @typehack */ import * as _typeHackStore from '@ngrx/store';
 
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/core/store/core_types.ts
```diff
@@ -35,3 +35,15 @@ export interface CoreState {
...
+export const initialState: CoreState = {
+  activePlugin: null,
+  plugins: {},
+  pluginsListLoaded: {
+    state: DataLoadState.NOT_LOADED,
+    lastLoadedTimeInMs: null,
+  },
+  reloadPeriodInMs: 30000,
+  reloadEnabled: false,
+  pageSize: 15,
```
caisq, line 48: This is partly pre-existing, but I'd like to use this opportunity to comment. Compared to the other fields of this object, `pageSize` is especially unclear in meaning. I think it's better to add a comment to where the field is defined (L32 above) to describe what it is.

tensorboard/webapp/core/store/core_initial_state_provider.ts
```diff
@@ -28,15 +29,8 @@ export const CORE_STORE_CONFIG_TOKEN = new InjectionToken<
 export function getConfig(deepLinker: HashDeepLinker): StoreConfig<CoreState> {
   return {
     initialState: {
+      ...initialState,
```
caisq, line 32: Is there a unit test covering getConfig()? There needs to be one that asserts `activePlugin` returned by the function reflects the return value of `deepLiner.getPluginId()` it is non-null. If there isn't a test like that, please add it.

tensorboard/webapp/core/store/core_types.ts
```diff
@@ -29,9 +29,22 @@ export interface CoreState {
   pluginsListLoaded: LoadState;
   reloadPeriodInMs: number;
   reloadEnabled: boolean;
+  // Size of a page in a paginated view.
```
caisq, line 32: nit: A paginated view of what?

stephanwlee: It is not defined concretely on purpose. It can be used for many UI elements.

caisq: Understood. Just state the fact that it's meant for general UI element use in the comment explicitly. :)

PR #3537, opened 2020-04-22
Refactor logic common to send Tensors and Scalars

bmd3k:

The future _TensorBatchedRequestSender behaves similarly to the existing _ScalarBatchedRequestSender. This change refactors some of the less-trivial logic out from _ScalarBatchedRequestSender to be reused in _TensorBatchedRequestSender.

Refactor logic for managing request byte budget into its own class and for pruning empty runs  and tags into its own function.

One test is improved to use mocked _ByteBudgetManager.


tensorboard/uploader/uploader_test.py
```diff
@@ -1039,6 +1039,11 @@ def test_break_at_scalar_point_boundary(self):
 
         self.assertGreater(len(requests), 1)
         self.assertLess(len(requests), point_count)
+        # This is the observed number of requests when running the test. There
+        # is no reasonable way to derive this value from just reading the code.
+        # The number of requests does not have to be 33 to be correct but if it
+        # changes it probably warrants some investigation or thought.
+        self.assertEqual(33, len(requests))
```
bmd3k, line 1046: I found this check reassuring while refactoring the code but I recognize it's a weird magic number.  Hopefully it's continued existence is justified by the comment.

tensorboard/uploader/uploader.py
```diff
@@ -536,12 +529,7 @@ def _create_run(self, run_name):
             request budget.
         """
         run_proto = self._request.runs.add(name=run_name)
-        # We can't calculate the proto key cost exactly ahead of time, as
```
davidsoergel, line 539: I don't quite understand this and the related comments below.  Should it be retained?  It seems to me that the total cost was being correctly summed up before across levels of the hierarchy, and still is, so you're right to delete the comments... right?

tensorboard/uploader/uploader_test.py
davidsoergel: ok with me

tensorboard/uploader/uploader.py
bmd3k: I've improved the comments to explain how proto cost is calculated.  Hopefully this is clearer.  Unfortunately I had to familiarize myself with this to convince myself that I could reuse this logic for Tensor sending.

PR #3539, opened 2020-04-22
Remove functionality of legacy DB mode

wchargin:
Summary:
TensorBoard typically reads from a logdir on disk, but the `--db_import`
and `--db` flags offered an experimental SQLite-backed read path. This
DB mode has always been clearly marked as experimental, and does not
work as advertised. For example, runs were broken in all dashboards.
A data set with *n* runs at top level would show as *n* runs in
TensorBoard, but all of them would have the name `.` and the same color,
and only one of them would actually render data. The images dashboard
was also completely broken and did not show images. And lots of plugins
just didn’t support DB mode at all. Furthermore, the existence of DB
mode is an active impediment to maintenance, and also accounts for some
of the few remaining test cases that require TensorFlow 1.x to run.

This patch removes support for DB mode from all plugins, removes the
`db_connection_provider` field from `TBContext`, and renders the `--db`
and `--db_import` flags functionally inoperative. The plumbing itself
will be removed in a separate commit; a modicum of care is required for
the `TensorBoardInfo`  struct.

Test Plan:
Running with a standard demo logdir, TensorBoard still works fine with
both `--generic_data false` and `--generic_data true`.

wchargin-branch: nodb-functionality


tensorboard/plugins/histogram/histograms_plugin.py
```diff
@@ -211,70 +167,6 @@ def histograms_impl(self, tag, run, experiment, downsample_to=None):
...
-                """,
-                {"run": run, "tag": tag, "plugin": metadata.PLUGIN_NAME},
-            )
-            row = cursor.fetchone()
-            if not row:
-                raise errors.NotFoundError(
-                    "No histogram tag %r for run %r" % (tag, run)
-                )
-            (tag_id,) = row
-            # Fetch tensor values, optionally with linear-spaced sampling by step.
```
wchargin, line 238: RIP cool linear sampling SQL query, 2018–2020.

nfelt: 💔⚰️

tensorboard/plugins/base_plugin.py
```diff
@@ -273,13 +272,6 @@ def __init__(
...
           data_provider: Instance of `tensorboard.data.provider.DataProvider`. May
             be `None` if `flags.generic_data` is set to `"false"`.
-          db_connection_provider: Function taking no arguments that returns a
-              PEP-249 database Connection object, or None if multiplexer should be
-              used instead. The returned value must be closed, and is safe to use in
-              a `with` statement. It is also safe to assume that calling this
-              function is cheap. The returned connection must only be used by a
-              single thread. Things like connection pooling are considered
-              implementation details of the provider.
           db_uri: The string db URI TensorBoard was started with. If this is set,
```
nfelt, line 275: Should `db_uri` also be removed here, or is it being used in the manager code somehow?

wchargin: Yep; thanks. Done.

PR #3542, opened 2020-04-23
Add optional --experiment_id parameter to upload subcommand

shawwn:

`tensorboard dev upload --logdir <MODEL_DIR> --experiment_id PgEl74YtR2KWjRSjsDtwiQ` would allow the user to upload additional data to the existing experiment `PgEl74YtR2KWjRSjsDtwiQ`.

I often leave a `tensorboard dev upload --logdir <MODEL_DIR>` command running during an active training session. However, sometimes the `tensorboard dev upload` command is interrupted before the training session ends. (For example, if the network dies, or if a training session is resumed at a later date.) The `--experiment_id` option allows the user to "resume uploading."

I modified `uploader_lib.TensorBoardUploader.create_experiment` to accept an optional `experiment_id` parameter. If it's not `None` (the default) then the passed ID is used. Otherwise, it retrieves a new `experiment_id` using `CreateExperimentRequest`, as usual.

image

Beyond that, I added `--experiment_id` to the flags parser:

image

... and some boilerplate to use the flag:

image

* Screenshots of UI changes

None

1. Create a new experiment as usual using `tensorboard dev upload --logdir <MODEL_DIR>`

2. Retrieve the experiment ID from the resulting URL

3. Interrupt the `tensorboard dev upload` command

4. Add additional training data to `<MODEL_DIR>`

5. Run `tensorboard dev upload --logdir <MODEL_DIR> --experiment_id <ID>`, where `<ID>` is the experiment ID retrieved in step 2

6. Visit the experiment URL and verify the new training data appears there

None

### DIFF
--- a/tensorboard/uploader/flags_parser.py
+++ b/tensorboard/uploader/flags_parser.py
@@ -87,13 +87,6 @@ def define_flags(parser):
         help="Directory containing the logs to process",
     )
     upload.add_argument(
-        "--experiment_id",
-        metavar="ID",
-        type=str,
-        default=None,
-        help="Upload to a specific experiment ID",
-    )
-    upload.add_argument(
         "--name",
         type=str,
         default=None,
@@ -104,6 +97,13 @@ def define_flags(parser):
         type=str,
         default=None,
         help="Experiment description. Markdown format.  Max 600 characters.",
+    )
+    upload.add_argument(
+        "--experiment_id",
+        metavar="ID",
+        type=str,
+        default=None,
+        help="ID of an existing experiment to upload to",
     )
     upload.add_argument(
         "--plugins",
--- a/tensorboard/uploader/flags_parser.py
+++ b/tensorboard/uploader/flags_parser.py
@@ -87,6 +87,13 @@ def define_flags(parser):
         help="Directory containing the logs to process",
     )
     upload.add_argument(
+        "--experiment_id",
+        metavar="ID",
+        type=str,
+        default=None,
+        help="Upload to a specific experiment ID",
+    )
+    upload.add_argument(
         "--name",
         type=str,
         default=None,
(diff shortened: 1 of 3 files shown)


bileschi:
Hi @shawwn !

Thanks for prototyping this!  Writing additional data to a known experiment is a good idea, and something we are definitely interested in.    Before we proceed, there are a few concerns because the uploader will re-upload everything it has already uploaded.  This causes a few problems, 

(1).  It's slower and uses more bandwidth
(2). It overwrites the update time (which may or may not be the user intention)
(3). The accounting system currently assumes each experiment is written only once, (which is a bug @Googlers : b/154841586 ).  The effect is that the uploader is charged new quota every time the same experiment is written.  The quota can be released by deleting the experiment, but it's certainly not a great experience.

We can probably live with (1) for the time being, until there is a way to reconcile the logdir view with the tbdev view. and only upload the diffs. (2) can be addressed via the reconciliation process, the experiment owner can overwrite the timestamp if they want, and the timestamp should definitely be changed if the data itself changes.   We can prioritize (3) and notify here when there is a fix.

Thanks again!
Sincerely,
Stan

shawwn:
(1) is fine, because uploads only take <10 minutes anyway. (The largest experiment I've ever uploaded still took <24h: https://tensorboard.dev/experiment/hajDPiVGTRylalOYfo73lg/#scalars)

For (2), are you saying it overwrites the "created" time?

image

If so, I didn't realize it. That's a reasonable concern, but doesn't matter much to me personally.

(3) is something I can live with until it's fixed internally. Uploading to a specific --experiment_id is probably a rare operation. Normally users just leave `tensorboard dev upload` running in parallel with the training session.

DavidGOrtega:
> (3) is something I can live with until it's fixed internally. Uploading to a specific --experiment_id is probably a rare operation.

Actually its a very interesting option for CI/CD MLOps space. Im trying your PR. Thanks!

malteos:
Is there any update on this or can I assume that uploading to existing experiments will not be possible anytime soon?

bmd3k:
> Is there any update on this or can I assume that uploading to existing experiments will not be possible anytime soon?

Hi @malteos. Unfortunately this will not be possible anytime soon.

PR #3544, opened 2020-04-23
[DebuggerV2] Revise and re-enable debugger_v2_plugin_test for Const ops

caisq:
  * This is a follow-up / clean-up task for adding Const ops to the set of instrumented ops by tfdbg2 (b/153722488)
  * In debugger_v2_plugin_test.py, several assertions related to the exact set of ops executed inside tf.functions (Graphs) are updated to reflect the Const ops that weren't instrumented previously but are now.
  * Disabling tags for the task are removed from the BUILD file.


caisq:
Thanks for the review, @bileschi !

PR #3545, opened 2020-04-23
Support writing tensors in uploader

bmd3k:

Allow upload of Tensor data to TensorBoard.dev so that we can later enable Tensor-based plugins like histograms.

Add uploader._TensorBatchedRequestSender, modelled very closely to uploader._ScalarBatchedRequestSender. It builds requests to WriteTensor.  Integrate it into uploader._BatchedRequestSender.

Wrote some new tests.

Local testing:
    * Update dev Storzy to allow histograms.
    * Run a local TensorBoard.dev instance, connected to the dev Storzy environment, and with histograms enabled.
    * Run the uploader with this command: 

   ```shell
    bazel run tensorboard -- dev \
    --origin http://localhost:8080 --api_endpoint api-dev.tensorboard.dev upload \
    --logdir <logdir with 2MB of histogram data> --plugins scalars,graphs,histograms \
    --verbose 0
   ```
    * Observe that the upload is broken into multiple requests below 131072 bytes:

image

    * Open new experiment on local frontend and observe histograms working:

image


tensorboard/uploader/uploader.py
```diff
@@ -330,12 +330,16 @@ def __init__(
         self._scalar_request_sender = _ScalarBatchedRequestSender(
             experiment_id, api, rpc_rate_limiter,
         )
+        self._tensor_request_sender = _TensorBatchedRequestSender(
+            experiment_id,
+            api,
+            # BDTODO: ASK REVIEWERS: Should we use a different rate limiter?
```
bmd3k, line 336: Question for reviewers: Should I reuse the rate limiter for _ScalarBatchedRequestSender? I think I should.

tensorboard/uploader/uploader.py
```diff
@@ -581,6 +581,157 @@ def _create_point(self, tag_proto, event, value):
         return point
 
 
+class _TensorBatchedRequestSender(object):
```
bmd3k, line 584: This class generously copies from _ScalarBatchedRequestSender. David, you mentioned offline that this is probably fine but would like to hear if you have an updated opinion after you've seen the code.  I did refactor out some common non-trivial logic in previous PRs and reuse it here.

tensorboard/uploader/uploader_test.py
```diff
@@ -94,7 +98,15 @@ def _create_mock_client():
     return mock_client
 
 
-_SCALARS_ONLY = frozenset((scalars_metadata.PLUGIN_NAME,))
+# By default allow at least one plugin for each upload type: Scalar, Tensor, and
+# Blobs.
+_SCALARS_HISTOGRAMS_AND_GRAPHS = frozenset(
```
bmd3k, line 103: It just seemed simpler to enable all of scalar, tensor, and blob functionality by default.

tensorboard/uploader/uploader_test.py
```diff
@@ -289,6 +314,42 @@ def scalar_event(tag, value):
             uploader.start_uploading()
         self.assertEqual(4 + 6, mock_client.WriteScalar.call_count)
         self.assertEqual(4 + 6, mock_rate_limiter.tick.call_count)
+        self.assertEqual(0, mock_blob_rate_limiter.tick.call_count)
+
+    def test_start_uploading_tensors(self):
```
bmd3k, line 319: The one "integration" test that exercises from the uploader level.

tensorboard/uploader/uploader_test.py
```diff
@@ -689,7 +728,7 @@ def test_upload_full_logdir(self):
 
 class BatchedRequestSenderTest(tf.test.TestCase):
```
bmd3k, line 729: Modifications in this Test class mostly integrate Tensors into existing tests.

tensorboard/uploader/uploader_test.py
```diff
@@ -1127,6 +1196,289 @@ def test_wall_time_precision(self):
         )
 
 
+class TensorBatchedRequestSenderTest(tf.test.TestCase):
```
bmd3k, line 1199: This Test class generously copies from ScalarBatchedRequestSenderTest and focuses on self-contained logic in TensorBatchedRequestSender without the other layers in uploader.py.

tensorboard/uploader/uploader.py
ericdnielsen: On server side, we can set those limits per rpc request, so we should either have that ability here, or we're stuck setting the limit to the lowest value of the rate limiter.  Makes me think long term, we probably don't want to have set the limit both places.  Server only with retry/backoff logic in the uploader would probably be nicer, but out of scope for this PR.

I'm personally OK with re-using the scalar batch, but I think its shortcut we'll need to remove in the future.

tensorboard/uploader/uploader.py
```diff
@@ -330,12 +330,16 @@ def __init__(
...
+            experiment_id,
+            api,
+            # BDTODO: ASK REVIEWERS: Should we use a different rate limiter?
+            rpc_rate_limiter,
+        )
         self._blob_request_sender = _BlobRequestSender(
             experiment_id, api, blob_rpc_rate_limiter, max_blob_size
         )
 
-        # TODO(nielsene): add tensor case here
```
ericdnielsen, line 337: Remove this todo as this is the implementation?

tensorboard/uploader/uploader.py
```diff
@@ -581,6 +581,157 @@ def _create_point(self, tag_proto, event, value):
...
+          _OutOfSpaceError: If adding the tag would exceed the remaining
+            request budget.
+        """
+        tag_proto = run_proto.tags.add(name=tag_name)
+        tag_proto.metadata.CopyFrom(metadata)
+        self._byte_budget_manager.add_tag(tag_proto)
+        return tag_proto
+
+    def _create_point(self, tag_proto, event, value):
+        """Adds a tensor point to the given tag, if there's space.
```
ericdnielsen, line 709: What happens if a single point (which is still a full tensor) is larger than the request limit?  It looks like that should result in an OutOfSpace event, as I don't see any support for splitting a single tensor between two uploads.  Does this need something similar to the blob chunking?  The limit sounds low enough that this might be an issue.

bmd3k: Good catch, a single point larger than 128Kb would just fail the upload.  I've started an internal discussion about reasonable maximum Tensor point size and filed an internal issue to track any necessary changes.   Hopefully it's just a matter of choosing a different max chunk size for WriteTensor and configuring this code and our servers to accept it.

tensorboard/uploader/uploader.py
davidsoergel: I agree that there are open questions about the best way to do this later, worth leaving a TODO about, but for now reusing the scalars rate limiter is fine.

tensorboard/uploader/uploader.py
davidsoergel: +1 we can raise the server-side limit and so forth, but the possibility always remains of a very large tensor exceeding whatever limit we set.  I think we can safely issue a warning, skip that tensor, and proceed-- but doing that needs some logic so that we don't just crash with RuntimeError at line 635.

davidsoergel: OK to punt to followup PR, per offline convo

PR #3553, opened 2020-04-27
pr_curves: compute available time entries on client

### LINKED ISSUE #3184: Simplify PR curve step selection logic for data provider performance
The PR curves dashboard renders a single step selector for each run
(rather than for each time series). The set of available steps is chosen
by picking an arbitrary tag and checking the steps for which that time
series has data. As different time series aren’t guaranteed to have the
same downsampling pattern if they include different numbers of steps,
this only sets the “target” step for each time series from the given
run. In the UI, each time series lists what step it’s _actually_ drawing
data from.

This data reading pattern doesn’t fit too nicely into the data provider
APIs, because to determine the exact set of downsampled steps one must
read the entire time series, so listing the steps for each run requires
performing one `read_tensors` queries for each run, scoped to the
“canonical” (arbitrarily chosen) tag for that run. This is the classic
“N+1 queries” performance bug.

We could make these queries less expensive by adding `RunTagFilter`
support for a set of `(run, tag)` pairs rather than just the cross
product of a set of `runs` with a set of `tags`; as discussed in [the
generic data APIs design doc][data-apis], we left this more flexible
form out of the initial design for simplicity. But we could also just
avoid doing this work in the first place: instead of viewing the actual
step pattern for an arbitrary time series, just take the max of all
`max_step` values for the run, and interpolate from step 0 up to that
max step. This can be done with just one `list_tensors` RPC.

This makes the “arbitrary tag” approximation more crude, but it was only
ever an approximation anyway.

[data-apis]: https://github.com/tensorflow/tensorboard/blob/cb4aa69cb2d71da119837787a98284035e5ffb81/docs/design/generic_data_apis.md

wchargin:
Summary:
This patch removes the `.../pr_curves/available_time_entries` route. The
same data can be computed on the client based on the results of the
individual `/pr_curves` routes, and there’s no ordering constraint. This
removes a particularly far-reaching query so that the PR curves plugin
can work more efficiently with data providers. Fixes #3184.

Test Plan:
Point TensorBoard at demo data for the PR curves plugin, and note that
the time axis sliders still work in step, relative, and wall-time axis
modes.

wchargin-branch: pr-curves-no-available-time-entries

### DIFF
--- a/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
+++ b/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
@@ -220,11 +220,9 @@ limitations under the License.
         this.reload();
       },
       reload() {
-        Promise.all([this._fetchTags()]).then(
-          () => {
-            this._reloadCards();
-          }
-        );
+        Promise.all([this._fetchTags()]).then(() => {
+          this._reloadCards();
+        });
       },
       _shouldOpen(index) {
         return index <= 2;
--- a/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
+++ b/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
@@ -279,12 +279,12 @@ limitations under the License.
         return tf_utils.aggregateTagInfo(runToTagInfo, defaultDisplayName);
       },
       _createDataChangeCallback(tag) {
-        return function onDataChange(runToData) {
-          this.set(
-            '_tagToRunToData',
-            Object.assign({}, this._tagToRunToData, {[tag]: runToData})
-          );
-        }.bind(this);
+        return (runToData) => {
+          this.set('_tagToRunToData', {
+            ...this._tagToRunToData,
+            [tag]: runToData,
+          });
+        };
       },
       _computeRunToAvailableTimeEntries(tagToRunToData) {
         const canonicalTag = {}; // map from run to canonical tag name
--- a/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
+++ b/tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
@@ -126,6 +126,7 @@ limitations under the License.
                   tag-metadata="[[_tagMetadata(_runToTagInfo, item.runs, item.tag)]]"
                   request-manager="[[_requestManager]]"
                   run-to-step-cap="[[_runToStep]]"
+                  on-data-change="[[_createDataChangeCallback(item.tag)]]"
                 ></tf-pr-curve-card>
               </template>
             </tf-category-paginated-view>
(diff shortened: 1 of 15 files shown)


tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
```diff
@@ -299,6 +278,35 @@ <h3>No precision–recall curve data was found.</h3>
         const defaultDisplayName = tag.replace(/\/pr_curves$/, '');
         return tf_utils.aggregateTagInfo(runToTagInfo, defaultDisplayName);
       },
+      _createDataChangeCallback(tag) {
+        return function onDataChange(runToData) {
```
stephanwlee, line 282: Do you preferred named function for better debugging? Arrow function may work better since you use `this`.

tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
```diff
@@ -299,6 +278,35 @@ <h3>No precision–recall curve data was found.</h3>
...
+            Object.assign({}, this._tagToRunToData, {[tag]: runToData})
+          );
+        }.bind(this);
+      },
+      _computeRunToAvailableTimeEntries(tagToRunToData) {
+        const canonicalTag = {}; // map from run to canonical tag name
+        for (const [tag, runToData] of Object.entries(tagToRunToData)) {
+          for (const [run, data] of Object.entries(runToData)) {
+            // arbitrary stable ordering: smallest tag name is canonical
+            if (canonicalTag[run] == null || tag < canonicalTag[run]) {
```
stephanwlee, line 294: No AI: `tag < canonicalTag[run]` may want to use the [compareTagNames](https://github.com/tensorflow/tensorboard/blob/b929faea0eb43b1fc61e8b30398423191a40ec73/tensorboard/components/vz_sorting/sorting.ts#L26) ideally but I assume our Python backend did not do that either.

tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
```diff
@@ -299,6 +278,35 @@ <h3>No precision–recall curve data was found.</h3>
         const defaultDisplayName = tag.replace(/\/pr_curves$/, '');
         return tf_utils.aggregateTagInfo(runToTagInfo, defaultDisplayName);
       },
+      _createDataChangeCallback(tag) {
+        return function onDataChange(runToData) {
+          this.set(
+            '_tagToRunToData',
+            Object.assign({}, this._tagToRunToData, {[tag]: runToData})
```
stephanwlee, line 285: `{...this._tagToRunToData, [tag]: runToData}` reads pretty well :)

tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
wchargin: You’re right; that should definitely work. (I was being a bit cautious
with the receiver because I don’t know exactly how Polymer binds `this`
in all the cases that matter, but you’re of course right that if an
explicit `.bind(this)` works then an arrow function should, too, so that
was silly of me.) Done.

tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
wchargin: Huh, I thought for some reason that we only supported array-spread and
not object-spread; there are no existing matches for `{...` in HTML
files. But indeed this seems to work fine. Done; thanks.

tensorboard/plugins/pr_curve/tf_pr_curve_dashboard/tf-pr-curve-dashboard.html
wchargin: Correct that the Python backend didn’t, and I don’t think it’s super
important. The intent is that it shouldn’t matter which tag is picked,
so I have a weak preference to keeping this simple.

PR #3560, opened 2020-04-29
DataFrame API: Initial implementation of ExperimentFromDev

caisq:
  * Start implementing the pandas DataFrame data access API
  * Add base class `BaseExperiment`
  * Add an implementation of `BaseExperiment` for tensorboard.dev, accessible via`tensorboard.data.experimental.ExperimentFromDev`.
  * Impelment `ExperimentFromDev.get_scalars()`
    * run and tag filtering will be implemented in future PRs.
    * DataFrame pivoting (using pandas `pivot_table`) with missing-value warning is implemented
  * Overriding api_endpoint is supported by `ExperimentFromDev`.
  * Unit tests added in experiment_from_dev_test.py
  * Manual verification against tensorboard.dev using the newly added test_binary.py.


tensorboard/data/experimental/experiment_from_dev.py
```diff
@@ -0,0 +1,162 @@
...
+            }
+        )
+        if pivot:
+            dataframe = dataframe.pivot_table(
+                ["value", "wall_time"], ["run", "step"], "tag",
+            )
+            if np.any(dataframe.isnull().values):
+                sys.stderr.write(
+                    "WARNING: pivoted DataFrame contains missing value(s). "
+                    "This is likely due to tags that contain uneven numbers of steps. "
```
wchargin, line 119: This is a bit ambiguous, and I parsed it wrong the first time (and it
took me a while to figure out what was meant). I read “tags that contain
uneven numbers of steps” to mean “two tags with different numbers of
steps”, as in;

```python
# Tags with uneven numbers of steps: tag `acc` has 3 steps for run
# `train` but only 1 step for run `test`.
df = pandas.DataFrame(
    {
        "run": ["train", "train", "train", "test"],
        "tag": ["acc", "acc", "acc", "acc"],
        "step": [1, 2, 3, 3],
        "wall_time": [1000, 2000, 3000, 3500],
        "value": [0.1, 0.7, 0.9, 0.8],
    }
)
```

But this pivots without any problem. What’s actually meant is something
more like “one run for which the step values are not the same for all
tags in that run”. Like this:

```python
# Runs with uneven steps across tags: run `train` has steps 1, 2, and 3
# for tag `xent` but only step 3 for tag `xent`.
df = pandas.DataFrame(
    {
        "run": ["train", "train", "train", "train"],
        "tag": ["xent", "xent", "xent", "acc"],
        "step": [1, 2, 3, 3],
        "wall_time": [1000, 2000, 3000, 3500],
        "value": [0.1, 0.7, 0.9, 0.8],
    }
)
```

This example *does* have the problem, which is properly detected by the
code. Perhaps we can update the message to be clearer?

tensorboard/data/experimental/experiment_from_dev.py
```diff
@@ -0,0 +1,162 @@
...
+                    "Consider calling `get_scalar()` with `pivot=False`"
```
wchargin, line 120: s/get_scalar/get_scalars/

tensorboard/data/experimental/experiment_from_dev.py
```diff
@@ -0,0 +1,162 @@
...
+                "step": steps,
+                "wall_time": wall_times,
+                "value": values,
+            }
+        )
+        if pivot:
+            dataframe = dataframe.pivot_table(
+                ["value", "wall_time"], ["run", "step"], "tag",
+            )
+            if np.any(dataframe.isnull().values):
```
wchargin, line 116: Looks like this check has false positives when the data contains actual
`NaN`s. Can we fix that? One way would be to check for missing values on
a copy where the data columns are known non-null:

```python
mask_frame = dataframe.copy()
mask_frame["value"] = [0] * len(mask_frame)
mask_frame["wall_time"] = [0] * len(mask_frame)
mask = df.pivot_table(["value", "wall_time"], ["run", "step"], "tag").isnull()
missing_data = np.any(mask.values)
# and `mask.values` lists exactly which are missing
```

…but feel free to suggest a better way (I have never used `pandas`; this
is just from my initial read of the docs).

tensorboard/data/experimental/experiment_from_dev.py
```diff
@@ -0,0 +1,162 @@
...
+    if compat.verdict == server_info_pb2.VERDICT_WARN:
+        sys.stderr.write("Warning [from server]: %s\n" % compat.details)
+        sys.stderr.flush()
+    elif compat.verdict == server_info_pb2.VERDICT_ERROR:
+        raise ValueError("Error [from server]: %s" % compat.details)
+    else:
+        # OK or unknown; assume OK.
+        if compat.details:
+            sys.stderr.write("%s\n" % compat.details)
+            sys.stderr.flush()
```
wchargin, line 162: This is never used; we probably want to nicely fail the initialization
on an error verdict, no?

tensorboard/data/experimental/test_binary.py
```diff
@@ -0,0 +1,49 @@
...
+def main(args):
+    experiment = experiment_from_dev.ExperimentFromDev(
+        args.experiment_id, api_endpoint=args.api_endpoint
+    )
+    dataframe = experiment.get_scalars()
+    print(dataframe)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser("Test run of ExperimentFromDev")
```
wchargin, line 35: Let’s try to keep all the logic—and in particular, all assignments—in a
function (fine to put this in `main`) so we don’t have to worry about
Python scoping issues.

tensorboard/pip_package/requirements_dev.txt
```diff
@@ -17,6 +17,7 @@
 
 # For tests
 grpcio-testing==1.24.3
+pandas==1.0.3
```
wchargin, line 20: OK for now; we may want this to really be a Pip “extra”.

tensorboard/data/experimental/BUILD
```diff
@@ -0,0 +1,73 @@
...
+
+package(default_visibility = ["//tensorboard:internal"])
+
+licenses(["notice"])  # Apache 2.0
+
+exports_files(["LICENSE"])
+
+py_library(
+    name = "experimental",
+    srcs = ["__init__.py"],
```
wchargin, line 12: We generally do this only for public APIs, and this is explicitly
intended to *not* yet be a public API; should we remove this and the
`__init__.py`?

tensorboard/data/experimental/base_experiment.py
```diff
@@ -0,0 +1,69 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import abc
+
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class BaseExperiment(object):
```
wchargin, line 27: This module is `srcs_version = "PY3"`, as it should be; let’s just
replace this with the Python 3 form (`class Foo(metaclass=Bar)`). We
shouldn’t add `six` to new code.

tensorboard/data/experimental/base_experiment.py
```diff
@@ -0,0 +1,69 @@
...
+    """Base class for experiment data access."""
+
+    # TODO(cais): Add list_scalar_runs().
+    # TODO(cais): Add list_scalar_tags().
+
+    @abc.abstractmethod
+    def get_scalars(self, runs_filter=None, tags_filter=None, pivot=True):
```
wchargin, line 34: Prefer using `None` as a default argument value rather than `True`, and
resolving inside the function itself. See, e.g.:
<https://unhashable.com/always-use-none-for-default-args-in-python/>

tensorboard/data/experimental/base_experiment.py
```diff
@@ -0,0 +1,69 @@
...
+          If `not pivot`:
+            A DataFrame with the following columns.
+              - run: (non-null object)
+              - tag: (non-null object)
+              - steps: (non-null int64)
+              - wall_time: (non-null object)
+              - value: (non-null float32)
+
+          The row sort order is: run start time >> tag >> steps >> wall_time
+          (ascending).
```
wchargin, line 67: Is this implemented? I don’t see any place where we sort by run start
time, either here or on the server.

tensorboard/data/experimental/experiment_from_dev_test.py
```diff
@@ -0,0 +1,250 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import sys
+from unittest import mock
+
+import numpy as np
+import grpc
```
wchargin, line 25: nit: abc

tensorboard/data/experimental/experiment_from_dev.py
```diff
@@ -0,0 +1,162 @@
...
+        dataframe = pandas.DataFrame(
+            {
+                "run": runs,
+                "tag": tags,
+                "step": steps,
+                "wall_time": wall_times,
+                "value": values,
+            }
+        )
+        if pivot:
```
wchargin, line 112: nit: Can we just `if not pivot: return dataframe` to avoid keeping two
codepaths in mind? (Also avoids indenting the whole block.)

tensorboard/data/experimental/experiment_from_dev_test.py
```diff
@@ -0,0 +1,250 @@
...
+        def fake_stderr_write(msg):
+            stderr_messages.append(msg)
+
+        with mock.patch.object(
+            experiment_from_dev,
+            "get_api_client",
+            lambda api_endpoint: mock_api_client,
+        ):
+            experiment = experiment_from_dev.ExperimentFromDev("789")
+            with mock.patch.object(sys.stderr, "write", fake_stderr_write):
```
wchargin, line 158: Per other comment, would really prefer to not write to `stderr`, but if
you must then please test this with `contextlib.redirect_stderr` rather
than a mock to at least get the right semantics.

tensorboard/data/experimental/experiment_from_dev_test.py
```diff
@@ -0,0 +1,250 @@
...
+
+        # Check values.
+        train_losses = list(dataframe.loc["train", ("value", "loss")])
+        self.assertAllClose(train_losses, 1.0 / (np.arange(0, 10) + 1.0))
+        train_accuracies = list(dataframe.loc["train", ("value", "accuracy")])
+        self.assertAllClose(train_accuracies, 1.0 / (10.0 - np.arange(0, 10)))
+        test_losses = list(dataframe.loc["test", ("value", "loss")])
+        self.assertAllClose(test_losses, -1.0 / (np.arange(0, 10) + 1.0))
+        test_accuracies = list(dataframe.loc["test", ("value", "accuracy")])
+        self.assertAllClose(test_accuracies, -1.0 / (10.0 - np.arange(0, 10)))
```
wchargin, line 99: Would it perhaps be easier to read both the test code and failure
messages if we just construct the expected dataframe and assert that
it’s equal/close to the actual dataframe? Likewise below.

tensorboard/data/experimental/experiment_from_dev_test.py
```diff
@@ -0,0 +1,250 @@
...
+        mock_api_client.StreamExperimentData = mock.Mock(
+            wraps=stream_experiment_data
+        )
+
+        with mock.patch.object(
+            experiment_from_dev,
+            "get_api_client",
+            lambda api_endpoint: mock_api_client,
+        ):
+            experiment = experiment_from_dev.ExperimentFromDev("789")
```
wchargin, line 210: It looks like this giant mock block is identical to the one in the
previous first case. Could you please combine the test cases, using
`self.subTest` for failure isolation?

tensorboard/data/experimental/experiment_from_dev_test.py
```diff
@@ -0,0 +1,250 @@
...
+import grpc_testing
+import tensorflow as tf
+
+from tensorboard.data.experimental import experiment_from_dev
+from tensorboard.uploader import test_util
+from tensorboard.uploader.proto import export_service_pb2
+from tensorboard.util import grpc_util
+
+
+class ExperimentFromDevTest(tf.test.TestCase):
```
wchargin, line 35: At a glance it doesn’t look like this uses any TensorFlow-specific
features; can it be a `tb_test.TestCase` instead, which runs
significantly faster due to skipping the TensorFlow import?

tensorboard/pip_package/requirements_dev.txt
caisq: Ack.

tensorboard/data/experimental/experiment_from_dev_test.py
caisq: `self.assertEqual()` doesn't work for two DataFrames. The `assertAllEqual()` and `assertAllClose()` from the `tf.test.TestCase` we're using doesn't work very well either. They succeed for equal/close DataFrames. But the big problem is that they don't check content of the indices. E.g., if you have two DataFrames with the same values and wall_times, but different run names, they won't catch the different. They'll just pass happily. Also, for failure conditions, they don't actually print out useful message about which column or which index differs between the two.

DataFrame has the `equals()` method. But it's not designed for unit testing. When two DataFrames are unequal, code like `self.assertEqual(dataframe.equals(expected))` won't tell you where the difference is.

The code here is a little long, but when failures occur, the line numbers gives you a clear idea which part of the DataFrame is the problem.

tensorboard/data/experimental/experiment_from_dev_test.py
caisq: Done.

tensorboard/data/experimental/experiment_from_dev_test.py
caisq: `pandas.testing.assert_frame_equal()` works well. Done revising the tests to use it. This shortens the test code quite a lot. Thanks.

caisq:
@wchargin I may have done something weird w.r.t. GitHub CR and caused my comments to be swallowed. But to summarized the latest changes:
1. I'm using `pandas.testing.assert_frame_equal()` now.
2. I removed the `__init__.py` file in the experimental/ folder. Let's leave the packaging discussion to a later PR.

tensorboard/BUILD
```diff
@@ -436,6 +436,16 @@ py_library(
...
 )
 
+py_library(
+    name = "expect_pandas_installed",
+    # This is a dummy rule used as a pandas dependency in open-source.
+    # We expect pandas to already be installed on the system, e.g. via
+    # `pip install pandas`.
+    # NOTE: Unlike other parallel dependencies in this file, pandas is an
+    # optional dependency.
+    visibility = ["//visibility:public"],
```
wchargin, line 446: When you rebase master, please remove this `//visibility:public` spec,
as all the others were removed in #3568.

tensorboard/data/experimental/experiment_from_dev_test.py
```diff
@@ -0,0 +1,206 @@
...
+                        pandas.testing.assert_frame_equal(
+                            dataframe,
+                            expected.pivot_table(
+                                ["value", "wall_time"], ["run", "step"], "tag"
+                            ),
+                            check_names=True,
+                        )
+                    else:  # pivot == False
+                        pandas.testing.assert_frame_equal(
+                            dataframe, expected, check_names=True
```
wchargin, line 120: fwiw, `check_names=True` is the default.

tensorboard/BUILD
caisq: Done.

tensorboard/data/experimental/experiment_from_dev_test.py
caisq: Yes, I'm aware of that. I prefer to be explicit and safer against potential future changes.

PR #3563, opened 2020-04-29
ng: fix infinite loop when on plugin selector

stephanwlee:
cause: mat-tab-group emits the changed event in the Angular's equivalent
to `componentDidUpdate`. The problem is that:
1. it emits the [change in a promise](https://github.com/angular/components/blob/58b8d30efb0e71f0f4bbc541405cda1adce26f16/src/material/tabs/tab-group.ts#L207-L213) (next tick)
2. it emits the change when the value is changed programmatically.

While lacking complete understanding of the flow, empirically, I found there to be some kind of event dispatch queue and the emission to not invoke our callback immediately.

Logging result
<details>

```
# Clicked on graphs, images, audio in quickly
11:50:44.922 tab-group: scheduling async
11:50:45.086 tab-group: triggering async
11:50:45.153 tab-group: scheduling async
11:50:45.202 tab-group: triggering async
11:50:45.249 tab-group: scheduling async
11:50:45.274 tab-group: triggering async
11:50:45.322 tab-group: scheduling async
11:50:45.430 tab-group: triggering async
# `onActivePluginSelectionChanged` is the callback that listens to
# mat-tab-group's changed event.
# Because flux cycle is synchronous, it makes sense to action to get dispatched
# then schedule a new update (update should change the value to `graphs` when
# the current value is audio).
11:50:46.085 onActivePluginSelectionChanged graphs
11:50:46.086 [Core] Plugin Changed
11:50:46.136 ngOnChanges graphs
11:50:46.139 tab-group: scheduling async
# Huh? which async is triggered? If this is the `graphs` from `11:50:46.139` or
# one from `11:50:45.202`?
11:50:46.440 tab-group: triggering async
# It looks like `triggering async` has almost nothing to do with the event we
# get.
11:50:46.510 onActivePluginSelectionChanged images
```

</details>

fix: no longer use the mat-tab-group's change event and listen directly
at the click event.

Fixes b/155104143.

### DIFF
--- a/tensorboard/webapp/header/plugin_selector_component.ng.html
+++ b/tensorboard/webapp/header/plugin_selector_component.ng.html
@@ -17,14 +17,20 @@ limitations under the License.
 <mat-tab-group
   class="active-plugin-list"
   [selectedIndex]="getActivePluginIndex()"
-  (selectedTabChange)="onActivePluginSelectionChanged($event)"
   animationDuration="100ms"
 >
-  <mat-tab
-    *ngFor="let plugin of activePlugins"
-    [label]="plugin.tab_name"
-    [disabled]="!plugin.enabled"
-  >
+  <mat-tab *ngFor="let plugin of activePlugins" [disabled]="!plugin.enabled">
+    <ng-template mat-tab-label>
+      <!-- Manually subscribe to the click event on the tab content element.
+      Cannot trust the selectedTabChange event since it is async and can cause
+      infinte loop. -->
+      <span
+        class="plugin-name"
+        (click)="onActivePluginSelection($event, plugin.id)"
+      >
+        {{plugin.tab_name}}
+      </span>
+    </ng-template>
   </mat-tab>
 </mat-tab-group>
 <mat-form-field floatLabel="never" *ngIf="disabledPlugins.length > 0">
--- a/tensorboard/webapp/header/plugin_selector_component.ts
+++ b/tensorboard/webapp/header/plugin_selector_component.ts
@@ -13,7 +13,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {Component, Input, Output, EventEmitter} from '@angular/core';
-import {MatTabChangeEvent} from '@angular/material/tabs';
 import {MatSelectChange} from '@angular/material/select';
 
 import {PluginId} from '../types/api';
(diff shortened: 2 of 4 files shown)


tensorboard/webapp/header/plugin_selector_component.ts
```diff
@@ -41,8 +40,9 @@ export class PluginSelectorComponent {
     return this.activePlugins.findIndex(({id}) => id === this.selectedPlugin);
   }
 
-  onActivePluginSelectionChanged({index}: MatTabChangeEvent) {
-    this.onPluginSelectionChanged.emit(this.activePlugins[index].id);
+  onActivePluginSelection(event: Event, pluginId: PluginId) {
+    event.stopPropagation();
+    this.onPluginSelectionChanged.emit(pluginId);
```
wchargin, line 45: Do I read correctly that we no longer emit `onPluginSelectionChanged`
when the plugin changes due to a hash change event, and that that’s
acceptable because this is only used to dispatch a `changePlugin`
action, and in the case of a hash change event a different container
will fire that action?

tensorboard/webapp/header/plugin_selector_component.ng.html
```diff
@@ -17,14 +17,20 @@
...
+  <mat-tab *ngFor="let plugin of activePlugins" [disabled]="!plugin.enabled">
+    <ng-template mat-tab-label>
+      <!-- Manually subscribe to the click event on the tab content element.
+      Cannot trust the selectedTabChange event since it is async and can cause
+      infinte loop. -->
+      <span
+        class="plugin-name"
+        (click)="onActivePluginSelection($event, plugin.id)"
+      >
+        {{plugin.tab_name}}
```
wchargin, line 31: Naïve question, but from the way that this is structured it looks like
the click event would only be fired when the user clicks exactly on the
text rather than elsewhere on the tab; should the `(click)` listener be
on the `mat-tab` instead?

stephanwlee: What you are saying is correct but something really odd is going on. I think mat-tab-group is using non-conventional pattern and is kinda treating the element as a `ng-template`. Anyways, the click handler could not be bound correctly.

To work around the issue that you mentioned, I made sure the `span` take the width and height of the entire `mat-tab` element.

tensorboard/webapp/header/plugin_selector_component.ts
stephanwlee: Correct but clarification is required.

The fact that we fire `onPluginSelectionChanged ` when we set the index programmatically was not a desired but okay behavior. This has almost nothing to do with the hash since literally grab the value from the store and set it as the value to the component.

> when the plugin changes due to a hash change event

This is handled by some other module called deep linker which is responsible for listening to hash change and notify store accordingly. If that event happens to change the store state, then plugin_selector_component will receive the new value and trigger `onPluginSelectionChanged`.

tensorboard/webapp/header/plugin_selector_component.ng.html
wchargin: Ah, I see. Okay. Well… if it works. :shipit:

tensorboard/webapp/header/plugin_selector_component.ts
wchargin: Yep; makes sense. Thanks!

stephanwlee:
Manually confirmed that the change does not impact internal app negatively. Will submit when the refactor storm subsides.

wchargin:
Thanks! Whenever you’re ready.

PR #3564, opened 2020-04-29
[DebuggerV2] Implement HTTP route /graphs/op_info

caisq:
  * Continue developing DebuggerV2 plugin
  * Implement the /graphs/op_info route by making use of the `graph_by_id()` and `read_graph_op_creation_stack_trace()` methods of the underlying `DebugDataReader` and conforming to the `DataProvider` abstraction.
  * The response object tracks one level of (immediate) inputs and one level of (immediate) downstream consumers.
  * Unit tests added to debugger_v2_plugin_test.py covering cases of
    * Op having both inputs and consumers.
    * Op with only inputs and no consumers.
    * Op with only consumers and no inputs.
    * Error condition: invalid graph_id
    * Error condition: invalid op_name


tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -283,6 +284,48 @@ def _parse_graph_execution_data_blob_key(blob_key):
...
+def _parse_graph_op_info_blob_key(blob_key):
+    """Parse the BLOB key for graph op info.
+
+    Args:
+      blob_key: The BLOB key to parse. By contract, it should have the format:
+       `${GRAPH_OP_INFO_BLOB_TAG_PREFIX}_${graph_id}_${op_name}`
+
+    Returns:
+      - run ID
+      - grahp_id
```
davidsoergel, line 316: typo

tensorboard/plugins/debugger_v2/debug_data_provider.py
```diff
@@ -283,6 +284,48 @@ def _parse_graph_execution_data_blob_key(blob_key):
...
+        tags=["%s_%s_%s" % (GRAPH_OP_INFO_BLOB_TAG_PREFIX, graph_id, op_name)],
+    )
+
+
+def _parse_graph_op_info_blob_key(blob_key):
+    """Parse the BLOB key for graph op info.
+
+    Args:
+      blob_key: The BLOB key to parse. By contract, it should have the format:
+       `${GRAPH_OP_INFO_BLOB_TAG_PREFIX}_${graph_id}_${op_name}`
```
davidsoergel, line 312: Apparently this also has `.{$run_name}`?  (Overall I'm not a fan of packing variables into structured strings like this; it seems very brittle.  But I think the approach is water under the bridge-- i.e. I know it was in the design doc a while back etc.-- so I'll try not to complain more about it here :))

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -446,6 +451,104 @@ def GraphExecutionData(self, run, begin, end, trace_id=None):
...
+          op_name: Name of the op.
+
+        Returns:
+          A JSON-serializable object containing the information regarding
+          the op's creation.
+
+        Raises:
+          NotFoundError if the graph_id or op_name does not exist.
+        """
+        runs = self.Runs()
```
davidsoergel, line 471: nit: variable used only once; could be inlined

davidsoergel: For that matter, do you even need this test?  What situation is it protecting you from?

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -446,6 +451,104 @@ def GraphExecutionData(self, run, begin, end, trace_id=None):
...
+        if run not in runs:
+            return None
+        try:
+            graph = self._reader.graph_by_id(graph_id)
```
davidsoergel, line 475: From this perspective, this call looks like it could be expensive, i.e. you wouldn't want to reload the graph for every per-op request.  In practice I see it's backed by a dict (i.e. the graphs are already in memory) so it's OK, but I wonder if the API could be clearer about this somehow.  Maybe the issue is that the call is on `self._reader`, so I expect it to *read* something.  Should that be called `self._debug_data` or something?  Or, should `graph_by_id` really by `graph_reader_by_id`, to suggest (IIUC) that what you get is basically a sub-reader?  (I don't know a good answer, just speculating)

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -446,6 +451,104 @@ def GraphExecutionData(self, run, begin, end, trace_id=None):
...
+            if digest.input_names and any(
+                tensor_name_to_op_name(input_name) == op_creation_digest.op_name
+                for input_name in digest.input_names
+            ):
+                data_object["consumers"].append(
+                    self._opCreationDigestToDataObject(digest)
+                )
+        return data_object
+
+    def _opCreationDigestToDataObject(self, op_creation_digest):
```
davidsoergel, line 517: This function name sounds like it's just a format conversion, when it's actually reading from disk.  Maybe `_loadOpDataObject` or similar?

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -446,6 +451,104 @@ def GraphExecutionData(self, run, begin, end, trace_id=None):
...
+        Args:
+          run: Name of the run.
+          graph_id: Debugger-generated ID of the graph that contains
+            the op in question. This ID is available from other methods
+            of this class, e.g., the return value of `GraphExecutionDigests()`.
+          op_name: Name of the op.
+
+        Returns:
+          A JSON-serializable object containing the information regarding
+          the op's creation.
```
davidsoergel, line 466: Add "and its immediate inputs and outputs"

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -446,6 +451,104 @@ def GraphExecutionData(self, run, begin, end, trace_id=None):
...
+                # TODO(b/155308456): Once Placeholder and Const ops have their
+                # ad hoc naming issues fixed, remove this `if` logic.
+                if input_op_digest:
+                    data_object["inputs"][
+                        slot
+                    ] = self._opCreationDigestToDataObject(input_op_digest)
+        # Populate data about immediate consuming ops.
+        data_object["consumers"] = []
+        op_creation_digests = list(graph._op_by_name.values())
+        for digest in op_creation_digests:
```
davidsoergel, line 507: This is an exhaustive search over the whole graph, repeated every time.  Since the reader has all the digests in memory already anyway, can you do this just once in the reader to populate 'op_creation_digest.output_names'?

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -446,6 +451,104 @@ def GraphExecutionData(self, run, begin, end, trace_id=None):
...
+            "device_name": op_creation_digest.device_name,
+            "op_type": op_creation_digest.op_type,
+            "op_name": op_creation_digest.op_name,
+            "input_names": op_creation_digest.input_names or [],
+            "num_outputs": op_creation_digest.num_outputs,
+            "host_name": host_name,
+            "stack_trace": stack_trace,
+        }
+
+    def _getGraphStackIds(self, graph_id):
```
davidsoergel, line 534: Similarly this whole function could run once in the reader to populate `graph.outer_graph_stack` or something.

caisq:
Thanks for the review, @davidsoergel . You have some very good points. In order to address them in a good way, I'll need to submit some changes to tensorflow/python/debug. I'll send you CLs soon.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Thanks for the comment. As you pointed out `graph_by_id()` is cheap for the current local-only case, because it's just a `dict` lookup. In a future hosted setting, this will be more complex. It'll most likely involve a database (Spanner query) with graph_id and op_name as a composite key or a single key, but the details of that is yet to be worked out.

The name `_reader` is appropriate in the sense that the object does read data from the tfdbg2-format files. There is some confusion because it does not only read the data, but organizes the data (like building an internal index for consumers of ops). So perhaps the name can be improved to reflect that, but I don't know what a good alternative name should be. In any case, the naming issue is pre-existing and we can keep discussing it without blocking this PR IMO.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Good point. This concern is now addressed by the newly added `_reader.get_op_consumers()` method.

tensorboard/plugins/debugger_v2/debug_data_provider.py
caisq: Right. We are trying to encompass multiple pieces of information in a single blob key. As you mentioned, this is a result of the design that conforms to the DataProvider abstraction, in which there is not a natural fit for efficient look up of 
 individual ops in graphs. So we encode information in blob keys, as an escape hatch. 

This approach is currently okay, because `run_name` is hard-coded to `__default_debugger_run__`. Each logdir currently is mandated to have at most only one run of tfdbg2 data. So there is no ambiguity when run name is included as the tail of the blob key, preceded by a `.`. When we support multiple debugger runs in the same logdir in the future, we can continue enforcing a run name pattern that doesn't contain `.`s.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
davidsoergel: OK.  Partly I guess I was wondering if this is the right place for the test (as opposed to doing it first in the caller) but nbd.  TODO SGTM, but I don't see it?

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -446,6 +452,87 @@ def GraphExecutionData(self, run, begin, end, trace_id=None):
             ],
         }
 
+    def GraphOpInfo(self, run, graph_id, op_name):
```
davidsoergel, line 455: Function names should start with lowercase.  I see there's already a local uppercase convention in this file, so OK I guess.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
davidsoergel: ok

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: The tensorboard code base as a whole has a many examples that violate this style guide, unfortunately:
E.g., https://github.com/tensorflow/tensorboard/search?q=%22def+Add%22&unscoped_q=%22def+Add%22

So I'll leave it as is for now.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Whoops. Added it.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Ack.

PR #3552, opened 2020-04-27
Partially revert "improve rerender of vz_line_chart (#3524)"

wchargin:
Summary:
This reverts commit 27d0023e728aba79d0129f890e41c6c2c4c436b9, then
reinstates stubs for the new method definitions due to Google-internal
code that expects them to exist.

Stopgap for #3551 pending real fix.

Test Plan:
Opening the PR curves dashboard and sliding one of the step sliders now
correctly renders the PR curve instead of showing just the final datum.

wchargin-branch: revert-3524

PR #3571, opened 2020-04-30
Implement commitChanges on vz-line-chart2

stephanwlee:
This is a roll forward of 46b6e615a (#3552). The commit partially rolls back
27d0023e7 (#3524) which introduces `commitChanges` which was supposed to be
invoked after setting data, visible series, or metadata.

This change addresses the regression caused by 27d0023e7 which forgot to
call the new API in all places where we use the vz-line-chart2.


stephanwlee:
I did few passes of `git grep`s:
1. `<vz-line-chart`
2. `setSeriesData|setSeriesMetadata|setVisibleSeries`

and manually inspected if they had appropriate `commitChanges`.

wchargin:
Sounds good; thanks!

PR #3575, opened 2020-05-01
Allow server to specify maximum tensor upload size

bmd3k:

Backends should be able to impose size limits for uploading individual tensor points, similar to how they can impose size limits for individual blobs.

Fallback to a reasonable default value for the time being while we wait for backends to be updated to provide this information.

Maximum size of tensor points is specified in the handshake response with the frontend. If the value is not present then fallback to some reasonable default. 

If a Tensor point larger than the maximum size is encountered, strip it from write requests and log a warning.

* Screenshots of UI changes

image

Wrote tests.

Ran uploader with a frontend that does not specify max limit and observed that default value is used.

Ran uploader with a frontend that does specify max limit and observed that specified value is used.

### DIFF
--- a/tensorboard/uploader/proto/server_info.proto
+++ b/tensorboard/uploader/proto/server_info.proto
@@ -109,4 +109,8 @@ message UploadLimits {
   // the server may enforce a *minimum* blob size, which is a separate issue).
   // If this is negative, no blobs should be uploaded.
   int64 max_blob_size = 1;
+  // The maximum allowed size for tensor point uploads.
+  // If this is 0, it is allowed to upload tensor points of size 0.
+  // If this is negative, no blobs should be uploaded.
+  int64 max_tensor_point_size = 2;
 }
--- a/tensorboard/uploader/server_info_test.py
+++ b/tensorboard/uploader/server_info_test.py
@@ -270,6 +270,7 @@ class MaxTensorPointSizeTest(tb_test.TestCase):
         actual = server_info.max_tensor_point_size(info)
         self.assertEqual(actual, 0)
 
+
 def _localhost():
     """Gets family and nodename for a loopback address."""
     s = socket
--- a/tensorboard/uploader/server_info_test.py
+++ b/tensorboard/uploader/server_info_test.py
@@ -262,10 +262,12 @@ class MaxTensorPointSizeTest(tb_test.TestCase):
         actual = server_info.max_tensor_point_size(info)
         self.assertEqual(actual, 42)
 
-    def test_upload_limits_provided_without_max_blob_size(self):
+    def test_upload_limits_provided_without_max_tensor_point_size(self):
         # This just shows that the proto3 default value of 0 is reported as
         # usual, not handled as a special case.
         info = server_info_pb2.ServerInfoResponse()
+        # Ensure upload_limits is set but do not explicitly set
+        # max_tensor_point_size.
         info.upload_limits.max_blob_size = 42
         actual = server_info.max_tensor_point_size(info)
         self.assertEqual(actual, 0)
--- a/tensorboard/uploader/server_info_test.py
+++ b/tensorboard/uploader/server_info_test.py
@@ -248,6 +248,28 @@ class MaxBlobSizeTest(tb_test.TestCase):
         self.assertEqual(actual, 0)
 
 
+class MaxTensorPointSizeTest(tb_test.TestCase):
+    """Tests for `max_tensor_point_size`."""
+
+    def test_old_server_no_upload_limits(self):
+        info = server_info_pb2.ServerInfoResponse()
+        actual = server_info.max_tensor_point_size(info)
+        self.assertEqual(actual, server_info._DEFAULT_MAX_TENSOR_POINT_SIZE)
+
+    def test_upload_limits_provided_with_max_tensor_point_size(self):
+        info = server_info_pb2.ServerInfoResponse()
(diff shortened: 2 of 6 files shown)


tensorboard/uploader/proto/server_info.proto
```diff
@@ -109,4 +109,8 @@ message UploadLimits {
   // the server may enforce a *minimum* blob size, which is a separate issue).
   // If this is negative, no blobs should be uploaded.
   int64 max_blob_size = 1;
+  // The maximum allowed size for tensor point uploads.
+  // If this is 0, it is allowed to upload tensor points of size 0.
+  // If this is negative, no blobs should be uploaded.
+  int64 max_tensor_point_size = 2;
```
bmd3k, line 115: Question about terminology: I try to use "tensor point" consistently but is there a better term?

tensorboard/uploader/server_info_test.py
```diff
@@ -248,6 +248,29 @@ def test_upload_limits_provided_without_max_blob_size(self):
...
+        actual = server_info.max_tensor_point_size(info)
+        self.assertEqual(actual, server_info._DEFAULT_MAX_TENSOR_POINT_SIZE)
+
+    def test_upload_limits_provided_with_max_tensor_point_size(self):
+        info = server_info_pb2.ServerInfoResponse()
+        info.upload_limits.max_tensor_point_size = 42
+        actual = server_info.max_tensor_point_size(info)
+        self.assertEqual(actual, 42)
+
+    def test_upload_limits_provided_without_max_blob_size(self):
```
davidsoergel, line 265: tensor_point

tensorboard/uploader/server_info_test.py
```diff
@@ -248,6 +248,29 @@ def test_upload_limits_provided_without_max_blob_size(self):
...
+        # This just shows that the proto3 default value of 0 is reported as
+        # usual, not handled as a special case.
+        info = server_info_pb2.ServerInfoResponse()
+        info.upload_limits.max_blob_size = 42
```
davidsoergel, line 269: Maybe add a comment like `# Ensure that upload_limits is set`

tensorboard/uploader/uploader_test.py
```diff
@@ -1411,6 +1424,57 @@ def test_break_at_tensor_point_boundary(self):
...
+        )
+        event_2.summary.value.add(
+            tag="two",
+            tensor=tensor_pb2.TensorProto(
+                tensor_content=b"\x01\x02\x03\x04\x05\x06\x07"
+            ),
+        )
+
+        run_proto = self._add_events_and_flush(
+            _apply_compat([event_1, event_2]), max_tensor_point_size=7 + 1
```
davidsoergel, line 1451: Why is this "7 + 1"?  Also, since the tensor_content at line 1446 is 7 bytes, one might expect that to pass the filter because 7 < 8.  Presumably the proto wire format overhead is the reason?  Anyway, please add some comments about what is going on here.

tensorboard/uploader/uploader_test.py
```diff
@@ -1411,6 +1424,57 @@ def test_break_at_tensor_point_boundary(self):
...
+    def test_strip_large_tensors(self):
+        # Generate test data with varying tensor point sizes. Use raw bytes.
+        event_1 = event_pb2.Event(step=1)
+        event_1.summary.value.add(
+            tag="one", tensor=tensor_pb2.TensorProto(tensor_content=b"\x01\x02")
+        )
+        event_1.summary.value.add(
+            tag="two",
+            tensor=tensor_pb2.TensorProto(
+                tensor_content=b"\x01\x02\x03\x04\x05\x06"
```
davidsoergel, line 1436: Maybe add a comment like "we'll filter this out in the second half of the test below"

tensorboard/uploader/uploader_test.py
```diff
@@ -1411,6 +1424,57 @@ def test_break_at_tensor_point_boundary(self):
...
+            ),
+        )
+        event_2 = event_pb2.Event(step=2)
+        event_2.summary.value.add(
+            tag="one", tensor=tensor_pb2.TensorProto(tensor_content=b"\x01\x02")
+        )
+        event_2.summary.value.add(
+            tag="two",
+            tensor=tensor_pb2.TensorProto(
+                tensor_content=b"\x01\x02\x03\x04\x05\x06\x07"
```
davidsoergel, line 1446: Similarly "We'll filter this out in both halves of the test below"

tensorboard/uploader/proto/server_info.proto
davidsoergel: "tensor point" sgtm.

tensorboard/uploader/server_info_test.py
bmd3k: Done.

tensorboard/uploader/uploader_test.py
bmd3k: Done.

tensorboard/uploader/uploader_test.py
bmd3k: Done.

tensorboard/uploader/server_info_test.py
bmd3k: Done.

tensorboard/uploader/uploader_test.py
bmd3k: You're intuition is correct. I've tried to improve the documentation.

PR #3576, opened 2020-05-01
Remove some non-critical uses of TensorFlow

wchargin:
Summary:
Most uses of `tf.compat.as_bytes` and `tf.compat.as_text` can be
replaced by direct equivalents, which is simpler because we only support
Python 3. These could instead depend on our `tensorflow_stub` module,
but generally there’s enough contextual type information (e.g., a given
proto field is known to be either `bytes` or `string`) that inlining the
appropriate branch seems simpler. Some uses are just dead code, anyway
(`_get_value`/`_get_values` helpers), and can be removed entirely.

Test Plan:
All existing tests pass, and a sanity check of affected dashboards
checks out. The affected targets have no remaining TensorFlow dep:

```
bazel query "somepath(
    same_pkg_direct_rdeps(
        set($(git show --format= --numstat | awk '{ print $3 }'))),
    //tensorboard/compat:tensorflow
)"
```

wchargin-branch: trim-noncritical-tf-20200501


tensorboard/backend/event_processing/event_accumulator.py
```diff
@@ -81,17 +82,15 @@
...
-COMPRESSED_HISTOGRAMS = "distributions"
-HISTOGRAMS = "histograms"
-IMAGES = "images"
-AUDIO = "audio"
-SCALARS = "scalars"
-TENSORS = "tensors"
-GRAPH = "graph"
-META_GRAPH = "meta_graph"
-RUN_METADATA = "run_metadata"
+COMPRESSED_HISTOGRAMS = tag_types.COMPRESSED_HISTOGRAMS
```
nfelt, line 85: Maybe add a comment like "Legacy aliases" or something?  Just to indicate why we're copying them all.

tensorboard/backend/event_processing/plugin_event_accumulator.py
```diff
@@ -42,12 +43,10 @@
 
 TensorEvent = namedtuple("TensorEvent", ["wall_time", "step", "tensor_proto"])
 
-## The tagTypes below are just arbitrary strings chosen to pass the type
-## information of the tag from the backend to the frontend
-TENSORS = "tensors"
-GRAPH = "graph"
-META_GRAPH = "meta_graph"
-RUN_METADATA = "run_metadata"
+TENSORS = tag_types.TENSORS
```
nfelt, line 46: Ditto here.

tensorboard/backend/event_processing/event_accumulator.py
wchargin: Done.

tensorboard/backend/event_processing/plugin_event_accumulator.py
wchargin: Done.

PR #3578, opened 2020-05-01
Factor tag type constants out of event accumulator

wchargin:
Summary:
Some code depends on `event_accumulator` only for its string constants,
which incurs an unnecessary transitive dependency on TensorFlow. We
factor out those constants into a dependency-free module.

Test Plan:
Straightforward refactor; unit tests suffice.

wchargin-branch: extract-tag-types


tensorboard/backend/event_processing/BUILD
```diff
@@ -179,6 +179,12 @@ py_test(
     ],
 )
 
+py_library(
+    name = "tag_types",
+    srcs = ["tag_types.py"],
+    srcs_version = "PY2AND3",
```
wchargin, line 185: (due to Google-internal transitive reverse dependencies)

PR #3582, opened 2020-05-05
 vulc: fix mangled DOM due to a bug in the printer

### LINKED ISSUE #3557: Graph info panel is broken
Symptom: after the #3476 refactor, the graph info panel is broken.

Cause:
After the change, the iron-list dom-module is missing from the document. When inspecting the elements more carefully, it appears like following:

```html
    <div id="numeric-alerts-table-container">
          <table id="numeric-alerts-table">
            <thead>
              <tr>
                <th>First Offense</th>
                <th>Tensor (Device)</th>
                <th id="event-counts-th">Event Counts</th>
              </tr>
            </thead>
            <tbody id="numeric-alerts-body"></tbody>
          </table>
        </div>
      </div>
      <template is="dom-if" if="[[!_hasDebuggerNumericAlerts(debuggerNumericAlerts)]]">
        <p class="no-numeric-alerts-notification">
          No numeric alerts so far. That is likely good. Alerts indicate the
          presence of NaN or (+/-) Infinity values, which may be concerning.

<dom-module id="iron-list">
  <template>
    <style>
```

As you can notice, the DOM is quite broken. There are multiple closing tags from tf-graph-debugger-data-card are missing. 

Upon more careful debugging, I have found out that the Vulcanization is actually responsible for such structure.

For instance, `<p>foo</p>` gets compiled into `<p>foo`. Similarly, 
```html
<template>
  <p>foo</p>
</template>
<dom-module id="iron-list">
</dom-module

<!-- gets compiled to -->
<template>
  <p>foo
</template>

<dom-module id="iron-list">
</dom-module
```

When debugged further, I found that there are some heuristic on HTML5Printer of jsoup that we use which omits closing tag of certain tags when they are the last one in current level <https://github.com/bazelbuild/rules_closure/blob/d1c0bb4862882e0444ee478aa2091d41a4df084e/java/org/jsoup/nodes/Html5Printer.java#L209-L214>.

I think the intention was good but if you parse then print below code using jsoup, you get quite wrong code.
```html
<div>
  <p>foo
</div>

<!-- converts to -->
<div>
  <p>foo
```

There are three ways to move forward:
1. fix jsoup not to wrongfully remove closing tag of parent node
2. never do multiple passes of jsoup parse and print
3. use genrule to inline polymer_lib into tensorboard.html since inlining is all we need.

cc @bmd3k

stephanwlee:
The issue is better described in #3557. Please refer to it.

Our HTML module traversal relies on the `document` so we cannot create
the document without nested document in one go. Instead, we do a post
processing of the document to prune away all the `document` so the
default jsoup's printer can print them without awkward `<#root>`
elements.

index.html size difference:
```
555323 before
554749 after
```

Fixes #3557.

### DIFF
--- a/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
+++ b/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
@@ -187,7 +187,6 @@ public final class Vulcanize {
     createFile(
         jsOutput, shouldExtractJs ? extractAndTransformJavaScript(document, jsPath) : "");
     Document normalizedDocument = getFlattenedHTML5Document(document);
-    normalizedDocument.outputSettings().syntax(Document.OutputSettings.Syntax.html);
     createFile(output, normalizedDocument.toString());
   }
 
--- a/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
+++ b/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
@@ -75,7 +75,6 @@ import org.jsoup.nodes.Comment;
 import org.jsoup.nodes.DataNode;
 import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Element;
-import org.jsoup.nodes.Html5Printer;
 import org.jsoup.nodes.Node;
 import org.jsoup.nodes.TextNode;
 import org.jsoup.parser.Parser;
@@ -182,10 +181,11 @@ public final class Vulcanize {
     }
 
     boolean shouldExtractJs = !jsPath.isEmpty();
+    // Write an empty file for shasum when all scripts are extracted out.
     createFile(
         jsOutput, shouldExtractJs ? extractAndTransformJavaScript(document, jsPath) : "");
-    // Write an empty file for shasum when all scripts are extracted out.
-    createFile(output, Html5Printer.stringify(document));
+    Document normalizedDocument = getFlattenedDocument(document);
+    createFile(output, normalizedDocument.toString());
   }
 
   private static void createFile(Path filePath, String content) throws IOException {
@@ -776,6 +776,55 @@ public final class Vulcanize {
     lastBody.appendChild(scriptElement);
 
     return scriptContent;
+  }
+
+  private static void cloneChildrenWithoutWhitespace(Element src, Element dest) {
+    List<Node> toMove = new ArrayList<Node>();
+    for (Node node : src.childNodes()) {
+      if (node instanceof TextNode && ((TextNode) node).isBlank()) {
+        continue;
+      }
+      toMove.add(node);
+    }
+    for (int i = 0; i < toMove.size(); i++) {
+      Node node = toMove.get(i);
+      dest.appendChild(node.clone());
+    }
+  }
+
+  // Refer to https://github.com/tensorflow/tensorboard/issues/3557.
+  private static Document getFlattenedDocument(Document document) {
+    Document flatDoc = new Document("/");
+    flatDoc.normalise();
+    Element rootDocumentHead = flatDoc.head();
+    Element rootDocumentBody = flatDoc.body();
+
+    Node currentNode = document;
+    while (currentNode != null) {
+      // Do not clone the element if it is a `document` inside `<head>`.
+      // We want to traverse further and get all elements from `<head>` and `<body>`
+      if (currentNode.parentNode() != null && currentNode.parentNode().nodeName().equals("head")
+          && !(currentNode instanceof Document)) {
+        rootDocumentHead.appendChild(currentNode.clone());
+      }
+
+      if (currentNode.nodeName().equals("body")) {
+        cloneChildrenWithoutWhitespace((Element) currentNode, rootDocumentBody);
+      }
+
+      if (currentNode.childNodeSize() > 0) {
+        currentNode = currentNode.childNode(0);
+      } else {
+        while (currentNode != null && currentNode.nextSibling() == null) {
+          currentNode = currentNode.parentNode();
+        }
+
+        if (currentNode != null) {
+          currentNode = currentNode.nextSibling();
+        }
+      }
+    }
+    return flatDoc;
   }
 
   private static final class JsPrintlessErrorManager extends BasicErrorManager {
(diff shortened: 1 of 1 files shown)


tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -778,6 +778,55 @@ private static String extractAndTransformJavaScript(Document document, Webpath j
...
+
+      if (currentNode.nodeName().equals("body")) {
+        cloneChildrenWithoutWhitespace((Element) currentNode, rootDocumentBody);
+      }
+
+      if (currentNode.childNodeSize() > 0) {
+        currentNode = currentNode.childNode(0);
+      } else {
+        while (currentNode != null && currentNode.nextSibling() == null) {
+          currentNode = currentNode.parentNode();
```
bmd3k, line 819: Is this the only place where currentNode might be set to null?  That would be worth mentioning to help understand when we can exit the top-level loop.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -182,10 +181,11 @@ public static void main(String[] args)
     }
 
     boolean shouldExtractJs = !jsPath.isEmpty();
+    // Write an empty file for shasum when all scripts are extracted out.
     createFile(
         jsOutput, shouldExtractJs ? extractAndTransformJavaScript(document, jsPath) : "");
-    // Write an empty file for shasum when all scripts are extracted out.
-    createFile(output, Html5Printer.stringify(document));
+    Document normalizedDocument = getFlattenedDocument(document);
+    createFile(output, normalizedDocument.toString());
```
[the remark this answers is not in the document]
stephanwlee: The patch exists in rules_closure and the change is unnecessary to already near tabled project.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -778,6 +778,79 @@ private static String extractAndTransformJavaScript(Document document, Webpath j
...
+      }
+      toMove.add(node);
+    }
+    for (int i = 0; i < toMove.size(); i++) {
+      Node node = toMove.get(i);
+      dest.appendChild(node.clone());
+    }
+  }
+
+  // When we inline the HTML based on `<link rel="import">` in `transform`, we
```
bmd3k, line 795: Thanks, this is very useful. Preferable to use JavaDoc format, though.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -778,6 +778,79 @@ private static String extractAndTransformJavaScript(Document document, Webpath j
...
+  // </html>
+  private static Document getFlattenedDocument(Document document) {
+    Document flatDoc = new Document("/");
+    flatDoc.normalise();
+    Element rootDocumentHead = flatDoc.head();
+    Element rootDocumentBody = flatDoc.body();
+
+    Node currentNode = document;
+    while (currentNode != null) {
+      // Do not clone the element if it is a `document` inside `<head>`.
```
bmd3k, line 825: Sorry, I feel like all my comments are just about documentation. (Which I guess is good - the code itself seems fine once you understand what it's trying to do).

This comment is mostly about what the next bit of code *does not do* but it probably would be clearer if rewritten so that it first explains what it *does do*.  It took me a while to figure it out.

Namely:

Copy most children of \<head\> elements into the flat doc without modification. The exception is if the node is of type `document` (or a <#root> element) in which case we want to traverse further and only copy the elements in its `body` and `head` elements.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -778,6 +778,79 @@ private static String extractAndTransformJavaScript(Document document, Webpath j
...
+      // We want to traverse further and get all elements from `<head>` and `<body>`
+      if (currentNode.parentNode() != null && currentNode.parentNode().nodeName().equals("head")
+          && !(currentNode instanceof Document)) {
+        rootDocumentHead.appendChild(currentNode.clone());
+      }
+
+      if (currentNode.nodeName().equals("body")) {
```
bmd3k, line 832: Is it possible to stop traversing deeper into body elements or do we have body elements nested inside body elements?  If the latter (nested body elements) then that's probably worth mentioning in the doc.

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -778,6 +778,79 @@ private static String extractAndTransformJavaScript(Document document, Webpath j
...
+  // documents and jsoup's Node.outerHtml (or Node.toString) are incapable of
+  // properly outputting that. Here, we flatten the document by combining all
+  // elements in `<head>` and `<body>` of nested document in one `<head>` and
+  // `<body>`.
+  //
+  // Examples:
+  // // Input
+  // <#root> <!-- document -->
+  //   <html>
+  //     <head><#root><html><body>foo</body></html></#root></head>
```
bmd3k, line 806: I think it would be worth expanding the example (or add another) to include a nested `head` and nested `body` element, if those are actual things we have to worry about (I think they are).

tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -778,6 +778,79 @@ private static String extractAndTransformJavaScript(Document document, Webpath j
...
+
+      if (currentNode.nodeName().equals("body")) {
+        cloneChildrenWithoutWhitespace((Element) currentNode, rootDocumentBody);
+      }
+
+      // Standard DFS.
+      if (currentNode.childNodeSize() > 0) {
+        currentNode = currentNode.childNode(0);
+      } else {
+        // Go up in the tree if there are no sibling elemnts and repeat.
```
bmd3k, line 840: elemnts => elements

stephanwlee:
I've made a considerable change to the PR. PTAL again. Sorry!

PR #3584, opened 2020-05-05
ng: set active plugin to first enabled plugin

stephanwlee:
Previously, we were wrongly setting it to the first plugin in the
object.

Now, we set it to the first active plugin. If nothing is active, set
it to `null`.

### DIFF
--- a/tensorboard/webapp/core/store/core_reducers.ts
+++ b/tensorboard/webapp/core/store/core_reducers.ts
@@ -57,11 +57,11 @@ const reducer = createReducer(
   on(
     actions.pluginsListingLoaded,
     (state: CoreState, {plugins}): CoreState => {
-      const firstActivePlugin =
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/core/store/core_reducers.ts
```diff
@@ -57,9 +57,11 @@ const reducer = createReducer(
   on(
     actions.pluginsListingLoaded,
     (state: CoreState, {plugins}): CoreState => {
-      const [firstPlugin] = Object.keys(plugins);
-      let activePlugin =
-        state.activePlugin !== null ? state.activePlugin : firstPlugin;
+      const firstActivePlugin =
```
caisq, line 60: Can we rename `firstActivePlugin` to `firstEnabledPlugin` to be clearer?

PR #3586, opened 2020-05-05
uploader: add num_tensors to Experiment proto

caisq:
The intent is to support listing the number of tensors in each experiment in the output of `tensorboard dev list`.

### DIFF
--- a/tensorboard/uploader/proto/experiment.proto
+++ b/tensorboard/uploader/proto/experiment.proto
@@ -33,9 +33,10 @@ message Experiment {
   // experiment, across all time series, including estimated overhead.
   // Output-only.
   int64 total_blob_bytes = 10;
-  // The number of tensors in this experiment, across all time series.
+  // The number of bytes used for storage of the contents of tensors in this
+  // experiment, across all time series, including estimated overhead.
   // Output-only.
-  int64 num_tensors = 11;
+  int64 total_tensor_bytes = 11;
 }
 
 // Field mask for `Experiment` used in get and update RPCs. The `experiment_id`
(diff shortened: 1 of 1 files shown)


tensorboard/uploader/proto/experiment.proto
```diff
@@ -33,6 +33,9 @@ message Experiment {
   // experiment, across all time series, including estimated overhead.
   // Output-only.
   int64 total_blob_bytes = 10;
+  // The number of tensors in this experiment, across all time series.
+  // Output-only.
+  int64 num_tensors = 11;
```
caisq, line 38: Question for reviewers: Do you think in addition to `num_tensors`, we should add `total_tensor_bytes`? How expensive is it to calculate `total_tensor_bytes` on the server side for each experiment?

ericdnielsen: I think if we only show one metric, it should be total_tensor_bytes.  Quota error messages refer to the number of bytes, not the number of tensors, so if people are looking for experiments they can delete to free up quota, the bytes is the usueful metric.  If we show both numbers that might be ok, but I'm not sure how a user would use the raw count.

Total tensor bytes are cached at both user and experiment level, so it should be a trivial query, but I don't recall offhand if its included in any existing query.  It might be needed to be added as a new query or to the metadata query.

caisq: @ericdnielsen What you wrote makes sense. I've changed `num_tensors` to `total_tensor_bytes` in this PR. PTAL.

caisq:
Closing this PR. `total_tensor_bytes` is already defined. I'll open another PR to hook up the logic for displaying # of tensor bytes.

PR #3587, opened 2020-05-05
ng: no active plugin warning page

stephanwlee:
Angular version of the TensorBoard never implemented the page.
The copy and UX is almost exactly that of Polymer's (misses data
location; angular version currently does not fetch environments)

Requires #3584 to see the change.

### DIFF
--- a/tensorboard/webapp/plugins/plugins_component.ng.html
+++ b/tensorboard/webapp/plugins/plugins_component.ng.html
@@ -49,7 +49,7 @@ limitations under the License.
     </p>
 
     <p>
-      <i>Last reload: {{lastUpdated | date}}</i>
+      <i>Last reload: {{lastUpdated | date: 'medium'}}</i>
     </p>
   </div>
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/plugins/plugins_component.ng.html
```diff
@@ -21,3 +21,35 @@
...
+      If you think TensorBoard is configured properly, please see
+      <a
+        href="https://github.com/tensorflow/tensorboard/blob/master/README.md#my-tensorboard-isnt-showing-any-data-whats-wrong"
+        >the section of the README devoted to missing data problems</a
+      >
+      and consider filing an issue on GitHub.
+    </p>
+
+    <p>
+      <i>Last reload: {{lastUpdated | date}}</i>
```
psybuzz, line 52: nit: perhaps `lastUpdated | date: 'medium'`, to get HH:MM:SS PM ?

PR #3589, opened 2020-05-05
ng: prevent double dashboard reload on mount 

stephanwlee:
Cause: dashboards, on Polymer's `ready` invoke `reload`. That means,
when the Angular's plugin_component mount the Polymer dashboard onto the
DOM, it will fetch the `data` right away. In addition to that, on
plugin_container mount, we fetch the `plugins_listing` and upon its
completion, invoke `reload` again.

To be clear, the fact that `reload` was getting triggered based on last
fetch time of `plugins_listing` was not ideal; it does imitate the way
Polymer based tf-tensorboard does (Promise.then) and it does tie with
the reloader but using the last loaded time as a proxy to user's action
(e.g., reload button click) is not correct. Ideally, we should only make
the data requests from `effects` but our Polymer code is ill-suited for
that.

Addressed the issue by ignoring first two triggers (initial values and the
first `pluginsListing` load) of the `pluginsListingLoaded`

### DIFF
--- a/tensorboard/webapp/plugins/plugins_component.ts
+++ b/tensorboard/webapp/plugins/plugins_component.ts
@@ -88,6 +88,9 @@ export class PluginsComponent implements OnChanges {
   @Input()
   lastUpdated?: number;
 
+  @Input()
+  reloadId!: number;
+
   readonly LoadingMechanismType = LoadingMechanismType;
 
   private readonly pluginInstances = new Map<string, HTMLElement>();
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/plugins/plugins_component.ts
```diff
@@ -96,7 +99,9 @@ export class PluginsComponent implements OnChanges {
     if (change['activePlugin'] && this.activePlugin) {
       this.renderPlugin(this.activePlugin!);
     }
-    if (change['lastUpdated']) {
+
+    if (change['reloadId'] && !change['reloadId'].firstChange) {
+      console.log('reload called');
```
wchargin, line 104: Remove?

stephanwlee:
> Just one possible approach; what do you think?

Made the change with `reload-on-ready`. If overall direction is okay, I intend to raise a PR with only the Polymer changes. PTAL.

stephanwlee:
> LGTM otherwise.

Got it. Will create separate PR with the `reload-on-ready` only.

PR #3594, opened 2020-05-06
[DebuggerV2] A few improvements to the /graphs/op_info route

caisq:
* List of improvements
  1. Add missing document string and unit tests for the `output_tensor_ids` field of the /graphs/op_info response.
  2. Handle actual cases discovered during prototyping in which certain special op types are not captured by tfdbg2's op callbacks and hence cannot be retrieved during op-input look up (e.g., a input op of the type `StatefulPartitionedCall`). This improves the robustness of this route overall.
  3. Change the `consumers` field from a dict mapping output slot index to consumer ops to an array, in which each element corresponds to an output slot of the op. It now has the schema
      `Array<Array<{op_name: string, input_slot: number: data?: GraphOpInfo}>>`
  4. Consolidate `input_names` and `inputs` to a single `inputs` field with the schema
      `Array<{op_name: string, output_slot: number: data?: GraphOpInfo}>`


caisq:
@davidsoergel As discussed offline, we consolidated 
- `input_names` into `inputs`
- `consumer_names_and_slots` into `consumers`

Doc strings and tests are updated accordingly. The PR description is also updated.

tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
```diff
@@ -254,19 +254,44 @@ def serve_graph_op_info(self, request):
...
+            symbolic tensors output by this op.
+            Each element of the array is an array of which the length equals
+            the number of downstream ops that consume the corresponding symbolic
+            tensor (only data edges are tracked).
+            Each element of the array is an object with the following fields:
+              - op_name: Name of the op receives the output tensor as an input.
+              - input_slot: 0-based input slot index at which the downstream
+                op receives this output tensor.
+              - data: A recursive data structure of this very schema.
+                In the rare case wherein the data for a consumer op cannot be
```
davidsoergel, line 289: also not rare

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
```diff
@@ -1091,62 +1092,77 @@ def testServeGraphOpInfoForOpWithInputsAndConsumers(self):
...
+        input0 = inputs[0]["data"]
+        input1 = inputs[1]["data"]
+        for inpt in (input0, input1):
+            self.assertEqual(inpt["op_type"], "Unpack")
+            self.assertNotIn("input_names", inpt)
+            self.assertEqual(inpt["num_outputs"], 4)
+            self.assertLen(inpt["output_tensor_ids"], 4)
+            self.assertEqual(inpt["host_name"], _HOST_NAME)
+            self.assertEqual(inpt["graph_ids"], data["graph_ids"])
+            self.assertLen(inpt["inputs"], 1)
```
davidsoergel, line 1119: Can you construct the entire expected input and do a deep compare all at once?  This sequence of individual comparisons is very hard to follow.  Deep compare could be done with `assertDictEqual()` or `repr()` if nothing else.

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -486,34 +492,66 @@ def GraphOpInfo(self, run, graph_id, op_name):
...
         del json_object["graph_id"]
         json_object["graph_ids"] = self._getGraphStackIds(
             op_creation_digest.graph_id
         )
         # TODO(cais): "num_outputs" should be populated in to_json() instead.
         json_object["num_outputs"] = op_creation_digest.num_outputs
+        del json_object["input_names"]
+
+        json_object["inputs"] = []
+        if op_creation_digest.input_names:
```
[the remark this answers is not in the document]
caisq: Done.

tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
```diff
@@ -254,19 +254,44 @@ def serve_graph_op_info(self, request):
...
+          - inputs: Specifications of all inputs to this op.
+            Currently only immediate (one level of) inputs are provided.
+            This is an array of length N_in, where N_in is the number of
+            data inputs received by the op. Each element of the array is an
+            object with the following fields:
+              - op_name: Name of the op that provides the input tensor.
+              - output_slot: 0-based output slot index from which the input
+                tensor emits.
+              - data: A recursive data structure of this same schema.
+                In the rare case wherein the data for an input cannot be
```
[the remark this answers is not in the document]
caisq: OK. There are two different cases
1. leaf nodes not having this field populated by design.
2. non-leaf nodes not having this populated due to op lookup failure.

I clarified this in the doc string here.

tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
```diff
@@ -254,19 +254,44 @@ def serve_graph_op_info(self, request):
...
+            This is an empty list for an op with no inputs.
+          - consumers: Specifications for all the downstream consuming ops of
+            this. Currently only immediate (one level of) consumers are provided.
+            This is an array of length N_out, where N_out is the number of
+            symbolic tensors output by this op.
+            Each element of the array is an array of which the length equals
+            the number of downstream ops that consume the corresponding symbolic
+            tensor (only data edges are tracked).
+            Each element of the array is an object with the following fields:
+              - op_name: Name of the op receives the output tensor as an input.
```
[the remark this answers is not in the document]
caisq: Done.

tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
caisq: Ditto. This is the same case as above.

tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
caisq: The difficulty in doing that is that some fields (like the detailed values of op_names) do not have concrete contracts. TensorFlow may change its internal op-naming algorithm any time and we'll be broken if we assert on a specific string name. In the TF2 world, op names are only meant for debugging. Examples like Lines 1120-1121 are written that way for exactly that reason.

I looked but didn't find a "wildcard" token supported by deep `assertEqual` - maybe a feature request for PyUnit.

PR #3599, opened 2020-05-07
Cache bleach.Cleaner for faster conversion of markdown to safe html

bmd3k:

Calls to tensorboard.dev's /data/plugin/scalars/tags endpoint can be surprisingly slow if there are many tags. 

The following call spends ~3.5 seconds in the code that converts tag description markdown into safe html, even though the description tags are empty:

http://tensorboard.dev/experiment/UlN9loBoTwGczi5FlWbYEQ/data/plugin/scalars/tags

It is the call to `bleach.clean()` that takes up almost all of this time because it is always reinstantiating `bleach.Cleaner`.

Similar to recent changes that cache `Markdown` instances per thread, we now cache `bleach.Cleaner` instances per thread.

### DIFF
--- a/tensorboard/plugin_util.py
+++ b/tensorboard/plugin_util.py
@@ -20,7 +20,7 @@ from __future__ import print_function
 
 import threading
 
-import bleach
+from bleach.sanitizer import Cleaner
 
 # pylint: disable=g-bad-import-order
 # Google-only: import markdown_freewisdom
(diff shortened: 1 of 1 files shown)


bmd3k:
> If empty descriptions were part of the problem here, maybe it's worth just adding an early exit to `markdown_to_safe_html` that skips the markdown conversion and bleach clean entirely and returns the empty string? That does seem like a very common case for descriptions that's worth optimizing for, since then we could potentially avoid ever initializing these objects in most request threads.

I was considering that as a next step. The change as is should considerably improve performance both for empty descriptions and non-empty descriptions. In my limited testing the early-exit-on-empty-markdown approach gains a little more performance improvement but not much.

PR #3600, opened 2020-05-07
[DebuggerV2] Add backend route /graphs/graph_info

caisq:
  * Continue developing the backend of DebuggerV2.
  * Add HTTP route `/graphs/graph`. This is at the same level as `/graphs/op`. But instead of serving information about individual ops, the new route serves information about the graph (its ID, name, relation to other graphs, etc.)
  * Use `DebuggedGraph.to_json()` to implement the serving of graph info.
  * Unit tests added.

### DIFF
--- a/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
+++ b/tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
@@ -453,6 +453,33 @@ class DebuggerV2EventMultiplexer(object):
             ],
         }
 
+    def GraphInfo(self, run, graph_id):
+        """Get the information regarding a TensorFlow graph.
+
+        Args:
+          run: Name of the run.
+          graph_id: Debugger-generated ID of the graph in question.
+            This information is available in the return values
+            of `GraphOpInfo`, `GraphExecution`, etc.
+
+        Returns:
+          A JSON-serializable object containing the information regarding
+            the TensorFlow graph.
+
+        Raises:
+          NotFoundError if the graph_id is not known to the debugger.
+        """
+        runs = self.Runs()
+        if run not in runs:
+            return None
+        try:
+            graph = self._reader.graph_by_id(graph_id)
+        except KeyError:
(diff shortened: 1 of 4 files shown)


caisq:
cc @davidsoergel

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
```diff
@@ -453,6 +453,33 @@ def GraphExecutionData(self, run, begin, end, trace_id=None):
...
+            the TensorFlow graph.
+
+        Raises:
+          NotFoundError if the graph_id is not known to the debugger.
+        """
+        runs = self.Runs()
+        if run not in runs:
+            return None
+        try:
+            graph = self._reader.graph_by_id(graph_id)
```
wchargin, line 476: Why is `run` not actually used in the query? Is it needed?

tensorboard/plugins/debugger_v2/debugger_v2_plugin.py
```diff
@@ -243,6 +244,50 @@ def serve_graph_execution_data(self, request):
...
 
+    @wrappers.Request.application
+    def serve_graph_info(self, request):
+        """Serve basic information about a TensorFlow graph.
+
+        The request specifies the debugger-generated ID of the graph being
+        queried.
+
+        The response contains a JSON object with the following fields:
+          - graph_id: The debugger-geneated ID (echoing the request).
```
wchargin, line 255: sp.: *generated*

tensorboard/plugins/debugger_v2/debug_data_multiplexer.py
caisq: Please recall that we currently support only up to one tfdbg2 run in each logdir, as read by `self._reader`. It is identified with the hardcoded run name `__default_debugger_run__`. You can see other public methods of this class do similar checks of `run` just to ensure it exists in the beginning, without using it for actual data look up subsequently.

PR #3605, opened 2020-05-08
DataFrame: Flatten MultiIndex; Set include_wall_time=False

caisq:
  * Improve DataFrame API based on the following user feedback
    * The MultiIndex DataFrame is hard to handle and not nice in terms of CSV round-trip compatibility
    * The wall_time column is usually not useful
  * When pivoting the DataFrame using `pivot_table()`, follow it with `reset_index()` to flatten the 2-level MultiIndex to a single level of index.
  * Then, call `.columns.name` to remove the `tag` columns name, which is an artifact of the tensorboard data model and not directly relevant to users.
  * Add `include_wall_time=False` kwarg to get_scalars()
  * Change the default value of `pivot` from `None` to `False` as per more standard and expected default value scheme.
* Screenshots of UI changes
  * From test_binary with default settings:
    * Before: image
   * After: image
  * New sub tests added

### DIFF
--- a/tensorboard/data/experimental/base_experiment.py
+++ b/tensorboard/data/experimental/base_experiment.py
@@ -28,7 +28,13 @@ class BaseExperiment(metaclass=abc.ABCMeta):
     # TODO(cais): Add list_scalar_tags().
 
     @abc.abstractmethod
-    def get_scalars(self, runs_filter=None, tags_filter=None, pivot=None):
+    def get_scalars(
+        self,
+        runs_filter=None,
+        tags_filter=None,
+        pivot=None,
+        include_wall_time=None,
+    ):
         """Export scalar data as a pandas.DataFrame.
 
         Args:
@@ -40,6 +46,9 @@ class BaseExperiment(metaclass=abc.ABCMeta):
             `pivot_data()` method to a “wide” format wherein the tags of a
             given run and a given step are all collected in a single row.
             If not provided, defaults to `True`.
+          include_wall_time: Include wall_time (timestamps in microseconds since
+            the epoch in float64) as a column in the returned DataFrame.
+            If not provided, defaults to `False`.
 
         Returns:
           If `pivot` (default):
--- a/tensorboard/data/experimental/base_experiment.py
+++ b/tensorboard/data/experimental/base_experiment.py
@@ -32,8 +32,8 @@ class BaseExperiment(metaclass=abc.ABCMeta):
         self,
         runs_filter=None,
         tags_filter=None,
-        pivot=None,
-        include_wall_time=None,
+        pivot=True,
+        include_wall_time=False,
     ):
         """Export scalar data as a pandas.DataFrame.
 
--- a/tensorboard/data/experimental/base_experiment.py
+++ b/tensorboard/data/experimental/base_experiment.py
@@ -46,7 +46,7 @@ class BaseExperiment(metaclass=abc.ABCMeta):
             `pivot_data()` method to a “wide” format wherein the tags of a
             given run and a given step are all collected in a single row.
             If not provided, defaults to `True`.
-          include_wall_time: Include wall_time (timestamps in microseconds since
+          include_wall_time: Include wall_time (timestamps in nanoseconds since
             the epoch in float64) as a column in the returned DataFrame.
             If not provided, defaults to `False`.
 
--- a/tensorboard/data/experimental/experiment_from_dev.py
+++ b/tensorboard/data/experimental/experiment_from_dev.py
@@ -70,7 +70,13 @@ class ExperimentFromDev(base_experiment.BaseExperiment):
         self._experiment_id = experiment_id
         self._api_client = get_api_client(api_endpoint=api_endpoint)
 
-    def get_scalars(self, runs_filter=None, tags_filter=None, pivot=None):
+    def get_scalars(
+        self,
+        runs_filter=None,
+        tags_filter=None,
+        pivot=None,
+        include_wall_time=None,
+    ):
         if runs_filter is not None:
             raise NotImplementedError(
                 "runs_filter support for get_scalars() is not implemented yet."
@@ -80,6 +86,9 @@ class ExperimentFromDev(base_experiment.BaseExperiment):
                 "tags_filter support for get_scalars() is not implemented yet."
             )
         pivot = True if pivot is None else pivot
+        include_wall_time = (
+            False if include_wall_time is None else include_wall_time
+        )
 
         request = export_service_pb2.StreamExperimentDataRequest()
         request.experiment_id = self._experiment_id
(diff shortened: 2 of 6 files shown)


caisq:
cc @shoyer

tensorboard/data/experimental/experiment_from_dev.py
```diff
@@ -107,33 +116,41 @@ def get_scalars(self, runs_filter=None, tags_filter=None, pivot=None):
...
         if num_missing_1 > num_missing_0:
             raise ValueError(
-                "pivoted DataFrame contains %d missing value(s). "
+                "pivoted DataFrame contains missing value(s). "
                 "This is likely due to two timeseries having different "
                 "sets of steps in your experiment. "
                 "You can avoid this error by calling `get_scalars()` with "
                 "`pivot=False` to disable the DataFrame pivoting."
             )
+        dataframe = dataframe.reset_index()
```
wchargin, line 152: I’m having a hard time understanding what this actually does (and the
test doesn’t really help, because it just does the same thing as the
code). It seems confusing to me to pivot a table and then immediately
flatten the indices again. What kind of structure are we going for here?
I’ve read the Pandas docs but still don’t quite follow.

tensorboard/data/experimental/base_experiment.py
```diff
@@ -40,6 +46,9 @@ def get_scalars(self, runs_filter=None, tags_filter=None, pivot=None):
             `pivot_data()` method to a “wide” format wherein the tags of a
             given run and a given step are all collected in a single row.
             If not provided, defaults to `True`.
+          include_wall_time: Include wall_time (timestamps in microseconds since
```
wchargin, line 49: They’re seconds (`ToNanoseconds() / 1e9`) since epoch, not microseconds.

tensorboard/data/experimental/experiment_from_dev_test.py
```diff
@@ -105,19 +101,29 @@ def stream_experiment_data(request, **kwargs):
...
+                            expected = expected.pivot_table(
+                                values=(
+                                    ["value", "wall_time"]
+                                    if include_wall_time
+                                    else "value"
+                                ),
+                                index=["run", "step"],
+                                columns="tag",
+                            ).reset_index()
+                            expected.columns.name = None
```
shoyer, line 124: I would suggest directly constructing the desired DataFrame here (i.e., with a direct call to `pandas.DataFrame`), rather than calling `pivot_table`, etc.

tensorboard/data/experimental/experiment_from_dev.py
```diff
@@ -80,6 +86,9 @@ def get_scalars(self, runs_filter=None, tags_filter=None, pivot=None):
                 "tags_filter support for get_scalars() is not implemented yet."
             )
         pivot = True if pivot is None else pivot
+        include_wall_time = (
+            False if include_wall_time is None else include_wall_time
+        )
```
shoyer, line 91: Is there a reason why you are setting default values to `None` rather than `True`/`False`?

Using a default value of `None` is recommended if the default would otherwise be a mutable variables (e.g., an empty `list` or `dict`), but booleans are immutable. It is clearer to use `True`/`False` directly than to set them inside the method.

Sometimes there are legitimate use-cases for defaulting a boolean variable to `None`, for detecting if a parameter was explicitly passed or not. But I don't think they apply here. (And if they do apply in the future, you can always change the default value later.)

tensorboard/data/experimental/experiment_from_dev_test.py
caisq: Great suggestion. Done!

tensorboard/data/experimental/experiment_from_dev.py
shoyer: Insisting that _all_ default values are `None` really is not standard practice in Python (e.g., you won't find it in Google's Python Style Guide). It doesn't really matter, but you might try inquiring with g/python-users or g/python-style for opinions of other experienced Python programmers.

Here's my take:

1. This extensibility concern about keeping default values in sync is only relevant if you're using _implementation inheritance_, but implementation inheritance is a maintenance nightmare that should be avoided in almost all case. If you're not using implementation inheritance, then you'll have to duplicate the logic for default values anyways.
2. Open source and data science users in particular are much more likely to see public _interfaces_ for code (e.g., with auto-complete or `SomeClass.some_method?` in a notebook) than Google engineers who will go straight to the original source code to figure out what it does. Even if there is a maintenance cost associated with duplicating default values, it would be worth it to make interfaces more self-documenting.

wchargin: Definitely agree about avoiding implementation inheritance, but the
extensibility concern is also relevant with composition; any kind of
wrapper function that just needs to forward arguments will run into
this. (I run into this fairly frequently, and having to write tests that
ensure that the `inspect.signature`s are kept in sync is pretty
annoying.) I don’t have super strong opinions about this and won’t block
over it, but I do think that the `None`-default form stands on its
merits—and I consider myself an open source user first and foremost. :-)

caisq: Thank you both for the discussion, William and Stephan. I'll take the liberty to be the tie breaker here :) 
I'll err on the side of making life easier for open-source users and change the default values to explicit `True`/`False`. An additional argument is that it's confusing for generated docs to say something like "pivot=None". This argument is similar to the notebook/IDE auto-complete argument, but is also somewhat standalone because there are users who don't use IDE but still look at docs.

caisq: Done. PTAL.

shoyer:
> * of course, a multi-dimensional axis can’t be naturally expressed
>   into CSV, but Pandas does a better(?) job of retaining the structure
>   when only the horizontal axis is multidimensional, and that’s the
>   primary motivation for this change.

It's not just pandas, but most CSV tools that handle reading flat column names better. If you set `include_wall_times=False` (now the default), they you get flat column names, which should avoid all the NaNs.

I'm not sure if there are real use-cases for including wall-times with the multi-index columns in the pivoted dataframes. In practice, if I wanted to analyze this data I would immediately "flatten" the hierarchical index and use only one column of wall-times. If wall-times or steps differ meaningfully for different variables, then those variables would be split into different dataframes. I doubt this sort of sophisticated reasoning can be handled well by heuristics, so perhaps those sort of manipulations are better left to user code.

caisq:
@wchargin Thanks for pointing out the caveat of CSV roundtripping when `inclue_wall_time=True`. Yes, in that case the roundtrip gives you extraneous bits such as the `NaN`s and the suffixes for `value` and `wall_time` such as `value.1`.

I think the use case of wall_time data will be rare. Also as @shoyer mentioned, in those rare use cases, users can write their own DataFrame manipulation logic to extract the time information they need.

caisq:
@wchargin I think in the documentation and colab examples, we should clarify these two things we found during this conversation: CSV roundtripping will be perfect only if these two conditions are satisifed:
1. use the `index=False` trick when calling `to_csv()` as mentioned by @shoyer
2. `include_wall_time` is `False`.

PR #3606, opened 2020-05-08
[Debugger] Replace some usage of innerHTML with innerText

caisq:
Use innerText in lieu of innerHTML in the debugger data card in the graphs plugin.


tensorboard/plugins/graph/tf_graph_debugger_data_card/tf-graph-debugger-data-card.html
```diff
@@ -554,7 +554,7 @@ <h2 id="numeric-alerts-header">Numeric Alerts</h2>
             if (alert.nan_event_count) {
               var nanCountSection = document.createElement('div');
               nanCountSection.classList.add('nan-mini-health-pill-section');
-              nanCountSection.innerHTML = alert.nan_event_count;
+              nanCountSection.innerext = alert.nan_event_count;
```
bmd3k, line 557: innerext => innerText

caisq: Oops. Thanks for the catch. 

Fixed.

PR #3614, opened 2020-05-10
uploader: add rpc method GetExperiment to ExporterService

caisq:
  * In the experimental `ExperimentFromDev` class, support getting metadata including
    experiment name and description.
  * In the `ExperimentalFromDev.get_scalars()`, support progress indicaor.
  * Add non-streaming rpc method `GetExperiment()` to `ExporterService`.
  * Add `Experiment` as a one-of response data type to `StreamExperimentDataResponse`
    * Con: In the current takeout paradigm, this leads to duplicate information.
    * Con: Wasteful when only the `Experiment` (metadata) is needed and all the scalars, tensor and blob sequences are not needed.

### DIFF
--- a/tensorboard/uploader/proto/export_service.proto
+++ b/tensorboard/uploader/proto/export_service.proto
@@ -16,8 +16,8 @@ service TensorBoardExporterService {
   // Get the metadata of an experiment, including the creation time, name,
   // description, counts of scalars, tensors and blob sequences, etc. See the
   // documentation string of `Experiment` for more details.
-  rpc GetExperimentMetadata(GetExperimentMetadataRequest)
-      returns (GetExperimentMetadataResponse) {}
+  rpc GetExperiment(GetExperimentRequest)
+      returns (GetExperimentResponse) {}
   // Stream scalars for all the runs and tags in an experiment.
   rpc StreamExperimentData(StreamExperimentDataRequest)
       returns (stream StreamExperimentDataResponse) {}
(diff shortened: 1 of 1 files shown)


tensorboard/uploader/proto/export_service.proto
```diff
@@ -13,6 +13,11 @@ service TensorBoardExporterService {
   // Stream the experiment_id of all the experiments owned by the caller.
   rpc StreamExperiments(StreamExperimentsRequest)
       returns (stream StreamExperimentsResponse) {}
+  // Get the metadata of an experiment, including the creation time, name,
+  // description, counts of scalars, tensors and blob sequences, etc. See the
+  // documentation string of `Experiment` for more details.
+  rpc GetExperimentMetadata(GetExperimentMetadataRequest)
+      returns (GetExperimentMetadataResponse) {}
```
wchargin, line 20: Per <https://aip.dev/131> and previous discussions, can we just make
this `GetExperiment`, since it returns an `Experiment`?

tensorboard/uploader/proto/export_service.proto
```diff
@@ -80,6 +85,19 @@ message StreamExperimentsResponse {
   repeated Experiment experiments = 2;
 }
 
+// Request to get the metadata of an experiment.
+message GetExperimentMetadataRequest {
+  // ID of the experiment to get.
+  string experiment_id = 1;
+}
```
wchargin, line 92: Let’s include an `experiment_mask` here so that we can avoid requesting
some of the more expensive fields if we don’t need to?

tensorboard/uploader/proto/export_service.proto
wchargin: The spec says that

> The response message **must** be the resource itself. (There is no
> `GetBookResponse`.)

—but it looks like this patch still returns `GetExperimentResponse`
rather than `Experiment`?

caisq: Oops. My bad. Didn't read the AIP spec carefully. Done.

PR #3625, opened 2020-05-13
Expand UploadLimits proto and use it in TensorBoardUploader.

bmd3k:

There are a number of new parameters that impact the throughput of uploading data that we want the TensorBoard.dev frontend to return to the uploader in the handshake. Each of scalar, tensor, and blob now have a max\_[scalars|tensor|blob]\_request\_size and min\_[scalars|tensor|blob]\_request\_interval parameters.

Update the UploadLimits proto to include the new fields. Allow an instance of UploadLimits to be passed to TensorBoardUploader and use this throughout the code instead of the constants that have been used.

None of the clients for TensorBoardUploader are yet updated to pass UploadLimits (there is one in this git repo and there is one internal to Google) so temporarily have TensorBoardUploader construct its own UploadLimits if None is passed, where values are all based on now-deprecated constants and arguments to the constructor.  We hope to update clients in the next day or two so this is intended to be very temporary.

Tests have been updated to use UploadLimits as arguments exclusively.

### DIFF
--- a/tensorboard/uploader/proto/server_info.proto
+++ b/tensorboard/uploader/proto/server_info.proto
@@ -114,11 +114,11 @@ message UploadLimits {
   // non-positive then no blobs should be uploaded.
   int64 max_blob_request_size = 5;
 
-  // The minimum interval between WriteScalar requests, in seconds.
+  // The minimum interval between WriteScalar requests, in milliseconds.
   int64 min_scalar_request_interval = 6;
-  // The minimum interval between WriteTensor requests, in seconds.
+  // The minimum interval between WriteTensor requests, in milliseconds.
   int64 min_tensor_request_interval = 7;
-  // The minimum interval between WriteBlob requests, in seconds.
+  // The minimum interval between WriteBlob requests, in milliseconds.
   int64 min_blob_request_interval = 8;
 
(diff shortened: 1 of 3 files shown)


tensorboard/uploader/proto/server_info.proto
```diff
@@ -104,6 +104,23 @@ message PluginControl {
...
+  // non-positive then no blobs should be uploaded.
+  int64 max_scalar_request_size = 3;
+  // The maximum allowed WriteTensor request size, in bytes. If this is
+  // non-positive then no blobs should be uploaded.
+  int64 max_tensor_request_size = 4;
+  // The maximum allowed WriteBlob request size, in bytes. If this is
+  // non-positive then no blobs should be uploaded.
+  int64 max_blob_request_size = 5;
+
+  // The minimum interval between WriteScalar requests, in seconds.
```
ericdnielsen, line 117: Is seconds the right granularity?  Would we ever want to be able to support 2 rps?

bmd3k: Done. Changed to milliseconds. I don't know if we'll want to configure sub-second intervals but we should allow it, just in case.

PR #3630, opened 2020-05-13
[DebuggerV2] Add basic reducers for graph ops

caisq:
  * Continue developing DebuggerV2, specifically its graph component
  * Add three actions: `graphOpFocused`, `graphOpInfoRequested` and `graphOpInfoLoaded`
  * Add reducers for them.
  * Unit tests are *not* added in the existing debugger_reducers_test.ts, because the file is getting too large in size. Instead, the new unit tests are in a new file in the same folder: debugger_graphs_reducers_test.

### DIFF
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_graphs_reducers_test.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_graphs_reducers_test.ts
@@ -1,0 +1,463 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import * as actions from '../actions';
+import {reducers} from './debugger_reducers';
+import {DataLoadState} from './debugger_types';
+import {
+  createDebuggerGraphsState,
+  createDebuggerState,
+  createTestGraphOpInfo,
+} from '../testing';
+
+describe('Debugger reducers', () => {
+  describe('graphOpFocused', () => {
+    it('sets focusedOp in graphs state from empty state', () => {
+      const state = createDebuggerState();
+      const nextState = reducers(
+        state,
+        actions.graphOpFocused({graph_id: 'g2', op_name: 'TestOp_12'})
+      );
+      expect(nextState.graphs.focusedOp).toEqual({
+        graphId: 'g2',
+        opName: 'TestOp_12',
+      });
+    });
+
+    it('sets focusedOp in graphs state from non-empty state', () => {
+      const state = createDebuggerState({
+        graphs: createDebuggerGraphsState({
+          focusedOp: {
+            graphId: 'g1',
+            opName: 'TestOp_1',
+          },
+        }),
+      });
+      const nextState = reducers(
+        state,
+        actions.graphOpFocused({graph_id: 'g2', op_name: 'TestOp_12'})
+      );
+      expect(nextState.graphs.focusedOp).toEqual({
+        graphId: 'g2',
+        opName: 'TestOp_12',
+      });
+    });
+  });
+
+  describe('graphOpInfoRequested', () => {
+    it('creates key for new graph_id', () => {
+      const state = createDebuggerState();
+      const nextState = reducers(
+        state,
+        actions.graphOpInfoRequested({graph_id: 'g8', op_name: 'x'})
+      );
+      expect(nextState.graphs.loadingOps).toEqual({
+        g8: {x: DataLoadState.LOADING},
+      });
+    });
+
+    it('adds op to existing graph key', () => {
+      const state = createDebuggerState({
+        graphs: createDebuggerGraphsState({
+          loadingOps: {
+            g1: {Op1: DataLoadState.LOADING},
+            g2: {Op2: DataLoadState.LOADING},
+          },
+        }),
+      });
+      const nextState = reducers(
+        state,
+        actions.graphOpInfoRequested({graph_id: 'g2', op_name: 'Op3'})
+      );
+      expect(nextState.graphs.loadingOps).toEqual({
+        g1: {Op1: DataLoadState.LOADING},
+        g2: {
+          Op2: DataLoadState.LOADING,
+          Op3: DataLoadState.LOADING,
+        },
+      });
+    });
+
+    it('no effect for an already-loading op', () => {
+      const state = createDebuggerState({
+        graphs: createDebuggerGraphsState({
+          loadingOps: {
+            g1: {Op1: DataLoadState.LOADING},
+            g2: {Op2: DataLoadState.LOADING},
+          },
+        }),
+      });
+      const nextState = reducers(
+        state,
+        actions.graphOpInfoRequested({graph_id: 'g2', op_name: 'Op2'})
+      );
+      expect(nextState.graphs.loadingOps).toEqual({
+        g1: {Op1: DataLoadState.LOADING},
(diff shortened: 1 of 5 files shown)


caisq:
cc @davidsoergel

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_graphs_reducers_test.ts
```diff
@@ -0,0 +1,463 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
stephanwlee, line 1: 2019 -> 2020.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_graphs_reducers_test.ts
```diff
@@ -0,0 +1,463 @@
...
+          {
+            op_name: opInfo2.op_name,
+            input_slot: 0,
+          },
+        ],
+      ];
+      opInfo2.inputs = [
+        {
+          op_name: opInfo1.op_name,
+          output_slot: 1,
```
stephanwlee, line 144: Sorry, I am missing a bit of context. 

This is the output_slot of the op1, right? How is the op1's output defined? Is it the `consumers`? If so, I only see one output. 

I guess another way to say that is what does this graph look like (visually)?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -674,6 +675,102 @@ const reducer = createReducer(
...
+          },
+          loadingOps: {
+            ...state.graphs.loadingOps,
+          },
+        },
+      };
+      for (const input of inputs) {
+        if (!input.data) {
+          continue;
+        }
```
stephanwlee, line 745: Help lazy person like me by documenting how this can happen.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -674,6 +675,102 @@ const reducer = createReducer(
...
+        if (newState.graphs.ops[graphId] === undefined) {
+          newState.graphs.ops[graphId] = {};
+        }
+        newState.graphs.ops[graphId][input.op_name] = input.data;
+        delete input.data;
```
stephanwlee, line 750: why do you need to modify `input`?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -674,6 +675,102 @@ const reducer = createReducer(
...
+          }
+          if (newState.graphs.ops[graphId] === undefined) {
+            newState.graphs.ops[graphId] = {};
+          }
+          newState.graphs.ops[graphId][consumer.op_name] = consumer.data;
+          delete consumer.data;
+        }
+      }
+      if (newState.graphs.ops[graphId] === undefined) {
+        newState.graphs.ops[graphId] = {};
```
stephanwlee, line 765: I see 
```ts
if (newState.graphs.ops[graphId] === undefined) {
        newState.graphs.ops[graphId] = {};
}
```

repeated thrice. Just put it near the `newState` creation?

```ts
graphs: {
          ...state.graphs,
          ops: {
            ...state.graphs.ops,
            [graphId]: {
              ...state.graphs.ops[graphId],
            },
          },
}
```

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -674,6 +675,102 @@ const reducer = createReducer(
...
+      }
+      newState.graphs.ops[graphId][opInfo.op_name] = opInfo;
+      // Remove the loading marker for the op.
+      newState.graphs.loadingOps[graphId][opInfo.op_name] =
```
stephanwlee, line 769: Logically impossible for this to be true but the check `if (newState.graphs.loadingOps[graph_id] === undefined) {` might lower cognitive overhead (I would not have to reason harder on other flows)

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_graphs_reducers_test.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_graphs_reducers_test.ts
caisq: This is a test glitch. Fixed. You're right, op1 only has one output. It didn't cause a test failure before because the affected part was not explicitly asserted on. The assertion was on the whole object. Note the lines that say:

```
     expect(nextState.graphs.ops).toEqual({
        // Verify the old graph op data hasn't changed.
        g0: {[opInfo0.op_name]: opInfo0},
        // 'g2' is the immediately-enclosing graph of the three ops.
        g2: {
          [opInfo1.op_name]: opInfo1,
          [opInfo2.op_name]: opInfo2,
          [opInfo3.op_name]: opInfo3,
        },
      });
```

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
caisq: Added a comment here. Basically, `input.data` can be undefined if the backend fails to lookup the data regarding the input op, e.g., due to certain special TF-internally generated ops such as `StatefulPartitionedCall`. This is explained in the documentation of the backend code: https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/debugger_v2/debugger_v2_plugin.py#L317

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
caisq: Added a comment to explain that. The `input` here is a field of `opInfo`. `opInfo` is put into `newState.graphs.ops[graphId][opInfo.op_name]` below. We don't want the store
to contain duplicate data between the op itself (described by `opInfo`) and the input op
(described by `input`).

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
caisq: Good suggestion. Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
caisq: I addressed this by using a similar approach as above:

```
        graphs: {
          ...state.graphs,
          ops: {
            ...state.graphs.ops,
            [graphId]: {
              ...state.graphs.ops[graphId],
            }
          },
          loadingOps: {
            ...state.graphs.loadingOps,
            [graphId]: {
              ...state.graphs.loadingOps[graphId],   // <--
            }
          },
        },
```

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
stephanwlee: I guess I am still lost as to why we are modifying the action payload (this delete call is only modifying the `inputs` which is from the `data` (really, action payload), right?) To ease my confusion, can you point me to a test that covers what you just said?

caisq: I agree that mutating the data payload is suboptimal, even though it doesn't cause any us any harm right away.

I've made the following revision. 
1. In stead of calling `delete` on the payload, I now shallow copy the payload into another object before using it.
    The shallow copy skips the things we don't want. So there is no more need for `delete`. (new L766-L782).
2. I added a few lines in the unit test to reflect the expected behavior.

PR #3633, opened 2020-05-14
Use tf-wbr-string in tf-node-info

bmd3k:

tf-node-info assigns some user-generated value to an element using innerHTML.  It does this because it augments the user-generated value with some \<wbr\> tags.  Need to find a way to do this that is less prone to attack.

There is a nice tf-wbr-string component that already does what we want. Refactor to a common location and allow callers to specify the delimiter for \<wbr\> insertion.

Run tensorboard and manually inspect DOM in the run selector and in the graph node info title to ensure \<wbr\>s are being properly inserted in text.

For the graph node, more specifically, have to drill down several layers in the debug model to generate a node title with sufficient length.

### DIFF
--- a/tensorboard/components/tf_runs_selector/tf-runs-selector.html
+++ b/tensorboard/components/tf_runs_selector/tf-runs-selector.html
@@ -39,7 +39,7 @@ Properties out:
       <h2>Data Location</h2>
       <tf-wbr-string
         value="[[dataLocation]]"
-        delimiter-pattern="[[_dataLocationDelimiterPattern]]"
+        delimiters="[[_dataLocationWbrDelimiters]]"
       />
     </paper-dialog>
     <div id="top-text">
@@ -60,7 +60,7 @@ Properties out:
       <div id="data-location">
         <tf-wbr-string
           value="[[_clippedDataLocation]]"
-          delimiter-pattern="[[_dataLocationDelimiterPattern]]"
+          delimiters="[[_dataLocationWbrDelimiters]]"
         /><!--
           We use HTML comments to remove spaces before the ellipsis.
         --><template
@@ -154,9 +154,9 @@ Properties out:
           value: 250,
           readOnly: true,
         },
-        _dataLocationDelimiterPattern: {
+        _dataLocationWbrDelimiters: {
           type: String,
-          value: '[/=_,-]',
+          value: '/=_,-',
           readOnly: true,
         },
         coloring: {
--- a/tensorboard/components/tf_runs_selector/tf-runs-selector.html
+++ b/tensorboard/components/tf_runs_selector/tf-runs-selector.html
@@ -39,7 +39,8 @@ Properties out:
       <h2>Data Location</h2>
       <tf-wbr-string
         value="[[dataLocation]]"
-        delimiter-pattern="[[_dataLocationDelimiterPattern]]" />
+        delimiter-pattern="[[_dataLocationDelimiterPattern]]"
+      />
     </paper-dialog>
     <div id="top-text">
       <h3 id="tooltip-help" class="tooltip-container">Runs</h3>
@@ -59,7 +60,8 @@ Properties out:
       <div id="data-location">
         <tf-wbr-string
           value="[[_clippedDataLocation]]"
-          delimiter-pattern="[[_dataLocationDelimiterPattern]]" /><!--
+          delimiter-pattern="[[_dataLocationDelimiterPattern]]"
+        /><!--
           We use HTML comments to remove spaces before the ellipsis.
         --><template
           is="dom-if"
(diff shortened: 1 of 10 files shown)


tensorboard/plugins/graph/tf_graph_op_compat_card/tf-graph-op-compat-card.html
```diff
@@ -227,11 +227,6 @@
         _getNode: function(nodeName, graphHierarchy) {
           return graphHierarchy.node(nodeName);
         },
-        _getPrintableHTMLNodeName: function(nodeName) {
```
bmd3k, line 230: I found this copy of _getPrintableHTMLNodeName() that appears to be unused. I don't think the wbr logic is necessary for the compatibility info section.

tensorboard/components/tf_runs_selector/tf-runs-selector.html
```diff
@@ -55,7 +58,10 @@ <h3 id="tooltip-help" class="tooltip-container">Runs</h3>
     </paper-button>
     <template is="dom-if" if="[[dataLocation]]">
       <div id="data-location">
-        <tf-wbr-string value="[[_clippedDataLocation]]" /><!--
+        <tf-wbr-string
+          value="[[_clippedDataLocation]]"
+          delimiter-pattern="[[_dataLocationDelimiterPattern]]"
```
wchargin, line 63: You can just inline the literal string here, which removes the need for
the extra (read-only, constant) property.

tensorboard/components/tf_wbr_string/tf-wbr-string.html
```diff
@@ -37,19 +37,21 @@
       properties: {
         /** The value to render with word breaks. */
         value: String,
+        /** The regular expression for specifying word break delimiters. */
```
wchargin, line 40: The code as written works only if `delimiterPattern` matches are always
exactly one character long (because of `idx + 1` in `_computeParts`).
Could we either (a) remove the footgun by interpreting this as a list of
characters rather than a full-fledged regex, (b) remove the footgun by
making the `_computeParts` code handle the general case, or (c) document
the footgun? or something else?

(If option (b) or (c), maybe clarify the docs to say “the string source
of a regular expression…” to avoid suggesting that it’s actually a
`RegExp` value.)

tensorboard/components/tf_runs_selector/tf-runs-selector.html
bmd3k: I use it twice in the file, though. When I was playing around with the tf-wbr-string API today it was good to have one less place to change it.

tensorboard/components/tf_wbr_string/tf-wbr-string.html
bmd3k: I implemented (a) this morning but was unhappy with the result:
https://github.com/tensorflow/tensorboard/pull/3633/commits/5f1b06f82341ebe98ec48b8229ec38baa3f0c494

So I've gone with option (b) instead. I added some basic tests to validate.

wchargin: Okay, that’s fine with me, though it definitely increases the complexity
of the change.

FWIW, you probably want to avoid `new RegExp('[' + delimiters + ']');`,
which suffers from regex injection and thus doesn’t actually fulfill the
stated contract; instead, you might want [`_.escapeRegExp`][ld], since
the powers that be have decided in their infinite wisdom that
[`RegExp.escape` is too dangerous for us mere mortals to handle][tc].

[ld]: https://lodash.com/docs#escapeRegExp
[tc]: https://github.com/benjamingr/RegExp.escape/issues/37#issue-98309281

tensorboard/components/tf_runs_selector/tf-runs-selector.html
```diff
@@ -148,6 +154,11 @@ <h3 id="tooltip-help" class="tooltip-container">Runs</h3>
           value: 250,
           readOnly: true,
         },
+        _dataLocationWbrDelimiters: {
```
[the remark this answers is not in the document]
bmd3k: Done.

tensorboard/components/tf_wbr_string/tf-wbr-string.html
bmd3k: I'm a little confused. I had changed things so that I was not wrapping the regex in '[' and ']'.  I think using _.escapeRegExp would mean callers couldn't specify patterns like "[a-g]" and get it to work as expected.

tensorboard/components/tf_wbr_string/tf-wbr-string.html
```diff
@@ -37,25 +37,30 @@
...
-          value = '';
+          return [];
         }
         while (true) {
-          const idx = value.search(delimiterPattern);
-          if (idx === -1) {
+          const delimiterRegExp = new RegExp(delimiterPattern, 'g');
+          delimiterRegExp.test(value);
+          if (delimiterRegExp.lastIndex === 0) {
             result.push(value);
```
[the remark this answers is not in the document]
wchargin: Yeah, the behavior wrt overlapping matches is exactly what I’d expect.
The behavior wrt empty matches isn’t what I’d expect, but it’s okay with
me given that it’s documented.

PR #3637, opened 2020-05-15
Migrate _parse_samples_per_plugin() into flag definition itself

nfelt:
This logic is used both by the WSGI app side (passed to TBContext as `sampling_hints`) and by the event ingestion / data provider side (used for `tensor_size_guidance`), which are going to be split apart in coming changes.

Rather than create a utility that both places have to reference, it seems straightforward enough just to do the parsing within the flag definition itself.  This will require a corresponding update to any code that set this flag after parsing (i.e. in `tensorboard.program.TensorBoard.configure()`) of which there is one instance internally.

### DIFF
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -514,7 +514,7 @@ considered to have stable support for generic data providers. (default:
 
         parser.add_argument(
             "--samples_per_plugin",
-            type=_parse_dict,
+            type=_parse_samples_per_plugin,
             default="",
             help="""\
 An optional comma separated list of plugin_name=num_samples pairs to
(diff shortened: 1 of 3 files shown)


tensorboard/plugins/core/core_plugin.py
```diff
@@ -576,3 +576,13 @@ def _gzip(bytestring):
     with gzip.GzipFile(fileobj=out, mode="wb", compresslevel=3, mtime=0) as f:
         f.write(bytestring)
     return out.getvalue()
+
+
+def _parse_dict(value):
```
wchargin, line 581: Optional, but maybe rename to `_parse_samples_per_plugin` or similar, to
avoid giving the impression that this parses arbitrary dicts or that it
accepts standard dict notation (`"{\"key\": \"value\"}"`)?

nfelt: Done.

PR #3639, opened 2020-05-16
Introduce reload-on-ready prop to dashboard

stephanwlee:
We now have different shell around dashboards. Especially the Angular
can now reason better on when the data should be fetched. Currently, all
dashboards `reload` as soon as mounted and we would like to be able to
control that.

Note that this change does not fix/address hparams since the logic is
deeply embedded into query-pane and makes request using observers which
fire on `ready` (props get set on ready and gets fired then).


wchargin:
Oops, should have read description more carefully. LGTM.

wchargin:
For reference, this comes from #3589.

Issue #3635, opened 2020-05-15
Run selector hidden when window is not tall enough


wchargin:
If the window height is less than about 520px (on my machine), the run
selector’s regex filter and list of runs are hidden entirely, and can’t
be shown by scrolling (there is no scrollbar):

![Screenshot of run selector without any content][ss]

The exact threshold depends on the amount of space taken up by other UI
elements, like the data location.

Related: #2325.

[ss]: https://user-images.githubusercontent.com/4317806/82066019-bb91d400-9683-11ea-9609-ae89fd598121.png

PR #3640, opened 2020-05-16
Give min-height to run selectors

stephanwlee:
When window height is small (especially on smaller laptop), run
selectors completely collapse and do not show any item for the run.
This resulted in suboptimal experience and we would like to solve it by
giving a minimum height to the runs selector and allowing option to be
scrollable.

image

For the review, I strongly recommend enabling hide whitespace:
image

### DIFF
--- a/tensorboard/components/tf_dashboard_common/dashboard-style.html
+++ b/tensorboard/components/tf_dashboard_common/dashboard-style.html
@@ -40,7 +40,6 @@ limitations under the License.
       }
 
       .settings {
-        contain: layout paint;
         min-height: 50px;
         overflow-x: hidden;
         overflow-y: auto;
@@ -78,10 +77,6 @@ limitations under the License.
 
       .sidebar-section:first-of-type {
         border: none;
-      }
-
-      .sidebar-section:last-of-type {
-        flex-grow: 1;
       }
 
       .sidebar-section paper-button {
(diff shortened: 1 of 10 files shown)


tensorboard/components/tf_dashboard_common/dashboard-style.html
```diff
@@ -66,7 +82,6 @@
 
       .sidebar-section:last-of-type {
         flex-grow: 1;
```
psybuzz, line 84: Now that .runs-selector has `flex-grow: 1`, can we drop this `last-of-type` rule?

tensorboard/components/tf_dashboard_common/dashboard-style.html
```diff
@@ -33,12 +33,26 @@
...
         height: 100%;
-        margin-right: 20px;
+        margin-right: 10px;
         overflow-x: hidden;
         padding: 5px 0;
         text-overflow: ellipsis;
       }
 
+      .settings {
+        contain: layout paint;
```
psybuzz, line 43: If overflow is already set to hide in x direction and scroll in y direction, is this `contain` adding anything?

stephanwlee: I thought it would not hurt to give browser some hint to not paint/layout out of the container in all cases but it was done without any proof that layout was causing issue. Removed.

stephanwlee:
Fixes #3635.

Also fixes b/156499704.

PR #3645, opened 2020-05-17
[DebuggerV2] Add effects to load graph ops & associated stack frames

caisq:
  * Continue developing DebuggerV2, specifically its graph visualization part.
  * In debugger_effects.ts, add effects to do two things:
    1. Load graph op info using the `fetchGraphOpInfo()` method on the `graphOpFocused()` action
        (if and only if the op isn't already loaded or currently being loaded)
    2. Load the creation stack frames of the op (only the frames that haven't been loaded).


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -653,6 +657,88 @@ export class DebuggerEffects {
...
+        this.store.select(getActiveRunId),
+        this.store.select(getLoadingGraphOps)
+      ),
+      filter(([actionData, runId, loadingOps]) => {
+        const {graph_id, op_name} = actionData;
+        return (
+          runId !== null &&
+          (loadingOps[graph_id] === undefined ||
+            loadingOps[graph_id][op_name] === undefined ||
+            loadingOps[graph_id][op_name] === DataLoadState.FAILED)
```
psybuzz, line 681: nit: perhaps use shorthand `!loadingOps[graph_id] || !loadingOps[graph_id][op_name]` ?

Can / will `loadingOps` ever store `DataLoadState.NOT_LOADED` ?

On a side note, I'm a little worried by how we use serializable objects as state to hold user values.  Someone can make an `op_name` called "toString" or "\__proto__".

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -653,6 +657,88 @@ export class DebuggerEffects {
...
+      }),
+      tap(([actionData]) =>
+        this.store.dispatch(graphOpInfoRequested(actionData))
+      ),
+      mergeMap(([actionData, runId]) => {
+        const {graph_id, op_name} = actionData;
+        return this.dataSource.fetchGraphOpInfo(runId!, graph_id, op_name).pipe(
+          map((graphOpInfoResponse) => {
+            this.store.dispatch(graphOpInfoLoaded({graphOpInfoResponse}));
+            return graphOpInfoResponse.stack_frame_ids;
```
psybuzz, line 692: nit: iirc, the pattern in most places is to separate `store.dispatch` calls into their own `tap()` operator.  This would be consistent with L684

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -653,6 +657,88 @@ export class DebuggerEffects {
...
+      })
+    );
+  }
+
+  private loadGraphOpStackFrames(
+    prevStream$: Observable<string[]>
+  ): Observable<void> {
+    return prevStream$.pipe(
+      withLatestFrom(
+        this.store.select(getActiveRunId),
```
psybuzz, line 705: In case the `activeRunId` might've changed since the `loadGraphOpInfo` stream started, should we make `loadGraphOpInfo` return an object that includes the non-null runId?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
caisq: This is unlikely, but for safety, I did what you suggested.

PR #3646, opened 2020-05-18
DataFrame API: Let pivot default to False

caisq:
  * Improve the usability of the (experimental) DataFrame API
  * Change the default value of the kwarg `pivot` to `False`.  Rationale: 
    * Many experiments don't have identical sets of steps across all tags. Raising an error at the end is not nice.
    * Raising a warning for missing values is potentially okay, but introduces users not seeing them and/or spamming stderr too much and hence users ignoring them. We don't want users to be misled.
    * As discussed else where, we will provide good examples and documentation to explain how to get the pivoted DataFrame that many users want (i.e., it's just setting `pivot=True` for experiments that meet the requirement.)
  * Manual verification using the test_binary.

### DIFF
--- a/tensorboard/data/experimental/base_experiment.py
+++ b/tensorboard/data/experimental/base_experiment.py
@@ -32,7 +32,7 @@ class BaseExperiment(metaclass=abc.ABCMeta):
         self,
         runs_filter=None,
         tags_filter=None,
-        pivot=True,
+        pivot=False,
         include_wall_time=False,
     ):
         """Export scalar data as a pandas.DataFrame.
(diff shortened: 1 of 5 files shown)


tensorboard/data/experimental/experiment_from_dev.py
```diff
@@ -74,9 +72,13 @@ def get_scalars(
         self,
         runs_filter=None,
         tags_filter=None,
-        pivot=True,
+        pivot=False,
         include_wall_time=False,
     ):
+        # NOTE(b/155980442): Import pandas early in this method, so if the
```
wchargin, line 78: Can this be a public issue reference, please? There’s nothing in the
linked bug that’s sensitive to Google.

caisq: OK. I'll create a public issue for this and move the pandas-related changes to a separate PR.

caisq: OK. This PR is now only about the default value of pivot.

tensorboard/data/experimental/base_experiment.py
```diff
@@ -45,7 +45,12 @@ def get_scalars(
           pivot: Whether to returned DataFrame will be pivoted (via pandas’
             `pivot_data()` method to a “wide” format wherein the tags of a
             given run and a given step are all collected in a single row.
-            If not provided, defaults to `True`.
+            Setting `pivot` to `True` stipulates that the sets of step values
+            are identical among all tags of the experiment (after any run and
```
wchargin, line 49: Is this among all tags or among all time series?

wchargin:
cc @GalOshri

tensorboard/data/experimental/base_experiment.py
caisq: Good question. It requires that in each run, all tags have the same set of steps. The steps may differ between to runs for the same tag. I clarified that in the doc string here.

PR #3649, opened 2020-05-19
ng: allow i18n in Angular

stephanwlee:
Fixes b/156957839.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "@angular/common": "^9.0.0",
     "@angular/core": "^9.0.0",
     "@angular/forms": "^9.0.0",
+    "@angular/localize": "^9.1.7",
     "@angular/material": "^9.0.0",
     "@angular/platform-browser": "^9.0.0",
     "@angular/platform-browser-dynamic": "^9.0.0",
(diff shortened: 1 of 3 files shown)


wchargin:
For context, the contents of the linked bug are:

> We currently do not use [i18n directives](https://angular.io/guide/i18n)
> in ng_tensorboard (including the code in tensorboard/webapp and the code
> in tensorboard/plugins/debugger_v2). However we may start using these
> directives in the future.
> 
> I discovered today that adding the i18n directive to ng_tensorboard
> cause test to hang and subsequently timeout.
> 
> Steps to reproduce:
> 1. In the OSS repo, add the i18n directive to any of the *.ng.html files in
>    tensorboard/plugins/debugger_v2, e.g., add `<div i18n>Foo</div>` to
>    debugger_component.ng.html
> 2. `bazel test tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/...`
>    and observe the timeout:
> 
> > 19 05 2020 01:35:52.145:INFO [Chrome Headless 68.0.3440.0 (Linux x86_64)]: Connected on socket 5bW84BYjoeb4wicsAAAA with id 60251317
> > 19 05 2020 01:36:22.147:WARN [Chrome Headless 68.0.3440.0 (Linux x86_64)]: Disconnected (0 times), because no message in 30000 ms.
> > Chrome Headless 68.0.3440.0 (Linux x86_64) ERROR
> >   Disconnected, because no message in 30000 ms.

stephanwlee:
Btw, confirmed the fix by adding i18n directive on header and running the test. Also confirmed that the directive was behaving correctly in ng_index.html.

PR #3653, opened 2020-05-19
Projector config at root becomes default config for any present runs

hfiller:
Update projector config to better handle multiple runs. If multiple runs are present, only one set of config/metadata needs to be present in the root directory, which will be used for any runs without config explicitly set.

### DIFF
--- a/tensorboard/plugins/projector/projector_plugin.py
+++ b/tensorboard/plugins/projector/projector_plugin.py
@@ -387,6 +387,10 @@ class ProjectorPlugin(base_plugin.TBPlugin):
             for tensor_name, tensor_shape in var_map.items():
                 if len(tensor_shape) != 2:
                     continue
+                # Optimizer slot values are the same shape as embeddings
+                # but are not embeddings.
+                if ".OPTIMIZER_SLOT" in tensor_name:
+                    continue
                 embedding = self._get_embedding(tensor_name, config)
                 if not embedding:
                     embedding = config.embeddings.add()
@@ -413,6 +417,9 @@ class ProjectorPlugin(base_plugin.TBPlugin):
     def _read_latest_config_files(self, run_path_pairs):
         """Reads and returns the projector config files in every run
         directory."""
+        """If no specific config exists, use the default config provided in
+        the root directory."""
+        default_config_fpath = os.path.join(self.logdir, metadata.PROJECTOR_FILENAME)
         configs = {}
         config_fpaths = {}
         for run_name, assets_dir in run_path_pairs:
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/projector/projector_plugin.py
```diff
@@ -422,6 +431,14 @@ def _read_latest_config_files(self, run_path_pairs):
                 with tf.io.gfile.GFile(config_fpath, "r") as f:
                     file_content = f.read()
                 text_format.Merge(file_content, config)
+            elif tf.io.gfile.exists(default_config_fpath):
+                with tf.io.gfile.GFile(default_config_fpath, "r") as f:
```
nfelt, line 435: This looks like it would read the same file contents over and over, once for each run that doesn't have its own projector config, which seems unnecessary. Can we restructure so that we only read the file at most once?  This isn't such a big deal for local filesystems but is more of a downside for e.g. GCS log directories.

tensorboard/plugins/projector/projector_plugin.py
```diff
@@ -422,6 +431,14 @@ def _read_latest_config_files(self, run_path_pairs):
...
+                    file_content = f.read()
+                text_format.Merge(file_content, config)
+                for embedding in config.embeddings:
```
nfelt, line 438: I'm a little unclear about why this branch (the default root config) needs to expand embedding metadata paths to absolute here, whereas the other branch (the per-run config) doesn't need to do this.  Can you add some brief commentary about this?  E.g. is it because the paths are interpreted relative to the run path, or something like that?

tensorboard/plugins/projector/projector_plugin.py
```diff
@@ -387,6 +387,10 @@ def _augment_configs_with_checkpoint_info(self):
             for tensor_name, tensor_shape in var_map.items():
                 if len(tensor_shape) != 2:
                     continue
+                # Optimizer slot values are the same shape as embeddings
+                # but are not embeddings.
+                if ".OPTIMIZER_SLOT" in tensor_name:
```
nfelt, line 392: I realize this is a small tweak, but since it doesn't appear to be related to the main change, and it's a functional change (as opposed to just refactoring or formatting), it would be ideal to break this out as its own PR.

tensorboard/plugins/projector/projector_plugin.py
hfiller: That's exactly it. It's interpreted relative to the run path. I'll add a comment explaining. :)

tensorboard/plugins/projector/projector_plugin.py
hfiller: 👍

hfiller: #3679

PR #3655, opened 2020-05-19
core: make TensorBoardWSGIApp args a bit easier to use

nfelt:
Small changes to make it easier to use `TensorBoardWSGIApp` directly:

- Clarify that `plugins` can accept several types, not just plugin loader instances (pursuant to the follow-up comments of https://github.com/tensorflow/tensorboard/pull/2884/files#r342324613)

- Change `assets_zip_provider` so that it's actually optional to omit, in which case we use a placeholder dummy zipfile that contains only an index.html

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -20,10 +20,12 @@ Provides TensorBoardWSGIApp for building a TensorBoard WSGI app.
 import base64
 import collections
 import hashlib
+import io
 import json
 import re
 import textwrap
 import time
+import zipfile
 
 import six
 from six.moves.urllib import (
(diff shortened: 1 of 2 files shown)


tensorboard/backend/application.py
```diff
@@ -546,3 +554,11 @@ def _clean_path(path):
     if path != "/" and path.endswith("/"):
         return path[:-1]
     return path
+
+
+def _placeholder_assets_zip_provider():
+    """Defines a default assets zip provider containing a dummy index.html."""
+    memfile = io.BytesIO()
```
wchargin, line 561: OK—this target has `srcs_version = "PY2AND3"`, but it looks to me like
the semantics of `io.BytesIO` are unchanged from 2 to 3, so seems fine.

nfelt: Which target? `:application` is only "PY3" now:

https://github.com/tensorflow/tensorboard/blob/f22060fb97423178ff2a67a363715a5dbb6735c4/tensorboard/backend/BUILD#L57

nfelt: Submitting but let me know if there's something I should follow up on here.

PR #3658, opened 2020-05-20
CI: Apply new Jupyter Notebook formatting checks (and fix the resulting failure)

davidsoergel:
`nbfmt.py` previously tolerated several errors that should have been failures.  That is fixed in https://github.com/tensorflow/docs/commit/f10110518e0fa60bc855e77354ae3eea156532c7.

Here we use the new version of `nbfmt.py`, and fix the remaining missing license issue to allow CI to pass.


davidsoergel:
For context, the previous formatting bug that alerted us to this CI problem was fixed in #3627.

davidsoergel:
Dunno what happened with the checks, but rerunning them resolved it.

PR #3659, opened 2020-05-20
Revert "uploader: add rpc method GetExperiment to ExporterService"

caisq:
Reverts tensorflow/tensorboard#3614 to resolve internal proto service
method name duplication. Will roll forward with fix later.


caisq:
@ericdnielsen Fortunately, not yet.

PR #3663, opened 2020-05-21
Reverts Vulcanization printer changes

stephanwlee:
Two commits (ebaf5b5 and 3777aec) introduced a bug where rel="import"
in the middle of the document causes the printer to print "<#root>".


stephanwlee:
cc @wchargin FYI.

PR #3665, opened 2020-05-21
DataFrame API: Add ExperimentFromDev to namespace tb.data.experimental.*

caisq:
  * Put the experimental `ExperimentFromDev` API (accessing tensorboard.dev
    scalars as `pandas.DataFrame`s) under a public import path:
    `tensorboard.data.experimental.ExperimentFromDev`.
  * Add the necessary `__init__.py` and the associated BUILD targets.
  * In build_pip_package.sh, add the line `pip install -r requirements.txt`,
     because we know have lines in the files imported from the
     `__init__.py` files that import third-party dependencies such as grpc and
     numpy.

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -64,6 +64,7 @@ py_library(
         ":errors",
         ":lazy",
         ":version",
+        "//tensorboard/data:lib_init_only",
     ],
 )
 
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -45,6 +45,7 @@ py_library(
         ":lib_init_only",
         ":notebook",
         ":program",
+        "//tensorboard/data:lib_init_only",
         "//tensorboard/summary",
         "//tensorboard/summary/writer",
     ],
@@ -64,7 +65,6 @@ py_library(
         ":errors",
         ":lazy",
         ":version",
-        "//tensorboard/data:lib_init_only",
     ],
 )
 
--- a/tensorboard/data/experimental/BUILD
+++ b/tensorboard/data/experimental/BUILD
@@ -22,7 +22,6 @@ py_library(
         "//tensorboard:expect_grpc_installed",
         "//tensorboard:expect_numpy_installed",
         "//tensorboard:expect_pandas_installed",
-        "//tensorboard:lazy",
         "//tensorboard/uploader:auth",
         "//tensorboard/uploader:server_info",
         "//tensorboard/uploader:util",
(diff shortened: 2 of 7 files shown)


tensorboard/pip_package/build_pip_package.sh
```diff
@@ -110,6 +110,7 @@ build() (
   # env markers are handled (https://github.com/pypa/setuptools/pull/1081)
   export PYTHONWARNINGS=ignore:DEPRECATION  # suppress Python 2.7 deprecation spam
   pip install -qU wheel 'setuptools>=36.2.0'
+  pip install -r requirements.txt
```
wchargin, line 113: We shouldn’t need `grpc`, etc. just to build the Pip package, just as we
currently don’t need to install requirements even though TensorBoard
code depends on lots of other non-vendored packages. (And of course the
rest of TensorBoard already uses numpy.) Could you please define `data`
as a lazy module, like all the other nontrivial public exports, to avoid
this?

tensorboard/data/experimental/BUILD
```diff
@@ -10,13 +10,13 @@ exports_files(["LICENSE"])
 py_library(
     name = "base_experiment",
     srcs = ["base_experiment.py"],
-    srcs_version = "PY3",
+    srcs_version = "PY2AND3",
```
wchargin, line 13: Why this change? It doesn’t look like there are any Google-internal
reverse dependencies of `//tensorboard:lib` that are still Python 2
binaries. This change is okay with me if it’s really necessary for some
reason, but in general we’ve been preferring to write new code as py3.

tensorboard/pip_package/build_pip_package.sh
wchargin: The `pip install -r requirements.txt` appears to still be here.

tensorboard/BUILD
```diff
@@ -64,6 +64,7 @@ py_library(
         ":errors",
         ":lazy",
         ":version",
+        "//tensorboard/data:lib_init_only",
```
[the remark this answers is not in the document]
caisq: Done.

tensorboard/data/experimental/BUILD
caisq: Yeah, see the comment I sent yesterday. This change is necessary because some internal build targets are PY2. Specifically the notebook-related builds. I discovered breakages in those builds during a test sync of this PR, which prompted this change.

tensorboard/pip_package/build_pip_package.sh
caisq: Sorry I missed that in the previous revision. Removed now.

caisq:
Gentle ping.

tensorboard/data/experimental/BUILD
caisq: After some further testing, I confirmed that the `PY2AND3` part isn't necessary. I have pushed a new commit to change them all to `PY3` in this file. PTAL @wchargin

PR #3674, opened 2020-05-27
Support jupyter-server-proxy (and other proxies)

zac-hopkinson:

Users will frequently want to run TensorBoard in environments where they are not allowed to open up or expose additional ports. For those cases [jupyter-server-proxy](https://github.com/jupyterhub/jupyter-server-proxy) allows forwarding web traffic to other services running on the system without opening new ports. (The Google Colab integration also works around this same issue of not being able to directly expose the port that TensorBoard runs on.)

This PR allows users to *optionally* pass the TensorBoard traffic through jupyter-server-proxy by setting the `TENSORBOARD_PROXY_URL` environment variable. When this variable is not set TensorBoard will behave as normal.

This adds support for a new optional environment variable `TENSORBOARD_PROXY_URL` which if set will be used as the URL's path for the IFRAME displayed in the notebook when running `%tensorboard`.

## Screenshots of UI changes

image

I've successfully run the Travis CI build on my fork of the project: https://travis-ci.com/github/zac-hopkinson/tensorboard

I've also built test environments of both JupyterLab and JupyterHub with the documentation that I used to both configure and test this feature. You can find this documentation here: https://github.com/zac-hopkinson/tensorboard-pr-test

You can skip reading the documentation and just `export TENSORBOARD_PROXY_URL="/proxy/%PORT%/"` to use this feature with JupyterLab or Jupyter Notebook

My first attempt at this added a new context to go along with _CONTEXT_COLAB, _CONTEXT_IPYTHON, and _CONTEXT_NONE in [notebook.py](https://github.com/tensorflow/tensorboard/blob/348ee57a437dd6bb63c4ee1917c52ecd0c3ae033/tensorboard/notebook.py#L53). After testing with different environments I decided that it wasn't a clean solution like I had hoped. I found that adding in the environment variable logic to _CONTEXT_IPYTHON and making sure it preserved backwards compatibility was a much cleaner solution. I'm open to ideas and feedback around this decision of course!

I also researched #3142. Initially I hadn't planned on supporting JupyterHub but after testing I found that this PR can handle both Jupyter Notebook/Lab and JupyterHub by just documenting how to configure each.

I also considered alternatives to the environment variable, such as a possible `~/.tensorboardrc` file for example, but that seemed like overkill.

### DIFF
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -395,17 +395,17 @@ def _display_ipython(port, height, display_handle):
         replacements = [
             ("%HTML_ID%", html_escape(frame_id, quote=True)),
             ("%JSON_ID%", json.dumps(frame_id)),
-            ("%URL%", json.dumps(proxy_url)),
-            ("%PORT%", "0"),
             ("%HEIGHT%", "%d" % height),
+            ("%PORT%", "0"),
+            ("%URL%", json.dumps(proxy_url)),
         ]
     else:
         replacements = [
             ("%HTML_ID%", html_escape(frame_id, quote=True)),
             ("%JSON_ID%", json.dumps(frame_id)),
-            ("%URL%", json.dumps("/")),
-            ("%PORT%", "%d" % port),
             ("%HEIGHT%", "%d" % height),
+            ("%PORT%", "%d" % port),
+            ("%URL%", json.dumps("/")),
         ]
 
     for (k, v) in replacements:
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -23,6 +23,7 @@ from __future__ import print_function
 import datetime
 import errno
 import json
+import os
 import random
 import shlex
 import sys
@@ -378,18 +379,31 @@ def _display_ipython(port, height, display_handle):
       <script>
         (function() {
           const frame = document.getElementById(%JSON_ID%);
-          const url = new URL("/", window.location);
-          url.port = %PORT%;
+          const url = new URL("%ABS_PATH%", window.location);
+          %PORT_CHANGE%
           frame.src = url;
         })();
       </script>
-  """
+    """
     replacements = [
         ("%HTML_ID%", html_escape(frame_id, quote=True)),
         ("%JSON_ID%", json.dumps(frame_id)),
-        ("%PORT%", "%d" % port),
         ("%HEIGHT%", "%d" % height),
     ]
+    if "TENSORBOARD_PROXY_URL" in os.environ:
+        abs_path = os.environ["TENSORBOARD_PROXY_URL"]
+        # Allow %PORT% in $TENSORBOARD_PROXY_URL
+        abs_path = abs_path.replace("%PORT%", "%d" % port)
+        replacements.extend([
+            ("%ABS_PATH%", abs_path),
+            ("%PORT_CHANGE%", ""),
+        ])
+    else:
+        replacements.extend([
+            ("%ABS_PATH%", "/"),
+            ("%PORT_CHANGE%", "url.port = %d;" % port),
+        ])
+
     for (k, v) in replacements:
         shell = shell.replace(k, v)
     iframe = IPython.display.HTML(shell)
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -394,15 +394,13 @@ def _display_ipython(port, height, display_handle):
         abs_path = os.environ["TENSORBOARD_PROXY_URL"]
         # Allow %PORT% in $TENSORBOARD_PROXY_URL
         abs_path = abs_path.replace("%PORT%", "%d" % port)
-        replacements.extend([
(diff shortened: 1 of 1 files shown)


zac-hopkinson:
Thanks Brian! Yeah, it's always been possible to get to TensorBoard manually when using jupyter-server-proxy. This PR is to get the IPython magic (`%tensorboard`) to be able to use it and for it to work seamlessly.

Part of the logic in notebook.py is to embed TensorBoard into Jupyter using an iframe. This PR allows that iframe to optionally support going through the proxy when it's not possible for the connection to go directly. I put together some examples showing how it's used here: https://github.com/zac-hopkinson/tensorboard-pr-test/#usingtesting

bmd3k:
Thanks for the explanation! The use case makes sense to me. I played around with your changes on a local Jupyter notebook install. I'll chat with some members on the team about the approach and get back to you.

tensorboard/notebook.py
```diff
@@ -378,18 +379,29 @@ def _display_ipython(port, height, display_handle):
       <script>
         (function() {
           const frame = document.getElementById(%JSON_ID%);
-          const url = new URL("/", window.location);
-          url.port = %PORT%;
+          const url = new URL("%ABS_PATH%", window.location);
```
bmd3k, line 382: The value of `ABS_PATH` could be a relative path.  So perhaps just name to `URL` - this matches the argument name in MDN documentation and the official spec.

tensorboard/notebook.py
```diff
@@ -378,18 +379,29 @@ def _display_ipython(port, height, display_handle):
...
+          %PORT_CHANGE%
```
bmd3k, line 383: I think it would be more readable if we didn't use PORT_CHANGE to possibly generate an entire javascript statement.  I'd prefer if we just replaced simple types. Can your changes be refactored so that this is:

```
if ("%PORT%") {
  url.port = %PORT%;
}
```

wchargin: +1; I like this idea (though I’d suggest something like `%PORT%` being
either an int or null and setting `const port = %PORT%` rather than
using the substitution `%PORT%` once in a string context and once in a
raw context, which is a bit iffy).

zac-hopkinson:
Thanks for the really good feedback! Suggested changes in the latest commit.

New generated JS and the iframe in the default config:
generated_standard

And with `TENSORBOARD_PROXY_URL` set:
generated_proxy

tensorboard/notebook.py
```diff
@@ -378,18 +379,35 @@ def _display_ipython(port, height, display_handle):
       <script>
         (function() {
           const frame = document.getElementById(%JSON_ID%);
-          const url = new URL("/", window.location);
-          url.port = %PORT%;
+          const url = new URL(%URL%, window.location);
+          const port = %PORT%;
+          if (port) {
```
wchargin, line 384: OK—this branches on `0` as well, but `0` isn’t a valid port for opening
a connection, so this is fine with me. No action required.

tensorboard/notebook.py
```diff
@@ -378,18 +379,35 @@ def _display_ipython(port, height, display_handle):
...
-    ]
+    """
+    proxy_url = os.environ.get("TENSORBOARD_PROXY_URL")
+    if proxy_url is not None:
+        # Allow %PORT% in $TENSORBOARD_PROXY_URL
+        proxy_url = proxy_url.replace("%PORT%", "%d" % port)
+        replacements = [
+            ("%HTML_ID%", html_escape(frame_id, quote=True)),
+            ("%JSON_ID%", json.dumps(frame_id)),
+            ("%URL%", json.dumps(proxy_url)),
```
wchargin, line 398: Could you move `%URL%` to the end of the substitutions list just to
avoid any issues in case the environment variable happens to contain the
strings `%PORT%` or `%HEIGHT%`? (likewise below since this substitution
table appears twice)

bmd3k:
The lint-build errors in your latest CI run is not the fault of your PR. Should be fixed now. You can push an empty commit to rerun.

zac-hopkinson:
Thank you both for the help and the great code review comments!

lsyarn:
For any one who was confused with how to use this feature on version after 2.3 (included):

Due to #4421——`tensorboard did follow --path_prefix`, this feature does not working well on jupyterhub.
This feature released in version 2.3.0, and #4421 occurred in version 2.3.0 too.

jameswex:
@zac-hopkinson I'm looking to add JupyterHub support for another tool (https://github.com/PAIR-code/lit/, not TensorBoard). But our notebook-mode code is basically identical in function to what TensorBoard has done. I've gotten it working in colab and jupyter (as seen in https://github.com/PAIR-code/lit/pull/163 and https://github.com/PAIR-code/lit/pull/171), but the widget fails to render in JupyterHub environments (I've tried Google Cloud AI Notebooks and AWS Sagemaker notebooks specifically). If you have time, I'd like to pick your brain about this as you seem to be really knowledgeable about it.

PR #3678, opened 2020-05-28
uploader: display upload statistics while uploading data

caisq:
  * This is the first of a series of PRs aimed at improving the usability of the tensorboard.dev uploader.
  * This PR specifically adds real-time display of upload statistics.
  * During uploading, the users sees 2 or 3 bars rendered with `tqdm`:
    1. [TIMESTAMP] Uploaded x scalars, y tensors (Y B), z binary objects (Z B)
    2. (if any data was skipped): Skipped n tensors (N B), m binary objects (M B)
    3. Uploading XYZ (this is the real-time status)
  * If the user wishes to see a less verbose UI, they can use `--verbose 0` to override
     the default value 1
  * When the uploading is idling, it displays "Listening for new data in logdir..."
  * Add a module `upload_tracker.py` to encapsulate the logic for stats tracking
     and display.
* Screenshots of UI changes
  * While uploading actively: image
  * While idling: image
  * Unit tests added for the new module and its usage inside the existing `uploader.py` module
  * Manual verification (see screenshots above)


caisq:
cc @GalOshri

tensorboard/uploader/uploader.py
```diff
@@ -503,7 +534,12 @@ def flush(self):
 
         self._rpc_rate_limiter.tick()
 
-        with _request_logger(request, request.runs):
+        with contextlib.ExitStack() as stack:
```
davidsoergel, line 537: Took me a bit to figure out why you did it this way :)  I think it's because `self._tracker` might be `None`, so you can't just `with self._tracker.scalars_tracker(...):`.

I think this would be clearer, more idiomatic, and more flexible if you don't allow `self._tracker` to be None, but instead push the verbosity argument down into UploadTracker and handle it there (i.e., checking the condition only when it comes time to actually print something).  Then all of these ExitStack things go away and you can just say

```
with _request_logger(request, request.runs):
    with self._tracker.scalars_tracker(self._num_values):
        ...
```

tensorboard/uploader/uploader.py
```diff
@@ -157,6 +163,9 @@ def create_experiment(self):
         response = grpc_util.call_with_retries(
             self._api.CreateExperiment, request
         )
+        self._tracker = (
```
davidsoergel, line 166: Per comment below, how about
```
self._tracker = upload_tracker.UploadTracker(self._verbosity)
```
(and all the consequences of that refactoring)

tensorboard/uploader/uploader.py
caisq: Done. This increases the maximum depth of ident by 1. But I think it's worth the benefits you pointed out.

tensorboard/uploader/uploader.py
caisq: Done.

tensorboard/uploader/uploader.py
```diff
@@ -961,7 +1007,12 @@ def flush(self):
                 # Note the _send_blob() stream is internally flow-controlled.
                 # This rate limit applies to *starting* the stream.
                 self._rpc_rate_limiter.tick()
-                sent_blobs += self._send_blob(blob_sequence_id, seq_index, blob)
+                with self._tracker.blob_tracker(len(blob)) as blob_tracker:
+                    sent_blobs += self._send_blob(
+                        blob_sequence_id, seq_index, blob
+                    )
+                    if self._tracker:
```
davidsoergel, line 1014: this condition is no longer needed

tensorboard/uploader/uploader.py
```diff
@@ -594,6 +622,7 @@ def __init__(
         rpc_rate_limiter,
         max_request_size,
         max_tensor_point_size,
+        tracker=None,
```
davidsoergel, line 625: Don't default to None, now that this is required

tensorboard/uploader/uploader.py
```diff
@@ -440,13 +457,21 @@ class _ScalarBatchedRequestSender(object):
...
     """
 
-    def __init__(self, experiment_id, api, rpc_rate_limiter, max_request_size):
+    def __init__(
+        self,
+        experiment_id,
+        api,
+        rpc_rate_limiter,
+        max_request_size,
+        tracker=None,
```
davidsoergel, line 466: Don't default to None, now that this is required

tensorboard/uploader/uploader.py
caisq: Done. Same for another line like this.

caisq:
I've removed the usage of tqdm, because of two reasons
1. The multi-progress-bar setup used previously in this PR doesn't work well in notebooks, as I discovered through manual testing. So I've reverted to a sys.stdout writing and flushing only approach.
2. We're currently not using the main progress-bar feature of tqdm. So it seems to be prudent to avoid adding this new dependency for now.

PR #3682, opened 2020-05-29
[DebuggerV2] Highlight same-file frames and stick to bottommost frame

caisq:
  * Continue developing DebuggerV2
  * Improve the StackTrace component and its interaction with eager timeline and graph ops
  * In the StackTrace component, use bold font weight to highlight frames that are in the focused
    file, but different from the focused stack frame.
    * Use a different background color (in conjunction with the bold font weight) to highlight 
       the focused stack frame in the focused file
  * Add a logic to stick to the bottommost frame of a focused source file when the user
    has previously clicked on a bottommost frame in the file.
    * This enables the user to effectively "step through" eager and graph executions and see
      how the corresponding code location changes in the focused source file.
  * Add logic in `StackTraceComponent` to scroll to
    * The last frame if no frame is focused on. This makes the experience more consistent with
      Python stack traces in consoles.
    * The focused frame if there is one.
* Screenshots of UI changes
  * image
  * Unit tests added for the new actions, selectors, reducers and UI action dispatching.
  * Manual testing against a real tfdbg2 logdir (see screenshots above)


caisq:
@psybuzz As we discussed offline, I'm going to make changes to the PR so that:
1. The stack trace will scroll to bottom by default.
2. It'll still have an implicit state about whether the bottommost frame in a file is stuck to. This will support the line-following feature when user navigates the eager / graph executions.
3. It'll probably have some visual difference for bottommost vs non-bottommost frames in a source file.

I'll give you a notice when the PR is ready for re-review.

caisq:
@psybuzz I've implemented proposals 1 and 2 as listed above. 

The size of the PR is a little larger than before. But the complexity has actually decreased. There is no longer a new effects module. Instead, the business logic are all carried out by reducers and selectors.

Given the size of the PR, I decided to defer task 3 (marking bottommost frames in the stack trace view) for later.

PTAL.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/stack_trace/stack_trace_component.ng.html
```diff
@@ -18,9 +18,21 @@
...
 
+    <div
+      class="stick-to-bottommost-frame"
+      *ngIf="stickToBottommostFrameInFocusedFile"
+      title="Sticking to the bottommost frame in the current source file when
+      navigating executions and graph ops. To remove this sticking, click
+      any non-bottommost stack frame."
+    >
+      Sticking to bottommost frame in file
+    </div>
```
psybuzz, line 34: Introducing the concept of "stickiness" seems like exposing internal complexity to users, which never alters user behavior (even without help text, I imagine users naturally click on a different frame when they have unwanted sticking).

Could we either drop the word "sticking to" from the text content or consider a visual indicator on the callframe element itself (like styling the bottommost callframe differently)?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/stack_trace/stack_trace_component.ng.html
```diff
@@ -18,9 +18,21 @@
 <div class="stack-trace-container">
   <div class="stack-trace-header">
     <span class="stack-trace-title">
-      Stack Trace
+      <div class="stack-trace-title-text">
```
psybuzz, line 21: This classname is unused.  If the extra wrapper is unnecessary, let's revert?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors.ts
```diff
@@ -602,8 +618,25 @@ export const getFocusedSourceFileContent = createSelector(
...
+  selectDebuggerState,
+  (state: DebuggerState): SourceLineSpec | null => {
+    const stackFrames = getFocusedStackFramesHelper(state);
+    const focusedLineSpec = state.sourceCode.focusLineSpec;
+    if (
+      state.stickToBottommostFrameInFocusedFile &&
+      stackFrames !== null &&
+      focusedLineSpec !== null
+    ) {
+      return getBottommostStackFrameInFocusedFile(stackFrames, focusedLineSpec);
```
psybuzz, line 630: This selector is a bit confusing now.  Before, if my source code had focusLineSpec (say for L15), the selector just returned L15.

Now, the selector says to return the line corresponding to the bottommost callframe in this file (say for L200), which is a very different contract.  Changing `sourceCode.focusLineSpec` used to always trigger this selector, but now it may not.

It might be easier to reason about if we keep the selector simple, and move logic related to updating the `sourceCode.focusLineSpec` into the reducer.  Wdyt?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/stack_trace/stack_trace_component.ts
```diff
@@ -59,5 +75,31 @@ export class StackTraceComponent {
...
+      return;
+    }
+    const lastFrameElement: HTMLElement | null = stackElement.querySelector(
+      '.stack-frame-container:last-child'
+    );
+    if (lastFrameElement) {
+      // Scroll the last frame into view only when there is no stack frame being
+      // focused on.
+      lastFrameElement.scrollIntoView({block: 'nearest'});
+    }
```
psybuzz, line 103: If you prefer, L91-L103 might be simplified by assuming the last element is always at the bottom.

```
if (focusedFrameElement) {
  focusedFrameElement.scrollIntoView({block: 'nearest'});
} else {
  stackElement.scrollTop = stackElement.scrollHeight;
}
```

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors.ts
caisq: Good question. In order to implement the "stick to the bottommost frame when user navigates 
executions or graph ops", we need an extra level of indirection. This extra level can be implemented
in one of two ways: 
1. As extra logic in a selector, as is done here.
2. As extra logic in reducers.

The downside of 2 is that it needs to touch upon multiple reducers, including
- `executionDigestFocused`, to accommodate user navigation of executions
- `graphOpFocused`, to accommodate user navigation of executions
- `stackFramesLoaded`, because stack frames of executions and graph ops
   are loaded lazily. The bottommost-frame calculation often needs to wait for this
   lazy loading.

But 1 is logically simpler because it utilizes the selector's ability to automatically
listen to changes in relevant bits of the ngrx store.

So in balance, I think approach 1 (the current approach) is better. To address your 
concern about confusion, I made the following changes:
1. I renamed this selector to `getMaybeBottommostStickingSourceLineSpec`, so that
    it'll be clearer that this doesn't return the value of `state.sourceCode.focusLineSpec`
     always.
2. I added a doc string to `state.sourceCode.focusLineSpec` to clarify that it is what the
    user has focused on and the UI may choose to focus on a different frame depending
   on other states of the store.

psybuzz: Hm, I see that the selector approach is convenient.  As mentioned, it involves less code, and if some new action X is introduced that can alter the current stack trace, things automatically work.

However, I'm still opposed to it (2), and would prefer the reducer (1) approach for these reasons:

Limits us in the future
- The "current stack frame seen by the user" is accessible by effects and views, but not reducers.  If we ever want to write a reducer that acts based on the current stack frame, it would be impossible(?) without duplicating code.
- For example, how might we build a feature that lets users press `ctrl+[` to move from the current stack frame to its parent?  The current stack frame !== `state.sourceCode.focusedLineSpec`, and we can't use `getMaybeBottommostStickingSourceLineSpec` from the reducer.  We'd have to duplicate the logic in this selector.

Makes the mental model complex
- It is misleading imo to have a `state.sourceCode.focusedLineSpec`, that does not represent the focused line seen by the user.  The meaning of this state property has now become `state.sourceCode.lastLineSpecFocusedFromStackTraceUI` or `state.sourceCode.lineSpecFocusedWhenIgnoringStickToBottom`, both of which are not easy to read or reason about.  The comment added in this PR helps, but does not fix the problem of a misleading name, imo.
- Does dispatching `actions.sourceLineFocused(sourceLineSpec1)` guarantee that `sourceLineSpec1` will become focused in the UI, or does it get overridden by stickToBottom logic in the selector?  I think that currently it is guaranteed to focus the provided line spec, but this 
- A selector named `getMaybeBottommostStickingSourceLineSpec` is hard to reason about.  If a view just wants the currently selected stack frame (as seen by the user), they first need to understand the sticking behavior, which they are not concerned about at all.

I'd prefer avoiding leaking the stickToBottom concept from reducers into selectors and views, if possible.  Duplicating a helper function across the 3 reducer fn's you mentioned seems maintainable, and has the benefit of making it clear to readers that `executionDigestFocused` may affect the current stack frame.

Might it be simple to touching the 3 reducers like so
```
on(executionDigestFocused, (state) => {
  // ...
  return {...state, sourceCode: {..., focusedLineSpec: getFocusedLineSpec(state)}};
}

on(stackFramesLoaded, (state) => {
  // ...
  return {...state, sourceCode: {..., focusedLineSpec: getFocusedLineSpec(state)}};
}

on(graphOpFocused, (state) => {
  // ...
  return {...state, sourceCode: {..., focusedLineSpec: getFocusedLineSpec(state)}};
}

function getFocusedLineSpec(state) {
  // does what getMaybeBottommostStickingSourceLineSpec is doing
}
```
?

caisq:
Thanks, @psybuzz . I just pushed another small commit to set `scrollTop` instead of calling `scrollIntoView()` in order to scroll the stack-frames list. I found `scrollIntoView()` causes the entire screen to scroll when parts of the list is out of the view, which I find less than ideal.

PR #3687, opened 2020-05-31
[DebuggerV2] Improve the scrolling experience of the eager timeline

caisq:
  * Improve the usability of DebuggerV2, specifically it's timeline component.
  * Previously, the mat-slider (see screenshot below) triggers actions on its
    `(change)` output, which means users won't see the eager-execution
    timeline update when dragging the slider thumb. They'll see it change
    only after they release the drag.
  * This PR changes the mat-slider output we listen to from `(change)` to
    `(input)`. This lets the eager-execution timeline update smoothly as
    the user drags the thumb of the mat-slider.
  * Previously, the loading state of all execution digests are lumped into a single
     boolean in the ngrx store, namely `state.executions.executionDigestsLoaded.state`.
     This is not adequate, especially given that the smoother scrolling experience
     makes simultaneous loading of multiple pages of execution digest much more
     likely. 
  * This PR replaces the `state` field with an array of begin-end ranges, .e.,
     `state.executions.executionDigestsLoaded.loadingRanges`. The selectors, reducers,
     effects, and their unit tests are adjusted accordingly.
    * The action `executionDigestsRequested`, which previously had no properties, now
       has a property that describes the requested range of execution digests.
    * This change makes the paged loading of execution digests more consistent with the
      paged loading of graph executions in DebuggerV2.
  * Also in this PR: improve the CSS of the timeline component to avoid line wrapping
     "Executions: <X> - <Y> of <ZZZ>" (see screenshot).
* Screenshots of UI changes
  * image
  * Some existing unit tests affected by this change are updated.
  * Manual testing against a real tfdbg2 logdir (see screenshot above)

### DIFF
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
@@ -85,7 +85,7 @@ import {
   getSourceFileListLoaded,
   getFocusedSourceFileIndex,
 } from '../store/debugger_selectors';
-import {arrayOfBeginEndRangesIncludes} from '../store/debugger_store_utils';
+import {beginEndRangesInclude} from '../store/debugger_store_utils';
 import {
   DataLoadState,
   DebuggerRunListing,
@@ -436,7 +436,7 @@ export class DebuggerEffects {
       filter(([{begin, end}, loaded]) => {
         return (
           end > begin &&
-          !arrayOfBeginEndRangesIncludes(loaded.loadingRanges, begin, end)
+          !beginEndRangesInclude(loaded.loadingRanges, begin, end)
         );
       }),
       tap(([{begin, end}]) => {
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
@@ -354,7 +354,7 @@ export class DebuggerEffects {
         const end = Math.min(numExecutions, pageSize);
         return (
           runId !== null && findRange(loaded.loadingRanges, begin, end) === -1
-        );
+        ); // TODO(cais): Add unit test for already loading.
       }),
       map(([, numExecutions, runId, pageSize]) => {
         const begin = 0;
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
@@ -349,13 +349,7 @@ export class DebuggerEffects {
         this.store.select(getExecutionPageSize),
         this.store.select(getExecutionDigestsLoaded)
       ),
-      filter(([, numExecutions, runId, pageSize, loaded]) => {
-        const begin = 0; // TODO(cais): Deduplicate with below.
-        const end = Math.min(numExecutions, pageSize);
-        return (
-          runId !== null && findRange(loaded.loadingRanges, begin, end) === -1
-        ); // TODO(cais): Add unit test for already loading.
-      }),
+      filter(([, , runId]) => runId !== null),
       map(([, numExecutions, runId, pageSize]) => {
         const begin = 0;
         const end = Math.min(numExecutions, pageSize);
@@ -402,10 +396,6 @@ export class DebuggerEffects {
         }
       ),
       withLatestFrom(this.store.select(getExecutionDigestsLoaded)),
-      filter(([props, loaded]) => {
-        const {begin, end} = props;
-        return findRange(loaded.loadingRanges, begin, end) === -1;
-      }),
       map(([props, loaded]) => {
         return {
           props,
@@ -443,11 +433,16 @@ export class DebuggerEffects {
     }>
   ): Observable<void> {
     return prevStream$.pipe(
-      filter(({begin, end}) => end > begin),
-      tap(({begin, end}) => {
+      withLatestFrom(this.store.select(getExecutionDigestsLoaded)),
+      filter(([{begin, end}, loaded]) => {
+        return (
+          end > begin && findRange(loaded.loadingRanges, begin, end) === -1
+        );
+      }),
+      tap(([{begin, end}]) => {
         this.store.dispatch(executionDigestsRequested({begin, end}));
       }),
-      mergeMap(({runId, begin, end}) => {
+      mergeMap(([{runId, begin, end}]) => {
         return this.dataSource.fetchExecutionDigests(runId, begin, end).pipe(
           tap((digests) => {
             this.store.dispatch(executionDigestsLoaded(digests));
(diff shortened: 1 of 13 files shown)


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -348,9 +349,7 @@ export class DebuggerEffects {
         this.store.select(getExecutionPageSize),
         this.store.select(getExecutionDigestsLoaded)
```
stephanwlee, line 350: This may not be used anymore.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_store_utils.ts
```diff
@@ -34,3 +34,20 @@ export function findFileIndex(
...
 }
+
+/**
+ * Find a range (with begin and end properties) in an array of ranges.
+ * @param range The ranges to search in.
+ * @param begin The begin of the range to search for.
+ * @param end The end of the range to search for.
+ * @returns Index (>=0) if found. -1 if not found.
+ */
+export function findRange(
```
stephanwlee, line 45: Few things about this util.

1. Word `Range` is overloaded as it can mean `Range` in HTML APIs. Prefer more concrete words
2. specify that the bounds are inclusive
3. return values should be boolean if the number greater or equal to zero is meaningless. If you change it, you should also change name of the util since `find` in JS API often mean return an instance (Range in this case)

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/timeline/timeline_component.css
```diff
@@ -63,8 +63,8 @@ limitations under the License.
   height: 48px;
   padding: 0;
   position: absolute;
-  right: 0;
-  width: 440px;
+  right: 40px;
+  width: calc(100% - 420px);
```
stephanwlee, line 67: Write a comment of where this magical `420` is coming from. Also, I personally think `left` is more readable in this case.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers_test.ts
```diff
@@ -676,6 +690,8 @@ describe('Debugger graphs reducers', () => {
     const state = createDigestsStateWhileLoadingExecutionDigests(
       pageSize,
       numExecutions,
+      0 /* Begin of loading range. */,
```
stephanwlee, line 693: nit: Prefer passing dict so the code reads better than describing the arg position with comments like this.

caisq: Done. This creates some inconsistency with other test-helper functions. But I think it's justified in this case because the arg list is indeed getting too long.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/timeline/timeline_component.css
caisq: Done. Using `left` now.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -435,11 +432,17 @@ export class DebuggerEffects {
...
   ): Observable<void> {
     return prevStream$.pipe(
-      filter(({begin, end}) => end > begin),
-      tap(() => {
-        this.store.dispatch(executionDigestsRequested());
+      withLatestFrom(this.store.select(getExecutionDigestsLoaded)),
+      filter(([{begin, end}, loaded]) => {
+        return (
+          end > begin &&
+          findBeginEndRangeIndex(loaded.loadingRanges, begin, end) === -1
```
stephanwlee, line 439: Btw, you may want smarter heuristic than this: 

if `loadingRanges` was [0, 100] and new begin and end were <0, 5>, you don't have to make the request again. Of course, the converse does not hold.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/timeline/timeline_component.css
```diff
@@ -61,10 +61,10 @@ limitations under the License.
 .timeline-slider {
   display: inline-block;
   height: 48px;
+  left: 340px; /* Total width of the two buttons and text div to the left. */
   padding: 0;
   position: absolute;
-  right: 0;
-  width: 440px;
+  width: calc(100% - 420px); /* `left` plus some room on the right. */
```
stephanwlee, line 67: I think there was a miscommunication. I think there should be `left` and `right` and no `width` with calc. If you want to keep it, please document where this magical number of `420` is coming from: (It looks like `left` + 80px of margin)

caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
caisq: Good point. This is relatively minor because the case you mentioned happens only at the end of the eager-execution timeline. But it's easy to implement: a simple change to the `findBeginEndRangeIndex()` suffices. So I did it.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_store_utils.ts
```diff
@@ -34,3 +34,54 @@ export function findFileIndex(
...
+        `but got begin=${begin}, end=${end}`
+    );
+  }
+  return ranges.findIndex(
+    (range) => range.begin === begin && range.end === end
+  );
+}
+
+/**
+ * Detemrine if an array of ranges contains a range that strictly includes
```
stephanwlee, line 64: `Detemrine` > `Determines`

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_store_utils.ts
```diff
@@ -34,3 +34,54 @@ export function findFileIndex(
...
+ * `[begin, end)`.
+ *
+ * @param ranges The ranges to search in.
+ * @param begin The begin of the range to search for (inclusive).
+ * @param end The end of the range to search for (exclusive).
+ * @returns `True` if and only if one of the ranges of `ranges` strictly
+ *   includes `[begin, end)`.
```
stephanwlee, line 71: AFAICT, it's both inclusive, right? i.e., `[begin, end]`

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_store_utils.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_store_utils.ts
caisq: Confirming the ranges are all left-inclusive-right-exclusive. This has been consistent in debugger-v2 code, and consistent with Python indexing as well. Keep in mind that the code `range.begin >= begin && range.end <= end` below is *not* about comparing a range with indices, but about comparing two ranges.

To give an example, if you have `range = [0, 3)` here and `begin = 0, end = 3`, this function should (and does) return
`true`.

PR #3690, opened 2020-06-01
build: fix all `load-on-top` lint warnings

wchargin:
Summary:
We currently suppress the `load-on-top` warning, but the Google-internal
linter still nags us about it. This patch fixes all existing occurrences
so that we can start enforcing the lint in CI in a subsequent patch.

Generated with:

```
git ls-files -z '*BUILD' third_party/js.bzl |
    xargs -0 ~/Downloads/buildifier --warnings=load-on-top --lint=fix
```

where `~/Downloads/buildifier` is at version 3.0.0.

Test Plan:
Running the CI Buildifier command without the `load-on-top` exemption
now passes:

```
git ls-files -z '*BUILD' third_party/js.bzl |
    xargs -0 ~/Downloads/buildifier --mode=check --lint=warn \
        --warnings=-native-py,-native-java,-constant-glob
```

Before this change, with Buildifier 3.0.0, that check failed.

wchargin-branch: build-fix-load-on-top

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,8 @@ on:
   pull_request: {}
 
 env:
-  BUILDIFIER: '0.29.0'
-  BUILDIFIER_SHA256SUM: '4c985c883eafdde9c0e8cf3c8595b8bfdf32e77571c369bf8ddae83b042028d6'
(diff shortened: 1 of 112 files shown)


wchargin:
Test sync passes: <http://cl/314165546>

PR #3696, opened 2020-06-02
data: specify more precise downsampling semantics

### LINKED ISSUE #3436: Require that downsampled reads include the most recent datum
The `DataProvider.read_*` methods admit a downsampling parameter, but
don’t specify its exact interpretation. Many implementations use the
semantics that the most recent datum is always included and the
remaining `k - 1` points are sampled uniformly at random. (This is the
behavior of the legacy event multiplexer, the `MultiplexerDataProvider`,
and some but not all Google-internal data providers.)

Showing the most recent datum is useful because it enables users to
accurately keep track of running experiments without having to wonder
when a recent point will happen to be sampled, which occurs increasingly
rarely as time goes on. It’s also necessary for the hparams plugin,
which reads the last datum to determine the “final” values of metrics,
and is therefore misleadingly wrong if this point is actually sampled
uniformly from the entire time series.

The `DataProvider` interface should be updated to require this, and all
data provider implementations should adapt as necessary.

wchargin:
Summary:
The data provider `read_*` methods have always taken an integer-valued
`downsample` parameter, but the interpretation of that parameter has
hitherto been left to the imagination. We now specify some of the
properties that we expect implementations to satisfy. These properties
(a) are historically supported by TensorBoard’s native event file
reading, (b) are motivated by user experience (i.e., users complain when
these aren’t satisfied), and (c) are relied upon by code that expects
the native TensorBoard logic, and so should be specified in the general
case, too.

Fixes #3436.

wchargin-branch: data-spec-downsampling

### DIFF
--- a/tensorboard/data/provider.py
+++ b/tensorboard/data/provider.py
@@ -73,6 +73,22 @@ class DataProvider(object):
     for 3 random inputs from this batch". A single blob can of course be
     represented as a blob sequence that always has exactly one element.
 
+    When reading time series, *downsampling* refers to selecting a
+    subset of the points in each time series. Downsampling only occurs
+    across the step axis (rather than, e.g., the blobs in a single blob
+    sequence datum), and occurs individually within each time series.
+    When downsampling, the latest datum should always be included in the
+    sample, so that clients have a view of metrics that is maximally up
+    to date. Implementations may choose to force the first (oldest)
+    datum to be included in each sample as well, but this is not
+    required. The remainder of the points in the sample should be
+    selected uniformly at random from available points. Downsampling
+    should be deterministic within a time series. It is also useful for
+    the downsampling behavior to depend only on the set of step values
+    within a time series, such that two "parallel" time series with data
+    at exactly the same steps also retain the same steps after
+    downsampling.
+
     Every time series belongs to a specific experiment and is owned by a
     specific plugin. (Thus, the "primary key" for a time series has four
     components: experiment, plugin, run, tag.) The experiment ID is an
@@ -187,7 +203,8 @@ class DataProvider(object):
           plugin_name: String name of the TensorBoard plugin that created
             the data to be queried. Required.
           downsample: Integer number of steps to which to downsample the
-            results (e.g., `1000`). Required.
+            results (e.g., `1000`). See `DataProvider` class docstring
+            for details about this parameter. Required.
           run_tag_filter: Optional `RunTagFilter` value. If provided, a time
             series will only be included in the result if its run and tag
             both pass this filter. If `None`, all time series will be
@@ -239,7 +256,8 @@ class DataProvider(object):
           plugin_name: String name of the TensorBoard plugin that created
             the data to be queried. Required.
           downsample: Integer number of steps to which to downsample the
-            results (e.g., `1000`). Required.
+            results (e.g., `1000`). See `DataProvider` class docstring
+            for details about this parameter. Required.
           run_tag_filter: Optional `RunTagFilter` value. If provided, a time
             series will only be included in the result if its run and tag
             both pass this filter. If `None`, all time series will be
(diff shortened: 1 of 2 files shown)


tensorboard/data/provider.py
```diff
@@ -187,7 +203,8 @@ def read_scalars(
           plugin_name: String name of the TensorBoard plugin that created
             the data to be queried. Required.
           downsample: Integer number of steps to which to downsample the
-            results (e.g., `1000`). Required.
+            results (e.g., `1000`). See `DataProvider` class docstring
+            for details about this parameter. Required.
```
davidsoergel, line 207: IIUC this is not required.  It defaults to None, meaning "don't downsample" (and similarly below).

tensorboard/data/provider.py
```diff
@@ -73,6 +73,22 @@ class DataProvider(object):
...
 
+    When reading time series, *downsampling* refers to selecting a
+    subset of the points in each time series. Downsampling only occurs
+    across the step axis (rather than, e.g., the blobs in a single blob
+    sequence datum), and occurs individually within each time series.
+    When downsampling, the latest datum should always be included in the
+    sample, so that clients have a view of metrics that is maximally up
+    to date. Implementations may choose to force the first (oldest)
+    datum to be included in each sample as well, but this is not
+    required. The remainder of the points in the sample should be
```
davidsoergel, line 84: Re the oldest datum: if this is useful, we should require it.  If it's not, why mention it at all?  The whole point of a contract is that the caller need know nothing of implementation details.  Also, if that detail is a side effect, so be it, but implementations should not go to extra lengths to add features that are not in the spec.  As we've already seen, that is a recipe for confusion.

tensorboard/data/provider.py
```diff
@@ -73,6 +73,22 @@ class DataProvider(object):
     for 3 random inputs from this batch". A single blob can of course be
     represented as a blob sequence that always has exactly one element.
 
+    When reading time series, *downsampling* refers to selecting a
+    subset of the points in each time series. Downsampling only occurs
+    across the step axis (rather than, e.g., the blobs in a single blob
+    sequence datum), and occurs individually within each time series.
+    When downsampling, the latest datum should always be included in the
```
davidsoergel, line 80: Can you add a TODO to test this at https://github.com/tensorflow/tensorboard/blob/168a9645d9d129267f65e4468c6177ccba65e30f/tensorboard/backend/event_processing/data_provider_test.py#L512 (and ideally the parallel internal places too)?

tensorboard/data/provider.py
wchargin: For clients to be blind to implementation details, we must clearly
define which parts of the interface *are* implementation details and
which are specified. Implementation details generally extend to
“everything not explicitly mandated by the spec”, but sometimes it can
be helpful to call these out explicitly. This is well established
practice (e.g., `java.util.Map` re: null entries; `java.util.Comparable`
re: compatibility with `equals`; Python <3.7 re: dictionary order;
Bazel re: `query` output sorting; basically all the POSIX utility specs;
C/C++ re: struct layout; all the non-normative notes in ECMAScript and C
and C++).

The concrete reason that I mentioned this is that some known data
provider implementations include “newest and `k - 1` uniform” while
others include “newest, oldest, and `k - 2` uniform”. Both of these are
fine. If we omit this sentence (“…may choose…”), then the latter kind of
implementation becomes invalid, because the points other than the newest
are clearly not selected uniformly at random.

If the implication is that the contract must specify every observable
behavior of the interface, then I must disagree. Every such constraint
by definition precludes some implementations, and often includes entire
classes of implementations with desirable properties. See above list for
lots of precedent.

tensorboard/data/provider.py
wchargin: Sure. Done.

tensorboard/data/provider.py
davidsoergel: Ah, gotcha.  Thanks for the explanation.  I was concerned that callers might not know this, so there could be `read_scalars(...)` calls that just don't mention `downsample`, but fortunately there are not.  +1 the PEP 3102 fix.

PR #3697, opened 2020-06-03
[DebuggerV2] Use Map for graph ops

### LINKED ISSUE #3661: [DebuggerV2] Avoid risk of user-generated op name conflicting with built-in JavaScript property names
In the plugins/debugger_v2 code, we are currently storing the graph op-related data in
types such as:

```
opsLoading: {
  [graph_id: string]: {
    [op_name: string]: DataLoadState
  }
}
```

While `graph_id`s are UUIDs and hence do not involve the risk of colliding with builtin
JavaScript object properites such as `toString`, op_names are not UUIDs and hence
do involve such risks.

We have considered using a `Map` in the inner level in #3645, but decided against that
approach for now because `Map`s are not JSON stringifiable and hinders ngrx's
dehydration/rehydration requirements.  

An alternative approach is to use prefixed keys. But it needs to be weighed against
other considerations.

In #3645, we are going with a plain object as typed above, with proper precautions
such that `toString` and similar built-in properties do not cause bugs in the code.

caisq:
  * Fix #3661: graph ops are indexed by TensorFlow op names, which are theoretically
    possible to collide with JavaScript Object's built-in properties such as `toString`
    and `hasOwnProperty`. This PR uses `Map`s to resolve this danger.
  * In debugger_types.ts, change the following two states in the ngrx state from
     `Object` to `Map`:
    * `graphs.ops`
    * `graphs.loadingOps`
  * All related reducer, selector, and effect code is updated.
  * All related unit tests are updated.
  * Made sure tests pass.
  * Manual verification against a real tfdbg2 logdir.

### DIFF
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
@@ -684,10 +684,10 @@ export class DebuggerEffects {
         return (
           runId !== null &&
           (loadingOps[graph_id] === undefined ||
-            loadingOps[graph_id][op_name] === undefined ||
+            !loadingOps[graph_id].has(op_name) ||
             !(
-              loadingOps[graph_id][op_name] === DataLoadState.LOADING ||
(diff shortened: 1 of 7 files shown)


caisq:
FYI @psybuzz

PR #3710, opened 2020-06-05
program: expose public `create_wsgi_app` method

### LINKED ISSUE #3709: Add Tensorboard server to Flask endpoint
I have a Flask App and a Tensorboad server. Is there a way by which I can map the Tensorboard server to one of the endpoints of Flask so that as soon as I hit that endpoint it triggers the Tensorboard server?

Flask application
```
from flask import Flask, jsonify, request
app = Flask(__name__)

@app.route('/hello-world', methods=['GET', 'POST'])
def say_hello():
    return jsonify({'result': 'Hello world'})

if __name__ == "__main__":
    app.run(host=host, port=5000)
```
Tensorboard server code:
```
from tensorboard.program import TensorBoard, setup_environment
def tensorboard_main(host, port, logdir):
    configuration = list([""])
    configuration.extend(["--host", host])
    configuration.extend(["--port", port])
    configuration.extend(["--logdir", logdir])

    tensorboard = TensorBoard()
    tensorboard.configure(configuration)
    tensorboard.main()

if __name__ == "__main__":
    host = "0.0.0.0"
    port = "7070"
    logdir = '/tmp/logdir'
    tensorboard_main(host, port, logdir)
```
I tried creating an endpoint in Flask app and then added tensorboard_main(host, port, logdir) in the hope that if I hit the endpoint then the server will start but I got no luck. 

I had posted this issue on StackOverflow but didn't get a valid reply hence posting it here.

wchargin:
Summary:
The `tensorboard.program` public API permits clients to specify a custom
WSGI server, which takes the TensorBoard WSGI app as an argument. But
the WSGI app is never exposed directly, so clients who want to embed
TensorBoard in a larger WSGI app need to put all their logic inside a
continuation, which is a bit awkward. This patch adds a public method to
directly create and return the WSGI app to simplify such clients.

Fixes #3709.

Test Plan:
In a new virtualenv with `flask` and the Pip packages from this commit,
the following simple program spins up a server with `/hello-world` and
`/tensorboard` routes:

```python
import os

import flask
import tensorboard as tb
from werkzeug import serving
from werkzeug.middleware import dispatcher

flask_app = flask.Flask(__name__)

@flask_app.route("/hello-world", methods=["GET", "POST"])
def say_hello():
    return flask.jsonify({"result": "Hello world"})

program = tb.program.TensorBoard()
program.configure(logdir=os.path.expanduser("~/tensorboard_data"))
tensorboard_app = program.create_wsgi_app()

app = dispatcher.DispatcherMiddleware(
    flask_app, {"/tensorboard": tensorboard_app}
)

if __name__ == "__main__":
    serving.run_simple("0.0.0.0", 7070, app)
```

wchargin-branch: program-create-wsgi-app

### DIFF
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -335,6 +335,33 @@ class TensorBoard(object):
         thread.start()
         return server.get_url()
 
+    def create_wsgi_app(self):
+        """Create a TensorBoard WSGI application.
+
+        The result of this method is a WSGI application (see PEP 3333)
+        that serves the TensorBoard interface. You should call this
+        method directly if you want to compose TensorBoard into a larger
+        WSGI application, e.g., with Werkzeug's `DispatcherMiddleware`.
+        If you're just trying to start a server that only serves
+        TensorBoard, you don't need to call this method; call `main` or
+        `launch` instead.
+
+        This method is heavyweight: e.g., it may start a new thread for
+        data loading. You probably should not call it more than once.
+
+        Returns:
+          A PEP 3333-compliant WSGI callable.
+        """
+        ingester = data_ingester.LocalDataIngester(self.flags)
+        ingester.start()
+        return application.TensorBoardWSGIApp(
+            self.flags,
+            self.plugin_loaders,
+            ingester.data_provider,
+            self.assets_zip_provider,
+            ingester.deprecated_multiplexer,
+        )
+
     def _register_info(self, server):
         """Write a TensorBoardInfo file and arrange for its cleanup.
 
@@ -406,15 +433,7 @@ class TensorBoard(object):
 
     def _make_server(self):
         """Constructs the TensorBoard WSGI app and instantiates the server."""
-        ingester = data_ingester.LocalDataIngester(self.flags)
-        ingester.start()
-        app = application.TensorBoardWSGIApp(
-            self.flags,
-            self.plugin_loaders,
-            ingester.data_provider,
-            self.assets_zip_provider,
-            ingester.deprecated_multiplexer,
-        )
+        app = self.create_wsgi_app()
         return self.server_class(app, self.flags)
 
 


lokesh005:
@nfelt can you please review this PR so that required action can be taken on the same. I am waiting for the PR to merge with the master so that I can use it in one of my use-cases.

wchargin:
It’s the weekend here; we’ll get to it when we can. :-)

nfelt:
I'm a little hesitant to expose this publicly, even though I realize it's not really functionally different from the existing server_class parameter support.

My main argument would be that the program.py API is not really the long-term right API for building TensorBoard WSGI apps (as we just did a lot of refactoring internally to get away from using it as a way to build WSGI apps).  Issue https://github.com/tensorflow/tensorboard/issues/2573 tracks making it easier to use the WSGI app directly, and although we hadn't determined if we wanted that to be a publicly supported API, I think if we're going to provide an option it makes sense to do it more in that vein than via program.py.  That way we have a cleaner path towards making additional decisions around e.g. whether the DataProvider should be parameterizable, and whether there should be side effects like starting the reloader background thread.

On that particular note, IMO one of the things that's a little undesirable about this approach versus the status quo is that it seems potentially surprising that an API to simply "create" the WSGI app also has heavyweight side effects like starting a background thread.  The server_class parameter, while more convoluted to use, has the benefit that it clearly communicates the notion that you are "running" the TensorBoard program, including any side effects or background work it does, and then only customizing the serving of the WSGI app as part of that program.

So given that it is possible to do this today, just not as convenient, I'm reluctant to add an additional API here.  If we see a lot of demand and feel like we can support a more generic WSGI app construction API then my inclination would be to do that via #2573.

wchargin:
Okay, that sounds reasonable to me! Thanks for the second opinion, and
I’m certainly happy to defer to you here given your extensive context.

Issue #3610, opened 2020-05-09
custom_scalar incorrectly toggles fullscreen chart (UI bug)


alshedivat:
## Environment information (required)

Please run `diagnose_tensorboard.py` (link below) in the same
environment from which you normally run TensorFlow/TensorBoard, and
paste the output here:

### Diagnostics

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version 724b56cee52e7d8eb89bbeec1f0d5ce3e38c9682

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=7, micro=3, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): posix.uname_result(sysname='Linux', nodename='Pooh', release='5.3.0-40-generic', version='#32~18.04.1-Ubuntu SMP Mon Feb 3 14:05:59 UTC 2020', machine='x86_64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: False
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
INFO: installed: tensorboard==2.2.1
WARNING: no installation among: ['tensorflow', 'tensorflow-gpu', 'tf-nightly', 'tf-nightly-2.0-preview', 'tf-nightly-gpu', 'tf-nightly-gpu-2.0-preview']
INFO: installed: tensorflow-estimator==2.2.0

--- check: tensorboard_python_version
INFO: tensorboard.version.VERSION: '2.2.1'

--- check: tensorflow_python_version
INFO: tensorflow.__version__: '2.2.0'
INFO: tensorflow.__git_version__: 'v2.2.0-0-g2b96f3662b'

--- check: tensorboard_binary_path
INFO: which tensorboard: b'$HOME/.pyenv/versions/3.7.3/bin/tensorboard\n'

--- check: addrinfos
socket.has_ipv6 = True
socket.AF_UNSPEC = <AddressFamily.AF_UNSPEC: 0>
socket.SOCK_STREAM = <SocketKind.SOCK_STREAM: 1>
socket.AI_ADDRCONFIG = <AddressInfo.AI_ADDRCONFIG: 32>
socket.AI_PASSIVE = <AddressInfo.AI_PASSIVE: 1>
Loopback flags: <AddressInfo.AI_ADDRCONFIG: 32>
[...]

PR #3719, opened 2020-06-08
Fix custom scalars full width mode

stephanwlee:
Cause: due to the limitation of the tf-dom-repeat, we were rendering
custom scalars card inside a container which got in the way of `width:
100%` (since the said container is written to wrap the content, not the
container width).

Fix: tf-dom-repeat used to assign a slot attribute to the DOM we stamp.
However, if the stamped DOM is "dom-if" or "dom-repeat", then DOM will
be dynamically appended to the container _without_ the `slot="items"`.
Without it, the DOM stamped by dom-if does not show up in the items slot
in the container. We fixed it by using MutationObserver and dynamically
appending the slot property to any DOM that is created in the
tf-dom-repeat container.

Test plan:
- tested with the custom scalars demo. 
- made sure other plugins (i.e., scalars and images) did not regress.

### DIFF
--- a/tensorboard/components/tf_paginated_view/tf-dom-repeat.html
+++ b/tensorboard/components/tf_paginated_view/tf-dom-repeat.html
@@ -165,6 +165,20 @@ limitations under the License.
         ) {
           return;
         }
+
+        const observer = new MutationObserver((mutationsList) => {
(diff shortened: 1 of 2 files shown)


stephanwlee:
Fixes #3610

PR #3721, opened 2020-06-09
[DebuggerV2] Fix a few CSS issues in overall layout

caisq:
  * Fix a few CSS issues in DebuggerV2's overall layout
  * The issues fixed include:
    1. The top-section and bottom-section of the overall layout doesn't have
        `box-sizing: border-box;` and hence overflows by the amount of border
        and padding, causing scroll bars to appear. This is fixed by applying the 
        said CSS property.
    2. The top section has three children. They some times wrap undesirably 
        (e.g., when zoom ratio is <100%) This is fixed by applying
        `white-space: nowrap`.
    3. The stack trace component at the lower-right corner doesn't have `height`
        set and hence grows downward beyond the bound. This is fixed by
        `height: 100%`
        * Before: image
        * After: image
    4. The scroll list for stack frames in stack trace component has a width
        that doesn't take into account the padding on the left of the parent element, 
        which causes the scroll bar to go out of bound.
        * Before: image
        * After: image

### DIFF
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/stack_trace/stack_trace_component.css
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/stack_trace/stack_trace_component.css
@@ -42,7 +42,6 @@ limitations under the License.
 }
 
 .stack-frame-array {
-  height: 300px;
   overflow-x: hidden;
   overflow-y: auto;
   /* Minus padding-left of the parent (stack-trace-container). */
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/stack_trace/stack_trace_component.css
```diff
@@ -45,6 +45,8 @@ limitations under the License.
   height: 300px;
   overflow-x: hidden;
   overflow-y: auto;
+  /* Minus padding-left of the parent (stack-trace-container). */
+  width: calc(100% - 8px);
```
stephanwlee, line 49: instead of keeping track of all the padding of parents in mind, just use flex (with justify-content: stretch) where possible.

caisq: Thanks for the suggestion. Normally yes, but StackTraceComponent is already flex in the column (vertical direction), which makes horizontal stretching a little complex. It's less complex to do it this way.

PR #3723, opened 2020-06-09
Roll forward vulcanization printer change with check for outside-head link importing

### LINKED ISSUE #3557: Graph info panel is broken

caisq:
  * Roll forward #3582 with change, in order to fix #3557.
* The solution is primarily due to @stephanwlee. When such midway imports are found,
   a `RuntimeException` is thrown with an error message describing the cause and the
   offending element.

### DIFF
--- a/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
+++ b/tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
@@ -74,13 +74,14 @@ import org.jsoup.nodes.Attribute;
 import org.jsoup.nodes.Comment;
 import org.jsoup.nodes.DataNode;
 import org.jsoup.nodes.Document;
+import org.jsoup.nodes.DocumentType;
 import org.jsoup.nodes.Element;
-import org.jsoup.nodes.Html5Printer;
 import org.jsoup.nodes.Node;
 import org.jsoup.nodes.TextNode;
 import org.jsoup.parser.Parser;
 import org.jsoup.parser.Tag;
 import org.jsoup.select.Elements;
+import org.jsoup.select.NodeVisitor;
 
 /** Simple one-off solution for TensorBoard vulcanization. */
 public final class Vulcanize {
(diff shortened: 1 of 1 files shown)


tensorboard/java/org/tensorflow/tensorboard/vulcanize/Vulcanize.java
```diff
@@ -778,6 +783,125 @@ private static String extractAndTransformJavaScript(Document document, Webpath j
...
+
+    document.traverse(new FlatDocumentCopier(flatDoc));
+
+    for (Element subdoc : flatDoc.getElementsByTag("#root")) {
+      if (subdoc != flatDoc) {
+        final int MAX_ELEMENT_STR_LEN = 200;
+        String elementStr = subdoc.toString();
+        if (elementStr.length() > MAX_ELEMENT_STR_LEN) {
+          elementStr = elementStr.substring(0, MAX_ELEMENT_STR_LEN) + "...";
+        }
```
stephanwlee, line 862: Not sure how useful it would be to see the document as stringified. Perhaps do the `outerHtml()` or `subdoc.parent().outerHTML()`?

caisq: Good point. Done.

caisq:
NOTE: Do not merge yet. Still resolving some internal issues.

PR #3725, opened 2020-06-10
ng: use replaceState for null pluginId to non-null

stephanwlee:
We are, when modifying the history to set the plugin id in the URL,
using replaceState as opposed to pushState on the initial plugin id set.
When done incorrectly, we can pushState and leave awkward intermediate
null plugin id in the URL.

Currently, our logic to use the `firstChange` of Angular's seem flaky
and often result in incorrect state. To remedy, we will replaceState
only when moving from `null` to non-null.

Test:
- tested by launching with `#images` and checked the history stack
- tested by launching with no hash and checked the history stack (after it changing to `#scalars`)

### DIFF
--- a/tensorboard/webapp/core/views/hash_storage_test.ts
+++ b/tensorboard/webapp/core/views/hash_storage_test.ts
@@ -98,7 +98,7 @@ describe('hash storage test', () => {
       {
         id: '',
         option: {
-          useLocationReplace: false,
+          useLocationReplace: true,
           defaultValue: '',
         },
       },
@@ -125,7 +125,7 @@ describe('hash storage test', () => {
     fixture.detectChanges();
 
     expect(setPluginIdSpy).toHaveBeenCalledWith('', {
-      useLocationReplace: false,
+      useLocationReplace: true,
       defaultValue: '',
     });
   });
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/core/views/hash_storage_test.ts
```diff
@@ -64,14 +65,58 @@ describe('hash storage test', () => {
...
+        id: '',
+        option: {
+          useLocationReplace: false,
+          defaultValue: '',
+        },
+      },
+    ]);
   });
 
   it('sets the hash to empty string when activePlugin is not set', () => {
```
nfelt, line 122: Just curious, can you remind me what the actual scenarios are for the follow two tests (where we no longer will use replaceState)?

My current understanding is that having an unset activePlugin would indicate a case where there are no active plugins at all (and it should show the no-dashboards-active message).  So in that case it would seem like there shouldn't be a navigation at all, and I assume pushState() itself would ignore the new state if it's the same URL, right?  So that seems fine.

However, I'm not sure what it means for activePlugin to be the empty string?

stephanwlee: > I assume pushState() itself would ignore the new state if it's the same URL, right? So that seems fine.

That's not strictly correct but it is true because we do `window.location.hash = <newPluginId>` and `hash = ""` is no-op when there is no change.

`activePlugin = null`, activePlugin is not set/bootstrapped yet
`activePlugin = ""`, user explicitly used "localhost:6006/#" 

Upon more careful reflection, I think it is proper to use the replaceState on the first render of the hash_storage_component so I kept the old behavior and modified the test accordingly.

PR #3737, opened 2020-06-14
[DebuggerV2] Introduce data reloading

caisq:
  * Improve DebuggerV2 usability when it takes a long timer for the backend to load
     all the debug data.
  * In debugger_effects.ts, add a `timer` and `repeatWhen` based Observable that drives
     repeating polling
    of "root-level" data, including active run ID, # of executions, # of graph executions, #
    of alerts and list of source-files.
  * This polling backs off when it has been long since polling yielded new data, with a
     minimum and maximum constraint on the time between polls.
  * Additionally, this CL lets debugger-v2 obey the auto reload and manual reload actions sent
    from tensorboard core.
  * Also in this CL:
    * Rename the "Eager Execution Timeline" to "Python Execution Timeline", as per suggestion
       from the tf-core team.
    * Remove the distracting "loading" message from the Python Execution Timeline. Now it's
       repeating, it would be even more distracting if not removed.
  * Unit tests added for the new actions, selectors, reducers and effects.
  * Manual testing on real tfdbg2 data logdirs of small and large sizes.


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors.ts
```diff
@@ -74,6 +74,21 @@ export const getActiveRunId = createSelector(
...
 
+/**
+ * Selectors related to data polling.
+ */
+
+/**
+ * Get the time elapsed from the last time at which a data polling
+ * yielded new data and the last polling time.
+ */
+export const getPollSilenceTime = createSelector(
```
stephanwlee, line 85: `PollSilence` is this a common phrase? It is quite foreign to me and I still don't understand it reading the comment above. 

Also, `lastNewPollDataTime` does not make much sense to me either. Can we do better property/variable naming here?

~Perhaps something like `getDebugTimeSinceMeaningfulData`? (I hope you can change `lastNewPollDataTime` to something like `lastDatafulLoadedTimeInMs`~

I think this selector should change to `getDebugDurationBetweenPollAndNonEmptyDataFetch` (FYI, `getDebug` prefix is important since all selectors ideally should be re-exported at the root level and it avoids collisions).

stephanwlee: Also, please specify the unit (of time) throughout this PR. Without it, code is much less readable.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -487,6 +487,19 @@ export interface DebuggerState {
   runs: DebuggerRunListing;
   runsLoaded: LoadState;
 
+  // Timestamp for the last data polling (including the
+  // initial data-loading event triggered by plugin loading,
+  // and those triggered by TensorBoard's core auto and manual
+  // reloading).
+  lastDataPollTime: number;
```
stephanwlee, line 494: Can you codified the contract? -1 means not loaded yet and this value can be equal to that of `lastNewPollDataTime`.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -487,6 +487,19 @@ export interface DebuggerState {
...
+  // Timestamp for the last non-empty data polling result.
+  // This is updated when any of the following is refreshed:
+  // - activeRunId
+  // - executions.executionDigestsLoaded.numExecutions
+  // - graphExecutions.executionDigestsLoaded.numExecutions
+  // - alerts.alertsLoaded.numAlerts
+  lastNewPollDataTime: number;
```
stephanwlee, line 501: Upon more careful inspection of the reducer code, this literally means `lastNonEmptyDataLoadedTimeInMs`. Can we just say that?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -111,6 +111,8 @@ const initialState: DebuggerState = {
     lastLoadedTimeInMs: null,
   },
   activeRunId: null,
+  lastDataPollTime: -1,
+  lastNewPollDataTime: -1,
```
stephanwlee, line 115: `-1` value here will result in off by one arithmetic on the selector since you are doing `lastDataPollTime` and `lastNewPollDataTime`.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -105,6 +113,68 @@ import {
...
+  minimumPollingInterval = minPollingInterval;
+  maximumPollingInterval = maxPollingInterval;
+  pollingBackoffFactor = backoffFactor;
+}
+
+export function resetDataPollingOptions(): void {
+  minimumPollingInterval = DEFAULT_MINIMUM_POLLING_INTERVAL;
+  maximumPollingInterval = DEFAULT_MAXIMUM_POLLING_INTERVAL;
+  pollingBackoffFactor = DEFAULT_POLLING_BACKOFF_FACTOR;
+}
```
stephanwlee, line 154: unused. remove.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -105,6 +113,68 @@ import {
...
+    throw new Error(
+      `minPollingInterval * backoffFactor ` +
+        `(${minPollingInterval * backoffFactor}) is expected to be < ` +
+        `maxPollingInterval (${maxPollingInterval}), but is not`
+    );
+  }
+  minimumPollingInterval = minPollingInterval;
+  maximumPollingInterval = maxPollingInterval;
+  pollingBackoffFactor = backoffFactor;
+}
```
stephanwlee, line 148: unused, too. Remoe me.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -105,6 +113,68 @@ import {
...
+const DEFAULT_POLLING_BACKOFF_FACTOR = 2;
+
+// Minimum polling interval in milliseconds.
+let minimumPollingInterval = DEFAULT_MINIMUM_POLLING_INTERVAL;
+// Maximum polling interval in miliseconds.
+let maximumPollingInterval = DEFAULT_MAXIMUM_POLLING_INTERVAL;
+// Backoff behavior takes effect when time since the elapsed time between the
+// most recent arrival of new polling result and the most recent polling event
+// exceeds `pollingBackoffFactor * minimumPollingInterval`.
+let pollingBackoffFactor = DEFAULT_POLLING_BACKOFF_FACTOR;
```
stephanwlee, line 127: 1. Please do not use have stateful module
2. These are probably unused after removing the deadcode?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -179,14 +277,38 @@ export class DebuggerEffects {
   /** @export */
   readonly loadData$: Observable<{}>;
 
-  /**
-   * When the debugger plugin is first loaded, request list of runs.
-   */
-  private onDebuggerLoaded() {
+  private onDebuggerLoaded(): Observable<void> {
```
stephanwlee, line 280: `onDebuggerLoaded` -> `onDebuggerDataPoll`

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -171,20 +173,33 @@ const reducer = createReducer(
...
-        activeRunId: runIds.length ? runIds[0] : null,
+        activeRunId: runIds.length > 0 ? runIds[0] : null,
         // TODO(cais): Handle multiple runs. We currently assumes there is only
         // one run, which is okay because the backend supports only one run
         // per experiment.
       };
     }
   ),
+  on(
+    actions.debuggerDataPoll,
```
stephanwlee, line 195: I totally understood this action wrongly when I only read the contract described in the action/type/variable-name (the documentation is really lacking so I went based off of the variable naming). This should read "debuggerDataPollTriggered" and the reducer state should read `lastDataPollTriggerTimeMs`.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
caisq: es, I'm subtracting `lastDataPollTime` from `lastNewPollDataTime`, i.e., `lastDataPollTime - lastNewDataTime`. These initial values are slightly arbitrary, but they serve the following case well.
- Initially, before any polling happens, the difference is 0, which causes the polling interval to be the lower bound, i.e., what we want.

Added a comment here to explain that.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
caisq: Done renaming this field and the related one. I gave it a slightly different (and shorter) name: `lastDataPollOnsetTimeMs`

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_selectors.ts
caisq: I'd argue that poll silence is a nice and short name for the slightly involved context here. It is how long polling has given empty (silent) responses. The name you suggested is precise but a mouthful and hence hard to type, think and say. I initially wrote a function name similar to it but decided against it. I personally believe that sometimes inventing new terminology helps understanding.

Units are added to the new fields and functions in this PR.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
caisq: Added the contract to the doc string here.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
caisq: This is now called `lastNonEmptyPollDataTimeMs`, to be slightly less verbose.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/actions/debugger_actions.ts
```diff
@@ -45,6 +45,12 @@ import {
  */
 export const debuggerLoaded = createAction('[Debugger] Debugger Loaded');
 
+export const debuggerUnloaded = createAction('[Debugger] Debugger Unloaded');
+
+export const debuggerDataPoll = createAction(
```
stephanwlee, line 50: I still think this should change to `debuggerDataPollStarted` or something like that.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -105,6 +113,40 @@ import {
...
+export const MIN_POLLING_INTERVAL_MS = 2e3;
+// Maximum polling interval in miliseconds.
+export const MAX_POLLING_INTERVAL_MS = 60e3;
+// Backoff behavior takes effect when time since the elapsed time between the
+// most recent arrival of new polling result and the most recent polling event
+// exceeds `pollingBackoffFactor * minimumPollingInterval`.
+export const POLLING_BACKOFF_FACTOR = 2;
+
+/**
+ * Get the current polling interval based on the time lapsed since
```
stephanwlee, line 126: code style: jsdoc should start with verb in third person tense, i.e., "Return*s* the current polling interval..."

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -179,14 +249,38 @@ export class DebuggerEffects {
...
+      }),
+      tap(() => this.store.dispatch(debuggerDataPoll())),
+      map(() => void null)
+    );
+  }
+
+  private onCoreReload(): Observable<void> {
+    return this.actions$.pipe(
+      ofType(manualReload, reload),
+      tap(() => this.store.dispatch(debuggerDataPoll())),
```
stephanwlee, line 274: You can decide: Instead, I think the reducers should listen to above actions but I can see how it would be harder to reason when `lastDataPollOnsetTimeMs`. 

Rationale: this reload is really different user's intent that "poll" per say. It should be up to reducer to change the correct state based on actions.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/actions/debugger_actions.ts
caisq: Renamed it to `debuggerDataPollOnset` for consistency.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
caisq: I prefer to keep this way because then all four ways of triggering a data poll in the dashboard will all go through a comment path in terms of triggering `lastDataPollOnsetTimeMs` change: via the action `debuggerDataPoll()`, which the reducer listens to.

PR #3741, opened 2020-06-16
Expose line chart "scaled to fit" state

psybuzz:
VzLineChart2 clients can already set the series data and manually
trigger `resetDomain`. However, it is hard to know when to redraw.

TensorBoard may update data every 30s, so redrawing upon every
data change may disrupt users who are interested in examining a
specific region they panned/zoomed into.

To allow better UX offerings, this exposes internal chart state, so
that VzLineChart2 clients may redraw only when the chart scales
have no transformation applied (e.g. not panned/zoomed).

See also cl/316522213

### DIFF
--- a/tensorboard/components/vz_line_chart2/vz-line-chart2.ts
+++ b/tensorboard/components/vz_line_chart2/vz-line-chart2.ts
@@ -296,23 +296,32 @@ namespace vz_line_chart2 {
     },
 
     /**
-     * Returns whether or not the current scales are fit to the visible data.
-     * When scales do not fit due to transformations (pan, zoom), calling
-     * resetDomain() will make them fit.
+     * Returns whether or not the current scales contain all visible data, after
+     * applying smoothing and outlier detection.
+     *
+     * This is true when there is no data, and false when data exists beyond the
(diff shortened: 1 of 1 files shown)


psybuzz:
Per an offline discussion, a behavior we'd like to try going forward is: when the extent of data fits within the current chart domain, resetDomain upon data updates.

So, the intended use is
- If this method returns true and data updates, then we resetDomain to fit the new data
- Explore showing a visual indicator when this method returns false, indicating that there may be data outside the visible region

> Can we, instead of this change, expose exactly that instead?

Whether the user zoomed is slightly different from knowing whether data is outside the viewport.  e.g. If the user zooms out, hasZoomed !== someDataIsOffscreen

Knowing whether user modified zoom/pan might be useful too, though after spending time reading the code, I don't know how to get this info.  Empirically, it seems that when a scale's `transformationDomain === transformationExtent`, that implies we're at the unaltered default zoom/pan.

psybuzz:
Per another offline discussion:

As you mention, our goal is for this method to indicate whether the user has zoomed/panned.  I've updated the equality checks and method name to reflect this!

PR #3745, opened 2020-06-18
Angular-based Text Plugin v2 backend (experimental)

pjfossee:
**Motivation for features / changes**

- Create backend implementation for new Angular-based text plugin, modifying it to move markdown parsing to the frontend.

**Technical description of changes**
- Forked original text plugin
- modified frontend metadata for angular plugin
- modified HTTP api to send raw markdown to frontend, where parsing will occur

**Screenshots of UI changes**

- No UI implemented yet, although in the end the angular UI should mirror the original Polymer UI fairly closely

**Detailed steps to verify changes work correctly (as executed by you)**

Tensorboard must be built with the `-c dbg` bazel arguments in order to enable the experimental plugins.

Once built and running, this url:

http://localhost:6006/ng_index.html?experimentalPlugin=text_v2#text_v2&run=.

should place the `TEXT_V2` plugin in the plugin dropdown menu, and should result in a blank (empty) plugin when opened (since the frontend is not implemented in angular yet).

Running the `text_demo.py` file and pointing tensorboard towards the run/tag at this link:

http://localhost:6006/data/plugin/text_v2/text?run=.&tag=higher_order_tensors%2Fmultiplication_table

should result in a successful HTTP response with the parameters:
```
    {
        "shape": array
        "step": int
        "string_array": nd_array
        "wall_time": float
        "squashed": boolean
    }
```
for each step in the run

**Alternate designs / implementations considered**

Considerations were made regarding whether to squash an n-dimensional array to 2-dimensions in the backend or the frontend.  While waiting to squash until the frontend may have provided a cleaner API, the added network load of sending arrays with many dimensions was though to be a more significant concern, since there is nothing to stop the user from attempting to upload a 10-dimensional arrays worth of data when only the lowest two dimensions are needed for rendering the visualizations in the frontend.  In the future, if visualizations for higher-dimensional arrays are desired for the text plugin, the API could easily be modified to allow this.

It was also considered whether to create an entirely new text plugin or to subclass the original text plugin.  It was decided to create a new text plugin.  The reasoning is that this plugin is more akin to a sibling of the original text plugin and the inheritance model did not fit the API that well, since most of the original text plugins methods were being overridden in the new plugin anyways.


tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,215 @@
...
+            self.plugin_name, _DEFAULT_DOWNSAMPLING
+        )
+        if context.flags and context.flags.generic_data != "false":
+            self._data_provider = context.data_provider
+        else:
+            self._data_provider = None
+
+    def frontend_metadata(self):
+        return base_plugin.FrontendMetadata(
+            is_ng_component=True, tab_name="Text v2", disable_reload=False
```
stephanwlee, line 141: `disable_reload=False`, there is no reason to disable the reload (unlike the debugger). Enable it :)

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,215 @@
...
+        "wall_time": wall_time,
+        "step": step,
+        "text": formatted_string_array,
+        "shape": shape,
+        "warning": warning,
+    }
+
+
+class TextV2Plugin(base_plugin.TBPlugin):
+    """Angular Text Plugin For Tensorboard"""
```
stephanwlee, line 120: Tensorboard -> TensorBoard

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,215 @@
...
+
+def create_event(wall_time, step, string_ndarray):
+    """Convert a text event into a JSON-compatible response with rank <= 2"""
+    formatted_string_array, shape, warning = reduce_and_jsonify(string_ndarray)
+    return {
+        "wall_time": wall_time,
+        "step": step,
+        "text": formatted_string_array,
+        "shape": shape,
+        "warning": warning,
```
stephanwlee, line 115: warning should be formed and presented in the FE (for internationalization and etc...). Please omit.

You should be able to detect the error by inspecting the shape and the `string_array`. If you think FE should not know about this, then, please change this to a boolean property that reads more like `has_partial_data`.

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,215 @@
...
+      text_arr: A numpy.ndarray containing strings.
+
+    Returns:
+        The JSON-compatible list
+        The shape of the array
+        A warning message (empty string if no warning)
+    """
+    warning = ""
+    if not text_arr.shape:
+        # It is a scalar. No need to put it in a table, just apply markdown
```
stephanwlee, line 97: I don't think you are creating a table. Update the stirng string.

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,215 @@
...
+    message is generated.  The list, shape of the array, and any warning
+    message is returned.
+
+    Args:
+      text_arr: A numpy.ndarray containing strings.
+
+    Returns:
+        The JSON-compatible list
+        The shape of the array
+        A warning message (empty string if no warning)
```
stephanwlee, line 93: You are returning a tuple of these. Say that you are returning a tuple (with these values)

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,215 @@
...
+
+    ndims = len(arr.shape)
+    if ndims < 2:
+        raise ValueError("reduce_to_2d requires an array of dimensionality >=2")
+    # slice(None) is equivalent to `:`, so we take arr[0,0,...0,:,:]
+    slices = ([0] * (ndims - 2)) + [slice(None), slice(None)]
+    return arr[slices]
+
+
+def reduce_and_jsonify(text_arr):
```
stephanwlee, line 76: text_arr -> text_ndarray? (because ndarray != list and it is clearer)

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,215 @@
...
+    return text_arr.tolist(), text_arr.shape, warning
+
+
+def create_event(wall_time, step, string_ndarray):
+    """Convert a text event into a JSON-compatible response with rank <= 2"""
+    formatted_string_array, shape, warning = reduce_and_jsonify(string_ndarray)
+    return {
+        "wall_time": wall_time,
+        "step": step,
+        "text": formatted_string_array,
```
stephanwlee, line 113: `"text"` should probably change to `string_array` or something like that.

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: `disable_reload=False` ======> reload enabled or disabled?  Semantics are a bit confusing here

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: right now there would be no way to detect on the frontend whether the `forrmatted_string_array` was squashed or not (because shape reflects its post-squashing shape, not its pre-squash shape).  I will change warning to a boolean.

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
stephanwlee: Ah, can we also have the raw shape? I think we can have better warning message if we had the original shape, too.

pjfossee:
> Great start! Could you also include the http_api.md so I can read/know about the contract before reading the implementation? Also, ideally, there should be a test with this.

was thinking about adding tests in a seperate PR, but i can work on making a couple and throw them in here

tensorboard/webapp/plugins/text/text_v2_plugin.py
stephanwlee: Also, the field "warning" is quite ambiguous. I know the http_api is supposed to help with that but having self descriptive name would be better.

tensorboard/webapp/plugins/text/http_api.md
```diff
@@ -0,0 +1,40 @@
...
+a dictionary with members `wall_time`, `step`, `text`, `shape`, and `warning`,
+where `wall_time` is a floating-point number of seconds since epoch, `step` is
+an integer step counter, `string` is an n-dimensional array, `shape` is an array
+of the sizes of each dimension, and `warning` is a boolean indicating whether
+the dimensionality of the array was reduced.
+
+Example:
+
+  {
+    shape: [2, 2]
```
stephanwlee, line 35: A real JSON has `"` on the key, too. :)

pjfossee: ope will fix.... done

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: sounds good, will include raw shape and will rename the `warning` field to something more descriptive

pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: resolved

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,209 @@
...
+
+    If the ndarray contains a single scalar string, then that ndarray is
+    converted to a list. If it contains an array of strings,
+    that array is converted to a list.  If the array contains dimensionality
+    greater than 2, all but two of the dimensions are removed, and a squashed
+    boolean is set to true.  Returned is a list, the shape of the original
+    array, and a boolean indicating squashsing has occured.
+
+    Args:
+      text_arr: A numpy.ndarray containing strings.
```
stephanwlee, line 82: nit: inconsistent whitespace from below returns block. use +4 whitespace.

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,209 @@
...
+
+    Returns:
+        a tuple containing:
+            The JSON-compatible list
+            The shape of the array
```
stephanwlee, line 87: indicate which shape this is referring to. I think now it refers to the original ndarray shape.

tensorboard/webapp/plugins/text/http_api.md
```diff
@@ -0,0 +1,40 @@
...
+      "higher_order_tensors/multiplication_table"
+    ]
+  }
+
+Runs without any text tags are omitted from the result.
+
+## `/data/plugin/text_v2/text?run=foo&tag=bar`
+
+Returns an array of text events for the given run and tag.  Each event is
+a dictionary with members `wall_time`, `step`, `text`, `shape`, and `warning`,
```
stephanwlee, line 26: update the `warning` part.

tensorboard/webapp/plugins/text/http_api.md
```diff
@@ -0,0 +1,40 @@
...
+where `wall_time` is a floating-point number of seconds since epoch, `step` is
+an integer step counter, `string` is an n-dimensional array, `shape` is an array
```
stephanwlee, line 28: `string` -> `string_array`?

tensorboard/webapp/plugins/text/http_api.md
```diff
@@ -0,0 +1,40 @@
...
+of the sizes of each dimension, and `squashed` is a boolean indicating whether
```
stephanwlee, line 29: better define what the `shape` is referring to :)

tensorboard/default.py
```diff
@@ -52,7 +52,7 @@
 from tensorboard.plugins.scalar import scalars_plugin
 from tensorboard.plugins.text import text_plugin
 from tensorboard.plugins.mesh import mesh_plugin
-
```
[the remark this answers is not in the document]
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: done

tensorboard/webapp/plugins/text/http_api.md
pjfossee: done

tensorboard/webapp/plugins/text/http_api.md
pjfossee: done

tensorboard/webapp/plugins/text/http_api.md
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,209 @@
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
stephanwlee, line 1: Sorry for not catching this earlier. Could you move the plugin python files to `tensorboard/plugins`? So it should end up in `tensorboard/plugins/text_v2`. The rationale is, sometimes, the plugins contain Python API that people use from the TensorFlow side. As such, it is likely that we will never move the plugin Python files out of the `tensorboard/plugins` while FE code have a bit more leeway. 

I think we are at a point where the benefit colocating FE code is stronger so we want the Angular frontend code to be in `tensorboard/webapp` but the Python counterpart should stay in the `tensorboard/plugins`.

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,209 @@
...
+
+# HTTP routes
+TAGS_ROUTE = "/tags"
+TEXT_ROUTE = "/text"
+
+_DEFAULT_DOWNSAMPLING = 100  # text tensors per time series
+
+
+def reduce_to_2d(arr):
+    """Given a np.npdarray with nDims > 2, reduce it to 2d.
```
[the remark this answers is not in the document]
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,209 @@
...
+        """Instantiates Angular TextPlugin via TensorBoard core.
+
+        Args:
+          context: A base_plugin.TBContext instance.
+        """
+        self._multiplexer = context.multiplexer
+        self._downsample_to = (context.sampling_hints or {}).get(
+            self.plugin_name, _DEFAULT_DOWNSAMPLING
+        )
+        if context.flags and context.flags.generic_data != "false":
```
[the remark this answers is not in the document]
pjfossee: done (maybe check the isActive() function, with this logic, we just return false always (but per the comment in the function this shouldn't matter)

tensorboard/webapp/plugins/text/http_api.md
```diff
@@ -0,0 +1,43 @@
...
+
+## `/data/plugin/text_v2/text?run=foo&tag=bar`
+
+Returns an array of text events for the given run and tag.  Each event is
+a dictionary with members `wall_time`, `step`, `string_array`, `shape`, and `squashed`,
+where `wall_time` is a floating-point number of seconds since epoch, `step` is
+an integer step counter, `string_array` is an n-dimensional array, `shape` is the
+size of each dimension in the tensor provided to the text-summarizer (Note this will
+be different from the shape of `string_array` if squashing has occured), and
+`squashed` is a boolean indicating whether the dimensionality of the array
```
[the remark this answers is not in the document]
pjfossee: done

tensorboard/webapp/plugins/text/http_api.md
```diff
@@ -0,0 +1,43 @@
...
+```
+Runs without any text tags are omitted from the result.
+
+## `/data/plugin/text_v2/text?run=foo&tag=bar`
+
+Returns an array of text events for the given run and tag.  Each event is
+a dictionary with members `wall_time`, `step`, `string_array`, `shape`, and `squashed`,
+where `wall_time` is a floating-point number of seconds since epoch, `step` is
+an integer step counter, `string_array` is an n-dimensional array, `shape` is the
+size of each dimension in the tensor provided to the text-summarizer (Note this will
```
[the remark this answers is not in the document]
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,209 @@
...
+import textwrap
+
+# Necessary for an internal test with special behavior for numpy.
+import numpy as np
+
+import six
+from werkzeug import wrappers
+
+from tensorboard import plugin_util
+from tensorboard.util import tensor_util
```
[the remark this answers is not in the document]
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
```diff
@@ -0,0 +1,209 @@
...
+"""Angular Version of Text Plugin."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import textwrap
+
+# Necessary for an internal test with special behavior for numpy.
+import numpy as np
```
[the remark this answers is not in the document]
pjfossee: done

tensorboard/plugins/text_v2/text_v2_plugin_test.py
```diff
@@ -0,0 +1,230 @@
...
+tf.compat.v1.disable_v2_behavior()
+
+GEMS = ["garnet", "amethyst", "pearl", "steven"]
+
+
+class TextPluginTest(tf.test.TestCase):
+    def setUp(self):
+        self.logdir = self.get_temp_dir()
+
+    def with_plugin(generate_testdata=True, include_text=True):
```
nfelt, line 48: The main point of this decorator was to support testing across both the bare multiplexer and the generic data provider; since we're not copying the support for the bare multiplexer, this decorator isn't adding any value here.

I'd just convert this into a regular test helper method that returns the initialized plugin:

```python
def create_plugin(generate_testdata=True, include_text=True):
    ...

# example usage
def testRoutesProvided(self):
    plugin = self.create_plugin()
    routes = plugin.get_plugin_apps()
    ...
```

You can at that point also remove the `with self.subTest()` call, since it was only useful when there were multiple calls to subTest but now there's just one.

tensorboard/plugins/text_v2/text_v2_plugin_test.py
```diff
@@ -0,0 +1,230 @@
...
+    def testText(self, plugin):
+        fry = plugin.text_impl("fry", "message", experiment="123")
+        leela = plugin.text_impl("leela", "message", experiment="123")
+        self.assertEqual(len(fry), 4)
+        self.assertEqual(len(leela), 4)
+        for i in range(4):
+            self.assertEqual(fry[i]["step"], i)
+            self.assertEqual(leela[i]["step"], i)
+
+        table = plugin.text_impl("fry", "vector", experiment="123")[0][
```
nfelt, line 136: Rather than calling `text_impl()` several times, a bit more compact to just do it like this:

```python
result = plugin.text_impl("fry", "vector", experiment="123")[0]
self.assertEqual(result["string_array"], [b"one", b"two", b"three", b"four"])
self.assertEqual(result["original_shape"], (4,))
...
```

Also I probably wouldn't name this variable `table` since while that made sense for the old code when it had been converted into a `<table>` already, in this case it's less clear.

tensorboard/plugins/text_v2/text_v2_plugin_test.py
```diff
@@ -0,0 +1,230 @@
...
+                [["foo", "bar"], ["zoink", "zod"]],
+                [["FOO", "BAR"], ["ZOINK", "ZOD"]],
+            ]
+        )
+
+        d3_expected = ([["foo", "bar"], ["zoink", "zod"]], (2, 2, 2), True)
+        self.assertEqual(convert(d3), d3_expected)
+
+    @with_plugin(generate_testdata=False)
+    def testPluginIsActiveWhenNoRuns(self, plugin):
```
nfelt, line 206: There's no real need for the three different IsActive tests now that they're all identical code, and the names are not really accurate at this point.

I'd just replace them all with the following:

```py
def test_is_active_returns_false(self):
    # `list_plugins` as called by TB core suffices
    self.assertFalse(plugin.is_active())

File: tensorboard/plugins/text_v2/text_v2_plugin_test.py [hunk 25]
```diff
@@ -0,0 +1,230 @@
...
+        self.assertFalse(plugin.is_active())
+
+    @with_plugin(include_text=False)
+    def testPluginIsActiveWhenRunsButNoText(self, plugin):
+        """The plugin should be inactive when there are runs but none has
+        text."""
+        self.assertFalse(plugin.is_active())
+
+    @with_plugin()
+    def testPluginIndexImpl(self, plugin):
```
nfelt, line 222: This test looks to be identical to `testIndex()` - I think it can just be deleted.

tensorboard/plugins/text_v2/text_v2_plugin_test.py
```diff
@@ -0,0 +1,230 @@
...
+            np.testing.assert_array_equal(actual, expected)
+
+    def test_reduce_and_jsonify(self):
+        convert = text_v2_plugin.reduce_and_jsonify
+        scalar = np.array("foo")
+        scalar_expected = ("foo", (), False)
+        self.assertEqual(convert(scalar), scalar_expected)
+
+        # Check that underscores are preserved correctly; this detects erroneous
+        # use of UTF-16 or UTF-32 encoding when calling markdown_to_safe_html(),
```
nfelt, line 178: We're no longer calling `markdown_to_safe_html()` so this comment (and the associated test logic) is obsolete; I would just remove this and the following three lines.

tensorboard/webapp/plugins/text/text_v2_plugin.py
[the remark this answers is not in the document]
nfelt: Yeah, that's expected that it would always return false now.  There are a few other plugins that do this:
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/plugins/pr_curve/pr_curves_plugin.py;l=170;drc=ba302fc490c014816068b7e4b3fd9d810e259cd5

Basically returning `False` here just means that the TensorBoard core logic is allowed to infer if the plugin is active or not, rather than asking the plugin directly.

tensorboard/plugins/text_v2/BUILD
```diff
@@ -22,3 +23,23 @@ py_library(
...
+    name = "text_v2_plugin_test",
+    size = "small",
+    srcs = ["text_v2_plugin_test.py"],
+    main = "text_v2_plugin_test.py",
+    srcs_version = "PY2AND3",
+    deps = [
+        ":text_v2_plugin",
+        "//tensorboard:expect_tensorflow_installed",
+        "//tensorboard:plugin_util",
+        "//tensorboard/backend:application",
```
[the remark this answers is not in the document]
pjfossee: done

tensorboard/plugins/text_v2/BUILD
```diff
@@ -22,3 +23,23 @@ py_library(
...
+        "//tensorboard/backend/event_processing:data_provider",
+        "//tensorboard/backend/event_processing:event_multiplexer",
+        "//tensorboard/plugins:base_plugin",
+        "//tensorboard/util:test_util",
+        "@org_pocoo_werkzeug",
```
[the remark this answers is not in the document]
pjfossee: done

tensorboard/plugins/text_v2/text_v2_plugin_test.py
pjfossee: done

tensorboard/plugins/text_v2/text_v2_plugin_test.py
pjfossee: done

tensorboard/plugins/text_v2/text_v2_plugin_test.py
pjfossee: done

tensorboard/plugins/text_v2/text_v2_plugin_test.py
pjfossee: done

tensorboard/plugins/text_v2/text_v2_plugin_test.py
pjfossee: done

tensorboard/webapp/plugins/text/text_v2_plugin.py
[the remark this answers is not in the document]
pjfossee: thanks for the explanation!

tensorboard/plugins/text_v2/text_v2_plugin_test.py
```diff
@@ -202,29 +174,11 @@ def test_reduce_and_jsonify(self):
         d3_expected = ([["foo", "bar"], ["zoink", "zod"]], (2, 2, 2), True)
         self.assertEqual(convert(d3), d3_expected)
 
-    @with_plugin(generate_testdata=False)
-    def testPluginIsActiveWhenNoRuns(self, plugin):
+    def testIsActiveReturnsFalse(self):
         """The plugin should be inactive when there are no runs."""
```
nfelt, line 178: The comment here is a bit stale since it's no longer the case that the plugin's inactive status depends on whether there are run or not.  I'd either remove it entirely, or replace it with `# `list_plugins` as called by TB core suffices` as commented in the code itself, or something else that indicates that it's intended to always return false.

tensorboard/plugins/text_v2/text_v2_plugin_test.py
```diff
@@ -0,0 +1,184 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
```
stephanwlee, line 2: This is a forked file so maybe you it should have `2020`?

tensorboard/plugins/text_v2/text_v2_plugin_test.py
```diff
@@ -0,0 +1,184 @@
...
+
+        ctx = base_plugin.TBContext(
+            logdir=self.logdir, multiplexer=multiplexer, data_provider=provider,
+        )
+
+        return text_v2_plugin.TextV2Plugin(ctx)
+
+    def generate_testdata(self, include_text=True, logdir=None):
+        tf.compat.v1.reset_default_graph()
+        sess = tf.compat.v1.Session()
```
stephanwlee, line 69: This is still v1 style. Did you consider using the v2/eager style for generating the data?

tensorboard/plugins/text_v2/text_v2_plugin_test.py
pjfossee: got it, good catch thanks, done

tensorboard/plugins/text_v2/text_v2_plugin_test.py
pjfossee: done

tensorboard/plugins/text_v2/text_v2_plugin_test.py
pjfossee: attempted but has issues when running bazel test (vs just running straight from python)
```
    tf.summary.text("message", message, step)
AttributeError: module 'tensorflow._api.v2.summary' has no attribute 'text'
```

pjfossee: fixed, done

PR #3757, opened 2020-06-24
cleanup: remove unused :application deps

nfelt:
Noticed some vestigial unused deps on `//tensorboard/backend:application` in various tests that no longer import it.

Generated with:

```
for label in $(bazel query 'rdeps(//..., //tensorboard/backend:application, 1)' --output label 2>/dev/null); do \
    src="$(bazel query 'labels(srcs, '"${label}"')' 2>/dev/null | cut -c3- | tr ':' '/')"; \
    if ! grep -q 'import.*application' "${src}"; then \
        buildozer 'remove deps //tensorboard/backend:application' "${label}"; \
    fi; \
done
```


nfelt:
Yeah, it was vestigial from when I had been passing `--output build` earlier in iteration.

Issue #3752, opened 2020-06-23
[RFC] Extending the Tensorboard Notebook Display API


jerrylian-db:
Three possible enhancements are proposed below to improve the Tensorboard experience on various notebooks environments. Please add your comments or concerns.

## Custom Display API
A [recent pull request](https://github.com/tensorflow/tensorboard/pull/3674) made it possible for users to run Tensorboard in environments where the Jupyter host machine is not directly exposed to the notebook user. We are proposing to make the notebook display API more flexible.

The proposed enhancement is an API that lets users to directly set an inline display function that works in their environment, providing more flexibility for Tensorboard to be used on various  notebook environments.

Here is an example use-case of this API.
```python
# In the notebook Python shell init script
def my_custom_display(port, height, display_handle)
    del display_handel
    html_template = …
    display_function(html_template) # this function is specific to the cloud environment

# import hook
Import wrapt
@wrapt.when_import(“tensorboard.notebook”)
def hook_tensorboard(module):
    module.register_custom_display(my_custom_display)

# In notebook cells
%load_ext tensorboard

%tensorboard --logdir logs
# uses the custom display function, which successfully renders the Tensorboard UI inline 
```

## Flag to prevent process reuse
Process reuse is a convenient feature when a single user runs Tensorboard on their personal machine. However, in multi-tenant environments where users share the same machine or container, reusing processes can lead to mutual interference. One possible scenario: user A starts a process on a log directory and user B starts tensorboard on the same directory, hence reusing user A's process. User B kills the Tensorboard process and user A's Tensorboard UI becomes stale (to user A’s surprise).

By adding a flag to Tensorboard that prevents process re-use, e.g.`--new_process`, users can set that flag by default to prevent interfering with each other on a multi-tenant environment

## Increasing the limit for the number of concurrent processes
Currently, there is a default and hard-coded limit of 10 concurrent Tensorboard processes in the port selection logic. For some environments, e.g. multi-tenant environments with powerful computing resources this is too restrictive. We propose increasing this default limit to 100.

wchargin:
Hi @jerrylian-db! Thanks for reaching out. A couple initial comments
inline:

> ```python
> @wrapt.when_imported("tensorboard.notebook")
> ```

With the understanding that this is just an example, in most cases we
would probably recommend just importing `tensorboard.notebook` and
registering the hook directly. The module is lightweight and has few
dependencies (in particular, no TensorFlow import).

> ```python
> def hook_tensorboard(module):
[...]

PR #3770, opened 2020-06-29
Custom display API

jerrylian-db:
    - This feature lets users to directly set an inline display function for the notebook API that works in their environment, providing more flexibility for Tensorboard to be used on various notebook environments. See RFC https://github.com/tensorflow/tensorboard/issues/3752 for more details.

    - Added a public API to the notebook module that allows users to set a custom display function.
    -  In `_get_context` ([link](https://github.com/jerrylian-db/tensorboard/blob/ed147694346da9c1553c2da1a705cc03c38b3298/tensorboard/notebook.py#L60)), we add a context type that is returned when the library detects that the custom display function is set. This new context type then causes the `_display` function ([link](https://github.com/jerrylian-db/tensorboard/blob/ed147694346da9c1553c2da1a705cc03c38b3298/tensorboard/notebook.py#L283)) to use the custom display function to display Tensorboard inline to python notebooks.

* Screenshots of UI changes
    - N/A

    1. Clone the repo.
    2. In the repo root directory, start a Jupyter notebook process and create a notebook inside of it. (Make sure that your environment has Tensorboard installed as the following code depends on having a Tensorboard binary in the environment.)
    3. Run the following code..
```python
import tensorboard
import random, IPython
def custom_display(port, height, display_handle):
    del display_handle

    frame_id = "tensorboard-frame-{:08x}".format(random.getrandbits(64))
    html_string = """
        <h1>Custom Inline Display</h1>
        <iframe id="%ID%" width="100%" height="%HEIGHT%" frameborder="0"></iframe>
        <script>
            (function() {
                const frame = document.getElementById("%ID%");
                const url = new URL("/", window.location);
                const port = %PORT%;
                if (port) {
                  url.port = port;
                }
                frame.src = url;
            })();
        </script>
    </html>
    """

    replacements = [
        ("ID%", frame_id),
        ("%HEIGHT%", "%d" % height),
        ("%PORT%", "%d" % port),
    ]

    for (k, v) in replacements:
        html_string = html_string.replace(k, v)
    iframe = IPython.display.HTML(html_string)
    IPython.display.display(iframe)
tensorboard.notebook.register_custom_display(custom_display)
tensorboard.notebook.start("--logdir test")
```
Result:
Screen Shot 2020-06-29 at 3 49 40 PM

    - We considered directly modifying the `_display` function to return the custom display if it detects it is set, rather than modifying `_get_context`. But we thought that the presented solution matches existing patterns better.

### DIFF
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -53,9 +53,11 @@ _CONTEXT_CUSTOM = "_CONTEXT_CUSTOM"
 
 _custom_display = None
 
+
 def register_custom_display(func):
     global _custom_display
     _custom_display = func
+
 
 def _get_context():
     """Determine the most specific context that we're in.
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -55,7 +55,7 @@ _custom_display = None
 
 
 def register_custom_display(func):
-    """Register a custom inline display function
+    """Register a custom function to display a running TensorBoard instance.
 
     Args:
       func: a function that takes in the following parameters and renders the
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -54,6 +54,7 @@ _CONTEXT_CUSTOM = "_CONTEXT_CUSTOM"
 _custom_display = None
 
 def register_custom_display(func):
+    global _custom_display
     _custom_display = func
 
 def _get_context():
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -49,12 +49,18 @@ from tensorboard import manager
 _CONTEXT_COLAB = "_CONTEXT_COLAB"
 _CONTEXT_IPYTHON = "_CONTEXT_IPYTHON"
 _CONTEXT_NONE = "_CONTEXT_NONE"
+_CONTEXT_CUSTOM = "_CONTEXT_CUSTOM"
 
+_custom_display = None
+
+def register_custom_display(func):
+    _custom_display = func
 
 def _get_context():
     """Determine the most specific context that we're in.
 
     Returns:
+      _CONTEXT_CUSTOM: If user has registered a custom display function.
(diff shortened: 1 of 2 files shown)


jerrylian-db:
Note: I did not find any automated tests for the `notebook.py` file. Please tell me if and how I should add test coverage for this feature.

jerrylian-db:
@googlebot I signed it!

tensorboard/notebook.py
```diff
@@ -145,7 +158,7 @@ def start(args_string):
     except ImportError:
         IPython = None
 
-    if context == _CONTEXT_NONE:
+    if context == _CONTEXT_NONE or context == _CONTEXT_CUSTOM:
```
falaki, line 161: Why should we print `Launching TensorBoard...` when custom display function has been registered?

tensorboard/notebook.py
```diff
@@ -49,19 +49,32 @@
 _CONTEXT_COLAB = "_CONTEXT_COLAB"
 _CONTEXT_IPYTHON = "_CONTEXT_IPYTHON"
 _CONTEXT_NONE = "_CONTEXT_NONE"
+_CONTEXT_CUSTOM = "_CONTEXT_CUSTOM"
+
+_custom_display = None
+
+
+def register_custom_display(func):
```
falaki, line 57: Nit: add some documentation for this function.

falaki: Specifically mention what is the signature of `func` argument.

tensorboard/notebook.py
jerrylian-db: This tells the user that the Tensorboard process is launching. It is standard behavior across the different python environments.

jerrylian-db:
@wchargin can you take a look at this? Thanks for your time.

wchargin:
@jerrylian-db yep—responded on the RFC:
<https://github.com/tensorflow/tensorboard/issues/3752#issuecomment-662622316>

PR #3773, opened 2020-06-30
Add "Upload" button to header for local users

bmd3k:

  Show the "Upload" button to users who load a TensorBoard on localhost. This Upload button is an entry point to the "Upload to TensorBoard.dev" dialog with instructions on how to create a TensorBoard.dev experiment with the event data.  We restrict to localhost users so we can be more certain that the user is the owner of the TensorBoard.

  In the header, check if window.location.hostname is a "localhost" or "127.0.0.1" value.  If so, show the button.

  Also, improve the styling of the button.

  Also, remove "?" from the "Upload to TensorBoard" dialog.

* Screenshots of UI changes

  On localhost:8080 or 127.0.0.1:
  image

  On non-localhost:
  image

  * Write some tests.
  * Manually verify:
    * Edit hosts file so www.nytimes.com points to localhost.
    * Load ng_index.html using localhost:<port>, 127.0.0.1:<port>, and www.nytimes.com:<port>
    * Observe in the the first two instances that the "Upload" button appears, with new styling.
    * Observce in the final instance that the "Upload" button does not appear.


tensorboard/webapp/header/header_component.ts
```diff
@@ -62,4 +71,12 @@ import {Component} from '@angular/core';
...
   ],
 })
-export class HeaderComponent {}
+export class HeaderComponent implements OnInit {
+  showUpload: boolean = false;
+
+  constructor(@Inject('window') private window: Window) {}
+
+  ngOnInit() {
+    this.showUpload = LOCAL_HOSTNAMES.includes(this.window.location.hostname);
```
bmd3k, line 80: My preference would actually have been to push this logic into the tbdev-upload-button component while keeping the "margin" css value in this file.  However, I don't understand how I would be able to conditionally apply the "margin" in this file based on whether tbdev-upload-button implementation decides to render or not.

stephanwlee: Yes, you should definitely move this business logic into the upload-button file. We can talk about the margin offline.

bmd3k: Offline you showed me how we could do this with @HostBinding.  So now the logic that determines whether to show the button is in the tbdev-upload-button component itself while the styling of margins when the button is shown still exists in the app-header.

tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
```diff
@@ -32,15 +36,25 @@ import {TbdevUploadDialogComponent} from './tbdev_upload_dialog_component';
...
     `,
   ],
 })
-export class TbdevUploadButtonComponent {
-  constructor(private dialog: MatDialog) {}
+export class TbdevUploadButtonComponent implements OnInit {
+  @HostBinding('class.shown') shown: boolean = false;
+
+  constructor(
+    @Inject('window') private window: Window,
```
stephanwlee, line 51: private readonly

tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
```diff
@@ -32,15 +36,25 @@ import {TbdevUploadDialogComponent} from './tbdev_upload_dialog_component';
...
+    private dialog: MatDialog
+  ) {}
+
+  ngOnInit() {
+    this.shown = LOCAL_HOSTNAMES.includes(this.window.location.hostname);
```
stephanwlee, line 56: this can be called in the constructor, I think. Does it need to happen in the OnInit? `OnInit` is the lifecycle method when you are guaranteed to have all `@Input()` passed but does not really buy you much.

tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
bmd3k: Done.

tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
bmd3k: Yes it seems to work in the constructor. I thought I might need to use ngOnInit to ensure the reference to `window` works but I guess it makes sense that it should be available in the constructor.

tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
```diff
@@ -40,7 +45,14 @@ import {TbdevUploadDialogComponent} from './tbdev_upload_dialog_component';
   ],
 })
 export class TbdevUploadButtonComponent {
-  constructor(private dialog: MatDialog) {}
+  @HostBinding('class.shown') shown: boolean = false;
+
+  constructor(
+    @Inject('window') private readonly window: Window,
```
stephanwlee, line 51: after the change, the `window` object is no longer used so we can remove `private readonly` and just use it as a local variable.

tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
```diff
@@ -40,7 +45,14 @@ import {TbdevUploadDialogComponent} from './tbdev_upload_dialog_component';
   ],
 })
 export class TbdevUploadButtonComponent {
-  constructor(private dialog: MatDialog) {}
+  @HostBinding('class.shown') shown: boolean = false;
```
stephanwlee, line 48: Since the `shown` is initialized at the constructor, we can remove the value here.

bmd3k: Done.

tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
bmd3k: Done.

ericdnielsen:
Reverting the PR as it break downstream workflows. Has been discussed with the author.

PR #3779, opened 2020-07-01
Change default FE to Angular TB

stephanwlee:
This commit changes the default route (e.g., localhost:6006/) to the
Angular based TensorBoard.

More an entire minor version, we intend to keep the legacy build around
as an escape hatch.

- This PR will be submitted only after July 6th for the tb-nightly on July 7th.
- Propose this change to make it into TensorBoard 2.3.0 release.
- We should cut TB 2.3.0 release by July 17th (should confirm the real date with TF).

WANT_LGTM=all+date

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -276,12 +276,6 @@ py_test(
     ],
 )
 
-config_setting(
-    name = "dev_build",
-    # consider changing this to fastbuild
-    values = {"compilation_mode": "dbg"},
-)
-
 tensorboard_zip_file(
     name = "webfiles",
     deps = [":assets"],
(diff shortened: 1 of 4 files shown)


tensorboard/BUILD
```diff
@@ -296,23 +290,16 @@ monaco_imports = [
...
+        "//tensorboard/webapp:svg_bundle",
+    ],
     path = "/",
     deps = [
         "@com_google_fonts_roboto",
-    ] + select({
-        "//tensorboard:dev_build": monaco_imports,
-        "//conditions:default": [],
-    }),
+    ] + monaco_imports,
```
nfelt, line 302: Any particular reason not to just inline the monaco imports at this point?

tensorboard/BUILD
```diff
@@ -296,23 +290,16 @@ monaco_imports = [
 tf_web_library(
```
nfelt, line 290: Can eliminate the TODO on the line above now, right?   🎉     🚀     🙈

PR #3780, opened 2020-07-01
program: increase port scan limit from 10 to 100

jerrylian-db:
    - Increase the port scanning range from 10 to 100 so that more users can start Tensorboard processes with default parameters in multi-tenant notebook environments.
    - See https://github.com/tensorflow/tensorboard/issues/3752 for more context.


wchargin:
(I’m adjusting the title and description to convey that this is actually
the main launching code path, which affects `tensorboard(1)` as well as
the notebook APIs.)

PR #3782, opened 2020-07-02
scalars: set data download filenames on backend, not `a` tag

nfelt:
Scalar chart data download URLs use custom filenames that include the user's run and tag names, for ease of identification when downloading from multiple charts.  This PR moves that custom filename support from the frontend, where it used the `download` attribute on `<a>` tags, to the backend, where we use the `Content-Disposition` header.

The intent is to sidestep issues with trying to sanitize the strings going into the `download` attribute so that they aren't intercepted by Polymer Resin; the easiest way I determined to do that was to stop binding to the attribute entirely.  It's a little less elegant in that if we had multiple download widgets for various charts, now each backend location must set the filename rather than having it centralized in the UI component, but right now it's only used in one place anyway and it doesn't seem too onerous even if we had multiple places.

In addition, I actually still implemented some sanitization of the attribute, to eliminate or replace with `_` all characters other than `[-_=A-Za-z0-9]`.  Chrome's auto-downloading behavior already seems to do something fairly similar (using `_` to replace `/`, for instance) but it seemed safest to ensure we're generating innocuous filenames in the first place.

Lastly, the scalar plugin tests for this were a mess, so I unrolled and flattened them a fair bit and added two new tests that explicitly test the `format=` URL behavior end-to-end.

(The first commit on this PR is split out as #3781 since it includes an unrelated behavior change.)

### DIFF
--- a/tensorboard/backend/http_util.py
+++ b/tensorboard/backend/http_util.py
@@ -88,6 +88,7 @@ def Respond(
     content_encoding=None,
     encoding="utf-8",
     csp_scripts_sha256s=None,
+    filename=None,
 ):
     """Construct a werkzeug Response.
 
@@ -128,6 +129,10 @@ def Respond(
         elements for script-src of the Content-Security-Policy. If it is None, the
         HTML will disallow any script to execute. It is only be used when the
         content_type is text/html.
+      filename: If non-empty, the `Content-Disposition: attachment` header will
+        be sent specifying this value as the filename. For most user agents this
+        will trigger a file download. Callers should take care to sanitize the
+        filename if derived from user-controlled strings.
 
     Returns:
       A werkzeug Response object (a WSGI application).
@@ -184,6 +189,8 @@ def Respond(
     headers.append(("X-Content-Type-Options", "nosniff"))
     if content_encoding:
         headers.append(("Content-Encoding", content_encoding))
+    if filename:
+        headers.append(("Content-Disposition", 'filename="%s"' % filename))
     if expires > 0:
         e = wsgiref.handlers.format_date_time(time.time() + float(expires))
         headers.append(("Expires", e))
--- a/tensorboard/backend/http_util.py
+++ b/tensorboard/backend/http_util.py
@@ -190,7 +190,10 @@ def Respond(
     if content_encoding:
         headers.append(("Content-Encoding", content_encoding))
     if filename:
-        headers.append(("Content-Disposition", 'filename="%s"' % filename))
+        quoted_filename = werkzeug.http.quote_header_value(
+            filename, allow_token=False  # Always use quotes even if not needed.
+        )
+        headers.append(("Content-Disposition", "filename=%s" % quoted_filename))
     if expires > 0:
         e = wsgiref.handlers.format_date_time(time.time() + float(expires))
         headers.append(("Expires", e))
--- a/tensorboard/plugins/scalar/scalars_plugin.py
+++ b/tensorboard/plugins/scalar/scalars_plugin.py
@@ -36,17 +36,11 @@ from tensorboard.backend import http_util
 from tensorboard.data import provider
 from tensorboard.plugins import base_plugin
 from tensorboard.plugins.scalar import metadata
+from tensorboard.util import string_util
 from tensorboard.util import tensor_util
 
 
 _DEFAULT_DOWNSAMPLING = 1000  # scalars per time series
-
-
-class OutputFormat(object):
-    """An enum used to list the valid output formats for API calls."""
-
-    JSON = "json"
-    CSV = "csv"
 
 
 class ScalarsPlugin(base_plugin.TBPlugin):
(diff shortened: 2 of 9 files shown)


tensorboard/backend/http_util.py
```diff
@@ -184,6 +189,8 @@ def Respond(
     headers.append(("X-Content-Type-Options", "nosniff"))
     if content_encoding:
         headers.append(("Content-Encoding", content_encoding))
+    if filename:
+        headers.append(("Content-Disposition", 'filename="%s"' % filename))
```
wchargin, line 193: Want `attachment; filename=…`?

wchargin, line 193: Can we use [`werkzeug.http.quote_header_value(filename)`][1] here?
Callers need to sanitize the logical filename, but shouldn’t need to
know about the RFC 2616 quoted-string grammar.

[1]: https://werkzeug.palletsprojects.com/en/1.0.x/http/#werkzeug.http.quote_header_value

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -127,7 +121,7 @@ def index_impl(self, experiment=None):
 
         return result
 
-    def scalars_impl(self, tag, run, experiment, output_format):
+    def scalars_impl(self, tag, run, experiment):
```
wchargin, line 124: Looks like legit test failure in the custom scalars plugin due to the
removal of this parameter.

tensorboard/util/string_util.py
```diff
@@ -0,0 +1,67 @@
...
+      Mapping object compatible with str.translate().
+    """
+    mapping = collections.defaultdict(lambda: None)
+    remapped_chars = string.punctuation
+    allowed_chars = string.ascii_letters + string.digits + "-_="
+    for char in remapped_chars:
+        mapping[ord(char)] = "_"
+    for char in allowed_chars:
+        mapping[ord(char)] = char
+    print(mapping[ord(" ")])
```
wchargin, line 63: Remove?

tensorboard/util/string_util.py
```diff
@@ -0,0 +1,67 @@
...
+import collections
+import string
+
+
+def sanitize_for_filename(s):
+    """Returns a sanitized filename-friendly version of the input string.
+
+    This discards all characters that are not ASCII letters, digits, or
+    punctuation, and then converts any remaining characters that are not in
+    [-_=,A-Za-z0-9] to "_".
```
wchargin, line 30: Doc comment includes comma as allowed character, but code doesn’t?

tensorboard/util/string_util_test.py
```diff
@@ -0,0 +1,40 @@
...
+
+        check("", "")
+        check("foo", "foo")
+        check("bad.exe", "bad_exe")
+        check(
+            "punct!#$%&'()*+,-./:;<=>?@[\]^_`{|}~\"uation",
+            "punct___________-_____=______________uation",
+        )
+        check("w h\ti\nt\re\x0bs\x0cpace", "whitespace")
+        check("emo\U0001F95Dji", "emoji")
```
wchargin, line 36: i’m honored

tensorboard/backend/http_util.py
nfelt: Ooh yes, thanks for pointing that out.  Done and test added.

nfelt: So I specifically left out `attachment` since with the current behavior it's possible in Chrome to open the .json file in the browser in a new tab if you want (the `.csv` is always forcibly downloaded; I think this is content-type specific browser behavior).  As a user I like it when it's possible to do this rather than requiring that I always download the file, so that's why I did it this way.

Setting the `download` attribute in the HTML tag but omitting `attachment` means that the behavior depends on whether you just click the link (download) or click to open in a new tab (browser dependent, but may stay in the browser), so that seemed like a nice option to preserve.

That said I agree it's overall a bit less consistent to reason about, so if you prefer I'm happy to add `attachment`.

tensorboard/util/string_util.py
nfelt: Great catch, thanks.  Updated the code to match docs.

tensorboard/plugins/scalar/scalars_plugin.py
nfelt: Thanks, will fix shortly.

tensorboard/backend/http_util.py
wchargin: That’s fine with me (I also like that behavior). In that case, let’s fix
the newly added docs to not mention `Content-Disposition: attachment`?

nfelt:
Update: I'm withdrawing this PR because after some discussion we decided that this isn't a great approach, since in the long term we want the flexibility to be able to generate the data to download wholly on the frontend (e.g. so that it can more easily reflect client-side filtering done by the user) and because it also didn't fix other cases of this problem - e.g. in the custom scalars download URLs (which we'd also have had to change) or in the scalars dashboard "Download chart as SVG" option which also uses a dynamic filename (the tag name) and is generated client-side, making it ineligible for this approach to sidestepping the problem.

Instead, after some discussion with the Polymer folks, it turns out they were fine changing Resin's policy for the `download` attribute to skip sanitization since the attribute value is only a hint to the browser anyway (which can more appropriately sanitize or reject the proposed filename in the context of actually writing it to disk, if it so chooses).

Note I already broke the still-useful test cleanup from this PR out into #3793.

PR #3783, opened 2020-07-02
Populate ${logdir} in "Upload to TensorBoard.dev" dialog

bmd3k:
When users copy the sample command to "Upload to TensorBoard.dev" we want the command to include the actual logdir value.

  * Use ngrx to call the existing `"data/environment"` endpoint to retrieve `Environment` data and store the result as part of `CoreState`.
  * Populate the dialog command using the `data_location` property.

* Screenshots of UI changes
  * When the logdir still has to be determined:
     image
  * When logdir has been determined and fits within the provided space:
    image
  * When logdir has been determined but does not fit within the provided space:
    image

  * Wrote some tests.
  * Opened dialog and pressed the copy button to copy text. Pasted it in a command shell to verify it works.

### DIFF
--- a/tensorboard/webapp/tbdev_upload/tbdev_upload_dialog_component.ts
+++ b/tensorboard/webapp/tbdev_upload/tbdev_upload_dialog_component.ts
@@ -32,7 +32,10 @@ export class TbdevUploadDialogComponent implements OnInit, OnDestroy {
 
   private ngUnsubscribe = new Subject();
 
-  constructor(private dialogRef: MatDialogRef<TbdevUploadDialogComponent>, private store: Store<State>) {}
+  constructor(
+    private dialogRef: MatDialogRef<TbdevUploadDialogComponent>,
+    private store: Store<State>
+  ) {}
 
   ngOnInit() {
     this.environment$
@@ -65,4 +68,5 @@ export class TbdevUploadDialogComponent implements OnInit, OnDestroy {
       // render a scrollbar underneath.
       return 'tensorboard dev upload --logdir \\\n    ' + logdir;
     }
+  }
 }
--- a/tensorboard/webapp/tbdev_upload/tbdev_upload_dialog_component.ts
+++ b/tensorboard/webapp/tbdev_upload/tbdev_upload_dialog_component.ts
@@ -26,7 +26,7 @@ import {takeUntil} from 'rxjs/operators';
   styleUrls: ['./tbdev_upload_dialog_component.css'],
 })
 export class TbdevUploadDialogComponent implements OnInit, OnDestroy {
-  commandText: string = 'tensorboard dev upload --logdir {logdir}';
+  commandText: string = this.getCommandText('');
   readonly environment$ = this.store.pipe(select(getEnvironment));
 
   private ngUnsubscribe = new Subject();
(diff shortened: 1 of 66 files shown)


tensorboard/webapp/tbdev_upload/tbdev_upload_dialog_component.ts
```diff
@@ -12,20 +12,61 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    if (!logdir) {
+      // Without logdir we print a literal '{logdir}' string that user will
+      // have to manually substitute.
+      return 'tensorboard dev upload --logdir {logdir}';
+    } else {
+      // With logdir we substitute the value into the command for the user.
+      // We assume that logdir is sufficiently long that we want to print it on
+      // the next line. If the logdir value is still too long then the CSS will
+      // render a scrollbar underneath.
+      return 'tensorboard dev upload --logdir \\\n    ' + logdir;
```
wchargin, line 69: As written, this isn’t escaped, and so the command can be wrong (if the
logdir contains spaces or some special characters) or even malicious:

![Screenshot of code injection with logdir named `--nope || ls /`][ss]

[ss]: https://user-images.githubusercontent.com/4317806/86337799-6f382e80-bc06-11ea-87c5-4c8cf24da2dc.png

Can we escape for POSIX `sh` here? (Probably wouldn’t be exactly right
on Windows, but that’s fine with me.)

bmd3k: Thanks for taking a close look. Would it be sufficient to wrap the logdir value in quotes?

bmd3k: I gave this a shot: Wrap the logdir value in (single) quotes and escape quote literals in the logdir.  I added a test for it, too.

tensorboard/webapp/tbdev_upload/tbdev_upload_dialog_component.ts
```diff
@@ -12,20 +12,62 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+  ) {}
+
+  ngOnInit() {
+    this.environment$
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe((environment) => {
+        if (environment && environment.data_location) {
+          this.commandText = this.getCommandText(environment.data_location);
+        }
+      });
```
stephanwlee, line 47: A component should not have access to the store. Please create a container and pass down the environment as input using the async pipe.

tensorboard/webapp/tbdev_upload/tbdev_upload_dialog_component.ts
stephanwlee: William, in what case does the logdir from `environment` contain such command? Does our tf-runs-selector handle the case properly?

wchargin: @bmd3k: Yep, that should work on POSIX shells; thanks!

@stephanwlee: I ran with `--logdir '--nope || ls /'` to get that
screenshot. The runs selector shows this path correctly and without
escaping (see bottom-left of screenshot), which is fine imho since it’s
not intended to be run as a command.

tensorboard/webapp/tbdev_upload/tbdev_upload_dialog_component.ts
bmd3k: Done.  I've tried to mimic what is being done in header/plugin_selector.container.ts.  Note that I originally mimicked what is being done in settings/dialog_component.ts so I don't know if we should update that component as well.

PR #3785, opened 2020-07-02
[DebuggerV2] Add screenshots to inactive page

caisq:
  * Flesh out the inactive / info page of Debugger V2
  * Add .png files with circular cutout and drop-shadow effects
  * Add a genrule to inline the images in the .ng.html file. The genrule uses a py_binary
    to manipulate the content string of the html file.
  * Adjust some wording and links in the .ng.html file
  * Make some CSS adjustments
* Screenshots of UI changes
  * image
  * The standard approach of including image assets (see [description](https://scotch.io/courses/build-your-first-angular-website/adding-an-imagelogo-in-angular))
    doesn't work with the open-source `ng_module` build rule.


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive/inline_images.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Replaces %*.png% in the input .html file with a data URI containing the
+# image content of the corresopnding .png files in the images/ subdirectory.
+
+mime_type='image/png'
+
+tmp_file="$(mktemp)"
+cp "$@" "${tmp_file}"
+
+for image_file in $0.runfiles/org_tensorflow_tensorboard/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive/images/*.png; do
```
stephanwlee, line 27: I don't think this path will be compatible in google3. Please just pass the whole image path as the input (srcs)?

caisq:
Gentle ping.

stephanwlee:
Sorry, I was waiting for your comment on https://github.com/caisq/tensorboard/pull/1. Sorry for not being explicit about that.

caisq:
@stephanwlee Sorry I overlooked your PR to my fork before. I just looked at it and here are my thoughts

1. Thanks for finding an approach that doesn't require copybara. I adopted the change in the BUILD file.
2. Also thanks for bringing my attention to differences in the `awk` command line between Linux and Mac. After looking into the differences related to the `-i inplace` argument (which is not supported on Mac) and multiple `-v` flags (which is also not supported on Mac). I decided to use Python for the genrule. This is more efficient than the previous approach because it replaces N in-place file writes (N being the number of images to inline) with no file writes at all (except the final one carried out by the genrule itself).

As such, there is not much shell code change in this PR anymore. So I don't think there is a need to bring in our residential shell expert. The one-liner in the BUILD file looks correct to me.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive/inline_images.py
```diff
@@ -0,0 +1,54 @@
...
+    input_html_path = sys.argv[1]
+    image_files = sys.argv[2:]
+    with open(input_html_path, "r") as f:
+        html = f.read()
+    for image_file in image_files:
+        with open(image_file, "rb") as f:
+            base64_content = base64.b64encode(f.read()).decode("utf-8")
+        data_uri = "data:%s;base64,%s" % (MIME_TYPE, base64_content)
+        image_basename = os.path.basename(image_file)
+        image_template = "%%%s%%" % image_basename
```
stephanwlee, line 43: %%wow%% :) Is there other ways to make it more readable? Perhaps `%{}%.format()`?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive/BUILD
```diff
@@ -4,6 +4,22 @@ package(default_visibility = ["//tensorboard:internal"])
 
 licenses(["notice"])  # Apache 2.0
 
+py_binary(
+    name = "inline_images",
+    srcs = ["inline_images.py"],
+    python_version = "PY3",
+    srcs_version = "PY2AND3",
+    deps = [],
```
stephanwlee, line 12: I forget... Is `deps` field required? If not, just drop it?

caisq: It is optional if empty (see https://docs.bazel.build/versions/master/be/python.html). Dropped it.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive/inline_images.py
caisq: `"%" + image_basename + "%"` does it. Done.

caisq:
Did a test sync and everything seems to work. Will merge the PR once the CI passes.

PR #3790, opened 2020-07-06
Add the run state in the core state

stephanwlee:
This change adds a concept of `runs` in Angular so we can render the runs selector.

## Goal
- We would like to port Polymer dashboards into Angular
- While transitioning, there will be a mixture of Angular and Polymer component
- For consistency, while transitioning, we want to use Polymer based tf-runs-selector
  rendered in the Angular
- tf-runs-selector takes current runs as an input and we should supply that in Angular
- Angular should, more and more, understand the notion of runs.

## Detail:
- Add a `runs` state in the store
- [in farther future] create a `runs` feature separate from the `core` feature.

### DIFF
--- a/tensorboard/webapp/core/actions/core_actions.ts
+++ b/tensorboard/webapp/core/actions/core_actions.ts
@@ -84,6 +84,11 @@ export const changePageSize = createAction(
   props<{size: number}>()
 );
 
+/**
+ * Action for when user wants to change the runs selection in the tf-runs-selector.
+ *
+ * Action is reserved for the wrapper for the Polymer based tf-runs-selector.
+ */
 export const polymerInteropRunSelectionChanged = createAction(
   '[Core] Run Selection Changed',
   props<{nextSelection: RunId[]}>()
--- a/tensorboard/webapp/core/actions/core_actions.ts
+++ b/tensorboard/webapp/core/actions/core_actions.ts
@@ -15,6 +15,8 @@ limitations under the License.
 import {createAction, props} from '@ngrx/store';
 import {Environment, PluginId, PluginsListing} from '../../types/api';
 
+import {Run, RunId} from '../types';
+
 // HACK: Below import is for type inference.
 // https://github.com/bazelbuild/rules_nodejs/issues/1013
 /** @typehack */ import * as _typeHackModels from '@ngrx/store/src/models';
@@ -80,4 +82,14 @@ export const changeReloadPeriod = createAction(
 export const changePageSize = createAction(
   '[Core] Page Size Change',
   props<{size: number}>()
+);
+
+export const polymerInteropRunSelectionChanged = createAction(
+  '[Core] Run Selection Changed',
+  props<{nextSelection: RunId[]}>()
+);
+
+export const fetchRunSucceeded = createAction(
+  '[Core] Run Fetch Successful',
+  props<{runs: Run[]}>()
 );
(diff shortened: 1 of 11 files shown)


tensorboard/webapp/core/store/core_reducers.ts
```diff
@@ -99,6 +99,21 @@ const reducer = createReducer(
       ...state,
       pageSize: nextPageSize,
     };
+  }),
+  on(actions.fetchRunSucceeded, (state, {runs}) => {
+    // Do not modify the runSelection since the Polymer component is the
+    // force of truth for the Polymer Interop.
```
bmd3k, line 105: force => source.

tensorboard/webapp/core/store/core_types.ts
```diff
@@ -32,6 +34,11 @@ export interface CoreState {
   // Size of a page in a general paginated view that is configurable by user via
   // settings.
   pageSize: number;
+
+  // TODO(stephanwlee): move these state to the `runs` features.
+  // For now, we want them here for Polymer interp states reasons, too.
```
bmd3k, line 39: interp => interop

tensorboard/webapp/core/actions/core_actions.ts
```diff
@@ -73,3 +75,13 @@ export const changePageSize = createAction(
   '[Core] Page Size Change',
   props<{size: number}>()
 );
+
+export const polymerInteropRunSelectionChanged = createAction(
```
bmd3k, line 79: Can you explain to me why this action is prefixed with polymerInterop but the other is not?  Does it have something to do with who you expect to dispatch the action?

tensorboard/webapp/core/store/core_types.ts
```diff
@@ -32,6 +34,11 @@ export interface CoreState {
...
+  polymerInteropRuns: Run[];
```
bmd3k, line 40: Can you explain to me (1) why you don't immediately structure the data model the way you want (with `runs` as a feature) and (2) why these are prefixed with polymerInterop.

stephanwlee: Ah, sure. I think the question here is related to the one on the core_actions.ts, too. 

The notion of a run is native on the internal application, but it is not the case for OSS. Although unclear when, we will eventually port the features into OSS. While waiting for the feature to be ported out, I wanted to unblock some of our OSS Angular development by making a notion of runs first class in OSS too. However, since the name of the selector/action will be about the same, I did not want them to eventually collide and make troubles. Also, I wanted to make sure the readers are aware that this state is temporary and is to be only used for interop with the Polymer component, tf-runs-selector.

While the shape and API is nearly congruent as the internal one for now, I did not want to make porting harder in case they go out of sync.

Now regarding #1, there is a cost associated with creating a new feature:
1. code bloat is one but...
2. unclear how it will reconcile with internal `runs` feature when the names are the same. 

Does this answer both of your questions?

bmd3k: Yes, I think so.  Still wondering a bit about the questions in my other comment but I think this offers enough clues.

tensorboard/webapp/core/actions/core_actions.ts
bmd3k: I think the answer here is "we expect polymerInteropRunSelectionChanged to be dispatched due to interaction with tf-run-selector" while "we expect fetchRunSuceeded to be fired by a fetch of the runs data in angular code".  Is that correct?  Whether I'm right or wrong, it could be worth documenting who you expect to be firing the events and when you expect them to do it.

stephanwlee: Added a stanza about how the `polymerInteropRunSelectionChanged` is expected to only be used by tf-runs-selector.

PR #3792, opened 2020-07-06
Depend on d3 via npm

stephanwlee:
We would like to start using d3 in Angular and others. This change makes sure the
dependency added in Angular does not deviate from the Polymer one by using the
same files.

Gotchas:
- While d3 can be used as a UMD module, we can also use them as regular esmodule.
- d3 in Polymer is used in UMD style (accessing global variable; e.g., `d3.select`)
  - we add the dependency by loading `d3.min.js`.
- d3 in Angular will be used via ESModule
  - `import {select} from 'd3';` gets bundled by rollup
- Because of two module systems, we will load multiple copies of d3 but the ESModule
  one is in the closure and does not seem to collide with global `d3`. While this will
  bloat the bundle size, we hope that this is on temporary solution that will be addressed
  when we bundle Polymer via rollup (after Polymer 3 migration).

Do note that it was not possible to defrag the d3 typings. 
Also note that the change introduces three JS errors when running the Angular app
which does not seem to negatively impact the app.

Necessary copybara config change: cl/320030660
Test change: cl/320024847

### DIFF
--- a/tensorboard/components/tf_imports/BUILD
+++ b/tensorboard/components/tf_imports/BUILD
@@ -135,6 +135,7 @@ tf_web_library(
     ],
 )
 
+# Separated out from ":d3" because of strip_prefix.
 tf_web_library(
     name = "d3_lib",
     srcs = [
--- a/tensorboard/components/tf_imports/BUILD
+++ b/tensorboard/components/tf_imports/BUILD
@@ -136,13 +136,24 @@ tf_web_library(
 )
 
 tf_web_library(
+    name = "d3_lib",
+    srcs = [
+        "@npm//:node_modules/d3/dist/d3.min.js",
+    ],
+    path = "/tf-imports",
+    strip_prefix = "node_modules/d3/dist",
+)
+
+tf_web_library(
     name = "d3",
     srcs = [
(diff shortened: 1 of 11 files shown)


stephanwlee:
+cc @sparkier

tensorboard/components/tf_imports/BUILD
```diff
@@ -135,14 +135,25 @@ tf_web_library(
     ],
 )
 
+tf_web_library(
```
nfelt, line 138: Might be worth a comment like "separated out from :d3 target to apply `strip_prefix`"?  (Especially useful internally when `strip_prefix` gets removed.)

tensorboard/defs/defs.bzl
```diff
@@ -71,6 +71,7 @@ def tf_ng_web_test_suite(runtime_deps = [], bootstrap = [], deps = [], **kwargs)
         ],
         runtime_deps = runtime_deps + [
             "//tensorboard/webapp/testing:initialize_testbed",
+            "@npm//d3",
```
nfelt, line 74: Useful to have a comment here why we need d3 as a general runtime dep for tests

PR #3796, opened 2020-07-06
data: make `DataProvider` method args keyword-only

wchargin:
Summary:
We plan to add a new `ctx` (“context”) parameter to all these methods at
the start of the positional parameters list, so we first ensure that all
existing call sites pass keyword arguments. This is a breaking change,
but the data provider API is marked experimental for exactly this
purpose. All Google-internal call sites have recently been fixed up.

(We have to fix up all implementations rather than just the interface
because Python doesn’t actually have a static type system.)

Test Plan:
Existing tests pass here and in a test sync (<http://cl/319881052>).

wchargin-branch: data-kwarg-only

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -372,7 +372,7 @@ class TensorBoardWSGI(object):
         response = collections.OrderedDict()
         eid = plugin_util.experiment_id(request.environ)
         plugins_with_data = frozenset(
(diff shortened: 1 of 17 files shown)


wchargin:
Syntax reference is [PEP 3102][1], in case you haven’t seen it before.

[1]: https://www.python.org/dev/peps/pep-3102/

PR #3800, opened 2020-07-07
Style "Upload" button to match UX designs

bmd3k:
* Motivation for changes
Style new "Upload" button to match UX designs

* Technical details
Override default angular styles with those from the UX designs

* Screenshot
image

### DIFF
--- a/tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
+++ b/tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
@@ -24,12 +24,7 @@ const LOCAL_HOSTNAMES: string[] = ['localhost', '127.0.0.1'];
 @Component({
   selector: 'tbdev-upload-button',
   template: `
-    <button
-      mat-stroked-button
-      *ngIf="shown"
-      (click)="openDialog()"
-      class="button"
-    >
+    <button mat-stroked-button *ngIf="shown" (click)="openDialog()">
       <span class="button-contents">
         <mat-icon svgIcon="info_outline_24px"></mat-icon>
         Upload
@@ -40,7 +35,7 @@ const LOCAL_HOSTNAMES: string[] = ['localhost', '127.0.0.1'];
     `
       /* More precise selector than angular material selectors to override
          styling. */
-      button.button.mat-stroked-button {
+      button.mat-stroked-button {
         background-color: rgba(198, 202, 209, 0.2);
         border: 1px solid #e9e9e9;
       }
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
```diff
@@ -33,6 +33,12 @@ const LOCAL_HOSTNAMES: string[] = ['localhost', '127.0.0.1'];
   `,
   styles: [
     `
+      /* More precise selector than angular material selectors to override
+         styling. */
+      button.mat-stroked-button {
```
bmd3k, line 38: I implemented this pretty naively. Is there a more angular-friendly way of accomplishing this?

These are the selectors I'm competing with:
image

tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
```diff
@@ -33,6 +33,12 @@ const LOCAL_HOSTNAMES: string[] = ['localhost', '127.0.0.1'];
...
+        background-color: rgba(198, 202, 209, 0.2);
+        border: 1px solid #e9e9e9;
```
stephanwlee, line 40: Please use SASS and the variables from the theme. Also for the rgba, use the rgba mixin and specify using the hex (prefer theme variable, too).

bmd3k: I refactored CSS and HTML to their own files.

There was much offline discussion and eventually we landed on making the button color Material Orange 500.  I've made some modification to the $tf-orange palette and $tb-primary to incorporate the new color. Those changes may require more discussion. 

Now I can set `background-color: mat-color($tb-primary, 500);` and the colors work with the tf-orange palette and sort of work with the tf-slate palette.

tensorboard/webapp/tbdev_upload/tbdev_upload_button_component.ts
```diff
@@ -17,32 +17,13 @@ import {MatDialog} from '@angular/material/dialog';
 
 import {TbdevUploadDialogContainer} from './tbdev_upload_dialog_container';
 
-// A list of hostname values that will trigger the tbdev-upload-button button to
-// appear.
+// A list of hostname values that will trigger the button to appear.
 const LOCAL_HOSTNAMES: string[] = ['localhost', '127.0.0.1'];
```
stephanwlee, line 21: Btw, of course this is a separate PR, but we need to also add the ipv6 version :)

PR #3801, opened 2020-07-07
scalars: remove legacy multiplexer codepaths

wchargin:
Summary:
In TensorBoard 2.2.0, we promoted the generic data codepaths to be on by
default (#3265). This was intended to be a silent change, and we haven’t
heard any complaints or bug reports. In this patch, we remove the legacy
codepaths entirely, simplifying the code and tests.

Recommend hiding whitespace changes when reviewing this patch.

Test Plan:
The scalars plugin still works with `--generic_data` unset or set to any
of true, false, or auto, both when there is data and when there isn’t.

wchargin-branch: scalars-generic-only


wchargin:
As promised: <https://github.com/tensorflow/tensorboard/pull/3265#issuecomment-588615473>

wchargin:
(Custom scalar plugin tests will break here… please hold.)

PR #3804, opened 2020-07-07
scalars: remove vestigial dual-test infrastructure

wchargin:
Summary:
Now that the scalars plugin can only read from a data provider, we can
remove the complexity that afforded testing both the data provider and
multiplexer codepaths. In particular, what was previously a test method
decorator that yielded sub-tests is now just a helper function.

Test Plan:
That the new tests pass suffices.

wchargin-branch: scalars-monotest


bileschi:
Nice!

Issue #3809, opened 2020-07-08
Dynamic plugin API method `getURLPluginData` is broken


psybuzz:
TensorBoard dynamic plugins - `tensorboard-plugin-fairness-indicators`, `tensorboard-plugin-wit` make use of the API method `getURLPluginData()`.  After the switch to the Angular entry point, this now returns an empty object, rather than the URL data.

Repro:
- Install `tb-nightly`, `tensorboard-plugin-wit`
- Open `http://localhost:6006/legacy.html#whatif&p.whatif.examplesPath=foobar` (legacy Polymer entry)
- Observe that the "path to examples" is properly filled
![image](https://user-images.githubusercontent.com/2322480/86951448-65329600-c106-11ea-9c63-a093b352e8fa.png)

- Open `http://localhost:6006/#whatif&p.whatif.examplesPath=foobar` (new Angular entry)

Actual: "path to examples" is empty
Expected: it should be populated as before

It seems that in Polymer, TB relied on `tf-tensorboard.html` to call `migrateLegacyURLScheme` in `storage.ts`, which did the initial reading of the URL hash after the document is loaded.  The Angular entry point is missing this call.

psybuzz:
Also manually tested with `bazel run tensorboard`, `tensorboard-plugin-fairness-indicators 0.0.6` and navigating to `http://localhost:6006/#fairness_indicators&p.fairness_indicators.evaluation_output_path=foobar` will actually make a network request to `http://localhost:6006/data/plugin/fairness_indicators/get_evaluation_result_from_remote_path?evaluation_output_path=foobar` as intended.

psybuzz:
Closing, as this is addressed by https://github.com/tensorflow/tensorboard/pull/3811

PR #3811, opened 2020-07-08
Fix getURLPluginData in Angular entry point

psybuzz:
This fixes the broken dynamic plugin API method `getURLPluginData`, which was
broken after the Angular entry point switch. `storage.ts` has an initial call to read
from the URL hash, but this is actually a no-op, since before `setUseHash(true)`
is called, it assumes it should use the "fake hash" by default (for tests only).

In the Polymer entry point world, `tf-tensorboard.html`'s `ready` callback will
`setUseHash(true)` before calling `migrateLegacyURLScheme`, which will
read from the real URL hash and normalize it.

This adds a similar call to the Angular side, in HashDeepLinker.  See #3809.

Manually checked that the repro steps now pass:
image

### DIFF
--- a/tensorboard/components/tf_storage/storage.ts
+++ b/tensorboard/components/tf_storage/storage.ts
@@ -49,7 +49,7 @@ namespace tf_storage {
   export const DISAMBIGUATOR = 'disambiguator';
 
   // Keep an up-to-date store of URL params, which iframed plugins can request.
-  export let urlDict: StringDict = componentToDict(readComponent());
+  export let urlDict: StringDict = {};
 
   tf_storage.addHashListener(() => {
     tf_storage.urlDict = componentToDict(readComponent());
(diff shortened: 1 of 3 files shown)


tensorboard/components/tf_storage/storage.ts
```diff
@@ -218,7 +218,7 @@ namespace tf_storage {
     return {get, set, getInitializer, getObserver, disposeBinding};
   }
 
-  export function migrateLegacyURLScheme() {
```
stephanwlee, line 221: Any reason we have to make the name change?  Also, you will need to update tf-tensorboard then.

psybuzz: Reverted back to the old name, and added a comment in tf-tensorboard to explain that order matters.

psybuzz:
Thanks, I've done a manual check to confirm that both the default Angular entry point and the /legacy.html one support reading the URL hash in both FI and WIT.

PR #3815, opened 2020-07-08
data: add `AuthProvider` and `AuthContext`

wchargin:
Summary:
This patch adds two new types for integrating authentication into the
data provider API. The `AuthProvider` type defines an interface for
extracting a specific kind of credential (say, an OAuth token) from a
WSGI environment. The `AuthContext` type lets clients ask for a
credential from a specific type of `AuthProvider`.

As written, `AuthContext` is not threadsafe. Making it truly threadsafe
is tricky if we assume that the `AuthProvider`s’ `authenticate` methods
do anything that Python code can, including obtaining a reference to the
very same `AuthContext` and calling its `get` method with a different
auth provider. (This would be a Bad Idea on the part of the caller, but
deadlocking would still be unfortunate.) Since most requests are handled
by a single thread, anyway, this shouldn’t be a problem.

Test Plan:
Basic unit tests added. Google-internal proof-of-concept code builds on
this work and shows that the `AuthProvider` interface suffices for now.

wchargin-branch: data-auth

### DIFF
--- a/tensorboard/data/BUILD
+++ b/tensorboard/data/BUILD
@@ -18,6 +18,25 @@ py_library(
 )
 
 py_library(
+    name = "auth",
+    srcs = ["auth.py"],
+    srcs_version = "PY3",
+    deps = [],
+)
+
+py_test(
+    name = "auth_test",
+    size = "small",
+    srcs = ["auth_test.py"],
+    srcs_version = "PY3",
+    tags = ["support_notf"],
+    deps = [
+        ":auth",
+        "//tensorboard:test",
+    ],
+)
+
+py_library(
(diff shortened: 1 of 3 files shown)


wchargin:
(Test failure is unrelated; #3814 should fix it.)

wchargin:
> I don't quite follow the thread safety discussion in the description.
> I think you mean to say AuthContext (rather than AuthProvider) is not
> thread safe, right?

Yes, thanks; fixed.

> But then I'm still not sure I understand what is meant about lock
> acquisition order.

Clarified a bit, but it requires some contrived client code and isn’t
really a big deal anyway.

PR #3816, opened 2020-07-08
backend: add `RequestContext`

wchargin:
Summary:
The `RequestContext` type stores request-specific information about
cross-cutting concerns. For instance, think:

  - authentication: an `AuthContext` for obtaining credentials
  - internationalization: the user’s home language
  - performance: a flag randomly set on 0.1% of requests to trigger more
    detailed profiling
  - deployment: a “debug”/“release” flag

A function call at any point in the stack can pull this information from
a `RequestContext` passed to it as an argument. A `RequestContext` is
first obtained in a request handler, where it can be extracted from the
WSGI environment.

As with the Go language’s `Context` type, this context isn’t just for
arbitrary values that we don’t want to bother passing around, like an
experiment ID or plugin name. Thus, the context is closed to a fixed set
of keys. For now, we include only an `AuthContext`.

The data provider API will soon be extended to require a context.

Test Plan:
Unit tests included.

wchargin-branch: request-context


tensorboard/backend/context.py
```diff
@@ -0,0 +1,99 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Request-scoped context."""
+
+from tensorboard.data import auth as auth_lib
```
nfelt, line 17: This feels a bit circular - we will end up with `tensorboard.data.provider` depending on `tensorboard.backend.context` which then depends back on `tensorboard.data.auth`.

See also the other comment on where `context` should live, but it would be ideal if there was a linearizable dependency structure between our high-level API packages.  It also feels a bit strange to say that `auth` lives in `data` if it's going to be the one true "auth" field on `RequestContext`?  As opposed to something that signifies it just being "auth for data providers".

(IMO the most obvious way to avoid that would probably be to generalize `auth` beyond just the data providers and make it part of the same top-level package as `context` and then having `tensorboard.data` take a dependency on that, but curious about your thoughts.)

tensorboard/plugin_util.py
```diff
@@ -150,6 +151,15 @@ def markdowns_to_safe_html(markdown_strings, combine):
     return warning + sanitized_combined
 
 
+def context(environ):
```
nfelt, line 154: Do we need this helper if we already have `context.from_environ()` itself, and that is public (assuming that the plugin is expected to have access to the `RequestContext` type itself, which is in the same module)?

tensorboard/backend/context.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+
+    def __init__(self, *, auth=None):
+        """Create a request context.
+
+        Args:
+          See "Fields" on class docstring. All arguments are optional
+          and will be replaced with default values if appropriate.
+        """
+        self.auth = auth if auth is not None else auth_lib.AuthContext.empty()
```
nfelt, line 49: If we're making this immutable, possibly nicer to use `@property` so that somebody can't wantonly reassign this.

tensorboard/backend/context.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Returns:
+      The `RequestContext` stored in the WSGI environment, or an empty
+      `RequestContext` if none is stored.
+    """
+    result = environ.get(_WSGI_KEY)
+    return result if result is not None else RequestContext()
+
+
+def update_environ(environ, **kwargs):
```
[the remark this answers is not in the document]
wchargin: Right again; this was bugging me, too. Definitely `update` feels
imperative (`dict.update`), though `replace` feels imperative to me,
too, so I picked `update` because in my experience it’s the standard
name for an immutable lens update (e.g., [Pickering, et al. 2017][1]).
(I almost called the dual `view_environ`, but thought that `from_*` was
clearer.)

I can go with your `set_in_environ`. I’m not a huge fan of the mutation
or the name, but if you prefer it then I don’t feel strongly about it.

[1]: https://arxiv.org/abs/1703.10857

tensorboard/plugin_util.py
wchargin: I like having this helper so that plugins don’t need to import yet
another TensorBoard utility module. Plugins already import this module
to get `plugin_util.experiment_id(environ)`, and having `context`
adjacent to that feels pretty natural to me.

tensorboard/backend/context.py
nfelt: (For posterity, #3826 moves `auth` to the top level.)

tensorboard/plugin_util.py
nfelt: Fair enough, as discussed this gives us a bit more flexibility later if we want to move `RequestContext`.  (I do wish ultimately we had a slightly less "util" name than `plugin_util` but we don't have to fix that now.)

tensorboard/backend/BUILD
```diff
@@ -7,6 +7,28 @@ licenses(["notice"])  # Apache 2.0
 
 exports_files(["LICENSE"])
 
+py_library(
+    name = "context",
```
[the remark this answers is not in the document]
nfelt: As discussed, we've agreed to move to `tensorboard.context`.

PR #3817, opened 2020-07-08
data: add `ctx` parameter to `DataProvider` methods

wchargin:
Summary:
All `DataProvider` methods now accept a `ctx: RequestContext` value as
the first positional argument. For now, the argument is optional and no
call sites provide it. Once internal data provider implementations are
also updated to accept this argument, it will be retrofitted onto all
call sites and made required.

Test Plan:
The addition of a positional parameter is safe since all existing
parameters are keyword-only, as of #3796. It suffices that tests pass.

wchargin-branch: data-optional-ctx-param


wchargin:
My understanding is that APIs that use context-style arguments typically
want the context to always be passed as the first parameter (e.g.,
[Golang `Context`][go]), so I don’t anticipate adding positional
parameters _before_ `ctx`. We can still add more positional parameters
after `ctx` or more keyword-only parameters. I agree that we may as well
keep the rest of the parameters keyword-only, and I don’t plan to change
them back.

I do think that `foo(ctx, ...)` looks nicer than `foo(ctx=ctx)`, though
of course this is subjective and aesthetic.

(OCaml’s lovely `foo ~ctx` syntax would be nice here, but alas[.])

[go]: https://golang.org/pkg/context/
[.]: https://github.com/tensorflow/tensorboard/pull/3819/files/5627ff92bd200b9c50279142196a912a8ba7db1f#diff-4d7386333484c768dc4f09fe0c18096b

nfelt:
Ack, fair enough that it's probably not going to change (although now that we've said that...).

PR #3819, opened 2020-07-09
data: require `ctx` and fix all call sites

wchargin:
Summary:
This patch changes the `MultiplexerDataProvider` to check that the `ctx`
argument is always passed, and fixes all call sites to pass it.

Test Plan:
Manually verified behavior of the following plugins:

  - scalars, incl. download links
  - custom scalars, incl. download links
  - images
  - audio
  - graphs (`--generic_data=true` required)
  - distributions
  - histograms
  - text
  - PR curves
  - hparams (automated tests do not suffice for any functionality)

Other plugins (debugger, profile, beholder, mesh, what-if tool) don’t
use the data provider interface and so are unaffected by this change.
The `text_v2` plugin does, but with `?experimentalPlugin=text_v2` I only
see a blank screen, even at head.

wchargin-branch: data-ctx-required


tensorboard/backend/event_processing/data_provider.py
```diff
@@ -48,6 +48,10 @@ def __init__(self, multiplexer, logdir):
         self._multiplexer = multiplexer
         self._logdir = logdir
 
+    def _validate_context(self, ctx):
+        if type(ctx).__name__ != "RequestContext":
+            raise TypeError("ctx must be a RequestContext; got: %r" % (ctx,))
```
nfelt, line 53: #ProblemsWeWouldNotHaveInARealProgrammingLanguage

tensorboard/plugins/graph/graphs_plugin.py
```diff
@@ -77,7 +78,8 @@ def is_active(self):
         if self._data_provider:
             return False  # `list_plugins` as called by TB core suffices
 
-        return bool(self.info_impl())
+        empty_context = context.RequestContext()  # not used
```
nfelt, line 81: No change needed, but honestly it seems simpler just to pass in `plugin_util.context(request.environ)` anyway even if it's not needed by `info_impl()` for this code path rather than to reason through why it's safe to omit it?

tensorboard/plugins/histogram/histograms_plugin.py
```diff
@@ -84,16 +85,17 @@ def is_active(self):
             return False  # `list_plugins` as called by TB core suffices
 
         if self._multiplexer:
-            return any(self.index_impl(experiment="").values())
+            empty_context = context.RequestContext()  # not used
```
nfelt, line 88: Ditto

tensorboard/plugins/graph/graphs_plugin.py
wchargin: We don’t have an `environ` because we don’t have a `request`. The bit of
`backend/application.py` that calls this method does, but we don’t have
access to it (and can’t easily get it without breaking `TBPlugin`
compatibility).

tensorboard/plugins/histogram/histograms_plugin.py
wchargin: [![dit]][to]

[dit]: https://user-images.githubusercontent.com/4317806/87107642-29d2bd00-c215-11ea-8d6e-2bcf4c8f6c99.png
[to]: https://i.kym-cdn.com/photos/images/original/000/303/457/f29.gif

tensorboard/plugins/graph/graphs_plugin.py
nfelt: Oops... duh :/ Sorry!

tensorboard/plugins/histogram/histograms_plugin.py
nfelt: I suppose making me snort unwillingly during a meeting is a fair punishment for having left a silly comment not just once but twice 🤣

PR #3823, opened 2020-07-09
ng: fetch and populate the runs in the store

stephanwlee:
Fetch and store the runs in the Angular app similar to that of environment.

### DIFF
--- a/tensorboard/webapp/core/effects/core_effects_test.ts
+++ b/tensorboard/webapp/core/effects/core_effects_test.ts
@@ -17,7 +17,7 @@ import {TestBed} from '@angular/core/testing';
 import {provideMockActions} from '@ngrx/effects/testing';
 import {Action, Store} from '@ngrx/store';
 import {MockStore, provideMockStore} from '@ngrx/store/testing';
-import {ReplaySubject, of} from 'rxjs';
+import {Subject, ReplaySubject, of} from 'rxjs';
 
 import {CoreEffects} from './core_effects';
 import * as coreActions from '../actions';
@@ -29,6 +29,7 @@ import {
   createState,
   createCoreState,
 } from '../testing';
+import {Run} from '../types';
 
 import {PluginsListing} from '../../types/api';
 import {DataLoadState} from '../../types/data';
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/core/effects/core_effects_test.ts
```diff
@@ -208,7 +214,20 @@ describe('core_effects', () => {
 
         action.next(onAction);
         httpMock.expectOne('data/plugins_listing').flush(pluginsListing);
-        expect(dispatchSpy).toHaveBeenCalledTimes(3);
+        fetchRunsSubjects[0].next([{id: '1', name: 'Run 1'}]);
```
bmd3k, line 217: Up to you but I don't think it's necessary to check all the actions here.  All you really care about in this test is "did something trigger?" rather than "How did it trigger?".

tensorboard/webapp/webapp_data_source/tb_server_data_source.ts
```diff
@@ -49,11 +50,23 @@ export class TBServerDataSource {
     return this.http.get<PluginsListing>(pathWithParams);
   }
 
-  fetchRuns() {
-    return from(this.tfBackend.runsStore.refresh());
+  fetchRuns(): Observable<Run[]> {
```
bmd3k, line 53: Would be nice if this implementation was consistent with fetchEnvironment().  Feel free to refactor fetchEnvironment() to look more like this.

I do think some explanation of the implications of these two calls (like the one in fetchEnvironment()) is valuable.

tensorboard/webapp/webapp_data_source/tb_server_data_source_test.ts
```diff
@@ -50,5 +69,35 @@ describe('tb_server_data_source', () => {
...
+        ]);
+      }));
+
+      it('calls the polymer API to refresh the polymer store, too', () => {
+        dataSource.fetchRuns().subscribe(() => {});
+        expect(tbBackend.tf_backend.runsStore.refresh).toHaveBeenCalled();
+      });
+    });
+
+    describe('fetchEnvironment', () => {
```
bmd3k, line 94: ah, thanks.

tensorboard/webapp/core/effects/core_effects_test.ts
stephanwlee: There are data manipulation (convert backend data is to the frontend data structure) so I think the "how" part is a bit important here.

tensorboard/webapp/webapp_data_source/tb_server_data_source.ts
stephanwlee: Done, made them look equal.

PR #3824, opened 2020-07-09
application: integrate auth providers and context

wchargin:
Summary:
This patch adds a new `auth_providers` argument to `TensorBoardWSGIApp`.
Clients can pass a mapping of auth providers that will be available to
plugin code on an `AuthContext`, available via the `RequestContext`.

Test Plan:
Unit tests included, demonstrating example usage.

wchargin-branch: application-auth-integration


tensorboard/backend/application.py
```diff
@@ -85,6 +88,8 @@ def TensorBoardWSGIApp(
       deprecated_multiplexer: Optional `plugin_event_multiplexer.EventMultiplexer`
           to use for any plugins not yet enabled for the DataProvider API.
           Required if the data_provider argument is not passed.
+      auth_providers: Optional mapping whose values are `AuthProvider` values
+          (with arbitrary keys). Defaults to `{}`.
```
nfelt, line 92: The keys are arbitrary in what actual object they can be, but not semantically arbitrary, if the expectation is that data providers will reference values from this mapping using those keys - maybe clarify that aspect?

tensorboard/backend/application.py
```diff
@@ -176,6 +186,8 @@ def __init__(
               experimentally. The corresponding plugins will only be activated for
               a user if the user has specified the plugin with the experimentalPlugin
               query parameter in the URL.
+          auth_providers: Optional mapping whose values are `AuthProvider`
+              values (with arbitrary keys). Defaults to `{}`.
```
nfelt, line 190: Ditto

tensorboard/backend/application.py
```diff
@@ -544,6 +558,18 @@ def wrapper(*args):
     return wrapper
 
 
+def _auth_context_middleware(wsgi_app, auth_providers):
+    def wrapper(*args):
```
nfelt, line 562: Can we simply specify these arguments as `environ` and `start_response`?

I see that `_handling_errors` doesn't do that, I assume to avoid eating the `self` parameter, but I think that's only relevant when trying to make a middleware that can be used as a decorator on a WSGI-app-as-method (as with `wrappers.Request()`) which we haven't bothered doing for our standalone middlewares like PathPrefixMiddleware, so there doesn't seem to be a ton of reason to do it here.

tensorboard/backend/application.py
wchargin: Good point; thanks. Will clarify.

tensorboard/backend/application.py
wchargin: I think you’re right: `_handling_errors` was previously used as a
decorator, but we refactored that in #2732. Will change.

tensorboard/backend/application.py
wchargin: Done.

tensorboard/backend/application.py
wchargin: Ditto.

tensorboard/backend/application.py
wchargin: Done, and likewise for `_handling_errors` in #3827.

PR #3825, opened 2020-07-09
Added the Angular wrapper for tf-runs-selector 

stephanwlee:
For the ease of our development in Angular, especially for porting a
fixing dashboard without any functional differences, we decided to just
wrap a Polymer based runs selector. This change introduced that.

However, do note that the selection state is in the ngrx store so you
must use the regular Angular pattern to render and access the data
related to runs.

Demo of it being applied to the app_container.ng.html.

image

### DIFF
--- a/tensorboard/webapp/core/store/core_selectors.ts
+++ b/tensorboard/webapp/core/store/core_selectors.ts
@@ -89,10 +89,14 @@ const selectSelection = createSelector(
   }
 );
 
+/**
+ * runSelection can be `null` when information is yet missing; e.g., run selection for an
+ * experiment that is not yet fetched.
+ */
 export const getRunSelection = createSelector(
   getRuns,
   selectSelection,
-  (runs: Run[], selection: Set<RunId>): Map<RunId, boolean> => {
+  (runs: Run[], selection: Set<RunId>): Map<RunId, boolean> | null => {
     const runSelection = new Map();
 
     for (const {id} of runs) {
--- a/tensorboard/webapp/core/store/core_selectors.ts
+++ b/tensorboard/webapp/core/store/core_selectors.ts
@@ -82,9 +82,23 @@ export const getRuns = createSelector(
   }
 );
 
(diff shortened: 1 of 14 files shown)


tensorboard/webapp/core/store/core_selectors.ts
```diff
@@ -82,9 +82,23 @@ export const getRuns = createSelector(
...
-  (state: CoreState): Map<RunId, boolean> => {
+  (state: CoreState): Set<RunId> => {
     return state.polymerInteropRunSelection;
   }
 );
+
+export const getRunSelection = createSelector(
+  getRuns,
+  selectSelection,
+  (runs: Run[], selection: Set<RunId>): Map<RunId, boolean> => {
```
psybuzz, line 95: The internal run store's version of this selector returns `Record<string, boolean>|null`, which is almost the same except for the `|null`.

Should we make this OSS return type nullable, so that we don't have to add guards at all the OSS locations calling this selector later on?

tensorboard/webapp/runs/views/legacy_runs_selector/legacy_runs_selector_component.ts
```diff
@@ -0,0 +1,49 @@
...
+export class LegacyRunsSelectorComponent implements AfterViewInit {
+  @Output()
+  onSelectionChange = new EventEmitter<string[]>();
+
+  @ViewChild('selector', {static: true})
+  private selector!: ElementRef;
+
+  ngAfterViewInit() {
+    this.selector.nativeElement.addEventListener(
+      'selected-runs-changed',
```
psybuzz, line 43: I don't see any code that dispatches this event.  Is the event firing coming in a followup PR?  If not, could you please point to it in case I missed?

stephanwlee: Left a comment on where it is coming from. Hint: it is a Polymer thing :)

PR #3828, opened 2020-07-10
Revert logic that preserves axis domains

psybuzz:
Domain preservation logic added in https://github.com/tensorflow/tensorboard/pull/3813
introduced a regression where changing the tag filter, tooltip sort settings
would reset the domain of axes to Plottable's default [0, 1] range.

This results in charts that appear blank, because it's out of view.


psybuzz:
Manually confirmed internally that tag filtering and changing xAxis type now properly resets the domain to fit data.

psybuzz:
Also manually confirmed in OSS that this fixes the issue as well.

PR #3831, opened 2020-07-11
reland: preserve domains in line chart when changing settings

psybuzz:
When clients of VzLineChart2 update the colorScale externally, the x/y axis
auto-scales to fit the domain. This may be unwanted behavior if clients want
to allow users to change a chart's color frequently without losing a carefully
crafted viewport domain.

This preserves the existing domains upon property changes that recreate
the chart, but do not affect the position of points.

Historical record:
- Introduced in https://github.com/tensorflow/tensorboard/pull/3813
- Reverted in https://github.com/tensorflow/tensorboard/pull/3828
  due to multiple regressions

1) Changing settings that affect position of chart points (e.g. xAxis type)
would not re-fit domain to data. Users may set xAxis to "Relative", and think
that the data was missing, when it was only outside of the viewport.

This is overcome by separating the preserve-domain logic into a separate
callback, which does not fire when properties affecting chart point positions
change.

2) Removing and re-attaching the same vz-line-chart element in DOM
could reset the domain to Plottable's default [0, 1] range. The exact cause of
this is unclear, but the axis domain reset occurs within
`Plottable.Components.Table.prototype.destroy`.

This is overcome by deleting the `_chart` LineChart reference when a
VzLineChart2 is detached from DOM. Dropping the reference is safe:

Re-attaching a previously detached VzLineChart2 does not reuse the old
Line chart:
- attach/detach always updates the `isAttached` property
- `isAttached` changes are wired to trigger `_createChart`
- When `_createChart` renders a chart via `renderTo`, it always renders a brand
  new instance of LineChart
- Thus, re-attaching an old VzLineChart2 element does not rely on its old
LineChart instance

Deleting the `_chart` property is safe:
- Polymer observers fire these methods when the property changes:
  _reloadFromCache, _smoothingChanged, _tooltipSortingMethodChanged,
  _outliersChanged
- All those methods guard against a falsy `_chart` before using it

Manually tested that the regressions are addressed and updating colorScale still
preserves domain.

### DIFF
--- a/tensorboard/components/vz_line_chart2/vz-line-chart2.ts
+++ b/tensorboard/components/vz_line_chart2/vz-line-chart2.ts
@@ -239,10 +239,9 @@ namespace vz_line_chart2 {
     },
 
     observers: [
-      '_makeChart(xComponentsCreationMethod, xType, yValueAccessor, yScaleType)',
-      // Properties that do not affect positions of rendered points should
-      // preserve the chart's previous viewport domain.
-      '_makeChartWithPreviousDomain(colorScale, isAttached, tooltipColumns)',
+      '_makeChart(xComponentsCreationMethod, xType, yValueAccessor, yScaleType, isAttached)',
+      '_colorScaleChanged(colorScale)',
+      '_tooltipColumnsChanged(tooltipColumns)',
       // Refer to the cache and, if available, load data of a new visible series.
       '_reloadFromCache(_chart, _visibleSeriesCache)',
       '_smoothingChanged(smoothingEnabled, smoothingWeight, _chart)',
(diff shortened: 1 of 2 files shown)


psybuzz:
> Would it be possible to not create a new chart at all when the colorScale changes?

Done for colorScale and tooltipColumns.

> Honestly, we should never recreate the LineChart instance if possible.

Preserving the instance on attach/re-attach is something to explore.  We'd need to require that clients manually call destroy() (currently called on both detach and re-attach).  If we don't care about removing event listeners, this can be skipped, but it expands the scope of this PR.

For now I'd like to keep this PR focused on colorScale, tooltipColumns.  sorting, smoothing, outlier changes already avoid recreating the LineChart.

tensorboard/components/vz_line_chart2/vz-line-chart2.ts
```diff
@@ -411,10 +411,10 @@ namespace vz_line_chart2 {
...
         var chart = new LineChart(
-          xComponentsCreationMethod,
+          normalXComponentsCreationMethod,
           this.yValueAccessor,
-          yScaleType,
-          colorScale,
+          this.yScaleType,
+          this.colorScale,
           this.$.tooltip,
           this.tooltipColumns,
```
stephanwlee, line 419: I might make sense to remove these from the constructor but definitely not a required change.

psybuzz: Ack, will keep these constructor params for now.

PR #3833, opened 2020-07-12
[DebuggerV2] Refactor and clean up stack-frame-related types

caisq:
  * Make the types related to stack frames in Debugger V2 clearer;
     Tackle a clean-up TODO item.
  * Rename `SourceLineSpec` to `StackFrame`, and use it throughout
    the ngrx store. 
  * Rename `StackFrame` to `StackFrameAsArray`; it is used only at the
    place where the ngrx store talks to the data source (backend).
  * The new `StackFrame` and `StackFrameAsArray` now both
     contain `function_name` as a field, which is an improvement in
     consistent from the previous state, where one of the two types
     contained function name.
* Testeds
  * Unit tests are revised accordingly and continue to pass.
  * Manual testing on tfdbg2 logdirs.


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -94,7 +94,10 @@ import {
   getPollSilenceTimeMs,
   getSourceFileListLoaded,
 } from '../store/debugger_selectors';
-import {beginEndRangesInclude} from '../store/debugger_store_utils';
+import {
+  beginEndRangesInclude,
+  stackFrameAsArray2StackFrame,
```
stephanwlee, line 99: `stackFrameAsArray2StackFrame` -> `stackFrameAsArrayToStackFrame`? `2`, to me, can mean either: (1) version 2 (2) `to` or (3) two dimensions of the array

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -638,8 +641,9 @@ export class DebuggerEffects {
             // TODO(cais): Do this reshaping in the backend and simplify
             // the frontend code here.
             for (let i = 0; i < stackFrameIds.length; ++i) {
-              stackFramesById[stackFrameIds[i]] =
-                stackFramesResponse.stack_frames[i];
+              stackFramesById[stackFrameIds[i]] = stackFrameAsArray2StackFrame(
+                stackFramesResponse.stack_frames[i]
+              );
```
stephanwlee, line 646: Convert the backend data structure into the FE at the data source level, please.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
```diff
@@ -451,10 +446,18 @@ export interface SourceFileSpec {
 }
 
 // A specific line of a source-code file, i.e., a stack frame.
-export interface SourceLineSpec extends SourceFileSpec {
+export interface StackFrame extends SourceFileSpec {
   lineno: number;
+  function_name: string;
```
stephanwlee, line 451: if this is truly FE only type, use camelCase? `focusLineSpec` is camel case :\  
A counter argument can be that `SourceFileSpec` already defines snake_cased props.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/testing/index.ts
```diff
@@ -109,15 +110,33 @@ export function createTestGraphOpInfo(
...
+    function_name:
+      function_name || `function_${Math.floor(Math.random() * 1e3)}`,
+  };
+}
+
+export function createTestStackFrameAsArray(
+  host_name?: string,
+  file_path?: string,
+  lineno?: number,
+  function_name?: string
```
stephanwlee, line 131: prefer an object to a builder: I don't want to refer to this code when understanding the positional argument; it is better to have the name as a key to the argument.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_store_utils.ts
```diff
@@ -42,37 +42,67 @@ export function findFileIndex(
...
+  return [
+    stackFrame.host_name,
+    stackFrame.file_path,
+    stackFrame.lineno,
+    stackFrame.function_name,
+  ];
+}
+
+/**
+ * Convertsthe array representation of a stack frame into the object
```
stephanwlee, line 63: `Convertsthe` -> `Converts the`

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_store_utils.ts
```diff
@@ -42,37 +42,67 @@ export function findFileIndex(
...
+ * representation.
+ * @param array
+ * @returns StackFrame
```
stephanwlee, line 66: These are already strongly typed in the parameter/returnType. Omit them (they aren't valid Closure types either; here and above)?

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_store_utils.ts
caisq: Done.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_store_utils.ts
caisq: Removed them.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/testing/index.ts
caisq: Done, and changed all the call sites as well.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_types.ts
caisq: I agree that for 100% consistency, these types should be all lowerCamelCase. However, I'd like to defer it to later because of the following reasons:
1. The LOC affected by this will be large.
2. There are several other data types in this file which use snake_case in their field names. If I change the field names only for this type, it'll create some inconsistency.

caisq:
@stephanwlee To answer your question, function name is always defined.
- For lines that do not belong to a function but inside a .py file, it is "<module>"
- For lines in lambdas, it is "<lambda>"
- For other special cases, such as lines inside a jupyter notebook cell, it is also populated as a non-empty string.

PR #3834, opened 2020-07-13
HParams: display "no data" message, like other plugins

davidsoergel:


tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
```diff
@@ -617,17 +635,36 @@
...
+            }
+            this.set('_experiment', experiment);
+            // Doing this immediately, not only in finally(), allows the
+            // parent template to start rendering the results table before the
+            // table contents have been fetched.
+            this._computeDataFound();
+            this._computeHParams();
+            this._computeMetrics();
+            this._queryServer();
+            this._resolveGetExperiment();
```
psybuzz, line 653: I don't fully understand why `computeDataFound` needs to happen before making the requests.  Doesn't the `finally` get called in the same task as the `then`?  Looking at devtools, the `finally` code executes before the network request for /session_groups completes.

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
```diff
@@ -447,6 +447,24 @@
...
+          type: Boolean,
+          value: false,
+          notify: true,
+        },
+
+        noDataFound: {
+          type: Boolean,
+          value: false,
+          notify: true,
+        },
```
psybuzz, line 466: Hm, could we avoid confusing cases (`noDataFound === false`) if we had:

- hasHparamsOrMetrics: default false, becomes true if the latest response has 1+ hparams/metrics
- hparamsOrMetricsDataLoaded: default false, becomes true after receiving our first response

?

tensorboard/plugins/hparams/tf_hparams_parallel_coords_plot/tf-hparams-parallel-coords-plot.html
```diff
@@ -196,7 +196,7 @@
       observers: ['_optionsOrSessionGroupsChanged(options.*, sessionGroups.*)'],
 
       _optionsOrSessionGroupsChanged() {
-        if (this.options === null) {
```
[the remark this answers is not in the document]
wchargin: FWIW, the general rule that I’m familiar with is “always use `===`
except specifically for `== null`, because it is the same\* as
`=== null || === undefined`, which is easy to understand and often what
you want”. This is common enough to have an [ESLint option][esl], and is
also [called out in the Google style guide][gs].

\* Except specifically that `document.any == null` because we simply
cannot have nice things.

[esl]: https://eslint.org/docs/rules/eqeqeq#allow-null
[gs]: https://google.github.io/styleguide/jsguide.html#features-equality-checks-exceptions

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
```diff
@@ -765,6 +802,9 @@
       // above. Returns a promise, for use in unit-tests, that resolves (with
       // 'undefined') when the ListSessionGroups RPC returns.
       _queryServerNoDebounce() {
+        if (!this._hparams) {
```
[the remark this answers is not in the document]
davidsoergel: That's not valid.  `_sendListSessionGroupsRequest` calls `_buildListSessionGroupsRequest`, which requires that `_hparams` is defined at line 930.  In fact, thanks for pointing out that `_metrics` must also be defined, at line 957, so this condition should be more stringent.

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
```diff
@@ -617,17 +635,36 @@
...
+          .finally(() => {
+            this._computeDataFound();
+          });
+      },
+
+      _computeDataFound() {
+        const result = Boolean(
+          this._experiment &&
+            this._experiment.hparamInfos &&
+            this._experiment.hparamInfos.length > 0
```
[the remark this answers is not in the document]
davidsoergel: As below, it's the converse: both must be populated, not just one or the other.  Strengthened here too.

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
davidsoergel: Oh!  Yes, great, thanks.

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
davidsoergel: That would indeed be clearer here.  The problem arises where these are used, in `tf-hparams-main`, where we have things like `<template is="dom-if" if="[[_noDataFound]]">`.  The issue is that the condition of `dom-if` can be a concrete property or a function call, or the negation of those, but not any more complicated expression.  Thus, if we use the clearer state properties here, we'd have to introduce functions like `dataFound(hasHparamsOrMetrics, hparamsOrMetricsDataLoaded) { return hasHparamsOrMetrics && hparamsOrMetricsDataLoaded; }`.  Preparing the properties concretely here is a bit hacky, but avoids that additional complication.

Added a comment about the confusing case; hope that helps.

tensorboard/plugins/hparams/tf_hparams_parallel_coords_plot/tf-hparams-parallel-coords-plot.html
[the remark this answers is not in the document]
davidsoergel: At some point in the sequence of state changes, `this.options` became `undefined`, not `null`.  This caused `this.options.configuration` (immediately below) to choke.  I didn't track down the root cause, but just figured that we should treat `undefined` the same anyway, and that seemed to work.  Indeed, `!this.options` works too.

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
psybuzz: Ah, Polymer template expressions are so limited :/

The extra complexity of `dataFound(hasHparamsOrMetrics, hparamsOrMetricsDataLoaded)` seems reasonable to me.  However if you prefer the simpler 2 properties, could we rename them to something more explicit?
e.g.
`dataFound` --> `hasNonEmptyHparamsOnDataLoaded` or `dataLoadedWithNonEmptyHparams`
`noDataFound` --> `hasEmptyHparamsOnDataLoaded` or `dataLoadedWithEmptyHparams`

or similar would be much less confusing to me, esp in tf-hparams-main.html.

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
```diff
@@ -765,6 +804,9 @@
       // above. Returns a promise, for use in unit-tests, that resolves (with
       // 'undefined') when the ListSessionGroups RPC returns.
       _queryServerNoDebounce() {
+        if (!this._hparams || !this._metrics) {
+          return null;
```
psybuzz, line 808: nit: `return null` --> `return`, since nothing depends on the return value?

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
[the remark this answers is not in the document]
psybuzz: Gotcha, thanks for checking

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
davidsoergel: Sure, the renaming sgtm; done.

tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.html
davidsoergel: Done

PR #3835, opened 2020-07-13
Scalars: Multiplex fetch (one tag, many runs).

davidsoergel:
This dramatically speeds up scalars dashboard loading when there are many runs, and reduces load on our hosted services as well.


wchargin:
(haven’t reviewed patch, but assuming it does what it says on the tin—)
this is awesome, thanks! Was planning to do this in the next few weeks,
but even better to have it in earlier. :-)

davidsoergel:
Haha @wchargin I think you just volunteered to review ;)

davidsoergel:
Ah, sorry, need to fix tests.  Something weird happened with my local tests so I didn't trust it, but I see the CI also failed for obvious reasons.  Will do today or tomorrow, depending.

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -129,12 +137,33 @@ def tags_route(self, request):
...
+            tag = request.form["tag"]
+            json_runs = request.form["runs"]
+            try:
+                runs = json.loads(json_runs)
+            except Exception as e:  # pylint: disable=broad-except
+                # Different JSON libs raise different exceptions, so we just do a
+                # catch-all here. This problem is complicated by how Tensorboard might be
+                # run in many different environments, as it is open-source.
+                # TODO(@caisq, @chihuahua): Create platform-dependent adapter to catch
+                # specific types of exceptions, instead of the broad catching here.
```
wchargin, line 154: I see that this is cribbed from the debugger plugin; do we know whether
this is still an issue? Historically, Python 2 and 3 raised different
errors here (`ValueError` vs. the new `JSONDecodeError`), but if that
was the only issue then we should be fine to do the sane thing now. Is
there some terrifying Google-internal `json` monkey patching going on?
(If so, a Buganizer link would be nice.)

As written, this `Exception` also catches, e.g., `NameError`s and other
things that should clearly be 500s, as well as control flow like
`KeyboardInterrupt`s, so I’d like to avoid it if possible.
<https://google.github.io/styleguide/pyguide.html#244-decision>

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
```diff
@@ -188,10 +188,7 @@
       properties: {
         tag: String,
 
-        /**
-         * @type {Array<Object>}
-         */
-        dataToLoad: Array,
+        tagAndRuns: Object,
```
wchargin, line 191: Please document shape, especially since it’s _not_ what I would expect
from the name (which would be `{tag: string, runs: string[]}`).

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -129,12 +137,33 @@ def tags_route(self, request):
     @wrappers.Request.application
     def scalars_route(self, request):
         """Given a tag and single run, return array of ScalarEvents."""
-        tag = request.args.get("tag")
-        run = request.args.get("run")
+        if request.method == "GET":
+            tag = request.args.get("tag")
+            run = request.args.get("run")
+            runs = [run]
+        else:
```
wchargin, line 144: Can we explicitly check for POST here and return 405 otherwise?

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -129,12 +137,33 @@ def tags_route(self, request):
     @wrappers.Request.application
     def scalars_route(self, request):
         """Given a tag and single run, return array of ScalarEvents."""
-        tag = request.args.get("tag")
-        run = request.args.get("run")
+        if request.method == "GET":
```
wchargin, line 140: I drafted #3838 on Friday but wasn’t sure whether to send it; after
seeing the increased scope of the HTTP-specific parts of the request
handling in this PR, I think that it would be useful to land that first
so that this PR can be better tested. Does that sound reasonable to you?

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -95,29 +96,36 @@ def index_impl(self, ctx, experiment=None):
...
-        if scalars is None:
-            raise errors.NotFoundError(
-                "No scalar data for run=%r, tag=%r" % (run, tag)
-            )
-        values = [(x.wall_time, x.step, x.value) for x in scalars]
+        result = {}
+        for run in all_scalars:
+            scalars = all_scalars.get(run, {}).get(tag, [])
+            # Note we do not raise an error if data for a given run was not
+            # found; we just return an empty array in that case.
```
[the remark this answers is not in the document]
nfelt: FWIW, minor clarification that #3781 only introduced a new error for the bare multiplexer path, which has subsequently been removed by #3801.

So the change here seems mostly orthogonal to the consistency concern that led me to introduce that other error; we might still argue it's nice to be able to distinguish 404 from empty array here but I don't think there's a consistency concern.

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -95,29 +96,36 @@ def index_impl(self, ctx, experiment=None):
...
+            values = [(x.wall_time, x.step, x.value) for x in scalars]
+            result[run] = values
+
         if output_format == OutputFormat.CSV:
+            if runs.length > 1:
```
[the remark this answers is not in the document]
nfelt: FWIW, this did fail a test (`test_download_url_csv`).

tensorboard/plugins/scalar/scalars_plugin.py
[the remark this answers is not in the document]
wchargin: Yes, good point; thanks. There’s definitely not a consistency concern
anymore, I just still think that it’s a good idea. (By the reference
I mostly meant to point out that in your consistency fix you chose to
standardize on the error rather than the silently empty result, but my
wording was ambiguous; sorry.)

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.html
```diff
@@ -218,14 +215,17 @@
...
-              scalarChart.setSeriesMetadata(name, datum);
-              scalarChart.setSeriesData(name, formattedData);
+              for (const run in data) {
+                const points = data[run];
+                const formattedData = points.map((point) => ({
+                  wall_time: new Date(point[0] * 1000),
+                  step: point[1],
+                  scalar: point[2],
+                }));
+                const name = this._getSeriesNameForRun(run, datum.experiment);
```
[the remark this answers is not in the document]
davidsoergel: Cleaned up these ones, at least.

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -81,13 +81,22 @@ namespace tf_dashboard_common {
       requestData: {
         type: Function,
         value: function() {
-          return (datum) =>
-            this.requestManager.request(this.getDataLoadUrl(datum));
+          return (datum) => {
+            const dataLoadUrl = this.getDataLoadUrl(datum);
+            var url;
```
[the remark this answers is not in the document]
davidsoergel: Done

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -81,13 +81,22 @@ namespace tf_dashboard_common {
...
+            var postdata = {};
```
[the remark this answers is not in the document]
davidsoergel: Repro'd, but I don't understand it, and can't investigate further right now.  Will get back to it.

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -95,7 +97,7 @@ def index_impl(self, ctx, experiment=None):
                 }
         return result
 
-    def scalars_impl(self, ctx, tag, run, experiment, output_format):
+    def _scalars_impl(self, ctx, tag, run, experiment, output_format):
```
wchargin, line 100: No, this actually is public. The hparams plugin uses it:
<https://github.com/tensorflow/tensorboard/blob/9f87f60c317b8591dc4803b9d923616380d004b3/tensorboard/plugins/hparams/list_metric_evals.py#L54>

Whether it _should_ be public is a different question, but please don’t
change it in this patch.

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -129,12 +150,44 @@ def tags_route(self, request):
...
-        run = request.args.get("run")
+        if request.method == "GET":
+            tag = request.args.get("tag")
+            run = request.args.get("run")
+        else:
+            response = (
+                "%s requests are forbidden by the scalars plugin."
+                % request.method
+            )
+            return http_util.Respond(request, response, "text/plain", code=405)
```
wchargin, line 161: HTTP 405 responses are required to contain an `Allow` header listing the
allowed methods. FWIW, given that you’ve moved the POST handler to a
separate route, it’s fine with me if you want to omit the 405 checking
altogether (few other routes implement it). But if you do implement it,
please conform to spec.

(Also, if you do implement it, it seems confusing to say that GET/POST
requests are “forbidden by the scalars plugin” in both cases. Clearly
neither GET nor POST requests are forbidden by the plugin in general.)

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -69,6 +70,7 @@ def get_plugin_apps(self):
         return {
             "/scalars": self.scalars_route,
             "/tags": self.tags_route,
+            "/scalarsmulti": self.scalars_multirun_route,
```
wchargin, line 73: I like your choice of the name `scalars_multirun` for the handler
method. Could we maybe call the route `/scalars_multirun`, too? It seems
clearer than `scalarsmulti`, since it’s only “multi” along one of the
two extra axes (and of course the normal `/scalars` is already “multi”
along the time axis).

(There’s inconsistent precedent for both `camelCase` and `snake_case` in
route names (`individualImage`, `plugins_listing`), but most things use
snake case, and we don’t tend to use `lowersmooshcase`.)

tensorboard/plugins/scalar/scalars_plugin.py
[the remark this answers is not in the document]
wchargin: The comment here that we return an empty array for not-found data is not
correct with the multiplexer provider:

```
$ curl localhost:6006/data/plugin/scalars/scalarsmulti -F tag=xent/xent_1 -F runs='["lr_1E-03,conv=1,fc=2"]'
{"lr_1E-03,conv=1,fc=2": [[1563406405.3048136, 2000, 0.01962822675704956]]}
$ curl localhost:6006/data/plugin/scalars/scalarsmulti -F tag=xent/xent_1 -F runs='["lr_1E-03,conv=1,fc=2","nope"]'
{"lr_1E-03,conv=1,fc=2": [[1563406405.3048136, 2000, 0.01962822675704956]]}
```

This is because the multiplexer provider follows the relevant portion of
the data provider contract for `read_scalars`:

> The result will only contain keys for run-tag combinations that
> actually exist, which may not include all entries in the
> `run_tag_filter`.

For what it’s worth, I think that this is the sensible behavior. It’s
fine not to fail the request, but returning an empty array when we have
the opportunity to just omit the key is just introducing error.

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -119,6 +121,25 @@ def scalars_impl(self, ctx, tag, run, experiment, output_format):
...
+        all_scalars = self._data_provider.read_scalars(
+            ctx,
+            experiment_id=experiment,
+            plugin_name=metadata.PLUGIN_NAME,
+            downsample=self._downsample_to,
+            run_tag_filter=provider.RunTagFilter(runs=runs, tags=[tag]),
+        )
+        result = {}
+        for run in all_scalars:
+            scalars = all_scalars.get(run, {}).get(tag, [])
```
wchargin, line 135: nit: The fallback in `all_scalars.get(run, {})` is superfluous, since
we iterate `for run in all_scalars`. The fallback in `….get(tag, [])`
is, I think, also superfluous, since `run` won’t appear in the output at
all if it doesn’t have data for `tag` (since your `RunTagFilter` has a
singleton `tag` axis). So this could just be `all_scalars[run][tag]`,
which I think is clearer because the current version suggests codepaths
that cannot actually exist.

tensorboard/plugins/scalar/scalars_plugin_test.py
```diff
@@ -171,6 +171,21 @@ def test_scalars_with_scalars(self):
         self.assertEqual("application/json", response.headers["Content-Type"])
         self.assertEqual(self._STEPS, len(json.loads(response.get_data())))
 
+    def test_scalarsmulti_with_scalars(self):
+        server = self.load_server([self._RUN_WITH_SCALARS])
+        response = server.post(
+            "/data/plugin/scalars/scalarsmulti",
+            data={
+                "runs": json.dumps([self._RUN_WITH_SCALARS]),
```
wchargin, line 179: Nice test; thanks! Maybe include a run that doesn’t have scalar data
and/or a run that doesn’t have any data to test those edge cases, since
it’d be easy?

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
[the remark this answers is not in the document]
davidsoergel: Fixed

tensorboard/plugins/scalar/scalars_plugin_test.py
davidsoergel: Done.

tensorboard/plugins/scalar/scalars_plugin.py
davidsoergel: yep, crossed signals-- done already

tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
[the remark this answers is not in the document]
davidsoergel: Fixed previously (crossed signals); resolving

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -119,6 +121,26 @@ def scalars_impl(self, ctx, tag, run, experiment, output_format):
...
+            plugin_name=metadata.PLUGIN_NAME,
+            downsample=self._downsample_to,
+            run_tag_filter=provider.RunTagFilter(runs=runs, tags=[tag]),
+        )
+        result = {}
+        for run in all_scalars:
+            scalars = all_scalars[run][tag]
+            # Note we do not raise an error if data for a given run was not
+            # found; we just omit the run that case.
+            if scalars:
```
wchargin, line 138: What’s the purpose of this `if scalars` check? We know that `scalars` is
a list of `ScalarDatum` values (e.g., it can’t be `None`), so this omits
the run if it has empty data. But then a missing time series is
indistinguishable from a time series that is simply empty.

It’s (currently) true that in both the multiplexer data provider and
TensorBoard.dev we don’t keep track of empty time series, but there’s no
such restriction in the data provider model,\* so I don’t see why we
should go out of our way to lose information here. Can we just drop the
`if`-condition?

Sorry if I was unclear earlier; what I meant to convey is that the
structure of the data provider methods already has a simple, lossless
way to keep track of which time series exist and whether they have data,
and that plugin code should try to stick to that when possible. In most
cases this should generally amount to just doing the straightforward
thing without extra conditions.

\* If you want a motivating example, imagine some tuning service that
knows ahead of time what metrics you want to track.

tensorboard/plugins/scalar/scalars_plugin_test.py
wchargin: This works; thanks! I actually meant that it would suffice to just set
`runs` to `[_RUN_WITH_SCALARS, _RUN_WITH_HISTOGRAMS]` and keep the same
expected output. (That way, you’re also actually testing the multi-run
functionality of the `/scalars_multirun` route.) But having multiple
test cases is also perfectly fine with me if you prefer that.

tensorboard/plugins/scalar/scalars_plugin.py
davidsoergel: OK, I cleaned up this case-- thanks for clarifying.  I see now that what confused me is line 109 in the single-run case, which (per your other comment) suggests impossible states, e.g. the run is populated but the tag is not.  IIUC, that one would also be clearer as

```
if !all_scalars:
    raise NotFoundError(...)
scalars = all_scalars[run][tag]
```

(Not changing it here, though, so as not to pollute the PR.)

davidsoergel:
Unfortunately this implementation breaks the caching in data_loader_behavior.ts.  Setting aside for now.

wchargin:
Superseded by #4050.

PR #3838, opened 2020-07-14
scalars: refactor tests to use public interfaces

wchargin:
Summary:
In #3819, I made a cross-cutting change to all the plugins, and was
struck by how much easier it was to update the plugins that were tested
via their public APIs (mostly WSGI APIs) rather than internal `*_impl`
methods. Tests for (e.g.) the images plugin didn’t require any changes
at all. These tests are also more comprehensive, as they cover the
pipeline including more plumbing, parsing, and serialization. This patch
refactors the tests for the scalars plugin to use that form; if we like
it, we can make similar changes for other plugins, too.

Test Plan:
Tests still pass, and fail when (e.g.) the `run` and `tag` arguments are
swapped in `scalars_route`’s call to `scalars_impl`.

wchargin-branch: scalars-public-test


wchargin:
cc @nfelt, cc @davidsoergel re: #3835

wchargin:
Thanks, @stephanwlee ! Will merge after #3839 fixes CI.

PR #3844, opened 2020-07-15
migration prep: introduce relaxed tsconfig

stephanwlee:
Our Polymer based UI is currently built by our custom tooling that
incorporates tsc and JSCompiler. The build macro had a built-in tsconfig
which was quite lax and allows improper typing. This change introduces
the relaxed tsconfig based on internal rule, the minimum bar we should
try to meet when migrating.

### DIFF
--- a/tensorboard/defs/defs.bzl
+++ b/tensorboard/defs/defs.bzl
@@ -45,10 +45,19 @@ def tf_ts_config(**kwargs):
 
     ts_config(**kwargs)
 
-def tf_ts_library(**kwargs):
-    """TensorBoard wrapper for the rule for a TypeScript library."""
+def tf_ts_library(strict_checks = True, tsconfig = None, **kwargs):
+    """TensorBoard wrapper for the rule for a TypeScript library.
 
-    ts_library(**kwargs)
+    Args:
+      strict_checks: whether to enable stricter type checking. Default is True.
+          Please use `strict_checks = False` for only Polymer based targets.
+      tsconfig: TypeScript configuration. If specified, it overrides strict_checks.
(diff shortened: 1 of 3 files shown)


tensorboard/defs/defs.bzl
```diff
@@ -45,10 +45,19 @@ def tf_ts_config(**kwargs):
...
+def tf_ts_library(strict_checks = True, tsconfig = None, **kwargs):
+    """TensorBoard wrapper for the rule for a TypeScript library.
+
+    Args:
+      strict_checks: whether to enable stricter type checking. Default is True.
+          Please use `strict_checks = False` for only Polymer based targets.
+      tsconfig: TypeScript configuration. If specified, it overrides strict_checks.
+      **kwargs: keyword arguments to ts_library build rule.
+    """
+    if not strict_checks and not tsconfig:
```
nfelt, line 57: Optional, this might be slightly clearer as:

```python
if not tsconfig:
  tsconfig = "//:tsconfig.json" if strict_checks else "//:tsconfig-lax"
ts_library(tsconfig = tsconfig, **kwargs)
```

tsconfig-lax.json
```diff
@@ -0,0 +1,18 @@
+{
+  "compilerOptions": {
```
nfelt, line 2: Is there a specific place that we came up with this list of options from?  If so, might be nice to document as a comment on the BUILD target.

stephanwlee: I left a short sentence on it but I don't think I can comment much on it more than I did.

nfelt: Ok, no problem.

PR #3847, opened 2020-07-16
data: add dispatching data provider

wchargin:
Summary:
A new `DispatchingDataProvider` composes multiple sub-providers under
different experiment ID prefixes. Thus, if `FooDataProvider` expects
experiment IDs like `123` and `BarDataProvider` expects IDs like `abc`,
you could use a `DispatchingDataProvider` to simultaneously expose
experiment IDs like `foo:123` and `bar:abc`.

Most method implementations of `DispatchingDataProvider` simply involve
determining the appropriate sub-provider and forwarding all arguments.
We implement those with a tiny bit of metaprogramming (since there are
nine such methods), stopping short of [more drastic measures][ddd].

[ddd]: https://wchargin.github.io/posts/dubious-dynamic-delegation-in-python/

Test Plan:
Unit tests included. For Google-internal use, <http://cl/321472166>
shows an application of this data provider.

wchargin-branch: data-dispatching-provider


tensorboard/data/dispatching_provider.py
```diff
@@ -0,0 +1,211 @@
...
+        result = sub_provider.read_blob_sequences(
+            *args, experiment_id=sub_eid, **kwargs
+        )
+        for tag_to_data in result.values():
+            for (tag, old_data) in tag_to_data.items():
+                new_data = [
+                    provider.BlobSequenceDatum(
+                        step=d.step,
+                        wall_time=d.wall_time,
+                        values=_convert_blob_references(prefix, d.values),
```
wchargin, line 123: I’m open to inlining this method if reviewers think that that would be
clearer. It’d drop a top-level method and docstring and shorten the code
overall, at the cost of a nested loop comprehension here.

tensorboard/data/dispatching_provider_test.py
```diff
@@ -0,0 +1,337 @@
...
+import base64
+
+from tensorboard import errors
+from tensorboard import context
+from tensorboard import test as tb_test
+from tensorboard.data import dispatching_provider
+from tensorboard.data import provider
+
+
+class FiniteProvider(provider.DataProvider):
```
wchargin, line 26: I’m curious to hear what reviewers think of this test pattern. My
thoughts:

  - I like that we’re testing the observable behavior of the real code
    with valid data providers, rather than relying on mocking or spying
    or anything brittle like that.
  - I considered using the existing multiplexer provider as sub-provider
    for these tests, but it felt inappropriate because the multiplexer
    provider doesn’t care about experiment IDs at all (not to mention
    that it touches disk).
  - Because the data provider interface is fairly large (many methods,
    many parameters), this test provider is fairly long, even though we
    skip tensors entirely.
  - In general, I like to introduce a kind of “configurable in-memory
    implementation” for things like this, so that in the test code you
    can say “add an experiment with this ID and this kind of data”. We
    don’t yet have one of those for the `DataProvider` interface itself.
    Would now be a good time to add one?

Overall, I think that these tests are reasonably robust, but I’m open to
ideas on how to improve them, along any axis.

tensorboard/data/dispatching_provider_test.py
```diff
@@ -0,0 +1,337 @@
...
+                _ctx(), experiment_id=experiment_id, plugin_name="images"
+            )
+            reading = dp.read_blob_sequences(
+                _ctx(),
+                experiment_id=experiment_id,
+                plugin_name="images",
+                downsample=10,
+                run_tag_filter=provider.RunTagFilter(
+                    runs=list(listing),
+                    tags=sorted(set().union(*listing.values())),
```
nfelt, line 278: This feels a wee bit fancy for a test helper?  I can parse it, but it might be simpler to just make run and tag parameters to the helper method [1] or to leave then empty to request all runs and tags.  See also other comments but I'd probably try to keep the test logic here fairly focused on the parts that DispatchingProvider does differently versus things that are more like generic DataProvider behavior expectations.

[1] Another option rather than making them parameters is just to down-scope the helper to postprocessing the `read_blob_sequences()` response and making the `read_blob()` calls, since that's where I think it has the most bang for the buck, and just leaving the `{list,read}_blob_sequences()` calls as being directly invoked in the test method.

tensorboard/data/dispatching_provider.py
```diff
@@ -0,0 +1,211 @@
...
+    Args:
+      prefix: The prefix of the sub-provider that generated the sub-key,
+        or `None` if this was generated by the unprefixed provider.
+      sub_key: The opaque key from the sub-provider.
+
+    Returns:
+      A string encoding `prefix` and `sub_key` injectively.
+    """
+    payload = [prefix, sub_key]
+    json_str = json.dumps(payload, separators=(",", ":"))
```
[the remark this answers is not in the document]
nfelt: Ah, that makes sense and I hadn't thought about the unprefixed case, good point.  Keeping as is SGTM.

tensorboard/data/dispatching_provider_test.py
nfelt: FWIW, optional especially since it's merged already but I think the "assert equality to the bare provider response" approach would also work for the blobs, right?  Since the only part of the response being altered is the blob keys, but those are invisible to the assertion since we feed them back into the same provider.

wchargin: Right again; thanks! Sent #3858 (but entirely non-urgent).

Issue #3840, opened 2020-07-15
Projector plugin hangs with "Fetching sprite image..."


davidsoergel:
Projector plugin hangs with "Fetching sprite image...", using our standard test demodir.  I don't know the cause yet, or how long this has been an issue.

![Screen Shot 2020-07-14 at 8 29 25 PM](https://user-images.githubusercontent.com/1347381/87491340-e497fc00-c615-11ea-8e44-c1b0cf0e932c.png)

psybuzz:
Using the team's demo logdir, I can reproduce this error using an older version from ~July 8th.

The sprite image request fails with 400, and surprisingly it takes 10+ minutes before I can see the network error.  For some reason it's taking a long time to fetch any run, tensor data.

![image](https://user-images.githubusercontent.com/2322480/87596980-da254300-c6a5-11ea-9468-8681c263a88c.png)

psybuzz:
The projector plugin backend responds with:
`.../demodir/__debugger_data__/sprite.png" does not exist or is directory`
Still investigating, but adding @hfiller and @caisq for projector + debugger expertise.

The demo logdir contains both projector data, and debugger v1 summaries (assumed from the "__debugger_data__" run appearing in the network logs).  I can confirm that
- running the latest TB against the demo logdir stalls when Projector tries to load the __debugger_data__ run
- running the latest TB in a separate logdir with only projector data works fine

Conclusion
- Either the Projector backend is incorrectly trying to access Debugger v1 runs, or Debugger v1 runs are confusing the Projector backend
- I still need to check how far back this has been happening

psybuzz:
Confirmed that this issue is present at least as far back as June 8, so it's not a recent regression.

psybuzz:
Manually bisected to the change https://github.com/tensorflow/tensorboard/pull/3653 (on June 1), which was not included in the recent TB 2.2.2 release (on May 28).

@hfiller is the PR above something we can revert safely, or is there a clear fix?

hfiller:
This PR can safely be reverted.

juliagong:
Hi! I'm having this issue in tensorboard (~28,000 images and embeddings to be displayed in the projector). I'm using tensorboard version 1.15.0 and am wondering what the issue could be? The loading of tensors and metadata is fast as usual, and then it hangs on "Fetching sprite image...". This didn't happen when I had only 128 images and embeddings. Thanks!

lutzzz:
[...]

PR #3850, opened 2020-07-16
Revert "[Projector] make root directory config default for available runs"

psybuzz:
This reverts commit 12f6234e0608d7b8703540fb3473372de88f788e.

The changes in https://github.com/tensorflow/tensorboard/pull/3653 allowed TensorBoard to
use the root directory's projector config and apply it to each subdirectory.

If the root dir's config specifies a "sprite" that does not exist in every subdirectory, it
causes the Projector frontend to hang when loading a run.
See https://github.com/tensorflow/tensorboard/issues/3840 for details on the issue.

Confirmed that this fixes the issue by running TB on a directory with:
- ./projector_config.pbtxt
```
embeddings {
  tensor_name: "EMNIST_Letters:0"
  metadata_path: "metadata.tsv"
  sprite {
    image_path: "sprite.png"
    single_image_dim: 28
    single_image_dim: 28
  }
}
```

- alongside a subdirectory `profile_demo_new` that contains a `.profile-empty` event file but no `sprite.png` file
image

### DIFF
--- a/tensorboard/plugins/projector/projector_plugin.py
+++ b/tensorboard/plugins/projector/projector_plugin.py
@@ -434,22 +434,6 @@ class ProjectorPlugin(base_plugin.TBPlugin):
     def _read_latest_config_files(self, run_path_pairs):
         """Reads and returns the projector config files in every run
         directory."""
-        """If no specific config exists, use the default config provided in
-        the root directory."""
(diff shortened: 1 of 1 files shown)


psybuzz:
cc @hfiller

PR #3851, opened 2020-07-16
lib: make `errors` submodule lazy

wchargin:
Summary:
All the other submodules are lazy. We originally made `errors` eager
because it seemed like a small surface (#2631), but some Google-internal
Python 2 binaries unnecessarily transitively depend on `errors` because
of this edge. Moving it to a lazy dependency makes it easier to make
`errors` use py3-only features.

Test Plan:
Existing smoke test in `test_pip_package.sh` suffices.

wchargin-branch: lib-lazy-errors

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -43,6 +43,7 @@ py_library(
     visibility = ["//tensorboard:internal"],
     deps = [
         ":lib_init_only",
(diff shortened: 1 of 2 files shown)


wchargin:
Googlers: test sync also passes: <http://cl/321625168>.

PR #3852, opened 2020-07-16
errors: add `UnauthenticatedError`

wchargin:
Summary:
We add a new type to distinguish authentication errors (“we can’t
confirm who you are”) from authorization errors (“we know who you are,
but you’re not allowed to perform this operation”). This was omitted
from the initial draft of the `errors` module because the natural status
code is HTTP 401, but HTTP 401 responses are required by RFC 7235, §3.1
to include a `WWW-Authenticate` challenge header. In this patch, we add
the necessary error handling machinery to support that header.

Test Plan:
Unit tests suffice for the core implementation. As a manual test,
changing a `NotFoundError` to an `UnauthorizedError` with an arbitrary
challenge in `scalars_plugin.py` properly sends a 401 with the intended
response and challenge header. Since this patch changes `:errors` to
include Python 3 syntax, Google-internal <http://cl/321666883> witnesses
that there are no more internal Python 2 dependents after #3851.

wchargin-branch: errors-unauthenticated

### DIFF
--- a/tensorboard/errors_test.py
+++ b/tensorboard/errors_test.py
@@ -36,6 +36,12 @@ class InvalidArgumentErrorTest(tb_test.TestCase):
     def test_http_code(self):
         self.assertEqual(errors.InvalidArgumentError().http_code, 400)
 
+    def test_headers(self):
+        e1 = errors.InvalidArgumentError()
+        e2 = errors.InvalidArgumentError()
+        self.assertEmpty(e1.headers)
+        self.assertIsNot(e1.headers, e2.headers)
+
 
 class NotFoundErrorTest(tb_test.TestCase):
     def test_no_details(self):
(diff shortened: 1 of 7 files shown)


tensorboard/errors_test.py
```diff
@@ -36,6 +36,12 @@ def test_with_details(self):
     def test_http_code(self):
         self.assertEqual(errors.InvalidArgumentError().http_code, 400)
 
+    def test_headers(self):
+        e1 = errors.InvalidArgumentError()
+        e2 = errors.InvalidArgumentError()
+        self.assertEmpty(e1.headers)
+        self.assertIsNot(e1.headers, e2.headers)
```
psybuzz, line 43: Why is it important to check that 2 equivalent errors do not have the exact same headers reference?

wchargin: The way I’ve written it, subclasses add headers by appending to
`self.headers`. (I figured that this was simplest.) If these references
were shared, then in

```python
e1 = errors.UnauthenticatedError(challenge="one")
e2 = errors.UnauthenticatedError(challenge="two")
```

both `e1` and `e2` would have a `headers` list with two entries (one
with challenge `one`, one with challenge `two`) due to the mutation
pollution.

In general, this is just “avoid shared mutable state”. The test fails if
you write the initialization as

```python
class PublicError(RuntimeError):
    http_code = 400
    headers = []
```

rather than

```python
class PublicError(RuntimeError):
    http_code = 400
    def __init__(...):
        self.headers = []
```

even though the former looks reasonable to an untrained eye.

wchargin:
> > Note that the <http://cl/321642352> tests might be failing?

> Oof, thank you. Will fix.

Ah: it’s only failing because I didn’t include #3851 in the test sync. A
new version in <http://cl/321666883> should pass, but I’ll wait for it
to make sure.

wchargin:
Actually passed now. Thanks again.

PR #3855, opened 2020-07-17
application: add injection point for middleware

wchargin:
Summary:
Clients of `TensorBoardWSGIApp` may now inject middleware around the
core TensorBoard app routing but still within the experiment ID, path
prefix, request context, etc. middlewares. This is useful for handlers
for extra routes that need to access the request context.

Test Plan:
Unit tests included.

wchargin-branch: application-middleware-injection

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -71,7 +71,7 @@ def TensorBoardWSGIApp(
     assets_zip_provider=None,
     deprecated_multiplexer=None,
     auth_providers=None,
-    middlewares=None,
+    experimental_middlewares=None,
 ):
     """Constructs a TensorBoard WSGI app from plugins and data providers.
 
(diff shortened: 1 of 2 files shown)


wchargin:
Renamed `extra_routes` to `middlewares` per suggestion of @nfelt .

wchargin:
(WANT_LGTM=any, I think)

tensorboard/backend/application.py
```diff
@@ -91,6 +92,13 @@ def TensorBoardWSGIApp(
       auth_providers: Optional mapping whose values are `AuthProvider` values
         and whose keys are used by (e.g.) data providers to specify
         `AuthProvider`s via the `AuthContext.get` interface. Defaults to `{}`.
+      middlewares: Optional list of WSGI middlewares (i.e., callables that take
```
nfelt, line 95: Maybe it would be prudent to explicitly mark this parameter as "experimental" for now?  The `TensorBoardWSGIApp` API is still basically a private API, but I think the rest of the signature is a bit more established so it might be worth differentiating this one, given that based on internal conversation it doesn't seem like we've fully settled how we want to manage the middleware tower in the long term.

(E.g. I could imagine that at some point perhaps we'd rather have extra routes implemented via an actual `extra_routes` parameter taking a mapping of path to WSGI app, so that would no longer require passing `middlewares` - or we could decide to peel off the outer middlewares from the return value of this function and instead require callers to apply those "by hand" at which point this becomes obsolete.)

wchargin: Sure, that’s fine with me. Agreed that I’m not sure exactly what the
right form here is. Done.

PR #3863, opened 2020-07-21
run selector interop: populate initial state

stephanwlee:
Bug: currently, the initial state of the legacy runs selector is not reflected in the store.
Fix: read the state from the Polymer component and fire an action.


stephanwlee:
@Sparkier

tensorboard/webapp/runs/views/legacy_runs_selector/legacy_runs_selector_component.ts
```diff
@@ -49,5 +49,9 @@ export class LegacyRunsSelectorComponent implements AfterViewInit {
         this.onSelectionChange.emit(event.detail.value as string[]);
       }
     );
+    setTimeout(() => {
+      // dispatch the initial value from the component.
```
bmd3k, line 53: Nit: Capital "D" "Dispatch".

PR #3864, opened 2020-07-21
Text demo v2

pjfossee:
Update the demo file to better test the text plugin's frontend UI.  This will help in creating/testing the new Angular frontend UI for the text plugin.  Also ports the demo to TF V2 to make it easier to use (no longer need to switch TensorFlow versions to use it)

Adds multiple runs and a function to test the pagination view in the frontend UI.

* Screenshots of UI changes
image

Run `python text_demo_v2.py` to create a logfile in the folder `/tmp/text_demo`

None, this is a pretty straightforward update to the original demo file.


tensorboard/plugins/text_v2/text_demo_v2.py
```diff
@@ -0,0 +1,146 @@
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
stephanwlee, line 1: Talked offline. We would like everything to be runnable by bazel. Please do one of two things:
1. remove `text_demo_v2.py` and just combine this into `text_demo.py` and test via `bazel run tensorboard/plugins/text:text_demo`
2. create a bazel target

tensorboard/plugins/text_v2/text_demo_v2.py
```diff
@@ -0,0 +1,146 @@
...
+    # The result, `table_full`, is a rank-2 string tensor of shape
+    # `[step + 1, step + 1]`. We can pass it directly to the summary, and
+    # we'll get a nicely formatted table in TensorBoard.
+    text("multiplication_table", table_full, step)
+
+
+def create_run(logdir, run_name):
+    # Creates a run specified by `run_name` populated with the text summaries
+    # generated by the function above.
+    writer = tf.summary.create_file_writer(logdir + "/" + run_name)
```
stephanwlee, line 117: Question to @nfelt: it seems like demos are hard coding logdir like `/tmp/images_demo` which probably won't work on Windows. Is there a reason why we did that instead of using https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp and `os.path.join`?

tensorboard/plugins/text_v2/text_demo_v2.py
```diff
@@ -0,0 +1,146 @@
...
+# Number of steps for which to write data.
+STEPS = 16
+
+
+def simple_example(step):
+    # Text summaries log arbitrary text. This can be encoded with ASCII or
+    # UTF-8. Here's a simple example, wherein we greet the user on each
+    # step:
+    step_string = tf.as_string(step)
+    greeting = tf.strings.join(["Hello from step ", step_string, "!"])
```
stephanwlee, line 35: I think `tf.constant('hello from step %d!' % step)` might be simpler to write and read. Here and below. Also, feel free to put a jovial emoji to showcase the unicode :)

tensorboard/plugins/text_v2/text_demo_v2.py
```diff
@@ -0,0 +1,146 @@
...
+    # unless those tags are sub-tags of another tag.  To create a subtag name,
+    # one would input MainTagName/SubTagName as the input argument to the `text`
+    # summary API.  Subtags will be displayed within the main tag's dropdown,
+    # and a pager will appear if there are more than 12 subtags.  This test
+    # displays that UI feature by creating 36 subtags within the main tag:
+    # "simple_example_with_pagination"
+    for i in range(36):
+        step_string = tf.as_string(step)
+        greeting = tf.strings.join(["Hello from step ", step_string, "!"])
+        text("Card" + str(i + 1) + "/greeting", greeting, step)
```
stephanwlee, line 51: Use the Python string format for better readability. Similar to previous comment, basically.

tensorboard/plugins/text_v2/text_demo_v2.py
```diff
@@ -0,0 +1,146 @@
...
+    # doesn't actually have to be neatly aligned, so we can create these
+    # pretty easily. Let's do so.
+    header_row = "Pounds of chocolate | Happiness"
+    chocolate = tf.range(step)
+    happiness = tf.square(chocolate + 1)
+    chocolate_column = tf.as_string(chocolate)
+    happiness_column = tf.as_string(happiness)
+    table_rows = tf.strings.join([chocolate_column, " | ", happiness_column])
+    table_body = tf.strings.reduce_join(inputs=table_rows, separator="\n")
+    table = tf.strings.join([header_row, "---|---", table_body], separator="\n")
```
stephanwlee, line 73: While I don't have strong feelings, if you think it is easier to form a string in Python, then use `tf.constant`, feel free to refactor this block.

(btw, while this method is readable, the `higher_order_tensors` one look more challenging)

tensorboard/plugins/text_v2/text_demo_v2.py
pjfossee: moved to original demo file

tensorboard/plugins/text_v2/text_demo_v2.py
pjfossee: done

tensorboard/plugins/text_v2/text_demo_v2.py
pjfossee: done

tensorboard/plugins/text_v2/text_demo_v2.py
pjfossee: I think for forming markdown tables, the current way is more clear.  But I do agree it gets a little more funky with the higher_order_tensors test, not sure if converting to python string formatting would help much tho.

tensorboard/plugins/text/text_demo.py
```diff
@@ -39,8 +32,24 @@ def simple_example(step):
...
+    # Text summaries can be grouped by Tag (the first argument to the `text`
+    # summary's API).  Unique tags will have their own dropdown in the frontend,
+    # unless those tags are sub-tags of another tag.  To create a subtag name,
+    # one would input MainTagName/SubTagName as the input argument to the `text`
+    # summary API.  Subtags will be displayed within the main tag's dropdown,
+    # and a pager will appear if there are more than 12 subtags.  This test
+    # displays that UI feature by creating 36 subtags within the main tag:
+    # "simple_example_with_pagination"
+    for i in range(36):
+        step_string = tf.as_string(step)
```
stephanwlee, line 49: unused.

tensorboard/plugins/text/text_demo.py
```diff
@@ -39,8 +32,24 @@ def simple_example(step):
     # UTF-8. Here's a simple example, wherein we greet the user on each
     # step:
     step_string = tf.as_string(step)
```
stephanwlee, line 34: unusd here, too.

tensorboard/plugins/text/text_demo.py
pjfossee: done

tensorboard/plugins/text/text_demo.py
pjfossee: done

tensorboard/plugins/text_v2/text_demo_v2.py
nfelt: I missed this question from before.  It would be better if these worked on windows, but using `mkdtemp` would have the disadvantage that the directory is no longer predictable on posix (makes it a bit harder to do things like run the demo and then open TensorBoard pointed at the logdir).  So might be nice to keep the `/tmp/foo_demo` naming structure for that reason, but we could use `mkdtemp` on Windows?

wchargin: Original author here. The purpose was indeed to put the output in a
predictable location. There was also a plan to have a single “mega-demo”
target that would have run all the demos and put the output in a common
directory, hence the common `run(logdir)` structure to the demos, but we
never got around to that. I’m not inclined to go out of our way to
support Windows for _development_ processes, rather than end users,
given that none of us actually uses it and we’ve had minimal demand.

PR #3862, opened 2020-07-21
Update protos for TF v2.3.0-rc2

psybuzz:
While testing the pip candidate locally, it fails on the proto_test.

The helpful message says this expected when TF's protos update.
This change was created by checking out `git checkout v2.2.0-rc0`
in a TensorFlow local repo, and running
`./tensorboard/compat/proto/update.sh PATH_TO_TENSORFLOW_REPO`.

PR #3860, opened 2020-07-20
TensorBoard 2.3.0

psybuzz:
Prepared for the 2.3.0 pip package release.

Created from HEAD @ 0c88c1f8c98e68898bd5eabfed47e90ae1267a44
Cherrypicks: none

fyi @jameswex @qiuminxu @jindalshivam09

PR #3865, opened 2020-07-21
TensorBoard 2.3.0 (attempt 2)

psybuzz:
The original attempt (https://github.com/tensorflow/tensorboard/pull/3860) did not include important 2 PRs needed to get the build
and tests passing: https://github.com/tensorflow/tensorboard/pull/3861, https://github.com/tensorflow/tensorboard/pull/3862.

This 2nd attempt rebases the 2.3 branch to include all PRs up to and including those 2.

Note: In the Github UI, the 2.3 branch says "You can't delete this protected branch".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ env:
     - BAZEL_SHA256SUM=e13581d44faad6ac807dd917e682fef20359d26728166ac35dadd8ee653a580d
   matrix:
     - TF_VERSION_ID=tensorflow==1.15.0
-    - TF_VERSION_ID=tf-nightly
+    - TF_VERSION_ID=tensorflow==2.3.0rc2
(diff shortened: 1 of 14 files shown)


wchargin:
> In the Github UI, the 2.3 branch says "You can't delete this protected
> branch".

If we rebase and merge this as is, the top three commits (the release
ones) will be empty and, I think, dropped entirely. We should force-push
the `2.3` branch back to master at a64659b4a170, then update this PR to
contain just the release commits; no cherry picks needed this time
because they’re in master. (Equivalently, you could force-push to just
before the proto update and py38 fix and then cherry-pick those in here,
but that dupes the hashes and I don’t see a reason to do that.)

This is equivalent to deleting the branch and just re-cutting it from
master.

@nfelt, can you sanity-check me here?

PR #3874, opened 2020-07-22
Scalars plugin: Return 400 if run or tag is not specified

manivaradarajan:

If the "/data/plugin/scalars/scalars" route is hit without specifying the run or tag query parameters, TensorBoard will return 500. Corresponding errors were noticed in tensorboard.dev server logs. 

Internal Google bug reference: b/161680845

Scalars plugin now checks that tag and run are not None before proceeding with the request. If either is None, it raises InvalidArgumentError.

* Screenshots of UI changes

None relevant.

1) Run local TensorBoard with changes
2) Visit "http://localhost:6006/data/plugin/scalars/scalars?run=foo", "http://localhost:6006/data/plugin/scalars/scalars?tag=foo" and "http://localhost:6006/data/plugin/scalars/scalars"
3) Ensure that "Invalid argument: Both run and tag must be specified: tag=X, run=Y" are returned as appropriate

None


wchargin:
(I can review this; let me know when.)

tensorboard/plugins/scalar/scalars_plugin.py
```diff
@@ -22,21 +22,18 @@
 from __future__ import division
 from __future__ import print_function
 
-import collections
```
wchargin, line 25: Appreciate the import cleanup, but it’d be best to separate that out,
particularly because it should be accompanied by removing the build
edges (`//tensorboard/util:tensor_util`, et al.). Can you revert this
part of the diff? I can take care of the cleanup if you like.

tensorboard/plugins/scalar/scalars_plugin_test.py
```diff
@@ -171,6 +167,28 @@ def test_scalars_with_scalars(self):
         self.assertEqual("application/json", response.headers["Content-Type"])
         self.assertEqual(self._STEPS, len(json.loads(response.get_data())))
 
+    def test_scalars_with_scalars_unspecified_run(self):
+        server = self.load_server([self._RUN_WITH_SCALARS])
+        response = server.get(
+            "/data/plugin/scalars/scalars",
+            query_string={
+                "run": None,
```
wchargin, line 175: (FWIW, this `"run": None` does nothing; okay with me to leave it in if
you want.)

manivaradarajan: It reads slightly better as a test to me to leave it in and be explicit.

tensorboard/plugins/scalar/scalars_plugin_test.py
```diff
@@ -171,6 +167,28 @@ def test_scalars_with_scalars(self):
...
+                "tag": "foo_tag"
+            },
+        )
+        print("Response: ", response.status_code)
```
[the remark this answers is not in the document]
manivaradarajan: Thank you - crept in by accident.

tensorboard/plugins/scalar/scalars_plugin.py
manivaradarajan: Removed.

PR #3876, opened 2020-07-23
Change our Plottable monkey patch to TypeScript

stephanwlee:
This change is made so that we can more easily migrate to tf_ts_library.
Background: we have plans to migrate our build system from
tf_web_library to tf_ts_library. In doing so, we will, without this
change, make ts_library depend on js_library (this file) but that is
impossible since there is no such thing as js_library.

Since the module largely monkey patches and accesses private methods
we treated many violations with casting to `any` for now.


tensorboard/components/vz_chart_helpers/plottable-interactions.ts
```diff
@@ -49,10 +49,11 @@ THE SOFTWARE.
  * To override few quick private/protected methods, we had to use JavaScript to
  * bypass TypeScript typechecks.
  */
-var vz_chart_helpers;
-(function(vz_chart_helpers) {
+namespace vz_chart_helpers {
   // HACK: parentElement does not work for webcomponents.
-  function getHtmlElementAncestors(elem) {
+  function getHtmlElementAncestors(
+    elem: Element | Node | HTMLElement | DocumentFragment
```
psybuzz, line 55: optional: I'm curious, does a simpler `elem: Node` also work?

PR #3881, opened 2020-07-24
chore: remove beholder

### LINKED ISSUE #3843: Deprecating the Beholder plugin
The TensorBoard team plans to deprecate the [Beholder plugin](https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/beholder/README.md) contributed a few years ago by Chris Anderson (cc @chrisranderson).

What this means concretely is that we will mark the plugin as deprecated in our next release, and remove it in the following release.  We don't consider this a breaking change that requires a major version bump since Beholder has been indicated as only in beta since it was contributed.

We recognize that Beholder has value for users, but we haven't had the bandwidth to maintain the plugin and make necessary changes to update it to work with TF 2.0 or eager mode (#1237), or to continue keeping the frontend web components up-to-date as the TensorBoard frontend evolves.

If you are still using Beholder and would like to continue using it in newer versions of TensorBoard, we recommend pursuing converting it to a dynamic plugin as outlined in [the plugin guide](https://github.com/tensorflow/tensorboard/blob/master/ADDING_A_PLUGIN.md), which would mean hosting it in a separate git repository and publishing it to PyPI as a `tensorboard_plugin_beholder` package, which users can then install if they wish to still use Beholder.

stephanwlee:
As part of the TB 2.3.0 release, TensorBoard has announced that we would
discontinue support for beholder.

We would definitely welcome any contribution to make it work with TF 2.x
and the plugin as a dynamic plugin.

🙇‍♂️ and 👋 beholder. 

Fixes #3843.


stephanwlee:
Done!

PR #3885, opened 2020-07-27
npmi: demo Data Generator

Sparkier:
To be able to experiment with the plugin, this now provides a generator for demo data.

Added a generator for demo data. It generates random data points and saves them for use with the plugin.

* Screenshots of UI changes
no UI changes

Run `bazel run //tensorboard/plugins/npmi:npmi_demo` to generate the data.
Then run `bazel run -c dbg tensorboard -- --logdir /tmp/npmi_demo` to start the plugin with the generated data.

-


tensorboard/plugins/npmi/README.md
```diff
@@ -3,3 +3,7 @@
 The nPMI plugin is a frontend to the techniques presented in a [paper](https://svn.spraakdata.gu.se/repos/gerlof/pub/www/Docs/npmi-pfd.pdf) by Gelof Bouma.
 
 The plugin is currently experimental and under active development.
+
+## Demo Data
+
+To experiment with the plugin without the need to generate real log data, simply generate demo data using `bazel run //tensorboard/plugins/npmi:npmi_demo` and then start the plugin with `bazel run -c dbg tensorboard -- --logdir /tmp/npmi_demo`.
```
stephanwlee, line 9: `-c dbg` can be dropped now.

tensorboard/plugins/npmi/npmi_demo.py
```diff
@@ -0,0 +1,703 @@
...
+from absl import app
+import tensorflow as tf
+import numpy as np
+from tensorboard.plugins.npmi import summary
+
+tf.compat.v1.enable_eager_execution()
+
+# Directory into which to write tensorboard data.
+LOGDIR = "/tmp/npmi_demo"
+MOUNTAINS = [
```
stephanwlee, line 32: would it be possible for this to come from some other data file? It is a bit lengthy and distracting.

Sparkier: Completely agree, just didnt know how you would like this to be. Should this be another file I read from e.g. txt or something like that or a python file just holding this var?

manivaradarajan: Putting it into a separate text file is reasonable. In code you could specify that the input file is a list of entities, one entity per line.

nfelt: I see you already went with the text file, but it might be a bit easier to just create an `npmi_demo_data.py` file with this as a variable.  It's slightly less clean in the sense that it's really more plain data than python, but getting the correct path to a data file (especially that works both open source and internally) is a little finicky.  The current approach depends on the current working directory being the Bazel root, and while that's often the case it's a bit unfortunate to have the script fail if it isn't run from there.

AFAIK the existing precedent we have to get a path reference to it that works both externally and internally is this hacky line here:
https://github.com/tensorflow/tensorboard/blob/7511300d6ca5d25d332459d89e5725a47baf225b/tensorboard/program.py#L99-L101

tensorboard/plugins/npmi/BUILD
```diff
@@ -43,6 +43,19 @@ py_test(
...
+py_binary(
+    name = "npmi_demo",
+    srcs = ["npmi_demo.py"],
+    srcs_version = "PY2AND3",
+    deps = [
+        ":summary",
+        "//tensorboard:expect_tensorflow_installed",
+    ],
+    data = [
+        "mountains.txt"
```
nfelt, line 55: Maybe worth calling this demo_mountains.txt or something so that folks looking at the directory don't wonder whether this plays some critical role in the operation of the plugin :)

tensorboard/plugins/npmi/npmi_demo.py
```diff
@@ -0,0 +1,76 @@
...
+
+tf.compat.v1.enable_eager_execution()
+
+# Directory into which to write tensorboard data.
+LOGDIR = "/tmp/npmi_demo"
+
+
+def setup_run(logdir, verbose, run_name):
+    # Getting a list of dummy annotations from a txt file.
+    f = open("./tensorboard/plugins/npmi/mountains.txt", "r")
```
nfelt, line 36: Generally preferable to use a context manager so that closing the file correctly is handled automatically, plus a little more concise.

```python
with open("./tensorboard/plugins/npmi/mountains.txt", "r") as f:
    mountains = f.readlines()

Comment (nfelt) [inline tensorboard/plugins/npmi/npmi_demo.py:36]: See other comment, but this will only work when the demo script is run from the root directory.  Maybe nice in that case to catch an exception if this file doesn't exist and tell the user they need to be in the Bazel root directory.

File: tensorboard/plugins/npmi/npmi_demo.py [hunk 5]
```diff
@@ -0,0 +1,76 @@
...
+import numpy as np
+from tensorboard.plugins.npmi import summary
+
+tf.compat.v1.enable_eager_execution()
+
+# Directory into which to write tensorboard data.
+LOGDIR = "/tmp/npmi_demo"
+
+
+def setup_run(logdir, verbose, run_name):
```
nfelt, line 34: Argument `verbose` isn't actually used here?

tensorboard/plugins/npmi/README.md
Sparkier: Done.

tensorboard/plugins/npmi/npmi_demo.py
Sparkier: Changed to Python file, so this is not relevant anymore.

Sparkier: Changed to python file, so this is not relevant anymore.

tensorboard/plugins/npmi/BUILD
Sparkier: Changed to python file, so this is not relevant anymore.

tensorboard/plugins/npmi/npmi_demo.py
Sparkier: Changed to python file for that reason.

tensorboard/plugins/npmi/npmi_demo.py
Sparkier: Removed.

Sparkier: Removed that.

PR #3893, opened 2020-07-28
migrate: components_polymer3/vz_sorting

psybuzz:


tensorboard/components_polymer3/vz_sorting/BUILD
```diff
@@ -1,17 +1,16 @@
...
     name = "vz_sorting",
     srcs = [
         "sorting.ts",
-        "vz-sorting.html",
     ],
-    path = "/vz-sorting",
+    strict_checks = False,
 )
 
 tensorboard_webcomponent_library(
```
stephanwlee, line 16: remove me. I don't think internal rules will be able to handle this dependency.

tensorboard/components_polymer3/vz_sorting/test/BUILD
```diff
@@ -11,6 +11,7 @@ tf_web_test(
     name = "test",
     src = "/vz-sorting/test/tests.html",
     web_library = ":test_web_library",
+    tags = ["manual"]
```
stephanwlee, line 14: buildifier? (trailing comma is missing.)

psybuzz: Done, trailing comma and order was off

tensorboard/components_polymer3/vz_sorting/BUILD
psybuzz: Done

PR #3898, opened 2020-07-28
Polymer style for tensor widget in TypeScript

stephanwlee:
Now, the TypeScript code can import the CSS for tensor widget in Polymer
3 compatible way.

### DIFF
--- a/tensorboard/tools/inline_file_content.py
+++ b/tensorboard/tools/inline_file_content.py
@@ -1,0 +1,44 @@
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""Inline file contents in destination file.
+
+Replaces special template syntax (e.g., `%filename.ext%`) in the destination file.
+"""
+
+import os
+import sys
+
+
+def inline_content():
+    dest_file_path = sys.argv[1]
+    src_paths = sys.argv[2:]
+    with open(dest_file_path, "r") as f:
+        html = f.read()
+    for src_path in src_paths:
+        with open(src_path, "rb") as f:
+            content = f.read().decode("utf-8")
+        src_basename = os.path.basename(src_path)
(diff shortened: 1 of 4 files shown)


tensorboard/tools/inline_file_content.py
```diff
@@ -0,0 +1,44 @@
...
+"""Inline file contents in destination file.
+
+Replaces special template syntax (e.g., `%filename.ext%`) in the destination file.
+"""
+
+import os
+import sys
+
+
+def inline_content():
```
nfelt, line 24: Slightly more conventional to call this `main()`.  Also can you provide (here or in module doc) a brief example of the CLI arguments?  E.g.  `usage: inline_file_content.py <dest-file> <src-file> [<src-file>...]`

tensorboard/tools/inline_file_content.py
```diff
@@ -0,0 +1,44 @@
...
+    dest_file_path = sys.argv[1]
+    src_paths = sys.argv[2:]
+    with open(dest_file_path, "r") as f:
+        html = f.read()
```
nfelt, line 28: If this isn't generally specific to HTML (and the naming of the file and usage doesn't indicate it) better to name this variable generically as well, like `dest_contents`

tensorboard/tools/inline_file_content.py
```diff
@@ -0,0 +1,44 @@
...
+    for src_path in src_paths:
+        with open(src_path, "rb") as f:
+            content = f.read().decode("utf-8")
```
nfelt, line 31: Seems easier to just open the file as text (`r` rather than `rb`) and skip the decode?

tensorboard/tools/inline_file_content.py
```diff
@@ -0,0 +1,44 @@
...
+        src_basename = os.path.basename(src_path)
+        template_string = "%" + src_basename + "%"
+        if template_string not in html:
+            raise ValueError(
+                "Cannot find %s in dest file %s"
+                % (template_string, dest_file_path)
+            )
+        html = html.replace(template_string, content)
```
nfelt, line 39: Optional, but I'd recommend structuring this replacement algo to avoid worrying about a few edge cases:

1. Edge case where two files have same base name (i.e. same replacement string)
2. Edge case where one file or more earlier replacements produce a valid replacement string for a later file, causing nested expansion

I think a reasonable way to do that would just be to build up a dict mapping replacement keys (i.e. basenames) to contents, which lets you raise an error for case 1 above, and then use `re.sub()` to do the replacements all in one go, which addresses case 2.  To do the latter you could pass in a regex that's the union of the replacement keys (each one escaped using `re.escape()`), and pass in a replacement function that does a simple lookup from the matching text into the dict of replacements.

stephanwlee:
Please TAL again. I made the changes according to your feedback. Hope my changes are correct.

Btw, I hope this to replace other appropriate genrules in the future.

stephanwlee:
Thanks a lot for the quick reviews!

PR #3900, opened 2020-07-28
migrate: components_polymer3/analytics

psybuzz:


psybuzz:
Darn, in order to keep polymer_lib building properly, we need to keep analytics.html around for now.  I've re-included the analytics.html alongside the converted analytics.ts.  Let me know if this works for you

PR #3897, opened 2020-07-28
migrate: components_polymer3/tf_storage

stephanwlee:

PR #3903, opened 2020-07-28
migrate: experimental/plugin util

psybuzz:
Diffbase against #3897.


tensorboard/components_polymer3/experimental/plugin_util/message.ts
```diff
@@ -31,110 +31,97 @@ limitations under the License.
...
-    | object
-    | object[];
-
-  export interface Message {
-    type: string;
-    id: number;
-    payload: PayloadType;
-    error: string | null;
-    isReply: boolean;
+export type PayloadType =
```
stephanwlee, line 34: can we actually deduplicate with plugin_lib/message.ts instead? We can do that in a follow up if you'd like that.

psybuzz: Yes!  Done in this PR.

PR #3905, opened 2020-07-29
migrate: components_polymer3/tf_card_heading

stephanwlee:


tensorboard/components_polymer3/tf_card_heading/tf-card-heading.ts
```diff
@@ -0,0 +1,186 @@
...
+  @property({type: String})
+  run: string = null;
+
+  @property({type: String})
+  description: string = null;
+
+  @property({type: String})
+  color: string = null;
+
+  @property({
```
nfelt, line 138: Just for my understanding - it seems like in the Polymer 3 style this might be more idiomatically converted to also be an `@computed` property on `_computeRunBackground()` (and likewise for `_runColor` below)?  (It sounds like that supports `observer` still, but at any rate it also seems like maybe `_updateHeadingStyle()` would be more idiomatic as an `@observer` anyway?)

I'm totally fine if our preference is to stick with the minimum working conversion and not bother trying to optimize, just curious for my own information whether that's an accurate reading here?

stephanwlee: I forgot the condition in which the tool automatically concerts that but something was not deemed safe or unhandled. 

I am going for minimal human intervention at this point since the decorator syntax is a mere sugar on top of this `property` decorator. When converting yourself, try to not make mechanical changes but do look out for incorrect typing/wrong conversions.

PR #3906, opened 2020-07-29
Include HParams in "Getting started with TensorBoard.dev" colab.

bmd3k:

TensorBoard.dev has launched support for the HParams Dashboard. We want to include some basic HParams logging in the TensorBoard.dev colab.

After discussing with PM, update the "Getting started with TensorBoard.dev" so that it includes basic HParams logging in the call to Model.fit().

The HParams output is not very interesting -- we are still only doing a single run -- but we decided it was more important to keep the colab brief. We provide a link to the HParams Dashboard colab for users who want to explore further.

Uploaded the colab changes to my Drive.
Opened the colab in colab.google.com.
Ran the colab without modifications.  It worked and had HParams in the Dashboard.

image


docs/tbdev_getting_started.ipynb
```diff
@@ -12,7 +12,7 @@
     },
```
GalOshri, line 12: <p>I don't think this is necessary: we want this tutorial to be as concise as possible and shouldn't try to explain things unrelated to TensorBoard.dev too much. If we want to keep it, perhaps the more concise: "Use the MNIST dataset and create a simple TensorFlow model."?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/tensorboard/pull/3906/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

GalOshri, line 12: <p>Similar to the above comment, although having one line to say the TensorBoard callback helps log to TensorBoard and the hp.KerasCallback logs the hparams seems fine. I don't think we need to mention the histogram_freq parameter and for the HParams link can just say "learn more <a href="https://www.tensorflow.org/tensorboard/hyperparameter_tuning_with_hparams" target="_blank">here</a>". </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/tensorboard/pull/3906/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

bmd3k: Done. I removed this section.

bmd3k: I simplified it. Can you give it another look?

docs/tbdev_getting_started.ipynb
```diff
@@ -79,7 +79,8 @@
       "outputs": [],
```
manivaradarajan, line 79: <p>I think this can be a clearer. Can you rephrase as:</p><p> </p><p>TensorBoard logs are created during training by passing the TensorBoard (link) and hyperparameters callbacks (link) to Keras' Model.fit(). These logs can then be uploaded to TensorBoard.dev.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/tensorboard/pull/3906/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

bmd3k: Done. Do you think "https://www.tensorflow.org/tensorboard/get_started" is the correct "TensorBoard" link?

PR #3916, opened 2020-07-30
Add hparams to list of launched TensorBoard.dev dashboards

bmd3k:
We will soon launch the hparams dashboard in TensorBoard.dev. It should be listed along with other dashboards available in TensorBoard.dev.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ Colab.
 You may also be interested in the hosted TensorBoard solution at
 [TensorBoard.dev][]. You can use TensorBoard.dev to easily host, track, and
 share your ML experiments for free. For example, [this experiment][] shows a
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,8 +13,8 @@ Colab.
 You may also be interested in the hosted TensorBoard solution at
 [TensorBoard.dev][]. You can use TensorBoard.dev to easily host, track, and
 share your ML experiments for free. For example, [this experiment][] shows a
```
GalOshri, line 15: This link does not currently include HParams. Will this experiment be updated to include HParams, or does the experiment link need to be changed?

bmd3k: Good catch. Done. I've updated with a link to a new experiment.

PR #3922, opened 2020-07-30
sync: unblock sync by not using location API

stephanwlee:
tsetse, internal conformance, complained about use of the location API.
It seems to ignore the history counterpart (although it looks equally bad :))

### DIFF
--- a/tensorboard/components_polymer3/tf_storage/storage.ts
+++ b/tensorboard/components_polymer3/tf_storage/storage.ts
@@ -273,7 +273,7 @@ function writeComponent(component: string, useLocationReplace = false) {
     if (useLocationReplace) {
       const url = new URL(window.location.href);
       url.hash = component;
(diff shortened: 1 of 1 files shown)


tensorboard/components_polymer3/tf_storage/storage.ts
```diff
@@ -271,7 +271,9 @@ function readComponent(): string {
 function writeComponent(component: string, useLocationReplace = false) {
   if (useHash()) {
     if (useLocationReplace) {
-      window.location.replace('#' + component);
+      const url = new URL(window.location.href);
+      url.hash = component;
+      window.history.replaceState(undefined, undefined, url.toString());
```
psybuzz, line 276: MDN suggests the first 2 parameters are non-optional.  Should we pass `replaceState(null, '', url.toString())`?

https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState

PR #3930, opened 2020-07-31
sync: remove dependency on polymer legacy el mixin and fixed typing on dom-repeat

stephanwlee:


stephanwlee:
Ah, I totally forgot to use another branch for that. Thanks for reviewing both commits. I will merge this as rebase since they are logically separate changes.

PR #3937, opened 2020-08-01
Allow tensorboard_html_binary to serve JS file in the devserver

stephanwlee:
With this change, when you `bazel run` a target, loading the HTML file in the browser will work since
extracted JavaScript would be served by the server.

Previously above flow failed since the server never served the JavaScript file.

### DIFF
--- a/tensorboard/defs/vulcanize.bzl
+++ b/tensorboard/defs/vulcanize.bzl
@@ -35,7 +35,7 @@ def _tensorboard_html_binary(ctx):
     manifests = depset(transitive=[manifests, dep.webfiles.manifests])
     webpaths = depset(transitive=[webpaths, dep.webfiles.webpaths])
     files = depset(transitive=[files, dep.data_runfiles.files])
-  webpaths = depset([ctx.attr.output_path, ctx.attr.js_path], transitive=[webpaths])
+  webpaths = depset([ctx.attr.output_path], transitive=[webpaths])
   closure_js_library=collect_js(
       unfurl(ctx.attr.deps, provider="closure_js_library"))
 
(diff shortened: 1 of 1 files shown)


tensorboard/defs/vulcanize.bzl
```diff
@@ -73,7 +73,10 @@ def _tensorboard_html_binary(ctx):
   # webfiles manifest
   manifest_srcs = [struct(path=ctx.outputs.html.path,
                           longpath=long_path(ctx, ctx.outputs.html),
-                          webpath=ctx.attr.output_path)]
+                          webpath=ctx.attr.output_path),
+                   struct(path=ctx.outputs.js.path,
+                          longpath=long_path(ctx, ctx.outputs.js),
+                          webpath=ctx.attr.js_path)]
```
wchargin, line 79: The `js_path` attribute is optional, and the following targets don’t
provide it:

```
//tensorboard/components/vz_line_chart2/microbenchmark:binary
//tensorboard/components:polymer_lib_binary
//tensorboard/components_polymer3/vz_line_chart2/microbenchmark:binary
//tensorboard/components_polymer3:polymer_lib_binary
//tensorboard/plugins/graph/polymer3/tf_graph_app:binary
//tensorboard/plugins/graph/tf_graph_app:binary
//tensorboard/plugins/projector/vz_projector:devserver
```

So it looks like this adds a struct with `webpath = ""`; curious how/why
that works? (And it does: the `vz_projector:devserver` still seems to
work fine for me.)

stephanwlee: You are right. I, too, am surprised that empty webpath is treated correctly.

I added some checks before adding the manifest.

PR #3940, opened 2020-08-02
migrate: mesh plugin

stephanwlee:
Do note that for ease of migration, we have copied two modules from the
tensorflow/graphics repository[1]. Do note that we made some manual
changes to better type (to best of our knowledge based on the closure
type annotations). Because version of THREE has been upgraded and
changed, we may be using deprecated & removed API. We will be able to
better tell whether we are doing the right thing only after when we do
the functional testing.

[1]: https://github.com/tensorflow/graphics/blob/master/tensorflow_graphics/tensorboard/mesh_visualizer/tf_mesh_dashboard/

WANT_LGTM=any


tensorboard/plugins/mesh/polymer3/tf_mesh_dashboard/tf-mesh-dashboard.ts
```diff
@@ -209,123 +205,123 @@ <h3>No point cloud data was found.</h3>
...
+  }
+
+  @computed('_runToTagInfo', '_selectedRuns', '_tagFilter')
+  get _categories(): unknown[] {
+    var runToTagInfo = this._runToTagInfo;
+    var selectedRuns = this._selectedRuns;
+    var tagFilter = this._tagFilter;
+    const runToTag = _.mapValues(runToTagInfo, (x) => Object.keys(x));
+    const baseCategories = categorizeRunTagCombinations(
+      runToTag as any,
```
nfelt, line 305: FWIW, this could be `runToTag as tf_categorization_utils.RunToTag` if we want to avoid the extra `as any`.

tensorboard/plugins/mesh/polymer3/tf_mesh_dashboard/tf-mesh-dashboard.ts
```diff
@@ -209,123 +205,123 @@ <h3>No point cloud data was found.</h3>
...
+    return index <= 2;
+  }
+
+  reload() {
+    this._fetchTags().then(this._reloadMeshes.bind(this));
+  }
+
+  _handleWindowResize() {
+    this._getAllChildren().forEach((g) => {
+      (g as any).redraw();
```
nfelt, line 275: Can we add `redraw()` to the `TfMeshLoader` public interface, so that this doesn't require `as any`?  It looks like `redraw()` is meant to be public anyway.

wchargin:
(started my review before @nfelt’s was posted and submitted without
refreshing; my approval isn’t meant to dismiss his comments :-) )

tensorboard/plugins/mesh/polymer3/tf_mesh_dashboard/tf-mesh-dashboard.ts
stephanwlee: Done.

tensorboard/plugins/mesh/polymer3/tf_mesh_dashboard/tf-mesh-dashboard.ts
stephanwlee: Done.

tensorboard/plugins/mesh/polymer3/tf_mesh_dashboard/mesh-loader.ts
```diff
@@ -0,0 +1,352 @@
...
+import '../../../../components_polymer3/polymer/irons_and_papers';
+import '../../../../components_polymer3/tf_card_heading/tf-card-heading';
+
+import {
+  ArrayBufferDataProvider,
+  ErrorCodes,
+} from './array-buffer-data-provider';
+import {MeshViewer} from './mesh-viewer';
+
+export abstract class TfMeshLoader extends HTMLElement {
```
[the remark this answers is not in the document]
stephanwlee: Changed it to an interface. Thanks.

stephanwlee:
@nfelt Changed to have three commits as you detailed. Given LGTM from William, I do not require another pass from you but I will wait a bit for your feedbacks.

PR #3947, opened 2020-08-03
add explicit global HTML style and add roboto dependency

stephanwlee:
Do note that after the library upgrade, we need to explicitly tell
Polymer not to fetch Roboto from CDN.

Manually ran the application to check for no regression.

Note that the roboto genrule file change does not impact Polymer 2.

### DIFF
--- a/third_party/fonts.bzl
+++ b/third_party/fonts.bzl
@@ -908,6 +908,7 @@ def tensorboard_fonts_workspace():
           "        '  unicode-range: U+0102-0103, U+1EA0-1EF9, U+20AB;',",
           "        '}',",
           "        '</style>',",
+          "        '<script>window.polymerSkipLoadingFontRoboto = true;</script>',",
           '        "EOF",',
           '    ]),',
           ')',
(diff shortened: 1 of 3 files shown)


third_party/fonts.bzl
```diff
@@ -908,6 +908,7 @@ def tensorboard_fonts_workspace():
           "        '  unicode-range: U+0102-0103, U+1EA0-1EF9, U+20AB;',",
           "        '}',",
           "        '</style>',",
+          "        '<script>window.polymerSkipLoadingFontRoboto = true;</script>',",
```
psybuzz, line 911: I have a couple questions to confirm my understanding:
- Before, did Polymer 2 come bundled with Roboto-loading code, which loaded the font even though index.uninlined.html didn't import it?
- Does this apply nicely in Polymer 2 world?  (e.g. this PR also makes sense even if we were not migrating)

stephanwlee: > Does this apply nicely in Polymer 2 world? (e.g. this PR also makes sense even if we were not migrating)

It noops in Polymer 2 world. "Manually ran the application to check for no regression.", by this, I meant existing webapp (`bazel run tensorboard`).

> Before, did Polymer 2 come bundled with Roboto-loading code, which loaded the font even though index.uninlined.html didn't import it?

Not sure. I suspect Polymer 2 did not come with the loading code.

stephanwlee: Confirmed that there is no match for `polymerSkipLoadingFontRoboto` in existing polymer_lib which makes me believe that it is a new invention in Polymer 3.

PR #3948, opened 2020-08-03
fix: call *Initializer with reference to

stephanwlee:
Our Polymer based storage utility required Polymer class instance for
the initializer (so we can set the component prop value when hash
changes). This broke in the class syntax so we reinstated it by
explicitly passing `this`.

Co-authored-by: William Chargin <wchargin@gmail.com>


nfelt:
Thanks for fixing up the runs-selector code for me!

PR #3953, opened 2020-08-04
sync: fixed projector for sync

stephanwlee:
Few issues:
1. depend on `numeric`. `numericjs` is not the popular package.
2. remove export if using `LegacyElementMixin`. Fixed component class
    imports accordingly
3. depend on internal legacy element mixin
4. remove innerHTML and change it to textContent since we are all
    setting textContent
5. replaced custom word break with tf-wbr-string
6. side-stepped issue with tsetse regarding hyperlinks

Made sure nothing regressed by manually running the standalone projector.


tensorboard/plugins/projector/polymer3/vz_projector/vz-projector-bookmark-panel.ts
```diff
@@ -40,7 +40,7 @@ export class BookmarkPanel extends LegacyElementMixin(PolymerElement) {
   @property({type: Number})
   selectedState: number;
 
-  private projector: Projector;
+  private projector: any;
```
nfelt, line 43: Might be nice to keep around the types as comments (`// actual type: Projector`) but up to you since it's extra effort already.

PR #3954, opened 2020-08-04
chore: add npm deps for hparams

psybuzz:
yarn add
- @polymer/iron-pages
- @polymer/paper-radio-button
- @polymer/paper-radio-group
- @vaadin/vaadin-grid
- @vaadin/vaadin-split-layout

PR #3955, opened 2020-08-04
migrate: plugins/hparams

psybuzz:
Diffbase:
- https://github.com/tensorflow/tensorboard/pull/3954: npm deps
- TBD: once scalar dashboard is migrated, hparams will rebase to depend on it

Tested with:
`TB_POLYMER=1 bazel run tensorboard` on `/index_polymer3.html`

Notable changes include:
- packages depending on the old api_d_ts now depend on api_pre_polymer3_d_ts target.
  - api.d.ts --> api-pre-polymer3.d.ts
  - api.d.ts (new file)
  - checked that this command succeeds:
    bazel build tensorboard/plugins/hparams/tf_hparams_parallel_coords_plot
(note the lack of polymer3/ in file path)
- dropped the unused dependency on tf-tensorboard/plugin-dialog
- increased CSS height of instances of `tf-hparams-session-group-values`.
  The new vaadin-grid has different default height, so it will be clipped
  otherwise.
- split-layout's `vertical` HTML attribute is now `orientation='vertical'`

image

image

image


wchargin:
By the way, I didn’t run a test sync; please ensure that we’re good for
that before syncing.

PR #3957, opened 2020-08-04
bug: fix broken contentActive in tf-dom-repeat

stephanwlee:
tf-dom-repeat and tf-category-paginated-view were weird because the sub
class was implementing `_contentActive` protected Polymer property via
getter (because it was a computed property). It looks like it is
impossible for subclass to override a property using a getter.

For instance, in pure JavaScript,
```js
(function() {
  class Foo {
    foo = 'foo';
  }

  class Bar extends Foo {
    foo =  'bar';
  }

  class Baz extends Foo {
    get foo() {
      return 'baz';
    }
  }

  console.log(new Bar().foo); // 'bar'
  console.log(new Baz().foo); // 'foo'
})();
```

This change no longer implements `_contentActive` setting via getter.


nfelt:
FWIW I'm kind of horrified that JS class getters don't override properties, but I'm even more horrified that after quite a bit of searching I can't find this gotcha thoroughly documented anywhere D:

PR #3958, opened 2020-08-04
polymer3: use hourglass import for paper & irons

stephanwlee:
Polymer's iron and paper components are hard to depend in a way that is
compatible both in OSS and google3. With this change, by creating an
hourglass dependency, we can resolve the discrepancy easily by making
the changes in only the new file.


tensorboard/components_polymer3/tf_dashboard_common/tf-multi-checkbox.ts
```diff
@@ -289,7 +286,7 @@ class TfMultiCheckbox extends PolymerElement {
   synchronizeColors(e) {
     this._setIsolatorIcon();
     const checkboxes = this.root.querySelectorAll('paper-checkbox');
-    checkboxes.forEach((p: PaperCheckboxElement) => {
+    checkboxes.forEach((p: any) => {
       const color = this.coloring.getColor((p as any).name);
```
nfelt, line 290: We no longer need `(p as any)` here now, same below

PR #2392, opened 2019-06-27
core: upgrade Polymer to v2.7

stephanwlee:
* Motivation for features / changes
Polymer 1 is quite old and we should migrate off to 2 and possibly to 3.
This is partially for its internal deprecation but also to revamp build tooling:
we are currently tied to a process called Vulcanization because we have to
build HTML files.

* Detailed steps to verify changes work correctly (as executed by you)
Ran/clicked/exercised all plugins.

* Warning
Although we tried to thorough in our testing, there may be few visual/functional
regressions. We intend to fix forward those bugs as they are identified. 

* Merge plan: squash

* Commit Message
```
core: update the dependencies with Polymer v2 (#2231)                           
core: bump up JSCompiler langout to ES2015 (#2231)                              
core: use v2 Polymer API (#2231)                                                
core: fixed Plottable CSS issue (#2231)                                         
core: <style> should appear inside template (#2231)                             
cleanup: Replaced paper-menu with paper-listbox (#2231)                         
core: fixed the limit setter (#2231)                                                                                                                                                                                                          
graph: fixed most visual regressions of tf-graph (#2251)                        
graph: fixed tf-graph-icons and moved it for reusability (#2251)                
graph: fix graph node toggle bug (#2251)                                        
graph: fix graph fitting in tf-graph-controls (#2251)                           
graph: Fixed input trace (#2251)                                                
graph: fixed broken context menu (#2251)                                        
graph: fixed info panel buttons (#2251)                                         
graph: fixed remaining global selector (#2251)                                  
core: fixed the Plottable interaction (#2249)                                   
profile: fixed mouse interaction (#2286)                                        
profile: fixed CSS issues (#2286)                                               
profile: fixed hosts not updating issue (#2286)                                 
profile: fix misc issues (#2286)                                                
profile: overview and pipline view fixes (#2286)                                
core: fix the line chart tooltip (#2277)                                        
core: fix the data-loader behavior (#2250)                                      
core: use proper CSS selector syntax for :host (#2250)                          
core: fix dashboard-style slot usage to property "slot" (#2250)                 
cpv: fixed custom templatizer in paginated-view2 (#2273)                        
cpv: Restructure files (#2273)                                                  
cpv: Implement category-paginated-view (#2273)                                  
cpv: Migrate to tf-category-paginated-view (#2273)                              
cpv: Fix test for tf-category-paginated-view (#2273)                            
cpv: Remove unused paginated-view and renamed test (#2273)                      
[...]

PR #3959, opened 2020-08-04
migrate: plugins/histogram/polymer3

wchargin:
This patch covers `vz_histogram_timeseries` and `tf_histogram_dashboard`.

Test Plan:
With `TB_POLYMER=1 bazel run //tensorboard`,

  - histograms render;
  - the histogram mode (overlay/offset) selector works;
  - the offset time axis works;
  - toggling runs in the run selector works;
  - entering a regex filter in the run selector works;
  - entering a regex tag filter works;
  - adjusting the pagination limit works.


tensorboard/plugins/histogram/polymer3/tf_histogram_dashboard/tf-histogram-core.ts
```diff
@@ -13,8 +11,4 @@
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
--->
-
-<link rel="import" href="../tf-imports/d3.html" />
-
-<script src="histogramCore.js"></script>
+==============================================================================*/
```
psybuzz, line 14: Can this unused file can be deleted?

tensorboard/plugins/histogram/polymer3/tf_histogram_dashboard/tf-histogram-dashboard.html
```diff
@@ -1,252 +0,0 @@
...
-        _dataNotFound: Boolean,
-        _tagFilter: String,
-        _restamp: {
-          type: Boolean,
-          value: false,
-        },
-
-        // Categories must only be computed after _dataNotFound is found to be
-        // true and then polymer DOM templating responds to that finding. We
-        // thus use this property to guard when categories are computed.
```
psybuzz, line 177: Preserve this comment?

tensorboard/plugins/histogram/polymer3/tf_histogram_dashboard/tf-histogram-dashboard.html
```diff
@@ -1,252 +0,0 @@
...
-        },
-
-        _requestManager: {
-          type: Object,
-          value: () => new tf_backend.RequestManager(),
-        },
-      },
-
-      listeners: {
-        'content-visibility-changed': '_redrawCategoryPane',
```
psybuzz, line 192: The auto-migration script doesn't cover `listeners`, so we should add something like

`this.addEventListener('content-visibility-changed', this._redrawCategoryPane.bind(this))`
in either the `ready()` or `connectedCallback()`

psybuzz:
Tested it out with the commits mentioned, and it looks great!

tensorboard/plugins/histogram/polymer3/tf_histogram_dashboard/tf-histogram-dashboard.html
wchargin: Done.

tensorboard/plugins/histogram/polymer3/tf_histogram_dashboard/tf-histogram-dashboard.html
wchargin: We looked into this a bit. This isn’t a standard DOM event; rather, it’s
a custom Polymer event that was originally added in #1465. But there
isn’t any code to actually fire the event; `git log -S` shows that it
was removed in #2273 via #2392. Thus, while the absence of this listener
may be a bug, it’s not new, so I’ll keep this migration patch as is to
make things simple. Thanks for the keen eye, though!

wchargin:
Test sync; <http://cl/325558004>.

PR #3960, opened 2020-08-04
migrate: fix irons_and_papers import in tf-multi-checkbox.ts

nfelt:
Followup to #3958, the import has a few two many `..`'s.


wchargin:
(thanks!)

PR #3963, opened 2020-08-04
migrate: enable projector polymer3 with TB_POLYMER3=1

nfelt:
This adds logic similar to #3961 but for the projector plugin (which requires a different approach since it's a dynamic plugin).  Once again we bundle both the Polymer 2 and Polymer 3 assets but under different paths, and then use the `TB_POLYMER3=1` env var condition to select which ones we serve under the canonical routes.

Tested by running locally and verifying that the projector loads normally without any special configuration, and with `TB_POLYMER3=1` it seems to load the Polymer 3 version (but it's still broken because of some three.js error, see below).  At least, the routes resolve and `projector_binary.js` includes what appears to be the polymer3 version.

I also tested building a pip package and installing it to check that the setup.py changes work.

The remaining error on the console is:

```
three.module.js:4102 Uncaught ReferenceError: p is not defined
    at new DownsampleCalculator (three.module.js:4102)
    at createModule (three.module.js:4096)
    at Object.3../globals (three.module.js:4096)
    at s (three.module.js:4091)
    at three.module.js:4091
    at Object.1../lib/globals (three.module.js:4092)
    at s (three.module.js:4091)
    at e (three.module.js:4091)
    at three.module.js:4091
    at three.module.js:4091
```

### DIFF
--- a/tensorboard/components_polymer3/BUILD
+++ b/tensorboard/components_polymer3/BUILD
@@ -42,6 +42,7 @@ tf_ts_library(
     name = "polymer3_ts_lib",
     srcs = ["polymer3_lib.ts"],
     deps = [
+        "//tensorboard/components_polymer3/experimental/plugin_util:plugin_host",
         "//tensorboard/components_polymer3/tf_backend",
         "//tensorboard/components_polymer3/tf_globals",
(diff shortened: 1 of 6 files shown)


nfelt:
Closing and re-opening for Travis.

PR #3965, opened 2020-08-05
sync: remove usage of THREE during module resolution

stephanwlee:
Due to internal quirks, threejs symbols are not available during the
module load time. To work around this issue, we will use it only during
component creation time.


stephanwlee:
They expect a symbol to exist on `window.THREE` then re-export it for some reason :( and window.THREE has to be loaded before anything which is hard to control with JSComp

PR #3967, opened 2020-08-05
Expose experiment metadata as part of the Environment object in the ngrx store.

bmd3k:

  Certain variants of TensorBoard (TensorBoard.dev, for example) retrieve experiment metadata when loading the "Environment". We want to properly expose that as part of the `Environment` that we store in the ngrx store.

  Add a few new properties to the `Environment` object. The data is already retrieved and stored by TensorBoard.dev - it just needs to be exposed officially in the internal API.

  Mainly make sure this change doesn't break TensorBoard.
  * Build and run TensorBoard from source.
  * Load TensorBoard in a browser.
  * Pay particular attention the TbdevUploadButton, which is the only current Angular component to use `Environment` object.  Make sure it is still functioning properly.

### DIFF
--- a/tensorboard/webapp/types/api.ts
+++ b/tensorboard/webapp/types/api.ts
@@ -72,6 +72,12 @@ export interface Environment {
   data_location: string;
   /** @export */
   window_title: string;
+  /** @export */
+  experiment_name?: string;
+  /** @export */
+  experiment_description?: string;
+  /** @export */
+  creation_time?: number;
 }
 
 export type GetRunsResponse = string[];


tensorboard/webapp/types/api.ts
```diff
@@ -72,6 +72,12 @@ export interface Environment {
   data_location: string;
   /** @export */
   window_title: string;
+  /** @export */
+  experiment_name?: string;
+  /** @export */
+  experiment_description?: string;
+  /** @export */
+  creation_time?: number;
```
bileschi, line 80: Does it make sense to also include `update_time` at this point?

bileschi: Also, is it common to include a units here?  I see the time is a number but is it seconds since epoch?  ms?  Is there a standard practice here?

bmd3k: > Does it make sense to also include update_time at this point?

TensorBoard.dev does not yet retrieve update_time so I don't think I should add it.

> Also, is it common to include a units here? I see the time is a number but is it seconds since epoch? ms? Is there a standard practice here?

Unfortunately this is out of my control. This declaration just mimics existing behavior. I'm not sure where these fields are originally declared (if anywhere) but this is where they are set:

https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/plugins/core/core_plugin.py;l=149;drc=fbdb2b7b674538ff86c4704dcd3f1f0ddc25737a

PR #3972, opened 2020-08-05
migrate: plugins/scalar/polymer3

nfelt:
Migrates the scalars dashboard to Polymer 3.

The first 4 commits in the PR are autogenerated or otherwise formulaic; they are broken out in order to have nicer diffs.

Tested via `TB_POLYMER3=1 bazel run tensorboard -- --bind_all --logdir <demodir>` and trying every UI affordance that I could think of (i.e. all the UI controls, downloading, chart pan/zoom, etc.). 

_Update: this is resolved by #3992._ ~There is one known issue, which is that `tf-line-chart-data-loader` doesn't reload data when `data-to-load` changes.  The most obvious way to trigger this is to try to add a new time series to the chart (i.e. if you start with runs deselected at load, and then select new runs, they don't appear on the chart).  This seems to be caused by the fact that the [`Node.isConnected`](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected) property on `tf-line-chart-data-loader` is returning `undefined` when it should be `true`, which is a problem since `DataLoaderBehavior` gates data loading in response to `dataToLoad` changes using `if (this.isConnected)`.  I assume this is probably a result of the complicated inheritance hierarchy for `TfLineChartDataLoader` and haven't finished debugging it, but it seems reasonable to try to get this merged anyway to avoid backsliding.~


tensorboard/plugins/scalar/polymer3/tf_scalar_dashboard/tf-scalar-card.ts
```diff
@@ -0,0 +1,354 @@
...
+  xType: string;
+
+  @property({type: Boolean})
+  active: boolean;
+
+  @property({type: Boolean})
+  ignoreYOutliers: boolean;
+
+  @property({type: Object})
+  requestManager: object;
```
wchargin, line 197: Can we type this as `RequestManager`? I’ve generally been trying to
include the appropriate types where possible; okay with me to drop it
for something like `tagMetadata` if you can’t be bothered, but since
`RequestManager` is a pure-TypeScript class anyway it seems worth imho.

tensorboard/plugins/scalar/polymer3/tf_scalar_dashboard/tf-scalar-card.ts
```diff
@@ -0,0 +1,354 @@
...
+  @property({type: Object})
+  colorScale: object = null;
+
+  @property({type: String})
+  tooltipSortingMethod: string;
+
+  // This function is called when data is received from the backend.
+  @property({type: Object})
+  _loadDataCallback: object = function() {
+    return (scalarChart, datum, data) => {
```
wchargin, line 222: This can just be:

```typescript
_loadDataCallback = (scalarChart, datum, data) => {
  …
};
```

tensorboard/plugins/scalar/polymer3/tf_scalar_dashboard/tf-scalar-card.ts
```diff
@@ -0,0 +1,354 @@
...
+    const ind = columns.findIndex((c) => c.title == 'Name');
+    columns.splice(ind, 1, {
+      title: 'Name',
+      evaluate: (d) => {
+        const datum = d.dataset.metadata().meta;
+        return this._getSeriesDisplayNameFromDatum(datum);
+      },
+    });
+    return columns;
+  }.call(this);
```
wchargin, line 281: Since arrow functions implicitly bind `this`, we’d generally prefer to
write this as:

```typescript
_tooltipColumns = (() => {
  // ...
  return columns;
})();
```

The Google JS style guide says:

> Prefer arrow functions over the `function` keyword, particularly for
> nested functions

and the internal-only TypeScript style guide is stronger with “Always
use arrow functions instead of pre-ES6 function expressions”.

tensorboard/plugins/scalar/polymer3/tf_scalar_dashboard/tf-scalar-card.ts
nfelt: Done, just overlooked it.

tensorboard/plugins/scalar/polymer3/tf_scalar_dashboard/tf-scalar-card.ts
nfelt: Done.

tensorboard/plugins/scalar/polymer3/tf_scalar_dashboard/tf-scalar-card.ts
nfelt: Done.

nfelt:
FYI, added a few more commits to this PR, including fixing the data loading issue (via #3992), and did functional testing externally and internally.  Let me know if you want to re-review, otherwise I'll merge later today

PR #3978, opened 2020-08-05
license: extract bh_tsne license out and bundle it

stephanwlee:
We currently have two licenses for our pip package: (1) one on LICENSE
in pip package and (2) inlined in JavaScript bundle. For ease of our
conformance to the rules, we decided to inline Karpathy's TSNE license
to (1).


stephanwlee:
Talked offline: we now better understand the process that we need to follow (which we may need to do anyways).
Move all forked code into `third_party/[foo]/` and create a METADATA file in the folder. That way, we can keep the license header in the source code (in this case, TypeScript file).

We will do a cleanup afterwards.

PR #3982, opened 2020-08-06
Adding back in a missing dependency

yfang1:
scalars_plugin.py is referencing metadata.py, but the dependency is not set up in the BUILD file.

### DIFF
--- a/tensorboard/plugins/scalar/BUILD
+++ b/tensorboard/plugins/scalar/BUILD
@@ -15,6 +15,7 @@ py_library(
     srcs = ["scalars_plugin.py"],
     srcs_version = "PY2AND3",
     deps = [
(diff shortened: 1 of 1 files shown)


yfang1:
@googlebot I signed it!

tensorboard/plugins/scalar/BUILD
```diff
@@ -15,6 +15,7 @@ py_library(
     srcs = ["scalars_plugin.py"],
     srcs_version = "PY2AND3",
     deps = [
+        ":metadta",
```
nfelt, line 18: typo: metadata

yfang1: That's embarrassing. I blame chrome remote desktop.

PR #3984, opened 2020-08-06
text_v2: add skeleton to the application

stephanwlee:
This creates a skeleton for text_v2 plugins upon which we will build features.

Because we only render hard coded string, it is as boring as this:

image

### DIFF
--- a/tensorboard/webapp/app_module.ts
+++ b/tensorboard/webapp/app_module.ts
@@ -28,6 +28,9 @@ import {PluginsModule} from './plugins/plugins_module';
 import {ROOT_REDUCERS, loggerMetaReducerFactory} from './reducer_config';
 import {ReloaderModule} from './reloader/reloader_module';
 import {SettingsModule} from './settings/settings_module';
+import {TextV2Module} from './plugins/text_v2/text_v2_module';
+
+const NG_PLUGIN_MODULES = [TextV2Module];
 
 @NgModule({
   declarations: [AppContainer],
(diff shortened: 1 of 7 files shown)


stephanwlee:
cc @pjfossee

tensorboard/webapp/app_module.ts
```diff
@@ -28,6 +28,9 @@ import {PluginsModule} from './plugins/plugins_module';
 import {ROOT_REDUCERS, loggerMetaReducerFactory} from './reducer_config';
 import {ReloaderModule} from './reloader/reloader_module';
 import {SettingsModule} from './settings/settings_module';
+import {TextV2Module} from './plugins/text_v2/text_v2_module';
+
+const NG_PLUGIN_MODULES = [TextV2Module];
```
psybuzz, line 33: Instead of here, shouldn't `TextV2Module` be imported in `plugins_module.ts`, next to the other DebuggerModule, NpmiModule?  Or is TextV2 different than other Angular plugins somehow?

stephanwlee: It looks like different applications will have different set of plugins supported so it should be closer to the application declaration, not plugins_module. Plugins_module provide plugin renderer which has not much to do with plugins an app supports.

psybuzz: Moving the place of configuration sounds reasonable to me.  If this is the place though, can we move Npmi and Debugger module imports here as well?

stephanwlee: I am inlined to do that in a quick follow up since refactoring plugins_module.ts is outside of scope of this PR.

stephanwlee: Created the follow up :) #3988.

psybuzz: lgtm % the comment added in the followup

PR #3987, opened 2020-08-07
text v2: add data source and reducers

stephanwlee:
This change introduces data_source (for interacting with backend) and
reducers. Please do note that, to allow for development, we are hard
coding the values in the reducers without any implementations of it.

In follow ups, we will add effects and actions to integrate with the reducers.

Diffbase #3984.


stephanwlee:
cc @pjfossee

tensorboard/webapp/plugins/text_v2/store/text_types.ts
```diff
@@ -0,0 +1,29 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {StepDatum} from '../data_source';
+
+export const TEXT_FEATURE_KEY = 'text';
+
+type RunId = string;
+type Tag = string;
```
bmd3k, line 20: Is `Tag` just the Id of the Tag? If so, should it be `TagId`?

tensorboard/webapp/plugins/text_v2/data_source/text_v2_types.ts
```diff
@@ -0,0 +1,42 @@
...
+limitations under the License.
+==============================================================================*/
+import {Injectable} from '@angular/core';
+import {Observable} from 'rxjs';
+
+export interface BackendRunToTagsMap {
+  [runName: string]: string[];
+}
+
+export interface BackendStepDatum {
```
bmd3k, line 22: It would be less confusing if we don't mix types for the TextDataSource interface (BackendRunToTagsMap, StepDatum) with types that are internal to the implementation (BackendStepDatum). Consider move this particular interface into the implementation class, where you don't have to `export` it, or to its own file?

tensorboard/webapp/plugins/text_v2/data_source/text_v2_types.ts
```diff
@@ -0,0 +1,42 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {Injectable} from '@angular/core';
+import {Observable} from 'rxjs';
+
+export interface BackendRunToTagsMap {
```
bmd3k, line 18: If you're exposing this as part of the interface of TextDataSource then perhaps remove 'Backend' from the name. That you happen to return the raw Backend response is an implementation detail worth hiding from callers.

tensorboard/webapp/plugins/text_v2/data_source/text_v2_types.ts
```diff
@@ -0,0 +1,42 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
bmd3k, line 1: I don't know if `types` is the correct name for this file. It's actually more your service interface - maybe this should just be text_data_source.ts. But I don't feel strongly about it.

tensorboard/webapp/plugins/text_v2/data_source/text_v2_data_source_test.ts
```diff
@@ -0,0 +1,81 @@
...
+        imports: [TBHttpClientTestingModule],
+        providers: [TextV2DataSource],
+      }).compileComponents();
+
+      httpMock = TestBed.inject(HttpTestingController);
+      dataSource = TestBed.inject(TextV2DataSource);
+    });
+
+    describe('fetchRunToTag', () => {
+      it('fetches from correct endpoint', () => {
```
bmd3k, line 40: Do you want to test that the data from the backend is returned to the spy?

tensorboard/webapp/plugins/text_v2/data_source/text_v2_types.ts
stephanwlee: Refactored to make backend types contained in the backend specific data provider but instead have only frontend types exported. Thanks for pointing that out.

tensorboard/webapp/plugins/text_v2/data_source/text_v2_types.ts
stephanwlee: Done, thanks.

PR #3988, opened 2020-08-07
refactor: tease out plugin enumeration from plugins_module

stephanwlee:
plugins_module is responsible for providing plugins registry and view
for rendering a plugin. Now, this module should not have enumeration of
all possible plugins in TensorBoard because whether an application wants
to embed a plugin or not is an application logic. For instance, we do
not intend to show debugger_v2 in TB.dev for now.

Diffbase #3984.

### DIFF
--- a/tensorboard/webapp/BUILD
+++ b/tensorboard/webapp/BUILD
@@ -7,6 +7,19 @@ package(default_visibility = ["//tensorboard:internal"])
 
 licenses(["notice"])  # Apache 2.0
 
+ng_module(
+    name = "oss_plugins_module",
+    srcs = [
+        "oss_plugins_module.ts",
+    ],
+    deps = [
+        "//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin:debugger_v2",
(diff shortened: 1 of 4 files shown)


stephanwlee:
cc @bmd3k

stephanwlee:
@bmd3k factored out reducer_config to its own target.

tensorboard/webapp/BUILD
```diff
@@ -31,6 +56,7 @@ ng_module(
         "//tensorboard/webapp/feature_flag",
         "//tensorboard/webapp/header",
         "//tensorboard/webapp/plugins",
+        "//tensorboard/webapp/plugins/text_v2",
```
psybuzz, line 59: please remove this `text_v2` dep from `:app`

stephanwlee: I'm sorry, I don't understand why that is correct. Can you please elaborate?

stephanwlee: Done

PR #3992, opened 2020-08-07
migrate: replace Node.isConnected with manually wired property

nfelt:
This fixes an issue in `DataLoaderBehavior` that was noticed with `tf-line-chart-data-loader` where it doesn't reload data when `data-to-load` changes (I haven't checked if it affects other elements that mix in `DataLoaderBehavior`).  The most obvious way to trigger this is to try to add a new time series to the chart (i.e. if you start with runs deselected at load, and then select new runs, they don't appear on the chart).  This seems to be caused by the fact that the [`Node.isConnected`](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected) property on `tf-line-chart-data-loader` is returning `undefined` when it should be `true`, which is a problem since `DataLoaderBehavior` gates data loading in response to `dataToLoad` changes using `if (this.isConnected)`.

My best guess is that this has something to do with the complicated inheritance hierarchy for `TfLineChartDataLoader` but it's a bit unclear.  @stephanwlee suggested just replacing `Node.isConnected` with a property wired up to the callbacks, so I've done that instead here.

### DIFF
--- a/tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
+++ b/tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
@@ -96,10 +96,12 @@ export function DataLoaderBehavior<Item, Data>(
     _isConnected = false;
 
     connectedCallback() {
+      super.connectedCallback();
       this._isConnected = true;
     }
 
     disconnectedCallback() {
+      super.disconnectedCallback();
       this._isConnected = false;
     }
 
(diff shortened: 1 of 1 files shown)


tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -91,12 +91,25 @@ export function DataLoaderBehavior<Item, Data>(
 
     dataLoading = false;
 
+    // The standard Node.isConnected doesn't seem to be set reliably, so we
+    // wire up our own property manually.
+    _isConnected = false;
+
+    connectedCallback() {
+      this._isConnected = true;
```
stephanwlee, line 99: Did you manually test this? You may need to call `super`.

nfelt: Thanks for the catch - I did test it manually and it definitely worked, but who knows...

PR #3993, opened 2020-08-07
Update Getting Started notebook to use `--one_shot` uploader.

bileschi:
--one_shot is the preferred workflow in this case, where the training is already complete before the upload begins.
--one_shot upload was introduced in tensorboard 2.3.0

Tested: 
*  ran through the notebook on Jupyter on my laptop using python 3.7.3 and tensorflow 2.3.0
*  ran through the notebook on colab by uploading the notebook.


bileschi:
sorry testing.

PR #3994, opened 2020-08-07
dep: add paper-radio-group dependency

stephanwlee:
Added the paper-radio-group to irons_and_papers.ts.


stephanwlee:
Confirmed that the change is safe with a sample sync.

PR #3995, opened 2020-08-07
custom_scalars: remove Polymer 3 `tf_web_library`

wchargin:
Summary:
Because of this `tf_web_library`, the `javascript` branch has the
compiled `*.js` version of a source `*.ts` file, which makes the custom
scalars migration diff apply less cleanly to that branch. This patch,
once merged and integrated into the `javascript` branch, should restore
the original `*.ts` file.

Test Plan:
Fingers crossed.

wchargin-branch: custom-scalars-remove-tfwl

### DIFF
--- a/tensorboard/plugins/custom_scalar/polymer3/tf_custom_scalar_dashboard/BUILD
+++ b/tensorboard/plugins/custom_scalar/polymer3/tf_custom_scalar_dashboard/BUILD
@@ -4,3 +4,44 @@ licenses(["notice"])  # Apache 2.0
 
 # `tf_web_library`/`tf_ts_library` here removed to facilitate internal
 # sync-testing scripts.
+
+dict(
+    name = "tf_custom_scalar_dashboard",
+    srcs = [
+        "tf-custom-scalar-card-style.html",
+        "tf-custom-scalar-dashboard.html",
+        "tf-custom-scalar-helpers.html",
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/custom_scalar/polymer3/tf_custom_scalar_dashboard/BUILD
```diff
@@ -1,46 +1,6 @@
...
-        "@org_polymer_paper_dropdown_menu",
-        "@org_polymer_paper_icon_button",
-        "@org_polymer_paper_input",
-        "@org_polymer_paper_item",
-        "@org_polymer_paper_listbox",
-        "@org_polymer_paper_slider",
-        "@org_polymer_paper_styles",
-    ],
-)
+# `tf_web_library`/`tf_ts_library` here removed to facilitate internal
```
nfelt, line 5: If you wanted a slightly nicer diff for when this is converted to `tf_ts_library`, I think you could disable this while keeping the contents of the rule by just replacing `tf_web_library` with `dict`.  Gross but works in my local testing :)

wchargin: Heheheh. Okay, sure. :-)

wchargin:
it works :-)

PR #3996, opened 2020-08-07
polymer3: add indirection for `dom-repeat`

wchargin:
Summary:
The `dom-repeat` module, which provides typings for the `DomRepeat`
interface, requires special treatment for syncing into google3, so we
route it through an indirection module.

Googlers, see <http://cl/325485393> for corresponding internal change.

Test Plan:
Work-in-progress migration for the custom scalars dashboard seems to be
happy with these type definitions.

wchargin-branch: polymer3-indirect-dom-repeat

### DIFF
--- a/tensorboard/components_polymer3/polymer/BUILD
+++ b/tensorboard/components_polymer3/polymer/BUILD
@@ -15,6 +15,16 @@ tf_ts_library(
 )
 
 tf_ts_library(
+    name = "dom-repeat",
+    srcs = [
+        "dom-repeat.ts",
+    ],
+    deps = [
+        "@npm//@polymer/polymer",
+    ],
+)
+
+tf_ts_library(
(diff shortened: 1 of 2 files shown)


wchargin:
We use `itemForElement` in a bunch of places:
<https://github.com/tensorflow/tensorboard/blob/8ddcceb27e677a9314aad3c2bda055691c12372e/tensorboard/plugins/custom_scalar/tf_custom_scalar_dashboard/tf-custom-scalar-multi-line-chart-card.html#L438>

It’s not strictly something that we _must_ use; we could of course
`any`-cast. But since policy is to avoid `any`-casting when possible,
and there’s enough precedent for shim modules like this, I thought this
approach reasonable.

stephanwlee:
I vaguely remember writing that code. Whoops. :) I don't object to adding the indirection :)

PR #4003, opened 2020-08-08
migrate: remove graph `tensorboard_webcomponent_library`s

wchargin:
Summary:
Unblocks sync. Some of these generate Google-internal test targets that
fail to build due to the sibling deps that no longer exist. These would
be removed in the migration process, anyway.

Follow-up to #3999.

Test Plan:
A test sync at this commit now succeeds: <http://cl/325560211>.

wchargin-branch: migrate-remove-graph-legacies


wchargin:
np!

PR #3932, opened 2020-07-31
migrate: convert graph externally part 2

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/3921
Followup: https://github.com/tensorflow/tensorboard/pull/4004

Migrates the graph plugin to polymer3.  Note that the tf-graph-app binary should still work.

When tested with an experimental branch (https://github.com/psybuzz/tensorboard/tree/mi-graph1test adapted from https://github.com/tensorflow/tensorboard/commit/fc4aba784b27b669d080f8325dead8a4e7a3bfc9), it works in Firefox.  The graph itself is hidden in Chrome, still investigating.

image

PR #4004, opened 2020-08-08
migrate: convert graph to pass internally part 3

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/3932
Followup: https://github.com/tensorflow/tensorboard/pull/4005

Migrating the graph dashboard requires some extra changes to make it
pass more strict conformance rules internally. Most of the changes
involve making type annotations looser.

Notably:
Fixed circular dependencies between JS modules.  `tf_graph_common`
contained lots and lots of cycles, including
- scene.ts --> edge.ts --> scene.ts
- node.ts --> annotation.ts --> node.ts

These were resolved by either moving methods into `common.ts`,
or in the case of `annotation.ts`, merging it into `node.ts`. The
`tf_graph_common` modules now follow a proper layering scheme
(higher layers can depend on lower layers, but never the other way
around).

layer 0: common, proto
layer 1: util
layer 2: parser, graph
layer 3: op, template
layer 4: hierarchy
layer 5: render, loader
layer 6: tf-graph-scene
layer 7: contextmenu, edge, layout
layer 8: scene
layer 9: annotation/node


psybuzz:
> Did you consider making dom_utils (would have selectChild and selectOrCreateChild or node_stats to break the circular dependency?

Although I opted to put those in `common.ts`, I could imagine it being split into `dom_utils.ts` and `types.ts`.  For now, I think common.ts suffices as a small and simple common module.

Btw, I'm not confident the layering I came up with is the right one, but a more holistic approach from someone with expertise in the graph would come up with a much better one.

PR #4006, opened 2020-08-08
histograms: use `irons_and_papers` indirection

wchargin:
Summary:
The initial version of the histograms migration was drafted before the
`irons_and_papers` hourglass dep. It actually works internally as
currently written, but using the indirection is more clearly correct.

Test Plan:
Re-tested the tag filter, offset time selector, and “big chart button”.

wchargin-branch: histograms-irons-and-papers


wchargin:
Test sync: <http://cl/325568479>.

PR #4010, opened 2020-08-11
Migrate plugins/hparams

stephanwlee:
Migrates the hparams plugin.

Do note that removed the vaadin-split-layout by simply using flex (824f9f5).

Relates to #3955 

Co-authored-by: Erik Luoe <luoe@chromium.org>


wchargin:
Non-bug, just noting that some of the styles appear to be different here:

![Minor style differences in old and new hparams][ss]

…so just a note to others who might notice it that a difference here
doesn’t mean that the CSS has been wired up incorrectly.

[ss]: https://user-images.githubusercontent.com/4317806/90557185-f148cc00-e14e-11ea-8f7e-bb3cb54aac18.png

PR #4014, opened 2020-08-11
Fix missing metadata in tbdev getting started notebook.

bileschi:
Missing metadata was breaking import.
Fixed by:

*  Open the notebook in colab
*  Download the notebook from colab using 'download ipynb'
*  Reformat the downloaded notebook using `nbfmt`

Tested:
*  Uploaded to colab and ran.


bileschi:
Tested:
*  Uploaded to colab and ran.

PR #4019, opened 2020-08-11
ng: dispatch changePlugin action when store change

stephanwlee:
This change on purpose introduces one of the anti-pattern of ngrx usage
in our app: composite action.

Background: plugins are rendered when user interacts with the plugins
selector on the header and plugins know to fetch data (if empty) when
`changePlugin` action fires. However, there is other condition in
which the activePlugin can change: (1) user launched with specific
plugin id deep link or (2) plugins listing is loaded. In the past, for
plugins to fetch the data on load, it had to write necessary code for
detecting such nuances which has been a source of pain and a source of
bug.

This change: in the core effects, that understands the core reducers the
best, detects the condition in which the initial non-null value of
activePlugin is set to dispatch `changePlugin` action without user
interaction on the header.

Alternative design: we can subscribe to the store value and listen to
the change but it is actually no different than other composite action:
a state only changes when an action fires and listening to a change in
the store essentially creates a chain between multiple actions (e.g.,
action A -> state change -> action B -> state change -> ...).

### DIFF
--- a/tensorboard/webapp/core/effects/core_effects.ts
+++ b/tensorboard/webapp/core/effects/core_effects.ts
@@ -88,12 +88,14 @@ export class CoreEffects {
   );
 
   /**
-   * HACK: COMPOSITE ACTION -- rationale: most plugins want to be able to tell
-   * when it becomes active in order to, for example, fetch necessary data. By
-   * firing changePlugin on activePlugin value set, we can prevent (1) other
-   * feature developer from responding to values changes from the store (creates
-   * composite actions) and (2) re-implement complex and brittle observable
-   * pattern.
+   * HACK: COMPOSITE ACTION -- Fire `changePlugin` on first truthy value of
+   * activePlugin on coreLoaded or pluginsListingLoaded.
+   *
+   * Rationale: most plugins want to be able to tell when it becomes active in
+   * order to, for example, fetch necessary data. By firing changePlugin on
+   * activePlugin first value set, we can prevent (1) other feature developer
+   * from responding to values changes from the store (creates composite
+   * actions) and (2) re-implement complex and brittle observable pattern.
    *
    * @export
    */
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/core/effects/core_effects_test.ts
```diff
@@ -245,4 +246,116 @@ describe('core_effects', () => {
...
+      expect(recordedActions).toEqual([
+        coreActions.changePlugin({
+          plugin: 'foo',
+        }),
+      ]);
+    });
+
+    it(
+      'ignores plugins listing loaded when activePlugin was present at the time of' +
+        ' coreLaoded',
```
bmd3k, line 333: coreLaoded => coreLoaded.

tensorboard/webapp/core/effects/core_effects.ts
```diff
@@ -84,6 +87,33 @@ export class CoreEffects {
     {dispatch: false}
   );
 
+  /**
+   * HACK: COMPOSITE ACTION -- rationale: most plugins want to be able to tell
```
bmd3k, line 91: Would be nice to first describe what this effect is trying to accomplish.  I think it's something along the lines of "Dispatches changePlugin action when first active plugin is loaded after page load."

tensorboard/webapp/core/effects/core_effects.ts
```diff
@@ -84,6 +87,33 @@ export class CoreEffects {
...
+   * when it becomes active in order to, for example, fetch necessary data. By
+   * firing changePlugin on activePlugin value set, we can prevent (1) other
```
bmd3k, line 93: "By firing changePlugin on activePlugin value set"

Do you mean "By firing changePlugin when activePlugin value is first set?"  Or is this supposed to be a reference to the overall behavior of changePlugin: "By always firing changePlugin when activePlugin value is set".?

PR #4029, opened 2020-08-14
polymer3: add web-animations-js shim to Polymer 3

stephanwlee:
In this change, we use genrule to inline the content since the module
cannot be safely imported from TypeScript internally. In order to have
congruent build system, we just use genrule to concatenate already
compiled sources.

### DIFF
--- a/tensorboard/components_polymer3/BUILD
+++ b/tensorboard/components_polymer3/BUILD
@@ -64,10 +64,21 @@ tf_ts_library(
 )
 
 tf_js_binary(
-    name = "polymer3_lib_binary",
+    name = "polymer3_lib_binary_no_shim",
     compile = True,
     entry_point = ":polymer3_lib.ts",
     deps = [":polymer3_ts_lib"],
+)
+
+genrule(
+    name = "polymer3_lib_binary",
+    srcs = [
+        # Do not sort. order is important.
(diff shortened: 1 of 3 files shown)


tensorboard/components_polymer3/BUILD
```diff
@@ -64,12 +64,23 @@ tf_ts_library(
...
 
+genrule(
+    name = "polymer3_lib_binary",
+    srcs = [
+        # Do not sort. order is important.
+        "@npm//:node_modules/web-animations-js/web-animations-next-lite.min.js",
+        ":polymer3_lib_binary_no_shim.js",
+    ],
+    outs = ["polymer3_lib_binary.js"],
+    cmd = "for f in $(SRCS); do cat \"$$f\"; echo; done > $@",
```
nfelt, line 81: Hm, can you just do `cat $(SRCS) >$@`?

stephanwlee: There is a sourcemap at the end of the file and simple concatenation actually result in wrong code. I wanted new line delimited concatenation. Please advise on how to do that more cleanly.

nfelt: Ah I see, that's fine then, I don't know of a better way offhand.

PR #4030, opened 2020-08-14
Integration with Databricks notebooks

jerrylian-db:

At Databricks, we want to support the `%tensorboard` API. In order to do so, we need specialized TensorBoard display functionality. Our current integration involves monkey patches private APIs of the notebook module. To prevent our current integration from regressing, we want to contribute a Databricks integration.

Our integration follows the pattern of the Colab integration. To detect that we are in a Databricks context, we check if `dbutils`, a Python module only available in Databricks notebooks, is loaded.
We wrote a `_display_databricks` display function that calls a Databricks API to display TensorBoard inline to Databricks notebook.

Screenshots of UI changes: N/A

In the following steps, we use end to end manual tests to verify that the contribution integrates with Databricks, while preserving previous integrations.

### Build TensorBoard
In root of my TensorBoard repo, run `bazel run //tensorboard/pip_package:extract_pip_package` to get a wheel file for TensorBoard.
 
### Databricks
1. Create an account with Databricks community edition. Log into to your Community edition workspace.
2. Go to the cluster creation page. Open the browser javascript console and run `window.prefs.set("enableCustomSparkVersions", true)` in it. Refresh the page. A "custom spark version" field should appear.
3. Paste this DBR (not MLR) runtime image into the custom spark version field: `custom:custom-local__7.x-snapshot-scala2.12__unknown__ml-11816__3e423ba__e0a3e49__jerry.liang__435a42c__format-2.lz4`. Start the cluster.
4. After the cluster has started, go to the libraries view of the cluster configuration UI. Click the "install new" button. A modal will pop-up. Use it to upload your TensorBoard wheel file and install it.
5. Run
    ```python
    from tensorboard import notebook
    notebook._get_context()
    ```
    which returns `'_CONTEXT_DATABRICKS'`
6. Run
    ```python
    %load_ext tensorboard
    %tensorboard --logdir logs
    ```
    Result screenshot:
Screen Shot 2020-08-14 at 2 30 26 AM

Note: this HTTP error is specific to the community edition Databricks workspace. The community edition is feature limited, so TensorBoard doesn't work. However, from the result of `_get_context()` and the "open in a new tab" link, we know that the Databricks TensorBoard display function has been called, as desired.

### Jupyter
1. Create a new python3 virtual environment on your computer. In it, install `Jupyter` and your wheelfile.
2. Open a Jupyter notebook.
3. `pip install` the wheel file.
4. Run
    ```python
    from tensorboard import notebook
    notebook._get_context()
    ```
    which returns `'_CONTEXT_IPYTHON'`
5. Run
    ```python
    %load_ext tensorboard
    %tensorboard --logdir logs
    ```
    Result screenshot:
Screen Shot 2020-08-14 at 12 05 08 AM

### IPython

1. In the same virtual environment as the Jupyter step above, run `IPython`.
2. In IPython shell, run
    ```python
    %load_ext tensorboard
    %tensorboard --logdir logs
    ```
    Result screenshots:
Screen Shot 2020-08-14 at 12 06 39 AM
Screen Shot 2020-08-14 at 12 06 35 AM

### Colab
1. Open a new notebook.
2. Run `!pip uninstall tensorboard`
3. Upload the TensorBoard wheel file to the file directory.
4. Run
    ```python
    from tensorboard import notebook
    notebook._get_context()
    ```
    which returns `'_CONTEXT_COLAB'`
5. Run
    ```python
    %load_ext tensorboard
    %tensorboard --logdir logs
    ```
    Result screenshots:
Screen Shot 2020-08-14 at 12 25 46 AM

In https://github.com/tensorflow/tensorboard/pull/3770, we proposed creating a new notebook API for TensorBoard users to could specify custom display functionality. This not only fulfill our own integration needs but also provide value to other vendors who might be facing similar problems. We concluded that approach is not appropriate because it locks down the display function signature, has a high API surface area, and adds state to the notebook module.

### DIFF
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -28,6 +28,7 @@ import random
 import shlex
 import textwrap
 import time
+import sys
 
 try:
     import html
@@ -46,6 +47,7 @@ from tensorboard import manager
 # Return values for `_get_context` (see that function's docs for
 # details).
 _CONTEXT_COLAB = "_CONTEXT_COLAB"
+_CONTEXT_DATABRICKS = "_CONTEXT_DATABRICKS"
 _CONTEXT_IPYTHON = "_CONTEXT_IPYTHON"
 _CONTEXT_NONE = "_CONTEXT_NONE"
 
@@ -73,6 +75,15 @@ def _get_context():
         if IPython.get_ipython() is not None:
             # We'll assume that we're in a Colab notebook context.
             return _CONTEXT_COLAB
+
+    # In Databricks, the `dbutils` module is preloaded.
+    try:
+        import IPython
+    except ImportError:
+        pass
+    else:
+        if IPython.get_ipython() is not None and "dbutils" in sys.modules:
+            return _CONTEXT_DATABRICKS
 
     # In an IPython command line shell or Jupyter notebook, we can
     # directly query whether we're in a notebook context.
@@ -314,6 +325,7 @@ def _display(port=None, height=None, print_message=False, display_handle=None):
 
     fn = {
         _CONTEXT_COLAB: _display_colab,
+        _CONTEXT_DATABRICKS: _display_databricks,
         _CONTEXT_IPYTHON: _display_ipython,
         _CONTEXT_NONE: _display_cli,
     }[_get_context()]
@@ -367,6 +379,11 @@ def _display_colab(port, height, display_handle):
     else:
         IPython.display.display(script)
 
+def _display_databricks(port, height, display_handle):
+    import IPython
+    ip_shell = IPython.get_ipython()
+    dbutils = ip_shell.ns_table["user_global"]["dbutils"]
+    dbutils.tensorboard.display(port, height, display_handle)
 
 def _display_ipython(port, height, display_handle):
     import IPython.display
--- a/tensorboard/notebook.py
+++ b/tensorboard/notebook.py
@@ -57,6 +57,7 @@ def _get_context():
 
     Returns:
       _CONTEXT_COLAB: If in Colab with an IPython notebook context.
+      _CONTEXT_DATABRICKS: If in Databricks with an IPython notebook context.
       _CONTEXT_IPYTHON: If not in Colab, but we are in an IPython notebook
         context (e.g., from running `jupyter notebook` at the command
         line).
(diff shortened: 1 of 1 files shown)


wchargin:
Hi @jerrylian-db—thanks for the PR! I’ve read through the code and the
context in the PR description (very helpful—thank you!) and we’re
discussing this internally. I hope to get back to you today, or tomorrow
at the latest.

falaki:
> Hi—thanks for your patience. Our primary concern with this approach is
> that it ties TensorBoard core to a proprietary, closed platform in ways
> that we can’t even manually test.
> 
> As [discussed earlier](https://github.com/tensorflow/tensorboard/issues/3752#issuecomment-662622316), I would be happy to entertain designs for a
> public API exposed by TensorBoard core, and it would be fine with me if
> Databricks provided a proprietary implementation of that API only
> accessible from within your platform. One way to implement that would be
> to use Python entry points to discover notebook environment providers.
> We use this mechanism to discover dynamic plugins, and it works well.
> (In particular, it doesn’t require adding any state; it’s declarative.)
> 
> To be clear, in my comments on #3770, I didn’t at all mean to suggest
> that a public API here isn’t appropriate. It’s okay to lock in the
> `display` function signature, as long as we’re at least somewhat
> thoughtful about making it forward-compatible, and I don’t think that
> the API surface area is too large. We’re not going to be able to proceed
> with direct coupling to a closed third-party platform, but I’d be happy
> to guide and advise you to provide an openly accessible API like this.
> 
> Does this seem reasonable to you? Let me know how you’d like to proceed.

@wchargin thanks for getting back on this. Will Google Colab's private APIs be migrated to use this future public API? Is there a timeline for this API? We proposed one which, unfortunately, was not agreeable. I think we will wait to see how you migrate Colab to use the new API and then we can follow.

wchargin:
> Will Google Colab's private APIs be migrated to use this future public
> API?

Yes, definitely. I’d be happy to migrate the Colab integration to use
this API, and in fact I’d consider it almost a requirement to validate
that the API is structured well enough to accommodate it.

I’m a bit more okay with the Colab integration than with the integration
in this PR, since Colab is free to use and the APIs that we use are also
publicly available, but you’re right that philosophically these are
similar and should be on the same level of abstraction.

> Is there a timeline for this API? We proposed one which,
> unfortunately, was not agreeable. I think we will wait to see how you
> migrate Colab to use the new API and then we can follow.

This isn’t something that we’re developing on the TensorBoard team. I’m
offering to help you develop it in collaboration with us, but other than
that there aren’t any plans to implement this (and thus no timelines).

wchargin:
Seeing no updates in the past two months, I’m going to close this for
now, but [my comment above still stands][1]. Thanks!

[1]: https://github.com/tensorflow/tensorboard/pull/4030#issuecomment-675171580

PR #4033, opened 2020-08-14
[Debuggger V1] Remove external reference to the deprecated plugin

caisq:
  * This PR removes the references to the deprecated debugger (v1) plugin.
    This plugin has been marked as deprecated and for removal in the 2.3.0
    release notes. The code itself will be removed in a follow-up change.
  * Removes external references to tensorboard/plugins/debugger, from
     BUILD files, Python modules and Polymer .html files that are not a part of
     plugins/debugger.
 * During testing, it was discovered that the BUILD files in plugins/audio and plugins/images
     needed to be fixed (i.e., dependency on the ":metadata" BUILD target added in
     each respective BUILD file) after plugins/debugger is removed. This probably was due
     to an unmasking of transitive dependencies of some sort.
  * Verified that Debugger V2 still functions properly

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -266,7 +266,6 @@ py_library(
         "//tensorboard/plugins/audio:audio_plugin",
         "//tensorboard/plugins/core:core_plugin",
         "//tensorboard/plugins/custom_scalar:custom_scalars_plugin",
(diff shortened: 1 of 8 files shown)


caisq:
@nickfelt Done.

PR #4037, opened 2020-08-15
cleanup: remove the legacy build in the TensorBoard

stephanwlee:
When TensorBoard 2.3.0 shipped Angular based entry, we supported the
legacy Polymer based entry via "/legacy.html" route. Since the endpoint
was meant to be used very temporary, and since the release is already
cut and released, we are now removing the legacy entry point.


stephanwlee:
Yup, that was going to be my follow up.

PR #4043, opened 2020-08-18
polymer3: cleanup Polymer 3 based components

stephanwlee:
This change removes folders and files in `tensorboard/components/*` (except tf_imports)
and `tensorboard/plugins/*/*` (except polymer3).


stephanwlee:
Created #4056 and #4057. Also cleaned up the description. 

Yes, I did want to reduce the cognitive load by removing separable changes in their separate PR and it ended up in #4049.

stephanwlee:
This change now only has changed to tensorboard/components and tensorboard/plugins/*/(?!polymer3)/.

PR #4045, opened 2020-08-18
data-loader-behavior: add fine-grained batching support

wchargin:
Summary:
The data loader behavior maintains a fine-grained cache of key-value
pairs. This is useful because when the set of requested keys expands,
only the new keys need to be fetched. But, until now, the behavior has
been hard-coded to fire a separate request for each key-value pair.
Clients can have either fine-grained cache invalidation or efficient
batched requests, but not both at the same time. This patch enriches the
behavior to support just that.

In a future change, the scalars dashboard will take advantage of this to
batch requests for multiple runs and a single tag. This will only
require changing the `requestData` function in `tf-scalar-card`.

In doing so, we need to shuffle around the API a bit. Instead of asking
clients to provide `getDataLoadUrl: (Item) => string` plus a separate
function `requestData: (url: string) => Promise<Data>` (where “`Item`”
and “`Data`” are the keys and values, respectively), clients now provide
a single function `requestData` that takes raw `Item`s (now plural),
performs the request(s), and returns the data. The function provides a
stream of key-value pairs, which is represented in callback style for
convenience. (We don’t want to drag Observable into this.)

The purpose of this approach, as opposed to a perhaps more natural
approach that simply adapts `getDataLoadUrl` to return some kind of
request struct with a callback to map a response into key-value pairs,
is to accommodate the variety of existing clients. The structures get
pretty wild: e.g., `tf-line-chart-data-loader` mixes in the behavior but
doesn’t actually provide the needed properties; they’re provided instead
by `tf-scalar-card`, but then `tf-hparams-session-group-details` further
overrides some of those properties of `tf-scalar-card` with an entirely
different backend. It’s a bit wordier for clients, but at least there
are fewer moving pieces to keep track of.

Test Plan:
The scalars, custom scalars, distributions, histograms, PR curves, and
hparams dashboards all work. The fine-grained invalidation on the
scalars dashboard works: e.g., set the tag filter to `mnist` and then to
`mnist|hparams`, and watch only the hparams demo data load; then, set it
to `hparams` and watch the MNIST charts disappear without any repaints
to the hparams demo charts. The post-load callback properly causes
scalar charts’ domains to adjust. The refresh button in the dashboard UI
properly invalidates and re-fetches data.

(Make sure to run with `TB_POLYMER3=1` until that’s the default.)

wchargin-branch: dlb-batch-finegrained

### DIFF
--- a/tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
+++ b/tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
@@ -15,7 +15,7 @@ limitations under the License.
 import {PolymerElement} from '@polymer/polymer';
 import * as _ from 'lodash';
 
-import {Canceller} from '../tf_backend/canceller';
+import {CancelResult, Canceller} from '../tf_backend/canceller';
 import {RequestManager} from '../tf_backend/requestManager';
 
 type CacheKey = string;
@@ -34,6 +34,23 @@ export interface DataLoaderBehaviorInterface<Item, Data>
   dataToLoad: Item[];
 }
 
+// A function that takes a list of items and asynchronously fetches the
+// data for those items. As each item loads, it should invoke the
+// `onLoad` callback with an `{item, data}` pair to update the cache.
+// After all items have finished loading, it should invoke the
+// `onFinish` callback. Conceptually, that this function accepts
+// `onLoad` and `onFinish` as arguments is as if it returned an
+// Observable-style stream of `{item, data}`-pairs, CPS-transformed.
+//
+// Used in `DataLoaderBehavior.requestData`.
+export interface RequestDataCallback<Item, Data> {
+  (
+    items: Item[],
+    onLoad: (kv: {item: Item; data: Data}) => void,
+    onFinish: () => void
+  ): void;
+}
+
 export function DataLoaderBehavior<Item, Data>(
   superClass: new () => PolymerElement
 ): new () => DataLoaderBehaviorInterface<Item, Data> {
@@ -47,16 +64,16 @@ export function DataLoaderBehavior<Item, Data>(
      */
     loadKey = '';
 
-    // List of data to be loaded. By default, a datum is passed to
+    // List of items to be loaded. By default, items are passed to
     // `requestData` to fetch data. When the request resolves, invokes
     // `loadDataCallback` with the datum and its response.
     dataToLoad: Item[] = [];
 
     /**
-     * A function that takes a datum as an input and returns a unique
+     * A function that takes an item as an input and returns a unique
      * identifiable string. Used for caching purposes.
      */
-    getDataLoadName = (datum: Item): CacheKey => String(datum);
+    getDataLoadName = (item: Item): CacheKey => String(item);
 
     /**
      * A function that takes as inputs:
@@ -66,28 +83,11 @@ export function DataLoaderBehavior<Item, Data>(
      * This function will be called when a response from a request to that
      * data URL is successfully received.
      */
-    loadDataCallback!: (component: this, datum: Item, data: Data) => void;
+    loadDataCallback!: (component: this, item: Item, data: Data) => void;
 
-    public requestManager!: RequestManager;
-
-    // A function that takes a datum as argument and makes the HTTP
-    // request to fetch the data associated with the datum. It should return
-    // a promise that either fullfills with the data or rejects with an error.
-    // If the function doesn't bind 'this', then it will reference the element
-    // that includes this behavior.
-    // The default implementation calls this.requestManager.request with
-    // the value returned by this.getDataLoadUrl(datum) (see below).
-    // The only place getDataLoadUrl() is called is in the default
-    // implementation of this method. So if you override this method with
-    // an implementation that doesn't call getDataLoadUrl, it need not be
-    // provided.
-    requestData = (datum: Item) => {
-      return this.requestManager.request(this.getDataLoadUrl(datum));
-    };
-
-    // A function that takes a datum and returns a string URL for fetching
-    // data.
-    getDataLoadUrl!: (datum: Item) => string;
+    // Function that actually loads data from the network. See docs on
+    // `RequestDataCallback` for details.
+    requestData: RequestDataCallback<Item, Data>;
 
     dataLoading = false;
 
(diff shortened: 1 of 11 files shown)


wchargin:
If helpful, cf. #4050 for the actual multiplexing, which I’ve done some
testing of but haven’t properly documented yet, and so isn’t yet ready
for offical review.

tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -66,28 +83,11 @@ export function DataLoaderBehavior<Item, Data>(
...
-    // an implementation that doesn't call getDataLoadUrl, it need not be
-    // provided.
-    requestData = (datum: Item) => {
-      return this.requestManager.request(this.getDataLoadUrl(datum));
-    };
-
-    // A function that takes a datum and returns a string URL for fetching
-    // data.
-    getDataLoadUrl!: (datum: Item) => string;
+    loadDataCallback!: (component: this, item: Item, data: Data) => void;
```
stephanwlee, line 86: Do you think there is a benefit in "batching" this API, too? i.e., call the callback once per all `items`.

tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -192,62 +192,51 @@ export function DataLoaderBehavior<Item, Data>(
...
-              // Promise.prototype.finally instead
-              () => {}
-            )
-            .then(
-              this._canceller.cancellable(({cancelled}) => {
-                if (cancelled) {
-                  return;
+          const dirtyItems = this.dataToLoad.filter((datum) => {
+            const cacheKey = this.getDataLoadName(datum);
+            return !this._dataLoadState.has(cacheKey);
```
stephanwlee, line 198: Is this condition sufficient? Shouldn't we also check whether the item is already loading?

tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -192,62 +192,51 @@ export function DataLoaderBehavior<Item, Data>(
...
-              }),
-              // TODO(stephanwlee): remove me when we can use
-              // Promise.prototype.finally instead
-              () => {}
-            )
-            .then(
-              this._canceller.cancellable(({cancelled}) => {
-                if (cancelled) {
-                  return;
+          const dirtyItems = this.dataToLoad.filter((datum) => {
```
stephanwlee, line 196: Sorry about pre-existing naming I created. Can we, in a follow up, rename these (or create an alias) to say `item` instead of `data`? i.e., we use both right now with `dataToLoad`, `getDataLoadName`, and `dataLoadState` but use `item` in requestData`.

stephanwlee, line 196: I do not remember your plan so I will ask here: what is our plan with the maximum number of runs we batch?

On the `reload`, we expunge the cache (maybe we should only expunge already loaded so we do not double fetch) and invoke `_loadData`. If user has thousands of runs, one batch request can ask for thousand entries which I don't think will scale nicely (if you want some anecdote, please ask offline). Do we plan on breaking the request down into chunks even in this model?

tensorboard/plugins/custom_scalar/polymer3/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.ts
```diff
@@ -347,12 +356,23 @@ class _TfCustomScalarMarginChartCard extends LegacyElementMixin(PolymerElement)
...
-      run,
-    });
+  _requestData: RequestDataCallback<RunItem, CustomScalarsDatum> = (
+    items,
+    onLoad,
+    onFinish
+  ) => {
+    const router = getRouter();
+    const baseUrl = router.pluginRoute('custom_scalars', '/scalars');
+    Promise.all(
```
stephanwlee, line 366: Hmm, I wonder if it makes more sense to just return the promise in this callback.

tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
wchargin: Yeah, fair question. The data loader behavior shouldn’t impose a max
batch size, but I was considering adding a batch size of ~100 in the
`requestData` implementation that calls `/scalars_multirun`. This is
just based off some napkin math: per time series, 1000 points, with
JSON-encoded wall time (~18 bytes), step (~5 bytes), value (~18 bytes),
and overhead (~8 bytes), for about 48 bytes per datum and thus 4.8 KB
per time series. A response size of 500 KB seems entirely reasonable,
and capping it somewhere seems wise to avoid unbounded server load.

> Do we plan on breaking the request down into chunks even in this
> model?

I think that it’s fine to call `requestData` with all the items and let
`requestData` decide if/how to break that up into smaller requests.

tensorboard/plugins/custom_scalar/polymer3/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.ts
wchargin: So, there were a couple models here that I considered. You could just
replace the `onFinish` continuation with “returns `Promise<void>`”,
which I presume is what you’re suggesting. That works, but I thought it
was a bit confusing to mix callbacks and promises. The alternative that
I’d be more interested in is requiring the function to just return an
array of `Promise<{item: Item, data: Data}>`—either that or an
early-bound array of `{item: Item, data: Promise<Data>}`. In some ways
this felt a bit more restrictive because `requestData` has to explicitly
farm out promises for each item, but of course it’s equally flexible
assuming that you don’t do unexpected things like call the `onLoad`
callback multiple times. This also has a nice way to reject individual
promises, which would be a strong argument if `DataLoaderBehavior` had a
notion of failed load states.

Do you have thoughts or preferences here?

tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
wchargin: Wait, no, this is right, right? We don’t want to re-fetch it if it’s
already loading, and so this evaluates to `false` if the load state is
either `LOADING` or `LOADED` (anything but absent).

wchargin:
Response to your action item here:
<https://github.com/tensorflow/tensorboard/pull/4045#discussion_r472655128>

PTAL; if you like, I’m happy to take a stab at any of the refactorings
that I mentioned, but I haven’t started any yet.

tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
stephanwlee: Ah, I think I made a mistake. I definitely meant to only commit the changes on `onLoadFinish`. I will send a separate PR shortly.

tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
stephanwlee: Yup, you are right. Whoops.

tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
stephanwlee: That plan sounds good. Thanks for sharing your plans.

tensorboard/plugins/custom_scalar/polymer3/tf_custom_scalar_dashboard/tf-custom-scalar-margin-chart-card.ts
stephanwlee: Ok, I buy the argument. While I think the node-like APIs are a bit dated, I agree now that maybe we may not want to mix callback + Promise.
 
I definitely think we can re-evaluate `onLoadDataCallback` and `onLoadFinish`. Not sure how that will change the API structure though. Since we have plans to redo scalars, maybe I can borrow your thoughts on the API organization then.

tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
wchargin: Forgot to comment here: I did play around with this and everything still
works, and I see a significant improvement in paint time.

It’d be slightly easier for me if I could send that PR after these two
land, just so that I don’t have to futz with conflicts; is that okay
with you?

stephanwlee: Awesome. It is awesome that it improves the paint time, too.

Please feel free to work on it.

Issue #1283, opened 2018-07-12
Error loading graph in tensorboard with placeholder named "length"


TanguyMoreau:
Tensorflox-gpu : 1.8.0
Tensorboard : 1.8.0
Python : 3.6.5
Distribution : Ubuntu 16.04.4 LTS

### Description of the problem : 
When adding a graph to visualize it in tensorboard, the graph does not print.
![image](https://user-images.githubusercontent.com/16269358/42632282-f4f38428-85dc-11e8-8b37-b28b103aedbf.png)
This problem does not seem to depend on the browser (we tested chrome, firefox and safari). Also the console log of the browser shows this error : 
<img width="267" alt="Browser console log
![tensorboard_bug](https://user-images.githubusercontent.com/16269358/42632368-3922ab42-85dd-11e8-8f09-4e230624fbe8.png)
" src="https://user-images.githubusercontent.com/16269358/42632350-222f19d4-85dd-11e8-81b9-a14f4b2067db.png">

### Minimal example of the problem :
```
Import tensorflow as tf

graph = tf.Graph()
with graph.as_default():
   var1 = tf.placeholder(dtype=tf.int32, shape=(None), name="wedontcare")
   var2 = tf.placeholder(dtype=tf.float32, shape=(None), name="length")
with tf.Session(graph=graph) as sess:
   sess.run(tf.global_variables_initializer())
   writer = tf.summary.FileWriter("graph/")
   writer.add_graph(sess.graph)
```

### What we tried : 
1) It seems the problem occurs when the name `name="length"` is specified in a placeholder. This example works fine : 
```
Import tensorflow as tf

graph = tf.Graph()
with graph.as_default():
   # No palceholder name "length"
   var1 = tf.placeholder(dtype=tf.int32, shape=(None), name="wedontcare")
   var2 = tf.placeholder(dtype=tf.float32, shape=(None), name="thisworks")  
with tf.Session(graph=graph) as sess:
   sess.run(tf.global_variables_initializer())
   writer = tf.summary.FileWriter("graph/")
   writer.add_graph(sess.graph)
```

2) Also it seems we need at least two placeholders to trigger the prolbem. This example works fine : 
```
Import tensorflow as tf

graph = tf.Graph()
with graph.as_default():
   # Only one placeholder
   var2 = tf.placeholder(dtype=tf.float32, shape=(None), name="length")
with tf.Session(graph=graph) as sess:
   sess.run(tf.global_variables_initializer())
[...]

PR #4050, opened 2020-08-18
scalars: multiplex data fetches within a tag

wchargin:
Summary:
As of this patch, a `tf-scalar-card` can make just one network request
to fetch its data, instead of one request per time series (i.e., one
request per run, since each scalar chart renders a single tag). This
reduces network overhead, improves throughput due to higher request
concurrency, and offers the opportunity for data providers to more
efficiently request the data in batch.

This is implemented with a new POST route `/scalars_multirun`, since the
list of run names may be long. The frontend is configured to batch
requests to at most 64 runs at once, so the multiplexing is bounded.

This only affects the scalars plugin. Other scalar chart sources, like
PR curves, custom scalars, and the hparams metrics views, are not
affected.

Supersedes #3835, with the same idea and same general backend approach,
but using the frontend APIs enabled by #4045.

Test Plan:
On the hparams demo with four charts showing, each fetching 50 runs, we
now make only four requests as opposed to 200. On a Google-internal
networked data provider, this improves end-to-end time (measured from
“first spinner appears” to “last spinner disappears”) by about 50%, from
22±1 seconds to 11±1 seconds. (Before this patch, the network requests
were being queued all the way to the end of the test period.)

Changing the batch size to 4 and then running on a dataset with 14 runs
shows that the requests are properly batched, including the last one
with just 2 runs.

Testing hparams, custom scalars, and PR curves shows that they continue
to work, even when multiple time series are requested.

wchargin-branch: scalars-mux-runs


wchargin:
> DO NOT SUBMIT until corresponding internal change is submitted.

No internal change actually needed. (The relevant internal ML dashboard
is isolated from these changes due to the Polymer 3 migration.) I still
do want to better document and profile this, though, so leaving as draft
for the moment.

wchargin:
With a networked data provider (Google-internally) on a dataset with
four scalar charts each with 50 runs, triggering load by using “Toggle
All Runs” to turn all runs from off to on (on clean load cache),
timing from “first spinner appears” to “last spinner disappears”,
I measure an improvement from 22±1 seconds to 11±1 seconds: about 50%.

tensorboard/plugins/scalar/polymer3/tf_scalar_dashboard/tf-scalar-card.ts
```diff
@@ -257,19 +261,56 @@ export class TfScalarCard extends PolymerElement {
...
-      items.map((item) => {
-        const url = addParams(baseUrl, {tag: item.tag, run: item.run});
-        return this.requestManager
-          .request(url)
-          .then((data) => void onLoad({item, data}));
+      requestGroups.map(({tag, runs}) => {
+        return this.requestManager.request(url, {tag, runs}).then((allData) => {
+          for (const run of runs) {
+            const item = {tag, run};
+            if (Object.prototype.hasOwnProperty.call(allData, run)) {
```
stephanwlee, line 307: Any reason why we have to use Object.prototype.hasOwnProperty? (i.e., why not `allData.hasOwnProperty`?)

wchargin: If you have a run called `hasOwnProperty`, then that would be a type
error.

```javascript
> ({train: 1}).hasOwnProperty("train")
true
> ({train: 1, hasOwnProperty: 2}).hasOwnProperty("train")
Thrown:
TypeError: {(intermediate value)(intermediate value)}.hasOwnProperty is not a function
> Object.prototype.hasOwnProperty.call({train: 1, hasOwnProperty: 2}, "train")
true
```

Yeah.

I prefer to write this kind of code defensively—even if it seems
unlikely, it sometimes happens (e.g.: #1283).

tensorboard/plugins/scalar/http_api.md
```diff
@@ -61,3 +61,25 @@ instead be in CSV format:
     1443856985.705543,1448,0.7461960315704346
     1443857105.704628,3438,0.5427092909812927
     1443857225.705133,5417,0.5457325577735901
+
+## `/data/plugin/scalars/scalars_multirun` (POST)
+
+Accepts form-encoded POST data with a (required) singleton key `tag` and a
+repeated key `runs`. Returns a JSON object mapping run names to arrays of the
```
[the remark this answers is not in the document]
wchargin: Zero-length `runs` is fine; it just returns an empty object. Missing
`tag` is a 400. Duplicate `tag`s will silently pick one of them
(probably last?). I can make that an error, though; probably a good
idea. Thanks.

> Also, do you think it is a good idea to perhaps write an example of
> the request?

Yep. Done.

tensorboard/plugins/scalar/polymer3/tf_scalar_dashboard/tf-scalar-card.ts
wchargin: One day <https://github.com/hasOwnProperty> will show up in a JSON
object mapping GitHub login names to database IDs, and someone will have
a fun time…

stephanwlee: Got it. Thanks.

tensorboard/plugins/scalar/http_api.md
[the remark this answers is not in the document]
stephanwlee: Ah, I expected to see FormData but this works too :)

PR #4056, opened 2020-08-20
polymer3: cleanup vestigial components

stephanwlee:
This change removes vestigial components (stuck in Polymer 2) defined in
polymer3 directory that will not be migrated to Polymer 3.


stephanwlee:
@nfelt FYI, forgot about tf_tensorboard folder so removed that too. 

Checked at the end by running `rg 'tensorboard/components/' -g 'tensorboard/components_polymer3/**/*'`.

PR #4059, opened 2020-08-20
[nPMI] Non-Backend Related State

Sparkier:
Adding the part of the state that does not come from the backend.

Added reducers, actions, and types for the part of the state that is user controlled/does not come from the backend.

`bazel test //tensorboard/webapp:karma_test_chromium-local`

### DIFF
--- a/tensorboard/webapp/plugins/npmi/actions/npmi_actions.ts
+++ b/tensorboard/webapp/plugins/npmi/actions/npmi_actions.ts
@@ -13,7 +13,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {createAction, props} from '@ngrx/store';
-import {AnnotationDataListing, MetricListing} from '../store/npmi_types';
+import {
+  AnnotationDataListing,
+  MetricListing,
+  AnnotationSorting,
+} from '../store/npmi_types';
 
 // HACK: Below import is for type inference.
 // https://github.com/bazelbuild/rules_nodejs/issues/1013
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/plugins/npmi/store/npmi_reducers.ts
```diff
@@ -80,6 +104,261 @@ const reducer = createReducer(
...
+      return {
+        ...state,
+        selectedAnnotations: [],
+      };
+    }
+  ),
+  on(
+    actions.flagAnnotations,
+    (state: NpmiState, {annotations}): NpmiState => {
+      let notFlagged: string[] = annotations.filter(
```
stephanwlee, line 157: Another suggestion: would it be possible to change this to a positive word? I had to think harder (I am definitely spoiled by dumb code that I write).

Very dumb code I would write goes like:
```ts

const existingFlaggedAnnotations = new Set(state.flaggedAnnotations);

const newFlags: string[] = annotations.filter(
   (annotation) => existingFlaggedAnnotations.has(annotation)
);

if (newFlags.length) {
  return {
    ...state,
    flaggedAnnotations: [...existingFlaggedAnnotations, ...newFlags],
  };
}

// If all annotations are already flagged, user wants to remove them
for (const annotation of annotations) {
  existingFlaggedAnnotations.delete(annotation);
}

return {
 ...state,
 flaggedAnnotations: [...existingFlaggedAnnotations],
};
```

tensorboard/webapp/plugins/npmi/store/npmi_reducers_test.ts
```diff
@@ -268,4 +268,485 @@ describe('npmi_reducers', () => {
...
+        expect(nextState.metricArithmetic).toEqual([
+          {kind: 'metric', metric: 'nPMI@test'},
+          {kind: 'operator', operator: Operator.AND},
+          {kind: 'metric', metric: 'nPMI@third'},
+        ]);
+      });
+    });
+
+    describe('Change a Filter', () => {
+      it('change a metric filter', () => {
```
stephanwlee, line 616: add a test that adds a new metric filter.

tensorboard/webapp/plugins/npmi/store/npmi_reducers_test.ts
```diff
@@ -268,4 +268,485 @@ describe('npmi_reducers', () => {
...
+        })
+      );
+      expect(nextState.flaggedAnnotations).toEqual([
+        'annotation_1',
+        'annotation_2',
+        'annotation_3',
+      ]);
+    });
+
+    it('flagging annotations with some already flagged', () => {
```
stephanwlee, line 341: say something about not changing the order of the existing flag annotation if that is part of the contract (or add a new spec that, even if it looks similar/same, codifies the contract)

tensorboard/webapp/plugins/npmi/actions/npmi_actions.ts
```diff
@@ -41,3 +45,83 @@ export const npmiPluginDataLoaded = createAction(
...
+export const clearSelectedAnnotations = createAction(
+  '[NPMI] Clearing the Annotation Selection'
+);
+
+export const flagAnnotations = createAction(
+  '[NPMI] Adding/Removing Annotations to/from Flagged',
+  props<{annotations: string[]}>()
+);
+
+export const hideAnnotations = createAction(
```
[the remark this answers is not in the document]
Sparkier: Renamed Action.

tensorboard/webapp/plugins/npmi/store/npmi_reducers.ts
Sparkier: Thank you for that tip.

tensorboard/webapp/plugins/npmi/store/npmi_reducers_test.ts
Sparkier: I think I have that. See lines 440-492.

tensorboard/webapp/plugins/npmi/store/npmi_reducers.ts
```diff
@@ -80,6 +104,250 @@ const reducer = createReducer(
...
+      return {
+        ...state,
+        flaggedAnnotations: [...combinedFlaggedAnnotations],
+      };
+    }
+  ),
+  on(
+    actions.npmiToggleAnnotationsHidden,
+    (state: NpmiState, {annotations}): NpmiState => {
+      const notHidden: string[] = annotations.filter(
```
stephanwlee, line 170: Could you take a similar approach as `npmiToggleAnnotationFlags` here?

tensorboard/webapp/plugins/npmi/store/npmi_reducers.ts
```diff
@@ -80,6 +104,250 @@ const reducer = createReducer(
...
+        metricsRegex: regex,
+      };
+    }
+  ),
+  on(
+    actions.npmiAddMetricFilter,
+    (state: NpmiState, {metric}): NpmiState => {
+      // Only add if not already in active filters
+      const filter = state.metricFilters[metric];
+      if (filter === undefined) {
```
stephanwlee, line 214: instead, do an early return. Here and elsewhere.
```ts
if (filter) {
  return state;
}

 const newContent: ArithmeticElement[] = [];
if (state.metricArithmetic.length !== 0) {
  newContent.push({kind: 'operator', operator: Operator.AND});
}
newContent.push({kind: 'metric', metric: metric});
return {
  ...state,
  metricArithmetic: [...state.metricArithmetic, ...newContent],
  metricFilters: {
    ...state.metricFilters,
    [metric]: {
      max: 1.0,
      min: -1.0,
      includeNaN: false,
    },
  },
};

tensorboard/webapp/plugins/npmi/store/npmi_reducers_test.ts
stephanwlee: Hmm, technically, though, I can pass in a new metric in the changeMetricFilter and the reducer will gladly add a new entry. Can we put an invariant check in the changeMetricFilter reducer? (i.e., the metric should exist in order to change)

tensorboard/webapp/plugins/npmi/store/npmi_reducers.ts
Sparkier: Sure, sorry I missed that.

PR #4060, opened 2020-08-20
npmi: add missing `expect_*` Angular deps

wchargin:
Summary:
Needed to unblock internal sync.

Test Plan:
Test sync CL: <http://cl/327722605>

wchargin-branch: npmi-data-source-angular-deps

### DIFF
--- a/tensorboard/webapp/angular/BUILD
+++ b/tensorboard/webapp/angular/BUILD
@@ -10,6 +10,7 @@ tf_ts_library(
     name = "expect_angular_common_http",
     srcs = [],
     visibility = [
+        "//tensorboard/webapp/plugins/npmi/data_source:__subpackages__",
         "//tensorboard/webapp/webapp_data_source:__subpackages__",
     ],
     deps = [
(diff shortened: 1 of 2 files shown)


wchargin:
(WANT_LGTM=any)

tensorboard/webapp/angular/BUILD
```diff
@@ -10,6 +10,7 @@ tf_ts_library(
     name = "expect_angular_common_http",
     srcs = [],
     visibility = [
+        "//tensorboard/webapp/plugins/npmi/data_source:__subpackages__",
```
wchargin, line 13: @stephanwlee: I’m just expanding this for now to fix the sync, but could
you fill me in on why this visibility is restricted in the first place?
The original change was here:
<https://github.com/tensorflow/tensorboard/pull/3223/files#diff-757b4eaabfbe41a380034bcfc1cbdef1>

The other `expect_*` rules aren’t visibility-restricted, and the actual
`@angular/common/http` dependency isn’t, either, so I wonder whether it
would be more appropriate to just let these be `//tensorboard:internal`
like the rest.

stephanwlee: Ah, that's because no one should directly depend on the http_client but instead use the tb_http_client (it uses relative path vs. absolute path depending on different app).

wchargin: Got it, okay. I figured that that might be it, but wanted to fix the
sync before making substantive changes to the code. Feel free to follow
up with the desired change (either you or @Sparkier), or maybe I can get
to it tomorrow.

PR #4061, opened 2020-08-21
data-loader-behavior: rename `Item`/`Data` to `K`/`V`

wchargin:
Summary:
These names were a bit confusing, especially since identifiers like
`dataToLoad` actually referred to “items” rather than “data”. Since the
module basically provides a key-value store, `K` and `V` should be
straightforward enough names.

Related replacements:

  - `s/dataToLoad/keysToLoad/`
  - `s/data-to-load/keys-to-load/`
  - `s/getDataLoadName/getCacheKey/`

Test Plan:
Checked the scalars, custom scalars, distributions, histograms, hparams,
and PR curves dashboards with their respective demo datasets.

wchargin-branch: dlb-rename-kv

### DIFF
--- a/tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
+++ b/tensorboard/components_polymer3/tf_dashboard_common/data-loader-behavior.ts
@@ -26,36 +26,35 @@ export enum LoadState {
   LOADED,
 }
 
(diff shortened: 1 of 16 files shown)


wchargin:
(Dropping because I’m not realistically going to resolve this soon. More
important things to do.)

PR #4077, opened 2020-08-24
polymer3: cleanup Polymer 3 based components (rollforward)

stephanwlee:
This reverts #4070 which is a revert of #4043.

Tested the sync with cl/327899075 which requires some config change.

Delta from #4043, `git checkout master -- tensorboard/components/tensor_widget`


bileschi:
Made a comment on cl/327899075

stephanwlee:
Given that the change was previously reviewed for correctness, this time, we have done prework to examine the syncability, I will merge it as is.

PR #4078, opened 2020-08-24
build: upgrade TensorFlow workspace dep to 2.3.0

wchargin:
Summary:
The `WORKSPACE` shuffling boils down to the change that TensorFlow now
pulls in `bazel_skylib`, and at a later version than we were using. This
appears to be a silent no-op(…) if any other dependency defines it
first. Some dependencies allow specifying `omit_bazel_skylib = True`;
others, like `rules_sass`, simply check the shared mutable state of
whether it’s been defined already, and define it if it’s not defined
yet. Thus, the workspace definition order must be “just so”.

Test Plan:
Brief check of the scalars, graphs, and projector plugins, plus
invocation of `tensorboard dev list`.

wchargin-branch: workspace-tf-2.3.0


wchargin:
Thanks to your cache fixes, this now finishes comfortably on time. :-)

PR #4083, opened 2020-08-25
refactor npmi dependency so no one depends on angular directly

stephanwlee:


Sparkier (approved):
Looks Good. Thanks for fixing this.

Issue #4091, opened 2020-08-27
[infra] Travis CI build on tf-nightly times out


psybuzz:
Several in-progress PRs have had a Travis CI run that failed only on the tf-nightly setup with a timeout error.
Even after removing lots of Polymer 2 code, we are seeing these timeouts.

See
https://travis-ci.org/github/tensorflow/tensorboard/builds/721612298
https://travis-ci.org/github/tensorflow/tensorboard/builds/721662030
https://travis-ci.org/github/tensorflow/tensorboard/builds/721479907

One option is to try re-building the Travis cache.  A current plan is the following modifications to .travis.yml:
- a PR `bazel build`, but disabled tests (now a good build is cached)
- a PR with both `bazel build` and the tests enabled (now a good build and tests are cached)

stephanwlee:
You may actually want to stage this:
1. only have bazel fetch
2. bazel fetch + build
3. bazel fetch + build + test
4. etc...

(1) should be relatively fast and end within 5-10 minutes.
(2) this is expected to take a long time. Given that our cache is growing to 8GB, download and upload alone will take a long time. (and may push to 1:40hrs limit)

psybuzz:
Closing now that the series has landed.

Each CI round currently runs 3 jobs, one per config.  For reference, here is the max running time spent per job, in different PRs:

Fetch = 17:03
Fetch + components_polymer3 = 22:38
Fetch + components_polymer3 + webapp = 39:55
Fetch + tensorboard/... = 50:51
Fetch + tensorboard/... + tests = 16:24

PR #4092, opened 2020-08-27
[infra] updating Travis cache with fetch-only flow (1/3)

psybuzz:
This change should have no visible effect to the product.
The Travis flow is updated to ignore the build, test steps that
come after the `bazel fetch`.

2 followups will re-add bazel build, test steps to rebuild the
Travis cache.

See https://github.com/tensorflow/tensorboard/issues/4091

PR #4093, opened 2020-08-27
[infra] update Travis cache with built components_polymer3 (part 2)

psybuzz:
This change should have no visible effect to the product.
The Travis flow is updated to reinstate the build step, but
keep ignored the test steps that come after the `bazel build`.

This PR adds the components_polymer3/ dir to the bazel build
step.  Followups will re-add remaining steps.

See https://github.com/tensorflow/tensorboard/issues/4091

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,11 +92,13 @@ before_script:
 # categorize them as 'failed', rather than 'error' for other sections.
 script:
   - elapsed "script"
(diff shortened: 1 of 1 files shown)


psybuzz:
Thanks, I hadn't thought to wait for the master cache for #4092 to finish.  I'll definitely wait for 4093's build on master to finish before step 3/3.

PR #4098, opened 2020-08-28
test_util: remove obsolete `run_v2_only` decorator

wchargin:
Summary:
This decorator was only used in one test, and that test is not
compatible with no-TensorFlow mode, so as of #2507 it will only ever run
in TF 2.x. We don’t plan to support any more 1.x-only tests in the
future, so this decorator has served its purpose and may now retire.

Test Plan:
CI suffices.

wchargin-branch: remove-v2only-decorator


tensorboard/util/test_util.py
```diff
@@ -209,10 +206,3 @@ def _impl(reason=None):
         return unittest.skipUnless(guard(), reason)
 
     return _impl
-
-
-run_v2_only = _run_conditionally(
```
nfelt, line 214: Might as well also remove _run_conditionally() since this is its last usage and I don't know that we have any immediately foreseeable need for it?

wchargin: Good point; thanks! Done.

PR #4099, opened 2020-08-28
core: fix `--window_title` in Angular TensorBoard

wchargin:
Summary:
This behavior is slightly different from Polymer TensorBoard, in that if
the window title is set but then the server restarts and the window
title is no longer set, it will revert to “TensorBoard” instead of
retaining its previous value.

This implementation uses a new view instead of an effect so that we can
read the environment information from the store without violating our
policies about composite actions.

Test Plan:
Launch TensorBoard with `--window_title one`, and note that the title is
properly set. Keep the frontend open, and relaunch the server with
`--window_title two`. Press the in-app “Reload” button, and verify that
the title updates. Finally, relaunch the server with no `--window_title`
and verify that the frontend resets the title to “TensorBoard”.

No TypeScript tests because I can’t think of anything that would go
wrong without being caught by the type system.

wchargin-branch: ng-window-title

### DIFF
--- a/tensorboard/webapp/core/views/page_title_component.ts
+++ b/tensorboard/webapp/core/views/page_title_component.ts
@@ -1,0 +1,37 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {
+  ChangeDetectionStrategy,
+  Component,
+  Input,
+  OnChanges,
+  SimpleChanges,
+} from '@angular/core';
+
+@Component({
+  selector: 'page-title-component',
+  template: '',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class PageTitleComponent implements OnChanges {
+  @Input()
+  title!: string | null;
+
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes['title'] != null) {
+      document.title = changes['title'].currentValue;
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/core/views/page_title_component.ts
```diff
@@ -0,0 +1,37 @@
...
+} from '@angular/core';
+
+@Component({
+  selector: 'page-title-component',
+  template: '',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class PageTitleComponent implements OnChanges {
+  @Input()
+  title!: string | null;
```
stephanwlee, line 30: Is the prop really nullable?

tensorboard/webapp/core/views/page_title_container.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import {getEnvironment} from '../store';
+import {State} from '../store/core_types';
+
+/** @typehack */ import * as _typeHackRxjs from 'rxjs';
+
+@Component({
+  selector: 'page-title',
+  template: `
+    <page-title-component [title]="title$ | async"> </page-title-component>
```
stephanwlee, line 27: omit extra whitespace: 

```
...async"> </page...>
          ^
          |
```

tensorboard/webapp/core/views/page_title_component.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes['title'] != null) {
```
stephanwlee, line 33: Can this never be `null`? I guess it can be uninitialized but I don't think it will have other non-truthy values, right?

tensorboard/webapp/core/views/page_title_component.ts
wchargin: Nope (was in earlier draft). Fixed; thanks.

tensorboard/webapp/core/views/page_title_container.ts
wchargin: Done; thanks.

tensorboard/webapp/core/views/page_title_component.ts
wchargin: I considered just using `this.props.title` here and skipping the check,
but figured that in case we add more attributes, like

```html
<page-title-component [title]="foo" [animateWithEmoji]="of-course" />
```

then we’d want to check the change record, so I included this.

Rewritten to just `if (changes['title'])` per your preference.

stephanwlee: Got it. Makes sense. New code reads/looks fine :)

PR #4112, opened 2020-08-31
chore: remove comments on BUILD licenses()

psybuzz:
Google internally now recommends avoiding comments on the
`licenses()` rule, arguing that they cause confusion and easily
become out of sync with the source-of-truth LICENSE file.

This adds a CI check in the GitHub Actions workflow to check
for license comments.

Tested that the CI action works in the test PR:
https://github.com/tensorflow/tensorboard/pull/4113
https://github.com/tensorflow/tensorboard/pull/4113/checks?check_run_id=1053651164

See also
https://opensource.google/docs/thirdparty/linter/#build-licenses-comment%20.no-toc
Googlers: cl/326102766


wchargin:
One approach would be to run [`buildozer ... remove_comment`][1] in CI
and assert that no files changed.

[1]: https://github.com/tensorflow/tensorboard/pull/4090#pullrequestreview-476058331

psybuzz:
Thanks for the tips on adding a check!

ci/download_buildozer.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
```
psybuzz, line 1: Note: this file is mainly copy/paste from `ci/download_buildifier.sh`.  I don't know shell scripts well enough, but the only difference is replacing s/buildifier/buildozer

.github/workflows/ci.yml
```diff
@@ -125,6 +130,9 @@ jobs:
         # Use | to start a literal so YAML doesn't complain about the '!' character.
         run: |
           ! git grep 'python_version = "PY2"' '*BUILD'
+      - name: 'No comments on licenses rule'
+        run: |
+          ! buildozer '//tensorboard/...:%licenses' remove_comment
```
nfelt, line 135: Buildozer exits non-zero when there was nothing to do (the intended case we want here) but it also exits non-zero for errors: https://github.com/bazelbuild/buildtools/blob/e8b2086fd9d7027b97ae8f49744d570fc6c299b4/buildozer/README.md#error-code

So just using `!` to invert the exit status is over-broad here and would hide legitimate errors.  I think instead it might work to just do

```shell
buildozer '//tensorboard/...:%licenses' remove_comment; test $? = 3
```

Worth checking that that does still correctly catch true positives though.

stephanwlee: `//tensorboard/...` would not catch ones in `//third_party/...` which is a mistake that I made.

psybuzz: Updated to check the specific error code.  Tested that it still correctly catches bad cases in https://github.com/tensorflow/tensorboard/pull/4113

psybuzz:
> Also, do note that `buildozer '//third_party/...:%licenses' remove_comment` has no effect. I am not sure if that works in js.bzl since it is not technically `licenses` "target" but they are defined with macros.
> 
> In the hindsight, maybe we should keep the licenses in the .bzl files since they are not licenses of ours but that of our dependencies and they help me understand what kind of licenses we are using.

The more I'm read, the more I think the way we organize our 3P deps is strange.  I think a restructuring should happen, perhaps something like:
```
- third_party/
  - dagre
    - src/
    - LICENSE
    - BUILD
```

However, that is out of scope for this PR, intended to just unblock the sync.  For now, I've preserved the comments in the macros under tensorboard/third_party/, but moved the one for bleach.BUILD to the top of the BUILD file, in case the linter is aggressive.

psybuzz:
It seems the lint-build check now always fails :(

```
buildozer '//tensorboard/...:%licenses' remove_comment; test $? = 3
##[error]Process completed with exit code 3.
```

This doesn't make sense to me, because locally the above command has exit code 0, and if we try running just the remove_comment part (without the test $? = 3), it still has exit code 3:
https://github.com/tensorflow/tensorboard/pull/4115/checks?check_run_id=1057450448

nfelt:
Sorry, my bad for not testing my suggestion thoroughly enough; GitHub actions runs the shell script steps with `set -e -o pipefail` to fail-fast on command failures, and I mistakenly didn't realize that this causes it to abort the script after the first part of the line (the `buildozer` call) without getting to the second part (the `test`).

This should fix it: `buildozer '//tensorboard/...:%licenses' remove_comment && false || test $? = 3

The logic is that adding the `&& false` basically says: if the preceding command exited cleanly, then make the whole command here fail; and the `|| test $? = 3` says "if the preceding command failed, check if the exit code was 3, and if so, make the whole command here pass".  Chaining with `&&` and `||` is exempt from the fail-fast behavior; we need the initial `&& false` so that the first part of the command before the `||` _always_ fails and thus always triggers the exit code test.

Tested this time:

```shell
$ (set -e -o pipefail; (exit 0) && false || test $? = 3); echo $?
1
$ (set -e -o pipefail; (exit 1) && false || test $? = 3); echo $?
1
$ (set -e -o pipefail; (exit 2) && false || test $? = 3); echo $?
1
$ (set -e -o pipefail; (exit 3) && false || test $? = 3); echo $?
0
```

psybuzz:
Thanks for the detailed explanation @nfelt !

PR #4116, opened 2020-09-01
Rename .*polymer3 folder to their original paths

stephanwlee:
As part of the Polymer 3 migration, we forked the files, migrated, and pointed the application to the migrated fork.
As we are done with the code migration and removal of Polymer 2 files, we are putting back the files where they were.

Please review the PRs commit by commit. First commit should mostly rename the files then we change the import/
package declaration paths.

Merge plan: squash (since intermediate commits do not build as is)


psybuzz (approved):
Thanks for splitting it into reviewable commits.  Might be worth doing a sync presubmit test before merging.

rebeccaann95531 (commented):
Yes that's right

PR #4117, opened 2020-09-01
chore: misc sync fixes (unused in workspace.bzl, license comments)

psybuzz:
This change unbreaks the sync with some misc fixes:
- Removes licenses() comments under npmi's metric_arithmetic/
- Removes unused 'loads' in workspace.bzl

Drive by changes to workspace.bzl:
- Added to CI lint checks, to catch future instances where it may
  not pass Buildifier
- Reformatting slightly via Buildifier
- Remove unused 'com_google_protobuf_js'

Googlers, see
example failure: cl/329551428
example success: cl/329569832

### DIFF
--- a/third_party/workspace.bzl
+++ b/third_party/workspace.bzl
@@ -16,7 +16,6 @@
 TensorBoard external dependencies that can be loaded in WORKSPACE files.
 """
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:java.bzl", "java_import_external")
 load("@io_bazel_rules_webtesting//web/internal:platform_http_file.bzl", "platform_http_file")
 load("@npm_bazel_typescript//:index.bzl", "ts_setup_workspace")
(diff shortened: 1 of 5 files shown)


psybuzz:
Thanks @stephanwlee for the comment to remove com_google_protobuf_js.  It seems to work fine

third_party/workspace.bzl
```diff
@@ -12,105 +12,96 @@
...
 load("//third_party:typings.bzl", "tensorboard_typings_workspace")
 
-def tensorboard_workspace():
-  """Add repositories needed to build TensorBoard."""
-  tensorboard_fonts_workspace()
-  tensorboard_polymer_workspace()
-  tensorboard_python_workspace()
-  tensorboard_typings_workspace()
-  tensorboard_js_workspace()
+def tensorboard_workspace(name = ""):
```
psybuzz, line 28: In reaction to Buildifier's complaint:
```
third_party/workspace.bzl:29: unnamed-macro: By convention "tensorboard_workspace" should have a keyword argument called "name".
```

third_party/workspace.bzl
```diff
@@ -12,105 +12,96 @@
...
+"""
+TensorBoard external dependencies that can be loaded in WORKSPACE files.
+"""
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:java.bzl", "java_import_external")
-load("@io_bazel_rules_closure//closure:defs.bzl", "filegroup_external")
-load("@io_bazel_rules_closure//closure:defs.bzl", "web_library_external")
-load("@io_bazel_rules_webtesting//web/internal:platform_http_file.bzl", "platform_http_file")
+load("@io_bazel_rules_webtesting//web/internal:platform_http_file.bzl", "platform_http_file")  # buildifier: disable=bzl-visibility
```
psybuzz, line 20: In reaction to Buildifier's complaint:
```
third_party/workspace.bzl:21: bzl-visibility: Module "@io_bazel_rules_webtesting//web/internal:platform_http_file.bzl" can only be loaded from files located inside "//web", not from "//third_party/workspace.bzl". (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#bzl-visibility)
```

stephanwlee: Do you know how/why bazel complains? I suspect it is because the path has "internal" or "private". If my hypothesis is correct, perhaps we should use this technique to hide all of our .bzls :)

psybuzz: Yup, it does a pretty simple check that the path contains those keywords:
https://github.com/bazelbuild/buildtools/blob/master/warn/warn_visibility.go

We could hide ours too, but it only protects us if consumers run and respect `buildifier` warnings!

PR #4120, opened 2020-09-02
infra: upgrade FE deps

psybuzz:
Changes a few frontend dependencies:

- tsconfigs now pull in 'es2020' rather than 'es2018' typedefs.
  Despite TS docs [1] not mentioning es2020 as an option, our
  version of TS supports this option [2]. Notably this adds
  typing support for `Array.prototype.flat()` (part of ES2019).
- tsconfigs have some more strict options that were not
  part of the flags [3] included by `"strict": true`.
- rxjs bumped to 7.0.0beta0. The changelog [4] notes several
  breaking changes, but none seem to affect us. Notably this
  new version includes the `combineLatestWith` operator.

Manually clicked around the UI, and behavior seems unaffected.

[1] https://www.typescriptlang.org/docs/handbook/compiler-options.html
[2] https://github.com/microsoft/TypeScript/pull/29167
[3] https://github.com/microsoft/TypeScript/blob/b0011feee125b8945f7ae97711fdacc38769e069/src/compiler/utilities.ts#L5922
[4] https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "numeric": "^1.2.6",
     "plottable": "^3.9.0",
     "requirejs": "^2.3.6",
-    "rxjs": "^6.5.5",
+    "rxjs": "7.0.0-beta.0",
     "three": "~0.108.0",
     "umap-js": "^1.3.2",
     "web-animations-js": "^2.3.2",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,8 +5,10 @@
     "experimentalDecorators": true,
     "importHelpers": true,
     "inlineSourceMap": true,
-    "lib": ["dom", "es2019"],
+    "lib": ["dom", "es2020"],
     "moduleResolution": "node",
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitReturns": true,
     "skipLibCheck": true,
     "strict": true,
     // Don't scan the node_modules/@types folder for ambient types.
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,7 +5,7 @@
     "experimentalDecorators": true,
     "importHelpers": true,
     "inlineSourceMap": true,
-    "lib": ["dom", "es2018"],
+    "lib": ["dom", "es2019"],
     "moduleResolution": "node",
     "skipLibCheck": true,
     "strict": true,
--- a/yarn.lock
+++ b/yarn.lock
@@ -5120,7 +5120,14 @@ rxjs@6.5.4:
   dependencies:
     tslib "^1.9.0"
 
-rxjs@^6.5.3, rxjs@^6.5.5:
+rxjs@7.0.0-beta.0:
+  version "7.0.0-beta.0"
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -120,7 +120,8 @@
     "numeric": "^1.2.6",
     "plottable": "^3.9.0",
     "requirejs": "^2.3.6",
-    "rxjs": "^6.5.5",
+    "rxjs": "^7.0.0-beta.0",
+    "static-eval": "^2.1.0",
```
stephanwlee, line 124: we don't have direct usage of static-eval, right? What is this achieving? If anything, it should be devDependency.

tsconfig.json
```diff
@@ -5,8 +5,10 @@
     "experimentalDecorators": true,
     "importHelpers": true,
     "inlineSourceMap": true,
-    "lib": ["dom", "es2018"],
+    "lib": ["dom", "es2020"],
     "moduleResolution": "node",
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitReturns": true,
```
stephanwlee, line 11: nice!

stephanwlee (changes requested):

package.json
psybuzz: Correct, removed for now.

static-eval, bl, are transitive deps of weblas, which doesn't have any newer version.  Might be worth migrating off weblas someday, as it's not actively updated and we only use it for GPU-matmul.

yarn.lock
```diff
@@ -5120,7 +5120,14 @@ rxjs@6.5.4:
...
 
-rxjs@^6.5.3, rxjs@^6.5.5:
+rxjs@7.0.0-beta.0:
+  version "7.0.0-beta.0"
+  resolved "https://registry.yarnpkg.com/rxjs/-/rxjs-7.0.0-beta.0.tgz#41f730761e00ba7b5a28955e0f177afba8e2a134"
+  integrity sha512-MMsqDczs2RzsTvBiH6SKjJkdAh7WaI6Q0axP/DX+1ljwFm6+18AhQ3kVT8gD7G0dHIVfh5hDFoqLaW79pkiGag==
+  dependencies:
+    tslib "^1.9.0"
+
+rxjs@^6.5.3:
```
stephanwlee, line 5130: I think this means one of our dependencies still lead to rxjs@6.6.0. We may want to identify and make sure it is fine.

stephanwlee (approved):

yarn.lock
psybuzz: The rxjs@6.5.3, 6.5.4 versions in the yarn.lock are only used from `@angular/cli`.  Although I confirmed that yarn install still installs rxjs 7 beta and the app builds/runs fine, it's possible some breaking change in rxjs 7 beta may silently affect `@angular/cli`.

psybuzz:
Thanks for the valuable feedback!

PR #4122, opened 2020-09-03
deps: upgrade Prettier to 2.1.1

### LINKED ISSUE #3989: 🧹: Upgrade prettier to 2.x
Upgrading formatter like prettier often creates unwanted changes. For prettier 2.0, too, configuration + formatting has changed a lot ([changelog](https://prettier.io/blog/2020/03/21/2.0.0.html)).

Now, we are now on TypeScript 3.8 using new JS/TS language features like optional chaining (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) but prettier fails to parse and fails violently. The support for 3.8 features were added at prettier 2.0 and, here, I propose we do another round of massive formatting changes (in ~2w).

wchargin:
This is done as three commits so that we can tell Git to ignore the
middle commit (a pure reformat) when showing blame. (A follow-up PR will
update the `.git-blame-ignore-revs` file; we need to know the commit
hash to do that, which will only be known after this PR is rebased and
merged.)

Closes #3989.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "postinstall": "ngcc -p main es2015",
     "build": "bazel build //...",
     "test": "ibazel test //...",
-    "lint": "prettier --check 'tensorboard/**/*.'{css,html,js,ts,scss} .github/**/*.yml",
+    "lint": "prettier --check 'tensorboard/**/*.'{css,html,js,ts,scss} .github/**/*.yml || echo 'Failures OK during Prettier upgrade'",
     "fix-lint": "prettier --write 'tensorboard/**/*.'{css,html,js,ts,scss} .github/**/*.yml"
   },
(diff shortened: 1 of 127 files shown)


stephanwlee (approved):
What is the intended merge strategy?

wchargin:
> What is the intended merge strategy?

Rebase and merge, then send a follow-up PR to add the middle commit’s
hash to `.git-blame-ignore-revs`.

stephanwlee:
Ah, it was written :) I expected to see the usual merge-plan section in the PR description

wchargin:
Ah, I forgot to blame the reformatting change to TensorBoard Gardener.
Ah, well; it’s fine.

PR #4127, opened 2020-09-05
[nPMI] Implemented Download

Sparkier:
Added the download button for report generation.

* Screenshots of UI changes
image

### DIFF
--- a/tensorboard/webapp/plugins/npmi/util/BUILD
+++ b/tensorboard/webapp/plugins/npmi/util/BUILD
@@ -111,3 +111,32 @@ tf_ts_library(
         "@npm//rxjs",
     ],
 )
+
+ng_module(
+    name = "csv_result",
+    srcs = [
+        "csv_result.ts",
+    ],
+    deps = [
+        "metric_type",
+        "//tensorboard/webapp/plugins/npmi/store:types",
+    ],
+)
+
+tf_ts_library(
+    name = "csv_result_test_lib",
+    testonly = True,
+    srcs = ["csv_result_test.ts"],
+    deps = [
+        ":csv_result",
+        "//tensorboard/webapp/angular:expect_angular_core_testing",
+        "//tensorboard/webapp/core/testing",
+        "//tensorboard/webapp/plugins/npmi/store:types",
+        "@npm//@angular/common",
+        "@npm//@angular/compiler",
+        "@npm//@angular/core",
+        "@npm//@angular/platform-browser",
+        "@npm//@types/jasmine",
+        "@npm//rxjs",
+    ],
+)
--- a/tensorboard/webapp/plugins/npmi/util/BUILD
+++ b/tensorboard/webapp/plugins/npmi/util/BUILD
@@ -118,7 +118,7 @@ ng_module(
         "csv_result.ts",
     ],
     deps = [
-        "metric_type",
+        ":metric_type",
         "//tensorboard/webapp/plugins/npmi/store:types",
     ],
 )
--- a/tensorboard/webapp/plugins/npmi/util/csv_result.ts
+++ b/tensorboard/webapp/plugins/npmi/util/csv_result.ts
@@ -19,16 +19,19 @@ export function convertToCSVResult(
   flaggedData: [string, ValueData[]][],
   run: string,
   metrics: string[]
-): string[][] {
+): string {
   const result = [[run, ...metrics]];
-  if (metrics.length === 0 || flaggedData.length === 0) {
-    return result;
+  if (!metrics.length || !flaggedData.length) {
+    const csvContent =
+      'data:text/csv;charset=utf-8,' +
+      result.map((e) => e.join(',')).join('\n');
+    return csvContent;
   }
   const strippedMetrics = metrics.map((metric) => stripMetricString(metric));
-  for (const element of flaggedData) {
-    const runValues = element[1].filter((values) => values.run === run);
-    if (runValues.length > 0) {
-      const elementResult = [element[0]];
+  for (const [annotation, valueDataList] of flaggedData) {
+    const runValues = valueDataList.filter((values) => values.run === run);
+    if (runValues.length) {
+      const elementResult = [annotation];
       for (const metric of strippedMetrics) {
         const metricValue = runValues.find((value) => value.metric === metric);
         if (metricValue === undefined) {
@@ -40,5 +43,7 @@ export function convertToCSVResult(
       result.push(elementResult);
     }
   }
-  return result;
+  const csvContent =
+    'data:text/csv;charset=utf-8,' + result.map((e) => e.join(',')).join('\n');
+  return csvContent;
 }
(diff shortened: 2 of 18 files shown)


tensorboard/webapp/plugins/npmi/views/data_selection/results_download/results_download_container.ts
```diff
@@ -0,0 +1,119 @@
...
+        map(([runToMetrics, activeRuns, metricFilters]) => {
+          let metrics: string[] = [];
+          for (const run of activeRuns) {
+            if (runToMetrics[run]) {
+              metrics = metrics.concat(
+                runToMetrics[run].filter((key) => metricIsNpmiAndNotDiff(key))
+              );
+            }
+          }
+          metrics = [...new Set([...Object.keys(metricFilters), ...metrics])];
```
psybuzz, line 87: Consider
L79: let metrics = Object.keys(metricFilters);
L87: return [...new Set(metrics)];

tensorboard/webapp/plugins/npmi/views/data_selection/results_download/results_download_test.ts
```diff
@@ -0,0 +1,84 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
psybuzz, line 1: If the `metricIsNpmiAndNotDiff` in results_download_container is an important piece of business logic, it might also be good to have a test case for that as well

tensorboard/webapp/plugins/npmi/views/data_selection/results_download/results_download_container.ts
```diff
@@ -0,0 +1,119 @@
...
+    this.flaggedAnnotations$
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe((x) => (this.flaggedAnnotations = x));
+    this.store
+      .select(getAnnotationData)
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe((x) => (this.annotationData = x));
+    this.activeRuns$
+      .pipe(takeUntil(this.ngUnsubscribe))
+      .subscribe((x) => (this.runs = x));
```
psybuzz, line 71: Hm, this is an interesting pattern.  Admittedly, I don't know whether it's a good/bad practice to subscribe manually and set component properties like this.

In many other containers, a custom `createSelector` manipulated the data into the form we need, and passes it via async pipe to a component.  Reasoning about when container/component properties update is simpler because only no local property state in the container needs to be kept.

Wdyt about replacing all the observable subscriptions with
```
annotationData$ = this.store.select(getAnnotationData
flaggedAnnotations$ = this.store.select(getFlaggedAnnotations
metrics$ = createSelector(getRunToMetrics, getMetricFilters, getRunSelection
    (runToMetrics, metricFilters, runSelection, annotationData, flaggedAnnotations) => {
        // all the logic to produce the metrics 
    }
```
then async-pipe'ing these into the component, and moving the entire `downloadRequested()` from container into component?

There would be no local state in this container to maintain, and no need to ngUnsubscribe, since we'd be using async pipes.

tensorboard/webapp/plugins/npmi/util/BUILD
```diff
@@ -111,3 +111,32 @@ tf_ts_library(
...
     ],
 )
+
+ng_module(
+    name = "csv_result",
+    srcs = [
+        "csv_result.ts",
+    ],
+    deps = [
+        "metric_type",
```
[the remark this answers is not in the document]
Sparkier: I built, so I guess it worked. Am surprised myself, though.

tensorboard/webapp/plugins/npmi/util/csv_result.ts
```diff
@@ -0,0 +1,44 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {ValueData} from '../store/npmi_types';
+import {stripMetricString} from './metric_type';
+
+export function convertToCSVResult(
```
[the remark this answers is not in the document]
Sparkier: Agreed.

tensorboard/webapp/plugins/npmi/views/data_selection/results_download/results_download_container.ts
```diff
@@ -0,0 +1,119 @@
...
+    const flagSet = new Set(this.flaggedAnnotations);
+    const flaggedData = Object.entries(this.annotationData).filter((entry) =>
+      flagSet.has(entry[0])
+    );
+    for (const run of this.runs) {
+      const csvResult = convertToCSVResult(flaggedData, run, this.metrics);
+      let csvContent =
+        'data:text/csv;charset=utf-8,' +
+        csvResult.map((e) => e.join(',')).join('\n');
+      var element = document.createElement('a');
```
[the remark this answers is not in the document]
Sparkier: Done.

tensorboard/webapp/plugins/npmi/views/data_selection/results_download/results_download_component.ts
```diff
@@ -0,0 +1,32 @@
...
+} from '@angular/core';
+
+@Component({
+  selector: 'results-download-component',
+  templateUrl: './results_download_component.ng.html',
+  styleUrls: ['./results_download_component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ResultsDownloadComponent {
+  @Input() flaggedAnnotations!: string[];
```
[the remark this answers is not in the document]
Sparkier: That does not work, as I need the number of flagged annotations. I changed it to numFlaggedAnnotations instead.

tensorboard/webapp/plugins/npmi/views/data_selection/results_download/results_download_test.ts
Sparkier: Yes, agreed. This is part of another PR that I requested Stephans review for where I implemented tests for all the remaining util functions.

tensorboard/webapp/plugins/npmi/views/data_selection/results_download/results_download_container.ts
Sparkier: Sure, I can do that. Just thought that the container should not have such methods as it should not assume anything about the state. And I guess I would implement it like this as I think we dont need a `createSelector` here:
```
readonly metrics$ = combineLatest([
      this.store.select(getRunToMetrics),
      this.activeRuns$,
      this.store.select(getMetricFilters)
    ])
      .pipe(
        map(([runToMetrics, activeRuns, metricFilters]) => {
          let metrics = Object.keys(metricFilters);
          for (const run of activeRuns) {
            if (runToMetrics[run]) {
              metrics = metrics.concat(
                runToMetrics[run].filter((key) => metricIsNpmiAndNotDiff(key))
              );
            }
          }
          metrics = [...new Set(metrics)];
          return metrics;
        })
      )
```

psybuzz: As per offline discussion: combineLatest instead of createSelector sounds good.  I'm more concerned about replacing the `.subscribe(callback)`s.  L108 - L114 can move into the component without giving it much business logic.

psybuzz (approved):
Thanks!

PR #4130, opened 2020-09-07
[nPMI] Fixed Run Sorting for Annotation and Height of Rows

Sparkier:
Fixed incorrect soorting and patched annotations to correct height.

### DIFF
--- a/tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
+++ b/tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
@@ -42,7 +42,7 @@ export class AnnotationComponent implements AfterViewInit, OnChanges {
   @Input() flaggedAnnotations!: string[];
   @Input() hiddenAnnotations!: string[];
   @Input() activeMetrics!: string[];
-  @Input() activeRuns!: string[];
+  @Input() numActiveRuns!: number;
   @Input() showCounts!: boolean;
   @Input() annotation!: string;
   @Input() runHeight!: number;
@@ -149,12 +149,14 @@ export class AnnotationComponent implements AfterViewInit, OnChanges {
     this.data.forEach((element) => {
       runs.add(element.run);
     });
-    this.runs = this.activeRuns.filter((run) => runs.has(run));
-    this.svg.style('height', `${this.activeRuns.length * this.runHeight}px`);
-    this.width = this.annotationContainer.nativeElement.clientWidth || 10;
-    this.chartWidth = this.width - this.margin.left - this.margin.right;
+    this.runs = [...runs];
+    // Needs to be numActiveRuns * runHeight because of cdk-virtual-croll
+    this.svg.style('height', `${this.numActiveRuns * this.runHeight}px`);
+    // Needs to be runs.length * runHeight as this is used for y-pos scales
     this.chartHeight =
       this.runs.length * this.runHeight - this.margin.top - this.margin.bottom;
+    this.width = this.annotationContainer.nativeElement.clientWidth || 10;
+    this.chartWidth = this.width - this.margin.left - this.margin.right;
   }
 
   private setTextClass() {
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
```diff
@@ -176,7 +176,7 @@ export class AnnotationComponent implements AfterViewInit, OnChanges {
 
     this.yScale
       .rangeRound([0, this.chartHeight - this.runHeight])
-      .domain(this.runs.map((d: string) => d));
```
psybuzz, line 179: I'm curious why this changes.

Before, we are creating a copy of `this.runs` before passing it to D3's scale.domain().  Are we fixing cases where making copies was unnecessary?  If so, should we also update the `domain([...this.runs])` callsites below in this file?

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
```diff
@@ -149,12 +149,12 @@ export class AnnotationComponent implements AfterViewInit, OnChanges {
...
-    this.runs = [...runs];
-    this.svg.style('height', `${this.runs.length * this.runHeight}px`);
-    this.height = this.runs.length * this.runHeight;
+    this.runs = this.activeRuns.filter((run) => runs.has(run));
+    this.svg.style('height', `${this.activeRuns.length * this.runHeight}px`);
     this.width = this.annotationContainer.nativeElement.clientWidth || 10;
     this.chartWidth = this.width - this.margin.left - this.margin.right;
-    this.chartHeight = this.height - this.margin.top - this.margin.bottom;
+    this.chartHeight =
+      this.runs.length * this.runHeight - this.margin.top - this.margin.bottom;
```
psybuzz, line 157: Is it intentional that `this.svg.style.height` is out of sync with `this.chartHeight`?  It's not clear to me why the former uses `activeRuns.length` and the latter uses `runs.length`.  Doesn't that lead to cases where the chartHeight is clipped?

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
```diff
@@ -42,6 +42,7 @@ export class AnnotationComponent implements AfterViewInit, OnChanges {
   @Input() flaggedAnnotations!: string[];
   @Input() hiddenAnnotations!: string[];
   @Input() activeMetrics!: string[];
+  @Input() activeRuns!: string[];
```
psybuzz, line 45: Is it necessary for the AnnotationComponent to be aware of the difference between active and inactive runs?  If the component never renders both at the same time in the UI, perhaps we can reduce the component's concerns.

An alternative to adding this field, is to just filter the `@Input() data` being passed into the component to only include data for the active runs.

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
Sparkier: The map function did not do anything and yes, there does not need to be a copy, that is why I removed it.

Agreed, the other copies also should not be made. Removed them.

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
Sparkier: The SVG should always be `activeRuns.length * runHeight` to ensure all rows have the same height. I need to do this to be able to make use of `cdk-virtual-scroll-viewport`.

The drawing area which also determines the scales and thus y-positions of all d3 elements should, however, only consider the runs that the current annotation contains.

I added code comments for that.

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
Sparkier: It actually already does only include data for active runs.

I thus only need this for the aforementioned svg height calculation and changed this to be numActiveRuns.

psybuzz (approved):

PR #4131, opened 2020-09-07
[nPMI] Util Tests and Annotations Row Filtered by Regex

Sparkier:
Added tests for utils and fixed annotation filtering.


tensorboard/webapp/plugins/npmi/util/filter_annotations.ts
```diff
@@ -26,27 +26,31 @@ export function filterAnnotations(
...
-        allMetrics.has(valueDataElement.metric)
-      );
-    });
-    if (
-      checkArithmeticParts(metricArithmetic, metricFilters, valueDataElements)
-    ) {
-      if (valueDataElements.length !== 0) {
-        data[entry[0]] = valueDataElements;
+    const filterRegex = new RegExp(annotationsRegex, 'i');
+    if (filterRegex.test(entry[0])) {
```
stephanwlee, line 39: use the early terminate strategy to avoid very nested blocks/ifs.

tensorboard/webapp/plugins/npmi/util/filter_annotations.ts
```diff
@@ -26,27 +26,31 @@ export function filterAnnotations(
...
+      let valueDataElements = entry[1];
+      // Remove all inactive runs and keep only metrics currently displayed
+      valueDataElements = valueDataElements.filter((valueDataElement) => {
+        return (
+          allRuns.has(valueDataElement.run) &&
+          allMetrics.has(valueDataElement.metric)
+        );
+      });
+      if (
+        checkArithmeticParts(metricArithmetic, metricFilters, valueDataElements)
```
stephanwlee, line 49: `checkArithmeticParts` does not read well in my head. Under what criterion is it returning true? Consider refactoring the name to capture the contract.

tensorboard/webapp/plugins/npmi/util/metric_type_test.ts
```diff
@@ -0,0 +1,80 @@
...
+import {
+  metricIsMetricCount,
+  metricIsNpmi,
+  metricIsNpmiAndNotDiff,
+  stripMetricString,
+} from './metric_type';
+
+describe('metric type utils', () => {
+  describe('metric is count', () => {
+    it('returns true if the metric starts with count@', () => {
```
stephanwlee, line 24: for thoroughness, add a test with input `foo@count` and/or `foo@count@` (to make sure it is not a substring match).

tensorboard/webapp/plugins/npmi/util/metric_type_test.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    it('returns false if the metric does not start with count@', () => {
+      const metric = 'nPMI@test';
+      expect(metricIsMetricCount(metric)).toBeFalse();
+    });
+  });
+
+  describe('metric is npmi', () => {
+    it('returns true if the metric starts with nPMI@', () => {
```
stephanwlee, line 36: is `nPMIFoobar` legal? How about `npmi@test`? Add them as tests.

tensorboard/webapp/plugins/npmi/util/filter_annotations_test.ts
```diff
@@ -0,0 +1,455 @@
...
+    });
+
+    it('returns correct result with annotations regex active', () => {
+      const result = filterAnnotations(
+        annotationData,
+        activeRuns,
+        metricArithmetic,
+        metricFilters,
+        metrics,
+        '1'
```
stephanwlee, line 344: Make sure it is regex with more interesting regex? `1$` or `Ann.+_1$` would do the job.

tensorboard/webapp/plugins/npmi/util/filter_annotations_test.ts
```diff
@@ -0,0 +1,455 @@
...
+  const metrics = ['nPMI@test', 'nPMI@other'];
+  const hiddenAnnotations = ['annotation_2', 'annotation_3'];
+
+  describe('filter annotations', () => {
+    it('returns correct result filtered by runs, active metrics, and metric filters', () => {
+      const result = filterAnnotations(
+        annotationData,
+        activeRuns,
+        metricArithmetic,
+        metricFilters,
```
stephanwlee, line 144: Could you add a case where the annotation is filtered away by the arithmetic operations?

tensorboard/webapp/plugins/npmi/util/filter_annotations_test.ts
```diff
@@ -0,0 +1,455 @@
...
+  AnnotationDataListing,
+  ArithmeticElement,
+  ArithmeticKind,
+  Operator,
+  MetricFilterListing,
+} from '../store/npmi_types';
+import {filterAnnotations, removeHiddenAnnotations} from './filter_annotations';
+
+describe('filter annotations utils', () => {
+  const annotationData: AnnotationDataListing = {
```
stephanwlee, line 25: General feedback: I think you are repeating this type of fixture creation in many tests.
For ease of refactor when the data structure changes later, it is often a good idea to use a builder/factory method.

Also, when reading the test specs, it is ideal to be able to tell why you get the result you are asserting without jumping too much around the file (i.e., metricFilters is used over and over and I have to remind myself how it is defined). 

If you think any of this is a good idea, feel free to refactor away here or in a follow up.

Sparkier: I think this is a good point. I guess a follow-up PR would make sense here.

Sparkier: Did that now for the annotation data.

tensorboard/webapp/plugins/npmi/testing/index.ts
```diff
@@ -58,3 +58,98 @@ export function appStateFromNpmiState(metricsState?: NpmiState): State {
     [NPMI_FEATURE_KEY]: metricsState || createNpmiState(),
   };
 }
+
+export function getSampleAnnotationData() {
```
stephanwlee, line 62: We tend to use prefix like `create` for creating a fixture. Similarly, we use prefix `build` for fixture builder.

tensorboard/webapp/plugins/npmi/util/filter_annotations.ts
stephanwlee: Did you forget to push a commit? I don't think this comment is resolved.

stephanwlee (approved):

PR #4141, opened 2020-09-08
Roll forward "Automatically Change sorting when Metric Filter Added (#4133)"

wchargin:
Summary:
This rolls forward commit 19e23ed79e18d1cc3e7d69368fc505d168d127bf with
minor name changes (`sorting` to `sort`, `SortingOrder` to `SortOrder`)
so that it builds.

Test Plan:
Running `bazel build //tensorboard` still completes successfully.

wchargin-branch: roll-forward-4133

### DIFF
--- a/tensorboard/webapp/plugins/npmi/store/npmi_reducers_test.ts
+++ b/tensorboard/webapp/plugins/npmi/store/npmi_reducers_test.ts
@@ -472,7 +472,7 @@ describe('npmi_reducers', () => {
         expect(nextState.metricArithmetic).toEqual([
           {kind: ArithmeticKind.METRIC, metric: 'nPMI@test'},
         ]);
-        expect(nextState.sorting).toEqual({
+        expect(nextState.sort).toEqual({
           metric: 'nPMI@test',
           order: SortOrder.DOWN,
         });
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/plugins/npmi/store/npmi_reducers_test.ts
```diff
@@ -472,6 +472,10 @@ describe('npmi_reducers', () => {
         expect(nextState.metricArithmetic).toEqual([
           {kind: ArithmeticKind.METRIC, metric: 'nPMI@test'},
         ]);
+        expect(nextState.sorting).toEqual({
```
Sparkier, line 475: This is what you missed. Should be `nextState.sort`.

Sparkier (approved):

tensorboard/webapp/plugins/npmi/store/npmi_reducers_test.ts
wchargin: Fixed; thanks. I ran `bazel build //tensorboard`, which worked, but
didn’t build the web tests. Everything actually passes now. :-)

PR #4144, opened 2020-09-08
Remove legacy multiplexer codepaths for plugins at parity

wchargin:
Summary:
In TensorBoard 2.3.0, we promoted the generic data codepaths to be on by
default (#3559) for plugins at parity. This was intended to be a silent
change, and we haven’t heard any complaints or bug reports. In this
patch, we remove the legacy codepaths entirely, simplifying the code and
tests. This follows the release plan previously set out:
<https://github.com/tensorflow/tensorboard/pull/3265#issuecomment-588615473>

Recommend hiding whitespace changes when reviewing this patch. Commits
separated for reviewer convenience; I plan to squash and merge, but
could be convinced to rebase.

Test Plan:

Checking for `context.multiplexer` only shows uses in places that
actually need it:

```
$ git grep -l -w context.multiplexer tensorboard/plugins/
tensorboard/plugins/custom_scalar/custom_scalars_plugin.py
tensorboard/plugins/graph/graphs_plugin.py
tensorboard/plugins/mesh/mesh_plugin.py
tensorboard/plugins/projector/projector_plugin.py
```


tensorboard/plugins/image/images_plugin_test.py
```diff
@@ -47,6 +45,15 @@
...
 
 class ImagesPluginTest(tf.test.TestCase):
+    def setUp(self):
+        super().setUp()
+        (logdir, multiplexer) = self._create_data()
+        provider = data_provider.MultiplexerDataProvider(multiplexer, logdir)
+        ctx = base_plugin.TBContext(logdir=logdir, data_provider=provider)
+        plugin = images_plugin.ImagesPlugin(ctx)
+        self._initialize_plugin_specific_attrs(plugin)
+        return plugin
```
nfelt, line 55: `return plugin` doesn't do much inside `setUp()`?

tensorboard/plugins/image/images_plugin_test.py
```diff
@@ -47,6 +45,15 @@
 
 
 class ImagesPluginTest(tf.test.TestCase):
+    def setUp(self):
+        super().setUp()
+        (logdir, multiplexer) = self._create_data()
+        provider = data_provider.MultiplexerDataProvider(multiplexer, logdir)
+        ctx = base_plugin.TBContext(logdir=logdir, data_provider=provider)
+        plugin = images_plugin.ImagesPlugin(ctx)
+        self._initialize_plugin_specific_attrs(plugin)
```
nfelt, line 54: This `setUp()` would probably read better with this call inlined now that it's only used once, but up to you since it's existing code

nfelt (approved):
Yay, exciting to see this code going away 🥳

tensorboard/plugins/image/images_plugin_test.py
wchargin: I was considering it, and am happy to do so. More red.

tensorboard/plugins/image/images_plugin_test.py
wchargin: Indeed. Some might say it does nothing at all.

PR #4149, opened 2020-09-10
cleanup: break projector dependency on weblas with tf

stephanwlee:
We were using weblas for the matrix multiplication in GPU. Since the
library is not maintained regularly and depends on vulnerable versions
of libraries, we break the dependency to weblas by replacing it with
TensorFlow.js.

When the demo was run with tf.js, I did not see anything out of
ordinary.

Plan afterwards: will remove dependency on weblas.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "@polymer/paper-toolbar": "^3.0.1",
     "@polymer/paper-tooltip": "^3.0.1",
     "@polymer/polymer": "^3.4.1",
+    "@tensorflow/tfjs": "^2.3.0",
     "@vaadin/vaadin-grid": "^5.6.6",
     "d3": "5.7.0",
     "dagre": "^0.8.5",
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-projections-panel.ts
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-projections-panel.ts
@@ -205,7 +205,7 @@ class ProjectionsPanel extends LegacyElementMixin(PolymerElement) {
       if (this.dataSet && this.projector) {
         this.dataSet.perturbTsne();
         this.projector.notifyProjectionPositionsUpdated();
-        this.perturbInterval = setInterval(() => {
+        this.perturbInterval = window.setInterval(() => {
           this.dataSet.perturbTsne();
           this.projector.notifyProjectionPositionsUpdated();
         }, 100);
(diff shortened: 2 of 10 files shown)


tensorboard/plugins/projector/vz_projector/vz-projector-projections-panel.ts
```diff
@@ -205,7 +205,7 @@ class ProjectionsPanel extends LegacyElementMixin(PolymerElement) {
       if (this.dataSet && this.projector) {
         this.dataSet.perturbTsne();
         this.projector.notifyProjectionPositionsUpdated();
-        this.perturbInterval = setInterval(() => {
+        this.perturbInterval = window.setInterval(() => {
```
psybuzz, line 208: I'm curious about this changing when it worked before.  Was there some new TS error?

package.json
```diff
@@ -111,6 +111,7 @@
     "@polymer/paper-toolbar": "^3.0.1",
     "@polymer/paper-tooltip": "^3.0.1",
     "@polymer/polymer": "^3.4.1",
+    "@tensorflow/tfjs": "^2.3.0",
     "@vaadin/vaadin-grid": "^5.6.6",
     "d3": "5.7.0",
     "dagre": "^0.8.5",
```
psybuzz, line 117: Can we drop the `"weblas": "^0.9.1"` on L128, and remove weblas from yarn.lock, too?  Or does something else still use it?

psybuzz (approved):

package.json
stephanwlee: > Plan afterwards: will remove dependency on weblas.

As described in the description, I have a plan to remove weblas everywhere afterwards (tf_imports/js.bzl/etc).

tensorboard/plugins/projector/vz_projector/vz-projector-projections-panel.ts
stephanwlee: It was surprising to me too.

If I were to guess, since TF.js works in node, TSC is confused as to which version of `setInterval` this is referring to (node version does not return number but an object or something else; I forget). As such, we need to be a bit more explicit about which version of setInterval we are using by referring to `window` (as opposed to `global`) in our ts_library. 

Thoughts? Is this change still good?

psybuzz: The change still lgtm to land.

I see others online that suggest simply using `window.setInterval` too [1].
[1] https://github.com/Microsoft/TypeScript/issues/30128

PR #4155, opened 2020-09-11
Plugin/Projector: Document renaming of standalone build target

aknoerig:
The mentioned `devserver` target does no longer exist in the build file. It has been renamed to `standalone`.

### DIFF
--- a/tensorboard/plugins/projector/README.md
+++ b/tensorboard/plugins/projector/README.md
@@ -2,7 +2,7 @@
 
 To develop the Embedding Projector, launch it in standalone mode:
 ```sh
-bazel run tensorboard/plugins/projector/vz_projector:devserver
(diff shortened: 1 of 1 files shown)


aknoerig:
@googlebot I signed it!

stephanwlee (approved):
Sorry about changing the target name. 
I did not realize people were relying on our development server.

If it is not any trouble, could you please tell us how you are using the binary?

aknoerig:
Thanks! The standalone build is very much appreciated. I think the projector, or a modified version of it, could be useful in the context of a professional, customer-facing AI tool. That's what I would like to explore.

luggie:
Well same here. I'm also very interested in the standalone build as I'm currently experimenting with it for my master's thesis.

PR #4157, opened 2020-09-12
projector: fix UMAP dependency issue

stephanwlee:
While doing Polymer 3 migration, we changed everything to use JavaScript
module system (as opposed to tf_web_library). In doing so, we forgot
about the UMAP dependency but TSC did not complain about it because of
our d.ts file.

This change fixes the import and also remove the d.ts file.


wchargin (approved):
Confirmed fixed for me; thanks! Great to hear that the more standard
TypeScript toolchain means that we can drop the manual `d.ts` hack.

Issue #4159, opened 2020-09-14
Projector Plugin: Bookmark loading is broken


aknoerig:
## Environment information (required)

I am using the standalone build of the projector plugin from master.

- bazel 2.1.0
- Chrome 85.0.4183.102

## Issue description

In the projector plugin, the selected bookmark fails to load.

Steps to recreate:

1. Build the projector plugin using the standalone target
1. Open in browser
1. Switch to the "Word2Vec All" sample dataset
1. Select the sample bookmark "Politics: Bad to Good"

Console output:

```
Uncaught TypeError: evt.path[i].getAttribute is not a function
    at HTMLElement.getParentDataIndex (vz-projector-inspector-panel.mjs:404)
    at HTMLElement._radioButtonHandler (vz-projector-inspector-panel.mjs:377)
    at handler (apply-shim.js:264)
    at HTMLElement.<anonymous> (iron-a11y-keys-behavior.js:295)
    at _fire (dom-if.js:63)
    at trackForward (dom-if.js:444)
    at Object.click (dom-if.js:403)
    at HTMLElement._handleNative (dom-repeat.js:644)
```

Probably this is another side-effect of the recent upgrade to Polymer3?

PR #4160, opened 2020-09-14
Projector Plugin: Fix bookmark loading (issue #4159)

aknoerig:

Bookmark loading currently fails in the projector plugin (see issue #4159)

The failure is because of a missing type check in `getParentDataIndex()`, crashing on `getAttribute()`. This change makes sure the object that is checked against is actually an HTMLElement.

Same as described in issue #4159 

There is probably be a better way to find the data-index of the selected bookmark. The current approach seems a bit hackish. But that's beyond the scope of the bug fixed here.

### DIFF
--- a/tensorboard/plugins/projector/vz_projector/vz-projector-bookmark-panel.ts
+++ b/tensorboard/plugins/projector/vz_projector/vz-projector-bookmark-panel.ts
@@ -197,9 +197,12 @@ class BookmarkPanel extends LegacyElementMixin(PolymerElement) {
    */
   private getParentDataIndex(evt: Event) {
     for (let i = 0; i < (evt as any).path.length; i++) {
-      let dataIndex = (evt as any).path[i].getAttribute('data-index');
-      if (dataIndex != null) {
-        return +dataIndex;
+      let elem = (evt as any).path[i];
+      if (elem instanceof HTMLElement) {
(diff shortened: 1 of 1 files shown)


bileschi:
@aknoerig sorry for the delay!  Assigning to @psybuzz for review.

hfiller (approved):
👍 there are other ways, but this is easy to read and just as efficient.

tensorboard/plugins/projector/vz_projector/vz-projector-bookmark-panel.ts
```diff
@@ -197,9 +197,12 @@ class BookmarkPanel extends LegacyElementMixin(PolymerElement) {
    */
   private getParentDataIndex(evt: Event) {
     for (let i = 0; i < (evt as any).path.length; i++) {
-      let dataIndex = (evt as any).path[i].getAttribute('data-index');
-      if (dataIndex != null) {
-        return +dataIndex;
+      let elem = (evt as any).path[i];
```
psybuzz, line 200: optional + minor: ideally these `let`s would be `const`s, since they don't get updated.

Since it was there before, also fine with the PR as is.

psybuzz (approved):
also lgtm!

psybuzz:
Thank you for contributing, @aknoerig !

PR #4162, opened 2020-09-14
proto_test: on failure, show all non-empty diffs

wchargin:
Summary:
We now list the diffs of all proto definitions that aren’t up to date
rather than just the first one.

Test Plan:
This test currently passes, but if you revert #4163 and use current
`tf-nightly`s then it will fail. Do so, then run the test manually
(`bazel test //tensorboard/compat/proto:proto_test`) and note that the
output contains diffs from 9 files.

wchargin-branch: proto-test-show-all-failures


wchargin:
@ericdnielsen: Friendly ping?

ericdnielsen (approved):

PR #4165, opened 2020-09-15
[oss-timeseries] add deps for typing, color picker, platform-browser

psybuzz:
Time Series code requires
- types/resize-observer-browser
- ngx-color-picker
- angular/platform-browser


package.json
```diff
@@ -112,12 +112,14 @@
     "@polymer/paper-tooltip": "^3.0.1",
     "@polymer/polymer": "^3.4.1",
     "@tensorflow/tfjs": "^2.3.0",
+    "@types/resize-observer-browser": "^0.1.3",
```
stephanwlee, line 115: move this to devDependency. type files are used for building the tool and is not to be used in the runtime (I know that for non-Node projects, it is kinda awkward and arbitrary distinction but devDep is consistent with other type imports)

yarn.lock
```diff
@@ -5830,6 +5842,11 @@ tslib@^1.10.0, tslib@^1.8.1, tslib@^1.9.0:
   resolved "https://registry.yarnpkg.com/tslib/-/tslib-1.13.0.tgz#c881e13cc7015894ed914862d276436fa9a47043"
   integrity sha512-i/6DQjL8Xf3be4K/E6Wgpekn5Qasl1usyw++dAA35Ue5orEn65VIxOA+YvNNl9HV3qv70T7CNwjODHZrLwvd1Q==
 
+tslib@^2.0.0:
```
stephanwlee, line 5845: Hmm, I wonder whether we need to upgrade our tslib, too (there is `tslib@~1.8.0` below that has no dep)

stephanwlee (approved):

package.json
psybuzz: Done.

yarn.lock
psybuzz: I tried deleting and regenerating yarn.lock, but the tslib@~1.8.0 remains, presumably due to Plottable's dep on it.

psybuzz (commented):
In the latest commit, I also downgraded `ngx-color-picker` to match the internal version 8.2.0.  This version does not introduce a new tslib version dep.

yarn.lock
stephanwlee: I guess it is no longer relevant but presumably, we would want to upgrade https://github.com/tensorflow/tensorboard/blob/3010b2c6afba6abbf68cf3f759c31d7b70d8daf9/package.json#L65 if we want to go 2.0 (and I think we eventually should; but do note that TS 3.8.3 is not compatible with tslib 2.x; https://github.com/microsoft/tslib/releases)

psybuzz: Gotcha, will keep in mind for the future

PR #4172, opened 2020-09-17
[oss-timeseries] add testing/, types/, util/

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/4171

Adds testing/, types/, util/ folders for Time Series.

Googlers, see test sync cl/332324994 for confidence that this
does not block the sync.

### DIFF
--- a/tensorboard/webapp/util/lang.ts
+++ b/tensorboard/webapp/util/lang.ts
@@ -16,8 +16,10 @@ type MapObjectValuesTransformer = (value: any, key: string) => any;
 
 /**
  * Returns a new object where all values have been transformed.
- * Similar to https://lodash.com/docs/#mapValues or
- * https://underscorejs.org/#mapObject
+ * For familiarity, the API signature was made to be as close as possible to the
+ * signatures of equivalent functions in popular libraries:
+ * - https://lodash.com/docs/#mapValues
+ * - https://underscorejs.org/#mapObject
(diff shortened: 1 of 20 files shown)


tensorboard/webapp/util/lang.ts
```diff
@@ -0,0 +1,32 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+type MapObjectValuesTransformer = (value: any, key: string) => any;
+
+/**
+ * Returns a new object where all values have been transformed.
+ * Similar to https://lodash.com/docs/#mapValues or
+ * https://underscorejs.org/#mapObject
```
stephanwlee, line 20: Hmm, did we copy the API signature or the implementation, too? If latter, we need to attribute this code to them, right?

stephanwlee (approved):
Please make sure to properly attribute borrowed code if it is truly borrowed.

tensorboard/webapp/util/lang.ts
psybuzz: The implementation is not borrowed, but I've updated the comment in the code to note that the API signature was intended to be similar to popular libraries' equivalents for familiarity.

PR #4173, opened 2020-09-17
[oss-timeseries] add widgets/, app_routing/

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/4172

Adds widgets/, app_routing/ folders for Time Series.

Googlers, see test sync cl/332324994 for confidence that this
does not block the sync.


tensorboard/webapp/widgets/resize_detector_test.ts
```diff
@@ -45,20 +45,20 @@ describe('resize detector', () => {
...
-        triggerResize = () => {
-          callback([], resizeObserver);
-        };
-        resizeObservers.push(resizeObserver);
-        return resizeObserver;
-      }
-    );
+    spyOn(window, 'ResizeObserver').and.callFake(function (
+      callback: ResizeObserverCallback
+    ) {
```
psybuzz, line 50: `function()` functions can be treated as constructors, while arrow functions cannot.  My hypothesis is that internally, some downleveling to ES5 converts the latter to the former, making it safe.  In OSS, the arrow function is preserved through to the test, causing it to fail.

stephanwlee: Hmm, perhaps there is a discrepancy in the typing shim?

stephanwlee (approved):

tensorboard/webapp/widgets/resize_detector_test.ts
psybuzz: While there is a typing discrepancy, the error happens at runtime not build time.  I suppose it's possible our OSS type checking is less strict somehow.  Either that, or something internally rewrites the arrow fn into a function fn.

PR #4180, opened 2020-09-18
[oss-timeseries] add experiment shim, upgrade deps for runs/

psybuzz:
In preparation for the runs/ migration, this change
- Adds an experiments/ folder, a basic shim with a default
  experiment id
- Upgrades ngx-color-picker once more. Version 8.2.0 has
  an anonymous AMD module, which fails at runtime in test
  environments. Version 9+ of the npm package comes
  with a named AMD module, which works out of the box
- Renames the existing runs/ to runs_legacy/, to avoid
   confusion / BUILD file edits

Note: it may be possible to get anonymous AMD modules
working in tests, but tweaking `require_js_karma_config.js`
and `tf_ng_web_test_suite` to accommodate is not trivial.


tensorboard/webapp/experiments/types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+export interface Experiment {
+  // int64 should be represented in string as JS can only represent number up to
+  // 53 bits.
```
stephanwlee, line 17: Consider omitting this; TB services may provide different types of experiment id and string should cover all cases (and this detail does not matter).

stephanwlee (approved):
> Upgrades ngx-color-picker once more. Version 8.2.0 has
an anonymous AMD module, which fails at runtime in test
environments. Getting it to work by tweaking
require_js_karma_config.js and tf_ng_web_test_suite
were not trivial. Version 9+ of the npm package comes
with a named AMD module, which works out of the box

Hmm, I do not see any changes to require_js_karma_config.js or defs.bzl. 
Is the change not contained in this PR?

psybuzz:
> Hmm, I do not see any changes to require_js_karma_config.js or defs.bzl.
Is the change not contained in this PR?

Ahh, I think the PR description I wrote was misleading.  Originally,
I wanted to tweak require_js_karma_config and defs.bzl, but failed
to get them to fix the issue.

As a result, I gave up on that approach (getting an anonymous
AMD module to load properly) and instead just upgraded the
ngx-color-picker.

I'll update the description

tensorboard/webapp/experiments/types.ts
psybuzz: Done

PR #4181, opened 2020-09-18
[oss-timeseries] add runs/, util/ui_selectors.ts

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/4180

Adds runs/, util/ui_selctors.ts for Time Series.

Googlers, see test sync cl/332542898 for confidence that this
does not block the sync.

### DIFF
--- a/tensorboard/webapp/runs/store/runs_selectors_utils.ts
+++ b/tensorboard/webapp/runs/store/runs_selectors_utils.ts
@@ -1,0 +1,137 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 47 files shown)


tensorboard/webapp/runs/store/runs_selectors_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+            `Metric, ${specB.tag}, datasetTypes have to match. Got: ` +
+              `${specA.datasetType} vs. ${specB.datasetType}`
+          );
+        }
+      }
+    }
+  }
+
+  if (validationError.length) {
+    // TODO(b/157733179): handle the error gracefully on the view side.
```
stephanwlee, line 118: I assume this is an intended change?

stephanwlee (approved):
Wow, this is a lot of code. I admit I have not carefully read through every line but since this is just a port/copy, I will trust test/lint mostly.

tensorboard/webapp/runs/store/runs_selectors_utils.ts
psybuzz: Yes, this line is intended.  It would be good to perhaps move all internal bug #s into new GitHub issues, but that can be done independently outside of this PR series.

runs_selectors_utils.ts was not manually altered in this PR's manual changes:
https://github.com/tensorflow/tensorboard/pull/4181/files/9ae8c76d4e128d1054b2cc138e250eba42b27a97..da8db23b5648e8d5268c1401f5c4b97d9e2a3bfc

PR #4182, opened 2020-09-19
[oss-timeseries] add metrics/

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/4181

Adds metrics/ for Time Series.

Googlers, see test sync cl/332574849 for confidence that this
does not block the sync.


tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
```diff
@@ -79,7 +79,7 @@
     <img
       alt="Image at step {{ stepValues[stepIndex] }}"
       src="{{ imageUrl }}"
-      [style.filter]="cssFilter()"
+      [ngStyle]="{'filter': cssFilter()}"
```
psybuzz, line 82: Angular complains on dynamic bindings to style.filter, but it is okay with ngStyle.
`'WARNING: sanitizing unsafe style value contrast(100%) brightness(2) (see http://g.co/ng/security#xss).`

tensorboard/webapp/metrics/effects/index.ts
```diff
@@ -42,14 +42,18 @@ import {
   TagMetadata,
   TimeSeriesRequest,
   TimeSeriesResponse,
-} from '../data_source';
+} from '../data_source/index';
```
psybuzz, line 45: Without this, requirejs fails at runtime with
`There is no timestamp for org_tensorflow_tensorboard/tensorboard/webapp/metrics/data_source.js!`

third_party/js.bzl
```diff
@@ -525,25 +525,39 @@ def tensorboard_js_workspace():
...
-            "ic_expand_more_24px.svg": "expand_more_24px.svg",
-            "ic_expand_less_24px.svg": "expand_less_24px.svg",
-            "ic_cancel_24px.svg": "cancel_24px.svg",
-            "ic_arrow_downward_24px.svg": "arrow_downward_24px.svg",
-            "ic_arrow_upward_24px.svg": "arrow_upward_24px.svg",
-            "ic_get_app_24px.svg": "get_app_24px.svg",
+            "http://mirror.tensorflow.org/raw.githubusercontent.com/google/material-design-icons/b3f05bfbf4329a5b63f50a720f867c2bac163f98/src/content/push_pin/materialicons/24px.svg": "keep_24px.svg",
+            "https://raw.githubusercontent.com/google/material-design-icons/b3f05bfbf4329a5b63f50a720f867c2bac163f98/src/content/push_pin/materialicons/24px.svg": "keep_24px.svg",
+            "http://mirror.tensorflow.org/raw.githubusercontent.com/google/material-design-icons/b3f05bfbf4329a5b63f50a720f867c2bac163f98/src/content/push_pin/materialiconsoutlined/24px.svg": "keep_outline_24px.svg",
+            "https://raw.githubusercontent.com/google/material-design-icons/b3f05bfbf4329a5b63f50a720f867c2bac163f98/src/content/push_pin/materialiconsoutlined/24px.svg": "keep_outline_24px.svg",
```
psybuzz, line 561: The rename attribute map's keys must either be the "after the last slash piece", or the full URL.

https://github.com/bazelbuild/rules_closure/blob/master/closure/filegroup_external.bzl#L167-L170

tensorboard/webapp/metrics/views/card_renderer/card_view_test.ts
```diff
@@ -93,7 +91,7 @@ describe('card view test', () => {
     fixture.componentInstance.pluginType = PluginType.SCALARS;
     fixture.detectChanges();
 
-    expect(fixture.debugElement.classes['full-width']).toBe(false);
+    expect(fixture.debugElement.classes['full-width']).not.toBeTruthy();
```
psybuzz, line 94: After CardViewContainer's `@HostBinding('class.full-width') showFullWidth: boolean = false` finishes change detection, we originally expected `debugElement.classes['full-width'] === false` internally.

In OSS, `debugElement.classes['full-width'] === undefined`, so we need to be a little more relaxed.
It's unclear what the exact discrepancy is, but most likely due to a different Angular version.

psybuzz: After some further hunting, this breaking change from Angular 9 (used in OSS) --> Angular 10 looks like a possible reason:

"DebugElement.classes returns undefined for classes that were added and then subsequently removed (previously, classes added and later removed would have a value of false)."
https://angular.io/guide/ivy-compatibility#less-common-changes

stephanwlee (approved):

psybuzz:
Thanks for being patient with all these PRs!

PR #4184, opened 2020-09-21
Turn TBServerDataSource errors into application state

bmd3k:

We want to improve error handling for cases where some top-level environment information cannot be loaded. This translates HTTP errors into application state.

Handle errors in requests to `data/environment`, `data/plugins_listing`, and `data/runs`. Transform HTTP error codes into an enum value for a new `failedCode` field in `pluginsListingLoaded` application state.


tensorboard/webapp/types/data.ts
```diff
@@ -22,8 +22,17 @@ export enum DataLoadState {
   FAILED,
 }
 
+export enum LoadFailureCode {
+  UNKNOWN = 'UNKNOWN',
```
stephanwlee, line 26: Just wondering, are you using the value of these enum in any meaningful way? If not, omit it so they remain as number for smaller bundle.

tensorboard/webapp/types/data.ts
```diff
@@ -22,8 +22,17 @@ export enum DataLoadState {
...
+  NOT_FOUND = 'NOT_FOUND',
+}
+
 export interface LoadState {
   state: DataLoadState;
-  // Time since epoch.
+  // Time of last successful load. Time since epoch.
   lastLoadedTimeInMs: number | null;
+  // Reason for failure of most recently completed request. This should not be
+  // set if there has not been a failure or if the most recently completed
```
stephanwlee, line 35: > This should not be set if there has not been a failure

Can you guarantee this with typing instead of comment?

stephanwlee (approved):

tensorboard/webapp/types/data.ts
bmd3k: I can guarantee for state === NOT_LOADED, LOADED, FAILED but there is still some ambiguity for state === LOADING.

I also made slight adjustments for typing of lastLoadedTimeInMs.

I feel like I'll have a bad time trying to get this into the internal repo - will start looking at that now.

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/store/debugger_reducers.ts
```diff
@@ -400,15 +400,18 @@ const reducer = createReducer(
...
         numExecutions > state.executions.executionDigestsLoaded.numExecutions;
-      const newState = {
+      return {
         ...state,
         lastNonEmptyPollDataTimeMs: numExecutionsIncreased
           ? Date.now()
           : state.lastNonEmptyPollDataTimeMs,
         executions: {
           ...state.executions,
+          focusIndex:
```
bmd3k, line 410: Note that there were some tricky situations for convincing the compiler that I'm properly using the types. This one took me the longest to figure out.

bmd3k:
After offline discussion we decided to limit this change to only the pluginsListLoaded state and not other similar states.  My git foo is not quite strong enough to have put that work into this PR so I created a new one (https://github.com/tensorflow/tensorboard/pull/4186) and will abandon this one.

Issue #4175, opened 2020-09-17
chore: bump chromium version, used by karma test runner


psybuzz:
Running `bazel test tensorboard/...` invokes web tests run by Karma.
The Karma launcher is setup to use a version of Chromium that is
defined in `third_party/workspace.bzl`. Currently, it's using old
snapshots of Chromium 68, which lacks support for modern language
features like `Array.prototype.flat`.

This issue is to track the chore of upgrading the dependency onto
a newer Chromium snapshot.

PR #4185, opened 2020-09-21
chore: bump Chromium version used in functional tests

psybuzz:
The functional tests rely on downloading and launching a
Chromium browser snapshot, currently pinned to version
68.0.3440.0 in the `org_chromium_chromium` target in
third_party/workspace.bzl.

Chromium 68 lacks some new modern features, such as
Array.prototype.flat(), so this change bumps the browser
snapshot version to a more recent 84.0.4147.0.

Tested on Mac via
`bazel test //tensorboard/webapp:karma_test_chromium-local`

For Chromium version information, see
https://omahaproxy.appspot.com/
https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html

Fixes https://github.com/tensorflow/tensorboard/issues/4175

### DIFF
--- a/third_party/workspace.bzl
+++ b/third_party/workspace.bzl
@@ -47,22 +47,22 @@ def tensorboard_workspace(name = ""):
     )
 
     platform_http_file(
-        name = "org_chromium_chromium",
+        name = "org_chromium_chromium",  # pinned to Chromium 84.0.4147.0
         licenses = ["notice"],  # BSD 3-clause (maybe more?)
         amd64_sha256 =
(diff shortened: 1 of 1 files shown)


psybuzz:
For the reviewer, or anyone interested, [Omahaproxy](https://omahaproxy.appspot.com/) can be used
to find the 6-digit commit # for a particular Chromium version
like so:
image

To verify that a 6-digit commit is part of a particular Chromium
version, it can be looked up like so:
image
image

bmd3k (approved):

PR #4186, opened 2020-09-23
Turn TBServerDataSource errors into application state

bmd3k:
* Motivation for features / changes

We want to improve error handling for cases where some top-level environment information cannot be loaded. This PR translates HTTP request errors into application state.

* Technical description of changes

(This is a new version of https://github.com/tensorflow/tensorboard/pull/4184 that limits the changes to just the `pluginstListLoaded` state field)

Handle errors in requests to `data/environment`, `data/plugins_listing`, and `data/runs`. The state for these requests is currently kept in `pluginsListLoaded` field. Change `pluginsListLoaded` to have custom type that extends `LoadState` with a `failureCode` field. `failureCode` initially has values `NOT_FOUND` or `UNKNOWN`.

PR #4192, opened 2020-09-24
Render PluginLoadState errors in the UI and allow parent components to override error UI

bmd3k:

  We want to render more helpful error messages for cases where the TensorBoard server is not responsive or has an error. We want to allow other flavors of TensorBoard to customize these errors.

  In https://github.com/tensorflow/tensorboard/pull/4186 we improved error handling for `data/runs`, `data/plugins_listing` and `data/environment` HTTP requests. Errors from those requests are translated into a `failureCode` field in the `pluginsListLoad` state.

  In this PR we render those errors in the `plugins` component, given that we want to render the errors in a similar way to the errors the component already renders.

  We also provide a mechanism for parent component to specify custom error templates. This will allow other flavors of TensorBoard to customize the error messages.

* Screenshots of UI changes

  image

### DIFF
--- a/tensorboard/webapp/plugins/plugins_container_test.ts
+++ b/tensorboard/webapp/plugins/plugins_container_test.ts
@@ -12,6 +12,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+import {Component} from '@angular/core';
 import {TestBed} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
 import {Store} from '@ngrx/store';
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/plugins/plugins_container_test.ts
```diff
@@ -467,6 +508,54 @@ describe('plugins_component', () => {
...
+        fixture.detectChanges();
+
+        expect(
+          fixture.debugElement.query(By.css('.environment-not-loaded'))
+        ).toBeNull();
+        expect(
+          fixture.debugElement.query(By.css('.custom-not-found-template'))
+        ).toBeNull();
+        expect(
+          fixture.debugElement.query(By.css('.custom-unknown-template'))
```
stephanwlee, line 554: can we do a basic content sanity check instead? If we rely on css, the test behavior may change when I add a CSS/class line to the template.

stephanwlee (approved):

tensorboard/webapp/plugins/plugins_container_test.ts
bmd3k: Done.

PR #4193, opened 2020-09-24
uploader: accept internal alternate experiment name flag

davidsoergel:
Work around argparse bug for internal users


wchargin (approved):
Okay: confirmed that `tensorboard dev upload -h` doesn’t show the thing,
and that both the public and internal arguments work (and, for what it’s
worth, whichever one is specified second takes precedence).

PR #4194, opened 2020-09-25
uploader: allow reporting the TB.dev URL via callback

davidsoergel:


tensorboard/uploader/uploader_subcommand.py
```diff
@@ -73,15 +73,15 @@ def _prompt_for_user_ack(intent):
     sys.stderr.write("\n")
 
 
-def _run(flags):
+def _run(flags, experiment_url_callback=None):
```
wchargin, line 76: Docstring has `Args:`; please update it.

tensorboard/uploader/uploader_subcommand.py
```diff
@@ -508,7 +512,7 @@ def execute(self, server_info, channel):
         )
 
 
-def _get_intent(flags):
+def _get_intent(flags, experiment_url_callback=None):
```
wchargin, line 515: Ditto.

tensorboard/uploader/uploader_subcommand.py
```diff
@@ -429,6 +431,8 @@ def execute(self, server_info, channel):
         )
         experiment_id = uploader.create_experiment()
         url = server_info_lib.experiment_url(server_info, experiment_id)
+        if self.experiment_url_callback:
```
wchargin, line 434: Should be `is not None`. (In principle, a user-supplied callable could
be falsy, and what you’re trying to check is whether one was provided,
not how truthy it is.)

wchargin (approved):
nit: Style for `area: description` PR/commit headlines is that both be
lowercase. See commits in torvalds/linux, git/git, golang/go,
tensorflow/tensorboard, ….

Approved modulo above and inlines.

tensorboard/uploader/uploader_subcommand.py
davidsoergel: Done

tensorboard/uploader/uploader_subcommand.py
davidsoergel: Done.

tensorboard/uploader/uploader_subcommand.py
davidsoergel: A falsy callable!?  Subtle, but ok :)

wchargin: Look no further than thine own backyard:

```python
>>> from absl import flags
>>> fv = flags.FlagValues()
>>> callable(fv) and not fv
True
```

Another, found with a bit of REPL-hacking:

```python
>>> from pygments import style
>>> style.Style._styles.clear()
>>> callable(style.Style) and not style.Style
True
```

And `unittest.mock.call` almost fits the bill: it’s a callable subclass
of `tuple`, so could in principle be falsy, but it always happens to be
non-empty.

If Python makes it possible, someone will have done it! The examples
seem far-fetched now, but if an `is not None` could prevent a debugging
nightmare then it seems justified to me—especially since it’s what you
really mean, anyway.

Issue #4169, opened 2020-09-15
Using TensorBoard in Notebooks fails on tensorflow:2.3.0-jupyter


bhlarson:
Running the "https://www.tensorflow.org/tensorboard/tensorboard_in_notebooks" example using the jupyter server.  It fails to load the tensorboard into the notebook
$ docker run -it -p 8888:8888 -p 6006:6006 tensorflow/tensorflow:2.3.0-jupyter 
"https://www.tensorflow.org/tensorboard/tensorboard_in_notebooks" succeeded in:
$ docker run -it -p 8888:8888 -p 6006:6006 tensorflow/tensorflow:1.15.2-py3-jupyter

This issue has been reported in the past but has not been addressed.

rmothukuru:
@bhlarson,
Can you please refer [Answer 1](https://stackoverflow.com/questions/41523005/how-to-use-tensorboard-in-a-docker-container-on-windows) and [Answer 2](https://stackoverflow.com/questions/33836728/view-tensorboard-on-docker-on-google-cloud/33838211#33838211) and let us know if it helps. Thanks!

bhlarson:
Thank you for your feedback.  I have worked through Answer 1 and Answer 2 and there is no change in behavior.  

- In my tests, the docker image is Ubuntu 18.04 run and tested on an Ubuntu 18.04 computer so the issue is not with windows.
- With tensorflow/tensorflow:1.15.2-py3-jupyter, "curl http://localhost:6006/" retrieves tensorboard from the host os.  Curl is missing from the docker image.  Running curl http://localhost:6006/ from the host os when launched with both of the following is successful: 
docker run -it -p 8887:8888 -p 0.0.0.0:6006:6006 tensorflow/tensorflow:1.15.2-py3-jupyter or 
docker run -it -p 8887:8888 -p 6006:6006 tensorflow/tensorflow:1.15.2-py3-jupyter

- With docker run -it -p 8888:8888 -p 0.0.0.0:6006:6006 tensorflow/tensorflow:2.3.0-jupyter,  curl http://localhost:6006/ returns the tensorboard page from within the container but: "curl: (56) Recv failure: Connection reset by peer" from the host.

The tensorboad sever is certainly running and talking on port 6006.  Although the port is available and mapped, something in the  tensorflow/tensorflow:2.3.0-jupyter container is blocking the communication and preventing Tensorboard from being displayed in Jupyter.

manivaradarajan:
[...]

PR #4195, opened 2020-09-28
docs: add bind_all to allow docker access

danielsiwiec:
This addresses #4169

This allows the local browser to access tensorboard running inside docker, as the guide does. Without this, tensorboard UI will not be accessible from outside the container

### DIFF
--- a/docs/tensorboard_in_notebooks.ipynb
+++ b/docs/tensorboard_in_notebooks.ipynb
@@ -95,18 +95,14 @@
         "id": "9w7Baxc8aCtJ"
       },
       "source": [
-        "**For Docker users:** In case you are running a [Docker](https://docs.docker.com/install/) image of [Jupyter Notebook server using TensorFlow's nightly](https://www.tensorflow.org/install/docker#examples_using_cpu-only_images), it is necessary to expose not only the notebook's port, but the TensorBoard's port.\n",
-        "\n",
-        "Thus, run the container with the following command:\n",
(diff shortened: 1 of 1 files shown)


wchargin (commented):
Hi @danielsiwiec! Thanks for the contribution. Can confirm that this is
fix is necessary and sufficient for use in Docker. That said, for
everyone else, this isn’t necessary, so it’s at best superfluous and at
worst an unintended privilege expansion, since the `--bind_all` will
cause their TensorBoard to be exposed to the local network.

Could we perhaps instead update the paragraph at the top of the Setup
section? Something like

> **For Docker users:** In case you are running […] TensorBoard's port.
> Also, pass `--bind_all` to `%tensorboard` to expose the port outside
> the container.

or similar. I’d also be okay with an inline note next to the paragraphs
that you’ve modified, if you prefer that, as long as the experience for
non-Docker users isn’t too cluttered.

danielsiwiec:
@wchargin - you're right. the fix only applies to Docker. I like your idea - let me update the PR to have a paragraph at the top instead

danielsiwiec:
@wchargin done

wchargin:
@googlebot I consent.

wchargin (approved):
@danielsiwiec: Great! I’ve removed a couple line breaks to make it
clearer where the “For Docker users” comment begins and ends, but this
looks good. Will merge once CI passes. Thank you!

danielsiwiec:
@googlebot I consent.

PR #4197, opened 2020-09-28
Fix URL hash loss upon initial navigation

psybuzz:
Enabling the upcoming Time Series dashboard [1] requires the
AppRoutingModule to be in effect with routes. Internal tests
caught a race condition, in which "setting the plugin hash",
e.g. `/#scalars` was done before the AppRouting effects.
AppRouting then replaced the entire URL hash on initial
navigation, losing important state.

This change adds an exception in AppRouting to allow the
initial navigation to preserve the hash. Non-initial navigations
can still safely discard, as we expect DeepLinkProviders to
manage things.

An alternative considered was to make HashStorageContainer
re-set the hash upon navigations. This also works, but the
drawback is that it introduces a dependency from Core on
AppRouting, and having the HashStorageContainer know about
routing side effects is non-ideal.

Tested both manually and via bazel commands, also internally
with --runs_per_test=20 and on TB embedders.

Internally, a test catching this race condition is sensitive to
execution options. Running
`bazel test ... --test_output=streamed` may pass, while
`bazel test ...` may fail.

Googlers, see test sync cl/334266882

[1] https://github.com/tensorflow/tensorboard/pull/4189

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -228,21 +228,39 @@ export class AppRoutingEffects {
         ofType(navigated),
         withLatestFrom(this.store.select(getActiveRoute)),
         filter(([, route]) => Boolean(route)),
-        map(([, route]) => route!),
-        filter((route) => {
+        map(([navigatedAction, route]) => {
+          return {
+            isFirstNavigation: navigatedAction.before === null,
+            route: route!,
+          };
+        }),
+        filter(({route}) => {
           return !areRoutesEqual(route, {
             pathname: this.location.getPath(),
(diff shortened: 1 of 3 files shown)


stephanwlee:
> Internal tests
caught a race condition, in which "setting the plugin hash",
e.g. /#scalars was done before the AppRouting effects.
AppRouting then replaced the entire URL hash on initial
navigation, losing important state.

To expound on above, when both AppRoutingEffects and CoreEffects initialize, they make some changes asynchronously; former makes modification to URL (with two debounce/nextTick) and latter makes XHR to fetch plugins listing. Nothing guarantees that the XHR to land after the URL is modified and the test did reveal the cases when the XHR finished before router getting initialized.

stephanwlee (approved):

PR #4199, opened 2020-09-29
uploader: move UploadIntent tests to uploader_subcommand_test.py

davidsoergel:
I thought UploadIntent was untested before because I didn't notice it at the bottom of `uploader_test.py` (and apparently neglected to search for it).

### DIFF
--- a/tensorboard/uploader/BUILD
+++ b/tensorboard/uploader/BUILD
@@ -84,6 +84,18 @@ py_library(
     ],
 )
 
+py_test(
+    name = "uploader_subcommand_test",
+    srcs = ["uploader_subcommand_test.py"],
+    srcs_version = "PY3",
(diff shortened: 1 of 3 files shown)


tensorboard/uploader/BUILD
```diff
@@ -145,6 +144,18 @@ py_test(
     ],
 )
 
+py_test(
+    name = "uploader_subcommand_test",
```
wchargin, line 148: nit: Could you move this build rule to just under `uploader_subcommand`, please?

wchargin (approved):

tensorboard/uploader/BUILD
davidsoergel: Oops, thanks :)

Issue #49, opened 2017-06-16
tensorboard not able to read large event file (~600MB)


chihuahua:
This issue had been migrated from tensorflow/tensorflow#836.

> I am training a network on 4 GPUs. The event file is too large to be parsed by tensorboard. How can I increase the limit?
> 
> RNING:tensorflow:IOError [Errno 2] No such file or directory: '/home/pc/anaconda2/lib/python2.7/site-packages/tensorflow/tensorboard/TAG' on path /home/pc/anaconda2/lib/python2.7/site-packages/tensorflow/tensorboard/TAG
> WARNING:tensorflow:Unable to read TensorBoard tag
> Starting TensorBoard  on port 6006
> (You can navigate to http://0.0.0.0:6006)
> [libprotobuf ERROR google/protobuf/src/google/protobuf/io/coded_stream.cc:207] A protocol message was rejected because it was too big (more than 67108864 bytes).  To increase the limit (or to disable these warnings), see CodedInputStream::SetTotalBytesLimit() in google/protobuf/io/coded_stream.h.
> Exception in thread Thread-2:
> Traceback (most recent call last):
>   File "/home/pc/anaconda2/lib/python2.7/threading.py", line 801, in __bootstrap_inner
>     self.run()
>   File "/home/pc/anaconda2/lib/python2.7/threading.py", line 1073, in run
>     self.function(*self.args, **self.kwargs)
>   File "/home/pc/anaconda2/lib/python2.7/site-packages/tensorflow/python/summary/event_accumulator.py", line 242, in Update
>     self.Reload()
>   File "/home/pc/anaconda2/lib/python2.7/site-packages/tensorflow/python/summary/event_accumulator.py", line 175, in Reload
>     for event in self._generator.Load():
>   File "/home/pc/anaconda2/lib/python2.7/site-packages/tensorflow/python/summary/impl/directory_watcher.py", line 82, in Load 
>     for event in self._loader.Load():
>   File "/home/pc/anaconda2/lib/python2.7/site-packages/tensorflow/python/summary/impl/event_file_loader.py", line 53, in Load 
>     event.ParseFromString(self._reader.record())
> DecodeError: Error parsing message

dandelionmane:
@digitalsword Could you give us a link to the 600mb event file, or post a script that will generate a repro?

Sakib-Ar-Rahman:
[...]

PR #4200, opened 2020-09-29
uploader: add test for experiment_url_callback

davidsoergel:
Add test neglected from #4194

### DIFF
--- a/tensorboard/uploader/uploader_subcommand_test.py
+++ b/tensorboard/uploader/uploader_subcommand_test.py
@@ -116,6 +116,41 @@ class UploadIntentTest(tf.test.TestCase):
             mock_stdout_write.call_args_list[-1][0][0], "\nDone.\n"
         )
 
+    def testUploadIntentWithExperimentUrlCallback(self):
+        """Test the upload intent with a callback."""
+        mock_server_info = mock.MagicMock()
+        mock_channel = mock.MagicMock()
+        upload_limits = server_info_pb2.UploadLimits(
+            max_scalar_request_size=128000,
+            max_tensor_request_size=128000,
+            max_tensor_point_size=11111,
+            max_blob_request_size=128000,
+            max_blob_size=128000,
+        )
+        tb_dev_url = "https://tensorboard.dev/experiment/test_experiment_id"
+        mock_experiment_url_callback = mock.MagicMock()
+        mock_stdout_write = mock.MagicMock()
+        with mock.patch.object(
+            server_info_lib,
+            "allowed_plugins",
+            return_value=_SCALARS_HISTOGRAMS_AND_GRAPHS,
+        ), mock.patch.object(
+            server_info_lib, "upload_limits", return_value=upload_limits
+        ), mock.patch.object(
+            server_info_lib, "experiment_url", return_value=tb_dev_url
+        ), mock.patch.object(
+            sys.stdout, "write", mock_stdout_write
+        ), mock.patch.object(
+            dry_run_stubs,
+            "DryRunTensorBoardWriterStub",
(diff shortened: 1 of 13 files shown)


tensorboard/uploader/uploader_subcommand_test.py
```diff
@@ -115,6 +115,44 @@ def testUploadIntentUnderDryRunOneShot(self):
...
+            max_scalar_request_size=128000,
+            max_tensor_request_size=128000,
+            max_tensor_point_size=11111,
+            max_blob_request_size=128000,
+            max_blob_size=128000,
+        )
+        tb_dev_url = "https://tensorboard.dev/experiment/test_experiment_id"
+        mock_experiment_url_callback = mock.MagicMock()
+        mock_stdout_write = mock.MagicMock()
+        with mock.patch.object(
```
wchargin, line 132: So much stuff is being mocked out here that I have no idea what code
paths under test are actually being tested.

Can we just use real values instead? `mock_server_info` shouldn’t be a
mock; it should be a `ServerInfoResponse` proto value. Ditto for
`upload_limits`, but you shouldn’t need that at all. Patching the
library functions in `server_info_lib` seems crazy to me—that’s real
code that we want to test, and it doesn’t even have side effects. Also,
please use `Mock` instead of `MagicMock` unless you specifically need
the magic methods, which you don’t here. (See internal “Python Tip of
the Week #049: Mocking pitfalls”.)

How about something like (sketch; adjust to taste):

```python
    def testUploadIntentWithExperimentUrlCallback(self):
        """Test the upload intent with a callback."""
        server_info = server_info_pb2.ServerInfoResponse()
        server_info.url_format.template = "https://tensorboard.dev/x/{}"
        server_info.url_format.id_placeholder = "{}"

        stub = dry_run_stubs.DryRunTensorBoardWriterStub()
        stub.CreateExperiment = lambda req, **__: write_service_pb2.CreateExperimentResponse(
            experiment_id="test_experiment_id", url="this URL is ignored"
        )

        expected_url = "https://tensorboard.dev/x/test_experiment_id"

        with mock.patch.object(
            dry_run_stubs, "DryRunTensorBoardWriterStub", wraps=lambda: stub,
        ), mock.patch.object(sys.stdout, "write"):
            mock_channel = mock.Mock()
            mock_experiment_url_callback = mock.Mock()
            intent = uploader_subcommand.UploadIntent(
                self.get_temp_dir(),
                dry_run=True,
                one_shot=True,
                experiment_url_callback=mock_experiment_url_callback,
            )
            intent.execute(server_info, mock_channel)
        mock_experiment_url_callback.assert_called_once_with(expected_url)
```

davidsoergel: Yeah happy to do all that, but I don't have bandwidth to similarly clean up all the other tests here from which I pasted the structure.  Local consistency is also a virtue, right?

wchargin: Sure, understood that it’s copy-pasted, and not asking you to clean up
the others. Local consistency is a virtue *ceteris paribus*, but in this
case (imho) the ceteris seem rather far from paribus, and so the
aphorism that springs more readily to my mind is to not let the perfect
be the enemy of the good. But I do welcome your thoughts—let me know if
you disagree with the direction that I’m suggesting.

davidsoergel: > the ceteris seem rather far from paribus

*Chef's kiss*

Since I like to paste things, I pasted your sketch.  Worked first time and looks correct to me.  So, thanks for ghostwriting this whole PR!

wchargin (approved):
Well, approved; I guess it really needs _your_ approval. :-)

davidsoergel:
LGTM

PR #4205, opened 2020-09-30
Refactor import of components necessary for Polymer <-> Angular interop.

bmd3k:

  We want to make it easier for other flavors of TensorBoard to identify and import polymer components that are necessary to run the TensorBoard UI.

  Move imports of polymer components necessary for Polymer <-> Angular interop into their own ts file, separate from the imports of dashboards.

  Ran a TensorBoard to ensure it works.
  Imported into an internal repo to ensure I can reuse the definitions.


tensorboard/components/polymer3_lib.ts
```diff
@@ -27,10 +27,4 @@ import '../plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-red
...
 import '../plugins/text/tf_text_dashboard/tf-text-dashboard';
 
-// Exported Polymer <-> Angular interop (to be removed).
-import './experimental/plugin_util/plugin-host';
-import './tf_backend/tf-backend-polymer';
-import './tf_globals/globals-polymer';
-import './tf_storage/tf-storage-polymer';
-import './tf_paginated_view/tf-paginated-view-store';
-import './tf_color_scale/tf-color-scale-polymer';
+import './polymer3_core_components';
```
stephanwlee, line 30: Optional: alternative name can be `polymer_interop_helper`

stephanwlee (approved):
Thanks!

tensorboard/components/polymer3_lib.ts
bmd3k: Done. Also updated documentation for the new file.

PR #4206, opened 2020-09-30
Get label colour from raw confusion matrix

gonnet:
**Motivation for features / changes**

This code was choosing the label colour based on the row-normalized confusion matrix values, but the background colour of each square is chosen from the un-normalized confusion matrix.

This causes problems for under-represented labels where all the squares in one row are light, but the label colour switches to white if any entry in the row has more than 0.5 relative to the row.

**Technical description of changes**

Use the un-normalized confusion matrix when deciding which colour to give a label.

**Screenshots of UI changes**

Before (left) and after (right) the change, using data from a different run than in the guide. Note how the text of the leftmost square in the middle row is now visible.

image


gonnet:
@nfelt Can you take a quick look at this? I was not able to assign a reviewer myself... Thanks!

nfelt (approved):
Thanks for the fix!

PR #4211, opened 2020-10-01
Defining a new build target for tensor.proto and its deps.

yfang1:
Changed suggested by @nfelt 

* Screenshots of UI changes

### DIFF
--- a/tensorboard/compat/proto/BUILD
+++ b/tensorboard/compat/proto/BUILD
@@ -26,24 +26,31 @@ tb_proto_library(
         "meta_graph.proto",
         "node_def.proto",
         "op_def.proto",
-        "resource_handle.proto",
         "rewriter_config.proto",
         "saved_object_graph.proto",
         "saver.proto",
         "step_stats.proto",
         "struct.proto",
         "summary.proto",
-        "tensor.proto",
         "tensor_description.proto",
-        "tensor_shape.proto",
         "tfprof_log.proto",
         "trackable_object_graph.proto",
-        "types.proto",
         "variable.proto",
         "verifier_config.proto",
         "versions.proto",
     ],
     visibility = ["//visibility:public"],
+    deps = [":tensor_proto_lib"],
+)
+
+tb_proto_library(
+    name = "tensor_proto_lib",
+    srcs = [
+        "resource_handle.proto",
+        "tensor.proto",
+        "tensor_shape.proto",
+        "types.proto",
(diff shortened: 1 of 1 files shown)


tensorboard/compat/proto/BUILD
```diff
@@ -26,24 +26,32 @@ tb_proto_library(
...
+
+tb_proto_library(
+    name = "tensor_proto_lib",
+    srcs = [
+        "resource_handle.proto",
+        "tensor.proto",
+        "tensor_shape.proto",
+        "types.proto",
+    ],
+    visibility = ["//visibility:public"],
```
nfelt, line 54: Let's leave this as "tensorboard internal" visibility (the default, i.e. drop this line) since as a new target it shouldn't have any dependencies, and the internal change will add the necessary narrow visibility.

tensorboard/compat/proto/BUILD
```diff
@@ -26,24 +26,32 @@ tb_proto_library(
...
-        "tensor_shape.proto",
         "tfprof_log.proto",
         "trackable_object_graph.proto",
-        "types.proto",
         "variable.proto",
         "verifier_config.proto",
         "versions.proto",
     ],
     visibility = ["//visibility:public"],
+    deps = [":tensor_proto_lib"],
```
nfelt, line 43: Rather than using deps here, let's just leave the files in this rule in addition to the tensor_proto_lib rule - it should be okay for them to be included in more than one tb_proto_library.  Otherwise it's kind of odd that protos_all needs to depend on `tensor_proto_lib` even for things that it uses directly, not just via TensorProto.

yfang1: Yeah that's what I was doing initially, but the Travis CI didn't like it, saying resource_handle.py is already generated by "protos_all".

tensorboard/compat/proto/BUILD
yfang1: done.

tensorboard/compat/proto/BUILD
yfang1: ... to handle the oddity properly, it would seem like we need to tear apart protos_all, define separate proto libs, and build the dependency graph.

nfelt: Ah my bad, you're right.  Never mind on this one then, sorry.

nfelt (approved):

Issue #4207, opened 2020-09-30
AppRoutingEffects' async initialization leads to race conditions, clearing URL params


psybuzz:
In https://github.com/tensorflow/tensorboard/pull/4197, we discovered a race condition and fixed it
in a non-ideal manner.

Background:

> when both AppRoutingEffects and CoreEffects initialize, they make some changes asynchronously; former makes modification to URL (with two debounce/nextTick) and latter makes XHR to fetch plugins listing. Nothing guarantees that the XHR to land after the URL is modified and the test did reveal the cases when the XHR finished before router getting initialized.

from https://github.com/tensorflow/tensorboard/pull/4197#issuecomment-701690582

The symptom is that after opening TensorBoard, the '#scalars' hash may get dropped
from the URL immediately.  Any actions that update the hash with a key=value, e.g.
changing smoothing (adds &smoothingWeight=... to URL) causes TensorBoard to
show "No active plugin".

The PR proposes a fix where AppRouting preserves any URL hash during its initial
'navigated' action.  However, it would be more ideal if AppRouting itself could
manage the URL hash entirely.

Alternatively, HashStorageContainer could be modified to wait for AppRouting
to initialize, before HashStorageContainer sets '#scalars' in the URL hash.
However, for this to land, it requires all TB embedders to use AppRouting.

Googlers, see mirrored issue b/169799696

psybuzz:
It turns out that this bug also affects URL queryParams, not just the hash.  AppRouting assumes it manages the URL queryParams entirely and can safely discard existing params because it is the source of truth.

Discarding initial queryParams turns out to break things like `?tensorboardColab=`, which are read directly from `window.location.search` at some arbitrary time during the TB session (see `tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts`).  This leads to TB thinking that the `?tensorboardColab=` param was not present, causing bad behavior in Colab.

There are a handful of other instances where we read directly from `window.location.search` at a time that is possibly after the AppRoutingEffects initialization.

stephanwlee:
I don't want to say our currently approach generally is a bad one; I think this is an unfortunate byproduct of our usage of query parameter in unorganized way. 

Of course in the long run, we can do what we do with the activePlugin and bootstrap the redux state correctly based on the initial query parameter but until all of our offering uses the router and redux state, it will not happen. 

For now, I think we need to find a workaround one of which can be populating some environment variable on the global scope (window) in the deep link state.

[...]

PR #4213, opened 2020-10-01
bug: fix AppRouting interference with URL queryParam-reading components

psybuzz:
See https://github.com/tensorflow/tensorboard/issues/4207 for details.

Several callsites in TB's Angular and Polymer codebases read
directly from `window.location.search`. For example, scalar cards
read the `?tensorboardColab=true` param at arbitrary times in
the TB session, to know whether TB is running inside Colab.

AppRoutingEffects has an initialization step that clears/discards
all existing URL queryParams, causing breaking behavior.

Since AppRouting is not yet a complete source of truth for URL
query params, this change makes all callsites that read from
`window.location.search` do so on the initial app load, before
AppRouting can interfere.

An alternative considered was making AppRoutingEffects aware
of a hardcoded list of reserved queryParam keys, and preserving
them whenever updating the URL. In that approach, there are
some unknown questions, e.g. if a user navigates to a different
route and returns to the original route, do we still need to preserve
queryParams?

Manually tested that running TB with `?tensorboardColab=true`
and opening scalar cards results in network requests to
the `/scalars` endpoint, not `/scalars_multirun.`

### DIFF
--- a/tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
+++ b/tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
@@ -16,9 +16,16 @@ import {Injectable} from '@angular/core';
 
 import {TBFeatureFlagDataSource} from './tb_feature_flag_data_source_types';
 
+/**
+ * Save the initial URL query params, before the AppRoutingEffects initialize.
+ * Ideally, AppRouting would manage all URL params.
+ * https://github.com/tensorflow/tensorboard/issues/4207
+ */
+const initialURLSearchParams = new URLSearchParams(window.location.search);
+
 const util = {
   getParams() {
-    return new URLSearchParams(window.location.search);
+    return initialURLSearchParams;
   },
 };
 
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -16,9 +16,16 @@ import {Injectable} from '@angular/core';
 
 import {TBFeatureFlagDataSource} from './tb_feature_flag_data_source_types';
 
+/**
+ * Save the initial URL query params, before the AppRoutingEffects initialize.
+ * Ideally, AppRouting would manage all URL params.
+ * https://github.com/tensorflow/tensorboard/issues/4207
```
stephanwlee, line 22: You can drop this line. Even if router gains the capability, we want this component to read from the URL to bootstrap certain feature flags states.

stephanwlee (approved):

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
psybuzz: Done, thanks!

psybuzz:
As per offline discussion, in a followup CL, I will write a regression test that checks that our app doesn't emit POST requests under ?tensorboardColab=true.

PR #4218, opened 2020-10-05
Use GET instead of POST requests for hparams in colab environment.

bmd3k:

  Hparams dashboard does not work in Google-internal colab environments because they do not support POST requests. 

  Reuse mechanism introduced in #4191 that provides a 'tensorboardColab' query parameter in the URL to indicate a colab environment. If the value for the parameter is true then the Hparams dashboard will send GET requests. Otherwise it sends POST requests.

  There already existed code that purported to determine if the hparams dashboard was running in a colab environment but this code has not worked for some time, possibly since #2798. The code in this PR simply replaces that code. The plumbing to pipe this through other layers had already been implemented.

* Tests

  I loaded a local tensorboard with the query parameter 'tensorboardColab=true' and inspected the network tab to confirm that hparams-related calls were GET requests.

  I loaded the local tensorboard without 'tensorboardColab' in the URL and inspected the network tab to confirm that hparams-related calls were POST requests.

### DIFF
--- a/tensorboard/plugins/hparams/tf_hparams_dashboard/tf-hparams-dashboard.ts
+++ b/tensorboard/plugins/hparams/tf_hparams_dashboard/tf-hparams-dashboard.ts
@@ -20,6 +20,10 @@ import '../tf_hparams_main/tf-hparams-main';
 import * as tf_hparams_backend from '../tf_hparams_backend/tf-hparams-backend';
 import {LegacyElementMixin} from '../../../components/polymer/legacy_element_mixin';
 
+const inColab =
+  new URLSearchParams(window.location.search).get('tensorboardColab') ===
+  'true';
+
 const PLUGIN_NAME = 'hparams';
 @customElement('tf-hparams-dashboard')
 class TfHparamsDashboard extends LegacyElementMixin(PolymerElement) {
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/hparams/tf_hparams_dashboard/tf-hparams-dashboard.ts
```diff
@@ -20,6 +20,10 @@ import '../tf_hparams_main/tf-hparams-main';
 import * as tf_hparams_backend from '../tf_hparams_backend/tf-hparams-backend';
 import {LegacyElementMixin} from '../../../components/polymer/legacy_element_mixin';
 
+const inColab =
+  new URLSearchParams(window.location.search).get('tensorboardColab') ===
```
wchargin, line 24: Could you leave a comment here that this is deliberately a side-effect
at module import time to work around the bug in `AppRouting`, as #4213
did for the scalars dashboard?

https://github.com/tensorflow/tensorboard/blob/5c483ac16a69da89a72c4d8e79b1ef42f435301a/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts#L33-L36

wchargin (approved):

wchargin:
@googlebot I consent.

tensorboard/plugins/hparams/tf_hparams_dashboard/tf-hparams-dashboard.ts
wchargin: (I’ve gone ahead and done this.)

PR #4219, opened 2020-10-06
AppRouting should preserve hash on same-RouteId navigations

psybuzz:
Followup: https://github.com/tensorflow/tensorboard/pull/4220

AppRoutingEffects recently added logic to preserve URL hash
upon the initial 'navigated' action (see [1]). While this fixed cases
where the URL needs to be read upon initial page load, it does
not address an upcoming case needed for deeplinking pins in
Time Series.

Flow: pin a card, DeepLinkProvider serializes `?pinnedCards=`
in the URL which drops `#timeseries`, reload the tab, the
selected dashboard is no longer Time Series.
Expected: Time Series remains selected upon tab reload.

This change overcomes this issue the DeepLinkProvider by
preserving the hash upon same-RouteId navigations.

Manually checked internal TB embedders that navigating to
a different route id still clears the hash.

Googlers, see sync cl/335688801

[1] https://github.com/tensorflow/tensorboard/pull/4213

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -229,15 +229,14 @@ export class AppRoutingEffects {
         withLatestFrom(this.store.select(getActiveRoute)),
         filter(([, route]) => Boolean(route)),
         map(([navigatedAction, route]) => {
-          // DeepLinkProviders may tell AppRoutingEffects to modify URL query
-          // params upon redux state changes. These updates can discard URL
-          // hash storage, set by tf-storage on the Polymer side. Due to async
-          // operators, we cannot guarantee that the initial 'navigated' action
-          // will be processed before any other URL-modifying logic.
+          // The URL hash can be set via HashStorageComponent (which uses
+          // Polymer's tf-storage). DeepLinkProviders also modify the URL when
+          // a provider's serializeStateToQueryParams() emits. These result in
+          // the URL updated without the previous hash. HashStorageComponent
+          // makes no attempt to restore the hash, so it is dropped.
 
-          // When Polymer-set URL hashes are dropped, this results in bad
-          // behavior when refreshing (e.g. lost active plugin) and when
-          // changing dashboards (e.g. lost tagFilter).
+          // This results in bad behavior when refreshing (e.g. lost active
+          // plugin) and when changing dashboards (e.g. lost tagFilter).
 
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -229,8 +229,27 @@ export class AppRoutingEffects {
         withLatestFrom(this.store.select(getActiveRoute)),
         filter(([, route]) => Boolean(route)),
         map(([navigatedAction, route]) => {
+          // DeepLinkProviders may tell AppRoutingEffects to modify URL query
+          // params upon redux state changes. These updates can discard URL
+          // hash storage, set by tf-storage on the Polymer side. Due to async
+          // operators, we cannot guarantee that the initial 'navigated' action
+          // will be processed before any other URL-modifying logic.
```
stephanwlee, line 236: I don't think it is purely about async operators. 

For instance, if you are navigated to timeseries (`#timeseries` is written by HashStorage) and if you interact with the pinning, this effect will overwrite and remove `#timeseries`. In this case, HashStorage was not even attempt to write any hash so there is nothing timing related about this.

Could you more accurately depict the problem please? 

Also I'd appreciate `hash storage` -> `HashStorageComponent`

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -651,7 +651,42 @@ describe('app_routing_effects', () => {
       expect(replaceStateSpy).toHaveBeenCalledWith('/experiments#foo');
     });
 
-    it('does not preserve hash upon replace for non-initial navigation', () => {
+    it('preserves hash upon navigations to the same route id', () => {
```
stephanwlee, line 654: in this behavioral spec, maybe specify that this is a temporary behavior until all apps use the router?

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
psybuzz: Updated the comment and dropped the part about async.

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
psybuzz: Done.

stephanwlee (approved):
Thanks for updating the comments.

PR #4222, opened 2020-10-07
Create dedicated build rule for mat_icon_module.ts

bmd3k:

  We want to decouple mat_icon_module.ts from the ":app" build target.

  A couple internal build targets depend on the "//tensorboard/webapp:app" built target but all they want is mat_icon_module.ts. The ":app" build target and its dependencies pull in unnecessary code like the NPMI and DebuggerV2 plugins.

  Create a new build target ":mat_icon" whose src is only "mat_icon_module.ts".  Name ":mat_icon" as a dep of ":app".

  There will have to be a copybara config chase to properly import the new target into the internal repo.

  This change will have to be accompagnied by a pseudo-atomic commit in the internal repo to update the affected internal build targets. Will send that out after the copybara changes are submitted.

  * Run a local tensorboard from the git codebase and ensure it still runs.
  * Import the change into an internal repo client, modify affected build targets to use the new target, and make sure affected targets build.


bmd3k:
Darn:
ERROR: file 'tensorboard/webapp/mat_icon_module.ngfactory.mjs' is generated by these conflicting actions:
Label: //tensorboard/webapp:mat_icon, //tensorboard/webapp:app

I have to reevaluate my approach. A pseudo-atomic commit with the internal repo might be necessary.

bmd3k:
> Darn:
> ERROR: file 'tensorboard/webapp/mat_icon_module.ngfactory.mjs' is generated by these conflicting actions:
> Label: //tensorboard/webapp:mat_icon, //tensorboard/webapp:app
> 
> I have to reevaluate my approach. A pseudo-atomic commit with the internal repo might be necessary.

I have to do a pseudo-atomic commit with the internal repo. PR updated. Will send you the corresponding internal change after my copybara changes are submitted and I can properly generate the internal changes.

stephanwlee (approved):

PR #4224, opened 2020-10-08
Avoid POST requests in TBHttpClient when in Colab

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/4223

Time Series relies mainly on POST requests to the /timeSeries
endpoint. This change updates the Python plugin backend to
support GET requests.

In general, we'd like to fallback to GET instead of POST for any
frontend components that make network requests while in a
Colab environment, until internal Colab supports POST.

Manually checked via the DevTools network panel that opening
the Time Series dashboard makes GET requests when the URL
contains `?tensorboardColab=true`, and POST requests when
it does not.

Googlers, see test sync cl/336394977

### DIFF
--- a/tensorboard/webapp/webapp_data_source/tb_http_client.ts
+++ b/tensorboard/webapp/webapp_data_source/tb_http_client.ts
@@ -61,22 +61,22 @@ export class TBHttpClient implements TBHttpClientInterface {
 
   post<ResponseType>(
     path: string,
-    body: any,
+    body: FormData,
     options: PostOptions = {}
   ): Observable<ResponseType> {
-    // Google-internal Colab does not support HTTP POST requests, so we fall
-    // back to HTTP GET (even though public Colab supports POST)
-    // See b/72932164.
     return this.store.select(getIsFeatureFlagsLoaded).pipe(
       filter((isLoaded) => Boolean(isLoaded)),
       take(1),
       withLatestFrom(this.store.select(getIsInColab)),
       mergeMap(([, isInColab]) => {
+        // Google-internal Colab does not support HTTP POST requests, so we fall
+        // back to HTTP GET (even though public Colab supports POST)
+        // See b/72932164.
         if (isInColab) {
-          const optionsForGet = {headers: options.headers} as GetOptions;
-          optionsForGet.params =
-            body instanceof FormData ? convertFormDataToObject(body) : body;
-          return this.http.get<ResponseType>(path, optionsForGet);
+          return this.http.get<ResponseType>(path, {
+            headers: options.headers,
+            params: convertFormDataToObject(body),
+          });
         } else {
           return this.http.post<ResponseType>(path, body, options);
         }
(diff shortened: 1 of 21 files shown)


tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "experimentalDecorators": true,
     "importHelpers": true,
     "inlineSourceMap": true,
-    "lib": ["dom", "es2020"],
+    "lib": ["dom", "es2020", "dom.iterable"],
```
psybuzz, line 8: Note: this was required to support for doing `for ... of` over a FormData in Typescript

tensorboard/webapp/webapp_data_source/tb_http_client.ts
```diff
@@ -26,9 +37,20 @@ import {
...
+  }
+  return result;
+}
+
 @Injectable()
 export class TBHttpClient implements TBHttpClientInterface {
-  constructor(private http: HttpClient) {}
+  constructor(
+    private readonly http: HttpClient,
+    private readonly store: Store<State>
```
bmd3k, line 52: Should TBHttpClient be ngrx-aware? It seems to fall into the category of files that we want to keep as a pure service ignorant of ngrx.  I don't have a strong opinion about this myself but wonder if it breaks some of the team's best practices.

stephanwlee: Ideally, no. It invites many bad patterns so it is generally advised against but I would not mind this "hack" if it can encapsulate away the details of colabness (which we hope to soon get rid of) for now.

tensorboard/webapp/webapp_data_source/tb_http_client.ts
```diff
@@ -42,7 +64,24 @@ export class TBHttpClient implements TBHttpClientInterface {
     body: any,
     options: PostOptions = {}
   ): Observable<ResponseType> {
-    return this.http.post<ResponseType>(path, body, options);
+    // Google-internal Colab does not support HTTP POST requests, so we fall
```
bmd3k, line 67: I think this comment would be better placed immediately under the line if(isInColab).  It does not accurately describe the entire code statement that follows it.

bmd3k (approved):

tensorboard/webapp/webapp_data_source/tb_http_client.ts
```diff
@@ -42,7 +64,24 @@ export class TBHttpClient implements TBHttpClientInterface {
...
+    // See b/72932164.
+    return this.store.select(getIsFeatureFlagsLoaded).pipe(
+      filter((isLoaded) => Boolean(isLoaded)),
+      take(1),
+      withLatestFrom(this.store.select(getIsInColab)),
+      mergeMap(([, isInColab]) => {
+        if (isInColab) {
+          const optionsForGet = {headers: options.headers} as GetOptions;
+          optionsForGet.params =
+            body instanceof FormData ? convertFormDataToObject(body) : body;
```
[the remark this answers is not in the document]
psybuzz: Done.

tensorboard/webapp/webapp_data_source/tb_http_client.ts
psybuzz: Done.

psybuzz (commented):
Thanks for the reviews!

PR #4227, opened 2020-10-09
fix timeseries color picker positioning

stephanwlee:
Color picker has different modes: inline vs. modal. In our case, since
we use the color picker in the scrollable area of the runs table, we
need it to escape the scrollable container and absolutely position by
injecting component on the body. For that, in order for ngx-color-picker
to properly position without using Angular cdk component, it requires to
know the root component.

This change injects ViewContainerRef which is what ngx-color-picker
requires but details of how that lets descendent components know of the
root component escape us at the moment.


psybuzz (approved):
Thanks for fixing this

PR #4229, opened 2020-10-09
Add overridable sanitize() step to tf-markdown-view.

bmd3k:

  Several Google-internal instances of TensorBoard require an html sanitization step for user-provided markdown.

  Define a sanitize() function for usage by tf-markdown-view before it renders user-provided markdown. The default implementation returns the input html without modification.

  Internally at Google we wil override the implementation of this sanitize() operation.

  Googlers please see cl/336163444 for corresponding internal CL, which will have to be submitted before this PR is merged.

  Ran local tensorboard for OSS with text plugin and ensured user markdown is still rendered properly.

  Patched the changes from this PR into a Google-internal repository with the changes from cl/336163444 and ensured that various versions of TensorBoard both build properly and render user markdown properly.

### DIFF
--- a/tensorboard/components/tf_markdown_view/BUILD
+++ b/tensorboard/components/tf_markdown_view/BUILD
@@ -9,8 +9,15 @@ tf_ts_library(
     srcs = ["tf-markdown-view.ts"],
     strict_checks = False,
     deps = [
+        ":sanitize",
         "//tensorboard/components/polymer:legacy_element_mixin",
         "@npm//@polymer/decorators",
         "@npm//@polymer/polymer",
     ],
+)
+
+tf_ts_library(
+    name = "sanitize",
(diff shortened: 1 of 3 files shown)


tensorboard/components/tf_markdown_view/BUILD
```diff
@@ -9,8 +9,15 @@ tf_ts_library(
...
         "//tensorboard/components/polymer:legacy_element_mixin",
         "@npm//@polymer/decorators",
         "@npm//@polymer/polymer",
     ],
 )
+
+tf_ts_library(
+    name = "sanitize",
+    srcs = ["sanitize.ts"],
+    deps = [],
```
stephanwlee, line 22: remove the empty deps.

stephanwlee (approved):

PR #4231, opened 2020-10-13
line-chart: create reusable worker

stephanwlee:
This is one of the foundational library for the new upcoming feature:
Gpu based line chart. In the new line chart, we would like to leverage
OffscreenCanvas which allows worker thread to manipulate the canvas.

In doing so, we are creating a small library that maintains worker
instances so we don't spawn off unbounded number of workers.

#gpu_line_chart


tensorboard/webapp/widgets/line_chart_v2/lib/worker_allocator.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Returns a worker-like object that can be used to offload computation like Worker.
+ * This method allocates new Worker upto 10 instances. Depending on its usage (not by CPU
+ * utilization but merely number of instance holder), it allocates the freest one. Upon
+ * disposal (e.g., Angular's ngOnDestroy), please invoke `free` for recycle.
+ *
+ * @param workerResourcePath path to worker JavaScript resource.
+ */
+export function getWorkerInstance(workerResourcePath: string): WorkerLike {
```
wchargin, line 72: Could we maybe clarify that this is a URL to a JavaScript file that the
worker will load and execute? As written, it kind of sounds like just a
cache key to me.

(Assuming that I’m reading the MDN docs correctly.)

wchargin, line 72: FWIW: if we expect that most clients will only want to create workers
for a single resource (or small few resources, as opposed to a dynamic
set), which it seems to me should be the case, then an alternate way to
factor this could be to expose `WorkerPool` itself. That way, there’s no
global mutable state, and, since its constructor could also take an
optional `workerFactory` defaulting to `getWorker`, no need to mock.

Clients could still bind `new WorkerPool(...)`s to module scope if the
global mutable state is deemed convenient enough, but at least then the
pool is scoped to a module rather than just a resource path.

Would also remove the need for `TEST_ONLY.cleanup`: each test just
allocates its own pool.

tensorboard/webapp/widgets/line_chart_v2/lib/worker.ts
```diff
@@ -0,0 +1,23 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+/**
+ * Returns a Worker instancefor a given resource url.
```
wchargin, line 17: Maybe “Creates a new worker that loads and executes the JavaScript at
the given URL”? to clarify the semantics of the resource path and also
that this is creating something rather than “getting” it from a cache.

tensorboard/webapp/widgets/line_chart_v2/lib/worker_allocator.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    const shouldAllocateNew =
+      this.workers.every(({activeCount}) => activeCount > 0) &&
+      this.workers.length < MAX_WORKER_INSTANCE;
+
+    if (shouldAllocateNew) {
+      const worker = getWorker(this.workerResourcePath);
+      workerLike = {
+        activeCount: 0,
+        postMessage: (mesasge: any, transfer: Transferable[]) => {
```
wchargin, line 44: s/mesasge/message/

tensorboard/webapp/widgets/line_chart_v2/lib/worker_allocator.ts
```diff
@@ -0,0 +1,87 @@
...
+    return workerLike;
+  }
+}
+const workerAllocators = new Map<string, WorkerAllocator>();
+
+/**
+ * Returns a worker-like object that can be used to offload computation like Worker.
+ * This method allocates new Worker upto 10 instances. Depending on its usage (not by CPU
+ * utilization but merely number of instance holder), it allocates the freest one. Upon
+ * disposal (e.g., Angular's ngOnDestroy), please invoke `free` for recycle.
```
wchargin, line 68: What are the implications of failing to `free`? Looks to me like it only
affects resource usage/load balancing policies, not correctness? (If so,
that’s pleasant.)

What are the implications of a double-`free` or use-after-`free`? Since
workers are never destructed from the pool, it looks like it would also
only affect load balancing, unless clients have a hard-requirement on
the number of concurrent tasks for a single worker. Also a pleasant
surprise.

Up to you whether to guarantee and/or document these properties, but we
should be sure to understand them. Any corrections to my assessments?

tensorboard/webapp/widgets/line_chart_v2/lib/worker_allocator_test.ts
```diff
@@ -0,0 +1,102 @@
...
+      expect(getWorkerInstance('test_30')).toBe(workers[5]);
+      // 6th one is still freest (only one user).
+      expect(getWorkerInstance('test_30')).toBe(workers[5]);
+      // Now, there is both 3rd and 6th one. Because 3rd one comes first in index,
+      // it wins.
+      expect(getWorkerInstance('test_30')).toBe(workers[2]);
+      // Lastly, 6th one is allocated.
+      expect(getWorkerInstance('test_30')).toBe(workers[5]);
+      // Now that every one has 3 active clients, 1st one gets returned.
+      expect(getWorkerInstance('test_30')).toBe(workers[0]);
```
wchargin, line 99: Very nice; thanks.

tensorboard/webapp/widgets/line_chart_v2/lib/worker_allocator.ts
```diff
@@ -0,0 +1,87 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+import {getWorker} from './worker';
+
+const MAX_WORKER_INSTANCE = 10;
+
+export interface WorkerLike {
```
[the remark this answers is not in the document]
stephanwlee: Tried to address the concern by not calling it `Like` but `Proxy`. I am not sure if the new name is a great improvement but hope it does abate part of your concerns.

tensorboard/webapp/widgets/line_chart_v2/lib/worker_allocator.ts
stephanwlee: Not sure if I did the best job at describing the property but hope my refactored comment makes it clearer.

tensorboard/webapp/widgets/line_chart_v2/lib/worker_pool.ts
```diff
@@ -0,0 +1,98 @@
...
+ * https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel
+ */
+export interface WorkerProxy {
+  activeCount: number;
+  postMessage: (message: any, transfer: Transferable[]) => void;
+  free: () => void;
+}
+
+/**
+ * Returns a worker-like object that can be used to offload computation like Worker.
```
wchargin, line 33: “Returns a worker-like object”—are these docs meant for `getNext` rather
than the class, or else should they be rephrased?

tensorboard/webapp/widgets/line_chart_v2/lib/worker_pool_test.ts
```diff
@@ -0,0 +1,94 @@
...
+  describe('allocation', () => {
+    it('allocates pools even when instantiated pool with the same path', () => {
+      new WorkerPool('a', 1, workerFactory).getNext();
+      new WorkerPool('a', 1, workerFactory).getNext();
+      expect(workerFactory).toHaveBeenCalledTimes(2);
+    });
+
+    it('allocates upto maxPoolSize then allocates from first one again', () => {
+      const pool = new WorkerPool('a', 3, workerFactory);
+      const workers = [...new Array(5)].map(() => {
```
wchargin, line 48: …wow, TIL that `new Array(5)` is not the same as `[...new Array(5)]`.
Like, I knew about empty slots, but I always did `Array(5).fill().map`.

How lovely that `[...xs]` and `xs.slice()` are in fact _not_ always
equivalent when `Array.isArray(xs)`.

tensorboard/webapp/widgets/line_chart_v2/lib/worker_allocator.ts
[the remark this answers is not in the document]
wchargin: Sure, that works for me.

wchargin (approved):
Lovely; thanks!

wchargin:
> I actually do not know if we have any immediate need for a worker so I will not move it. While I am very reluctant to move it today, it is very easy to refactor/move so you can feel free to do so when necessary.

Yeah, I’m generally happy to default to “only put it in `util/` if it
has 3+ clients”, lest `util/` become a speculative dumping ground,
though I don’t hold this as a hard rule.

PR #4233, opened 2020-10-14
Temporarily disable usage of scalars_multirun endpoint

psybuzz:
The /scalars_multirun endpoint triggers errors when TensorBoard is
hosted with a Google Cloud Spanner backend. This change reverts
TensorBoard back to using `_requestDataGet` unconditionally.

Once [1] is synced into Google, we plan to re-enable the
`_requestDataPost` path for non-Colab environments.

Manually checked that the scalars dashboard makes GET requests
to `/data/plugin/scalars/scalars` with and without the
`?tensorboardColab=true` flag.

Googlers, see b/170675710.

[1] https://github.com/googleapis/python-spanner/pull/144

### DIFF
--- a/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
+++ b/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
@@ -277,7 +277,7 @@ export class TfScalarCard extends PolymerElement {
 
     // TODO(b/170675710): Stop unconditionally forcing legacy endpoint when
     // fixed.
-    const forceLegacyEnpoint = true;
+    const forceLegacyEndpoint = true;
     if (inColab || forceLegacyEndpoint) {
       return this._requestDataGet(items, onLoad, onFinish);
     } else {
--- a/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
+++ b/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
@@ -36,18 +36,6 @@ type RunTagItem = {run: string; tag: string};
 const initialURLSearchParams = new URLSearchParams(window.location.search);
 
 /**
- * Whether TensorBoard is running in a Colab environment.
- *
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
```diff
@@ -270,14 +282,19 @@ export class TfScalarCard extends PolymerElement {
...
   ) => {
-    const inColab = initialURLSearchParams.get('tensorboardColab') === 'true';
-    if (inColab) {
-      // Google-internal Colab doesn't support HTTP POST requests, so we fall
-      // back to HTTP GET (even though public Colab supports POST).
-      return this._requestDataGet(items, onLoad, onFinish);
-    } else {
-      return this._requestDataPost(items, onLoad, onFinish);
-    }
+    // TODO(psybuzz): Ideally, we would always make use of `_requestDataPost`,
```
nfelt, line 285: Could we get away with a smaller delta here and leave most of the code intact so it's easy to revert?  E.g.

```typescript
// TODO(b/170675710): Stop unconditionally forcing legacy endpoint when fixed.
const forceLegacyEnpoint = true;
if (inColab || forceLegacyEndpoint) {
  return this._requestDataGet(items, onLoad, onFinish);
} else {
  return this._requestDataPost(items, onLoad, onFinish);
}
```

That makes it obvious that we'll still need to the `inColab` check after resolving the TODO, which is better IMO than stating this fact in a comment.  And I'm not sure we need to link to the spanner PR given that the gating factor isn't anything to do with the PR per se but rather with Google-internal release details which aren't particularly visible or meaningful outside Google anyway.

nfelt (approved):

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
psybuzz: That's much more concise!  Done.

nfelt (approved):
LGTM

PR #4235, opened 2020-10-14
cleanup: NO_ERRORS_SCHEMA to avoid noise in test

stephanwlee:
Angular test specs have their own way of compiling and specifying
modules. As a result, when building Angular template, it can resolve to
a different implementation or nothing (e.g., <mat-icon> without
importing MatIconModule resolves to nothing). This makes Angular's *Ivy*
compiler write warning (instead of throwing exception and halting the
build) and litters our test.

Ideally, there should be a way to assert on console.warn and fail the
test but we will not invest much time in infrastructure in this PR, at
least.

Used this script to modify BUILD files (which resulted in no-op)
```sh
test_files="$(git show --pretty="format:" --name-only HEAD | grep _test.ts)"

for test_file in $test_files; do
  fullname=$(bazel query "$test_file")
  target=$(bazel query "attr('srcs', $fullname, ${fullname//:*/}:*)")
  buildozer 'add deps @npm//@angular/core' $target
done
```

### DIFF
--- a/tensorboard/webapp/plugins/npmi/npmi_container_test.ts
+++ b/tensorboard/webapp/plugins/npmi/npmi_container_test.ts
@@ -15,6 +15,7 @@ limitations under the License.
 /**
  * Unit tests for the NPMI Container.
  */
+import {NO_ERRORS_SCHEMA} from '@angular/core';
 import {TestBed} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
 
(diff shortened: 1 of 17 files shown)


bmd3k (approved):
Can you update the PR title NO_SCHEMA_ERROR => NO_ERRORS_SCHEMA?

stephanwlee:
Thanks for catching the typo

PR #4236, opened 2020-10-14
cleanup: remove extraneous console.log in test

stephanwlee:
Found the instance with: `rg 'console.log' -g '**/*_test.ts'`
There are no other instance of console.log in test.


psybuzz (approved):
Thanks (I think this log was probably me?)

PR #4240, opened 2020-10-15
line chart: layout system

stephanwlee:
This change adds a concept of layout. Layout is, as depicted in README,
a class that is rectangular (i.e., width and height) and can be
positioned or laid out based on parent dimensions.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/composite_layout.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/composite_layout.ts
@@ -1,0 +1,86 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/composite_layout.ts
```diff
@@ -0,0 +1,86 @@
...
+limitations under the License.
+==============================================================================*/
+
+import {RectLayout, LayoutStrategy, LayoutOption} from './layout';
+import {Rect} from './types';
+
+export class CompositeLayout extends RectLayout {
+  private readonly superposedLayouts: RectLayout[];
+
+  constructor(config: LayoutOption, contentColumns: RectLayout[]) {
```
wchargin, line 22: A `RectLayout` is also a composite layout, since it takes a rectangular
array of child `RectLayout`s. In what way is a `CompositeLayout`
different? Why does it only take columns?

wchargin (commented):
I understand the motivation for having a layout manager abstraction; that
sounds good to me. I don’t understand this particular incarnation of layout
manager. Main comment is
<https://github.com/tensorflow/tensorboard/pull/4240#discussion_r509477386>.

wchargin (commented):
Per discussion: it turns out that we don’t actually need this code for
the sequence of work that we’re planning on checking in. We need to
render a chart with an *x* axis, a *y* axis, a set of grid lines, and
the actual data series. There are two approaches:

 1. Render all elements in an `OffscreenCanvas` in a worker thread.
 2. Render the data series in an `OffscreenCanvas`, render the others in
    SVG, and composite them together with normal CSS grid in an Angular
    component.

We’re proceeding with option (2) because it’s simpler and gives better
output (3D canvas text rendering is weird), which means that we don’t
need our own layout engine at all.

PR #4241, opened 2020-10-15
line chart: added scale libraries

stephanwlee:
This module is anaologous to d3.scale. While doing performance
explorations, we found d3.scale operations to cause a lot of GC
impacting performance very significantly; a simple math operations were
taking a significant time that we would be spending more time converting
coordinate system than drawing.

This change introduces two kinds of scale; linear and log based 10. Some
utility method on the scales are pure wrapper around d3 method but for
the critical path, we wroute our own.

For posterity, we did look into using GPGPU to support mass coordinate
conversions but it introduced more noises and, near half the time, it
was slower than simple CPU based implementation.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
@@ -1,0 +1,196 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {scaleLinear, scaleLog} from 'd3';
+
+import {ScaleType} from './scale_types';
+
+export interface Scale {
+  forward(domain: [number, number], range: [number, number], x: number): number;
+
+  reverse(domain: [number, number], range: [number, number], x: number): number;
+
+  nice(minAndMax: [number, number]): [number, number];
+
+  /**
+   * @param lowAndHigh bounds of the tick
+   * @param sizeGuidance approximate number of the ticks. Depending on the lowAndHigh, it
+   * may return less or more ticks.
+   */
+  ticks(lowAndHigh: [number, number], sizeGuidance: number): number[];
+}
+
+export function createScale(type: ScaleType): Scale {
+  switch (type) {
+    case ScaleType.LINEAR:
+      return new LinearScale();
+    case ScaleType.LOG10:
+      return new Log10Scale();
+    default:
+      throw new RangeError(`ScaleType ${type} not supported.`);
+  }
+}
+
+const PADDING_RATIO = 0.05;
+const MIN_SIGNIFICANT_PADDING = 0.01;
+
+class LinearScale implements Scale {
+  private transform(
+    inputSpace: [number, number],
+    outputSpace: [number, number],
+    x: number
+  ): number {
+    const [inputMin, inputMax] = inputSpace;
+    const inputSpread = inputMax - inputMin;
+    const [outputMin, outputMax] = outputSpace;
+    const outputSpread = outputMax - outputMin;
+
+    if (inputSpread === 0) {
+      return outputMin;
+    }
+
+    return (outputSpread / inputSpread) * (x - inputMin) + outputMin;
+  }
+
+  forward(
+    domain: [number, number],
+    range: [number, number],
+    x: number
+  ): number {
+    return this.transform(domain, range, x);
+  }
+
+  reverse(
+    domain: [number, number],
+    range: [number, number],
+    x: number
+  ): number {
+    return this.transform(range, domain, x);
+  }
+
+  nice(minAndMax: [number, number]): [number, number] {
+    let [min, max] = minAndMax;
+    if (max < min) {
+      throw new Error('Unexpected input: min is larger than max');
+    }
+
+    const scale = scaleLinear();
+    const padding =
+      max === min
+        ? // In case both `min` and `max` are 0, we want some padding.
+          Math.max(min * PADDING_RATIO, MIN_SIGNIFICANT_PADDING)
+        : (max - min + Number.EPSILON) * PADDING_RATIO;
+    const [niceMin, niceMax] = scale
+      .domain([min - padding, max + padding])
+      .nice()
+      .domain();
+    return [niceMin, niceMax];
+  }
+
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  ticks(lowAndHigh: [number, number], sizeGuidance: number): number[] {
+    return scaleLinear().domain(lowAndHigh).ticks(sizeGuidance);
+  }
+}
+
+class Log10Scale implements Scale {
+  private transform(x: number): number {
+    return Math.log10(x > 0 ? x : Number.MIN_VALUE);
```
wchargin, line 108: This transform surprises me on two counts. First, it’s discontinuous and
non-monotonic: `transform(0.001) === -3` but `transform(0) === 5e-324`.
(`Number.MIN_VALUE` is the smallest *positive* float.) Second, why do we
want to override the standard behavior of `log` on edge cases? I always
thought that `log(0) = -Inf`, `log(x where x < 0) is NaN` was pretty
sensible.

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
```diff
@@ -0,0 +1,196 @@
...
+}
+
+export function createScale(type: ScaleType): Scale {
+  switch (type) {
+    case ScaleType.LINEAR:
+      return new LinearScale();
+    case ScaleType.LOG10:
+      return new Log10Scale();
+    default:
+      throw new RangeError(`ScaleType ${type} not supported.`);
```
wchargin, line 41: Let’s bind `const _: never = type;` so that we get a type error here if
we ever add a new scale type. (And then still throw the error, since
TypeScript isn’t sound.)

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
```diff
@@ -0,0 +1,196 @@
...
+import {scaleLinear, scaleLog} from 'd3';
+
+import {ScaleType} from './scale_types';
+
+export interface Scale {
+  forward(domain: [number, number], range: [number, number], x: number): number;
+
+  reverse(domain: [number, number], range: [number, number], x: number): number;
+
+  nice(minAndMax: [number, number]): [number, number];
```
wchargin, line 24: Okay, so I’m reading this as: “A `Scale` takes a `domain` and a `range`
(real intervals) and returns an iso between them. This is uncurried to
reduce allocations, for performance reasons.”

So far, so good. This immediately makes `forward` and `reverse` clear
(though nit: maybe call the parameter to `reverse` `y` rather than `x`).
But what is `nice`? Doubting [Nice][fr] or [`nice`][unix], but I can’t
guess from the signature or the implementations.

[fr]: https://en.wikipedia.org/wiki/Nice
[unix]: https://en.wikipedia.org/wiki/Nice_(Unix)

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * @param lowAndHigh bounds of the tick
```
wchargin, line 27: Unclear: what are these? Values in the domain? Values in the range?
Bounds of a single tick (like, the thing that’s 1–2 px wide)? Is this
just the extent of the axis, or only counting from the first tick mark?
What about the results—are they in domain-space or range-space?

This method seems useful and helpful; just a bit more clarity here would
be nice.

wchargin (commented):
This looks good! Sending preliminary review now; will review
implementation shortly, but I expect to approve it.

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
stephanwlee: First question: `transform(0.001) = -3` and `transform(0) = Math.log10(Number.MIN_VALUE) ~ -323.3` right?

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
```diff
@@ -0,0 +1,196 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {scaleLinear, scaleLog} from 'd3';
+
+import {ScaleType} from './scale_types';
+
+export interface Scale {
```
wchargin, line 19: Does the API require that `domain[1] > domain[0]` or `domain[1] >= domain[0]`?
Must `domain[i]` be finite, or at least non-NaN? Same for `range`.

My intuition is that it might make sense to require

  - all values finite (and thus non-NaN)
  - `domain[1] >= domain[0]`
  - `range[1] > range[0]`

since this is what the calling code can reasonably guarantee: i.e., it’s
filtered NaNs/infs from the dataset and is painting into a non-empty
region. The scales have to nicely handle the case where the domain is
empty/singleton.

What do you think?

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
wchargin: Also, `minAndMax` but `lowAndHigh`—do they mean the same thing? If so,
let’s be consistent. (And descriptive: perhaps `domain`?)

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
wchargin: Ah, yes, of course. Sounds good then. Thanks.

tensorboard/webapp/widgets/line_chart_v2/lib/scale_test.ts
```diff
@@ -0,0 +1,296 @@
...
+      it('converts value from domain space to range space', () => {
+        expect(scale.forward([0, 1], [-100, 100], 0)).toBe(-100);
+        expect(scale.forward([0, 1], [-100, 100], 0.5)).toBe(0);
+        expect(scale.forward([0, 1], [-100, 100], 1)).toBe(100);
+
+        expect(scale.forward([0, 1], [-100, 100], -1)).toBe(-300);
+        expect(scale.forward([0, 1], [-100, 100], 5)).toBe(900);
+      });
+
+      it('allows flippping order of the range', () => {
```
wchargin, line 37: sp.: *flipping*

wchargin (approved):
Test cases convince me that this looks pretty reasonable.

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
stephanwlee: Changed the name and left some comment about how it relates to `d3.nice`.

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
stephanwlee: Changed everything to `domain`.

PR #4244, opened 2020-10-16
Introduce alert snackbar UI

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/4221
Followup: https://github.com/tensorflow/tensorboard/pull/4245

Introduces an alert snackbar UI, which surfaces application
errors to the user in the corner of the screen.

Technical design discussion:

On the surface, there are many ways to architect an alert/error
system in ngrx, where our goal is to allow any feature in TB or
a TB embedder to trigger alerts in the snackbar UI.

Creating an 'Alert' feature store, whose reducer listens to other
features' actions would work for TB, but not embedders, since
TB core does not have access to ngrx actions from external
features.

One way to workaround this composability problem is to force
embedders to override the Alerts module and `composeReducers`.
Upon receiving an embedder-specific error, the embedder's
Alerts module can set the latest error accordingly. However,
this approach assumes that a reducer can determine whether
an action should produce an alert or not, based on the action's
payload.

In the followup, the 'cardPinStateToggled' action provides a
counterexample: it should only trigger an alert if there are too
many current pinned cards.

One way to workaround this is to introduce AlertEffects as a
layer that listens to actions from all features, and converts them
into generic 'alertReported' actions (assuming effects can be
composed, similar to reducers). However, this leads us to a world
where AlertModule depends heavily on other features. For
example, if AlertModule depends on feature F, it becomes
difficult to create a TB application / embedder that needs Alerts
without needing F. This is a design constraint we'd like to follow,
as it gives flexibility in bundling TB with only a subset of the
current features.

One might imagine that specific features can do the translation
themselves, using specific effects to convert actions into
'alertReported' actions. However, this synchronous dispatch of
composite actions goes against a best practice. In other
scenarios, it is known to lead to hard-to-reason-about in-between
states, and makes undoing or reverting state changes difficult
due to the processing of multiple actions.

Thus, we finally land on the existing patch. Essentially, it hides
away the composite action using Angular's DI framework. The
actions that should trigger alerts are registered once, and are
dispatched by the AlertEffects, rather than specific features.

Googlers, see test sync cl/338344258

### DIFF
--- a/tensorboard/webapp/alert/effects/index.ts
+++ b/tensorboard/webapp/alert/effects/index.ts
@@ -1,0 +1,49 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {Injectable} from '@angular/core';
+import {Actions, createEffect} from '@ngrx/effects';
+import {Store} from '@ngrx/store';
+import {tap} from 'rxjs/operators';
+import {State} from '../../app_state';
+import {alertReported} from '../actions';
+import {AlertActionModule} from '../alert_action_module';
+
+/** @typehack */ import * as _typeHackNgrxEffects from '@ngrx/effects/effects';
+/** @typehack */ import * as _typeHackStore from '@ngrx/store';
+/** @typehack */ import * as _typeHackRxjs from 'rxjs';
+
+@Injectable()
+export class AlertEffects {
+  constructor(
+    private readonly actions$: Actions,
+    private readonly store: Store<State>,
+    private readonly alertActionModule: AlertActionModule
+  ) {}
+
+  /** @export */
+  reportRegisteredActionAlerts$ = createEffect(
+    () => {
+      return this.actions$.pipe(
+        tap((action) => {
+          const alertInfo = this.alertActionModule.getAlertFromAction(action);
+          if (alertInfo) {
+            this.store.dispatch(alertReported(alertInfo));
+          }
+        })
+      );
+    },
+    {dispatch: false}
+  );
+}
--- a/tensorboard/webapp/alert/types.ts
+++ b/tensorboard/webapp/alert/types.ts
@@ -1,6 +1,4 @@
-<!--
-@license
-Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -13,13 +11,15 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
--->
+==============================================================================*/
+/**
+ * An alert structure used when creating newly reported alerts.
+ */
+export interface AlertReport {
+  details: string;
+}
 
-<app-header></app-header>
-<main #main>
-  <router-outlet></router-outlet>
-</main>
-<alert-snackbar></alert-snackbar>
-<hash-storage></hash-storage>
-<page-title></page-title>
-<settings-polymer-interop></settings-polymer-interop>
+/**
+ * An alert exposed by the feature's selectors.
+ */
+export type AlertInfo = AlertReport & {created: number};
--- a/tensorboard/webapp/alert/types.ts
+++ b/tensorboard/webapp/alert/types.ts
@@ -16,7 +16,7 @@ limitations under the License.
  * An alert structure used when creating newly reported alerts.
  */
 export interface AlertReport {
-  details: string;
+  localizedMessage: string;
 }
 
 /**
--- a/tensorboard/webapp/alert/views/alert_snackbar_container.ts
+++ b/tensorboard/webapp/alert/views/alert_snackbar_container.ts
@@ -21,7 +21,7 @@ import {
 import {MatSnackBar} from '@angular/material/snack-bar';
 import {Store} from '@ngrx/store';
 import {Subject} from 'rxjs';
-import {filter, takeUntil, tap} from 'rxjs/operators';
+import {filter, takeUntil} from 'rxjs/operators';
 
 import {State} from '../../app_state';
 import {getLatestAlert} from '../../selectors';
--- a/tensorboard/webapp/alert/views/alert_snackbar_container.ts
+++ b/tensorboard/webapp/alert/views/alert_snackbar_container.ts
@@ -1,0 +1,70 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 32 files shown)


tensorboard/webapp/app_module.ts
```diff
@@ -23,14 +23,15 @@ import {AppRoutingModule} from './app_routing/app_routing_module';
...
+import {ErrorModule} from './error/error_module';
+import {ErrorSnackbarModule} from './error/views/error_snackbar_module';
 import {HashStorageModule} from './core/views/hash_storage_module';
 import {PageTitleModule} from './core/views/page_title_module';
 import {FeatureFlagModule} from './feature_flag/feature_flag_module';
 import {HeaderModule} from './header/header_module';
 import {MatIconModule} from './mat_icon_module';
 import {PluginsModule} from './plugins/plugins_module';
 import {ROOT_REDUCERS, loggerMetaReducerFactory} from './reducer_config';
-import {ReloaderModule} from './reloader/reloader_module';
```
stephanwlee, line 33: Is this an unused import cleanup?

stephanwlee (commented):
> This PR introduces the concept of error "details", a string which
is defined in a TS module that will be surfaced in the UI.

This did not sufficiently describe your design choice; it looks like the change is recommending putting what should be localized in the action payload which kind of means that, for instance, there is a network error or whatever, you would be firing an action, presumably synchronously. That is conceptually regarding an action as a mutation/command, not event.

Are you willfully making the exception here? Does it need to be be done that way?

tensorboard/webapp/app_module.ts
psybuzz: Yes.  Although not related to the change, it is a small drive-by that fixes a lint warning upon sync, since this file is changed.

psybuzz:
In the followup, https://github.com/tensorflow/tensorboard/pull/4245, this approach requires us to include 'canCreateNewPins' and 'wasPinned' to the cardPinStateToggled action payload.

Overall, my impression of this approach is that it's slightly more complex than it might deserve to be.  I would be fine with this approach, or doing the simpler composite-action (noting that it would be an exception to our best practice.  Also open to any other ideas, if we have any.

tensorboard/webapp/alert/effects/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    private readonly alertActionModule: AlertActionModule
+  ) {}
+
+  /** @export */
+  reportRegisteredActionAlerts$ = createEffect(
+    () => {
+      return this.actions$.pipe(
+        tap((action) => {
+          const alertInfo = this.alertActionModule.getAlertFromAction(action);
+          if (alertInfo) {
```
stephanwlee, line 41: in this case, we would be subscribing to all actions. Would it make sense to use https://www.learnrxjs.io/learn-rxjs/operators/creation/defer and use `ofType`? 

It may have the same effect as your code so I would not insist on it.

tensorboard/webapp/alert/types.ts
```diff
@@ -0,0 +1,25 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+/**
+ * An alert structure used when creating newly reported alerts.
+ */
+export interface AlertReport {
+  details: string;
```
stephanwlee, line 19: I do not think this is sufficient to migrate existing code. We will need to have an optional actionType for an alert that can have actions (you may also want to have localized string for those actions).

stephanwlee, line 19: `details` -> `localizedMessage`?

tensorboard/webapp/reducer_config.ts
```diff
@@ -37,11 +37,12 @@ function logger(reducer: ActionReducer<any>): ActionReducer<any> {
...
-    ? (reducer) => (state, action) => {
-        return reducer(state, action);
-      }
-    : logger;
+  return logger;
+  // return !isDevMode()
+  //   ? (reducer) => (state, action) => {
+  //       return reducer(state, action);
+  //     }
+  //   : logger;
```
stephanwlee, line 45: revert please.

tensorboard/webapp/app_container.ng.html
```diff
@@ -19,6 +19,7 @@
 <main #main>
   <router-outlet></router-outlet>
 </main>
+<alert-snackbar></alert-snackbar>
```
stephanwlee, line 22: Hmm, this is not great for composability but I have no good recommendation here.

tensorboard/webapp/alert/views/alert_snackbar_test.ts
```diff
@@ -0,0 +1,96 @@
...
+            open: snackBarOpenSpy,
+          },
+        },
+      ],
+      declarations: [AlertSnackbarContainer],
+    }).compileComponents();
+    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
+  });
+
+  it('should open the snackbar on each alert', () => {
```
stephanwlee, line 49: omit the world "should" in test; it is redundant and not needed.

tensorboard/webapp/reducer_config.ts
psybuzz: Whoops, done.

tensorboard/webapp/alert/types.ts
psybuzz: Done.

tensorboard/webapp/alert/types.ts
```diff
@@ -0,0 +1,25 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+/**
+ * An alert structure used when creating newly reported alerts.
+ */
+export interface AlertReport {
+  localizedMessage: string;
```
stephanwlee, line 19: I may have left a comment in the other PR but I don't think I find any comment.

How would you plan on supporting different actions in the snackbar?  i.e., in some cases, we want to have button to retry sending a request. In order to replace our current error handling, we would need that facility. Without it, we will have two ways to achieve the same thing.

tensorboard/webapp/alert/views/alert_snackbar_container.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  ngOnInit() {
+    this.store
+      .select(getLatestAlert)
+      .pipe(
+        takeUntil(this.ngUnsubscribe),
+        filter((alert) => Boolean(alert)),
+        tap((alert) => {
+          this.showAlert(alert!.localizedMessage);
+        })
```
stephanwlee, line 53: maybe do this in the subscribe instead.

stephanwlee (approved):
LGTM. I don't think we immediately in this PR require the actionable snackbar so I will be okay if you want to do a follow up.

tensorboard/webapp/alert/types.ts
psybuzz: I'd like to do this in a separate PR, if that works for you.  This is what I had in mind:

extend `AlertReport` with a new field:
```
export interface AlertReport {
  localizedMessage: string;
  retryCallback?: (store: Store) => void;
}
```

used like so:
```
AlertActionModule.registerActionAlerts([
  {
    localizedMessage: "Fetch failed",
    retryCallback: (store) => {
      // ...
      store.dispatch(fetchAction(...));
    }
  },
])
```

with the AlertSnackbarContainer using `.openFromComponent`, instead of `.open`,
to open an AlertSnackbarComponent, which contains:

```
// template
<div *ngIf="!!alertInfo.retryCallback" (click)="onRetryClicked">Retry</div>

// component
@Input() alertInfo: AlertInfo

onRetryClicked() { this.alertInfo.retryCallback(); }
```

tensorboard/webapp/alert/views/alert_snackbar_container.ts
psybuzz: Done.

psybuzz (commented):
Thanks for the review!

tensorboard/webapp/alert/types.ts
stephanwlee: I wish it to be more generic than just retryCallback (and it, I don't think should be a callback since it opens the API for just about anything) but I will look forward to seeing the change afterwards.

stephanwlee (approved):

Issue #4242, opened 2020-10-15
Increase the max number of pinned cards in Time Series


psybuzz:
The Time Series dashboard allows you to pin cards, but is limited to
a fixed number to avoid overloading the URL.  If TensorBoard had
a more robust user settings storage system, e.g. filesystem,
we should be able to increase the max number of pins.

abhi-glitchhg:
any update on this?

afrozenator:
Why's there such an arbitrary limit?

Mekadrom:
This message is showing for me even though I only have one pin. I believe it is due to the url query args containing many more pins than are actually present.

jacobjennings:
I managed to clear outdated pins that kept coming back by navigating to http://localhost:6006/?pinnedCards=

gnulnx:
@dongdongtong  I'm not sure.  I run tensorboard outside of vscode from a different terminal.   I also typically do my training runs outside vscode as I have had the vscode terminal crash on me several times at this point.

PR #4245, opened 2020-10-16
[time series] prevent pinning cards from exceeding limit

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/4244

Prevents the "toggle pin" button from creating a new pin,
if we already have the max number of pins allowed.
Unresolved pins still count towards the quota.

This enforced limit prevents the URL from overflowing,
see https://github.com/tensorflow/tensorboard/issues/4242

Googlers, see test sync cl/338344567

image

### DIFF
--- a/tensorboard/webapp/metrics/metrics_module.ts
+++ b/tensorboard/webapp/metrics/metrics_module.ts
@@ -15,7 +15,7 @@ limitations under the License.
 import {CommonModule} from '@angular/common';
 import {NgModule} from '@angular/core';
 import {EffectsModule} from '@ngrx/effects';
-import {StoreModule} from '@ngrx/store';
+import {Action, StoreModule} from '@ngrx/store';
 import {PluginRegistryModule} from '../plugins/plugin_registry_module';
 
 import {RunsSelectorModule} from '../runs/views/runs_selector/runs_selector_module';
@@ -25,6 +25,13 @@ import {MetricsEffects} from './effects';
 import {METRICS_FEATURE_KEY, reducers} from './store';
 import {MetricsDashboardContainer} from './views/metrics_container';
 import {MetricsViewsModule} from './views/metrics_views_module';
+import {AlertActionModule} from '../alert/alert_action_module';
+import * as actions from './actions';
+
+const CREATE_PIN_MAX_EXCEEDED_TEXT =
+  `Max pin limit exceeded. Remove existing` +
+  ` pins before adding more. See ` +
+  `https://github.com/tensorflow/tensorboard/issues/4242`;
 
 @NgModule({
   imports: [
@@ -38,6 +45,20 @@ import {MetricsViewsModule} from './views/metrics_views_module';
     MetricsViewsModule,
     StoreModule.forFeature(METRICS_FEATURE_KEY, reducers),
     EffectsModule.forFeature([MetricsEffects]),
+    AlertActionModule.registerAlertActions([
+      {
+        actionCreator: actions.cardPinStateToggled,
+        alertFromAction: (action: Action) => {
+          const {wasPinned, canCreateNewPins} = action as ReturnType<
+            typeof actions.cardPinStateToggled
+          >;
+          if (!wasPinned && !canCreateNewPins) {
+            return {details: CREATE_PIN_MAX_EXCEEDED_TEXT};
+          }
+          return null;
+        },
+      },
+    ]),
   ],
   entryComponents: [MetricsDashboardContainer],
 })
--- a/tensorboard/webapp/metrics/metrics_module.ts
+++ b/tensorboard/webapp/metrics/metrics_module.ts
@@ -28,10 +28,32 @@ import {MetricsViewsModule} from './views/metrics_views_module';
 import {AlertActionModule} from '../alert/alert_action_module';
 import * as actions from './actions';
 
+/** @typehack */ import * as _typeHackModels from '@ngrx/store/src/models';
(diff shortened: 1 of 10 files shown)


tensorboard/webapp/metrics/views/card_renderer/card_view_container.ts
```diff
@@ -62,6 +70,35 @@ export class CardViewContainer {
...
+          this.store.dispatch(
+            actions.cardPinStateToggled({
+              cardId: this.cardId,
+              canCreateNewPins,
+              wasPinned: wasPinned,
+            })
+          );
+        })
+      )
+      .subscribe(() => {});
```
stephanwlee, line 94: alternatively, can we just use `this.store.select().pipe(take(1),...)`? down below in `onPinStateChanged`? Is this pattern better?

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -291,3 +291,17 @@ export function buildOrReturnStateWithPinnedCopy(
...
+ * finite at the moment to mitigate super long URL lengths, until there is more
+ * durable value storage for pins.
+ */
+const MAX_PIN_COUNT = 10;
+
+export function canCreateNewPins(state: MetricsState) {
+  const pinCountInURL =
+    state.pinnedCardToOriginal.size +
+    state.unresolvedImportedPinnedCards.length;
+  return pinCountInURL < MAX_PIN_COUNT;
```
stephanwlee, line 306: there should be an accompanying test.

psybuzz: Done

tensorboard/webapp/metrics/views/card_renderer/card_view_container.ts
```diff
@@ -88,7 +88,22 @@ export class CardViewContainer {
   }
 
   onPinStateChanged() {
-    this.store.dispatch(actions.cardPinStateToggled({cardId: this.cardId}));
+    this.store
+      .select(selectors.getCardPinnedState, this.cardId)
+      .pipe(
+        take(1),
+        withLatestFrom(this.store.select(selectors.getCanCreateNewPins)),
+        tap(([wasPinned, canCreateNewPins]) => {
```
stephanwlee, line 96: Maybe do what's inside the `tap` in subscribe instead.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -433,4 +439,61 @@ describe('metrics store utils', () => {
...
+          ['pinnedCard3', 'card3'],
+        ]),
+        unresolvedImportedPinnedCards: [],
+      });
+
+      expect(canCreateNewPins(state)).toBe(false);
+    });
+
+    it('returns false when unresolved pins reaches the limit', () => {
+      TEST_ONLY.util.MAX_PIN_COUNT = 3;
```
stephanwlee, line 471: umm, you are mutating the global state and it won't restore itself. Maybe use `spyOnProperty` which should restore the original value.

tensorboard/webapp/metrics/views/card_renderer/card_view_container.ts
psybuzz: Done.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
psybuzz: Ah, added an `afterEach` to restore the original value.

It seems `spyOnProperty` does not work for object literal properties, only getter properties or ones created with Object.defineProperty.

tensorboard/webapp/metrics/views/card_renderer/card_view_container.ts
```diff
@@ -20,7 +20,7 @@ import {
 } from '@angular/core';
 import {Store} from '@ngrx/store';
 import {Observable} from 'rxjs';
-import {map, throttleTime} from 'rxjs/operators';
+import {map, take, tap, throttleTime, withLatestFrom} from 'rxjs/operators';
```
stephanwlee, line 23: `tap` likely to be unused.

tensorboard/webapp/metrics/metrics_module.ts
```diff
@@ -25,6 +25,13 @@ import {MetricsEffects} from './effects';
 import {METRICS_FEATURE_KEY, reducers} from './store';
 import {MetricsDashboardContainer} from './views/metrics_container';
 import {MetricsViewsModule} from './views/metrics_views_module';
+import {AlertActionModule} from '../alert/alert_action_module';
+import * as actions from './actions';
+
+const CREATE_PIN_MAX_EXCEEDED_TEXT =
+  `Max pin limit exceeded. Remove existing` +
+  ` pins before adding more. See ` +
+  `https://github.com/tensorflow/tensorboard/issues/4242`;
```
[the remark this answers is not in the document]
stephanwlee: Thanks for the screenshot. I was wondering how a generic action would be handled in the snackbar but linkification could be a fair start.

stephanwlee (approved):

tensorboard/webapp/metrics/views/card_renderer/card_view_container.ts
psybuzz: Thanks, removed

PR #4249, opened 2020-10-19
line chart: introduce renderer

stephanwlee:
Renderer is an abstraction over how we are going to draw line, text, and
rect. For now, we only support two renderer, SVG and three.js based
WebGL.

Some design choices:
- decided to make cacheId as required field as part of the renderer API
    for better cache controls.
- decided to namespace the cacheIds with a renderGroup; a cacheId has to
    be unique within a render group.


tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
```diff
@@ -0,0 +1,39 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {Paths, Rect} from '../types';
+
+export interface IRenderer {
```
wchargin, line 17: The interface is more important, so it should get the good name:
`interface Renderer`, `class BaseRenderer` or something.

tensorboard/webapp/widgets/line_chart_v2/lib/types.ts
```diff
@@ -31,3 +31,8 @@ export interface Extent {
   x: [number, number];
   y: [number, number];
 }
+
+/**
+ * Flattened array of 2d coordinates: [x0, y0, x1, y1, ..., xn, yn].
+ */
+export type Paths = Float32Array;
```
wchargin, line 38: From the comment I can see that it’s a sequence of points. Why is it
called `Paths`? Is it really a `Points`, or a [`Polyline`][pl], or a
`Polygon`, or something else?

[pl]: https://en.wikipedia.org/wiki/Polyline

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  drawLine(cacheId: string, paths: Paths, spec: LineSpec): void;
+
+  flush(): void;
+
+  renderGroup(groupName: string, renderBlock: () => void): void;
+}
+
+export interface LineSpec {
+  visible: boolean;
```
wchargin, line 34: Why would I call `drawLine` with `visible: false`? Is that supposed to
somehow erase the line? (Seems dubious.)

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
```diff
@@ -0,0 +1,39 @@
...
+  onResize(domRect: Rect): void;
+
+  drawLine(cacheId: string, paths: Paths, spec: LineSpec): void;
+
+  flush(): void;
+
+  renderGroup(groupName: string, renderBlock: () => void): void;
+}
+
+export interface LineSpec {
```
wchargin, line 33: IMHO these are more “options”/“paint options”/“draw options” than a
spec; surely a spec must include the actual coordinates to be drawn.

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
```diff
@@ -0,0 +1,39 @@
...
+   *
+   * @param domRect Container dimensions
+   */
+  onResize(domRect: Rect): void;
+
+  drawLine(cacheId: string, paths: Paths, spec: LineSpec): void;
+
+  flush(): void;
+
+  renderGroup(groupName: string, renderBlock: () => void): void;
```
wchargin, line 30: I can guess from the name and type signature what “kind” of thing this
is (a scoped bracketing operation), but what are the semantics? Is this…
composited onto an intermediate buffer so that opacity transforms can be
applied atomically? rendered transactionally with opportunities for
internal performance improvements? just a helper for `flush`? Why would
I call this method, what must I ensure as a caller, and what does the
method guarantee?

What, also, are the semantics of the `groupName`? From the abstract
class, it looks like a key. Do the keys last just for the duration of
the render group, or are they owned externally to the callback? Is it
valid to reuse the key, like

```typescript
r.renderGroup("foo", () => {
  r.renderGroup("foo", () => {
  });
});
```

? If not, if the key is meant to always be unique, would it make more
sense for `renderGroup` to gensym a group name and pass it to the
`renderBlock` callback? Surely the answer depends on how the group name
is used, so… how is it used?

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/svg_renderer.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (!cache?.data || !arePathsEqual(paths, cache?.data)) {
+      const data = this.createPathDString(paths);
+      svgPath.setAttribute('d', data);
+      renderCache.set(cacheId, {cacheable: svgPath, data: paths});
+    }
+
+    svgPath.style.display = cssDisplayValue;
+    svgPath.style.stroke = color;
+    svgPath.style.opacity = String(opacity ?? 1);
```
wchargin, line 80: > ```typescript
> opacity ?? 1
> ```

Neat that we can actually use this.

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface IRenderer {
+  /**
+   * Certain renderer requires DOM dimensions for correct density and operations. The
+   * method is invoked when container is resized.
+   *
+   * @param domRect Container dimensions
+   */
+  onResize(domRect: Rect): void;
+
+  drawLine(cacheId: string, paths: Paths, spec: LineSpec): void;
```
wchargin, line 26: So, after reading the types, `paths` and `spec` make sense. But I’m not
used to rendering contexts having a built-in notion of caching. What’s
the `cacheId` for, and what are its semantics?

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_test.ts
```diff
@@ -0,0 +1,356 @@
...
+          DEFAULT_LINE_OPTIONS
+        );
+      });
+
+      expect(scene.children.length).toBe(1);
+      const lineAfter = scene.children[0] as THREE.Line;
+      assertPaths(lineAfter, new Float32Array([0, 1, 0, 1]));
+    });
+
+    it('skips renderering if render starts with visibility=false ', () => {
```
wchargin, line 345: sp.: too many *er*s

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
wchargin, line 17: I note no `clear` method. Is that not needed? Or just coming later?

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer.ts
```diff
@@ -0,0 +1,74 @@
...
+  private currentRenderGroupCache: RenderGroupMap<Cacheable> | null = null;
+  protected cacheIdsToRemove = new Set<string>();
+
+  protected getRenderCache(): RenderGroupMap<Cacheable> {
+    if (!this.currentRenderGroupCache) {
+      throw new RangeError(
+        'Invariant error: expected getRenderCache to be invoked inside a renderGroup'
+      );
+    }
+    return this.currentRenderGroupCache!;
```
wchargin, line 49: This unsafe `!` isn’t needed; the type is refined due to the `if`.

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer.ts
```diff
@@ -0,0 +1,74 @@
...
+    string,
+    RenderGroupMap<Cacheable>
+  >();
+  private currentRenderGroupCache: RenderGroupMap<Cacheable> | null = null;
+  protected cacheIdsToRemove = new Set<string>();
+
+  protected getRenderCache(): RenderGroupMap<Cacheable> {
+    if (!this.currentRenderGroupCache) {
+      throw new RangeError(
+        'Invariant error: expected getRenderCache to be invoked inside a renderGroup'
```
wchargin, line 46: nit: This is a precondition failure, not an invariant error, since your
abstraction cannot guarantee that the desired invariant holds (since
subclasses exist).

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/svg_renderer.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  drawLine(cacheId: string, paths: Paths, spec: LineSpec) {
+    if (!paths.length) {
+      return;
+    }
+
+    super.drawLine(cacheId, paths, spec);
+
+    const renderCache = this.getRenderCache();
```
wchargin, line 50: Won’t this throw if a client calls `drawLine` outside a `renderGroup`?

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer.ts
```diff
@@ -0,0 +1,74 @@
...
+  abstract flush(): void;
+
+  private groupToCacheIdToCacheable = new Map<
+    string,
+    RenderGroupMap<Cacheable>
+  >();
+  private currentRenderGroupCache: RenderGroupMap<Cacheable> | null = null;
+  protected cacheIdsToRemove = new Set<string>();
+
+  protected getRenderCache(): RenderGroupMap<Cacheable> {
```
wchargin, line 43: From subclass use, it looks like the mutable render cache map and set of
cache IDs to remove are pretty important. How are subclasses supposed to
use them? Please document.

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export type RenderGroupMap<Cacheable> = Map<
+  string,
+  {
+    data: Paths;
+    cacheable: Cacheable;
+  }
+>;
+
+export abstract class Renderer<Cacheable> implements IRenderer {
```
wchargin, line 27: From subclass use, it looks like the mutable render cache map and set of
cache IDs to remove are pretty important. How are subclasses supposed to
use them? Please document.

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer.ts
```diff
@@ -0,0 +1,74 @@
...
+limitations under the License.
+==============================================================================*/
+
+import {LineSpec, IRenderer} from './renderer_types';
+import {Paths, Rect} from '../types';
+
+export type RenderGroupMap<Cacheable> = Map<
+  string,
+  {
+    data: Paths;
```
wchargin, line 22: Is this a cache specifically for `drawLine`? When there are other
drawing functions, what will go here?

Relatedly, I wonder whether it makes sense to include `data` in this
struct at all. Couldn’t `data` just be part of the `Cacheable` type
argument?

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer.ts
wchargin, line 27: From the name, `Cacheable` really sounds like a key: strings are
cacheable, numbers are cachable, arrays are not cacheable. But it looks
like it’s actually being used as a cache _value_. I was pretty confused
for a while. Can we clarify?

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
wchargin, line 30: Follow-up question: I’m getting the impression that this interface
expects its clients to call `drawLine` with the same cache ID many times
in the same render group. Why would they do that? If the render group
lives not longer than a frame (just a guess), and the cache ID
corresponds to one line segment on screen (just a guess), why would a
client draw a given line segment more than once per frame? and thus why
cache at all instead of just painting immediately?

wchargin (commented):
In your mind there is a picture, but I see only glimpses. I need to
illuminate the dark corners before I can review the implementation.

Stopping here to avoid comment spam.

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
stephanwlee: Done. Thanks for that.

tensorboard/webapp/widgets/line_chart_v2/lib/types.ts
stephanwlee: I like "Polyline" so I applied it everywhere. Thanks for the suggestion.

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
stephanwlee: In case of canvas based renderer, it would mean nothing but for object based renderer, it means we can just "`display: none`" and everything is still correct.

This is so that toggling a run does not remove/recreate same object over and over.

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
stephanwlee: Ah, you cannot nest a renderGroup so added some check and test for that. Thanks for catching that.

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
stephanwlee: I did not yet find a reason to have `clear` method for an object based renderer. We may add it later though.

stephanwlee:
I tried to capture some of the details you have required in the Renderer interface. Hope it contains enough information and please let me know if I should revise my abstraction.

wchargin (commented):
Okay. Thanks for the discussion. Here are my comments.

Since this machinery was developed with constraints/plans different from
our current ones, my goal is to reduce the _conceptual surface area_ of
this interface as much as possible, but not further. (For instance, by
removing the layout system in #4240.)

## Group names

Currently, elements are keyed by both render group name and cache key.
Render group names are intended to accommodate having one renderer with
multiple “drawables”: e.g., a scalar chart that with a `SeriesLineView`
but also an `InteriorTickView` and `ErrorBarView`. But: (a) we have no
current plans to introduce such patterns, and (b) this only punts the
problem to figuring out unique names for the render groups. (I’m not
convinced that using `this.constructor.name` for the group is a great
solution. Not only is it [unhygienic][hy], it also means that you can
only have one of each class of drawable per renderer, so I don’t think
that it really solves the problem.)

Since we don’t intend to expose this API publicly, if we do decide that
we need it later, we can add it then.

[hy]: https://en.wikipedia.org/wiki/Hygienic_macro 

Thus, I propose removing group names, at which point “`renderGroup`”
just becomes (per your suggestion) “`updateScene`”:

```typescript
interface Renderer {
  // Invoke a callback that calls `drawLine` to populate all objects
  // currently in the scene. Then, update the rendered view to match the
  // described state. Callback must not call `updateScene`, `onResize`.
  updateScene(renderBlock: () => void): void;

  // Draw a line. Must only be called inside `updateScene`. If called
  // multiple times for the same `cacheId` inside one `updateScene`
  // block, the last call wins.
  drawLine(cacheId: string, polyline: Polyline, paintOpt: LinePaintOption);

  // Repaint current scene state in response to viewport change.
  onResize(domRect: Rect): void;

  // ???
  flush();
}
```

## Remove `flush`

I don’t see why this needs to be part of the public API. It might make
sense as an abstract method on `BaseCachedObjectRenderer` to be called
at the end of `renderGroup`, but clients should never call it directly,
right?

## Scope drawing methods

You’re only allowed to call `drawLine` from within a render group. Make
the invalid states unrepresentable:

```typescript
interface Paintbrush {  // TODO(@stephanwlee): name?
  // Draw a line. If called multiple times for the same `cacheId`, the
  // last call wins.
  drawLine(cacheId, polyline, paintOpt);
}

interface Renderer {
  // Invoke a callback that calls `drawLine` to populate all objects
  // currently in the scene. Then, update the rendered view to match the
  // described state. Callback must not call any method on this instance.
  updateScene(renderBlock: (Paintbrush) => void);

  // Repaint current scene state in response to viewport change.
  onResize(Rect);
}
```

## _Consider_ making scene elements data, not calls

There is a sense in which these are equivalent:

```typescript
// Imperative API
renderer.updateScene((p) => {
  p.drawLine("foo", fooPts, fooOpts);
  p.drawLine("bar", barPts, barOpts);
  p.drawEllipse("baz", bazPts, bazOpts);
});

// Declarative API
const scene = new Map();
scene.set("foo", {type: "LINE", points: fooPts, opts: fooOpts});
scene.set("bar", {type: "LINE", points: barPts, opts: barOpts});
scene.set("baz", {type: "ELLIPSE", points: bazPts, opts: bazOpts});
renderer.updateScene(scene);
```

The declarative approach has some benefits:

  - While constructing the scene, you can `scene.delete` to “un-paint”
    an element. Once you’ve called `drawLine`, you can’t do that.
  - The “last element wins” semantics are obvious from the semantics of
    `Map`. Sure, the last `drawLine` call for an ID wins, but what if
    I call `drawLine("foo", ...); drawEllipse("foo", ...)`? Does it
    override the shape type, or is this illegal, or are they separate?
  - No need for callbacks, eliminating issues like “the callback passed
    to `updateScene` must not itself call `updateScene`”
  - With this API, it is actually somewhat easy to re-implement render
    group scoping compositionally by implementing this signature:

    ```typescript
    scopedRenderers(delegate: Renderer): (groupName: string) => Renderer;
    ```

    (You can do it the other way, too, but you basically end up needing
    to transform it into the declarative form.)

Over chat you seemed less than enthusiastic about this idea, so I’m
proposing it here merely as an idea.

---

Thoughts? I think this is the extent of my large interface-level
suggestions, so after these are implemented probably just one more round
of review should suffice.

stephanwlee:
Based on the feedback received, I decided to change the interface. I realized renderer was quite full of logic and managing cache on top of the busy rendering can be unpleasant. After the change, the renderer no longer maintains the cache but simply expects the larger framework to maintain and pass a cached (or recycled) object. This change allowed me to delete ~300 LOC partially due to a simpler test.

Now, I believe the refactor made it even harder to see the forest since I defined yet another tree that looks very estranged from the final implementation. As such, I decided to write a bit more context in README even if not all pieces are introduced yet to help understanding. Also, I decided to make the next implementation more concrete and push the next implementation here <https://github.com/tensorflow/tensorboard/commit/de92610184ae115b6d038013d30656fd91385f43>. 

Now, going over the [comment](https://github.com/tensorflow/tensorboard/pull/4249#pullrequestreview-519117072):
- **Group name**: this is no longer needed. I decided to make a small stateful container that Drawable controls. 
- **flush**: as argued offline, while I accept that it is a premature optimization (every DataDrawable can invoke the `renderer.flush()` instead of doing it just once), I have changed the contract so `setLine` has semantic of "drawing or enqueuing drawing operations depending on a renderer implementation". With this change, I think `flush` should commit the enqueued operations to pixels. Perhaps, should I call it `commit` instead?
- Data vs. imperative: still thinking about it. I think we can take a different approach if the imperative one leads to poorer code later, but, for now, I think I want to stick with the imperative one.

Am I heading into a better direction or did I make a wrong step?

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_test.ts
```diff
@@ -0,0 +1,226 @@
...
+    let renderer: ThreeRenderer;
+    let scene: THREE.Scene;
+
+    function assertLine(line: THREE.Line, polyline: Polyline) {
+      const geometry = line.geometry as THREE.BufferGeometry;
+      const positions = geometry.getAttribute(
+        'position'
+      ) as THREE.BufferAttribute;
+      let positionIndex = 0;
+      for (let index = 0; index < polyline.length; index += 2) {
```
wchargin, line 127: optional: maybe `polylineIndex`, to disambiguate?

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/svg_renderer.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return dBuilder.join('');
+  }
+
+  createOrUpdateLineObject(
+    cachedLine: CacheValue | null,
+    polyline: Polyline,
+    paintOpt: LinePaintOption
+  ): CacheValue | null {
+    if (!polyline.length) return null;
```
wchargin, line 62: If `cachedLine` is an existing cache key with more than zero points,
don’t we still need to do something here to hide the line?

wchargin (approved):

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/svg_renderer.ts
stephanwlee: Thanks for catching that. It definitely is a bug and I added explicit tests for that.

stephanwlee:
Thanks a lot for being patient with me and giving me valuable feedbacks on concepts I am introducing. Looking forward to make it even more intuitive!

PR #4253, opened 2020-10-23
Re-enable scalars multirun endpoint

psybuzz:
After the /scalars_multirun endpoint was introduced, it was
disabled due to b/170675710. Now that the bug is fixed on
the server side, we can re-enable the endpoint call on the
client side.


nfelt:
FWIW, we probably want to hold off on merging this until we know that the backend is fixed, so that we can still release the frontend from HEAD without needing to worry about regressions.

bmd3k (approved):

PR #4258, opened 2020-10-27
sync: depend on third_party d3 indirection

stephanwlee:
Due to the build quirk, our webapp cannot depend on d3 directly. To work
around, we can depend on our indirection under
//tensorboard/webapp/third_party. This change does that.


nfelt (approved):

nfelt:
Travis died fetching pip packages, so I restarted the job, FYI.

PR #4260, opened 2020-10-27
Linkify text in alert message UI

psybuzz:
Diffbase: https://github.com/tensorflow/tensorboard/pull/4257

Note to reviewer: please review only the commits *after* the diffbase.
e.g. all commits after the one titled "unused":
https://github.com/tensorflow/tensorboard/pull/4260/commits/39313de17480687eb24d87326c9dc4638e1db0bf
I made a mistake and should have pushed these branches to the source
repo rather than my own.

In the recently added Alert snackbar view, messages may include
URLs that could be linkified to benefit the user. This change
automatically linkifies URLs within the message.

Googlers, see test sync cl/339528257

image

### DIFF
--- a/tensorboard/webapp/alert/views/alert_display_snackbar_container.ng.html
+++ b/tensorboard/webapp/alert/views/alert_display_snackbar_container.ng.html
@@ -14,7 +14,18 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-<div class="message">{{ alert.localizedMessage }}</div>
+<div class="message">
+  <ng-container *ngFor="let piece of splitByURL(alert.localizedMessage)">
+    <ng-container *ngIf="!piece.isURL; else linkPiece">
+      {{ piece.text }}
+    </ng-container>
+    <ng-template #linkPiece>
+      <a href="{{ piece.text }}" rel="noreferrer noopener" target="_blank"
+        >{{ piece.text }}</a
+      >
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/alert/views/alert_display_snackbar_container.ng.html
```diff
@@ -14,7 +14,18 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-<div class="message">{{ alert.localizedMessage }}</div>
+<div class="message">
```
bmd3k, line 17: Just wondering: Did you consider a solution where the messages can include markup? It would be up to the source of the message to explain how it should be formatted. The source would be responsible for specying <a> tags but would not just be limited to linkifying explicity urls.

psybuzz: I had not deeply considered any richer alert UI, beyond text, links, and a followup action button.  Providing a string-based message is relatively simple, and there are a handful of immediate use cases we forsee: alert upon most errors (e.g. fetch runs failed, updating experiment description failed + retry, copied to clipboard would even work).

If we did need to render a more expressive component w/ tags, I would consider whether that should be a one-off modal, since it already requires creating a specific NgComponent.  If there are good reasons to support that in the same Alert ngrx feature, I could imagine us extending the state to store a NgComponent, and the view would use `componentFactory` to instantiate it.

Right now, I'm not aware of any use cases that warrant alerts with custom HTML, but please let me know if there any.

bmd3k (approved):

PR #4264, opened 2020-10-28
[nPMI] Embeddings Similar

Sparkier:
Users can now sort the annotations list by similarity when clicking on an annotation in the list.

* Screenshots of UI changes
image

`bazel test //tensorboard/webapp:karma_test_chromium_local`


Sparkier:
@stephanwlee did this change: `http://localhost:6007/?experimentalPlugin=npmi`
I get redirected to `http://localhost:6007/#npmi` but cannot reload from there.

Sparkier:
Also, @stephanwlee, I am wondering about this: `import {getCurrentRouteRunSelection} from '../../../../../selectors';`.

I often filter stuff based on the runs. E.g. I only want to display metrics for active runs. And I have `runToMetrics`, which is currently broken with the changed path. Can you give me a hint about the correct way to fix that?

stephanwlee:
Sorry for the late reply. I was on vacation on Friday and I was focused on this renderer work.

About redirection: Yes, this is kind of reality. We recently introduced our app router and it overwrites everything it does not know about. We can teach it the `experimentalPlugin` but given that it is not the normative user flow, we have not prioritize that.

About the new selector: would you be able to provide more information on what is broken about it? It returns a `Map`.

Sparkier:
To give a precise example:
Before, the run was `.`, now it is `defaultExperimentId/.` for  `getCurrentRunSelection`, but not for my `runToMetrics`. If all URLs change the same way, I could just do `.split('/').last()`, but I was wondering if this is the way to do this.

Sparkier:
@stephanwlee could you tell me your best practice for when and how to interpolate between runs in the form of `first` and `defaultExperimentId/first`?

stephanwlee:
In the Angular TensorBoard, we are experiment aware so it would be the best if you can, at the data source level, create a `runId` from a runName (Polymer TensorBoard used to treat runName as `id`s but they are not unique at all so we need to have a notion of `runId` separate from a displayName). More concretely, you need something like this: https://github.com/tensorflow/tensorboard/blob/6708b6b938fffec6c4d4f86547f8f81024de3c2e/tensorboard/webapp/runs/data_source/runs_data_source.ts#L39-L46

Sparkier:
Ok I think I got most of it now. What I did is convert my data in the frontend so that all runIds are prefixed with the experimentIds. What I am wondering now is if I should also use that in the color scale. I was thinking about using this:
https://github.com/tensorflow/tensorboard/blob/74aeb448e2e898c479d47da708e9471697140052/tensorboard/webapp/metrics/views/card_renderer/card_view_container.ts#L65-L80
But then it does not match the color of the run selector anymore. What would be best practice here?

Sparkier:
Got it, thank you. Needed to change the colorScale and runSelector as well, so will need a little more time to fix all loose ends.

tensorboard/webapp/plugins/npmi/util/sort_annotations.ts
```diff
@@ -15,47 +15,126 @@ limitations under the License.
...
+function calculateDistance(
+  reference: number[],
+  toEmbedding: number[],
+  faultCase: number = Number.POSITIVE_INFINITY
+): number {
+  if (reference.length != toEmbedding.length) return faultCase;
+  const subtracted = toEmbedding.map((i, n) => i - reference[n]);
+  const powered = subtracted.map((e) => Math.pow(e, 2));
+  const sum = powered.reduce((total, current) => total + current, 0);
+  return Math.sqrt(sum);
```
stephanwlee, line 139: I forget but is sqrt important for the cartesian distance? If it is only used for sorting, do you need to sqrt?

tensorboard/webapp/plugins/npmi/util/sort_annotations_test.ts
```diff
@@ -41,21 +42,95 @@ describe('sort annotations utils', () => {
...
+    };
+    const annotations = sortAnnotations(annotationData, sort, embeddingData);
+    expect(annotations).toEqual([
+      'annotation_2',
+      'annotation_1',
+      'annotation_3',
+    ]);
+  });
+
+  it('sorts annotations by similar embeddings with empty embedding', () => {
```
stephanwlee, line 91: would you add a case where `annotation_1` `embeddingData` is null

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
```diff
@@ -270,7 +273,7 @@ export class AnnotationComponent implements AfterViewInit, OnChanges {
         }.bind(this)
       )
       .attr('class', `hint-text ${this.textClass}`)
-      .text((d: string) => d);
+      .text((d: string) => d.split('/').slice(1).join());
```
stephanwlee, line 276: Instead of this, you may want to get the displayName for a run separate from the id. In other words, don't derive the display name from the id.

tensorboard/webapp/plugins/npmi/data_source/npmi_data_source.ts
```diff
@@ -124,23 +175,91 @@ export class NpmiHttpServerDataSource implements NpmiDataSource {
     );
   }
 
-  private fetchAnnotations() {
-    return this.http.get<AnnotationListing>(
-      this.httpPathPrefix + '/annotations'
+  private fetchAnnotations(experimentIds: string[]) {
+    const fetches = experimentIds.map((experimentId) => {
+      const url = `/experiment/${experimentId}/${this.httpPathPrefix}/annotations`;
```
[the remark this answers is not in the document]
psybuzz: I would keep as-is `/experiment/${experimentId}/${this.httpPathPrefix}/annotations` to be consistent with the Time Series data source.  Relying on the backend to strip the experiment/:id is safe for now, as changing that behavior would break Time Series as well.

tensorboard/webapp/plugins/npmi/store/npmi_types.ts
```diff
@@ -66,6 +66,8 @@ export interface AnnotationSort {
 export enum SortOrder {
   DOWN,
   UP,
+  SIMILAR,
+  DISSIMILAR,
```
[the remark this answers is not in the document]
Sparkier: I think I dont get what you are trying to say here. Changing `SIMILAR` to `ASCENDING`? I think that would not be very readable, no?

tensorboard/webapp/plugins/npmi/util/sort_annotations.ts
Sparkier: In our case, we can remove it, as we dont use the exact distance for anything.

Sparkier: It only transforms the domain that the function operates on, so not needed in our case.

tensorboard/webapp/plugins/npmi/util/sort_annotations.ts
```diff
@@ -15,47 +15,126 @@ limitations under the License.
...
+        if (annotation === sort.metric) {
+          distance[annotation] = Number.NEGATIVE_INFINITY;
+        } else {
+          distance[annotation] = embeddingData[annotation]
+            ? calculateDistance(
+                embeddingData[sort.metric],
+                embeddingData[annotation]
+              )
+            : Number.POSITIVE_INFINITY;
+        }
```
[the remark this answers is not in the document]
Sparkier: Id I understood you correctly, that is what I did.

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
Sparkier: I think I dont completely understand. This is what happens in the metrics plugin:
```
function parseRunId(runId: string): {run: string; experimentId: string} {
  const slashIndex = runId.indexOf('/');
  return {
    run: runId.substring(slashIndex + 1),
    experimentId: runId.substring(0, slashIndex),
  };
}
```
How should I get the runId instead?

tensorboard/webapp/plugins/npmi/store/npmi_types.ts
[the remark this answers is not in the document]
stephanwlee: Sorry, I was referring to how DOWN/UP are describing directionality but SIMILAR/DISSIMILAR is describing _what_ AND directionality.

tensorboard/webapp/plugins/npmi/util/sort_annotations.ts
```diff
@@ -15,47 +15,116 @@ limitations under the License.
...
 export function sortAnnotations(
   annotationData: AnnotationDataListing,
-  sort: AnnotationSort
+  sort: AnnotationSort,
+  embeddingData: EmbeddingListing
 ): string[] {
   let result = Object.keys(annotationData);
-  const strippedMetric = stripMetricString(sort.metric);
-  if (sort.metric === '') {
+  let similarityBased =
```
stephanwlee, line 29: prefer to use `const`.

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ng.html
```diff
@@ -24,7 +24,12 @@
     [checked]="selectedAnnotations.includes(annotation)"
     (click)="$event.preventDefault()"
   ></mat-checkbox>
-  {{annotation}}
+  <div
+    [ngClass]="hasEmbedding ? 'clickable-annotation' : ''"
+    (click)="similaritySort($event)"
```
stephanwlee, line 29: Prefer to use `<button>` for an interactable element (for accessibility)

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
stephanwlee: Metrics plugin is doing it right: it is defining that function and is being only used in the data source. Data source is the layer where we are more aware of a specific backend logic and in that case, we needed to derive runId and runDisplayName. 

I will put it in another way. Previously, there was no difference between runId vs. run display name; everything was just a string and they were not unique (you've already handled this in the data source). Now, in the experiment aware Angular code, we now differentiate name vs. id explicitly and regard the runId as something opaque. Data structure like the ValueData should only use the `runId` and never hold onto the `runName`. Here, the issue is, `runId`, depending on a backend, can be literally just numbers or random bytes, not composed of `eid/rid` yet here, you are extracting the name from this opaque id. 

Now, moving on to remediation: runSelector provides a data structure that gives you more metadata about a run including its displayName. You should use it to grab the data from the container, form a data structure, and pass it down to this component. I suspect it will be a large for this already large PR so I recommend fixing it in a quick follow up.

What do you think?

tensorboard/webapp/plugins/npmi/store/npmi_types.ts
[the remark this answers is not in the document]
Sparkier: Ahh. Ok,  I changed up and down to ascending and descending. That probably describes better that we are sorting by value and thedirection?

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_component.ts
Sparkier: Sure, sounds good. Now I think I also finally fully understand the differentiation between runName and runId. 

To clarify for this next PR, after looking into `tensorboard/webapp/runs/store/runs_selectors.ts`, there is no "get metadata for all runs", so I assume I would have to fetch all active runs and then use this:
```
/**
 * Returns Observable that emits run object.
 */
export const getRun = createSelector(
  getDataState,
  (state: RunsDataState, props: {runId: string}): Run | null => {
    return state.runMetadata[props.runId] ?? null;
  }
);
```
to get the run information for all the IDs?

stephanwlee (approved):

PR #4265, opened 2020-10-29
experiments: add `related_links` field

wchargin:
Summary:
Google-internal TensorBoard instances want to show some related links in
the UI for the header card: think, “link to logdir in online filesystem
viewer”, or “link to training job health status”. This patch updates the
types to permit the backend to send down such links.

Test Plan:
That TypeScript is still happy suffices.

wchargin-branch: experiments-related-links

### DIFF
--- a/tensorboard/webapp/experiments/types.ts
+++ b/tensorboard/webapp/experiments/types.ts
@@ -20,4 +20,5 @@ export interface Experiment {
   description?: string;
   hparams?: string;
   tags?: string[];
+  related_links?: Array<{name: string; url: string}>;
 }


tensorboard/webapp/experiments/types.ts
```diff
@@ -20,4 +20,5 @@ export interface Experiment {
   description?: string;
   hparams?: string;
   tags?: string[];
+  related_links?: Array<{name: string; url: string}>;
```
stephanwlee, line 23: nit: while start_time is snake case already, can we use camelCase?

stephanwlee (approved):

tensorboard/webapp/experiments/types.ts
wchargin: I indeed used snake case here to be consistent with `start_time`, since
these responses come directly from the backend (which uses snake case).

Since we don’t have a backend-to-frontend translation, either camel case
or snake case will be inconsistent with the conventions on one end.
Given that, I’d like to keep this consistent with the existing fields,
so that it’s not inconsistent with _both_ ends! If this bothers you, the
right solution imho is to teach either the backend or frontend to
automatically translate from one form to the other.

nfelt: image

wchargin: kebab-case for Racket!

nfelt: I need to start a trend to call names joined with periods `domain.case` - though really it'd be better as `case.domain` to do it right-to-left just to confuse everyone the same way DNS does.

PR #4266, opened 2020-10-29
Show alert on fetch run failures

psybuzz:
The recently introduced Alerts ngrx feature offers an affordance
for modules to register alerts. This registers the `fetchRunsFailed`
action to show a message to the user when runs fail to fetch.

In non-Ivy Angular, the entryComponents is required so that
the component factory will be generated, even though it is loaded
on-demand (by snackbar).

Googlers, see sync cl/339774223


bmd3k (approved):

psybuzz:
Thanks for the review!

Issue #4208, opened 2020-10-01
blurry images in Chrome despite `image-rendering: pixelated`


nfelt:
This tracks a bug that appears under recent versions of Chrome (at least, `85.0.4183.121 (Official Build) (64-bit)` on `macOS Version 10.15.4 (Build 19E287)`).  The bug manifests as transient blurriness in rendering upscaled images in the image dashboard that are deliberately meant to be pixelated; aka, the desired `image-rendering: pixelated` CSS property is not being upheld 100% of the time.

Steps to reproduce:

1. Write image summary data as follows.  The known important parts are A) including 2 steps so that the step slider is shown and B) omitting a description (this is critical since if a description is included, an "info" icon will be rendered in the card header, and this seems to suppress the bug; note that the image demo includes descriptions for all tags so it can't be used to repro the bug).

     ```
    import tensorflow as tf
    with tf.summary.create_file_writer("/tmp/blurry").as_default():
        for step in range(2):
            tf.summary.image("test", tf.expand_dims(tf.eye(16, batch_shape=[1]), axis=-1), step=step)
    ```

2. Launch TensorBoard at the resulting logdir `/tmp/blurry`.  The bug seems longstanding; I checked at least TensorBoard versions 1.15.0, 2.2.1, and current nightly (2.4.0a20200930).

3. Open the images dashboard.  The pixelated image summary should be shown.

4. Resize the browser window.  This part is slightly finicky; the bug seems to repro most reliably if doing this using a grab handle on the window that is not adjacent to the viewport or address bar (e.g. use one of the top corners attached to the OS level window header, rather than the sides or bottom corners).  Sometimes a resize doesn't trigger the bug (e.g. maybe for window widths that are too small), if so, just keep trying.

5. Mouse over the viewport or address bar.  As soon as the cursor enters this region, the pixelated image should suddenly turn blurry.

6. Generally the image will remain blurry until it a repaint is triggered somehow (e.g. by toggling the actual size view, or sliding the step back and forth, by collapsing and re-opening the clickable "category pane", or by resizing the browser window again).  Occasionally, it "snaps back" to being crisp before this though.

I failed to reproduce the bug in the following circumstances:

- If there's only a single step, and no step slider is shown.
- If the tag has a description, i.e. the "info" icon is shown in the card header.
- Using Firefox - it behaves correctly.
[...]

PR #4271, opened 2020-11-03
[time series] render images as pixelated

psybuzz:
Image cards rendered in Time Series may appear blurry, which
can be misleading if the written image is composed of discrete
values-per-pixel. This changes the Time Series image cards to
use the same 'image-rendering' CSS prop as the Image dashboard.

See [plugins/image/tf_image_dashboard/tf-image-loader.ts](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/plugins/image/tf_image_dashboard/tf-image-loader.ts;l=169?q=image-rendering%20pixelated&sq=)

Manually checked in Chrome and Firefox that the image looks
pixelated.

Googlers, see b/32560663, cl/113307097.

(Screenshot of image-rendering: pixelated for above image (below
image is default) in Chrome).
image

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_component.scss
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_component.scss
@@ -113,6 +113,9 @@ $_second-row-height: 15px;
 }
 
 .img-container img {
+  // Avoid the 'smooth' image scaling provided in browsers by default.
+  // Instead, choose rendering modes that are closer to pixel representation.
+  // See b/32560663.
   image-rendering: -moz-crisp-edges;
   image-rendering: pixelated;
 }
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/metrics/views/card_renderer/image_card_component.scss
```diff
@@ -112,6 +112,11 @@ $_second-row-height: 15px;
   position: relative;
 }
 
+.img-container img {
+  image-rendering: -moz-crisp-edges;
```
stephanwlee, line 116: attach a repro case or a logic in the comment so we do not revert this.

stephanwlee (approved):

tensorboard/webapp/metrics/views/card_renderer/image_card_component.scss
psybuzz: Great idea, done.

wchargin:
Cross-referencing #4208, since using the same CSS may mean that we’re
also subject to that bug here. No action required since we don’t have a
fix.

PR #4282, opened 2020-11-06
rust: implement TFRecord CRC masking operation

wchargin:
Summary:
The TFRecord format uses checksums given by applying a specific masking
permutation to a CRC-32C (Castagnoli) checksum. The permutation is to
improve error detection when checksums are embedded in buffers that are
themselves checksummed. This patch implements the TFRecord checksum
logic behind a `MaskedCrc` type.

Test Plan:
Unit tests included.

wchargin-branch: rust-masked-crc


tensorboard/data/server/lib.rs
```diff
@@ -14,3 +14,5 @@ limitations under the License.
 ==============================================================================*/
 
 //! Core functionality for TensorBoard data loading.
+
+pub mod masked_crc;
```
nfelt, line 18: I might be misunderstanding, but do we want `pub` here?  It seems like we'd want `masked_crc` to be a private utility for rustboard-lib rather than part of a public API (were we to imagine publishing it as its own crate).

nfelt (approved):

tensorboard/data/server/lib.rs
wchargin: Fair question, and you understand correctly. I made it `pub` as a
practical matter, because otherwise its contents are flagged as dead
code, since indeed they cannot be used:

```
$ git diff -U0
diff --git a/tensorboard/data/server/lib.rs b/tensorboard/data/server/lib.rs
index f6bf80b01..9da7547ac 100644
--- a/tensorboard/data/server/lib.rs
+++ b/tensorboard/data/server/lib.rs
@@ -18 +18 @@ limitations under the License.
-pub mod masked_crc;
+mod masked_crc;
$ cargo check
warning: constant is never used: `CRC_MASK_DELTA`
  --> masked_crc.rs:37:1
   |
37 | const CRC_MASK_DELTA: u32 = 0xa282ead8;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: function is never used: `mask`
  --> masked_crc.rs:40:4
   |
40 | fn mask(crc: u32) -> MaskedCrc {
   |    ^^^^

warning: associated function is never used: `compute`
  --> masked_crc.rs:57:12
   |
57 |     pub fn compute(bytes: &[u8]) -> Self {
   |            ^^^^^^^

warning: 3 warnings emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
```


We can silence this warning—

```rust
#[allow(dead_code)]
mod masked_crc;
```

—but this would also permit code that’s dead _within_ the module, which
currently would be caught:

```
$ git diff -U1
diff --git a/tensorboard/data/server/masked_crc.rs b/tensorboard/data/server/masked_crc.rs
index eb5bf9d11..a95851843 100644
--- a/tensorboard/data/server/masked_crc.rs
+++ b/tensorboard/data/server/masked_crc.rs
@@ -86 +86,3 @@ mod tests {
 }
+
+fn some_dead_code() {}
$ cargo check
warning: unused manifest key: raze
warning: function is never used: `some_dead_code`
  --> masked_crc.rs:88:4
   |
88 | fn some_dead_code() {}
   |    ^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
```

Once we have a `pub` entry point that’s used in the binary, we can go
back and mark the internals as non-`pub`. I think it’s not a big deal,
since we don’t intend to publish a half-baked crate.

wchargin (commented):
thanks!

tensorboard/data/server/lib.rs
nfelt: Ack, sounds reasonable, thanks for the explanation.

PR #4284, opened 2020-11-06
line chart: refactor to separate internal types

stephanwlee:
There are few types that are used internal to the charting library and
there are few public ones. Since the public one is a subset of the
internal one, we are turning `types`, which is too generic, into
`internal_types` and selectively re-export types in the `public_types`.


tensorboard/webapp/widgets/line_chart_v2/lib/renderer/BUILD
```diff
@@ -25,7 +25,7 @@ tf_ts_library(
         "renderer_types.ts",
```
psybuzz, line 25: Can we visibility restrict this target to lib:subpackages?  `public_types.ts` only re-exports specific symbols from `renderer_types.ts`, suggesting that some of its symbols may be non-public.

tensorboard/webapp/widgets/line_chart_v2/lib/public_types.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export {
+  DataSeries,
+  DataSeriesMetadata,
+  DataSeriesMetadataMap,
+  Dimension,
+  Extent,
+  Point,
+  Rect,
+} from './internal_types';
```
psybuzz, line 24: Looking ahead at [1], I see evidence why most of these need to be public, except for `Rect`.  It seems that `Rect` is useful for anything that deals with a `Coordinator`, which line_chart_component and above should not be exposed to.

Should we drop `Rect`?

[1] https://github.com/tensorflow/tensorboard/blob/swl_gpu_11/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/BUILD
stephanwlee: Done.

tensorboard/webapp/widgets/line_chart_v2/lib/public_types.ts
stephanwlee: Done

psybuzz (approved):

PR #4286, opened 2020-11-06
refactor: runs-table from mat-table to ngFor

stephanwlee:
This change is in preparation for potential virtual scrolling in the
runs table.

About virtual scroll: Run selector can possibly render many runs; in
real user cases, the run selector had to render more than a thousand
runs. While minor in the grand scheme of things, Angular rendering
thousands of elements with nested components can contribute to AFT or
initial meaning paint.

Virtual scroll module does not seemingly support mat-table and it
requires the items to be rendered via cdkVirtualFor. This change, for
now, renders using ngFor which is API compatible with cdkVirtualFor.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
@@ -30,9 +30,13 @@ limitations under the License.
   <div role="table">
     <ng-container *ngTemplateOutlet="header"></ng-container>
     <ng-container *ngTemplateOutlet="selectAll"></ng-container>
-    <ng-container *ngFor="let item of pageItems; trackBy: tableTrackBy">
-      <ng-container *ngTemplateOutlet="row; context:{item: item}"></ng-container
-    ></ng-container>
+    <div role="rowgroup">
+      <ng-container *ngFor="let item of pageItems; trackBy: tableTrackBy">
+        <ng-container
+          *ngTemplateOutlet="row; context:{item: item}"
+        ></ng-container
+      ></ng-container>
+    </div>
   </div>
 
   <div *ngIf="loading" class="loading">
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
```diff
@@ -27,186 +27,125 @@
...
-      <th mat-header-cell *matHeaderCellDef mat-sort-header class="name">
-        Experiment
-      </th>
-      <td mat-cell *matCellDef="let item" class="name">
-        <span [attr.title]="item.experimentName"
-          >{{ item.experimentAlias }}</span
+        <span
+          *ngSwitchCase="RunsTableColumn.EXPERIMENT_NAME"
+          [mat-sort-header]="RunsTableColumn.EXPERIMENT_NAME"
+          [class]="columnId"
```
psybuzz, line 102: ditto: drop class columnId, here and below

tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
```diff
@@ -27,186 +27,125 @@
...
-      <td mat-cell *matCellDef="let item">
-        <mat-checkbox
-          [checked]="item.selected"
-          (change)="onSelectionToggle.emit(item)"
-        ></mat-checkbox>
-      </td>
-    </ng-container>
+    <span
+      *ngFor="let columnId of columns"
+      role="cell"
```
psybuzz, line 87: According to [1], role=cell elements "MUST be nested in a container element with role="row""

[1] https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Cell_Role

psybuzz (commented):
Manually tested this in open source TB, and it looks good

tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
```diff
@@ -27,186 +27,125 @@
...
+      [ngClass]="['column', columnId]"
```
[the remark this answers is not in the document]
stephanwlee: It is used in the existing code. It is used in CSS and it is used in the test. I will _not_ drop it in this PR since DOM structure refactor is the major point, fixing for a potential bug.

tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
stephanwlee: ditto.

tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
```diff
@@ -27,186 +27,125 @@
...
+      </ng-container>
+    </span>
 
-    <ng-container
-      *ngFor="let column of metricColumns; trackBy: trackByMetricColumn"
-      [matColumnDef]="getMetricColumnId(column)"
+    <span
+      *ngFor="let column of hparamColumns; trackBy: trackByHparamColumn"
+      role="cell"
+      [ngClass]="['column', column.name]"
```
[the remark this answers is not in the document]
stephanwlee: ditto

stephanwlee:
Tested using VoiceOver. While it is not ideal, it was not horrible; you can navigate around the table in grid-like fashion. I think it is good enough.

tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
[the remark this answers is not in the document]
stephanwlee: I mis-remembered the code I wrote some time ago. You are right. This `columnId` business is a new addition and is now removed.

tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
```diff
@@ -16,33 +16,28 @@ limitations under the License.
...
+  background-color: #fff;
   display: block;
+  font-size: $_font-size;
   overflow: hidden;
-  background-color: #fff;
 }
 
 .table-container {
   contain: layout paint;
+  height: 100%;
```
psybuzz, line 30: [No action needed necessarily]

Let's make sure that this works internally.  For RunsTableContainers that use `useFlexibleLayout=false` in an experiment list type of view, a runs table's height should grow depending on the number of rows within.

tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
```diff
@@ -51,38 +46,35 @@ $_table-content-row-min-height: 43px;
...
-  tr:not(.select-all) th {
     white-space: nowrap;
   }
 
-  .mat-header-row,
-  .mat-row {
+  .header,
+  .run-row {
+    contain: strict;
+    display: table-row;
```
psybuzz, line 59: No action necessary, just wanted to comment that I'm rather unfamiliar with when to use the `display: table-*` CSS values and the details on how they are rendered/layout-ed differently from `display: block`.  I'd be happy to learn what benefits of using these over block/flex are.

psybuzz (approved):
Thanks for checking with VoiceOver!

tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
stephanwlee: If you see current TensorBoard and see their role and their computed `display` values, you will get some sense of what it should be. This change basically implements `table` without using `table`.

tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
stephanwlee: I did check both internally/externally.

tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
```diff
@@ -148,8 +144,8 @@ table {
 
 // Prevents the table column for checkbox and run_color from growing beyond
 // their sizes.
-.mat-column-checkbox,
-.mat-column-run_color {
```
psybuzz, line 152: Ah, I was not aware that Angular Material automatically created classnames based on the provided column text and attached them.  I now see why we need the `tb-column-*` analogs.

PR #4295, opened 2020-11-09
rust: add `ScriptedReader` test helper

wchargin:
Summary:
This is a [`Read`][r] implementation (i.e., input stream) that takes a
“script” of responses and yields data from at most one at a time. The
standard library has a [`Cursor`][c] type for much the same purpose, but
we need to simulate files that are being continually written.

[r]: https://doc.rust-lang.org/std/io/trait.Read.html
[c]: https://doc.rust-lang.org/std/io/struct.Cursor.html

Test Plan:
Unit tests included; run `bazel test //tensorboard/data/server/...`.

wchargin-branch: rust-scriptedreader

### DIFF
--- a/tensorboard/data/server/scripted_reader.rs
+++ b/tensorboard/data/server/scripted_reader.rs
@@ -1,0 +1,89 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+//! Test helper for simulating input streams that grow over time.
+
+use std::collections::VecDeque;
+use std::io::{self, Cursor, Read};
+
+/// A reader that delegates to a sequence of cursors, reading from each in turn and simulating
+/// EOF after each one.
+pub struct ScriptedReader(VecDeque<Cursor<Vec<u8>>>);
+
+impl ScriptedReader {
+    /// Creates a reader that reads from each of the given buffers in turn. The given iterator is
+    /// exhausted eagerly.
+    pub fn new<I: IntoIterator<Item = Vec<u8>>>(vecs: I) -> Self {
+        ScriptedReader(vecs.into_iter().map(Cursor::new).collect())
+    }
+}
+
+impl Read for ScriptedReader {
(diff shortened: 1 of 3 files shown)


tensorboard/data/server/scripted_reader.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl ScriptedReader {
+    /// Creates a reader that reads from each of the given buffers in turn. The given iterator is
+    /// exhausted eagerly.
+    pub fn new<I: IntoIterator<Item = Vec<u8>>>(vecs: I) -> Self {
+        ScriptedReader(vecs.into_iter().map(Cursor::new).collect())
+    }
+}
+
+impl Read for ScriptedReader {
```
stephanwlee, line 33: Question: I expect `ScriptedReader` to have `read` method since, after all, it is a reader. Is it more idiomatic to break a struct down into semantic functionalities (e.g., `Read`)? Or am I reading this incorrectly?

stephanwlee (approved):
Thanks for the PR. I poked around and learned a bit already. Will look forward to future PRs.

stephanwlee:
I forgot to add 🦀 EVERYWHERE!

tensorboard/data/server/scripted_reader.rs
wchargin: It’s not just idiomatic: It’s The Law™. To implement a trait, you need a
separate `impl SomeTrait for SomeType` block, and that’s where you put
the required methods.

Part of this is because trait methods are namespaced. A `Read` isn’t
just “something that happens to define `read` with the right signature”;
it’s something that defines specifically `Read::read`. This avoids a
problem with Java/TS/...-style interfaces where you have conflicts
between different interfaces that you want to implement, or existing
methods on your type.

So, [the following is legal][1]:

```rust
trait Foo { fn method() -> i32; }
trait Bar { fn method() -> String; }

struct S;

impl Foo for S { fn method() -> i32 { 123 } }
impl Bar for S { fn method() -> String { "hello".to_string() } }

fn main() {
    dbg!(<S as Foo>::method()); // 123
    dbg!(<S as Bar>::method()); // "hello"
}
```

[1]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8e889bfbc74fca98ddd5013c88b586a3

…and while clearly that’s a bit confusing if you control `Foo`, `Bar`,
and `S`, you can see how if any one or more of those is controlled by a
third party, this enables you to implement the interfaces that you want
without _all_ the library authors needing to coordinate about unique
names.

As the Zen of Python says, “namespaces are one honking great idea—let’s
do more of those!” So in Rust, to implement `Read::read`, you need an
`impl Read for ...` block that defines `fn read`.

wchargin (commented):
🦀

tensorboard/data/server/scripted_reader.rs
stephanwlee: Oh! `Read` here is a trait that you imported from `std::io` and is `std::io::Read`. Things started to click in my head! Thanks!

stephanwlee (approved):

wchargin:
Yep! That’s right.

PR #4296, opened 2020-11-09
workspace: upgrade to Bazel 3.0.0, rules_rust latest

wchargin:
Summary:
Bazel rule repositories (Go, Rust, protobuf, …) are migrating to a new
linker API introduced as a breaking Bazel change:
<https://github.com/bazelbuild/bazel/issues/10860>

This API requires Bazel 3.0.0, which is fine for us.

Test Plan:
Everything builds, all tests pass, and TensorBoard looks okay enough,
including the dynamically loaded projector plugin.

wchargin-branch: bazel-3.0.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ env:
   # Grab the BAZEL_SHA256SUM from the Bazel releases page; e.g.:
   # bazel-0.20.0-linux-x86_64.sha256
   global:
-    - BAZEL=2.1.0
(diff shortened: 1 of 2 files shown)


wchargin:
Sent upstream <https://github.com/bazelbuild/rules_rust/pull/484> for CI
error, but @stephanwlee tells me that we need Bazel 3.5.0 anyway, so we
might as well just do that instead rather than as a follow-up. Easier on
CI caches this way, too. Will supersede with #4301.

PR #4297, opened 2020-11-09
rust: add `rand` and `rand_chacha` dependencies

wchargin:
Summary:
The [`rand`] crate is the standard Rust library for random number
generation, and [`rand_chacha`] is a standard CSPRNG. Many applications
could just use `rand::rngs::StdRng`, but reproducibility of the specific
random sequence is somewhat important to TensorBoard, so [we depend on
`rand_chacha` directly, as suggested by `StdRng`’s docs][stdrng].

We set `targets` to a list that excludes `wasm32-wasi` to avoid having
to update our `rules_rust` dep for now, since the Bazel upgrade for that
is looking like it might take a while… [Workaround from @UebelAndre][w];
thanks!

[`rand`]: https://crates.io/crates/rand
[`rand_chacha`]: https://crates.io/crates/rand_chacha
[stdrng]: https://docs.rs/rand/0.7.3/rand/rngs/struct.StdRng.html
[w]: https://github.com/google/cargo-raze/pull/280#issuecomment-724380062

Test Plan:
Both crates build: `bazel build //third_party/rust:{rand,rand_chacha}`.

wchargin-branch: rust-dep-rand


nfelt (approved):
🎲 🎲 💃 🕺 💃 🕺

PR #4299, opened 2020-11-09
rust: support reservoir preemption

wchargin:
Summary:
This patch extends the bipartite reservoir sampling data structure with
support for preemption. A *preemption* occurs when an record is enqueued
with a step that is not higher than the last-seen step. In a preemption,
all records whose step is not lower than the new record’s step are
evicted.

Test Plan:
Unit tests expanded.

wchargin-branch: rust-reservoir-preemption


tensorboard/data/server/reservoir.rs
```diff
@@ -231,6 +233,53 @@ impl<T, C: ReservoirControl> StageReservoir<T, C> {
...
+                self.staged_preemption = true;
+                preempted += committed_preempted;
+            }
+        }
+        if preempted == 0 {
+            return; // No need to adjust `seen`.
+        }
+        let old_stored = self.committed_steps.len() + self.staged_items.len() + preempted;
+        let fac_preempted = (preempted as f64) / (old_stored as f64);
+        self.seen = (self.seen as f64 * (1.0 - fac_preempted)).ceil() as usize;
```
wchargin, line 280: @nfelt: When preempting back to the original step or before, I think
that it actually doesn’t have any rounding error as written.

Since `preempt` is called first in `offer`, before even incrementing
`seen`, when you preempt to step 0 we get `fac_preempted == 1.0` exactly
(since `preempted == old_stored`), and thus set `self.seen = 0`.

This test passes:

```rust
    #[test]
    fn test_preempt_long() {
        let mut rsv = StageReservoir::new(10);
        let mut head = Basin::new();
        for i in 0..100_000 {
            rsv.offer(Step(i), ());
        }
        rsv.commit(&mut head);
        // head.as_slice()[0] happens to be step 25630 here
        assert_eq!(rsv.seen, 100_000);
        rsv.offer(Step(0), ());
        assert_eq!(rsv.seen, 1); // exactly correct: just the newest record
        for i in 1..9 {
            rsv.offer(Step(i), ());
        }
        rsv.commit(&mut head);
        assert_eq!(
            head.as_slice()
                .iter()
                .map(|&(Step(s), _)| s)
                .collect::<Vec<i64>>(),
            (0..9).collect::<Vec<i64>>()
        );
        assert_eq!(rsv.seen, 9); // exactly correct: just 0 and 1..9
    }
```

Supposing that we include a test something like this, does it seem
reasonable to you to keep the code as is, without special casing for
preempting back to start?

tensorboard/data/server/reservoir.rs
```diff
@@ -234,6 +236,53 @@ impl<T, C: ReservoirControl> StageReservoir<T, C> {
...
+    }
+
+    /// Preempts any records whose step does not precede the given step.
+    fn preempt(&mut self, step: Step) {
+        let staged_nonpreempted = self
+            .staged_items
+            .iter()
+            .enumerate()
+            .rev()
+            .find_map(|(i, (s, _))| if *s < step { Some(i + 1) } else { None })
```
nfelt, line 254: I struggled a little to grok the iterator chain here - some of that is my having to look up APIs like `find_map()` but then it also just felt a bit convoluted to have this predicate checking for non-preempted items in order to count those (but not by directly counting them per se; instead by returning the enumerated index), all in order to subtract those from the length and then truncate accordingly.

Is there a reason not to just use `binary_search_by_key()` which looks like it would directly return the truncation point?  I'm not sure if I'm missing some reason that doesn't work, or if there's a performance concern about binary search or something else.

If we did want to linearly scan from the end of the vector, having looked at the Iter APIs for a while, what seems more "natural" to me would be:

```rust
let staged_preempted = self.staged_items.iter().rev().take_while(|(s, _)| s >= step).count();
```

Is there a reason the version you have is preferable?  (Honest question.)

tensorboard/data/server/reservoir.rs
```diff
@@ -234,6 +236,55 @@ impl<T, C: ReservoirControl> StageReservoir<T, C> {
...
+            if committed_preempted > 0 {
+                self.committed_steps.truncate(committed_nonpreempted);
+                self.staged_preemption = true;
+                preempted += committed_preempted;
+            }
+        }
+        if preempted == 0 {
+            return; // No need to adjust `seen`.
+        }
+        let old_stored = self.committed_steps.len() + self.staged_items.len() + preempted;
```
nfelt, line 281: Seems a little more self-evident to just do `let old_stored = self.len()` at the top of the function?  Then I think we might also not have to do the bookkeeping to track `preempted` at all.

tensorboard/data/server/reservoir.rs
```diff
@@ -413,13 +518,47 @@ mod tests {
...
                 assert_eq!(head.as_slice(), &[]);
             }
         }
     }
+
+    #[test]
+    fn test_preempt_to_start() {
+        let mut rsv = StageReservoir::new(100);
+        let mut head = Basin::new();
+        for i in 0..100_000 {
```
nfelt, line 539: I gather that this loop over 100,000 random samples is fast enough to not really be noticeable when running tests?  (I assume the point of going all the way up to 100,000 is to minimize the chance of the test passing by accident if we hadn't reset `self.seen` all the way to 0.)

tensorboard/data/server/reservoir.rs
```diff
@@ -307,52 +357,80 @@ mod tests {
         rsv.offer(Step(0), "zero");
         rsv.offer(Step(1), "one");
         rsv.offer(Step(2), "two");
-        rsv.offer(Step(3), "three");
+        rsv.offer(Step(3), "three???"); // this funny-looking record will be preempted
```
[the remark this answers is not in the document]
wchargin: I hadn’t considered that, only because I originally wrote the reservoir
with preemption from the beginning, and wrote the test all at once. But
on revisiting it I agree with your assessment. I also think that the
extensions to the property-based test are pretty reasonable to cover
normal cases of preemption, and we now have extra tests for more edge
cases. So I’ve gone and just stripped the preemption from this test.
Work for you?

wchargin (commented):
Thanks much! It is better now.

tensorboard/data/server/reservoir.rs
nfelt: Yeah I think it's the "citation needed" part where I'm a bit unsure.  I agree it's a bit more obvious to think of preempting only a few trailing events, but in practice I think I've also seen jobs that preempt back to 0 several times (think of a job failing repeatedly to even get to the first checkpoint) so I figured binary search avoids the worst case, especially for large reservoirs.  But I'm 100% happy to wait until further along and benchmark against some actual data.

nfelt (approved):
Nice!

tensorboard/data/server/reservoir.rs
nfelt: (And yes, it is very cool that the compiler can optimize through all those layers!)

tensorboard/data/server/reservoir.rs
```diff
@@ -119,6 +119,12 @@ impl<T> Basin<T> {
     pub fn as_slice(&self) -> &[(Step, T)] {
         &self.0[..]
     }
+
+    /// Extracts the steps in this basin. Convenient for tests.
+    #[cfg(test)]
```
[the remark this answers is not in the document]
wchargin: Okay; I think you’re right that it’s more cleanly separated to provide a
test-only helper. I liked the method syntax, and we could still get that
same API with an extension trait—

```diff
@@ -362,6 +356,17 @@ mod tests {
         }
     }
 
+    /// Convenience extension trait to extract the steps from a basin.
+    trait Steps {
+        fn steps(&self) -> Vec<Step>;
+    }
+
+    impl<T> Steps for Basin<T> {
+        fn steps(&self) -> Vec<Step> {
+            self.0.iter().map(|(s, _)| *s).collect()
+        }
+    }
+
     #[test]
     fn test_sampling_no_preemptions() {
         let mut rsv = StageReservoir::with_control(7, ScriptedControl::new());
```

—but I don’t actually use it in any fluent chains, so it’s probably
clearer still to just provide a helper function, `steps(&head)`. Done.

In cases where this isn’t possible, I like `#[cfg(test)]` more than the
JavaScript/Python `pub fn _TEST_ONLY_foo` pattern. But maybe best of all
is a simple [`pub(crate)`]. No strong feelings here.

[`pub(crate)`]: https://doc.rust-lang.org/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself

tensorboard/data/server/reservoir.rs
```diff
@@ -307,52 +357,80 @@ mod tests {
...
             ],
         );
 
-        rsv.ctl.extend(vec![3, 7, 6]);
-        rsv.offer(Step(8), "eight"); // evict index 3 (now "four")
-        rsv.offer(Step(9), "nine"); // 7 >= 7, so drop (evict most recent)
-        rsv.offer(Step(10), "ten"); // evict index 6 (now "nine")
+        rsv.ctl.extend(vec![4]);
+        rsv.offer(Step(6), "six");
+        rsv.ctl.extend(vec![1]); // the first that matters: evict step 1
```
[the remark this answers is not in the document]
wchargin: Cool; glad that this sits well with you. I have probably gone a bit
overboard with “proofs in code” in the past (though not on TensorBoard
that I can think of), so the second opinion is welcome.

PR #4306, opened 2020-11-10
rust: add TFRecord reading module

wchargin:
Summary:
We implement a module for reading a file that contains a growing stream
of TFRecords. The record contents are opaque; we don’t require that they
be protos or anything else. The module can detect a partially written
record and wait until more data is written to the file.

Test Plan:
Unit tests included.

wchargin-branch: rust-tf-record


tensorboard/data/server/tf_record.rs
```diff
@@ -0,0 +1,327 @@
...
+        let data_crc = MaskedCrc(LittleEndian::read_u32(&data_crc_buf));
+        self.header.clear(); // reset; caller may use this again
+        Ok(TfRecord { data, data_crc })
+    }
+}
+
+/// Fills `buf`'s remaining capacity from `reader`, or fails with `Truncated` if the reader is dry.
+fn read_remaining<R: Read>(reader: &mut R, buf: &mut Vec<u8>) -> Result<(), ReadRecordError> {
+    let want = buf.capacity() - buf.len();
+    reader.take(want as u64).read_to_end(buf)?;
```
wchargin, line 206: Here’s the `?` operator again. This time, `.read_to_end(buf)` is
a `Result<_, io::Error>`, but note that the enclosing function returns a
`Result<_, ReadRecordError>`—the types don’t line up. So the `?` on this
line is equivalent to:

```rust
match reader.take(want as u64).read_to_end(buf) {
    Err(io_error) => Err(ReadRecordError::from(io_error)),
    Ok(_) => (),
}
```

This is what the `impl From<io::Error> for ReadRecordError` above is
for: it describes how to convert an `io::Error` to our own error type,
so that we can use the `?` operator here.

wchargin (commented):
96 lines of code, 120 lines of test, 76 lines of docs.

(I half-want a bot/UI to automatically show this diffstat breakdown…)

Left a few comments inline to discuss some things that you might wonder
about, and maybe to prompt further questions.

tensorboard/data/server/tf_record.rs
stephanwlee: Interesting. I tend to think explicit statements (in this case, `::from` call) to be a good thing but I guess I will grow to like this.

tensorboard/data/server/tf_record.rs
```diff
@@ -0,0 +1,327 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::scripted_reader::ScriptedReader;
+    use std::io::Cursor;
+
+    #[test]
+    /// Tests a happy path with multiple records, one of which is truncated.
+    fn test_success() {
```
stephanwlee, line 221: Is it hypothetically possible that the writer is so borked that it writes more data than alleged length? If so, what would be a possible recovery mechanism? I cannot think of one so answer can be, writer **should never** be borked and, when it is, we will fail hard.

stephanwlee (approved):
Questions are for my own benefit and are not about your code.

tensorboard/data/server/tf_record.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// buf.extend(b"\x18\x00\x00\x00\x00\x00\x00\x00"); // length: 24 bytes
+    /// buf.extend(b"\xa3\x7f\x4b\x22"); // length checksum (0x224b7fa3)
+    /// let contents = b"\x09\x00\x00\x80\x38\x99\xd6\xd7\x41\x1a\x0dbrain.Event:2";
+    /// buf.extend(&contents[..5]); // file truncated mid-write
+    ///
+    /// let mut state = TfRecordState::new();
+    ///
+    /// // First attempt: read what we can, then encounter truncation.
+    /// assert!(matches!(
+    ///     state.read_record(&mut Cursor::new(buf)),
```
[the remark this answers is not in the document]
wchargin: > I must confess that I am not used to thinking data first; I find Rust
> particularly interesting since you think data struct first then
> traits, then method/impls later (exact ordering is not important but
> emphasis on thinking about type flowing before class in OOP).

Definitely. This is a good description.

> Now, I find it weird for RecordState to be passed with a buffer that
> changes its internal state. Since I am definitely new to this, I would
> value your opinion/code a lot more, but I just want to make sure why
> it is wrong to have, for example, something like TfRecordReader
> instead. Is current code more natural when thinking the data/state
> first?

This is also acceptable! A `TfRecordReader` could take an underlying
file reader as an argument and own it, and then just expose a
`read_record(&mut self)` method that calls the underlying reader.

For instance, `BufReader` in the standard library does this:
<https://doc.rust-lang.org/std/io/struct.BufReader.html#method.new>

If you let a `TfRecordReader` own and wrap an underlying reader, then
you have an additional guarantee: nothing else can access the underlying
reader anymore. For instance, in Java, you could write this:

```java
FileReader fr = new FileReader("/etc/hosts");
BufferedReader br = new BufferedReader(br);
br.readLine();
fr.read();  // poke into the underlying reader!
br.readLine();  // undefined behavior, probably! def. not what you want
```

…but in Rust [that’s a borrow-checker error][1]:

```rust
use std::fs::File;
use std::io::{BufRead, BufReader, Read};

fn main() -> std::io::Result<()> {
    let mut fr: File = File::open("/etc/hosts")?;
    let mut br: BufReader<File> = BufReader::new(fr);
    let mut line = String::new();
    br.read_line(&mut line)?;
    dbg!(line);
    fr.read(&mut [])?;
    Ok(())
}
```

```
error[E0382]: borrow of moved value: `fr`
  --> src/main.rs:10:5
   |
5  |     let mut fr: File = File::open("/etc/hosts")?;
   |         ------ move occurs because `fr` has type `std::fs::File`, which does not implement the `Copy` trait
6  |     let mut br: BufReader<File> = BufReader::new(fr);
   |                                                  -- value moved here
...
10 |     fr.read(&mut [])?;
   |     ^^ value borrowed here after move
```

For a `BufReader`, this is definitely a good thing, since it keeps an
internal buffer of extra data read from the file, and reading directly
from the underlying stream would break its invariants. What about for a
`TfRecordReader`? It feels like the same thing is true: if you just read
ahead in the file, then our bookkeeping about where in the record layout
we are will be incorrect. Preventing that is nice.

So, sure! Let’s let a `TfRecordReader` own an underlying reader.

There’s one place in our code where we _do_ want to poke the underlying
reader, though—in the documentation example for `read_record`. See: we
call `read_record(&mut Cursor::new(buf))`, then we change `buf`, then we
call it again with `&mut Cursor::new(buf)`, which is a different reader.
So one approach would be to expose

```rust
impl<R> TfRecordReader<R> {
    /// Borrows the underlying reader.
    pub fn as_inner(&mut self) -> &mut R {
        &mut self.reader
    }
}
```

which still lets our reader own the underlying reader but enables
callers to borrow it. Of course, if they borrow it, they can muck with
its state again, so we lose some of the guarantee that we’ve gained. So
instead what I’ve done is work around it a bit in the example, by
defining a reader type that actually does grow, just like a file does.
I’ve used [leading `#` comments][dt] to hide this part from the docs, so
that it still powers the doctest but doesn’t distract readers. You can
check out the doc server, if you like.

What do you think?

[1]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bd57b5c90cd06bcc697cf355698b20cd
[dt]: https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example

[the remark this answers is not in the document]
wchargin: > I find Rust particularly interesting since you think data struct first
> then traits, then method/impls later

You may also be interested in [this koan, “Behavioural modeling”][bm]:
<https://users.rust-lang.org/t/rust-koans/2408/2>

I didn’t fully understand it the first time I read it. Later, while
doing something entirely unrelated (in JavaScript!), I was enlightened.

[bm]: https://users.rust-lang.org/t/rust-koans/2408/2

tensorboard/data/server/tf_record.rs
```diff
@@ -0,0 +1,327 @@
...
+==============================================================================*/
+
+//! Resumable reading for TFRecord streams.
+
+use byteorder::{ByteOrder, LittleEndian};
+use std::io::{self, Read};
+
+use crate::masked_crc::MaskedCrc;
+
+// From TensorFlow `record_writer.cc` comments:
```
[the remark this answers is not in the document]
wchargin: Sure. Thanks for finding the link; I’ve copied around this comment block
through four iterations of this project and haven’t read the original cc source
in a while. :-)

wchargin (commented):
Thanks very much! Took your suggestion about a `TfRecordReader`. See
discussion below. Rust is not an OO language, but sometimes OO-familiar
practices can work well—and same for s/OO/functional/.

tensorboard/data/server/tf_record.rs
stephanwlee: Yup, that's what I was asking. Thanks for the response.

I assumed you can craft the data so that it contains correct "header" with very large length (with matching length CRC). In this case, since we forbid allocating memory larger than 2 GiB, the impact is minimal and will not cause our program to allocate 64 GiB and will not cause resource starvation.  LGTM!

stephanwlee (approved):

PR #4314, opened 2020-11-10
data: add `DataProvider` gRPC service spec

wchargin:
Summary:
This patch adds a gRPC interface that mirrors the `DataProvider` Python
interface. The intent is to define a `GrpcDataProvider` Python type that
is backed by a gRPC service.

The particular service definition here is what I’ve used in drafts of
data provider implementations [in Go][go] and [in Rust][rust], so I can
attest that it’s at least mostly reasonable on both the client and the
server. It’s mostly a one-to-one port of the Python interface and admits
an easy Python client.

[go]: https://github.com/wchargin/tensorboard-data-server
[rust]: https://github.com/wchargin/rustboard-server

Test Plan:
It builds: `bazel build //tensorboard/data/proto:protos_all_py_pb2_grpc`.

wchargin-branch: data-provider-proto


wchargin:
I’ve included proto support for all three data classes here, but I’m
also happy to strip it back to just scalars to start. If you’d prefer
that, let me know.

tensorboard/data/proto/data_provider.proto
```diff
@@ -0,0 +1,295 @@
...
+}
+
+message TagFilter {
+  // Only match tags with exactly one of these names. In particular, if this
+  // list is empty, no tags match.
+  repeated string tags = 1;
+}
+
+message ScalarMetadata {
+  // Largest step value of any datum in this time series. Non-negative.
```
nfelt, line 97: Will we actually enforce that this is non-negative?  AFAICT we use int64 everywhere so it certainly can be negative on the wire; if a time series only has negative steps, does that mean we return 0 here?

tensorboard/data/proto/data_provider.proto
```diff
@@ -0,0 +1,295 @@
...
+    string run_name = 1;
+    repeated TagEntry tags = 2;
+  }
+  message TagEntry {
+    string tag_name = 1;
+    ScalarMetadata time_series = 2;
+  }
+}
+
+message PluginFilter {
```
nfelt, line 72: IMO better to put all the submessages that apply to multiple dataclasses (i.e. the filters and `Downsample`) either above or below all the dataclass-specific messages (like Request/Responses or the metadata and data messages).

Right now they're sort of interleaved among the scalar List and Read calls which is slightly confusing.

tensorboard/data/proto/data_provider.proto
```diff
@@ -0,0 +1,295 @@
...
+  message TagEntry {
+    string tag_name = 1;
+    ScalarData data = 2;
+  }
+}
+
+// A column-major sequence of scalar points. Arrays `step`, `wall_time`, and
+// `value` have the same lengths.
+message ScalarData {
+  repeated int64 step = 1 [packed = true];
```
nfelt, line 139: In proto3 `packed=true` is on by default, so this isn't really needed: https://developers.google.com/protocol-buffers/docs/encoding#packed

tensorboard/data/proto/data_provider.proto
```diff
@@ -0,0 +1,295 @@
...
+}
+
+message ReadTensorsRequest {
+  // ID of experiment in which to query data.
+  string experiment_id = 1;
+  // Required filter for plugin name. If omitted, an empty message is implied.
+  PluginFilter plugin_filter = 2;
+  // Optional filter for time series. If omitted, all time series match.
+  RunTagFilter run_tag_filter = 3;
+  // Required downsampling specification describing how many points to return
```
nfelt, line 181: Again, it's a bit confusing to me to describe this as "Required" and then specify what happens if it's omitted to be something other than returning an error that the request is invalid.

tensorboard/data/proto/data_provider.proto
```diff
@@ -0,0 +1,295 @@
...
+    string run_name = 1;
+    repeated TagEntry tags = 2;
+  }
+  message TagEntry {
+    string tag_name = 1;
+    TensorData data = 2;
+  }
+}
+
+// A column-major sequence of tensor points. Arrays `step`, `wall_time`, and
```
nfelt, line 199: I assume the main motivation for doing this column-major unlike TB.dev is compactness of the wire representation?  Or was there another motivation?

I imagine it's significant space overhead for scalars where each point is very small, but it seems less significant for tensors and blobs, so I'd be a bit curious to hear some reasoning for this, since it does have the downside of not guarding against having arrays of different lengths.

nfelt (commented):
Looking pretty good, just have some fairly low-level questions.

wchargin (commented):
Thanks! Took most of your suggestions wholesale; left the others
unresolved for easy reading (but replied to all).

tensorboard/data/proto/data_provider.proto
nfelt: I'm moderately opposed to leaving it in, since I feel like it'll create confusion regarding its effect (namely, none, for proto3 files, and all of our proto files in OSS are proto3).  If some repeated varint fields are labeled and some aren't, it could mislead the reader into assuming there's a functional difference or at least some meaningful difference in intent.  Of course we could resolve that by _requiring_ the annotation everywhere, but that seems overzealous given it has no functional effect, and surveying our existing code, the only places that use it are a handful of older TF protos (newer TF protos generally don't seem to use it):

https://cs.opensource.google/search?ss=tensorflow%2Ftensorboard&q=case:yes%20lang:proto%20repeated%5C%20%5Ba-z%5D

tensorboard/data/proto/data_provider.proto
nfelt: It's fine as-is, I was just curious.  I forgot `zip()` can do more than 2 iterables which indeed makes it pretty straightforward to reconstruct.

nfelt (approved):

tensorboard/data/proto/data_provider.proto
```diff
@@ -0,0 +1,295 @@
...
+  string name = 2;
+  // Wall time of earliest recorded event, as floating-point seconds since
+  // epoch (same as event file format).
+  double start_time = 3;
+}
+
+message ListScalarsRequest {
+  // ID of experiment in which to query data.
+  string experiment_id = 1;
+  // Required filter for plugin name. If omitted, an empty message is implied.
```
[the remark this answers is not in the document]
wchargin: Agh… I’m so torn. I agree, and have been burned by this myself when
using `grpc_cli` to send requests (to TensorBoard.dev, which has the
same semantics) without specifying `downsample` and wondering why the
response was empty. But I’ve also been burned so many times by systems
that treat zero as not a number when there’s clearly a continuous
interpretation of the specification.

I’m happy to make `plugin_name` an error, since requiring that plugin
names not be empty seems perfectly reasonable. We already require that
_dashboard_ names not be empty, which is related, though distinct. But
empty and non-empty strings differ in relevant ways (e.g., disappearing
from path components, not making inductive progress) that zero and
positive integers don’t. And it’s so easy to imagine user flows that
_do_ want to set `downsample { num_points: 0 }`:

> **TensorBoard Export-Your-Data™ Wizard**
>
> How many points do you want from each of these time series?
>
>   - *accuracy*: `1000`
>   - *gan_output/image/0*: `10`
>   - *layer0/weights*: `0` &#x2003;**💥 kaboom!**
>
> <kbd>Download Archive</kbd>

I’ll break the tie by deferring to your intuition and making both cases
an error, but I would be happy to change `downsample` back to continuous
if we change our minds.

tensorboard/data/proto/data_provider.proto
wchargin: Cool.

wchargin (commented):
Thanks for the review!

tensorboard/data/proto/data_provider.proto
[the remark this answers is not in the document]
nfelt: Actually I wasn't suggesting that we make `downsample { num_points: 0 }` an error; I was suggesting only that we make the absence of `downsample` entirely an error (sorry if this was ambiguous).  So code that populates the `Downsample` message has no discontinuity at zero; it's only code that doesn't set `Downsample` entirely.  That takes care of the `grpc_cli` case.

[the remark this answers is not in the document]
wchargin: Ah, got it. Yeah, that sounds pretty reasonable. I generally like the
semantics of “omitted messages are implied to be empty”, but I’m happy
to deviate from it where required or sensible (e.g., `RunTagFilter`),
and this seems like such an occasion. Sent you #4333.

PR #4318, opened 2020-11-11
rust: add demo Tonic server

wchargin:
Summary:
This patch defines a simple “demo” service with one RPC, which adds a
sequence of numbers. It includes a Tonic server for this service to
demonstrate the end-to-end setup.

Test Plan:
In one shell, run `bazel run -c opt //tensorboard/data/server`. In
another shell, use `grpc_cli` to send an RPC to localhost port 6789:

```
grpc_cli --channel_creds_type=insecure \
    --protofiles tensorboard/data/server/demo.proto \
    call localhost:6789 Demo.Add "term: 2 term: 2"
```

This should print a response like `sum: 4`. On my machine, it completes
in 5.2 ± 2.6 ms (methodology: run `hyperfine` on the above command).
This seems reasonably fast given that it has to establish a connection,
whereas a Python gRPC client will keep a connection open. It’s also well
below the 40ms magic number for `TCP_NODELAY` issues.

wchargin-branch: rust-demo-tonic-server


tensorboard/data/server/gen_protos_tool.rs
```diff
@@ -28,6 +28,7 @@ fn main() -> std::io::Result<()> {
     };
     tonic_build::configure()
         .out_dir(&out_dir)
+        .format(false) // don't run `rustfmt`; shouldn't be needed to build
```
stephanwlee, line 31: ah :) I knew something was weird about the generated file

stephanwlee (approved):

tensorboard/data/server/main.rs
```diff
@@ -13,4 +13,38 @@ See the License for the specific language governing permissions and
...
+            ..Default::default()
+        };
+        Ok(Response::new(response))
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let addr = "[::0]:6789".parse::<std::net::SocketAddr>()?;
+    let handler = DemoHandler::default();
```
stephanwlee, line 44: One part that is iffy in my head is, how does it know to invoke `async fn add` for the `Demo.Add` service method?

tensorboard/data/server/gen_protos_tool.rs
wchargin: Yeah—I had `format(true)` (the default) originally, but that only worked
on my machine because I have `rustfmt` installed; it failed on Travis.
It would be nice for it to be formatted. We could format it pretty
easily if bazelbuild/rules_rust#388 were resolved, or we could copy what
bazelbuild/rules_rust#479 does and do it the hard way. I figured that to
start we could just accept the ugly generated code: we have the Rustdoc
server for normal API usage questions, and you can always run `rustfmt`
on it locally if you want to look at the implementation.

PR #4324, opened 2020-11-12
ci: don’t update Cargo lockfile

wchargin:
Summary:
I didn’t realize that this also updates locks for transitive deps, which
can cause the Raze check to fail as it pulls in new versions.

Test Plan:
At master, running `cargo generate-lockfile` is *not* clean, even though
it passed at submit time, because `tokio` has been updated. Running
`cargo fetch` does not change the lockfile, but does add to it if you
add a new dep (e.g., `anyhow = "1.0.33"`).

wchargin-branch: ci-no-generate-lockfile

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
       - name: 'Check cargo-raze freshness'
         run: |
           rm -rf third_party/rust/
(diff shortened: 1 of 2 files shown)


stephanwlee (approved):

wchargin:
Thanks! and sorry for the Travis delays.

PR #4325, opened 2020-11-13
refactor: feature flag typing remove key-value

stephanwlee:
Key-value allows just about anything and it is not that great for
typing. This change removes the key-value part of the FeatureFlag and
forces all apps to strongly type it. We can reintroduce it when we
absolutely need it.

### DIFF
--- a/tensorboard/webapp/feature_flag/effects/feature_flag_effects_test.ts
+++ b/tensorboard/webapp/feature_flag/effects/feature_flag_effects_test.ts
@@ -62,6 +62,7 @@ describe('feature_flag_effects', () => {
     it('loads features from the data source on init', () => {
       spyOn(dataSource, 'getFeatures').and.returnValue({
         enabledExperimentalPlugins: ['foo', 'bar'],
+        inColab: false,
       });
 
       actions.next(effects.ngrxOnInitEffects());
@@ -70,6 +71,7 @@ describe('feature_flag_effects', () => {
         featuresLoaded({
           features: {
             enabledExperimentalPlugins: ['foo', 'bar'],
+            inColab: false,
           },
         }),
       ]);
--- a/tensorboard/webapp/feature_flag/testing.ts
+++ b/tensorboard/webapp/feature_flag/testing.ts
@@ -1,4 +1,4 @@
-/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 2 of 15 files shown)


tensorboard/webapp/feature_flag/effects/feature_flag_effects_test.ts
```diff
@@ -62,6 +62,7 @@ describe('feature_flag_effects', () => {
     it('loads features from the data source on init', () => {
       spyOn(dataSource, 'getFeatures').and.returnValue({
         enabledExperimentalPlugins: ['foo', 'bar'],
+        inColab: false,
```
psybuzz, line 65: Perhaps we can use `buildFeatureFlags({enabledExperimentalPlugins: ['foo', 'bar']})` and drop the `inColab` here and below?

psybuzz (commented):
Thanks for making typing safer :)

I sprinkled a few comments with the purpose of reducing object literals
that require 'all feature flags as keys' into as few files as possible.
It would be great if, when we need to introduce 1 more feature flag,
we would only have to touch `feature_flag/types.ts` and no other file
under `feature_flag/`.

tensorboard/webapp/feature_flag/testing.ts
```diff
@@ -0,0 +1,26 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+import {FeatureFlags} from './types';
+
+export function buildFeatureFlag(
```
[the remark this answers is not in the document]
stephanwlee: I want types to be exclusively about the types, not for test utils.

tensorboard/webapp/feature_flag/effects/feature_flag_effects_test.ts
stephanwlee: I will not remove it.

I mean, I am not going to pass every single possible thing but this test spec is neither specifically about the `enabledExperimentalPlugins` or `inColab`. However, exercising both is a valid test imo.

tensorboard/webapp/feature_flag/testing.ts
```diff
@@ -0,0 +1,26 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
psybuzz, line 1: nit: 2020

tensorboard/webapp/feature_flag/effects/feature_flag_effects_test.ts
psybuzz: This list of files that require all feature flags (selectors.ts, effects_test.ts, testing.ts, types.ts) is small enough that the PR as-is looks reasonable to me.

psybuzz (approved):

PR #4326, opened 2020-11-13
feat: add an ability to enable gpu line chart

stephanwlee:
This change adds a new unconnected flag `fastChart`.  The flag is intended to used for enabling
new GPU based line chart in time series in the future.

### DIFF
--- a/tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
+++ b/tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
@@ -38,6 +38,7 @@ describe('tb_feature_flag_data_source', () => {
         expect(dataSource.getFeatures()).toEqual({
           enabledExperimentalPlugins: ['a', 'b'],
           inColab: false,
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
```diff
@@ -38,6 +38,7 @@ describe('tb_feature_flag_data_source', () => {
         expect(dataSource.getFeatures()).toEqual({
           enabledExperimentalPlugins: ['a', 'b'],
           inColab: false,
+          enableGpuChart: false,
```
psybuzz, line 41: Can we reduce boilerplate by using `buildFeatureFlags` in these tests?  Or perhaps be more explicit with what we're expect'ing on, as shown in the newly added unit test?

If we flip a feature flag's default from false to true, or introduce 1 new flag, I wouldn't want to update every single unit test in this file.

psybuzz (approved):

PR #4328, opened 2020-11-13
infra: persist experimentalPlugin to URL

stephanwlee:
This change reads the state from the store and use our DeepLinkProvider
to add the query parameter to the URL.

Note that no navigation will cause the state to change since we want to
keep it immutable.


psybuzz (approved):
A very readable PR, lgtm!

PR #4331, opened 2020-11-14
rust: build `tensorboard.data` protos

wchargin:
Summary:
In #4314, we added a proto definition for the data provider interface.
We now generate the corresponding Rust bindings.

Generated with `bazel run //tensorboard/data/server:update_protos`.

Test Plan:
Run the `:rustboard_core_doc_server`, and note that the `proto` package
has bindings for both the normal messages and the gRPC service.

wchargin-branch: rust-data-protos

### DIFF
--- a/tensorboard/data/server/BUILD
+++ b/tensorboard/data/server/BUILD
@@ -9,6 +9,7 @@ licenses(["notice"])
 _proto_packages = [
     "demo",
     "tensorboard",
+    "tensorboard.data",
 ]
 
 # Generated files with Rust protobuf bindings. These only exist in the build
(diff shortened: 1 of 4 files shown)


tensorboard/data/server/BUILD
```diff
@@ -85,6 +86,7 @@ genrule(
     srcs = [
```
psybuzz, line 86: To double check, humans never need to manually run this `gen_protos` genrule; `update_protos` is the only target in this folder that we ever need to manually `bazel run`?

psybuzz (approved):

tensorboard/data/server/BUILD
wchargin: Right. I think that’s true in general—it should never be necessary to
run a genrule, since all a genrule does is build its outputs, and those
will be built automatically if they’re transitively required.

Issue #1929, opened 2019-03-01
Graph visualization failed: GraphDefs failed to reconcile.


stephanwlee:
In TensorFlow v2, below code can cause GraphDef reconciliation error.

```py
@tf.function
def foo(x):
  return x ** 2

with writer.as_default():
  tf.summary.trace_on()
  foo(1)
  foo(2)
  tf.summary.trace_export("foo")
```

Depending on the argument, `tf.function` (really, auto-graph) creates ops that are unique within GraphDef but is not globally unique. In the example above, two GraphDefs (on from `foo(1)` and another from `foo(2)`) will be written out and they can collide badly in names and content.

In such case, instead of showing wrong graph content, TensorBoard throws an error.

n-chennakeshava:
I get the same error when I have multiple "@tf.function". I am  working on a distributed learning project across multiple GPUs. I have one @tf.function for the train loop, and another for the test loop. 

with strategy.scope():
    @tf.function
    def distributed_train_step(dataset_inputs):
          (...)
    @tf.function
    def distributed_test_step(dataset_inputs):
          (...)

     stamp = datetime.now().strftime("%Y%m%d-%H%M%S")
     logdir = 'logs/func/%s' % stamp
     writer = tf.summary.create_file_writer(logdir)

.
.
.
.

with writer.as_default():
       tf.summary.trace_export(name="my_func_trace",step=0)

stephanwlee:
How are you invoking your `tf.function`s? Is it writing to the same writer? If so, this is working as intended. Two tf.functions have graphdefs which may have the same node name but of different type/metadata.

luvwinnie:
@stephanwlee  If we use multiple Gpus which have train_step and test_step tf.functions, how should we resolve this problem? I am facing the same problem which shows it has below errors.

```
Traceback (most recent call last):
  File "/home/usr1/.virtualenvs/tensorflow-2.1/lib/python3.7/site-packages/tensorboard/plugins/graph/graph_util.py", line 118, in combine_graph_defs
    lambda n: n.name)
  File "/home/usr1/.virtualenvs/tensorflow-2.1/lib/python3.7/site-packages/tensorboard/plugins/graph/graph_util.py", line 85, in _safe_copy_proto_list_values
    raise _SameKeyDiffContentError(key)
tensorboard.plugins.graph.graph_util._SameKeyDiffContentError: sparse_categorical_crossentropy/Shape

During handling of the above exception, another exception occurred:

[...]

PR #4334, opened 2020-11-17
feat: combine graph by prefixing with unique name

stephanwlee:
Previously, graph plugin combined multiple graphs traced by creating one
monolith of a GraphDef. In doing so, we checked whether, for example,
node names are unique to detect a case when our graph vis can result in
faulty UI.

To alleviate the poor UX, we decided, instead, to duplicate all nodes in
one giant GraphDef container  prefixing all names. While this creates
some bloat, (1) users should see the confusing error less and (2)
combined graphs make it very clear that we have traced multiple graphs.

image

### DIFF
--- a/tensorboard/plugins/graph/graph_util.py
+++ b/tensorboard/plugins/graph/graph_util.py
@@ -22,14 +22,26 @@ def _prefixed_op_name(prefix, op_name):
 
 
 def _prefixed_func_name(prefix, func_name):
-    # TODO(stephanwlee): add business logic to strip "__inference_".
+    """Returns function name prefixed with `prefix`.
+
+    For function libraries, which are often created out of autographed Python
+    function, are factored out in the graph vis. They are grouped under a
+    function name which often has a shape of
+    `__inference_[py_func_name]_[numeric_suffix]`.
+
+    While it does not have some unique information about which graph it is from,
+    creating another wrapping structure with graph prefix and "/" is less than
+    ideal so we join the prefix and func_name using underscore.
+
+    TODO(stephanwlee): add business logic to strip "__inference_" for more user
+    friendlier name
+    """
     return "%s_%s" % (prefix, func_name)
 
 
-def _prepend_names(prefix, orig_graph_def):
-    mut_graph_def = graph_pb2.GraphDef()
-    for node in orig_graph_def.node:
-        new_node = mut_graph_def.node.add()
+def _add_with_prepended_names(prefix, graph_to_add, destination_graph):
+    for node in graph_to_add.node:
+        new_node = destination_graph.node.add()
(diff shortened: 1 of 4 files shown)


tensorboard/plugins/graph/graph_util.py
```diff
@@ -14,152 +14,110 @@
...
-                "Cannot combine GraphDefs because gradients share a gradient_func name "
-                "but map to different functions: %s"
+    if graph_defs[0].versions.producer:
+        dst_graph_def.versions.CopyFrom(graph_defs[0].versions)
+
+    for index, graph_def in enumerate(graph_defs):
+        if dst_graph_def.versions.producer != graph_def.versions.producer:
+            raise ValueError("Cannot combine GraphDefs of different versions.")
+
+        mapped_graph_def = _prepend_names("graph_%d" % (index + 1), graph_def)
```
nfelt, line 112: You might consider a function signature more like

```python
def _add_with_prepended_names(prefix, graph_to_add, destination_graph)
```

That way we can avoid the overhead of constructing the temporary graphdef that's returned here, only to then copy its nodes and functions into the destination graph def.  Since proto python objects aren't immutable, the python code [has to do a deep copy](https://developers.google.com/protocol-buffers/docs/reference/python-generated#repeated-message-fields) each time we move the nodes around, so it would be nice not to do that twice when we could do it just once.

tensorboard/plugins/graph/graph_util.py
```diff
@@ -14,152 +14,110 @@
...
+        if new_node.op == "PartitionedCall" and new_node.attr["f"]:
+
+            new_node.attr["f"].func.name = _prefixed_func_name(
+                prefix, new_node.attr["f"].func.name,
+            )
+
+    for func in orig_graph_def.library.function:
+        new_func = mut_graph_def.library.function.add()
+        new_func.CopyFrom(func)
+        # Not creating a structure out of factored out function. They already
```
nfelt, line 50: Does this comment refer to the fact that we don't use a `/` in the function name prefix, just an underscore?  If so, the comment might make more sense as a comment on `_prefixed_func_name()` itself.

nfelt (approved):
Looks like this addresses https://github.com/tensorflow/tensorboard/issues/1929?  If so, worth adding `Fixes #1929` to the description?

stephanwlee:
While I am cautiously optimistic that this change would alleviate much of #1929, I suspect this change has not covered few cases such as differing GraphDef version and gradient function support (while I have some code to handle it, I could not get my hands on a realistic example and had to just go by the comments on the definition). I will keep the issue for the time being and maybe remove the UI instruction to refer to #1929 when I truly think it is solved.

PR #4335, opened 2020-11-17
rust: add `tempfile` dependency

wchargin:
Summary:
The [`tempfile`] crate is the standard Rust library for handling
temporary files and directories. We anticipate using this for tests that
want to set up a directory of event files.

[`tempfile`]: https://crates.io/crates/tempfile

Test Plan:
It builds: `bazel build //third_party/rust:tempfile`.

wchargin-branch: rust-dep-tempfile


stephanwlee (approved):
Does this belong in dev-dep?

wchargin:
Good point; thanks! Done.

PR #4338, opened 2020-11-17
rust: add `clap` dependency

wchargin:
Summary:
The Rust ecosystem has a number of crates for command-line argument
parsing; [`clap`] is one of them. I’ve used `clap` in the past and
thought it pleasant and easy to use.

I’ve expressed our dependency on `clap`’s v3 beta branch because it has
the wonderful feature of generating type-safe argument parsing simply
from a struct definition. This is both safer and more ergonomic, with
significantly less client code. Here’s a [patch upgrading `clap` from v2
to v3 on an old version of RustBoard][upg]. I’m willing to handle
breaking API changes as they come along in the beta series; [usually,
they’re just renames][chg].

[`clap`]: https://crates.io/crates/clap
[chg]: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md
[upg]: https://github.com/wchargin/rustboard/commit/55d4505d78c9afb0c06aa8aa19c74ccd8eb1e3c2

Test Plan:
It builds: `bazel build //third_party/rust:clap`.

wchargin-branch: rust-dep-clap


tensorboard/data/server/Cargo.toml
```diff
@@ -67,6 +68,9 @@ additional_flags = [
     "--cfg=has_std",
 ]
 
+[package.metadata.raze.crates.proc-macro-error.'1.0.4']
+gen_buildrs = true
+
```
stephanwlee, line 73: Is this change required since clap can emit error?

stephanwlee (approved):
EZ 👏 

Does rust support utf-8 in source code? Can I write a macro that aliases `clap` to 👏 ? :D

tensorboard/data/server/Cargo.toml
wchargin: Yes, I expect that the `clap` dep on `proc-macro-error` is to make it
easy for `clap` to report compile-time errors. This line is needed
because `proc-macro-error` uses a `build.rs` script to detect your Rust
version and enable certain features if they’re stable.

wchargin (commented):
> Does rust support utf-8 in source code?

Yep! Unicode is permitted in source code files, and moreover it is
[permitted in identifiers][play]:

```rust
let 유니코드 = true;
assert!(유니코드);
```

[play]: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=132fb2ac089e96cc383164029dc8d0cd

> Can I write a macro that aliases `clap` to  ? :D

Not quite. You can use ~arbitrary Unicode in string literals, comments,
etc., but identifiers must meet the [Unicode identifier contraints][s].
This includes basically “things that appear in words in human
languages”, but does not include emoji.

[s]: http://www.unicode.org/reports/tr31/#Default_Identifier_Syntax

PR #4343, opened 2020-11-18
rust: create multi-file run loader

wchargin:
Summary:
This patch adds a `RunLoader` that reads from multiple event files
within a single run. For now, it only keeps track of scalars, and it
doesn’t do anything with the data that it reads. But it persists
everything that we’ll need for the rest of the pipeline.

Test Plan:
Unit tests included: they create a test log directory and write some
event files, then ensure that they can be read back properly.

wchargin-branch: rust-run-loader


wchargin:
This patch is against the combined diffbase of #4340, #4341, and #4342.
Please ignore diffbase commits. (Currently, the only commit under review
is 997ddfa.) Thanks!

tensorboard/data/server/run.rs
```diff
@@ -0,0 +1,363 @@
...
+        self.reload_files();
+    }
+
+    /// Updates the active key set of `self.files` to match the given filenames.
+    ///
+    /// After this function returns, `self.files` may still have keys not in `filenames`, but they
+    /// will all map to [`EventFile::Dead`].
+    fn update_file_set(&mut self, filenames: Vec<PathBuf>) {
+        // Remove any discarded files.
+        let new_file_set: HashSet<&PathBuf> = filenames.iter().collect();
```
wchargin, line 131: Of some interest: you have probably seen us write `.collect()` before to
turn an iterator into a vector. But here we’re turning one into a hash
set! In fact, [`collect`] can generate any collection type, as long as
it implements [`FromIterator`]. The specific type in any instance is
determined by context, and the compiler will complain if there’s not
enough information to uniquely identify a type. This is sometimes called
[target typing].

[`collect`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect
[`FromIterator`]: https://doc.rust-lang.org/std/iter/trait.FromIterator.html
[target typing]: https://docs.oracle.com/javase/tutorial/java/javaOO/lambdaexpressions.html#target-typing

wchargin (commented):
A couple tour-guide comments…

tensorboard/data/server/run.rs
```diff
@@ -0,0 +1,363 @@
...
+        write_event(writer, &event)
+    }
+
+    #[test]
+    fn test() -> Result<(), Box<dyn std::error::Error>> {
+        let logdir = tempfile::tempdir()?;
+        let f1_name = logdir.path().join("tfevents.123");
+        let f2_name = logdir.path().join("tfevents.456");
+        let mut f1 = BufWriter::new(File::create(&f1_name)?);
+        let mut f2 = BufWriter::new(File::create(&f2_name)?);
```
stephanwlee, line 297: No action required.

Interesting; I thought the module would consider an inevitable abstraction of the file system and operate on testable EventFile or something akin to that. 

I assume this is actually writing these events out to the temp directory (at `sync_all` call in L315)?

tensorboard/data/server/run.rs
```diff
@@ -0,0 +1,363 @@
...
+        f1.into_inner()?.sync_all()?;
+        f2.into_inner()?.sync_all()?;
+
+        let mut loader = RunLoader::new();
+        loader.reload(vec![f1_name, f2_name]);
+
+        // Start time should be that of the file version event, even though that didn't correspond
+        // to any time series.
+        assert_eq!(loader.start_time, Some(WallTime::new(1234.0).unwrap()));
+        assert_eq!(loader.time_series.keys().collect::<Vec<_>>(), vec![&tag]);
```
stephanwlee, line 324: `<_>`! One day I will see a code with my favorite emoji, `>_<`

tensorboard/data/server/run.rs
```diff
@@ -0,0 +1,363 @@
...
+            }
+            panic!("step {:?}: {:?}", step, payload);
+        }
+        assert_eq!(
+            actual_points,
+            vec![
+                (Step(0), WallTime::new(1235.0).unwrap(), 0.25),
+                (Step(1), WallTime::new(1236.0).unwrap(), 0.50),
+                (Step(2), WallTime::new(2346.0).unwrap(), 0.75),
+                (Step(3), WallTime::new(2347.0).unwrap(), 1.00),
```
stephanwlee, line 357: Testing my understanding: I assume the `loader.time_series` does not actually sort/order events by step, right? It is just happens to be sorted since we wrote as such in L311-L314. Don't think I've read the reservoir module thoroughly to know that on top of my head, but it looks like `rsv.offer` simply pushes onto the vec.

stephanwlee (approved):

tensorboard/data/server/run.rs
```diff
@@ -0,0 +1,363 @@
...
+        let new_file_set: HashSet<&PathBuf> = filenames.iter().collect();
+        for (k, v) in self.files.iter_mut() {
+            if !new_file_set.contains(k) {
+                *v = EventFile::Dead;
+            }
+        }
+
+        // Open readers for any new files.
+        for filename in filenames {
+            use std::collections::btree_map;
```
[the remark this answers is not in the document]
wchargin: Yeah, this is fine!

First, note that Rust imports are always pure. Modules do not have side
effects. So you can freely reorder imports (`rustfmt` sorts them) or
declare them at any scope that you like without worrying about bugs.

So, then, it’s all about scope. The general Rust convention is to import
most types and traits that you use at top level, and modules if you
want, too. Importing bare functions is rarer, I think, but not unheard
of. I like to be able to see `data_compat::initial_metadata(...)` to
clarify that it’s not just a helper function in this module.

So I could have imported `btree_map::Entry` at top level. But I think
that that would be a wee bit confusing, because if you see `Entry` in
some arbitrary place in the code, it’s not obvious what that refers to
(as opposed to, e.g., `HashMap`).

I also could have imported `btree_map` at top level, and that would also
have been reasonable imho. I don’t really have strong opinions here.

Some examples of my conventions:

```rust
// OK: types and traits that are clear from context
use std::collections::{HashMap, HashSet};
use std::io::{Read, Write};

// OK: modules, aliases
use std::io;
use std::path;
use crate::proto::tensorboard as pb;

// Avoid: types not clear from context
// use std::io::Error;  // a fn returns "Error"---what kind of error?

// Weak-avoid: standalone functions
// use std::iter::once;  // seeing "iter::once" would be clearer
```

tensorboard/data/server/run.rs
wchargin: That was certainly my plan, to parameterize the reader:

```rust
pub struct RunLoader<R> {
    files: HashMap<PathBuf, EventFile<R>>,
    // ...
}
impl<R: Read> for RunLoader { /* ... */ }
```

This works swimmingly for reading, but the problem is that `Read` is
just a stream type, and doesn’t actually give you a way to _open_ a
file. So I would have needed to define something like:

```rust
trait Filesystem {
    type File: Read + Seek;
    fn open(p: &Path) -> io::Result<Self::File>;
}

struct RealFilesystem;
impl Filesystem for RealFilesystem {
    type File = std::fs::File;
    fn open(p: &Path) -> io::Result<Self::File> { std::fs::File::open(p) }
}
```

And this works perfectly fine, and we probably _will_ want it
eventually, but I didn’t really want to bother implementing my own fake
in-memory file system when the real filesystem will do just fine.

> I assume this is actually writing these events out to the temp
> directory (at `sync_all` call in L315)?

Yep. The `sync_all` is an `fsync(2)`. Without it, the files would still
be written when they’re dropped, but here we needed to keep a reference
to them later, so this flushes the `BufWriter`’s buffer.

tensorboard/data/server/run.rs
wchargin: Hmm… I’m having trouble thinking of a piece of standard Rust syntax that
contains `>_<`, since `_` is not a valid identifier by itself. :-(
But this time, [macros *can* save the day][play]:

```rust
macro_rules! swlmoji {
    (<_>) => { "meep" };
    (>_<) => { "mrrp" };
}
fn main() {
    println!("{}", swlmoji!(<_>));
    println!("{}", swlmoji!(>_<));
}
```

[play]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cac3ec76b5687aff4fb8775b19217dc0

tensorboard/data/server/run.rs
wchargin: It is sorted because the reservoir preempts any non-preceding records
*before* pushing onto the vector, so at the time that it pushes the step
of the new record is always larger than the steps of all other records.

Including a preemption in this test is probably a pretty reasonable
idea. I’ll go ahead and do that.

tensorboard/data/server/run.rs
wchargin: (oh, and the `_` in `Vec<_>` means “please infer type parameter :-) )

PR #4344, opened 2020-11-18
line chart: temporarily add demo component

stephanwlee:
This change adds a demo module that tests integration of the line chart
component. After integrating all the features into time series
dashboard, the module will be removed.

Since it is a temporary demo change, we are not introducing tests.

PR #4345, opened 2020-11-18
line chart: fix issues with line chart

stephanwlee:
When the demo was exercised, it revealed several issues with the
GpuLineChart.

Issues are:
- worker resource files are not served by the app
- requestAnimationFrame used by chart.ts in context of worker misses
  reference to `window` and causes invocation error as is.
  - refactored the "MainThreadChart" to "ChartImpl" since it is not
    meant to only be used in the "main thread"


wchargin (approved):
Changes look fine, though I would like to understand how to actually run
them (see comment on #4344).

PR #4346, opened 2020-11-18
line chart: add utility for smoothing line chart

stephanwlee:
This change ports logic in the vz_line_chart for our consumption in new
GpuLineChart.

Future integration note: unlike vz_line_chart, smoothing is not built-in
to the line chart; previous pattern made the line chart more logic-ful
and made it harder to adopt other smoothing strategies or data
transformation logic. Instead, we decided to have a notion of `aux` or
auxiliary data series and let the consumer pass the smoothed data
series.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/data_transformer.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/data_transformer.ts
@@ -25,7 +25,7 @@ import {DataSeries} from './lib/public_types';
 export async function classicSmoothing(
   data: DataSeries[],
   smoothingWeight: number
-): Promise<Array<{srcId: string; points: DataSeries['points']}>> {
+): Promise<DataSeries[]> {
   if (!data.length) [];
 
   if (!Number.isFinite(smoothingWeight)) {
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/line_chart_v2/data_transformer.ts
```diff
@@ -0,0 +1,99 @@
...
+        points: [],
+      });
+      continue;
+    }
+
+    let last = series.points.length > 0 ? 0 : NaN;
+    let numAccum = 0;
+
+    const initialYVal = series.points[0].y;
+    const isConstant = series.points.every((point) => point.y == initialYVal);
```
wchargin, line 51: Optional, but maybe consider extracting `smoothSeries(series.points)`?
imho makes it easier to see the nested loop structure; I misread it the
first time and thought that these `isConstant`s were being computed for
every point, which is not the case (good).

tensorboard/webapp/widgets/line_chart_v2/data_transformer.ts
```diff
@@ -0,0 +1,99 @@
...
+ * Smoothes data series in y axis using smoothing algorithm from classical TensorBoard
+ * circa 2019-2020. 1st-order IIR low-pass filter to attenuate the higher-frequency
+ * components of the time-series.
+ * @param data DataSeries to smooth
+ * @param smoothingWeight Degree of smoothing. Number between 0 and 1, inclusive.
+ */
+export async function classicSmoothing(
+  data: DataSeries[],
+  smoothingWeight: number
+): Promise<Array<{srcId: string; points: DataSeries['points']}>> {
```
wchargin, line 28: Curious why we don’t return a `DataSeries[]` here? Is it useful to
rename the `id` field to `srcId`?

Not necessarily objecting; just wondering about the rationale.

wchargin (approved):
Nice tests; thanks! This makes sense and was easy to review.

tensorboard/webapp/widgets/line_chart_v2/data_transformer.ts
```diff
@@ -0,0 +1,99 @@
...
+import {DataSeries} from './lib/public_types';
+
+/**
+ * Smoothes data series in y axis using smoothing algorithm from classical TensorBoard
+ * circa 2019-2020. 1st-order IIR low-pass filter to attenuate the higher-frequency
+ * components of the time-series.
+ * @param data DataSeries to smooth
+ * @param smoothingWeight Degree of smoothing. Number between 0 and 1, inclusive.
+ */
+export async function classicSmoothing(
```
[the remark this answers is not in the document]
stephanwlee: I want an option to use gpgpu or other ways of accelerating this later in the future (changing something async -> sync is easier than the other way around, imo)

tensorboard/webapp/widgets/line_chart_v2/data_transformer.ts
stephanwlee: I am a bit leaning towards keeping the signature as is; while I do not have a proof handy, I suspect transforming data will be a major area of work later and will likely be a hot path for performance. When such time comes, we should really look into WASM or other ways of doing the transformation and, in such case, asynchronous function that operates in a larger batch (entire DataSeries instead of one series) would be beneficial (to avoid frequent shuffling of data between host RAM to VRAM).

PR #4347, opened 2020-11-18
feat: selector to read fastChart feature flag

stephanwlee:
This change trivially adds a selector to already existing state in the
feature flag reducer. With this change, we are able to discern whether
user manually enabled the faster line chart implementation.

### DIFF
--- a/tensorboard/webapp/feature_flag/store/feature_flag_selectors_test.ts
+++ b/tensorboard/webapp/feature_flag/store/feature_flag_selectors_test.ts
@@ -65,7 +65,7 @@ describe('feature_flag_selectors', () => {
       );
       const actual1 = selectors.getIsGpuChartEnabled(state1);
 
-      expect(actual1).toEqual(false);
+      expect(actual1).toBe(false);
 
       const state2 = buildState(
         buildFeatureFlagState({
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/feature_flag/store/feature_flag_selectors_test.ts
```diff
@@ -53,4 +53,30 @@ describe('feature_flag_selectors', () => {
...
+      const state1 = buildState(
+        buildFeatureFlagState({
+          features: buildFeatureFlag({
+            enableGpuChart: false,
+          }),
+        })
+      );
+      const actual1 = selectors.getIsGpuChartEnabled(state1);
+
+      expect(actual1).toEqual(false);
```
wchargin, line 68: I don’t mind at all either way, but I thought that you liked using
`.toBe` here? Just curious whether there’s some nuance that I’m missing.

wchargin (approved):

tensorboard/webapp/feature_flag/store/feature_flag_selectors_test.ts
stephanwlee: Yup, that's possible. It was simply my oversight.

PR #4348, opened 2020-11-18
feat: persist fastChart feature flag to url

stephanwlee:
This change builds on top of already existing feature of deep linking to
persist a state in the feature flag reducer for convenience of a user.

With this change, when launching tensorboard with `?fastChart=true`, it
will stay in the URL across navigations.


psybuzz (approved):
Nice!

PR #4353, opened 2020-11-19
data: add `ListPlugins` RPC

wchargin:
Summary:
The data provider API uses `list_plugins` to populate the list of active
dashboards. I forgot to add a corresponding RPC in #4314, since I’d
hacked around it in my prototypes, but we do want to be able to
implement this cleanly.

Test Plan:
It builds: `bazel build //tensorboard/data/proto:protos_all_py_pb2_grpc`.

wchargin-branch: data-list-plugins-rpc


psybuzz (approved):
I'm curious, since the `TensorBoardDataProvider` service doesn't contain an rpc for data_location() or experiment_metadata(), is the plan to eventually add them later on?  Or will we have some other provider to handle those?

wchargin:
Thanks! Yes, I plan to add those later, for two reasons. First, it’s
just easier to start with fewer things at once. Second, I have always
wanted to combine `data_location` and `experiment_metadata` into one
data provider method (I was on vacation when `experiment_metadata` was
introduced and didn’t get a chance to review it). They happen at the
same point in the query path and with the same information, and they’re
semantically quite similar—the data location _is_ experiment metadata.
So I’ll want to think about how to structure those RPCs.

For now, the data provider reports a location of `grpc://localhost:6806`
and no experiment metadata, which is perfectly fine.

PR #4356, opened 2020-11-19
data: implement gRPC data provider for scalars

wchargin:
Summary:
A new `GrpcDataProvider` talks over the `TensorBoardDataProvider` gRPC
interface to query for data. This enables it to talk to a RustBoard
process running concurrently. Pass `--grpc_data_provider ADDR` to
connect to a server running on the given address, like `localhost:6806`.

For now, this data provider only supports scalars. We’ll add more data
classes as we go along. (Easier to review and test this way.)

Test Plan:
Unit tests included. For an end-to-end test, run:

```
bazel run -c opt //tensorboard/data/server -- LOGDIR &
bazel run -c opt //tensorboard -- \
    --bind_all \
    --generic_data true \
    --grpc_data_provider localhost:6806 \
    --verbosity 0 \
    ;
```

…and you should see a working scalars dashboard for the given logdir.

Also tested that normal TensorBoard still works both in open-source and
after a test sync into Google.

wchargin-branch: data-grpc-provider


wchargin:
Despite the **core:rustboard** tag, this is just Python. :-)

Still of course happy to answer any questions, Rust- or otherwise.

tensorboard/data/grpc_provider.py
```diff
@@ -0,0 +1,155 @@
...
+            raise errors.PermissionDeniedError(e.details())
+        raise
+
+
+def _populate_rtf(run_tag_filter, rtf_proto):
+    """Copies `run_tag_filter` into `rtf_proto`."""
+    if run_tag_filter is None:
+        return
+    if run_tag_filter.runs is not None:
+        rtf_proto.runs.names[:] = sorted(run_tag_filter.runs)
```
psybuzz, line 153: Does traditional TB also sort the runs/tags, or is this sorting logic new?

tensorboard/data/grpc_provider.py
```diff
@@ -0,0 +1,155 @@
...
+        return [p.name for p in res.plugins]
+
+    def list_runs(self, ctx, *, experiment_id):
+        req = data_provider_pb2.ListRunsRequest()
+        req.experiment_id = experiment_id
+        with _translate_grpc_error():
+            res = self._stub.ListRuns(req)
+        return [
+            provider.Run(
+                run_id=run.name, run_name=run.name, start_time=run.start_time,
```
psybuzz, line 65: Are `run_id`s a new concept introduced in this grpc design, which is supposed to be unique "across all experiments"?

If the plan is to support globally unique run ids in a world with multiple experiments (e.g. upgrade `run_id=run.name` into `run_id=get_run_id(exp_id, run.name)` or something), it might be worth adding a comment saying that this is intended to change in the future.

tensorboard/program.py
```diff
@@ -407,16 +409,34 @@ def _fix_mime_types(self):
...
+        channel = grpc.insecure_channel(addr, options=options)
+        stub = grpc_provider.make_stub(channel)
+        provider = grpc_provider.GrpcDataProvider(addr, stub)
+        return provider
+
+    def _make_data_provider(self):
+        """Returns `(data_provider, deprecated_multiplexer)`."""
+        grpc_addr = self.flags.grpc_data_provider
+        if grpc_addr:
+            return (self._make_grpc_provider(grpc_addr), None)
```
psybuzz, line 425: For my own understanding, how does the `grpc_data_provider` know about the logdir, if we don't pass it the CLI flags?

tensorboard/data/grpc_provider.py
wchargin: No, on the contrary: the gRPC design does not have run IDs, because we
do not yet know what the right semantics are. It is the Python data
provider API that (optimistically) has run IDs, which are a bit of a
wart: they’re returned from `list_runs`, but ignored, and not used in
any of the other methods.

tensorboard/data/grpc_provider.py
wchargin: Traditional TensorBoard doesn’t send an RPC at all here, so there’s no
corresponding code. This sorting logic just makes the wire format—which
is logically a set—deterministic in representation. I think that that’s
a good habit when performance isn’t critical. It doesn’t really matter.

tensorboard/program.py
wchargin: The `grpc_data_provider` doesn’t know about the logdir, nor does it need
to. The data location is populated as `grpc://localhost:6801` based on
the address to which it connects. The logdir that the RPC _server_ reads
from is given as a command-line flag to the server, which is a separate
process entirely.

wchargin (commented):
Thanks! Responded, but I don’t see any actionable changes. PTAL.

tensorboard/data/grpc_provider.py
psybuzz: Gotcha, I also see that `MultiplexerDataProvider` also uses the run_name as the run_id.

tensorboard/program.py
psybuzz: Ahh right, the technical design doc even clarifies that the data loading server is in a separate process.  I forgot about that.

In the longer term, is the plan to keep these 2 processes launched separately?  Or will we have the main TB process spawn the RPC serving process, and pass it a subset of the command line flags?

psybuzz (approved):
After trying it out manually, and it works!

Note: if copying the description into the commit message, I would change the first line from
`bazel run -c opt //tensorboard/data/server &`
to
`bazel run -c opt //tensorboard/data/server <logdir> &`

tensorboard/program.py
wchargin: In at least the medium term, I plan for them to remain separate
processes (rather than an in-process C extension), but I do plan for
TensorBoard to launch the RPC server. A sibling of `LocalDataIngester`
will probably take care of that.

tensorboard/data/grpc_provider.py
wchargin: Right. (And same for TensorBoard.dev, fwiw.)

wchargin:
> Note: if copying the description into the commit message, I would change the first line from
> `bazel run -c opt //tensorboard/data/server &`
> to
> `bazel run -c opt //tensorboard/data/server <logdir> &`

It was correct when the PR was originally sent, and then it was
`<logdir>`; as of #4397, it’s `--logdir <logdir>`. Progress is fast. :-)

PR #4360, opened 2020-11-20
cleanup: use Parse instead of Merge

stephanwlee:
lint tool suggested using `Parse` instead of `Merge` for creating protobuf
from a txtproto. This change simply mechanically changes the API we use.


wchargin:
For context, the difference is that if your data has multiple values for
a singleton field, `Merge` silently drops all but one value, whereas
`Parse` raises an exception.

Might be nice to put this in the commit message for posterity.

bmd3k (approved):

PR #4361, opened 2020-11-20
build: set `srcs_version = "PY3"` everywhere

wchargin:
Summary:
We dropped official support for Python 2.7 early this year, but our
Python build rules still had `srcs_version = "PY2AND3"` in many places
due to lingering Google dependents. Now is the time to fix that, so that
we may move forward with even more conviction.

Generated with:

```
buildozer '//...:%py_library' 'set srcs_version PY3' &&
bazel query 'filter(expect.*installed, //tensorboard:all)' \
    | xargs buildozer 'remove srcs_version' &&
git ls-files -z | xargs -0 sed -i s/PY2AND3/PY3/
```

…which skips adding it to binaries and tests that don’t yet have it.
(They don’t pose a problem with reverse dependencies.) The global `sed`
hits `third_party/*.BUILD` and the `*.bzl` files.

Test Plan:
Sync into Google, run a global presubmit, and see what happens. Last
time I did this, only three targets failed (out of all our transitive
reverse dependencies), and they all looked like false positives.

wchargin-branch: build-srcs-version-py3


wchargin:
Googlers, see: <http://cl/343401587>

wchargin:
We’re happy with the global presubmit results. Plan to merge this on
Monday.

bileschi (approved):

PR #4364, opened 2020-11-20
rust: extract test utils for writing event files

wchargin:
Summary:
The run loader tests write sample events to files. These utilities will
also be useful for testing the logdir loader, so we factor them out into
a test-only helper module.

This uses the *[extension trait]* pattern: we define a new interface and
then implement it for _all_ other writers. This way, as long as you have
`SummaryWriteExt` in scope, you can call `.write_event(...)` on any
implementation of `Write`, like a file, a cursor, or a buffered writer.
Extension traits aren’t special or magic; they’re just a useful
consequence of how traits work in Rust.

[extension trait]: https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html

Test Plan:
Unit tests for the run loader updated to use the new helpers, which
themselves are also tested.

wchargin-branch: rust-writer-utils

### DIFF
--- a/tensorboard/data/server/writer.rs
+++ b/tensorboard/data/server/writer.rs
@@ -1,0 +1,126 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+//! Test helpers for writing event files.
+
+use std::io::Write;
+
+use crate::proto::tensorboard as pb;
+use crate::types::{Step, Tag, WallTime};
+
+/// Extends [`Write`] with methods for writing summary event files.
+pub trait SummaryWriteExt: Write {
+    /// Writes a TFRecord containing an `Event` proto into this writer.
+    fn write_event(&mut self, event: &pb::Event) -> std::io::Result<()> {
+        use prost::Message;
+        let mut data = Vec::new();
+        event.encode(&mut data)?;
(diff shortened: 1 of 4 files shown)


tensorboard/data/server/writer.rs
```diff
@@ -0,0 +1,126 @@
...
+                }],
+                ..Default::default()
+            })),
+            ..Default::default()
+        };
+        self.write_event(&event)
+    }
+}
+
+impl<W: Write> SummaryWriteExt for W {}
```
wchargin, line 59: Here’s the secret sauce: we implement `SummaryWriteExt` for all types
`W` that implement `Write`. Since `SummaryWriteExt` only has methods
with default implementations that are implemented in terms of the
existing methods on `Write`, we don’t actually need to put anything in
this block. Thus, any caller can access these methods by just bringing
`SummaryWriteExt` into scope.

tensorboard/data/server/writer.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    fn read_all_events<R: Read>(reader: R) -> Result<Vec<pb::Event>, event_file::ReadEventError> {
+        let mut result = Vec::new();
+        use crate::event_file::ReadEventError::ReadRecordError;
+        use crate::tf_record::ReadRecordError::Truncated;
+        let mut reader = EventFileReader::new(reader);
+        loop {
+            match reader.read_event() {
+                Ok(event) => result.push(event),
+                Err(ReadRecordError(Truncated)) => return Ok(result),
```
stephanwlee, line 76: Interesting; I expected the opposite. Truncated gets called when buffer is not full, right? 

https://github.com/tensorflow/tensorboard/blob/708ab6186b8399f6e6238b5e43a540994b15a041/tensorboard/data/server/tf_record.rs#L287-L290

I may be reading this wrongly :)

stephanwlee (approved):

tensorboard/data/server/writer.rs
wchargin: Yes, which includes the case where you’ve reached the end of file, and
so the “next record” is truncated at 0 bytes:

https://github.com/tensorflow/tensorboard/blob/708ab6186b8399f6e6238b5e43a540994b15a041/tensorboard/data/server/tf_record.rs#L124-L126

We could thread through some information to distinguish “EOF for now”
from “truncated after reading at least one byte”, but given that we
treat those two cases equivalently, anyway, I haven’t seen a need to do
so.

Does this answer your question?

PR #4367, opened 2020-11-20
infra: upgrade Bazel to 3.7.0 (1 of 4)

stephanwlee:
This change disables large part of CI so we can safely populate the
Travis cache to unblock other infrastructural upgrade.

Note that there are total of 4 PRs to be submitted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,14 +13,6 @@ node_js:
   - "11"
 
 env:
-  # Keep this Bazel version in sync with the `versions.check` directive
-  # near the top of our WORKSPACE file.
-  #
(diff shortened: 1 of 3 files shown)


nfelt (approved):

wchargin:
I was getting [HTTP 429 Too Many Requests][429] on the raw GitHub
release URL, so I’ve copied it over to the TF mirror. New builds should
transparently pick that up.

(Our `mirror_urls_test.sh` doesn’t pick that up currently; I’m inclined
to leave that as is, since it’d be a slight pain and we don’t update
this URL frequently.)

[429]: https://http.cat/429

Issue #4362, opened 2020-11-20
Constant time series with small values don't have proper axes


atishagarwala:
If I have a time series with a large, constant value f (e.g. 10.0), the y-axis scale reflects the constant value (scale goes from 0 to ~2f). However, if my value f is small (e.g. 0.01), the y-axis goes from -1.0 to 1.0, and does not at all reflect the scale of the constant.

It appears this issue was raised [here](https://github.com/tensorflow/tensorboard/pull/1507#discussion_r226418475) but not changed. I think constant curves should be made consistent - in the code where I noticed it, deviations from the known constant value were informative, and I spent a bunch of time thinking that I had a deviation (value of 0 instead of small but non-zero) due to not understanding how the plot axes were set.

Good:

![image](https://user-images.githubusercontent.com/23422890/99750012-76b48300-2a94-11eb-80f7-807897c35ee9.png)

Bad (real value of 1e-3):

![image](https://user-images.githubusercontent.com/23422890/99749978-64d2e000-2a94-11eb-9a9c-4b24cc564217.png)

bmd3k:
Thanks, i was able to reproduce. I'll see if I can improve things for this case.

nfelt:
@stephanwlee it looks like the domain scaling approach for gpu line chart is a bit different:
https://github.com/tensorflow/tensorboard/blob/ad652dacb11a5793ff7127709604e3842d7aa441/tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts#L114-L131

We probably want to be using a consistent domain "niceification" rule in both places, right?  So if the one for GPU line chart is better, maybe we should just backport that?

PR #4371, opened 2020-11-21
line chart: fix the nice domain for small constant

stephanwlee:
Also, this change improves testing around `niceDomain` and we assert on
non-positive value clipping.

This change partially address #4362 for the new gpu line chart. We will later
integrate the new scale to the legacy vz_line_chart.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
@@ -155,7 +155,7 @@ class Log10Scale implements Scale {
       return [Number.MIN_VALUE, 1];
     }
 
-    return [Math.max(Number.MIN_VALUE, adjustedMin * 0.5), adjustedMax * 1.5];
+    return [Math.max(Number.MIN_VALUE, adjustedMin * 0.5), adjustedMax * 2];
   }
 
   ticks(domain: [number, number], sizeGuidance: number): number[] {
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
@@ -28,7 +28,7 @@ export function createScale(type: ScaleType): Scale {
 }
 
 const PADDING_RATIO = 0.05;
-const MIN_SIGNIFICANT_PADDING = 0.01;
+const PADDING_FOR_ZERO = 0.01;
 
 class LinearScale implements Scale {
   private transform(
@@ -74,7 +74,9 @@ class LinearScale implements Scale {
     const padding =
       max === min
         ? // In case both `min` and `max` are 0, we want some padding.
-          Math.max(min * PADDING_RATIO, MIN_SIGNIFICANT_PADDING)
+          min === 0
+          ? PADDING_FOR_ZERO
+          : Math.abs(min) * PADDING_RATIO
         : (max - min + Number.EPSILON) * PADDING_RATIO;
     const [niceMin, niceMax] = scale
       .domain([min - padding, max + padding])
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
```diff
@@ -74,7 +74,9 @@ class LinearScale implements Scale {
     const padding =
       max === min
         ? // In case both `min` and `max` are 0, we want some padding.
-          Math.max(min * PADDING_RATIO, MIN_SIGNIFICANT_PADDING)
+          min === 0
+          ? PADDING_FOR_ZERO
+          : Math.abs(min) * PADDING_RATIO
```
nfelt, line 79: FWIW, I'm ok with this padding approach, but historically we had added more generous padding for constant plots (basically using `[0, 2c]` as the range) to ensure that we were always showing the x axis, while still adjusting the overall scale to match the magnitude of the constant.  Did you have a preference for the tighter zoom?  Could see an argument either way, just curious.

nfelt (approved):
Thanks for the fix!

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
stephanwlee: I see some reference to that time, but I am not sure whether that's what we have today (comment does not agree with the code, at least not in an obvious way).

https://github.com/tensorflow/tensorboard/blob/c7a4c73ca389318b200de157dda253f07279eeb8/tensorboard/components/vz_line_chart2/linear-scale.ts#L45-L50 

Regardless of what is happening in vz_line_chart, decided to include `0` in all cases.

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
```diff
@@ -161,10 +164,7 @@ class Log10Scale implements Scale {
       spreadInLog > 0
         ? spreadInLog * PADDING_RATIO
         : // In case `minLogValue` is 0 (i.e., log_10(1) = 0), we want some padding.
-          Math.max(
-            Math.abs(minLogValue * PADDING_RATIO),
-            MIN_SIGNIFICANT_PADDING
-          );
+          Math.max(Math.abs(minLogValue * PADDING_RATIO), PADDING_FOR_ZERO);
```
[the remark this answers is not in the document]
stephanwlee: Put some constant padding in the log scale (halving and doubling in log scale is just `log_10(x) ± log_10(2)`. I am not sure if it is visually pleasing just yet.

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
```diff
@@ -148,28 +149,13 @@ class Log10Scale implements Scale {
...
-          Math.max(
-            Math.abs(minLogValue * PADDING_RATIO),
-            MIN_SIGNIFICANT_PADDING
-          );
-
-    return [
-      this.untransform(Math.max(numericMinLogValue, minLogValue - padInLog)),
-      this.untransform(maxLogValue + padInLog),
-    ];
+    return [Math.max(Number.MIN_VALUE, adjustedMin * 0.5), adjustedMax * 1.5];
```
nfelt, line 158: You mentioned "Put some constant padding in the log scale (halving and doubling in log scale...)" but this is halving and then adding 50% - should it be `adjustedMax * 2.0`?  Or do we want the upper padding to be smaller for some reason?

This also seems like it doesn't match the tests which might be causing the failures (although weirdly, only for one of the two travis jobs).

nfelt (approved):
LGTM.  Yeah I agree the code in `vz-line-chart` is hard to parse, but the new code here looks great.

tensorboard/webapp/widgets/line_chart_v2/lib/scale.ts
stephanwlee: Yup, that was a mistake on my part. Fixed the test and code.

PR #4372, opened 2020-11-21
Limit max scalar smoothing to 0.99 / 0.999

psybuzz:
The smoothing slider setting affects line charts. In the Scalars
dashboard, the maximum value has traditionally been 0.999, not
1.0. The Time Series dashboard allows the slider to go all the way
to 1, which "flattens the curve" onto the x-axis, often much less
meaningful than a "close to 1" number, which still shows the
general shape of the curves.

This updates the Time Series' scalar cards to be use a max
smoothing of 0.99 when dragging the slider, and 0.999 when
entering a value in the input field.

Before:
image

After:
image

### DIFF
--- a/tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
+++ b/tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
@@ -37,7 +37,7 @@ limitations under the License.
       <mat-slider
         aria-labelledby="scalars-smoothing-label"
         color="primary"
-        [max]="MAX_SMOOTHING_VALUE"
+        [max]="MAX_SMOOTHING_SLIDER_VALUE"
         [min]="0"
         [step]="0.01"
         [value]="scalarSmoothing"
@@ -50,7 +50,7 @@ limitations under the License.
         type="number"
         [max]="MAX_SMOOTHING_VALUE"
         min="0"
-        step="0.01"
+        step="0.001"
         [value]="scalarSmoothing"
         (input)="onScalarSmoothingInput($event)"
       />
--- a/tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
+++ b/tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
@@ -39,10 +39,9 @@ limitations under the License.
         color="primary"
         [max]="MAX_SMOOTHING_VALUE"
         [min]="0"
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
```diff
@@ -37,20 +37,21 @@ <h3 class="section-title">Scalars</h3>
       <mat-slider
         aria-labelledby="scalars-smoothing-label"
         color="primary"
-        [max]="1"
+        [max]="MAX_SMOOTHING_VALUE"
         [min]="0"
-        [step]="0.01"
+        [step]="0.001"
```
stephanwlee, line 42: Do we absolutely need the triple precision? It looks really bad on the fob. I think it can stay `0.01` and max `0.99` (and revert the formatting which I think is a bit janky) unless people are giving us feedback otherwise.

stephanwlee (approved):
Thanks for proactively fixing this.

tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
psybuzz: I think you're right, the extra precision is likely too aggressive, and yeah, the formatting fn is unfortunate.

Reduced to max=0.99 with step=0.01

wchargin (commented):
> the maximum value has traditionally been 0.999

You note correctly. In my experience, there is often a meaningful
smoothing difference between `0.99` and `0.999`, and I often like to set
it to around `0.995`. Why use `0.99` for time series instead of `0.999`?

(I have seen the comment above.)

Aside: I wonder what it would feel like if we interpreted the smoothing
value logarithmically: say, apply `log_2(1 - x)` to the current value.
So `0.5` → 1, `0.75` → 2, `0.99` → 6.6, `0.999` → 10. Might be nice
since I often find myself fiddling in the very extremes.

psybuzz:
Proposing the middle-ground:

- Draggable slider has a max of 0.99, moves in 0.01 increments.
- Text input field has a max of 0.999, moves in 0.001 increments.

A slider with 0.001 increments is "too fine grained".  On the existing
Scalar dashboard smoothing slider, my mouse isn't even able to make
deltas of < 0.05.

With the small amount of extra complexity in this PR, we give granular
control only to the numeric text input field.

PR #4373, opened 2020-11-21
bug: apply new scale niceDomain to vz-line-chart

### LINKED ISSUE #4362: Constant time series with small values don't have proper axes

stephanwlee:
We are in midst of working on the new version of the line chart. In
doing so, we rewrote a scale library with a bit more tests.

This change integrates the new scale library so small constant values do
not result (e.g., 0.05) in  [-1, 1] domain.

image

Fixes #4362.

### DIFF
--- a/tensorboard/components/vz_line_chart2/linear-scale.ts
+++ b/tensorboard/components/vz_line_chart2/linear-scale.ts
@@ -15,7 +15,11 @@ limitations under the License.
 import * as d3 from 'd3';
 import * as Plottable from 'plottable';
 
-import {ValueProviderForDomain, ITfScale} from './tf-scale';
+import {
+  createScale,
+  ScaleType,
+} from '../../webapp/widgets/line_chart_v2/lib/scale';
+import {ITfScale, ValueProviderForDomain} from './tf-scale';
 
 export class LinearScale extends Plottable.Scales.Linear implements ITfScale {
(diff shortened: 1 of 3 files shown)


tensorboard/components/vz_line_chart2/linear-scale.ts
```diff
@@ -41,27 +45,8 @@ export class LinearScale extends Plottable.Scales.Linear implements ITfScale {
    */
```
nfelt, line 45: Can you remove the comment above, now that it's not accurate any more?

nfelt (approved):
LGTM.  Is there anything worth porting in the log scale?  Or I guess since it's not fixing the bug maybe not worth backporting.

PR #4375, opened 2020-11-23
cleanup: debugger to use tf_ng_module

stephanwlee:
ng_module will change shortly and we require some indirection for our
convenience. This change makes changes to the debugger_v2 code and
change our CI lint check to guard against any usages of ng_module.


wchargin (approved):
Nice; thanks!

Up to you, but is this worth test-syncing?

stephanwlee:
The risk is low for this PR; this is not the first PR we had converted to use the indirection.

PR #4377, opened 2020-11-23
line chart: export types for consumer of line chart

stephanwlee:
lib/public_types.ts is not really meant to be imported by consumer of
the line chart as (1) peeking into `lib` folder is a bit messy and (2)
it exports more types than what consumers really need.

The new `types` module is provided as part of the line_chart_v2
dependency and it only contains the types that consumers would need.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/types.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/types.ts
@@ -1,0 +1,23 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/line_chart_v2/types.ts
```diff
@@ -0,0 +1,23 @@
...
+==============================================================================*/
+
+// Only selectively export types that would be used by the user of the line chart.
+export {
+  DataSeries,
+  RendererType,
+  ScaleType,
+  DataSeriesMetadataMap,
+  DataSeriesMetadata,
+} from './lib/public_types';
```
wchargin, line 23: Thanks for the explicit export! Much easier to read these when poking
around the codebase than the `export *`s that we use in some places,
especially when there are multiple `export *`s in the same module.

wchargin (approved):

PR #4380, opened 2020-11-24
rust: discover event files in log directories

wchargin:
Summary:
A new `LogdirLoader` type owns `RunLoader`s for each run and wires them
up to a shared `Commit`. It is the logdir loader that finds event files
in the filesystem and feeds them to the run loaders.

For ease of review, this patch implements event file discovery but does
not actually load data. Stay tuned.

Test Plan:
Unit tests included. One test covers an edge case involving directories
whose names are not valid Unicode. This test only runs on Unix. We could
make it run on Windows, too, but I don’t have a Windows box to test on,
and the code under test is portable: the only platform-specific code is
the test setup. Same for some symlink-related tests. So run tests on a
`cfg(unix)` platform, such as Linux or macOS, for best coverage.

wchargin-branch: rust-logdir-loader-discovery


tensorboard/data/server/logdir.rs
```diff
@@ -0,0 +1,384 @@
...
+struct RunState {
+    /// Stateful loader for this run.
+    #[allow(dead_code)] // under construction
+    loader: RunLoader,
+    /// Path to this run's directory relative to the root logdir.
+    relpath: PathBuf,
+    /// Logdir-relative paths to other directories that normalize to the same name as this run
+    /// after lossy Unicode conversion. This is only non-empty when there are multiple filesystem
+    /// paths that are invalid Unicode and collide. For deduplicating warnings.
+    merged_relpaths: HashSet<PathBuf>,
```
stephanwlee, line 42: Subjective: I found it hard to conceptualize this and struggled a bit to understand how to understand this. It did click and ultimately understood what it means. 

Would it be better/worse to call this `collided_relpaths` or something like it?

tensorboard/data/server/logdir.rs
```diff
@@ -0,0 +1,384 @@
...
+                continue;
+            }
+            let filename = dirent.file_name().to_string_lossy();
+            if !filename.contains(EVENT_FILE_BASENAME_INFIX) {
+                continue;
+            }
+            let run_dir = match dirent.path().parent() {
+                Some(parent) => parent,
+                None => {
+                    // I don't know of any circumstance where this can happen, but I would believe
```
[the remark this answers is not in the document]
wchargin: Perhaps? But `dirent.path().parent()` doesn’t do any I/O; it’s supposed
to only fail “if the path terminates in a root or prefix”, where a root
is like `/` and a prefix is like `C:\`. So I think the filesystem would
have to say that `/` or `C:\` is a regular file. Not sure.

tensorboard/data/server/logdir.rs
```diff
@@ -0,0 +1,384 @@
...
+    pub fn reload(&mut self) {
+        let discoveries = self.discover();
+        self.synchronize_runs(&discoveries);
+        self.load_runs(discoveries);
+    }
+
+    /// Finds all event files under the log directory and groups them by run.
+    fn discover(&self) -> Discoveries {
+        let mut run_map: HashMap<Run, Vec<EventFileDiscovery>> = HashMap::new();
+        let walker = WalkDir::new(&self.logdir).sort_by(|a, b| a.file_name().cmp(b.file_name()));
```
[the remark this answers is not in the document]
wchargin: True; thanks! Added `.follow_links(true)` and tests.

tensorboard/data/server/logdir.rs
wchargin: Sure—thanks for the feedback. Renamed per your suggestion, and expanded
docs. Hopefully this helps.

wchargin (commented):
Thanks for the review!

tensorboard/data/server/logdir.rs
```diff
@@ -0,0 +1,442 @@
...
+            vec![&Run("test".to_string()), &Run("train".to_string())]
+                .into_iter()
+                .collect::<HashSet<_>>(),
+        );
+        Ok(())
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn test_symlink_loop() -> Result<(), Box<dyn std::error::Error>> {
```
stephanwlee, line 430: Yay :) I was wondering how the cycle is handled and you wrote a test for that!

stephanwlee (approved):

tensorboard/data/server/logdir.rs
wchargin: Yep—[`WalkDir::follow_links`] returns an error, which we print and then
continue:

```
$ cargo test --show-output
---- logdir::tests::test_symlink_loop stdout ----
error walking log directory: IO error for operation on /tmp/.tmpAeanMV/dir1: Too many levels of symbolic links (os error 40)
error walking log directory: IO error for operation on /tmp/.tmpAeanMV/dir2: Too many levels of symbolic links (os error 40)

---- logdir::tests::test_bad_unicode_collision stdout ----
merging directories "test\x99.run" and "test\xEE.run", which both normalize to run "test�.run"
```

[`WalkDir::follow_links`]: https://docs.rs/walkdir/2.3.1/walkdir/struct.WalkDir.html#method.follow_links

PR #4382, opened 2020-11-24
rust: add `anyhow` dependency

wchargin:
Summary:
The [`anyhow`] crate facilitates returning *unstructured* error values
in applications. It makes it easy to define freeform errors and add
context to them. It’s a sibling to [`thiserror`], by the same author.

[`anyhow`]: https://crates.io/crates/anyhow
[`thiserror`]: https://crates.io/crates/thiserror

Test Plan:
It builds: `bazel build //third_party/rust:anyhow`.

wchargin-branch: rust-dep-anyhow

### DIFF
--- a/tensorboard/data/server/Cargo.lock
+++ b/tensorboard/data/server/Cargo.lock
@@ -700,6 +700,7 @@ dependencies = [
 name = "rustboard"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "async-stream 0.3.0",
  "async-trait",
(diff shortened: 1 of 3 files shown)


wchargin:
You know, on second thought, I’m actually not yet sure that we’ll need
this. A normal CLI application would find it quite helpful, so I kind of
added it by habit, but a server might not find it necessary.

Instead, I’ll send you a PR for a logging interface, which we definitely
will want. :-)

PR #4383, opened 2020-11-24
rust: add `log` and `env_logger` dependencies

wchargin:
Summary:
The [`log`] crate provides a [façade] with familiar logging macros:
`info!`, `warn!`, `error!`, etc. To use `log`, you should bring your own
logging sink to handle log records. The [`env_logger`] crate provides
one such sink, which emits simple command-line output color-coded by log
level. It can optionally filter down to log output from specified
modules via an environment variable at runtime, which can be handy for
debugging.

Both these crates are popular and standard: even though they’re not part
of the Rust standard library, they’re maintained by the Rust core team.

[`env_logger`]: https://crates.io/crates/env_logger
[`log`]: https://crates.io/crates/log
[façade]: https://en.wikipedia.org/wiki/Facade_pattern

Test Plan:
They build: `bazel build //third_party/rust:{log,env_logger}`.

wchargin-branch: rust-dep-log


bileschi (approved):
Thanks for the clear PR message!

PR #4386, opened 2020-11-24
rust: implement `ListScalars` and `ReadScalars`

wchargin:
Summary:
The Rust gRPC server now implements the RPCs required to serve scalar
data to the TensorBoard frontend.

There is some repetition in the implementation: `ListScalars` and
`ReadScalars` have very similar bodies. There will be more repetition
with the tensor RPCs, and to a lesser degree with those for blob
sequences. I plan to consider refactoring that later, once all the raw
materials are there, but I’m open to feedback.

Test Plan:
Unit tests included. For an end-to-end test, cherry-pick #4356, launch
the data server (`bazel run //tensorboard/data/server -- LOGDIR`), and
concurrently run TensorBoard with `--grpc_data_provider localhost:6806`
and `--generic_data true` flags.

wchargin-branch: rust-listscalars-readscalars


tensorboard/data/server/server.rs
```diff
@@ -312,15 +507,72 @@ mod tests {
...
+                plugin_name: "scalars".to_string(),
+                ..Default::default()
+            }),
+            downsample: Some(data::Downsample {
+                num_points: 0,
+                ..Default::default()
+            }),
+            ..Default::default()
+        });
+        handler.read_scalars(req).await.unwrap();
```
nfelt, line 576: Do we not want to actually assert anything on the response?  Seems like it might not hurt to check that there are indeed no points.

tensorboard/data/server/server.rs
```diff
@@ -92,44 +93,122 @@ impl TensorBoardDataProvider for DataProviderHandler {
...
-            metadata: Some(data::ScalarMetadata {
-                max_step: 5,
-                max_wall_time: FAKE_START_TIME + 5.0,
-                ..Default::default()
-            }),
-            ..Default::default()
-        });
-        res.runs.push(run);
+        for (run, data) in runs.iter() {
+            if !run_filter.want(run) {
```
nfelt, line 105: FWIW, iterating over the filter instead of the keyset would seem likely to save effort on net (assuming single-run lookups against a big experiment are fairly plausible, while sending a filter with many non-existent runs is unlikely; I guess most optimal would be to iterate over the shorter one).  Thoughts?  It probably does make the code a little messier so maybe better to wait.

tensorboard/data/server/server.rs
```diff
@@ -92,44 +93,122 @@ impl TensorBoardDataProvider for DataProviderHandler {
...
+                    .as_ref()
+                    .map(|pd| pd.plugin_name.as_str());
+                if plugin_name != Some(&want_plugin) {
+                    continue;
+                }
+                let max_step = match ts.valid_values().last() {
+                    None => continue,
+                    Some((step, _, _)) => step,
+                };
+                let max_wall_time = ts
```
nfelt, line 128: Maybe worth a TODO to track this on `TimeSeries`?  Would be nice not to have to rescan every single scalar wall time each time we hit `ListScalars`.  I guess it's hard to get this entirely for "free" on `TimeSeries` due to preemptions, but still seems overall cheaper to compute it there (where we could take a cheap path if no preemption happened).

nfelt (approved):

tensorboard/data/server/server.rs
wchargin: Okay, on thinking a bit more I guess a simpler formulation is


```rust
impl<T: Hash + Eq> Filter<T> {
    fn pluck<'f, 'm: 'f, V: 'm>(
        &'f self,
        it: impl Iterator<Item = (&'m T, &'m V)> + 'm,
    ) -> impl Iterator<Item = (&'m T, &'m V)> + 'f
    where
        T: Eq + Hash + 'm,
    {
        let which: Option<&'f HashSet<T>> = match self {
            Filter::All => None,
            Filter::Just(which) => Some(which),
        };
        it.filter(move |(k, _)| match which {
            None => true,
            Some(which) => which.contains(k),
        })
    }
}
```

which should about have the same runtime properties as what I sketched
above: the `match self` on `Either::next` in that version corresponds to
the `match which` in the `it.filter` closure in this one.

So less complicated (almost fits on a screen), but the lifetime
annotations are still a bit much. But I could at least reasonably be
convinced now. Let me know if you have any thoughts.

wchargin (commented):
PTAL re: `run_tag_map!`; the rest is mostly just done.

tensorboard/data/server/server.rs
```diff
@@ -286,6 +425,36 @@ mod tests {
         );
     }
 
+    /// Converts a list-of-lists back into a map, for easy assertions that don't depend on
```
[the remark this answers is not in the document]
nfelt: Heh, I'll take a "clever" :) Yeah, I realize it depends on the orphan rule; originally I was thinking it would just be a new test-only trait with `fn get()` but then I realized `Index` could be used and makes it even pithier.

But I think the macro version is quite nice and addresses the main things I was lukewarm about with `unroll_map` so I'm perfectly happy to keep it the way you have it now.

tensorboard/data/server/server.rs
nfelt: Thanks for the detailed thoughts!  Didn't mean to make this too much of a digression, totally fine to leave as just a potential TODO for now.  Maybe if I have any particularly good ideas I'll take a stab at this myself :)

nfelt (approved):

tensorboard/data/server/server.rs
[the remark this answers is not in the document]
wchargin: Yeah, I would also be totally fine with a test-only extension trait (no
orphan rule issues since the trait is in-crate). And I would probably
implement those with a macro.

I still might switch to that. It has type information earlier, so it’s
easier to return a `HashMap<Run, HashMap<Tag, V>>` given a mapping
function `FnMut(Self::TagEntry) -> V`. And that would be nice to avoid
the newly introduced `.as_ref()` dances. You can do it with the macro,
but it’s a bit annoying because of inference details. (Basically: you
can’t write `(|x| x.foo())(some_x)` without explicitly specifying the
type for `x`, but you can work around this by capturing `x`’s type, Java
style: `std::iter::once(some_x).map(|x| x.foo()).nth(0).unwrap()` is
one awkward way to do so.)

tensorboard/data/server/server.rs
wchargin: Yep, TODO included. And, sure, feel free to grab it. :-)

PR #4388, opened 2020-11-25
rust: install logging framework

wchargin:
Summary:
Our main function now installs an `env_logger` log sink that prints to
stderr. By default, logs at INFO level and above are printed. This can
be customized by setting the `RUST_LOG` environment variable to a
filter: e.g., `RUST_LOG=warn` shows only logs at WARN level and up,
whereas `RUST_LOG=warn,mio=trace` shows WARN logs for most modules but
TRACE logs for modules that start with `mio`, such as `mio::poll`. For
more details about the filter grammar, see the [`env_logger` docs][1].

This patch changes all our `eprintln!`s into log statements, either
`warn!`s or `error!`s, as appropriate. We may still want to clean those
up to be more friendly: e.g., deduplicating errors. But now at least the
`RUST_LOG` variable provides a recourse against spammy logs.

[1]: https://docs.rs/env_logger/0.8.2/env_logger/#enabling-logging

Test Plan:
Run `bazel run //tensorboard/data/server -- path/to/logdir`, pointing to
a logdir with some scalar data and graphs. Note that INFO messages are
printed to bracket the load cycles, and WARN messages are printed
because `graph_def` events are not yet handled.

Then, re-run with `RUST_LOG=warn,mio=trace`, and note that the INFO logs
disappear but a TRACE log “registering with poller” from `mio::poll`
appears instead.

A `git grep eprintln` shows that the only remaining usage is the exit
from `main` when no arguments are passed, which is appropriate.

wchargin-branch: rust-add-logging


stephanwlee (approved):
For some reason, I thought you wrote the whole logger. Seems straight-forward :)

wchargin:
Renamed to “install logging framework”. :-)

PR #4389, opened 2020-11-25
cleanup: remove TypeScript build from tf_web_library

stephanwlee:
After our build system upgrade, we no longer use tf_web_library to build
the TypeScript files. Instead, we use the rules based on rules_nodejs
and ts_library.

Test CL: cl/344187444


wchargin (approved):
Thanks!

PR #4390, opened 2020-11-25
cleanup: remove unused tf_import and workspace dep

stephanwlee:
With the Polymer 3 migration, we no longer depend on many deps defined in tf_imports. This change
removes unused targets and WORKSPACE configs to manually pull in the dependency.

Test cl: cl/344187444


bileschi (approved):

wchargin (approved):
hi, sorry, just saw this: LGTM as long as sync is okay.

PR #4391, opened 2020-11-25
cleanup: refactor chai/mocha test to jasmine

stephanwlee:
Angular works the best when using jasmine and most of our codebase
already adopted jasmine. To avoid using multiple libraries, we decided
to refactor existing tests into jasmine.


psybuzz (approved):
thanks for this!

PR #4392, opened 2020-11-25
cleanup: remove dep to chai/sinon

stephanwlee:
After 357863f, we no longer require dependency on sinon and chai.


psybuzz (approved):
nice!

PR #4395, opened 2020-11-25
data: represent scalars as `f32`

wchargin:
Summary:
In-memory and wire representations of scalars are now 32-bit rather than
64-bit floats. We no longer support reading `DT_DOUBLE` tensors, which
is fine because TensorBoard has never written such tensors to event
files. This is mostly a no-op on the Python side, except that Python
`float`-to-`float` conversions that go through the data provider layer
are now lossy.

Per a comment of @nfelt on #4352.

Test Plan:
All the Rust changes follow from the types. Tests suffice, as usual.

wchargin-branch: data-scalars-f32


wchargin:
It looks like this has no significant effect on load time (giant dataset
goes from 190.0s to 189.9s, *N* = 1; probably noise).

tensorboard/data/server/run.rs
```diff
@@ -365,7 +365,7 @@ mod test {
         );
 
         // Points should be as expected (no downsampling at these sizes).
-        let scalar = |f: f32| commit::ScalarValue(f64::from(f));
+        let scalar = commit::ScalarValue;
```
nfelt, line 368: Could probably be inlined now, but fine if you want to leave as is. Just FYI.

nfelt (approved):

tensorboard/data/server/run.rs
wchargin: Yeah, but rustfmt wraps it if you inline it—

```rust
            vec![
                (
                    Step(0),
                    WallTime::new(1235.0).unwrap(),
                    &commit::ScalarValue(0.25)
                ),
```

—which makes it less readable, IMO.

(This is one thing that I like about `gofmt`’s “never break lines”
philosophy: sometimes it really is clearer to just overrun by 10 columns
or so. But being able to just write exorbitantly long lines and have
them fully beautified is certainly nice, too.)

nfelt: Huh okay. My intuition about when `rustfmt` is going to line-wrap something must be a little off since I had eyeballed it relative to other long lines and it looked fine, and now when I actually check the length it looks like it would have been 87 chars and I thought the line limit was 100?  Does this have something to do with the width heuristics configuration option?  Or am I missing something more obvious than that?

wchargin: > Does this have something to do with the width heuristics configuration
> option?

Plausibly? I don’t know. I don’t configure `rustfmt`; I leave it at its
defaults, which are good enough for me, and let it do its job. Unlike
Black and Prettier, Rustfmt _does_ have a bunch of options, and I guess
I’m not _strongly_ opposed to flipping some of them, but there is value
in just using the defaults, since things will just work by default in
more contexts.

(I have run into issues where a project’s `rustfmt.toml` specifies
dozens of options, some of which my `rustfmt` honors and others of which
it doesn’t. No idea why. Probably shouldn’t happen. But definitely
doesn’t happen if you just leave the defaults be.)

wchargin: FWIW:
<https://fuchsia.googlesource.com/fuchsia/+/719e5812fc273d1e32ac74d6a8e2c45862b52873/rustfmt.toml#7>
(plus more Google-internal references to the same copied file)

nfelt: Yeah, I read some of the GitHub issues talking about the width heuristics but wasn't sure exactly which one would be causing this and was figuring you might know.

`<soapbox>`
I don't love the width heuristics approach; for autoformatters I'd rather that the main rules they follow be clear enough that I can reliably predict what it's going to do ahead of time in my head, and having various different line length limits that are context-dependent makes that much harder. But based on the issue linked from the fuchsia rustfmt.toml (https://github.com/rust-lang/rustfmt/issues/3120#issuecomment-432485199) it sounds like they're more likely to move in the opposite direction and adopt an even more complex heuristic in hopes the output is better.  If that's still the plan it's a bit disappointing since it feels like curing the symptom to me...
`</soapbox>`

PR #4397, opened 2020-11-25
rust: add nice command-line argument parsing

wchargin:
Summary:
This patch outfits `//tensorboard/data/server` with a nice CLI, courtesy
of [`clap`]. The flags are a subset of the main TensorBoard flags, but
with kebab-case rather than snake-case (per everyone-other-than-Google
style, via `clap` defaults). Since we now accept a `--port` argument,
which may be `0`, we also print the port to which the server is bound.

The `clap` approach to argument parsing is “define a struct all of whose
fields implement `FromStr`, then mark it `#[derive(Clap)]` to generate
an argument parser”. This is great on both the static typing and runtime
performance fronts. To add more arguments, we just add more fields to
`struct Opts`. See <https://docs.rs/clap/3.0.0-beta.2> for details (this
syntax is introduced in `clap` v3.x, which is still in beta).

[`clap`]: https://crates.io/crates/clap

Test Plan:
Run `bazel run //tensorboard/data/server` with `-h` and `--help` to see
short and long help text, respectively. Then play around with each
argument in turn.

wchargin-branch: rust-cli

### DIFF
--- a/tensorboard/data/server/main.rs
+++ b/tensorboard/data/server/main.rs
@@ -13,50 +13,99 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-use log::{info, LevelFilter};
+use clap::Clap;
+use log::{debug, info, LevelFilter};
+use std::net::{IpAddr, SocketAddr};
 use std::path::PathBuf;
+use std::str::FromStr;
 use std::time::{Duration, Instant};
+use tokio::net::TcpListener;
 use tonic::transport::Server;
 
 use rustboard_core::commit::Commit;
 use rustboard_core::logdir::LogdirLoader;
-use rustboard_core::proto::tensorboard::data::tensor_board_data_provider_server::TensorBoardDataProviderServer;
+use rustboard_core::proto::tensorboard::data;
 use rustboard_core::server::DataProviderHandler;
 
-const RELOAD_INTERVAL: Duration = Duration::from_secs(5);
+use data::tensor_board_data_provider_server::TensorBoardDataProviderServer;
+
+#[derive(Clap, Debug)]
+#[clap(name = "rustboard", version = "0.1.0")]
+struct Opts {
+    /// Log directory to load
+    ///
+    /// Directory to recursively scan for event files (files matching the `*tfevents*` glob). This
+    /// directory, its descendants, and its event files will be periodically polled for new data.
+    #[clap(long)]
+    logdir: PathBuf,
+
+    /// Bind to this IP address
+    ///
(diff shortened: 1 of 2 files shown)


tensorboard/data/server/main.rs
```diff
@@ -13,50 +13,99 @@ See the License for the specific language governing permissions and
...
+    /// directory, its descendants, and its event files will be periodically polled for new data.
+    #[clap(long)]
+    logdir: PathBuf,
+
+    /// Bind to this IP address
+    ///
+    /// IP address to bind this server to. May be an IPv4 address (e.g., 127.0.0.1 or 0.0.0.0) or
+    /// an IPv6 address (e.g., ::1 or ::0).
+    #[clap(long, default_value = "::0")]
+    host: IpAddr,
```
stephanwlee, line 47: While I am glad validation is built-in to clap and typing can make sure we do not have, say, negative port number, I am wondering who is in charge of validating the IpAddr. Is it part of `IpAddr#FromStr`?  If so, how about `PathBuf`?

stephanwlee (approved):
Looks good but I would like to hear your response :)

tensorboard/data/server/main.rs
wchargin: > Is it part of `IpAddr#FromStr`? If so, how about `PathBuf`?

Yep: in all cases, the [`FromStr`] implementation is used, so, for example,
[`<IpAddr as FromStr>`], [`<PathBuf as FromStr>`]. (On the doc pages,
you can click `[src]` at right to see the implementations if you want.)

This is also why we `impl FromStr for Seconds` below.

[`FromStr`]: https://doc.rust-lang.org/std/str/trait.FromStr.html
[`<IpAddr as FromStr>`]: https://doc.rust-lang.org/std/net/enum.IpAddr.html#impl-FromStr
[`<PathBuf as FromStr>`]: https://doc.rust-lang.org/std/path/struct.PathBuf.html#impl-FromStr

wchargin: Or, maybe this helps: you can tell that the `FromStr` implementations
are responsible for validation because of the types:

```rust
fn from_str(s: &str) -> Result<Self, Self::Err>;
```

The docs for [`from_str`] also say this explicitly, but the signature is
enough to imply it: a method that returns a `Result` is generally
expected to perform validation of some kind and signal that through the
return value rather than panicking.

[`from_str`]: https://doc.rust-lang.org/std/str/trait.FromStr.html#tymethod.from_str

stephanwlee: Got it, thanks.

PR #4398, opened 2020-11-26
rust: add `--verbose` to control log level

wchargin:
Summary:
We now only show WARN logs by default, showing INFO at `-v` or all logs
at `-vv` or higher.

Test Plan:
Run `bazel run //tensorboard/data/server` with a `--logdir` that has
graphs. By default, that should print a warning that graphs are not yet
supported. With `-v` or `--verbose` or `--verbose=1`, it should also
print load cycle timing information. With `-vv` or `--verbose --verbose`
or `--verbose=2` (or more), it should print a debug log including the
parsed options, and also some `mio::poll` internals.

wchargin-branch: rust-verbose-flag


tensorboard/data/server/main.rs
```diff
@@ -58,6 +58,10 @@ struct Opts {
     /// does not include the time for the reload itself.
     #[clap(long, default_value = "5")]
     reload_interval: Seconds,
+
+    /// Use verbose output (-vv for very verbose output)
+    #[clap(long = "verbose", short, parse(from_occurrences))]
+    verbosity: u32,
```
stephanwlee, line 64: jw, not that it matters, do we need u32 or does u8 suffices?

stephanwlee (approved):

tensorboard/data/server/main.rs
wchargin: Yeah, `u8` would work fine here. I tend to default to 32- or 64-bit
integers for general numeric storage to make overflow-like issues less
of a problem. Using `u8` wouldn’t make this struct any smaller (because
of alignment), and I would surprised if it made anything faster at all:
even ignoring that performance obviously isn’t an issue in this case,
you can often get better performance by using 32-bit or 64-bit types.
Though it looks like my machine actually might like `u8` here, since my
[`sizeof(uint_fast8_t)`][fast] is `1`:

```
$ cat test.c
#include <stdint.h>
#include <stdio.h>
#define TEST(n) printf("% 24s = %zu\n", \
        "sizeof(uint_fast"#n"_t)", sizeof(uint_fast##n##_t))
int main() { TEST(8); TEST(16); TEST(32); TEST(64); }
$ make test && ./test
cc     test.c   -o test
    sizeof(uint_fast8_t) = 1
   sizeof(uint_fast16_t) = 8
   sizeof(uint_fast32_t) = 8
   sizeof(uint_fast64_t) = 8
```

[fast]: https://en.cppreference.com/w/c/types/integer

Anyway, any unsigned type is probably fine. :-)

PR #4402, opened 2020-11-30
line chart: adjust opacity in webgl using HSL

stephanwlee:
Currently, we are making things "brighter" in rgb color space which does
not look correct. Since brightness or opacity is better represented via
light in HSL, we use the hsl color space.


psybuzz (approved):
The resulting, adjusted color is now discontinuous over the opacity domain (adjusted color for opacity 0.99 is close to that of opacity 0.98, but very different from opacity 1).  lgtm if this is the desired effect.

stephanwlee:
Yup, it is discontiguous and I was okay with it (at least for now; given that we have hardcoded to opacity = 0.3)

PR #4403, opened 2020-11-30
rust: implement RPC-time downsampling

wchargin:
Summary:
A new `downsample` function performs TensorBoard-style downsampling of a
vector: random subsequence, relative order preserved, last element
always kept. The RPC server uses this to honor the downsampling
specifications for `ReadScalars` RPCs.

Test Plan:
Unit tests included for both the utility function and the RPC.

wchargin-branch: rust-downsample


tensorboard/data/server/downsample.rs
```diff
@@ -0,0 +1,109 @@
...
+        let xs: Vec<char> = "abcdefg".chars().collect();
+        let expected = downsample_cloned(&xs, 4);
+        assert_eq!(expected.len(), 4);
+        for _ in 0..100 {
+            assert_eq!(downsample_cloned(&xs, 4), expected);
+        }
+    }
+
+    #[test]
+    fn test_underlong_ok() {
```
nfelt, line 79: what is "underlong"?

nfelt (approved):

tensorboard/data/server/downsample.rs
wchargin: Converse of [*overlong*][1], which is somewhat common in encoding
contexts (e.g., “Modified UTF-8 is UTF-8 but with the 2-overlong
encoding of NUL”). Granted, I made up *underlong*, but it sounds better
than *overshort*. :-)

I can go with `test_k_less_than_n_ok` or something if you prefer.

[1]: https://stackoverflow.com/questions/7113117/what-is-exactly-an-overlong-form-encoding

tensorboard/data/server/server.rs
```diff
@@ -288,7 +292,7 @@ fn parse_downsample(downsample: Option<data::Downsample>) -> Result<i64, Status>
             num_points
         )));
     }
-    Ok(num_points)
+    Ok(num_points.try_into().unwrap_or(usize::MAX))
```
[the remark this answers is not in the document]
wchargin: This only matters if you’re on a 32-bit machine *and* you request at
least 2^32 points *and* your reservoir capacity is at least 2^32 *and*
that many points are actually in the reservoir, right? If any one of
those fails, this is just a lossless conversion (or at least that was
the intent—did I mess it up?).

Do we care about that case? As you say, truncating an RPC with 5 billion
points down to merely 4 billion points feels like the least of your
worries in that scenario, and I wasn’t really even intending to publish
32-bit binaries. TensorFlow doesn’t:
<https://pypi.org/project/tensorflow/#files>

(I would actually believe a claim that it can’t possibly matter, since
protos greater than 2 GiB are unsupported basically everywhere due to
legacy parser bugs.)

Somehow I got the idea that we should prefer 64-bit types to 32-bit
types as a general guideline, though looking through <https://aip.dev>,
I don’t see that advice given. Still, without a good reason to make this
32-bit, I’m weakly inclined to keep it 64-bit (and of course it makes no
difference on the wire, since they’re all varints). WDYT?

tensorboard/data/server/downsample.rs
nfelt: Yeah, I'd prefer a more straightforward name - even "overlong" would be obscure IMO given that it's not a particularly common term (as the SO question itself indicates) and its applicability to this scenario is far from obvious to me.  For that matter isn't this `test_k_greater_than_n_ok`?  So I'm not sure I even understand the metaphor despite the intention having been explained to me :)

tensorboard/data/server/server.rs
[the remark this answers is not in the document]
nfelt: \*shrug\* I agree it's a marginal edge case that won't occur in practice.  It just seemed like we might as well not even need to spend a line of code doing something to handle the "too large value" edge case when we could avoid it entirely by changing `downsample` and communicate up-front in the RPC API that it can't be unreasonably large.

I guess I'd be a little happier if we are going to keep this line if we change it to just hard-fail a request that sets downsample above a certain limit (either `usize::MAX`, or we could set a lower limit that's closer to the practical range, like 1000 or 10k).  It just feels like doing the truncation to `usxize::MAX` is in principle a weird thing to do since it subverts what the user's requested behavior was in a way that wouldn't be readily obvious to the caller / isn't documented.

nfelt (approved):

tensorboard/data/server/server.rs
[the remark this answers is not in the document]
wchargin: Sure; it feels awfully marginal to me, indeed. Another reason that this
is impossible to hit: on a 32-bit system, this would just fail to
allocate memory for the response, since even with unbounded swap your
address space simply isn’t large enough. So the potentially confusing
behavior never happens.

But no problem: we can hard-fail the request. Done.

tensorboard/data/server/downsample.rs
wchargin: Ah, my bad; “*k* greater than *n*” is indeed what I meant. See, this is
why I like having a name for the concept rather than fiddling with the
algebra. ;-)

Done, as `test_ok_when_k_greater_than_n`.

PR #4405, opened 2020-11-30
line chart: implement ignoreOutlier

stephanwlee:
This change backports some logic from vz-chart-helpers to ignore y
dimension outliers when computing the default viewBox.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/line_chart_internal_utils.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/line_chart_internal_utils.ts
@@ -23,20 +23,19 @@ import {isWebGl2Supported} from './lib/utils';
 /**
  * Returns extent, min and max values of each dimensions, of all data series points.
  *
- * When ignoreYOutliers is true, it will filter out values that are approximately [1] less
- * than 5th and greater than 95th percentile when calculating y-extent.
+ * When ignoreYOutliers is true, it will calculate extent using values within 5th and 95th
+ * quantiles.
  *
  * Note that it excludes auxillary data points and invisible data series.
- *
- * [1]: Uses R-7 method for approximation: https://github.com/d3/d3-array#quantile
  */
 export function computeDataSeriesExtent(
   data: DataSeries[],
   metadataMap: DataSeriesMetadataMap,
   ignoreYOutliers: boolean
 ): {x: [number, number] | undefined; y: [number, number] | undefined} {
+  let xMin: number | null = null;
+  let xMax: number | null = null;
   let yPoints: number[] = [];
-  const xPoints: number[] = [];
 
   let pointIndex = 0;
   for (const {id, points} of data) {
@@ -46,7 +45,8 @@ export function computeDataSeriesExtent(
     for (let index = 0; index < points.length; index++) {
       const {x, y} = points[index];
       if (Number.isFinite(x)) {
-        xPoints.push(x);
+        xMin = xMin === null || x < xMin ? x : xMin;
+        xMax = xMax === null || x > xMax ? x : xMax;
       }
       if (Number.isFinite(y)) {
         yPoints.push(y);
@@ -56,21 +56,16 @@ export function computeDataSeriesExtent(
   }
 
   yPoints.sort(d3.ascending);
+  let yMin = yPoints[0];
+  let yMax = yPoints[yPoints.length - 1];
 
   if (ignoreYOutliers && yPoints.length > 2) {
-    const aY = d3.quantile(yPoints, 0.05)!;
-    const bY = d3.quantile(yPoints, 0.95)!;
-    yPoints = yPoints.filter((val) => {
-      return aY <= val && val <= bY;
-    });
+    yMin = yPoints[Math.ceil((yPoints.length - 1) * 0.05)];
+    yMax = yPoints[Math.floor((yPoints.length - 1) * 0.95)];
   }
 
-  const [xMin, xMax] = d3.extent(xPoints);
-  const yMin = yPoints[0];
-  const yMax = yPoints[yPoints.length - 1];
-
   return {
-    x: xMin !== undefined && xMax !== undefined ? [xMin, xMax] : undefined,
+    x: xMin !== null && xMax !== null ? [xMin, xMax] : undefined,
     y: yMin !== undefined && yMax !== undefined ? [yMin, yMax] : undefined,
   };
 }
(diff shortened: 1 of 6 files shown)


nfelt (commented):
Hmm, so I thought you had a good argument here for keeping smoothing out of the line chart: https://github.com/tensorflow/tensorboard/pull/4359#issuecomment-730706272

Do we not have the same reasons to keep outlier pruning out of the line chart?  I guess perhaps it's slightly different since it doesn't change the actual data, but it also seems like something that's logically a little separate.  In the post above you made the point that PR curves don't need smoothing; they also don't need to ignore outliers (and they do need to set the Y-range to a pre-determined value, so if we expose that option, the ignore-outliers mode could just be code at the callsite that generates the extent to use?).

Also, the motivation of making it easier to change the outlier logic seems like a strong one, since I've always disliked how we interpret "ignore outliers" as just truncating 10% of the data no matter what the distribution is; I would be much happier if we could do something that's more sensitive to the actual distribution, like excluding points beyond X% of the inter-quartile range, or beyond 2 standard deviations, etc.  At that level it might even make sense as a tunable parameter like smoothing.

stephanwlee:
> Do we not have the same reasons to keep outlier pruning out of the line chart?

Unfortunately, while current code does not handle it properly, the extent calculation needs to take scale into account and ignoring the outlier should be also a bit aware of it.

For instance, let's use log scale as an example with the data series spanning negative to positive values. Of course, we need to filter out non-positive values (which we do not right now which is why you see `[Number.MIN_VALUE, max_val]` with any non-positive values and cause a very janky graph) and calculate the extent of the legal values (you can imagine doing the same for non-finite values for linear scale). At this point, I am sure you understand the issue that I am about to say but, I will say it out loud.

Doing `ignore_outlier(scale_aware_value_clip(data_series))` results in a very different results than doing `scale_aware_value_clip(ignore_outlier(data_series))`. In the former case, both lower and upper outliers get pruned out, while, in the latter case, only the upper outliers get pruned out since lower bounds are mostly negative values which will be clipped away.

I can see a different formulation of the line chart and expect consumers to calculate extent and always pass `viewBox` and expose scale to a public API. However, that would weaken the ergonomics of the line chart. Given that we do not have any plans to give options to change outlier pruning algorithm (as opposed to data transformation/smoothing), I was okay with ignore outlier being a first party concept of the chart.

tensorboard/webapp/widgets/line_chart_v2/line_chart_internal_utils.ts
```diff
@@ -22,44 +23,55 @@ import {isWebGl2Supported} from './lib/utils';
...
     if (!meta || meta.aux || !meta.visible) continue;
 
     for (let index = 0; index < points.length; index++) {
       const {x, y} = points[index];
-      if (!Number.isNaN(x)) {
-        xMin = Math.min(xMin, x);
-        xMax = Math.max(xMax, x);
-        xExtentChanged = true;
+      if (Number.isFinite(x)) {
+        xPoints.push(x);
```
nfelt, line 49: Optional but since we still only need the min and max, it seems more efficient to stick with manually tracking those, rather than using an extra O(n) memory to create xPoints and then another O(n) compute to traverse it in `d3.extent` below?  I guess in practice maybe it makes little difference but if we're trying to emphasize perf here it seems a little weird to make this slower.

perhaps slightly unnecessary to accumulate a copy of the array of points?  I guess in practice this is probably pretty cheap but if we're trying to emphasize performance it

tensorboard/webapp/widgets/line_chart_v2/line_chart_internal_utils.ts
```diff
@@ -22,44 +23,55 @@ import {isWebGl2Supported} from './lib/utils';
...
+      pointIndex++;
     }
   }
 
+  yPoints.sort(d3.ascending);
+
+  if (ignoreYOutliers && yPoints.length > 2) {
+    const aY = d3.quantile(yPoints, 0.05)!;
+    const bY = d3.quantile(yPoints, 0.95)!;
+    yPoints = yPoints.filter((val) => {
```
nfelt, line 63: We could skip the extra O(n) traversal here by just replicating the logic in `quantileSorted` ourselves and leaving out the approximation step - we aren't actually using it, given that we define the `extent` based not on the quantile value itself, but on the filtered data.  So it's the same as if we made the approximation use the index 5% of the way into the array (rounded up) and 95% of the way into the array (rounded down), since the result of the filtering is just to truncate the array to that middle section anyway.

So this should be functionally equivalent to what we have now:

```typescript
if (ignoreYOutliers && yPoints.length > 2) {
    yMin = yPoints[Math.ceil((yPoints.length - 1) * 0.05)];
    yMax = yPoints[Math.floor((yPoints.length - 1) * 0.95)];
}
```

Optional since if we change this to use the inter-quartile range or something else a little bit more statistically justified (IMO) we'll probably have to go back to using `filter()`, but just pointing it out in case we stick with this outlier definition.

tensorboard/webapp/widgets/line_chart_v2/line_chart_internal_utils.ts
```diff
@@ -22,44 +23,55 @@ import {isWebGl2Supported} from './lib/utils';
 /**
  * Returns extent, min and max values of each dimensions, of all data series points.
  *
+ * When ignoreYOutliers is true, it will filter out values that are approximately [1] less
+ * than 5th and greater than 95th percentile when calculating y-extent.
+ *
  * Note that it excludes auxillary data points and invisible data series.
  *
- * TODO(stephanwlee): add support for ignoreOutlier.
+ * [1]: Uses R-7 method for approximation: https://github.com/d3/d3-array#quantile
```
nfelt, line 31: FWIW, no change needed but as discussed below, the way we actually use the quantiles I think it doesn't really matter what the approximation algorithm is since we're just using the result to truncate anyway.

nfelt (approved):
Thanks for the explanation, I hadn't thought about how this is scale-dependent so I agree it that case it's more reasonable to think of it as built-in.  (I guess the PR curve charts also don't have log scale, so maybe there's some decomposition in which both the non-linear scaling and the outlier detection are sort of a layer on top of an even more stripped-down chart, but I can see how maybe it's not worth trying to unbundle that.)

tensorboard/webapp/widgets/line_chart_v2/line_chart_internal_utils.ts
```diff
@@ -22,44 +23,55 @@ import {isWebGl2Supported} from './lib/utils';
...
+        yPoints.push(y);
       }
+      pointIndex++;
     }
   }
 
+  yPoints.sort(d3.ascending);
+
+  if (ignoreYOutliers && yPoints.length > 2) {
+    const aY = d3.quantile(yPoints, 0.05)!;
```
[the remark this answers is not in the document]
stephanwlee: Quantile sorted is a recent addition that we cannot use since we are stuck on d3 5.7.0 (internal version)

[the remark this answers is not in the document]
nfelt: Ah, got it.  In that case though I think that's a stronger case to just reimplement the index-calculation logic (which is just one line) and do a direct lookup into the sorted array rather than doing two unnecessary scans.

PR #4409, opened 2020-12-01
Upgrade Black to v20.8b1 and drop py27 and p35

### LINKED ISSUE #4406: Black updates: drop py27+py35; upgrade from 19.x to 20.x
Our `pyproject.toml` has:

```toml
target-version = ["py27", "py35", "py36", "py37", "py38"]
```

Now that we’ve flipped `srcs_version` to `PY3` and let that soak for a
bit with no problems (#4361), we can safely drop py27 compatibility, and
we can also drop py35 since TensorFlow has done so and we’ve announced
as much in our release notes (#4320).

The main changes are removing `u""` sigils and permitting trailing
commas after splats, since [bpo-9232] was fixed. Diffstat is +112/−118
across 25 files.

I also propose that we upgrade from Black 19.x to 20.x. One concrete
improvement on the tooling side is that editor integrations now respect
`pyproject.toml`, and this depends on a helper added in 20.x. And, as
usual, staying reasonably up to date is nice to have.

The main changes seem to be adding newlines where there was an inline
trailing comma:

```diff
         self.assertTagsEqual(
             accumulator.Tags(),
-            {ea.IMAGES: tags, ea.GRAPH: True, ea.META_GRAPH: False,},
@@ note this comma ------------------------------------------------^
+            {
+                ea.IMAGES: tags,
+                ea.GRAPH: True,
+                ea.META_GRAPH: False,
+            },
         )
```

I don’t know why we have quite so many of those, and I’m not sure
whether they all should really exist, but given that they’re there, the
change seems reasonable to me. Diffstat is +854/−259 over 84 files.

As usual, I propose that we do this in three steps: (a) update configs
and temporarily disable Black enforcement, (b) reformat everything, and
(c) re-enable enforcement. This way, we can mark the commit(s) of (b) as
ignored for `git-blame` purposes via our `.git-blame-ignore-revs` file.
We can still handle these with a single PR, just merged with rebase.

[bpo-9232]: https://bugs.python.org/issue9232

wchargin:
Fixes #4406. See that issue for discussion.

To be merged with rebase.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,8 +65,7 @@ jobs:
       - run: pip freeze --all
       - name: 'Lint Python code for style with Black'
         # You can run `black .` to fix all Black complaints.
-        # TODO(#4406): Re-enable this check.
-        run: black --check --diff . || true
+        run: black --check --diff .
 
   lint-rust:
     runs-on: ubuntu-16.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
         # flake8 should run on each Python version that we target,
         # because the errors and warnings can differ due to language
         # changes, and we want to catch them all.
-        python_version: ['3.5', '3.7']
+        python_version: ['3.6', '3.7']
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-python@v1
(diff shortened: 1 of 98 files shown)


wchargin:
(Force-push to attribute reformats to TensorBoard Gardener.)

wchargin:
CLAs are okay because I wrote or generated all these commits.

psybuzz (approved):
Nice!

wchargin:
Thanks!

PR #4412, opened 2020-12-01
Revert "cleanup: move source code component from debugger-v2 to widgets"

nfelt:
Reverts tensorflow/tensorboard#4396 since it broke the sync into Google.

WANT_LGTM=any


wchargin (approved):
Approving since reverting this locally has sufficed to unbreak the sync
for me. All else equal, I have a tiny bit of weight in favor of fixing
the Copybara config if that’s easy, because then we could sync the Black
reformat in in its own cleanup CL, but not enough to object to this PR,
which certainly is following standard practices.

nfelt:
OK by me to fix forward in Copybara; github actions is hosed today anyway: https://www.githubstatus.com/incidents/sxthxvdlljk4

psybuzz (approved):
Thanks for reverting.  I'm looking into the fix-forward, and will post here when I have something.

If GH actions frees up faster than me fixing forward, feel free to land this first.

psybuzz:
@Googlers, internal pending change cl/345086597 is a fix-forward.  It has not yet passed internal tests (waiting on cl/345087038).  Once it does pass internal tests, this PR should be obsolete.  Depending on whichever finishes first, this PR's GH action tasks or the CL's internal tests, we must land one and close the other.

psybuzz (changes requested):
Removing approval because the internal fix-forward passed tests and is ready to land.

Thanks for this revert PR though :)

PR #4416, opened 2020-12-02
line chart: improve worker doc and check for assumption

stephanwlee:
Internally, due to conformance reasons, we cannot support a relative
path for the worker resource. As such, we are now modifying the contract
and accompany the change with a check.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/worker.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/worker.ts
@@ -20,7 +20,7 @@ limitations under the License.
  *
  * @param workerJsFilename Worker JavaScript resource file name. Exact resource path
  *   depends on context. It can be relative or absolute depending on a app and its
- *   configuration (internally, it is absolute; i.e., served at "/<resource_name.js>").
+ *   configuration. Please see the internal version of this module for details.
  */
 export function getWorker(workerJsFilename: string): Worker {
   if (workerJsFilename.includes('/')) {
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/worker.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/worker.ts
@@ -18,10 +18,15 @@ limitations under the License.
  *
  * This module exists to conform to internal requirements.
  *
- * @param workerResourcePath URL pathanme to the JavaScript resource served by
- *   TensorBoard. TensorBoard disallows fetching JavaScript resources from a differnt
- *   origin.
+ * @param workerJsFilename Worker JavaScript resource file name. Exact resource path
+ *   depends on context. It can be relative or absolute depending on a app and its
+ *   configuration (internally, it is absolute; i.e., served at "/<resource_name.js>").
  */
-export function getWorker(workerResourcePath: string): Worker {
-  return new Worker(workerResourcePath);
+export function getWorker(workerJsFilename: string): Worker {
+  if (workerJsFilename.includes('/')) {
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/worker.ts
```diff
@@ -18,10 +18,15 @@ limitations under the License.
...
- *   TensorBoard. TensorBoard disallows fetching JavaScript resources from a differnt
- *   origin.
+ * @param workerJsFilename Worker JavaScript resource file name. Exact resource path
+ *   depends on context. It can be relative or absolute depending on a app and its
+ *   configuration (internally, it is absolute; i.e., served at "/<resource_name.js>").
  */
-export function getWorker(workerResourcePath: string): Worker {
-  return new Worker(workerResourcePath);
+export function getWorker(workerJsFilename: string): Worker {
+  if (workerJsFilename.includes('/')) {
```
psybuzz, line 26: I see that today we are serving the script via the assets `tf_web_library` with `path = "/"`, but is there any real reason why it must be this way, rather than served from, say, "/workers/chart_worker.ts" in the future?  What is the bad case that this check helps us avoid?

tensorboard/webapp/widgets/line_chart_v2/lib/worker.ts
```diff
@@ -18,10 +18,15 @@ limitations under the License.
  *
  * This module exists to conform to internal requirements.
  *
- * @param workerResourcePath URL pathanme to the JavaScript resource served by
- *   TensorBoard. TensorBoard disallows fetching JavaScript resources from a differnt
- *   origin.
+ * @param workerJsFilename Worker JavaScript resource file name. Exact resource path
+ *   depends on context. It can be relative or absolute depending on a app and its
+ *   configuration (internally, it is absolute; i.e., served at "/<resource_name.js>").
```
psybuzz, line 23: Perhaps consider replacing "internally, it is absolute..." with a redirection "please see the internal version of this file for details", in case this comment goes out of sync with internal implementation?

psybuzz (approved):

psybuzz (changes requested):
Sorry, I was a little hasty with buttons and wanted to click "Comment" to get info before approving.

tensorboard/webapp/widgets/line_chart_v2/lib/worker.ts
stephanwlee: In OSS, we have to support path_prefix which we really don't internally (I guess it is still possible to use the flag internally but manual invocation of TB is not the normative case anymore).

psybuzz: So calling `getWorker('/foo.js')` is an error because it wouldn't work with path_prefix.  If someone called `getWorker('foo/bar.js')`, if would be valid as long as we used a tf_web_library in OSS with `path="foo"` (I think?).  Does it make sense to change the condition to `workerJsFileName.startsWith('/')`?

stephanwlee: While it would be valid, there is no easy way to support even "foo/bar.js" so we are giving up and just putting a constraint that it has to be a name of the file (i.e., "bar.js" in your example) and that's it.

psybuzz: Gotcha, thanks for the explanation

psybuzz (approved):

PR #4418, opened 2020-12-02
ui: allow disabled dropdown options

psybuzz:
This introduces a new field on the `DropdownOption` public interface
to let clients of the `<tb-dropdown>` component specify options that
are non-interactable.

image

### DIFF
--- a/tensorboard/webapp/widgets/dropdown/dropdown_component.ts
+++ b/tensorboard/webapp/widgets/dropdown/dropdown_component.ts
@@ -18,8 +18,7 @@ export interface DropdownOption {
   value: any;
   displayText: string;
   /**
-   * Whether the option should appear non-interactable. All options are
-   * 'enabled' by default.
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/widgets/dropdown/dropdown_component.ts
```diff
@@ -17,16 +17,28 @@ import {Component, EventEmitter, Input, Output} from '@angular/core';
 export interface DropdownOption {
   value: any;
   displayText: string;
+  /**
+   * Whether the option should appear non-interactable. All options are
```
stephanwlee, line 21: nit: drop "All options are"?

stephanwlee (approved):

tensorboard/webapp/widgets/dropdown/dropdown_component.ts
psybuzz: Done, thanks

PR #4419, opened 2020-12-02
ui: fix styles in dropdowns, smoothing input

psybuzz:
The Time Series settings panel uses dropdowns and text
input fields. This fixes 2 issues:
- When the dropdown is focused, there is no visual indication
- In Firefox, the smoothing input flows offscreen

Before
image

After
image

### DIFF
--- a/tensorboard/webapp/metrics/views/right_pane/settings_view_component.scss
+++ b/tensorboard/webapp/metrics/views/right_pane/settings_view_component.scss
@@ -62,6 +62,12 @@ section .control-row:not(:last-child) {
   }
 }
 
+.scalars-smoothing .slider-input {
+  flex: none;
+  // Wide enough to show 3 digits after a decimal.
+  width: 60px;
+}
+
 mat-slider {
   flex: 1;
   // Reset mat-slider's internal extra space on left/right sides
--- a/tensorboard/webapp/metrics/views/right_pane/settings_view_component.scss
+++ b/tensorboard/webapp/metrics/views/right_pane/settings_view_component.scss
@@ -64,8 +64,9 @@ section .control-row:not(:last-child) {
 
 .scalars-smoothing .slider-input {
   flex: none;
-  // Wide enough to show 3 digits after a decimal.
-  width: 60px;
+  // Wide enough to show 3 digits after a decimal. Required since Firefox does
+  // not shrink input fields smaller than some intrinsic size.
+  width: 5em;
 }
 
 mat-slider {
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/views/right_pane/settings_view_component.scss
```diff
@@ -62,6 +62,12 @@ section .control-row:not(:last-child) {
   }
 }
 
+.scalars-smoothing .slider-input {
+  flex: none;
+  // Wide enough to show 3 digits after a decimal.
+  width: 60px;
```
stephanwlee, line 68: This is a bit surprising. Do we need it because of the flex container? If so, can we have a wrapper DOM to not hardcode a width like this? I think without this, it will have the browser default input width which may not be what you want either. In any case, maybe `5c` or something that can change with the font size like `5em` (not precisely correct but at least it is not a constant)?

tensorboard/webapp/widgets/dropdown/dropdown_component.scss
```diff
@@ -20,3 +20,10 @@ mat-select {
   box-sizing: border-box;
   padding: 6px;
 }
+
+mat-select:focus {
```
stephanwlee, line 24: Is this truly good enough? I guess it is irrelevant with current usage but if the dropdown appeared below the `<select>`, you may still want to show the focus ring when you use keyboard to select a `<option>`.

psybuzz: imo it is enough.  The default browser `<select>` doesn't show the focus ring when the list of options is opened (via keyboard or mouse).

tensorboard/webapp/metrics/views/right_pane/settings_view_component.scss
psybuzz: Changed to `em`s.

I was also surprised; even without flex, the problem still occurs.  A quick search shows others have filed bugs on behavior like this, so it seems to be a known browser incompatibility.
https://bugzilla.mozilla.org/show_bug.cgi?id=1242914

tensorboard/webapp/widgets/dropdown/dropdown_component.scss
psybuzz: imo it is enough.  The default browser `<select>` doesn't show the focus ring when the list of options is opened (via keyboard or mouse), so we'd be consistent with browser behavior.

stephanwlee (approved):

Issue #4421, opened 2020-12-03
--path_prefix broken in 2.4.0+


qrtt1:
**Describe the current behavior**

Some tensorboards installed with tensorflow couldn't access the right URLs.

Reproducible steps

1. install a tf 2.3.1
2. start tensorboard with prefix

```
tensorboard --logdir logs --port 5001 --bind_all  --path_prefix /abc/def
```

We got `failed to fetch runs`

![image](https://user-images.githubusercontent.com/193223/100859743-93a47980-34ca-11eb-9fe8-74ac99ca12f6.png)

Because some requests did not add `abc/def` prefix to the base URL:

![image](https://user-images.githubusercontent.com/193223/100860177-22b19180-34cb-11eb-9150-97855bf88f5e.png)

 
**Describe the expected behavior**

All tensorboard requests should add prefix to the base URL if `--path_prefix` has given.

**Standalone code to reproduce the issue**

I created a regression tests project to investigate the problem

https://github.com/qrtt1/tensorboard-regression/runs/1487646153

It got worse after tf 2.3.0 (they use tensorboard `2.4.0`):

![image](https://user-images.githubusercontent.com/193223/100905996-a1c2bc00-3503-11eb-9f60-27407e10d947.png)

stephanwlee:
Thanks for the report. I was able to reproduce it and it is a serious regression. It is quite interesting to see a puppeteer based tests that catches regression for TensorBoard :) 

For TensorBoarder:

This change is related to the router in the 2.4.0. In the release, we introduced a handrolled router which is unaware of notion of path prefix and eagerly redirects "/foo/bar/index.html" to "/" which breaks all the subsequent requests. Few ways to move forward:
- make the request to "/environments" before bootstrapping components to fetch the path_prefix
  - this is a bit nasty because environment has an experiment specific data_location. I think we should disentangle global setting vs. per-exp settings
  - period of blank page until the request is successful
- invest in the dehydration/rehydration scheme to populate some state when flushing down the HTML

MrBago:
@stephanwlee I just ran into this and it's nice to see that you just merged #4423, thank you! Any chance we might be able to get a patch release of 2.4 with this fix backported?

wchargin:
We do plan to publish a patch release. I’ve filed #4547 to track it.

wchargin:
[TensorBoard 2.4.1][tb] shipped with this patch.

[tb]: https://pypi.org/project/tensorboard/2.4.1/

PR #4423, opened 2020-12-04
ng: support path_prefix at router level

stephanwlee:
Problem:
TensorBoard has a flag, path_prefix, which serves frontend and the data
endpoints at your specified, if the flag is passed, pathname. While
exact use case slips my mind, it is certainly used by dependents. Now,
the problem is that new Angular based TensorBoard has a custom router
for in-app navigation (we may add new routes for experiment list and
others) it is unaware of the path prefix.

Router works by matching set of predefined routes to current pathname
and finds a match. However, the configuration is unaware (rightfully) of
the path prefix and it results in zero match, which in turn cause
navigation to the default path, "/". This totally breaks the assumption
of the path prefix as all the request from the frontend will be sent to
"/data/..." as opposed to "[path_prefix]/data/..."

Remediation:
Previously, there was no reason to propagate the prefix information to
the frontend. Application only did hash changes and never required a
navigation. With this change, we now:

- server side render meta tag that depicts the relative path to the root of the
  application
- bootstrap Angular app from the relative root path and uses that information
  for matching routes and navigation

Tests:
- manually tested with/without path_prefix
- tested dynamic plugin with/without path_prefix (projector)

### DIFF
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -139,9 +139,9 @@ class CorePlugin(base_plugin.TBPlugin):
         # try to measure gzip and measure content length if a truthy value is
         # passed.
         res = http_util.Respond(request, None, "text/html")
-        res.stream.write(
-            '<meta name="tb-relative-root" content="%s/" />' % (relpath)
-        )
+        # res.stream.write(
+        #     '<meta name="tb-relative-root" content="%s/" />' % (relpath)
+        # )
         res.stream.write(index_asset_bytes)
         return res
 
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -135,15 +135,17 @@ class CorePlugin(base_plugin.TBPlugin):
             if self._path_prefix
             else "."
         )
-        # Technically, it is possible to flush parts using yields but http_utils
-        # try to measure gzip and measure content length if a truthy value is
-        # passed.
-        res = http_util.Respond(request, None, "text/html")
-        # res.stream.write(
-        #     '<meta name="tb-relative-root" content="%s/" />' % (relpath)
-        # )
-        res.stream.write(index_asset_bytes)
-        return res
+        meta_header = (
+            '<!doctype html><meta name="tb-relative-root" content="%s/">'
+            % relpath
+        )
+        content = meta_header.encode("utf-8") + index_asset_bytes
+        # By passing content_encoding, disallow gzipping. Bloats the content
+        # from ~25 kiB to ~120 kiB but reduces CPU usage and avoid 3ms worth of
+        # gzipping.
+        return http_util.Respond(
+            request, content, "text/html", content_encoding="identity"
+        )
 
     @wrappers.Request.application
     def _serve_environment(self, request):
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -133,7 +133,7 @@ class CorePlugin(base_plugin.TBPlugin):
         relpath = (
             posixpath.relpath(self._path_prefix, request.script_root)
             if self._path_prefix
-            else ""
+            else "."
         )
         # Technically, it is possible to flush parts using yields but http_utils
         # try to measure gzip and measure content length if a truthy value is
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -18,6 +18,7 @@
 import functools
 import gzip
 import mimetypes
+import posixpath
 import zipfile
 
 import six
@@ -55,6 +56,7 @@ class CorePlugin(base_plugin.TBPlugin):
         logdir_spec = context.flags.logdir_spec if context.flags else ""
         self._logdir = context.logdir or logdir_spec
         self._window_title = context.window_title
+        self._path_prefix = context.flags.path_prefix if context.flags else None
         self._assets_zip_provider = context.assets_zip_provider
         self._data_provider = context.data_provider
 
@@ -88,7 +90,15 @@ class CorePlugin(base_plugin.TBPlugin):
         with self._assets_zip_provider() as fp:
             with zipfile.ZipFile(fp) as zip_:
                 for path in zip_.namelist():
-                    gzipped_asset_bytes = _gzip(zip_.read(path))
+                    content = zip_.read(path)
+                    # Opt out of gzipping index.html
+                    if path == "index.html":
+                        apps["/" + path] = functools.partial(
+                            self._serve_index, content
+                        )
+                        continue
+
+                    gzipped_asset_bytes = _gzip(content)
                     wsgi_app = functools.partial(
                         self._serve_asset, path, gzipped_asset_bytes
                     )
@@ -111,6 +121,29 @@ class CorePlugin(base_plugin.TBPlugin):
         return http_util.Respond(
             request, gzipped_asset_bytes, mimetype, content_encoding="gzip"
         )
+
+    @wrappers.Request.application
+    def _serve_index(self, index_asset_bytes, request):
+        """Serves index.html content.
+
+        Note that we opt out of gzipping index.html to write preamble before the
+        resource content. This inflates the resource size from 2x kiB to 1xx
+        kiB, but we require an ability to flush preamble with the HTML content.
+        """
+        relpath = (
+            posixpath.relpath(self._path_prefix, request.script_root)
+            if self._path_prefix
+            else ""
+        )
+        # Technically, it is possible to flush parts using yields but http_utils
+        # try to measure gzip and measure content length if a truthy value is
+        # passed.
+        res = http_util.Respond(request, None, "text/html")
+        res.stream.write(
+            '<meta name="tb-relative-root" content="%s/" />' % (relpath)
+        )
+        res.stream.write(index_asset_bytes)
+        return res
 
     @wrappers.Request.application
     def _serve_environment(self, request):
--- a/tensorboard/plugins/core/core_plugin_test.py
+++ b/tensorboard/plugins/core/core_plugin_test.py
@@ -165,7 +165,7 @@ class CorePluginNoDataTest(tf.test.TestCase):
         html = response.get_data()
         self.assertEqual(
             html,
-            b'<meta name="tb-relative-root" content="/" />' + FAKE_INDEX_HTML,
+            b'<meta name="tb-relative-root" content="./" />' + FAKE_INDEX_HTML,
         )
 
     def testDataPaths_disableAllCaching(self):
--- a/tensorboard/plugins/core/core_plugin_test.py
+++ b/tensorboard/plugins/core/core_plugin_test.py
@@ -165,7 +165,8 @@ class CorePluginNoDataTest(tf.test.TestCase):
         html = response.get_data()
         self.assertEqual(
             html,
-            b'<meta name="tb-relative-root" content="./" />' + FAKE_INDEX_HTML,
+            b'<!doctype html><meta name="tb-relative-root" content="./">'
+            + FAKE_INDEX_HTML,
         )
 
     def testDataPaths_disableAllCaching(self):
@@ -398,8 +399,8 @@ class CorePluginPathPrefixTest(tf.test.TestCase):
         html = response.get_data()
 
         expected_meta = (
-            '<meta name="tb-relative-root" content="%s" />'
-            % (expected_tb_relative_root)
+            '<!doctype html><meta name="tb-relative-root" content="%s">'
+            % expected_tb_relative_root
         ).encode()
         self.assertEqual(
             html,
--- a/tensorboard/plugins/core/core_plugin_test.py
+++ b/tensorboard/plugins/core/core_plugin_test.py
@@ -163,7 +163,10 @@ class CorePluginNoDataTest(tf.test.TestCase):
         self.assertEqual(200, response.status_code)
         self.assertStartsWith(response.headers.get("Content-Type"), "text/html")
         html = response.get_data()
-        self.assertEqual(html, FAKE_INDEX_HTML)
+        self.assertEqual(
+            html,
+            b'<meta name="tb-relative-root" content="/" />' + FAKE_INDEX_HTML,
+        )
 
     def testDataPaths_disableAllCaching(self):
         """Test the format of the /data/runs endpoint."""
--- a/tensorboard/plugins/core/core_plugin_test.py
+++ b/tensorboard/plugins/core/core_plugin_test.py
@@ -375,6 +375,77 @@ class CorePluginTestBase(object):
             )
 
 
+class CorePluginPathPrefixTest(tf.test.TestCase):
+    def _send_request(self, path_prefix, pathname):
+        multiplexer = event_multiplexer.EventMultiplexer()
+        logdir = self.get_temp_dir()
+        provider = data_provider.MultiplexerDataProvider(multiplexer, logdir)
+        context = base_plugin.TBContext(
+            assets_zip_provider=get_test_assets_zip_provider(),
+            logdir=logdir,
+            data_provider=provider,
+            window_title="",
+            flags=FakeFlags(path_prefix=path_prefix),
+        )
+        plugin = core_plugin.CorePlugin(context)
+        app = application.TensorBoardWSGI([plugin], path_prefix=path_prefix)
+        server = werkzeug_test.Client(app, wrappers.BaseResponse)
+        return server.get(pathname)
+
+    def _assert_index(self, response, expected_tb_relative_root):
+        self.assertEqual(200, response.status_code)
+        self.assertStartsWith(response.headers.get("Content-Type"), "text/html")
+        html = response.get_data()
+
+        expected_meta = (
+            '<meta name="tb-relative-root" content="%s" />'
+            % (expected_tb_relative_root)
+        ).encode()
+        self.assertEqual(
+            html,
+            expected_meta + FAKE_INDEX_HTML,
+        )
+
+    def testIndex_no_path_prefix(self):
+        self._assert_index(self._send_request("", "/"), "./")
+        self._assert_index(self._send_request("", "/index.html"), "./")
+
+    def testIndex_path_prefix_foo(self):
+        self._assert_index(self._send_request("/foo", "/foo/"), "./")
+        self._assert_index(self._send_request("/foo", "/foo/index.html"), "./")
+
+    def testIndex_path_prefix_foo_exp_route(self):
+        self._assert_index(
+            self._send_request("/foo", "/foo/experiment/123/"), "../../"
+        )
+
+    def testIndex_path_prefix_foo_incorrect_route(self):
+        self.assertEqual(
+            404, (self._send_request("/foo", "/foo/meow/").status_code)
+        )
+        self.assertEqual(404, (self._send_request("/foo", "/").status_code))
+        self.assertEqual(
+            404, (self._send_request("/foo", "/index.html").status_code)
+        )
+
+        # Missing trailing "/" causes redirection.
+        self.assertEqual(301, (self._send_request("/foo", "/foo").status_code))
+        self.assertEqual(
+            301, (self._send_request("/foo", "/foo/experiment/123").status_code)
+        )
+
+    def testIndex_path_prefix_foo_bar(self):
+        self._assert_index(self._send_request("/foo/bar", "/foo/bar/"), "./")
+        self._assert_index(
+            self._send_request("/foo/bar", "/foo/bar/index.html"), "./"
+        )
+
+    def testIndex_path_prefix_foo_bar_exp_route(self):
+        self._assert_index(
+            self._send_request("/foo/bar", "/foo/bar/experiment/123/"), "../../"
+        )
+
+
 def get_test_assets_zip_provider():
     memfile = six.BytesIO()
     with zipfile.ZipFile(
--- a/tensorboard/webapp/app_routing/app_root.ts
+++ b/tensorboard/webapp/app_routing/app_root.ts
@@ -34,7 +34,7 @@ function metaElAppRootExtractor(location: Location): string {
   );
   if (!relativeRoot) return '';
   const {pathname} = new URL(relativeRoot.content, location.getHref());
-  return pathname.replace(/\/+$/, '');
+  return pathname.replace(/\/*$/, '/');
 }
 
 export const AppRootProvider: Provider = {
--- a/tensorboard/webapp/app_routing/app_root.ts
+++ b/tensorboard/webapp/app_routing/app_root.ts
@@ -38,6 +38,9 @@ export class AppRootProvider {
     return pathname.replace(/\/*$/, '/');
   }
 
+  /**
+   * Returns actual full pathname that includes path prefix of the application.
+   */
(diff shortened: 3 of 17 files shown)


wchargin:
A passing thought and question…

One technique that I’ve used in the past is to simply inspect
`window.location.pathname` to get the path prefix on the frontend when
the app starts, before any navigation. This is especially simple for
TensorBoard since we’re a single-page app, but it’s also easy to extend
to multi-page apps, even with client side routing. Then you don’t need
any special backend logic, and in fact this works perfectly with a
static server even when you don’t know the path prefix upfront.

It seems to me like that would be a bit simpler. Is there a reason that
we prefer introducing a new `/data/config` route?

Or, in other words—

> Previously, there was no reason to propagate the prefix information to
> the frontend. Application only did hash changes and never required a
> navigation.

—why does this have to change?

stephanwlee:
> One technique that I’ve used in the past is to simply inspect
window.location.pathname to get the path prefix on the frontend when
the app starts, before any navigation

There is no guarantee that application will only have "/" route and it is legal for users to start the app at "/foo/bar/experiment/123/"

wchargin:
> There is no guarantee that application will only have "/" route and it is legal for users to start the app at "/foo/bar/experiment/123/"

If I launch with `--path_prefix /foo/bar` at this patch and navigate to
`/foo/bar/experiment/123/`, it redirects me to just `/foo/bar/`, and
makes initial requests to `/foo/bar/experiment/123/data/runs` but then
later makes requests to just `/foo/bar/data/runs`. Is that working as
intended? (Like, it doesn’t matter for the OSS data providers, but it
doesn’t seem right to me.)

Anwyay, what I’ve typically done for this is to include in each page’s
static HTML a relative path reference from that page to the root of the
application:

```html
<!-- /index.html -->
<body initial-root=".">
  <h1>Home</h1>
</body>
```

```html
<!-- /about/roadmap/index.html -->
<body initial-root="../..">
  <h1>Our roadmap</h1>
</body>
```

Then, when `index.js` starts, it grabs `body.dataset.initialRoot` and
records ``normalize(`${pathname}/${initialRoot}/`)`` as the path prefix,
which works independent of the pathname, can be hosted statically, and
doesn’t add another RPC to the critical path.

But I recognize that this is probably a bit harder for us since we don’t
really do proper server side rendering, so okay with me to proceed with
your approach.

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -69,13 +73,29 @@ export class AppRoutingEffects {
...
   private readonly onInit$: Observable<Navigation> = this.actions$
     .pipe(ofType(initAction))
     .pipe(
       delay(0),
+      switchMap(() => {
+        return this.store.select(getPathPrefixConfig).pipe(
+          filter((pathPrefix) => pathPrefix !== null),
+          take(1),
+          tap((pathPrefix) => {
+            this.pathPrefix = pathPrefix;
```
psybuzz, line 92: Would things be more maintainable if we didn't store pathPrefix on AppRoutingEffects, but made the Location's getPath, onPopState, replaceState methods aware of it?  e.g.
```ts
this.location.setPathPrefix(pathPrefix);
```

I think Location's method surface is much smaller to check than making sure every usage of `this.location.getPath()` is covered in this file.  In case some other file in app_routing needed to inject Location, it wouldn't need to do all this wrapping as well.

tensorboard/webapp/core/store/core_reducers.ts
```diff
@@ -23,6 +23,12 @@ import {CoreState, initialState} from './core_types';
 
 const reducer = createReducer(
   initialState,
+  on(
+    actions.coreConfigLoaded,
+    (state: CoreState, {config}): CoreState => {
+      return {...state, config};
```
psybuzz, line 29: Given that `Config` represents application level info that is derived from the CLI flags, I think it'd be reasonable to assert that these can never change, or perhaps that they are only ever set once.  e.g. `if (!state.config.equals(config)) throw new Error`

psybuzz (commented):
This PR seems to break Time Series in standalone TB :/

The app 404s on a request to `https://.../experiment/defaultExperimentId/data/plugin/timeseries/tags` because metrics_data_source uses a url starting with '/'.  I wonder, should we patch `TBHttpClient` to add the path_prefix whenever its methods are called with an absolute-looking url?

Re: stamping the initial HTML with path_prefix (or the relative path to it, as per wchargin)
I don't have a strong preference between that approach and the current PR.  If we end up needing to switch to the stamped HTML approach for performance reasons later, most of this PR's frontend code doesn't have to change.

nfelt:
I'm not crazy about baking the `path_prefix` string itself into the actual XHR responses of the app.  Conceptually the functioning of the app should be independent of the path prefix it's "mounted" under - everything about the `path_prefix` should be able to be handled by normal WSGI middleware (as we started doing in https://github.com/tensorflow/tensorboard/pull/2733; this is also a task in https://github.com/tensorflow/tensorboard/issues/2573).

What William suggests above nicely avoids this dependency on the prefix itself:

> Anwyay, what I’ve typically done for this is to include in each page’s
> static HTML a relative path reference from that page to the root of the
> application

We can use this aspect of the suggestion even if we don't do server-side HTML modification, since we can just change `data/environment` to return something that indicates the relative path of that particular endpoint to the root (I think it would suffice to either return the literal subpage path like `experiment/1234/`, an expression representing the path to the root like `../..`, or even just an integer for the number of path segments we are below the root).

If we do that, we actually do preserve the property that `data/environment` is still page-specific, and we avoid introducing an extra XHR in the critical path.

stephanwlee:
> If I launch with --path_prefix /foo/bar at this patch and navigate to
/foo/bar/experiment/123/, it redirects me to just /foo/bar/, and
makes initial requests to /foo/bar/experiment/123/data/runs but then
later makes requests to just /foo/bar/data/runs. Is that working as
intended? (Like, it doesn’t matter for the OSS data providers, but it
doesn’t seem right to me.)

No, that is not what actually happens. If you navigate to `/foo/bar/experiment/123`, it should remain there (no redirection to /foo/bar/` if your app is truly aware of /experiment/123/) and run fetch data from `/foo/bar/experiment/123/data/*`. Are you referring to an actual behavior? If so, it is not intended and I would appreciate if you can kindly point me to the env that you've used to reproduce this.

> Then, when index.js starts, it grabs body.dataset.initialRoot and
records normalize(`${pathname}/${initialRoot}/`) as the path prefix,
which works independent of the pathname, can be hosted statically, and
doesn’t add another RPC to the critical path.

sounds like what I had recommended in the "dehydrate" section in original comment (https://github.com/tensorflow/tensorboard/issues/4421#issuecomment-737704486).
I avoided it for the reason you've mentioned.

> What William suggests above nicely avoids this dependency on the prefix itself:

>> Anwyay, what I’ve typically done for this is to include in each page’s
static HTML a relative path reference from that page to the root of the
application

> We can use this aspect of the suggestion even if we don't do server-side HTML modification, since we can just change data/environment to return something that indicates the relative path of that particular endpoint to the root (I think it would suffice to either return the literal subpage path like experiment/1234/, an expression representing the path to the root like ../.., or even just an integer for the number of path segments we are below the root).

Sorry, I don't understand the approach at all. `/data/environment` returning some information about path_prefix (directly or indirectly) sound about the same and they are on the critical path of application bootstrap no matter what. If you are protesting against the name "path_prefix", I can make some changes but I think it is the best option we've got since (1) we want to surface the app config information anyways and (2) semantically meaningful and straightforward.

Before I address any comments, I want to hear back from bypassers.

wchargin:
> No, that is not what actually happens. If you navigate to `/foo/bar/experiment/123`, it should remain there (no redirection to /foo/bar/`if your app is truly aware of /experiment/123/) and run fetch data from`/foo/bar/experiment/123/data/*`. Are you referring to an actual behavior? If so, it is not intended and I would appreciate if you can kindly point me to the env that you've used to reproduce this.

Yes, this is what I actually observed at d0a9f847c, running:

```
bazel run //tensorboard -- --logdir ~/tensorboard_data/mnist/ --path_prefix /foo/bar --bind_all
```

and navigating to `/foo/bar/experiment/123/` in my browser. I see these
requests:

![Screenshot of network panel showing the requests described in previous
comment][ss]

(I have `tf-nightly-20201204-py3.8`, but that shouldn’t matter.)

[ss]: https://user-images.githubusercontent.com/4317806/101229327-887d5500-3654-11eb-86b9-72a86ebdd585.png

wchargin:
Let me know if you need more info to repro.

tensorboard/http_api.md
```diff
@@ -157,6 +157,18 @@ Example response:
...
+Example response:
+
+    {
+      "path_prefix": "/foo/bar"
+      "window_title": "Custom Name",
+    }
+
 ## `data/environment`
 
 Returns environment in which the TensorBoard app is running.
```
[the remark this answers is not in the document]
stephanwlee: In the broad strokes, we are about aligned.

While `/data/environment` is not the most meaningful name, it is actually being used for experiment metadata.
https://github.com/tensorflow/tensorboard/blob/83ce2a296a76ae733784583d1f3319c8a4d8bfbf/tensorboard/plugins/core/core_plugin.py#L140-L147

My ideal would be to rename `/data/experiment/[:id]/metadata` but "environment", in some sense, about the same thing and I would not advocate for the change unless others feel strongly about it.

I would really not want to roll all experiment metadata under experiment-specific endpoint that is meant to serve lists or runs (besides those endpoints don't exist in a meaningful way? /experiments return an empty array all the time) 

One follow up work that is required is to remove `window_title` from `/environment` and its usage in the page title: https://github.com/tensorflow/tensorboard/blob/83ce2a296a76ae733784583d1f3319c8a4d8bfbf/tensorboard/webapp/core/views/page_title_container.ts#L39-L43

wchargin:
> Sorry, I don't understand the approach at all. `/data/environment` returning some information about path_prefix (directly or indirectly) sound about the same

Suppose that we’reat `/foo/bar/experiment/123` and request the path
prefix info. Say that in case (a), it returns `/foo/bar/` and we
interpret that accordingly. In case (b), it returns `experiment/123/`,
or, equivalently, `../..`, and we interpret that accordingly.

@nfelt and I are saying that case (b) is nicer than case (a) because
TensorBoard behaves the same way independent of the `/foo/bar` prefix.
For instance, you can run this server and simply scrape all the HTML and
XHRs, then save them to static files to “pickle the TensorBoard server”.
In case (b), TensorBoard continues to work if you move that directory
somewhere else relative to the web server. In case (a), it doesn’t.

> and they are on the critical path of application bootstrap no matter what.

Agreed, but `/data/environment` is already on the critical path. If we
add a new `/data/config` XHR, then that’s an extra XHR on the critical
path, right? This is independent of the (a)/(b) discussion above.

stephanwlee:
@wchargin, Thanks, I can repro. It is very surprising to me that we serve `index.html` on the `/foo/bar/experiment/123/` as opposed to `/foo/bar/baz/`. I will look into the plugins_listing and subsequent calls being made under "/foo/bar/experiment/123/data` separately (since it is not related to path_prefix).

> @nfelt and I are saying that case (b) is nicer than case (a) because
TensorBoard behaves the same way independent of the path prefix.
All the XHR responses are the same with any --path_prefix value.
For instance, you can run this server and simply scrape all the HTML and
...

That benefit seems very hypothetical and marginal. Response changing given current route (`/experiment/123/` vs. `/experiment/`) and handling such stateful response seems more painful.

> Agreed, but /data/environment is already on the critical path. If we
add a new /data/config XHR, then that’s an extra XHR on the critical
path, right? This is independent of the (a)/(b) discussion above.

Is `/data/environment` really on the critical path? I think `/data/plugins_listing` and `/data/runs` are but I am not so sure about the environments. I mean, you still see the spinner while the environment is loading but as soon as plugins listing is loaded, the dashboard view is visible on the screen. Also, in case of multi-experiments world, `/data/config` is only relevant
on the initial load and it will not matter on the subsequent dashboard navigations.

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
stephanwlee: I thought about that but was a bit weary of putting business logic into what seems like a clear API, location!

According to our usages of the state, we need to be able to
- read synchronously
- discern was modified 

Those one or two are separate APIs that really do not belong in the Location, imo.

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -44,6 +45,7 @@ import {Navigation, Route} from '../types';
 /** @typehack */ import * as _typeHackNgrxEffects from '@ngrx/effects/effects';
 /** @typehack */ import * as _typeHackModels from '@ngrx/store/src/models';
 /** @typehack */ import * as _typeHackStore from '@ngrx/store';
+import {getPathPrefixConfig} from '../../core/store';
```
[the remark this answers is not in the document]
stephanwlee: Thanks for doing that. I was going to run the sync test after LGTM but you beat me to it!

wchargin:
> Is `/data/environment` really on the critical path? I think `/data/plugins_listing` and `/data/runs` are but I am not so sure about the environments.

Good point. I stand corrected.

LukasMasuch:
As a tensorboard user, I would very much prefer the solution discussed under case b. Allowing the frontend to work under any relative path independent from the configured path-prefix (e.g. by using `window.location.pathname`) makes it a lot easier and flexible to integrate into many different setups. For me, it always was a big strength of the previous versions of tensorboard to not have this limitation. Requiring the frontend to use the configured path-prefix of the backend server might have unplanned consequences for many integration scenarios. 

Other complex tools, such as code-server, are also able to have their frontend run under a relative path. As far as I know, they are also using the `window.location.pathname` to construct the base URL (see [here](https://github.com/cdr/code-server/blob/master/src/common/util.ts#L50)).

psybuzz (changes requested):
Marking as "request changes", to await feedback on a new approach using server side provided path.

stephanwlee (commented):
@nfelt PTAL and see if this aligns with what you were thinking of.

@psybuzz do note that after index.html, our header appeared a bit different :\ I could not discern exactly what has caused the regression so I had to apply flex.

image

tensorboard/webapp/app_routing/app_root_test.ts
```diff
@@ -0,0 +1,58 @@
...
+    {href: 'https://tb.dev/', content: './', expected: ''},
+    {
+      href: 'https://tb.dev/foo/bar/experiment/1/',
+      content: '../../',
+      expected: '/foo/bar',
+    },
+    // wrong relative content but we handle it correctly.
+    {href: 'https://tb.dev/', content: '../../', expected: ''},
+  ].forEach(({content, href, expected}) => {
+    it(`using <meta> content, returns an absolute path: ${href} and ${content}`, () => {
```
nfelt, line 53: Maybe also worth testing the case of `https://tb.dev/index.html` to make sure we end up with the right result?

tensorboard/plugins/core/core_plugin_test.py
```diff
@@ -172,7 +172,10 @@ def testIndex_returnsActualHtml(self):
         self.assertEqual(200, response.status_code)
         self.assertStartsWith(response.headers.get("Content-Type"), "text/html")
         html = response.get_data()
-        self.assertEqual(html, FAKE_INDEX_HTML)
+        self.assertEqual(
+            html,
+            b'<meta name="tb-relative-root" content="/" />' + FAKE_INDEX_HTML,
```
nfelt, line 177: Can we add a couple more tests for the path prefix behavior?  We could add a new TestCase that doesn't construct the WSGI app in `setUp`, and instead have each test method call `TensorBoardWSGI` with a different value of the `path_prefix` parameter - maybe e.g. empty, one segment, and two segments.

It would also be good to try at least a couple different values for the path we pass to `get()` - e.g. "/", "/index.html", "/experiment/123/", "/experiment/123" (no trailing slash), etc.

nfelt (commented):
Yeah this looks good, thanks.  (I think we'll eventually want to avoid CorePlugin directly using path_prefix but I'm happy to clean that up later.)

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -88,7 +104,11 @@ export class AppRoutingEffects {
...
       map((navigation) => {
-        return {...navigation, browserInitiated: true};
+        return {
+          pathname: this.getAppRootlessPathname(navigation.pathname),
+          replaceState: navigation.replaceState,
+          browserInitiated: true,
+        };
       })
     )
   ).pipe(
```
psybuzz, line 114: I think things work when the pathname emitted by the 3 source observables on L103 start with a "/", but if navigationRequested is able to emit relative paths like "../experiment/1", my suspicion is that things may break:

- The app's routes are configured without appRoot ("/experiment", "/compare", etc)
- L103, the 3 source observables emit appRootLess pathnames ("/experiment/1") and relative pathnames ("../experiment/1")
- L115 maps those into non-relative pathnames that include the appRoot ("/appRoot/experiment/1")
- L121 uses the routeConfigs to match these strings (that include appRoot) against the app's routes, which is guaranteed to fail to match

Can we add a test case for dispatching something like `navigationRequested({pathname: '../experiments'})` under the presence of path prefix?

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -88,7 +104,11 @@ export class AppRoutingEffects {
     this.onInit$,
```
psybuzz, line 104: Might be good to add a comment, either one saying that the 3 source observables in the start of this chain must be appRootLess, or a comment at the top of `userInitNavRoute$`, saying that it must emit data that is appRootLess.

psybuzz (commented):
Side note: I think we may need `router_link_directive_container.ts` to become aware of AppRoots as well.  Currently, when an `<a>` has `[routerLink]=['foo']`, we guarantee that the "href" attribute is a full, correct URL so things like ctrl+click properly open "http://tb/foo" in a new tab.
With path_prefix, I would imagine those hrefs now point to incorrect appRoot-less URLs.

nfelt (commented):
Python changes and tests LGTM, thanks. Will defer to Erik's remaining comments before approving though.

tensorboard/webapp/app_routing/app_root.ts
```diff
@@ -0,0 +1,44 @@
...
+);
+
+function metaElAppRootExtractor(location: Location): string {
+  const metaElements = [
+    ...document.querySelectorAll('head meta'),
+  ] as HTMLMetaElement[];
+
+  const relativeRoot = metaElements.find(
+    (el) => el.name === 'tb-relative-root'
+  );
```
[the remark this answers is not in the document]
stephanwlee: Great idea. Done.

tensorboard/webapp/app_routing/app_root.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (!relativeRoot) return '';
```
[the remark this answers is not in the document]
stephanwlee: Made appRoot non-nullable and returned `/` as per the contract from backend.

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
stephanwlee: Great catch. I definitely forgot about the programmatical navigation (which is not currently being used in OSS) and did not properly account for abs/rel path in it. Handled it and added test.

tensorboard/plugins/core/core_plugin.py
```diff
@@ -115,6 +125,29 @@ def _serve_asset(self, path, gzipped_asset_bytes, request):
...
+        )
+        # Technically, it is possible to flush parts using yields but http_utils
+        # try to measure gzip and measure content length if a truthy value is
+        # passed.
+        res = http_util.Respond(request, None, "text/html")
+        res.stream.write(
+            '<meta name="tb-relative-root" content="%s/" />' % (relpath)
+        )
+        res.stream.write(index_asset_bytes)
+        return res
```
wchargin, line 149: I don’t quite understand why this change implies the accompanying change
to `http_util.Respond`. The alternative that comes to mind is:

```python
meta_header = '<meta name="tb-relative-root" content="%s/" />' % (relpath)
content = meta_header.encode("utf-8") + index_asset_bytes
return http_util.Respond(content)
```

Judging from the comment, it sounds like you want to avoid that because
it automatically gzips, since the content type is `text/html`. That
makes sense. But then can’t we just pass `content_encoding="identity"`
rather than adding new mechanisms? I think that that already does the
right thing:

```python
>>> from tensorboard.backend import http_util
>>> from werkzeug import test as wtest
>>> from werkzeug import wrappers
>>> def req():
...     headers = [("Accept-Encoding", "gzip")]
...     environ = wtest.EnvironBuilder(headers=headers).get_environ()
...     return wrappers.Request(environ)
... 
>>> res = http_util.Respond(req(), b"<html>\n", "text/html")
>>> res.headers.get("Content-Encoding"), res.get_data()
('gzip', b'\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\xff\xb3\xc9(\xc9\xcd\xb1\xe3\x02\x00\x994\xcb3\x07\x00\x00\x00')
>>> res = http_util.Respond(req(), b"<html>\n", "text/html", content_encoding="identity")
>>> res.headers.get("Content-Encoding"), res.get_data()
```

Failing that, I might prefer to add a new `no_gzip` flag or something
rather than introducing (a) the edge case in `http_util` and (b) the
stream-writing code in this method.

stephanwlee:
Tested the sync (cl/347955446) and tested against the internal version of TB.

psybuzz (approved):
The frontend code lgtm, thanks for checking internally.

I also did a quick test sync and saw 1 projector test fail, but I'm not 100% sure its related: cl/348085083

Approving % wchargin@'s approval on the content_encoding=identity topic.

wchargin (approved):
Python revisions look good to me. Ideally I would like `html.escape`
around the HTML interpolation; even though there’s probably an argument
that it’s safe in this case due to provenance of `relpath`, it’s nice
for it to be “clearly correct” rather than “merely correct”.

tensorboard/webapp/app_routing/views/router_link_directive_container.ts
```diff
@@ -47,9 +55,12 @@ export class RouterLinkDirectiveContainer {
   get href() {
     if (!this.pathname) return null;
 
-    return this.location.getFullPathFromRouteOrNav({
-      pathname: this.pathname,
-    });
+    return (
+      this.appRoot.slice(0, -1) +
```
nfelt, line 59: Would it maybe make sense to expose `getAbsPathnameWithAppRoot()` somewhere to kind of share this logic?  As it is, it's not terribly obvious here why we slice the last char off appRoot unless you can see the bigger picture.  Maybe at least a comment would help.

nfelt (approved):
Took a quick look and the updates and generally LGTM.  +1 to William's point that it would be a bit better to escape the inserted relative path, just in case.

tensorboard/webapp/app_routing/views/router_link_directive_container.ts
stephanwlee: Done. Moved it to AppRootProvider instead.

stephanwlee:
@psybuzz made significant changes to the FE code to adopt the feedback. Please TAL one last time.

psybuzz (approved):

PR #4425, opened 2020-12-05
metrics: gracefully degrade without tensors/blobs

wchargin:
Summary:
Not all data providers support tensors and blob sequences. The methods
are optional in the `DataProvider` interface, and (e.g.) the gRPC data
provider does not yet implement them, nor does the RustBoard server.
This patch teaches the time series dashboard to gracefully omit image
and histogram data if the needed data classes aren’t available.

Test Plan:
Point TensorBoard at an MNIST dataset with all three kinds of data, and
note that scalars appear if `--load_fast` is passed, and all data
appears otherwise. Prior to this patch, the requests would 500 under
`--load_fast`, and so nothing would be displayed at all.

wchargin-branch: metrics-degrade-tensors-blobs

### DIFF
--- a/tensorboard/plugins/metrics/metrics_plugin.py
+++ b/tensorboard/plugins/metrics/metrics_plugin.py
@@ -321,6 +321,11 @@ class MetricsPlugin(base_plugin.TBPlugin):
             experiment_id=experiment,
             plugin_name=image_metadata.PLUGIN_NAME,
         )
+        # Not all data providers support tensors and/or blob sequences.
+        if histogram_mapping is None:
+            histogram_mapping = {}
+        if image_mapping is None:
+            image_mapping = {}
 
         result = {}
         result["scalars"] = _format_basic_mapping(scalar_mapping)
--- a/tensorboard/data/provider.py
+++ b/tensorboard/data/provider.py
@@ -105,6 +105,8 @@ class DataProvider(object):
 
     Unless otherwise noted, any methods on this class may raise errors
     defined in `tensorboard.errors`, like `tensorboard.errors.NotFoundError`.
+
+    If not implemented, optional methods may return `None`.
     """
 
     def data_location(self, ctx=None, *, experiment_id):


tensorboard/plugins/metrics/metrics_plugin.py
```diff
@@ -321,6 +321,11 @@ def _tags_impl(self, ctx, experiment=None):
             experiment_id=experiment,
             plugin_name=image_metadata.PLUGIN_NAME,
         )
+        # Not all data providers support tensors and/or blob sequences.
+        if histogram_mapping is None:
+            histogram_mapping = {}
+        if image_mapping is None:
+            image_mapping = {}
```
psybuzz, line 328: If unimplemented DataProvider methods may return None, can we add the corresponding `if scalar_mapping is None` check as well, so that metrics_plugin.py doesn't rely on knowing which specific dataclasses have implemented `list_*`?

On the other hand, should `GrpcDataProvider` be responsible for implementing `list_tensors` that returns `{}`?  [1] suggests that instances of classes with `@abc.abstractmethod` must have "all of its abstract methods and properties are overridden".  Would it make sense to add GrpcDataProvider.list_tensors/blobs, or update the `data/provider.py` docstrings to indicate that they may return `None`?

[1] https://docs.python.org/3/library/abc.html#abc.abstractmethod

psybuzz (approved):

tensorboard/plugins/metrics/metrics_plugin.py
wchargin: The scalars methods are `abc.abstractmethods` whereas the methods for
other data classes are not. You cannot instantiate a `DataProvider`
without implementing `read_scalars`. It’ll hard-error.

That these methods pass when not implemented is (in my experience)
fairly standard, but I’m happy to document it explicitly.

psybuzz: > The scalars methods are abc.abstractmethods whereas the methods for
other data classes are not

Ahh, thanks for pointing this out.  I think my comment earlier was invalid, so feel free to keep or remove the data/provider.py change as you see fit!

PR #4427, opened 2020-12-05
line chart: calc extent without illegal values

stephanwlee:
Illegal value is defined by a scale where, in some situation, we can
support all finite values while, in other cases, we can only support
positive finite numbers.

This change takes predicate on the extent calculation code and properly
calculate extent of log scale.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/scale_types.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/scale_types.ts
@@ -62,4 +62,12 @@ export interface Scale {
    *    may return less or more ticks.
    */
   ticks(domain: [number, number], sizeGuidance: number): number[];
+
+  /**
+   * Returns true when a value, x, is a value, can be safely transformed symmetrically*,
+   * and is not an extremum.
+   *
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/scale_types.ts
```diff
@@ -62,4 +62,12 @@ export interface Scale {
...
    */
   ticks(domain: [number, number], sizeGuidance: number): number[];
+
+  /**
+   * Returns true when a value, x, is a value, can be safely transformed symmetrically*,
+   * and is not an extremum.
+   *
+   * *: e.g., f(1) = 2 and f'(2) = 1 vs. f(1) = NaN and f'(NaN) = ?.
+   */
+  isSafeNumber(x: number): boolean;
```
wchargin, line 72: (drive-by) thanks; these docs are helpful! I read the first file in this
PR and wondered what “safe” meant (since it can mean many things in the
context of floats…), and this answered my question exactly.

psybuzz (approved):

PR #4428, opened 2020-12-05
sync: add @ngrx/store dep on app since app_container uses it

stephanwlee:


wchargin (approved):
Confirmed that this builds in a test sync; thanks!

PR #4429, opened 2020-12-05
rust: support loading data server from Pip package

wchargin:
Summary:
With `--load_fast`, TensorBoard can now find the data server binary by
asking the new `tensorboard_data_server` Pip package. That Pip package
bundles a platform-specific binary and provides a `server_binary`
function to locate it.

The Pip package can be easily built with Bazel, but it can also be built
with plain Python if you bring your own RustBoard binary: say, after
building with Cargo. This is so that we can easily build Pip packages
for Mac and Windows without having to get Bazel working on those
platforms.

Test Plan:
For the Pip package, an end-to-end test is included, but it’s tagged as
a manual test because it depends on the enclosing Python environment, so
run it explicitly:

```
bazel test //tensorboard/data/server/pip_package:build_test
```

For an end-to-end test, follow the new instructions in `DEVELOPMENT.md`.
If you want to be extra sure that it’s really running the new binary,
you can add `std::process::exit(77)` to the top of `main.rs` and see
that TensorBoard fails accordingly.

wchargin-branch: rust-tds-pip-package


tensorboard/data/server/pip_package/setup.py
```diff
@@ -0,0 +1,56 @@
...
+import tensorboard_data_server
+
+
+setuptools.setup(
+    name="tensorboard_data_server",
+    version=tensorboard_data_server.__version__.replace("-", ""),
+    description="Fast data loading for TensorBoard",
+    long_description="Fast data loading for TensorBoard",
+    url="https://github.com/tensorflow/tensorboard/tree/master/tensorboard/data/server",
+    author="Google Inc.",
```
stephanwlee, line 27: nit: Google LLC is more correct but I bet we have made this mistake elsewhere too.

tensorboard/data/server/pip_package/BUILD
```diff
@@ -0,0 +1,39 @@
+package(default_visibility = ["//tensorboard:internal"])
```
stephanwlee, line 1: Not a big harm but does this need to be visible to other packages given that it has no libraries?

tensorboard/data/server/DEVELOPMENT.md
```diff
@@ -75,6 +75,24 @@ As an alternative to `--define=link_data_server=true`, you can set the
...
+and then run TensorBoard with `--load_fast`. You’ll need to re-generate and
+re-install the Pip package each time that you update the Rust binary.
+
+With `--load_fast`, the order of precedence for the different binary locations
+is:
+
+-   an explicit `TENSORBOARD_DATA_SERVER_BINARY` environment variable is honored
+    if present and non-empty; else
+-   the `tensorboard_data_server` package is queried if it is installed; else
+-   the server bundled with `--define=link_data_server=true` is used.
```
stephanwlee, line 94: In your opinion, is this the best way to develop the data server? If so, I think this line should be a bit expanded to denote that and maybe have a line like `ibazel run --define= link_data_server=true tensorboard -- --logdir [logdir]`.

stephanwlee (commented):
Code looks good. Just want to hear replies.

wchargin (commented):
Thanks for the early review, which is also a late review!

tensorboard/data/server/DEVELOPMENT.md
stephanwlee: Thanks. I did not carefully read the code outside of this changelist so I have missed that. I see that you have added/clarified few texts here and there and they LGTM.

stephanwlee (approved):

tensorboard/data/server/DEVELOPMENT.md
wchargin: Great; thanks for checking!

wchargin:
Double-checked test sync. Looks fine.

PR #4430, opened 2020-12-05
ng: dispatch coreLoaded after router is initialized

stephanwlee:
Only after the router is initialized, we can make sure we are at the
correct routes only at which point we should start to initialize and
make requests. Otherwise, we can make requests at the wrong paths which
may/may-not be correct.

With this change, we fire the coreLoaded after router is initialized
which is confirmed by checking for non-nullness of the activeRoute.

### DIFF
--- a/tensorboard/webapp/BUILD
+++ b/tensorboard/webapp/BUILD
@@ -115,6 +115,7 @@ tf_ng_module(
         "@npm//@angular/core",
         "@npm//@angular/platform-browser",
         "@npm//@ngrx/store",
+        "@npm//rxjs",
     ],
 )
 
--- a/tensorboard/webapp/BUILD
+++ b/tensorboard/webapp/BUILD
@@ -86,8 +86,10 @@ tf_ng_module(
         "app_container.ng.html",
     ],
     deps = [
+        ":app_state",
         ":mat_icon",
         ":oss_plugins_module",
+        ":selectors",
         ":store_module",
         "//tensorboard/webapp/alert",
         "//tensorboard/webapp/alert/views:alert_snackbar",
@@ -113,6 +115,25 @@ tf_ng_module(
         "@npm//@angular/core",
         "@npm//@angular/platform-browser",
         "@npm//@ngrx/store",
+    ],
+)
+
+tf_ng_module(
+    name = "app_test",
+    testonly = True,
+    srcs = [
+        "app_test.ts",
+    ],
+    deps = [
+        ":app",
+        ":app_state",
+        ":selectors",
+        "//tensorboard/webapp/app_routing:testing",
+        "//tensorboard/webapp/core/actions",
+        "@npm//@angular/common",
+        "@npm//@angular/core",
+        "@npm//@ngrx/store",
+        "@npm//@types/jasmine",
     ],
 )
 
@@ -239,6 +260,7 @@ tf_resource_digest_suffixer(
 tf_ng_web_test_suite(
     name = "karma_test",
     deps = [
+        ":app_test",
         "//tensorboard/webapp/alert:test_lib",
         "//tensorboard/webapp/alert/store:test_lib",
         "//tensorboard/webapp/alert/views:views_test",
--- a/tensorboard/webapp/app_container.ts
+++ b/tensorboard/webapp/app_container.ts
@@ -12,26 +12,38 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-import {Component, OnInit, ViewContainerRef} from '@angular/core';
+import {Component, ViewContainerRef} from '@angular/core';
 import {Store} from '@ngrx/store';
+import {filter, take} from 'rxjs/operators';
+
+import {getActiveRoute} from './selectors';
 import {coreLoaded} from './core/actions';
-import {State} from './core/store';
+import {State} from './app_state';
 
 @Component({
   selector: 'tb-webapp',
   templateUrl: './app_container.ng.html',
   styleUrls: ['./app_container.css'],
 })
-export class AppContainer implements OnInit {
+export class AppContainer {
   // vcRef is required by ngx-color-picker in order for it to place the popup
   // in the root node in a modal mode.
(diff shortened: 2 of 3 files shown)


stephanwlee:
After this change, when using path_prefix and navigating to "/foo/bar/experiment/123" should not make any requests to "/foo/bar/experiment/123/data/*" _after_ making one to "/foo/bar/experiment/123/data/config". We still need to make one for config to figure out the real route of the app, not a potential route like `/experiment/123/` (in reality, either the ng router should understand the notion of `/experiment/:id` or backend should not serve index.html under that route).

wchargin (commented):
> After this change, when using path_prefix and navigating to "/foo/bar/experiment/123" should not make any requests to "/foo/bar/experiment/123/data/*" _after_ making one to "/foo/bar/experiment/123/data/config".

I’m confused. Why don’t we want to make such XHRs? How else will we
fetch data for the correct experiment?

For instance, consider a TensorBoard.dev experiment:
<https://tensorboard.dev/experiment/EDZb7XgKSBKo6Gznh3i8hg/>

This makes requests to:

```
https://tensorboard.dev/experiment/EDZb7XgKSBKo6Gznh3i8hg/data/plugins_listing
https://tensorboard.dev/experiment/EDZb7XgKSBKo6Gznh3i8hg/data/runs
https://tensorboard.dev/experiment/EDZb7XgKSBKo6Gznh3i8hg/data/plugin/scalars/tags
```

etc. And likewise for an internal “corporate TensorBoard”.

Yet in open source TensorBoard, navigating to `/experiment/123/` causes
the app to redirect to `/` and make subsequent data XHRs without an
experiment ID, both at master and even more so at this commit.

So I don’t understand: (a) why isn’t this behavior breaking our other
TensorBoard instances? aren’t they also using this Angular shell? and
(b) why are we moving farther away from what seems to me to be the
correct behavior?

stephanwlee:
> (a) why isn’t this behavior breaking our other
TensorBoard instances? aren’t they also using this Angular shell? and

Sadly, no. We have 5 instances of the app component because Angular does not allow easy composition without creating an entire NgComponent so we have 5 different entry points right now. This would not impact it.

> (b) why are we moving farther away from what seems to me to be the
correct behavior?

Maybe I need to convince you a little bit here that this is the right behavior. The logic is _**not**_ to navigate to "/" whenever application bootstraps; instead, it navigates to a default route (in TB.oss, it currently is "/") if current pathname does not match any of the known route. So...

*In TB.dev*
router is not installed there and app_container does something completely different but it eventually should (1) use the router and (2) have routes for experiment list vs. experiment view.

*In a TB variant that I won't name (crap, you already named it)*
It is the most interesting to me; even if you navigate to an unknown route like `/foo/bar/baz/qaz`, the backend serves index.html (actually I need to fix something... hmm) and let the frontend deal with the illegal route. In this example though, with the code at master, it is possible to make the request at `/foo/bar/baz/qaz/data/plugins_listing` which is disallowed. Instead, we need to first match the route to a list of known routes, navigate to the default route, then bootstrap the application and this PR is exactly doing that. In reality, in this variant of TB, we force the requests to be made with the absolute url (which I can describe in a separate channel) so we never make the request at `/foo/bar/baz/qaz/data/plugins_listiing`.

Now, addressing my admittedly confusing phrases. 

>> After this change, when using path_prefix and navigating to "/foo/bar/experiment/123" should not make any requests to "/foo/bar/experiment/123/data/*" after making one to "/foo/bar/experiment/123/data/config".
> I’m confused. Why don’t we want to make such XHRs? How else will we
fetch data for the correct experiment?

I meant we need to know the prefix_path in order for us to correctly discern whether `/foo/bar/experiment/123/data` is legal or not (e.g., path_prefix is `/foo/bar/experiment/123/`. To do so, we _must_ make the request to `/data/config` but by the time we know the path_prefix, we should navigate away from `/foo/bar/experiment/123` and must not make the request to the incorrect path name. Hope this clarification helps.

wchargin:
> Maybe I need to convince you a little bit here that this is the right behavior. The logic is _**not**_ to navigate to "/" whenever application bootstraps; instead, it navigates to a default route (in TB.oss, it currently is "/") if current pathname does not match any of the known route.

This makes sense to me. But I would think that `/experiment/123/` should
be a valid route, and should serve `index.html`. Is it not? This is how
I usually tell open-source TensorBoard to load experiment 123. It’s true
that most of TensorBoard doesn’t actually care much about the ID, but
the ID _is_ passed through to the backend, plugins, and data providers.

At the very least, it’s useful for testing features that do care about
the experiment ID. From my perspective, “TensorBoard” (this codebase)
*does* offer experiment functionality, as used by TBdev and others, so
it makes sense that we have a way to exercise it.

> _In a TB variant that I won't name (crap, you already named it)_

What? I named nothing. :-)

> Instead, we need to first match the route to a list of known routes, navigate to the default route, then bootstrap the application and this PR is exactly doing that.

Okay, this by itself makes sense to me. So I think we agree that the
app should match to a valid route before sending more XHRs, and the
question of whether `/experiment/123/` is a valid route in OSS TB is a
separate question. Is that right?

wchargin (approved):

stephanwlee:
>  the question of whether /experiment/123/ is a valid route in OSS TB is a
separate question. Is that right?

Yeah, did not know that this was the point of confusion before. Thanks for making it very explicit. 
I can see it both ways. But my ideal would be one of two things:
1. remove the experiment aware middleware and let the downstream apps compose it (and remove weird support for `/experiment/123/index.html`, or
2. add support for "/experiment/:id/`in the frontend (we can have a route that just redirects).

wchargin:
> my ideal would be one of two things:
> 
> 1. remove the experiment aware middleware and let the downstream apps compose it (and remove weird support for `/experiment/123/index.html`, or
> 2. add support for "/experiment/:id/`in the frontend (we can have a route that just redirects).

Both of these sound like they’re treating `/experiment/123/` as
something weird or to be avoided. My ideal would be:

3.  Treat `/experiment/123/` as a normal experiment root. Don’t redirect
    away from it. Make XHRs to `/experiment/123/data/runs`, et al.
    Behave just as (e.g.) TBdev would. The root route `/` acts just like
    a special case of this when the experiment ID happens to be empty.

PR #4431, opened 2020-12-05
[*.py] Rename "Arguments:" to "Args:"

SamuelMarks:
I've written custom parsers and emitters for everything from docstrings to classes and functions. However, I recently came across an issue with the TensorFlow codebase: inconsistent use of `Args:` and `Arguments:` in its docstrings. It is easy enough to extend my parsers to support both variants, however it looks like `Arguments:` is wrong anyway, as per:

  - https://google.github.io/styleguide/pyguide.html#doc-function-args @ [`ddccc0f`](https://github.com/google/styleguide/blob/ddccc0f/pyguide.md)

  - https://chromium.googlesource.com/chromiumos/docs/+/master/styleguide/python.md#describing-arguments-in-docstrings @ [`9fc0fc0`](https://chromium.googlesource.com/chromiumos/docs/+/9fc0fc0/styleguide/python.md)

  - https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html @ [`c0ae8e3`](https://github.com/sphinx-contrib/napoleon/blob/c0ae8e3/docs/source/example_google.rst)

Therefore, only `Args:` is valid. This PR replaces them throughout the codebase.

PS: For related PRs, see tensorflow/tensorflow/pull/45420

### DIFF
--- a/tensorboard/backend/application.py
+++ b/tensorboard/backend/application.py
@@ -598,7 +598,7 @@ def _auth_context_middleware(wsgi_app, auth_providers):
 def _clean_path(path):
     """Removes a trailing slash from a non-root path.
 
-    Arguments:
+    Args:
       path: The path of a request.
 
     Returns:
--- a/tensorboard/backend/event_processing/event_accumulator_test.py
+++ b/tensorboard/backend/event_processing/event_accumulator_test.py
@@ -977,7 +977,7 @@ class RealisticEventAccumulatorTest(EventAccumulatorTest):
     def _writeMetadata(self, logdir, summary_metadata, nonce=""):
         """Write to disk a summary with the given metadata.
 
-        Arguments:
+        Args:
           logdir: a string
           summary_metadata: a `SummaryMetadata` protobuf object
           nonce: optional; will be added to the end of the event file name
--- a/tensorboard/backend/event_processing/plugin_event_accumulator_test.py
+++ b/tensorboard/backend/event_processing/plugin_event_accumulator_test.py
@@ -792,7 +792,7 @@ class RealisticEventAccumulatorTest(EventAccumulatorTest):
     def _writeMetadata(self, logdir, summary_metadata, nonce=""):
         """Write to disk a summary with the given metadata.
 
-        Arguments:
+        Args:
           logdir: a string
           summary_metadata: a `SummaryMetadata` protobuf object
           nonce: optional; will be added to the end of the event file name
--- a/tensorboard/data_compat.py
+++ b/tensorboard/data_compat.py
@@ -55,7 +55,7 @@ def migrate_value(value):
     disk; this method converts them to new-style values so that further
     code need only deal with one data format.
 
-    Arguments:
+    Args:
       value: A `Summary.Value` object. This argument is not modified.
 
     Returns:
--- a/tensorboard/encode_png_benchmark.py
+++ b/tensorboard/encode_png_benchmark.py
@@ -93,7 +93,7 @@ def _image_of_size(image_size):
 def _format_line(headers, fields):
     """Format a line of a table.
 
-    Arguments:
+    Args:
       headers: A list of strings that are used as the table headers.
       fields: A list of the same length as `headers` where `fields[i]` is
         the entry for `headers[i]` in this row. Elements can be of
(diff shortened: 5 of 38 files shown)


wchargin (commented):
Hmm. You’re attempting to establish a global invariant about docstring
format so that your code doesn’t break. But you’ve provided no way to
enforce this invariant. The fact that you’ve needed to send PRs for
thousands of such changes across dozens of repositories makes it clear
that absent such an enforcement mechanism, regressions will quickly
be introduced. Yet I don’t think that we really *want* such an
enforcement mechanism. More tooling, more flakiness, more rules, for
little practical benefit. (And we don’t use Pylint, so there’s no “just
enable `extensions.docparams` and everything will be fine”.)

Internally, Google also has no enforcement mechanism for this. Some of
our core libraries have `Arguments:` docstrings—a minority, certainly,
but we’ve found no need to be stricter, and our tooling works just fine.
[TensorFlow doc generation also works with `Arguments:`.][tf]

It sounds like it would be much easier for everyone if you could just
make your own parsers more permissive. (Easier for us to not deal with
it, and easier for you so that you don’t have to keep dealing with these
regressions.) As you say, it should be easy enough. Could you give that
a try?

One of the things that I have learned from TensorBoard is that when
writing parsing/loading/reading logic, it is valuable to treat the data
as it is, not how you think it should be. :-)

[tf]: https://www.tensorflow.org/api_docs/python/tf/compat/dimension_value#arguments

wchargin:
<https://github.com/tensorflow/tensorboard/pull/4431#pullrequestreview-545648269>

SamuelMarks:
@wchargin Yeah for sure I can extend my parser to support this. It won't be difficult even; maybe a line or two of code change.

In that case, and considering you're at Google, maybe you can ask them to change the official style guide for Google docstrings? - That way others won't expect the same strictness.

It might still be beneficial to enable some linting, so that people aren't using reserved keywords when they shouldn't, e.g., I still run into the odd issue with parsing out defaults, e.g., compare these two lines [callbacks.py#L953](https://github.com/tensorflow/tensorflow/blob/66d5177/tensorflow/python/keras/callbacks.py#L953) [callbacks.py#L1219](https://github.com/tensorflow/tensorflow/blob/66d5177/tensorflow/python/keras/callbacks.py#L1219). Without my `try`/`except SyntaxError` hack from last night it would set the `"default"` to `'epoch'` (correctly) and to ```the `Model`'s metrics``` (incorrectly). Linting would pickup this issue, as `defaults to` should be considered a keyword.

_Good thing I've written my own parser for ReST, Google, and numpy formats; otherwise the likely way around this would be sending these PRs._

wchargin:
> @wchargin Yeah for sure I can extend my parser to support this. It won't be difficult even; maybe a line or two of code change.

Great!

> In that case, and considering you're at Google, maybe you can ask them to change the official style guide for Google docstrings? - That way others won't expect the same strictness.

There’s nothing to change. You’re right that `Args:` is preferred and by
far the most common. This doesn’t imply that it’s used universally.

> Linting would pickup this issue, as `defaults to` should be considered a keyword.

No, I don’t think so. Docstrings are prose written for humans. This
docstring says, “If not provided, defaults to the `Model`'s metrics.”
This is perfectly acceptable and standard, and not an error. There is no
issue to pick up, because “defaults to” is not a keyword.

If you want programmatic access to default values, you can evaluate the
symbols and check `.__defaults__`:

```python
>>> import tensorflow as tf
>>> tf.keras.callbacks.ProgbarLogger.__init__.__defaults__
('samples', None)
```

SamuelMarks:
@wchargin Yeah I know, in my library I use `inspect` to get dynamic stuff and `ast` to parse the body itself, either from `inspect.getsource` or from an actual file. (technically I only use functions, so just pass the right string to the right place)

But the question is about the outer docstring; which may or may not be merged with the `__init__` signature and docstring. Something the `defaults to` text includes information not present in the function signature, e.g., the function signature argument defaults to `None` but the `default to` text actually tells you what it falls back to if None. So parsing out the `defaults to` is useful, and considering it to be just regular prose remains an issue.

In terms of `Args:` vs `Arguments:`, if your argument is that we shouldn't be so strict with names, and considering TensorFlow is a Google project (and you're a Googler); then maybe including a note in the style guide to defend your position would be warranted?

PR #4432, opened 2020-12-07
ci: add nightly scheduled GitHub Actions job

wchargin:
Summary:
This extends our current GitHub Actions CI workflow to run nightly at
05:00 PST (06:00 PDT, 13:00 UTC). The [docs for scheduled events][1]
indicate that such a job will run on the default (master) branch. This
job should run with write access to the build cache.

[1]: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#scheduled-events

Test Plan:
Note that the workflow parses and normal CI still runs on this job.
We’ll wait for tomorrow to see if it actually runs.

wchargin-branch: ci-gh-nightly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,12 @@ on:
       - '[0-9]+.*'
       - 'ci-*'
   pull_request: {}
+  schedule:
+    # 13:00 UTC is 05:00 in Pacific standard time (UTC-8), which is well
+    # after nightly TensorFlow wheels are released (around 2--3 AM) and
+    # just after nightly TensorBoard wheels are released (around 04:15).
+    # (cron syntax: minute hour day-of-month month day-of-week)
+    - cron: '0 13 * * *'
 
 env:
   # Keep this Bazel version in sync with the `versions.check` directive


psybuzz (commented):
Code looks good, but I'm curious: since the team checks that the `tb-nightly` pip package is pushed successfully to pypi every day, is this daily GH actions job adding necessary benefit?

wchargin:
Yeah; the point is to make sure that our code still works against the
latest version of `tf-nightly`. Nightly-breaking changes used to be
quite common, and so we want to know about them long before the next
release.

This won’t alert us if a `tb-nightly` push fails, since it’ll just grab
yesterday’s version. The issues are unrelated.

More concretely, this brings us to parity with Travis, where we do have
a daily cron job. (Not expressed in configuration anywhere, just in the
Travis settings.)

psybuzz (approved):
Thanks for the explanation!  Wasn't aware we had a Travis cron job setup before.

wchargin:
Yeah. The Travis cron was kind of awful because you can only set it up
by saying “trigger scheduled job”, and it runs at that time and every 24
hours thereafter. So if you want it to run at 05:00, you have to stay up
until then and push the button. A simple cron spec is much nicer.

wchargin:
[Looks like this worked fine:][run]

![Screenshot of the successful scheduled workflow run, which ran “6
hours ago” at time of writing][ss]

[run]: https://github.com/tensorflow/tensorboard/actions/runs/408289546
[ss]: https://user-images.githubusercontent.com/4317806/101529279-39317000-3945-11eb-89b2-3918d1591fc9.png

The logs show that this did indeed use a writable cache, as desired.

PR #4433, opened 2020-12-07
rust: replace static fixture with test helpers

wchargin:
Summary:
In #4385, @nfelt suggested using test helpers to easily populate a
commit rather than using a single fixture across all test cases. This
patch implements such test helpers and uses them in the server tests.
The net diff to the test code (not the helpers) is −32 lines.

Test Plan:
Existing tests pass. The new test helpers mostly don’t have explicit
tests, but that’s okay since they’re exercised by the main test code.

wchargin-branch: rust-test-data-builders


tensorboard/data/server/commit.rs
```diff
@@ -136,3 +136,166 @@ mod tests {
...
+        ///
+        /// By default, this maps step `i` to `0.5 ** (i + 1)`.
+        eval: Box<dyn Fn(Step) -> f32>,
+    }
+
+    impl Default for ScalarTimeSeriesBuilder {
+        fn default() -> Self {
+            ScalarTimeSeriesBuilder {
+                step_start: Step(0),
+                wall_time_start: WallTime::new(1235.0).unwrap(),
```
nfelt, line 246: Optional, but left to my own devices I would probably either omit some of these default values (aka require them to be set on the builder) or at least try to have the defaults be sort of the "minimally arbitrary" values (e.g. start wall time at 0.0 incrementing by 1.0, make len 1, make `eval` just be 0 always, etc.).

Otherwise we still have `test_read_scalars()` making assertions like `assert_eq!(xent_data.wall_time, vec![1235.0, 1236.0, 1237.0]);` that can't be understood to be correct without consulting this default implementation.  By contrast, `test_list_scalars()` is explicit at least about the `wall_time_start` and `len` values and thus its assertions are self-contained.

nfelt (approved):
Thanks!

tensorboard/data/server/commit.rs
wchargin: Thanks. You have inspired me to be less lazy. Done.

nfelt: `<meme>what is my purpose?</meme>`

PR #4434, opened 2020-12-08
ci: build and store data server Pip package

wchargin:
Summary:
A new CI job in our existing workflow builds the RustBoard binary and
its Pip package, and uploads the Pip package as an artifact on the job.
(The package is not uploaded to PyPI or anything like that.)

For now, we only build on Linux. Once the package build script supports
macOS, we should hopefully be able to just add that to the platform
matrix in this job config. Windows will probably need some adjustments
to deal with the shell in the action workflow and the `.exe` suffix on
the binary. We’ll handle that later.

Test Plan:
Check that CI properly builds a package. Download and install it into
your virtualenv. Then, run TensorBoard (with #4429) and check that it
locates the RustBoard server.

wchargin-branch: rust-tds-pip-ci


.github/workflows/ci.yml
```diff
@@ -97,6 +97,47 @@ jobs:
...
+            -c ./tensorboard/pip_package/requirements_dev.txt \
+            setuptools wheel
+      - name: 'Build'
+        run: cd tensorboard/data/server/ && cargo build --release
+      - name: 'Test'
+        run: cd tensorboard/data/server/ && cargo test --release
+      - name: 'Package'
+        run: |
+          rm -rf /tmp/pip_package && mkdir /tmp/pip_package
+          python tensorboard/data/server/pip_package/build.py \
```
stephanwlee, line 131: Just wondering, you do have a [bazel target for this](https://github.com/tensorflow/tensorboard/pull/4429/files#diff-cd744ef136a1ace4eea485e8399ac9f8b4aa457ba39334f7c0bf04daa8fbd2aaR6-R13), right? Are you manually invoking it here to (1) not require bazel and (2) specify output like code here?

stephanwlee:
This is kinda cool 
image

.github/workflows/ci.yml
wchargin: Exactly. Not requiring Bazel is the main point. We can build the whole
Rust package from scratch faster `bazel fetch //tensorboard/...` with
cache, and by building with Cargo rather than Bazel we avoid needing to
get Bazel set up on macOS and Windows CI (with any caching headaches
that that might entail).

wchargin:
thanks!

stephanwlee (approved):

PR #4436, opened 2020-12-08
rust: add macOS Pip package builds

wchargin:
Summary:
This patch adds support for macOS to the build script and the CI
workflow. Remains to be seen how to get around Gatekeeper issues.

Test Plan:

 1. Acquire access to a macOS environment.
 2. Wait for the GitHub Actions workflow to succeed.
 3. Make a new virtualenv.
 4. Download the `rustboard-macos-10.15.zip` artifact.
 5. Unzip it: `unzip rustboard-macos-10.15.zip`.
 6. Install TensorBoard: `pip install tb-nightly`. (If #4429 is not yet
    in nightlies, instead install a Pip package from head.)
 7. Install the downloaded wheel: `pip install *.whl`.
 8. Launch TensorBoard with these flags:

    ```
    tensorboard --logdir ~/tensorboard_data/mnist --bind_all \
        --load_fast --verbosity 0
    ```

    This should print a log line, “Server binary (from Python package):
    [path to venv]/tensorboard_data_server/bin/server”, and the normal
    TensorBoard UI should properly display data.

wchargin-branch: rust-tds-pip-macos

PR #4440, opened 2020-12-09
rust: bind server to `::1` by default

wchargin:
Summary:
Previously, the data server defaulted to host `::0`, the IPv6 wildcard
address and analogue of `0.0.0.0`, ~a.k.a. `--bind_all`. But we can
safely restrict this to just `::1`, the loopback address and analogue of
`127.0.0.1`, allowing localhost connections only. @psybuzz reports that
macOS shows a security dialog about incoming connections to the data
server, and that this fixes it.

Test Plan:
Running `tensorboard --load_fast` still works fine, since TensorBoard is
connecting to a data server running on the same machine.

wchargin-branch: rust-default-bind-loopback

### DIFF
--- a/tensorboard/data/server/cli.rs
+++ b/tensorboard/data/server/cli.rs
@@ -48,7 +48,7 @@ struct Opts {
     ///
     /// IP address to bind this server to. May be an IPv4 address (e.g., 127.0.0.1 or 0.0.0.0) or
     /// an IPv6 address (e.g., ::1 or ::0).
-    #[clap(long, default_value = "::0")]
+    #[clap(long, default_value = "::1")]
     host: IpAddr,
 
     /// Bind to this port


wchargin:
@psybuzz: Do you know a way to trigger again the “allow network
connections” popup that you mentioned in #4436? (Maybe remove `server`
from a list of firewall rules in sysprefs or something?) If so, could
you check whether you can reproduce the warning, and if so whether this
patch fixes it?

If you can’t easily reproduce, no worries: this patch is nice anyway, so
we can just submit it blind.

psybuzz (approved):
Just checked using the new mac artifact from this PR, and the "server" security popup is gone!  (under --bind_all, the one-time "python" popup still triggers).

wchargin:
Great; thanks! I assume that the one for `python` is by design, since
TensorBoard really is starting a web server. Appreciate the testing.

PR #4443, opened 2020-12-09
line chart: make threejs coordinator match DOM size

stephanwlee:
Previously, we had a camera coordinate system where we had a fixed camera
view point. This was great because irrespective of the actual DOM size,
we could keep the objects position static. However, this approach has a
major problem. Now that we want to render shapes like triangle and
circle, geometry that is defined in the camera coordinate system has to
account for mismatch in camera coordinate aspect ratio vs. DOM camera
coordinate aspect ratio.

We want circle that is not an oval in browser pixel location. To do
that, when creating a geometry in the camera coordinates, we need to
account for discrepancy in aspect ratio, which generally is harder than
just matching the camera view point to that of the DOM size. As such,
this change removes the camera coordinate system separate from DOM but
supports the y-axis flipping which is the only remaining difference in
the coordinate systems we support.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/coordinator.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/coordinator.ts
@@ -62,21 +62,18 @@ export class Coordinator {
   }
 
   /**
-   * Returns directionality of the y axis of the output space.
+   * Returns whether y axis is pointing down in the output space.
    *
    * ↑
-   * | viewBoxYDirection = +1
+   * | isYAxisPointedDown = false (e.g., cartesian coordinates, 3d scene)
    * |
    * |-------------→
    * |
-   * | viewBoxYDirection = -1
+   * | isYAxisPointedDown = true (e.g., DOM)
    * ↓
-   *
-   * +1: Suitable in 3-D aware context where you have full cartesian coordinates.
-   * -1: Suitable for DOM where <0, 0> is top-left corner of a DOM.
    */
-  getYAxisDirection(): -1 | 1 {
-    return -1;
+  isYAxisPointedDown(): boolean {
+    return true;
   }
 
   setXScale(scale: Scale) {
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/coordinator.ts
```diff
@@ -102,7 +121,9 @@ export class Coordinator {
       ),
       this.yScale.forward(
         domain.y,
-        [rect.y + rect.height, rect.y],
+        this.getYAxisDirection() === 1
+          ? [rect.y, rect.y + +rect.height]
```
psybuzz, line 125: Is the 2nd '+' needed?  `rect.height` already seems to be a number

tensorboard/webapp/widgets/line_chart_v2/lib/coordinator_test.ts
```diff
@@ -159,7 +159,7 @@ describe('line_chart_v2/lib/coordinator test', () => {
     });
 
     describe('#transformDataToUiCoord', () => {
-      beforeEach(() => {
+      it('converts with y-axis pointing up', () => {
```
psybuzz, line 162: Maybe I'm confusing myself, but this test uses the default coordinator with getYAxisDirection -1.  Should the text say "converts with y-axis pointing down"?

tensorboard/webapp/widgets/line_chart_v2/lib/coordinator.ts
```diff
@@ -60,6 +61,24 @@ export class Coordinator {
...
+   * |
+   * | viewBoxYDirection = -1
+   * ↓
+   *
+   * +1: Suitable in 3-D aware context where you have full cartesian coordinates.
+   * -1: Suitable for DOM where <0, 0> is top-left corner of a DOM.
+   */
+  getYAxisDirection(): -1 | 1 {
+    return -1;
+  }
```
psybuzz, line 80: I really do like the helpful ASCII comment, it's really clear!  An optional alternative is to name this `isYAxisPointedDown(): boolean`, which may help when reading callsites without having to go back to the ascii for reference.

psybuzz (approved):
lgtm, the description makes a lot of sense.  Also I didn't realize TS allowed return types `-1 | 1`!

tensorboard/webapp/widgets/line_chart_v2/lib/coordinator_test.ts
stephanwlee: This is for the ThreeCoordinator where y-axis is pointing up (L142 reads: `describe('ThreeCoordinator', () => {`)

psybuzz: Ah, gotcha.  Silly me for not scrolling up to read around this line!

PR #4444, opened 2020-12-09
line chart: omit nan from interactive view

stephanwlee:
Previously, we were trying to render NaN values on the line chart
interactive layer (circle to indicate where the tooltip is at). This
caused SVG to complain and populate error messages on the console.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_interactive_view.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_interactive_view.ts
@@ -406,8 +406,8 @@ export class LineChartInteractiveViewComponent
     );
   }
 
-  shouldRenderTooltipPoint(point: Point | undefined): boolean {
-    return point !== undefined && !isNaN(point.x) && !isNaN(point.y);
+  shouldRenderTooltipPoint(point: Point | null): boolean {
+    return point !== null && !isNaN(point.x) && !isNaN(point.y);
   }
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_interactive_view.ts
```diff
@@ -405,6 +406,10 @@ export class LineChartInteractiveViewComponent
     );
   }
 
+  shouldRenderTooltipPoint(point: Point | undefined): boolean {
+    return point !== undefined && !isNaN(point.x) && !isNaN(point.y);
```
psybuzz, line 410: Each datum within `cursoredData: TooltipDatum[]` is typed with a `point: {x: number; y: number} | null` property.  Shouldn't this function's signature be: `shouldRenderTooltipPoint(point: Point | null): boolean` ?

Can we use `!point` instead of `point !== undefined`, to avoid thinking about the difference?

psybuzz (approved):

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_interactive_view.ts
stephanwlee: `!point` is not good enough for type refinement and it causes type error for `point.x` (`Object is possibly 'null'.`). Changed from `undefined` to `null`.

PR #4447, opened 2020-12-10
rust: add logdir loading benchmark

wchargin:
Summary:
A new `bench` binary takes a `--logdir` parameter, loads the logdir to
completion, and then exits. This should report similar times to those
given by the info logs on the server itself, but is useful as a
standalone target because it exits upon completion. This makes it easier
to use with tools like `time(1)` or `hyperfine(1)` for automated studies
of performance over time.

I looked into using [`criterion`], which provides nice statistical
analysis over time. It works as advertised, but we want to be able to
point this benchmark at arbitrary logdirs and use it as a benchmarking
*tool* rather than just a fixed benchmark of a function. Same
justification for not using `cargo bench`.

The difference between the time reported in the `info!` logs and the
process runtime as reported by `time(1)` or `hyperfine(1)` tends to be
on the order of 5 ms on my machine, which I consider negligible enough
to not merit further investigation.

[`criterion`]: https://crates.io/crates/criterion

Test Plan:
To run with Cargo:

```
$ cargo run --release --bin bench -- --logdir ~/tensorboard_data/mnist/
```

To run with Bazel:

```
$ bazel run -c opt :bench -- --logdir ~/tensorboard_data/mnist/
```

For example, the gs://tensorboard-bench-logs/edge_cgan dataset completes
in 10.12 ± 0.04 seconds on my machine (Xeon W-2135 CPU, 64 GB RAM).

Note that these both run and report similar times. Note that `cargo run`
without `--bin` still runs the server.

wchargin-branch: rust-bench

### DIFF
--- a/tensorboard/data/server/BUILD
+++ b/tensorboard/data/server/BUILD
@@ -68,6 +68,18 @@ rust_test(
     ],
 )
 
+rust_binary(
+    name = "bench",
+    srcs = ["bench.rs"],
+    edition = "2018",
+    deps = [
+        ":rustboard_core",
+        "//third_party/rust:clap",
+        "//third_party/rust:env_logger",
+        "//third_party/rust:log",
+    ],
+)
+
 rust_doc_test(
     name = "rustboard_core_doc_test",
     dep = ":rustboard_core",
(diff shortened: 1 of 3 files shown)


psybuzz (approved):
Tried it and it works nicely!

PR #4450, opened 2020-12-10
Introduce a source code diff view component

psybuzz:
The existing SourceCode components are used in the Debugger V2 dashboard
to provide a readonly view of source code. This change branches the
NgComponent into a variant used solely for rendering a diff view. Source code
diffs may render 2 text contents side-by-side or as 1 unified inline diff.

Drive-by: migrated the existing (non-diff) SourceCode component to also
use the ResizeDetector.

Manually tested that resizing the Debugger V2 plugin still works:
image

Notes:

Why isn't this just part of the existing SourceCodeComponent?

The underlying Monaco classes are different. While they share common
functionality (autoresize, scrolling text into view, selecting regions [1]), the
2 subclasses have some fundamental differences.
- DiffEditor manages 2 source code models, each of which can have its
  own separate gutters, settings, etc.  A unified SourceCodeComponent
  supporting diff/non-diff would need all inputs to specify which editor
  it applies to.
- Diff editor has unique input options, e.g. renderSideBySide.

Until there is a stronger use case to benefit from reuse, I think keeping
these separate for now makes sense.

[1] https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.ieditor.html

### DIFF
--- a/tensorboard/webapp/widgets/source_code/source_code_component.ts
+++ b/tensorboard/webapp/widgets/source_code/source_code_component.ts
@@ -17,22 +17,15 @@ import {
   Component,
   ElementRef,
   Input,
+  OnChanges,
   SimpleChanges,
   ViewChild,
-  OnChanges,
-  OnDestroy,
-  OnInit,
 } from '@angular/core';
-import {fromEvent, interval, Subject} from 'rxjs';
-import {debounce, takeUntil, tap} from 'rxjs/operators';
-
 import {
-  DEFAULT_CODE_LANGUAGE,
   DEFAULT_CODE_FONT_SIZE,
+  DEFAULT_CODE_LANGUAGE,
   RESIZE_DEBOUNCE_INTERVAL_MS,
 } from './editor_options';
-
-/** @typehack */ import * as _typeHackRxjs from 'rxjs';
 
 @Component({
   selector: 'source-code-component',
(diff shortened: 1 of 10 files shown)


tensorboard/webapp/widgets/source_code/source_code_component.ts
```diff
@@ -67,7 +68,7 @@ export class SourceCodeComponent implements OnInit, OnChanges, OnDestroy {
     const resizePipe = fromEvent(window, 'resize')
       .pipe(
         takeUntil(this.ngUnsubscribe),
-        debounce(() => interval(RESIZE_DEBOUNCE_INTERAVL_MS)),
+        debounce(() => interval(RESIZE_DEBOUNCE_INTERVAL_MS)),
```
stephanwlee, line 71: Would it be terrible to use the ResizeDetector?  https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/widgets/resize_detector_directive.ts

tensorboard/webapp/widgets/source_code/source_code_diff_component.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  // When a diff is shown, the two versions can be rendered in one of 2 modes:
+  // - side by side: 2 scrollable frames with a vertical separator
+  // - inline: 1 scrollable frame showing modified and original lines
+  // This component does not respond to changes in this property.
+  @Input()
+  renderSideBySide: boolean = true;
+
+  @Input()
+  monaco: any;
```
stephanwlee, line 54: Is it possible to better type Monaco?

stephanwlee (approved):

psybuzz:
Thanks for the review!

PR #4453, opened 2020-12-11
make TF deps include //tensorboard/summary:tf_summary

nfelt:
This makes our `//tensorboard:expect_tensorflow_installed` placeholder target (reflecting a dep on the TF pip package from the virtualenv) include a dep on `//tensorboard/summary:tf_summary`.

This fixes "for once and for all" the problem wherein code within TensorBoard that depends on TF but not on our `tf_summary` shim logic prints the annoying `Limited tf.compat.v2.summary API due to missing TensorBoard installation.` message, and then fails to actually include the `tf.compat.v2.summary.scalar` endpoint.  Consequently we can remove ad-hoc additions of `tf_summary` deps from a few places.

This arrangement is also a better analog to the internal situation.

Confirmed that this syncs internally with a copybara change - see cl/346894347.

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -16,10 +16,6 @@ package_group(
 )
 
 # The standard TensorBoard binary that serves the webapp.
-# The two //tensorboard/summary lines are to suppress an
-# annoying startup warning emitted by TensorFlow that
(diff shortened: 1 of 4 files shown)


wchargin (approved):
Ah, that’s actually pretty sensible. Nice.

PR #4455, opened 2020-12-11
cleanup: remove unused tf_web_test

stephanwlee:
tf_web_test is a build rule specialized to test Polymer 1 components.
While we want to have tests, we are (1) making great strides rewriting
critical pieces in Angular (2) not writing much code in Polymer and (3)
hard to maintain tf_web_test which is largely built around
tf_web_library and rules_closure (which is deprecated).

Since we are not actively working to rewrite the tests, it would make
sense to remove these deadcodes.

WANT_LGTM=all


nfelt (approved):
LGTM since I agree we don't really want to be making big changes/investments to this codebase, and it seems unlikely we're going to have the extra bandwidth to rewrite these.

That said maybe worth getting another +1 from a teammate in case anyone objects.

stephanwlee:
Added other folks (not adding everyone I can think of though)

wchargin (approved):

bileschi (approved):

PR #4456, opened 2020-12-11
Do not specify default values in QueryParamsFeatureFlagDataSource. Rely on initialState defaults instead.

bmd3k:

  In the near future we will want feature flag default values to be customizable by different flavors of Tensorboard (TensorBoard.dev vs OSS TensorBoard vs TensorBoard.corp).

  Currently default values are, in practice, set by the QueryParamsFeatureFlagDataSource. This is not the mechanism that we want to use to specify default values so we remove this responsibility out of the DataSource.

  * Change TBFeatureFlagDataSource interface so that a Partial<FeatureFlags> can be returned. 
  * QueryParamsFeatureFlagDataSource returns Partial<FeatureFlags> only if corresponding query parameter was specified on URL.
  * Reducers ignore unspecified FeatureFlag properties from the data source and instead fall back to those already in State. In practice, for the moment, this is just the reducer initialState.

  * Wrote tests.
  * Bring up a tensorboard and load scalars dashboard.  Ensure default value for inColab is used (by looking at Time Series dashboard requests).
  * Override with `tensorboardColab=true` query param and see that Time Series dashboard now uses GET instead of POST for requests.
  * Override with `tensorboardColab` query param and see that Time Series dashboard again uses POST.

### DIFF
--- a/tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
+++ b/tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
@@ -27,8 +27,8 @@ import {FeatureFlags} from '../types';
  * does not have enough information to provide values. In that case, the
  * corresponding feature flag values should remain unchanged in the State.
  */
-export const featuresLoaded = createAction(
-  '[FEATURE FLAG] Features Loaded',
+export const partialFeatureFlagsLoaded = createAction(
+  '[FEATURE FLAG] Partial Feature Flags Loaded',
   props<{
     features: Partial<FeatureFlags>;
   }>()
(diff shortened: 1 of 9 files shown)


tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
```diff
@@ -20,9 +20,16 @@ import {FeatureFlags} from '../types';
...
 
+/**
+ * Signals that a data source has loaded feature flag values.
+ *
+ * Some or all feature flag properties can be unspecified if the data source
+ * does not have enough information to provide values. In that case, the
+ * corresponding feature flag values should remain unchanged in the State.
+ */
 export const featuresLoaded = createAction(
   '[FEATURE FLAG] Features Loaded',
```
stephanwlee, line 31: Perhaps call this "Partial Feature Flags Loaded"?

stephanwlee (approved):
I love it! :D

PR #4457, opened 2020-12-11
line chart: fix layout in Safari

stephanwlee:
In FF and Chrome, block element under grid takes the full height with
`height: 100%` but it seems to behave a bit differently with Safari.
This change changes block to flex which fixes short axes, small grid,
and weird interactivity.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
@@ -48,7 +48,7 @@ import {
   styles: [
     `
       :host {
-        display: block;
+        display: flex;
         overflow: hidden;
       }
 
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_grid_view.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_grid_view.ts
@@ -43,7 +43,7 @@ import {
   styles: [
     `
       :host {
-        display: block;
+        display: flex;
         overflow: hidden;
       }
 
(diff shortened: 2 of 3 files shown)


stephanwlee (commented):
(to reviewer: sorry about the noisy formatter change. For some reason, prettier on my different machine decided to format the template but it passes the frontend lint so... what's up?)

wchargin (approved):
On my machine, Prettier 2.1.1 is happy to leave the original code as is,
and is also happy to reformat it to the version in this PR if you add
newlines at the start and end of the template. I.e., if you apply this
diff and run Prettier, then it generates the output in the PR:

```diff
diff --git a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
index 26c835acf..365a7b939 100644
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
@@ -25,3 +25,4 @@ import {
   selector: 'line-chart-axis',
-  template: `<svg [ngClass]="axis">
+  template: `
+  <svg [ngClass]="axis">
     <ng-container *ngIf="axis === 'x'; else yAxisLine">
@@ -46,3 +47,4 @@ import {
     </ng-container>
-  </svg>`,
+  </svg>
+  `,
   styles: [
```

Can you check that you have the right version of Prettier and, in any
case, revert the formatting change?

wchargin:
In general, that Prettier is literal in your string literals is
expected, I think, so this just feels to me like a “revert spurious
change” request that’s independent of the formatter.

PR #4458, opened 2020-12-11
Provide and consume FeatureFlagState StoreConfig, allowing other Tensorboards to override default feature flag values.

bmd3k:

  Some instances of TensorBoard (notably, TensorBoard.dev), might want to provide different default
  values for feature flags.

  Define injection token that allows us to provide and consume a StoreConfig<FeatureFlagState>. StoreConfig
  allows an initialState value to be specified for the state.

  Provide a fallback StoreConfig that simply defines initialState to be. . . the initialState we already use.

  * Load local TensorBoard and ensure Scalars and TimeSeries dashboards successfully load scalar data.
  * Test that tensorboardColab query parameter impacts how TimeSeries dashboard loads scalar data.
  * Patch changes into internal repo and modify TensorBoard.dev to provide its own StoreConfig with an initial state
     that includes a non-empty value for enabledExperimentalPlugins. Load the TensorBoard.dev and ensure those
     experimental plugins appear by default.

### DIFF
--- a/tensorboard/webapp/feature_flag/store/feature_flag_reducers.ts
+++ b/tensorboard/webapp/feature_flag/store/feature_flag_reducers.ts
@@ -15,18 +15,9 @@ limitations under the License.
 import {Action, createReducer, on} from '@ngrx/store';
 
 import * as actions from '../actions/feature_flag_actions';
-import {FeatureFlagState} from './feature_flag_types';
+import {FeatureFlagState, initialState} from './feature_flag_types';
 
 /** @typehack */ import * as _typeHackStore from '@ngrx/store';
-
-const initialState: FeatureFlagState = {
-  isFeatureFlagsLoaded: false,
-  features: {
-    enabledExperimentalPlugins: [],
-    inColab: false,
-    enableGpuChart: false,
-  },
-};
 
 const reducer = createReducer<FeatureFlagState>(
   initialState,
@@ -42,6 +33,6 @@ const reducer = createReducer<FeatureFlagState>(
   })
 );
 
-export function reducers(state: FeatureFlagState, action: Action) {
+export function reducers(state: FeatureFlagState | undefined, action: Action) {
   return reducer(state, action);
 }
--- a/tensorboard/webapp/feature_flag/store/feature_flag_reducers.ts
+++ b/tensorboard/webapp/feature_flag/store/feature_flag_reducers.ts
@@ -15,7 +15,8 @@ limitations under the License.
 import {Action, createReducer, on} from '@ngrx/store';
 
 import * as actions from '../actions/feature_flag_actions';
-import {FeatureFlagState, initialState} from './feature_flag_types';
+import {initialState} from './feature_flag_store_config_provider';
+import {FeatureFlagState} from './feature_flag_types';
 
 /** @typehack */ import * as _typeHackStore from '@ngrx/store';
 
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/feature_flag/store/feature_flag_store_config_provider.ts
```diff
@@ -0,0 +1,41 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {InjectionToken} from '@angular/core';
+import {StoreConfig} from '@ngrx/store';
+
+import {FeatureFlagState, initialState} from './feature_flag_types';
+
+// /** @typehack */ import * as _typeHackStore from '@ngrx/store';
```
stephanwlee, line 20: Remove me :)

tensorboard/webapp/feature_flag/store/feature_flag_store_config_provider.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Injection token for providing feature flag StoreConfig.
+ */
+export const FEATURE_FLAG_STORE_CONFIG_TOKEN: InjectionToken<StoreConfig<
+  FeatureFlagState
+>> = new InjectionToken<StoreConfig<FeatureFlagState>>(
+  'Feature Flag Store Config'
```
stephanwlee, line 28: For consistency: `[Feature Flag] Store Config` (this pattern is from ngrx docs)

tensorboard/webapp/feature_flag/store/feature_flag_types.ts
```diff
@@ -25,3 +25,12 @@ export interface FeatureFlagState {
...
 }
+
+export const initialState: FeatureFlagState = {
+  isFeatureFlagsLoaded: false,
+  features: {
+    enabledExperimentalPlugins: [],
+    inColab: false,
+    enableGpuChart: false,
+  },
+};
```
stephanwlee, line 36: It is a bit weird that this is defined in the types file. Can this just move to the default provider?

tensorboard/webapp/feature_flag/store/feature_flag_store_config_provider.ts
bmd3k: Done.

tensorboard/webapp/feature_flag/store/feature_flag_store_config_provider.ts
bmd3k: Done.

tensorboard/webapp/feature_flag/store/feature_flag_types.ts
bmd3k: Done.

tensorboard/webapp/feature_flag/store/feature_flag_reducers.ts
```diff
@@ -15,7 +15,8 @@ limitations under the License.
 import {Action, createReducer, on} from '@ngrx/store';
 
 import * as actions from '../actions/feature_flag_actions';
-import {FeatureFlagState, initialState} from './feature_flag_types';
+import {initialState} from './feature_flag_store_config_provider';
```
stephanwlee, line 18: No action required: FYI, initialState is already populated when the store is created when we pass the third argument to the `StoreModule.forFeature` yet we have to define the `initialState` here. I find that so funky :\

stephanwlee (approved):

tensorboard/webapp/feature_flag/store/feature_flag_reducers.ts
bmd3k: Agreed, I spent some time trying to understand the best practice. I had considered not setting initialState in the default StoreConfig provider (and therefore actually making the initialState here meaningful) but that just seemed silly for other reasons.

PR #4459, opened 2020-12-11
chore: typecheck 'monaco' types in source code component

psybuzz:
This updates a few type annotations in the SourceCode component,
mainly relating to the use of `monaco`. Now, Monaco editor uses are
typechecked.

Test plan:
Manually changed an instance of `this.monaco.editor.create` to
`this.monaco.editor.create2` and saw that `bazel build tensorboard`
fails.

Googlers, see associated Copybara change: cl/347056094.

### DIFF
--- a/tensorboard/webapp/widgets/source_code/source_code_component.ts
+++ b/tensorboard/webapp/widgets/source_code/source_code_component.ts
@@ -46,9 +46,8 @@ export class SourceCodeComponent implements OnInit, OnChanges, OnDestroy {
   @Input()
   focusedLineno: number | null = null;
 
-  // TODO(cais): Explore better typing by depending on external libraries.
   @Input()
-  monaco: any | null = null;
+  monaco: typeof monaco | null = null;
 
   @ViewChild('codeViewerContainer', {static: true, read: ElementRef})
   private readonly codeViewerContainer!: ElementRef<HTMLDivElement>;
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/source_code/source_code_component.ts
```diff
@@ -46,9 +46,8 @@ export class SourceCodeComponent implements OnInit, OnChanges, OnDestroy {
   @Input()
   focusedLineno: number | null = null;
 
-  // TODO(cais): Explore better typing by depending on external libraries.
   @Input()
-  monaco: any | null = null;
+  monaco: typeof monaco | null = null;
```
stephanwlee, line 50: Is this referring to an ambient monaco variable? :\ Hate how it does that :(

stephanwlee (approved):

nfelt (approved):

tensorboard/webapp/widgets/source_code/source_code_component.ts
psybuzz: Yup.  Unfortunately the official `monaco.d.ts` uses TS Namespaces for a few top-level things, like `monaco` and `monaco.editor`.  I couldn't find other projects that pass around the `window.monaco` object around directly _and_ want to properly type it.  The only way I found to type those was to use this typeof keyword, as suggested by a StackOverflow post.

https://stackoverflow.com/questions/42815932/use-namespace-as-type-in-typescript

PR #4466, opened 2020-12-14
Pass FeatureFlags to CUSTOM_ELEMENT dashboards; use in Scalars dashboard for inColab and scalarsBatchSize.

bmd3k:

  We want to be able to use the FeatureFlags mechanism to modify behavior in dashboards.
  The Scalars dashboard, for example, could use two FeatureFlags:
  * `isColab`, which maps to the `tensorboardColab`  query parameter. The Scalars dashboard already independently
     reads this query parameter but it is more consistent to get it from FeatureFlags.
  * `scalarsBatchSize`, which is new functionality. TensorBoard.dev will use FeatureFlags to modify the /scalars_multirun
     batch size.

  * plugins_component on Angular side, which manages dashboard creation, now waits for FeatureFlags to be
     loaded before loading dashboards. It passes the featureFlags property of each dashboard, which can choose to
     use or ignore it as a polymer property.
  * introduce the `scalarsBatchSize` feature flag and map it to the query parameter of the same name.
  * scalars dashboard code uses the `featureFlags` property to read `isColab` and `scalarsBatchSize` feature flags
    and uses the values to modify behavior.

  * Wrote some tests.
  * Brought up a local tensorboard and ensured that `tensorboardColab` and `scalarsBatchSize` query parameters
     properly modify scalars dashboard behavior.
  * Temporarily synthetically delay query parameter loading by 10 seconds. Load tensorboard and observe that dashboards
     are not loaded until the 10 seconds elapses and that they load correctly.

### DIFF
--- a/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
+++ b/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
@@ -334,7 +334,7 @@ export class TfScalarCard extends PolymerElement {
       runs.push(run);
     }
 
-    const batchSize = this.batchSize || DEFAULT_BATCH_SIZE;
+    const batchSize = this.batchSize ?? DEFAULT_BATCH_SIZE;
 
     const requestGroups = [];
     for (const [tag, runs] of runsByTag) {
--- a/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
+++ b/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
@@ -30,6 +30,21 @@ import {DEFAULT_TOOLTIP_COLUMNS} from '../../../components/vz_line_chart2/vz-lin
 
 type RunTagItem = {run: string; tag: string};
 
+// Request at most this many runs at once.
+//
+// Back-of-the-envelope math: each scalar datum JSON value contains
+// two floats and a small-ish integer. Floats are about 18 bytes,
+// since f64s have -log_10(2^-53) ~= 16 digits of precision plus
+// decimal point and leading zero. Small-ish integers (steps) are
+// about 5 bytes. Add JSON overhead `[,,],` and you're looking at
+// about 48 bytes per datum. With standard downsampling of
+// 1000 points per time series, expect ~50 KB of response payload
+// per requested time series.
+//
+// Requesting 64 time series warrants a ~3 MB response, which seems
+// reasonable.
+const DEFAULT_BATCH_SIZE = 64;
+
 /**
  * Save the initial URL query params, before the AppRoutingEffects initialize.
  */
@@ -228,6 +243,12 @@ export class TfScalarCard extends PolymerElement {
   @property({type: String})
   tooltipSortingMethod: string;
 
+  @property({type: Number})
+  batchSize: number;
+
+  @property({type: Boolean})
+  inColab: number;
+
   // This function is called when data is received from the backend.
   @property({type: Object})
   _loadDataCallback: object = (scalarChart, item, maybeData) => {
@@ -273,8 +294,7 @@ export class TfScalarCard extends PolymerElement {
     // Google-internal Colab doesn't support HTTP POST requests, so we fall
     // back to HTTP GET (even though public Colab supports POST).
     // See b/126387106.
-    const inColab = initialURLSearchParams.get('tensorboardColab') === 'true';
-    if (inColab) {
+    if (this.inColab) {
       return this._requestDataGet(items, onLoad, onFinish);
     } else {
       return this._requestDataPost(items, onLoad, onFinish);
@@ -314,25 +334,12 @@ export class TfScalarCard extends PolymerElement {
       runs.push(run);
     }
 
-    // Request at most this many runs at once.
-    //
-    // Back-of-the-envelope math: each scalar datum JSON value contains
-    // two floats and a small-ish integer. Floats are about 18 bytes,
-    // since f64s have -log_10(2^-53) ~= 16 digits of precision plus
-    // decimal point and leading zero. Small-ish integers (steps) are
-    // about 5 bytes. Add JSON overhead `[,,],` and you're looking at
-    // about 48 bytes per datum. With standard downsampling of
-    // 1000 points per time series, expect ~50 KB of response payload
-    // per requested time series.
-    //
-    // Requesting 64 time series warrants a ~3 MB response, which seems
-    // reasonable.
-    const BATCH_SIZE = 64;
+    const batchSize = this.batchSize || DEFAULT_BATCH_SIZE;
 
     const requestGroups = [];
     for (const [tag, runs] of runsByTag) {
-      for (let i = 0; i < runs.length; i += BATCH_SIZE) {
-        requestGroups.push({tag, runs: runs.slice(i, i + BATCH_SIZE)});
+      for (let i = 0; i < runs.length; i += batchSize) {
+        requestGroups.push({tag, runs: runs.slice(i, i + batchSize)});
       }
     }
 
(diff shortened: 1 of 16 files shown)


tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-dashboard.ts
```diff
@@ -166,6 +167,8 @@ class TfScalarDashboard extends LegacyElementMixin(ArrayUpdateHelper) {
                   tag="[[item.tag]]"
                   tooltip-sorting-method="[[_tooltipSortingMethod]]"
                   x-type="[[_xType]]"
+                  batch-size="[[featureFlags.scalarsBatchSize]]"
```
stephanwlee, line 170: How does this behave when the `featureFlags` is undefined? (since it can be undefined as per L213)

tensorboard/webapp/feature_flag/types.ts
```diff
@@ -20,4 +20,10 @@ export interface FeatureFlags {
   inColab: boolean;
   // Whether to enable our experimental GPU line chart.
   enableGpuChart: boolean;
+  // Maximum number of runs to include in a request to get scalar data.
+  // `undefined` indicates that we should rely on defaults defined in the
+  // dashboards code.
+  //
+  // See: https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
```
stephanwlee, line 27: You may also want to tag https://github.com/tensorflow/tensorboard/blob/648ffe001eb910e68493506cbe468f9caab0f9c6/tensorboard/webapp/metrics/effects/index.ts#L192-L201.

tensorboard/webapp/plugins/plugins_component.ts
```diff
@@ -135,13 +148,17 @@ export class PluginsComponent implements OnChanges {
   private readonly pluginInstances = new Map<string, HTMLElement>();
 
   ngOnChanges(change: SimpleChanges): void {
+    if (!this.isFeatureFlagsLoaded || !this.activeKnownPlugin) {
```
stephanwlee, line 151: I think it technically is possible for the app to go from truthy `this.activeKnownPlugin` to falsy `this.activeKnownPlugin` when you have experiments (navigate from an exp with data to one without data). I think it is more correct to hide the previous activeKnownPlugin in that case. WDYT?

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
```diff
@@ -34,3 +34,5 @@ export abstract class TBFeatureFlagDataSource {
 export const EXPERIMENTAL_PLUGIN_QUERY_PARAM_KEY = 'experimentalPlugin';
 
 export const GPU_LINE_CHART_QUERY_PARAM_KEY = 'fastChart';
+
+export const SCALARS_BATCH_SIZE_PARAM_KEY = 'scalarsBatchSize';
```
stephanwlee, line 38: If you use a bit more generic name such as "maxDataBatchSize`, then you can reuse this under timeseries plugin.

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-dashboard.ts
bmd3k: batchSize and inColab end up being undefined and the scalars dashboard behaves as though they aren't set. That wasn't explicitly my intention but seems like the correct behavior? Thanks for asking me to check.

tensorboard/webapp/feature_flag/types.ts
bmd3k: I unfortunately don't really understand.  It doesn't look to me like there is anything related to batching in that code.

tensorboard/webapp/plugins/plugins_component.ts
bmd3k: I don't think the ngOnChanges() implementation at HEAD does anything when this.activeKnownPlugin is falsy. Is this a case we hadn't previously considered or am I introducing a bug?  Would I have to reproduce in tensorboard.corp?

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
bmd3k: Should it only tune scalars-related calls in Time Series dashboard?  In that case I think the name is fine (don't have a strong opinion though - 'maxDataBatchSize' might be just as good).

I don't think we should use this to tune anything non-scalars related since we don't (yet) know whether other things have similar performance characteristics.  It's entirely possible we will want a different value for batching histograms or blobs, for example.

tensorboard/webapp/feature_flag/types.ts
stephanwlee: Timeseries batches the request by default and it is not tunable today. This may prevent you from adopting TimeSeries but I just wanted to mention that this flag can be reused for timeseries.

tensorboard/webapp/plugins/plugins_component.ts
stephanwlee: True, I thought L160 would have handled that case, but it did not (i.e., your code does not cause regression). Could you leave a TODO or, if you think it is small enough change, just fix the bug?

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-dashboard.ts
stephanwlee: Got it. I am just surprised that `undefined.scalarsBatchSize` is legal in Polymerism but I forget Polymer :)

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
```diff
@@ -314,25 +334,12 @@ export class TfScalarCard extends PolymerElement {
...
-    // decimal point and leading zero. Small-ish integers (steps) are
-    // about 5 bytes. Add JSON overhead `[,,],` and you're looking at
-    // about 48 bytes per datum. With standard downsampling of
-    // 1000 points per time series, expect ~50 KB of response payload
-    // per requested time series.
-    //
-    // Requesting 64 time series warrants a ~3 MB response, which seems
-    // reasonable.
-    const BATCH_SIZE = 64;
+    const batchSize = this.batchSize || DEFAULT_BATCH_SIZE;
```
stephanwlee, line 337: I think it is more correct to write `this.batchSize ?? DEFAULT_BATCH_SIZE` since otherwise,
- if `this.batchSize` = 0, batchSize = DEFAULT_BATCH_SIZE
- if `this.batchSize` = -1, batchSize = -1
- if `this.batchSize` = 1, batchSize = 1

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
stephanwlee: Fair. SG.

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-dashboard.ts
bmd3k: Yes, I am also (pleasantly) surprised.

tensorboard/webapp/plugins/plugins_component.ts
bmd3k: Added TODO.

tensorboard/plugins/scalar/tf_scalar_dashboard/tf-scalar-card.ts
bmd3k: Yes. We could consider also enforcing that the batchSize > 0 but since this is for internal use only I'll wait to see if anybody ever complains.

stephanwlee (approved):

PR #4470, opened 2020-12-14
graph: use DataProvider-compatible APIs

wchargin:
Summary:
Events with top-level `tagged_run_metadata` are now transformed at read
time into blob sequence summaries. As a result, the graph plugin can get
all its data via tensor summaries. This also includes replacing calls to
`multiplexer.Graph` with reads of the `__run_graph__` tensor.

Since there’s an existing `PLUGIN_NAME_RUN_METADATA`, I had initially
hoped to use that for the summaries. But the graphs plugin actually
expects that data to include a graph (even though there is a separate
plugin called `PLUGIN_NAME_RUN_METADATA_WITH_GRAPH`) and there’s no way
to tell it apart from just the metadata. Thus, we implement this via a
new plugin name, which means that there are few structural changes to
the graph plugin code.

As written, the code still uses the multiplexer, but is set up to make
it easy to read from the data provider instead. We make that change in a
follow-up PR (see #4473).

Test Plan:
All data from the graphs demo (added in #4469) still works. Grepping for
`_multiplexer` in the graphs plugin code shows that the only calls are
to `PluginRunToTagToContent` and `Tensors`, both of which have
equivalents in the data provider API.

wchargin-branch: graph-data-provider-compatible


stephanwlee (approved):
I see some repetition with #4473. Am I reading it right?

wchargin:
> I see some repetition with #4473. Am I reading it right?

Not sure exactly what you mean? That PR is diffbased against this one.
They certainly are related. I separated the two PRs out because the
second one is pretty straightforward given this one, whereas I figured
that you might have input about the “new plugin name” approach or other
details in this PR. In other words, this PR makes the hard change easy,
and the next PR makes the easy change.

Does this help clarify?

stephanwlee:
I guess I still don't understand how to read your commits; some commit labels/names feel arbitrary and I cannot always make the correlation.

wchargin:
Ah. The names are somewhat arbitrary (but imho yours are, too, like
`sync` or `pp`). The easiest way to see the diffbase is just to look at
the base branch:

![Screenshot of the base branch view for PRs 4470 and 4473, showing that
4470 merges branch `foo` into `master` and 4473 merges branch `bar` into
`foo`, thus 4473 depends on 4470][ss]

But, if it would be helpful, I can try to more
consistently call out the dependencies in the description text (“builds
on #xyz”/“in a follow-up, …”). Or, if you have any suggestions or
examples of things that you find hard to understand, please let me know;
I’d be more than happy to adjust my workflow to make it easier to grok.

[ss]: https://user-images.githubusercontent.com/4317806/102176503-5f2fa680-3e56-11eb-960d-51ac9a649178.png

stephanwlee:
To be clear, I am not talking about the branch name. Rather, I am referring to your commit titles.

For instance, I see this bit repeated in this PR but looking at the commits, it is impossible to infer that #4473 was built on top of #4470 (or they share a common root)
 https://github.com/tensorflow/tensorboard/pull/4473/files#diff-65f6808e40d634101e802940b9c39d45301ea75826c4016f8156b658e6130d3dR129-R136 

#4470 
image

#4473 
image

(

wchargin:
Got it. The commits for #4473 don’t show that it’s built on top of this
PR because the base branch is this branch, so it’s comparing
`wchargin-graph-data-provider-compatible...wchargin-graph-data-provider-only`,
which indeed does not contain any commits from this PR. So I recommend
looking at the base branch to determine the order. I also try to open
the PRs in topological order, so #4470 should precede #4473 logically
just because it does so numerically.

As to the commit titles, my “update patch” is your “cr address” or
other people’s “ac” (address changes). I agree that they are mostly
devoid of information. (“Update diffbase” is just “merge changes from
parent branch into this branch”, or “rebase”, or “sync”. If you have a
suggestion for a clearer wording, let me know; ideally these wouldn’t
require much special thought on your part.)

And the extra commits in this PR are just because it hasn’t yet been
updated since its parent #4469 was merged, so `master` has an extra
commit (the squashed version of #4469) that this branch doesn’t have yet
and vice versa (the unsquashed version).

stephanwlee:
Ah, `wchargin-graph-data-provider-compatible...wchargin-graph-data-provider-only`... I totally missed that. My bad!

wchargin:
No problem! :-) there are lots of small fixes that I would like to make
to these UIs, like showing the diffbase graph à la Phabricator/Gerrit…

PR #4474, opened 2020-12-14
line chart: introduce formatter

stephanwlee:
Different use cases require you to format numbers a bit differently. For
instance, when viewing line chart in the time series dashboard, with the
wall time x axis type, you want to represent the numbers in time. This
change formalizes a notion of formatter that tooltip and axis can reuse.


tensorboard/webapp/widgets/line_chart_v2/lib/formatter_test.ts
```diff
@@ -0,0 +1,143 @@
...
+limitations under the License.
+==============================================================================*/
+import {
+  numberFormatter,
+  relativeTimeFormatter,
+  timeFormatter,
+  TEST_ONLY,
+} from './formatter';
+
+describe('line_chart_v2/sub_view/axis_formatter test', () => {
```
wchargin: Now `line_chart_v2/lib/formatter test`?

tensorboard/webapp/widgets/line_chart_v2/lib/formatter.ts
```diff
@@ -0,0 +1,122 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {format, scaleTime} from '../../../third_party/d3';
+
+export interface Formatter {
+  formatTick(x: number): string;
+  formatShort(x: number): string;
+  formatLong(x: number): string;
```
wchargin, line 20: It looks like the idea here is that `formatShort` is allowed to be lossy
and imprecise (`100001` → `1e5`) whereas `formatLong` is supposed to
give an ~exact value at the cost of a potentially longer string. Is that
right? If so, could you briefly document it?

tensorboard/webapp/widgets/line_chart_v2/lib/formatter.ts
```diff
@@ -0,0 +1,122 @@
...
+    builder.push(d3FloatFormatter(absX / SECOND_IN_MS), 'sec');
+  } else if (absX < HOUR_IN_MS) {
+    builder.push(d3FloatFormatter(absX / MINUTE_IN_MS), 'min');
+  } else if (absX < DAY_IN_MS) {
+    builder.push(d3FloatFormatter(absX / HOUR_IN_MS), 'hr');
+  } else if (absX < YEAR_IN_MS) {
+    builder.push(d3FloatFormatter(absX / DAY_IN_MS), 'day');
+  } else {
+    builder.push(d3FloatFormatter(absX / YEAR_IN_MS), 'yr');
+  }
```
wchargin, line 73: It looks like we always have `builder.length === 1` by the time that we
use `builder`. Why not just `let builder;` (maybe modulo renaming) and
assign to it rather than using an array?

tensorboard/webapp/widgets/line_chart_v2/lib/formatter.ts
```diff
@@ -0,0 +1,122 @@
...
+  },
+};
+
+const SECOND_IN_MS = 1000;
+const MINUTE_IN_MS = 60 * SECOND_IN_MS;
+const HOUR_IN_MS = 60 * MINUTE_IN_MS;
+const DAY_IN_MS = 24 * HOUR_IN_MS;
+const YEAR_IN_MS = 365 * DAY_IN_MS;
+
+function formatTime(x: number): string {
```
wchargin, line 55: Can we call this `formatRelativeTime`? It’s clear below from the name of
`relativeTimeFormatter`, but when I first read this, I had assumed that
this would take an epoch timestamp (like `1607992034.250`).

Could we also maybe document the implicit units of `x: number`, which
appear to be milliseconds?

tensorboard/webapp/widgets/line_chart_v2/lib/formatter.ts
```diff
@@ -0,0 +1,122 @@
...
+  formatTick: formatTime,
+  formatShort: formatTime,
+  formatLong: formatTime,
+};
+
+const d3TimeTickFormat = scaleTime().tickFormat();
+
+let localeOverride: string | undefined = undefined;
+
+export const timeFormatter: Formatter = {
```
wchargin, line 87: Similarly, maybe `wallTimeFormatter` disambiguates?

wchargin (commented):
submitting partial review since force-push means that I can’t write any
more comments

tensorboard/webapp/widgets/line_chart_v2/lib/formatter_test.ts
```diff
@@ -0,0 +1,143 @@
...
+          '2019'
+        );
+      });
+    });
+
+    describe('formatShort', () => {
+      it('formats using localization', () => {
+        expect(
+          timeFormatter.formatShort(new Date('2020-1-5 13:23').getTime())
+        ).toBe('1/5/20, 1:23:00 PM');
```
wchargin, line 130: Can we get this to spit out unambiguous dates? `1/5/20` is a good
example of the problem. I’d prefer `2020-01-05` myself, but also
amenable to `Jan 5, 2020` or similar. <https://xkcd.com/1179/>

Even if this is meant to use “the user’s current locale”, I personally
don’t find that very compelling. If you and I both use TensorBoard.dev,
and I load the plot and it says `1/5/20`, and I take a screenshot and
send it to you, and when you open the plot it says `5/1/20`, that’s
confusing.

tensorboard/webapp/widgets/line_chart_v2/lib/formatter.ts
```diff
@@ -0,0 +1,122 @@
...
+const SECOND_IN_MS = 1000;
+const MINUTE_IN_MS = 60 * SECOND_IN_MS;
+const HOUR_IN_MS = 60 * MINUTE_IN_MS;
+const DAY_IN_MS = 24 * HOUR_IN_MS;
+const YEAR_IN_MS = 365 * DAY_IN_MS;
+
+function formatTime(x: number): string {
+  if (x === 0) return '0';
+
+  const sign = x <= 0 ? -1 : 1;
```
wchargin, line 58: This can just be `Math.sign(x)` (equivalent since you return early on
zeros anyway). But since your call site needs a conditional, anyway,
maybe easier to inline?

```javascript
  return `${x < 0 ? '-' : ''}${builder.join('')}`;
```

Seems more direct; fewer things to keep track of.

tensorboard/webapp/widgets/line_chart_v2/lib/formatter.ts
wchargin: Ah, not quite; I missed that this was using the multi-argument form of
`push` (which I’m familiar with but not accustomed to). In that case,
what do you think about instead using

```javascript
  builder = d3FloatFormatter(absX / SECOND_IN_MS) + 'sec';
```

etc., or with template strings if you prefer? I still think the array
isn’t necessary since its length is constant and it’s only ever modified
once.

tensorboard/webapp/widgets/line_chart_v2/lib/formatter_test.ts
```diff
@@ -101,41 +103,41 @@ describe('line_chart_v2/lib/formatter test', () => {
...
       });
     });
 
     describe('formatShort', () => {
       it('formats using localization', () => {
         expect(
-          timeFormatter.formatShort(new Date('2020-1-5 13:23').getTime())
-        ).toBe('1/5/20, 1:23:00 PM');
+          wallTimeFormatter.formatShort(new Date('2020-1-5 13:23').getTime())
+        ).toBe('Jan 5, 20, 1:23:00 PM');
```
wchargin, line 132: “Jan 5, 20” feels unfamiliar to me. I would probably expect that to mean
“the two days January 5 and January 20” rather than “January 5, 2020”.
Is “Jan 5, 2020” too long? What if we only include the year when it’s
not the current year? (just providing some ideas; not blocking)

tensorboard/webapp/widgets/line_chart_v2/lib/formatter.ts
```diff
@@ -41,65 +65,81 @@ function formatNumberShort(x: number): string {
...
+
+/**
+ * Formats relative time in milliseconds with human readable unit.
+ */
+function formatRelativeTime(x: number): string {
   if (x === 0) return '0';
 
-  const sign = x <= 0 ? -1 : 1;
-  const builder = [];
+  let str = Math.sign(x) === 1 ? '' : '-';
```
wchargin, line 93: nit: why `Math.sign(x) === 1` rather than the simpler `x > 0`?

wchargin (approved):

PR #4475, opened 2020-12-15
line chart: transfer data with 64 bit precision

stephanwlee:
We premature optimized thinking that we should allocate less memory
using Float32Array where possible instead of Float64Array. The thing
is, with 32 bit precision, it is not possible to represent the date
serialized to integer.

This changes the data transfer layer only. We will diagnose and change
other places where we use Float32Array as needed (for instance, WebGL
uses 32 bit precision and threejs internally uses Float32Array a lot.
Having more precision does not help in all cases)

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/worker/compact_data_series.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/worker/compact_data_series.ts
@@ -34,7 +34,7 @@ export function compactDataSeries(
     return len + data.points.length;
   }, 0);
   let seriesIndex = 0;
-  const flattenedSeries = new Float32Array(totalLength * 2);
+  const flattenedSeries = new Float64Array(totalLength * 2);
   const idsAndLengths: Array<{id: string; length: number}> = [];
 
   for (const series of dataSeriesArr) {
@@ -57,7 +57,7 @@ export function decompactDataSeries(
   compactDataSeries: CompactDataSeries
 ): DataSeries[] {
   const {flattenedSeries, idsAndLengths} = compactDataSeries;
-  const rawData = new Float32Array(flattenedSeries);
+  const rawData = new Float64Array(flattenedSeries);
   const data: DataSeries[] = [];
 
   if (rawData.length % 2 !== 0) {
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/worker/compact_data_series.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/worker/compact_data_series.ts
@@ -17,7 +17,7 @@ import {DataSeries} from '../internal_types';
 
 export interface CompactDataSeries {
   idsAndLengths: Array<{id: string; length: number}>;
-  // buffer of packed f64s
+  // buffer of packed f64s that represents 2d coordinates, [x1, y1, x2, y2, ..., xn, yn].
   flattenedSeries: ArrayBuffer;
 }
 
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/worker/compact_data_series.ts
```diff
@@ -34,7 +34,7 @@ export function compactDataSeries(
     return len + data.points.length;
   }, 0);
   let seriesIndex = 0;
-  const flattenedSeries = new Float32Array(totalLength * 2);
+  const flattenedSeries = new Float64Array(totalLength * 2);
```
wchargin, line 37: It’s a bit suspicious to me that this change doesn’t need a change in
the types or documentation of `CompactDataSeries`. Could you add a brief
doc comment there indicating what `flattenedSeries` represents? i.e., a
buffer of packed f64s in some particular format?

tensorboard/webapp/widgets/line_chart_v2/lib/worker/compact_data_series_test.ts
```diff
@@ -104,49 +104,33 @@ describe('line_chart_v2/lib/compact_data_series', () => {
...
-          id: 'foo',
-          points: [{x: 1, y: 0}],
-        },
-      ]);
+      expect(actual).toEqual(dataSeries);
     });
 
-    it('does not round fractiional numbers', () => {
+    it('transfers date', () => {
+      const time = new Date('2020-01-01').getTime();
```
wchargin, line 124: Cool; looks like the imprecision for realistic dates is about 3 minutes,
since `f32::EPSILON * 1577836800_000.0 == 188092.8`. Change LGTM.

wchargin (approved):
> we should allocate less memory using Float32Array where possible
> instead of Float64Array

fwiw, in my experience the main perf boost from using f32 over f64 is
due to improved vectorization since you have twice the SIMD throughput.
For example, single-precision [`ADDPS`] adds up to sixteen f32s in one
go, whereas double-precision [`ADDPD`] adds up to only eight f64s, but
the two instructions [both have a throughput of 1/cycle][fog]. The raw
memory allocation isn’t too much either way in our case: 1000 scalar
time series each with 1000 scalars is on the order of 8 MB with f64s,
which is negligible compared to the amount of memory that TensorBoard
typically uses.

Also, in #4395, we changed the data provider gRPC from f64 to f32 for
scalar data to reflect that we really only guarantee preserving scalars
as single-precision (and since we never write double-precision scalars,
in either TF 1.x or TF 2.x). The change in this PR is fine with me; it’d
also be fine with me to switch back to f32 for the scalar data if you
desire.

[`ADDPS`]: https://www.felixcloutier.com/x86/addps
[`ADDPD`]: https://www.felixcloutier.com/x86/addpd
[fog]: https://www.agner.org/optimize/instruction_tables.pdf#page=229

stephanwlee:
> 1000 scalar time series each with 1000 scalars is on the order of 8 MB with f64s,

Good point. It was an unsubstantiated claim. One more thing I was considering is amount of memory commands you end up doing and wasting with JavaScript. Allocating a shared memory (the 8 MB of f64s is transferred from main to the worker process) only to be thrown away to create another data structure, only to be thrown out and turned into ThreeJS representation, which will be transformed and transferred to GRAM, etc...

> Also, in #4395, we changed the data provider gRPC from f64 to f32 for
scalar data to reflect that we really only guarantee preserving scalars
as single-precision (and since we never write double-precision scalars,
in either TF 1.x or TF 2.x). The change in this PR is fine with me; it’d
also be fine with me to switch back to f32 for the scalar data if you
desire.

Yeah, that is the reason I chose to use f32. I think I can keep everything f32 if we decide to encode timestamps in seconds but I don't want to encode details of scalars plugin to the generic charting library. I am sure there are other ways to transfer datetime in more compact manner, though, as you said is not that important.

wchargin:
Yep. All makes sense to me.

dstiers11 (approved):

PR #4477, opened 2020-12-15
line chart: add temporal scale

stephanwlee:
This change implements temporal scale using d3.scaleTime.

While it _can_ be a performance hotspot, transforming data in temporal
dimension is non-trivial so, for now, we decided to adopt d3.scaleTime
as is.


psybuzz (approved):
lgtm

Outside of this PR, I am a little concerned that we now have value formatting logic in 2 places:
`webapp/widgets/line_chart_v2/lib/formatter.ts`
`webapp/util/value_formatter.ts`

stephanwlee:
Thanks for raising it. Will look to defrag at the right time.

PR #4478, opened 2020-12-15
graph: simplify tests

wchargin:
Summary:
Now that the data provider code path has replaced the multiplexer code
path entirely, we can remove the test indirection in favor of more
simple straightline setup. This follows the pattern of tests for the
scalars plugin.

Test Plan:
That tests pass suffices; if the data weren’t getting set up properly,
tests would return the wrong data and fail.

wchargin-branch: graph-simplify-tests

### DIFF
--- a/tensorboard/plugins/graph/graphs_plugin_v2_test.py
+++ b/tensorboard/plugins/graph/graphs_plugin_v2_test.py
@@ -65,17 +65,17 @@ class GraphsPluginV2Test(
         self.assertEqual(mime_type, "text/x-protobuf")
         return text_format.Parse(graph_pbtxt, graph_pb2.GraphDef())
 
-    @graphs_plugin_test.with_runs(
-        [
-            graphs_plugin_test._RUN_WITH_GRAPH_WITH_METADATA,
-            graphs_plugin_test._RUN_WITHOUT_GRAPH_WITH_METADATA,
-        ]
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/graph/graphs_plugin_v2_test.py
```diff
@@ -65,17 +65,17 @@ def _get_graph(self, plugin, *args, **kwargs):
...
-        [
-            graphs_plugin_test._RUN_WITH_GRAPH_WITH_METADATA,
-            graphs_plugin_test._RUN_WITHOUT_GRAPH_WITH_METADATA,
-        ]
-    )
-    def test_info(self, plugin):
+    def test_info(self):
         raise self.skipTest(
             "TODO: enable this after tf-nightly writes a conceptual graph."
         )
```
stephanwlee, line 71: dang, this is an old TODO that I can remove.

stephanwlee (approved):

tensorboard/plugins/graph/graphs_plugin_v2_test.py
wchargin: Yep. :-) am considering improving these tests while I’m here, both to
cover more classes of data and to use the public APIs rather than the
internals. If so, I’ll poke at this.

PR #4482, opened 2020-12-16
Use `nonlocal` rather than `box[0]` in tests

wchargin:
Summary:
To set a value in an outer scope from within a nested function, we’ve
historically used the pattern of “boxing” the value in a length-1 list:

```python
count_box = [0]

def foo():
    count_box[0] += 1
    ...

foo()
assert count_box[0] == 1
```

Now that we use Python 3 everywhere, we can avoid the box and just use
the `nonlocal` keyword instead:

```python
count = 0

def foo():
    nonlocal count
    count += 1
    ...

foo()
assert count == 1
```

This patch fixes up uses of this pattern, all of which are in tests.

Part of #4488.

Test Plan:
Running `git grep -F '_box = ['` no longer matches any results.

wchargin-branch: box-to-nonlocal

### DIFF
--- a/tensorboard/auth_test.py
+++ b/tensorboard/auth_test.py
@@ -20,25 +20,27 @@ from tensorboard import test as tb_test
 
 class AuthContextTest(tb_test.TestCase):
     def test_cache_success(self):
-        call_count_box = [0]
+        call_count = 0
 
         class UsernameProvider(auth.AuthProvider):
             def authenticate(self, environ):
-                call_count_box[0] += 1
+                nonlocal call_count
+                call_count += 1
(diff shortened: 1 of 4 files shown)


nfelt (approved):
Oh python :/  Thanks for the cleanup.  If you wanna create an issue for "thank goodness we finally can stop caring about Python 2" and log cleanup like this to it you might as well get credit for it.

wchargin:
Sure. Credit attribution aside, that will be a nice place to keep a
to-do list of “changes that I want to make in low-traffic periods to
avoid merge conflicts”, like “remove all `__future__` imports”. #4488

PR #4483, opened 2020-12-16
line chart: temporal axis

stephanwlee:
This change treats time axis a bit specially and hardcodes logic for
rendering times. Time ticks, unlike simple numbers, can result in
massive confusion if we encode one to fit into 50px. For instance, if I
were representing 2020/12/31 12:31:12.442 as "12.442", you lose all the
important date, hour, and minute. We are definitely using `<title>` so
user can see more complete information but we lose the "glance-ability".

To mitigate the issue (not perfectly solving the issue), show a major
tick only for date which shows a bit higher level information (datetime)
when certain conditions are right. This business logic definitely is not
perfect but it is a great start and found it to be helpful.

I find special treatment for time scale encoded in line_chart_v2 a bit
unfortunate but it is the easiest way out for now.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
@@ -45,18 +45,10 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
       <ng-container *ngFor="let tick of getTicks()">
         <g>
           <text [attr.x]="textXPosition(tick)" [attr.y]="textYPosition(tick)">
-            {{
-              customFormatter
-                ? customFormatter.formatTick(tick)
-                : scale.defaultFormatter.formatTick(tick)
-            }}
+            {{ getFormatter().formatTick(tick) }}
           </text>
           <title>
-            {{
-              customFormatter
-                ? customFormatter.formatReadable(tick)
-                : scale.defaultFormatter.formatReadable(tick)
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -96,16 +139,35 @@ export class LineChartAxisComponent {
...
+      this.scale instanceof TemporalScale &&
+      this.axisExtent[1] - this.axisExtent[0] < DAY_IN_MS &&
+      minorTickCount > majorTickCount &&
+      this.getMajorTicks().length <= 2
     );
   }
 
-  getTickString(tick: number): string {
-    return numberFormatter.formatTick(tick);
+  getMajorTicks() {
```
psybuzz, line 163: The `: number[]` return type annotation on L167 is nice for me, who often forgets d3 function signatures.  Can we add it here, too?

psybuzz (approved):
I tested out the PR locally and set the X axis type to "Wall" time, and observed no visual diff.  Just double checking: a future change will actually add styles to the `.major` class and hook things up?

stephanwlee:
> I tested out the PR locally and set the X axis type to "Wall" time, and observed no visual diff. Just double checking: a future change will actually add styles to the .major class and hook things up?

You are right. It is not wired up as it introduces noises to already medium sized change. I have a quick follow up that I will add you to.

PR #4484, opened 2020-12-16
Identify which tensor is too large, when dropping from upload

ericdnielsen:
When tensor are dropped for being too long, the end-user has no way of identifying which tensor was dropped.  The user needs to know in case it is a critical tensor to their visualizations.  And we'd like to get concrete reports from users about if our current limits work or are too restrictive, so we need users able to say which tensors they needed, and which ones were ok to skip.

Adding additional data to the log message.
Before: "Tensor too large; skipping.  Size NNNNN exceeds limit of 16384 byte."
After "Tensor (name, step: NNN) too large ...."

The line wraps,  it wrapped before though.

Ran a --dry-run before and after

None, simple change.

### DIFF
--- a/tensorboard/plugins/text/text_demo.py
+++ b/tensorboard/plugins/text/text_demo.py
@@ -24,7 +24,7 @@ tf.compat.v1.enable_v2_behavior()
 LOGDIR = "/tmp/text_demo"
 
 # Number of steps for which to write data.
-STEPS = 16
+STEPS = 160
 
 
 def simple_example(step):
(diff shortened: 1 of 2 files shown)


bileschi (approved):
Thanks!   So it tells you the name of the tag, and the step, but it does not print out the (very big) tensor?   SGTM.

ericdnielsen:
Yes, sample output (thought I had included this in the PR description)

W1216 11:18:32.671410 140383102875456 uploader.py:797] Tensor (higher_order_tensors/multiplication_table, step: 159) too large; skipping. Size 154751 exceeds limit of 16384 bytes.

wchargin (commented):
Perhaps consider also including the run? We generally try to surface
(run, tag) pairs, since tag and step are not sufficient to uniquely
determine the point.

ericdnielsen:
I don't think we have access to run_name at this point, without plumbing it through a series of calls.  I'll see if there's a clean way to add it.

ericdnielsen:
Only had to add a single extra argument

W1216 11:51:21.001875 140322372941632 uploader.py:798] Tensor (run:run2, tag:higher_order_tensors/multiplication_table, step: 159) too large; skipping. Size 154751 exceeds limit of 16384 bytes.

ericdnielsen:
Need to fix the unit tests now with the extra parameter.

Issue #4422, opened 2020-12-03
RustBoard tasks


wchargin:
## Feature tasks

If you’re interested, let me know, and I’ll assess whether it’s a good idea and
provide guidance if appropriate.

### Small-ish

  - [x] @bileschi: Plugin listing (#4487)
      - touches: server
      - straightforward; good starter task
  - [x] @wchargin: Data location and experiment metadata (#4751)
      - touches: server, Python data provider, RPC protos
      - pretty straightforward; good starter task
  - [x] @wchargin: Make CRC checking optional (#4573)
      - touches: core, CLI
      - straightforward; good starter task
  - [x] @nfelt: Online commits (commit data periodically while runs are loading) (#4604)
      - touches: core only
      - fairly straightforward; good starter task
  - [x] @psybuzz: Samples-per-plugin (#4689)
      - touches: core, CLI, Python ingester
      - probably pretty straightforward? could be a good starter task
      - [x] optional extra: support unbounded reservoirs (#4754)

### Involved

  - [x] @nfelt: Tensors core (#4726)
      - touches: core, server, Python data provider
  - [x] @nfelt: Tensors data compat
      - [x] histograms (#4740)
      - [x] PR curves (#4734)
      - [x] hparams (#4735)
      - [x] text (#4734)
  - [x] @wchargin: Blob sequences (#4563, et al.)
      - touches: core, server, Python data provider
      - needs generic blob sequence support plus compat for images,
        audio, graphs
      - expect some moderately tricky stuff on the server since
        `ReadBlob` is a server-streaming RPC
  - [x] @wchargin: Blob data compat
      - [x] images (#4564)
      - [x] audio (#4566)
      - [x] graphs (#4544)
      - [x] graphs sub-plugins (Keras, etc.) (#4568, #4569)
  - [x] @wchargin: Parallelize across runs (#4634)
      - touches: core only
      - expect to think about lifetimes
      - consider ~[`crossbeam`] for scoped threads and maybe [`SegQueue`]~ [`rayon`]!
  - [ ] Optimize run-tag filters over small input sets
      - i.e.: iterate over the filter, not the map
      - touches: server only
      - will be harder than it sounds :-)
      - mostly for learning purposes; will be faster, but won’t matter
        in practice
  - [ ] Reload-multifile
      - touches: core, CLI
      - note: current logic _does_ always read from all files; it just
        never marks old files as dead

[`crossbeam`]: https://docs.rs/crossbeam
[...]

PR #4487, opened 2020-12-16
rust: implement `ListPlugins` RPC

bileschi:
https://github.com/tensorflow/tensorboard/issues/4422 Plugin listing task

    * Implements plugin listing by iterating over all Runs / Tags in the commit and reducing to a set of pluginnames, which are moved into a list in the response.
    * Adds tests.

### DIFF
--- a/tensorboard/data/server/server.rs
+++ b/tensorboard/data/server/server.rs
@@ -373,7 +373,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_list_plugins_multiple_timeseries() {
+    async fn test_list_plugins_multiple_timeseries_same_type() {
         let commit = CommitBuilder::new()
             .scalars("train", "xent2", |b| b.build())
             .scalars("train", "xent", |b| b.build())
@@ -388,6 +388,9 @@ mod tests {
             vec!["scalars"]
         );
     }
+
+    // TODO(@bileschi): Add a test for the case of multiple different types of
+    // tags once sample CommitBuilders exist for plugins beyond "scalar".
 
     #[tokio::test]
     async fn test_list_plugins_no_data() {
--- a/tensorboard/data/server/server.rs
+++ b/tensorboard/data/server/server.rs
@@ -52,28 +52,29 @@ impl TensorBoardDataProvider for DataProviderHandler {
         &self,
         _request: Request<data::ListPluginsRequest>,
     ) -> Result<Response<data::ListPluginsResponse>, Status> {
-        let mut res: data::ListPluginsResponse = Default::default();
-        let runs = self.commit.runs.read().unwrap();
-        let mut plugin_names_set: HashSet<String> = HashSet::new();
-        for run_data in runs.values() {
-            let run_data = run_data.read().unwrap();
-            for scalar_value in run_data.scalars.values() {
-                let metadata: &pb::SummaryMetadata = scalar_value.metadata.as_ref();
-                let plugin_data = &metadata.plugin_data.as_ref();
-                match plugin_data {
-                    Some(d) => {
-                        plugin_names_set.insert(d.plugin_name.clone());
-                    }
-                    _ => {
-                        plugin_names_set.insert(String::new());
-                    }
-                }
+        let runs = self.read_runs()?;
+        // Collect set of plugin names.
+        let mut plugin_names = HashSet::new();
+        for (run, data) in runs.iter() {
+            let data = data
+                .read()
+                .map_err(|_| Status::internal(format!("failed to read run data for {:?}", run)))?;
+            for time_series in data.scalars.values() {
+                let metadata: &pb::SummaryMetadata = time_series.metadata.as_ref();
+                let plugin_name = match &metadata.plugin_data.as_ref() {
+                    Some(d) => d.plugin_name.clone(),
+                    None => String::new(),
+                };
+                plugin_names.insert(plugin_name);
             }
         }
-        // Move out of set into response's list.
-        for s in plugin_names_set {
-            res.plugins.push(data::Plugin { name: s });
-        }
+        // Move out of set into a new ListPluginsResponse.
+        let res = data::ListPluginsResponse {
+            plugins: plugin_names
+                .into_iter()
+                .map(|name| data::Plugin { name })
+                .collect(),
+        };
         Ok(Response::new(res))
     }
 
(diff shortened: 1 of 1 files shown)


tensorboard/data/server/server.rs
```diff
@@ -52,9 +53,27 @@ impl TensorBoardDataProvider for DataProviderHandler {
...
+                    Some(d) => {
+                        plugin_names_set.insert(d.plugin_name.clone());
+                    }
+                    _ => {
+                        plugin_names_set.insert(String::new());
+                    }
+                }
+            }
+        }
+        // Move out of set into response's list.
```
wchargin, line 73: Indeed—these strings are being moved *out* of the set, and ownership of
them given to the vector. You can tell statically that these are being
moved, not copied (no `.clone()`) or borrowed (types are `T`, not `&T`).

tensorboard/data/server/server.rs
```diff
@@ -350,6 +369,44 @@ mod tests {
...
+            experiment_id: "123".to_string(),
+        });
+        let res = handler.list_plugins(req).await.unwrap().into_inner();
+        let expected: Vec<String> = vec![];
+        assert_eq!(
+            res.plugins
+                .into_iter()
+                .map(|p| p.name)
+                .collect::<Vec<String>>(),
+            expected
```
wchargin, line 406: Can inline `expected` as `Vec::<String>::new()` (and use `collect::<Vec<_>>()` again).

tensorboard/data/server/server.rs
```diff
@@ -350,6 +369,44 @@ mod tests {
...
+        });
+        let res = handler.list_plugins(req).await.unwrap().into_inner();
+        assert_eq!(
+            res.plugins.into_iter().map(|p| p.name).collect::<Vec<_>>(),
+            vec!["scalars"]
+        );
+    }
+
+    // TODO(@bileschi): Add a test for the case of multiple different types of
+    // tags once sample CommitBuilders exist for plugins beyond "scalar".
```
wchargin, line 390: You can do this now. The `ScalarTimeSeriesBuilder` lets you set the
metadata of the time series, so you can construct a `SummaryMetadata`
with a different plugin name. See sibling methods for examples of
calling methods on the builder, but it should look something like:

```rust
let mut custom_metadata = pb::SummaryMetadata::default();
custom_metadata.plugin_name = "custom_scalars".to_string();
let commit = CommitBuilder::new()
    .scalars("train", "xent", |b| b.build())
    .scalars("train", "meta", |b| b.metadata(custom_metadata).build())
    .build();
```

I recommend including this test case now, because it raises a question
about the determinism of the output order, which we should address one
way or the other.

wchargin (commented):
Nice! This should work as is, except for the panics. Some improvements
suggested inline.

dstiers11 (approved):

tensorboard/data/server/server.rs
```diff
@@ -52,9 +53,27 @@ impl TensorBoardDataProvider for DataProviderHandler {
         _request: Request<data::ListPluginsRequest>,
     ) -> Result<Response<data::ListPluginsResponse>, Status> {
         let mut res: data::ListPluginsResponse = Default::default();
-        res.plugins.push(data::Plugin {
-            name: "scalars".to_string(),
-        });
+        let runs = self.commit.runs.read().unwrap();
```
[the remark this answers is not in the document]
bileschi: Switched to `self.read_runs()?`

tensorboard/data/server/server.rs
```diff
@@ -52,9 +53,27 @@ impl TensorBoardDataProvider for DataProviderHandler {
...
+                    _ => {
+                        plugin_names_set.insert(String::new());
+                    }
+                }
+            }
+        }
+        // Move out of set into response's list.
+        for s in plugin_names_set {
+            res.plugins.push(data::Plugin { name: s });
+        }
```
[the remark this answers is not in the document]
bileschi: Done (last version)

tensorboard/data/server/server.rs
```diff
@@ -52,9 +53,27 @@ impl TensorBoardDataProvider for DataProviderHandler {
         _request: Request<data::ListPluginsRequest>,
     ) -> Result<Response<data::ListPluginsResponse>, Status> {
         let mut res: data::ListPluginsResponse = Default::default();
-        res.plugins.push(data::Plugin {
-            name: "scalars".to_string(),
-        });
+        let runs = self.commit.runs.read().unwrap();
+        let mut plugin_names_set: HashSet<String> = HashSet::new();
```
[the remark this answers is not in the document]
bileschi: Removed.  And read up on turbofish.   TMYK 🌈

tensorboard/data/server/server.rs
```diff
@@ -52,9 +53,27 @@ impl TensorBoardDataProvider for DataProviderHandler {
...
+        for run_data in runs.values() {
+            let run_data = run_data.read().unwrap();
+            for scalar_value in run_data.scalars.values() {
```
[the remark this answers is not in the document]
bileschi: Done

tensorboard/data/server/server.rs
```diff
@@ -386,8 +387,32 @@ mod tests {
...
+        });
+        let res = handler.list_plugins(req).await.unwrap().into_inner();
+        assert_eq!(
+            res.plugins
+                .into_iter()
+                .map(|p| p.name)
+                .collect::<Vec<_>>()
+                .sort(),
+            vec!["custom_scalars", "scalars"].sort()
+        );
```
wchargin, line 414: `sort()` returns unit, not the vector itself (same as Python):

```rust
fn main() {
    assert_eq!(vec![1, 2].sort(), vec![3, 4].sort()); // passes!
}
```

```python
assert [1, 2].sort() == [3, 4].sort()  # passes!
```

([Clippy catches this][unit-cmp], but it looks like we don’t run Clippy
on test code. I’ll fix that separately, since we have a few existing
Clippy warnings in tests.)

[unit-cmp]: https://rust-lang.github.io/rust-clippy/master/index.html#unit_cmp

You’ll want to sort the vectors and then compare them (as variables).

tensorboard/data/server/server.rs
```diff
@@ -52,28 +52,29 @@ impl TensorBoardDataProvider for DataProviderHandler {
...
+        let runs = self.read_runs()?;
+        // Collect set of plugin names.
+        let mut plugin_names = HashSet::new();
+        for (run, data) in runs.iter() {
+            let data = data
+                .read()
+                .map_err(|_| Status::internal(format!("failed to read run data for {:?}", run)))?;
+            for time_series in data.scalars.values() {
+                let metadata: &pb::SummaryMetadata = time_series.metadata.as_ref();
+                let plugin_name = match &metadata.plugin_data.as_ref() {
```
wchargin, line 64: Can just match on `&metadata.plugin_data`. Then the `d` in `Some(d)` is
of type `&PluginData`, so `d.plugin_name.clone()` is still fine.

(Could also just match on `metadata.plugin_data.as_ref()` without the
extra indirection, which amounts to the same thing in this case but is
longer.)

tensorboard/data/server/server.rs
wchargin: Not done?

tensorboard/data/server/server.rs
bileschi: updated to capture the sorted lists

tensorboard/data/server/server.rs
bileschi: done

wchargin (approved):

PR #4489, opened 2020-12-17
line chart: implement tooltip sort order

stephanwlee:
This change re-implements a feature of vz-line-chart, tooltip sorting.
We sort the tooltip based on few criterions:
- ascending/descending of `y`
- nearest: distance to the cursor
- default: just the order of the runs from... (unclear and it confuses
  people)

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -47,6 +47,7 @@ import {ScalarStepDatum} from '../../data_source';
 import {TooltipSort, XAxisType} from '../../types';
 import {
   ScalarCardDataSeries,
+  ScalarCardSeriesMetadata,
   ScalarCardSeriesMetadataMap,
 } from './scalar_card_types';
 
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -185,7 +193,51 @@ export class ScalarCardComponent {
...
+  ): ScalarTooltipDatum[] {
+    const scalarTooltipData = tooltipData.map((datum) => {
+      return {
+        ...datum,
+        metadata: {
+          ...datum.metadata,
+          closest: false,
+          distSqToCursor:
+            (datum.point.x - cursorLoc.x) ** 2 +
+            (datum.point.y - cursorLoc.y) ** 2,
```
wchargin, line 212: Optional: could use `Math.hypot` here, which you might find simpler? On
the perf side, you have an unneeded `fsqrt`, but in exchange all the
arithmetic is in native code. May or may not matter either way.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -1285,20 +1308,245 @@ describe('scalar card', () => {
...
+              aux: false,
+            },
+            {
+              x: 10000,
+              step: 10000,
+              y: 3,
+              value: 3,
+              wallTime: new Date('2021-01-01').getTime(),
+            }
+          ),
```
wchargin, line 1446: At some point we may want a set of test helpers to build these without
duplicating the same 51 lines in each function… but no action required
for this PR.

wchargin (approved):
> * default: just the order of the runs from... (unclear and it confuses
>   people)

:upside_down_face:

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
stephanwlee: Thanks, I actually forgot about existence of hypot :) Applied.

Issue #4488, opened 2020-12-16
Replace py2isms with py3


wchargin:
We have by now firmly and securely dropped Python 2 support. There
remain many contortions in our code that are vestiges of the
bicompatibility period. Most of these fall under “mostly harmless, but
should be cleaned up incrementally or eventually”.

Off the top of my head:

  - [x] Replace `u"xxx"` literals with just `"xxx"` strings (#4409)
  - [ ] Replace `box = [None]` closure access pattern with `nonlocal` (#4482)
  - [ ] Remove now-useless `__future__` imports
  - [ ] Use `os.makedirs(exist_ok=True)` rather than `EEXIST` dances
  - [ ] Remove fallbacks for `sys.stderr.buffer`, et al.
  - [ ] Simplify conditional imports of `unittest.mock`
  - [ ] Kill the `unittest.mock` backport dependency

Feel free to add more.

PR #4493, opened 2020-12-17
py3: remove `argparse_util` helper

wchargin:
Summary:
In #2793, we added a helper to ensure that you can run `tensorboard`
either with or without an explicit subcommand. This works by default in
Python 3; the helper was only needed for Python 2. Thus, we remove it.

Part of #4488.

Test Plan:
The test plan from #2793 still passes. Instead of patching in a `demo`
subcommand, you can now use `tensorboard dev list [--json]`.

wchargin-branch: py3-no-argparseutil


nfelt (approved):
🧹 🧹 🧹 🎉

PR #4494, opened 2020-12-17
projector: read run names from data provider

wchargin:
Summary:
The projector plugin no longer depends on the event multiplexer. It
finds run directories by calling `list_runs` to get run names and
adjoining those to the value of the `--logdir` flag. This still only
works with a local logdir (as was also the case before this patch), but
now it works even when the data provider is not backed by a multiplexer.

The data provider interface does not mandate that run names be valid
relative paths under a logdir, but that does hold for the two data
providers that we care about: the multiplexer provider and the RustBoard
provider. Making the projector plugin truly portable will require much
deeper changes, so we don’t bother with anything more principled here.

This will certainly not work with `--logdir_spec`; the old version may
or may not have worked. The docs for `--logdir_spec` clearly disclaim
that it may cause some features to stop working, so I’m okay with this.

Test Plan:
Tested on a logdir whose `projector_config.pbtxt` is under plugin assets
and one whose config is in the main run directory. You can tell that
it’s properly reading the config because otherwise the sprite sheet
won’t show up. The projector plugin no longer contains any occurrences
of the word `multiplexer`, and it still works under `--load_fast`, which
does not create a multiplexer at all. Test sync also passes.

wchargin-branch: projector-data-provider-run-names


wchargin:
I’ve rolled back #4479 and rolled it forward as #4508, so this pull
request now depends on #4508.

tensorboard/plugins/projector/projector_plugin.py
```diff
@@ -524,7 +533,8 @@ def _append_plugin_asset_directories(self, run_path_pairs):
                 metadata.PLUGIN_ASSETS_NAME,
```
psybuzz, line 533: nit: maybe replace this with `plugin_assets_name`, since we've extracted it in L525?

psybuzz (approved):

tensorboard/plugins/projector/projector_plugin.py
wchargin: Ah, indeed. Thanks.

PR #4497, opened 2020-12-18
cleanup: remove dev print for graph

stephanwlee:
color histograms method spammed "no pairs found" onto console which
makes console quite unusable for a large graph. This change removes it.


psybuzz (approved):
lgtm

PR #4502, opened 2020-12-21
Add a lengths argument to tensorboard.summary.audio.

rryan:

Audio summaries have always assumed that examples in the dense
tensor shaped [batch_size, num_frames, num_channels] all have the same
length. This is frequently not the case when building audio models, such as
text-to-speech models, where the output length of the model varies per batch
item. When ragged tensors of substantially different shapes are batched
together, this can be wasteful. Additionally, the padding after the end of each
record is included in the resulting WAV file, which is misleading because these
samples were not present in the source signal. The consumer of the audio summary
has no way to determine what the original true lengths of each example were.

This change adds a lengths keyword argument to tensorboard.summary.audio, which
allows the caller to indicate the per-batch-item lengths of each signal in the
dense input tensor.

* Screenshots of UI changes

No UI changes -- adds keyword argument to tensorboard.summary.audio.

Run unit tests.

Support `tf.RaggedTensor` instead of adding a lengths argument. RaggedTensor is fairly new and adoption is low, so it would be nice to meet users where they are. I think RaggedTensor support would be a great future improvement.

### DIFF
--- a/tensorboard/plugins/audio/summary_test.py
+++ b/tensorboard/plugins/audio/summary_test.py
@@ -250,6 +250,39 @@ class SummaryV2OpTest(SummaryBaseTest, tf.test.TestCase):
             # Reset to default state for other tests.
             tf2.summary.experimental.set_step(None)
 
+    def test_lengths(self):
+        data = np.array([[[1.0], [0.0]], [[1.0], [-1.0]]]).astype(np.float32)
+        lengths = np.array([1, 2]).astype(np.int32)
+        event = self.audio_event("a", data, 44100, step=333, lengths=lengths)
+        self.assertEqual(event.step, 333)
+        self.assertLen(event.summary.value, 1)
+        tensor = tf.io.parse_tensor(
+            event.summary.value[0].tensor.SerializeToString(),
+            out_type=tf.string,
+        )
+
+        self.assertEqual(tensor.shape, [2, 2])
+        # If lengths are respected, the first example will be 2 bytes shorter
+        # than the second.
+        self.assertLen(tensor[0, 0].numpy(), 46)
+        self.assertLen(tensor[1, 0].numpy(), 48)
+
+    def test_no_lengths(self):
+        data = np.array([[[1.0], [0.0]], [[1.0], [-1.0]]]).astype(np.float32)
+        event = self.audio_event("a", data, 44100, step=333, lengths=None)
+        self.assertEqual(event.step, 333)
+        self.assertLen(event.summary.value, 1)
+        tensor = tf.io.parse_tensor(
+            event.summary.value[0].tensor.SerializeToString(),
+            out_type=tf.string,
+        )
+
+        self.assertEqual(tensor.shape, [2, 2])
+        # If no lengths are provided, both examples contain all samples in the
+        # dense tensor.
+        self.assertLen(tensor[0, 0].numpy(), 48)
+        self.assertLen(tensor[1, 0].numpy(), 48)
+
 
 if __name__ == "__main__":
     tf.test.main()
--- a/tensorboard/plugins/audio/summary_test.py
+++ b/tensorboard/plugins/audio/summary_test.py
@@ -251,37 +251,33 @@ class SummaryV2OpTest(SummaryBaseTest, tf.test.TestCase):
             tf2.summary.experimental.set_step(None)
 
     def test_lengths(self):
-        data = np.array([[[1.0], [0.0]], [[1.0], [-1.0]]]).astype(np.float32)
-        lengths = np.array([1, 2]).astype(np.int32)
-        event = self.audio_event("a", data, 44100, step=333, lengths=lengths)
-        self.assertEqual(event.step, 333)
-        self.assertLen(event.summary.value, 1)
-        tensor = tf.io.parse_tensor(
-            event.summary.value[0].tensor.SerializeToString(),
-            out_type=tf.string,
+        data = np.array(
+            [[[1.0], [0.0]], [[1.0], [-1.0]], [[0.0], [0.0]]]
+        ).astype(np.float32)
+        lengths = np.array([1, 2, 0]).astype(np.int32)
+        pb = self.audio(
+            "a", data, 44100, step=333, lengths=lengths, max_outputs=2
         )
+        results = tensor_util.make_ndarray(pb.value[0].tensor)
 
-        self.assertEqual(tensor.shape, [2, 2])
+        self.assertEqual(results.shape, (2, 2))
         # If lengths are respected, the first example will be 2 bytes shorter
         # than the second.
-        self.assertLen(tensor[0, 0].numpy(), 46)
-        self.assertLen(tensor[1, 0].numpy(), 48)
+        self.assertLen(results[0, 0], 46)
+        self.assertLen(results[1, 0], 48)
 
     def test_no_lengths(self):
-        data = np.array([[[1.0], [0.0]], [[1.0], [-1.0]]]).astype(np.float32)
-        event = self.audio_event("a", data, 44100, step=333, lengths=None)
-        self.assertEqual(event.step, 333)
-        self.assertLen(event.summary.value, 1)
-        tensor = tf.io.parse_tensor(
-            event.summary.value[0].tensor.SerializeToString(),
-            out_type=tf.string,
-        )
+        data = np.array(
+            [[[1.0], [0.0]], [[1.0], [-1.0]], [[0.0], [0.0]]]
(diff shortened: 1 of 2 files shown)


wchargin (commented):
Hi @rryan! Thanks for the patch, and good to see you again. We’ll
probably want to discuss this with some team members who are on vacation
right now. But I feel generally positive on this change (clear purpose,
no workaround, backward compatible, doesn’t affect serialized form).

rryan:
Happy new year! Gentle ping :) who would the most appropriate reviewer be?

wchargin:
Yes; thanks! I think @nfelt and I will both want to take a look at this.

tensorboard/plugins/audio/summary_v2.py
```diff
@@ -103,15 +107,31 @@ def lazy_tensor():
...
-                encode_fn,
-                limited_audio,
-                dtype=tf.string,
-                name="encode_each_audio",
-            )
+            if lengths is not None:
+                tf.debugging.assert_rank(lengths, 1)
+                limited_lengths = lengths[:max_outputs]
+
+                def encode_with_length(datum_length):
```
wchargin, line 118: nit: Rename to `datum_and_length`? I was confused on first read, parsing
this as “length of datum”.

tensorboard/plugins/audio/summary_test.py
```diff
@@ -250,6 +250,39 @@ def test_default_step(self):
...
             tf2.summary.experimental.set_step(None)
 
+    def test_lengths(self):
+        data = np.array([[[1.0], [0.0]], [[1.0], [-1.0]]]).astype(np.float32)
+        lengths = np.array([1, 2]).astype(np.int32)
+        event = self.audio_event("a", data, 44100, step=333, lengths=lengths)
+        self.assertEqual(event.step, 333)
+        self.assertLen(event.summary.value, 1)
+        tensor = tf.io.parse_tensor(
+            event.summary.value[0].tensor.SerializeToString(),
```
wchargin, line 260: Can we use `tensor_util.make_ndarray(pb.value[0].tensor)` instead? (See
other tests in this file for examples.)

tensorboard/plugins/audio/summary_test.py
```diff
@@ -250,6 +250,39 @@ def test_default_step(self):
             # Reset to default state for other tests.
             tf2.summary.experimental.set_step(None)
 
+    def test_lengths(self):
+        data = np.array([[[1.0], [0.0]], [[1.0], [-1.0]]]).astype(np.float32)
+        lengths = np.array([1, 2]).astype(np.int32)
+        event = self.audio_event("a", data, 44100, step=333, lengths=lengths)
```
wchargin, line 256: Maybe fine to just use `pb = self.audio(...)` here? We don’t really care
about hte step or wall time (the tests that use `audio_event` are
specifically testing that behavior), and this makes it slightly easier
to pluck out the tensor proto.

tensorboard/plugins/audio/summary_v2.py
```diff
@@ -103,15 +107,31 @@ def lazy_tensor():
...
             encode_fn = functools.partial(
                 audio_ops.encode_wav, sample_rate=sample_rate
             )
-            encoded_audio = tf.map_fn(
-                encode_fn,
-                limited_audio,
-                dtype=tf.string,
-                name="encode_each_audio",
-            )
+            if lengths is not None:
```
wchargin, line 114: All else equal, it would be nice to avoid introducing this new control
flow branch. One alternative could be to handle the `lengths is None`
case by synthesizing something like `lengths = int32_max * ones([k])`.
Something like this:

```diff
diff --git a/tensorboard/plugins/audio/summary_v2.py b/tensorboard/plugins/audio/summary_v2.py
index 0efd6e2e4..82cb4ba53 100644
--- a/tensorboard/plugins/audio/summary_v2.py
+++ b/tensorboard/plugins/audio/summary_v2.py
@@ -111,27 +111,22 @@ def audio(
             encode_fn = functools.partial(
                 audio_ops.encode_wav, sample_rate=sample_rate
             )
+            k = tf.shape(limited_audio)[0]
             if lengths is not None:
-                tf.debugging.assert_rank(lengths, 1)
-                limited_lengths = lengths[:max_outputs]
+                limited_lengths = lengths[:k]
+            else:
+                limited_lengths = tf.ones([k], dtype=tf.int64) * (2 ** 63 - 1)
 
-                def encode_with_length(datum_length):
-                    datum, length = datum_length
-                    return encode_fn(datum[:length])
+            def encode_with_length(datum_length):
+                datum, length = datum_length
+                return encode_fn(datum[:length])
 
-                encoded_audio = tf.map_fn(
-                    encode_with_length,
-                    (limited_audio, limited_lengths),
-                    dtype=tf.string,
-                    name="encode_each_audio",
-                )
-            else:
-                encoded_audio = tf.map_fn(
-                    encode_fn,
-                    limited_audio,
-                    dtype=tf.string,
-                    name="encode_each_audio",
-                )
+            encoded_audio = tf.map_fn(
+                encode_with_length,
+                (limited_audio, limited_lengths),
+                dtype=tf.string,
+                name="encode_each_audio",
+            )
             # Workaround for map_fn returning float dtype for an empty elems input.
             encoded_audio = tf.cond(
                 tf.shape(input=encoded_audio)[0] > 0,
```

What do you think?

tensorboard/plugins/audio/summary_test.py
```diff
@@ -250,6 +250,39 @@ def test_default_step(self):
             # Reset to default state for other tests.
             tf2.summary.experimental.set_step(None)
 
+    def test_lengths(self):
```
wchargin, line 253: Could you add a test to cover the case where `max_outputs` actually
limits the batch size? (i.e., `len(lengths) > max_outputs`)

tensorboard/plugins/audio/summary_v2.py
rryan: I'm happy to go with that if you prefer -- the length-less codepath produces a smaller and slightly more efficient graph with only a little bit of increased Python complexity so that's why I did it this way.

tensorboard/plugins/audio/summary_test.py
rryan: Done.

tensorboard/plugins/audio/summary_test.py
rryan: Done.

tensorboard/plugins/audio/summary_v2.py
rryan: Done.

tensorboard/plugins/audio/summary_test.py
rryan: Done -- if you don't mind I just added max_outputs to the existing 2 test cases I added :).

rryan (commented):
Thanks for the review!

wchargin (commented):
Code looks good to me. Withholding approval until @nfelt has a chance to
voice any objections.

tensorboard/plugins/audio/summary_v2.py
```diff
@@ -65,6 +66,9 @@ def audio(
         default, so if you want "wav" in particular, set this explicitly.
       description: Optional long-form description for this summary, as a
         constant `str`. Markdown is supported. Defaults to empty.
+      lengths: Optional lengths `Tensor`, shaped `[k]`, where each entry
```
nfelt, line 69: Could you slightly expand the documentation here so it's a little more obvious why one would want to use this parameter and exactly what it does?  e.g. something like the following (added wording in italics)

> Optional lengths `Tensor`, shaped `[k]`, where each entry indicates the length of the k'th example in frames. _Useful when `data` should be interpreted as ragged rather than fixed-length; each example will be truncated accordingly prior to audio encoding._ k may be statically unknown (i.e., `None`).

I would suggest adding a note to `data` as well, something like:

> A `Tensor` representing audio data with shape `[k, t, c]`, where `k` is the number of audio clips, `t` is the number of frames, and `c` is the number of channels. _To pass batches of data where the frame count is variable, use the `lengths` parameter._ Elements should be floating-point values in `[-1.0, 1.0]`. Any of the dimensions may be statically unknown (i.e., `None`).

nfelt (approved):
LGTM, just asking for slightly expanded docs.  Thanks for the well-justified contribution.

tensorboard/plugins/audio/summary_v2.py
rryan: Good suggestions! Done.

wchargin (approved):
(explicit LGTM)

PR #4507, opened 2020-12-22
py3: remove `mock` backport

wchargin:
Summary:
We now always import `unittest.mock` directly, rather than conditionally
depending on the `@org_pythonhosted_mock` backport. This simplifies both
the Python code and the BUILD structure, which was a bit precarious.

Part of #4488.

Test Plan:
All tests pass here and in a test sync (<http://cl/348690501>).

wchargin-branch: py3-no-mock-backport


psybuzz (approved):
Nice to see so many deletions!

PR #4508, opened 2020-12-23
projector: fix `RunPaths` result aliasing

wchargin:
Summary:
This rolls forward #4479 (see that commit for context) with changes due
to Google-internal, non-TensorBoard code that monkey patches some of the
projector internals:

  - The `_run_paths` field is now updated only during `_update_configs`,
    to ensure that each change to `_run_paths` gives `_configs` the
    opportunity to update.
  - Configs/run paths are no longer updated in `__init__`, since that
    may be expensive and involve file access (even with a multiplexer).
  - The `_run_paths` field is now always a `dict` rather than `None`,
    fixing an existing bug wherein accessing `configs` on a plugin with
    no multiplexer would raise an `AttributeError`.

This reverts commit 4c9a699392809e2ec015698a32abe771931a0f67, and
therefore reinstates commit 638014e1c12e43aeb090e8b101e0cf41c518c650.

Test Plan:
Test plan from #4479 still passes, and a test sync now passes as well.
Also tested pointing the projector at a logdir with no runs but with
checkpoint data at root, as fixed in #3694; this worked in the original
version of this change (#4479), but it works now, too.

Test sync: <http://cl/350223680>.

wchargin-branch: projector-fix-aliasing

### DIFF
--- a/tensorboard/plugins/projector/projector_plugin.py
+++ b/tensorboard/plugins/projector/projector_plugin.py
@@ -242,9 +242,8 @@ class ProjectorPlugin(base_plugin.TBPlugin):
         self.multiplexer = context.multiplexer
         self.logdir = context.logdir
         self.readers = {}
-        self.run_paths = None
+        self._run_paths = None
         self._configs = {}
-        self.old_num_run_paths = None
         self.config_fpaths = None
         self.tensor_cache = LRUCache(_TENSOR_CACHE_CAPACITY)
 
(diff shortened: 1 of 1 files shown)


psybuzz (approved):
lgtm.  By the way, is there any link to the internal Google CL that has the failure that led to the revert, for reference?

wchargin:
Thanks! No submitted CL (since it was broken). Sent some internal
resources offline.

PR #4509, opened 2020-12-24
ts: pass explicit args to promise resolvers

wchargin:
Summary:
When creating a new promise in JavaScript, calling `resolve()` is
equivalent to calling `resolve(undefined)`. But [TypeScript 4.1 forbids
the implicit form][ts-relnotes]:

> Expected 1 arguments, but got 0. Did you forget to include 'void' in
> your type argument to 'Promise'?

This patch updates our code to explicitly patch `Promise<void>` where
appropriate, which for some reason suppresses the error, and to
explicitly pass `undefined` in places where `Promise<void>` is wrong.
Call sites found by manual `git grep -F 'resolve()'` and filtering out
matches for `Promise.resolve`, which may still be invoked without
argument.

[ts-relnotes]: https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#resolves-parameters-are-no-longer-optional-in-promises

Googlers, see <http://cl/348490334> for context.

Test Plan:
Spot-checked the scalars, images, and projector dashboards, including
with `?fastChart=true`.

wchargin-branch: ts-promise-resolve-explicit

### DIFF
--- a/tensorboard/components/tf_backend/requestManager.ts
+++ b/tensorboard/components/tf_backend/requestManager.ts
@@ -13,8 +13,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 interface ResolveReject {
-  resolve: Function;
-  reject: Function;
+  resolve: (x: unknown) => void;
+  reject: (x: unknown) => void;
 }
 
 /**
@@ -203,7 +203,7 @@ export class RequestManager {
       this._queue.length > 0
     ) {
       this._nActiveRequests++;
-      this._queue.pop().resolve();
+      this._queue.pop().resolve(undefined);
     }
   }
   /**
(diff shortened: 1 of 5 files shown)


wchargin:
cc @brad4d

stephanwlee (approved):
Would it be possible to prefer the "prettier" (subjective) version of this? 

```ts
// good
new Promise<void>((resolve) => {
    resolve();
})

// bad
new Promise((resolve) => {
    resolve();
})
```
https://www.typescriptlang.org/play?#code/HYUw7gBACgTg9gWwJYGcQB4BuckBMB8AFITCCnADaYgCUEAvPhAN4BQEHEp5VIhNAblYBfGq1ahIsRKj4kylanUYt2nbor6CRNIA

wchargin:
Thanks! Done, in some places (I don’t think it makes sense to do that
for `requestManager`, since it should usually resolve to a non-null
value). I hadn’t read that far in the release notes, and it’s pretty
surprising to me that this works (what type is being inferred for `T` in
`new Promise((resolve) => resolve())` if not `void`?), but given that it
does, I guess it’s okay with me.

stephanwlee (approved):

brad4d:
Thanks for the quick turnaround on this!

PR #4510, opened 2020-12-24
py3: excise `six`

wchargin:
Armed with a handful of eldritch macros and more patience than Python
deserves, we cast `six` out into the void.

This is a mixture of automated and manual changes. See individual
commits for details. Plan to merge with rebase.

Test sync passes: <http://cl/348866797>.

Part of #4488.

### DIFF
--- a/tensorboard/compat/tensorflow_stub/io/gfile.py
+++ b/tensorboard/compat/tensorflow_stub/io/gfile.py
@@ -24,7 +24,6 @@ from collections import namedtuple
 import glob as py_glob
 import io
 import os
-import six
 import sys
 import tempfile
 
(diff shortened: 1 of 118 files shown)


tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -158,7 +157,7 @@ def _write(self, filename, file_content, mode):
 
     def glob(self, filename):
         """Returns a list of files that match the given pattern(s)."""
-        if isinstance(filename, six.string_types):
+        if isinstance(filename, (str,)):
```
nfelt, line 160: It looks like in every\* place that we had used `six.string_types` we're just passing it into isinstance(), so it seems better to just replace it with `str` rather than `(str,)` since the latter is entirely vestigial?

\* The one exception being in a docstring, but in that case I think `str` would also be more intelligible than `(str,)`

nfelt (approved):
Thanks for doing this! 🧹🧹🧹  Those vim macros are indeed scary...

tensorboard/compat/tensorflow_stub/io/gfile.py
wchargin: Good point. Done.

PR #4512, opened 2020-12-28
rust: run Clippy on tests and fix existing errors

wchargin:
Summary:
We run Clippy to catch common errors, but by default Clippy only runs on
non-test code. We noticed this in #4487, when Clippy would have caught a
real error (test doing nothing due to `unit_cmp`). This patch adds the
`--tests` flag to our Clippy invocation and fixes all existing warnings.

Test Plan:
The first draft of this PR included only the workflow change, and thus
properly failed on CI. Once the fixes were included, CI passed.

wchargin-branch: rust-clippy-tests

### DIFF
--- a/tensorboard/data/server/lib.rs
+++ b/tensorboard/data/server/lib.rs
@@ -17,10 +17,6 @@ limitations under the License.
 
 #![allow(clippy::needless_update)] // https://github.com/rust-lang/rust-clippy/issues/6323
 
-pub fn zero(x: f32) -> bool {
-    x == 0.0
-}
-
 pub mod cli;
 pub mod commit;
 pub mod data_compat;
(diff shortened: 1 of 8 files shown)


tensorboard/data/server/lib.rs
```diff
@@ -17,6 +17,10 @@ limitations under the License.
 
 #![allow(clippy::needless_update)] // https://github.com/rust-lang/rust-clippy/issues/6323
 
+pub fn zero(x: f32) -> bool {
+    x == 0.0
+}
+
```
bileschi, line 23: What is the purpose of this `zero` function?

bileschi (approved):
Approved.  Confused about the zero impl.  Don't see anything in the lint about problems with `lib.rs` either?

PR #4513, opened 2020-12-28
rust: fix test warning on macOS

wchargin:
Summary:
We import `OsStr` but only use it in tests that don’t run on macOS,
which causes an `unused_imports` warning: harmless, but should be fixed.

Test Plan:
Note that the `build-data-server-pip` macOS CI job’s “Test” step emits a
warning before this patch but not after it.

wchargin-branch: rust-unused-import-macos


bileschi (approved):
Thanks for the clear description and instructions on how to find the (fixed) warning

PR #4524, opened 2021-01-08
infra: FE development environment

stephanwlee:
With this change, we introduce a new Bazel rule, `tf_dev_js_binary`. It
creates a JavaScript bundle like `tf_js_binary` but instead of being
smart about module dependencies, it simply concatenates all JavaScript
(e.g., compiled TypeScript) sources in the Bazel dependency graph.

This greatly reduces the time we spent for bundling resources and makes
development easier. For instance, making a template change to the run
selector takes <5s of incremental build time (tsc + concat + Vulcanization)
which used to take about 18s with tsc + rollup + Vulcanization.

Do note that this change is imperfect. For instance, we really do not
have to restart the Python webserver when a web resources change as long
as we know to load the new version from the .runfiles. That change would
require a different strategy in loading an asset in the core_plugin
which is out of scope of this change. Another optimization opportunity
is to apply tf_dev_js_binary to every place where we use tf_js_binary.
While simple, since it adds noise, it is postponed to subsequent PRs.

Lastly, do note that this makes a bundle larger during development. For
example, index.js: 4.0 MiB gzipped (18.7 MiB uncompressed) to 5.3 MiB
(25.8 MiB uncompressed)

Before merging, we need to:
- [x] fix sync
- ~[ ] fix nightly releases to use the optimized mode (`-c opt`)~


stephanwlee:
@wchargin I am not too confident about the bash command I am using here. PTAL at it more carefully.

tensorboard/defs/internal/js.bzl
```diff
@@ -0,0 +1,146 @@
...
+
+    entry_point_module_name = ctx.workspace_name + "/" + _remove_ext(ctx.file.entry_point.short_path)
+
+    concat_command = ";".join(
+        [
+            "awk 'BEGINFILE {print \"// file: \"FILENAME}{print}' * " + " ".join([file.path for file in js_files]) + " >" + ctx.outputs.js.path,
+            # `require` the entry module name so it is evaluated.
+            "echo ';require([\"%s\"]);' >>" % entry_point_module_name + ctx.outputs.js.path,
+        ],
+    )
```
wchargin, line 83: Can we use shell arguments rather than manually interpolating (analogous
to using parameters in SQL queries)? Could write your command as:

```python
    concat_command = """
        output="$1" && shift
        entry_point="$1" && shift
        {
            awk 'BEGINFILE { print "// file: " FILENAME } { print }' "$@"
            printf ';require(["%s"]);\n' "${entry_point}"
        } >"${output}"
    """
    concat_args = (
        [ctx.outputs.js.path, entry_point_module_name]
        + [file.path for file in js_files]
    )
    ctx.actions.run_shell(..., arguments = concat_args)
```

tensorboard/defs/internal/js.bzl
```diff
@@ -0,0 +1,146 @@
...
+    # out.
+    js_files = [
+        f
+        for f in files.to_list()
+        if f.path.endswith(".js")
+    ]
+
+    ctx.actions.write(
+        output = ctx.outputs.manifest,
+        content = "\n".join([file.path for file in js_files]),
```
wchargin, line 71: This doesn’t emit a trailing newline; should it?

tensorboard/defs/internal/js.bzl
```diff
@@ -0,0 +1,146 @@
...
+        mnemonic = "ConcatJs",
+        progress_message = "concatenating JavaScript files from dependencies",
+        inputs = js_files,
+        outputs = [ctx.outputs.js],
+        command = concat_command,
+    )
+
+def _remove_ext(path):
+    ext_ind = path.rfind(".")
+    return path[:ext_ind]
```
wchargin, line 95: I don’t mind this; FWIW, skylib has `paths.replace_extension(path, "")`,
in case you prefer that.

tensorboard/defs/internal/js.bzl
```diff
@@ -0,0 +1,146 @@
...
+
+def _tf_dev_js_binary_impl(ctx):
+    files_depsets = []
+
+    bootstrap_and_deps = ctx.attr._ambient_deps + ctx.attr.deps
+    for dep in bootstrap_and_deps:
+        if JSNamedModuleInfo in dep:
+            # Collect UMD modules compiled by tf_ts_library
+            files_depsets.append(dep[JSNamedModuleInfo].sources)
+        elif not NpmPackageInfo in dep and hasattr(dep, "files"):
```
wchargin, line 26: nit: `NpmPackageInfo not in dep` (`not x in y` → `x not in y`)

tensorboard/defs/internal/js.bzl
```diff
@@ -0,0 +1,146 @@
...
+        module_name = ctx.attr._anonymous_umd_deps[target]
+        named_file = ctx.actions.declare_file(file.path + ".named_umd.js")
+
+        # Patch anonymous umd modules to have named in their declarations. For instance,
+        # it converts `define(['exports'], ...` to `define('d3', ['exports'], ...`.
+        # `define`'s argument behaves differently based on arity. For instance:
+        # 1: expects the argument to be a factory function to be invoked. Anonymous and
+        #    no dependency.
+        # 2: expects an array then a function. First arguments define dependencies to be
+        #    injected itno the factory. Anonymous with dependencies.
```
wchargin, line 41: sp.: s/itno/into/

wchargin (commented):
Interesting change. Thanks. Some comments included; will probably want
to think on it a bit more.

> I am not too confident about the bash command I am using here.

The `run_shell`, I presume? Done. (Your `cat $(SRCS) >$@` is good.)

tensorboard/defs/internal/js.bzl
stephanwlee: Omitted the manifest file generation. I was replicating closure rules but we really don't need it.

stephanwlee:
Thanks a lot for better looking command. It reads much better!

Please take your time thinking about the ramification. I think this is a stepping stone to better environment so I don't think this is _the_ environment. I also would not mind folding this change as I am happy to learn a bit more about Bazel-ism with this exercise.

wchargin (commented):
Code looks good to me. I’m mostly wondering what kinds of potential
failure modes this introduces. How likely do we think it is that there
might be bugs that only repro in `-c opt`, and how tricky would it be to
track that down?

Do we think that it would be useful to use a config setting other than
`-c opt` for this? I always build with `-c opt` so that native code is
faster (makes a huge difference internally; in OSS, it affects Rust but
maybe not much else). And if trying to track down or diagnose a bug that
might manifest differently in these environments, it would be nice to
not have to also rebuild all the native code (e.g., protoc). But I can
also see an argument that `-c fastbuild`/`-c opt` is exactly the
distinction being made and so we should use that. Thoughts?

(For usability, just put `--define=tf_use_concatjs=true` in your
bazelrc and you don’t have to think about this; you would want to do
this for `-c opt`, anyway.)

cc @psybuzz @japie1235813 @bmd3k @nfelt

stephanwlee:
> Code looks good to me. I’m mostly wondering what kinds of potential
failure modes this introduces. How likely do we think it is that there
might be bugs that only repro in -c opt, and how tricky would it be to
track that down?

Yes, that's the only concern I have. This change definitely is susceptible to the cases when there is compiler optimizer bug that only surfaces with `-c opt`. I do not have much mitigation handy but enforcing correctness at expense of bad developer productivity is not great either (teammates have expressed not willing to work in OSS because of the poor build toolchain).

>  I always build with -c opt so that native code is
faster (makes a huge difference internally; in OSS, it affects Rust but
maybe not much else)

FWIW, internal FE toolchain also relies on `-c opt` and uses different Closure compiler options; so I am very sure `-c opt` is not only limited to the native libraries/binaries. 

> (For usability, just put --define=tf_use_concatjs=true in your
bazelrc and you don’t have to think about this; you would want to do
this for -c opt, anyway.)
> ...if trying to track down or diagnose a bug that
might manifest differently in these environments

Back to an ability to detect a regression.

Regression would naturally be caused by a TypeScript change by someone who would most naturally work with `--define=tf_use_concatjs=true` in bazelrc (assuming that the person mainly touches web code). In that case, the developer would not be able to find a bug at the dev time especially given that the unit tests are ran with requirejs without rollup (or closure compiler), so whether you use `--define` vs. `-c`, it is not all that different. If possible, I would want to not introduce another `--define` which was recommended against by the Bazel team.

stephanwlee:
>> I always build with -c opt so that native code is
faster (makes a huge difference internally; in OSS, it affects Rust but
maybe not much else)
>
> FWIW, internal FE toolchain also relies on -c opt and uses different Closure compiler options; so I am very sure -c opt is not only limited to the native libraries/binaries.

@bmd3k am I correctly remembering js_library and js_binary? It might be the module set server thing.

nfelt:
Not that I have a huge stake in this, but isn't the pattern often that the devserver is just a separate build target entirely, rather than using a build flag (whether `--define` or `-c opt`)?  So basically `tf_dev_js_binary` makes sense to me on its own; the part that seems more questionable is using config settings to swap it in for `tf_js_binary`.

There would certainly be boilerplate we'd need to copy if we had a separate entry point entirely for running TB with the dev js binary UI (although maybe this is a good argument for attempting to separate out an FE server from the main python server, down the road), but I could see it being simpler overall to understand, and a bit more obvious that we expect that there may be functional differences / you might want to be running the prod version occasionally side-by-side with dev to check for regressions, versus using a build flag which (IMO) kind of suggests that we trust the compiler to produce an artifact with the same correctness / functional characteristics regardless of the flag value.

wchargin (approved):
Cool; thanks for the discussion. Approving for code, and I don’t object
to the approach. Feel free to let others weigh in if you want.

stephanwlee:
> , but isn't the pattern often that the devserver ... the part that seems more questionable is using config settings to swap it in for tf_js_binary.
>
> There would certainly be boilerplate we'd need to copy if we had a separate entry point entirely for running TB with the dev js binary UI (although maybe this is a good argument for attempting to separate out an FE server from the main python server,

I think that's a bit orthogonal and will be a natural evolution of the build tools. For instance, for a dev target, you can still pass `-c opt` and use different Closure compiler options internally (I agree that it is a bit confusing). To share where I am going, after this change, I plan on making the `tf_js_binary` even slower by adding terser which minifies/mangles and makes our sources/pip-package ~20% smaller (We can also remove sourcemaps in `tf_js_bianry` which is a big win, too). After that, depending on how many complains, I plan on adding dev_asset_middleware that removes need for `tensorboard_html_binary` and `webfiles.zip`, which for some reason strips away all the nice sourcemaps and boost developer productivity on the way. As you can guess, these changes require a bit more than just Bazel configurations and require us to have different targets. Do note that, even in this case, we will have discrepancy between dev vs. prod environments and, without good e2e tests, we will cause regressions.

nfelt:
Okay, thanks - but I'm a bit confused still.  If we think the natural evolution of the build tools will be towards separate targets anyway (so we can have dev_asset_middleware), then what I was trying to suggest is that maybe we should just use separate targets here?  I.e. we'd have two different tensorboard binaries (one with prod FE, one with dev FE) rather than depending on any config setting to automatically switch the same binary between using `tf_js_binary` versus `tf_dev_js_binary`.

stephanwlee:
And I was arguing that we can do that incrementally.

psybuzz:
Sorry for the late reply on this thread.  I like the approach, and I am so looking forward to going from >12s to ~5s incremental builds!

> it would be nice to not have to also rebuild all the native code (e.g., protoc). But I can also see an argument that -c fastbuild/-c opt is exactly the distinction being made and so we should use that

This is a good point.  In normal development, I don't imagine we'll need to switch back and forth too often, so the cost of rebuilding all the other native code might be fine.  I see that other JS-related Bazel rules like https://bazelbuild.github.io/rules_nodejs/Rollup.html will change depending on `-c opt`, so it seems reasonable for us to do the same here.

One non-blocking question I have, is: after this change, do we need to add `-c opt` to our CI workflow and pip package scripts as well?  e.g. whenever CI or a publish script does `bazel test tensorboard/...` for sanity checking, do those need to be updated to `bazel test tensorboard/... -c opt` now?

stephanwlee:
> One non-blocking question I have, is: after this change, do we need to add -c opt to our CI workflow and pip package scripts as well?

Yes, that is what the TODO is about.

> e.g. whenever CI or a publish script does bazel test tensorboard/... for sanity checking, do those need to be updated to bazel test tensorboard/... -c opt now?

No, not really, at least for now. `bazel test` for frontend does not leverage `tf_js_binary` at all and there is no tests against the optimized code today. I am still assessing whether we absolutely need to do that for other types of correctness though.

nfelt:
> And I was arguing that we can do that incrementally and even in that future world, we _can_ have a different behavior with `-c opt`.

Okay... I guess in a world where we end up with separate prod vs dev server targets, it seems kind of like a detour / somewhat confusing to have previously been using `-c opt` to switch between prod mode vs dev mode (assuming that once we have separate targets we would want prod vs dev mode to be controlled by the target, rather than by `-c opt`, so that they are orthogonal rather than interdependent).  But it sounds like you have a plan for how you want this to work so I won't object.

stephanwlee (commented):
Ok, now that we have the `//tensorboard:dev` target, I have removed the `config_setting` and `select`. FE behavior should not change at all with `-c opt`. 

Since it has changed a bit since last reviewed, PTAL.

wchargin (approved):
Nice! With `//tensorboard`, my end-to-end time from “make trivial change
to (non-Polymer) TypeScript file” to “orange header bar renders in
refreshed page” is 21 seconds; with `//tensorboard:dev`, 11 seconds.
Certainly noticeable.

tensorboard/webapp/bootstrap.ts
```diff
@@ -20,6 +20,13 @@ import {AppModule} from './app_module';
...
+  document.readyState === 'interactive' ||
+  document.readyState === 'complete'
+) {
   platformBrowser().bootstrapModule(AppModule);
-});
+} else {
+  window.addEventListener('DOMContentLoaded', () => {
+    platformBrowser().bootstrapModule(AppModule);
+  });
+}
```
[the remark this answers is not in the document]
stephanwlee: I cannot be certain but current working theory is that requirejs (for `:dev`) has internal mechanism before invoking the entry point. For reference, Vulcanize/JSComp flattens entire script (without IIFE) and does not have any special bootstrapping logic.

Issue #4313, opened 2020-11-10
Projector plugin has triple scrollbar(!!!) on Firefox


wchargin:
The projector plugin’s “no data” screen has three vertical scrollbars in
Firefox:

![Screenshot of the projector plugin’s “No checkpoint was found” screen
in Chrome and Firefox, with three consecutive vertical scrollbars in
Firefox][ss]

The main view (i.e., with data) has two vertical scrollbars. There
shouldn’t be any in either case.

[ss]: https://user-images.githubusercontent.com/4317806/98718440-4f96dc80-2343-11eb-88b9-895f78282ee1.png

PR #4525, opened 2021-01-08
Fix unnecessary scroller bar in firefox and chrome

japie1235813:
https://github.com/tensorflow/tensorboard/issues/4313
Unnecessary scrollers in Chrome and firefox 

1. Replace margin with padding for the content of projector text block to prevent the body height exceeds the iframe
2. The script tag ahead of nested iframe creates a small extra height for firefox. Apply line-heigh 0 to the body 

* Screenshots of UI changes
* Chrome
chrome
Firefox
firefox

### DIFF
--- a/tensorboard/plugins/projector/tf_projector_plugin/index.js
+++ b/tensorboard/plugins/projector/tf_projector_plugin/index.js
@@ -19,6 +19,7 @@ html,
 body,
 iframe {
   border: 0;
+  display: block;
   height: 100%;
   margin: 0;
   width: 100%;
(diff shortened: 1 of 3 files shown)


stephanwlee:
On my mac, when I enable scrollbar always, I still see one scrollbar.
image

image

Instead of `line-height: 0;`, I think more correct thing to do is to convert `iframe` into `display: block` (default is inline it seems, at least on FF). 

WDYT?

tensorboard/plugins/projector/tf_projector_plugin/index.js
```diff
@@ -22,6 +22,9 @@ iframe {
   height: 100%;
   margin: 0;
   width: 100%;
+}
+body {
+  line-height: 0;
```
stephanwlee, line 27: I think this should be avoided if possible and is something that can be replaced with `display: block`.

tensorboard/plugins/projector/tf_projector_plugin/index.js
```diff
@@ -22,6 +22,7 @@ iframe {
   height: 100%;
   margin: 0;
   width: 100%;
+  display: block;
```
stephanwlee, line 25: nit: Not a rule but prefer to a-z sort if not inconvenient. Here and plugins_component.ts

stephanwlee (approved):

PR #4526, opened 2021-01-08
refactor: tweak graph const. measurement

stephanwlee:
Polymer's property setter invokes binding and causes ripple of changes.
Previously, with new Hierarchy() measurement, we were measuring time it
takes to propagate `renderHierarchy` changes which internally redoes
layout and rendering. As such, `new Hierarchy` was measuring,
"tf-graph-scene (layout)" and "tf-graph-scene (build scene)".

### DIFF
--- a/tensorboard/plugins/graph/tf_graph/tf-graph.ts
+++ b/tensorboard/plugins/graph/tf_graph/tf-graph.ts
@@ -249,9 +249,13 @@ class TfGraph extends LegacyElementMixin(PolymerElement) {
     );
   }
   _buildRenderHierarchy(graphHierarchy) {
-    tf_graph_util.time(
+    // Certain Polymer property setter are dynamically generated and is not properly
+    // typed.
+    const anyThis = this as any;
+
+    const renderGraph = tf_graph_util.time(
       'new tf_graph_render.Hierarchy',
-      function () {
+       () => {
         if (graphHierarchy.root.type !== tf_graph.NodeType.META) {
           // root must be metanode but sometimes Polymer's dom-if has not
           // remove tf-graph element yet in <tf-node-info>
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/graph/tf_graph/tf-graph.ts
```diff
@@ -298,12 +302,12 @@ class TfGraph extends LegacyElementMixin(PolymerElement) {
...
-        this.async(function () {
-          this.fire('rendered');
-        });
-      }.bind(this)
+
+        return renderGraph;
+      }
     );
+
+    anyThis._setRenderHierarchy(renderGraph);
```
psybuzz, line 310: Before, `_buildRenderHierarchy` was a no-op when `graphHierarchy.root.type !== tf_graph.NodeType.META`.  Now, we will end up calling `anyThis._setRenderHierarchy(undefined)`.  Is this change to logic intentional and safe?

Or does it make sense to pull our the non-meta-root-node check to the very top of `_buildRenderHierarchy` and do an early return?

tensorboard/plugins/graph/tf_graph/tf-graph.ts
```diff
@@ -298,12 +302,12 @@ class TfGraph extends LegacyElementMixin(PolymerElement) {
             };
           }),
         });
-        this._setRenderHierarchy(renderGraph);
-        this.async(function () {
-          this.fire('rendered');
-        });
```
psybuzz, line 304: I found no places where we listen for a 'rendered' event or have `on-rendered` in a template, so I'm assuming this was just unused.

psybuzz (approved):
lgtm % comment about non-root-meta-node.

tensorboard/plugins/graph/tf_graph/tf-graph.ts
stephanwlee: I think so. I cannot find any place where we use it.

PR #4528, opened 2021-01-09
infra: better type Map

stephanwlee:
When you create a Map without values, TypeScript has to infer the type
and, sometimes, it is imperfect and types cannot unsafe. This change
forces valueless Map constructor to specify the types so we rely less on
the inference.

To emulate the internal enforcement, we are making a necessary changes
to `tf_ts_library` but do note that it is imperfect and sometimes misses the
cases. Also do note that it somehow leaks its way into targets with
`strict_check = False` which is unintended but bearable.

Do note that this requires sync changes.
Test CL with necessary change patched: cl/350867815

### DIFF
--- a/tensorboard/defs/strict_type_check.d.ts
+++ b/tensorboard/defs/strict_type_check.d.ts
@@ -1,0 +1,20 @@
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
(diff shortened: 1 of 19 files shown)


tensorboard/defs/strict_type_check.d.ts
```diff
@@ -0,0 +1,20 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+// Type override of lib.es2015.collection.d.ts.
+// Make sure value is not inferred and take anything.
```
psybuzz, line 17: nit: 'Make sure key/value types are not inferred as `any`.'?

psybuzz (commented):
Nice!  Internally, is the introduction of this new rule something that we are driving ourselves?

stephanwlee:
No, we are not driving it. For context, PTAL cl/350606198.

psybuzz (approved):
Thanks for the internal CL context, this PR lgtm

PR #4531, opened 2021-01-11
infra: suffix index.js and worker url with md5

stephanwlee:
TensorBoard resources, especially JavaScripts, are not tiny yet they are
not being cached at all by clients. As such, users are downloading large
amount of content every time browser refreshes and it impacts:
- above the fold time load latency
- server utilization

This change mitigates above two problems by giving us an ability to
leverage caching by creating a unique resource url with md5 hash of the
file.

Do note that browsers treat, for example, `index.js` differently than
`index.js?random_query`.

Internal change: cl/351171360
Test CL: cl/351209032


tensorboard/webapp/BUILD
```diff
@@ -207,13 +207,27 @@ tf_web_library(
 
 # A Vulcanized html binary for the complete app (both Angular and Polymer parts)
 tensorboard_html_binary(
-    name = "index",
+    name = "index_binary",
     input_path = "/index_polymer3.inlined.html",
     js_path = "/index.js",
-    output_path = "/index.html",
+    output_path = "/index_binary.html",
```
wchargin, line 213: Maybe give this a name that more clearly indicates the difference
compard to `index.html` (which *isn’t* that one contains more or less
binary content)? e.g., `/index.no_resource_digests.html` following our
existing patterns, but I’m open to suggestions.

tensorboard/defs/defs.bzl
```diff
@@ -149,3 +149,29 @@ def tf_ng_module(assets = [], **kwargs):
         angular_assets = assets,
         **kwargs
     )
+
+def tf_res_digest_suffixer(name, resources, template, out):
```
wchargin, line 153: Can we use the full word `resource` in this name?

I’m not in love with `tf_resource_digest_suffixer`, because most rules
are named based on what the output is (a binary, a library, a test)
rather than a “«verb»er”. But I don’t immediately have a suggestion, so
no action required.

tensorboard/defs/defs.bzl
```diff
@@ -149,3 +149,29 @@ def tf_ng_module(assets = [], **kwargs):
...
+
+        <script src="index.js?_file_md5=486c31fc"></script>
+    """
+
+    native.genrule(
+        name = name,
+        srcs = [template] + resources,
+        outs = [out],
+        cmd = "$(execpath //tensorboard/defs:res_digest_suffixer) $(SRCS) > $@",
+        tools = [
```
wchargin, line 174: Should be `exec_tools`, as this tool is meant to be run on the [platform
on which Bazel actions run, not the one on which Bazel itself runs.][1]
Cf. #4031.

[1]: https://docs.bazel.build/versions/master/platforms.html

tensorboard/defs/defs.bzl
```diff
@@ -149,3 +149,29 @@ def tf_ng_module(assets = [], **kwargs):
         angular_assets = assets,
         **kwargs
     )
+
+def tf_res_digest_suffixer(name, resources, template, out):
+    """Query parameter suffixer for a resource in a template.
```
wchargin, line 154: Cool. I’m okay with a new binary and macro for this. The more we have,
the more I consider defining something like a custom Jinja context with
filters for inlining base64 images, computing file digests, transcluding
files, etc. That way we could have fewer ad hoc genrules everywhere.

tensorboard/BUILD
```diff
@@ -317,6 +317,7 @@ tf_web_library(
         "//tensorboard/webapp/widgets/line_chart_v2/lib/worker:chart_worker.js",
     ],
     path = "/",
+    suppress = ["*"],
```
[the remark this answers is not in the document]
stephanwlee: Webfiles checker is incapable of understanding the URL and matches filename against a path with query parameter. Because we are not using Closure Compiler, this is effectively only suppressing the Webfiles checker which I think is good enough.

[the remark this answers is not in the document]
wchargin: Hmm, okay. It seemed strange to me because I would have guessed that if
the webfiles validator can’t resolve query strings, whatever actually
does the webfile merging will also fail. And if we’re not doing webfile
merging, then I don’t see why the webfiles validator is even being run.
But I’ll defer to you if you think that this is okay.

Might we perhaps at least change to `suppress = ["strictDependencies"]`?

tensorboard/defs/defs.bzl
```diff
@@ -150,28 +150,45 @@ def tf_ng_module(assets = [], **kwargs):
...
+    Args:
+        name: Name of the rule.
+        resources: dict of replacement keyword to resource File.
+        template: File that should be replaced
+        out: Name of the output file.
     """
 
+    srcs = [template]
+    args = ["$(location %s)" % template]
+    for res_template_name in resources:
```
wchargin, line 177: `for (template_name, label) in resources.items()`?

tensorboard/defs/defs.bzl
```diff
@@ -150,28 +150,45 @@ def tf_ng_module(assets = [], **kwargs):
...
         <script src="index.js"></script>
 
     For resources of ["index.js"], it will be replaced with:
 
-        <script src="index.js?_file_md5=486c31fc"></script>
+        <script src="index.js?_file_hash=486c31fc"></script>
+
+    Args:
+        name: Name of the rule.
+        resources: dict of replacement keyword to resource File.
```
wchargin, line 170: Writing `File` here seems a bit confusing, since [`File`] is a Bazel
concept but these values are just strings (this is a macro, not a rule).
Maybe “file” or “label”?

[`File`]: https://docs.bazel.build/versions/master/skylark/lib/File.html

tensorboard/defs/resource_digest_suffixer_test.py
```diff
@@ -0,0 +1,117 @@
...
+            f.write(template)
+
+        return full_path
+
+    def tearDown(self):
+        try:
+            shutil.rmtree(self.get_temp_dir())
+        except OSError:
+            # Some tests delete the directory.
+            pass
```
wchargin, line 40: We can remove this, I think? The point of `get_temp_dir()` is that it’s
automatically cleaned up.

wchargin (approved):
Lovely; thanks!

stephanwlee:
Moved the macro to `tensorboard/defs/internal/resource.bzl`.
Rationale: having everything stuffed into defs.bzl or large bzl file makes it harder to do internal transformations so let's break .bzl down?

tensorboard/defs/internal/resource.bzl
```diff
@@ -0,0 +1,57 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""rules/macros for manipulating resources."""
+
+def tf_resource_digest_suffixer(name, resources, template, out):
+    """Query parameter suffixer for a resource in a template.
+
+    In order to facilitate resource caching, the macro suffixes resource declaration in
+    a template file by replacing it with one with query parameter,
+    "?_file_hash=[len_8_truncated_md5]".
```
wchargin, line 21: s/md5/hash/

wchargin (commented):
I don’t entirely understand the internal/external structure—why not move
the contents of `defs/internal/resource.bzl` to `defs/js.bzl`?—but no
major objection.

stephanwlee:
Any folder with word `internal` is regarded as an internal set of rules and it disallows other packages like What-If tool to reach in. `defs/js.bzl` re-exports only what is supposed to be public but in this case, I am not even sure if I want this to be a public rule/macro. Did that address your question?

wchargin:
> Any folder with word `internal` is regarded as an internal set of rules and it disallows other packages like What-If tool to reach in.

Interesting; I didn’t know this. I don’t entirely understand, since we
load `@io_bazel_rules_webtesting//web/internal:platform_http_file.bzl`
in our `third_party/workspace.bzl`. Isn’t that an external package?

I also don’t quite get why it’s valuable to declare this macro in an
internal `.bzl` file given that we re-export it from `js.bzl`. Wouldn’t
the What-If Tool be able to access it through `js.bzl` anyway?

Agreed that it would be nice to avoid random Google packages taking a
dependency on the macro, but my understanding was that [this isn’t
possible][bzlvis]. That said, since the expanded macro depends on
`:resource_digest_suffixer`, does it suffice to limit the visibility of
that binary to `//tensorboard:internal` (which you’ve done)?

[bzlvis]: https://docs.bazel.build/versions/master/visibility.html#visibility-of-bzl-files

stephanwlee:
> Interesting; I didn’t know this. I don’t entirely understand, since we
load @io_bazel_rules_webtesting//web/internal:platform_http_file.bzl
in our third_party/workspace.bzl. Isn’t that an external package?

Technically, it is not a violation at Bazel level but buildozer would scream foul. If you disregard those lint warnings, I guess it has no powers. In other words, it is not Barlog proof.

wchargin:
I see. In any case, like I said, no major objection; please don’t feel
blocked on my behalf.

PR #4532, opened 2021-01-11
timeseries: stamp scalar only when first visible

stephanwlee:
In the new gpu line chart, we have an option to use Worker process to
render the line chart but its initialization can be heavy as it requires
to download the worker resources (or even parse script and create a
process). To make the AFT better, we only stamp and create line chart
component when it is actually visible on the screen.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_selectors.ts
+++ b/tensorboard/webapp/metrics/store/metrics_selectors.ts
@@ -140,11 +140,18 @@ export const getCardMetadata = createSelector(
   }
 );
 
-// A cheap identity selector to skip recomputing selectors.
-const getVisibleCardIdSet = createSelector(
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/metrics/store/metrics_selectors.ts
```diff
@@ -140,14 +140,21 @@ export const getCardMetadata = createSelector(
...
   }
 );
 
+export const getVisibleCardIdSet = createSelector(
+  selectVisibleCardIdSet,
+  (cardIdSet): Set<CardId> => {
+    return cardIdSet;
+  }
+);
+
```
psybuzz, line 157: Hm, if we expose the Set, it's probably simpler if we got rid of the Array variant, and avoid keeping 2 selectors that do almost the same thing.  wdyt?

psybuzz (approved):

PR #4534, opened 2021-01-11
bug: fix requests to absolute paths under path_prefix

psybuzz:
Some frontend components, including the Time Series dashboard, make
HTTP requests using absolute paths, which fail under --path_prefix, since
the TensorBoard application may not be served under '/'.

This change amends TBHttpClient to add the 'app root' as a prefix when
the requested resource is an absolute path. Internal schemes of the form
'/experiment/some_dynamic_id/...' (e.g. TB.dev) should be compatible, if
the 'app root' is '/' in those cases.

Manually tested with/without `--path_prefix=/foo/bar` that the Time Series
dashboard loads data. Pointing the logdir at npmi demo data did not load
anything in the dashboard, so I was not able to confirm that it works as
intended.

Note that Debugger V2's code viewer still fails to load Monaco under path
prefix. It does not use TBHttpClient for fetching Monaco assets.

No sync changes should be required.  Googlers, see test sync cl/351684147.

### DIFF
--- a/tensorboard/webapp/app_routing/BUILD
+++ b/tensorboard/webapp/app_routing/BUILD
@@ -41,6 +41,7 @@ tf_ng_module(
     name = "app_root",
     srcs = [
         "app_root.ts",
+        "app_root_module.ts",
     ],
     deps = [
         ":location",
--- a/tensorboard/webapp/app_routing/app_root_module.ts
+++ b/tensorboard/webapp/app_routing/app_root_module.ts
@@ -1,4 +1,4 @@
-/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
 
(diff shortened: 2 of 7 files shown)


psybuzz:
Sorry for thrashing, I just realized this approach (having Time Series request `data/plugins/...`) won't work.  '/compare' routes really do need to fetch tags for different experiments using the expId in the URL.  I'm thinking we might just want to add the AppRoot in Time Series' OSS requests.

psybuzz:
Ready for a review.

I think `loadMonaco()` should be addressed in a separate PR.  Internally, the `load_monaco_shim.ts` is forked, so a possible route to address it would be PRs/CLs to introduce
```ts
class MonacoProvider{
  constructor(appRootProvider: AppRootProvider /* dep injected */)
  loadMonaco()
}
```
in both OSS and internal, followed by cutovers and changes to delete the old `loadMonaco()`.

stephanwlee (approved):

PR #4535, opened 2021-01-12
rust: build universal data server with no binary

### LINKED ISSUE #4533: Figure out optional platform-specific PyPI deps
We want the Python distribution package `tensorboard` to depend on
`tensorboard-data-server` so that `--load_fast` can be accessible by
default. But `tensorboard` is cross-platform (pure-Python) while the
data server has native deps. We want to still be able to install Python
on platforms without a compatible data server package.

For example, in JavaScript-land, if you run `yarn add chokidar` (a
high-level file watcher interface) on Linux, you see:

```
info fsevents@2.3.1: The platform "linux" is incompatible with this module.
info "fsevents@2.3.1" is an optional dependency and failed compatibility check. Excluding it from installation.
```

…because `fsevents` is macOS-only.

What we _don’t_ want is for `pip install tensorboard` to suddenly fail
on Windows because we don’t have a Windows wheel.

Probably this can be achieved with something like “provide a `bdist`
wheel for each platform and also a generic `bdist` wheel”, or “also
provide an `sdist` wheel”. We just have to be careful to get the
precedence right.

A fallback option is to include binaries for all platforms in a single
wheel and determine which one to use (if any) at runtime. Might be
simpler (might not, since we can’t say “build one wheel one each
platform in CI”), but requires shipping a few extra megs of binaries to
all machines.

Falls under “Pip packaging” for #4422.

wchargin:
Summary:
The `tensorboard-data-server` package now supports a universal build
configuration. When built in this configuration, the package does not
actually contain a binary, and returns `None` when interrogated for its
path. But it can be installed on any platform, so it serves as a
fallback. Thus, `tensorboard` can depend on `tensorboard-data-server`
while maintaining install-time compatibility on all platforms.

Closes #4533; see that issue for details.

Test Plan:
Running `build.py` with `--universal` successfully creates a Pip package
called `tensorboard_data_server-0.1.0a0-py3-none-any.whl` that contains
no binary. Running TensorBoard with `--load_fast` prints the new error
message and exits. When re-installing the native wheel, TensorBoard can
still find it. The existing codepaths for a bundled server and a server
taken from environment variable also still work.

Uploaded the three wheels built in CI to Test PyPI as `0.1.0a20210111`,
in the order (1) macOS (2) universal (3) Linux, waiting a minute or so
in between each. When installing a patched TensorBoard wheel with a new
`tensorboard-data-server` dep, it behaved as expected per #4533.

wchargin-branch: rust-universal-pip

### DIFF
--- a/tensorboard/data/server/pip_package/build.py
+++ b/tensorboard/data/server/pip_package/build.py
@@ -37,18 +37,27 @@ import tempfile
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument(
+        "--out-dir",
+        help="Output directory into which to place the wheel; should already exist",
+        required=True,
+    )
+    binary_group = parser.add_mutually_exclusive_group(required=True)
+    binary_group.add_argument(
         "--server-binary",
         help="Path to the //tensorboard/data/server binary to bundle into the package",
     )
-    parser.add_argument(
-        "--out-dir",
-        help="Output directory into which to place the wheel; should already exist",
+    binary_group.add_argument(
+        "--universal",
+        help="Build a universal package with no server",
+        action="store_true",
     )
     args = parser.parse_args()
 
-    server_binary = os.path.abspath(args.server_binary)
-    if not os.path.isfile(server_binary):
-        raise RuntimeError("No such file: %s" % server_binary)
+    server_binary = None
+    if not args.universal:
+        server_binary = os.path.abspath(args.server_binary)
+        if not os.path.isfile(server_binary):
+            raise RuntimeError("No such file: %s" % server_binary)
 
     TDS = "tensorboard_data_server"  # convenience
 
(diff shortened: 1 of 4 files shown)


wchargin:
@stephanwlee: Would you mind running

```
pip install --extra-index-url https://test.pypi.org/simple/ tensorboard-data-server &&
python -c 'print(__import__("tensorboard_data_server").server_binary())'
```

in a new virtualenv, and verifying that it prints a path, not `None`?
Just double-checking that it’s picking up the platform-specific wheel on
macOS as well.

tensorboard/data/server/pip_package/build.py
```diff
@@ -37,18 +37,27 @@
...
         "--server-binary",
         help="Path to the //tensorboard/data/server binary to bundle into the package",
     )
-    parser.add_argument(
-        "--out-dir",
-        help="Output directory into which to place the wheel; should already exist",
+    binary_group.add_argument(
+        "--universal",
+        help="Build a universal package with no server",
+        action="store_true",
```
stephanwlee, line 52: Is `store_true` a reserved keyword of setup tools or soemthing?

stephanwlee (commented):
The command returns a path on macos.

stephanwlee (approved):

tensorboard/data/server/pip_package/build.py
wchargin: `store_true` is `argparse`’s strange syntax for “this is a boolean flag
that does not take an argument; when the flag is passed, store `True` as
its value”. You can see it also in (e.g.) `--bind_all` in `core_plugin`.
It’s not related to `setuptools` specifically.

wchargin (commented):
> The command returns a path on macos.

Great; thank you for checking!

PR #4536, opened 2021-01-12
py3: remove `six` from requirements

wchargin:
Summary:
As of #4510, we don’t need `six`, so we can remove it from the Pip
package requirements.

Part of #4488.

Test Plan:
Running `:extract_pip_package` and installing it into a virtualenv still
installs `six` as a transitive dep, so this is actually no functional
change for now. Maybe one day.

wchargin-branch: py3-norequire-six


wchargin:
(Oops, forgot to request review.)

nfelt (approved):
You could say `six` has been... deep-sixed 🥁

PR #4540, opened 2021-01-13
line chart: update only visible charts

stephanwlee:
In the timeseries dashboard, we keep track of cards that are in the
view. With this information, we can opt out of updating the charts that
are outside of the viewport and save CPU cycles.

This change propagates the visibility information via `enableUpdate`
flag and make sure to apply changes only when the card is visible in the
viewport.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
@@ -109,7 +109,7 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
 
   /**
    * Optional parameter to tweak whether to propagate update to line chart implementation.
-   * When not specified, it defaults to `true`. When it is `false`, it remembers what has
+   * When not specified, it defaults to `false`. When it is `true`, it remembers what has
    * changed and applies the change when the update is enabled.
    */
   @Input()
--- a/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
@@ -108,6 +108,14 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
   tooltipTemplate?: TooltipTemplate;
 
   /**
+   * Optional parameter to tweak whether to propagate update to line chart implementation.
+   * When not specified, it defaults to `true`. When it is `false`, it remembers what has
+   * changed and applies the change when the update is enabled.
+   */
+  @Input()
+  enableUpdate?: boolean;
+
+  /**
    * Whether to ignore outlier when computing default viewBox from the dataSeries.
    *
    * Do note that we only take values in between approxmiately 5th to 95th percentiles.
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
```diff
@@ -107,6 +107,14 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
   @Input()
   tooltipTemplate?: TooltipTemplate;
 
+  /**
+   * Optional parameter to tweak whether to propagate update to line chart implementation.
+   * When not specified, it defaults to `true`. When it is `false`, it remembers what has
+   * changed and applies the change when the update is enabled.
+   */
+  @Input()
+  enableUpdate?: boolean;
```
psybuzz, line 116: Could we invert the meaning and have a property `disableUpdates` instead?  My default assumption when reading optional properties is that `enableXYZ = false` would be equivalent to not specifying a value.

Then L319 can be `if (!this.lineChart || this.disableUpdate) return;`

stephanwlee: I generally prefer positive APIs but I can see a case of `disableUpdate` here, too.

I do not exactly recall but I think there was ToTT or code style guide around this. It says something like how `disableUpdate=false` making you think more.

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
```diff
@@ -307,7 +316,7 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
    * Minimally and imperatively updates the chart library depending on prop changed.
    */
   private updateLineChart() {
-    if (!this.lineChart) return;
+    if (!this.lineChart || this.disableUpdate === true) return;
```
psybuzz, line 319: nit: Can we do a simpler `if (!this.lineChart || this.disableUpdate) return;`?

tensorboard/webapp/widgets/line_chart_v2/lib/drawable.ts
```diff
@@ -23,15 +23,26 @@ import {
...
-  private prevFrameCache = new Map<string, any>();
-  private currFrameCache = new Map<string, any>();
+type Cacheable = {};
 
-  getFromPreviousFrame(key: string): any {
-    return this.prevFrameCache.get(key);
+export interface RenderCachePublicApi {
+  getFromPreviousFrame(key: string): Cacheable | null;
+  setToCurrentFrame(key: string, value: Cacheable): void;
+}
```
psybuzz, line 31: Would it be cleaner to move this interface definition into `paint_brush.ts` or another common file, to avoid a circular dependency between `drawable.ts` and `paint_brush.ts`?

tensorboard/webapp/widgets/line_chart_v2/lib/paint_brush.ts
```diff
@@ -40,7 +38,9 @@ export class PaintBrush {
       polyline,
       paintOpt
     );
-    this.context.renderCache.setToCurrentFrame(cacheId, newCacheValue);
+    if (newCacheValue) {
```
psybuzz, line 41: Looking at paint_brush.ts's signatures alone, I'd expect that `setLine(/* args for valid line */)` followed by `setLine(/* args for null line */)` to result in null or no entry in the current frame map.  With this new condition, we are saying that if `setLine` fails to create a valid line, it will preserve the existing one.

Could you help me understand when why this change is needed?

stephanwlee:
Sorry about the confusion. I mistakenly added commit that appeared in a different PR that is already merged. Rebased and omitted and removed unnecessary noise.

psybuzz (approved):
Looks good!

PR #4544, opened 2021-01-13
rust: load graphs into memory as blob sequences

wchargin:
Summary:
The `Commit` structure now has a `blob_sequences` store in addition to
its `scalars` store. For now, we only actually load run graphs into
memory (since it’s trivial, and everything else is merely easy).

Test Plan:
Unit tests included. There are no user-facing changes to the RPC layer.
Loading an MNIST dataset with scalars, graphs, and images no longer
prints warnings about `graph_def`s, and still serves scalars properly.

wchargin-branch: rust-load-graphs

### DIFF
--- a/tensorboard/data/server/commit.rs
+++ b/tensorboard/data/server/commit.rs
@@ -56,6 +56,9 @@ pub struct RunData {
 
     /// Scalar time series for this run.
     pub scalars: TagStore<ScalarValue>,
+
+    /// Blob sequence time series for this run.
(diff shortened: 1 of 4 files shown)


wchargin:
@psybuzz: Friendly ping for #4544 and #4546? Let me know if you have any
questions; happy to work through them with you.

psybuzz (approved):

psybuzz:
Thanks for walking thru these PRs with me offline!

PR #4545, opened 2021-01-13
rust: define blob key encoding

wchargin:
Summary:
The data provider RPC protocol requires servers to emit “blob keys”,
opaque, URL-safe tokens that uniquely identify blobs. This patch defines
an in-memory `BlobKey` struct with encoding and decoding. The wire
format is URL-safe base64 (with no padding) over JSON.

The seemingly unrelated changes are needed because loading `serde_json`
adds trait implementations that make some type inference inadmissible:
<https://github.com/rust-lang/rust/issues/80964>

Test Plan:
Unit tests included.

wchargin-branch: rust-blob-keys


tensorboard/data/server/blob_key.rs
```diff
@@ -0,0 +1,145 @@
...
+            ParseBlobKeyError::BadBase64(_) => (),
+            other => panic!("expected BadBase64(_), got {:?}", other),
+        };
+    }
+
+    #[test]
+    fn test_bad_json() {
+        match "AAAAAA".parse::<BlobKey>().unwrap_err() {
+            ParseBlobKeyError::BadJson(_) => (),
+            other => panic!("expected BadBase64(_), got {:?}", other),
```
stephanwlee, line 142: `BadBase64` -> `BadJson`?

stephanwlee (approved):

tensorboard/data/server/blob_key.rs
wchargin: Indeed; thanks!

Issue #4547, opened 2021-01-13
Patch release: 2.4.1


wchargin:
Planned cherry-picks:

  - [x] #4423: ng: support path_prefix at router level (done in #4549)
  - [x] #4332: Remove `frame-ancestors *` CSP directive (done in #4550)

wchargin:
Shipped: <https://pypi.org/project/tensorboard/2.4.1/>

PR #4549, opened 2021-01-14
2.4: ng: support path_prefix at router level (#4423)

wchargin:
Backport of #4423 to 2.4. Cf. #4547.

---

Problem:
TensorBoard has a flag, path_prefix, which serves frontend and the data
endpoints at your specified, if the flag is passed, pathname. While
exact use case slips my mind, it is certainly used by dependents. Now,
the problem is that new Angular based TensorBoard has a custom router
for in-app navigation (we may add new routes for experiment list and
others) it is unaware of the path prefix.

Router works by matching set of predefined routes to current pathname
and finds a match. However, the configuration is unaware (rightfully) of
the path prefix and it results in zero match, which in turn cause
navigation to the default path, "/". This totally breaks the assumption
of the path prefix as all the request from the frontend will be sent to
"/data/..." as opposed to "[path_prefix]/data/..."

Remediation:
Previously, there was no reason to propagate the prefix information to
the frontend. Application only did hash changes and never required a
navigation. With this change, we now:

- server side render meta tag that depicts the relative path to the root of the
  application
- bootstrap Angular app from the relative root path and uses that information
  for matching routes and navigation

Tests:
- manually tested with/without path_prefix
- tested dynamic plugin with/without path_prefix (projector)

Co-authored-by: Stephan Lee <stephanwlee@gmail.com>


stephanwlee (approved):
You write beautiful code :P

image

wchargin:
And so fast, too!

PR #4550, opened 2021-01-14
2.4: Remove `frame-ancestors *` CSP directive (#4332)

wchargin:
Backport of #4332 to 2.4. Cf. #4547.

---

This PR essentially reverses https://github.com/tensorflow/tensorboard/pull/2797.

Currently this doesn't work because the `frame-ancestors *` directive prevents VS Code from framing TensorBoard. This is because VS Code is an Electron application, and Electron appears to be unable to frame websites which set `frame-ancestors *` in its response headers: https://github.com/electron/electron/issues/26369

If I'm reading the CSP specification correctly, omitting the frame-ancestors directive altogether is equivalent to setting `frame-ancestors *`, so to my knowledge this PR should not result in a behavior change for environments which correctly implement the CSP spec. From https://w3c.github.io/webappsec-csp/2/#directive-frame-ancestors:

> The term allowed frame ancestors refers to the result of parsing the frame-ancestors directive’s value as a source list. If a frame-ancestors directive is not explicitly included in the policy, then allowed frame ancestors is "*".

Co-authored-by: Joyce Er <joyceerhuiling@gmail.com>


wchargin:
CLAs are okay since this is just a backport of #4332, whose original
author has signed the CLA.

stephanwlee (approved):

PR #4551, opened 2021-01-14
line chart: set cache when object is returned

stephanwlee:
There was a bug where when we tried to remove a line while the data is
not yet fetched, the line chart threw a JSE for trying to access
`null.obj3d` in the remove object function. This happened since renderer
can, in some case, early exit by returning null and we populated the
cache with null values. With this change, we disallow adding null value
to the cache with better typing. We previously had the bug because it
was typed to take `any` which includes `null`. Now, we type it to extend
an object which is an okay constraint since it is an Object drawer.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/drawable.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/drawable.ts
@@ -23,15 +23,26 @@ import {
 import {PaintBrush} from './paint_brush';
 import {ObjectRenderer} from './renderer/renderer_types';
 
-class RenderCacheContainer {
-  private prevFrameCache = new Map<string, any>();
-  private currFrameCache = new Map<string, any>();
+type Cacheable = {};
 
-  getFromPreviousFrame(key: string): any {
-    return this.prevFrameCache.get(key);
+export interface RenderCachePublicApi {
+  getFromPreviousFrame(key: string): Cacheable | null;
+  setToCurrentFrame(key: string, value: Cacheable): void;
+}
+
+class RenderCacheContainer implements RenderCachePublicApi {
+  private prevFrameCache = new Map<string, Cacheable>();
+  private currFrameCache = new Map<string, Cacheable>();
+
+  getFromPreviousFrame(key: string): Cacheable | null {
+    const value = this.prevFrameCache.get(key);
+    return value ?? null;
   }
 
-  setToCurrentFrame(key: string, value: any) {
+  setToCurrentFrame(key: string, value: Cacheable) {
+    if (!value) {
+      console.log(key);
+    }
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/drawable.ts
```diff
@@ -23,15 +23,26 @@ import {
...
 import {ObjectRenderer} from './renderer/renderer_types';
 
-class RenderCacheContainer {
-  private prevFrameCache = new Map<string, any>();
-  private currFrameCache = new Map<string, any>();
+type Cacheable = {};
 
-  getFromPreviousFrame(key: string): any {
-    return this.prevFrameCache.get(key);
+export interface RenderCachePublicApi {
```
bmd3k, line 28: Maybe just call this "RenderCache". Or "RendererCache".

tensorboard/webapp/widgets/line_chart_v2/lib/paint_brush.ts
```diff
@@ -21,11 +22,8 @@ import {
...
-export interface PaintBrushContext<T = any> {
-  renderCache: {
-    getFromPreviousFrame(cacheKey: string): T;
-    setToCurrentFrame(cacheKey: string, obj: T): void;
-  };
+export interface PaintBrushContext {
+  renderCache: RenderCachePublicApi;
 }
 
 export class PaintBrush {
```
bmd3k, line 29: I think it would be slightly clearer if the constructor just took an object of type `RenderCachePublicAPI` (or, preferably `RenderCache` or `RendererCache`) and we got rid of the `PaintBrushContext` interface. It's not clear whey we need the extra level of indirection.

tensorboard/webapp/widgets/line_chart_v2/lib/drawable.ts
```diff
@@ -23,15 +23,26 @@ import {
...
+
+  getFromPreviousFrame(key: string): Cacheable | null {
+    const value = this.prevFrameCache.get(key);
+    return value ?? null;
   }
 
-  setToCurrentFrame(key: string, value: any) {
+  setToCurrentFrame(key: string, value: Cacheable) {
+    if (!value) {
+      console.log(key);
```
bmd3k, line 44: Did you intend to keep this here? If so, could you make the message a bit clearer?

bmd3k (approved):

tensorboard/webapp/widgets/line_chart_v2/lib/drawable.ts
stephanwlee: Nope, it was for my own debugging. Remove it. Thanks for catching it

PR #4555, opened 2021-01-14
line chart: improve color opacity interpolation

stephanwlee:
Previously, we were using d3.color.brighter but its algorithm created
a bit too salient color for certain colors (low luminance like our slate
color) while being light enough for others.

To alleviate this unevenness in brightening, we decided to use the color
interpolation instead. It results in more visually correct result.

Note: changed opacity from 0.4 to 0.25. It is 0.2 in vz-line-chart but
it looked a bit too faint in GPU line chart so we up-ed it by 0.05

Old:
image

New:
image


psybuzz (approved):
Thanks for the screenshots.  The slate aux lines definitely look better.

PR #4558, opened 2021-01-15
core: cache on JS resources when hash is present

stephanwlee:
Previously, TensorBoard never cached the resource even when its contents
never change in a given pip package. While this is tolerable in the
local instance setting, it is not the best especially when we start to
use worker where the resources would have to be fetched per worker
instance,

This change enables caching for one day (for now) when JavaScript
resources with `_file_hash` query parameter is requested. While this
approach works for any resources, there is no strong reason to do so and
we will start with restrictive caching at first. In the future, we may
apply the query parameter to more resources and enable caching of SVG
and other resources.

### DIFF
--- a/tensorboard/plugins/core/core_plugin_test.py
+++ b/tensorboard/plugins/core/core_plugin_test.py
@@ -377,9 +377,9 @@ class CorePluginTestBase(object):
             )
 
 
-class CorePluginResourceTEst(tf.test.TestCase):
+class CorePluginResourceTest(tf.test.TestCase):
     def setUp(self):
-        super(CorePluginResourceTEst, self).setUp()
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/core/core_plugin_test.py
```diff
@@ -374,6 +377,45 @@ def FirstEventTimestamp_stub(run_name):
             )
 
 
+class CorePluginResourceTEst(tf.test.TestCase):
```
wchargin, line 380: s/TEst/Test/

wchargin (approved):
Nice. Idea and implementation both seem good to me. Nice tests.

PR #4565, opened 2021-01-16
rust: pass summary metadata to `into_blob_sequence`

wchargin:
Summary:
For some data types, data-compat conversions of _values_ depend on the
time series _metadata_. For instance, we need to adjust the shapes of
audio tensors to remove labels, and we need to update graph sub-plugins
(Keras models, etc.) to convert scalars to tensors. This patch updates
the `EventValue::into_blob_sequence` interface to require a reference to
the initial summary metadata so that we can add support for such data
types. It’s not yet used.

Test Plan:
Existing unit tests suffice; for an end-to-end test the graphs and
images plugins still work.

wchargin-branch: rust-blob-sequence-metadata


tensorboard/data/server/run.rs
```diff
@@ -104,7 +104,7 @@ impl StageTimeSeries {
     fn commit(&mut self, tag: &Tag, run: &mut commit::RunData) {
         use pb::DataClass;
         match self.data_class {
-            DataClass::Scalar => self.commit_to(tag, &mut run.scalars, EventValue::into_scalar),
+            DataClass::Scalar => self.commit_to(tag, &mut run.scalars, |ev, _| ev.into_scalar()),
```
psybuzz, line 107: I'm not sure I understand why this had to change.  Are we expecting to modify this enrich function, in some way that doesn't also apply to L120's `EventValue::into_blob_sequence`?

psybuzz (approved):

tensorboard/data/server/run.rs
wchargin: It changes because of this line of the diff:

```diff
diff --git a/tensorboard/data/server/run.rs b/tensorboard/data/server/run.rs
--- a/tensorboard/data/server/run.rs
+++ b/tensorboard/data/server/run.rs
@@ -125,7 +125,7 @@ impl StageTimeSeries {
 
     /// Helper for `commit`: writes staged data for this time series into storage for a statically
     /// known data class.
-    fn commit_to<V, F: FnMut(EventValue) -> Result<V, commit::DataLoss>>(
+    fn commit_to<V, F: FnMut(EventValue, &pb::SummaryMetadata) -> Result<V, commit::DataLoss>>(
         &mut self,
         tag: &Tag,
         store: &mut commit::TagStore<V>,
```

The callback now takes `(EventValue, &pb::SummaryMetadata)` rather than
just `EventValue`. But `EventValue::into_scalar` is still a function of
only an `EventValue`. So this lambda drops the unused parameter.

Line 120 doesn’t change because `EventValue::into_blob_sequence` used to
take `EventValue` and now takes `(EventValue, &pb::SummaryMetadata)`,
just like the callback, so it typechecks both before and after.

wchargin (commented):
thanks for the quick reviews!

tensorboard/data/server/run.rs
psybuzz: Ah, thanks.  Yeah, both commit_to and into_* have changed, it makes sense

PR #4569, opened 2021-01-16
rust: support graph sub-plugins

wchargin:
Summary:
This patch adds support for data read by the graphs dashboard but
written under the following plugin names:

  - `graph_run_metadata`
  - `graph_run_metadata_graph`
  - `graph_keras_model`

(source of truth in `plugins/graph/metadata.py`).

Test Plan:
Unit tests included. As an end-to-end test, all the data written by the
graphs plugin demo now works with `--load_fast`.

wchargin-branch: rust-graph-subplugins


tensorboard/data/server/data_compat.rs
```diff
@@ -242,7 +263,11 @@ impl SummaryValue {
                     Some(SCALARS_PLUGIN_NAME) => {
                         md.data_class = pb::DataClass::Scalar.into();
                     }
-                    Some(IMAGES_PLUGIN_NAME) | Some(AUDIO_PLUGIN_NAME) => {
+                    Some(IMAGES_PLUGIN_NAME)
+                    | Some(AUDIO_PLUGIN_NAME)
+                    | Some(GRAPH_RUN_METADATA_PLUGIN_NAME)
+                    | Some(GRAPH_RUN_METADATA_WITH_GRAPH_PLUGIN_NAME)
+                    | Some(GRAPH_KERAS_MODEL_PLUGIN_NAME) => {
```
stephanwlee, line 270: Future plan question: are we going to trust the `data_class` property on the proto in the future? I believe the graph_run* and graph_keras* data are all annotated correctly with the data_class.

tensorboard/data/server/data_compat.rs
```diff
@@ -671,6 +696,48 @@ mod tests {
...
+                    pb::SummaryMetadata {
+                        plugin_data: Some(PluginData {
+                            plugin_name: plugin_name.to_string(),
+                            content: b"1".to_vec(),
+                            ..Default::default()
+                        }),
+                        data_class: pb::DataClass::BlobSequence.into(),
+                        ..Default::default()
+                    },
+                );
```
stephanwlee, line 734: No AI required: this test feels too mechanical :\

stephanwlee (approved):

tensorboard/data/server/data_compat.rs
wchargin: We already trust the data class: the first clause of this `match` is

```rust
        match (md, &*self.0) {
            // Any summary metadata that sets its own data class is expected to already be in the right
            // form.
            (Some(md), _) if md.data_class != i32::from(pb::DataClass::Unknown) => Box::new(md),
```

But I’m not sure if the graph data does set data classes? The data
written by `:graphs_demo` doesn’t seem to have them, and I don’t see
data classes in [the writing code][w]. In any case, data classes were
introduced more recently than the `graph_*` writing code, so old data
must not have them, I think.

So eventually I would probably like plugins to write summaries with data
class explicitly set, but I expect to maintain this compatibility
indefinitely, and that seems okay to me. WDYT?

[w]: https://github.com/tensorflow/tensorflow/blob/d7485ac5c3b7c76100fb5854af50b76322d8ec58/tensorflow/python/ops/summary_ops_v2.py#L1156-L1161

tensorboard/data/server/data_compat.rs
wchargin: Like, too much of “assert that `f(x)` equals the result of manually
computing `f(x)`”? I agree; it does. I think that I feel okay with it
since it is actually meaningful coverage—dropping a single line in the
`match` would cause parts of the graph plugin to silently not render.

Or maybe you meant that the test implementation spends a lot of lines on
proto manipulation? Agreed there, too. We could refactor to reuse proto
sub-structures, but I generally like optimizing for readability in tests
like this. In #4552, for example, I made the tests more explicit because
if they had been written like that in the first place, the bug would
likely have been caught in code review: `new Date(1610613017)` is an
obvious bug—too few digits—whereas `new Date(fakePoints1[0].x)` is not
obviously correct and just repeats the implementation code, which was
also wrong.

Noted that you said “no AI required”, so feel free to reply or not, as
you like. Glad that this passes your minimum bar, but also always
interested in doing better than that.

wchargin:
Merging, but happy to discuss more.

PR #4571, opened 2021-01-16
rust: upgrade to Tokio 1.0 ecosystem

wchargin:
Summary:
The `tokio` ecosystem recently released version 1.0, and as of today,
`prost` and `tonic` have both released compatible upgrades as well. This
patch upgrades our Rust toolchain to use the new versions. The only
breaking change affecting us just needs a 2-line change (`cli.rs`), and
Tokio 1.0 comes with a [stability guarantee] of no breaking changes for
a minimum of 3 years, plus long-term support for 5 years.

Performance-wise, we see about a 10% overall improvement in load times:
12% (±6%) on `mnist`, 8% (±3%) on `edge_cgan`. Prost in particular also
has new performance-oriented features for `bytes` deserialization, which
we can look into separately.

[stability guarantee]: https://tokio.rs/blog/2020-12-tokio-1-0#a-stability-guarantee

Test Plan:
Unit tests all still pass, and an end-to-end test shows that the scalars
dashboard still works with `--load_fast`.

wchargin-branch: rust-tokio-1.0


wchargin:
> Prost in particular also has new performance-oriented features for
> `bytes` deserialization, which we can look into separately.

I looked into using [`bytes::Bytes`] for all our `bytes` proto fields.
In benchmarks, it appears to make things 0%(±3%) faster. So I’ll leave
that alone for now.

[`bytes::Bytes`]: https://docs.rs/bytes/1.0.1/bytes/struct.Bytes.html

nfelt (approved):
Oops, sorry for overlooking this.

wchargin:
no problem! Nothing urgent. Thanks.

PR #4573, opened 2021-01-19
rust: make CRC checking optional

wchargin:
Summary:
Instead of computing checksums for every record, we can now compute
checksums only for records that fail proto parsing. This means that bit
flips in payloads like string fields may not be caught. But it comes
with a significant performance improvement:

  - 2.42× (±0.24×) faster for `mnist` (125 ms → 50 ms)
  - 1.20× (±0.03×) faster for `edge_cgan` (10.4 s → 8.7 s)
  - 1.23× (±0.01×) faster for `nndiv` (182 s → 148 s)

Both the main RustBoard binary and the `bench` binary now take
`--checksum`/`--no-checksum` flags, defaulting to `--no-checksum`.

Test Plan:
Unit tests extended. Benchmarked with:

```
hyperfine -w3 \
    './target/release/bench --logdir ~/testdata/edge_cgan/ {cksum}' \
    -L cksum --checksum,--no-checksum
```

wchargin-branch: rust-optional-crc

### DIFF
--- a/tensorboard/data/server/event_file.rs
+++ b/tensorboard/data/server/event_file.rs
@@ -32,6 +32,8 @@ pub struct EventFileReader<R> {
     last_wall_time: Option<f64>,
     /// Underlying record reader owned by this event file.
     reader: TfRecordReader<R>,
+    /// Whether to compute CRCs for records before parsing as protos.
+    checksum: bool,
 }
 
 /// Error returned by [`EventFileReader::read_event`].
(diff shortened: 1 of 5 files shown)


stephanwlee (approved):

tensorboard/data/server/event_file.rs
```diff
@@ -68,14 +70,30 @@ impl<R: Read> EventFileReader<R> {
...
         let record = self.reader.read_record()?;
-        record.checksum()?;
-        let event = Event::decode(&record.data[..])?;
+        let event = if self.checksum {
+            record.checksum()?;
+            Event::decode(&record.data[..])?
+        } else {
+            match Event::decode(&record.data[..]) {
+                Ok(proto) => proto,
+                Err(e) => {
```
stephanwlee, line 91: Nice. Checking checksum on error!

wchargin: Yep, I figured that it could come in handy, and it’s easy enough and
costs nothing in the happy path.

PR #4574, opened 2021-01-19
tensorboard-data-server 0.1.0

wchargin:
This PR bumps the data server version number from 0.1.0a0 to 0.1.0, and
then to 0.2.0a0. Plan: merge with rebase, then release from the data
server binaries at the first commit as built by GitHub CI.

Notes:

  - Rust versions use <https://semver.org>, so prerelease specifiers are
    like `0.2.0-alpha.0` rather than Python `0.2.0a0`.
  - We could use `version = clap::crate_version!()` in the Rust code to
    avoid having to duplicate across `Cargo.toml` and `cli.rs`. This
    works fine when building with Cargo. But `CARGO_PKG_VERSION` is not
    set when building with Bazel, so for now I’m doing the simple thing
    and duplicating the string manually.

Test Plan:
Note that `cargo run -- -h` and `bazel run . -- -h` both print the
correct version.

### DIFF
--- a/tensorboard/data/server/Cargo.lock
+++ b/tensorboard/data/server/Cargo.lock
@@ -756,7 +756,7 @@ dependencies = [
 
 [[package]]
 name = "rustboard"
-version = "0.1.0"
+version = "0.2.0-alpha.0"
 dependencies = [
  "async-stream 0.3.0",
  "async-trait",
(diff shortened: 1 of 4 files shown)


psybuzz (approved):
🎉

PR #4575, opened 2021-01-20
line chart: fix axis layout

stephanwlee:
bug: x-axis did not extend to the full width of the line chart.
fix: use width: 100% and height: 100% on the inner element.

Before
image

After
image

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
@@ -70,6 +70,7 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
   styles: [
     `
       :host {
+        contain: strict;
         display: flex;
         overflow: hidden;
       }
@@ -87,6 +88,12 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
       text {
         font-size: 11px;
         user-select: none;
+      }
+
+      .x-axis,
+      .y-axis {
+        height: 100%;
+        width: 100%;
       }
 
       .x-axis {


tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -70,6 +70,7 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
   styles: [
     `
       :host {
+        contain: strict;
```
bmd3k, line 73: TIL about the`contain` property.  Apparently not supported by Safari.

Are you using it strictly for performance gains or are you also relying on it to impact how the axis appears to the user?

stephanwlee: It is not strictly (no pun intended) related to the layout issue. It is just telling browser that layout, paint, and etc... will not bleed out of the container of this DOM and can be considered independently which tends to help with paints (i.e., your whole web page should not re-paint just because you modified one DOM) and layouts.

bmd3k: No, given that rendering performance is an important goal of your project, I wouldn't ask you to remove it.

I am just hoping to understand what this means for Safari users - whether we risk breaking layout for them with use of these properties.

bmd3k (approved):

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
stephanwlee: Got it. While `contain` _can_ break layouts, it often does not (if it does, we can enumerate out what we want to contain such as `paint`). I will do a functional test on Safari afterwards.

PR #4577, opened 2021-01-20
rust: plumb reload interval to data server

wchargin:
Summary:
The TensorBoard `--reload_interval` flag now affects the flags passed to
the data server. To faithfully interpret the flag, we also teach the
data server to accept `--reload once`, corresponding to what TensorBoard
calls `--reload_interval 0`. We introduce this separate syntax to keep
the interpretation of `--reload N` continuous in `N`.

Test Plan:
Run with `--load_fast --verbosity 0 --reload_interval N` for various
values of `N`. Note that with `--reload_interval 0`, the data server
prints info logs for only one loading cycle, whereas with positive
intervals it delays for the correct amount of time. Note that fractional
intervals don’t cause a crash (they truncate).

wchargin-branch: rust-plumb-reload-interval

### DIFF
--- a/tensorboard/data/server_ingester.py
+++ b/tensorboard/data/server_ingester.py
@@ -52,9 +52,10 @@ class ExistingServerDataIngester(ingester.DataIngester):
 class SubprocessServerDataIngester(ingester.DataIngester):
     """Start a new data server as a subprocess."""
 
-    def __init__(self, logdir):
+    def __init__(self, logdir, reload_interval):
         self._data_provider = None
         self._logdir = logdir
+        self._reload_interval = reload_interval
 
     @property
     def data_provider(self):
@@ -70,9 +71,15 @@ class SubprocessServerDataIngester(ingester.DataIngester):
         tmpdir = tempfile.TemporaryDirectory(prefix="tensorboard_data_server_")
         port_file_path = os.path.join(tmpdir.name, "port")
 
+        if self._reload_interval <= 0:
+            reload = "once"
+        else:
+            reload = str(int(self._reload_interval))
+
         args = [
             server_binary,
             "--logdir=%s" % (self._logdir,),
+            "--reload=%s" % reload,
             "--port=0",
             "--port-file=%s" % (port_file_path,),
             "--die-after-stdin",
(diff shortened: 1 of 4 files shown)


tensorboard/data/server_ingester.py
```diff
@@ -70,9 +71,15 @@ def start(self):
         tmpdir = tempfile.TemporaryDirectory(prefix="tensorboard_data_server_")
         port_file_path = os.path.join(tmpdir.name, "port")
 
+        if self._reload_interval <= 0:
```
psybuzz, line 74: For `reload_interval` values < 0, based on [1], I think we used to avoid loading/ingesting completely, which differs from `reload_interval` 0.  Should we preserve this old behavior?

[1] https://github.com/tensorflow/tensorboard/blob/3c1503df6a6c2130a96c60804007065bcccb6537/tensorboard/backend/event_processing/data_ingester.py#L92

psybuzz (commented):
Code and test look good, only the 1 comment

tensorboard/data/server_ingester.py
wchargin: Indeed (and it’s even documented). I considered it, but it didn’t seem
very useful to me to teach the data server to admit a mode that does not
serve data. The history shows that this was added in #1001 to accomodate
DB mode, but that kind of accommodation is now performed by the
ingestion system.

Maybe in a follow-up PR I can revert the special treatment for negative
reload intervals altogether, now that DB mode is gone.

wchargin: Sent #4583.

psybuzz: Thanks for referencing the original commit.  If we think it's a vestige of DB mode and has no other real utility, dropping support sounds good to me.

If we are dropping the negative interval logic in this PR, can we either add a TODO or file an issue on GH to track removing it in the other places?  I think it's also probably worth mentioning in the next release notes as well.

psybuzz (approved):

tensorboard/data/server_ingester.py
psybuzz: Wow that was fast :)

Since the followup PR is out, a TODO/issue isn't really needed.

wchargin: Yep. Merged.

PR #4579, opened 2021-01-20
diagnose: print data server info

wchargin:
Summary:
If `tensorboard-data-server` is installed, the diagnostic script will
now print the Python version, the binary path, and the data server
version (as determined by invoking the server binary with `--version`).

Test Plan:
Tested: with no data server package; with the universal wheel, which
provides the data server Python package but no actual binary; and with
the Linux wheel, which does have a server binary.

wchargin-branch: diagnose-data-server


psybuzz (approved):
lgtm!

PR #4580, opened 2021-01-20
introduce py_repl() rule for a convenient Bazel-aware python shell

nfelt:
This adds a `py_repl` build rule that makes it simpler to spin up an interactive Python shell that includes a specified set of Bazel deps, which is useful for ad-hoc exploration of a library interface.

In a conventional Python setup a regular old `python` would work, but with Bazel this is more of a pain to get the correct runfiles, including files potentially generated by Bazel itself (e.g. `_pb2.py` files generated from protos, or genrule-created files).  Using a `py_binary` solves that, but it's a little annoying to clutter the codebase with a nearly empty main.py, and you have to remember the right invocation of `PYTHONINSPECT=1` to have it drop you into the REPL.  This build rule essentially just prepackages all of that for you.

### DIFF
--- a/tensorboard/defs/py_repl.bzl
+++ b/tensorboard/defs/py_repl.bzl
@@ -1,0 +1,80 @@
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Defines `py_repl` rule for creating a `bazel run`-able python REPL."""
+
+# Minimal wrapper over ctx.actions.write to write a string to a file.
+# Simpler than the genrule equivalent since we don't need to escape.
+def _write_file_impl(ctx):
+    ctx.actions.write(ctx.outputs.out, ctx.attr.content)
+
+_write_file = rule(
+    implementation = _write_file_impl,
+    attrs = {
+        "out": attr.output(mandatory = True),
+        "content": attr.string(),
+    },
+)
+
+def py_repl(name, preamble = None, deps = None):
+    """Executable target that runs the python interpeter interactively.
+
+    This provides a convenient way to interactively explore Python library
+    code that must be run under Bazel (e.g. it depends on Bazel-generated
+    Python dependencies, so cannot be run using `python -i` alone). It is
+    effectively just a wrapper over defining a `py_binary` with your desired
+    library deps plus a main.py with a few "preamble" lines and then running
+    the resulting built binary using `PYTHONINSPECT=1` to drop you in a REPL.
+
+    Args:
+      name: the name of this target
+      preamble: list of strings definining lines of Python code
+        that should be executed when starting the interpreter
+      deps: py_library targets that should be available as dependencies
+        to import into the interpreter
+    """
+
+    # Print each line of the preamble before executing it.
+    full_preamble = "\n".join(
+        ["print(" + repr(line) + ")\n" + line for line in preamble],
+    )
+
+    _write_file(
+        name = name + "_py_gen",
+        out = name + ".py",
+        content = full_preamble,
+    )
+
+    native.py_binary(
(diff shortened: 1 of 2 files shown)


tensorboard/defs/py_repl.bzl
```diff
@@ -0,0 +1,80 @@
...
+    code that must be run under Bazel (e.g. it depends on Bazel-generated
+    Python dependencies, so cannot be run using `python -i` alone). It is
+    effectively just a wrapper over defining a `py_binary` with your desired
+    library deps plus a main.py with a few "preamble" lines and then running
+    the resulting built binary using `PYTHONINSPECT=1` to drop you in a REPL.
+
+    Args:
+      name: the name of this target
+      preamble: list of strings definining lines of Python code
+        that should be executed when starting the interpreter
```
wchargin, line 43: This suggests to me that I could include arbitrary blocks, like:

```python
py_repl(
    ...,
    preamble = [
        "for x in (1, 2, 3):",
        "    print(x)",
    ],
)
```

But, iiuc, this doesn’t quite work, since the full preamble ends up
being:

```python
print('for x in (1, 2, 3):')
for x in (1, 2, 3):
print('    print(x)')
    print(x)
```

…which has indentation errors. Similarly, any triple-quoted strings in
the preamble would have extra text interposed. Maybe at least document
this constraint?

tensorboard/defs/py_repl.bzl
```diff
@@ -0,0 +1,80 @@
...
+        name = name + "_py",
+        srcs = [name + ".py"],
+        main = name + ".py",
+        deps = deps,
+    )
+
+    _write_file(
+        name = name + "_sh_gen",
+        out = name + ".sh",
+        content = "PYTHONINSPECT=1 exec $1",
```
wchargin, line 69: TIL about `python -i` and `PYTHONINSPECT`. Nice.

wchargin (approved):
Do you think that it would be useful to have a `//tensorboard:repl` that
depends on `//tensorboard:lib`, for general-purpose proto debugging?
Would enable me to forget my `PYTHONPATH` hacks once and for all.

If so, could be nice to include it in this PR so that we have a smoke
test that the Bazel rule is well defined.

wchargin:
> Do you think that it would be useful to have a `//tensorboard:repl` that
> depends on `//tensorboard:lib`, for general-purpose proto debugging?

Having used a hacked-in version twice since writing this comment, and
having wanted it again this morning, I’ll +1 my own suggestion. :-)

tensorboard/defs/py_repl.bzl
```diff
@@ -0,0 +1,80 @@
...
+      name: the name of this target
+      preamble: list of strings definining lines of Python code
+        that should be executed when starting the interpreter
+      deps: py_library targets that should be available as dependencies
+        to import into the interpreter
+    """
+
+    # Print each line of the preamble before executing it.
+    full_preamble = "\n".join(
+        ["print(" + repr(line) + ")\n" + line for line in preamble],
```
wchargin, line 50: As written, this is “`NoneType` is not iterable” if `preamble = []` is
not given explicitly. Can we default to an empty list?

tensorboard/defs/py_repl.bzl
nfelt: Good point.  Changed it to be a list of statements instead, and use `exec()` to prevent statements from leaking out into the generated script.  PTAL?

tensorboard/defs/py_repl.bzl
nfelt: Done.

nfelt (commented):
> Do you think that it would be useful to have a `//tensorboard:repl` that
> depends on `//tensorboard:lib`, for general-purpose proto debugging?
> Would enable me to forget my `PYTHONPATH` hacks once and for all.
> 
> If so, could be nice to include it in this PR so that we have a smoke
> test that the Bazel rule is well defined.

Done.

tensorboard/defs/py_repl.bzl
```diff
@@ -0,0 +1,105 @@
...
+def _dedent(text):
+    """Like `textwrap.dedent()` but only supports space indents."""
+    indents = []
+    for line in text.splitlines():
+        stripped = line.lstrip(" ")
+        if not stripped:
+            continue
+        indents.append(len(line) - len(stripped))
+    if not indents:
+        return
```
wchargin, line 39: `return ""`, not just `return`?

wchargin (approved):

tensorboard/defs/py_repl.bzl
nfelt: Thanks - `return text` actually is what I intended when input is all whitespace.  (In this context the `rstrip()` on the way in makes it equivalent, but in the general case I think probably better to preserve.)

Issue #4581, opened 2021-01-20
Provide a TensorBoard-native summary writer API


nfelt:
This issue tracks work to provide, within TensorBoard proper, a summary "writer" API intended for end-users to log data for visualization in TensorBoard.  This is intended to be a lightweight way to log to TensorBoard without necessarily depending on TensorFlow, although it should also be usable by eager-mode TensorFlow users.

Note that this API isn't intended to replace `tf.summary` - the latter will continue to be the recommended approach for TensorFlow users who want to log data from within graphs or tf.functions.

Plan of work. Some of these steps may be done on-demand / as need arises.

- [ ] Define initial `Writer` object and `Output` interface with scalar support
- [ ] Port `tf.summary` summary types
    - [ ] Scalars 
    - [ ] Images
    - [ ] Audio
    - [ ] Histograms
    - [ ] Text
    - [ ] HParams
    - [ ] Graph
    - [ ] PR curves
    - [ ] Custom scalars
    - [ ] Mesh
    - [ ] Embedding
    - [ ] NPMI
- [ ] Add support for generic / extensible summary types
- [ ] Add support for tracking a step within the writer itself
- [ ] Implement other `Output` types
    - [ ] In-memory output
    - [ ] Logging to stdout
- [ ] Implement filesystem support for `DirectoryOutput`
    - [ ] Local filesystem
    - [ ] tf.io.gfile support
    - [ ] GCS support (without TF)
- [ ] Add API documentation
- [ ] Add demo (or retrofit existing demos)

jpauwels:
Isn't this what [TensorboardX](https://github.com/lanpa/tensorboardX) is trying to do as well? Might be more efficient if both efforts could be joined (if that works out licence-wise).

nfelt:
Thanks for asking.  It's a bit of a long story, but basically we are aware of TensorboardX and have collaborated in the past, and some of the code in the repo today was contributed by the author of TensorboardX - see #2065 for example.  The work described in this issue will be building on that foundation and essentially "closing the loop" relative to the initial foray in this direction from a few years ago.

[...]

PR #4582, opened 2021-01-20
summary: initial code for TensorBoard-native summary writer

nfelt:
This introduces a summary writer API, `tensorboard.summary.Writer`, to allow users to log data for visualization in TensorBoard without additional dependencies, as outlined in #4581.  This PR contains the bare-bones definition of the `Writer` object (with support only for scalars), the definition of the `Output` interface, and corresponding tests sufficient to show the API in use end-to-end.

Remaining work to be done is outlined in #4581.


tensorboard/summary/_output.py
```diff
@@ -0,0 +1,114 @@
...
+        summary_metadata = summary_pb2.SummaryMetadata(
+            plugin_data=summary_pb2.SummaryMetadata.PluginData(
+                plugin_name=plugin_name, content=tag_metadata
+            ),
+            summary_description=description,
+            data_class=summary_pb2.DataClass.DATA_CLASS_SCALAR,
+        )
+        event = event_pb2.Event(wall_time=wall_time, step=step)
+        event.summary.value.add(
+            tag=tag, simple_value=data, metadata=summary_metadata
```
wchargin, line 102: With `DATA_CLASS_SCALAR`, [time series are expected to have `tensor`
values rather than `simple_value`s][spec]. I think technically both
RustBoard and the Python reading logic will happen to accept your data,
but it would be nice to conform to spec. Should we change this to emit
tensors, or do you think that we should make the spec more permissive?

[spec]: https://github.com/tensorflow/tensorboard/blob/742781d94413533df799143772dd6fd4bdc5091a/tensorboard/compat/proto/summary.proto#L70-L73

wchargin (commented):
This looks pretty good to me! Mostly questions inline, and I’d like to
take another look later.

tensorboard/summary/BUILD
```diff
@@ -1,20 +1,81 @@
...
+        ":output",
+        "//tensorboard:expect_numpy_installed",
+    ],
+)
+
+py_test(
+    name = "writer_test",
+    srcs = ["writer_test.py"],
+    srcs_version = "PY3",
+    deps = [
```
wchargin, line 48: Do we want `tags = ["support_notf"]` on these?

wchargin (commented):
Yep, I continue to be happy with this.

nfelt (commented):
Thanks for the review! PTAL

nfelt:
So the test failure in `DirectoryOutputTest.test_flush_and_close` happens in no-TF testing because our stub of `tf.io.gfile` does unbuffered writes, so the first event we write to the writer appears immediately on disk prior to calling `flush()`, whereas the real implementation (I assume) ends up using standard Python binary I/O which is buffered by default.

We have a few options:

1. Allow the behavior to be divergent and just relax the test.

2. Change `EventFileWriter` / `_AsyncWriter` so that their built-in buffering applies to the first event added - it [currently doesn't](https://github.com/tensorflow/tensorboard/blob/475a659931a3678f8637f6ad71c43bc72dafc351/tensorboard/summary/writer/event_file_writer.py#L208).  This is an observable behavior change for no-TF users and plausibly could be considered a regression, though I don't think we make any guarantees about this.  I originally wrote the test assuming that this logic is what we were testing (rather than platform/filesystem-level buffering), not realize it doesn't apply to the first event.

3. Change our `tf.io.gfile` stub to better mimic the real implementation by buffering binary writes.  This is a little non-trivial since we'd have to determine whether it makes sense to implement the buffering ourselves in our `GFile` object (making it cross-platform in a way that I'm not sure it actually is in the real implementation; we might want to confirm), or try to reuse the native Python binary file buffering (making it platform-specific but possibly more accurate).  The latter option is complicated by the fact that right now we implement methods on `GFile` by using the functional API on the underlying `FileSystem`, so `LocalFileSystem.write()` opens a new file object each time we call `write()` (and hence automatically flushes it each time too, which eliminates the buffering effect).  This is obviously subpar for repeated small writes to local disk (like we are doing here), but fixing it properly requires restructuring the `FileSystem` abstraction so there's the ability to pass around filesystem-specific state.  I'd meant to redo some parts of the `FileSystem` abstraction anyway but obviously that's the most work, so it'd be nice to untangle if possible.

WDYT?

tensorboard/summary/output_test.py
```diff
@@ -0,0 +1,97 @@
...
+            data=np.float32(0),
+            step=np.int64(step),
+            wall_time=0.0,
+            tag_metadata=b"meta",
+            description="desc",
+        )
+        emit_scalar(1)
+        events = test_util.read_tfevents(logdir)
+        # Only the file version event should be present.
+        self.assertLen(events, 1)
```
wchargin, line 51: Hmm, looks like this is failing in notf—not obvious to me why, but could
be genuine?

wchargin (approved):
Looks good; thanks!

tensorboard/summary/output_test.py
nfelt: Fixed by relaxing the test and leaving a comment, as discussed.

tensorboard/summary/output_test.py
```diff
@@ -0,0 +1,97 @@
...
+        events = test_util.read_tfevents(logdir)
+        self.assertLen(events, 2)
+        event = events[1]
+        self.assertEqual(event.step, 12)
+        self.assertEqual(event.wall_time, 123.456)
+        summary = event.summary.value[0]
+        self.assertEqual(summary.tag, "tag")
+        self.assertEqual(
+            tensor_util.make_ndarray(summary.tensor), np.array(42.0)
+        )
```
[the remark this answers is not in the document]
nfelt: Done, confirmed the newly added assertion fails when I change the dtype to float64.

tensorboard/summary/BUILD
```diff
@@ -1,20 +1,81 @@
...
-        ":summary_v1",
-        ":summary_v2",
+        ":output",
+        "//tensorboard:expect_numpy_installed",
+    ],
+)
+
+py_test(
+    name = "writer_test",
+    srcs = ["writer_test.py"],
```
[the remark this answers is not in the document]
nfelt: Ok, I've updated them all to have a leading underscore.  It's slightly awkward-feeling in practice IMO, since the underscore starts getting kind of viral (leaks into BUILD targets, for the test util also into imports in the other files, etc.).  But I agree it is more consistent overall.

nfelt (commented):
Thanks for the review!

wchargin (approved):
Cool; thanks! If we end up finding the underscores too awkward, I’m
happy to backtrack/reconsider.

PR #4583, opened 2021-01-20
core: drop support for negative `--reload_interval`

wchargin:
Summary:
In #1001, we changed the interpretation of `--reload_interval` such that
negative values caused event files not to be read at all. This was to
accommodate DB mode: when loading from a database, reading event files
was superfluous. But DB mode is gone, and the corresponding distinction
for a Rust-based data provider is handled by the ingester system
instead. That is, `event_processing/data_ingester.py` now always loads
event files, and `data/server_ingester.py` never loads event files.

This is technically a backward-incompatible change, but DB mode was
always experimental, so the only stable use of this was to launch a
TensorBoard instance that never loaded any data at all. I’m okay with
breaking such use cases.

Test Plan:
Running with `--reload_interval` set to `0` or `1` or `inf` still works;
setting it to `-1` or `nan` or `zzz` gives an appropriate error message.

wchargin-branch: no-negative-reload-interval

### DIFF
--- a/tensorboard/backend/event_processing/data_ingester.py
+++ b/tensorboard/backend/event_processing/data_ingester.py
@@ -89,8 +89,6 @@ class LocalDataIngester(ingester.DataIngester):
 
     def start(self):
         """Starts ingesting data based on the ingester flag configuration."""
-        if self._reload_interval < 0:
-            return
 
         def _reload():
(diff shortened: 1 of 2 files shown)


nfelt (approved):
Fair enough.

PR #4584, opened 2021-01-21
core: fix reload interval parsing

wchargin:
Summary:
Fixes an error in #4583. Not sure how this slipped my testing; perhaps
I committed incorrectly.

Test Plan:
Follow the test plan from #4583, and double-check that data actually
loads in the UI.

wchargin-branch: fix-reload-interval


nfelt (approved):
tfw it would have worked in Rust

wchargin:
tfw type checking

PR #4586, opened 2021-01-21
tensorboard-data-server 0.1.1

wchargin:
Changes since 0.1.0:

  - performance improvements: checksums off by default, Tokio 1.0
  - support for `--reload_interval`
  - suppressed “listening on [::1]:PORT” spam

Merge plan: rebase and merge, then release from the unique commit where
the version numbers are 0.1.1.

PR #4587, opened 2021-01-21
tensorboard-data-server 0.2.0

wchargin:
Changes since 0.1.0:

  - performance improvements: checksums off by default; Tokio 1.0
  - `--reload once`, to support TensorBoard `--reload_interval`
  - suppressed “listening on [::1]:PORT” spam

Merge plan: rebase and merge, then release from the unique commit where
the version numbers are 0.2.0.

(Supersedes 0.1.1 from #4586. Let’s use minor releases for now so that
the version numbers advance monotonically with commits.)

### DIFF
--- a/tensorboard/data/server/Cargo.lock
+++ b/tensorboard/data/server/Cargo.lock
@@ -747,7 +747,7 @@ dependencies = [
 
 [[package]]
 name = "rustboard"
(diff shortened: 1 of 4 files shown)


nfelt (approved):
🚀

PR #4588, opened 2021-01-21
infra: create dev target

stephanwlee:
New `tensorboard:dev` target provides a development instance of the
application which optimizes for development experience. While the
changes are omitted in this commit for brevity, we are looking to stream
line the FE building to optimize for its incremental build time.

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -357,6 +357,7 @@ tensorboard_zip_file(
     deps = [":dev_assets"],
 )
 
+# Should keep the asset dependencies in sync with :assets.
 tf_web_library(
     name = "dev_assets",
     srcs = [
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -19,6 +19,7 @@ package_group(
 py_binary(
     name = "tensorboard",
     srcs = ["main.py"],
+    data = ["webfiles.zip"],
     main = "main.py",
     srcs_version = "PY3",
     deps = [
@@ -42,6 +43,25 @@ py_library(
     deps = [
         "//tensorboard:expect_absl_logging_installed",
         "//tensorboard/compat:tensorflow",
+    ],
+)
+
+py_binary(
+    name = "dev",
+    srcs = ["main_dev.py"],
+    data = ["dev_webfiles.zip"],
+    main = "main_dev.py",
+    srcs_version = "PY3",
+    deps = [
+        ":default",
+        ":dynamic_plugins",
+        ":lib",
+        ":program",
+        "//tensorboard:expect_absl_flags_installed",
+        "//tensorboard/plugins:base_plugin",
+        "//tensorboard/uploader:uploader_subcommand",
+        "//tensorboard/util:tb_logging",
+        "//tensorboard/util:timing",  # non-strict dep, for patching convenience
     ],
 )
 
@@ -248,7 +268,6 @@ py_library(
 py_library(
     name = "default",
     srcs = ["default.py"],
-    data = ["webfiles.zip"],
     srcs_version = "PY3",
     deps = [
         "//tensorboard:expect_pkg_resources_installed",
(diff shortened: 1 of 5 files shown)


tensorboard/main_dev.py
```diff
@@ -0,0 +1,77 @@
...
+        os.path.dirname(inspect.getfile(sys._getframe(1))), "dev_webfiles.zip"
+    )
+
+    tensorboard = program.TensorBoard(
+        default.get_plugins(),
+        lambda: open(path, "rb"),
+        subcommands=[uploader_subcommand.UploaderSubcommand()],
+    )
+
+    app.run(tensorboard.main, flags_parser=tensorboard.configure)
```
nfelt, line 72: Can we preserve the handling for FlagsError?  Nice to not have extra behavior differences between dev and non-dev versions, and it's only a few lines.

tensorboard/main_dev.py
```diff
@@ -0,0 +1,77 @@
...
+from tensorboard.uploader import uploader_subcommand
+from tensorboard.util import tb_logging
+
+
+logger = tb_logging.get_logger()
+
+
+def run_main():
+    """Initializes flags and calls main()."""
+    logger.info("TensorBoard running in development mode.")
```
nfelt, line 52: The next line reverts to not sending INFO messages to the logs at all, so a bit confusing IMO to allow this one message at INFO first.  Can we just write directly to stderr?  A bit more consistent with our other messages as well (the TF not found one below, and the one that prints out the serving address).

tensorboard/BUILD
```diff
@@ -19,11 +19,12 @@ package_group(
 py_binary(
     name = "tensorboard",
     srcs = ["main.py"],
+    data = ["webfiles.zip"],
```
nfelt, line 22: So I agree this is where the data dep belongs, but right now the code in `main.py` actually reads it from `program.get_default_assets_zip_provider()`.  It used to live in `default` but was moved to program back in #1631 (not really sure why, at this point).

IMO better at this point to just inline it into `main.py` and remove the fallback logic in `program.TensorBoard()` that calls it, so that `assets_zip_provider` must be specified by the caller.  That way we don't have skew between where the data dep is defined in the BUILD file, and where the code actually expects it to exist.  WDYT?

tensorboard/BUILD
```diff
@@ -325,6 +342,28 @@ tf_web_library(
     ],
 )
 
+tensorboard_zip_file(
+    name = "dev_webfiles",
+    deps = [":dev_assets"],
+)
+
+tf_web_library(
+    name = "dev_assets",
```
nfelt, line 351: I assume the idea is that we'll replace this fairly soon with something more substantially different?  Otherwise, would be good to have a comment saying to keep the srcs/deps in sync with the non-dev `assets`.

tensorboard/BUILD
stephanwlee: Makes sense. Done.

tensorboard/BUILD
stephanwlee: Yup, #4589 gives you glimpse of what the difference might be.

tensorboard/program.py
```diff
@@ -107,7 +88,8 @@ def __init__(
           plugins: A list of TensorBoard plugins to load, as TBPlugin classes or
             TBLoader instances or classes. If not specified, defaults to first-party
             plugins.
-          assets_zip_provider: Delegates to TBContext or uses default if None.
+          assets_zip_provider: A function that provides a zip file containing assets to
```
nfelt, line 91: Move up one to match arg order?

tensorboard/program_test.py
```diff
@@ -26,37 +26,51 @@
 from tensorboard.plugins.core import core_plugin
 
 
+def fake_asset_provider():
```
nfelt, line 29: If we wanted to avoid the boilerplate here we could always make the argument still optional, but if omitted, we would generate a dummy zip file that just contains `index.html` with `Hello TensorBoard` or something.  Probably fine to make it required, but just a thought.

nfelt (approved):
Pip package smoke tests seems to be broken by this change :/  If I had to guess, maybe either moving the data dependency, or moving the `webfiles.zip` reference (I would think referencing from `program.py` would work the same as from `main.py` but maybe for some strange reason it's different).

tensorboard/program.py
stephanwlee: Done. Thanks

tensorboard/program_test.py
stephanwlee: I prefer things to be a bit more explicit so I will keep it as is.

PR #4590, opened 2021-01-21
line chart: restore thicker line width

psybuzz:
WebGL enabled line charts have been using a thinner line thickness
by default, due to a limitation in WebGL [1]. A common workaround
is to render line thickness manually, turning each line segment into
a rectangle composed of 2 triangles. This change uses the approach
to restore the 2 pixel line width found in stable TensorBoard.

Notably, line segments used to be representable by 2 vertices are
now represented with 6 vertices (2 triangles), so performance is
expected to degrade.

This change does not take into account the "end caps" of each line,
so minor artifacts are noticeable when examining the joint between
two consecutive line segments carefully. Addressing the end caps
may require adding more triangles, which adds a decent amount
of complexity without significant noticeable benefit.

Manually checked by:
Adding `?fastChart=true#timeseries` to the URL to view WebGL enabled
line charts in Time Series.

[1] https://github.com/mrdoob/three.js/issues/14627

Before (with ?fastChart=true)
image

After (with ?fastChart=true)
image

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
@@ -42,8 +42,9 @@ enum CacheType {
 
 interface LineCacheValue {
   type: CacheType.LINE;
-  obj3d: THREE.Line;
+  obj3d: THREE.Mesh;
   data: Polyline;
+  width: number;
 }
 
 interface TriangleCacheValue {
@@ -61,9 +62,10 @@ interface CircleCacheValue {
 type CacheValue = LineCacheValue | TriangleCacheValue | CircleCacheValue;
 
 /**
- * Updates BufferGeometry with Float32Array that denotes flattened array of Vec2 (<x, y>).
+ * Updates BufferGeometry with Float32Array that denotes flattened array of Vec2
+ * (<x, y>) representing points that form a connected set of line segments.
  */
-function updateGeometryWithVec2Array(
+function updatePolylineGeometry(
   geometry: THREE.BufferGeometry,
   flatVec2: Float32Array
 ) {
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
```diff
@@ -92,14 +94,112 @@ function updateGeometryWithVec2Array(
...
+ * THREE.LineBasicMaterial is ignored [1]. Each line segment is a rectangle
+ * that is split into 2 triangles [A->B->C] and [C->B->D]. Each triangle has
+ * 3 components (x, y, z).
+ *
+ * Assuming a line segment is as follows:
+ *              A             C
+ *              |             |
+ * (startPoint) |-------------| (endPoint)
+ *              |             |
+ *              B             D
```
stephanwlee, line 113: Not sure if this is better but how about this? 

````
   ^   A+--+C
   |   +   /+
dy |   | /  |
   |   +/   +
   v   B+--+D

       <---->
         dx
```

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
```diff
@@ -92,14 +94,112 @@ function updateGeometryWithVec2Array(
...
+ * [1] https://github.com/mrdoob/three.js/issues/14627
+ */
+function updateThickPolylineGeometry(
+  geometry: THREE.BufferGeometry,
+  flatVec2: Float32Array,
+  thickness: number
+) {
+  const numSegments = flatVec2.length / 2 - 1;
+  const numVertices = numSegments * 2 * 3;
+  const numComponents = numVertices * 3;
```
stephanwlee, line 124: Component == xyz coordinate for each vertex? Hmm, how about coordinates?

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
```diff
@@ -92,14 +94,112 @@ function updateGeometryWithVec2Array(
...
+  for (let i = 0; i < numSegments; i++) {
+    const [x1, y1, x2, y2] = [
+      flatVec2[2 * i],
+      flatVec2[2 * i + 1],
+      flatVec2[2 * i + 2],
+      flatVec2[2 * i + 3],
+    ];
+    const startPointVec = new THREE.Vector2(x1, y1);
+    const endPointVec = new THREE.Vector2(x2, y2);
+    const unitNormalVec = new THREE.Vector2(x2 - x1, y2 - y1)
```
stephanwlee, line 147: Hmm rotation feels a bit more crazier than what I had in mind. If we can calculate `dy / 2`, should it not be quite easy to calculate the coordinates ourselves? I am sure rotation is a simple matrix operation but doing rotation then scaling sounds like one affine transformation that _can_ be parallelized (but we don't here) and is more complex than simple arithmetic. WDYT?

stephanwlee (commented):
Cool idea, it mostly looks good to me. I want to hear the response to my feedback before approving.

psybuzz (commented):
re testing: now that rotations are in play, it no longer makes much sense to compare exact coordinate values, because the coordinates drawn no corresponds 1 to 1 with the polyline endpoints, and values are offset due to rotation.

Perhaps the ideal test for catching regressions in these charts is a screenshot test, which exists internally but not in OSS yet.

stephanwlee (approved):

PR #4592, opened 2021-01-21
[npmi] Run ID to RunName

Sparkier:
Clean up the hacky way we previously handled the conversion from runID to name.


tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_container.ts
```diff
@@ -83,6 +86,21 @@ export class AnnotationContainer {
...
+    switchMap((experimentIds) => {
+      return combineLatest(
+        experimentIds!.map((experimentId) => {
+          return this.store.select(selectors.getRuns, {experimentId});
+        })
+      );
+    }),
+    map((runMaps) => {
+      let maps: Run[] = [];
+      return maps.concat(...runMaps);
```
stephanwlee, line 101: there is `Array.prototype.flat` you can use. Also, this is an array, not a map (object/dict/Map).

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_container.ts
```diff
@@ -51,6 +53,7 @@ import {RunColorScale} from '../../../../../types/ui';
       [showCounts]="showCounts$ | async"
       [sidebarWidth]="sidebarWidth$ | async"
       [colorScale]="runColorScale$ | async"
+      [runMaps]="runMaps$ | async"
```
psybuzz, line 56: iiuc, the container has @Input bindings for `data` (an array of items, each of which has a runId), and `runMaps` (an array of `Run`s, which we use to convert runIds into run names).  In this PR, we jump through some hoops to get the actual run names:
- first getting current experiment ids
- then getting runs based on that
- then doing a linear lookup in the Runs array for a matching run (and `drawRunHintTexts` is called on every redraw call?)

I'd like to propose something which is hopefully more convenient:
- Introduce a new `getRunMap()` selector in `webapp/runs/store/runs_selectors.ts`, whose only purpose is to convert `state.runMetadata` into an ES6 Map, and return it
- `annotation_container.ts` uses the selector, and passes the `runMap` Map directly into `annotation_component`
- Instead of a linear `this.runMaps.find`, the component does `this.runMap.get(run)?.name || ''` to get the run name from the runId

With this, we skip the experimentId/route business, and have an actual Map to use, rather than an array.

wdyt?

psybuzz (commented):
Thanks for the code cleanup!

tensorboard/webapp/plugins/npmi/data_source/npmi_data_source.ts
```diff
@@ -124,7 +124,7 @@ export class NpmiHttpServerDataSource implements NpmiDataSource {
         for (const run of Object.keys(annotations)) {
           for (const annotationIndex in annotations[run]) {
             const annotation = annotations[run][annotationIndex];
-            if (embeddings[run][annotationIndex]) {
+            if (embeddings[run] && embeddings[run][annotationIndex]) {
```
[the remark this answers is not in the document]
Sparkier: It is possible (and with my data in fact the case) that the list of embeddings does not match the list of annotations. It is also possible that there are no embeddings provided for a run or in general.
The frontend handles this absence and just does not display embedding-related functionality.

tensorboard/webapp/plugins/npmi/views/annotations_list/annotation/annotation_container.ts
Sparkier: That was something I thought about as well, but I assume there was a reason why this was not an option. Otherwise the whole run fetching would have been much easier. Can you comment on this @stephanwlee?

stephanwlee: > With this, we skip the experimentId/route business, and have an actual Map to use, rather than an array.

You still need to supply it with the `experimentId` information (unless you are suggesting making `getRunsForCurrentRoute`.

psybuzz: I was thinking of a `getRunMap()` that just returns all the Runs it knows about, without requiring any `experimentId`. 
 Although, I do see the reason to avoid offering the map in a selector.  Giving a map makes it easy for consumers to mistake the keys as a list of "all the runs", when they actually want only runs for the current route.

If we really don't want an experiment-agnostic `getRunMap()`, could we take another approach where
- annotation_container takes all the runIds from its `data` input prop
- calls `this.store.select(getRun, runId).pipe(take(1))` on each one, and collects the results
- assembles the runIdToName map itself, and passes that to annotation_component

?

The annotation_container already has runIds which are specific enough, so it has no need to be aware of experimentIds.

stephanwlee: Sorry, I still don't understand. The runs_reducer contain many data including `eid -> RunId[]` and `RunId -> RunMetadata`. Are you saying we should just expose the `RunId -> RunMetadata`?  I have no big issue with that except I don't think that change is necessary to unblock this PR.

psybuzz: > Are you saying we should just expose the RunId -> RunMetadata?

Yup, that's my suggestion.  My intent is not to block the PR on that change, but rather to call out that `annotation_container` should not have to fetch experimentIds and that passing a "runIdToRun" Map seems more appropriate to pass into the component instead of an Array.

stephanwlee: > ...to pass into the component 

Do you mean big C Component? (i.e., Container vs. Component)? Ideally, you should pass neither array nor map but instead a UI data structure that you form in the Container into the Component with only minimal information that you require for rendering.

psybuzz: In https://github.com/tensorflow/tensorboard/pull/4609, I'm proposing a new selector `getRunMap` returning a `Map<string, Run>`, which should make things simpler in this PR.

Assuming folks like the idea, the change in this PR could look like:
```ts
// annotation_container.ts
<annotation-component [runIdToRuns] = "runIdToRuns$ | async">
// ...
runIdToRuns$ = this.store.select(getRunMap);
```
```ts
// annotation_component.ts
@Input runIdToRuns;
// ...
      .text((d: string) => {
        return this.runIdToRuns.get(d)?.name || '';
      });
```

> Ideally, you should pass neither array nor map but instead a UI data structure that you form in the Container into the Component with only minimal information that you require for rendering.

That's true.  Alternatively, one might want `data: ValueData[]` to be changed to `dataWithRunDisplayNames: ValueDataWithRunDisplayNames[]`, and annotations_container only has to pass to its `annotation_component` just `data`, instead of `data` and `runNames`.  I'll leave that up to the PR author, both are fine to me.

psybuzz (approved):
As noted in https://github.com/tensorflow/tensorboard/pull/4609#pullrequestreview-576029813, passing the entire map from container -> component is less pure than passing only the required info needed for rendering.

Without knowing how `this.runs` is used by the component in detail, I don't have a good sense of how much complexity is added by trying to filter the map or merge its info with `this.data`.  As such, I think its reasonable to pass the map as is for convenient lookup, but will defer to @stephanwlee for final lgtm.

stephanwlee (approved):

PR #4593, opened 2021-01-21
feature flag: persist only overridden flags

stephanwlee:
Currently, TensorBoard reflects the feature flags as is to the URL. This
means, we cannot differentiate overrides from defaults and makes
changing the defaults harder.

For instance, soon, we want to change fastChart to default true but, in
that case, we don't want the URL to say anything about fastChart. Also,
in that world, we would like to support fastChart=false which should
remain in the URL since it is an explicit override user has requested.

This change now differentiate overrides from default values in the
reducer.

### DIFF
--- a/tensorboard/webapp/feature_flag/store/feature_flag_selectors.ts
+++ b/tensorboard/webapp/feature_flag/store/feature_flag_selectors.ts
@@ -15,6 +15,7 @@ limitations under the License.
 
 import {createSelector, createFeatureSelector} from '@ngrx/store';
 
+import {FeatureFlags} from '../types';
 import {
   FeatureFlagState,
   FEATURE_FLAG_FEATURE_KEY,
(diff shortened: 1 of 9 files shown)


tensorboard/webapp/feature_flag/store/feature_flag_selectors.ts
```diff
@@ -36,25 +37,32 @@ export const getIsFeatureFlagsLoaded = createSelector(
...
+export const getOverridenFeatureFlags = createSelector(
   selectFeatureFlagState,
+  (state: FeatureFlagState): Partial<FeatureFlags> => {
+    return state.flagOverrides;
+  }
+);
+
+export const getEnabledExperimentalPlugins = createSelector(
+  getFeatureFlags,
   (state) => {
```
psybuzz, line 54: now that the parameter is no longer `FeatureFlagState`, can we rename these `state` params to `flags`, here and below?

psybuzz (approved):

PR #4594, opened 2021-01-21
Expose getParams() in QueryParamsFeatureFlagDataSource

bmd3k:

  We want to add application-specific feature flags for varieties of TensorBoard. Part of this work requires implementing application-specific children of QueryParamsFeatureFlagDataSource. There is some logic from the base class that we'd like to share with children, so we can avoid duplicating some testing hooks.

  We expose a new protected method getParams() in QueryParamsFeatureFlagsDataSource. It returns util.getParams(), which can be overriden in tests for the child classes.

  Googlers, see http://cl/353022802 for some more context.


psybuzz (approved):
Thanks for the CL link for context

PR #4595, opened 2021-01-21
ci: run on pushes to tags

wchargin:
Summary:
Running CI on tags seems sensible on its own merits, and also enables us
to trigger CI on data server releases by tagging them rather than using
a temporary branch.

Test Plan:
The workflow run for this PR should show that the definition is still
valid. After merging, push a new tag for a commit that includes this
patch, and check that it starts a CI run.

wchargin-branch: ci-trigger-on-tags

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,8 @@ on:
       - master
       - '[0-9]+.*'
       - 'ci-*'
+    tags:
+      - '*'
   pull_request: {}
   schedule:
     # 13:00 UTC is 05:00 in Pacific standard time (UTC-8), which is well


wchargin:
I tried this out on a fork. I think what’s happening is that it runs the
workflow when I tag a commit that includes this workflow definition
change, but not when I tag an earlier commit. That is, when a tag is
pushed, it determines whether to run the workflow based on the CI at
that tag rather than at head, which makes sense.

This should be fine for us. It does mean that to test the functionality
post-merge, I’ll need to create a temporary tag and then delete it.
Altering tag history is generally frowned upon, but this feels different
from “oops, let’s change the `3.0.0` tag to also include a cherry-pick”,
so I can be okay with it.

nfelt (approved):

PR #4603, opened 2021-01-25
rustboard: propagate --verbosity=debug to tensorboard-data-server

nfelt:
This extends the current propagation of `INFO`-level logging to also apply to `DEBUG` - we just pass `--verbose` a second time if the Python code is logging at `DEBUG`.

### DIFF
--- a/tensorboard/data/server_ingester.py
+++ b/tensorboard/data/server_ingester.py
@@ -86,6 +86,8 @@ class SubprocessServerDataIngester(ingester.DataIngester):
         ]
         if logger.isEnabledFor(logging.INFO):
             args.append("--verbose")
+        if logger.isEnabledFor(logging.DEBUG):
+            args.append("--verbose")  # Repeat arg to increase verbosity.
 
         logger.info("Spawning data server: %r", args)
         popen = subprocess.Popen(args, stdin=subprocess.PIPE)


wchargin (approved):
Nice; thanks.

wchargin:
In case it’s new to you: you can set the `RUST_LOG` env variable for
more precise logging, like:

```
RUST_LOG=rustboard_core::cli=info,rustboard_core::run=debug bazel run ...
```

or just `RUST_LOG=rustboard` to show all logs from our code. (It’s a
prefix match on the module name: <https://docs.rs/env_logger>.) This can
be helpful since the actual gRPC server has a log of `TRACE`-level logs.

(I still approve of this change.)

nfelt:
> In case it’s new to you: you can set the `RUST_LOG` env variable for
> more precise logging, like:
> 
> ```
> RUST_LOG=rustboard_core::cli=info,rustboard_core::run=debug bazel run ...
> ```
> 
> or just `RUST_LOG=rustboard` to show all logs from our code. (It’s a
> prefix match on the module name: https://docs.rs/env_logger.) This can
> be helpful since the actual gRPC server has a log of `TRACE`-level logs.

Ah, thanks, useful!  I wondered briefly if this existed but not long enough to go digging for it (in part since I saw the propagation code was already there).

PR #4604, opened 2021-01-25
rust: make RunLoader commit periodically while loading a large run

nfelt:
Addresses the online commits task from #4422.

RunLoader's `reload()` method now ensures that we commit all time series data staged from the run at a regular interval (which I've hardcoded to 5 seconds), rather than waiting until all event files in the run have been read to exhaustion before committing any data.  I did a little refactoring to facilitate making this change which I've broken out into separate commits.

Based on a minimal amount of benchmarking against the ~20GiB event file from `gs://tensorboard-bench-logs/nndiv_100m/events.out.tfevents.1549560561.hostname`, it looks like this adds about 1% overhead in reading the file to completion (first is unmodified code; second is with this PR):

```
$ hyperfine -m 5 'bazel-bin/tensorboard/data/server/bench --logdir ~/scratch/tensorboard-bench-logs/'
Benchmark #1: bazel-bin/tensorboard/data/server/bench --logdir ~/scratch/tensorboard-bench-logs/
  Time (mean ± σ):     168.505 s ±  2.442 s    [User: 164.278 s, System: 4.212 s]
  Range (min … max):   166.091 s … 171.609 s    5 runs

$ hyperfine -m 5 'bazel-bin/tensorboard/data/server/bench --logdir ~/scratch/tensorboard-bench-logs/'
Benchmark #1: bazel-bin/tensorboard/data/server/bench --logdir ~/scratch/tensorboard-bench-logs/
  Time (mean ± σ):     171.296 s ±  2.586 s    [User: 167.006 s, System: 4.275 s]
  Range (min … max):   167.432 s … 174.439 s    5 runs
```

Test plan: ran against the aforementioned logdir and verified that we now see new data steadily appearing at 5 second intervals.

I haven't modified the unit tests to cover this behavior because I was a bit unsure about how to go about it.  My main thought was parameterizing `reload()` (or `RunLoader` at construction) in order to inject some kind of fake clock object, but I wasn't sure if there was a standard technique used for this.  Open to suggestions.

### DIFF
--- a/tensorboard/data/server/run.rs
+++ b/tensorboard/data/server/run.rs
@@ -15,19 +15,20 @@ limitations under the License.
 
 //! Loader for a single run, with one or more event files.
 
-use log::warn;
+use log::{debug, warn};
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fs::File;
 use std::io::BufReader;
 use std::path::PathBuf;
 use std::sync::RwLock;
+use std::time::Instant;
 
 use crate::commit;
 use crate::data_compat::{EventValue, GraphDefValue, SummaryValue, TaggedRunMetadataValue};
 use crate::event_file::EventFileReader;
 use crate::proto::tensorboard as pb;
 use crate::reservoir::StageReservoir;
-use crate::types::{Step, Tag, WallTime};
+use crate::types::{Run, Step, Tag, WallTime};
 
 /// A loader to accumulate reservoir-sampled events in a single TensorBoard run.
 ///
@@ -35,6 +36,10 @@ use crate::types::{Step, Tag, WallTime};
 /// parameterized over a filesystem interface.
 #[derive(Debug)]
 pub struct RunLoader {
+    /// The run name associated with this loader. Used primarily for logging; the run name is
+    /// canonically defined by the map key under which this RunLoader is stored in LogdirLoader.
+    run: Run,
+
     /// The earliest event `wall_time` seen in any event file in this run.
     ///
     /// This is `None` if and only if no events have been seen. Its value may decrease as new
@@ -146,8 +151,9 @@ impl StageTimeSeries {
 }
 
 impl RunLoader {
-    pub fn new() -> Self {
+    pub fn new(run: Run) -> Self {
         Self {
+            run: run,
             start_time: None,
             files: BTreeMap::new(),
             time_series: HashMap::new(),
@@ -171,9 +177,17 @@ impl RunLoader {
     ///
     /// If we need to access `run_data` but the lock is poisoned.
     pub fn reload(&mut self, filenames: Vec<PathBuf>, run_data: &RwLock<commit::RunData>) {
+        let run_name = self.run.0.clone();
+        debug!("Starting load for run {:?}", run_name);
+        let start = Instant::now();
         self.update_file_set(filenames);
         self.reload_files();
         self.commit_all(run_data);
+        debug!(
+            "Finished load for run {:?} ({:?})",
+            run_name,
+            start.elapsed()
+        );
     }
 
     /// Updates the active key set of `self.files` to match the given filenames.
@@ -329,12 +343,6 @@ fn read_event(
     }
 }
 
-impl Default for RunLoader {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 #[cfg(test)]
 mod test {
     use super::*;
@@ -390,7 +398,7 @@ mod test {
         f1.into_inner()?.sync_all()?;
         f2.into_inner()?.sync_all()?;
 
-        let mut loader = RunLoader::new();
+        let mut loader = RunLoader::new(run.clone());
         let commit = Commit::new();
         commit
             .runs
--- a/tensorboard/data/server/run.rs
+++ b/tensorboard/data/server/run.rs
@@ -189,7 +189,7 @@ impl RunLoader {
         debug!("Starting load for run {:?}", run_name);
         let start = Instant::now();
         self.update_file_set(filenames);
-        self.reload_files();
+        self.reload_files(read_event);
         commit_all(&mut self.data, run_data);
         debug!(
             "Finished load for run {:?} ({:?})",
@@ -235,8 +235,8 @@ impl RunLoader {
         }
     }
 
-    /// Reads data from all active event files.
-    fn reload_files(&mut self) {
+    /// Reads data from all active event files, and calls a handler for each event.
+    fn reload_files<F: FnMut(&mut RunLoaderData, pb::Event)>(&mut self, mut handle_event: F) {
         for (filename, ef) in self.files.iter_mut() {
             let reader = match ef {
                 EventFile::Dead => continue,
@@ -256,7 +256,7 @@ impl RunLoader {
                         break;
                     }
                 };
-                read_event(&mut self.data, event);
+                handle_event(&mut self.data, event);
             }
         }
     }
--- a/tensorboard/data/server/run.rs
+++ b/tensorboard/data/server/run.rs
@@ -37,7 +37,7 @@ use crate::types::{Run, Step, Tag, WallTime};
 #[derive(Debug)]
 pub struct RunLoader {
     /// The run name associated with this loader. Used primarily for logging; the run name is
-    /// canonically defined by the map key under which this `RunLoader` is stored in LogdirLoader.
+    /// canonically defined by the map key under which this `RunLoader` is stored in `LogdirLoader`.
     run: Run,
     /// The event files in this run.
     ///
@@ -164,7 +164,7 @@ const COMMIT_INTERVAL: Duration = Duration::from_secs(5);
 impl RunLoader {
     pub fn new(run: Run) -> Self {
         Self {
-            run: run,
+            run,
             files: BTreeMap::new(),
             checksum: true,
             data: RunLoaderData::default(),
--- a/tensorboard/data/server/run.rs
+++ b/tensorboard/data/server/run.rs
@@ -190,7 +190,7 @@ impl RunLoader {
         let start = Instant::now();
         self.update_file_set(filenames);
         self.reload_files();
-        self.commit_all(run_data);
+        commit_all(&mut self.data, run_data);
         debug!(
             "Finished load for run {:?} ({:?})",
             run_name,
@@ -260,13 +260,17 @@ impl RunLoader {
             }
         }
     }
+}
 
-    fn commit_all(&mut self, run_data: &RwLock<commit::RunData>) {
-        let mut run = run_data.write().expect("acquiring tags lock");
-        run.start_time = self.data.start_time;
-        for (tag, ts) in &mut self.data.time_series {
-            ts.commit(tag, &mut *run);
-        }
+/// Commits all data staged in the run loader into the given run of the commit.
+///
+/// This is a standalone function because it's called from `reload_files` in a context that already
+/// has an exclusive reference into `self.files`, and so can't call methods on `&mut self`.
+fn commit_all(run_loader_data: &mut RunLoaderData, run_data: &RwLock<commit::RunData>) {
+    let mut run = run_data.write().expect("acquiring tags lock");
+    run.start_time = run_loader_data.start_time;
+    for (tag, ts) in &mut run_loader_data.time_series {
+        ts.commit(tag, &mut *run);
     }
 }
 
--- a/tensorboard/data/server/run.rs
+++ b/tensorboard/data/server/run.rs
@@ -39,14 +39,6 @@ pub struct RunLoader {
     /// The run name associated with this loader. Used primarily for logging; the run name is
     /// canonically defined by the map key under which this RunLoader is stored in LogdirLoader.
     run: Run,
-
-    /// The earliest event `wall_time` seen in any event file in this run.
-    ///
-    /// This is `None` if and only if no events have been seen. Its value may decrease as new
-    /// events are read, but in practice this is expected to be the wall time of the first
-    /// `file_version` event in the first event file.
-    start_time: Option<WallTime>,
-
     /// The event files in this run.
     ///
     /// Event files are sorted and read lexicographically by name, which is designed to coincide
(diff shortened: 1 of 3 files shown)


tensorboard/data/server/run.rs
```diff
@@ -69,6 +66,20 @@ enum EventFile<R> {
     Dead,
 }
 
+/// Holds data staged by a RunLoader that will be commited to the Commit.
+#[derive(Debug)]
+struct RunLoaderData {
```
wchargin, line 71: I like the idea of extracting this type.

When first writing this module, I considered separating `RunLoader` into
`files` and `data` and putting methods on `data`, to cut down on
free-floating functions taking a bunch of `&mut` arguments. It seems
like that’s about what you’re doing here, so, some observations:

  - You’ve written this semantically as “the data that will be
    committed”, which makes sense. You could also write it more
    structurally as “the data that isn’t in `files`, since that’s the
    field into which we like to take long-lived mutable references”.
    I.e.: you could include `self.run` and `self.checksum` here.

    Doing so avoids the clone of `run_name`, which obviously isn’t a big
    performance hit, but makes me wonder whether the structure might be
    nice. It also simplifies the callback structure a bit (see below).

  - The free-standing `fn commit_all` and `fn read_event` functions are
    nicely set up to be `&mut self` methods on `RunLoaderData`, which I
    viewed as one of the benefits of this structure in the first place.

Thoughts?

tensorboard/data/server/run.rs
```diff
@@ -171,9 +188,33 @@ impl RunLoader {
...
     pub fn reload(&mut self, filenames: Vec<PathBuf>, run_data: &RwLock<commit::RunData>) {
+        let run_name = self.run.0.clone();
+        debug!("Starting load for run {:?}", run_name);
+        let start = Instant::now();
         self.update_file_set(filenames);
-        self.reload_files();
-        self.commit_all(run_data);
+        let mut n = 0;
+        let mut last_commit_time = Instant::now();
+        self.reload_files(|run_loader_data, event| {
```
wchargin, line 197: Thoughts on inlining this callback into `reload_files`? I don’t mind
callbacks in general, but they are a bit more complicated than not using
them. If `self.run` moves to `self.data.run`, then you only need to pass
in `self.reload_files(run_data, &start)`, which feels pretty reasonable
to me: it’s natural that `reload_files` should take the destination, and
`start` is just for instrumentation.

Then `reload` reads like:

```rust
let start = Instant::now();

self.update_file_set(filenames);
self.reload_files(run_data, &run_name, &start);
self.data.commit_all(run_data);
// ^ maybe move this into `reload_files`, too? maybe; could go either way

debug!(
    "Finished load for run {:?} ({:?})",
    self.data.run.0,
    start.elapsed()
);
```

tensorboard/data/server/run.rs
```diff
@@ -15,33 +15,30 @@ limitations under the License.
...
 pub struct RunLoader {
-    /// The earliest event `wall_time` seen in any event file in this run.
-    ///
-    /// This is `None` if and only if no events have been seen. Its value may decrease as new
-    /// events are read, but in practice this is expected to be the wall time of the first
-    /// `file_version` event in the first event file.
-    start_time: Option<WallTime>,
-
+    /// The run name associated with this loader. Used primarily for logging; the run name is
+    /// canonically defined by the map key under which this RunLoader is stored in LogdirLoader.
```
wchargin, line 40: nit: Since Rustdoc is Markdown, it’s [convention][1] to use backticks
around code atoms like `` `RunLoader` `` and `` `Commit` ``. You can
also write doc crosslinks like ``[`RunLoader`]``, but that’s more at
your discretion as to whether it’s useful.

(Throughout.)

tensorboard/data/server/run.rs
```diff
@@ -69,6 +66,20 @@ enum EventFile<R> {
     Dead,
 }
 
+/// Holds data staged by a RunLoader that will be commited to the Commit.
```
wchargin, line 69: sp.: s/commited/committed/

tensorboard/data/server/run.rs
```diff
@@ -269,8 +310,12 @@ fn read_event(
         }
         Some(wt) => wt,
     };
-    if start_time.map(|start| wall_time < start).unwrap_or(true) {
-        *start_time = Some(wall_time);
+    if data
+        .start_time
+        .map(|start| wall_time < start)
+        .unwrap_or(true)
```
wchargin, line 316: Can be `.map_or(true, |start| wall_time < start)`? (Could have been
originally, but presumably I forgot; now, it makes the code short enough
to not have to spill onto four lines, so worth changing imho.)

tensorboard/data/server/run.rs
```diff
@@ -269,8 +310,12 @@ fn read_event(
...
+    {
+        (*data).start_time = Some(wall_time);
```
wchargin, line 318: Can just be `data.start_time`; no need to explicitly deref.

(i.e., C’s `operator->` and `operator.` are basically both just `.`.)

tensorboard/data/server/run.rs
```diff
@@ -145,13 +156,19 @@ impl StageTimeSeries {
...
         Self {
-            start_time: None,
+            run: run,
             files: BTreeMap::new(),
-            time_series: HashMap::new(),
             checksum: true,
+            data: RunLoaderData {
+                start_time: None,
+                time_series: HashMap::new(),
+            },
```
wchargin, line 171: Optional: if you let `RunLoaderData` derive `Default`, then this is its
default value, so can be just `data: RunLoaderData::default()`.

wchargin (commented):
Lovely! Works on my machine. Some comments inline.

I’m okay with a ~1% overhead, especially since your μ±σ confidence
intervals overlap a bunch, and since this feature is clearly important
for non-small datasets.

> (first is unmodified code; second is with this PR):

One trick is to copy the old and new binaries to `./bench-{old,new}`
and then ask `hyperfine` to do a parameter sweep:

```
hyperfine '{tool} --logdir logs' -L tool ./bench-old,./bench-new
```

The main benefit is that it’ll give you a confidence interval around the
relative performance, something like “old is 1.02(±0.06)× faster”.

> I haven't modified the unit tests to cover this behavior because I was a bit unsure about how to go about it. My main thought was parameterizing `reload()` (or `RunLoader` at construction) in order to inject some kind of fake clock object, but I wasn't sure if there was a standard technique used for this. Open to suggestions.

That’s about right. The standard technique often involves doing so with
type parameters such that it’s all optimized out: [`StageReservoir`] and
its `ReservoirControl` do this. As to clock objects specifically: I’ve
not done this, and I don’t know of common patterns, but I imagine that a
simple `trait Clock { fn now(&self) -> Instant }` would suffice here,
and let the test clock own an `Rc<Instant>` or `AtomicI32` or something
that ticks forward each time that it’s called.

I’m okay with omitting dedicated tests for this. It seems unlikely to be
subtly broken, and the failure mode is not catastrophic (read-only, and
the right data still shows up, just later). Feel free to dive in if
you’re interested, but no action requested.

[`StageReservoir`]: https://github.com/tensorflow/tensorboard/blob/a906f7191137c1816c0544c00432a308fc627d2f/tensorboard/data/server/reservoir.rs#L70

> I did a little refactoring to facilitate making this change which I've broken out into separate commits.

Thanks; this was helpful. Feel free to merge however you want (rebase,
squash, separate PRs).

tensorboard/data/server/run.rs
nfelt: Good point about how the functions could be methods, I hadn't thought that far.  Can do.

Re: the scope of the data structure, I did briefly think about which fields to include and it seemed like a more semantically coherent boundary to define it as the data that will be committed.  Defining as just "everything except files" felt odd to me even if functionally it's the most convenient.  In particular even though cloning `run` is a bit ugly, semantically it just feels quite weird to make `run` a field on `self.data` instead of the top-level `RunLoader`, so I guess I preferred putting the weirdness inside the method bodies versus into the struct layout.  Similarly, it seems like semantically `self.checksum` would be more naturally located alongside `self.files` since it only affects the file-reading half of `RunLoader`, not the data-buffering half.

I guess in the purest sense we want something like a tripartite structure in which we have "effectively constant metadata" like `self.run` and `self.checksum` grouped in one substructure, and and then two different mutable substructures, `self.files` and `self.data`, so that all three can be reborrowed independently from the parent struct, but that seems... overcomplicated?  Would be nice to have some way to help the borrow checker resolve this without as much reshuffling of the actual struct fields themselves.

Thoughts?

tensorboard/data/server/run.rs
nfelt: Hmm.  So I arrived at this because I liked `reload_files()` better with the business logic of what to do with the event and how often to commit abstracted out of it.  In particular part of my reasoning was that if we wanted to parameterize the 5 second interval or do something fancier than a fixed interval (like triggering based on the event details itself), that configuration would all need to be plumbed into `reload_files()` either as their own parameters or as fields on `self`, and encapsulating it via the callback felt prettier.

As a halfway solution, I originally did have `commit_all()` called in `reload_files()` and was trying to pass in some kind of smaller "check if I need to commit yet" predicate instead, but then I remembered that abstractions are supposed to be free in Rust so I figured hey maybe I can just get away with calling a callback closure on every event :)

I'll grant it makes `reload()` a little messier; but I liked being able to read the sequence of commit steps and the timing details all in one function.  So if you aren't particularly opposed I think I'm happier with it this way.

nfelt (commented):
Thanks for the review!  Will address other comments after resolving the bigger questions.

tensorboard/data/server/run.rs
wchargin: Yep, that makes sense and seems reasonable to me. No strong opinion;
feel free to proceed as desired.

> Would be nice to have some way to help the borrow checker resolve this without as much reshuffling of the actual struct fields themselves.

You may be interested in [@nikomatsakis’s article on interprocedural
conflicts][1]. (Context: “NLL” = “non-lexical lifetimes”, referring to
a set of changes to the borrow checker taught it to accept more kinds of
valid code, making it much easier to write programs that compile.)

[1]: https://smallcultfollowing.com/babysteps/blog/2018/11/01/after-nll-interprocedural-conflicts/

tensorboard/data/server/run.rs
wchargin: This also seems reasonable to me. Sold.

> abstractions are supposed to be free in Rust so I figured hey maybe I can just get away with calling a callback closure on every event :)

[@chandlerc might contort that there are no zero-cost abstractions][1],
and that (Rust) closures have no run-time cost but do have human cost.
:-) But yeah, I’m okay with them here.

[1]: https://youtu.be/rHIkrotSwcc?t=9

wchargin (commented):
GitHub is still showing this in my queue; not sure why. Responded here
earlier:

  - <https://github.com/tensorflow/tensorboard/pull/4604/files#r564807244>
  - <https://github.com/tensorflow/tensorboard/pull/4604/files#r564812906>

tensorboard/data/server/run.rs
nfelt: Converted the free functions into methods.

Yeah, that article pretty much sums up the understanding I had come to as well.  The "View" thing is interesting, I could see that being fairly effective for a more complex situation but it definitely isn't lightweight.  Thanks for sharing!

tensorboard/data/server/run.rs
nfelt: No such thing as a free lunch after all huh 😿

nfelt (commented):
PTAL

> One trick is to copy the old and new binaries to `./bench-{old,new}`
> and then ask `hyperfine` to do a parameter sweep:
> 
> ```
> hyperfine '{tool} --logdir logs' -L tool ./bench-old,./bench-new
> ```
> 
> The main benefit is that it’ll give you a confidence interval around the
> relative performance, something like “old is 1.02(±0.06)× faster”.

Thanks!  Yeah, I realized after I had already run the first one that I could probably just copy the binaries into different places (unlike python...) but was too lazy to re-run it 😬 

> > I haven't modified the unit tests to cover this behavior because I was a bit unsure about how to go about it. My main thought was parameterizing `reload()` (or `RunLoader` at construction) in order to inject some kind of fake clock object, but I wasn't sure if there was a standard technique used for this. Open to suggestions.
> 
> That’s about right. The standard technique often involves doing so with
> type parameters such that it’s all optimized out: [`StageReservoir`](https://github.com/tensorflow/tensorboard/blob/a906f7191137c1816c0544c00432a308fc627d2f/tensorboard/data/server/reservoir.rs#L70) and
> its `ReservoirControl` do this. As to clock objects specifically: I’ve
> not done this, and I don’t know of common patterns, but I imagine that a
> simple `trait Clock { fn now(&self) -> Instant }` would suffice here,
> and let the test clock own an `Rc<Instant>` or `AtomicI32` or something
> that ticks forward each time that it’s called.
> 
> I’m okay with omitting dedicated tests for this. It seems unlikely to be
> subtly broken, and the failure mode is not catastrophic (read-only, and
> the right data still shows up, just later). Feel free to dive in if
> you’re interested, but no action requested.

Thanks, that technique makes sense!  I also agree with your last two points though, so if you're fine with it I'll omit the dedicated tests.

> > I did a little refactoring to facilitate making this change which I've broken out into separate commits.
> 
> Thanks; this was helpful. Feel free to merge however you want (rebase,
> squash, separate PRs).

I'm fine just squashing if you are (mostly broke out commits for ease of review).

tensorboard/data/server/run.rs
```diff
@@ -15,33 +15,30 @@ limitations under the License.
...
 pub struct RunLoader {
-    /// The earliest event `wall_time` seen in any event file in this run.
-    ///
-    /// This is `None` if and only if no events have been seen. Its value may decrease as new
-    /// events are read, but in practice this is expected to be the wall time of the first
-    /// `file_version` event in the first event file.
-    start_time: Option<WallTime>,
-
+    /// The run name associated with this loader. Used primarily for logging; the run name is
+    /// canonically defined by the map key under which this `RunLoader` is stored in LogdirLoader.
```
wchargin, line 40: nit if you’re already updating: `` `LogdirLoader` ``

wchargin (approved):
Feels nice to enable “Hide whitespace changes” without worrying that I’m
missing a load-bearing control flow bug!

tensorboard/data/server/run.rs
```diff
@@ -145,13 +158,16 @@ impl StageTimeSeries {
...
 
+/// Minimum time to wait between committing while a run is still loading.
+const COMMIT_INTERVAL: Duration = Duration::from_secs(5);
+
 impl RunLoader {
-    pub fn new() -> Self {
+    pub fn new(run: Run) -> Self {
         Self {
-            start_time: None,
+            run: run,
```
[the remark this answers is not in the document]
nfelt: Done.

tensorboard/data/server/run.rs
nfelt: I thought I got'em all but no. Thanks.

PR #4605, opened 2021-01-26
line chart: fix tick text clipping at axis edges

psybuzz:
In the new `?fastChart=true` line charts, tick labels on the x/y axes
show number values as strings. When the text is too close to the
left/right/top/bottom edge of the chart, the text can be visually
"cut off" in the middle.

This change automatically hides tick text if the tick is too close to
the edge of the chart.

Before
image

After
image

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
@@ -45,7 +45,7 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
           ></line>
         </ng-template>
         <ng-container *ngFor="let tick of getTicks()">
-          <g>
+          <g *ngIf="shouldShowMinorTick(tick, axis)">
             <text [attr.x]="textXPosition(tick)" [attr.y]="textYPosition(tick)">
               {{ getFormatter().formatTick(tick) }}
             </text>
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -162,6 +162,17 @@ export class LineChartAxisComponent {
...
+    // Hide the tick text when it is too close to the left/right edge.
+    const fractionOfFullExtent =
+      (tick - this.axisExtent[0]) / (this.axisExtent[1] - this.axisExtent[0]);
+    const margin = axis === 'x' ? 0.075 : 0.03;
+    if (fractionOfFullExtent < margin || fractionOfFullExtent > 1 - margin) {
+      return false;
+    }
+    return true;
+  }
+
```
stephanwlee, line 175: This feels too brittle. 1. There is no guarantee that the text would not bleed out even with these margins and (2) constants defined up feels too magical. Can we just start without this? If you find it important, use canvas to measure and properly quell the axis? 

Regardless, we need to update the test.

psybuzz: Sure, I can work on a proper fix + test, but I definitely prefer having this noticeable regression fixed before enabling the fastChart flag by default.

Regarding the approach, measurement sounds good to me.  To make sure we're on the same page, it sounds like the process will be:
- Get all the tick positions
- Measure the first and last tick text
- If the last tick goes beyond the right edge, hide it (vice versa for first tick/left edge)

Also, I'm not sure how to turn a non-draft PR back into a draft, so I'll just update the header title.

stephanwlee: Regression is not a functional but cosmetic. I think we should not be blocking something based on this mere minor "inconvenience".

PR #4609, opened 2021-01-26
Introduce a selector providing a RunId to Run map

psybuzz:
Existing ngrx selectors in the Run feature store offer ways to get a `Run`
from its associated `runId`, and a list of `Run`s for a specific
`experimentId`. However, it would be convenient for consumers who
wish to convert a batch of `runId`s into a batch of `Run`s without having
to do a combination of multiple `getRun()` calls.

This introduces a convenience selector that returns a map that
consumers can use for easy lookup.  For example,

```ts
// Container
<component [handyMap] = "store.select(getRunMap) | async">
```
```ts
// Component
@Input handyMap;
// ...
const run = this.handyMap.get(runId);
```

This change is intended to offer support for
https://github.com/tensorflow/tensorboard/pull/4592

### DIFF
--- a/tensorboard/webapp/runs/store/runs_selectors.ts
+++ b/tensorboard/webapp/runs/store/runs_selectors.ts
@@ -76,6 +76,17 @@ export const getRuns = createSelector(
 );
 
 /**
+ * Returns an Observable that emits a map from RunIds to Runs. Note: the keys
+ * do NOT necessarily correspond to the current route's runs.
+ */
+export const getRunMap = createSelector(
+  getDataState,
+  (state: RunsDataState): Map<string, Run> => {
+    return new Map(Object.entries(state.runMetadata));
+  }
+);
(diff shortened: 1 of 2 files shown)


stephanwlee (commented):
As I mentioned in the other PR, I don't think the container should pass entire map to a Component, but form a data structure that the UI Component understands with it. With that minor difference, this code LGTM.

Is there any places in metrics plugin where we can benefit from this selector?

stephanwlee (approved):

psybuzz:
> but form a data structure that the UI Component understands with it
👍 

> Is there any places in metrics plugin where we can benefit from this selector?
I did a quick scan, but didn't see an immediate place to drop it in.  I could imagine this being handy when a "run grouping" feature in the left sidepane comes together, though.

Thanks for the quick review!

PR #4613, opened 2021-01-26
proto: automatically update BUILD file

wchargin:
Summary:
This is the long-term equivalent to #4608: we teach the `update.sh`
script to also update the BUILD structure for individual proto targets.
All `tb_proto_library`s in `tensorboard/compat/proto/BUILD` are now
automatically generated.

If this starts to skew, we could add a `--build-only` flag to the update
script and run it in CI. I think it’s okay to omit that for now.

New code based off the script in the description of #4516, plus
additions to generate `protos_all` and integrate it into the build file.

Test Plan:
Running `update.sh` against TensorFlow commit b565087c5cd7 generates no
diff. Removing any of the deps in the BUILD file and re-running the
update script correctly reinstates the deps.

wchargin-branch: proto-update-deps


tensorboard/compat/proto/BUILD
```diff
@@ -12,6 +12,23 @@ filegroup(
...
+        "manual",
+        "notap",
+    ],
+    deps = [
+        ":protos_all_py_pb2",
+        "//tensorboard:expect_tensorflow_installed",
+    ],
+)
+
+# Rest of file AUTOMATICALLY GENERATED by tensorboard/compat/proto/update.sh.
```
psybuzz, line 30: Perhaps it might be good to add a line to this comment, saying not to modify the "AUTOMATICALLY GENERATED" substring, which our script explicitly checks for.  e.g. maybe a "do not modify anything below this line manually", and ensure the "AUTOMATICALLY GENERATED" string is below that line.

psybuzz (approved):
Overall this looks good, I ran it locally and it properly regenerated the individual proto targets (also picked up a recent comment change in TF https://github.com/tensorflow/tensorflow/commit/953dea002ee6136a99cec9f98852b4fda8e72101#diff-e94adf32e895c059f47c60b066d1a8723c8091cc3494214b488bbb0b14f718fb)

I should note that I do not consider myself familiar with the details of shell script syntax, awk, and sed.  For example, I'm not really sure I understand the use of '#' characters in:
`sed -e 's#.*compat/proto/\([^.]*\).*#\t\t":\1",#'`.  Please feel free to add another reviewer, who can offer a more proper review of `update.sh`, if needed.

Seeing that it works manually, properly sorts deps, and recreates with nice indentation, I believe it works, though.

tensorboard/compat/proto/BUILD
wchargin: I’m not sure that I understand. Are you saying that the “automatically
generated; do not modify” comment should itself have a comment saying
not to modify the comment, either? IMHO, this seems more confusing than
it’s worth. If it does become a problem, having a `--build-only` mode
and CI check would more robustly ensure that the structure is kept
intact, as mentioned in the PR description.

wchargin (commented):
> I do not consider myself familiar with the details of shell script
> syntax, awk, and sed […] Please feel free to add another reviewer

Okay, added R=nfelt.

> I'm not really sure I understand the use of '#' characters

The `#` characters are just delimiters: `s#foo#bar#`. You can use any
single character; `/` is a common choice, but if there are lots of
slashes in the pattern, it can be nicer to use a different symbol. This
way, you don’t need to escape the slashes, and it’s easier to see which
parts of the expression are which.

tensorboard/compat/proto/update.sh
```diff
@@ -41,4 +41,57 @@ find tensorboard/compat/proto/ -type f  -name '*.proto' -exec perl -pi \
...
+    printf '\tname = "protos_all",\n'
+    printf '\tsrcs = [],\n'
+    printf '\tvisibility = ["//visibility:public"],\n'
+    printf '\tdeps = [\n'
+    for f in *.proto; do
+      printf '\t\t":%s",\n' "${f%.proto}"
+    done | sort
+    printf '\t],\n'
+    printf ')\n'
+  } | expand -t4 >BUILD.new
```
nfelt, line 85: Optional, but I'd rather just use 4 spaces per tab inline.  It's simpler, more like the final output, and IMO it reads better to not have the lines start with `\t` which makes it harder for me to parse the word (i.e. I see `tname` and `tdeps`).

The one argument I could see maybe is the places where we have 8 spaces and it's not visually aligned, but in that case maybe just pipe those individual lines to `expand`?  That way the readability overall doesn't suffer.

nfelt (approved):

tensorboard/compat/proto/BUILD
wchargin: Sure; done.

tensorboard/compat/proto/update.sh
```diff
@@ -41,4 +41,57 @@ find tensorboard/compat/proto/ -type f  -name '*.proto' -exec perl -pi \
   -e 's|tensorflow\.TensorShapeProto|tensorboard.TensorShapeProto|g;' \
   {} +
 
+# Update dependency graph.
+(
+  cd tensorboard/compat/proto/
+
+  {
+    # Keep all non-`tb_proto_library` paragraphs from the build file...
```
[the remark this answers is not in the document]
wchargin: Done. (Indeed; this looks simpler.)

tensorboard/compat/proto/update.sh
wchargin: Hmm… okay. It’s easier for me to count `\t` than `␣␣␣␣`, but sure; done.

PR #4614, opened 2021-01-26
line chart: put fades around start&end of text

stephanwlee:
The GPU line chart does not account for space when creating tick string.
This can of course lead to clipping which is kind of fine except it can
sometimes lead to a confusing chart when the substring can make a lot of
sense. While this is mitigated by `<title>` and tooltips, we lose the
glance-ability.

To give user a hint that the tick string might be clipped, we now use
faded mask on the tick strings.

image

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
@@ -32,18 +32,15 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
   selector: 'line-chart-axis',
   template: `
     <div [class]="axis + '-axis'">
-      <svg class="minor">
+      <svg class="line">
         <ng-container *ngIf="axis === 'x'; else yAxisLine">
           <line x1="0" y1="0" [attr.x2]="domDim.width" y2="0"></line>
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -105,14 +103,48 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
...
+          #0000 0%,
+          #000 10%,
+          #000 90%,
+          #0000 100%
+        );
+      }
+
       .y-axis {
-        flex-direction: row;
+        flex-direction: row-reverse;
```
psybuzz, line 124: Just checking, is this "-reverse" needed?

psybuzz (approved):
Thanks for doing this!

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
stephanwlee: Yup, in case of y-axis, the text should appear before the line while for, x-axis, the line should appear before (on top of) the text. Since, DOM order wise, the line is placed before the text, we need the row-reverse.

PR #4618, opened 2021-01-27
data_compat: mark converted image and audio data

wchargin:
Summary:
In TensorFlow 1.x, image and audio data are represented with a separate
time series for each batch item (`input_image/image/0`, etc.). In
TensorFlow 2.x, a time series may include multiple batch items. This
patch adds a field to image and audio metadata protos to indicate that
they have been automatically converted (by `data_compat`) from TF 1.x to
TF 2.x. This way, downstream code can tell that a TF 1.x time series may
be part of a larger batch even though it only contains a single sample.

Since this new annotation is a proto field that is `true` only for
legacy data, summary metadata for TF 2.x summaries is unchanged.

Test Plan:
As end-to-end tests, verified that the images and audio dashboards both
still work with TF 1.x data, and checked with `grpc_cli` that the data
server returns summary metadata with non-empty `plugin_data.content`:

```
grpc_cli --channel_creds_type=local \
    --protofiles tensorboard/data/proto/data_provider.proto \
    call localhost:41905 TensorBoardDataProvider.ListBlobSequences \
    'plugin_filter { plugin_name: "audio" }'
```

wchargin-branch: mark-tf1x-image-audio

### DIFF
--- a/tensorboard/data/server/data_compat.rs
+++ b/tensorboard/data/server/data_compat.rs
@@ -15,6 +15,7 @@ limitations under the License.
 
 //! Conversions from legacy formats.
 
+use prost::Message;
 use std::convert::TryInto;
 use std::fmt::Debug;
 
@@ -262,8 +263,8 @@ impl SummaryValue {
             // form.
             (Some(md), _) if md.data_class != i32::from(pb::DataClass::Unknown) => Box::new(md),
             (_, Value::SimpleValue(_)) => blank(plugin_names::SCALARS, pb::DataClass::Scalar),
-            (_, Value::Image(_)) => blank(plugin_names::IMAGES, pb::DataClass::BlobSequence),
-            (_, Value::Audio(_)) => blank(plugin_names::AUDIO, pb::DataClass::BlobSequence),
+            (_, Value::Image(_)) => tf1x_image_metadata(),
+            (_, Value::Audio(_)) => tf1x_audio_metadata(),
             (Some(mut md), _) => {
                 // Use given metadata, but first set data class based on plugin name, if known.
                 match md.plugin_data.as_ref().map(|pd| pd.plugin_name.as_str()) {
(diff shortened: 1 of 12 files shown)


tensorboard/data/server/data_compat.rs
```diff
@@ -286,6 +287,50 @@ impl SummaryValue {
     }
 }
 
+fn tf1x_image_metadata() -> Box<pb::SummaryMetadata> {
+    let plugin_content = pb::ImagePluginData {
```
nfelt, line 291: Just wondering: in principle we could just construct the proto and serialize it once, and just copy the serialized bytes into each metadata, right?  I guess it's probably not worth it given that this is only initial metadata so we don't have to do this on every event.

tensorboard/data/server/data_compat.rs
```diff
@@ -286,6 +287,50 @@ impl SummaryValue {
...
+        converted_to_tensor: true,
+        ..Default::default()
+    };
+    let mut encoded_content = Vec::new();
+    plugin_content
+        .encode(&mut encoded_content)
+        // vectors are resizable, so should always be able to encode
+        .expect("failed to encode image metadata");
+
+    Box::new(pb::SummaryMetadata {
```
nfelt, line 301: Optional, but might be nice to have `blank_metadata_with_plugin_content()` to factor out the shared bit of these two helpers and `blank()`?  I guess I could go either way; up to you.

nfelt (approved):

tensorboard/data/server/data_compat.rs
wchargin: Indeed. You have to worry a bit about how/when that’s constructed and
accessed if you do that. The whole thing is easy if your computation can
be `const` (C++ `constexpr`), but anything involving allocation won’t be
(yet). More generally, Rust is picky about shared mutable globals—you
know, memory safety and all. [`once_cell::sync::OnceCell`] is a pretty
standard solution here, and ergonomic enough. But given that, as you
note, this is not on any critical path, I’d prefer to keep it simple.

[`once_cell::sync::OnceCell`]: https://docs.rs/once_cell/1.4.0/once_cell/sync/struct.OnceCell.html

tensorboard/data/server/data_compat.rs
wchargin: Sure; no strong preference. Works out to net ±0 LOC. :-)

PR #4620, opened 2021-01-28
graph: update metadata docs on plugin names

wchargin:
Summary:
These docs attempted to guess the future. They guessed wrong, but the
future is now here, so we can guess the present instead: much easier.

wchargin-branch: graph-metadata-docs


nfelt (approved):
I think I liked the old future better than the present 💁

wchargin:
> I think I liked the old future better than the present 💁

Always seems to happen, doesn’t it? :-)

As the person who both specified the old future and implemented the
present, I agree; I probably would have preferred a namespacing solution
over a proliferation of plugins. The problem is that the data model
isn’t as simple as “run graph/tag graph”. The existing case structure of
how the graphs plugin looks for data is much more complicated, and
I didn’t see a good way to actually set up the namespacing. 💁

nfelt:
> Always seems to happen, doesn’t it? :-)

Oof. Right in the feels.  Somehow the old future always made the present seem like it was gonna be simpler, huh.

PR #4621, opened 2021-01-28
graph: fix file upload control when no initial data loaded

### LINKED ISSUE #4556: graph: unable to upload a graph for viewing, when no graph in logdir
Repro:
- Find a .pbtxt graph file on the filesystem
- Launch TensorBoard at a log dir that does not contain the graph
- Open TensorBoard's Graph dashboard (after https://github.com/tensorflow/tensorboard/pull/4538, the sidebar should be visible)
- Click "Upload file" and open the .pbtxt from filesystem

Actual: the graph file does not load or get rendered
Expected: the .pbtxt should load

japie1235813:
Fix file upload control function (fixes #4556 #1613)

Previously the board render either the graphboard or no-data messages. Now it renders both and uses css display to control their appearances.
* Screenshots of UI changes
Select file:
selectFile
Normal loading:
Screen Shot 2021-01-28 at 1 52 56 PM

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_dashboard/tf-graph-dashboard.ts
+++ b/tensorboard/plugins/graph/tf_graph_dashboard/tf-graph-dashboard.ts
@@ -87,7 +87,10 @@ class TfGraphDashboard extends LegacyElementMixin(PolymerElement) {
         on-fit-tap="_fit"
         trace-inputs="{{_traceInputs}}"
       ></tf-graph-controls>
-      <div class="center" slot="center">
+      <div
+        class$="center [[_getGraphDisplayClassName(_selectedFile, _datasets)]]"
+        slot="center"
+      >
         <tf-graph-dashboard-loader
           id="loader"
           datasets="[[_datasets]]"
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/graph/tf_graph_dashboard/tf-graph-dashboard.ts
```diff
@@ -88,62 +88,58 @@ class TfGraphDashboard extends LegacyElementMixin(PolymerElement) {
...
+          selected-file="[[_selectedFile]]"
+          out-graph-hierarchy="{{_graphHierarchy}}"
+          out-graph="{{_graph}}"
+          out-stats="{{_stats}}"
+          progress="{{_progress}}"
+          hierarchy-params="[[_hierarchyParams]]"
+          compatibility-provider="[[_compatibilityProvider]]"
+        ></tf-graph-dashboard-loader>
+        <div
+          class$="no-data-message [[_showComponent(_selectedFile, _datasets, 'false')]]"
```
stephanwlee, line 104: Two related things.

1. Since we render `.no-data-message` and `.graphboard` all the time anyways, can we just set a class name on the outer DIV or whatever and use CSS to control whether to show `.graphboard` or `.no-data-message`?
2.  `_showComponent` is a misnomer. Sure, that is the logic you are implementing but that is not what the method is doing (it is returning a class name). Also, `_getContainerClassName` method should return a class name that describes a state instead of a generic verb, `.remove`.

stephanwlee (changes requested):
Change makes sense but would like changes to the method name and CSS organization. Please refer to the comment inline.

tensorboard/plugins/graph/tf_graph_dashboard/tf-graph-dashboard.ts
japie1235813: Thanks for the review! Please take another look. Thanks!

stephanwlee (approved):
LGTM. In the future, please press the reload icon there to re-request a review :) 

image

PR #4625, opened 2021-01-29
line chart: auto reset viewBox when scaleType

stephanwlee:
Wne user manually changes the viewBox (by scrolling or dragging), we
tend to honor the fixed viewBox even when data/visibility/etc...
changes. While this is reasonable in most cases, when scaleType changes,
it can lead to an illegal domain (e.g., log scale cannot render viewBox
that is negative) which is not great for UX.

As such. this change resets and overwrites the fixed viewBox when
user changes the viewBox.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
@@ -174,8 +174,13 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
       this.isMetadataUpdated = true;
     }
 
+    if (this.scaleUpdated) {
+      this.isViewBoxOverridden = false;
+    }
+
     this.isViewBoxChanged =
       this.isViewBoxChanged ||
+      this.scaleUpdated ||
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
```diff
@@ -174,8 +174,13 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
       this.isMetadataUpdated = true;
     }
 
+    if (this.scaleUpdated) {
+      this.isViewBoxOverridden = false;
```
psybuzz, line 178: Is it worth adding a test case to cover the scenario: "it updates the viewBox upon data changes, when a manual change is followed by a scale change"
?

psybuzz (approved):

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
stephanwlee: Done.

PR #4627, opened 2021-01-29
rust: add `crossbeam` dependency

wchargin:
Summary:
The [`crossbeam`] crate provides utilities for concurrent programming.
It’s best known for its [scoped threads, which enable safe sharing of
local variables][2p], often considered too dangerous in C++. For loading
runs concurrently, we’ll used scoped threads and also concurrent queues.

[2p]: https://matklad.github.io/2020/07/15/two-beautiful-programs.html
[`crossbeam`]: https://crates.io/crates/crossbeam

Test Plan:
It builds: `bazel build //third_party/rust:crossbeam`.

wchargin-branch: rust-dep-crossbeam


stephanwlee (approved):
Where do we intend to use the concurrency?

wchargin:
For loading runs concurrently within a logdir (i.e., every run loads
sequentially, but there is parallelism across runs).

wchargin:
If curious, you can see the [draft implementation][1] of concurrent run
loading, which uses Crossbeam in roughly the same manner that I expect
to use it for in TensorBoard.

[1]: https://github.com/wchargin/rustboard-server/blob/05550d7b356b94c7e0dea335030d4d11ab6510b8/src/logdir.rs#L139-L158

PR #4628, opened 2021-01-29
program: restore default assets zip provider

wchargin:
Summary:
This patch reverts the backward-incompatible changes to `program` made
in #4588, while retaining the new functionality in that change.
Specifically, we restore the calling convention of `TensorBoard`
(reordering the parameters was a breaking change, as was removing the
default value). For clarity, the data dependency to `webfiles.zip` lives
in a new Python module, `//tensorboard:assets_lib`. The `program` module
late-imports this only if no assets zip provider is given, in which case
the caller is expected to have a build dep on `:assets_lib`. The main
`//tensorboard` binary does so; `//tensorboard:dev` does not.

Test Plan:
Tested `//tensorboard`, `//tensorboard:dev`, and the Pip package in a
new virtualenv. There is no build path from `:dev` to the prod assets:

```
$ bazel query 'somepath(//tensorboard:dev, //tensorboard:webfiles.zip)'
INFO: Empty results
```

wchargin-branch: program-default-assets

### DIFF
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -88,8 +88,10 @@ class TensorBoard(object):
           plugins: A list of TensorBoard plugins to load, as TBPlugin classes or
             TBLoader instances or classes. If not specified, defaults to first-party
             plugins.
-          assets_zip_provider: A function that provides a zip file containing assets to
-            the application.
+          assets_zip_provider: A function that provides a zip file containing
+            assets to the application. If `None`, the default TensorBoard web
+            assets will be used. (If building from source, your binary must
+            explicitly depend on `//tensorboard:assets_lib` if you pass `None`.)
           server_class: An optional factory for a `TensorBoardServer` to use
             for serving the TensorBoard WSGI app. If provided, its callable
             signature should match that of `TensorBoardServer.__init__`.
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -77,19 +77,19 @@ class TensorBoard(object):
 
     def __init__(
         self,
-        assets_zip_provider,
         plugins=None,
+        assets_zip_provider=None,
         server_class=None,
         subcommands=None,
     ):
         """Creates new instance.
 
         Args:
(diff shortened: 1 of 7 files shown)


wchargin:
(Oops; meant to send to @nfelt since we’d talked about this.
@stephanwlee: feel free to review, but no obligation.)

tensorboard/program.py
```diff
@@ -104,6 +104,18 @@ def __init__(
...
 
             plugins = default.get_plugins()
+        if assets_zip_provider is None:
+            try:
+                from tensorboard import assets
+            except ImportError as e:
+                # `tensorboard.assets` is not a strict Bazel dep; clients are
+                # required to either depend on `//tensorboard:assets_lib` or
+                # pass a valid assets provider.
+                raise ImportError(
```
stephanwlee, line 114: No action required (or just relief):

I was a bit worried program.py would have hard dep on `webfiles.zip` but this of course make it better.

stephanwlee (approved):

tensorboard/program.py
```diff
@@ -77,19 +77,19 @@ class TensorBoard(object):
...
         """Creates new instance.
 
         Args:
-          assets_zip_provider: A function that provides a zip file containing assets to
-            the application.
           plugins: A list of TensorBoard plugins to load, as TBPlugin classes or
             TBLoader instances or classes. If not specified, defaults to first-party
             plugins.
+          assets_zip_provider: A function that provides a zip file containing assets to
+            the application.
```
nfelt, line 92: Maybe mention what happens if this is omitted/None?

nfelt (approved):
Thanks!

tensorboard/program.py
wchargin: Done.

tensorboard/program.py
wchargin: Yeah—it took a bit of thinking to come up with a build/import structure
that I liked, but I think that it turned out okay.

PR #4632, opened 2021-01-29
line chart: fix line chart not updating bug

stephanwlee:
When the line chart is initialized but never resized, currently our
`domDimension` and `viewBox` which is updated lazily after content is
initialized is not reflected to the template and their children. This is
because we use OnPush change detector and we never explicitly ask
Angular to detect changes.

This bug manifests especially when you spam on the tag filter and get
the timing right.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
@@ -187,16 +187,13 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
   }
 
   ngAfterViewInit() {
-    const dimPropChanged = this.readAndUpdateDomDimensions();
     this.initializeChart();
-    const viewboxPropChanged = this.updateLineChart();
+    this.updateLineChart();
 
     // After view is initialized, if we ever change the Angular prop that should propagate
     // to children, we need to retrigger the Angular change. Since we lazily update the
     // property, these may return false.
-    if (dimPropChanged || viewboxPropChanged) {
-      this.changeDetector.detectChanges();
-    }
+    this.changeDetector.detectChanges();
   }
 
   onViewResize() {
@@ -266,6 +263,8 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
     const callbacks: ChartCallbacks = {onDrawEnd: () => {}};
     let params: ChartOptions | null = null;
 
+    this.readAndUpdateDomDimensions();
+
     switch (rendererType) {
       case RendererType.SVG: {
         params = {
@@ -306,8 +305,7 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
     return getRendererType(this.preferredRendererType);
   }
 
-  private readAndUpdateDomDimensions(): boolean {
-    const prevDomDimension = this.domDimensions;
+  private readAndUpdateDomDimensions(): void {
     this.domDimensions = {
       main: {
         width: this.seriesView.nativeElement.clientWidth,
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
```diff
@@ -182,8 +182,16 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
...
     this.initializeChart();
-    this.updateLineChart();
+    const viewboxPropChanged = this.updateLineChart();
+
+    // After view is initialized, if we ever change the Angular prop that should propagate
+    // to children, we need to retrigger the Angular change. Since we lazily update the
+    // property, these may return false.
+    if (dimPropChanged || viewboxPropChanged) {
+      this.changeDetector.detectChanges();
+    }
```
psybuzz, line 194: iiuc, the idea is that when Angular calls our `ngAfterViewInit` lifecycle hook, it assumes we won't alter any props used in the template.  Since we do alter props used in the template, we need the explicit `detectChanges()` call.

In this PR, we are careful to only call detectChanges when necessary, but I think this adds a bit much complexity.
- L186 seem to be order-dependent, since `initializeChart` internally calls `readAndUpdateDomDimensions`, so it must come before L185.
- `updateLineChart` returns `ngStateUpdated`, but it only depends on the `viewBox` not any other state changes
- `ngStateUpdated` is not true when the viewBox changes to the 'fixed' or 'overridden' value.  Are we sure this is safe?

I think it would be a lot simpler to just call `detectChanges()` here in `ngAfterView` unconditionally, without checking whether the dim/viewbox actually changed.  If we ever updated the code to remove the `viewBox` from the template, or added new props to the template that can be updated from `ngAfterViewInit`, it would be less fragile to those changes as well.

wdyt?

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
```diff
@@ -182,8 +182,16 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
...
+    const dimPropChanged = this.readAndUpdateDomDimensions();
     this.initializeChart();
-    this.updateLineChart();
+    const viewboxPropChanged = this.updateLineChart();
+
+    // After view is initialized, if we ever change the Angular prop that should propagate
+    // to children, we need to retrigger the Angular change. Since we lazily update the
+    // property, these may return false.
+    if (dimPropChanged || viewboxPropChanged) {
+      this.changeDetector.detectChanges();
```
psybuzz, line 193: When the component initializes but never resizes, I'm curious, why do we not need to call `this.lineChart.resize(this.domDimensions.main);` here, just like in `onViewResize` on L201?

psybuzz: Ah, nvm I see now that `ChartImpl` calls its own `resize()` in its constructor, so LineChartComponent's `initializeChart` should take care of that.

tensorboard/webapp/widgets/line_chart_v2/line_chart_component_test.ts
```diff
@@ -237,6 +250,34 @@ describe('line_chart_v2/line_chart test', () => {
...
+      ],
+      seriesMetadataMap: {foo: buildMetadata({id: 'foo', visible: true})},
+      yScaleType: ScaleType.LINEAR,
+    });
+    fixture.detectChanges();
+
+    const grid = fixture.debugElement.query(By.directive(FakeGridComponent));
+    expect(grid.componentInstance.domDim).toEqual({
+      width: 150,
+      height: 70,
```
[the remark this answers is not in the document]
stephanwlee: If you look at the top of the file, on the `testable-comp`, we have hard coded the dimension (using `position: fixed`). Left a comment.

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
stephanwlee: Reverted most changes and only kept the unconditional `detectChanges`.

tensorboard/webapp/widgets/line_chart_v2/line_chart_component_test.ts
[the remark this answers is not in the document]
psybuzz: Ohh I missed that, thanks!

psybuzz (approved):

PR #4633, opened 2021-01-30
rust: add `rayon` dependency

wchargin:
Summary:
The [`rayon`] crate provides easy data parallelism: change `iter()` to
`par_iter()` and you’re off to the races—but not the data races, which
are prevented at compile time. We can use it to load runs concurrently.

[`rayon`]: https://crates.io/crates/rayon

Test Plan:
It builds: `bazel build //third_party/rust:rayon`.

wchargin-branch: rust-dep-rayon

### DIFF
--- a/tensorboard/data/server/Cargo.lock
+++ b/tensorboard/data/server/Cargo.lock
@@ -798,6 +798,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "rayon"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


wchargin:
I’m investigating this as an alternative to `crossbeam` (introduced in
#4627). I’ve used Crossbeam in the past, and it works. But I think that
Rayon could be easier to use, because:

  - it provides a higher level API, more like “do these things in
    parallel” than “chain these concurrent queues together”; and
  - it will be easier to reuse a pool of reloader threads across reload
    cycles, because of some details of the Crossbeam queues approach.

I can send out both patches and we can see which one we like better.

nfelt (approved):
Rayon looks slick and the PR is very clean, so LGTM.

wchargin:
Thanks!

PR #4635, opened 2021-02-01
Revert "introduce py_repl() rule for a convenient Bazel-aware python shell"

nfelt:
Reverts tensorflow/tensorboard#4580 since it breaks the internal sync.  Will fix and roll-forward.


stephanwlee (approved):

stephanwlee:
Impeccable code.

PR #4637, opened 2021-02-02
Instrument graph code with debug events

psybuzz:
This introduces a new utility under "tb_debug" that allows for
traced events throughout Polymer code to trigger a global function.
Additionally, this adds instrumentation within the Graph dashboard's
frontend that hooks up events that are important to collect.

By adding a `console.log` in `notifyActionEventFromPolymer`, one
can log each instance a Graph completed an interesting action,
and how long it took.

Googlers, see cl/355497543, which should be landed first. Also see
a test sync cl/355488337.


tensorboard/plugins/graph/tf_graph_common/BUILD
```diff
@@ -43,3 +44,18 @@ tf_ts_library(
...
+    testonly = True,
+    srcs = [
+        "common_test.ts",
+    ],
+    deps = [
+        ":tf_graph_common",
+        "//tensorboard/webapp/tb_debug",
+        "@npm//@types/jasmine",
+        "@npm//dagre",
+        "@npm//lodash",
```
stephanwlee, line 59: Do you need dagre and lodash here? It seems to violate strict_deps :\

tensorboard/plugins/graph/tf_graph_common/common_test.ts
```diff
@@ -0,0 +1,124 @@
...
+    );
+
+    const debugListenerSpy = spyOn(tb_debug, 'notifyActionEventFromPolymer');
+
+    try {
+      await tf_graph_loader.fetchAndConstructHierarchicalGraph(
+        mockTracker,
+        null /* remotePath */,
+        null /* pbTxtFile */
+      );
```
stephanwlee, line 110: I think you should fail the test if it does not throw?
```ts
expect(async () => await tf_graph_loader.fetchAndConstructHierarchicalGraph(
        mockTracker,
        null /* remotePath */,
        null /* pbTxtFile */
      )).toThrowError();
const firstArgs = debugListenerSpy.calls
        .allArgs()
        .map((args: any[]) => args[0]);
      expect(firstArgs).toEqual([
        {
          eventCategory: tb_debug.GRAPH_DEBUG_TIMING_EVENT_CATEGORY,
          eventAction: tb_debug.GraphDebugEventId.GRAPH_LOAD_FAILED,
          eventValue: jasmine.any(Number),
        },
      ]);
```

Above should work, I think.

tensorboard/webapp/tb_debug/BUILD
```diff
@@ -0,0 +1,13 @@
+load("//tensorboard/defs:defs.bzl", "tf_ts_library")
```
stephanwlee, line 1: Do you intend this package to be used by Angular app too? If not, I think it would be confusing to place it here. Move it to tensorboard/components?

tensorboard/plugins/graph/tf_graph_common/BUILD
psybuzz: Done!

tensorboard/plugins/graph/tf_graph_common/common_test.ts
psybuzz: Good catch.  The example given unfortunately did not work because the 2nd `expect` call synchronously fires after the first `expect`, but before the fetchAndConstructHierarchicalGraph finishes.

Done using a `await expectAsync(promise).toBeRejected()` pattern.

tensorboard/webapp/tb_debug/BUILD
psybuzz: Done.

stephanwlee (approved):

PR #4642, opened 2021-02-02
rust: upgrade `cargo-raze` to v0.8.0

wchargin:
Summary:
This patch upgrades our `cargo-raze` to v0.8.0, which is necessary to
support more recent versions of `rules_rust`, which we would like to
pull in for some quality-of-life fixes.

We only upgrade to v0.8.0 because v0.9.0 seems to have some issues
around dev dependencies (see google/cargo-raze#355), but this should be
sufficient for our needs.

Test Plan:
It suffices that everything still builds with both Cargo and Bazel.
For good measure, a quick test of `--load_fast` still seems to work.

wchargin-branch: raze-v0.8.0


wchargin:
Hmm… never mind. v0.8.0 _should_ be sufficient for our needs, but it
doesn’t include the bug fix in google/cargo-raze#331. So we’ll have to
wait until google/cargo-raze#355 is fixed and released.

No problem. Upgrading `rules_rust` was nice to have, but not blocking.

PR #4643, opened 2021-02-03
rust: add `reqwest` dependency

wchargin:
Summary:
The [`reqwest`] crate provides a high-level HTTP interface, similar in
spirit to Python’s `requests` package. It has built-in support for JSON
serialization and deserialization via `serde` (which we already use) and
is based on the `hyper` stack (which we also already use), so it should
fit in nicely. We’ll use it to make requests to GCS.

[`reqwest`]: https://crates.io/crates/reqwest

Test Plan:
It builds: `bazel build //third_party/rust:reqwest`.

wchargin-branch: rust-dep-reqwest


wchargin:
Oh, I should note: this dep introduces a bit of build log spam. That’s
already been fixed in [newer `rules_rust`][rr-548]. Newer `rules_rust`
made a [breaking change][rr-499] that would require us to also upgrade
to `cargo-raze` at least v0.8.0. In fact, we need v0.9.0 for [a fix to
that same change][cr-331]. But v0.9.0 also has a [breaking bug][cr-355]
that prevents us from upgrading. So, for now, we are stuck and must live
with the build spam until the bug is fixed so that we can upgrade the
thing so we can effect the entirely useless workspace name change so
that we can upgrade the other thing. *Give us this day our daily yaks…*

[cr-331]: https://github.com/google/cargo-raze/pull/331
[cr-355]: https://github.com/google/cargo-raze/pull/355
[rr-499]: https://github.com/bazelbuild/rules_rust/issues/499
[rr-548]: https://github.com/bazelbuild/rules_rust/pull/548

nfelt (approved):
> _Give us this day our daily yaks…_

image

PR #4644, opened 2021-02-03
rust: read from `CARGO_PKG_VERSION`

wchargin:
Summary:
Rust programs expect the `CARGO_PKG_VERSION` environment variable to be
defined at compile time; Cargo sets this to the package version from
`Cargo.toml`, and Bazel sets it to the `version` attribute of the
library. This patch moves our version number from being hard-coded in
`cli.rs` to being hard-coded in the BUILD file (and also in `Cargo.toml`
in either case). This way, we can use the version in more places, like
user agent strings without needing to keep more things in sync.

Test Plan:
Both `cargo run -- -h` and `bazel run . -- -h` show the correct version
number in the help information.

wchargin-branch: rust-env-version

### DIFF
--- a/tensorboard/data/server/BUILD
+++ b/tensorboard/data/server/BUILD
@@ -43,6 +43,7 @@ rust_library(
         "writer.rs",
     ] + _checked_in_proto_files,
     edition = "2018",
+    version = "0.3.0-alpha.0",
     deps = [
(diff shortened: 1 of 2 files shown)


wchargin:
*Sigh*. <https://github.com/bazelbuild/rules_rust/issues/573>

PR #4647, opened 2021-02-03
rust: add refresh token authentication

wchargin:
Summary:
Our GCS integration can now read standard Google credentials, read from
either the `GOOGLE_APPLICATION_CREDENTIALS` environment variable or the
well-known path under `~/.config/gcloud`. Refresh tokens are exchanged
for access tokens, which are cached until they’re about to expire.
Outgoing requests use a cached access token or fetch a new one if
necessary.

Test Plan:
Run TensorBoard with `--load_fast` and a `--logdir` pointing to a bucket
with restricted access. Note (with `RUST_LOG=rustboard=debug`) that
access tokens are only fetched once and are then reused.

Use `gcloud auth application-default login` to generate credentials.

Note also that if you cap access token lifetime to 30 seconds—

```diff
diff --git a/tensorboard/data/server/gcs/auth.rs b/tensorboard/data/server/gcs/auth.rs
--- a/tensorboard/data/server/gcs/auth.rs
+++ b/tensorboard/data/server/gcs/auth.rs
@@ -296,3 +296,3 @@ impl RefreshTokenCreds {
             access_token: res.access_token,
-            expires: Instant::now() + Duration::from_secs(res.expires_in),
+            expires: Instant::now() + Duration::from_secs(res.expires_in.min(30)),
         })
```

—(and change the margin from 60 seconds to 10 seconds) then access
tokens are still only refreshed *once* every 30 seconds, even among
concurrent loads.

wchargin-branch: rust-gcs-auth


wchargin:
> Run TensorBoard with `--load_fast` and a `--logdir` pointing to a bucket
> with restricted access.

Ping me internally if you want a pointer.

tensorboard/data/server/gcs/auth.rs
```diff
@@ -0,0 +1,300 @@
...
+    client_id: String,
+    client_secret: String,
+    refresh_token: String,
+}
+impl Debug for RefreshTokenCreds {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("RefreshTokenCreds")
+            .field("client_id", &self.client_id)
+            .field("client_secret", &self.client_secret)
+            .field("refresh_token", &format_args!("<redacted>"))
```
psybuzz, line 251: Just checking: if the `<redacted>` here is to cover the case when we accidentally save/log our refresh tokens somewhere, should we redact the client_secret too?

tensorboard/data/server/gcs/auth.rs
```diff
@@ -0,0 +1,300 @@
...
+        &self,
+        rb: RequestBuilder,
+        http: &HttpClient,
+        lifetime: Duration,
+    ) -> RequestBuilder {
+        if self.creds.anonymous() {
+            return rb;
+        }
+        let token = self.token.read().expect("failed to read auth token");
+        if let Some(t) = Self::check_expiry(&*token, lifetime) {
```
psybuzz, line 115: Thinking out loud:
At first glance, I was confused by what '&*' meant, but I think it makes sense now.  The type of `token` is a `RwLockReadGuard<Option<BoundedToken>>`, and what we really want to check is the `Option<BoundedToken>`.  Doing the deref '*' removes the guard, and the next '&' gets a reference to the Option, since we don't want to do a value copy.

If we simply did a `&token`, the code still compiles, but then we're just getting a reference to the guard, which is not the same as the reference to what it is guarding (the Option).

psybuzz (commented):
Overall, everything makes sense.  It was fun to read about things I wasn't familiar with: XDG_CONFIG_HOME, RwLocks in action, etc!

When I try to run
`bazel run tensorboard:dev --define=link_data_server=true -- --load_fast --logdir gs://tensorboard-bench-logs/edge_cgan --bind_all`

I get:
```
thread 'main' panicked at 'Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.', external/raze__tokio__1_0_2/src/runtime/blocking/shutdown.rs:51:21
```

Presumably, the blocking is because of the Reqwest HttpClient, but I'm not sure why Rust thinks blocking is not allowed here.  Have you seen this before?

tensorboard/data/server/gcs/auth.rs
```diff
@@ -0,0 +1,300 @@
...
+        };
+        match serde_json::from_reader(reader) {
+            Ok(refresh_token) => {
+                info!(
+                    "Using refresh token GCS credentials from {}",
+                    creds_file.display()
+                );
+                Credentials::RefreshToken(RefreshToken(refresh_token))
+            }
+            Err(e) => {
```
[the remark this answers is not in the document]
wchargin: Good question. Indeed: that [becomes a decode error][1], which we
propagate up (as an `io::ErrorKind::InvalidData` error, through the
mapping in `gcs::logdir::reqwest_to_io_error` from #4646).

[1]: https://docs.rs/reqwest/0.11.0/src/reqwest/async_impl/response.rs.html#242

wchargin (commented):
Thanks for the review!

wchargin:
> When I try to run
> `bazel run tensorboard:dev --define=link_data_server=true -- --load_fast --logdir gs://tensorboard-bench-logs/edge_cgan --bind_all`
> 
> I get:
> 
> ```
> thread 'main' panicked at 'Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.', external/raze__tokio__1_0_2/src/runtime/blocking/shutdown.rs:51:21
> ```
> 
> Presumably, the blocking is because of the Reqwest HttpClient, but I'm not sure why Rust thinks blocking is not allowed here. Have you seen this before?

Thanks much for catching this! Looks like it only happens in debug mode,
which is why I didn’t catch it. Fixed in #4646.

tensorboard/data/server/gcs/auth.rs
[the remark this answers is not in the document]
psybuzz: Interesting, thanks

psybuzz (approved):

tensorboard/data/server/gcs/auth.rs
wchargin: …but, you know, on second thought, it costs nothing to hide the client
“secret”, so I might as well, just in case my understanding is missing
some pieces. Done.

nfelt: Yeah - it's not secret for well-known installed apps like `gcloud` or our uploader, since it's not possible to hide it, but I think it could be a real secret for a generic `GOOGLE_APPLICATION_CREDENTIALS` path.

tensorboard/data/server/gcs/auth.rs
```diff
@@ -0,0 +1,298 @@
...
+                t.expires.saturating_duration_since(Instant::now())
+            );
+            t.access_token.authenticate(rb)
+        } else {
+            rb
+        }
+    }
+
+    /// Checks whether `token` represents a token that will still be valid for at least the given
+    /// `lifetime`, and if so returns a reference to the inner access token.
```
nfelt, line 139: Slightly non-obvious IMO from the name that `check_expiry` also unwraps the token.  Optional but maybe consider making it a method on BoundedToken (could replace `valid_at`) like `unwrap_if_valid_for()`?

wchargin: Okay, sure: changed as suggested.

PR #4649, opened 2021-02-04
rust: upgrade to latest `rules_rust`, `cargo-raze`

wchargin:
Summary:
We upgrade to latest `rules_rust` mainly for bazelbuild/rules_rust#572.
This requires upgrading `cargo-raze` to the newly released v0.9.2, plus
a herd of interdependent changes:

  - update to `rules_rust` latest
  - rename `@io_bazel_rules_rust` to `@rules_rust`
  - update Prost `protoc` wiring (not sure exactly why the old one
    doesn’t work anymore, but the new one is nicer, anyway)
  - update doc server to remove arch name, which also makes it work when
    compiled with `-c opt`

Test Plan:
Build changes only, so that `bazel test` passes and a `--load_fast`
smoke screen looks good suffices for correctness.

To see that the `openssl-sys` build spam is gone:

```
rm -rf bazel-out/k8-opt/bin/external/raze__openssl_sys__0_9_60  # ignore EACCES
bazel build @raze__openssl_sys__0_9_60//:openssl_sys
```

To test the doc server:

```
bazel run //tensorboard/data/server:rustboard_core_doc_server
```

…and navigate to the listed port.

wchargin-branch: rules-rust-20210203


wchargin:
Test sync passes: <http://cl/355526616>.

wchargin:
(WANT_LGTM=any)

tensorboard/data/server/cargo/BUILD.bazel
```diff
@@ -0,0 +1,246 @@
+"""
+@generated
+cargo-raze generated Bazel file.
+
+DO NOT EDIT! Replaced on runs of cargo-raze
+"""
```
stephanwlee, line 6: Is this going to be added to `.gitattributes`?

stephanwlee: Also, which command generated this file?

stephanwlee (approved):

tensorboard/data/server/cargo/BUILD.bazel
wchargin: Thanks for the review!

> Is this going to be added to `.gitattributes`?

I considered marking it as binary in `.gitattributes`. I like it as is
because it’s human-readable, it only reflects our direct deps (not the
whole lockfile) and so won’t get too much churn, and its diffs are
human-readable and the blame on it will be useful. I’ll merge as is just
for expediency, but I’d be happy to change course if you’d like that or
if it becomes a problem later on.

> Also, which command generated this file?

This is generated with the workflow described in `server/DEVELOPMENT.md`
under the “Adding third-party dependencies” section. The specific
command that I always pull from my shell history is

```
(tb && rm -rf third_party/rust/ && cd tensorboard/data/server && cargo fetch && cargo raze)
```

(where `tb` is a `cd` to the repository root), which is basically an
abbreviation of the steps in that doc.

The file is new because `cargo-raze` now supports “Cargo workspaces”,
which are a way to have multiple packages in one repo. (Think of it kind
of like Lerna in the NPM world.) This means that the global namespace of
`//third_party/rust` is replaced with a per-`Cargo.toml` namespace. We
only have one `Cargo.toml`, so it’s just a move for us—but now that we
have workspace support I might actually want to take advantage of it to
separate out the gRPC code from the core.

PR #4650, opened 2021-02-04
Revert "types: make `features` omittable by typing"

stephanwlee:
Reverts tensorflow/tensorboard#4648

This change breaks the sync as it can violate certain types defined internally. Until internal codes can take this change, this PR is reverted.


psybuzz (approved):
D:

PR #4651, opened 2021-02-04
Disable terser on polymer bundle in dev assets mode

psybuzz:
While the Angular bundle incremental build time was improved
with the `tensorboard:dev` mode, the Polymer 3 bundle still
relies on Rollup. After Terser was introduced in the build, the
Polymer incremental build time can take 30+ sec on top of the
TypeScript compilation time.

This change removes the Terser step from the Polymer bundling
phase, only for the dev assets mode.

### DIFF
--- a/tensorboard/components/BUILD
+++ b/tensorboard/components/BUILD
@@ -1,4 +1,4 @@
-load("//tensorboard/defs:defs.bzl", "tf_js_binary", "tf_ts_library")
+load("//tensorboard/defs:defs.bzl", "tf_dev_js_binary", "tf_js_binary", "tf_ts_library")
 
 package(default_visibility = ["//tensorboard:internal"])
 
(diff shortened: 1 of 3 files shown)


stephanwlee (commented):
Would it make more sense to use the tf_dev_js_binary on the polymer bundle?

psybuzz:
> Would it make more sense to use the tf_dev_js_binary on the polymer bundle?

Ideally yes, and I find the PR as-is quite unfortunate.  Sadly, I couldn't find any UMD-style bundle of Polymer (the library) itself.  Without that, `@polymer/polymer` imports from our components won't be resolved without the Rollup step first.

tensorboard/components/BUILD
```diff
@@ -1,4 +1,4 @@
-load("//tensorboard/defs:defs.bzl", "tf_js_binary", "tf_ts_library")
+load("//tensorboard/defs:defs.bzl", "tf_dev_js_binary", "tf_js_binary", "tf_ts_library")
```
stephanwlee, line 1: remove unused dev_js_binary.

tensorboard/components/BUILD
```diff
@@ -38,13 +38,33 @@ tf_ts_library(
...
 
+# A target for development only. By dropping 'compile', this skips the expensive
+# minification step.
+tf_js_binary(
+    name = "polymer3_lib_binary_no_shim_dev",
+    compile = False,
+    entry_point = ":polymer3_lib.ts",
+    deps = [":polymer3_ts_lib"],
+)
+
```
stephanwlee, line 49: This might belong in `tensorboard/webapp/dev_assets` maybe?

stephanwlee (commented):
Alternative is to build `@polymer/polymer` using rollup but we need to export polymer symbols outside of that bundle.

stephanwlee (approved):

tensorboard/components/BUILD
psybuzz: Done.

tensorboard/components/BUILD
psybuzz: Done.

psybuzz (commented):
Thanks for the review!

PR #4654, opened 2021-02-05
Reland "Disable terser on polymer bundle in dev assets mode (#4651)"

psybuzz:
**Status**: this PR was originally a roll-forward, but that was
abandoned after we reverted (#4655) the offending sync blocker.
This PR is now a re-land of the original #4651 with modifications.

Re-lands the change in #4651 that was reverted in #4655.
The original change used the `compile` option on `tf_js_binary`
as a signal to skip the Terser minification step. However, this
broke internally because `compile=False` has special semantics
in Google.

Now, the reland does the same minification skip when a new
rule attribute `dev_mode_only=True` is present. Internally,
this attribute is ignored.

Googlers, see test sync cl/355919112.

Before this is landed, cl/355772570 must be submitted first,
to fix syncing issues.

### DIFF
--- a/tensorboard/defs/defs.bzl
+++ b/tensorboard/defs/defs.bzl
@@ -26,7 +26,14 @@ def tensorboard_webcomponent_library(**kwargs):
     """Rules referencing this will be deleted from the codebase soon."""
     pass
 
-def tf_js_binary(name, compile, deps, visibility = None, **kwargs):
+def tf_js_binary(
+    name,
+    compile,
+    deps,
+    visibility = None,
+    dev_mode_only = False,
+    **kwargs
+):
     """Rules for creating a JavaScript bundle.
 
     Please refer to https://bazelbuild.github.io/rules_nodejs/Built-ins.html#rollup_bundle
@@ -53,7 +60,7 @@ def tf_js_binary(name, compile, deps, visibility = None, **kwargs):
         **kwargs
     )
 
-    if compile:
+    if not dev_mode_only:
         internal_result_name = name + "_terser_internal_min"
         terser_minified(
             name = internal_result_name,
--- a/tensorboard/defs/defs.bzl
+++ b/tensorboard/defs/defs.bzl
@@ -35,8 +35,7 @@ def tf_js_binary(name, compile, deps, visibility = None, **kwargs):
 
     # `compile` option is used internally but is not used by rollup_bundle.
     # Discard it.
-    internal_rollup_name = name + "_terser_internal_dbg"
-    internal_min_name = name + "_terser_internal_min"
+    internal_rollup_name = name + "_rollup_internal_dbg"
     rollup_bundle(
         name = internal_rollup_name,
         config_file = "//tensorboard/defs:rollup_config.js",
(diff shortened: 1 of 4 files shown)


wchargin (commented):
Hmm; my expectation is generally that tags are not load-bearing by Bazel
rules themselves, so this feels surprising to me. Might we instead pass
`dev_only = True` to the macro itself, and use that signal to control
the behavior? (Might need an internal `kwargs.pop`, but that can be made
without having to land this atomically.)

psybuzz:
Avoiding a load-bearing tag sounds good.  I've replaced it with a "dev_mode_only" attribute and added a `kwargs.pop` internally at cl/355772570.

psybuzz:
Actually, just in case there's anything controversial in this PR, it's probably better to immediately revert and unblock the sync.  If it's more comfortable, I have a revert PR out at #4655

tensorboard/defs/defs.bzl
```diff
@@ -53,7 +55,7 @@ def tf_js_binary(name, compile, deps, visibility = None, **kwargs):
         **kwargs
     )
 
-    if compile:
+    if kwargs.get("dev_mode_only") == True:
```
wchargin, line 58: I’m confused about a bunch of things, so maybe I’m just misunderstanding
it all. It looks to me like, as written, `dev_mode_only` will be passed
through to `rollup_bundle` as `kwargs`, which is an error. I expected to
see something like:

```python
def tf_js_binary(
    name,
    compile,
    deps,
    visibility = None,
    dev_mode_only = False,
    **kwargs
):
    ...  # now `rollup_bundle(..., **kwargs)` is okay
```

But it also looks to me like the `dev_mode_only` condition here is
backward, and the real `tf_js_binary` target doesn’t actually set
`dev_mode_only = True`, so I’m not convinced that I understand…

In any case, agreed to start with a rollback.

psybuzz: Sorry for the confusion, I made some mistakes and didn't actually upload the patch with the right changes.  You're totally correct that the `tf_js_binary` wasn't setting the flag, and didn't work.

Now that the offending PR was reverted, I've updated this PR to be a re-land of the original.  The original commit is the first one, and the 2nd commit in this branch contains new modifications.

The test sync cl has also been updated: cl/355900608

tensorboard/webapp/dev_assets/BUILD
```diff
@@ -40,19 +40,42 @@ tf_web_library(
     path = "/tb-webapp",
 )
 
+# By dropping 'compile', this skips the expensive minification step.
```
wchargin, line 43: We keep `compile` now; update comment?

(Or, is the comment needed now? This is in the `dev_assets` package, and
sets `dev_mode_only = True`, so is it self-explanatory? I defer to you.)

tensorboard/webapp/dev_assets/BUILD
```diff
@@ -40,19 +40,42 @@ tf_web_library(
...
+)
+
+# The index.html used for dev mode needs to use the polymer3 lib specifically
+# for dev mode.
 genrule(
     name = "rename_index_html",
     srcs = ["//tensorboard/webapp:index_polymer3.inlined.html"],
     outs = ["index.html"],
-    cmd = "cat $(SRCS) > $@",
+    cmd = "sed 's/polymer3_lib_binary/polymer3_lib_binary_dev/g' $(SRCS) > $@",
```
wchargin, line 70: …fwiw, this feels pretty sketch to me; it could have false-positive
matches, and if `index.html` changes structurally then this replacement
could become a silent no-op. But I see that this was approved in #4651,
so I won’t block on it.

It would make me very slightly more comfortable to anchor the pattern to
the `\.js` extension so that it’s at least idempotent.

tensorboard/defs/defs.bzl
```diff
@@ -54,19 +60,23 @@ def tf_js_binary(name, compile, deps, visibility = None, **kwargs):
...
     )
 
-    terser_minified(
-        name = internal_min_name,
-        src = internal_rollup_name,
-        config_file = "//tensorboard/defs:terser_config.json",
-        visibility = ["//visibility:private"],
-        sourcemap = False,
-    )
+    if not dev_mode_only:
```
wchargin, line 63: nit: Inverting the control flow here (`if dev_mode_only:`) both makes
the shorter block come first and avoids a negation, both of which are
generally considered easier to read.

Other than that, this logic looks good to me.

wchargin (approved):
LGTM once <http://cl/355772570> is submitted and a test sync passes.

psybuzz (commented):
Thanks for the review

tensorboard/webapp/dev_assets/BUILD
wchargin: > I also a comment in the *.html saying that a substitution may take place.

Good idea; appreciate it.

PR #4657, opened 2021-02-06
graph: instrument debug events for user actions

psybuzz:
The Graph code includes instrumentation to notify when timing events
occur for debugging. This change adds new instrumentation points
for non-timing related user-initiated actions, such as expanding a node,
performing a search, and toggling the "Trace Inputs" control.

Notably, these non-timing debug events have their own category,
and use the `eventLabel` optionally, instead of the `eventValue` field
required for timing-related events.

Test plan:
Manually checked that events are notified by adding a `console.log`
from inside `notifyActionEventFromPolymer()`, and observed the
output

image

Googlers, see test sync cl/355952339

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_common/util.ts
+++ b/tensorboard/plugins/graph/tf_graph_common/util.ts
@@ -20,41 +20,45 @@ import * as _ from 'lodash';
 import {notifyActionEventFromPolymer} from '../../../components/tb_debug';
 import {
   GraphDebugActionEventId,
-  GraphDebugEventId,
   GraphDebugTimingEventId,
   GRAPH_DEBUG_ACTION_EVENT_CATEGORY,
   GRAPH_DEBUG_TIMING_EVENT_CATEGORY,
-  isDebugTimingEventId,
 } from '../../../components/tb_debug/types';
 import {NodeStats, ProgressTracker} from './common';
 
 const ASYNC_TASK_DELAY = 20;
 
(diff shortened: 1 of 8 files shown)


tensorboard/plugins/graph/tf_graph_common/util.ts
```diff
@@ -19,26 +19,44 @@ import * as _ from 'lodash';
...
-  // milliseconds for a timing event.
-  eventValue?: number;
+interface DebugTimingEvent {
+  eventId: GraphDebugTimingEventId;
+  // An associated duration in milliseconds for a timing event.
+  eventValue: number;
 }
 
+interface DebugActionEvent {
+  eventId: GraphDebugActionEventId;
```
stephanwlee, line 40: Instead of `isDebugTimingEventId`, I think it makes more sense to make these discriminate-able by typing differently. 

```ts
interface DebugTimingEvent {
  timingId: GraphDebugTimingEventId;
  eventValue: number;
}

interface DebugActionEvent {
  actionId: GraphDebugActionEventId;
  eventLabel?: string;
}

export function notifyDebugEvent(debugEvent: DebugEvent) {
  if ('timingId' in debugEvent) {...}
  else {...}
  ...
}
```

Would that work?

stephanwlee (approved):
Thanks for very thorough logging.

psybuzz (commented):
Thanks for the review!

PR #4658, opened 2021-02-06
line chart: add trapezoid renderer

stephanwlee:
## Motivation 
When there is a NaN value in the line chart, we render a triangle in place of the NaN based on some business logic. However, when number of NaN becomes high (e.g., entire dataset has millions of NaNs), the new line chart mostly chokes because it simply creates way too many three.js objects. While we have not delved too deeply into it, it spends considerable amount of time in our updater as well as three.js renderer when having too many objects at least compared to one object with many vertices.

To supply the assessment with an imprecise data, when rendering 20 runs, we achieve about 15 FPS with lines (it draws many triangles for the thickness) while it achieves about 2 FPS (~600ms per frame) when drawing all points as triangles. 

## Solution
For consecutive NaNs, instead of drawing many number of triangles, we collapse them into a trapezoid (many dense triangles already looked like one).

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
@@ -69,6 +69,16 @@ export interface ObjectRenderer<CacheValue = {}> {
     paintOpt: CirclePaintOption
   ): CacheValue | null;
 
+  /**
+   * Draws or enqueues trapezoid drawing operations depending on a renderer.
+   */
+  createOrUpdateTrapezoidObject(
+    cached: CacheValue | null,
+    start: Point,
+    end: Point,
+    paintOpt: TrapezoidPaintOption
+  ): CacheValue | null;
+
   flush(): void;
 
   destroyObject(cachedValue: CacheValue): void;
@@ -86,6 +96,10 @@ export interface LinePaintOption extends PaintOption {
 
 export interface TrianglePaintOption extends PaintOption {
   size: number;
+}
+
+export interface TrapezoidPaintOption extends PaintOption {
+  width: number;
 }
 
 export interface CirclePaintOption extends PaintOption {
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
```diff
@@ -406,6 +418,41 @@ export class ThreeRenderer implements ObjectRenderer<CacheValue> {
...
+      new THREE.Vector2(start.x - width / 2, start.y - altitude / 3),
+      new THREE.Vector2(start.x, start.y + (altitude * 2) / 3),
+      new THREE.Vector2(end.x, end.y + (altitude * 2) / 3),
+      new THREE.Vector2(end.x + width / 2, end.y - altitude / 3),
+    ]);
+    shape.autoClose = true;
+    const geom = new THREE.ShapeBufferGeometry(shape);
+
+    if (!cached) {
+      const mesh = this.createMesh(geom as any, paintOpt);
```
psybuzz, line 444: `as any` can be dropped, since ShapeBufferGeometry extends BufferGeometry

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
```diff
@@ -88,6 +98,10 @@ export interface TrianglePaintOption extends PaintOption {
   size: number;
 }
 
+export interface TrapezoidPaintOption extends PaintOption {
+  width: number;
```
psybuzz, line 102: iiuc, the meaning of this `width` option is the length of the "slanted side" of the trapezoid?

I find it a bit strange to specify this slanted length, rather than the "height" or "altitude" of the trapezoid.  Although both are two ways to express what is functionally the same parameter, I think the altitude, used in the calculations below, is a bit more natural.

Otherwise if specifying the slanted length is preferred, could we rename this variable to `slantedSideLength`?

psybuzz (approved):
I'm not able to manually verify what it looks like visually, but code lgtm

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
stephanwlee: Ah, I did not mean to use "width" as the slant side length. Changed it to `altitude` instead.

PR #4663, opened 2021-02-08
line chart: scale aware scroll to zoom

stephanwlee:
Previously, in the interactive_view, we used simple arithmetic to
propose new viewExtent based on scroll magnitude and velocity, but it
did not account for the scale. As such, we are now using scale to
compute the next extent based on the position and speed of the cursor
and scroll.


psybuzz (approved):
Manually tested and it works well.  I think this fixes b/178748308?

PR #4668, opened 2021-02-09
Expand palette definitions to include all weights.

bmd3k:

  Some Google-internal libraries require some additional weights in our color palette to be defined. Otherwise they fail on compile.

  Expand our palette color definitions to include all weights that are multiples of 100.


tensorboard/webapp/theme/_tb_palette.scss
```diff
@@ -45,15 +49,19 @@ $tf-orange: (
   300: mat-color($mat-orange, 500),
   400: mat-color($mat-orange, 500),
   500: mat-color($mat-orange, 500),
+  600: mat-color($mat-orange, 700),
```
stephanwlee, line 52: For the tf-slate, 600 was matching 500 yet, here, it matches 700. Is this intended?

stephanwlee (approved):
I am a bit wary of explosion of all shades of colors in general but I assume there is a good usage of this. LGTM.

tensorboard/webapp/theme/_tb_palette.scss
bmd3k: No, it was not intentional. Now 600 matches 700 in both tf-slate and tf-orange.

PR #4670, opened 2021-02-09
Move the tfo-only images into comments.

MarkDaoust:
These will be uncommented in the site-import pipeline.

The goal here is to make the output less confusing when run in colab.

The code to make these replacements was:

```
import pathlib
import re

for path in pathlib.Path('.').rglob('*.ipynb'):
  content = path.read_text()
  new_content = re.sub('(<.*?tfo-display-only-on-site.*?/>)', r'<!-- \1 -->', content)
  if content != new_content:
    print(path)

  path.write_text(new_content)
```

see cl/356365619 , b/124898762 for context.

### DIFF
--- a/docs/get_started.ipynb
+++ b/docs/get_started.ipynb
@@ -238,7 +238,7 @@
         "id": "MCsoUNb6YhGc"
       },
       "source": [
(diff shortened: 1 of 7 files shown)


nfelt (approved):

MarkDaoust:
Thanks.

PR #4677, opened 2021-02-11
rust: use `bytes::Bytes` for proto fields

wchargin:
Summary:
This patch changes our proto bindings for `bytes` fields from `Vec<u8>`
to [`Bytes`]. The advantage is that they can be shared and sliced
cheaply. For example, the `ReadBlob` RPC no longer needs to clone the
full contents of the blob to refer to it after dropping the read lock,
and it also no longer needs to clone each sent chunk.

Correspondingly, we now represent blob sequences as `Vec<Bytes>` (seems
pretty sensible). Another option is `Vec<Arc<[u8]>>` here, but creating
an `Arc<[u8]>` from a `Vec<u8>` or `Bytes` [requires a full copy].

This patch is made mostly by updating `gen_protos_tool` and pushing
through the type changes.

[`Bytes`]: https://docs.rs/bytes/1.0.1/bytes/struct.Bytes.html
[requires a full copy]: https://stackoverflow.com/a/44641572

Test Plan:
For most of these changes, type-checking is sufficient; for the changes
to `ReadBlob`, unit tests are relevant, too.

wchargin-branch: rust-bytes


wchargin:
An alternate description of this change would be “avoid 2*n* bytes worth
of copies in `ReadBlob`, and plumb `bytes::Bytes` through as necessary”.
Most of the mass of this patch is just pushing types through, but the
actual payoffs are the `Bytes::clone` and `slice_ref(...)` in `server`.

wchargin:
@psybuzz: Friendly ping?

psybuzz (approved):
Sorry for the long delay, lgtm!

wchargin:
No problem; thanks!

PR #4686, opened 2021-02-16
timeseries: reflect smoothing setting in URL

stephanwlee:
This change adds "smoothing" to query parameter for convenience of a
user. When user changes to the scalars smoothing setting to a
non-default value, it will get reflected to the URL in the query
parameter (not hash)_as per our state persistence doc.

Do note that existing Polymer code uses the hash which can be convenient
to users but kind of makes our set up more convoluted as both Polymer
and Angular app will be overwriting it.


stephanwlee:
Strongly recommend enabling "Ignore whitespace changes" when reviewing.

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -325,9 +310,21 @@ const reducer = createReducer(
       state.pinnedCardToOriginal,
       state.cardStepIndex
     );
+
+    const hydratedSmoothing = hydratedState.metrics.smoothing;
+    const newSmoothing =
+      Number.isFinite(hydratedSmoothing) && hydratedSmoothing !== null
+        ? Math.max(0, Math.min(0.999, hydratedSmoothing))
```
psybuzz, line 317: Can we pull out the same hardcoded 0.999 value from 'webapp/metrics/views/right_pane/settings_view_component.ts' and reference `MAX_SMOOTHING_VALUE` here?

tensorboard/webapp/routes/core_deeplink_provider.ts
```diff
@@ -112,15 +122,22 @@ export class CoreDeepLinkProvider extends DeepLinkProvider {
     queryParams: SerializableQueryParams
   ): DeserializedState {
     let pinnedCards = null;
+    let smoothing = null;
     for (const {key, value} of queryParams) {
-      if (key === 'pinnedCards') {
-        pinnedCards = extractPinnedCardsFromURLText(value);
-        break;
+      switch (key) {
+        case 'pinnedCards':
```
psybuzz, line 128: Unrelated to PR, but since we are here, could we also extract `PINNED_CARDS_KEY`?

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -325,9 +310,21 @@ const reducer = createReducer(
...
+        : state.settings.scalarSmoothing;
+    const newSettings = {
+      ...state.settings,
+      scalarSmoothing: newSmoothing,
+    };
```
psybuzz, line 322: Maybe my suggestion leans a bit towards overoptimizing, but in the case when `newSmoothing` has not changed and we can reuse the old settings, should we return `state.settings` instead of a new object?

tensorboard/webapp/routes/core_deeplink_provider_test.ts
```diff
@@ -61,175 +66,222 @@ describe('core deeplink provider', () => {
...
-      store.overrideSelector(selectors.getUnresolvedImportedPinnedCards, []);
-      store.refreshState();
+        expect(queryParamsSerialized.length).toBe(0);
+      });
 
-      expect(queryParamsSerialized[queryParamsSerialized.length - 1]).toEqual(
-        []
-      );
-    });
+      it('deserializes the state in the URL without much sanitization', () => {
```
psybuzz, line 91: I'm not sure if we have any need to pass the raw, unsanitized smoothing value from the URL all the way to the reducer.  Would it be better to do the isFinite check, etc here in the core_deeplink_provider, given that `pinnedCards` is also sanitized here?

stephanwlee: I am inclined to keep it as is. That is because at the end of the day, reducer should do more input validations anyways since we really don't to spread validation logic all over the places and reducer is where the business logic should normally go.

Even without `Number.isFinite` call in the reducer and in the provider, since it is `null`able, we still need to do similar kind of guards in the reducer.

psybuzz: That's fair enough.  If this is our stance, perhaps some of the `pinnedCards` sanitization logic should really belong in its reducer as well, then (unrelated to this PR of course).

psybuzz (approved):

tensorboard/webapp/routes/core_deeplink_provider_test.ts
stephanwlee: In fact, I would have added the smoothing change validation logic in the reducer, too :)

PR #4689, opened 2021-02-18
Plumb --samples_per_plugin to Rust data server

psybuzz:
Introduces a `--samples-per-plugin` flag on the Rust data server, which
takes the `--samples_per_plugin` flag from standard TensorBoard's CLI.
At the moment, this change makes the value of `0` interpreted as
"create a reservoir with capacity 0", while traditionally the value would be
interpreted as "create a reservoir with unbounded capacity". This
discrepancy will be fixed in a separate PR.

Test plan:
Added unit tests and ran with `cargo test`.
Ran `bazel run tensorboard:dev --define=link_data_server=true -- --load_fast --logdir <logdir> --bind_all --samples_per_plugin=scalars=5,images=0` and observed that scalar charts that normally show lots of points now only show 5, while images do not appear at all.
image

Associated issue: "samples per plugin" item in the Rustboard task list #4422

### DIFF
--- a/tensorboard/data/server/cli.rs
+++ b/tensorboard/data/server/cli.rs
@@ -22,6 +22,7 @@ use std::io::{Read, Write};
 use std::net::{IpAddr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::sync::Arc;
 use std::thread;
 use std::time::{Duration, Instant};
 use tokio::net::TcpListener;
@@ -32,6 +33,7 @@ use crate::commit::Commit;
 use crate::logdir::LogdirLoader;
 use crate::proto::tensorboard::data;
 use crate::server::DataProviderHandler;
+use crate::types::PluginSamplingHint;
 
 use data::tensor_board_data_provider_server::TensorBoardDataProviderServer;
 
@@ -114,6 +116,17 @@ struct Opts {
     )]
     #[allow(unused)]
     no_checksum: bool,
+
+    /// Set explicit series sampling
+    ///
+    /// A comma separated list of plugin_name=num_samples pairs to explicitly specify how many
+    /// samples to keep per tag for the specified plugin. For unspecified plugins, series are
+    /// randomly downsampled to reasonable values to prevent out-of-memory errors in long running
+    /// jobs. Setting a plugin's num_samples to `0` will keep all samples for that plugin. For
+    /// instance, `--samples_per_plugin=scalars=500,images=0` keeps 500 events in each scalar series
+    /// and preserves all images.
+    #[clap(long)]
+    samples_per_plugin: Option<PluginSamplingHint>,
 }
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
@@ -174,7 +187,7 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Leak the commit object, since the Tonic server must have only 'static references. This only
     // leaks the outer commit structure (of constant size), not the pointers to the actual data.
     let commit: &'static Commit = Box::leak(Box::new(Commit::new()));
-
+    let psh_ref = Arc::new(opts.samples_per_plugin);
     thread::Builder::new()
         .name("Reloader".to_string())
         .spawn({
@@ -185,7 +198,7 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 // Create the logdir in the child thread, where no async runtime is active (see
                 // docs for `DynLogdir::new`).
                 let logdir = DynLogdir::new(raw_logdir).unwrap_or_else(|| std::process::exit(1));
-                let mut loader = LogdirLoader::new(commit, logdir, 0);
+                let mut loader = LogdirLoader::new(commit, logdir, 0, psh_ref);
                 // Checksum only if `--checksum` given (i.e., off by default).
                 loader.checksum(checksum);
                 loop {
--- a/tensorboard/data/server/cli.rs
+++ b/tensorboard/data/server/cli.rs
@@ -119,14 +119,13 @@ struct Opts {
 
     /// Set explicit series sampling
     ///
-    /// A comma separated list of plugin_name=num_samples pairs to explicitly specify how many
+    /// A comma separated list of `plugin_name=num_samples` pairs to explicitly specify how many
     /// samples to keep per tag for the specified plugin. For unspecified plugins, series are
     /// randomly downsampled to reasonable values to prevent out-of-memory errors in long running
-    /// jobs. Setting a plugin's num_samples to `0` will keep all samples for that plugin. For
-    /// instance, `--samples_per_plugin=scalars=500,images=0` keeps 500 events in each scalar series
-    /// and preserves all images.
-    #[clap(long)]
-    samples_per_plugin: Option<PluginSamplingHint>,
+    /// jobs. For instance, `--samples_per_plugin=scalars=500,images=0` keeps 500 events in each
+    /// scalar series and keeps none of the images.
+    #[clap(long, default_value = "", setting(clap::ArgSettings::AllowEmptyValues))]
+    samples_per_plugin: PluginSamplingHint,
 }
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
--- a/tensorboard/data/server/run.rs
+++ b/tensorboard/data/server/run.rs
@@ -81,7 +81,7 @@ struct RunLoaderData {
     time_series: HashMap<Tag, StageTimeSeries>,
 
     /// A map defining how many samples per plugin to keep.
-    plugin_sampling_hint: Arc<Option<PluginSamplingHint>>,
+    plugin_sampling_hint: Arc<PluginSamplingHint>,
 }
 
 #[derive(Debug)]
@@ -104,8 +104,22 @@ struct StageValue {
 impl StageTimeSeries {
     fn new(
         metadata: Box<pb::SummaryMetadata>,
-        plugin_sampling_hint: Arc<Option<PluginSamplingHint>>,
+        plugin_sampling_hint: Arc<PluginSamplingHint>,
     ) -> Self {
+        let data_class =
+            pb::DataClass::from_i32(metadata.data_class).unwrap_or(pb::DataClass::Unknown);
+        let capacity = StageTimeSeries::capacity(&metadata, plugin_sampling_hint);
+        Self {
+            data_class,
+            metadata,
+            rsv: StageReservoir::new(capacity),
+        }
+    }
+
+    fn capacity(
+        metadata: &Box<pb::SummaryMetadata>,
+        plugin_sampling_hint: Arc<PluginSamplingHint>,
+    ) -> usize {
         let data_class =
             pb::DataClass::from_i32(metadata.data_class).unwrap_or(pb::DataClass::Unknown);
         let mut capacity = match data_class {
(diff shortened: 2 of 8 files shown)


tensorboard/data/server/types.rs
```diff
@@ -89,13 +92,36 @@ impl Borrow<str> for Run {
...
+
+impl FromStr for PluginSamplingHint {
+    type Err = <usize as FromStr>::Err;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut result = HashMap::new();
+        for pair_str in s.split(',') {
+            let pair: Vec<_> = pair_str.split('=').collect();
+            if pair.len() != 2 {
+                error!("While parsing plugin sampling hint: {}", s);
+                std::process::exit(1);
```
wchargin, line 107: `FromStr` should be a pure function and should not exit or panic. It
should return an `Err` if parsing fails. Feel free to define your own
error type; something like:

```rust
#[derive(Debug, thiserror::Error)]
enum ParsePluginSamplingHintError {
    #[error("component should have exactly 1 \"=\" sign, but found {part:?}")]
    SyntaxError { part: String },
    #[error(transparent)]
    ParseIntError(#[from] std::num::ParseIntError),
}

impl FromStr for PluginSamplingHint {
    type Err = ParsePluginSamplingHintError;
    // ...
}
```

tensorboard/data/server/types.rs
```diff
@@ -138,4 +163,18 @@ mod tests {
         ];
         assert_eq!(actual, expected);
     }
+
+    #[test]
+    fn test_plugin_sampling_hint() {
```
wchargin, line 168: This test looks good; can we also test the cases where parsing fails?

tensorboard/data/server/types.rs
```diff
@@ -89,13 +92,36 @@ impl Borrow<str> for Run {
...
+#[derive(Debug)]
+pub struct PluginSamplingHint(pub HashMap<String, usize>);
+
+impl FromStr for PluginSamplingHint {
+    type Err = <usize as FromStr>::Err;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut result = HashMap::new();
+        for pair_str in s.split(',') {
+            let pair: Vec<_> = pair_str.split('=').collect();
+            if pair.len() != 2 {
```
wchargin, line 105: If I’m reading this right, it looks like `from_str("")` fails to parse.
IMHO it would be nice for that to successfully yield an empty map: it’s
more consistent, you could drop the `if samples_per_plugin:` on the
Python side, and the CLI parser could just set `default_value = ""`\*.
Thoughts?

\* Due to clap-rs/clap#1740, you may also need to explicitly indicate
that empty values are allowed for this argument, due to a misfeature of
Clap 3.x. Cf. [`gsutil.rs`] for an example. Hopefully this is fixed
before 3.0 final is released.

[`gsutil.rs`]: https://github.com/tensorflow/tensorboard/blob/e68b9646ce245749fd61cb28818284290581501f/tensorboard/data/server/gcs/gsutil.rs#L43

tensorboard/data/server/run.rs
```diff
@@ -99,15 +102,30 @@ struct StageValue {
...
             pb::DataClass::BlobSequence => 10,
             _ => 0,
         };
+        // Override the capacity using the plugin-specific hint.
+        if capacity > 0 {
+            if let Some(ref pd) = metadata.plugin_data {
+                if let Some(hint) = &*plugin_sampling_hint {
+                    if let Some(&num_samples) = hint.0.get(&pd.plugin_name) {
+                        // TODO(psybuzz): if the hint prescribes 0 samples, the reservoir should ideally
+                        // be unbounded. For now, it simply creates a reservoir with capacity 0.
```
wchargin, line 123: Two questions to simplify this call site:

 1. From this logic, it looks like a `PluginSamplingHint` with an empty
    hash map serves as a nice “zero value”. What do you think about
    making the plugin sampling hint required to remove one layer of
    conditionals?

 2. What do you think about encapsulating this logic into a method on
    `PluginSamplingHint`? Maybe:
    
    ```rust
    /// Gets the capacity for a given time series.
    pub fn capacity(&self, metadata: &pb::SummaryMetadata) -> usize;
    // ^ would include the default-values logic
    ```

    In some ways, this just moves the code around, but I think that you
    might find that the implementation simplifies, too. For instance, it
    enables you to just `return 0` from the `match` clause for `_`.

tensorboard/data/server/cli.rs
```diff
@@ -114,6 +116,17 @@ struct Opts {
     )]
     #[allow(unused)]
     no_checksum: bool,
+
+    /// Set explicit series sampling
+    ///
+    /// A comma separated list of plugin_name=num_samples pairs to explicitly specify how many
```
wchargin, line 122: nit: Rustdoc is Markdown; backticks around `` `plugin_name=num_samples` ``
and around `` `num_samples` `` below?

tensorboard/data/server/run.rs
```diff
@@ -525,4 +547,76 @@ mod test {
...
+        )?;
+        f1.write_graph(
+            Step(2),
+            WallTime::new(1237.0).unwrap(),
+            Bytes::from_static(b"<sample model graph>"),
+        )?;
+        // flush, so that the data's there when we read it
+        f1.into_inner()?.sync_all()?;
+
+        let plugin_sampling_hint = PluginSamplingHint::from_str("scalars=2").unwrap();
```
wchargin, line 585: More idiomatic to write `"scalars=2".parse().unwrap()` and not bother
bringing `FromStr` into scope. The [`parse`] method is target-typed,
like `collect` on iterators; it works for any target type that
implements `FromStr`.

[`parse`]: https://doc.rust-lang.org/std/primitive.str.html#method.parse

tensorboard/data/server/run.rs
```diff
@@ -525,4 +547,76 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn test_plugin_sampling_hint() -> Result<(), Box<dyn std::error::Error>> {
```
wchargin, line 552: Nice test.

tensorboard/data/server/run.rs
```diff
@@ -525,4 +547,76 @@ mod test {
...
+
+        let runs = commit.runs.read().expect("read-locking runs map");
+        let run_data: &commit::RunData = &*runs
+            .get(&run)
+            .expect("looking up data for run")
+            .read()
+            .expect("read-locking run data map");
+
+        let scalar_ts = run_data.scalars.get(&tag).unwrap();
+        let scalar_ts_values = scalar_ts.valid_values().collect::<Vec<_>>();
```
wchargin, line 609: Since you only care about the count, you can use `.count()` instead of
`.collect()`.

tensorboard/data/server/run.rs
```diff
@@ -525,4 +547,76 @@ mod test {
...
+
+        // Points for plugins specified in the sampling hint should be downsampled.
+        assert!(scalar_ts_values.len() == 2);
```
wchargin, line 612: Maybe use `assert_eq!(..., 2)` for better error messages?

wchargin (commented):
This looks excellent. Apart from the error handling for the parser, I’m
quite happy with this; nice work.

tensorboard/data/server/run.rs
psybuzz: > What do you think about making the plugin sampling hint required to remove one layer of conditionals?

Sounds good!

> What do you think about encapsulating this logic into a method on `PluginSamplingHint`?

Hm, if we extracted the logic into a method that accepted `metadata`, it would be able to compute the capacity entirely, so there would be less reason to keep the `DataClass` `match` around in `StageTimeSeries::new`.  In that case, `PluginSamplingHint::capacity(metadata)` should contain all the capacity logic.

If the concern is that the capacity logic is a bit much to have in the constructor, would it suffice to just move the code into a method `StageTimeSeries::capacity`?

aside: also switched the `if capacity > 0` to `if data_class != Unknown`, since even if a known data class had 0 default capacity, the user hint should still override.

psybuzz (commented):
Ready for another look

tensorboard/data/server/run.rs
wchargin: > Hm, if we extracted the logic into a method that accepted `metadata`, it would be able to compute the capacity entirely, so there would be less reason to keep the `DataClass` `match` around in `StageTimeSeries::new`. In that case, `PluginSamplingHint::capacity(metadata)` should contain all the capacity logic.

Agreed.

> If the concern is that the capacity logic is a bit much to have in the constructor, would it suffice to just move the code into a method `StageTimeSeries::capacity`?

I’m not too worried by the number of lines of code in the constructor.
I was wondering whether it makes sense from a conceptual standpoint that
the plugin sampling hint should know how to determine the capacity. For
instance, that would provide a natural place to unit-test that logic if
we so desired.

That said, I don’t feel strongly about it; the question was genuine.
This is fine with me, and I do think that having it in a separate
function (as you’ve done) is nice, because you can return early.

tensorboard/data/server/run.rs
```diff
@@ -161,12 +188,15 @@ impl StageTimeSeries {
...
-    pub fn new(run: Run) -> Self {
+    pub fn new(run: Run, plugin_sampling_hint: Arc<PluginSamplingHint>) -> Self {
         Self {
             run,
             files: BTreeMap::new(),
             checksum: true,
-            data: RunLoaderData::default(),
+            data: RunLoaderData {
+                plugin_sampling_hint,
+                ..RunLoaderData::default()
```
wchargin, line 198: Do you think that it would make sense to remove the `Default`
implementation and instead provide `RunLoaderData::new` that requires
an `Arc<PluginSamplingHint>`? No strong preference, and happy to
defer to you; just curious for your thoughts.

tensorboard/data/server_ingester.py
```diff
@@ -92,13 +102,19 @@ def start(self):
...
+        samples_per_plugin = ",".join(sample_hint_pairs)
+
         args = [
             server_binary,
             "--logdir=%s" % (self._logdir,),
             "--reload=%s" % reload,
             "--port=0",
             "--port-file=%s" % (port_file_path,),
             "--die-after-stdin",
+            "--samples-per-plugin=%s" % samples_per_plugin,
```
wchargin, line 117: I’m noticing that while this works fine with the data server at head, it
will fail for the v0.3.0 data server (no such flag), which is what’s
currently released. What do you think about either:

  - only including this flag when `samples_per_plugin` is non-empty, as
    you had previously? then `--load_fast` would still be broken with
    v0.3.0 with explicit `--samples_per_plugin`, but at least that’s not
    in the default path; or

  - explicitly gating the flag on `tensorboard_data_server.__version__`?
    you could use `pkg_resources.parse_version` to ensure that it’s at
    least `0.4.0a0` (cf. `version_test.py`)

I appreciate that this kind of dependency constrant is less than ideal,
and can think about ways to improve it in the general case.

wchargin (approved):

tensorboard/data/server_ingester.py
psybuzz: > only including this flag when samples_per_plugin is non-empty, as you had previously?

Done.

Agreed that it's non ideal, but hopefully the `--undefok` Clap may go through before we do the next release of `tensorboard-data-server`.

Explicitly gating the flag on version # would work for us, but I'm not sure what the story would be for users writing their own custom data server using it via `TENSORBOARD_DATA_SERVER_BINARY`.

Besides `--undefok`, the only thing I have thought of is trying to detect support for the flag by running `<binary> --help` and parsing the output to see whether `--samples-per-plugin` is there.  Which is also very non-ideal!

tensorboard/data/server/run.rs
psybuzz: Gotcha.  In that case, keeping the logic here in `StageTimeSeries` still seems preferable than moving to `PluginSamplingHint::capacity` imo, so I'll keep as-is for now.

psybuzz (commented):
Thanks for the review!

tensorboard/data/server_ingester.py
wchargin: > Explicitly gating the flag on version # would work for us, but I'm not sure what the story would be for users writing their own custom data server using it via `TENSORBOARD_DATA_SERVER_BINARY`.

Ah, good point, yes. You could invoke `rustboard --version` and see what
it prints, which is not *too* hard but certainly less convenient.

> Besides `--undefok`, the only thing I have thought of is trying to detect support for the flag by running `<binary> --help` and parsing the output to see whether `--samples-per-plugin` is there. Which is also very non-ideal!

Also a good point. Certainly not ideal, and only really works for
feature additions (e.g., would be harder to detect if we add support for
something like `--samples-per-plugin=scalars=100,images=all`).
But I don’t mind it that much if we want to go that route.

wchargin (approved):
LGTM; thanks!

Issue #4685, opened 2021-02-13
TB2.4+ POST to scalar_multirun omits XSRF header, blocking use of TensorBoard under JupyterLab


cliffwoolley:
### Diagnostics

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version e43767ef2b648d0d5d57c00f38ccbd38390e38da

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=8, micro=5, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): posix.uname_result(sysname='Linux', nodename='xpl-dvt-52', release='4.15.0-20-generic', version='#21-Ubuntu SMP Tue Apr 24 06:16:15 UTC 2018', machine='x86_64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: False
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
WARNING: no installation among: ['tb-nightly', 'tensorboard', 'tensorflow-tensorboard']
WARNING: no installation among: ['tensorflow', 'tensorflow-gpu', 'tf-nightly', 'tf-nightly-2.0-preview', 'tf-nightly-gpu', 'tf-nightly-gpu-2.0-preview']
INFO: installed: tensorflow-estimator==2.4.0

--- check: tensorboard_python_version
INFO: tensorboard.version.VERSION: '2.4.1'

--- check: tensorflow_python_version
2021-02-12 23:29:05.329875: I tensorflow/stream_executor/platform/default/dso_loader.cc:49] Successfully opened dynamic library libcudart.so.11.0
INFO: tensorflow.__version__: '2.4.0'
INFO: tensorflow.__git_version__: 'unknown'

--- check: tensorboard_data_server_version
INFO: no data server installed

--- check: tensorboard_binary_path
INFO: which tensorboard: b'/usr/local/bin/tensorboard\n'

--- check: addrinfos
socket.has_ipv6 = True
socket.AF_UNSPEC = <AddressFamily.AF_UNSPEC: 0>
socket.SOCK_STREAM = <SocketKind.SOCK_STREAM: 1>
socket.AI_ADDRCONFIG = <AddressInfo.AI_ADDRCONFIG: 32>
socket.AI_PASSIVE = <AddressInfo.AI_PASSIVE: 1>
[...]

PR #4691, opened 2021-02-19
http_client: add XSRF header to non-GET requests

wchargin:
Summary:
All POST, PUT, and DELETE requests sent through `TBHttpClient` now
include a header `X-XSRF-Protected: 1`. The name and value are constant;
this is not a XSRF token. Instead, servers can determine from the
presence of this non-standard header that it was sent through JavaScript
(rather than a form) and thus is subject to the same-origin policy.

This is currently intended for Google-internal surfaces, and is
unrelated to #4685.

Test Plan:
Unit tests included. Test sync passes: <http://cl/358305731>.

wchargin-branch: http-xsrf-header


tensorboard/webapp/webapp_data_source/tb_http_client_test.ts
```diff
@@ -103,6 +104,39 @@ describe('TBHttpClient', () => {
     });
   });
 
+  describe('XSRF header', () => {
+    it('is not attached to GET requests', () => {
```
stephanwlee, line 108: we tend to drop `is` as it becomes very repetitive without much information (and some reporters naturally places `is` sometimes). So, `it('attaches to non-GET requests'` would be good.

tensorboard/webapp/webapp_data_source/BUILD
```diff
@@ -64,12 +64,14 @@ tf_ng_module(
...
         ":http_client",
         ":http_client_testing",
+        "//tensorboard/webapp/angular:expect_angular_common_http",
         "//tensorboard/webapp/angular:expect_angular_core_testing",
         "//tensorboard/webapp/angular:expect_ngrx_store_testing",
         "//tensorboard/webapp/app_routing:app_root",
         "//tensorboard/webapp/core:types",
         "//tensorboard/webapp/feature_flag/store",
         "//tensorboard/webapp/feature_flag/store:types",
+        "@npm//@angular/common",
```
stephanwlee, line 74: Do you need this given that you have `expect_angular_common_http`?

stephanwlee (approved):

tensorboard/webapp/webapp_data_source/tb_http_client_test.ts
wchargin: > `it('attaches to non-GET requests'`

But “XSRF header attaches to non-GET requests” isn’t right; it’s the
client that does the attaching, not the header itself.

If you just don’t like the verb _is_, I suppose that I could write
something like “XSRF header appears on POST requests”. But “attaches to
non-GET requests” is tested in the other test cases: what we’re testing
here is that it is _not_ attached to _GET_ requests.

I don’t see a clear way to phrase that without _is not_ or _does not_,
and I’m not sure that I really understand the problem with _is_ in the
first place, so I’ve left this as is. Let me know if I misunderstand.

> some reporters naturally places `is` sometimes

That doesn’t make much sense to me, since often we put other verbs
there: e.g., “does not clobber unrelated headers” below. Any reporter
that changes “XSRF header does not […]” to “XSRF header is does not […]”
seems to broken to be worth designing around, imho.

tensorboard/webapp/webapp_data_source/BUILD
wchargin: Huh; I was getting strict-deps errors, and adding this fixed it, but
I can’t seem to reproduce that now. Removed and test-synced; thanks.

tensorboard/webapp/webapp_data_source/tb_http_client_test.ts
stephanwlee: But often `it` refers to the subject under test. So the complete sense would be `TBHttpClient XSRF header appears on POST requests` which is very awkward because of describe in between. 

Alternative could be `TBHttpClient for XSRF attaches custom header to non-GET requests`.

wchargin: Okay: I don’t see what’s wrong with “`TBHttpClient` XSRF header appears
on POST requests” (reads like fine English to me), but I’ve taken the
structure of your suggestion in case it brings you joy. :-)

PR #4695, opened 2021-02-22
line chart: enable GPU line chart by default

stephanwlee:
This change enables the GPU line chart, the new line chart
implementation, on by default in the time series dashboard. We will, in
the future*, remove the old implementation.

To disable the new implementation, you can pass `?fastChart=false` in the query param.

*: Probably after major/minor release of TensorBoard.


bmd3k (approved):
Congrats!

PR #4699, opened 2021-02-23
timeseries: implement runs x-axis splitter

stephanwlee:
Please refer to #4696 for definition of the monotonic step and what
problem we are trying to solve.

This change adds integration of the user setting to detect non-monotonic
step changes and split runs into sub runs. While this change is only
applied to the new GPU line chart, it can easily transfer to the legacy
one.

image

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
@@ -55,6 +55,7 @@ import {
   getCardMetadata,
   getCardTimeSeries,
   getMetricsIgnoreOutliers,
+  getMetricsScalarPartitionNonMonotonicX,
   getMetricsScalarSmoothing,
   getMetricsTooltipSort,
   getMetricsXAxisType,
@@ -65,12 +66,14 @@ import {CardRenderer} from '../metrics_view_types';
 import {getTagDisplayName} from '../utils';
 import {LegacySeriesDataList} from './scalar_card_component';
 import {
+  PartialSeries,
+  PartitionedSeries,
   ScalarCardDataSeries,
   ScalarCardPoint,
   ScalarCardSeriesMetadataMap,
   SeriesType,
 } from './scalar_card_types';
-import {getDisplayNameForRun} from './utils';
+import {getDisplayNameForRun, partitionSeries} from './utils';
 
 type ScalarCardMetadata = CardMetadata & {
   plugin: PluginType.SCALARS;
@@ -98,11 +101,6 @@ function areSeriesDataListEqual(
       })
     );
   });
-}
-
-interface PartialSeries {
-  runId: string;
-  points: ScalarCardPoint[];
 }
 
 function areSeriesEqual(
@@ -315,7 +313,7 @@ export class ScalarCardContainer implements CardRenderer, OnInit {
       return JSON.stringify(['smoothed', seriesId]);
     }
 
-    this.dataSeries$ = partialSeries$.pipe(
+    const normalizedPartialSeries$ = partialSeries$.pipe(
       combineLatestWith(this.store.select(getMetricsXAxisType)),
       // Normalize time and, optionally, compute relative time.
       map(([partialSeries, xAxisType]) => {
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -332,16 +342,84 @@ export class ScalarCardContainer implements CardRenderer, OnInit {
...
+        }
+        return partitionedRunsData;
+      }),
+      shareReplay(1)
+    );
+
+    this.dataSeries$ = partitionedRuns$.pipe(
+      combineLatestWith(this.store.select(getMetricsScalarSmoothing)),
+      switchMap(([runsData, smoothing]) => {
+        const cleanedRunsData = runsData.map(({id, points}) => ({id, points}));
```
psybuzz, line 415: The variable naming is starting to get really complex, imo.  If `partitionedRunsData` is a list, a name like shouldn't it contain `partitionedRunsDatum`s?

L327 `runsData` is a `{runId: string; points: ScalarCardPoint[]}`
L357 `runsData` is a `{id: string; runId: string; points: ScalarCardPoint[]}`
L370 `currentPartition` is a `{id: string; runId: string; points: ScalarCardPoint[]}`
L415 `cleanedRunsData` is a `{id: string, points: ScalarCardPoint[]}[]`
L441 `runIdAndPoints` is a `LineChartDatum`

iiuc, there are 5 different forms of "cleaning" that should occur, and more than one could be interpreted as "cleanedRunsData":
{runId,points} =>
{runId,pointsWithWallTime} =>
{runId,pointsWithWallTime,partitionIndex/size} ==>
{id,runId,(un)smoothedPointsWithWallTime,partitionIndex/size} ==>
{id,runId,(un)smoothedPointsWithWallTime}

It is fine with me to use the same name to refer to more than one of those 5 types, but could we make the names overall more consistent?

psybuzz (commented):
Since the partitioning logic doesn't really have any difficult dependencies, could we extract it into a pure function and add a unit test?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_types.ts
```diff
@@ -52,3 +52,16 @@ export interface ScalarCardPoint extends Point {
...
 
 export type ScalarCardDataSeries = DataSeries<ScalarCardPoint>;
+
+export interface PartialSeries {
+  seriesId: string;
+  points: ScalarCardPoint[];
+}
+
+export interface PartitionedSeries {
+  seriesId: string; // dataset id.
```
psybuzz, line 62: iiuc, `PartitionedSeries.seriesId` and `PartialSeries.seriesId` are **not interchangeable**.  For `PartialSeries`, the seriesId must always correspond to a runId, while for `PartitionedSeries`, it does not.  e.g.
- `runSelectionMap.get(partialSeries.seriesId)` is ok
- `runSelectionMap.get(partitionedSeries.seriesId)` is not

I'm wondering how we might avoid someone in the future calling a runId-function on a non-runId.  Should this PR rename the `PartialSeries.seriesId` into `runId`?

tensorboard/webapp/metrics/views/card_renderer/utils_test.ts
```diff
@@ -38,4 +39,134 @@ describe('metrics card_renderer utils test', () => {
...
+            seriesId: 'a',
+            points: buildPoints([0, Infinity, NaN, 1, -1]),
+          },
+        ]);
+
+        expect(actual).toEqual([
+          {
+            seriesId: '["a",0]',
+            runId: 'a',
+            points: buildPoints([0, Infinity, NaN, 1]),
```
psybuzz, line 138: Ah, nice test case, I hadn't considered partitioning with NaN/Inf.  Seems to behave as I would expect.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -342,7 +340,31 @@ export class ScalarCardContainer implements CardRenderer, OnInit {
...
+        this.store.select(getMetricsScalarPartitionNonMonotonicX)
+      ),
+      map<[PartialSeries[], boolean], PartitionedSeries[]>(
+        ([normalizedSeries, enablePartition]) => {
+          if (enablePartition) return partitionSeries(normalizedSeries);
+
+          return normalizedSeries.map((series) => {
+            return {
+              ...series,
+              runId: series.seriesId,
```
psybuzz, line 357: Upon first read, I think it is a little non-obvious that converting `seriesId` back into `runId` is valid.  It is valid here, because `normalizedSeries$` returns an observable of `PartialSeries`, and we are implicitly guaranteeing that seriesId === runId for `PartialSeries` (but not for `PartitionedSeries`).

See comment below.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_types.ts
stephanwlee: As per refactor PR, changed it to `runId`.

tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -29,3 +30,52 @@ export function getDisplayNameForRun(
...
+ * steps.
+ */
+export function partitionSeries(series: PartialSeries[]): PartitionedSeries[] {
+  const partitionedSeries: PartitionedSeries[] = [];
+  for (const datum of series) {
+    const currentPartition: Array<Omit<
+      PartitionedSeries,
+      'partitionSize' | 'partitionIndex'
+    >> = [];
+    let lastXValue = datum.points[0]?.x ?? 0;
```
psybuzz, line 45: I just tried writing a scalar `step=-1`, and it seems negative steps work in existing TensorBoard.  Should this `lastXValue` be `-Infinity`?

tensorboard/webapp/metrics/views/card_renderer/utils_test.ts
```diff
@@ -38,4 +39,134 @@ describe('metrics card_renderer utils test', () => {
...
+          runId: 'b',
+          points: buildPoints([1, 1, 1]),
+        },
+      ]);
+
+      expect(actual).toEqual([
+        {
+          seriesId: '["a",0]',
+          runId: 'a',
+          points: buildPoints([1, 2, 2]),
```
psybuzz, line 66: The more I think about it, the more I wonder whether this should actually be split into [1, 2] and [2].  If we see the same step value 2x, could that also indicate a process that died/restarted from checkpoint as well?

On the other hand, perhaps users might also intentionally "overwrite" the same last step as a legitimate use case?  I'm not sure, so the PR as-is seems reasonable, just interested in your thoughts.

psybuzz (approved):
Thanks for this UX improvement!

tensorboard/webapp/metrics/views/card_renderer/utils.ts
stephanwlee: Just wondering, how would that manifest as a bug?

PR #4701, opened 2021-02-23
timeseries: refactor scalar container data fetch

stephanwlee:
Previously, types flow through Observable in the scalars card container
was quite confusing with its typing and naming. This change refactors a
bit so it is easier by reorganizing and added certain types to be more
explicit.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
@@ -63,7 +63,7 @@ import {
 import {CardId, CardMetadata, XAxisType} from '../../types';
 import {CardRenderer} from '../metrics_view_types';
 import {getTagDisplayName} from '../utils';
-import {SeriesDataList, SeriesPoint} from './scalar_card_component';
+import {LegacySeriesDataList} from './scalar_card_component';
 import {
   ScalarCardDataSeries,
   ScalarCardPoint,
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -246,128 +246,152 @@ export class ScalarCardContainer implements CardRenderer, OnInit {
...
+      >((runToSeries) => {
+        const runIds = Object.keys(runToSeries);
+        if (!runIds.length) {
           return of([]);
         }
 
         return combineLatest(
-          runIdAndPoints.map((runIdAndPoint) => {
-            return this.getRunDisplayNameAndPoints(runIdAndPoint);
+          runIds.map((seriesId) => {
```
psybuzz, line 389: nit: same as above, might be clearer to see the runId to seriesId conversion if we have the parameter still be named `runId` and then make L392 `return {seriesId: runId, displayName}`.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -100,14 +100,14 @@ function areSeriesDataListEqual(
   });
 }
 
-interface RunIdAndPoints {
-  runId: string;
+interface PartialSeries {
+  seriesId: string;
```
psybuzz, line 104: In this PR, calls like `runSelectionMap.get(seriesData.seriesId)` and `getRunDisplayName(runIdAndPoint.seriesId)` assume that seriesId and runId are exactly the same.  Even though this assumption changes in the followup, could we add a comment in this PR here, noting that each `seriesId` is equal to 1 unique `runId`?

Alternatively, this PR could preserve the `runId` name, and the followup can introduce `seriesId`s.

psybuzz (approved):

psybuzz:
Thanks for this refactor.  Marking the legacy series as "Legacy*" is especially helpful!

PR #4703, opened 2021-02-24
lint: yarn lint ancient code

japie1235813:
These changes are there when running `yarn lint`, which is not related to any feature I am developing. Submit a pr to fix it and avoid `git restore` them every time.


japie1235813:
Closed the pr. the yarn might pick different version of prettier.

wchargin:
Yeah—sometimes you might run into this if your editor is automatically
picking an incompatible version of Prettier, in which case the right fix
is to tell it to pull from our `yarn.lock` or equivalent. To my
knowledge, Yarn itself should pick a consistent version, since Prettier
is hard-pinned in our manifest and lockfile.

PR #4704, opened 2021-02-25
graph: create initial README with user instructions

psybuzz:
This introduces a README in `tensorboard/plugins/graph/README.md`,
similar to other plugins. The main motivation is to:

- Catch users inspecting the repo who may want to know how the Graph
  dashboard works, and may have missed the tutorial.
- Have a place where we can add technical details that document the
  dashboard's internals.

It is **not** meant to be a replacement to the TensorFlow documentation
site's guide.  The official tutorial contains a Colab, is more stable (no TF1
references), and no implementation details of the dashboard's internals.

A followup change expands upon the technical details section:
https://github.com/tensorflow/tensorboard/pull/4705

### DIFF
--- a/tensorboard/plugins/graph/README.md
+++ b/tensorboard/plugins/graph/README.md
@@ -1,0 +1,129 @@
+# Graph Dashboard
+
+The Graph dashboard provides a visual representation of a computation graph,
+designed for understanding ML models.
+
+## What are graphs?
+
+Programs written with TensorFlow and other frameworks may transform the source
+code into a computational graph first, before actually executing the operations.
+To illustrate the idea, this short snippet:
+
+```
+w = 2
+result = sqrt(w)
+```
+
+might be represented as a graph with
+
+-   3 nodes: w, sqrt, result
+-   2 edges: [w, sqrt], [sqrt, result]
+
+Frameworks such as TF2 allow users to manually annotate functions with the
+`@tf.function` decorator to indicate to TensorFlow that a graph should be
+created. Other frameworks may require users to explicitly call a utility
+function to write an implicitly created graph.
+
+Notably, the graph itself does not contain variable values (e.g. weights),
+assets, or signatures (e.g. input and output values).
+
+### Creating a graph
+
+For details guides on creating graphs, please see the docs for
+[TensorFlow](https://www.tensorflow.org/guide/intro_to_graphs) or
+[PyTorch](https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter.add_graph)
+
+**Saving to an event file**
+
+In TensorFlow 2, wrap a `@tf.function` decorated function with
+`tf.summary.trace_on(graph=true)` and `tf.summary.trace_export()`. This will add
+the graph data to the event files in the log directory.
+
+```python
+@tf.function
+def my_func():
+  return tf.random.uniform((3, 3))
+
+writer = tf.summary.create_file_writer('logs')
+tf.summary.trace_on(graph=True)
+
+my_func()
+
+with writer.as_default():
+  tf.summary.trace_export(name="my_func", step=0)
+```
+
+In TensorFlow 1, the graph of a session is manually added with `add_graph()`.
+
+```python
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/graph/README.md
```diff
@@ -0,0 +1,129 @@
...
+result = sqrt(w)
+```
+
+might be represented as a graph with
+
+-   3 nodes: w, sqrt, result
+-   2 edges: [w, sqrt], [sqrt, result]
+
+Frameworks such as TF2 allow users to manually annotate functions with the
+`@tf.function` decorator to indicate to TensorFlow that a graph should be
```
japie1235813, line 23: "to indicate to" sounds some grammar error to me

tensorboard/plugins/graph/README.md
```diff
@@ -0,0 +1,129 @@
...
+created. Other frameworks may require users to explicitly call a utility
+function to write an implicitly created graph.
```
japie1235813, line 25: what does "implicitly created" mean exactly? On the other hand, what are examples of "explicitly created"?

tensorboard/plugins/graph/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+The dashboard offers a variety of features, including:
+
+**Inspection**
+
+-   Viewing details, attributes of a selected node by clicking it.
+-   Searching for a node by name.
+-   A "Trace inputs" mode to highlight all nodes that may have some effect on
+    the selected node.
+-   Node coloring modes: by structure, device, TPU compatibility, etc.
```
japie1235813, line 98: I recalled last time you said same color means the nodes are the same? Does this item mean the same thing?

tensorboard/plugins/graph/README.md
```diff
@@ -0,0 +1,129 @@
...
+-   Indicating "size" of output tensors visually with thicker edges.
+
+**Organization**
+
+-   Expand and collapse group nodes.
+-   Organize nodes automatically by name. TensorBoard uses the "/" forward slash
+    in node names to determine node groups.
```
japie1235813, line 105: I think it's better to align the items as either starting with verb or noun

tensorboard/plugins/graph/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+The `tensorboard/plugins/graph/tf_graph_*` directories contain the frontend
+code, while most of the non-view related processing and core types are defined
+in `tensorboard/plugins/graph/tf_graph_common`. This document is not exhaustive,
+but will highlight important components of graph processing and cover the
+following lifecycle:
+
+-   Loading graph data into runtime
+-   Parsing the graph data
+-   Processing an internal representation
```
japie1235813, line 129: What is "internal" representation?

tensorboard/plugins/graph/README.md
psybuzz: I think the grammar sounds natural to me, but it definitely sounds wordy.  Since this paragraph essentially outlines how graphs are created (next section), I've removed this paragraph and folded it into the ones below.

tensorboard/plugins/graph/README.md
psybuzz: Dropped.

Actually, I don't know any specific ML frameworks that let users explicitly create nodes and edges in the graph.

tensorboard/plugins/graph/README.md
psybuzz: Done.

tensorboard/plugins/graph/README.md
psybuzz: Yes, this coloring mode feature is intended to help users tell apart group nodes with similar internal structure.  I've expanded on this bullet point to hopefully be a bit clearer.

tensorboard/plugins/graph/README.md
psybuzz: Dropped these 3 list items.

I think it might be clearer to just have the full details in the followup PR, and drop this not-so-meaningful lifecycle list.

japie1235813 (approved):
Thanks for addressing the questions and making changes.

PR #4705, opened 2021-02-25
graph: update README with technical details

psybuzz:
This adds technical details to the Graph plugin's README.md.
It includes a glossary of terms, how a graph's data is loaded
and parsed, and discusses key processing steps.

It is not exhaustive and leaves many TODOs deliberately.

Diffbase: https://github.com/tensorflow/tensorboard/pull/4704

### DIFF
--- a/tensorboard/plugins/graph/README.md
+++ b/tensorboard/plugins/graph/README.md
@@ -120,3 +120,383 @@ The `tensorboard/plugins/graph/tf_graph_*` directories contain the frontend
 code, while most of the non-view related processing and core types are defined
 in `tensorboard/plugins/graph/tf_graph_common`. This document is not exhaustive,
 but will highlight important phases of graph processing.
+
+Loading a graph involves these core transformations:
+
+-   Fetch: path | pbTxtFile --> rawPbtxtString
+-   Parse: rawPbTxtString --> GraphDef
+-   Build: GraphDef --> SlimGraph
+-   Build: SlimGraph --> Hierarchy
+-   Build: Hierarchy --> RenderGraphInfo
+-   Build (lazily): SubHierarchy
+    -   Creates the BridgeGraph, stored on Node
+    -   Creates RenderNodeInfo for each rendered Node
+
+The GraphDef --> SlimGraph phase adapts the TensorFlow-specific GraphDef format
+into the simplest TensorBoard-owned, view-agnostic representation. SlimGraph's
+nodes are OpNodes, while Hierarchy contains both OpNodes and MetaNodes within.
+View components ingest Hierarchy's nodes throughout the lifetime of the app.
+
+The 'build sub hierarchy' phase happens lazily whenever the visible graph's
+structure changes. For example, expanding a group node or jumping to a deep node
+from the 'Search' results will trigger this.
+
+### Fetch and parse graph data
+
+Graph data is either loaded from a `fetch()` to the TensorBoard HTTP server's
+endpoint `data/plugins/graphs/graph` or when users manually upload a *.pbtxt
+from their filesystem in the UI. The following sketches out the pieces with
+pseudocode.
+
+**tf-graph-dashboard-loader**
+
+-   slimGraph, hierarchy = loader.fetchAndConstructHierarchicalGraph()
+-   Pass the resulting SlimGraph, Hierarchy values to view components
+    (`tf-graph-board`) via 2-way Polymer data bindings
+
+**tf_graph_common/loader:
+fetchAndConstructHierarchicalGraph(remotePath|pbTxtFile): SlimGraph, Hierarchy**
+
+-   graphDef = parser.fetchAndParseGraphData(remotePath, pbTxtFile)
+-   slimGraph = build(graph)
+-   checkOpsForCompatibility(slimGraph)
+    -   Updates nodes objects on SlimGraph
+-   hierarchy = hierarchy.build(slimGraph)
+-   return slimGraph, hierarchy
+
+**tf_graph_common/parser: fetchAndParseGraphData(pbtxtString): GraphDef**
+
+-   contents: string = fetchFromServer(filepath)
+
+    ```
+    node {
+      name: "adder/Add"
+      op: "Add"
+      input: "Placeholder"
+      input: "Placeholder_1"
+      attr {
+        key: "T"
+        value {
+          type: DT_FLOAT
+        }
+      }
+      attr {
+        ...
+      }
+    }
+    ...
+    ```
+
+-   graphDef = parsePbtxtFile(input)
+
+    -   Parses input in 1 MB chunks, line by line. It looks for the 'node'
+        fields and creates an array entry for each. The Protobuf text format
+        does not indicate which protobuf attributes are 'repeated', so a
+        hardcoded list `GRAPH_REPEATED_FIELDS` in the frontend is used to
+        determine which attributes to group as arrays.
+    -   Otherwsie, if the .pbtxt has 2 declarations with the same name, the two
+        '{}' values are combined into an array '[{}, {}]'.
+
+-   Returns a GraphDef object
+
+    ```
+    {
+      "node": [
+        {
+          "name": "adder/Add",
+          "op": "Add",
+          "input": ["Placeholder", "Placeholder_1"],
+          "attr": [...],
+        },
+        ...
+      ],
+      "versions": {"producer":561}
+    }
+    ```
+
+**tf_graph_common/graph: build(GraphDef): SlimGraph**
+
+-   Consider all 'Const' ops as 'in-embedding' nodes.
+-   Consider all '*Summary' ops as 'out-embedding' nodes.
+-   Convert GraphDef's nodes (NodeDefs) into OpNodes
+
+    -   Remove prefixes/suffixes from names
+        -   '^Placeholder' means "there is an input to this node called
+            'Placeholder' and that is a control dependency". Remove the caret.
+        -   'outputTensorKey: 0' means "this input has outputTensorKey 0".
+            Remove the suffix.
+    -   Finds the '_output_shapes' attribute and creates a TensorShape.
+
+-   'library.function' is a special, repeated GraphDef property we process.
+
+    -   Create an artificial OpNode with name `'__function_library__' +
+        func.signature.name` and an empty op type.
+    -   Create an artificial OpNode for each func.signature.input_arg with
+        properties
+
+        ```
+        {
+          name: '__function_library__' + func.signature.name + '/' + arg.name,
+          op: 'input_arg',
+          attr: [ {key: 'T', value: {type: arg.type}}, ],
+          functionInputIndex: currentInputIndex++,
+        }
+        ```
+
+    -   Create an artificial OpNode for each func.signature.output_arg, using
+        func.node_def nodes. We assume that output_args and node_defs are 1:1.
+
+        ```
+        {
+          ...node_def,
+          name: '__function_library__' + func.signature.name + '/' + node_def.name,
+          input: [node_def.input],
+          functionOutputIndex: currentOutputIndex++,
+        }
+        ```
+
+-   'Building the data structure' (SlimGraph)
+
+    -   Add nodes to SlimGraph
+
+        -   Set the name to be the normalized name.
+        -   Set the `inEmbeddings`, `outEmbeddings` properties on the OpNode.
+
+    -   Add edges to SlimGraph
+
+        -   Detect reference edges using a hardcoded list (e.g. Assign,
+            AssignAdd).
+        -   Move the inputs of the (Const) in-embedding node into incoming edges
+            of the main node.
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+        root. If node is named 'a/b/c', create metanodes 'a' and 'a/b'.
+    -   Update embedding nodes to point to the OpNode they are embedded to:
+        `embedding.parentNode = node;`.
+-   'Detect series'
+    -   Recursively traverse metanodes starting from the top-level. At each
+        metanode,
+        -   Take the direct children nodes and group them into clusters by
+            op-type.
+        -   For each op cluster, detect series by greedily looking for numeric
+            suffix changes. For example, 3 OpNodes with op 'Add': `add_1`,
```
stephanwlee, line 327: It might be interesting to explicitly call out implicit `_0` which this logic understands.

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+            `add_2`, `add_3` will be grouped into a SeriesNode 'add_[1-3]'.
+        -   Series groups require at least 2 nodes (later we filter out groups
+            with less than 5 nodes) where the numeric suffix is in increasing
+            order.
+        -   Children within a SeriesNode have a reference: `child.owningSeries =
+            seriesName`.
+    -   If series has 5+ members,
```
stephanwlee, line 334: I asked myself this question reading section above. Perhaps call this out before L329?

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+        -   Add it to hierarchy and the metagraph, updating references,
+            histograms, etc. The metagraph no longer considers the original
+            OpNodes as direct children, if they are owned by a SeriesNode.
+-   'Adding edges'
+    -   For each BaseEdge
```
stephanwlee, line 339: Where is the BaseEdge from and who creates it? Is it a class in the SlimGraph?

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+            -   When a metaedge contains multiple base edges, it sums their
+                total sizes
+        -   Update the MetaEdge with a reference to the BaseEdge:
+            `metaedge.addBaseEdge(baseEdge, h)`. A metanode may contain multiple
+            BaseEdges if there are multiple connections between two metanodes.
+    -   Note, the Hierarchy does not have a direct reference to metaedges. It
+        simply has a root metanode with a metagraph, and edges of the metagraph
+        are metaedges.
+-   'Finding similar subgraphs'
+
```
stephanwlee, line 358: nit: remove this for consistency.

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+            store the maxMetaEdgeSize on the Hierarchy itself.
+            -   For each outputShape tensor, get the "size": [4, 2, 1] is 4 *
+                2 * 1 = 12.
+            -   When a metaedge contains multiple base edges, it sums their
+                total sizes
+        -   Update the MetaEdge with a reference to the BaseEdge:
+            `metaedge.addBaseEdge(baseEdge, h)`. A metanode may contain multiple
+            BaseEdges if there are multiple connections between two metanodes.
+    -   Note, the Hierarchy does not have a direct reference to metaedges. It
+        simply has a root metanode with a metagraph, and edges of the metagraph
```
stephanwlee, line 355: Hmm, so who owns the metaedge? Is it metanode?

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+        -   Create a map from `Record<Signature, MetaNode[]>`, and keep only the
+            clusters that contain multiple MetaNodes, unless it is a library
+            function. For functions, we may "add more nodes with the template
+            later" (TODO: why?).
+    -   let templates = groupTemplateAndAssignId(nnGroups, verifyTemplate);
+    -   For each MetaNode in a nnGroup
+
+        -   For all previous clusters, try to add the current MetaNode to the
+            previous cluster if their metagraphs are similar. Note: this likely
+            results in a quadratic+ time complexity.
```
stephanwlee, line 375: Right, actual guestimation of time complexity would be nice in terms of |cluster| and complexity of clustering with |N| and |E|. 

Do you think there is some chance of saving compute time if we went bottom to up instead to achieve memoization?

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+            Signatures over all nodes in the entire graph are required to
+            determine the templates for node coloring.
+        -   Signatures come from serializing a node's structural info:
+            -   "depth=1 |V|=3 |E|=0 [ops] Add=3"
+            -   "depth=2 |V|=6 |E|=7 [ops] Placeholder=2,Add=9,Pow=1"
+        -   Create a map from `Record<Signature, MetaNode[]>`, and keep only the
+            clusters that contain multiple MetaNodes, unless it is a library
+            function. For functions, we may "add more nodes with the template
+            later" (TODO: why?).
+    -   let templates = groupTemplateAndAssignId(nnGroups, verifyTemplate);
```
stephanwlee, line 370: So there was some mentions of template earlier (when defining hierarchy) and previous section described signatures. Are they related at all? If not, how is a template defined?

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+
+## Glossary
+
+Communicating the inner workings of the Graph dashboard requires introducing
+some dashboard-specific concepts used throughout the web frontend code. Much of
+this terminology comes from
+[existing literature](https://www.cs.ubc.ca/labs/imager/tr/2008/Archambault_GrouseFlocks_TVCG/grouseFlocksSub.pdf).
+
+-   **GraphDef**: A 'base graph' structure, loosely tied to TensorFlow's
+    [graph.proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/graph.proto).
```
stephanwlee, line 443: You don't like the TB version? :D https://github.com/tensorflow/tensorboard/blob/master/tensorboard/compat/proto/graph.proto

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+-   **NodeDef / RawNode**: A 'base node' structure, loosely tied to TensorFlow's
+    [node_def.proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/node_def.proto).
+-   **OpNode**: The result of 'normalizing' a raw `NodeDef`.
```
stephanwlee, line 446: What does it "normalize"? I assume op names and edge info, right?

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+-   **BaseEdge**: An edge between two `OpNodes`.
+-   **SlimGraph**: A lightweight structure containing `OpNode`s and `BaseEdge`s.
```
stephanwlee, line 448: Meta question: do you think we need this data structure or can we go to Hierarchy straight from GraphDef? Does it have actual benefit from the abstraction? 

Also, what do you think about the name? Slim was type of the models for TFv1 and it confuses me :\

tensorboard/plugins/graph/tf_graph_common/hierarchy.ts
```diff
@@ -914,7 +914,11 @@ function groupSeries(
     });
   });
 }
-/** cluster op-nodes with similar op */
+/**
+ * Cluster op-nodes with similar op. This examines only the direct children of
+ * the metagraph, does not recursively check descendants.
+ * @return A map from op to a list of node names.
```
stephanwlee, line 920: Can we do this via better typing or named types?

stephanwlee (approved):
I left bunch of comments but they are general asks for clarification, your opinion, and your future plan with information given in the doc.

"Code" wise, I have no problem and any improvements to documentation is a good thing  so LGTM. Feel free to take more iterative approach in upgrading the doc.

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+
+Note: building a Hierarchy modifies the SlimGraph's OpNodes in place!
+
+**tf_graph_common/hierarchy.ts: build(SlimGraph): Hierarchy**
+
+-   Create an artificial MetaNode for the root.
+-   Create an index of nodes as `Record<nodeId: string,
+    OpNode|SeriesNode|MetaNode>`
+-   'Adding nodes'
+    -   Iterate over OpNodes, keep track of device names, XLAClusters,
```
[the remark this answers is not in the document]
psybuzz: Today, the use cases are definitely close to the RenderGraphInfo's view logic.  Some pieces, such as calculating the op/TPU compat histograms, only need to be calculated if the user explicitly turns on the appropriate "color by" mode.  I think it depends on whether there are use cases for knowing those expensive stats that do not depend on the "rendered graph".

Even if we compute these lazily, what I like about storing info like histograms, device names on Hierarchy is that it allows us to think about other UI views besides a graph.  e.g. what if there was a tabular UI that showed Op/Device stats?  Such ideas are hypothetical and definitely a stretch, though.

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+### Building the Hierarchy
+
+View components create the RenderGraph mainly from the Hierarchy. As we will
+see, the key difference between a Hierarchy and a SlimGraph is that a Hierarchy
+contains more nodes beyond the OpNodes: MetaNodes are created to group other
+nodes, and MetaEdges are a layer on top of the SlimGraph's BaseEdges.
+
+Furthermore, the Hierarchy building phase will compute expensive information
+about the graph and store it on the data structure, such as templates. Template
+ids are stored on MetaNodes, and are used by the view components to determine
```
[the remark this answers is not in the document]
psybuzz: Dropped the reference to "view components".

tensorboard/plugins/graph/README.md
```diff
@@ -121,9 +121,384 @@ describing the motivations, research, and process behind it.
...
+node coloring. When 'Color by structure' is enabled, MetaNodes with similar
+structure are assumed to have the same templateId, and thus assigned the same
+color.
+
+The BridgeGraph is not computed at this phase, but instead on-demand by the view
+components.
+
+Note: building a Hierarchy modifies the SlimGraph's OpNodes in place!
+
+**tf_graph_common/hierarchy.ts: build(SlimGraph): Hierarchy**
```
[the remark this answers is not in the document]
psybuzz: Replaced this and other occurrences with `tf_graph_common/hierarchy.ts: SlimGraph -> Hierarchy`.

Hm I see why they are confusing.  In the source code, there is a function named `build` with a `SlimGraph` parameter and it returns a `Hierarchy`.  Maybe the `x -> y` format without any function name would be clearer?

tensorboard/plugins/graph/README.md
psybuzz: Done.

tensorboard/plugins/graph/README.md
psybuzz: Added note.

tensorboard/plugins/graph/README.md
psybuzz: Good call, it's the metagraph.  Updated.

tensorboard/plugins/graph/README.md
psybuzz: Done.

tensorboard/plugins/graph/README.md
psybuzz: These are the same term.  Updated.

tensorboard/plugins/graph/README.md
psybuzz: Updated the above line to say 5, which is a configurable parameter.

psybuzz (commented):
Thanks for the thorough review!

Taking an iterative approach on these docs sounds good.  I'll take any further comments today, and plan to address/submit sometime this PR tomorrow at the latest.

tensorboard/plugins/graph/README.md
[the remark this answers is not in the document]
stephanwlee: 👍  thanks!

stephanwlee (approved):

Issue #4702, opened 2021-02-24
Fix link in README - Graph Explorer / Graph Visualiser


AlexAndrascu:
The link pointing to the Graph Explorer / Visualiser tutorial is broken:

(https://www.tensorflow.org/get_started/graph_viz)

stephanwlee:
Thanks for raising this. When using `git grep`, there were three hits for `graph_viz`.

```
README.md:graph visualizer tutorial](https://www.tensorflow.org/get_started/graph_viz).
docs/r1/overview.md:graph visualizer tutorial](https://www.tensorflow.org/get_started/graph_viz).
tensorboard/plugins/graph/tf_graph_dashboard/tf-graph-dashboard.ts:            <a href="https://www.tensorflow.org/get_started/graph_viz"
```

nfelt:
https://github.com/tensorflow/docs/blob/master/site/en/r1/guide/graph_viz.md is where the original content now lives (the docs from TF 1 are no longer rendered on tensorflow.org, they're only in GitHub).  Might be better to instead replace with https://www.tensorflow.org/tensorboard/graphs which is our newer Colab-based tutorial.

PR #4706, opened 2021-02-25
Update README.md

ymodak:
Fixes https://github.com/tensorflow/tensorboard/issues/4702 (broken graph visualizer link).


stephanwlee:
As noted, there is another reference to the graph_viz in the codebase. Can we fix that, too, please?

nfelt (approved):
Thanks for the fix!

ymodak:
I have updated the pr with other changes mentioned in the issue thread as suggested. Thanks!

stephanwlee (approved):

PR #4708, opened 2021-02-25
alerts: plumb followupAction into store

psybuzz:
The `Alerts` feature's views are responsible for handling and surfacing
alerts from actions in a "toast" UI. Modules may register a `followup`
action to explicitly show a "retry" style button that triggers a followup
action.

The retry button did not appear as expected, since the reducer was
only storing the `localizedMessage` field. This change includes the
optional `followupAction` as well.

Tested by manually modifying `metrics_module.ts`'s `alertFromAction`
registration to return a followup action:
```ts
return {
  localizedMessage: 'whoah message',
  followupAction: {
    localizedLabel: 'whoah label',
    getFollowupAction: async () => {
      console.log('yay')
      return actions.cardStepSliderChanged({cardId: 'blah', stepIndex: 0});
    },
  }
};
```
then after clicking the 'Pin' icon on a card to trigger the action, observed
that the followup action produces the desired retry button.
image

### DIFF
--- a/tensorboard/webapp/alert/store/BUILD
+++ b/tensorboard/webapp/alert/store/BUILD
@@ -48,6 +48,7 @@ tf_ts_library(
         ":testing",
         ":types",
         "//tensorboard/webapp/alert/actions",
+        "@npm//@ngrx/store",
         "@npm//@types/jasmine",
     ],
 )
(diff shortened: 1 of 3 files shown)


wchargin (approved):
Nice; thanks for looking into this!

psybuzz:
Thanks for identifying this bug!

Issue #1557, opened 2018-10-27
Specify limited x-axis range for all plots


pwohlhart:
It would be nice if we could have a way to specify a specific range of steps/wall time to look at.
Often you are only interested in the last few steps.
Even more often the very first few (hundred to thousand) steps are not very interesting and/or have a very different value ranges than everything beyond that, which makes the automatic zooming much less useful.
(see e.g, the graph in nikonikolov's comment on issue #405.

Issue #77, opened 2017-06-16
Scalars: Simple Way of Setting y axis to start at 0


decentralion:
Migrated from https://github.com/tensorflow/tensorflow/issues/7631

> I have a number of scalars that I plot in tensorboard that are non negative values (ie accuracy, etc). It would be cool if there was a quick way to have the y axis start at 0 rather than allocating unused space in the negative y direction.
> ![image](https://cloud.githubusercontent.com/assets/51059/23080071/7b1aa63c-f51d-11e6-8ae0-82bbf299a01f.png)
>

dandelionmane:
I think the right way to do this would be to add another argument to the scalar_summary:

`tb.plugins.scalar.summary(name, value, ... , y_axis=[0, None])`

y_axis would take a tuple of numbers, where None means default behavior, and a number means fix the bound to that exact number. The first number would be the lower bound, the second would be the upper bound.

dandelionmane:
Oh, and the extra argument would get proxied to the scalars plugin via SummaryMetadata. Note we still need to get support for SummaryMetadata into released TensorFlow (right now it is in HEAD but not 1.2).

chris-chris:
@dandelionmane 
May I try this issue? I'll follow your suggestions above.

dandelionmane:
Sure. Just so you know, it will probably be a tricky change that will require a few iterations to get right.
Here are the steps, (can be done in different order):
- Modify the [vz-line-chart](https://github.com/tensorflow/tensorboard/tree/master/tensorboard/plugins/scalars/vz_line_chart) so that there is a setting for explicitly configuring the y scale domain
- Modify the [scalars plugin](https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/scalars/scalars_plugin.py) so that it has a route for providing metadata on what the y scale domain is for a given tag. It would be nice to do this in a way where we can use it as a general pattern across plugins for retrieving per-tag metadata (we might want to include it in the index route)
- Modify the [tf-scalars-dashboard](https://github.com/tensorflow/tensorboard/tree/master/tensorboard/plugins/scalars/tf_scalar_dashboard) so that it uses this new metadata from the route to properly set the setting on individual charts
[...]

PR #4711, opened 2021-02-25
line chart: support manual extent changes

stephanwlee:
One of the requested feature in the scalars chart is being able to
change the axis extent manually. For example, when plotting a
probability, you want to see the chart from [0, 1] since there is
absolutely no reason to look beyond that. This change gives a UI
affordance to do so.

Clicking on an axis, you will see a popover with controls that allows
you to manually set the extents.

Future UX tweaks (including how the UI control is presented) will
follow.

image
image

~Something to be improved (in a follow up): when the dropdown is visible, maybe make the edit button still visible.~


stephanwlee:
FYI @atishagarwala

stephanwlee:
Relevant GitHub issues:
#77, #1557,

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -31,7 +32,12 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
 @Component({
   selector: 'line-chart-axis',
   template: `
-    <div [class]="axis + '-axis'">
+    <div
+      [class]="axis + '-axis axis'"
+      [matMenuTriggerFor]="manualControl"
```
psybuzz, line 37: Having the axis-es themselves be click targets is certainly convenient.  However, this would directly conflict with a future plan to make the x-axis clickable for a "linked time" mode.  In that separate proposal, the x-axis may be clickable to set a chart's "current step" to a specific value, or to a range if click+drag.  I think it's also reasonable that in the future, TB might adopt gestures from line charts in other tools such as showing a vertical/horizontal line when hovering over the x/y axis.

WDYT about starting out without the "axis as click target", and instead starting with these menus triggered from a new option inside 3-dot menu on each scalar card?

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -56,6 +62,40 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
...
+      xPosition="before"
+      [yPosition]="axis === 'y' ? 'above' : 'below'"
+    >
+      <div
+        class="extent-edit-input"
+        (click)="$event.stopPropagation()"
+        (keydown)="keydownPreventClose($event)"
+      >
+        <label>min</label>
+        <input #minInput tabindex="1" type="number" [value]="axisExtent[0]" />
```
psybuzz, line 76: When manually testing these inputs:
- I open the menu, the min input shows X (current min)
- I edit the min value to X2, then I dismiss the menu
- I reopen the menu

Actual: input still shows X2, even though current value is X
Expected: input should show X, because I expect the form to reflect the current state of the chart

psybuzz, line 76: Panning the graph via alt+drag seems to set the value to extremely high precision, and by default the beginning of the number is not visible.  Can we round values when the form is hydrated?

image

psybuzz (commented):
I'm pleasantly excited to see this PR!

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -56,6 +62,40 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
...
+        <label>max</label>
+        <input #maxInput tabindex="2" type="number" [value]="axisExtent[1]" />
+      </div>
+      <div class="extent-edit-control" (keydown)="keydownPreventClose($event)">
+        <button mat-stroked-button tabindex="4">Cancel</button>
+        <button
+          mat-raised-button
+          color="primary"
+          tabindex="3"
+          class="change"
```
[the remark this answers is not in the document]
stephanwlee: With the style encapsulation, what does that really achieve?

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -56,6 +62,40 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
         </g>
       </svg>
     </div>
+    <mat-menu
+      #manualControl="matMenu"
+      xPosition="before"
+      [yPosition]="axis === 'y' ? 'above' : 'below'"
+    >
```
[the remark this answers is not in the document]
stephanwlee: > If the mat-menu is what's handling the dismissal on click/keydown, can we remove all the stopPropagation's from the div/buttons inside, and instead move those event listeners to the <mat-menu> itself?

Does not seem to work. Specifically, mat-menu uses portal and content inside the mat-menu gets projected onto the overlay, not the `mat-menu` itself. We can put a wrapper around all of our element but I think that defeats keyboard navigation or some of the assumption of the mat-menu. Will keep it as is.

> If Cancel/Change could call something like menu.close, it would be more explicit.

At least for testing, added that explicit close.

> Also, it seems click(+drag) on the mat-menu itself will dismiss it, when it should not.

Might be annoying but it is okay as is.

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -56,6 +62,40 @@ const DAY_IN_MS = 24 * 1000 * 60 * 60;
...
+        <label>min</label>
+        <input #minInput tabindex="1" type="number" [value]="axisExtent[0]" />
+      </div>
+      <div
+        class="extent-edit-input"
+        (click)="$event.stopPropagation()"
+        (keydown)="keydownPreventClose($event)"
+      >
+        <label>max</label>
+        <input #maxInput tabindex="2" type="number" [value]="axisExtent[1]" />
```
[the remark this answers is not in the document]
stephanwlee: I used tabindex to optimize the tabbing UX; wanted the `change` button to take precedence over `cancel`. Achieved the same thing without tabindex using `reverse` for now.

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
stephanwlee: We can optimize those UX consideration in a follow up. Depending on a value, rounding can be quite janky thus is not a simple one or two line change.

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
stephanwlee: > However, this would directly conflict with a future plan to make the x-axis clickable for a "linked time" mode. In that separate proposal, the x-axis may be clickable to set a chart's "current step" to a specific value, or to a range if click+drag

I think we could fix this when we get there. Overflow menu is very cumbersome and imo, counter-intuitive in current manifestation of the feature..

stephanwlee:
@psybuzz updated the PR and description according to the offline discussion.

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.scss
```diff
@@ -92,3 +104,53 @@ text {
...
   }
 }
+
+.extent-edit-button {
+  background-color: mat-color($mat-gray, 200);
+  display: none;
+  font-size: 0;
+  height: 24px;
+  line-height: 24px;
+  margin: 5px;
```
psybuzz, line 114: nit: setting margin together with pos:abs seems a bit redundant, maybe just specify the right and drop margin?  e.g. `right: 5px`

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.scss
```diff
@@ -92,3 +104,53 @@ text {
     );
   }
 }
+
+.extent-edit-button {
+  background-color: mat-color($mat-gray, 200);
+  display: none;
+  font-size: 0;
```
psybuzz, line 111: This element has no text, and line-height is fixed.  Checking: can `font-size` be dropped?

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -101,4 +110,34 @@ export class LineChartAxisComponent {
       ? LineChartAxisComponent.TEXT_PADDING
       : this.getDomPos(tick);
   }
+
+  keydownPreventClose(event: KeyboardEvent) {
+    if (event.key !== 'Escape') {
+      event.stopPropagation();
```
psybuzz, line 116: Can we add a TODO / comment explaining why we need to stopPropagation to prevent MatMenu's default behavior?  I think it's not super obvious without having manually playing with this feature.

psybuzz (approved):
Thanks for addressing the feedback!  Only minor comments.

> With the style encapsulation, what does that really achieve?

True, there actually isn't much real risk.  A quick search shows only 3 components explicitly using `ViewEncapsulation.None`, but since we prefer encapsulation by default, no need to worry.

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.scss
stephanwlee: I cannot fathom internals of mat-icon-button but without font-size, it pushes the mat-icon down to near ~line-height which makes me think that there is some invisible whitespace in the template causing the UI jank.

PR #4721, opened 2021-03-02
line chart: extract HTML and style from axis

stephanwlee:
As we add more features, the module is getting out of hand. This change
refactors and extracts template and style part of the module into
separate files.


japie1235813 (approved):

stephanwlee:
Thanks for a quick review.

PR #4722, opened 2021-03-03
Add a toggle that turns automatic high-degree node extraction off

martinwicke:
This adds a toggle (next to "trace inputs") which turns automatic high-degree node extraction off.

Automatic aux node creation can break the graph up in unwanted ways. For some graphs, the high-degree heuristic is not useful (graphs that have high-degree internal nodes, e.g., if you concatenate lots of towers or some such)

This adds an argument to the hierarchy constructor which if passed, avoids calling the high-degree node extraction code. It also removes the hard-coded PARAM entry for node extraction (which turned off all node extraction, but wasn't user settable).

  - build
  - upload graph known to be broken apart verify it's broken apart
  - toggle "auto-extract high-degree nodes" 
  - verify graph changes, and high-degree nodes are not extracted

  - Make the in/out degree triggering extraction configurable (probably not worth it)
  - Turn off high-degree extraction altogether instead of providing the option (would work for me, but I'm hesitant to break others in this manner)
  - Rely more heavily on op list extraction (this might work, but I lack the expertise to come up with a good op list)

### DIFF
--- a/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
+++ b/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
@@ -44,8 +44,6 @@ class TfGraphScene2 extends LegacyElementMixin(PolymerElement) {
   colorBy: string;
   @property({type: Boolean})
   traceInputs: boolean;
-  @property({type: Boolean})
-  extractNodes: boolean;
 
   // For each render hierarchy, we only fit it to the viewport once (when the scene is attached to
   // the DOM). We do not fit the hierarchy again (unless the user clicks the reset button). For
@@ -664,11 +662,5 @@ class TfGraphScene2 extends LegacyElementMixin(PolymerElement) {
       this.selectedNode,
       this.traceInputs
     );
-  }
-    
-  // Re-do hierarchy if extraction changes
-  @observe('extractNodes')
-  _updateExtractNodes() {
-    this.renderHierarchy
   }
 }
--- a/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
+++ b/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
@@ -665,4 +665,10 @@ class TfGraphScene2 extends LegacyElementMixin(PolymerElement) {
       this.traceInputs
     );
   }
+    
+  // Re-do hierarchy if extraction changes
+  @observe('extractNodes')
+  _updateExtractNodes() {
+    this.renderHierarchy
+  }
 }
--- a/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
+++ b/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
@@ -44,6 +44,8 @@ class TfGraphScene2 extends LegacyElementMixin(PolymerElement) {
   colorBy: string;
   @property({type: Boolean})
   traceInputs: boolean;
+  @property({type: Boolean})
+  extractNodes: boolean;
 
   // For each render hierarchy, we only fit it to the viewport once (when the scene is attached to
   // the DOM). We do not fit the hierarchy again (unless the user clicks the reset button). For
--- a/tensorboard/plugins/graph/tf_graph/tf-graph.ts
+++ b/tensorboard/plugins/graph/tf_graph/tf-graph.ts
@@ -77,7 +77,6 @@ class TfGraph extends LegacyElementMixin(PolymerElement) {
           health-pill-step-index="{{healthPillStepIndex}}"
           handle-edge-selected="[[handleEdgeSelected]]"
           trace-inputs="[[traceInputs]]"
-          extract-nodes="[[extractNodes]]"
         ></tf-graph-scene>
       </div>
     </div>
@@ -137,7 +136,7 @@ class TfGraph extends LegacyElementMixin(PolymerElement) {
   @property({type: Boolean})
   traceInputs: boolean;
   @property({type: Boolean})
-  extractNodes: boolean;
+  autoExtractNodes: boolean;
   @property({type: Array})
   nodeContextMenuItems: unknown[];
   @property({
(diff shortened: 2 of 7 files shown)


tensorboard/plugins/graph/tf_graph/tf-graph.ts
```diff
@@ -173,6 +176,7 @@ class TfGraph extends LegacyElementMixin(PolymerElement) {
     (this.$$('tf-graph-scene') as any).panToNode(nodeName);
   }
   @observe(
+    'extractNodes',
```
psybuzz, line 179: When manually testing this PR, I discovered that:
- Set the toggle Off (no auto-extract)
- Load a graph (observe nothing was extracted)
- Set the toggle On (observe nodes auto-extracted)

Then,
- Set the toggle Off again (no auto-extract)

Expected: nothing should be extracted
Actual: high degree nodes still extracted, even though the toggle is Off

I think this surprising behavior manifests because each node's `.include` property is not cleared upon changing the toggle. To correct this, could we split it this L179 observer into a new one separate from the rest ('graphHierarchy', 'edgeWidthFunction', ...), which resets each node's `.include` to be unspecified?

Something like:

```ts
@observe('autoExtractNodes')
_onAutoExtractNodesChanged() {
    if (!this.graphHierarchy) return;
    for (const node of Object.values(this.graphHierarchy.getNodeMap())) {
        node.include = tf_graph.InclusionType.UNSPECIFIED;
    }
    this._buildRenderHierarchy(this.graphHierarchy);
  }
```
?

tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.ts
```diff
@@ -1357,6 +1369,7 @@ class TfGraphControls extends LegacyElementMixin(PolymerElement) {
     this._selectedTagIndex = 0;
     this._selectedGraphType = this._getDefaultSelectionType();
     this.traceInputs = false; // Set trace input to off-state.
+    this.extractNodes = true;
```
psybuzz, line 1372: Optional: I would also be okay with dropping this.  Presumably, if a user is inspecting a graph that doesn't require high degree extraction, they might have several graphs of this type in their logdir.  Would it make sense to preserve the state, instead of resetting it to `true`, upon run index changed?

tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
```diff
@@ -44,6 +44,8 @@ class TfGraphScene2 extends LegacyElementMixin(PolymerElement) {
   colorBy: string;
   @property({type: Boolean})
   traceInputs: boolean;
+  @property({type: Boolean})
+  extractNodes: boolean;
```
psybuzz, line 48: nit: throughout this PR, wdyt about renaming `extractNodes` to `autoExtractNodes`?

In case we build out more node extraction logic in the future, I think `autoExtractNodes` would be a clearer name to avoid confusion with manual node extraction, for example.

psybuzz (commented):
Thanks for contributing this PR!

tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
martinwicke: Yup, will do.

tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
```diff
@@ -663,4 +665,10 @@ class TfGraphScene2 extends LegacyElementMixin(PolymerElement) {
       this.traceInputs
     );
   }
+    
+  // Re-do hierarchy if extraction changes
+  @observe('extractNodes')
+  _updateExtractNodes() {
+    this.renderHierarchy
+  }
```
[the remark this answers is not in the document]
martinwicke: Yes.

tensorboard/plugins/graph/tf_graph/tf-graph.ts
```diff
@@ -77,6 +77,7 @@ class TfGraph extends LegacyElementMixin(PolymerElement) {
           health-pill-step-index="{{healthPillStepIndex}}"
           handle-edge-selected="[[handleEdgeSelected]]"
           trace-inputs="[[traceInputs]]"
+          extract-nodes="[[extractNodes]]"
```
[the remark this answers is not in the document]
martinwicke: Honestly, I do not understand where these are needed and where they are not.

tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
martinwicke: Will do.

tensorboard/plugins/graph/tf_graph/tf-graph.ts
martinwicke: Ok. This will reset manually included/excluded nodes, but I agree that is better than the current behavior.

tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.ts
martinwicke: I think that's more intuitive, I agree. It does look like a global UI setting, not a per-graph setting. I'll revert this reset.

martinwicke:
There's some lint left, I'll fix this when it's otherwise ready. 

Can you take a look whether this is better?

tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.ts
```diff
@@ -1079,9 +1086,14 @@ class TfGraphControls extends LegacyElementMixin(PolymerElement) {
     notify: true,
   })
   traceInputs: boolean = false;
+  @property({
+    type: Boolean,
+    notify: true,
+  })
+  autoExtractNodes: boolean = false;
```
psybuzz, line 1093: For now, I'd prefer preserving today's existing default behavior and keep `autoExtractNodes = true` by default. In the future, we can always change the default easily, based on user feedback. Would that be reasonable?

psybuzz (approved):
LGTM % minor comments, I'm pleased to see this happen!

If the Prettier linting ends up being too painful, let me know and I can push a commit from my end as well.

tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.ts
martinwicke: Oops. Sorry. Leftover from testing.

martinwicke:
Happy to run prettier. It does propose some changes that are unrelated to my edits though -- do you want those?

martinwicke:
I didn't add the unrelated lint things for now.

psybuzz:
> It does propose some changes that are unrelated to my edits though -- do you want those?

I'm not sure why there would be unrelated edits, unless maybe Prettier is run from another directory (running from "tensorboard_repo_dir/tensorboard/" might not pickup our settings).

That said, I checked out your latest patch, ran Prettier locally, and it passed the linter locally, so it seems good to me

martinwicke:
Cool. Merge at will then. I can, for historical reasons, but I don't think I should.

PR #4726, opened 2021-03-04
rust: implement tensor RPCs and client

nfelt:
Implements RPC-level support for the tensor data class in Rustboard (part of #4422).  In particular we include tensor time series when generating the `ListPlugins` response, and we wire up `ListTensors` and `ReadTensors` on both the server and client.  The latter RPCs are almost identical to the scalars ones; the only nontrivial differences involve handling the actual `TensorProto` values, where we have to explicitly clone them out of the `Commit` on the server side, and convert them into numpy arrays on the client side.

This doesn't yet render any data on its own, since nothing populates the `tensors` fields in the commit (which were introduced by #4710), but with coming PRs that implement the population side, I've tested end-to-end that this correctly renders the expected dashboards.  (There are also unit tests for server and client modeled on those for scalars.)

### DIFF
--- a/tensorboard/data/server/server.rs
+++ b/tensorboard/data/server/server.rs
@@ -234,16 +234,114 @@ impl TensorBoardDataProvider for DataProviderHandler {
 
     async fn list_tensors(
         &self,
-        _request: Request<data::ListTensorsRequest>,
+        req: Request<data::ListTensorsRequest>,
     ) -> Result<Response<data::ListTensorsResponse>, Status> {
-        Err(Status::unimplemented("not yet implemented"))
+        let req = req.into_inner();
+        let want_plugin = parse_plugin_filter(req.plugin_filter)?;
+        let (run_filter, tag_filter) = parse_rtf(req.run_tag_filter);
+        let runs = self.read_runs()?;
+
+        let mut res: data::ListTensorsResponse = Default::default();
+        for (run, data) in runs.iter() {
+            if !run_filter.want(run) {
+                continue;
+            }
+            let data = data
+                .read()
+                .map_err(|_| Status::internal(format!("failed to read run data for {:?}", run)))?;
+            let mut run_res: data::list_tensors_response::RunEntry = Default::default();
+            for (tag, ts) in &data.tensors {
+                if !tag_filter.want(tag) {
+                    continue;
+                }
+                if plugin_name(&ts.metadata) != Some(&want_plugin) {
+                    continue;
+                }
+                let max_step = match ts.valid_values().last() {
+                    None => continue,
+                    Some((step, _, _)) => step,
+                };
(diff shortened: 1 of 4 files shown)


tensorboard/data/server/server.rs
```diff
@@ -234,16 +235,114 @@ impl TensorBoardDataProvider for DataProviderHandler {
...
+                let mut points = ts.valid_values().collect::<Vec<_>>();
+                downsample::downsample(&mut points, num_points);
+                let n = points.len();
+                let mut steps = Vec::with_capacity(n);
+                let mut wall_times = Vec::with_capacity(n);
+                let mut values = Vec::with_capacity(n);
+                for (step, wall_time, value) in points {
+                    steps.push(step.into());
+                    wall_times.push(wall_time.into());
+                    // Clone the TensorProto to get a copy to send in the response.
```
wchargin, line 326: Cool, this seems fine. If we wanted to be really tight with the locks,
we could change the commit’s tensor values to hold `Arc<TensorProto>`s
and then clone them only after dropping the lock—and maybe one day Prost
and Tonic might [learn about `Arc<_>`s][arc] and remove the need to clone
at all. But certainly no action required.

[arc]: https://github.com/danburkert/prost/issues/429

tensorboard/data/server/server.rs
```diff
@@ -820,6 +922,107 @@ mod tests {
...
+                .as_ref()
+                .unwrap()
+                .plugin_name,
+            "tensors".to_string()
+        );
+    }
+
+    #[tokio::test]
+    async fn test_read_tensors() {
+        fn make_string_tensor_proto<T: Into<Bytes>>(value: T) -> pb::TensorProto {
```
wchargin, line 971: This is perfectly fine; it’s also idiomatic to write

```rust
fn make_string_tensor_proto(value: impl Into<Bytes>) -> ...
```

using an existential instead of a type parameter. These are mostly
equivalent; I don’t have a strong preference here.

wchargin (approved):
LGTM—a couple passing comments, but no suggestions.

tensorboard/data/server/server.rs
nfelt: Done; thanks for the tip.

tensorboard/data/server/server.rs
nfelt: Yeah, I realize it's not exactly ideal to be cloning this here, but SGTM to defer further optimization for now.  I'd found myself wondering whether TensorProto was even the right type for the commit to hold in the first place (since we're forcing it into an ndarray ultimately, it could be slimmed down quite a bit, and maybe made more ergonomic to manipulate).  If we revisit that, we could also consider reference counting to avoid the clone-inside-critical-section here.

PR #4727, opened 2021-03-04
line chart: improve major axis render

stephanwlee:
In very limited use cases of the wall time or temporal scale, the new
line chart shows a major axis to avoid overcrowding of the minor ticks
with long labels.

Previously, the major tick labels were placed only on the "beginning"
which often can be outside of the viewport. For example, for minor ticks
from "4:23" to "4:59", if the major tick goes from "<Date> 4:00" and
"<Date> 5:00", you may not be able to see the label for "4:00" at all.

This change fixes that by putting a border around, using previous
example, 4:00 to 5:00, and centering the label in the available space.

Other things you might notice are:
- refactor how ticks are generated and its data structure
- instead of using `x` and `y` property on SVG, we use padding and
  margin to position axes now (with little to no visual change).

Before:
image

After:
image

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ng.html
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ng.html
@@ -26,7 +26,7 @@ limitations under the License.
           [attr.x]="textXPosition(tick.value)"
           [attr.y]="textYPosition(tick.value)"
         >
-          {{tick.tickFormattedString }}
+          {{ tick.tickFormattedString }}
         </text>
         <title>{{ getFormatter().formatReadable(tick.value) }}</title>
       </g>
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ng.html
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ng.html
@@ -18,14 +18,18 @@ limitations under the License.
   <div class="line"></div>
   <div class="minor">
     <svg class="ticks">
-      <ng-container *ngFor="let tick of getTicks()">
-        <g>
-          <text [attr.x]="textXPosition(tick)" [attr.y]="textYPosition(tick)">
-            {{ getFormatter().formatTick(tick) }}
-          </text>
-          <title>{{ getFormatter().formatReadable(tick) }}</title>
-        </g>
-      </ng-container>
+      <g
+        class="minor-tick-label"
+        *ngFor="let tick of minorTicks; trackBy: trackByMinorTick"
+      >
+        <text
+          [attr.x]="textXPosition(tick.value)"
+          [attr.y]="textYPosition(tick.value)"
+        >
+          {{tick.tickFormattedString }}
+        </text>
+        <title>{{ getFormatter().formatReadable(tick.value) }}</title>
+      </g>
     </svg>
     <button
       [class.extent-edit-button]="true"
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ng.html
```diff
@@ -18,14 +18,18 @@
...
-      </ng-container>
+      <g
+        class="minor-tick-label"
+        *ngFor="let tick of minorTicks; trackBy: trackByMinorTick"
+      >
+        <text
+          [attr.x]="textXPosition(tick.value)"
+          [attr.y]="textYPosition(tick.value)"
+        >
+          {{tick.tickFormattedString }}
```
japie1235813, line 29: nit: a space before tick

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -59,40 +69,62 @@ export class LineChartAxisComponent {
...
+  private getMajorMinorTicks(): {major: MajorTick[]; minor: MinorTick[]} {
+    const formatter = this.getFormatter();
     const domSize = this.axis === 'x' ? this.domDim.width : this.domDim.height;
     const maxTickSize = getDomSizeInformedTickCount(domSize, this.gridCount);
-    return this.scale.ticks(this.axisExtent, maxTickSize);
+    const minorTicks: MinorTick[] = [];
+    for (const tickValue of this.scale.ticks(this.axisExtent, maxTickSize)) {
+      minorTicks.push({
+        tickFormattedString: formatter.formatTick(tickValue),
+        value: tickValue,
```
japie1235813, line 115: nit: I prefer the same order as the interface defines (alphabetically)

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -59,40 +69,62 @@ export class LineChartAxisComponent {
...
+      });
+    }
+
+    const majorTicks: MajorTick[] = [];
+    for (const tickValue of this.getMajorTickValues()) {
+      majorTicks.push({
+        tickFormattedString: formatter.formatShort(tickValue),
+        start: tickValue,
```
japie1235813, line 123: ditto

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -29,6 +29,16 @@ import {
 
 const DAY_IN_MS = 24 * 1000 * 60 * 60;
 
+interface MinorTick {
+  value: number;
+  tickFormattedString: string;
+}
+
+interface MajorTick {
+  start: number;
```
japie1235813, line 38: Could you explain more about 'start' vs 'value'? why are they named differently?

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ng.html
```diff
@@ -43,14 +47,19 @@
...
-      <text [attr.x]="textXPosition(tick)" [attr.y]="textYPosition(tick)">
-        {{ getFormatter().formatShort(tick) }}
-      </text>
-      <title>{{ getFormatter().formatReadable(tick) }}</title>
-    </g>
-  </svg>
+  <div *ngIf="majorTicks.length" class="major ticks">
+    <span
+      *ngFor="let tick of majorTicks; index as i; last as isLast; trackBy: trackByMajorTick"
+      [class.major-label]="true"
```
japie1235813, line 53: if this is always true, why don't we just use `class="major-label"`?

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
japie1235813, line 115: nit: I prefer to make the order aligned with the interface definition

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
japie1235813, line 123: ditto

japie1235813 (commented):
Except some questions, most of them I do not have strong opinions.

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ng.html
stephanwlee: The next line defines `class.last` and I am afraid that Angular does not exactly handle `class="foo"` and `[class.bar]="something"` correctly.

I am leaning towards keeping as is.

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
stephanwlee: This is alphabetical as is and I do not think keeping every object defined in the order of definition in an interface is tenable/more-readable/enforceable. Will keep it as is.

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
stephanwlee: Ditto to what I said ;)

japie1235813 (approved):

PR #4728, opened 2021-03-04
line chart: improve small number on linear scale 

stephanwlee:
When you zoom in a lot on a line chart, current line chart tries to
cramp labels with many digits into the small space causing them to
overlap and cause major visual jank.

With recent improvements to the major ticks to work on y-axis too, when
we detect small differences in the extent, we now show a major ticks for
the linear scale, too, and remove a lot of redundant long prefixes from
the minor ticks.

image

This PR contains #4727

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_utils.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_utils.ts
@@ -39,7 +39,7 @@ function getNumLeadingZerosInFractional(value: number): number {
   return 0;
 }
 
-export function getTicks(
+export function getStandardTicks(
   scale: Scale,
   formatter: Formatter,
   maxMinorTickCount: number,
@@ -66,9 +66,11 @@ export function getTicksForTemporalScale(
   const [low, high] = lowAndHigh;
   let majorTicks = scale.ticks(lowAndHigh, 2);
   if (high - low >= DAY_IN_MS || majorTicks.length > 2) {
-    majorTicks = [];
+    // Return standard ticks.
+    return getStandardTicks(scale, formatter, maxMinorTickCount, lowAndHigh);
   }
 
+  // Do something special for this special case.
   const minorTickVals = scale.ticks(lowAndHigh, maxMinorTickCount);
   return {
     major: majorTicks.map((tickVal) => {
@@ -95,9 +97,11 @@ export function getTicksForLinearScale(
   const [low, high] = lowAndHigh;
   const diff = Math.abs(high - low);
   if (diff > 1e-3) {
-    return getTicks(scale, formatter, maxMinorTickCount, lowAndHigh);
+    // Return standard ticks.
+    return getStandardTicks(scale, formatter, maxMinorTickCount, lowAndHigh);
   }
 
+  // Do something special for this special case.
   const minorTickVals = scale.ticks([low, high], maxMinorTickCount);
   const numFractionalToKeep = getNumLeadingZerosInFractional(diff);
   const majorTickVals = scale.ticks([low, high], 2);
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+    return Number(maybeExponential[1]) - 1;
+  }
+  return 0;
+}
+
+export function getTicksForLinearScale(
+  scale: LinearScale,
+  formatter: Formatter,
+  maxMinorTickCount: number,
+  low: number,
```
bmd3k, line 44: For consistency, would it make sense to just pass type `range: [number, number]`. Seems like this is a common type in your Scale and Axis code.

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+      minor: minorTickVals.map((tickVal) => {
+        return {
+          value: tickVal,
+          tickFormattedString: formatter.formatTick(tickVal),
+        };
+      }),
+    };
+  }
+
+  const numFranctionalToKeep = getNumLeadingZerosInFractional(diff);
```
bmd3k, line 62: Franctional -> Fractional

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+  formatter: Formatter,
+  maxMinorTickCount: number,
+  low: number,
+  high: number
+): {minor: MinorTick[]; major: MajorTick[]} {
+  const minorTickVals = scale.ticks([low, high], maxMinorTickCount);
+
+  const diff = Math.abs(high - low);
+  if (diff > 1e-3) {
+    return {
```
bmd3k, line 51: From what I can reason out, this is the same return value for any other Scale that is not TemporalScale. Can you refactor the code so you only have one place that generates this?

Something like:

```
if (scale instance of TemporalScale) {
  // generate TemporalScale specific thing
} else if (scale instance of LinearScale and diff <= 1e-3) {
  // generate small LinearScale specific thing 
} else {
  // Generate default thing.
  const minorTickVals = scale.ticks(this.axisExtent, maxMinorTickCount);
  return {
      major: [],
      minor: minorTickVals.map((tickVal) => {
        return {
          value: tickVal,
          tickFormattedString: formatter.formatTick(tickVal),
        };
      }),
    };
}
```

It would also address what I think is some odd asymmetry in how you change behavior based on Scale type. You branch for LinearScale at ngOnChanges() but you branch for TemporalScale further down in call hierarchy at getMajorTickValues(). It's a bit difficult to follow.

stephanwlee: I initially had a code like that but the "small LinearScale specific thing" was not so small and it was just generally hard to follow the logic when done that way. In my trial, division by the scale type was the easiest to follow as long as all implementations conform to the interface.

bmd3k: I think it's worth illustrating my suggestion:
https://github.com/tensorflow/tensorboard/compare/bb3ce0eaa...1b41b0c29

I find that significantly easier to read:
1. It's clear that different Scales are handled differently.
2. It's clearer that LinearScale and TemporalScale can (often?) generate the standard tick values but there are clear special cases where they do not.
3. There is less code duplication in generating the standard tick values. Actually, even stronger: It is nice that we don't use two entirely different ways to generate the same standard result.
4. There is less code duplication before and after generating the ticks: Only calculdate maxMinorTickCount once; only translate the results into this.minorTicks and this.majorTicks once.

stephanwlee: Yeah, that's not that far from my ideal situation; to clarify, I did not like the code when minor and major generations broken up and each requiring some switch statements. I did not find temporal and default scale generation to be complex to warrant the move but I definitely can see why that is correct. I will add tests to the utility promptly.

bmd3k (approved):
Thanks for your patience!

stephanwlee:
Thanks for the review!

PR #4731, opened 2021-03-04
npmi: remove vz_projector dependency

stephanwlee:
Internally, we have conformance check to not include certain sources
into our service binaries and vz_projector dependency violates the
conformance.

This change simply copies one method and reimplement the range.


psybuzz (approved):
Thanks for removing this npmi -> projector dep!

PR #4733, opened 2021-03-05
runs table: support hparam and metrics sorting

stephanwlee:
This change completes the implementation of the runs table sorting.
Previously, we only supported sorting by experiment name and runs name.
However, with this change, we want to sort by an arbitrary hparam or
metric key user provided.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
@@ -53,6 +53,8 @@ import {
   DiscreteHparamValues,
   DomainType,
   IntervalFilter,
+  SortKey,
+  SortType,
 } from '../../types';
 
 import {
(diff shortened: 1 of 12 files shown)


tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
```diff
@@ -111,6 +129,10 @@ function sortRunTableItems(
         continue;
       }
 
+      if (valA === undefined || valB === undefined) {
+        return valB === undefined ? -1 : 1;
+      }
```
psybuzz, line 134: Just checking: should this also take into account the `sort.direction`?  I don't have a strong opinion, also fine with keeping as is.

psybuzz (approved):
I haven't tried it manually, but code looks good!

tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
stephanwlee: I intended for undefined to always appear at the end after non-undefined values regardless of the sort direction. It felt very natural to me and I can be convinced otherwise.

stephanwlee:
Tested the sync and tested with the internal version of TB. Seems nominal. cl/361576880.

PR #4734, opened 2021-03-05
rust: implement data_compat.rs basic tensor support

nfelt:
This wires up `data_compat.rs` handling for tensors, as part of the tensor support subtask of #4422, which includes:

- adding `initial_metadata()` special-cases for tensor-dataclass plugins
- adding `into_tensor()` for enriching tensor values
- calling `into_tensor()` from within RunLoader
- tests for the above

I also adjusted some naming of other tests within `data_compat.rs` to be (IMO) slightly more precise and clear now that we have the new tensor test methods.

Note that we don't yet handle A) events with no summary value that should nonetheless be treated as having implicit Tensor values (the use case being hparams metadata persistence), and B) legacy histogram data stored in `HistogramProto`.  I've left those for later PRs.

Test plan: I confirmed that with this change and the preceding RPC PR, we can now render the "normal" tensor dashboards.   In particular, what now works is: text, histograms/distributions for data written with the tensor-based histogram summary APIs, PR curves, and NPMI (the visualization seems somewhat broken, at least with the demo data, but it's equivalently broken with and without `--load_fast` now). Hparams doesn't work yet per the note above. The mesh plugin hasn't been DataProviderized; I included it for completeness anyway but it's untested.

### DIFF
--- a/tensorboard/data/server/data_compat.rs
+++ b/tensorboard/data/server/data_compat.rs
@@ -38,6 +38,12 @@ pub(crate) mod plugin_names {
     pub const GRAPH_RUN_METADATA: &str = "graph_run_metadata";
     pub const GRAPH_RUN_METADATA_WITH_GRAPH: &str = "graph_run_metadata_graph";
     pub const GRAPH_KERAS_MODEL: &str = "graph_keras_model";
+    pub const HISTOGRAMS: &str = "histograms";
+    pub const TEXT: &str = "text";
+    pub const PR_CURVES: &str = "pr_curves";
+    pub const HPARAMS: &str = "hparams";
+    pub const MESH: &str = "mesh";
+    pub const NPMI: &str = "npmi";
 }
 
 /// The inner contents of a single value from an event.
@@ -82,6 +88,23 @@ impl EventValue {
                 Some(f) => Ok(ScalarValue(f)),
                 None => Err(DataLoss),
             },
+            _ => Err(DataLoss),
+        }
+    }
+
+    /// Consumes this event value and enriches it into a tensor.
+    ///
+    /// This supports summaries with `tensor` populated.
+    //
+    // TODO(#4422): support conversion of other summary types to tensors.
+    pub fn into_tensor(self, _metadata: &pb::SummaryMetadata) -> Result<pb::TensorProto, DataLoss> {
+        let value_box = match self {
+            EventValue::GraphDef(_) => return Err(DataLoss),
+            EventValue::TaggedRunMetadata(_) => return Err(DataLoss),
+            EventValue::Summary(SummaryValue(v)) => v,
+        };
+        match *value_box {
+            pb::summary::value::Value::Tensor(tp) => Ok(tp),
             _ => Err(DataLoss),
         }
     }
@@ -272,6 +295,14 @@ impl SummaryValue {
                     Some(plugin_names::SCALARS) => {
                         md.data_class = pb::DataClass::Scalar.into();
                     }
+                    Some(plugin_names::HISTOGRAMS)
+                    | Some(plugin_names::TEXT)
+                    | Some(plugin_names::HPARAMS)
+                    | Some(plugin_names::PR_CURVES)
+                    | Some(plugin_names::MESH)
+                    | Some(plugin_names::NPMI) => {
+                        md.data_class = pb::DataClass::Tensor.into();
+                    }
                     Some(plugin_names::IMAGES)
                     | Some(plugin_names::AUDIO)
                     | Some(plugin_names::GRAPH_RUN_METADATA)
@@ -446,7 +477,7 @@ mod tests {
         }
 
         #[test]
-        fn test_enrich_valid_tensors() {
+        fn test_enrich_rank_0_tensors() {
             let tensors = vec![
                 pb::TensorProto {
                     dtype: pb::DataType::DtFloat.into(),
@@ -480,7 +511,7 @@ mod tests {
         }
 
         #[test]
-        fn test_enrich_short_tensors() {
+        fn test_enrich_rank_0_tensors_corrupted_with_short_data() {
             let tensors = vec![
                 pb::TensorProto {
                     dtype: pb::DataType::DtFloat.into(),
@@ -508,7 +539,7 @@ mod tests {
         }
 
         #[test]
-        fn test_enrich_long_tensors() {
+        fn test_enrich_rank_0_tensors_corrupted_with_long_data() {
             let tensors = vec![
                 pb::TensorProto {
                     dtype: pb::DataType::DtFloat.into(),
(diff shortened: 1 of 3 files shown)


tensorboard/data/server/data_compat.rs
```diff
@@ -38,6 +38,12 @@ pub(crate) mod plugin_names {
     pub const GRAPH_RUN_METADATA: &str = "graph_run_metadata";
     pub const GRAPH_RUN_METADATA_WITH_GRAPH: &str = "graph_run_metadata_graph";
     pub const GRAPH_KERAS_MODEL: &str = "graph_keras_model";
+    pub const HISTOGRAMS: &str = "histograms";
+    pub const TEXT: &str = "text";
+    pub const PR_CURVES: &str = "pr_curves";
+    pub const HPARAMS: &str = "hparams";
+    pub const MESH: &str = "mesh";
```
wchargin, line 45: Oh, interesting. I hadn’t planned to accommodate `mesh` because it
wasn’t obvious to me that the data fit well into the existing storage
classes. It’s not clearly `Tensor` because the data may be arbitrarily
large, and plausibly over our recommended limits. It’s not clearly
`BlobSequence` because it’s the wrong shape—though we could faithfully
reinterpret-cast without too much trouble, since all dimensions but one
are constant.

But I’m happy to just make it `Tensor` as you do here. I’m not sure how
widely used it is in conjunction with environments that would actually
care about the size limit (e.g., TensorBoard.dev uploads), so it may
just be a non-issue. As you note, the plugin backend still uses the
multiplexer, but we can fix that.

tensorboard/data/server/data_compat.rs
```diff
@@ -38,6 +38,12 @@ pub(crate) mod plugin_names {
...
+    pub const NPMI: &str = "npmi";
```
wchargin, line 46: I think [nPMI data has always had `DATA_CLASS_TENSOR`][1] at rest. Can
we get away with dropping the special case here?

[1]: https://github.com/tensorflow/tensorboard/blame/1e7a722c49829351235308d8fdc31e28d4a04383/tensorboard/plugins/npmi/metadata.py#L45

tensorboard/data/server/data_compat.rs
```diff
@@ -619,6 +650,76 @@ mod tests {
...
+        fn test_metadata_tensor_with_dataclass() {
+            let md = pb::SummaryMetadata {
+                plugin_data: Some(PluginData {
+                    plugin_name: "rando".to_string(),
+                    content: Bytes::from_static(b"preserved!"),
+                    ..Default::default()
+                }),
+                data_class: pb::DataClass::Tensor.into(),
+                ..Default::default()
+            };
```
wchargin, line 666: nit: We do have `blank_with_plugin_content` (per your suggestion), in
case you’d like to use that? Also suffices for the next test case, since
you can pass `DataClass::Unknown`.

tensorboard/data/server/data_compat.rs
```diff
@@ -86,6 +92,23 @@ impl EventValue {
         }
     }
 
+    /// Consumes this event value and enriches it into a tensor.
+    ///
+    /// This supports summaries with `tensor` populated.
+    //
+    // TODO(#4422): support conversion of other summary types to tensors.
+    pub fn into_tensor(self, _metadata: &pb::SummaryMetadata) -> Result<pb::TensorProto, DataLoss> {
```
wchargin, line 100: Curious what the thought process is behind taking `_metadata` here?
I omitted it from `into_scalar` because it’s not actually needed, but
that does make the signatures different and makes the `RunLoader`
integration ever so slightly more awkward. I don’t feel strongly either
way, just wondering about the rationale.

tensorboard/data/server/data_compat.rs
```diff
@@ -619,6 +650,76 @@ mod tests {
...
+                string_val: vec![Bytes::from_static(b"foo")],
+                ..Default::default()
+            })));
+            let result = v.initial_metadata(Some(md.clone()));
+            assert_eq!(*result, md);
+        }
+
+        #[test]
+        fn test_metadata_tensor_without_dataclass() {
+            for plugin_name in vec![
```
wchargin, line 678: Clippy says:

```
warning: useless use of `vec!`
   --> data_compat.rs:678:32
    |
678 |               for plugin_name in vec![
    |  ________________________________^
679 | |                 plugin_names::HISTOGRAMS,
680 | |                 plugin_names::TEXT,
681 | |                 plugin_names::PR_CURVES,
...   |
684 | |                 plugin_names::NPMI,
685 | |             ] {
    | |_____________^
    |
    = note: `#[warn(clippy::useless_vec)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_vec
help: you can use a slice directly
    |
678 |             for plugin_name in &[plugin_names::HISTOGRAMS,
679 |                 plugin_names::TEXT,
680 |                 plugin_names::PR_CURVES,
681 |                 plugin_names::HPARAMS,
682 |                 plugin_names::MESH,
683 |                 plugin_names::NPMI] {
    |
```

(In case you’re wondering, the reason that these checks don’t show up
inline is that [our Cargo package is not at repository root][1].)

[1]: https://github.com/actions-rs/clippy-check/issues/28#issuecomment-680401392

wchargin (approved):
Nice; thanks! Only request is to remove nPMI special-casing if we can;
if that doesn’t work for some reason, happy to keep it. (I do want to
support it; I just think that we might get it for free.)

tensorboard/data/server/data_compat.rs
nfelt: > (per your suggestion)

Done, lol.

tensorboard/data/server/data_compat.rs
nfelt: Done, thanks.

tensorboard/data/server/data_compat.rs
wchargin: > I mistakenly figured I was going to need it later (to handle empty hparams conversions)

Me, too. :-)

No preference.

tensorboard/data/server/data_compat.rs
wchargin: Yeah—in the past, I think, we’d agreed that it was “reasonable” to
assume that text would be small, even though it’s not bounded. The
unbounded-size tensor is also approximately what I was thinking, as
“basically sugar for a length-1 blob sequence, but we handle the
serialization for you”. From that lens, I don’t know that we actually
need a separate data type—`emit_tensor(..., as_blob=True)` or something
might suffice.

Happy to leave as is for now.

wchargin (approved):

tensorboard/data/server/data_compat.rs
nfelt: Ack. Will leave as-is in that case.

tensorboard/data/server/data_compat.rs
nfelt: True, I guess it basically is just sugar for a length-1 blob sequence, so it could be persisted as such.  But I think we'd probably need a new data type at the `DataProvider` level if we wanted to both handle the Tensor serialization and also support a streaming read, right?

PR #4738, opened 2021-03-05
graph: fix `NotFoundError` typo

wchargin:
Summary:
When requesting a blob that doesn’t exist, we intend to 404, but instead
500 because the error type is called `NotFoundError`, not `NotFound`.

Test Plan:
Regression test included.

wchargin-branch: graph-notfound


wchargin:
cc @azhmogin-google

stephanwlee (approved):

PR #4740, opened 2021-03-05
rust: support legacy TF 1.x histogram summaries

nfelt:
This implements support in Rustboard for converting legacy TF 1.x histogram summaries that use `HistogramProto` into the k-by-3 tensor format that TensorBoard expects.  We adopt the same logic as `data_compat.py` for consistency, even though the semantics are a bit dubious.

Test plan: unit tests, plus I manually generated an organic TF 1.x histogram summary and confirmed that it now A) shows up in Rustboard-backed TB and B) looks the same as in slow-loading TB.

Part of #4422 tensor support sub-task.

### DIFF
--- a/tensorboard/data/server/data_compat.rs
+++ b/tensorboard/data/server/data_compat.rs
@@ -391,6 +391,7 @@ fn tf1x_audio_metadata() -> Box<pb::SummaryMetadata> {
 mod tests {
     use super::*;
     use pb::summary::value::Value;
+    use std::iter::FromIterator;
 
     fn tensor_shape(dims: &[i64]) -> pb::TensorShapeProto {
         pb::TensorShapeProto {
@@ -403,6 +404,24 @@ mod tests {
                 .collect(),
             ..Default::default()
         }
+    }
+
+    /// Macro to construct `Bytes` from an array of numerics having a `to_le_bytes()` method.
+    ///
+    /// Use like: `let b = to_le_bytes![1.0, 2.0f32];`
+    ///
+    /// Note the suffixed literal, which is necessary so the compiler knows the intended type.
+    //
+    // TODO(Rust 1.51): I think we can avoid intermediate Vecs with array::IntoIter from
+    // https://github.com/rust-lang/rust/pull/80470, e.g. with:
+    //   Bytes::from_iter([].iter().flat_map(|v| array::IntoIter::new(v.to_le_bytes())))
+    macro_rules! to_le_bytes {
+        ($($x:expr),+ $(,)?) => (
+            Bytes::from_iter([$($x),+].iter()
+                .map(|v| v.to_le_bytes())
+                .collect::<Vec<_>>()
+                .concat())
+        );
     }
 
     mod scalars {
@@ -484,13 +503,13 @@ mod tests {
                 pb::TensorProto {
                     dtype: pb::DataType::DtFloat.into(),
                     tensor_shape: Some(tensor_shape(&[])),
-                    tensor_content: Bytes::from(f32::to_le_bytes(0.125).to_vec()),
+                    tensor_content: to_le_bytes![0.125f32],
                     ..Default::default()
                 },
                 pb::TensorProto {
                     dtype: pb::DataType::DtFloat.into(),
                     tensor_shape: None, // no explicit tensor shape; treated as rank 0
-                    tensor_content: Bytes::from(f32::to_le_bytes(0.125).to_vec()),
+                    tensor_content: to_le_bytes![0.125f32],
                     ..Default::default()
                 },
             ];
@@ -518,7 +537,7 @@ mod tests {
                 pb::TensorProto {
                     dtype: pb::DataType::DtFloat.into(),
                     tensor_shape: Some(tensor_shape(&[])),
-                    tensor_content: Bytes::from(f32::to_le_bytes(0.125)[..2].to_vec()),
+                    tensor_content: to_le_bytes![0.125f32].slice(..2),
                     ..Default::default()
                 },
             ];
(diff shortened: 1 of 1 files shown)


tensorboard/data/server/data_compat.rs
```diff
@@ -103,6 +105,51 @@ impl EventValue {
...
+                                size: num_buckets as i64,
+                                ..Default::default()
+                            },
+                            pb::tensor_shape_proto::Dim {
+                                size: 3,
+                                ..Default::default()
+                            },
+                        ],
+                        ..Default::default()
+                    }),
```
wchargin, line 148: Optional: this is just `tensor_shape(&[num_buckets as i64, 3])` if you
promote `fn tensor_shape` from a test helper to a real helper.

tensorboard/data/server/data_compat.rs
```diff
@@ -103,6 +105,51 @@ impl EventValue {
         };
         match *value_box {
             pb::summary::value::Value::Tensor(tp) => Ok(tp),
+            pb::summary::value::Value::Histo(hp) => {
+                // Migrate legacy TF 1.x HistogramProto to TensorProto. The "spec" in summary.proto
+                // says `bucket` and `bucket_limit` are parallel arrays encoding bucket counts and
+                // bucket right edges; the first bucket's left edge is assumed to be -DBL_MAX and
+                // subsequent left edges are defined as the right edge of the preceeding bucket.
+                //
+                // Our conversion logic in data_compat.py however disobeys this and instead sets the
```
wchargin, line 114: Sigh—my bad, probably. Put it on the list of “ways in which the
histogram transformation pipeline makes zero sense”.

tensorboard/data/server/data_compat.rs
```diff
@@ -103,6 +105,51 @@ impl EventValue {
...
+                // leftmost and rightmost edges to the `min` and `max` values, respectively. This
+                // will result in the outermost buckets having left edge > right edge if they were
+                // originally empty. Apparently the histogram visualization can't handle buckets
+                // extending to -/+ DBL_MAX, but can handle buckets of negative width?
+                //
+                // For consistency with the status quo, we replicate this questionable logic here.
+                let num_buckets = hp.bucket.len();
+                let bucket_edges = || hp.bucket_limit.iter().take(num_buckets - 1).copied();
```
wchargin, line 122: Shall we check that `bucket.len() == bucket_limit.len()` and signal data
loss otherwise? or at least take `num_buckets` to be the shorter of the
two lengths? As written, it looks like `bucket_lefts` could be much
shorter than `bucket_counts`, which would make the output shape not
correct.

Also, if `hp.bucket.len() == 0`, the `num_buckets - 1` is an error
(panic in debug mode), so let’s do something nicer in that case?
Maybe worth a test (up to you).

In case you’re interested, my [sketch of this routine][1] took quite a
different approach (imperative instead of streamy). I think that I like
your idea to use `tensor_content`, since at least cloning a `Bytes` is
cheap compared to cloning a `Vec<f64>`.

[1]: https://github.com/wchargin/rustboard-server/blob/05550d7b356b94c7e0dea335030d4d11ab6510b8/src/run.rs#L257

tensorboard/data/server/data_compat.rs
```diff
@@ -103,6 +105,51 @@ impl EventValue {
...
+                                ..Default::default()
+                            },
+                            pb::tensor_shape_proto::Dim {
+                                size: 3,
+                                ..Default::default()
+                            },
+                        ],
+                        ..Default::default()
+                    }),
+                    tensor_content: tensor_content,
```
wchargin, line 149: nit: Clippy says to pun this (`tensor_content,`).

tensorboard/data/server/data_compat.rs
```diff
@@ -720,6 +768,61 @@ mod tests {
...
+        fn test_enrich_tf1x_histogram() {
+            let hp = pb::HistogramProto {
+                min: -1.999,
+                max: 1.999,
+                num: 60.0,
+                bucket_limit: vec![-2.0, -1.0, 0.0, 1.0, 2.0, f64::MAX],
+                bucket: vec![0.0, 10.0, 20.0, 20.0, 10.0, 0.0],
+                ..Default::default()
+            };
+            let v = EventValue::Summary(SummaryValue(Box::new(Value::Histo(hp.clone()))));
```
wchargin, line 782: nit: Clippy says these clones are useless (`hp.clone()` → `hp`, and below).

wchargin (commented):
> I manually generated an organic TF 1.x histogram summary

You can use `gs://tensorboard-bench-logs/mnist` here, too. Can confirm
that it shows histograms at this commit but not before it, even with
your core tensor support.

nfelt (commented):
PTAL

tensorboard/data/server/data_compat.rs
wchargin: Lovely; thanks! Yeah, I’m perfectly happy with the streams here. I’m
part of the “use streams judiciously” crowd, and this looks like a
pretty easy sell. It’s definitely more readable than mine.

tensorboard/data/server/data_compat.rs
wchargin: Sure, sounds good. Appreciate the awareness of the yak density and happy
that you feel comfortable calling that out.

wchargin (approved):

tensorboard/data/server/data_compat.rs
```diff
@@ -118,10 +118,15 @@ impl EventValue {
...
+                if hp.bucket.len() != hp.bucket_limit.len() {
+                    return Err(DataLoss);
+                }
                 let num_buckets = hp.bucket.len();
-                let bucket_edges = || hp.bucket_limit.iter().take(num_buckets - 1).copied();
-                let bucket_lefts = iter::once(hp.min).chain(bucket_edges());
-                let bucket_rights = bucket_edges().chain(iter::once(hp.max));
+                // Skip the last `bucket_limit`; it gets replaced by `hp.max`. It's okay to ignore
+                // the edge case at 0 since `.zip()` will stop immediately in that case anyway.
+                let bucket_edges = &hp.bucket_limit[..usize::saturating_sub(num_buckets, 1)];
```
wchargin, line 127: This can be `num_buckets.saturating_sub(1)` if you want, but I don’t mind it like this.

nfelt: Ack, thanks for the observation.  I guess I didn't really think about whether `num_buckets` is already `usize`.  Will leave as-is just to avoid re-pushing only for this.

PR #4741, opened 2021-03-06
graph: notify consumers when Hierarchy's templates update (3)

psybuzz:
The motivation behind these changes is to allow the `.templates` field,
currently stored in the Hierarchy, to be computed lazily, e.g. only if/when
the user selects the "Color by structure" mode.

This introduces a "Dispatcher" class, used to give the `Hierarchy` the
traits to allow it to dispatch events and let Hierarchy clients register to
listen to those events. Doing so allows components inside `<tf-node-info>`
and `<tf-graph-scene>`, who currently read the templates, to listen for
changes.

For reference, the hierarchy of Polymer components looks like:
```html
- <tf-graph-dashboard>
  - <tf-graph-dashboard-loader {{graphHierarchy}}>; creates GraphDef, SlimGraph, Hierarchy, also used externally
  - <tf-graph-controls>; also used externally
  - <tf-graph-board [[graphHierarchy]]>; also used externally
    - <tf-graph {{graphHierarchy}}>; creates Hierarchy, RenderHierarchy
      - <tf-graph-scene>
        - <tf-graph-minimap>
    - <tf-graph-info [[graphHierarchy]]> (floating info panels on the right)
      - <tf-node-info>
        - <tf-node-icon [[_templateIndex]]>
        - <tf-node-list-item [[_templateIndex]]>
```

Our use case requires
- some way to inform all components who depend on the templates,
  that they have changed.
- one component to be responsible for the business logic
  (only hierarchy.updateTemplates() if colorBy === structure)

Currently, the templateIndex, when computed, is not passed via
Polymer data bindings. Instead, consumers receive the Hierarchy and call its
`.getTemplateIndex()` method.

This Dispatcher trait approach achieves the goals without having to do a
major refactoring, without modifying `<tf-graph-board>`'s public fields,
and without making Hierarchy aware of the `ColorBy` concept.

Alternatives considered
- Move Hierarchy to `<tf-graph-board>`, make it responsible for propagating 
  all Hierarchy event notifications to descendants. Notably, changing
  `<tf-graph-board>` would affect internal consumers who use it publicly
  (unfortunately).
- Make the `<tf-graph-controls>` responsible for updating the hierarchy's
  templates, and use data binding to notify everyone of the `[[templateIndex]]`.
  Polymer's order of events made this complex.
- Don't pass the Hierarchy from the `<tf-graph>` to the `<tf-graph-info>`.

Diffbase: https://github.com/tensorflow/tensorboard/pull/4725
Followup: https://github.com/tensorflow/tensorboard/pull/4742

### DIFF
--- a/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
+++ b/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
@@ -33,10 +33,9 @@ import {template} from './tf-graph-scene.html';
 import {LegacyElementMixin} from '../../../components/polymer/legacy_element_mixin';
 import {TfGraphScene} from '../tf_graph_common/tf-graph-scene';
 import {ColorBy} from '../tf_graph_common/view_types';
-import {HierarchyEvent} from '../tf_graph_common/hierarchy';
 
 @customElement('tf-graph-scene')
-class TfGraphScene2
+export class TfGraphScene2
   extends LegacyElementMixin(PolymerElement)
   implements TfGraphScene {
   static readonly template = template;
@@ -258,13 +257,6 @@ class TfGraphScene2
   /** Main method for building the scene */
   _build(renderHierarchy: tf_graph_render.RenderGraphInfo) {
     this.templateIndex = renderHierarchy.hierarchy.getTemplateIndex();
-    renderHierarchy.hierarchy.addListener(
-      HierarchyEvent.TemplatesUpdated,
-      () => {
-        this.templateIndex = renderHierarchy.hierarchy.getTemplateIndex();
-        this._nodeColorsChanged();
-      }
-    );
     tf_graph_util.time(
       'tf-graph-scene (layout):',
       function () {
@@ -480,8 +472,12 @@ class TfGraphScene2
    * UI controls.
    */
   @observe('colorBy')
-  _nodeColorsChanged() {
+  nodeColorsChanged() {
     if (this.renderHierarchy != null) {
+      // Formatters will read `sceneElement.templateIndex` directly.
+      // Ensure that it is up to date.
+      this.templateIndex = this.renderHierarchy.hierarchy.getTemplateIndex();
+
       // We iterate through each svg node and update its state.
       _.each(this._nodeGroupIndex, (nodeGroup, nodeName) => {
         this._updateNodeState(nodeName);
--- a/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
+++ b/tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
@@ -33,6 +33,7 @@ import {template} from './tf-graph-scene.html';
 import {LegacyElementMixin} from '../../../components/polymer/legacy_element_mixin';
 import {TfGraphScene} from '../tf_graph_common/tf-graph-scene';
 import {ColorBy} from '../tf_graph_common/view_types';
+import {HierarchyEvent} from '../tf_graph_common/hierarchy';
 
 @customElement('tf-graph-scene')
 class TfGraphScene2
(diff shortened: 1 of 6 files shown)


tensorboard/plugins/graph/tf_graph_common/hierarchy.ts
```diff
@@ -48,10 +48,17 @@ import * as tf_graph from './graph';
 import * as tf_graph_proto from './proto';
 import * as tf_graph_util from './util';
 
+export enum HierarchyEvent {
+  /**
+   * Fired when the templates may have been updated. No event payload attached.
+   */
+  TemplatesUpdated = 'TemplatesUpdated',
```
stephanwlee, line 55: ENUMs are often uppercased. `TEMPLATES_UPDATED = 'TemplatesUpdated',`.

Also, when the value of the enum is not useful, use the default numeric one.

tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
```diff
@@ -257,6 +258,13 @@ class TfGraphScene2
   /** Main method for building the scene */
   _build(renderHierarchy: tf_graph_render.RenderGraphInfo) {
     this.templateIndex = renderHierarchy.hierarchy.getTemplateIndex();
+    renderHierarchy.hierarchy.addListener(
+      HierarchyEvent.TemplatesUpdated,
+      () => {
+        this.templateIndex = renderHierarchy.hierarchy.getTemplateIndex();
+        this._nodeColorsChanged();
+      }
+    );
```
stephanwlee, line 267: (We've talked about this once before; don't know if there was a conclusion we drew from it) I prefer there being explicit destruct/removeListener if possible at all. Did you consider using a removeListener? If you were to add one, it is often easier to remove with a `key` instead of a pointer to a function instance.

stephanwlee (commented):
I did not review `colorbynone2` so my comment may be off basis but having a listener attached to a data structure this way (given two way binding) feels a bit unfortunate.

There were few things you've considered but did you consider having a module level singleton dispatcher object that you use to subscribe and unsubscribe?

```ts
// dispatcher.ts

type EventKey = string;
const listeners: Map<EventKey, Function> = new Map();

export function addListener(type: string, fn: Function) {
  ...
}

export function dispatch(type: string) {
  ...
}
```

tensorboard/plugins/graph/tf_graph_common/hierarchy.ts
psybuzz: Thanks, done.

tensorboard/plugins/graph/tf_graph/tf-graph-scene.ts
psybuzz: I do recall us discussing whether to call `removeEventListener()` in the context of a browser event listener.  iirc, the concern was that browsers would not clean up properly and cause extra memory to be retained.  I don't think this is the case, but I wasn't able to prove its veracity at that time.

In this particular PR, the context is slightly different.  `Hierarchy` is a plain JS class, and listener management is done by us (the application), not the browser's renderer.  As long as there is a path from the application's root to the `Hierarchy`, the `Hierarchy` will not be garbage collected, and neither will its listeners.  However, once the `Hierarchy` changes, due to loading a new graph, or expanding a grouped series node, we lose all references to it.  In this PR, the listeners on the Hierarchy are no longer accessible from the application's root, so they should be garbage collected as well.

I think the pattern in this PR boils down to something like https://gist.github.com/psybuzz/f368886b99b542156e5f7280cc9966b7
e.g. clients can add listeners to the hierarchy and never remove them, and once the hierarchy changes, the listeners will not be marked as collectable by the GC

psybuzz (commented):
> I did not review colorbynone2 so my comment may be off basis

The diffbase PRs were purely about adding/removing types, interfaces, no new patterns.

> having a module level singleton dispatcher object that you use to subscribe and unsubscribe?

An interesting idea. That would create a module-level pubsub letting anything inside graph/ listen to events such as TEMPLATES_UPDATED. For now, I prefer the class-level Dispatcher, mainly because it is more tightly scoped. There are only a handful of classes that need to be able to listen to Hierarchy events so far, and they all already own a Hierarchy.

I realized I made a mistake in my last patch, where the tf-graph-scene adds a listener, when it doesn't own a Hierarchy. Other classes, such as tf-graph-minimap, tf-graph-controls, are similar.

If there was a use case for other modules/features outside of graph/ to listen for changes in, say, selectedNode, then I think having a singleton dispatcher at the graph/ level makes more sense.

PR #4745, opened 2021-03-06
refactor: move hparams state out of runs

stephanwlee:
This change largely takes hparams and its filter data from runs_reducer
and reimplements `eid` indexed data.

New reducer is not yet integrated into the app and will be done in a follow up.

### DIFF
--- a/tensorboard/webapp/hparams/_redux/hparams_actions.ts
+++ b/tensorboard/webapp/hparams/_redux/hparams_actions.ts
@@ -1,0 +1,55 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+/**
+ * @fileoverview Hparams Ngrx actions.
+ */
+
+import {createAction, props} from '@ngrx/store';
+
+import {DiscreteHparamValues} from '../types';
+
+/** @typehack */ import * as _typeHackModels from '@ngrx/store/src/models';
+
+export const hparamsDiscreteHparamFilterChanged = createAction(
+  '[Runs] Run Discrete Hparam Filter Changed',
+  props<{
+    experimentId: string;
+    hparamName: string;
+    filterValues: DiscreteHparamValues;
+    includeUndefined: boolean;
+  }>()
+);
+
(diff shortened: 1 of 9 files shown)


tensorboard/webapp/hparams/_redux/hparams_actions.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import {createAction, props} from '@ngrx/store';
+
+import {DiscreteHparamValues} from '../types';
+
+/** @typehack */ import * as _typeHackModels from '@ngrx/store/src/models';
+
+export const hparamsDiscreteHparamFilterChanged = createAction(
+  '[Runs] Run Discrete Hparam Filter Changed',
```
psybuzz, line 26: "[Runs]" --> "[Hparams]"?

tensorboard/webapp/hparams/_redux/hparams_selectors.ts
```diff
@@ -0,0 +1,101 @@
...
+  (data: HparamsMetricsAndFilters | undefined): Map<string, IntervalFilter> => {
+    if (!data) return new Map();
+    return data.metric.filters;
+  }
+);
+
+/**
+ * Returns Observable that emits map of hparam name to filter values.
+ */
+export const getRunHparamFilterMap = createSelector(
```
psybuzz, line 81: nit: When this selector was under the runs/ feature, "Run" made sense as something to namespace the selector.  If the "Run" in "getRunHparamFilterMap" is no longer relevant, let's drop it?

tensorboard/webapp/hparams/_redux/hparams_types.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+import {DiscreteFilter, HparamSpec, IntervalFilter, MetricSpec} from '../types';
+
+/**
+ * Key used to namespace the hparams reducer.
+ */
+export const HPARAMS_FEATURE_KEY = 'hparams';
+
+export type ExperimentId = string;
```
psybuzz, line 26: I see that this type is a direct copy from runs/, but I don't think it makes sense for us to add a new "public" `ExperimentId`.  Can we drop the "export" on this type?  The runs/ variant of ExperimentId is not used outside of runs_types.ts, afaik.

psybuzz (commented):
Overall the code looks good, but I'm not aware of the motivation behind the refactor.  Are there new "hparam" features that are not relevant to "runs", that are planned, and this change prepares us by avoiding too much bloat in runs/?

stephanwlee:
Ah, I indeed forgot to externalize my thoughts. As you have suspected, I wanted to debloat the runs reducer with its massive state. Hparams, after all, is its own entity as it should be usable by the hparams plugin.

psybuzz (approved):
Rationale is fair, lgtm!

Note: even though the hparams_reducers are not hooked up in this PR, the missing hparams_reducers_test.ts could still be ported in this PR as well, I'm guessing?  I'm also fine with moving those in a separate PR, if more convenient.

stephanwlee:
Yup, that's correct. Size of the PR was getting concerning so I think I will do it in a quick follow up before I integrate this into the views.

PR #4746, opened 2021-03-07
[npmi] Simplified UMAP Data Structure

Sparkier:
As we only use umap and only in 2D for now, I simplified the data structure.

Changed how projections are stored in our embedding datasets.


Sparkier:
@stephanwlee I changed the data structure for UMAP. Not sure if this matches your intention, though.

stephanwlee (approved):

stephanwlee:
FYI: the test failure seems real 
image

Issue #4744, opened 2021-03-06
Graph dashboard pixelated after zoom


wchargin:
Help! I need more pixels. Does anybody have more pixels?

![Screenshot of the graphs dashboard. The viewport is zoomed in around
one node. The node is heavily pixelated, by about a factor of 10, and
it’s hard to read the writing. Some other parts of the viewport appear
fine: although the node’s rounded rectangle and its label pixelated, the
arrows connecting nodes are crisp, as are their labels.][ss]

This is in 2021-03-05 nightlies, but isn’t super recent. Will try to
bisect and report back.

[ss]: https://user-images.githubusercontent.com/4317806/110192023-9459ee00-7de0-11eb-991d-4146d88d7293.png

wchargin:
Oh, looks like it’s actually a Chrome thing:

  - Chrome 89: broken all the way back to `tb-nightly==2.5.0a20210217`,
    and I don’t think that we’ve just missed it for that long;
  - Firefox 86: works on today’s `tb-nightly` (and in December, too).

wchargin:
Filed <https://bugs.chromium.org/p/chromium/issues/detail?id=1185435>.

psybuzz:
Just tested with tb-nightly==2.5.0a20210309, Chrome 91, MacOS, and things are no longer pixelated.  Closing for now, but for anyone who may continue to see blurriness, especially on other platforms, please add a comment.

For TensorBoard.dev users, we expect this fix to land in roughly 1 week.

PR #4748, opened 2021-03-08
graph: fix pixelated SVG for Chromium

psybuzz:
Starting in Chromium 89, zooming in on nodes in the Graph dashboard
yields very pixelated rendering artifacts. See details in [1]. This
regression is a result of changes to Chromium's SVG renderer [2].

To workaround the symptom, this change replaces the
`will-change: transform` CSS with `transform: translateZ(1px)`. Newer
Chromium versions take the former `will-change` as a hint to use the
rasterized result at initial scale, which renders faster at the cost of
image quality.

Note that we intentionally do not do `transform: translateZ(0)`. The
`1px` is important to avoid introducing a blurry text regression in
Firefox [4], for unknown reasons. The "#root" node does not have
`transform: translateZ(1px)` because this would interfere with
panning the graph.

The `will-change` rule was added in TB PR [3], but removing it has no
noticeable effect on load performance or panning the viewport.

[1] https://github.com/tensorflow/tensorboard/issues/4744
[2] https://bugs.chromium.org/p/chromium/issues/detail?id=1185435
[3] https://github.com/tensorflow/tensorboard/pull/4495
[4] https://imgur.com/a/irk0jB7
image

EDIT: original approach just removed will-change, but is now updated
to avoid regressing Firefox.

### DIFF
--- a/tensorboard/plugins/graph/tf_graph/tf-graph-scene.html.ts
+++ b/tensorboard/plugins/graph/tf_graph/tf-graph-scene.html.ts
@@ -278,6 +278,15 @@ export const template = html`
 
     /* --- Node label --- */
 
+    .node {
+      /* Provide a hint to browsers to avoid using their static rasterization
+      at initial scale, which looks very pixelated on Chromium. Note that we
+      intentionally do *not* use 'will-change: transform' and 'translateZ(0)
+      here, which introduce blurry on Firefox.
+      See https://github.com/tensorflow/tensorboard/issues/4744 */
+      transform: translateZ(1px);
+    }
+
     .node > text.nodelabel {
       cursor: pointer;
       fill: #444;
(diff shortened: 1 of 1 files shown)


wchargin:
> This is not ideal, but alternatives
> were not found to work perfectly across both browsers (e.g.
> `backface-visibility: hidden`, `transform: translateZ(0)`).

Any chance of `-moz-transform`/`-webkit-transform`?

It seems unjust to introduce a bug into Firefox to work around a
regression in Chrome.

wchargin (approved):
Lovely; thanks very much for the investigation and the screenshots!
I’m glad that you found my comment helpful.

> `animation: max_animation_raster_scale 10s 9999999s`

Help, I started my TensorBoard a few months ago and all of a sudden
everything is huge!

PR #4750, opened 2021-03-08
data_provider: add `GetExperiment` RPC

wchargin:
Summary:
This new RPC is intended to back both the `data_location` and the
`experiment_metadata` data provider methods (which really should be
combined into one). It takes an experiment ID and returns all available
metadata. If needed, we could add a fieldmask, but we leave that out
until we have a need for it.

The Rust implementation of the data server will only populate the data
location field, but we still include the rest to mirror the Python
interface.

This is the first use of `google.protobuf.Timestamp` or any of the
well-known types, so some build tweaks are needed.

Test Plan:
The protos and Rust code all build, and the newly updated instructions
in `DEVELOPMENT.md` for still work. With `grpc_cli`, the new RPC can be
seen to successfully return a non-OK `"not yet implemented"` status.

wchargin-branch: data-getexperiment-rpc


tensorboard/data/proto/data_provider.proto
```diff
@@ -31,6 +35,38 @@ service TensorBoardDataProvider {
...
+  string description = 3;
+
+  // Time that the experiment was created.
+  //
+  // This does not necessarily have any relation to times of events within an
+  // experiment. A new experiment may contain old data, or data may be freshly
+  // written to an experiment after it is created.
+  //
+  // May be unset if no creation time is known.
+  google.protobuf.Timestamp creation_time = 4;
```
wchargin, line 67: @nfelt: Could you take a quick look at this use of `Timestamp` in light
of the build changes in this PR and its diffbase, #4749? I recall that
in #4314 you lobbied for using `Timestamp` rather than `double`, and in
this case I agree that it’s the right tradeoff as far as the proto
definition goes. It incurs a bit of build complexity; IMHO it’s worth
it, but I would also not mind just using `double` if we prefer that.

tensorboard/data/server/BUILD
```diff
@@ -144,9 +145,32 @@ genrule(
...
+        # directory under the genfiles tree for its workspace. Bazel variable
+        # expansion doesn't provide a way to get that directly (it's not a
+        # target), but we can extract it from the files themselves.
+        f=
+        for f in $(locations @com_google_protobuf//:well_known_protos); do
+            break  # just take one arbitrarily
+        done
+        [ -n "$${f}" ] || die 'no well-known protos in fileset'
+        dir="$$(dirname "$${f}")"
+        while true; do
```
nfelt, line 164: This while loop feels like a bit of a complicated way to effectively just remove the suffix `/src/.*?` and replace it with `/src`.  Maybe consider just doing string munging for that?

tensorboard/data/proto/data_provider.proto
nfelt: I'm ok with the build complexity since it seems fairly contained, although boy do I wish `com_google_protobuf` were a nicer structured Bazel dependency to have.

tensorboard/data/proto/data_provider.proto
```diff
@@ -2,12 +2,16 @@ syntax = "proto3";
...
 
+import "google/protobuf/timestamp.proto";
+
 import "tensorboard/compat/proto/summary.proto";
 import "tensorboard/compat/proto/tensor.proto";
 
 option go_package = "github.com/tensorflow/tensorboard/tensorboard/data/proto/data_provider_proto";
 
 service TensorBoardDataProvider {
+  // Get metadata about an experiment.
```
nfelt, line 13: I don't object to evolving this interface a little away from Python `DataProvider` but FWIW we should anticipate that we might want to change this further depending on the outcome of experiment-ifying Python `DataProvider` as part of upcoming work.

nfelt (approved):

PR #4751, opened 2021-03-09
rust: implement data location

wchargin:
Summary:
This patch implements `GetExperiment` for the Rust data server and the
Python data provider. The server only populates the data location field,
because we don’t actually have any other experiment metadata.

Test Plan:
Tested with `--load_fast` using a local logdir and a GCS logdir.

wchargin-branch: rust-data-location

### DIFF
--- a/tensorboard/data/grpc_provider.py
+++ b/tensorboard/data/grpc_provider.py
@@ -47,7 +47,25 @@ class GrpcDataProvider(provider.DataProvider):
         self._stub = stub
 
     def data_location(self, ctx, *, experiment_id):
-        return "grpc://%s" % (self._addr,)
+        req = data_provider_pb2.GetExperimentRequest()
+        req.experiment_id = experiment_id
+        with _translate_grpc_error():
+            res = self._stub.GetExperiment(req)
+        return res.data_location
+
+    def experiment_metadata(self, ctx, *, experiment_id):
+        req = data_provider_pb2.GetExperimentRequest()
+        req.experiment_id = experiment_id
+        with _translate_grpc_error():
+            res = self._stub.GetExperiment(req)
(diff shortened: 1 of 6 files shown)


tensorboard/data/grpc_provider.py
```diff
@@ -47,7 +47,25 @@ def __init__(self, addr, stub):
         self._stub = stub
 
     def data_location(self, ctx, *, experiment_id):
-        return "grpc://%s" % (self._addr,)
```
nfelt, line 50: No action needed, just musing that it was kind of useful actually to see this when testing Rustboard (easy to remember which grpc port it was on).  I wonder if maybe it'd be worth formatting the data location as something like `<logdir> served by grpc://<address>`?

nfelt (approved):

tensorboard/data/grpc_provider.py
wchargin: Yeah, I also miss it. I feel weakly inclined to keep it to just the
logdir to avoid confusing users. It’s nice to be able to say that
`--load_fast` is meant to be a straight upgrade.

You can get the address from `--verbosity 0` or `RUST_LOG`, and I’d be
happy to patch `/data/environment` to return `str(data_provider)` for
debug use. Could even let that `str` be rendered on hover, if we want to
be really aggressive. That doesn’t cover the use case of “a user sent us
only a screenshot and we want to divine what data provider they’re
using”, but perhaps it gets most of the value—what do you think?

nfelt: I like the idea of patching `/data/environment`.  If we want user screenshots to be more useful probably the first thing we want is a version number anyway, so it's not critical that it be visible.  But nice if the frontend has some way to indicate to a curious soul "what DataProvider am I actually talking to".

PR #4754, opened 2021-03-09
rust: support unbounded reservoirs

wchargin:
Summary:
The `--samples-per-plugin` argument now understands inputs like
`scalars=100,images=all`. This indicates that some reservoirs should
have unbounded capacity.

Reservoir constructors take `impl Into<Capacity>` so that passing a bare
`usize` is still valid, enjoying all the concision of dynamic typing
alongside the safety of sound static analysis.

Test Plan:
Due to #4752, passing `--samples_per_plugin=scalars=0` to TensorBoard
does not actually work, with or without `--load_fast`. But unit tests
have decent coverage, and manual end-to-end tests against a live data
server show that it really does return lots of data.

wchargin-branch: rust-unbounded-reservoir

### DIFF
--- a/tensorboard/data/server/cli.rs
+++ b/tensorboard/data/server/cli.rs
@@ -121,9 +121,10 @@ struct Opts {
     ///
     /// A comma separated list of `plugin_name=num_samples` pairs to explicitly specify how many
     /// samples to keep per tag for the specified plugin. For unspecified plugins, series are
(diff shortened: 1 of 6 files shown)


psybuzz (approved):
Looks good!  It's neat to see a real example of using `impl Trait` in the argument position: `pub fn new(capacity: impl Into<Capacity>)`

Issue #2801, opened 2019-10-18
Reorganize argparse flag definitions to reflect actual CLI structure


nfelt:
Tracking bug for cleaning up our argparse flag definitions so they semantically align with the structure we want our CLI to have.  Factored partially out of #2573.

- [ ] Move CorePlugin flag definitions into other places as follows:
  - program.TensorBoard
    - version_tb
    - host
    - port
    - inspect
    - event_file
    - tag
    - logdir
    - path_prefix  # apply middleware in program.main()
  - backend.ingestion (??)
    - purge_orphaned_data
    - db
    - db_import
    - db_import_use_op
    - max_reload_threads
    - reload_interval
    - reload_task
    - reload_multifile
    - reload_multifile_inactive_secs
    - samples_per_plugin
  - backend.application
    - window_title
- [ ] Move CorePlugin fix_flags into business logic
  - Can we get rid of fix_flags everywhere?
- [ ] Consider adding requirement that plugins prefix their flag names with the name of the plugin, to avoid conflicts and improve clarity

PR #4755, opened 2021-03-09
core: dump some debug info in `/data/environment`

wchargin:
Summary:
In #4751, we implemented `data_location` support for RustBoard, which
means that the location no longer includes the gRPC address. But while
this is an improvement for users, the address was convenient for
development, and there’s no other way to get it from the UI. This patch
augments `/data/environment` to dump some basic debug info: the
TensorBoard version, the data provider, and the command-line flags.

This behavior can be turned off. It’s enabled in `tensorboard(1)`, but
disabled by default on the `CorePlugin` and `CorePluginLoader` APIs.
Thus, users who build their own TensorBoards will not find themselves
unexpectedly leaking information about their setup. For normal users of
`tensorboard(1)`, the flags are fine to render; the only potentially
sensitive thing is the logdir path, which is of course already
prominent.

Sample response, passed through `jq .`:

```
{
  "version": "2.5.0a0",
  "data_location": "logs/mnist/",
  "window_title": "",
  "debug": {
    "data_provider": "GrpcDataProvider(addr='localhost:32877')",
    "flags": {
      "logdir": "logs/mnist/",
      "logdir_spec": "",
      "host": null,
      "bind_all": true,
      "port": null,
      "reuse_port": false,
      "load_fast": true,
      "grpc_creds_type": "local",
      "grpc_data_provider": "",
      "purge_orphaned_data": true,
      "db": "",
      "db_import": false,
      "inspect": false,
      "version_tb": false,
      "tag": "",
      "event_file": "",
      "path_prefix": "",
      "window_title": "",
      "max_reload_threads": 1,
      "reload_interval": 5,
      "reload_task": "auto",
      "reload_multifile": null,
      "reload_multifile_inactive_secs": 86400,
      "generic_data": "auto",
      "samples_per_plugin": {},
      "custom_predict_fn": "",
      "wit_data_dir": "",
      "__tensorboard_subcommand": "serve"
    }
  }
}
```

Test Plan:
Manually navigated to `/data/environment` and verified that the flags
render and the data provider appears for both `--load_fast` and the
legacy multiplexer. Checked that when the change to `default.py` is
reverted, no debug data shows up.

wchargin-branch: environment-debug-info


tensorboard/plugins/core/core_plugin.py
```diff
@@ -50,18 +51,24 @@ class CorePlugin(base_plugin.TBPlugin):
...
         """Instantiates CorePlugin.
 
         Args:
           context: A base_plugin.TBContext instance.
+          include_debug_info: If true, `/data/environment` will include some
+            basic information like the TensorBoard server version. Disabled by
+            default to prevent surprising information leaks in custom builds of
+            TensorBoard.
         """
+        self._flags = context.flags
```
wchargin, line 64: Apropos of #2801 and comments on #4411: I don’t mind this because it
really does want to get all of `flags` rather than just using it as a
shortcut for some specific properties, and also because this
functionality really could be part of the _actual_ core, anyway.
But open to pushback if you disagree.

nfelt (commented):
Let's also update http_api.md? 
 https://github.com/tensorflow/tensorboard/blob/master/tensorboard/http_api.md#dataenvironment
 
 Maybe worth specifying there and possibly in the code that the format of the `debug` payload is subject to change and intended for human consumption, not machine consumption, so that we can change it in the future without being told we [broke someone's workflow](https://xkcd.com/1172/).

tensorboard/plugins/core/core_plugin.py
nfelt: Yeah, I agree it's fine for now, although if you don't mind editing #2801 after this goes in so it reflects the current state that'd be helpful.

tensorboard/plugins/core/core_plugin.py
```diff
@@ -193,12 +194,38 @@ def _serve_environment(self, request):
                     "creation_time": experiment_metadata.creation_time,
                 }
             )
+        if self._include_debug_info:
+            environment["debug"] = {
+                "version": version.VERSION,
```
nfelt, line 199: Maybe worth including `version` unconditionally at the top level?  That seems safe, and it's more structured than the rest, so we could exempt it from the "human consumption only" constraint in case we do want to render it in the frontend.

nfelt (approved):

wchargin (commented):
> Let's also update http_api.md? […]

Done; thanks.

PR #4758, opened 2021-03-09
line chart: conditionally disable fit to domain

stephanwlee:
The new line chart now differentiate the state between viewBox overridden
vs. default. When it is put on the default, any changes in the data will
automatically be reflected in the new domain calculation while when
overridden, the viewBox will be "stuck".

This was one of the older feature requests from the users where they
wanted the line chart to clearly indicate whether it will "auto scroll"
to fit the data or it will be stuck.

With this change, we will improve the feature in a minor way by
disabling the "fit to domain" button when it is already at the default
to indicate that the data updates will trigger a domain update as well.

Do note that the `title` is actually not working inside a mat-menu on a
`disabled` button but we will find a way to make it work in a follow-up.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
@@ -66,11 +66,11 @@ limitations under the License.
       </button>
       <button
         mat-menu-item
-        [disabled]="!seriesDataList || !seriesDataList.length || isExtentAtDefault"
+        [disabled]="!seriesDataList || !seriesDataList.length || (newLineChart && !newLineChart.getIsViewBoxOverridden())"
         (click)="resetDomain()"
         [title]="
-            isExtentAtDefault ?
-                'Line chart is already fitted to data. When data updates, line chart '
+            (newLineChart && !newLineChart.getIsViewBoxOverridden) ?
+                'Line chart is already fitted to data. When data updates, the line chart '
                 + 'will auto fit to its domain.' :
                 ''
         "
@@ -101,7 +101,6 @@ limitations under the License.
       [customXFormatter]="xAxisType === XAxisType.RELATIVE ? relativeXFormatter : undefined"
       [ignoreYOutliers]="ignoreOutliers"
       [tooltipTemplate]="tooltip"
-      (isExtentAtDefault)="isExtentAtDefault = $event.atDefault"
     ></line-chart>
 
     <ng-template #legacyChart>
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
@@ -66,8 +66,14 @@ limitations under the License.
       </button>
       <button
         mat-menu-item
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
```diff
@@ -66,8 +66,14 @@
       </button>
       <button
         mat-menu-item
-        [disabled]="!seriesDataList || !seriesDataList.length"
+        [disabled]="!seriesDataList || !seriesDataList.length || isExtentAtDefault"
         (click)="resetDomain()"
+        [title]="
+            isExtentAtDefault ?
+                'Line chart is already fitted to data. When data updates, line chart '
```
psybuzz, line 73: nit: "When data updates, line chart" -> "When data updates, the line chart"

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
```diff
@@ -123,6 +125,9 @@ export class LineChartComponent implements AfterViewInit, OnChanges, OnDestroy {
   @Input()
   ignoreYOutliers: boolean = false;
 
+  @Output()
+  readonly isExtentAtDefault = new EventEmitter<{atDefault: boolean}>();
```
psybuzz, line 129: Is there a reason why the payload is an object, rather than just the boolean?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
```diff
@@ -95,6 +101,7 @@
       [customXFormatter]="xAxisType === XAxisType.RELATIVE ? relativeXFormatter : undefined"
       [ignoreYOutliers]="ignoreOutliers"
       [tooltipTemplate]="tooltip"
+      (isExtentAtDefault)="isExtentAtDefault = $event.atDefault"
```
psybuzz, line 104: nit: `isExtentAtDefaultChanged`?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
```diff
@@ -66,8 +66,14 @@
       </button>
       <button
         mat-menu-item
-        [disabled]="!seriesDataList || !seriesDataList.length"
+        [disabled]="!seriesDataList || !seriesDataList.length || isExtentAtDefault"
```
psybuzz, line 69: Looking at the LineChart's impl, it seems like `isExtentAtDefault === !isViewBoxOverridden` always.  If so, would it be simpler to have the scalar card component's template just call `!this.lineChart?.getIsViewboxOverridden()`?

That way, we might need to worry about making sure the event emitter is properly called whenever `isViewBoxOverridden` changes.

stephanwlee: Done. I am a bit afraid of the dirty check being imperfect when you invoke a method but it seems negligible and not too terrible right now (ngAfterContentChecked or whatever gets invoked on every single mousemove event; the method is invoked on mouse move as well)

Issue #3714, opened 2020-06-06
Graph "Download PNG" button not working with TensorBoard 2.2.2


carlocantcook:
I am using TensorBoard 2.2.2 with PyTorch 1.4.0 and Python 3.7.4 on Windows 10 Home 1909

I added a model graph, which is rendered correctly, but I cannot download the image using the Download PNG button. 

I get different results with different browsers:

Firefox 77.0.1 (64-bit): Download PNG button does nothing, no download is triggered.

Chrome 83.0.4103.61: Produces a file called png.html, which is blank when viewed in the browser an contains some HTML Code.

Opera 68.0.3618.125: same as Chrome

Microsoft Edge 44.18362.449.0: produces a file named ..png which is not actually PNG encoded (can't be opened as an image)

I have attached some samples. The graph should display a simple sequential model with 3 consecutive 2D convolutional layers.

![microsoft_edge](https://user-images.githubusercontent.com/66512012/83940159-d4a12700-a7e2-11ea-902f-6026945ebcd3.png)
[png.html.txt](https://github.com/tensorflow/tensorboard/files/4739686/png.html.txt)

wchargin:
Hi @carlocantcook! Thanks for the clear report. I can reproduce this.
Glancing at the code, I suspect that the `d3.select("#graphdownload")`
is not matching anything because of shadow DOM changes. I’ll take a look
on Monday.

wchargin:
Okay, I can confirm that if we hack in the correct `#graphdownload` link
then the downloaded file is valid—

```diff
diff --git a/tensorboard/plugins/graph/tf_graph_common/minimap.ts b/tensorboard/plugins/graph/tf_graph_common/minimap.ts
index 1e2d138ad..892917e84 100644
--- a/tensorboard/plugins/graph/tf_graph_common/minimap.ts
+++ b/tensorboard/plugins/graph/tf_graph_common/minimap.ts
@@ -164,7 +164,7 @@ module tf.scene {
         // detached from the dom.
         return;
       }
-      let $download = d3.select('#graphdownload');
+      let $download = d3.select((window as any).graphDownloadLink);
       this.download = <HTMLLinkElement>$download.node();
       $download.on('click', (d) => {
         // Revoke the old URL, if any. Then, generate a new URL.
diff --git a/tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.ts b/tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.ts
[...]

PR #4759, opened 2021-03-09
graph: fix PNG download button

psybuzz:
The "Download as PNG" button in the Graph dashboard was broken
during the Polymer 2 -> 3 migration. Before, `<tf-graph-minimap>`
relied on assumptions that `<tf-graph-controls>` would provide an
`<a>` tag for downloading in the DOM. This is broken functionally
because the element is not accessible from another shadow subtree.

This fixes the download button by properly plumbing a 'click' event
handler from the controls to trigger code in the minimap.

Manually tested that clicking the button properly opens or saves
a PNG to the filesystem using Chrome and Firefox.

Googlers, see test sync cl/362151624

Fixes https://github.com/tensorflow/tensorboard/issues/3714

For reviewers, the component tree structure looks like:
```html
- <tf-graph-dashboard>
  - <tf-graph-controls>
  - <tf-graph-board>
    - <tf-graph>
      - <tf-graph-scene>
        - <tf-graph-minimap>
```

image

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_app/tf-graph-app.ts
+++ b/tensorboard/plugins/graph/tf_graph_app/tf-graph-app.ts
@@ -16,6 +16,7 @@ import {PolymerElement, html} from '@polymer/polymer';
 import {customElement, property} from '@polymer/decorators';
 
 import '../tf_graph_board/tf-graph-board';
+import {TfGraphBoard} from '../tf_graph_board/tf-graph-board';
 import '../tf_graph_controls/tf-graph-controls';
 import '../tf_graph_loader/tf-graph-loader';
 import * as tf_graph_render from '../tf_graph_common/render';
@@ -98,6 +99,7 @@ class TfGraphApp extends LegacyElementMixin(PolymerElement) {
             on-fit-tap="_fit"
             trace-inputs="{{_traceInputs}}"
             auto-extract-nodes="{{_autoExtractNodes}}"
+            on-download-image-requested="_onDownloadImageRequested"
           ></tf-graph-controls>
           <tf-graph-loader
             id="loader"
@@ -199,6 +201,11 @@ class TfGraphApp extends LegacyElementMixin(PolymerElement) {
     }
   }
   _fit() {
-    (this.$$('#graphboard') as any).fit();
+    ((this.$$('#graphboard') as unknown) as TfGraphBoard).fit();
+  }
+  _onDownloadImageRequested(filename: string) {
+    ((this.$$('#graphboard') as unknown) as TfGraphBoard).downloadAsImage(
+      filename
+    );
   }
 }
--- a/tensorboard/plugins/graph/tf_graph_app/tf-graph-app.ts
+++ b/tensorboard/plugins/graph/tf_graph_app/tf-graph-app.ts
@@ -16,7 +16,6 @@ import {PolymerElement, html} from '@polymer/polymer';
 import {customElement, property} from '@polymer/decorators';
 
 import '../tf_graph_board/tf-graph-board';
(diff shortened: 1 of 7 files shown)


tensorboard/plugins/graph/tf_graph_app/tf-graph-app.ts
```diff
@@ -199,6 +201,11 @@ class TfGraphApp extends LegacyElementMixin(PolymerElement) {
     }
   }
   _fit() {
-    (this.$$('#graphboard') as any).fit();
+    ((this.$$('#graphboard') as unknown) as TfGraphBoard).fit();
```
wchargin, line 204: Why do we need to cast through `unknown`s here? It works fine for me
locally if casting directly to `TfGraphBoard`.

tensorboard/plugins/graph/tf_graph_board/tf-graph-board.ts
```diff
@@ -293,7 +294,15 @@ class TfGraphBoard extends LegacyElementMixin(PolymerElement) {
...
   fit() {
-    (this.$.graph as any).fit();
+    ((this.$.graph as unknown) as TfGraph).fit();
+  }
+  async downloadAsImage(filename: string) {
+    const blob = await ((this.$.graph as unknown) as TfGraph).getImageBlob();
+    const element = document.createElement('a');
+    element.href = URL.createObjectURL(blob);
+    element.download = filename;
+    element.click();
```
wchargin, line 304: I’ve tested this and it seems to work for me, but I’m a bit surprised.
I expected a synthetic `e.click()` event to not trigger a download.
Do you know what I’m misunderstanding?

tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.ts
```diff
@@ -1131,6 +1130,8 @@ export class TfGraphControls extends LegacyElementMixin(PolymerElement) {
   })
   _legendOpened: boolean = true;
 
+  private _downloadFilename = '';
```
wchargin, line 1133: Should this default to `'graph.png'` for consistency with the setter?

wchargin (approved):
Recommend test-syncing this, but the code looks good to me.

tensorboard/plugins/graph/tf_graph_app/tf-graph-app.ts
psybuzz: Ah true, it was actually unnecessary.  Even with `as TfGraphBoard`, the internal sync seems to not like this, so I've set it back to `as any` for now.

On that note, I still haven't figured out how to trigger the internal check in OSS, as simply adding compiler option `declaration: true` doesn't do the trick.

tensorboard/plugins/graph/tf_graph_board/tf-graph-board.ts
psybuzz: My understanding is that `el.click()`, while synthetic, still triggers the browser's "default behavior", as well as any 'click' listeners registered on the element.  I'm not sure that really answers the root of your question, I looked briefly at the spec but didn't see anything definitive.  Similar to `download`, calling `.click` on other types of elements may trigger native behaviors such as page navigation, opening the browser file selector dialog, submitting a form, etc.

tensorboard/plugins/graph/tf_graph_controls/tf-graph-controls.ts
psybuzz: Done.

psybuzz:
Added a test sync cl to the description: cl/362151624

Ended up needing to go back to `as any` for a few things.

PR #4760, opened 2021-03-09
sync: fix graph symbols exported improperly

psybuzz:
Some Polymer element classes in the Graph dashboard are exported
for use as types in other files. This breaks the sync inside Google, so
this change removes the `export`s. Note, there is no functional change.

A working hypothesis is that TSC is run with `--declaration` internally,
and we might want to consider adding that in OSS as well.

Googlers, see test sync cl/361900919


stephanwlee (approved):

wchargin:
> However, unrelated
> errors still prevent the sync from passing.

Fixed.

PR #4763, opened 2021-03-10
accumulator: ignore first `SessionLog.START` event

### LINKED ISSUE #4743: Graph events inconsistently evicted after session log `START` event
The following event file (posted with permission) renders incorrectly in
the graphs dashboard:
[graph_failure.tfevents.zip](https://github.com/tensorflow/tensorboard/files/6094530/graph_failure.tfevents.zip)

The `/info` route thinks that a graph exists, but the `/graph` route
404s when trying to fetch it.

The salient property of this event file is that it has a `graph_def`
event followed by a `session_log { status: START }` event. The latter
event indicates to [purge all preceding events][sl] across all tags.
This docstring is twice wrong:

  - it says “purge all previously seen events with larger steps”, but
    the code actually purges events with larger-or-equal steps, which
    matters because both the `graph_def` and the `session_log` are at
    step 0; and
  - it says “all” events, but actually only purges tensors.

The fact that this purge can happen at all means that it is possible for
a time series to have summary metadata but not actually have any data,
which is something that we assumed could not happen (since a normal
preemption event always leaves at least one point in the reservoir). And
the fact that it only purges tensors means that this only started
affecting graphs after #4470, which changed the read path from `Graph()`
to `Tensors("__run_graph__")`. These have equivalent input streams due
to `dataclass_compat`, but only `Tensors(...)` gets purged.

We should fix the data provider implementation to not list blob
sequences that do not actually have tensor data, so that the graphs
dashboard is internally consistent. But we should maybe also consider
the implications of this more broadly.

(This does not affect `--load_fast`, because RustBoard does not care
about `session_log` events.)

[sl]: https://github.com/tensorflow/tensorboard/blob/9b8b527ee6f2e784b5007d03ea4ed5667f7d3ec2/tensorboard/backend/event_processing/event_accumulator.py#L629-L632

wchargin:
Summary:
Fixes #4743. A supervisor may use `SessionLog.START` events whenever it
starts, so the first such event does not indicate a *restart* and does
not require preemption. Events like graphs sometimes(?) appear before
the first `SessionLog.START`. This patch changes the accumulator to only
preempt on non-initial `SessionLog.START` events to retain that data.

Test Plan:
The updated unit test fails before this patch and passes after it. Also,
the dataset in #4743 now shows a graph in the graphs dashboard rather
than an internal 404. For posterity, the salient structure of that event
file comprises this sequence of events, all implicitly at step 0:

    { file_version: "brain.Event:2" }
    { graph_def: "..." }
    { meta_graph_def: "..." }
    { session_log { status: START } }

wchargin-branch: prune-session-restarts-only

### DIFF
--- a/tensorboard/backend/event_processing/plugin_event_accumulator.py
+++ b/tensorboard/backend/event_processing/plugin_event_accumulator.py
@@ -159,6 +159,7 @@ class EventAccumulator(object):
         self._generator_mutex = threading.Lock()
 
         self.purge_orphaned_data = purge_orphaned_data
+        self._seen_session_start = False
 
         self.most_recent_step = -1
         self.most_recent_wall_time = -1
@@ -510,11 +511,16 @@ class EventAccumulator(object):
     def _CheckForRestartAndMaybePurge(self, event):
         """Check and discard expired events using SessionLog.START.
 
-        Check for a SessionLog.START event and purge all previously seen events
-        with larger steps, because they are out of date. Because of supervisor
-        threading, it is possible that this logic will cause the first few event
-        messages to be discarded since supervisor threading does not guarantee
-        that the START message is deterministically written first.
+        The first SessionLog.START event in a run indicates the start of a
+        supervisor session. Subsequent SessionLog.START events indicate a
+        *restart*, which may need to preempt old events. This method checks
+        for a session restart event and purges all previously seen events whose
+        step is larger than or equal to this event's step.
+
+        Because of supervisor threading, it is possible that this logic will
+        cause the first few event messages to be discarded since supervisor
+        threading does not guarantee that the START message is deterministically
+        written first.
 
         This method is preferred over _CheckForOutOfOrderStepAndMaybePurge which
         can inadvertently discard events due to supervisor threading.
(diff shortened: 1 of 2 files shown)


tensorboard/backend/event_processing/plugin_event_accumulator.py
```diff
@@ -510,11 +511,16 @@ def _MaybePurgeOrphanedData(self, event):
...
+        The first SessionLog.START event in a run indicates the start of a
+        supervisor session. Subsequent SessionLog.START events indicate a
+        *restart*, which may need to preempt old events. This method checks
+        for a session restart event and purges all previously seen events whose
+        step is larger than or equal to this event's step.
+
+        Because of supervisor threading, it is possible that this logic will
+        cause the first few event messages to be discarded since supervisor
+        threading does not guarantee that the START message is deterministically
+        written first.
```
psybuzz, line 523: I might be mistaken, but isn't this 2nd paragraph no longer relevant?    The "first few event messages" can still be discarded, but that is because of multiple START messages appearing, not because of a START message appearing after the first few messages.

psybuzz (approved):

tensorboard/backend/event_processing/plugin_event_accumulator.py
wchargin: Hmm, I’m not sure that I follow. I interpreted this as meaning, “at the
start of a supervisor invocation, the order of the `START` event and the
initial summaries may not be deterministic”. So if we have

```
file_version
graph_def [step: 0]
START [step: 0]
scalar [step: 0]
scalar [step: 1]
// preempted!
scalar [step: 2]
START [step: 2]
scalar [step: 3]
```

where the step-2 `START` should have been written before the step-2
scalar, but wasn’t because of “threading reasons”, then I think that
this paragraph is still relevant.

That said, I did some light digging and wasn’t able to find more details
about the circumstances that this refers to, so I could well be missing
something.

PR #4767, opened 2021-03-11
pip: require tensorboard-data-server 0.4.0

wchargin:
Summary:
We’ve released `tensorboard-data-server==0.4.0`. This patch updates the
dependency from TensorBoard to the data server.

Test Plan:
wheel into a new virtualenv. Note that the new data server is installed
and that `--load_fast` works as desired.

wchargin-branch: require-data-server-0.4.0


psybuzz (approved):
Also did a basic test on my side and looks good

PR #4773, opened 2021-03-15
cli: add `--load_fast=auto` option

wchargin:
Summary:
The `--load_fast` argument now takes a tri-state value: one of `false`,
`true`, or `auto`. If set to `auto`, it uses a data server iff one is
available and installed. This is equivalent to “try to use `--load_fast`
but degrade gracefully if that would fail”.

With this change, `--load_fast` *must* take a value. The old behavior of
`--load_fast` is now spelled `--load_fast true`. The default value is
still `false`, so the default behavior is unchanged.

Test Plan:
Try with the following configurations:

  - via Bazel with `--define=link_data_server=true`
  - without Bazel-bundled binary, but with `tensorboard-data-server`
  - with universal wheel of `tensorboard-data-server` installed (should
    fail, because this does not actually include a data server)
  - with nothing installed (should fail)

In each failure case, `--load_fast true` should hard-error, while
`--load_fast auto` should degrade.

wchargin-branch: load-fast-auto

### DIFF
--- a/tensorboard/data/server/DEVELOPMENT.md
+++ b/tensorboard/data/server/DEVELOPMENT.md
@@ -57,8 +57,8 @@ useful in conjunction with `ibazel` to restart the server when you make changes.
 The server doesn’t have to be running when TensorBoard starts.
 
 To tell TensorBoard to start the server as a subprocess, build with
-`--define=link_data_server=true` and pass `--load_fast` to TensorBoard along
-with a normal `--logdir`. Thus:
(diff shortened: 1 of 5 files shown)


nfelt (approved):

nfelt (commented):
Should also update this doc:

https://github.com/tensorflow/tensorboard/blob/master/tensorboard/data/server/DEVELOPMENT.md#running-under-tensorboard

wchargin:
Done; thanks.

PR #4775, opened 2021-03-15
mesh: add generic data support

wchargin:
Summary:
This patch replaces the multiplexer code in the mesh plugin with data
provider code. It is not gated behind a flag.

We treat all mesh data as tensors, since that’s most expedient at this
time and the local data providers (multiplexer and RustBoard) don’t
actually have any size constraints. See also discussion in #4734.

Test Plan:
The dashboard still works with the standard demo data, both with and
without `--load_fast=true`.

wchargin-branch: mesh-generic


psybuzz (approved):

wchargin:
(I’ve rotated this below #4774, since that PR is still awaiting review
and these two are logically independent.)

PR #4776, opened 2021-03-15
histogram: remove unnecessary downsampling

wchargin:
Summary:
As the comment says, the histograms plugin used to manually downsample
because the multiplexer data provider didn’t. As of #3272, it does, so
this code is vestigial.

Test Plan:
Note that changing the default histogram sampling constants (in the same
file) to very small numbers causes the frontend to show less data. Unit
tests for the histograms plugin also cover this: removing downsampling
from the multiplexer data provider causes the histogram tests to fail.

wchargin-branch: histogram-rm-downsample

### DIFF
--- a/tensorboard/plugins/histogram/histograms_plugin.py
+++ b/tensorboard/plugins/histogram/histograms_plugin.py
@@ -19,8 +19,6 @@ this plugin.
 """
 
 
-import random
-
 from werkzeug import wrappers
 
 from tensorboard import errors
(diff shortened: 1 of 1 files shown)


wchargin:
For context, I found this with `git grep -l multiplexer ':*plugin.py'`
while looking for plugins that don’t yet have data provider support.
The word “multiplexer” appeared in a comment removed in this diff.

stephanwlee (approved):

Issue #4777, opened 2021-03-15
Merge `data_location` into `experiment_metadata`


wchargin:
The data location field is a piece of metadata scoped to an experiment,
so one would expect it to be returned by `experiment_metadata`. But it
is not so: `data_location` is a separate call. This forces two separate
database queries, which is unnecessary since `core_plugin` calls them
consecutively at the [unique call site][cs]. Thus, it’s more than just a
cleanliness issue.

Suggested migration path:

1.  Make `ExperimentMetadata.__init__` take kwargs only.
2.  Add a new `data_location` field with default value `""`.
3.  Update data providers to populate the new field, and update
    `core_plugin` to prefer that field if populated.
4.  Remove the `data_location` method and make the field non-optional.
5.  Remove dead `data_location` implementations from data providers.

[cs]: https://github.com/tensorflow/tensorboard/blob/b16ed7cd56ad4e8bee3a0c6cc9ca76042d6e3449/tensorboard/plugins/core/core_plugin.py#L178-L183

wchargin:
> 4.  Remove the `data_location` method and make the field non-optional.

Alternatively, we could keep it optional and make them all optional.

PR #4778, opened 2021-03-15
data: make metadata required, providing defaults

wchargin:
Summary:
The `experiment_metadata` call is now required to return a non-`None`
value, but the fields now all have default values. This way, we can more
easily add new fields in a backward-compatible way. See #4777.

A few Google-internal data providers still have cases where they can
return `None`, so `core_plugin` (the unique call site) still tolerates
that deviation.

Test Plan:
Note that TensorBoard still looks the same: the `/data/environment`
response has some extra fields, but we don’t render a header bar or
anything.

wchargin-branch: metadata-required


stephanwlee (commented):
I assume you've tested the sync. LGTM.

stephanwlee (approved):

wchargin:
Yep: <http://cl/363021369>, <http://cl/363021896>.

PR #4779, opened 2021-03-15
npmi: fix the undefined in the selector

stephanwlee:
`getEmbeddingDataSet` is typed as something that can return an undefined
but component is typed as non-nullable which is incorrect. Angular does
not correctly type check across the component tree and resulted in
incorrect typing.

This change fixes the case when we try to access property in an
undefined in a template.


japie1235813 (approved):

wchargin:
@googlebot I consent.

stephanwlee:
Thanks a lot for bumping the CI 👊

PR #4786, opened 2021-03-17
cli: detect filesystems for `--load_fast=auto`

wchargin:
Summary:
If you pass `--logdir s3://... --load_fast=auto`, we no longer spawn a
data server. Same if you use `--logdir_spec` instead of `--logdir`.

GCS filesystems are considered supported. We don’t check the format of
Google application credentials, so this isn’t perfect.

It’s a little weird to have this logic in `program.py` rather than the
data server itself. Alternatives considered:

 1. Add a `logdir_supported` function to the data server Pip package.
    This is nice because we could push a new data server release with
    new filesystems without having to push a new TensorBoard release.
    But it doesn’t suffice for the cases where the data server binary
    doesn’t come from the Pip package (i.e., when it comes from a Bazel
    `--define` or an environment variable).

 2. Add a `--check-logdir` flag to the data server to directly ask it to
    check whether it can handle a given filesystem. This is nice because
    it has the benefit of (1) without the aforementioned drawback. It
    also makes it more natural to do subtler checks, like “GCS logdirs
    are okay if the user’s credentials are of a supported variety”. But
    it requires launching the data server twice and throwing away the
    first result ([parse, don’t validate][pdv]). Also, it’d require
    releasing a new data server binary, which is inconvenient in the
    short term.

 3. Teach the data server to exit with a special exit code if it can
    tell that the log directory is unsupported. Then, try starting a
    data server ingester, and fall back to a local ingester if that
    fails. This has the benefits of (1) and (2), and avoids the problem
    of validating the input twice. It would still require a release, and
    is also a bit more complicated mechanically.

I think that I like (3) for the long term and will consider implementing
it. In the meantime, this is simple, prevents a de facto regression, and
doesn’t incur much architectural debt, so let’s do it.

[pdv]: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

Test Plan:
Both `~/tensorboard_data` and `gs://tensorboard-bench-logs/mnist`
activate the data server as desired, whereas `wat://something` falls
back and gives a TensorFlow error (also as desired).

wchargin-branch: cli-loadfast-detect-fs


wchargin:
@nfelt: Interested in your thoughts on the considered alternatives?

nfelt (approved):

nfelt:
> @nfelt: Interested in your thoughts on the considered alternatives?

I agree with your assessment that this is fine for now and (3) is probably nicer in the long term.  Nothing in particular to add, but thanks for outlining the options.

wchargin:
Cool; thanks for your input!

PR #4788, opened 2021-03-18
rust: improve GCS credential error handling

wchargin:
Summary:
Failure cases when obtaining GCS credentials are now handled in a more
structured manner. Instead of just logging ad hoc warnings, we return
error details to the caller. In particular, this lets the caller
distinguish the case where the user has credentials of a format that we
don’t support, such as service account private keys.

The log messages are nicer, too: in case of a service account key, we
recognize and explain the problem instead of just emitting a JSON error.

Test Plan:
To test each error case, set `GOOGLE_APPLICATION_CREDENTIALS` to:

  - `/enoent` or `/etc/shadow` to test `Unreadable(...)`;
  - `/dev/null` to test `Unparseable(...)`;
  - a service account credentials file to test `Unsupported(...)`.
    (Or, just a valid JSON file without the needed fields.)

Then note that both anonymous access and properly authenticated access
still work.

wchargin-branch: rust-gcs-error-handling

### DIFF
--- a/tensorboard/data/server/gcs/auth.rs
+++ b/tensorboard/data/server/gcs/auth.rs
@@ -149,6 +149,7 @@ impl BoundedToken {
 
 /// The user's persistent credentials, if any. This represents all the information needed to
 /// request access tokens.
+#[derive(Debug)]
 pub enum Credentials {
     Anonymous,
     RefreshToken(RefreshToken),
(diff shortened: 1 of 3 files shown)


tensorboard/data/server/gcs/auth.rs
```diff
@@ -235,6 +242,42 @@ impl Credentials {
     }
 }
 
+/// Partial structure of the `GOOGLE_APPLICATION_CREDENTIALS` file.
+#[derive(Deserialize)]
+struct ApplicationCreds {
+    r#type: Option<String>,
```
wchargin, line 248: The JSON field is called `type`, but `type` is a keyword in Rust, so we
use `r#type` as a *raw identifier*:
<https://doc.rust-lang.org/edition-guide/rust-2018/module-system/raw-identifiers.html>

An alternative would be to call the field something else and ask serde
to rename while deserializing:

```rust
#[derive(Deserialize)]
struct ApplicationCreds {
    #[serde(rename = "type")]
    creds_type: Option<String>,
    ...,
}
```

I don’t have a strong preference.

tensorboard/data/server/gcs/auth.rs
```diff
@@ -235,6 +242,42 @@ impl Credentials {
...
+            (Some(client_id), Some(client_secret), Some(refresh_token)) => {
+                Ok(Credentials::RefreshToken(RefreshToken(RefreshTokenCreds {
+                    client_id,
+                    client_secret,
+                    refresh_token,
+                })))
+            }
+            _ => Err(UnsupportedCredentialsError {
+                creds_type: self.r#type.unwrap_or_default(),
+            }),
```
stephanwlee, line 276: I am already forgetting a lot of Rust so help me understand here: so we get UnsupportedCredentialsError when the JSON does not contain all of `client_id`,`client_secret`, and `refresh_token`, right? In which scenario would we have that?  

If I am reading the description correctly, service account would fall into this case? In such case, I am not 100% sure if just printing `type` (which can be missing, too) is actionable/comprehendible. 

Did I understand the flow correctly?

stephanwlee (commented):
Mostly looks good. I just have a UX question.

tensorboard/data/server/gcs/auth.rs
wchargin: You understand correctly. A refresh token creds file looks like:

```json
{
  "type": "authorized_user",
  "client_id": "...",
  "client_secret": "...",
  "refresh_token": "..."
}
```

and a service account creds file looks like:

```json5
{
  "type": "service_account",
  "project_id": "<GCP project ID>",
  "private_key": "...",
  /* also some more string fields here (but not `refresh_token`, etc.) */
}
```

I suspect that `type` should always be there; I just made it optional
because I couldn’t find actual docs/specs for this file format, so
wanted to be a bit conservative.

Note that we also print the file name. So the actual error message, on a
real service account file, is:

> unsupported GCS credentials of type "service_account"; only OAuth refresh tokens supported (credentials file /tmp/my-service-account-key.json)

I think that this is pretty readable. WDYT?

stephanwlee: I see, so there is some implicit trust that the parsable JSON is correctly formed like `type: service_account` always misses `refresh_token` and etc... 

Not sure if I want to trust random file on the file system but probability of it being corrupted is low, I guess.

LGTM.

stephanwlee (approved):

tensorboard/data/server/gcs/auth.rs
wchargin: Yeah—I think the failure modes here are all pretty tolerable. Even if
you have a somehow-corrupted credentials file, we’ll emit a clear error
and exit cleanly (or, with #4794, fall back to `--load_fast=false`).
And if something goes wrong, we just fail to authenticate; no
possibility of losing, misrepresenting, or corrupting the actual data.

In case it helps, the [TensorFlow logic][1] is similarly ad hoc. It just
assumes that if there’s a `refresh_token` field, the client ID and
secret must also be there, and likewise that if there’s a `private_key`
field then `private_key_id`, et al. must be present.

[1]: https://github.com/tensorflow/tensorflow/blob/da86421df896248401341c9f1a6b23cefb267a3a/tensorflow/core/platform/cloud/google_auth_provider.cc#L219-L234

PR #4769, opened 2021-03-12
what's new: redux foundation

japie1235813:
* Motivation for features / changes
More note-reading part to _redux/  and build up the selector-reducer-action foundation
 
* Technical description of changes
  * Modify a bunch of visibility to outside _redux/ for its usage in tensorboard app
  * Leave tests for reducer cause actions is not brought into the picture yet.

PR #4791, opened 2021-03-18
sync: fix `@ngrx/store/testing` import

wchargin:
Summary:
Recent changes in #4769 added an import of `@ngrx/store/testing`, which
works fine here but requires an extra build dep internally.

Test Plan:
Test sync now builds: <http://cl/363586736>.

wchargin-branch: sync-20210317-ngrx-store-testing


japie1235813 (approved):
Thanks for the fix!

PR #4792, opened 2021-03-18
tensorboard-data-server 0.5.0

psybuzz:
Changes since 0.4.0:

- Added support for Custom Scalars, Mesh plugin data
- Improved error messages for unsupported GCS credentials
- Added `--error-file` option to extract startup errors

Merge plan: rebase and merge, then release from the unique commit where
the version numbers are 0.5.0.

### DIFF
--- a/tensorboard/data/server/pip_package/tensorboard_data_server/__init__.py
+++ b/tensorboard/data/server/pip_package/tensorboard_data_server/__init__.py
@@ -19,7 +19,7 @@ import os
 
 # Version of this Python package. This may differ from the versions of
 # both TensorBoard and the data server.
-__version__ = "0.5.0a0"
+__version__ = "0.5.0"
 
 
 def server_binary():
--- a/tensorboard/data/server/pip_package/tensorboard_data_server/__init__.py
+++ b/tensorboard/data/server/pip_package/tensorboard_data_server/__init__.py
@@ -19,7 +19,7 @@ import os
 
 # Version of this Python package. This may differ from the versions of
 # both TensorBoard and the data server.
-__version__ = "0.5.0"
+__version__ = "0.6.0a0"
 
 
 def server_binary():
--- a/tensorboard/data/server/Cargo.lock
+++ b/tensorboard/data/server/Cargo.lock
@@ -1080,7 +1080,7 @@ dependencies = [
 
 [[package]]
 name = "rustboard"
-version = "0.5.0-alpha.0"
+version = "0.5.0"
 dependencies = [
  "async-stream",
  "async-trait",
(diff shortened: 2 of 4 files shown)


wchargin:
Thanks! If we could wait for #4793, that’d be great.

wchargin:
(ready :-) )

psybuzz:
Did a confidence check of the wheel on Mac (using the mac wheel) and Linux (using the manylinux wheel) in an env with tf-nightly, and manually checked the GCS bucket (gs://tensorboard-bench-logs/edge_cgan), --load_fast=true,auto,false, and saw that Custom Scalars and Mesh loaded without any plugin warnings in the terminal.

tensorboard/data/server/pip_package/tensorboard_data_server/__init__.py
```diff
@@ -19,7 +19,7 @@
 
 # Version of this Python package. This may differ from the versions of
 # both TensorBoard and the data server.
-__version__ = "0.5.0"
+__version__ = "0.6.0-alpha.0"
```
wchargin, line 22: Can we change this to `0.6.0a0` to keep the consistent Python format?

I’m actually seeing that Python treats these as equivalent—

```python
>>> from pkg_resources import parse_version as pv
>>> pv("0.6.0a0") == pv("0.6.0-alpha.0")
True
```

—which is news to me. But this isn’t mentioned in PEP 440:

> The canonical public version identifiers MUST comply with the following scheme:
>
> ```
> [N!]N(.N)*[{a|b|rc}N][.postN][.devN]
> ```

So I think that I’d prefer to keep it strictly compliant, at least until
we can find some reference saying that the `-alpha.0` form is okay.

wchargin (approved):
> Did a confidence check of the wheel on Mac (using the mac wheel) and Linux (using the manylinux wheel) in an env with tf-nightly, and manually checked the GCS bucket (gs://tensorboard-bench-logs/edge_cgan), --load_fast=true,auto,false, and saw that Custom Scalars and Mesh loaded without any plugin warnings in the terminal.

Lovely; thanks! I’ve also kicked the tires a bit on Linux, and it seems
to work for me. Approved, but please fix inline before merging. (And
friendly reminder to rebase-and-merge.)

tensorboard/data/server/pip_package/tensorboard_data_server/__init__.py
psybuzz: Done.

Ah, nice catch, my original change was unintentional.  I also prefer the strict compliance with PEP 440.  I've force pushed and will do the rebase-and-merge once CI passes.

Thanks!

PR #4793, opened 2021-03-18
rust: improve startup error handling

wchargin:
Summary:
RustBoard currently writes normal logs to stderr, and also writes any
fatal startup errors to stderr. This makes it impossible to detect and
capture startup errors without also suppressing the log output. As of
this patch, you can pass `--error-file FILE` to ask RustBoard to write
fatal errors to `FILE` instead, so you can leave stderr unencumbered.

In this patch, we also improve error handling for unknown protocols
(like `--logdir wat://mnist`), which complements `--error-file` by
making the startup errors themselves more helpful.

This will be used in `tensorboard(1)` to gracefully fall back to the
legacy multiplexer in case of unsupported invocations: i.e., option (3)
as discussed in #4786.

Test Plan:
Run the data server with `--logdir wat://`, or point to a GCS logdir
while `GOOGLE_APPLICATION_CREDENTIALS` is set to something unreadable.
Note that without `--error-file`, these both continue to write an error
to stderr, and with `--error-file`, they both write to the given file.

wchargin-branch: rust-improve-startup-errors

### DIFF
--- a/tensorboard/data/server/cli.rs
+++ b/tensorboard/data/server/cli.rs
@@ -47,7 +47,9 @@ struct Opts {
     ///
     /// Directory to recursively scan for event files (files matching the `*tfevents*` glob). This
     /// directory, its descendants, and its event files will be periodically polled for new data.
-    #[clap(long)]
+    ///
+    /// If this log directory is invalid or unsupported, exits with status 8.
+    #[clap(long, setting(clap::ArgSettings::AllowEmptyValues))]
     logdir: PathBuf,
 
     /// Bind to this IP address
@@ -95,6 +97,14 @@ struct Opts {
     /// when the server starts.
     #[clap(long)]
     port_file: Option<PathBuf>,
+
+    /// Write startup errors to this file
+    ///
+    /// If the logdir is invalid or unsupported, write the error message to this file instead of to
+    /// stderr. That way, you can capture this output while still keeping stderr open for normal
+    /// logging.
+    #[clap(long)]
+    error_file: Option<PathBuf>,
 
     /// Checksum all records (negate with `--no-checksum`)
     ///
@@ -147,6 +157,10 @@ impl FromStr for ReloadStrategy {
     }
 }
 
+/// Exit code for failure of [`DynLogdir::new`].
+// Keep in sync with docs on `Opts::logdir`.
+const EXIT_BAD_LOGDIR: i32 = 8;
+
 #[tokio::main]
 pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let opts = Opts::parse();
@@ -156,6 +170,16 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
         _ => LevelFilter::max(),
     });
     debug!("Parsed options: {:?}", opts);
+    let data_location = opts.logdir.display().to_string();
+
+    // Create the outside an async runtime (see docs for `DynLogdir::new`).
+    let (raw_logdir, error_file) = (opts.logdir, opts.error_file);
+    let logdir = tokio::task::spawn_blocking(|| DynLogdir::new(raw_logdir))
+        .await?
+        .unwrap_or_else(|e| {
+            write_startup_error(error_file.as_ref().map(|p| p.as_ref()), e);
+            std::process::exit(EXIT_BAD_LOGDIR);
+        });
 
     if opts.die_after_stdin {
         thread::Builder::new()
@@ -184,8 +208,6 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
         eprintln!("listening on {:?}", bound);
     }
 
-    let data_location = opts.logdir.display().to_string();
-
     // Leak the commit object, since the Tonic server must have only 'static references. This only
     // leaks the outer commit structure (of constant size), not the pointers to the actual data.
     let commit: &'static Commit = Box::leak(Box::new(Commit::new()));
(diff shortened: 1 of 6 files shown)


wchargin:
R=@nfelt since we’d talked about this kind of thing a bit previously;
feel free to revert to me if busy.

tensorboard/data/server/cli.rs
```diff
@@ -147,6 +157,10 @@ impl FromStr for ReloadStrategy {
     }
 }
 
+/// Exit code for failure of [`DynLogdir::new`].
+// Keep in sync with docs on `Opts::logdir`.
+const EXIT_BAD_LOGDIR: i32 = 8;
```
nfelt, line 162: No action needed but any particular reason for choosing 8 versus just, say, 2?

tensorboard/data/server/cli.rs
```diff
@@ -156,6 +170,16 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
         _ => LevelFilter::max(),
     });
     debug!("Parsed options: {:?}", opts);
+    let data_location = opts.logdir.display().to_string();
+
+    // Create the outside an async runtime (see docs for `DynLogdir::new`).
```
nfelt, line 175: typo: "Create the outside an"?

tensorboard/data/server/cli.rs
```diff
@@ -254,6 +272,23 @@ fn write_port_file(path: &Path, port: u16) -> std::io::Result<()> {
     Ok(())
 }
 
+/// Writes error to the given file, or to stderr as a fallback.
+fn write_startup_error(path: Option<&Path>, error: dynamic_logdir::Error) {
```
nfelt, line 276: Optional and feel free to tell me this is less idiomatic but I find the `match` here a little more complicated to think about than the relatively more prosaic nested if with fall-through:

```
if let Some(p) = path {
  let write_to_file = ...
  if write_to_file(p).is_ok() {
    return;
  }
}
eprintln!("fatal: {}", error);
```

tensorboard/data/server/cli.rs
```diff
@@ -254,6 +272,23 @@ fn write_port_file(path: &Path, port: u16) -> std::io::Result<()> {
...
+    let write_to_file = |path: &Path| -> std::io::Result<()> {
+        let mut f = File::create(path)?;
+        writeln!(f, "{}", error)?;
+        f.sync_all()?;
+        Ok(())
+    };
+    let should_write_to_stderr = match path {
+        None => true,
+        Some(p) => write_to_file(p).is_err(),
```
nfelt, line 285: Is it worth surfacing the error writing to the error file in some way?

tensorboard/data/server/cli/dynamic_logdir.rs
```diff
@@ -52,29 +67,43 @@ impl DynLogdir {
...
-            }
-            Ok(c) => c,
-        };
-        Some(DynLogdir::Gcs(gcs::Logdir::new(client, bucket, prefix)))
+        let client = gcs::Client::new(gcs::Credentials::from_disk()?)?;
+        Ok(DynLogdir::Gcs(gcs::Logdir::new(client, bucket, prefix)))
+    }
+}
+
+fn is_protocol(s: &str) -> bool {
```
nfelt, line 101: Link to https://tools.ietf.org/html/rfc3986#section-3.1 if that's what we're implementing here?

FWIW, the logic in TB proper (for no-TF mode; haven't checked real TF) doesn't validate the scheme/protocol; it only splits on "://".  Not a big deal but maybe worth trying to harmonize these two since it's a little odd to have the behavior differ.

tensorboard/data/server/cli/dynamic_logdir.rs
```diff
@@ -52,29 +67,43 @@ impl DynLogdir {
...
-        let gcs_path = match path_str.strip_prefix("gs://") {
-            // Assume that anything not starting with `gs://` is a path on disk.
-            None => return Some(DynLogdir::Disk(DiskLogdir::new(path))),
-            Some(p) => p,
-        };
+        if path_str.is_empty() {
+            return Err(Error::EmptyLogdir);
+        }
+
+        let protocol_parts: Vec<&str> = path_str.splitn(2, "://").collect();
```
nfelt, line 76: Heh. I feel like every time I look an an API and think it could be improved, I go looking and find out it's there but only [in Rust nightly](https://doc.rust-lang.org/std/string/struct.String.html#method.split_once).

nfelt (approved):

tensorboard/data/server/cli.rs
nfelt: Okay, fair enough.

tensorboard/data/server/cli/dynamic_logdir.rs
nfelt: Yeah, 3986 is meant to update 1738 so I think it's the better reference.  SGTM to do it the correct way especially since TF also does that, and we can fix the no-TF one at some point.

tensorboard/data/server/cli.rs
nfelt: LGTM.  I just figured it was a bit weird to fail to fail to do what was instructed and not have any way for the user to assess why.  Also I guess my thought was that `--error-file` was more about being able to easily extract the critical error w/out including potentially unrelated stuff from stderr rather than about keeping stderr free from a critical error that is semi-expected, but I can see that it sort of serves both purposes.

tensorboard/data/server/cli/dynamic_logdir.rs
wchargin: Ah, got it. I see the “Updates” tag now. I’ve always had trouble
navigating the maze of URL RFCs, so this will help; thanks.

PR #4795, opened 2021-03-18
docs: update `tf.summary.image()`

psybuzz:
Update documentation for `tf.summary.image`.

#API_FIXIT_3_18_21

From what I understand, OSS is the source of truth, but the
TF Devsite is not updated immediately without manually doing
an internal process to update the devsite.

### DIFF
--- a/tensorboard/plugins/image/summary_v2.py
+++ b/tensorboard/plugins/image/summary_v2.py
@@ -37,28 +37,32 @@ def image(name, data, step=None, max_outputs=3, description=None):
 
     This example writes 2 random grayscale images:
 
-    >>> w = tf.summary.create_file_writer('test/logs')
-    >>> with w.as_default():
-    ...   image1 = tf.random.uniform(shape=[8, 8, 1])
-    ...   image2 = tf.random.uniform(shape=[8, 8, 1])
-    ...   tf.summary.image("grayscale_noise", [image1, image2], step=0)
+    ```python
+    w = tf.summary.create_file_writer('test/logs')
+    with w.as_default():
+      image1 = tf.random.uniform(shape=[8, 8, 1])
+      image2 = tf.random.uniform(shape=[8, 8, 1])
+      tf.summary.image("grayscale_noise", [image1, image2], step=0)
+    ```
 
     To avoid clipping, data should be converted to one of the following:
 
     - floating point values in the range [0,1], or
     - uint8 values in the range [0,255]
 
-    >>> # Convert the original dtype=int32 `Tensor` into `dtype=float64`.
-    >>> rgb_image_float = tf.constant([
-    ...   [[1000, 0, 0], [0, 500, 1000]],
-    ... ]) / 1000
-    >>> tf.summary.image("picture", [rgb_image_float], step=0)
+    ```python
+    # Convert the original dtype=int32 `Tensor` into `dtype=float64`.
+    rgb_image_float = tf.constant([
+      [[1000, 0, 0], [0, 500, 1000]],
+    ]) / 1000
+    tf.summary.image("picture", [rgb_image_float], step=0)
 
-    >>> # Convert original dtype=uint8 `Tensor` into proper range.
-    >>> rgb_image_uint8 = tf.constant([
-    ...   [[1, 1, 0], [0, 0, 1]],
-    ... ], dtype=tf.uint8) * 255
-    >>> tf.summary.image("picture", [rgb_image_uint8], step=1)
+    # Convert original dtype=uint8 `Tensor` into proper range.
+    rgb_image_uint8 = tf.constant([
+      [[1, 1, 0], [0, 0, 1]],
+    ], dtype=tf.uint8) * 255
+    tf.summary.image("picture", [rgb_image_uint8], step=1)
+    ```
 
     Arguments:
       name: A name for this summary. The summary tag used for TensorBoard will
--- a/tensorboard/plugins/image/summary_v2.py
+++ b/tensorboard/plugins/image/summary_v2.py
@@ -29,37 +29,39 @@ def image(name, data, step=None, max_outputs=3, description=None):
 
     See also `tf.summary.scalar`, `tf.summary.SummaryWriter`.
 
-    Logs a set of images to the current log directory for later analysis
-    in TensorBoard's 'Images' dashboard. Like `tf.summary.scalar` points,
-    each set of images is associated with a `step` and a `name`.  All the
-    image sets with the same `name` constitute a time series of image sets.
+    Writes a collection of images to the current default summary writer. Data
+    appears in TensorBoard's 'Images' dashboard. Like `tf.summary.scalar` points,
+    each collection of images is associated with a `step` and a `name`.  All the
+    image collections with the same `name` constitute a time series of image
+    collections.
 
     This example writes 2 random grayscale images:
 
     ```python
-    test_summary_writer = tf.summary.create_file_writer('test/logs')
-    with test_summary_writer.as_default():
-      image1 = tf.random.normal(shape=[8, 8])
-      image2 = tf.random.normal(shape=[8, 8])
-      tf.summary.image("grayscale noise", [image1, image2], step=0)
+    w = tf.summary.create_file_writer('test/logs')
+    with w.as_default():
+      image1 = tf.random.uniform(shape=[8, 8, 1])
+      image2 = tf.random.uniform(shape=[8, 8, 1])
+      tf.summary.image("grayscale_noise", [image1, image2], step=0)
     ```
 
     To avoid clipping, data should be converted to one of the following:
-    - floating point values in the range [0,1]
-    or
+
+    - floating point values in the range [0,1], or
     - uint8 values in the range [0,255]
 
     ```python
-    # Using division to convert the origional dtype=Int32 value into
-    # dtype=float64.
-    rgb_image_scaled = tf.constant([
-      [[255, 127, 0], [0, 127, 255]],
-    ]) / 255
+    # Convert the original dtype=int32 `Tensor` into `dtype=float64`.
+    rgb_image_float = tf.constant([
+      [[100, 0, 0], [0, 50, 100]],
+    ]) / 100
+    tf.summary.image("picture", [rgb_image_float], step=0)
 
+    # Convert original dtype=uint8 `Tensor` into proper range.
     rgb_image_uint8 = tf.constant([
-      [[255, 255, 0], [0, 127, 255]],
-    ], dtype=tf.uint8)
-    tf.summary.image("rgb picture", [rgb_image_scaled, rgb_image_uint8], step=0)
+      [[1, 1, 0], [0, 0, 1]],
+    ], dtype=tf.uint8) * 255
+    tf.summary.image("picture", [rgb_image_uint8], step=1)
     ```
 
     Arguments:
@@ -71,7 +73,8 @@ def image(name, data, step=None, max_outputs=3, description=None):
         should be 1, 2, 3, or 4 (grayscale, grayscale with alpha, RGB, RGBA).
         Any of the dimensions may be statically unknown (i.e., `None`).
         Floating point data will be clipped to the range [0,1]. Other data types
-        may be clipped into an allowed range for safe casting to uint8.
+        will be clipped into an allowed range for safe casting to uint8, using
+        `tf.image.convert_image_dtype`.
       step: Explicit `int64`-castable monotonic step value for this summary. If
         omitted, this defaults to `tf.summary.experimental.get_step()`, which must
         not be None.
--- a/tensorboard/plugins/image/summary_v2.py
+++ b/tensorboard/plugins/image/summary_v2.py
@@ -53,8 +53,8 @@ def image(name, data, step=None, max_outputs=3, description=None):
     ```python
     # Convert the original dtype=int32 `Tensor` into `dtype=float64`.
     rgb_image_float = tf.constant([
-      [[100, 0, 0], [0, 50, 100]],
-    ]) / 100
+      [[1000, 0, 0], [0, 500, 1000]],
+    ]) / 1000
     tf.summary.image("picture", [rgb_image_float], step=0)
 
     # Convert original dtype=uint8 `Tensor` into proper range.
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,29 @@
...
+
+    ```
+    writer = tf.summary.create_file_writer('logs')
+    with writer.as_default():
+      image1 = tf.random.normal(shape=[8, 8])
+      image2 = tf.random.normal(shape=[8, 8])
+      tf.summary.image("grayscale noise", [image1, image2], step=0)
+    ```
+
+    When image data is outside of the range [0, 1), please rescale the data
```
bileschi, line 43: Consider rephrasing to describe a requirement, rather than make a request:

"Data should be scaled to be within the [0, 1) range."

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,29 @@
...
+    before writing the summary, to avoid a clipped result.
+
+    ```
+    rgb_image = tf.constant([
```
bileschi, line 47: This image violates the above constraint.  Specifically, the constraints specifies the range [0, 1) with the open value at the top of the range.  However, 255/255 is exactly 1.   Did you mean `[0, 1]` (note square bracket).

Also, be sure those values are floats and not ints, otherwise you will end up with an image which has values exactly 0 or 1.

bileschi (changes requested):

tensorboard/plugins/image/summary_v2.py
psybuzz: Done.

tensorboard/plugins/image/summary_v2.py
psybuzz: Good catch, thanks.  I did some testing and found that floats are clipped, uint8's are preserved (already in range [0,255]), and things are strange when converting from a signed type.

Internally we use https://www.tensorflow.org/api_docs/python/tf/image/convert_image_dtype, whose docs say that it expects floating point values to have range [0,1), with an exclusive end of the range.  However, testing out the summary API, float32 and float64 values of `1.0` are indeed accepted (treated like rgb 255 in range [0,255]), so I've updated the docs to say `[0,1]` inclusive.

bileschi (approved):

tensorboard/plugins/image/summary_v2.py
nfelt: FWIW, I think the fact that it accepts `1.0` is an artifact of the saturation / wrap-around.  So I don't think the docs for `convert_image_dtype` are incorrect; semantically the range probably is supposed to be `[0, 1)`, but `1.0` still works because it gets clipped, as does any float that's `>= 1.0`.

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,41 @@
...
+
+    Logs a set of images to the current log directory for later analysis
+    in TensorBoard's 'Images' dashboard. Like `tf.summary.scalar` points,
+    each set of images is associated with a `step` and a `name`.  All the
+    image sets with the same `name` constitute a time series of image sets.
+
+    This example writes 2 random grayscale images:
+
+    ```python
+    test_summary_writer = tf.summary.create_file_writer('test/logs')
```
nfelt, line 40: Optional but I might use a shorter name like `w` here to be a bit more colloquial/brief for an example.

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,41 @@
...
+    with test_summary_writer.as_default():
+      image1 = tf.random.normal(shape=[8, 8])
```
nfelt, line 42: I'd probably use [`uniform`](https://www.tensorflow.org/api_docs/python/tf/random/uniform) instead so that it emits data that's within the valid range, since `normal` will be evenly distributed around 0.0 so half the image will get clipped.

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,41 @@
...
+      image2 = tf.random.normal(shape=[8, 8])
+      tf.summary.image("grayscale noise", [image1, image2], step=0)
+    ```
+
+    To avoid clipping, data should be converted to one of the following:
+    - floating point values in the range [0,1]
```
nfelt, line 48: Did we check the markdown rendering for this list?  (Not sure if there's an easy way to do that, maybe we should ask the docs folks.)  May work better to format as

```
To avoid clipping, data should be converted to one of the following:

- floating point values in the range [0,1], or
- uint8 values in the range [0,255]
```

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,41 @@
...
+    or
+    - uint8 values in the range [0,255]
+
+    ```python
+    # Using division to convert the origional dtype=Int32 value into
```
nfelt, line 53: typo: original

typo: `Int32` should be `int32`?  (lowercase i)

nfelt, line 53: This seems like a bit of an odd example.  Our original values here are already valid (to convert directly into uint8) so it seems a bit odd to show converting them to floating point.

Maybe a better example to convert from values between 1 and 100, or values between -1.0 and 1.0?

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,41 @@
...
     """Write an image summary.
 
+    See also `tf.summary.scalar`, `tf.summary.SummaryWriter`.
+
+    Logs a set of images to the current log directory for later analysis
+    in TensorBoard's 'Images' dashboard. Like `tf.summary.scalar` points,
+    each set of images is associated with a `step` and a `name`.  All the
+    image sets with the same `name` constitute a time series of image sets.
+
+    This example writes 2 random grayscale images:
```
nfelt, line 37: When I execute this example in eager mode against recent `tf-nightly` I get an error:

```
ValueError: .  Tensor  must have rank 4.  Received rank 3, shape (2, 8, 8)
```

I think you can resolve this by generating the images with shape `[8, 8, 1]` instead.

Not to be a drag about this, but obviously examples are only going to help users if they work :)  So let's make sure we double-check them.

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,41 @@
 def image(name, data, step=None, max_outputs=3, description=None):
     """Write an image summary.
 
+    See also `tf.summary.scalar`, `tf.summary.SummaryWriter`.
```
nfelt, line 30: Putting this "see also" at the top seems a bit odd to me, but it looks like this was discussed on #4796 so I'll comment there.

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,41 @@
...
+
+    Logs a set of images to the current log directory for later analysis
```
nfelt, line 32: I'd say "Writes a collection of images to the current default summary writer..." here:

- "writes" vs "logs" for consistency
- "collection" vs "set" because set has a mathematical meaning (uniqueness) that's not relevant here. ditto for usage of "set" below
- while we only have a file writer / logdir writer right now, the API is a bit more abstract than that, so better to say writer vs log directory

tensorboard/plugins/image/summary_v2.py
```diff
@@ -35,7 +70,8 @@ def image(name, data, step=None, max_outputs=3, description=None):
         width of the images, and `c` is the number of channels, which
         should be 1, 2, 3, or 4 (grayscale, grayscale with alpha, RGB, RGBA).
         Any of the dimensions may be statically unknown (i.e., `None`).
-        Floating point data will be clipped to the range [0,1).
+        Floating point data will be clipped to the range [0,1]. Other data types
+        may be clipped into an allowed range for safe casting to uint8.
```
nfelt, line 74: s/may be/will be/ - I think better to be definitive here.  We can even always give a pointer to `convert_image_dtype` if people want to know exactly what we do.

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,41 @@
...
+    image sets with the same `name` constitute a time series of image sets.
+
+    This example writes 2 random grayscale images:
+
+    ```python
+    test_summary_writer = tf.summary.create_file_writer('test/logs')
+    with test_summary_writer.as_default():
+      image1 = tf.random.normal(shape=[8, 8])
+      image2 = tf.random.normal(shape=[8, 8])
+      tf.summary.image("grayscale noise", [image1, image2], step=0)
```
nfelt, line 44: nit, but I generally try to avoid tag names with spaces in them, since it's kind of "un-idiomatic" for typical TensorBoard usage. In my experience `lower_with_underscores` is the typical style.  Ditto for other examples.

nfelt (commented):
Thanks very much for taking this on!  Feel free to have a TF docs person take a glance at it too (probably better than tagging them on the sync CL).

tensorboard/plugins/image/summary_v2.py
nfelt: Ok, caught up to the conversation on https://github.com/tensorflow/tensorboard/pull/4797#discussion_r597312820, so this is fine as is.

tensorboard/plugins/image/summary_v2.py
psybuzz: Done, all.

tensorboard/plugins/image/summary_v2.py
psybuzz: Ahh, that's embarrassing G_G.  I've double checked the latest examples.
image

tensorboard/plugins/image/summary_v2.py
psybuzz: Done, and will do in the .histogram PR as well.

tensorboard/plugins/image/summary_v2.py
psybuzz: Done, good idea.

tensorboard/plugins/image/summary_v2.py
psybuzz: Done.

tensorboard/plugins/image/summary_v2.py
psybuzz: Updated, but I haven't verified the markdown rendering yet.  @wolffg would you happen to know if this markdown would render properly, or if there's a way for us to check?

I tried running some internal "docs to devsite" target but ran into a "Seems like you don't have permissions to run these kokoro builds".

tensorboard/plugins/image/summary_v2.py
psybuzz: Done.

psybuzz: Updated the 2 rgb examples to both involve scaling, and make the 1st one use an initial range of [0,1000] outside of uint8 range.

tensorboard/plugins/image/summary_v2.py
psybuzz: Done, and mentioned `convert_image_dtype`.

tensorboard/plugins/image/summary_v2.py
```diff
@@ -27,6 +27,43 @@
...
+    Writes a collection of images to the current default summary writer. Data
+    appears in TensorBoard's 'Images' dashboard. Like `tf.summary.scalar` points,
+    each collection of images is associated with a `step` and a `name`.  All the
+    image collections with the same `name` constitute a time series of image
+    collections.
+
+    This example writes 2 random grayscale images:
+
+    ```python
+    w = tf.summary.create_file_writer('test/logs')
```
wolffg, line 41: Hey, for these examples, can you consider making these testable?  See: https://www.tensorflow.org/community/contribute/docs_ref for details.

psybuzz: I updated it to use "doctest style" `>>>`, `...`s, but I can't seem to get the test to run at all.

Digging into `bazel run //tensorflow/tools/docs:tf_doctest`, it seems that passing `--file` only works when the file is in the tensorflow repo, which TensorBoard's V2 summaries are not.  Passing `--module` and building from source didn't work for me, as I got some unrelated mlir error during the bazel build.

I did do a test sync cl/364014007 and saw that TAP presubmit passed.  Is this sufficient?

MarkDaoust: I had a quick look at this and confirmed @psybuzz's findings that there's no way to run it using the current `--file` or `--module` options. 

It looks like those two interactive options could be relatively simple to fix.

But, AFAIKT, it still wouldn't run as an actual test. The main  reason is that it's setup to find modules within `tensorflow.python`, so I don;t think tf_doctest would find this.

I think the least resistance way to get this working would be to copy tf_doctest into your package, like they did with [tf_estimator_doctest.py](https://github.com/tensorflow/estimator/blob/3fb3b55b05b3eb75f3ad284b010f964796ef7cae/tensorflow_estimator/python/estimator/tf_estimator_doctest.py)

>  cl/364014007  and saw that TAP presubmit passed. Is this sufficient?

I don't see the doctest in the list of tests that ran. Since TensorFlow's a separate project, this probably has something to do with the blueprint setup. If it had run, don't think it would test this, so that's fine.

psybuzz: Thanks for confirming!  For now, I would like to keep the scope of these PRs on the content and structure of the docs, and make the task of improving our infra by introducing TensorBoard doctests a separate PR.  These doctests, if run, would actually change the filesystem which is not ideal, so we might want to consider a doctest setup that mocks out filesystem effects.

I wasn't aware of tf_estimator_doctest.py, it's nice to know another team has done this pattern.

tensorboard/plugins/image/summary_v2.py
nfelt: LGTM, thanks!

nfelt (approved):

psybuzz:
Thanks for the reviews!

> Hmm, would you be able to follow up internally about how we can effectively preview these changes?

Will do.  @MarkDaoust shared with me some commands that I will try out.  For now, landing these PR as is.  After sync, we should be able to see changes appear in nightly docs shortly.

PR #4796, opened 2021-03-18
docs: update `tf.summary.scalar()`

bileschi:
update documentation for `tf.summary.scalar`.

Includes an example and a link to our featured demo notebook.

 #API_FIXIT_3_18_21

### DIFF
--- a/tensorboard/plugins/scalar/summary_v2.py
+++ b/tensorboard/plugins/scalar/summary_v2.py
@@ -30,6 +30,34 @@ from tensorboard.util import tensor_util
 def scalar(name, data, step=None, description=None):
     """Write a scalar summary.
 
+    See also `tf.summary.image`, `tf.summary.histogram`, `tf.summary.SummaryWriter`.
+
+    Logs simple numeric values for later analysis in TensorBoard.  Each summary
+    point is associated with an integral `step` value. This enables the incremental
+    logging of time series data.  A common usage of this API is to log loss during
+    training to produce a loss curve.
+
+    For example:
+
+    ```python
+    test_summary_writer = tf.summary.create_file_writer('test/demo_log_dir')
+    with test_summary_writer.as_default():
+        tf.summary.scalar('loss', 0.345, step=1)
+        tf.summary.scalar('loss', 0.234, step=2)
+        tf.summary.scalar('loss', 0.123, step=3)
+    ```
+
+    Multiple independent time series may be logged by giving each series a unique
+    `name` value.
+
+    See https://www.tensorflow.org/tensorboard/get_started for more examples
+    of effective usage of `tf.summary.scalar`.
+
+    There is no constraint to prevent the user of this API from logging multiple
+    scalar values to the same step of the same time series.  The way this
+    is interpreted depends on the consumer of the logs.
+
+
     Arguments:
       name: A name for this summary. The summary tag used for TensorBoard will
         be this name prefixed by any active name scopes.
--- a/tensorboard/plugins/scalar/summary_v2.py
+++ b/tensorboard/plugins/scalar/summary_v2.py
@@ -32,10 +32,11 @@ def scalar(name, data, step=None, description=None):
 
     See also `tf.summary.image`, `tf.summary.histogram`, `tf.summary.SummaryWriter`.
 
-    Logs simple numeric values for later analysis in TensorBoard.  Each summary
-    point is associated with an integral `step` value. This enables the incremental
-    logging of time series data.  A common usage of this API is to log loss during
-    training to produce a loss curve.
+    Writes simple numeric values for later analysis in TensorBoard.  Writes go to
+    the current default summary writer. Each summary point is associated with an
+    integral `step` value. This enables the incremental logging of time series
+    data.  A common usage of this API is to log loss during training to produce
+    a loss curve.
 
     For example:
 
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/scalar/summary_v2.py
```diff
@@ -30,6 +30,34 @@
...
+
+    ```python
+    test_summary_writer = tf.summary.create_file_writer('test/demo_log_dir')
+    with test_summary_writer.as_default():
+        tf.summary.scalar('loss', 0.345, step=1)
+        tf.summary.scalar('loss', 0.234, step=2)
+        tf.summary.scalar('loss', 0.123, step=3)
+    ```
+
+    Multiple independent time series may be logged by giving each series a unique
```
psybuzz, line 50: This paragraph and the "There is no constraint..." paragraph below on L56 apply to all the tf.summary.* APIs, not just scalars.  I wonder if it makes more sense to move these 2 paragraphs out of here, and maybe more appropriate in the "overview" page for the entire tf.summary module?  (https://www.tensorflow.org/api_docs/python/tf/summary)

tensorboard/plugins/scalar/summary_v2.py
```diff
@@ -30,6 +30,34 @@
 def scalar(name, data, step=None, description=None):
     """Write a scalar summary.
 
+    See also `tf.summary.image`, `tf.summary.histogram`, `tf.summary.SummaryWriter`.
```
psybuzz, line 33: Similar to the other PR https://github.com/tensorflow/tensorboard/pull/4797, wdyt about either changing the text to something like "See the rest of the tf.summary.* methods for ways to write other data types", or perhaps moving these comments to the tf.summary module "overview" doc page?

bileschi: Agreed that all the tf.summary symbols are relevant, some more than others, but these are here the way they are to be conformant with the "See also" functionality in the TensorFlow devsite, lending consistency.

tensorboard/plugins/scalar/summary_v2.py
bileschi: I suspect, as the most fundamental type, that the `tf.scalar` symbol will be most users' first interaction with the summary API.  I support adding this information to in the higher`tf/summary` documentation, but think it is valuable here also.

tensorboard/plugins/scalar/summary_v2.py
```diff
@@ -30,6 +30,34 @@
...
+
+    Logs simple numeric values for later analysis in TensorBoard.  Each summary
+    point is associated with an integral `step` value. This enables the incremental
+    logging of time series data.  A common usage of this API is to log loss during
+    training to produce a loss curve.
+
+    For example:
+
+    ```python
+    test_summary_writer = tf.summary.create_file_writer('test/demo_log_dir')
```
psybuzz, line 43: In the .text() PR, it looks like the logdir was renamed to "test/logdir".  Let's update this one as well, and I'll update the .histogram(), .image() PRs to be consistent as well.

tensorboard/plugins/scalar/summary_v2.py
psybuzz: Sounds reasonable to me.

psybuzz (approved):
lgtm!

tensorboard/plugins/scalar/summary_v2.py
```diff
@@ -30,6 +30,34 @@
 def scalar(name, data, step=None, description=None):
     """Write a scalar summary.
 
+    See also `tf.summary.image`, `tf.summary.histogram`, `tf.summary.SummaryWriter`.
+
+    Logs simple numeric values for later analysis in TensorBoard.  Each summary
```
nfelt, line 35: Useful for consistency to mention something here about the current default writer, e.g. like wording suggested in https://github.com/tensorflow/tensorboard/pull/4795/files#r597993135

Would also recommend s/Logs/Writes/ just so we maintain a consistent style throughout.

tensorboard/plugins/scalar/summary_v2.py
```diff
@@ -30,6 +30,34 @@
...
+    with test_summary_writer.as_default():
+        tf.summary.scalar('loss', 0.345, step=1)
+        tf.summary.scalar('loss', 0.234, step=2)
+        tf.summary.scalar('loss', 0.123, step=3)
+    ```
+
+    Multiple independent time series may be logged by giving each series a unique
+    `name` value.
+
+    See https://www.tensorflow.org/tensorboard/get_started for more examples
```
nfelt, line 53: I'm not 100% sure if bare links linkify automatically for tensorflow.org.  Maybe ask the docs people?  E.g. here's one example that doesn't: https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/python/distribute/tpu_strategy.py;l=246;drc=adf69bf65356f810fd4c98cd6d3914ca1d451ca0

Compare with rendering: https://www.tensorflow.org/api_docs/python/tf/distribute/TPUStrategy?version=nightly

tensorboard/plugins/scalar/summary_v2.py
```diff
@@ -30,6 +30,34 @@
...
+    of effective usage of `tf.summary.scalar`.
+
+    There is no constraint to prevent the user of this API from logging multiple
```
nfelt, line 56: This comment seems potentially a little confusing?  Maybe worth framing this as something a little more explicit:

"In general, this API expects that data points are logged with a monotonically increasing step value.  Duplicate points for a single step or points logged out of order by step are not guaranteed to display as desired in TensorBoard."

tensorboard/plugins/scalar/summary_v2.py
nfelt: I agree that we probably do want some general docs on the module about common behavior.  Maybe after the individual symbol updates we can take a pass over the entire module to harmonize the docstrings a little and link to common module-level docs as need be.

nfelt (approved):
Thanks for doing this!  LGTM pending a couple questions for the docs people / to verify ourselves.

tensorboard/plugins/scalar/summary_v2.py
nfelt: Ok, caught up to the conversation on https://github.com/tensorflow/tensorboard/pull/4797#discussion_r597312820.  I guess it is a reasonably common pattern (though I'm not sure I see that it's specially handled by the tf.org software?) so in that case fine as is.

tensorboard/plugins/scalar/summary_v2.py
bileschi: Updated with Markdown style link.

PR #4797, opened 2021-03-18
docs: update `tf.summary.text()`

bileschi:
Update documentation for `tf.summary.text`.

Adds some examples and examples of what features are supported.

#API_FIXIT_3_18_21

### DIFF
--- a/tensorboard/plugins/text/summary_v2.py
+++ b/tensorboard/plugins/text/summary_v2.py
@@ -28,9 +28,10 @@ def text(name, data, step=None, description=None):
 
     See also `tf.summary.scalar`, `tf.summary.SummaryWriter`, `tf.summary.image`.
 
-    Logs text Tensor values for later visualization and analysis in TensorBoard.  Like
-    `tf.summary.scalar` points, text points are each associated with a `step` and a
-    `name`.  All the points with the same `name` constitute a time series of text values.
+    Writes text Tensor values for later visualization and analysis in TensorBoard.
+    Writes go to the current default summary writer.  Like `tf.summary.scalar`
+    points, text points are each associated with a `step` and a `name`.
+    All the points with the same `name` constitute a time series of text values.
 
     For Example:
     ```python
@@ -45,20 +46,23 @@ def text(name, data, step=None, description=None):
 
     ```python
     with test_summary_writer.as_default():
-        table_as_markdown_text = '''
+        text_data = '''
               | *hello* | *there* |
               |---------|---------|
               | this    | is      |
               | a       | table   |
         '''
-        tf.summary.text('markdown_text', table_as_markdown_text, step=0)
+        text_data = '\n'.join(l.strip() for l in text_data.splitlines())
+        tf.summary.text('markdown_text', text_data, step=0)
     ```
 
     Since text is Tensor valued, each text point may be a Tensor of string values.
     rank-1 and rank-2 Tensors are rendered as tables in TensorBoard.  For higher ranked
-    Tensors, you'll see just a 2D slice of the data.
+    Tensors, you'll see just a 2D slice of the data.  To avoid this, reshape the Tensor
+    to at most rank-2 prior to passing it to this function.
 
-    Demo notebook at https://www.tensorflow.org/tensorboard/text_summaries.
+    Demo notebook at
+    ["Displaying text data in TensorBoard"](https://www.tensorflow.org/tensorboard/text_summaries).
 
     Arguments:
       name: A name for this summary. The summary tag used for TensorBoard will
--- a/tensorboard/plugins/text/summary_v2.py
+++ b/tensorboard/plugins/text/summary_v2.py
@@ -58,6 +58,8 @@ def text(name, data, step=None, description=None):
     rank-1 and rank-2 tensors are rendered as tables in TensorBoard.  For higher ranked
     tensors, you'll see just a 2D slice of the data.
 
+    Demo notebook at https://www.tensorflow.org/tensorboard/text_summaries.
+
     Arguments:
       name: A name for this summary. The summary tag used for TensorBoard will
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/text/summary_v2.py
```diff
@@ -26,6 +26,40 @@
 def text(name, data, step=None, description=None):
     """Write a text summary.
 
+    See also `tf.summary.scalar`, `tf.summary.SummaryWriter`, `tf.summary.image`.
```
psybuzz, line 29: Maybe personal taste, but I think that seeing '.scalar, .SummaryWriter, .image' is a bit redundant with the left sidepane for navigation in the devsite, especially if all tf.summary.* pages will mention this.

wdyt about something more generic along the lines of: "See the rest of the `tf.summary.*` methods for ways to write other data types"?

imo, this "see also" might make more sense at the bottom, near the mention of the demo notebook.
wdyt?

tensorboard/plugins/text/summary_v2.py
```diff
@@ -26,6 +26,40 @@
...
+        table_as_markdown_text = '''
+              | *hello* | *there* |
+              |---------|---------|
+              | this    | is      |
+              | a       | table   |
+        '''
+        tf.summary.text('markdown_text', table_as_markdown_text, step=0)
+    ```
+
+    Since text is tensor valued, each text point may be a tensor of string values.
```
psybuzz, line 57: Our docs aren't always consistent about writing "tensor" vs "\`Tensor\`", but I see the latter in many of the TF devsite docs.  Should we replace instances in this paragraph with "\`Tensor\`"?

tensorboard/plugins/text/summary_v2.py
```diff
@@ -26,6 +26,40 @@
 def text(name, data, step=None, description=None):
     """Write a text summary.
 
+    See also `tf.summary.scalar`, `tf.summary.SummaryWriter`, `tf.summary.image`.
+
+    Logs text tensor values for later visualization and analysis in TensorBoard.  Like
+    `tf.summary.scalar` points, text points are each associated with a `step` and a
+    `name`.  All the points with the same `name` constitute a time series of text values.
```
psybuzz, line 33: Nice.  Much of this paragraph applies to tf.summary.image, so I will copy/paste if that is alright :)

psybuzz (approved):
overall lgtm, I will update the .image() PR to be more consistent with this.

we might want to wait for nfelt to lgtm before landing as well

tensorboard/plugins/text/summary_v2.py
```diff
@@ -26,6 +26,40 @@
...
+
+    For Example:
+    ```python
+    test_summary_writer = tf.summary.create_file_writer('test/text_demo_log_dir')
```
psybuzz, line 37: nit: personally, I prefer shorter logdir like "logs" or "logdir".  wdyt?

tensorboard/plugins/text/summary_v2.py
bileschi: I put these here to be consistent with the ui pattern for TensorFlow's https://www.tensorflow.org/api_docs/python/tf/concat

Note that the See Also section is handled specially by the docsite software.

tensorboard/plugins/text/summary_v2.py
bileschi: done

tensorboard/plugins/text/summary_v2.py
bileschi: done

tensorboard/plugins/text/summary_v2.py
psybuzz: Gotcha, after seeing tf.concat's docs, this makes sense.  I'll also add similar "See also"s for .image, .histogram!

tensorboard/plugins/text/summary_v2.py
```diff
@@ -26,10 +26,44 @@
 def text(name, data, step=None, description=None):
     """Write a text summary.
 
+    See also `tf.summary.scalar`, `tf.summary.SummaryWriter`, `tf.summary.image`.
+
+    Logs text Tensor values for later visualization and analysis in TensorBoard.  Like
```
nfelt, line 31: Same comment as other PR re: mentioning default writer, consistency w.r.t. choice of leading verb.

tensorboard/plugins/text/summary_v2.py
```diff
@@ -26,10 +26,44 @@
...
+              | a       | table   |
+        '''
+        tf.summary.text('markdown_text', table_as_markdown_text, step=0)
+    ```
+
+    Since text is Tensor valued, each text point may be a Tensor of string values.
+    rank-1 and rank-2 Tensors are rendered as tables in TensorBoard.  For higher ranked
+    Tensors, you'll see just a 2D slice of the data.
+
+    Demo notebook at https://www.tensorflow.org/tensorboard/text_summaries.
```
nfelt, line 61: Ditto here again about whether we've checked a bare link works?

tensorboard/plugins/text/summary_v2.py
```diff
@@ -26,10 +26,44 @@
...
+              |---------|---------|
+              | this    | is      |
+              | a       | table   |
+        '''
+        tf.summary.text('markdown_text', table_as_markdown_text, step=0)
+    ```
+
+    Since text is Tensor valued, each text point may be a Tensor of string values.
+    rank-1 and rank-2 Tensors are rendered as tables in TensorBoard.  For higher ranked
+    Tensors, you'll see just a 2D slice of the data.
```
nfelt, line 59: Maybe add "to avoid this, reshape the Tensor to at most rank-2 prior to passing it to this function."?

nfelt (approved):
Thanks for doing this!

PR #4798, opened 2021-03-19
docs: update `tf.summary.histogram()`

psybuzz:
Update documentation for `tf.summary.histogram`.

#API_FIXIT_3_18_21

From what I understand, OSS is the source of truth, but the
TF Devsite is not updated immediately without manually doing
an internal process to update the devsite.

### DIFF
--- a/tensorboard/plugins/histogram/summary_v2.py
+++ b/tensorboard/plugins/histogram/summary_v2.py
@@ -49,8 +49,8 @@ def histogram(name, data, step=None, buckets=None, description=None):
     each histogram is associated with a `step` and a `name`. All the histograms
     with the same `name` constitute a time series of histograms.
 
-    When higher order `Tensor`s are flattened to produce a histogram of all
-    entries within.
+    The histogram is calculated over all the elements of the given `Tensor`
+    without regard to its shape or rank.
 
     This example writes 2 histograms:
 
@@ -83,7 +83,8 @@ def histogram(name, data, step=None, buckets=None, description=None):
     Arguments:
       name: A name for this summary. The summary tag used for TensorBoard will
         be this name prefixed by any active name scopes.
-      data: A `Tensor` of any shape. Must be castable to `float64`.
+      data: A `Tensor` of any shape. The histogram is computed over its elements,
+        which must be castable to `float64`.
       step: Explicit `int64`-castable monotonic step value for this summary. If
         omitted, this defaults to `tf.summary.experimental.get_step()`, which must
         not be None.
--- a/tensorboard/plugins/histogram/summary_v2.py
+++ b/tensorboard/plugins/histogram/summary_v2.py
@@ -41,6 +41,8 @@ DEFAULT_BUCKET_COUNT = 30
 def histogram(name, data, step=None, buckets=None, description=None):
     """Write a histogram summary.
 
+    See also `tf.summary.scalar`, `tf.summary.SummaryWriter`.
+
     Logs a histogram to the current log directory for later analysis in
     TensorBoard's 'Histograms' and 'Distributions' dashboards (one histogram
     will appear in both). Like `tf.summary.scalar` points, each histogram is
@@ -50,8 +52,8 @@ def histogram(name, data, step=None, buckets=None, description=None):
     This example writes 2 histograms:
 
     ```python
-    writer = tf.summary.create_file_writer('logs')
-    with writer.as_default():
+    test_summary_writer = tf.summary.create_file_writer('test/logs')
+    with test_summary_writer.as_default():
       tf.summary.histogram("activations", [-10, 0, 10], step=0)
       tf.summary.histogram("initial weights", tf.random.normal([5]), step=0)
     ```
@@ -60,8 +62,8 @@ def histogram(name, data, step=None, buckets=None, description=None):
     thereof) at specific layers in a neural network, over time.
 
     ```python
-    writer = tf.summary.create_file_writer('logs')
-    with writer.as_default():
+    test_summary_writer = tf.summary.create_file_writer('test/logs')
+    with test_summary_writer.as_default():
       for step in steps:
         # After computing activations...
         tf.summary.histogram("layer1/activate", layer1_activate, step=step)
--- a/tensorboard/plugins/histogram/summary_v2.py
+++ b/tensorboard/plugins/histogram/summary_v2.py
@@ -54,26 +54,31 @@ def histogram(name, data, step=None, buckets=None, description=None):
 
     This example writes 2 histograms:
 
-    >>> w = tf.summary.create_file_writer('test/logs')
-    >>> with w.as_default():
-    ...     tf.summary.histogram("activations", tf.random.uniform([100, 50]), step=0)
-    ...     tf.summary.histogram("initial_weights", tf.random.normal([1000]), step=0)
+    ```python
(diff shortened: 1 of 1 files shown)


bileschi:
Not merging, as PR author has credentials to do so and may want to make additional tweaks.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -41,6 +41,36 @@
 def histogram(name, data, step=None, buckets=None, description=None):
     """Write a histogram summary.
 
+    See also `tf.summary.scalar`, `tf.summary.SummaryWriter`.
+
+    Logs a histogram to the current log directory for later analysis in
```
nfelt, line 46: Same comment as on #4795  re: opening line.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -41,6 +41,36 @@
...
+    TensorBoard's 'Histograms' and 'Distributions' dashboards (one histogram
```
nfelt, line 47: slightly clearer imo for the parenthetical to say "(data written using this API will appear in both places)".

Otherwise "one histogram" makes me think "only one of them? which one?"

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -41,6 +41,36 @@
...
+    associated with a `step` and a `name`. All the histograms with the same
+    `name` constitute a time series of histograms.
+
+    This example writes 2 histograms:
+
+    ```python
+    test_summary_writer = tf.summary.create_file_writer('test/logs')
+    with test_summary_writer.as_default():
+      tf.summary.histogram("activations", [-10, 0, 10], step=0)
+      tf.summary.histogram("initial weights", tf.random.normal([5]), step=0)
```
nfelt, line 58: ditto re: spaces in tag names

nfelt, line 58: I'd consider using a bigger size than 3 and 5 which is A) a bit more realistic and B) will result in a nicer histogram, rather than showing how the histogram smoothing is wonky when there are only a few values.  Something >30 (default number of buckets) would be best.  If we just use `tf.random.normal()` you might as well pass `[1000]` or something so the plot looks pretty.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -41,6 +41,36 @@
...
+    ```
+
+    A common use case is to examine the changing activation patterns (or lack
+    thereof) at specific layers in a neural network, over time.
+
+    ```python
+    test_summary_writer = tf.summary.create_file_writer('test/logs')
+    with test_summary_writer.as_default():
+      for step in steps:
```
nfelt, line 67: Would be a little nicer if this example was runnable code.  Could do that via e.g. s/steps/range(10)/, and then passing in something like `tf.random.normal([1000], mean=step, stddev=N)` where N is the layer number.  That gives you a more appealing result like this:

Screen Shot 2021-03-19 at 3 37 50 PM

nfelt (changes requested):
Thanks for doing this!

tensorboard/plugins/histogram/summary_v2.py
psybuzz: Done.

tensorboard/plugins/histogram/summary_v2.py
psybuzz: Done.

tensorboard/plugins/histogram/summary_v2.py
psybuzz: Done.

psybuzz: Done.

tensorboard/plugins/histogram/summary_v2.py
psybuzz: Nice idea, done.

psybuzz (commented):
Note: also added a note about how higher order tensors just get flattened.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -41,6 +41,45 @@
...
+    each histogram is associated with a `step` and a `name`. All the histograms
+    with the same `name` constitute a time series of histograms.
+
+    When higher order `Tensor`s are flattened to produce a histogram of all
+    entries within.
+
+    This example writes 2 histograms:
+
+    ```python
+    w = tf.summary.create_file_writer('test/logs')
```
wolffg, line 58: Is it possible to make this a testable example?  See https://www.tensorflow.org/community/contribute/docs_ref

psybuzz: (response copied from https://github.com/tensorflow/tensorboard/pull/4795)

I updated it to use "doctest style" `>>>`, `...`s, but I can't seem to get the test to run at all.

Digging into `bazel run //tensorflow/tools/docs:tf_doctest`, it seems that passing `--file` only works when the file is in the tensorflow repo, which TensorBoard's V2 summaries are not.  Passing `--module` and building from source didn't work for me, as I got some unrelated mlir error during the bazel build.

I did do a test sync cl/364023724, which runs an TAP presubmit.  Is this sufficient?

psybuzz: Similar to the tf.summary.image() PR, I've reverted the doctest style for now, with the intent of keeping these PRs narrowly scoped.  Introducing a doctest process in our workflow sounds like a good, separate task.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -43,32 +43,41 @@ def histogram(name, data, step=None, buckets=None, description=None):
...
-    will appear in both). Like `tf.summary.scalar` points, each histogram is
-    associated with a `step` and a `name`. All the histograms with the same
-    `name` constitute a time series of histograms.
+    Writes a histogram to the current default summary writer, for later analysis
+    in TensorBoard's 'Histograms' and 'Distributions' dashboards (data written
+    using this API will appear in both places). Like `tf.summary.scalar` points,
+    each histogram is associated with a `step` and a `name`. All the histograms
+    with the same `name` constitute a time series of histograms.
+
+    When higher order `Tensor`s are flattened to produce a histogram of all
```
nfelt, line 52: This line seems like it's missing a word?  Maybe you meant something like "When _called with_ higher order tensors, they are flattened..."?

If that's the case, it's a bit imprecise what "higher order" actually means here.  If we do mention flattening I might do so in the Args section below, e.g. extend the `data` comment to

> A `Tensor` of any shape. The histogram is computed over its elements, which must be castable to `float64`.

Or if we want a general comment up here, maybe phrase it more like "The histogram is calculated over the elements of the given tensor without regard to its shape or rank."

nfelt (approved):

tensorboard/plugins/histogram/summary_v2.py
psybuzz: I like the wording, done.

tensorboard/plugins/histogram/summary_v2.py
psybuzz: For posterity, b/183515215 tracks the internal effort to disentangle the tf.summary API from the TensorBoard repo, which might make the tensorflow/ repo the natural place for these doctests in the future.

PR #4803, opened 2021-03-19
graph: relax the aggressive numeric node series detection

psybuzz:
Before, these nodes [add_0, add_1, add_2, ...] could be collapsed into
a single "series group node", while [add1, add2, add3, ...] could not.
This change relaxes the strict regex check to allow names without
underscores.

Note that the "minimum 5 node series length" constraint still applies,
as the main heuristic that guards against overly aggressive series
grouping.

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_common/common_test.ts
+++ b/tensorboard/plugins/graph/tf_graph_common/common_test.ts
@@ -14,9 +14,11 @@ limitations under the License.
 ==============================================================================*/
 import * as tb_debug from '../../../components/tb_debug';
 import * as tf_graph from './graph';
+import * as tf_graph_common from './common';
 import * as tf_graph_hierarchy from './hierarchy';
 import * as tf_graph_loader from './loader';
 import * as tf_graph_parser from './parser';
+import * as tf_graph_proto from './proto';
 
 describe('graph tests', () => {
   let mockTracker: jasmine.SpyObj<any>;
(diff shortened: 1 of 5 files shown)


tensorboard/plugins/graph/tf_graph_common/common_test.ts
```diff
@@ -192,5 +194,197 @@ describe('graph tests', () => {
         ])
       );
     });
+
+    it('groups numeric series of nodes without underscores', async () => {
```
stephanwlee, line 198: to codify the expected behavior for the future reader, add tests for:
- non-sequential numbering (e.g., 2 → 3 → 1 and 5 → 6 → 7 and 10 → 20 → 21) 
- mixed format (e.g., `foo_2` → `foo3` or `conv3` → `conv2` → `conv2_1`)
- any others? like `foo` → `bar` → `foo_2` → `foo3` → `bar_2`

psybuzz: Added more test cases

tensorboard/plugins/graph/tf_graph_common/common_test.ts
```diff
@@ -192,5 +194,364 @@ describe('graph tests', () => {
...
+  graphDef: tf_graph_proto.GraphDef
+): Promise<tf_graph.SlimGraph> {
+  return tf_graph.build(
+    graphDef,
+    tf_graph.DefaultBuildParams,
+    createProgressTracker()
+  );
+}
+
+function createProgressTracker(): tf_graph_common.ProgressTracker {
```
stephanwlee, line 551: `createProgressTracker` → `createNoopProgressTracker`

stephanwlee (approved):
Thanks, I did not know how the series grouping would work under the cases depicted in the test and TIL.

tensorboard/plugins/graph/tf_graph_common/common_test.ts
psybuzz: Done

psybuzz (commented):
Thanks!

PR #4804, opened 2021-03-19
rust: support and default to `--host localhost`

### LINKED ISSUE #4801: In Docker, RustBoard can't bind to `::1`
On current `tensorflow/tensorflow@nightly`:

```
$ docker run --rm -it tensorflow/tensorflow@sha256:31775f136e5fe2d50ae075100dff335e9ae0954b8f9b529791a8bf739f3dd415 sh
# rustboard="$(python -c 'import tensorboard_data_server as tds; print(tds.server_binary())')"
# "${rustboard}" --version
rustboard 0.4.0
# mkdir logs
# "${rustboard}" --logdir logs  
Error: Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" }
# "${rustboard}" --logdir logs --host ::0
listening on [::]:6806
^C
# "${rustboard}" --logdir logs --host ::1
Error: Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" }
# "${rustboard}" --logdir logs --host 0.0.0.0
listening on 0.0.0.0:6806
^C
# "${rustboard}" --logdir logs --host 127.0.0.1
listening on 127.0.0.1:6806
^C
```

It looks like we can bind to IPv4 wildcard and loopback, and IPv6
wildcard, but not IPv6 loopback. Poking around the internet, this seems
to be a common issue with Docker, but I can’t find an explanation;
everyone just says to bind to IPv4 instead.

We should do something nicer here: presumably, attempt to bind to IPv6,
and fall back to IPv4.

(Thanks to @tgolsson for the [original report]!)

[original report]: https://github.com/tensorflow/tensorboard/issues/4784#issuecomment-802706654

wchargin:
Summary:
We now interpret `--host` as a string that can be `getaddrinfo`d rather
than requiring it to be a literal IP address. Thus, we can support
hostnames like `localhost` that [may have multiple resolutions][play]:

```
[src/main.rs:3] ("localhost", 6006).to_socket_addrs().unwrap().collect::<Vec<_>>() = [
    127.0.0.1:6006,
    [::1]:6006,
]
```

[play]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=15504b49f87e983b4748ca976dcb3bd1

Thus, we change the default `--host` from `::1` to `localhost` to
support systems where `::1` may not be available. Should fix #4801.

We also improve the error message on bind failure: it’s pretty-printed
instead of `Debug`-formatted, and it includes the problematic address.

Test Plan:
From `tensorboard/data/server/`, run `cargo build`, then reproduce the
failure in Docker:

```
$ img=tensorflow/tensorflow@sha256:31775f136e5fe2d50ae075100dff335e9ae0954b8f9b529791a8bf739f3dd415
$ v="$PWD/target/debug:/rustboard-bin:ro"
$ docker run --rm -it -v "$v" "$img" /rustboard-bin/rustboard --logdir /tmp --host ::1
fatal: failed to bind to ("::1", 6806): Cannot assign requested address (os error 99)
```

…then try again with default `--host` and note that it works:

```
$ docker run --rm -it -v "$v" "$img" /rustboard-bin/rustboard --logdir /tmp
listening on 127.0.0.1:6806
```

(You may need to `docker kill $(docker ps -q)` the thing afterward.)

wchargin-branch: rust-host-getaddrinfo

### DIFF
--- a/tensorboard/data/server/cli.rs
+++ b/tensorboard/data/server/cli.rs
@@ -19,7 +19,6 @@ use clap::Clap;
 use log::{debug, error, info, LevelFilter};
 use std::fs::File;
 use std::io::{Read, Write};
-use std::net::{IpAddr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
@@ -52,12 +51,12 @@ struct Opts {
     #[clap(long, setting(clap::ArgSettings::AllowEmptyValues))]
     logdir: PathBuf,
 
-    /// Bind to this IP address
+    /// Bind to this host name
     ///
-    /// IP address to bind this server to. May be an IPv4 address (e.g., 127.0.0.1 or 0.0.0.0) or
-    /// an IPv6 address (e.g., ::1 or ::0).
-    #[clap(long, default_value = "::1")]
-    host: IpAddr,
+    /// Host to bind this server to. May be an IPv4 address (e.g., 127.0.0.1 or 0.0.0.0), an IPv6
+    /// address (e.g., ::1 or ::0), or a string like `localhost` to pass to `getaddrinfo(3)`.
+    #[clap(long, default_value = "localhost")]
+    host: String,
 
     /// Bind to this port
     ///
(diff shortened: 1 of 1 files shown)


wchargin:
If it’s not too much trouble, would you mind giving this a quick test on
macOS? No Docker tomfoolery required, just:

```
bazel run --define=link_data_server=true //tensorboard -- --logdir ~/tensorboard_data/ --bind_all --load_fast=auto --verbosity 0
```

and make sure that you can see some data in TensorBoard. It should all
be fine, but with network-y things it can be helpful to double-check.

stephanwlee (approved):

PR #4805, opened 2021-03-19
what's new: display red dot when there are unread messages

japie1235813:
To handle more logics on the front-end component. Includes:
  * display red dot when there are unread messages
  * remove red dot when the user click the bell
  *  read/write last read timestamp from local storage 

* Screenshots of UI changes
first time loading
https://user-images.githubusercontent.com/1131010/111849167-32ae7f00-88ca-11eb-962f-a83afdcb0483.mov

no red dot for the second time when the local storage shows the last read timestamp


tensorboard/webapp/notification_center/_redux/notification_center_reducers.ts
```diff
@@ -14,16 +14,38 @@ limitations under the License.
...
+const reducer = createReducer(
+  initialState,
+  on(
+    actions.notificationBellClicked,
+    (state: NotificationState): NotificationState => {
+      const timeNow = Date.now();
+      window.localStorage.setItem(
+        NOTIFICATION_LAST_READ_TIME_KEY,
+        timeNow.toString()
+      );
```
stephanwlee, line 41: Are you going to move this to a DataSource? Please add a TODO to indicate that this is not the final state.

tensorboard/webapp/notification_center/_redux/notification_center_selectors_test.ts
```diff
@@ -57,5 +58,15 @@ describe('notification_center_selectors', () => {
         },
       ]);
     });
+
+    it('get last read timestamp', () => {
```
stephanwlee, line 62: `get` → `returns`.

stephanwlee, line 62: can it be a `null`? If so, write a simple test.

tensorboard/webapp/notification_center/_redux/notification_center_types.ts
```diff
@@ -29,9 +29,11 @@ export interface Notification {
 }
 
 export const NOTIFICATION_FEATURE_KEY = 'notification';
+export const NOTIFICATION_LAST_READ_TIME_KEY = 'lastReadNotificaitonTimestamp';
 
 export interface NotificationState {
   notifications: Notification[];
+  lastReadTimestamp: number;
```
stephanwlee, line 36: I think this should be nullable to indicate that it has not been loaded yet.

For instance, if we were persist this information in the backend, we want to make this state fetchable from the backend and we have to account for the flight time and should be able to tell difference between 'not fetched yet' and UTC time = 0.

tensorboard/webapp/notification_center/_views/notification_center_component.scss
```diff
@@ -13,6 +13,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
+.red-dot {
+  background-color: #d93025;
```
stephanwlee, line 17: If possible, use a material color variable; something like `mat-color($mat-red, 700)` 

In a retrospect, you should change `category-icon` as well

tensorboard/webapp/notification_center/_views/notification_center_component.scss
```diff
@@ -13,6 +13,16 @@ See the License for the specific language governing permissions and
...
+  border-radius: 5px;
+  height: 10px;
+  position: absolute;
```
stephanwlee, line 20: Is this component a `relative` container? If not, you may want to put one since it can wrongly place the red dot.

tensorboard/webapp/notification_center/_views/notification_center_component.scss
```diff
@@ -13,6 +13,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
+.red-dot {
+  background-color: #d93025;
+  border-radius: 5px;
```
stephanwlee, line 18: `5px` → `50%` or use a SASS variable like
```scss
$_dim: 10px;
border-radius: $_dim / 2;
height: $_dim;
width: $_dim;
```

tensorboard/webapp/notification_center/_views/notification_center_component.ts
```diff
@@ -25,4 +25,7 @@ export class NotificationCenterComponent {
   readonly CategoryEnum = CategoryEnum;
 
   @Input() notifications!: ViewNotificationExt[];
+  @Input() hasUnreadMessages!: boolean;
+
+  @Output() bellIconClicked = new EventEmitter<void>();
```
stephanwlee, line 30: `bellIconClicked` → `bellButtonClicked`?

tensorboard/webapp/notification_center/_views/notification_center_test.ts
```diff
@@ -47,6 +47,8 @@ describe('notification center', () => {
       ],
     }).compileComponents();
     store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
```
stephanwlee, line 49: Test for the interaction (bell button click) is missing.

japie1235813: add 3 more tests, please help to check. Thanks!

tensorboard/webapp/notification_center/_redux/notification_center_types.ts
japie1235813: added null type. then return `-1` in `getLastReadTime` selector. Please let me know any suggestions.

tensorboard/webapp/notification_center/_redux/notification_center_selectors_test.ts
japie1235813: added.

tensorboard/webapp/notification_center/_redux/notification_center_reducers.ts
japie1235813: added

tensorboard/webapp/notification_center/_redux/notification_center_reducers_test.ts
```diff
@@ -12,3 +12,19 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+import * as notificationActions from './notification_center_actions';
+import * as notificationReducers from './notification_center_reducers';
+import {buildNotificationState} from './testing';
+
+describe('notification reducers', () => {
+  it('set lastReadTimestamp to current after bell icon clicked', () => {
```
stephanwlee, line 20: `it('set` → `it('sets` as per our convention.

tensorboard/webapp/notification_center/_redux/notification_center_selectors.ts
```diff
@@ -32,3 +32,10 @@ export const getNotifications = createSelector(
     return state.notifications;
   }
 );
+
+export const getLastReadTime = createSelector(
+  selectNotifications,
+  (state: NotificationState): number => {
+    return state.lastReadTimestamp ? state.lastReadTimestamp : -1;
```
stephanwlee, line 39: 1. it should be `state.lastReadTimestamp ?? -1;` (otherwise lastReadTimestamp = 0 would return -1)
2. can you specify time unit in the state and the selector
3. I think this thing should return `null` to let the UI know that it is not loaded yet. WDYT? Not a strong opinion here.

tensorboard/webapp/notification_center/_views/notification_center_component.scss
```diff
@@ -12,6 +12,23 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+@import 'tensorboard/webapp/theme/tb_theme';
+
+$_dim: 10px;
```
stephanwlee, line 17: if `dim` is not used outside, prefer to define it inside the `.red-dot`

tensorboard/webapp/notification_center/_views/notification_center_component.scss
```diff
@@ -12,6 +12,23 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+.notification-icon-button {
+  position: relative;
+}
+
+.red-dot {
+  background-color: mat-color($mat-red, 700);
+  border-radius: $_dim / 2;
+  height: $_dim;
+  position: absolute;
+  right: $_dim;
```
stephanwlee, line 28: the fact that `right` == `$_dim` was a coincidence but not by design, right? If not, could you tell me why it is?

tensorboard/webapp/notification_center/_views/notification_center_test.ts
```diff
@@ -82,4 +91,49 @@ describe('notification center', () => {
...
+        title: 'test title',
+        content: 'test content',
+      },
+    ]);
+    store.overrideSelector(selectors.getLastReadTime, 1);
+    const fixture = TestBed.createComponent(NotificationCenterContainer);
+    fixture.detectChanges();
+    const reddot = fixture.debugElement.query(By.css('.red-dot'));
+    expect(reddot).toBeNull();
+  });
```
stephanwlee, line 138: put these under describe block:

```ts
describe("unread dot (red dot)", () => {
  it('appears when unread notifications exist', () => {
    ...
})
```

tensorboard/webapp/notification_center/_views/view_types.ts
```diff
@@ -17,7 +17,12 @@ import {Notification} from '../_redux/notification_center_types';
...
-
 export interface ViewNotificationExt extends Notification {
   icon: string | null;
+  hasRead: boolean;
 }
+
+/**
+ * The key used to record the latest read timestamp in local storage.
+ */
+export const NOTIFICATION_LAST_READ_DATE_KEY = 'notificaitonLastReadDateInMs';
```
stephanwlee, line 28: prob should be defined in the store type OR data_source type. Definitely should not be defined in the view types.

tensorboard/webapp/notification_center/_views/notification_center_component.scss
japie1235813: You're right. I misplaced it.

tensorboard/webapp/notification_center/_views/view_types.ts
japie1235813: This was accidentally left when I move the key to _redux. It is not used anywhere now. Removed.

tensorboard/webapp/notification_center/_redux/notification_center_selectors.ts
japie1235813: I think the UI only cares if the red dot should appear or not. If there is no timestamp in local storage, it is equivalent to `-1` which makes the condition (notification.dateInMs < lastReadTime) be true
I prefer to leave the logics on _redux side. The UI could be quite clean this way.
Unless... there are some cases that the `getLastReadTime` returns `null` for other reasons?

japie1235813: 1. addressed. Thanks!
2. Renamed `lastReadTimestamp` to `lastReadTimestampInMs`

tensorboard/webapp/notification_center/_redux/notification_center_selectors.ts
```diff
@@ -32,3 +32,10 @@ export const getNotifications = createSelector(
     return state.notifications;
   }
 );
+
+export const getLastReadTime = createSelector(
```
stephanwlee, line 36: specify the unit here for sure.

tensorboard/webapp/notification_center/_views/notification_center_component.ng.html
```diff
@@ -18,7 +18,10 @@
   mat-icon-button
   [matMenuTriggerFor]="menu"
   aria-label="Display notification messages"
+  class="notification-icon-button"
```
stephanwlee, line 21: consideration for future:  you may want to make this button disabled while the data is being fetched.

stephanwlee (approved):

tensorboard/webapp/notification_center/_views/notification_center_component.ng.html
japie1235813: Noted. Thanks!

PR #4806, opened 2021-03-19
pip: require tensorboard-data-server 0.5.0

psybuzz:
Summary:
We’ve released tensorboard-data-server==0.5.0. This patch updates the
dependency from TensorBoard to the data server.

Test Plan:
wheel into a new virtualenv. Note that the new data server is installed
and that --load_fast with Custom Scalars, Mesh dashboards works as
desired.


wchargin (approved):
Works on my machine.

PR #4811, opened 2021-03-22
cli: add `--extra_data_server_flags`

wchargin:
Summary:
The main `tensorboard(1)` now proxies any `--extra_data_server_flags` to
the data server binary invoked as a subprocess. This would have been
useful as a less invasive workaround for #4801, and may serve as a
similar escape hatch in the future.

The argument to `--extra_data_server_flags` undergoes POSIX word
splitting, so you can pass multiple arguments by separating them by
spaces, quoting appropriately. Thus:

```
tensorboard --extra_data_server_flags="--host 'localhost or something'"
```

You can’t override flags that `tensorboard(1)` already passes, though,
because the data server will complain about conflicting arguments.
Thus, this interface is marked as unstable and is not intended to ever
be stabilized.

Test Plan:
Unit tests suffice.

wchargin-branch: cli-extra-data-server-flags

### DIFF
--- a/tensorboard/data/server_ingester.py
+++ b/tensorboard/data/server_ingester.py
@@ -68,6 +68,7 @@ class SubprocessServerDataIngester(ingester.DataIngester):
         reload_interval,
         channel_creds_type,
         samples_per_plugin=None,
+        extra_flags=None,
     ):
         """Initializes an ingester with the given configuration.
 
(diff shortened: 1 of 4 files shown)


psybuzz (approved):
Looks good for debugging purposes

PR #4813, opened 2021-03-22
cli: enable fast Rust data loading by default

wchargin:
Summary:
Running `tensorboard` now defaults to the fast data loading codepath
when available. This *does not* affect systems where RustBoard is not
installed, including on platforms where it’s unavailable. It also does
not affect invocations in which RustBoard detects that it is not
suitable for the job, such as an unsupported remote filesystem or
credential set. In such cases, TensorBoard will automatically fall back
to the legacy multiplexer paths. This selection is silent unless
`--verbosity 0` is given.

Because of this automatic fallback detection, this change is expected to
be “safe”. There may of course be undiscovered bugs in RustBoard, in
which case users have `--load_fast=false` as an escape hatch.

Users can also set `--load_fast=true` to bypass the selection logic,
suppress the advisory note that new loading logic is being used, and
hard-fail if RustBoard is not available.

Test Plan:
`bazel run //tensorboard` loads fast.

wchargin-branch: load-fast-auto

### DIFF
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -394,9 +394,9 @@ port to TensorBoard as a subprocess.(default: %(default)s).\
             help="""\
 Use alternate mechanism to load data. Typically 100x faster or more, but only
 available on some platforms and invocations. Defaults to "auto" to use this new
-mode only if only if available, otherwise falling back to the legacy loading
-path. Set to "true" to suppress the advisory note and hard-fail if the fast
-codepath is not available. Set to "false" to always fall back. Feedback/issues:
+mode only if available, otherwise falling back to the legacy loading path. Set
+to "true" to suppress the advisory note and hard-fail if the fast codepath is
+not available. Set to "false" to always fall back. Feedback/issues:
 https://github.com/tensorflow/tensorboard/issues/4784
 (default: %(default)s)
 """,
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -392,11 +392,13 @@ port to TensorBoard as a subprocess.(default: %(default)s).\
             default="auto",
             choices=["false", "auto", "true"],
             help="""\
-Use alternate mechanism to load data. Typically 100x faster or more. Set to
-"auto" to use this new mode only if only if installed and supported for this
-invocation. Set to "true" to suppress the advisory note and hard-fail if the
-fast codepath is not available. Set to "false" to always fall back to legacy
-loading. (default: %(default)s)
+Use alternate mechanism to load data. Typically 100x faster or more, but only
+available on some platforms and invocations. Defaults to "auto" to use this new
+mode only if only if available, otherwise falling back to the legacy loading
+path. Set to "true" to suppress the advisory note and hard-fail if the fast
+codepath is not available. Set to "false" to always fall back. Feedback/issues:
+https://github.com/tensorflow/tensorboard/issues/4784
+(default: %(default)s)
 """,
         )
 
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -389,11 +389,14 @@ port to TensorBoard as a subprocess.(default: %(default)s).\
         parser.add_argument(
             "--load_fast",
             type=str,
-            default="false",
+            default="auto",
             choices=["false", "auto", "true"],
             help="""\
-Experimental. Use a data server to accelerate loading. Set to "auto" to use a
-data server only if installed and supported for this invocation.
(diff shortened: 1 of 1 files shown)


wchargin:
Please DO NOT SUBMIT until the following are merged:

  - #4812: pulls in a data server version that works more seamlessly in
    Docker (v0.5.0 will always fall back to multiplexer)

  - #4809: cosmetic fix, but important

tensorboard/plugins/core/core_plugin.py
```diff
@@ -389,11 +389,14 @@ def define_flags(self, parser):
         parser.add_argument(
             "--load_fast",
             type=str,
-            default="false",
+            default="auto",
             choices=["false", "auto", "true"],
             help="""\
-Experimental. Use a data server to accelerate loading. Set to "auto" to use a
-data server only if installed and supported for this invocation.
+Use alternate mechanism to load data. Typically 100x faster or more. Set to
```
nfelt, line 395: Wording: right now this is worded in a way that suggests to me as though simply passing the flag enables the alternate mechanism (like `store_true`), which is a bit confusing given that it's actually taking a value, and the default value is in fact `auto` rather than on or off.

Would recommend something more like

> Whether to use an accelerated mechanism to load data. Typically 100x faster or more, but only available on some platforms and for some invocations. Defaults to "auto" to use this new mode only if available (generally, if `tensorboard-data-server` is installed) and if supported for this invocation, otherwise falls back to the legacy loading path. Set to "true" to suppress the advisory note and hard-fail if the fast codepath is not available. Set to "false" to always fall back.

Also, maybe worth mentioning the GH issue for feedback here as well.

nfelt (approved):

tensorboard/plugins/core/core_plugin.py
wchargin: Sounds good; thanks. Dropped “if `tensorboard-data-server` is installed”
to avoid confusion because the Python package is always installed (it
just might have `server_binary() is None`).

It’s also helpful that `argparse` shows the set of valid choices:

```
  --load_fast {false,auto,true}
                        Use alternate mechanism to load data. Typically 100x
                        faster or more, but only available on some platforms
                        and invocations. Defaults to "auto" to use this new
                        mode only if only if available, otherwise falling back
                        to the legacy loading path. Set to "true" to suppress
                        the advisory note and hard-fail if the fast codepath
                        is not available. Set to "false" to always fall back.
                        Feedback/issues:
                        https://github.com/tensorflow/tensorboard/issues/4784
                        (default: auto)
```

…so I think that the tri-state status is probably sufficiently clear.

tensorboard/plugins/core/core_plugin.py
```diff
@@ -389,11 +389,16 @@ def define_flags(self, parser):
...
             type=str,
-            default="false",
+            default="auto",
             choices=["false", "auto", "true"],
             help="""\
-Experimental. Use a data server to accelerate loading. Set to "auto" to use a
-data server only if installed and supported for this invocation.
+Use alternate mechanism to load data. Typically 100x faster or more, but only
+available on some platforms and invocations. Defaults to "auto" to use this new
+mode only if only if available, otherwise falling back to the legacy loading
```
nfelt, line 397: `only if` repeated twice here

tensorboard/plugins/core/core_plugin.py
nfelt: Oh right... hm. It would be nice if there was something slightly more concrete we could say here about when the fast path will be taken, since it feels fairly vague right now, but I guess now that we've outsourced that decision to the pip package or to the included server binary itself, we can't really do that without this potentially becoming stale.

tensorboard/plugins/core/core_plugin.py
wchargin: Oops; thanks. Fixed.

tensorboard/plugins/core/core_plugin.py
wchargin: Agreed; I have wondered whether it will be confusing that `auto`
sometimes uses the fast path and sometimes doesn’t (sounds silly to put
it that way, since that’s its purpose, but I think the point stands).
I contemplated adding a `WARNING: Can't use fast path because`... when
`--load_fast=auto` is given (maybe only when given explictly?), but
worried that it would be too spammy for people who always use
incompatible features like `--logdir_spec` (deprecated) or unsupported
filesystems. Maybe it’s worth it? Open to suggestions.

nfelt: No particular suggestions.  I agree it would be easy for that to be spammy.  To some extent I think there's just a tension here between wanting a clean separation of concerns and wanting to be able to present the user with an enumeration of what is/isn't supported.  Coming down on the side of the former is probably fine and if it's a usage impediment we can revisit.

PR #4814, opened 2021-03-22
chore: upgrade threejs to 0.125.0

psybuzz:
Upgrading to the newer Three.js 0.125.0 to silence dependabot warnings.
Notable changes:
- `Material`'s `vertexColors` is now a boolean
- `BufferAttribute.setArray` was removed in favor of
   `BufferedGeometry.setAttribute`
- Words like `texture` became reserved keywords when used in Three.js'
   shaders

Manually checked that the mesh demo, projector demo, and some custom
projector example with sprite images look as expected. I did not test the
"polylines that connect multiple points in a dataset" feature in the Projector
since I do not know how it works. Line charts in the Time Series look normal
as well.

See the Three.js changelog: https://github.com/mrdoob/three.js/wiki/Migration-Guide
Will be rendered obsolete: https://github.com/tensorflow/tensorboard/pull/4715
Googlers, see b/150083780, test sync cl/364427588, and copybara
cl/365095009.

image

### DIFF
--- a/tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
+++ b/tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
@@ -437,10 +437,23 @@ export class MeshViewer extends THREE.EventDispatcher {
     }) as any;
     const geometry = new THREE.BufferGeometry();
     // Flattens the [N, 3] shaped `vertices` into a N*3 length array.
+    // THREE.BufferAttribute expects values to be a shallow list, with the
+    // itemSize specified.
     const flatPoints = new Float32Array(vertices.flat());
     geometry.setAttribute('position', new THREE.BufferAttribute(flatPoints, 3));
 
-    // Flattens the [N, 3] shaped `faces` into a N*3 length array of indices.
+    /**
+     * Flattens the [N, 3] shaped `faces` into a N*3 length array of indices.
+     * This array of integers corresponds to triplets in the 'position'
+     * attribute. For example,
+     *   flatPoints = new Float32Array([7.5, 0, 0, 8.5, 0, 0, 9.5, 0, 0]);
+     *   vertexIndicesPerFace = new Uint16Array([2, 0, 1]);
+     *
+     * corresponds to 1 face with vertices
+     *   vertex1 at (9.5, 0, 0)
+     *   vertex2 at (7.5, 0, 0)
+     *   vertex3 at (8.5, 0, 0)
+     */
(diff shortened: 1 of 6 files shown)


tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
```diff
@@ -441,46 +435,37 @@ export class MeshViewer extends THREE.EventDispatcher {
...
-            vertex_color[2] / 255
-          );
-          face.vertexColors.push(color);
-        }
-      }
-      geometry.faces.push(face);
-    });
+    const geometry = new THREE.BufferGeometry();
+    // Flattens the [N, 3] shaped `vertices` into a N*3 length array.
+    const flatPoints = new Float32Array(vertices.flat());
```
japie1235813, line 440: Could you add more explanation why we flattens here?
Not sure this is equivalent to L445-452.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
```diff
@@ -441,46 +435,37 @@ export class MeshViewer extends THREE.EventDispatcher {
...
+    geometry.setAttribute('position', new THREE.BufferAttribute(flatPoints, 3));
+
+    // Flattens the [N, 3] shaped `faces` into a N*3 length array of indices.
+    const vertexIndicesPerFace = new Uint16Array(faces.flat());
```
japie1235813, line 444: Why "face" is an integer? (while points are float) What is the meaning of it?

japie1235813 (approved):

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
psybuzz: Updated the comment to explain that THREE.BufferAttribute expects a flat array.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
psybuzz: Updated the comment.

This array contains indices which correspond to "items" in the 'positions' attribute.  Added an example in the comment to illustrate how each "item" is a triplet of values (XYZ).

Most of my understanding was gained through reading this helpful site, and debugging
https://threejsfundamentals.org/threejs/lessons/threejs-custom-buffergeometry.html

psybuzz (commented):
Thanks for reviewing!

PR #4815, opened 2021-03-23
timeseries: customizable settings w/ DI

stephanwlee:
This change introduces DI for the time series' redux state to allow
different application to provide the default settings. Do note that this
pattern CANNOT be used for a route aware states.

Also do note that this change introduces some infrastructure for
integrational style testing using Jasmine to better test the behavior
provided by the metrics_module which otherwise is not really tested.

Extraneous looking changes are due to the integration test which,
for instance, does not provide certain values from the store (due to a
lacking experimentId from the route) and break certain assumptions.


psybuzz (approved):
Neat patterns!

I didn't know about `TestBed.overrideProvider` before, will keep it in mind.

Issue #4732, opened 2021-03-04
graph: "Ungroup this series of nodes" button is broken


psybuzz:
Repro:
- Open the Graph dashboard
- Load a graph that contains a collapsible series of nodes, e.g. [series_graph.txt](https://github.com/tensorflow/tensorboard/files/6086986/series_graph.txt)
- Select the series group, and click "Ungroup this series of nodes" button in the panel
![image](https://user-images.githubusercontent.com/2322480/110039747-9051a180-7cf6-11eb-9a10-5233732784aa.png)

Expected: it should expand the series group
![image](https://user-images.githubusercontent.com/2322480/110039898-c68f2100-7cf6-11eb-86ba-44580560f278.png)

Actual: Nothing happens, and the console shows an error
![image](https://user-images.githubusercontent.com/2322480/110039884-be36e600-7cf6-11eb-8727-93ae6cabc914.png)

Note that double-clicking the grouped series node to expand it (different from ungrouping it) still works.

psybuzz:
I manually checked with tensorboard==2.2.2 (broken), and tensorboard==1.14 (good).

PR #4817, opened 2021-03-24
graph: fix 'ungroup this series' button

psybuzz:
TensorBoard detects series of nodes (e.g. 'add_1', 'add_2', 'add_3', ...)
and visually collapses them into a single "series node". Currently, the
"ungroup this series" button in the right, floating info pane is broken.
Vestiges indicate that at one time, the button would modify the
`HierarchyParams.seriesMap` object, and rebuild a new `Hierarchy`
using the modified `HierarchyParams`.

This change properly wires up the handlers, but instead of modifying
the `seriesMap` in place, it creates a fresh new `seriesMap` and new
`HierarchyParams` for the new `Hierarchy` to use. This avoids a bug
where ungrouping a series then selecting a different graph to view
would cause the stale `seriesMap` from the previous graph to carry
over.

Fixes https://github.com/tensorflow/tensorboard/issues/4732

To test manually, feel free to check out this sample graph with a series:
https://github.com/tensorflow/tensorboard/files/6086986/series_graph.txt

Added a unit test and manually checked that the button now works.
image

### DIFF
--- a/tensorboard/plugins/graph/tf_graph_common/common_test.ts
+++ b/tensorboard/plugins/graph/tf_graph_common/common_test.ts
@@ -13,6 +13,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import * as tb_debug from '../../../components/tb_debug';
+import * as tf_graph from './graph';
+import * as tf_graph_hierarchy from './hierarchy';
 import * as tf_graph_loader from './loader';
 import * as tf_graph_parser from './parser';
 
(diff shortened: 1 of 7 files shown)


tensorboard/plugins/graph/tf_graph_common/common_test.ts
```diff
@@ -127,4 +129,68 @@ describe('graph tests', () => {
       },
     ]);
   });
+
+  describe('Graph Hierarchy', () => {
+    it('does not mutate the provided seriesMap by reference', () => {
+      const readonlySeriesMap = new Map([
+        ['fooNode', tf_graph.SeriesGroupingType.UNGROUP],
+        ['barNode', tf_graph.SeriesGroupingType.GROUP],
+      ]);
```
stephanwlee, line 138: idea: another way to have this test would be to `Object.freeze` it.

tensorboard/plugins/graph/tf_graph_common/hierarchy.ts
```diff
@@ -578,10 +600,7 @@ export function joinAndAggregateStats(
     }
   });
 }
-export function getIncompatibleOps(
-  hierarchy: Hierarchy,
-  hierarchyParams: HierarchyParams
-) {
+export function getIncompatibleOps(hierarchy: Hierarchy) {
```
stephanwlee, line 603: would you mind specifying the return type? :D

tensorboard/plugins/graph/tf_graph_op_compat_card/tf-graph-op-compat-card.ts
```diff
@@ -219,16 +217,12 @@ class TfGraphOpCompatCard extends LegacyElementMixin(PolymerElement) {
       list.fire('iron-resize');
     }
   }
-  @computed('graphHierarchy', 'hierarchyParams')
+  @computed('graphHierarchy')
   get _incompatibleOpNodes(): object {
```
stephanwlee, line 221: if you type the `getIncompatibleOps`, you can better type the return types.

tensorboard/plugins/graph/tf_graph_common/hierarchy.ts
```diff
@@ -884,13 +900,20 @@ function groupSeries(
     // If the series contains less than the threshold number of nodes and
     // this series has not been adding to the series map, then set this
     // series to be shown ungrouped in the map.
-    if (nodeMemberNames.length < threshold && !(seriesNode.name in map)) {
-      map[seriesNode.name] = tf_graph.SeriesGroupingType.UNGROUP;
+    if (
+      nodeMemberNames.length < threshold &&
+      hierarchy.getSeriesGroupType(seriesNode.name) ===
+        tf_graph.SeriesGroupingType.GROUP
```
stephanwlee, line 906: Previous code reads `!(seriesNode.name in map)` and it makes me think that this code is basically checking to see if the `hierarchy.getSeriesGroupType(seriesNode.name)` has the default value in case it does not exist. I think it is more prudent to `null` to `Map.prototype.has` than matching against the default value defined in various places in this module.

Would it not read better if the `hierarchy.getSeriesGroupType(seriesNode.name)` were to return `null`?

tensorboard/plugins/graph/tf_graph_op_compat_card/tf-graph-op-compat-card.ts
psybuzz: Good idea, done.

tensorboard/plugins/graph/tf_graph_common/hierarchy.ts
psybuzz: I would still prefer to hide the fact that the group types are stored implicitly, because it is a concern that is specific to `Hierarchy` internals.  External consumers of the `Hierarchy` actually do not need to know what `null` values mean, so exposing only Group/Ungroup keeps consumers less complex and easier to reason about, imo.

While the previous code does check whether a node has been explicitly grouped/ungrouped, the new code matches the same intent (if a grouped series is too short, then ungroup it), and does not require readers to understand anything about the default group type.

I've updated the comment above on L900 accordingly.

stephanwlee (approved):

PR #4828, opened 2021-03-31
plugin_util: factor out summary metadata version checker

wchargin:
Summary:
This abstracts over the metadata warnings for histograms and PR curves,
recently added in 5e220deb33ca and fea085ba5c5b, respectively. The
plugin code is now just a few lines. The new helper is in `plugin_util`,
but underscore-named to show that it’s meant for TensorBoard-internal
use only.

This functionality is now supported by the scalars, custom scalars,
images, audio, distributions (implicitly), histograms, text, PR curves,
mesh, and time series dashboards.

It’s not supported by the graphs dashboard, since graphs have a
non-standard summary metadata version format. It’s not supported in
hparams just because the backend is more complex with more moving parts,
so the fruit hangs not quite as low. It’s not supported in the debugger,
profile, What-If Tool, or projector dashboards because they don’t use
summaries.

Test Plan:
As is, data still renders as expected. After adding `version += 1` to
the top of `_MetadataVersionChecker.ok`, logs show:

```
plugin_util.py:219] Some scalar data is too new to be read by this version of TensorBoard. Upgrading TensorBoard may fix this. (sample: run 'hparams_demo/2/validation', tag 'epoch_accuracy', data version 1)
plugin_util.py:219] Some histogram data is too new to be read by this version of TensorBoard. Upgrading TensorBoard may fix this. (sample: run 'mnist/lr_1E-04,conv=2,fc=2', tag 'fc2/biases', data version 1)
plugin_util.py:219] Some PR curve data is too new to be read by this version of TensorBoard. Upgrading TensorBoard may fix this. (sample: run 'pr_curve_demo/mask_every_other_prediction', tag 'green/pr_curves', data version 1)
```

…and no data appears in the dashboards. Tested in all the dashboards
listed above.

wchargin-branch: plugin-common-version-warning

### DIFF
--- a/tensorboard/BUILD
+++ b/tensorboard/BUILD
@@ -592,6 +592,7 @@ py_library(
     deps = [
         ":context",
         "//tensorboard/backend:experiment_id",
+        "//tensorboard/util:tb_logging",
         "@org_mozilla_bleach",
         "@org_pythonhosted_markdown",
     ],
--- a/tensorboard/plugin_util.py
+++ b/tensorboard/plugin_util.py
@@ -25,7 +25,10 @@ import markdown
 
 from tensorboard import context as _context
 from tensorboard.backend import experiment_id as _experiment_id
+from tensorboard.util import tb_logging
 
+
+logger = tb_logging.get_logger()
 
 _ALLOWED_ATTRIBUTES = {
     "a": ["href", "title"],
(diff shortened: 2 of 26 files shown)


wchargin (commented):
@nfelt: Let me know once you’re satisfied with the API, and I can add
this to all the plugins. (Deferring just to reduce churn.)

nfelt:
LGTM in general on the API.  Not blocking, but looking at this I wonder if ultimately a nicer API could be obtained by centralizing the version tracking, so e.g. we add standard fields on `SummaryMetadata` or `PluginData` that have an integer "data version" plus a string for the minimum TensorBoard version that can read that data version.  Then we could handle all the warning / skipping logic in a centralized place versus per plugin.  That is slightly less flexible though in case we ever want the condition to be more complicated than a minimum TensorBoard version.

wchargin:
If you don’t mind, would appreciate a brief review for the time series
code, since it’s slightly more than a trivial copy-paste.

nfelt:
> If you don’t mind, would appreciate a brief review for the time series
> code, since it’s slightly more than a trivial copy-paste.

Hmmm, it's pretty convoluted needing to plumb `parse_metadata` and `version_checker` through all the layers of functions.  It looks like in the end, all it needs to do is filter out tags from the original `mapping` based on their metadata, so it seems much simpler if we just add a step to `_tags_impl()` in between retrieving the mappings and passing them to the `_format_X_mapping()` helpers that preprocesses the mappings to remove those tags?  It is an extra traversal of the overall mapping, but it also avoids doing the check twice (once for run to tag output, once for tag to description output) so it seems plausibly a wash in terms of overhead.

wchargin:
Oof, yes, thank you. Much better.

PR #4830, opened 2021-03-31
rust: use `std::array::IntoIter` where appropriate

wchargin:
Summary:
Follows up to discussion in #4739 by using the by-value iterator for
arrays, which was stabilized along with const generics in Rust 1.51.0.

Also, drive-by fix for a new syntax warning.

Test Plan:
Type-checking and unit tests suffice.

wchargin-branch: rust-array-intoiter

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,6 +145,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           toolchain: ${{ matrix.rust_version }}
+          default: true
       - name: 'Install Python packaging deps'
         run: |
           python -m pip install -U pip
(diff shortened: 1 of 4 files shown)


wchargin:
Huh, seems like a proper subset of build actions are installing the 1.50
toolchains instead of 1.51. Will investigate.

wchargin:
Ah, it’s because the GitHub Linux VMs have Rust 1.50.0 [installed by
default][1] (neat!) and we need to ask `rustup` to set the new
toolchains as the default versions. Will send out a separate fix.

[1]: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

nfelt (approved):
Thanks! 🦀

PR #4831, opened 2021-03-31
Update standalone_lib.html

ymodak:
Fixes https://github.com/tensorflow/tensorflow/issues/48173


lamberta (approved):

lamberta:
cc @wchargin

wchargin (commented):
Thanks for the patch. This won’t actually fix projector.tensorflow.org
as described in the original issue, because we don’t actively deploy to
that site (and haven’t in years). The right fix here is to solve the
underlying problem by adding or reinstating a redirect from the original
URL. See <http://cl/360984871> (Google-internal) for an example.

@ymodak, do you want to take care of that?

ymodak:
Thanks for the information. I will take a look at the example. Closing this PR for now.

PR #4833, opened 2021-04-01
infra: upgrade FE dependencies

stephanwlee:
This simply upgrades all the transitive dependencies using `yarn
upgrade` using the version spec specified in the package.json.

Do note that this change upgrade `y18n@4.0.0` (to 5.0.5) which is known
to have a security vulnerability, CVE-2020-7774.

### DIFF
--- a/tensorboard/webapp/core/effects/core_effects_test.ts
+++ b/tensorboard/webapp/core/effects/core_effects_test.ts
@@ -75,7 +75,7 @@ describe('core_effects', () => {
 
     fetchPolymerRunsSubjects = [];
     createElementSpy = spyOn(document, 'createElement');
-    createElementSpy.withArgs('tf-backend').and.returnValue({
+    createElementSpy.withArgs('tf-backend').and.returnValue(({
       tf_backend: {
         runsStore: {
           refresh() {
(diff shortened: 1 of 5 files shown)


stephanwlee:
Partially includes #4827.

japie1235813 (approved):

tensorboard/webapp/core/effects/core_effects_test.ts
```diff
@@ -85,7 +85,7 @@ describe('core_effects', () => {
           },
         },
       },
-    });
+    } as unknown) as HTMLElement);
```
japie1235813, line 88: What is the rational to cast the obj to unknown?

stephanwlee: ah, this spy is stubbing `createElement` which is supposed to return `HTMLElement` but we are returning an object here. An object is not an instance of a `HTMLElement` but we are asking TypeScript to forget the type of the object and coerce `unknown` to `HTMLElement`.

stephanwlee:
Thanks for already looking at additional type changes; was going to ping after CI passes. 

It looks like typings have changed with the upgrade but the failures actually looked more correct in my eyes than it was so I fixed them.

I will ping again when CI passes.

stephanwlee:
Got offline 🆗  to merge the PR. Thanks @japie1235813

PR #4834, opened 2021-04-05
rust: hold commit in `Arc` instead of leaking

wchargin:
Summary:
While watching @jonhoo’s latest [“Crust of Rust” video on atomics][jh],
I realized that we can just convert this irritating `&'static Commit`
leaked reference into a jointly owned `Arc<Commit>` while still meeting
the `T: 'static` bound for the RPC handler type. It’s not really a big
functional difference, because the commit would live for the whole
program anyway, but it’s nice to not have to worry about it. It should
also reduce the peak memory usage during tests, since we used to box up
and leak a new commit in each test case.

(In fact, this probably makes server shutdown a tad slower. If we really
wanted to recover that performance, we could make the server generic
over `T: Deref<Target = Commit>`, and pass `Box::leak(...)` in the main
program and an `Arc<Commit>` in tests. But that seems overkill; dropping
a few hash maps is probably not the end of the world.)

[jh]: https://youtu.be/rMGWeSjctlY?t=1181

wchargin-branch: rust-no-leak-commit

### DIFF
--- a/tensorboard/data/server/cli.rs
+++ b/tensorboard/data/server/cli.rs
@@ -214,17 +214,16 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
         eprintln!("listening on {:?}", bound);
     }
 
-    // Leak the commit object, since the Tonic server must have only 'static references. This only
-    // leaks the outer commit structure (of constant size), not the pointers to the actual data.
-    let commit: &'static Commit = Box::leak(Box::new(Commit::new()));
+    let commit = Arc::new(Commit::new());
     let psh_ref = Arc::new(opts.samples_per_plugin);
     thread::Builder::new()
         .name("Reloader".to_string())
         .spawn({
             let reload_strategy = opts.reload;
             let checksum = opts.checksum;
+            let commit = Arc::clone(&commit);
             move || {
-                let mut loader = LogdirLoader::new(commit, logdir, 0, psh_ref);
+                let mut loader = LogdirLoader::new(&commit, logdir, 0, psh_ref);
                 // Checksum only if `--checksum` given (i.e., off by default).
                 loader.checksum(checksum);
                 loop {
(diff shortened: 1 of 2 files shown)


psybuzz (approved):
Thanks for linking the video, it was helpful to watch.

> It should
also reduce the peak memory usage during tests, since we used to box up
and leak a new commit in each test case.

This is neat!

> In fact, this probably makes server shutdown a tad slower

I don't see why this would be, so I could use some help with understanding.  How is the leaked Box approach faster at shutdown?

wchargin:
> > In fact, this probably makes server shutdown a tad slower
> 
> I don't see why this would be, so I could use some help with understanding. How is the leaked Box approach faster at shutdown?

Sure.

Right before any function returns, its owned values are dropped. For
some data types (like primitives), this doesn’t actually do anything.
But for others, it does: heap-allocated types need to free memory,
reference-counted pointers need to decrement counts (and maybe free
memory if the count has hit zero), RAII lock guards need to release the
lock, etc.

So, when at the end of your function you hold an `Arc<Commit>` with no
other copies, returning from the function will cause the `Arc<_>` to be
dropped, which will cause the `Commit` to be dropped, which will entail
freeing memory for the commit’s heap-allocated hash maps and their
contents (the reservoirs, summary metadata, etc.). This is true for all
functions, including `main`.

The exact cost to free memory depends on your allocator, but it’s
generally not… well, free. And on top of the allocator cost, you’re at
minimum walking a potentially large structure and calling `drop` on all
its members.

By contrast, if you just hold a `&'static Commit`, only the reference is
dropped. Dropping a reference doesn’t do anything: it’s just a pointer,
whose underlying value is owned separately. So if you leak a `Commit`,
you never have to walk its contents and go through and free them.

After your program exits, the OS will reclaim all the memory anyway, so,
for types whose implementation of `Drop` is pure, just leaking them can
be a low-hanging optimization. This kind of thing matters enough that
it’s well-known and common for some kinds of tools, like compilers.
Some implementations of command-line tools like `sed` take an extreme
but simple approach of just not freeing any memory at all.

Some more discussion:
<https://softwareengineering.stackexchange.com/questions/305930/if-i-need-to-use-a-piece-of-memory-throughout-the-lifespan-of-my-program-is-it>

PR #4835, opened 2021-04-06
Add http context to request context.

### LINKED ISSUE #4820: Add ip address to RequestContext
Currently RequestContext holds AuthContext that can help data provider get the auth information from the end user. Similarly we need ability to pass IP address of the end user to the data provider that further forward that information to down stream services for security, quota, and fraud checks.

karthikv2k:
Fixes #4820 

Added a new filed to RequestContext for http attributes like ip address and x_forwarded_for header value.

* Screenshots of UI changes
NA

Added unit tests.
bazel test tensorboard:context_test
bazel test tensorboard:http_context_test

NA

### DIFF
--- a/tensorboard/webapp/notification_center/_redux/notification_notes.ts
+++ b/tensorboard/webapp/notification_center/_redux/notification_notes.ts
@@ -19,7 +19,25 @@ export const notificationNotes = [
     category: CategoryEnum.WHATS_NEW,
     dateInMs: 1579766400000,
     title: '2.4 release',
-    content:
-      '<li>Visualize Scalars, Images, and  Histograms in one place</li><li>Custom colors for runs</li><li>Group previews</li>',
+    content: `
+  # Hello there
+
+  ## This is the markdown with \`code\`.
+
+  \`this is a pretty long inline code block. It may wrap around without any special CSS. \`
+
+  helloworldhelloworldhelloworldhelloworldhelloworldhelloworldhelloworldhelloworldhelloworldhelloworld
+
+  \`\`\`
+  code test support hehe super long
+  \`\`\`
+
+  \`\`\`
+  thisisevenworseasthereisnowhitespacebetweenwordsright
+  \`\`\`
+
+  - Visualize Scalars, Images, and  Histograms in one place
+  - Custom colors for runs
+  - Group previews`,
   },
 ] as Notification[];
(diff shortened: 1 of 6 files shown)


tensorboard/http_context.py
```diff
@@ -0,0 +1,53 @@
...
+
+        The argument list is sorted and may be extended in the future;
+        therefore, callers must pass only named arguments to this
+        initializer.
+
+        Args: See "Fields" on class docstring. All arguments are optional and will
+        be replaced with default values if appropriate.
+        """
+        self._remote_ip = remote_ip
+        self._x_forwarded_for = x_forwarded_for
```
wchargin, line 41: Why introduce a new nested container instead of putting these on the
top-level context? It seems reasonable to me to put properties of the
HTTP request on the `RequestContext`, and it would be nice to avoid a
proliferation of types and modules if not necessary. It’s clear from the
docs and implementation that `HttpContext` and `RequestContext` have the
same structure, so it seems like an easy improvement to me.

tensorboard/http_context.py
```diff
@@ -0,0 +1,53 @@
...
+class HttpContext:
+    """Container of values related to caller's HTTP request.
+
+    This context is for providing security, and quota concerns.
+
+    `HttpContext` values are to be treated as immutable.
+
+    Fields:
+      remote_ip: Best possible guess of the caller's IP Address.
+      x_forwarded_for: Value of X-Forwarded-For header.
```
wchargin, line 27: Please document syntax: presumably these are strings representing either
IPv4 or IPv6 addresses? are they known to be in canonical form? could
they also be hostnames, or empty? are they permitted to be `None`? (It
may be that the answers are different for the two fields.)

tensorboard/http_context_test.py
```diff
@@ -0,0 +1,34 @@
...
+class HttpContextTest(tb_test.TestCase):
+    def test_args(self):
+        http = http_context.HttpContext(remote_ip="foo", x_forwarded_for="bar")
+        self.assertEqual(http.remote_ip, "foo")
+        self.assertEqual(http.x_forwarded_for, "bar")
+
+    def test_empty(self):
+        http = http_context.HttpContext()
+        self.assertEqual(http.remote_ip, None)
+        self.assertEqual(http.remote_ip, None)
```
wchargin, line 30: Dupe: want `x_forwarded_for` in one?

PR #4838, opened 2021-04-06
example: raw_scalars: use DataProvider

stephanwlee:
Raw Scalars example plugin was using PluginMultiplexer which is
deprecated and now throws RuntimeError.

Do note that the deprecation was not caught because the tests were run
manually, not in CI.

### DIFF
--- a/tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py
+++ b/tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py
@@ -122,23 +122,22 @@ class ExampleRawScalarsPlugin(base_plugin.TBPlugin):
           the data series.
 
         Raises:
-          errors.NotFoundError: if run+tag pair has no scalar data.
+          NotFoundError if there are no scalars data for provided `run` and
+          `tag`.
         """
-        try:
-            all_scalars = self._data_provider.read_scalars(
-                ctx,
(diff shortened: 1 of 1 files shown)


tensorboard/examples/plugins/example_raw_scalars/tensorboard_plugin_example_raw_scalars/plugin.py
```diff
@@ -122,15 +125,15 @@ def scalars_impl(self, tag, run):
...
-                for tensor_event in tensor_events
-            ]
+            all_scalars = self._data_provider.read_scalars(
+                ctx,
+                experiment_id=experiment,
+                plugin_name=metadata.PLUGIN_NAME,
+                downsample=5000,
+                run_tag_filter=provider.RunTagFilter(runs=[run], tags=[tag]),
+            )
+            scalars = all_scalars.get(run, {}).get(tag, None)
```
psybuzz, line 135: Now that these dict getters have default values, it seems the `KeyError` should never happen, but a `TypeError: 'NoneType' object is not iterable` might occur if it tries to iterate over `None`.  Can we update the docstring's `Raises: ...` and change the exception type?

psybuzz (approved):

psybuzz:
Thanks for taking on this upgrade!

wchargin (approved):
Thanks! There also exists `example_basic`, but okay with me to keep this
PR self-contained, because then it also serves as a nice upgrade guide.

PR #4839, opened 2021-04-07
plugin api: refactor to use Angularism

stephanwlee:
Currently the PluginApis are implemented a bit funnily; it uses
Polymer's RunsStore but exposes API to Angular via a Polymer component
hack.

This is not ideal for our long term code health, testability, and API
expandability. This change refactors to use NgModule, DI, and Angular
view patterns to better implement the host side of the APIs.


tensorboard/components/experimental/plugin_util/plugin_api_host_test.ts
```diff
@@ -0,0 +1,267 @@
...
+          );
+        runApi.init();
+        expect(broadcastSpy).toHaveBeenCalledWith(MessageId.RUNS_CHANGED, [
+          'hello',
+          'world',
+          'TensorBoard',
+        ]);
+      });
+
+      it('broadcasts only when runs changes', () => {
```
psybuzz, line 140: Ah, I see that the existing behavior was only checking run names, since the Polymer `runsStore` did not know about `ids`.  Now that we are hooking into the Ngrx store, I think it makes sense to use `id`s now that we can, as that is closer to the intent of the API.  e.g. if someone deletes a run from filesystem and creates a new one with the same name, an author probably wants to be notified.

Wdyt about changing `setOnRunsChanged` to emit whenever run `id`s change, but preserve `getRuns()` behavior of returning run names, in this PR?

psybuzz (commented):
overall looks good

psybuzz (approved):
Thanks!

PR #4843, opened 2021-04-08
fix(doc) : fix API demo of debugger V2

kevint324:
The demo code is not consistent with the API and
caller will get a "TypeError: enable_dump_debug_info()
got an unexpected keyword argument 'logdir'" error.


bmd3k (approved):
Great! I confirmed that `logdir` is no longer a keyword argument. Thanks for the fix!

PR #4846, opened 2021-04-08
plugin api: implement reload listener

stephanwlee:
This change introduces `tb_plugin_lib.experimental.core.setOnReload(()
=> {})` API. It gets invoked when TensorBoard finishes reloading the
data automatically and manually.

### DIFF
--- a/tensorboard/components/experimental/plugin_lib/core.ts
+++ b/tensorboard/components/experimental/plugin_lib/core.ts
@@ -12,7 +12,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-import {sendMessage} from './plugin-guest';
+import {listen, sendMessage} from './plugin-guest';
 
 /**
  * When called from a plugin with plugin_name `N`, it returns a promise which
@@ -22,4 +22,12 @@ import {sendMessage} from './plugin-guest';
  */
 export async function getURLPluginData() {
   return sendMessage('experimental.GetURLPluginData');
+}
+
+/**
+ * Listen to when data reloads on TensorBoard. There are two ways in which
+ * data reloads--(1) automatic refresh and (2) user clicks on a reload button.
+ */
+export function setOnReload(callback: () => void) {
+  listen('experimental.DataReloaded', callback);
 }
--- a/tensorboard/components/experimental/plugin_util/core-host-impl.ts
+++ b/tensorboard/components/experimental/plugin_util/core-host-impl.ts
@@ -17,10 +17,10 @@ limitations under the License.
  */
 import {Injectable} from '@angular/core';
 import {Store} from '@ngrx/store';
-import {distinctUntilChanged, filter, map} from 'rxjs/operators';
+import {distinctUntilChanged, filter} from 'rxjs/operators';
 
 import {State} from '../../../webapp/app_state';
-import {getLastLoadedTimeInMs} from '../../../webapp/selectors';
+import {getAppLastLoadedTimeInMs} from '../../../webapp/selectors';
 import * as tf_storage from '../../tf_storage';
 import {MessageId} from './message_types';
 import {Ipc} from './plugin-host-ipc';
@@ -52,7 +52,7 @@ export class PluginCoreApiHostImpl {
     });
 
     this.store
-      .select(getLastLoadedTimeInMs)
+      .select(getAppLastLoadedTimeInMs)
       .pipe(
         filter((lastLoadedTimeInMs) => lastLoadedTimeInMs !== null),
         distinctUntilChanged()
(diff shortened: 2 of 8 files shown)


tensorboard/components/experimental/plugin_lib/core.ts
```diff
@@ -23,3 +23,11 @@ import {sendMessage} from './plugin-guest';
 export async function getURLPluginData() {
   return sendMessage('experimental.GetURLPluginData');
 }
+
+/**
+ * Listen to when data reloads on TensorBoard. There are two ways in which
+ * data reloads--(1) automatic refresh and (2) user clicks on a reload button.
```
psybuzz, line 29: I wonder whether there is any change this might trigger on initial load as well.

After /plugins_listing succeeds, CoreEffects dispatches `pluginsListingLoaded`:
(A)
Something creates + registers the plugin's iframe.  In possibly another thread, the iframe starts listening on a MessageChannel.

(B)
Reducer receives `pluginsListingLoaded`, and updates the timestamp.  CoreHostImpl subscribes to changes in the timestamp, and broadcasts a message to any registered iframes.

Are we guaranteed that (A) will always happen after (B)?

tensorboard/components/experimental/plugin_util/core-host-impl.ts
```diff
@@ -43,5 +50,18 @@ export class PluginCoreApiHostImpl {
       }
       return result;
     });
+
+    this.store
+      .select(getPluginsListLoaded)
```
psybuzz, line 55: While `getPluginsListLoaded` changes happen to correspond exactly(?) to reload/manualReload actions today, these seem like fundamentally different actions.  e.g. in the future, the contract between updating the plugins list and reload events can change in another file, and the author may not know that this file also needs to change.

Could we add a `lastReloadTimeInMs` in the store?  With that, we could also guarantee that the callback will never fire on the initial load.

stephanwlee: > Could we add a lastReloadTimeInMs in the store? 

Sorry, I don't follow. It is already a state from the store?

tensorboard/components/experimental/plugin_lib/core.ts
stephanwlee: I must admit, I don't get the concern. Currently contract is that we will call the callback on data reload, not on initial load. I think you are talking about initial load and this callback getting invoked there? 

I expect plugins to read the initial data using a necessary API on its initial load. Kind of like how `setOnRunsChanged` vs. `getRuns` are two separate APIs while `setOnRunsChanged` already provides you with the list of runs.

psybuzz: > Currently contract is that we will call the callback on data reload, not on initial load.

That is a reasonable contract, but is that contract upheld in this PR?  Looking at https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/core/effects/core_effects.ts#L81, the initial `coreLoaded` will also trigger pluginsListingLoaded.

How do we know that the initial load does not cause this API to fire?

tensorboard/components/experimental/plugin_util/core-host-impl.ts
psybuzz: No, it does not exist currently.  I am suggesting that we should add this new field in the Core store's state along with the appropriate reducer and selector, so that core-host-impl can do `this.store.select(getLastReloadTimeInMs)`.

Less preferably, if we want plugin listing loads to be bound 1 to 1 to reloads, then we should explicitly add a comment in `webapp/core/actions/core_actions.ts` next to `pluginsListingRequested`, so that anyone who wants to use it understands that it must only ever be dispatched on coreLoaded, reload, manualReload.

tensorboard/components/experimental/plugin_lib/core.ts
stephanwlee: There is a difference between:
1. TensorBoard data loaded for the first time vs.
2. Plugin freshly stamped. 

The contract is that we will trigger the callback when data ever reloads which includes (1).

tensorboard/components/experimental/plugin_util/core-host-impl.ts
stephanwlee: Ah, I wasn't sure what was being communicated here. You are protesting against using the `PluginListing` as a proxy to figuring out whether the application load state. Yes, that's correct and that's something to be fixed eventually.

To move forward, since that refactor actually has larger impact than just scope of this PR (plugins_container is actually using that proxy state incorrectly), I added a selector that is to be later correctly implemented. Does that sound good?

psybuzz: Thanks, this sounds good to me.  From the CoreHostImpl's perspective, it is nicely unaware of the pluginsListing!

psybuzz (approved):

PR #4849, opened 2021-04-09
core: fix the header pagination

### LINKED ISSUE #4841: Plugin in header bar can be hidden with no chevron to scroll
Browser: Chrome on Chromebook
  Version 88.0.4324.186 (Official Build) (64-bit)

On TB.dev's example experiment, with a maximized chrome window:
https://tensorboard.dev/experiment/QFRIzZJpTZCNRzi8N7zomA/#scalars

The Text plugin does not appear in the plugin header list, and there are no scroll bars/chevrons.

If you reduce the font-size to 90% it pops out from behind the Send Feedback button (no chevrons are needed in that case).

stephanwlee:
TensorBoard customizes the material tab header component to have less
visual jank and have consistent behavior as the Polymer based
TensorBoard. This caused the scroll calculation to be incorrect (about
36 * 2 pixels) and a plugin like "text" where the tab label is very
short caused the bug to manifest quite obviously.

This change fixes that by changing how the mat-header is styled by
overriding few of its internals.

Fixes #4841.

### DIFF
--- a/tensorboard/webapp/header/plugin_selector_component.scss
+++ b/tensorboard/webapp/header/plugin_selector_component.scss
@@ -12,6 +12,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+@import 'tensorboard/webapp/theme/tb_theme';
+
 :host {
   align-items: center;
   display: flex;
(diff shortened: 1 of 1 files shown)


stephanwlee:
Gentle ping! :D

psybuzz (approved):
lgtm!  I'm assuming that all TB variants have a header color of `$tb-primary, 700`

stephanwlee:
> all TB variants have a header color of $tb-primary, 700

Currently, they are all set to "primary" at the Material level.

PR #4850, opened 2021-04-09
example_raw_scalars: include plugin API

stephanwlee:
TensorBoard has an experimental iframe API to get runs and get notified
when data refreshes. This change adds the iframe API library to the
example so we can better test and showcase the APIs.

Now that we need to include a built library, we now need to do some
JavaScript "building" requiring a bit more involved
`python setup.py develop` incantation. This change introduces bazel target,
`ibazel run tensorboard/examples/plugins/example_raw_scalars:develop`
that helps with the development of the example plugin.

Reviewer: this may violate the principle of the example; the example is no
longer portable, i.e., you cannot just copy the folder and expect to run the
plugin. While we lose the portability here, the basic example still has that
capability and I think it is nice to exemplify our plugin APIs in an example.
Please let me know if this PR is something to be avoided.

### DIFF
--- a/tensorboard/components/experimental/plugin_lib/BUILD
+++ b/tensorboard/components/experimental/plugin_lib/BUILD
@@ -1,8 +1,15 @@
-load("//tensorboard/defs:defs.bzl", "tf_ts_library")
+load("//tensorboard/defs:defs.bzl", "tf_js_binary", "tf_ts_library")
 
 package(default_visibility = ["//tensorboard:internal"])
 
 licenses(["notice"])
+
+tf_js_binary(
+    name = "plugin_lib_bin",
+    compile = True,
+    entry_point = "polymer-interop.ts",
+    deps = [":plugin_lib_polymer_interop_internal"],
+)
 
 # TODO(psybuzz): create a NPM package when a better requirement comes up using
 # tf_js_binary.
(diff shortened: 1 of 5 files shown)


psybuzz (commented):
Based on the intent of this PR, I suggest that we add instructions on how users can build, include, and call the plugin API in their own plugin, rather than introducing `develop.sh`.

What I like about this PR:
- The TensorBoard team has 1 bazel command to manually test it.
- Brings awareness to setOnReload and how to call it.

I think the main benefit of having an example_raw_scalars plugin is to provide some template that users can copy. If users run the example and see `setOnReload` working, I think the natural followup one asks is "let me try using it!". At that point, they are stuck because it is unclear how './lib' is built and included.

It would also be nice for users to be able to use the plugin lib without having to understand Bazel, genrules, where "org_tensorflow_tensorboard" comes from, "tf_js_binary", etc. The necessary concepts to use the plugin API are really:

- The entry point is `components/experimental/plugin_lib/index.ts`
- It needs to be bundled with one's favorite bundler (e.g. rollup, webpack, parcel, esbuild)
- API methods are namespaced, e.g. `setOnReload` is exposed on `lib.core.setOnReload`

I think for now, it might be best to just add documentation in the README explaining those 3 points, or even checking in a pre-built .js bundle for users.  Wdyt?

Issue #4844, opened 2021-04-08
[2.5] Failed to pick subchannel


n-gao:
### Diagnostics

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version e43767ef2b648d0d5d57c00f38ccbd38390e38da

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=9, micro=2, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): posix.uname_result(sysname='Linux', nodename='fs', release='5.4.0-42-generic', version='#46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020', machine='x86_64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: True
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
INFO: installed: tb-nightly==2.5.0a20210407
WARNING: no installation among: ['tensorflow', 'tensorflow-gpu', 'tf-nightly', 'tf-nightly-2.0-preview', 'tf-nightly-gpu', 'tf-nightly-gpu-2.0-preview']
INFO: installed: tf-estimator-nightly==2.5.0.dev2021032601
INFO: installed: tensorboard-data-server==0.6.0

--- check: tensorboard_python_version
INFO: tensorboard.version.VERSION: '2.5.0a20210407'

--- check: tensorflow_python_version
Traceback (most recent call last):
  File "/nfs/homedirs/gaoni/diagnose_tensorboard.py", line 522, in main
    suggestions.extend(check())
  File "/nfs/homedirs/gaoni/diagnose_tensorboard.py", line 75, in wrapper
    result = fn()
  File "/nfs/homedirs/gaoni/diagnose_tensorboard.py", line 278, in tensorflow_python_version
    import tensorflow as tf
ModuleNotFoundError: No module named 'tensorflow'

--- check: tensorboard_data_server_version
INFO: data server binary: '/nfs/homedirs/gaoni/miniconda3/envs/wfnet/lib/python3.9/site-packages/tensorboard_data_server/bin/server'
INFO: data server binary version: b'rustboard 0.6.0'

--- check: tensorboard_binary_path
INFO: which tensorboard: b'/nfs/homedirs/gaoni/miniconda3/envs/wfnet/bin/tensorboard\n'

--- check: addrinfos
socket.has_ipv6 = True
[...]

PR #4851, opened 2021-04-09
data: check liveness before blessing data server

wchargin:
Summary:
Issue #4844 shows that there are circumstances when communication
between TensorBoard and the local data server process cannot be
established. As of this patch, we send a trivial RPC to the data server
and wait for its response before committing to use the new loading
paths. This adds about 1–5ms total time to the happy path on my machine,
with a worst case penalty of 5s due to the timeout. If the server is not
reachable, we print a warning and fall back to the legacy paths.

Test Plan:

  - Test with normal working TensorBoard.

  - Simulate failed data server connection by changing the definition of
    `addr` to `"localhost:%d" % (port + 777)` on line 175. This should
    print the `UNAVAILABLE`/“failed to connect” message from #4844.

  - Simulate slow data server by adding the following to `cli.rs`:

    ```rust
    tokio::time::sleep(Duration::from_secs(3)).await;
    ```

    Add this right before the `Server::builder().(...).await?` call at
    the end of `main`: i.e., after we write the port file, but before we
    actually respond to requests. Note that TensorBoard still works with
    a 3-second delay, but that it actually delays printing the startup
    message for those 3 seconds as it determines which data provider to
    use.

  - Simulate extra-slow data provider as above but waiting 6 seconds,
    and note that TensorBoard prints a `DEADLINE_EXCEEDED` error, falls
    back to the legacy paths, and shows valid data.

wchargin-branch: data-liveness-check

### DIFF
--- a/tensorboard/data/BUILD
+++ b/tensorboard/data/BUILD
@@ -95,6 +95,7 @@ py_library(
         ":ingester",
         "//tensorboard:expect_grpc_installed",
         "//tensorboard:expect_pkg_resources_installed",
+        "//tensorboard/data/proto:protos_all_py_pb2",
         "//tensorboard/util:tb_logging",
     ],
 )
--- a/tensorboard/data/server_ingester.py
+++ b/tensorboard/data/server_ingester.py
@@ -26,6 +26,7 @@ import pkg_resources
 
 from tensorboard.data import grpc_provider
 from tensorboard.data import ingester
+from tensorboard.data.proto import data_provider_pb2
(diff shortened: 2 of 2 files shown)


wchargin:
If all looks good, I’d like to cherry-pick this into 2.5.0.

stephanwlee (approved):

stephanwlee:
Unusual merge: for sake of expediency for 2.5.0 release, I am merging it on behalf of wchargin.

PR #4856, opened 2021-04-10
time series: move the fit button to header

stephanwlee:
The scalar line chart in the time series has had the feature where we
automatically fit domain to the new data when it is unaltered. However,
because it was hidden inside the overflow menu and because there was no
clear visual indication when the chart will fit to the domain or not, it
was largely confusing. With the recent changes to make it more salient
with `title` and `disabled` on the button, we improved in this area but
was still not too obvious.

To fix this problem, we are now moving the control out of the overflow
menu into the chart header so there is a clearer visual difference
between domains overridden vs. not.

image

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -779,6 +779,56 @@ describe('scalar card', () => {
     expect(visibleRunIds).toEqual(['run3']);
   }));
 
+  describe('controls', () => {
+    const ByCss = {
+      FIT_TO_DOMAIN: By.css('[aria-label="Fit line chart domains to data"]'),
+    };
+
+    it('resets domain when user clicks on reset button', fakeAsync(() => {
+      const runToSeries = {
+        run1: [{wallTime: 100, value: 1, step: 333}],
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -779,13 +779,16 @@ describe('scalar card', () => {
...
   }));
 
-  describe('overflow menu', () => {
-    beforeEach(() => {
+  describe('controls', () => {
+    const ByCss = {
+      FIT_TO_DOMAIN: By.css('[aria-label="Fit line chart domains to data"]'),
+    };
+
+    it('resets domain when user clicks on reset button', fakeAsync(() => {
```
psybuzz, line 787: Just checking, since this test case doesn't do any `flush()`ing, can `fakeAsync` be dropped?

psybuzz (approved):
Thanks, a nice UX change indeed!

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
stephanwlee: It is used in the `createComponent` or one of these utils (I tried and it threw an error).

PR #4857, opened 2021-04-10
runs: prevent sort when filter button is clicked

stephanwlee:
Currently, the interactable elements are stacked on top of each other on
the runs table; sort -> filter button. Without the stopPropagation,
clicking on a filter menu button will trigger a sort which is not what
we want.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
@@ -2453,6 +2453,45 @@ describe('runs_table', () => {
           );
         });
       });
+
+      it('does not sort because you click on the filter menu button', () => {
+        store.overrideSelector(
+          getRunHparamFilterMap,
+          buildHparamFilterMap([
+            [
+              'foo',
+              buildDiscreteFilter({
+                possibleValues: ['faz', 'bar', 'baz'],
+              }),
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -2453,6 +2453,45 @@ describe('runs_table', () => {
...
+          ])
+        );
+        const fixture = createComponent(TEST_HPARAM_SPECS, TEST_METRIC_SPECS);
+        fixture.detectChanges();
+
+        const columnHeaders = fixture.nativeElement.querySelectorAll(
+          Selector.HEADER_COLUMN
+        );
+        columnHeaders[3].querySelector('button').click();
+        columnHeaders[4].querySelector('button').click();
```
japie1235813, line 2489: It looks like there are only 4 elements with [role="columnheader"] 
Did I miss anything?

stephanwlee: 0th: checkbox
1st: name
2nd: something I am forgetting (time?)
3rd: hparam
4th: metric.

(since I added the `getRunMetricFilterMap`).

japie1235813 (approved):

PR #4861, opened 2021-04-13
notification center: integrate markdown renderer

stephanwlee:
This change integrates the markdown renderer with the notification
center. Now the content of the notification can be any markdown string
as long as it does not contain certain dangerous elements like script
tag.

This change also makes few visual modification to the markdown renderer
for the optimal rendering in the small widthed screen.

image

### DIFF
--- a/tensorboard/webapp/notification_center/_views/notification_center_test.ts
+++ b/tensorboard/webapp/notification_center/_views/notification_center_test.ts
@@ -13,6 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {CommonModule} from '@angular/common';
+import {NO_ERRORS_SCHEMA} from '@angular/core';
 import {TestBed} from '@angular/core/testing';
 import {MatButtonModule} from '@angular/material/button';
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/notification_center/_views/notification_center_test.ts
```diff
@@ -48,6 +49,7 @@ describe('notification center', () => {
           initialState: [],
         }),
       ],
+      schemas: [NO_ERRORS_SCHEMA],
```
japie1235813, line 52: why `NO_ERRORS_SCHEMA` is needed?

japie1235813 (approved):

tensorboard/webapp/notification_center/_views/notification_center_test.ts
stephanwlee: Schema and NO_ERROR_SCHEMA specifically allows us to omit certain Angular components defined in the template. 

To be more specific, we do not absolutely need the real implementation of the `markdown-renderer` to test the `notification-center` so the real implementation is omitted.

Issue #4862, opened 2021-04-13
build: genrules that use a `py_binary` in `exec_tools` are broken by TF python toolchain


nfelt:
The problem:

1. We have genrules that use `py_binary` targets listed in `exec_tools`.
    - Motivating case: [`//tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin:generate_dtypes`](https://github.com/tensorflow/tensorboard/blob/1501f462ad61eba0a8b93910d3995e50fbaab784/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/BUILD#L91-L108)
2. Because they are listed in `exec_tools` these `py_binary` targets get built in exec mode, using a special exec build configuration (rather than the host mode, which is what happens if you pass them in `tools`, and rather than the target mode, which is what happens if you just build the `py_binary` directly).
    - Quoting @wchargin: "host" = the machine on which Bazel runs, "execution" = the machine on which Bazel actions execute, "target" = end-user architecture
    - See background note at bottom.
3. Due to our use of the TF Bazel workspace, we inherit a bunch of complex configs that affect the exec build configuration, and in particular, we inherit a special TF python toolchain that overrides the default Bazel autodetecting python toolchain.
4. The special TF python toolchain finds the full path to the `python` binary on the build machine, and then "bakes" this into the Python toolchain configuration, which results in the exec-mode build of the `py_binary` rule's wrapper script hardcoding the path to whatever `python` was in use when it was first built.
5. This build artifact has an environment dependency - most obviously on `PATH` - but unlike other build artifacts that rebuild whenever `PATH` changes (https://github.com/tensorflow/tensorboard/issues/3213), this build artifact doesn't get marked as stale if `PATH` changes (possibly because it's assumed to be an exec mode artifact and thus "independent" of the host machine, even though it's certainly not), and consequently seems to persist until you do `bazel clean --expunge` or possibly some other scenarios (in practice it seemed to occasionally get cleared for no apparent reason at all).
6. When `PATH` changes - for example, switching to a new virtualenv - everything else in our build will reflect the new `python` interpreter path (since it uses the Bazel auto-detecting python toolchain), **_except_** for these `py_binary` targets, which continue to be hardcoded to point to the old path.
7. At this point, not only is the build not hermetic, it's invisibly dependent on the prior build environment, and as soon as something breaks there (for example: `numpy` gets uninstalled) you have a breakage that is not reliably reproducible and generally maddening to debug.
[...]

PR #4864, opened 2021-04-13
build: patch TF python_configure to fix genrule toolchain issues

nfelt:
This is a workaround for #4862 - see that issue for all the gory details.

The tl;dr is that we want to fix an omission in the TF python configuration rules, where they only list a small number of environment variables that their [repository rules are sensitive to](https://docs.bazel.build/versions/master/skylark/lib/globals.html#repository_rule). In particular they don't include `PATH` (or `PYTHONPATH` or `PYTHONHOME`), even though these are all actually used by the toolchain resolution rules to find the active python binary and python lib directory.  So we just patch them in, which ameliorates #4862.  (In theory we could try to upstream this, but we're currently pinned to TF 2.3.0 rules and having to upgrade will probably be painful.)

Tested: ran repro script from #4862 and verified that after this change, while it still bakes the local python path into the py_binary wrapper script that's invoked by the genrule, it at least regenerates the wrapper when `PATH` changes (i.e. when switching virtualenvs) which mostly sidesteps the problem.  It would still feel more foolproof to never bake in that path at all and continue using Bazel's autodetecting python toolchain, but to do that I think we might need to migrate off of the TF workspace rules entirely.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -90,6 +90,13 @@ http_archive(
         "http://mirror.tensorflow.org/github.com/tensorflow/tensorflow/archive/v2.3.0.tar.gz",  # 2020-07-23
         "https://github.com/tensorflow/tensorflow/archive/v2.3.0.tar.gz",
     ],
+    patches = [
+        # Patch TF's python_configure.bzl to ensure it reconfigures its python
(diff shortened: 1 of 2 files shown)


wchargin (approved):

wchargin (approved):
No, I insist.

PR #4865, opened 2021-04-14
plugin api: cut tie to lodash and polymer

stephanwlee:
Core API requires dependency on tf_storage for reading the hash and that
module depends on lodash. Lodash, technically, is not an issue but it
seems to double bundle on the dev target of the internal application and
breaks the build.

Also, this change guts the bazel dependency of plugin_host to the
polymer binary (saving changes to the plugin_host code does not trigger
rebuild of Polymer anymore!)


psybuzz (approved):
Sorry for the delay, lgtm

PR #4866, opened 2021-04-14
Gracefully handle cases where App Routing state is not available.

bmd3k:
#4839 introduced a runtime dependency from plugins component on tensorboard app router. This breaks some internal versions of TensorBoard that have not yet integrated the app router.

Rather than unwind the dependency or rollback the change, we instead handle no-app-router case more gracefully. It means that there is some functionality of the plugins component that is not available to no-app-router tensorboards. I think this is ok given that the app router is becoming more-and-more critical for integrating core tensorboard functionality. I think it's ok to expect that tensorboards will need to integrate the app router if they want to unlock this (and other) functionality.


bmd3k:
This is superceded by #4869.

PR #4869, opened 2021-04-14
Import PluginApiHostModule at app level. Do not assume it exists in PluginsComponent.

bmd3k:
#4839 introduced a transitive runtime dependency from PluginsComponent on tensorboard app router and runs module. This breaks some internal versions of TensorBoard that have not yet integrated the app router or the runs module.

We relax the dependency of PluginsComponent on PluginApiHostModule, making it optional. PluginApiHostModule is no longer imported by PluginsModule, thus breaking the dependency completely for internal versions of TensorBoard. In the case of core OSS TensorBoard, we instead import PluginApiHostModule from the top-level AppModule.

This supercedes #4866.

### DIFF
--- a/tensorboard/webapp/plugins/plugins_module.ts
+++ b/tensorboard/webapp/plugins/plugins_module.ts
@@ -23,10 +23,6 @@ import {PluginRegistryModule} from './plugin_registry_module';
 @NgModule({
   declarations: [PluginsContainer, PluginsComponent],
   exports: [PluginsContainer],
-  imports: [
-    CoreModule,
-    CommonModule,
-    PluginRegistryModule,
-  ],
+  imports: [CoreModule, CommonModule, PluginRegistryModule],
 })
 export class PluginsModule {}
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/plugins/plugins_module.ts
```diff
@@ -19,16 +19,10 @@ import {PluginsContainer} from './plugins_container';
 import {PluginsComponent} from './plugins_component';
 import {CoreModule} from '../core/core_module';
 import {PluginRegistryModule} from './plugin_registry_module';
-import {PluginApiHostModule} from '../../components/experimental/plugin_util/plugin_api_host_module';
```
stephanwlee, line 22: Please remove plugin_util:plugin_host dependency in the BUILD file of this module.

stephanwlee (approved):
Ideally, the test should be updated or changed to include the `undefined` case but, for interest of time, it LGTM.

tensorboard/webapp/plugins/plugins_module.ts
bmd3k: The dependency still exists in plugins_component.ts.

PR #4873, opened 2021-04-15
Test new IFRAME plugin behavior for PluginsContainer

bmd3k:
#4869 fixed a bug under time pressure but did not include tests.

Add a test that ensures that when there is no PluginApiHostModule available to PluginsContainer then attempting to load IFRAME plugins will result in an error.

### DIFF
--- a/tensorboard/webapp/plugins/plugins_container_test.ts
+++ b/tensorboard/webapp/plugins/plugins_container_test.ts
@@ -151,7 +151,7 @@ describe('plugins_component', () => {
         PluginsContainer,
         PluginRegistryModule,
         getTestingProvider(),
-        ...providersOverride ? providersOverride : [],
+        ...(providersOverride ? providersOverride : []),
       ],
       declarations: [
         PluginsContainer,
--- a/tensorboard/webapp/plugins/plugins_container_test.ts
+++ b/tensorboard/webapp/plugins/plugins_container_test.ts
@@ -99,7 +99,6 @@ interface TbElement extends HTMLElement {
 describe('plugins_component', () => {
   let store: MockStore<State>;
   let createElementSpy: jasmine.Spy;
-  let pluginApiHost: PluginApiHostModule;
 
   const PLUGINS = {
     bar: {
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/plugins/plugins_container_test.ts
```diff
@@ -359,6 +345,46 @@ describe('plugins_component', () => {
...
+      expectPluginIframe(pluginElement, 'foo');
+      expect(registerPluginIframeSpy).toHaveBeenCalledWith(
+        pluginElement,
+        'foo'
+      );
+    });
+
+    it('throws error for IFRAME type of plugin when no PluginApiHostModule', async () => {
+      TestBed.overrideProvider(PluginApiHostModule, {useValue: null});
+      configureMockStore();
```
stephanwlee, line 376: Is there any importance to this order? If not, was `configureMockStore` refactor required?

bmd3k: Yes, order is important. I tried to document this effectively in configureMockStore() but also have now added a comment here.

tensorboard/webapp/plugins/plugins_container_test.ts
```diff
@@ -140,6 +139,28 @@ describe('plugins_component', () => {
     store.refreshState();
   }
 
+  /**
+   * Configures default behavior of the MockStore.
+   * Call this only after all providers have been configured as it will freeze
+   * the TestBed configuration.
```
stephanwlee, line 145: I find this refactor a bit unwieldy and I think it can be improved in two ways:
```ts
// 1. factor everything in beforeEach as `setup` method that does both TestBed configuration AND mocking the store with default values
function setup(providersOverride: someType[]|undefined) {
      providers: providersOverride ?? [
     ...
 });
 store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
  store.overrideSelector(getPlugins, PLUGINS);
  ...
}

// Method 2
describe('plugin_component with PluginApiModule', () > {
  // current test block... 
});

describe('plugin_container without PluginApiHostModule', () => {
  beforeEach(() => {
    // do not provide the PluginApiModule and see how entire component behave without it
    // currently, we are just testing one method so even if other places break, we would not know.
  });
});
```

WDYT?

bmd3k: I did method 1 with some inspiration from method 2 as well.

tensorboard/webapp/plugins/plugins_container_test.ts
```diff
@@ -231,6 +239,7 @@ describe('plugins_component', () => {
     });
 
     it('creates an element for IFRAME type of plugin', async () => {
+      const pluginApiHost = TestBed.inject(PluginApiHostModule);
```
stephanwlee, line 242: Hmm how does this actually work? I see no providers for this.

stephanwlee (approved):

tensorboard/webapp/plugins/plugins_container_test.ts
bmd3k: The beforeEach calls setup(). setup(), by default, injects the PluginApiHostModule. Only in the new test do we choose not to inject PluginApiHostModule.

stephanwlee: Ah, `getTestingProvider` includes the PluginApiHostModule. Make sense.

bmd3k: Ah, right, ya, unfortunately that's hidden by a layer of indirection.

PR #4875, opened 2021-04-15
Factor public uploader errors into their own module.

bileschi:
Precondition for factoring `*BatchedRequestSender` into their own modules, so as to simplify uploader logic.  That factoring, in turn, is a precondition for enabling faster asynchronous uploads.'

Moves public errors into their own module.

Tests still pass

There is also the internal _OutOfSpaceError.  It will be moved into the individual BatchedRequestSender.  There is no need to expose this error type or that is must be common across the BatchedRequestSenders

### DIFF
--- a/tensorboard/uploader/BUILD
+++ b/tensorboard/uploader/BUILD
@@ -101,6 +101,7 @@ py_library(
     deps = [
         ":logdir_loader",
         ":upload_tracker",
+        ":uploader_errors",
         ":util",
         "//tensorboard:expect_grpc_installed",
         "//tensorboard/backend:process_graph",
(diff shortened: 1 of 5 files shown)


bmd3k (approved):

wchargin:
This breaks internal code that relies on those symbols.
See presubmit failures for <http://cl/369217127>.

An acceptable solution to get this synced cleanly: export the symbols
from both modules, then update the internal code, then remove the
re-exports when possible. Thus:

```python
# uploader.py
ExperimentNotFoundError = uploader_errors.ExperimentNotFoundError
...
```

PR #4879, opened 2021-04-16
Explicitly list ngrx dependencies in PluginApiHostModule.

bmd3k:
TensorBoard frontend team is experimenting with a new policy to explicitly `import` ngrx-based dependencies in NgModules that otherwise don't need to be listed.

An example is code in one part of the codebase that relies on ngrx selectors in a different codebase. The corresponding NgModule for the calling code technically does not have to list the selector's corresponding NgModule as an import. However, we think it should be policy that the NgModule dependency is explicitly listed.

We first apply this experimental policy to PluginApiHostModule. Recent changes to that module to depend on selectors in other parts of the codebase, and the surprising runtime breakages it created in Google-internal TensorBoard services, are what provoked this discussion in the first place.


tensorboard/components/experimental/plugin_util/plugin_api_host_module.ts
```diff
@@ -14,13 +14,16 @@ limitations under the License.
...
+import {AppRoutingModule} from '../../../webapp/app_routing/app_routing_module';
+import {RunsModule} from '../../../webapp/runs/runs_module';
 
 import {PluginCoreApiHostImpl} from './core-host-impl';
 import {Ipc, registerPluginIframe} from './plugin-host-ipc';
 import {PluginRunsApiHostImpl} from './runs-host-impl';
 
 @NgModule({
   providers: [Ipc, PluginCoreApiHostImpl, PluginRunsApiHostImpl],
+  imports: [AppRoutingModule, RunsModule],
```
psybuzz, line 26: +CoreModule

PluginCoreApiHostImpl --> `getAppLastLoadedTimeInMs` selector from CoreModule.

psybuzz (approved):

stephanwlee (approved):
Assuming the CoreMoudle is included; it LGTM.

japie1235813 (approved):
Lgtm with CoreModule present

PR #4883, opened 2021-04-19
Revert "Factor public uploader errors into their own module. (#4875)"

bileschi:
This reverts commit 25a050f4a467322ffaccfa12213e1130c1cd830c.

Import errors due to breaking changes introduced by moving symbols without reexporting them


bmd3k (approved):

wchargin (approved):
(I can’t test-sync this right now because Reasons, but a clean revert
seems fine to me.)

PR #4887, opened 2021-04-20
core: remap /data/notifications to static path if present

nfelt:
This wires in logic that remaps the path `/data/notifications` at the server side to serve a static asset (found at the path `notifications_note.json`) if one is available, otherwise it falls back to the existing implementation of returning a JSON response containing no notifications.  Effectively this is similar to serving a redirect, but it avoids an extra server-client roundtrip.

Test plan: added a unit test, and tested end-to-end by patching in this commit and running TensorBoard locally: https://github.com/nfelt/tensorboard/commit/89597d613f73a727444fc49eab1dcd1e72375fc2


japie1235813 (approved):

tensorboard/plugins/core/core_plugin.py
```diff
@@ -90,10 +91,18 @@ def get_plugin_apps(self):
             "/histograms": self._redirect_to_index,
             "/images": self._redirect_to_index,
         }
-        apps.update(self.get_resource_apps())
+        apps.update(self._static_handlers)
```
japie1235813, line 94: why not just call the function here if `self._static_handlers` is not used in other places?

japie1235813 (approved):

tensorboard/plugins/core/core_plugin.py
```diff
@@ -90,10 +91,18 @@ def get_plugin_apps(self):
...
+        remappings = {
+            # Serve notifications from static JSON file if present.
+            "/data/notifications": "/notifications_note.json",
```
japie1235813, line 97: I tried to verify the change and 
`...com/data/notifications_note.json` redirect to `.com/data/notifications_note.json`
which is not the correct path (omit 'data/')

tensorboard/plugins/core/core_plugin.py
nfelt: Hmm good point, earlier I was going to need the property but now I don't.  Will inline.

tensorboard/plugins/core/core_plugin.py
nfelt: I don't quite follow, can you provide a little more detail about the exact way you were testing this and exactly what response you got?  I'm not sure I understand the `...com` versus `.com` part and also I wouldn't expect this remapping to affect an incoming path with `notifications_note.json` (since it only maps the path `/data/notifications`).

tensorboard/plugins/core/core_plugin.py
```diff
@@ -90,10 +91,18 @@ def get_plugin_apps(self):
...
+        }
+        for src, dest in remappings.items():
+            handler = apps.get(dest)
```
japie1235813, line 100: the handler is None

PR #4888, opened 2021-04-20
timeseries: preserve select all on page when new runs arrive

psybuzz:
Before, the auto-selection logic was:
- For the 1st load, auto-select only if runCount <= N
- On subsequent loads, new runs are always unselected

For users who frequently want new runs to be automatically
selected, we
- Increase N to 500
- Show a toast when the run set exceeds N, informing the user
  that new runs might not be selected.

image

Manually tested by running a script, pressing the in-app reload.

```python
import tensorflow as tf

for i in range(50):
  writer = tf.compat.v2.summary.create_file_writer('./logs' + str(i))
  writer.set_as_default()
  tf.summary.scalar("cool tag", 1, step=1)
```

Googlers, see b/184284696.


tensorboard/webapp/runs/store/runs_reducers.ts
```diff
@@ -98,26 +99,38 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
...
+     *
+     * Known caveat: subsequent reloads after the first may result
+     * in selecting a large number of runs automatically, incuring a
+     * performance cost.
+     */
+    let nextShouldAutoSelectRunsMap = state.shouldAutoSelectRuns;
+    const runCount = action.runsForAllExperiments.length;
+    if (runCount > 0) {
+      const eidsBasedKey = serializeExperimentIds(action.experimentIds);
+      let selectNewRunsByDefault =
```
stephanwlee, line 114: Didn't think that this variable would be a boolean type. Perhaps `shouldSelectNewRuns`?

tensorboard/webapp/runs/store/runs_reducers.ts
```diff
@@ -171,6 +185,10 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
     return {
       ...state,
       selectionState: nextSelectionState,
+      shouldAutoSelectRuns: new Map([
+        ...state.shouldAutoSelectRuns,
+        [stateKey, nextValue],
```
stephanwlee, line 190: Similar to this, shouldn't we set the `shouldAutoSelectRuns` to false when one is toggled off? 

If the goal is to be consistent with the Scalars plugin, then we should not select by default when at least one is toggled off by user.

tensorboard/webapp/runs/store/runs_reducers.ts
```diff
@@ -98,26 +99,38 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
...
-        MAX_NUM_RUNS_TO_ENABLE_BY_DEFAULT;
-      for (const run of action.runsForAllExperiments) {
-        selectionMap.set(run.id, runSelected);
+    /**
+     * Auto-select new runs based on the saved preference, or based on the
+     * number of runs.
+     *
+     * Known caveat: subsequent reloads after the first may result
+     * in selecting a large number of runs automatically, incuring a
+     * performance cost.
```
stephanwlee, line 108: Q: would it make sense to have some upper limit like 500 (random number off of my head)? 
OTOH, perhaps these rules are too random and may confuse users more than help them.

nfelt: $0.02: I think raising the limit significantly to N=100 or N=500 would probably make sense (especially for time series where we have the different query pattern, and the intersection observer, and the gpu line chart).

If we do keep a limit at all, it might be nice to make the behavior "select up to the limit, then activate a toast that explains not all runs were selected" rather than "past the limit, select 0 runs" like we have now.  That would also work a little better with the "new runs arrive" case IMO since we could just activate the toast again.  So e.g.

Initial page load:
> You have a lot of runs! Selected first N for display. [Dismiss]

Manual or auto-reload which loads new runs when N are already selected:
> New runs loaded, but not auto-selected since N are already selected. [Dismiss]

tensorboard/webapp/runs/views/runs_selector/runs_selector_test.ts
```diff
@@ -111,4 +118,104 @@ describe('runs selector test', () => {
...
+      store.overrideSelector(
+        getCurrentRouteRunSelection,
+        createRunSelectionMap(MAX_NUM_RUNS_TO_ENABLE_BY_DEFAULT + 1)
+      );
+      store.refreshState();
+
+      expect(snackBarOpenSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('does not show when there are too few runs', () => {
```
nfelt, line 151: This test is just the first ~10 lines of the previous test, right?  It seems fine to omit (maybe just reword the previous test to something like `it("activates when number of runs exceeds limit")` or something).

nfelt (approved):
Overall approach SGTM. Thanks for addressing this!

tensorboard/webapp/runs/views/runs_selector/runs_selector_container.ts
```diff
@@ -49,6 +71,51 @@ export class RunsSelectorContainer {
...
+    runsExceedsLimitForRoute$.subscribe(() => {
+      const text =
+        `The number of runs is over ` +
+        `${MAX_NUM_RUNS_TO_ENABLE_BY_DEFAULT}. New runs are unselected ` +
+        `for performance reasons.`;
+      this.snackBar.open(text, 'DISMISS', {
+        duration: 5000,
+        horizontalPosition: 'start',
+        verticalPosition: 'bottom',
+      });
```
stephanwlee, line 113: Can we undo this and use the alerts module? This is taking the imperative approach of a view instead of more declarative version: e.g., you are using business logic to determine whether to show a snackbar or not then are making an imperative API call to the snackbar. Instead, the view should react to the data change which we already have in the alerts component. Just add an alert in this case?

tensorboard/webapp/runs/store/runs_types.ts
```diff
@@ -68,10 +68,17 @@ export interface RunsDataState {
...
+  /**
+   * Map from a 'key' to a `RunId` to boolean whether the run is selected.
+   * The 'key' is a serialization of multiple experiment IDs.
+   *
+   * Run selection is tied to a list of experimentIds which is somewhat related
+   * to route but not strictly. For instance, if we want to render the
+   * run-selector in both experiment list and dashboard routes sharing the
+   * state, they need to share a key.
+   *
+   * TODO: this belongs in UI state not data state.
```
stephanwlee, line 80: Avoid a TODO without a name of a bug. Either put your name or just say "hack".

stephanwlee (changes requested):

tensorboard/webapp/runs/store/runs_types.ts
psybuzz: Done.

tensorboard/webapp/runs/views/runs_selector/runs_selector_container.ts
psybuzz: Done.

tensorboard/webapp/runs/views/runs_selector/runs_selector_test.ts
psybuzz: Done.

tensorboard/webapp/runs/views/runs_selector/runs_selector_container.ts
```diff
@@ -49,6 +71,44 @@ export class RunsSelectorContainer {
...
+        );
+      })
+    );
+    runsExceedsLimitForRoute$.subscribe(() => {
+      const text =
+        `The number of runs exceeds ` +
+        `${MAX_NUM_RUNS_TO_ENABLE_BY_DEFAULT}. New runs are unselected ` +
+        `for performance reasons.`;
+      this.store.dispatch(alertReported({localizedMessage: text}));
+    });
```
stephanwlee, line 107: this is effectively creating a composite action; please do this in the alerts feature.

stephanwlee (changes requested):

tensorboard/webapp/runs/views/runs_selector/runs_selector_container.ts
psybuzz: As per offline discussion, moved the composite action to RunsTable, and added a comment to indicate that this should be considered an exception, which goes against recommended patterns.

tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
```diff
@@ -353,9 +376,51 @@ export class RunsTableContainer implements OnInit {
...
+        })
+      );
+      runsExceedLimitForRoute$.subscribe(() => {
+        const text =
+          `The number of runs exceeds ` +
+          `${MAX_NUM_RUNS_TO_ENABLE_BY_DEFAULT}. New runs are unselected ` +
+          `for performance reasons.`;
+        this.store.dispatch(
+          alertActions.alertReported({localizedMessage: text})
+        );
```
stephanwlee, line 412: Just wondering, does any asynchronusity disrupt the UX? Adding what is effectively `setTimeout(0)` would make me feel tiny bit better about the synchronous composite action.

tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
```diff
@@ -353,9 +376,51 @@ export class RunsTableContainer implements OnInit {
...
+     * For consumers who show checkboxes, notify users that new runs may not be
+     * selected by default. Avoid showing it more than once per route, since it
+     * would be annoying to see the alert on every auto-reload (assuming a new
+     * run per reload).
+     *
+     * Warning: this pattern is not recommended in general. Dispatching
+     * `alertReported` would be better handled in a Ngrx Reducer in response
+     * to `fetchRunsSucceeded` or via the declared alert registrations using
+     * `alertFromAction`. Unfortunately, those currently have no way of knowing
+     * whether a run table is actually shown (with checkboxes), so we make a
```
stephanwlee, line 389: Just so we are 100% aligned; since the view is a representation of the state in the store, you _can_ reason with the state in the store to know. However, it is quite cumbersome and, now, you have to ingrain presentation logic into the store which is very unpleasant.

tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
```diff
@@ -353,9 +376,51 @@ export class RunsTableContainer implements OnInit {
...
+          `The number of runs exceeds ` +
+          `${MAX_NUM_RUNS_TO_ENABLE_BY_DEFAULT}. New runs are unselected ` +
+          `for performance reasons.`;
+        this.store.dispatch(
+          alertActions.alertReported({localizedMessage: text})
+        );
+      });
+    }
+
     this.store.dispatch(runTableShown({experimentIds: this.experimentIds}));
```
stephanwlee, line 416: Wow, when did we do this :(

stephanwlee (approved):

tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
psybuzz: Googlers, see original cl/330969935.  Back then, it was also noted that this is kind of an anti-pattern, but a tradeoff was made for convenience.  For the same reason as you mentioned, ingraining presentation logic into the store would be unpleasant, so this tradeoff still probably makes sense.

tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
psybuzz: I manually added it to try it out, and the UX seems fine, e.g. Alert still appears and the rest of the UI is functional.  Not sure how valuable it is to go add setTimeouts to all cases like these, but perhaps worth considering in a separate PR.

PR #4890, opened 2021-04-20
timeseries: fix fit-to-domain state issue

stephanwlee:
In a rare case, the ScalarCardComponent does not read the correct value
from the `newLineChart && newLineChart.getIsViewBoxOverridden()` causing
the fit-to-domain button to have the wrong state (enabled when it should
be disabled). This bug was identified in our screenshot based test where
this hard to reproduce timing issue was found.

Fixed the issue by manually propagating the view box overridden state
using the `Output()` which seems to have made the screenshot test very
stable (150 runs).

Relates to b/185595590.


stephanwlee:
Thanks for the quick review!

PR #4891, opened 2021-04-20
types: improve TbHttpClient types

stephanwlee:
Currently, we manually typed the bodies of `put` and `post`. Instead, we
want to extend that of the `HttpClient` so we now do that.

### DIFF
--- a/tensorboard/webapp/webapp_data_source/tb_http_client.ts
+++ b/tensorboard/webapp/webapp_data_source/tb_http_client.ts
@@ -85,7 +85,8 @@ export class TBHttpClient implements TBHttpClientInterface {
 
   post<ResponseType>(
     path: string,
-    body: Parameters<typeof HttpClient.prototype.post>[1],
+    // Angular's HttpClient is typed exactly this way.
+    body: any | null,
     options: PostOptions = {}
   ): Observable<ResponseType> {
     options = withXsrfHeader(options);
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/webapp_data_source/tb_http_client.ts
```diff
@@ -85,7 +85,7 @@ export class TBHttpClient implements TBHttpClientInterface {
 
   post<ResponseType>(
     path: string,
-    body: FormData,
+    body: Parameters<typeof HttpClient.prototype.post>[1],
```
bmd3k, line 88: Does this just evaluate to `any`?

From: https://angular.io/api/common/http/HttpClient
image

stephanwlee: Seems to be that way: https://github.com/angular/angular/blob/11.2.10/packages/common/http/src/client.ts#L2033-L2306

Intention, though, is, if they ever upgrade and better type the parameters, we will get them for free. Would you rather have us just type them as `any` instead?

bmd3k: Personally I would prefer if we were just explicit about the type and just put `any` or `any|null` here. The gain in adaptibility is not worth the readability and cognitive cost. `any` is a very easy type to understand. `Parameters<>` is not easy to understand and requires one to go read the  `Parameters` documentation and then the `HttpClient` documentation to reason it out.

If Angular ever upgrades the type in a way that matters, which I think is unlikely, it will be a stricter type and we will be alerted to it when our BUILD fails.

bmd3k (approved):

tensorboard/webapp/webapp_data_source/tb_http_client.ts
stephanwlee: Changed it to `any | null`.

bmd3k (approved):

PR #4900, opened 2021-04-22
timeseries: correctly fix fit-to-domain

stephanwlee:
In #4890, we incorrectly fixed the issue where scalar-card-component
does not update its DOM based on changes.

Background
----------------
while gory details are not yet known personally to me,
Angular, roughly speaking, has very strict condition in which it will
update the DOM. To name a few I know, they are (1) browser event, (2)
async (e.g., setTimeout), and (3) async pipe. In the earlier change, we
assumed that the `@Output` is one of them but it does not appear so.

Assumption
---------------
Angular updates DOM top to bottom (component tree wise) and
even if child component communicate information upwards when it mounts,
it _can_ be too late for parent component to actually apply the update
to the DOM.

Fix: we now use the AsyncPipe and Observable to communicate the
information upwards from a child component.

Why is this correct?
--------------------
AsyncPipe, on next value, interacts with the change detector[1]. Also,
this is a pattern that, sometimes, CDK components use[2]. Please see
public methods like `attachments` or `backdropClick`. At first, I was
curious why CDK did not use `Output` but now it starts to click a
little.

Test Case
---------
1. Launch TensorBoard and go to TimeSeries
2. Set the tag filter so only one scalar card matches the filter
3. Type a typo so nothing matches
4. Remove that single character typo
5. Repeat 3-4 to see the fit-to-domain turn solid (without this patch)

[1]: https://github.com/angular/angular/blob/master/packages/common/src/pipes/async_pipe.ts#L146
[2]: https://material.angular.io/cdk/overlay/api

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
@@ -23,10 +23,10 @@ limitations under the License.
   <span class="controls">
     <button
       mat-icon-button
-      [disabled]="!isViewBoxOverridden"
+      [disabled]="!lineChart || !(lineChart.getIsViewBoxOverridden() | async)"
       (click)="resetDomain()"
       [title]="
-          isViewBoxOverridden ?
+          !lineChart || !(lineChart.getIsViewBoxOverridden() | async) ?
               'Fit line chart domains to data' :
               'Line chart is already fitted to data. When data updates, the line chart '
                 + 'will auto fit to its domain.'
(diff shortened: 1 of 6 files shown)


psybuzz (approved):
Thanks for finding and fixing this.  For reference, the repro case only work before commit 9519153ed2501bd3d6ca1ee5bb9c95e1d6975520 (that commit changes the default state, so the symptom isn't visible).

It saddens me that I believed `@Output` would work with OnPush, as described in many articles like [1].  I'd like to better understand my misunderstanding there, in my own time.

[1] https://blog.angular-university.io/onpush-change-detection-how-it-works/

psybuzz:
After being bothered by this and doing some digging, I found that dropping the `ngIf=isEverVisible` check makes the symptom go away.  I think the root cause is that Angular doesn't trigger CD on event listeners when the component is made visible in the same task(?) as an IntersectionObserver callback.

Filed https://github.com/angular/angular/issues/41771

Issue #4901, opened 2021-04-22
Only Visualization of Loss is demonstrated in Tensorboard Scalars Documentation


worldpeaceaspirer:
## Issue description:
This is an issue to improve [documentation of Tensorboard Scalars](https://www.tensorflow.org/tensorboard/scalars_and_keras). In [this Tutorial](https://www.tensorflow.org/tensorboard/scalars_and_keras), only visualization of **`Loss`** is demonstrated but usually, we visualize **`Metrics`** as well (like `Accuracy`).

So, there can be a mention on how to visualize Metrics.

wchargin:
Hi @worldpeaceaspirer! The TensorBoard Keras callback used in those docs
will automatically log all the metrics that you’re tracking. So, if you
want to track accuracy, just add `metrics=["accuracy"]` to your model
creation code:

```python
model.compile(
    loss='mse',
    optimizer=keras.optimizers.SGD(lr=0.2),
    metrics=['accuracy'],
)
```

…and you’ll see `epoch_accuracy` appear in TensorBoard.

Does this answer your question?

worldpeaceaspirer:
@wchargin,
Thank you for your response and sorry for the delayed response. My point is, since the purpose of documentation is to guide the community, it will be better to cover all the Cases like **`Metrics`** and **`Loss`**, instead of just **`Loss`**, so that the documentation would be complete.

PR #4902, opened 2021-04-22
Update scalars_and_keras.ipynb

rmothukuru:
Added a statement to visualize Metrics, in Tensorboard Scalars Documentation

* Motivation for features / changes : Point raised in [this issue](https://github.com/tensorflow/tensorboard/issues/4901)

* Technical description of changes : Explanation to Visualizing Metrics was missing in the documentation.

* Screenshots of UI changes

* Detailed steps to verify changes work correctly (as executed by you) : Verified it in [my Branch](https://github.com/rmothukuru/tensorboard/blob/Tensorboard_Scalars_Doc/docs/scalars_and_keras.ipynb)

* Alternate designs / implementations considered : This is just a Documentation change

### DIFF
--- a/docs/scalars_and_keras.ipynb
+++ b/docs/scalars_and_keras.ipynb
@@ -319,6 +319,15 @@
     {
       "cell_type": "markdown",
       "metadata": {
+        "id": "bom4MdeewRKW"
+      },
+      "source": [
+        "**Important Note**: If you want to visualize `Metrics` (example, `Accuracy`) as well, you have to add `metrics` argument in `model.compile`."
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
         "id": "vvwGmJK9XWmh"
       },
       "source": [


wchargin (commented):
Instead of a loud disclaimer, could we just add `metrics=['accuracy']`
to our sample model above? It seems likely that users will already be
tracking metrics that they’re interested in, so no need to draw their
attention to an **Important Note**; we can just demonstrate the right
workflow.

rmothukuru:
> Instead of a loud disclaimer, could we just add `metrics=['accuracy']`
> to our sample model above? It seems likely that users will already be
> tracking metrics that they’re interested in, so no need to draw their
> attention to an **Important Note**; we can just demonstrate the right
> workflow.

@wchargin,
We can do that but then the Tensorboard Graph, which shows only Loss, has to be updated, and even [the Heading](https://www.tensorflow.org/tensorboard/scalars_and_keras#training_the_model_and_logging_loss), **`training_the_model_and_logging_loss`** needs to be changed. 

Please let me know how to proceed. Thanks!

psybuzz:
To chime in, I think it is reasonable to add `metrics=['accuracy']` and update the docs accordingly, with updated headings, instructions, and screenshots as needed.

PR #4906, opened 2021-04-23
infra: remove JSCompiler from Vulcanize

stephanwlee:
TensorBoard no longer uses JSCompiler in a meaningful way. For the most
TypeScript/JavaScript source, we use rollup via rules_nodejs. Hence, we
now remove the JSCompiler portion from the Vulcanization.

Soon after, vulcanize.bzl will be completely removed accordingly.

- [x] test sync: http://cl/370090328

### DIFF
--- a/tensorboard/defs/internal/html.bzl
+++ b/tensorboard/defs/internal/html.bzl
@@ -128,7 +128,13 @@ tb_combine_html = rule(
             """,
         ),
         "data": attr.label_list(allow_files = True),
-        "deps": attr.label_list(aspects = [closure_js_aspect], mandatory = True),
+        "deps": attr.label_list(
+            aspects = [closure_js_aspect],
+            mandatory = True,
+            doc = """Dependencies of `input_path` that provides `webfiles`.
+                Normally, they should be targets using `tf_web_library`s.
+            """,
+        ),
         "_Vulcanize": attr.label(
             default = Label("//tensorboard/java/org/tensorflow/tensorboard/vulcanize:Vulcanize"),
             executable = True,
--- a/tensorboard/defs/internal/html.bzl
+++ b/tensorboard/defs/internal/html.bzl
@@ -1,0 +1,133 @@
+# Copyright 2017 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


tensorboard/defs/internal/html.bzl
```diff
@@ -0,0 +1,133 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Rule for building the HTML binary."""
+
+load("@io_bazel_rules_closure//closure:defs.bzl", "closure_js_aspect")
+load("@io_bazel_rules_closure//closure/private:defs.bzl", "collect_js", "long_path", "unfurl")  # buildifier: disable=bzl-visibility
```
nfelt, line 18: `# buildifier: disable=bzl-visibility` lol

> KEEP OUT or don't I'm a sign not a cop

tensorboard/defs/internal/html.bzl
```diff
@@ -0,0 +1,133 @@
...
+        manifests = depset(transitive = [manifests, dep.webfiles.manifests])
+        webpaths = depset(transitive = [webpaths, dep.webfiles.webpaths])
+        files = depset(transitive = [files, dep.data_runfiles.files])
+    webpaths = depset([ctx.attr.output_path], transitive = [webpaths])
+    closure_js_library = collect_js(
+        unfurl(ctx.attr.deps, provider = "closure_js_library"),
+    )
+
+    # vulcanize
+    ignore_regexs_file_set = depset()
```
nfelt, line 42: This looks unused?

tensorboard/defs/internal/html.bzl
```diff
@@ -0,0 +1,133 @@
...
+            transitive_files = transitive_runfiles,
+        ),
+    )
+
+tb_combine_html = rule(
+    implementation = _tb_combine_html_impl,
+    attrs = {
+        "input_path": attr.string(mandatory = True),
+        "output_path": attr.string(mandatory = True),
+        # If specified, it extracts scripts into {name}.js and inserts <script src="{js_path}">.
```
nfelt, line 119: Could we use `doc` parameters for these attrs to provide a brief description of each one, just for posterity so we can better understand how this is used?  They can just be a single line.

One thing that I had to go read `Vulcanize.java` to remember is the distinction between `input_path` and `deps` - i.e. that `input_path` is where we start the crawl from, and `deps` has to contain all the webfiles that `input_path` is referencing.  Just brief mentions of things like that would help contextualize what's happening here.

nfelt (approved):
Thanks for doing this!

tensorboard/defs/internal/html.bzl
stephanwlee: Removed. Thanks.

tensorboard/defs/internal/html.bzl
stephanwlee: I gave it a stab. Thanks for the suggestion.

nfelt:
Thanks, LGTM!

PR #4907, opened 2021-04-23
core: fix the plugin selector color

stephanwlee:
To fix the wrong width calculation on the mat-tab-header, we use a weird
technique with opaque background. The color we picked wrongly set the
`700` causing a visual jank in an internal version.

Was able to confirm that omitting `700` suffices.

Example of the bad color:
image


japie1235813:
any screenshot for the visual deficient?

stephanwlee:
Added a screenshot of a violatoin.

japie1235813 (approved):

PR #4908, opened 2021-04-26
routing: improve compare navigation

stephanwlee:
When wanting to navigate to a compare route with a programmatical API,
we need to form the navigation parameter correctly, knowing its encoding
scheme. Not wanting to leak the business logic for encoding alias and
id, we now changed the interface for the compare route navigation with a
translation to an AppRouting Route in the AppRouting effects.


psybuzz (approved):

stephanwlee:
Thanks for the quick review.

PR #4909, opened 2021-04-26
Fix `--one_shot` for remote file system.

bileschi:
--one_shot uploading should work for non local filesystem

Googlers, this is internal bug b/186244973

### DIFF
--- a/tensorboard/uploader/uploader_subcommand.py
+++ b/tensorboard/uploader/uploader_subcommand.py
@@ -23,7 +23,7 @@ import textwrap
 
 from absl import logging
 import grpc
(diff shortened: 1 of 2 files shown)


tensorboard/uploader/uploader_subcommand.py
```diff
@@ -23,6 +23,7 @@
 
 from absl import logging
 import grpc
+from tensorboard.compat import tf
```
nfelt, line 26: Can you add a BUILD dep for `"//tensorboard/compat:tensorflow"`

nfelt (approved):

PR #4912, opened 2021-04-27
Ensure name and description are set in the exporter mask

ericdnielsen:
Ensure that both name and description are returned when using the tensorboard dev list command.

A recent refactoring that enforced a more consistent use of the fieldmask wasn't aware of this call site, and we had no tests.

### DIFF
--- a/tensorboard/uploader/uploader_subcommand_test.py
+++ b/tensorboard/uploader/uploader_subcommand_test.py
@@ -204,7 +204,7 @@ class UploadIntentTest(tf.test.TestCase):
             mock_stdout_write.call_args_list[-1][0][0],
         )
 
-    def testListIntendSetsExperimentMask(self):
+    def testListIntentSetsExperimentMask(self):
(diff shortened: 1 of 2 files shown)


tensorboard/uploader/uploader_subcommand_test.py
```diff
@@ -202,6 +204,30 @@ def testUploadIntentNonDryRunNonOneShotInterrupted(self):
             mock_stdout_write.call_args_list[-1][0][0],
         )
 
+    def testListIntendSetsExperimentMask(self):
```
bileschi, line 207: Should be ListIntent not ListIntend

bileschi (approved):
LGTM.  looks like you need to also run `black` to format.

tensorboard/uploader/uploader_subcommand_test.py
ericdnielsen: Done

PR #4915, opened 2021-04-28
refactor: use router event to fetch data.

stephanwlee:
Context: previously, we were using a composite action[1] to know when the
dashboard view updated and thus its data needs to be refreshed.

Fix: now that TensorBoard has a router and can determine what views are
rendered given a route, we can use a purer redux pattern to determine
whether the dashboard data has to be reloaded. Since not all versions of
the app uses the router, we will temporarily support the coreLoaded.

[1]: composite actioness: view was subscribing to the state changes to
determine to fire an action which in turn triggers another action.

Confirmed that this is a noop for internal app by running the screenshot tests
5 times each.


tensorboard/webapp/core/effects/core_effects.ts
```diff
@@ -70,28 +86,55 @@ export class CoreEffects {
...
+      distinctUntilChanged(([, beforeRouteId], [, afterRouteId]) => {
+        return beforeRouteId === afterRouteId;
+      })
+    )
+  ).pipe(
+    throttleTime(DATA_LOAD_COND_THROTTLE_IN_MS, undefined, {leading: true})
+  );
+
+  // Emits when data should be refreshed.
+  // HACK: currently, plugins list loaded state is used as a proxy to know
```
bmd3k, line 109: I think this comment could be a bit more explicit. Why is state === LOADING grounds for not updating the data?

tensorboard/webapp/core/effects/core_effects_test.ts
```diff
@@ -153,14 +167,19 @@ describe('core_effects', () => {
...
 
+        fetchPolymerRunsSubjects[0].next([{id: '1', name: 'Run 1'}]);
+        fetchPolymerRunsSubjects[0].complete();
+
         expect(recordedActions).toEqual([
           coreActions.pluginsListingRequested(),
           coreActions.environmentLoaded({
             environment: createEnvironment(),
           }),
+          coreActions.polymerRunsFetchRequested(),
```
bmd3k, line 178: Should we have a similar test but where polymerRunsFetchFailed and pluginsListingFetchSucceeded?

tensorboard/webapp/core/state.ts
```diff
@@ -0,0 +1,19 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {State as AppRoutingState} from '../app_routing/store/app_routing_types';
+import {State as FeatureFlagState} from '../feature_flag/store/feature_flag_types';
+import {State as CoreState} from './store/core_types';
+
+export type State = CoreState & AppRoutingState & FeatureFlagState;
```
bmd3k, line 19: Seems like we need further discussion with the team before merging this part of the PR.

bmd3k (commented):
I got through part of the PR.

tensorboard/webapp/core/effects/core_effects.ts
stephanwlee: Since I am not really changing the existing behavior and the behavior is codified in the test, I will be terse with the description.

tensorboard/webapp/core/state.ts
stephanwlee: For now, I took the easy route of re-exporting the symbol. I can undo all these changes but... a bit lazy :)

tensorboard/webapp/core/effects/core_effects.ts
```diff
@@ -70,28 +86,55 @@ export class CoreEffects {
...
+  private readonly onDataReload$ = merge(
+    this.onDashboardLoad$,
+    this.actions$.pipe(ofType(reload, manualReload))
+  ).pipe(
+    withLatestFrom(
+      this.store.select(getPluginsListLoaded),
+      this.store.select(getEnabledExperimentalPlugins)
+    ),
+    filter(([, {state}]) => state !== DataLoadState.LOADING),
+    share()
```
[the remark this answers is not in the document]
stephanwlee: When creating N observable streams out of one "common one" and, if one common one is expensive, you can use `share` or `shareReplay`. Btw,:
```ts
const expensiveStreamA$ = new BehaviorSubject<any>();
const b$ = expensiveStreamA$.pipe(...);
const c$ = expensiveStreamA$.pipe(...);

// without share on the streamA, expensiveStreamA$ will be repeated twice, once for each b$ and c$. In order to
// fork off of one stream, you need to use a `share`.
```

`share` was used while debugging something and we can do without it so I removed it.

tensorboard/webapp/core/store/core_selectors.ts
```diff
@@ -35,13 +36,59 @@ export const getPluginsListLoaded = createSelector(
...
   getPluginsListLoaded,
-  (loadedState: PluginsListLoadState): number | null =>
-    loadedState.lastLoadedTimeInMs
+  getPolymerRunsLoadState,
+  (
+    pluginListLoadState: PluginsListLoadState,
+    runsLoadState: LoadState
+  ): number | null => {
+    if (
+      pluginListLoadState.lastLoadedTimeInMs === null &&
```
bmd3k, line 81: Should this be || instead of &&? If some of the data is not yet loaded (as returned by getCoreDataLoadedState()), then I would intuitively expect that lastLoadedTime is null. I guess my interpretation of this selector is "returns the last time that all data was successfully loaded" but the implementation is "returns the last time some of the data was successfully loaded".

bmd3k (approved):

tensorboard/webapp/core/store/core_selectors.ts
stephanwlee: I can see it go both ways actually. 

Let's take the case when both were loaded once and we are reloading. In this case, Let us further assume that one of those two responses came back earlier than the other. To denote this in a pseudo-code,
```
// The case when, for some reason, only one is initialized to `t`.
State = {
  A = t
  B = null
}
// In above case, we return `t`.
// Something happened and now both A and B are being loaded.
State = {
  A: t // still loading 
  B: t+1 // B is refreshed and is now at `t+1`
}
// In above case, if we were to have the semantics of "last successful loaded time of all data", then I think we should return `t`
// but it is not that easy.
// Current selector returns `t+1`.
```

What do you think about the reloading case? Please advise.

stephanwlee: Addressed it by better codifying the expected behavior and updated the specs accordingly. Please take a look.

tensorboard/webapp/core/store/core_selectors.ts
```diff
@@ -35,13 +36,74 @@ export const getPluginsListLoaded = createSelector(
...
+    pluginListLoadState: PluginsListLoadState,
+    runsLoadState: LoadState
+  ): number | null => {
+    if (
+      pluginListLoadState.lastLoadedTimeInMs === null ||
+      runsLoadState.lastLoadedTimeInMs === null
+    ) {
+      return null;
+    }
+    if (pluginListLoadState.state === runsLoadState.state) {
```
bmd3k, line 95: There seem to be scenarios where the value returned by getAppLastLoadedTimeInMs can go backwards in time during a reload, which I find really confusing.

I think if you want to solve the problem accurately then it needs to be modeled in the state with a new `coreLoadState: LoadState` property and handled in the reducer layer.

Would this work?

```
on(pluginListFetchRequested):
  pluginListLoadState = {...pluginListLoadState, state: LOADING}
  coreLoadState = {...coreLoadState, state: LOADING}
on(polymerRunsFetchRequested):
  polymerRunsLoadState = {...polymerRunsLoadState, state: LOADING}
  coreLoadState = {...coreLoadState, state: LOADING}
on(pluginListFetchSuccess):
  pluginListLoadState = {state: LOADED, lastLoadTime: Date.now()}
  if (pluginLoadState.state === LOADED && polymerRunsLoadState.state === LOADED):
    coreLoadState = {state: LOADED, lastLoadTime: pluginListLoad.lastLoadTime}
on(polymerRunsFetchSuccess):
  polymerRunsLoadState = {state: LOADED, lastLoadTime: Date.now()}
  if (pluginLoadState.state === LOADED && polymerRunsLoadState.state === LOADED):
    coreLoadState = {state: LOADED, lastLoadTime: polymerRunsLoadState.lastLoadTime}
// TODO: Handle failed states
```

Then getCoreDataLoadedState and getAppLastLoadedTimeInMs become trivial.

stephanwlee: Thanks for catching the error in my logic. Adopted your change.

tensorboard/webapp/core/store/core_selectors.ts
```diff
@@ -35,13 +36,27 @@ export const getPluginsListLoaded = createSelector(
...
+export const getCoreDataLoadedState = createSelector(
+  selectCoreState,
+  (state: CoreState): DataLoadState => {
+    return state.coreDataLoadState.state;
+  }
+);
+/**
+ * Returns last _successful_ loaded time of the applicational state where an
+ * applicational state is defined as combinations of plugins listing, runs, and
+ * environment.
```
bmd3k, line 53: Is it worth clarifying that `environment` does not, in reality, impact CoreDataLoadedState? IIRC you previously modified the `TODO(tensorboard-team)` to clarify that.

bmd3k (approved):

tensorboard/webapp/core/store/core_selectors.ts
stephanwlee: Added the TODO back. `environment` does impact `CoreDataLoadedState`. If it fails to load, the effects will dispatch `fetchPluginsListingFailed` and it ultimately changes the `CoreDataLoadedState` to failed.

PR #4916, opened 2021-04-28
cleanup(line-chart): use un-deprecated API

stephanwlee:
In the new GPU line chart, we were using the `addAttribute` API of
three's which is now throwing a deprecation warning on the console when
testing. This change uses the new `setAttribute`.


psybuzz (approved):
Nice

PR #4918, opened 2021-04-28
build: add `polymer_bin` to tf_js_binary

stephanwlee:
For non-Polymer JavaScript, we can apply more compilation optimizations
we are missing out on. This change annotates the Polymer binary with a
property so we can apply a different build options.


tensorboard/defs/defs.bzl
```diff
@@ -32,11 +32,21 @@ def tf_js_binary(
...
-    Please refer to https://bazelbuild.github.io/rules_nodejs/Built-ins.html#rollup_bundle
-    for more details.
+    Args:
+        name: Name of the target.
+        compile: whether to compile when bundling.
+        deps: dependencies of the js_binary.
+        visibility: visibility of the target.
+        dev_mode_only: whether the binary is for development. When True, it will
+          omit the Terser.
+        includes_polymer: whether this binary is contains Polymer.
```
psybuzz, line 46: nit: "is contains" -> "contains"

tensorboard/defs/defs.bzl
```diff
@@ -32,11 +32,21 @@ def tf_js_binary(
...
         dev_mode_only = False,
+        includes_polymer = False,
         **kwargs):
     """Rules for creating a JavaScript bundle.
 
-    Please refer to https://bazelbuild.github.io/rules_nodejs/Built-ins.html#rollup_bundle
-    for more details.
+    Args:
+        name: Name of the target.
+        compile: whether to compile when bundling.
```
psybuzz, line 41: Maybe worth noting in these docs that `compile` and `includes_polymer` are only used internally.

psybuzz (commented):
Could you please explain or attach a link to something that explains what `includes_polymer` will be used for?

stephanwlee:
Relates to cl/371010264.

psybuzz (approved):
Thanks for the context

PR #4919, opened 2021-04-28
type: better type Polymer-Angular interops

stephanwlee:
Currently, the Polymer interop components like `tf-storage` are poorly
typed. While we try to type them better, they are not properly
typed/annotated that more advanced compiler wrongly mangles the symbol.

This change makes the interop components Klutz friendly.

### DIFF
--- a/tensorboard/webapp/core/effects/core_effects.ts
+++ b/tensorboard/webapp/core/effects/core_effects.ts
@@ -12,6 +12,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+import '../../tb_polymer_interop_types';
+
 import {Injectable} from '@angular/core';
 import {Store} from '@ngrx/store';
 import {Actions, ofType, createEffect} from '@ngrx/effects';
@@ -26,6 +28,7 @@ import {
   distinctUntilChanged,
   take,
 } from 'rxjs/operators';
+
 import {
   coreLoaded,
   environmentLoaded,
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/core/effects/core_effects.ts
```diff
@@ -54,7 +54,7 @@ export class CoreEffects {
   // Ngrx assumes all Effect classes have properties that inherit from the base
   // JS Object. `tf_backend` does not, so we wrap it.
   private readonly tfBackend = {
-    ref: (document.createElement('tf-backend') as any).tf_backend,
+    ref: document.createElement('tf-backend').tf_backend,
```
psybuzz, line 57: Why is it that the other files creating one of our Polymer elements need to import `'../tb_polymer_interop_types'` but this file does not?

tensorboard/webapp/tb_polymer_interop_types.ts
```diff
@@ -0,0 +1,73 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+declare global {
+  interface Document {
+    createElement(name: 'tf-backend'): TfBackendElement;
+    createElement(name: 'tf-globals'): TfGlobalsElement;
+    createElement(name: 'tf-storage'): TfStorageElement;
```
psybuzz, line 19: Neat pattern.

psybuzz (approved):
Nice to drop `any`s!

tensorboard/webapp/tb_polymer_interop_types.ts
stephanwlee: found a neater pattern :)

tensorboard/webapp/core/effects/core_effects.ts
stephanwlee: Included.

PR #4924, opened 2021-04-30
types: further improve Polymer interop types.

stephanwlee:
This change adds better typing pattern for tf-paginated-view and
vz-histograms.

### DIFF
--- a/tensorboard/webapp/widgets/histogram/histogram_component.ts
+++ b/tensorboard/webapp/widgets/histogram/histogram_component.ts
@@ -12,6 +12,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+import '../../tb_polymer_interop_types';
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/widgets/histogram/histogram_component.ts
```diff
@@ -12,6 +12,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+import '../../tb_polymer_interop_types';
```
psybuzz, line 15: Are we now able to drop the `as VzHistogramTimeSeries;` on L79 in this file?

psybuzz (approved):

tensorboard/webapp/widgets/histogram/histogram_component.ts
stephanwlee: Ah drats. Forgot to add those changes to the commit. Thanks.

PR #4926, opened 2021-05-03
rust: switch `sync_all` calls to `flush`

wchargin:
Summary:
For `--port-file` and `--error-file`, we only need output data to be
flushed to the operating system, not to actual disk. In addition to
making things a bit faster, `/dev/null` can now be used with these
arguments without raising `EINVAL`.

Uses not in `cli.rs` are all in tests.

Test Plan:
Running the data server with `--logdir /tmp/foo --port-file /dev/null`
now works as desired, and normally running `tensorboard` works, too.

wchargin-branch: rust-flush

### DIFF
--- a/tensorboard/data/server/cli.rs
+++ b/tensorboard/data/server/cli.rs
@@ -279,7 +279,7 @@ fn die_after_stdin() {
 fn write_port_file(path: &Path, port: u16) -> std::io::Result<()> {
     let mut f = File::create(path)?;
     writeln!(f, "{}", port)?;
-    f.sync_all()?;
+    f.flush()?;
     Ok(())
 }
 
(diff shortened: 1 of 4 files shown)


psybuzz (approved):
Neat!

At first I ran the server with `--port-file port.txt` to double check and was confused that it didn't write a port.txt.  Then realized it was written to `bazel-bin/tensorboard/data/server/server.runfiles/org_tensorflow_tensorboard/port.txt`

wchargin:
Using `bazel run` sets the current working directory to under the
runfiles tree. If you invoke the binary directly, or use `cargo run`,
the working directory is unchanged. In either case, our code is behaving
correctly.

PR #4929, opened 2021-05-04
debugger-v2: prevent runs request on reload

stephanwlee:
Currently, the debugger-v2 effects fetch its runs on `manualReload` or
`autoReload` without taking the active plugin into account.

This change fixes that and only fetch the runs when the plugin is
active.


psybuzz (approved):
Can we add a test that checks that it prevents requests when the plugin is 'inactive'?

Also to confirm: after this change, when an auto-reload occurs on a non-debugger dashboard, switching back to the debugger dashboard will result in stale data shown.

stephanwlee:
> Also to confirm: after this change, when an auto-reload occurs on a non-debugger dashboard, switching back to the debugger dashboard will result in stale data shown.

That is mostly consistent with how other dashboards works so I don't feel too bad about it. We can easily change the effects to load the data on plugins change which is not what TimeSeries does IIRC.

stephanwlee:
My bad; I misread the PR title and unintentionally merged this PR. Since the test can be added in a follow up, I will do so immediately.

PR #4930, opened 2021-05-04
types: declare types to be public

stephanwlee:
- experiments/types which is used to type the backend response should
  not be mangled.
- route parameter, since defined as strings in the route configurations,
  cannot be mangled.

### DIFF
--- a/tensorboard/webapp/experiments/types.ts
+++ b/tensorboard/webapp/experiments/types.ts
@@ -12,7 +12,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-export interface Experiment {
+export declare interface Experiment {
   id: string;
   name: string;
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/experiments/types.ts
```diff
@@ -12,7 +12,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-export interface Experiment {
+export declare interface Experiment {
```
psybuzz, line 15: Why does `Experiment` need to be declared non-mangleable / where Is it serialized?

psybuzz (approved):

tensorboard/webapp/experiments/types.ts
psybuzz: Sorry, I should've read the description closely.  It's a backend response, so we'd like to ensure we can read properties off it.

Issue #4928, opened 2021-05-04
tensorboard_data_server is not conform to manylinux2010


DavHau:
In the manylinux2010 tagged .whl release of tensorboard_data_server, the following file is not manylinux conform:

`tensorboard_data_server/bin/server` dynamically links against `libcrypto.so.1.1` and `libssl.so.1.1` which are not part of the allowed dependencies according to [PEP 571 -- The manylinux2010 Platform Tag](https://www.python.org/dev/peps/pep-0571/)

On systems with explicit dependency management like NixOS this leads to trouble.

I guess either the manylinux2010 tag or the dependency on the mentioned libraries should be removed.

wchargin:
Hi @DavHau—thanks for the report! We might be able to drop the
dependencies by switching from `openssl-sys` to `rustls`; I'll look into
that. If not feasible (or not sufficient), I'll look into changing the
platform tag; I'm a bit more cautious about this, but we'll see what we
can do.

wchargin:
@DavHau: Sent #4931 to fix this. If you want to give this a shot, you
can download the pre-built `tensorboard-data-server` wheel from the
GitHub Actions CI job:
<https://github.com/tensorflow/tensorboard/suites/2650138713/artifacts/58249455>

Or, you can of course build from source.

DavHau:
Thanks for the quick fix!

wchargin:
@DavHau: You're welcome; thanks again for the clear report! I've pushed
`tensorboard-data-server==0.6.1` with a manylinux2010-compliant binary.
It's compatible with TensorBoard 2.5.0.

<https://pypi.org/project/tensorboard-data-server/0.6.1/>

DavHau:
BTW, just out of curiosity I was running `auditwheel` on your .whl file:
```
❯❯❯ auditwheel show tensorboard_data_server-0.6.1-py3-none-manylinux2010_x86_64.whl

tensorboard_data_server-0.6.1-py3-none-manylinux2010_x86_64.whl is
consistent with the following platform tag: "manylinux_2_24_x86_64".

The wheel references external versioned symbols in these system-
provided shared libraries: libdl.so.2 with versions {'GLIBC_2.2.5'},
libm.so.6 with versions {'GLIBC_2.2.5'}, librt.so.1 with versions
{'GLIBC_2.2.5'}, libgcc_s.so.1 with versions {'GCC_3.3', 'GCC_4.2.0',
[...]

PR #4931, opened 2021-05-04
rust: switch from OpenSSL to `rustls`

wchargin:
Summary:
The RustBoard data server now uses `rustls` as a TLS backend instead of
using OpenSSL. This culls the `libcrypto.so.1.1` and `libssl.so.1.1`
shared library dependencies, fixing #4928.

Test Plan:
Run with `--logdir gs://tensorboard-bench-logs/edge_cgan` to verify that
RustBoard still works. Then, dump the `ldd` output:

```
$ cd tensorboard/data/server/
$ cargo build --release
$ ldd target/release/rustboard
	linux-vdso.so.1 (0x00007ffe8afbd000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f422533e000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f422531c000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f42251d8000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f42251d2000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f422500d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f4225b66000)
```

…and verify that all the shared libraries are on the [manylinux2010
whitelist].

[manylinux2010 whitelist]: https://www.python.org/dev/peps/pep-0571/#the-manylinux2010-policy

wchargin-branch: rust-rustls


wchargin:
@psybuzz: Would you mind giving this a sanity check on macOS? I think
that it should be pretty safe—this *removes* a system-level dependency
on OpenSSL, and `rustls`/`ring` don't link against system libraries as
far as I can tell. But it would be nice to double-check.

The following should suffice:

```
bazel run //tensorboard -- --load_fast true --logdir gs://tensorboard-bench-logs/edge_cgan --bind_all
```

psybuzz (approved):
Checked that with this PR, running `bazel run //tensorboard -- --load_fast true --logdir gs://tensorboard-bench-logs/edge_cgan --bind_all` on MacOS 11.3 works properly; scalars and images load fine.

wchargin:
Thank you!

PR #4934, opened 2021-05-04
debugger-v2: refresh data on change plugin

stephanwlee:
This change is a follow up to #4929 where we are:
- adding a test for reload actions whilst the plugin is inactive
- adds `changePlugin` as a condition for refreshing the data to follow the
  timeseries.

### DIFF
--- a/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
+++ b/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
@@ -279,8 +279,23 @@ export class DebuggerEffects {
   }
 
   private onCoreReload(): Observable<void> {
-    return this.actions$.pipe(
-      ofType(manualReload, reload, changePlugin),
+    return merge(
+      this.actions$.pipe(ofType(manualReload, reload)),
+      // Load the debugger data if it was never loaded before and now becomes
+      // an active plugin.
+      this.actions$.pipe(ofType(changePlugin)).pipe(
+        // We use debugger runs loaded state as a proxy to the debugger state
+        // being or not.
+        withLatestFrom(this.store.select(getDebuggerRunsLoaded)),
+        filter(([, loadState]) => {
+          return (
+            loadState.state === DataLoadState.NOT_LOADED ||
+            (loadState.state === DataLoadState.FAILED &&
+              loadState.lastLoadedTimeInMs === null)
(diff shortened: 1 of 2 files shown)


psybuzz (commented):
Thanks for adding the test!

The TimeSeries behavior is a bit more nuanced than just reloading upon "changePlugin to timeseries".  I think it avoids loading data if no reload occurred in the time while the plugin was inactive.

The Time Series Effect has an extra condition on changePlugin, to check "tagLoadState === DataLoadState.NOT_LOADED".  Importantly, its Reducer makes sure to set the loaded flag back to `NOT_LOADED` when a reload occurs [1].  In this way, it avoids reloading more often than needed.  This behavior is not really obvious because it is split across the Effect + Reducer.

Ideally, I think we would just make Debugger have the same behavior.  However, since the Debugger doesn't seem to have much attention atm, adding a TODO would also suffice for me.  wdyt?

[1] https://github.com/tensorflow/tensorboard/blob/83574eb51248d5db0a29b65684e185ff533c7e2a/tensorboard/webapp/metrics/store/metrics_reducers.ts#L341

stephanwlee:
> The TimeSeries behavior is a bit more nuanced than just reloading upon "changePlugin to timeseries". I think it avoids loading data if no reload occurred in the time while the plugin was inactive.

Debugger data model is a bit nebulous for current me so I decided to take the easy route of using the runs loaded state (why does it have its own notion of runs?!) to achieve the similar effect. On top of what timeseries did, I made some changes to fetch the data even in the `FAILED` state.

psybuzz (approved):

tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/effects/debugger_effects.ts
```diff
@@ -275,8 +279,23 @@ export class DebuggerEffects {
...
   private onCoreReload(): Observable<void> {
-    return this.actions$.pipe(
-      ofType(manualReload, reload),
+    return merge(
+      this.actions$.pipe(ofType(manualReload, reload)),
+      // Load the debugger data if it was never loaded before and now becomes
+      // an active plugin.
+      this.actions$.pipe(ofType(changePlugin)).pipe(
+        // We use debugger runs loaded state as a proxy to the debugger state
+        // being or not.
```
psybuzz, line 288: nit: typo in this comment?  "state being ___ or not"

PR #4938, opened 2021-05-05
timeseries: replace matInput with reusable FilterInputComponent

psybuzz:
The Time Series tag filter input and run selector were using Angular's
`matInput` whose styles are only included when another `mat-form-field`
has been rendered on the page before [1].

To make the filter input styles deterministic, drop the `matInput` directive
and replace it with a custom, styled widget for filter inputs.

[1] https://github.com/angular/components/issues/21871


tensorboard/webapp/theme/_tb_theme.template.scss
```diff
@@ -52,3 +52,12 @@ $tb-background: map-get($tb-theme, background);
...
 @include angular-material-theme($tb-theme);
+
+.tb-text-input {
+  -webkit-appearance: none;
+  box-sizing: content-box;
+  font: inherit;
+  border: none;
+  outline: none;
+  padding: 0;
+}
```
stephanwlee, line 63: Instead of creating a CSS based components, just create an Angular widget.

psybuzz: Creating a widget wrapper around `<input>` seems to add unnecessary boilerplate, imo.  Every attribute like `autocomplete` or event binding like `(input)` now requires extra @Input's and @Output's on this new widget.  Also, every module that uses it must now import another NgModule.

What do we gain from a TBTextInputWidget, from this extra boilerplate?

stephanwlee: If it has very consistent UX that we can supply with it, I think it is very well worth it. Kind of like mat-input; for me, I gain almost nothing by using it yet for the consistent UX, we use it. 

CSS based widgetization is creating a new pattern unnecessarily.

psybuzz: Switched to the custom widget approach, PTAL

tensorboard/webapp/testing/dom.ts
```diff
@@ -126,9 +126,16 @@ export function sendKey<T>(
     el.setSelectionRange(startingCursorIndex, startingCursorIndex);
   }
 
+  // Use to override default options on programmatic events to be closer to
+  // user-initiated events, which bubble and cancel.
+  const baseEventOptions = {
```
stephanwlee, line 131: Cool. Would you be able to spend a little bit of time to maybe find the TypeScript HTML typing for the event options? Since it is already checked when you pass it as the second argument to Event constructor, I guess it is okay as is, too.

tensorboard/webapp/widgets/filter_input/filter_input_component.ts
```diff
@@ -0,0 +1,59 @@
...
+      autocomplete="off"
+      [placeholder]="placeholder"
+      [matAutocomplete]="matAutocomplete"
+      [matAutocompleteDisabled]="!matAutocomplete"
+    />
+  `,
+  styles: [
+    `
+      :host {
+        display: flex;
```
stephanwlee, line 35: just wondering, is there a reason why this element has to be a non-inline element?

tensorboard/webapp/widgets/filter_input/filter_input_component.ts
```diff
@@ -0,0 +1,59 @@
...
+})
+export class FilterInputComponent {
+  @Input() value!: string;
+  @Input() matAutocomplete?: string;
+  @Input() placeholder?: string;
+
+  @ViewChild(MatAutocompleteTrigger)
+  autocompleteTrigger!: MatAutocompleteTrigger;
+
+  getAutocompleteTrigger(): MatAutocompleteTrigger {
```
stephanwlee, line 56: Consider just exposing `closeAutocomplete` unless the trigger has other APIs that must be used.

tensorboard/webapp/widgets/filter_input/filter_input_component.ts
```diff
@@ -0,0 +1,59 @@
...
+import {Component, Input, ViewChild} from '@angular/core';
+import {MatAutocompleteTrigger} from '@angular/material/autocomplete';
+
+/**
+ * A text input field intended for filtering items.
+ */
+@Component({
+  selector: 'tb-filter-input',
+  template: `
+    <input
```
stephanwlee, line 24: I noticed that the two applications both have the `<mat-icon svgIcon="search_24px">`. Would it maybe make sense to include the icon since it is literally for filters?

It would be close to this (on older safari): https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search
image

stephanwlee (approved):
Thanks for your willingness to compromise. 

It mostly LGTM and the comments are for clarifications and they are not blockers.
One thing though: I do notice that the tests are missing. I understand that the component is quite trivial (mostly a visual component). Was the omission an oversight or an intentional one?

psybuzz (commented):
Thanks for the review!

> I understand that the component is quite trivial (mostly a visual component). Was the omission an oversight or an intentional one?

In the previous commit, I was thinking it wasn't too valuable to test that this component properly passes props to 1 template element.  However, writing the test ended up catching missed things, and I think checking that autocomplete overlays fire made it worth adding a test.

tensorboard/webapp/testing/dom.ts
```diff
@@ -126,9 +126,16 @@ export function sendKey<T>(
...
+  // user-initiated events, which bubble and cancel.
+  const baseEventOptions: EventInit = {
+    bubbles: true,
+    cancelable: true,
+  };
+
   // Convert typing to object. Sadly, because keyCode is deprecated, it is not
   // typed properly.
-  const keyboardEventArg = {key, keyCode} as {};
+  const keyboardEventArg = {...baseEventOptions, key, keyCode} as {};
```
stephanwlee, line 138: FYI, `as {}` might have made you miss that typo.

PR #4942, opened 2021-05-06
core: fetch runs when alias changes

stephanwlee:
Polymer based TensorBoard is unaware of notion of an experiment and to
naively support it, we have made changes to the runs endpoint so it
returns a unique name of a run per experiment by prefixing name of an
experiment. However, it is not as simple as just that.

Because name of an experiment can be arbitrary large and can contain
illegal character like "/", we have a notion of alias per experiment.

The run-selector is unaware of anything experiment simply knows about
list of run names (not ids) and communicates with other Polymer based
components using the run names. As a result, whenever alias changes, we
need to refresh the runs data and reload the whole Polymer dashboard so
all components display and speak the same run name.

This change triggers Polymer run load whenever alias changes.

### DIFF
--- a/tensorboard/webapp/core/effects/core_effects.ts
+++ b/tensorboard/webapp/core/effects/core_effects.ts
@@ -178,6 +178,9 @@ export class CoreEffects {
             return of([]);
           }
 
+          // If alias map changes, we need to refetch the list of runs as
+          // Polymer's run selector and tags rely on run names including the
+          // alias.
           return this.store.select(getExperimentIdToAliasMap).pipe(
             distinctUntilChanged((beforeAliasDict, afterAliasDict) => {
               const entries = Object.entries(beforeAliasDict);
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/core/effects/core_effects.ts
```diff
@@ -172,11 +171,54 @@ export class CoreEffects {
...
-      const runsReload$ = this.onDataReload$.pipe(
+      const runsReload$ = this.onDashboardLoad$.pipe(
+        map(([, routeKind]) => routeKind),
+        switchMap((routeKind) => {
+          if (routeKind !== RouteKind.COMPARE_EXPERIMENT) {
+            return of([]);
+          }
+
+          return this.store.select(getExperimentIdToAliasMap).pipe(
+            distinctUntilChanged((beforeAliasDict, afterAliasDict) => {
```
bmd3k, line 182: Can you add a (possibly brief) comment explaining what the code is trying to achieve here? Or move into its own function whose name describes the code?

tensorboard/webapp/core/effects/core_effects.ts
```diff
@@ -172,11 +171,54 @@ export class CoreEffects {
...
+              }
+              for (const [experimentId, alias] of entries) {
+                if (afterAliasMap.get(experimentId) !== alias) {
+                  return false;
+                }
+              }
+              return true;
+            }),
+            // HACK: arbitrary microtask delay.
+            // aliasChange leads to route changes and route changes causes
```
bmd3k, line 196: aliasChange does not exist as a symbol, but maybe you just want "An alias change..."

bmd3k, line 196: Do you mind just taking another look at this comment for spelling?

bmd3k (approved):

PR #4944, opened 2021-05-07
sync: add marked indirection package

stephanwlee:
Internal marked is a bit funky so, to paper over the differences, we
introduce the indirection.


japie1235813 (approved):
Thanks for helping resolve this!

tensorboard/webapp/third_party/marked.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * @fileoverview This is an marked interop that papers over the differences
+ * within google internal repository and the external repository. Please depend
+ * on this module instead of depending on the marked directly.
+ */
+import * as markedImport from 'marked';
+
+// You cannot `export * from 'marked';` due to below error[1].
+// [1]: 'marked' uses 'export =' and cannot be used with 'export *'.
+export const marked = markedImport;
```
evmar, line 24: Try:

export import marked = markedImport;

(The difference is that 'const' doesn't propagate any defined types)

stephanwlee: Done.

PR #4952, opened 2021-05-10
Import AppRoutingModule and FeatureFlagModule from CoreModule

bmd3k:
Modify CoreModule so that it explicitly imports AppRoutingModule and FeatureFlagModule.

This implements a new policy where we like to make dependencies explicit in the module graph. In this case, core_effects depends on selectors and actions from both app_routing and feature_flag.

This also fixes a recent regression where some Google-internal tensorboards are unable to successfully load experiments where path does not contain final '/' after the experiment id.  Googlers, see b/187695455.

For the broken tensorboards, the observed behavior is as follows:

1. app_routing_effects dispatches navigated() action shortly after experiment page load.
    * https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/app_routing/effects/app_routing_effects.ts;l=265;drc=ea145a40a0c50092fecdf83d7c934260a3e15c0b
2. core_effects reacts to navigated() event by attempting to load data. page has not yet been updated with final '/' and so the data requests do not contain the experiment id.
    * https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/core/effects/core_effects.ts;l=96;drc=658ca67762c2509ccdea2bb54d3849c6faf5be9e
3. app_routing_effects reacts to navigated() event by updating the browser URL with the final '/'. Too late!
    * https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/app_routing/effects/app_routing_effects.ts;l=275;drc=ea145a40a0c50092fecdf83d7c934260a3e15c0b

We actually need (3) to happen before (2) if we want the data requests to contain the experiment id.

It seems sufficient to register app_routing_effects before core_effects. Instead of modifying individual tensorboard services to guarantee that registration order (Googlers, see cl/372979868), we force that order by having CoreModule import AppRoutingModule before it registers its effects.


psybuzz (approved):
Thanks for the clear description!

PR #4953, opened 2021-05-10
Revert "refactor: use router event to fetch data (again) (#4941)"

psybuzz:
This reverts commit 658ca67762c2509ccdea2bb54d3849c6faf5be9e.

The offending commit introduced a hard dependency on the `AppRoutingModule`,
which was previously optional. By reverting, TensorBoards can now operate
without AppRouting if needed. In the long term, we plan to reland.

Googlers, this fixes an internal-only issue, see b/187747192 and test
sync cl/373010383.


psybuzz:
Appreciate the fast review :)

PR #4958, opened 2021-05-11
color grouping: add feature flag

japie1235813:
add enabledColorGroup flag for color grouping function switch


stephanwlee (commented):
You may want to add serialize/deserializer so you don't have to modify the URL all the time.

PR #4962, opened 2021-05-12
plugin-api: fix getURLPluginData()

psybuzz:
Currently, `getURLPluginData()` always returns empty data, because its
code path has `useHash() === false`. "core-host-impl.ts" directly imports
Polymer's `tf_storage_utils`, producing a separate scope with state that
is not shared with the Polymer side. When `deeplink/hash.ts` sets the flag
via `document.createElement('tf-globals').tf_globals.setUseHash(true)`,
the variable that updates is different from the one that is read by the
plugin host.

This change makes the plugin host read Polymer state via manually
instantiating a Polymer CustomElement, rather than via direct import.

Googlers, see b/187854773 and test sync cl/373276716

For reference, there are 2 ways for an Angular TS module to use a Polymer TS module.
A) direct import
```ts
import * as globals from 'tf_globals';
globals.useHash();
```

B) access through CustomElement
```ts
import * as tf_globals from 'tf_globals';
@customElement('tf-globals')
class TfGlobals extends PolymerElement {
  _template = null;
  tf_globals = tf_globals;
}
document.createElement('tf-globals').tf_globals.useHash();
```

After TS+Closure compilation, these 2 methods result in separate copies of the Polymer code, each with its individual scope. Angular code using method A) will not be sharing the same state as Polymer components.  We should prefer doing method B).

### DIFF
--- a/tensorboard/components/experimental/plugin_util/plugin_api_host_test.ts
+++ b/tensorboard/components/experimental/plugin_util/plugin_api_host_test.ts
@@ -24,7 +24,6 @@ import {
   getExperimentIdsFromRoute,
   getRuns,
 } from '../../../webapp/selectors';
-import * as tf_storage_utils from '../../tf_storage/storage_utils';
 import {PluginCoreApiHostImpl} from './core-host-impl';
 import {MessageId} from './message_types';
 import {Ipc} from './plugin-host-ipc';
@@ -236,7 +235,6 @@ describe('plugin_api_host test', () => {
       let triggerGetUrlData: (context: {
         pluginName: string;
       }) => Record<string, string> = () => ({});
-      let getUrlDictSpy: jasmine.Spy;
 
       beforeEach(() => {
         listenSpy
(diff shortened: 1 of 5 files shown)


tensorboard/components/experimental/plugin_util/plugin_api_host_test.ts
```diff
@@ -246,17 +244,23 @@ describe('plugin_api_host test', () => {
...
+            getUrlHashDict: () => {
+              return {
+                'globalThing': 'hey',
+                'p.plugin_id.a': '1',
+                'p.plugin_id.b': 'b',
+                'p.another_plugn.b': '2',
+              }
+            },
+          },
+        } as any);
```
stephanwlee, line 263: Is there a reason not to use the tb_polymer_interop_type and undo `any` here?

stephanwlee (approved):
Besides the lint error, the code roughly lgtm.

tensorboard/components/experimental/plugin_util/plugin_api_host_test.ts
psybuzz: Switched to `as unknown as TfStorageElement`.

When trying to directly cast/declare a `TfStorageElement`, TS complains that our plain object is too different from `HTMLElement`, so it requires the unknown first.  This unknown casting loses actual typechecking benefits, but at least it is clearer to readers what type it is.

PR #4965, opened 2021-05-12
what's new: export testing.ts for functions reuse

japie1235813:
to reuse `buildNotificationState` and `buildStateFromNotificationState` in other tests

### DIFF
--- a/tensorboard/webapp/notification_center/_redux/BUILD
+++ b/tensorboard/webapp/notification_center/_redux/BUILD
@@ -55,6 +55,7 @@ tf_ts_library(
     srcs = [
         "testing.ts",
     ],
+    visibility = ["//tensorboard:internal"],
     deps = [
         ":types",
     ],


tensorboard/webapp/notification_center/_redux/BUILD
```diff
@@ -55,6 +55,7 @@ tf_ts_library(
     srcs = [
         "testing.ts",
     ],
+    visibility = ["//tensorboard:internal"],
```
stephanwlee, line 58: From which feature are you trying to use this target from?

japie1235813: for testing in cl/373274039

stephanwlee: There is almost no point of providing the correct structure especially when the effects will be only using the public API (selector) and won't need to know much the store itself. Please omit this change and just use `store.overrideSelector` especially across the feature boundary.

japie1235813: I see. Drop this pr.

PR #4966, opened 2021-05-12
runs: implement Runs and Experiments grouping

stephanwlee:
This change implements utility for grouping by experiments and runs.
From the experiment-less users, this change should be a noop.

This change also reorganizes the default color assignment to operate
based on the group generated by the utility.

Alternative considered: thought about only remembering `groupId` to
color instead of `groupToColor` and `defaultRunColor`. With this mode,
we would read the `runIds` state and ultimately create run-to-color map
but this requires:
1. stitching information in the selector making the memoization trickier
2. users of the selector pass current `experimentIds` so we can obtain
  proper `runIds`. This would increase blast radius of the change a bit
  too large.


tensorboard/webapp/runs/store/runs_reducers.ts
```diff
@@ -36,16 +36,17 @@ import {
...
   initialState: dataInitialState,
   reducers: dataRouteContextReducers,
 } = createRouteContextedState<RunsDataRoutefulState, RunsDataRoutelessState>(
   {
-    colorOverride: new Map(),
-    defaultColor: new Map(),
-    nextGroupColorIndex: 0,
+    runColorOverride: new Map(),
+    defaultRunColor: new Map(),
+    groupToColor: new Map(),
```
japie1235813, line 48: It's a bit hard to tell these two maps from the naming.
IIUC, `defaultRunColor` means the mapping of run to color
while `groupToColor` is the mapping of group to color, a group could be a run or a set of runs (like group by experiment)
Can we rename this to be more a pair-like?

tensorboard/webapp/runs/store/runs_reducers.ts
```diff
@@ -180,29 +181,36 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
...
+      state.runIdToExpId
+    );
+    Object.entries(groups).forEach(([groupId, runs]) => {
+      const color =
+        groupToColor.get(groupId) ??
+        CHART_COLOR_PALLETE[groupToColor.size % CHART_COLOR_PALLETE.length];
+      groupToColor.set(groupId, color);
+
+      for (const run of runs) {
+        defaultRunColor.set(run.id, color);
```
japie1235813, line 199: so.. after grouping (by exp/regex etc). The colors for each run would just change. Even when users go back to color by run, it would just stay there?

I think the naming "default" is confusing me.

tensorboard/webapp/runs/store/runs_reducers_test.ts
```diff
@@ -238,7 +247,67 @@ describe('runs_reducers', () => {
...
+
+    describe('advanced grouping', () => {
+      it('assigns default color to by experiment', () => {
+        const state = buildRunsState({
+          groupBy: {
+            key: GroupByKey.EXPERIMENT,
+          },
+          defaultRunColor: new Map([
+            ['foo', '#aaa'],
+            ['bar', '#aaa'],
```
japie1235813, line 260: does this runs these two runs ('foo', 'bar') are a group (they're on the same experiment?)

tensorboard/webapp/runs/store/runs_reducers_test.ts
```diff
@@ -238,7 +247,67 @@ describe('runs_reducers', () => {
...
+            buildRun({id: 'beta'}),
+            buildRun({id: 'gamma'}),
+            buildRun({id: 'lambda'}),
+          ],
+          newRunsAndMetadata: {
+            eid1: {
+              runs: [
+                buildRun({id: 'baz'}),
+                buildRun({id: 'foo'}),
+                buildRun({id: 'qaz'}),
```
japie1235813, line 280: 'foo' is not in this experiment but 'foo' is assigned the same color? Or the default status shows the 'foo' is also under the same eid?

stephanwlee: This is testing the case when a `run` is already present in the store with its value assigned.

tensorboard/webapp/runs/store/runs_reducers_test.ts
stephanwlee: Wrote a comment; I was simulating data inconsistency.

tensorboard/webapp/runs/store/runs_reducers.ts
stephanwlee: Renamed the property name to be more explicit and left a comment in the interface. PTAL.

japie1235813 (approved):

stephanwlee:
Resolved some comments offline; while I do agree that `runColor` & `runOverrideColor` separation serves no particular purposes at the moment, I am keeping the status quo because:
- it is unrelated to this change
- easier to debug and inspect the state progression when the overrides are separate from the default assignment
- we can have richer UX around overrides (e.g., put an indicator that a color was overridden and introduce an UI affordance to put it back to the default assignment)

PR #4967, opened 2021-05-13
flag: add feature flag for enableColorGroup

stephanwlee:
Currently, the feature flag is not configurable via URL and this change
enables the serialization and deserialization of the flag.

### DIFF
--- a/tensorboard/webapp/routes/dashboard_deeplink_provider.ts
+++ b/tensorboard/webapp/routes/dashboard_deeplink_provider.ts
@@ -28,7 +28,10 @@ import {
 import {CardUniqueInfo, METRICS_SETTINGS_DEFAULT} from '../metrics/types';
 import * as selectors from '../selectors';
 import {getMetricsScalarSmoothing} from '../selectors';
-import {EXPERIMENTAL_PLUGIN_QUERY_PARAM_KEY} from '../webapp_data_source/tb_feature_flag_data_source_types';
+import {
+  ENABLE_COLOR_GROUP_QUERY_PARAM_KEY,
+  EXPERIMENTAL_PLUGIN_QUERY_PARAM_KEY,
+} from '../webapp_data_source/tb_feature_flag_data_source_types';
 import {
   DeserializedState,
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/routes/dashboard_deeplink_provider.ts
```diff
@@ -86,6 +89,12 @@ export class DashboardDeepLinkProvider extends DeepLinkProvider {
         const queryParams = experimentalPlugins.map((pluginId) => {
           return {key: EXPERIMENTAL_PLUGIN_QUERY_PARAM_KEY, value: pluginId};
         });
+        if (typeof overriddenFeatureFlags.enabledColorGroup === 'boolean') {
```
japie1235813, line 92: What if the type is not boolean?
As the test below shown, we want to have "enabledColorGroup=foo" be true? Which part handle this logic?

stephanwlee: > we want to have "enabledColorGroup=foo" be true? Which part handle this logic?

Yes, that's my intention. When user specifies a value that is not a stringified JavaScript boolean, I think it is safe to treat it as no-value and enable the feature. Since it is easily reversible action (on user's part), I don't think this is a bad call. Do note that we have been doing this for other flags already.

japie1235813: so... if it's `enabledColorGroup=foo` it will not do queryParams key-value setup and then `if (params.has(ENABLE_COLOR_GROUP_QUERY_PARAM_KEY))` will be false
how do we check the `!== 'false'` condition to still enable the `enabledColorGroup` feature?

japie1235813 (approved):

tensorboard/webapp/routes/dashboard_deeplink_provider.ts
stephanwlee: Resolved offline by explaining how this change integrates with FeatureFlagDataSource -> store -> Router (DashboardDeepLinkProvider)

PR #4973, opened 2021-05-14
Use an r-string so the \n doesn't break the site.

MarkDaoust:
Fixes: https://www.tensorflow.org/api_docs/python/tf/summary/text

The \n is interpreted by python as a new-line.
When building the site, we dedent the text. After that newline there is zero indentation.
0 is the floor, so none of the text is dedented.
When converted to markdown the 4-space indent is interpreted as preformatted text.


nfelt (approved):
Thank you!

PR #4974, opened 2021-05-14
router: change URL in between navigating & navigated

stephanwlee:
With this change, we now modify the URL before `navigated` is fired so
all the downstream can send request right away without having to confirm
that the route change is indeed reflected in the URL bar.

Context: previously, our router modified the URL only when the
`navigated ` action is fired. Often, this action is listened to by many
views/effects so they can do a cleanup or a data fetch for the new
route. Now, because the URL was only modified after `navigated ` is
fired, there could be a timing issues where, depending on the order of
execution, we can be making requests to the unmodified URL, causing
issues.

Now, we are changing the semantics of the `navigated` to really really
mean, 'hey, the navigation is now complete and everything is settled'.

Tested this change internally.

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
@@ -642,7 +642,7 @@ describe('app_routing_effects', () => {
       });
 
       function navigateAndExpect(
-        navigation: Navigation,
+        navigation: Navigation | Route,
         expected: {pushStateUrl: null | string; replaceStateUrl: null | string}
       ) {
         fakeAsync(() => {
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
@@ -153,7 +153,7 @@ describe('app_routing_effects', () => {
     store.overrideSelector(getActiveRoute, null);
   });
 
-  describe('fireNavigatedIfValidRoute$', () => {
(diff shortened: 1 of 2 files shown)


stephanwlee:
To reviewer: recommend turning on 'ignore white space change' option.

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -587,190 +633,183 @@ describe('app_routing_effects', () => {
...
 
-    it('pushes state if path and search do not match new route on navigated', () => {
-      const activeRoute = buildRoute({
-        routeKind: RouteKind.EXPERIMENTS,
-        pathname: '/experiments',
-        queryParams: [],
-        navigationOptions: {
-          replaceState: false,
-        },
+        navigateAndExpect(activeRoute, {
```
psybuzz, line 683: Based on this call, the type for the first parameter to `navigateAndExpect` on L645 should be `Navigation | Route`, not `Navigation`.

psybuzz (approved):
Thanks for testing it internally.  I tested it in OSS and looks good

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
stephanwlee: Fixed; thanks for catching it.

stephanwlee:
Tested once more; since unclear whether this will break our odd variant, I will merge this, sync it, then simmer it so it can be reverted in isolation.

PR #4977, opened 2021-05-17
refactor: settings code structure to new format

stephanwlee:
As the team, we decided to refactor our folder structure to denote
private modules with underscore prefix. This change just changes the
folder structure without other functional changes.


tensorboard/webapp/BUILD
```diff
@@ -342,7 +342,6 @@ tf_ng_web_test_suite(
         "//tensorboard/webapp/runs/store:testing",
         "//tensorboard/webapp/runs/views/runs_selector:runs_selector_test",
         "//tensorboard/webapp/runs/views/runs_table:runs_table_test",
-        "//tensorboard/webapp/settings:test_lib",
```
psybuzz, line 345: There might've one a long time ago that I don't remember, but was there an offline discussion about moving away from the hourglass-style for testing targets?

tensorboard/webapp/settings/_views/settings_module.ts
```diff
@@ -0,0 +1,47 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
psybuzz, line 1: This file's naming scheme is misaligned with notifications_center/'s.

- `notification_center/_views/views_module.ts`
- `settings/_views/settings_module.ts`

Which naming scheme is correct?

psybuzz (approved):

tensorboard/webapp/settings/_views/settings_module.ts
stephanwlee: There was no prescription around that. I'm leaning towards latter (thus my PR :D)

tensorboard/webapp/BUILD
stephanwlee: Hourglass makes any changes in webapp folder painfully slow for `//tensorboard/webapp:karma_test` so we decided to have per-feature "hourglass".

PR #4980, opened 2021-05-17
runs: re-assign colors on runGroupByChanged event

stephanwlee:
This change implements the color reassignment when the groupBy state
changes.

Product consideration:
- while we can keep the run color override, we are starting by resetting
all the manual changes for now.

### DIFF
--- a/tensorboard/webapp/runs/store/runs_reducers.ts
+++ b/tensorboard/webapp/runs/store/runs_reducers.ts
@@ -208,41 +208,48 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
       groupKeyToColorString,
     };
   }),
-  on(runsActions.runGroupByChanged, (state: RunsDataState, {experimentIds, groupBy}): RunsDataState => {
-    // Reset the groupKeyToColorString
-    const groupKeyToColorString = new Map<string, string>();
-    const defaultRunColorForGroupBy = new Map(state.defaultRunColorForGroupBy);
+  on(
+    runsActions.runGroupByChanged,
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/runs/store/runs_reducers.ts
```diff
@@ -208,6 +208,48 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
...
+      );
+
+      const allRuns = experimentIds
+        .flatMap((experimentId) => {
+          return state.runIds[experimentId];
+        })
+        .map((runId) => {
+          return state.runMetadata[runId];
+        })
+        .filter(Boolean);
```
japie1235813, line 227: IIUC, Line 224-226 return a list of Run, which is defined here
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/runs/store/runs_types.ts;l=30-36?q=RunsDataState&ss=tensorflow%2Ftensorboard
What does the `filter(Boolean)` do? Can the list include invalid runs? (like None?)

japie1235813 (approved):

tensorboard/webapp/runs/store/runs_reducers.ts
stephanwlee: Ah thanks a lot for catching this. I was wrong in thinking that the type would be `undefined` because I am accessing a random property on an object but indeed I was wrong; TypeScript does not complain even if I remove `filter(Boolean)`. Removed it accordingly.

stephanwlee:
Thank you for a quick review!

PR #4982, opened 2021-05-17
Add JSDoc to DeepLinkProvider interface definition

bileschi:
Interfaces should have JSDoc.

Just adds JSDoc.

### DIFF
--- a/tensorboard/webapp/app_routing/deep_link_provider.ts
+++ b/tensorboard/webapp/app_routing/deep_link_provider.ts
@@ -54,8 +54,8 @@ export abstract class DeepLinkProvider {
   /**
    * When the URL changes, the router dispatches a `stateRehydratedFromUrl`
    * action [1], and calls this method to generate the action's payload.
-   * Specifically, the resulting objection of this method becomes the
-   * `partialState` field on the payload.
+   * Specifically, the result of this method becomes the `partialState`
+   * field on the payload.
    *
    * [1] webapp/app_routing/actions/app_routing_actions.ts
    *
--- a/tensorboard/webapp/app_routing/deep_link_provider.ts
+++ b/tensorboard/webapp/app_routing/deep_link_provider.ts
@@ -20,11 +20,30 @@ import {State} from '../app_state';
 
 import {SerializableQueryParams} from './types';
 
+/**
+ * The `DeepLinkProvider` interface standardizes the relationship between the
+ * app state and the URL parameters.  It is expected that when either an update
+ * to the URL parameters should change app state, or, in the other direction,
+ * when an update to the state should cause a change in the URL parameters,
+ * that logic should be invoked through a `DeepLinkProvider`.
+ */
 @Injectable()
 export abstract class DeepLinkProvider {
+
+  /**
+   * Modifies the URL to match the state of the app.
+   * @param store The ngrx store containing the state of the app.
+   * @return An observable which will emit the query parameters matching the state.
+   */
   abstract serializeStateToQueryParams(
     store: Store<State>
   ): Observable<SerializableQueryParams>;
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/app_routing/deep_link_provider.ts
```diff
@@ -20,11 +20,30 @@ import {State} from '../app_state';
 
 import {SerializableQueryParams} from './types';
 
+/**
+ * The `DeepLinkProvider` interface standardizes the relationship between the
+ * app state and the URL parameters.  It is expected that when either an update
+ * to the URL parameters should change app state, or, in the other direction,
+ * when an update to the state should cause a change in the URL parameters,
+ * that logic should be invoked through a `DeepLinkProvider`.
```
stephanwlee, line 28: While statement is correct, I think it is more correct to say below:
```
TensorBoard rotuer considers URL as a view where a subset of state gets projected onto (like a HTML input element) and the DeepLinkProvider is analogous to a Container of the view. Since the state to be reflected onto the URL bar differs by a route, we have a DeepLinkProvider per route.
```

tensorboard/webapp/app_routing/deep_link_provider.ts
```diff
@@ -20,11 +20,30 @@ import {State} from '../app_state';
...
+ */
 @Injectable()
 export abstract class DeepLinkProvider {
+
+  /**
+   * Modifies the URL to match the state of the app.
```
stephanwlee, line 34: This method is not intended to do the mutation but only grab certain state from the store into query parameters. Also, the resulting `Observable` should emit when the query parameter should change.

tensorboard/webapp/app_routing/deep_link_provider.ts
```diff
@@ -20,11 +20,30 @@ import {State} from '../app_state';
...
+   * @param store The ngrx store containing the state of the app.
+   * @return An observable which will emit the query parameters matching the state.
+   */
   abstract serializeStateToQueryParams(
     store: Store<State>
   ): Observable<SerializableQueryParams>;
 
+  /**
+   * Given the query params, updates the ngrx state to match.
```
stephanwlee, line 43: Again, this method does not do the mutation; rather, it only returns a data structure, as per definition in the interface, based on current query parameter.

tensorboard/webapp/app_routing/deep_link_provider.ts
```diff
@@ -20,11 +20,30 @@ import {State} from '../app_state';
...
+   * @param queryParams TensorBoard URL SeralizableQueryParams that should be used
+   *     to update the app state.
+   * @return A JS object describing a patch that should be made to the ngrx state.
```
stephanwlee, line 46: > A JS object describing a patch that should be made to the ngrx state

Reducers do not apply this object directly onto the ngrx state. e.g., the interface that this method returns can be so that it is wildly different from that of the reducers.

tensorboard/webapp/app_routing/deep_link_provider.ts
```diff
@@ -20,11 +20,49 @@ import {State} from '../app_state';
...
+   *     the URL.
+   */
   abstract serializeStateToQueryParams(
     store: Store<State>
   ): Observable<SerializableQueryParams>;
 
+  /**
+   * When the URL changes, the router dispatches a `stateRehydratedFromUrl`
+   * action [1], and calls this method to generate the action's payload.
+   * Specifically, the resulting objection of this method becomes the
```
psybuzz, line 57: One minor objection: `resulting objection` --> `resulting object` or `result`

psybuzz (approved):
Thanks for documenting this

PR #4984, opened 2021-05-18
settings: add redux state for global state

stephanwlee:
Motivation
----------
In the multi-experiment version of TensorBoard, settings in the metrics
dashboard "resets" to its default values every time a new experiment
opens. Users got a bit uncomfortable doing that so we want to improve
the UX by remembering user's inter-experiment global settings.

This change
-----------
This change simply adds the redux state but we eventually want to
remember the settings via localStorage. While this solution is not a
panacea (we still forget when user rotates port number), this can
improve certain experience drastically.

Again, this change only adds redux state, reducers, actions, and
selectors.

Discussion
----------
For consumer of the selector, it is quite hard to now mix the
globalSetting + per-experiment override. In fact, you really need three
selectors that purportedly do the similar things:
- one for getting the global setting (needed for
  settings view component).
- one for getting and rendering per-experiment setting (needed for rendering right nav).
- one for mixing the two and just getting the, for example, smoothing.

For providing the third kind of selector, we may wanto to provide a
composite selector that grabs state from two different features.
However, placement of said selector is murky and such selector creates
dependencies on multiple DataNgModule.

Plan
----
- write the persistence via LocalStorage
- read the persisted data from LocalStorage and bootstrap the data from it
- change the runs_type to have `null`able values for redundant setting
  - change right nav and views to mix the data accordingly
- Introduce UI affordance to change the global settings in the settings dialog
 (conditionally rendered based on plugins listing)

### DIFF
--- a/tensorboard/webapp/settings/BUILD
+++ b/tensorboard/webapp/settings/BUILD
@@ -18,6 +18,7 @@ tf_ng_module(
 tf_ng_web_test_suite(
     name = "karma_test",
     deps = [
+        "//tensorboard/webapp/settings/_redux:test_lib",
         "//tensorboard/webapp/settings/_views:test_lib",
     ],
 )
--- a/tensorboard/webapp/settings/BUILD
+++ b/tensorboard/webapp/settings/BUILD
@@ -9,6 +9,7 @@ tf_ng_module(
     ],
     deps = [
         "//tensorboard/webapp/core",
+        "//tensorboard/webapp/settings/_redux",
         "//tensorboard/webapp/settings/_views",
         "@npm//@angular/core",
     ],
--- a/tensorboard/webapp/settings/_redux/BUILD
+++ b/tensorboard/webapp/settings/_redux/BUILD
@@ -1,0 +1,40 @@
+load("//tensorboard/defs:defs.bzl", "tf_ng_module")
+
+package(default_visibility = ["//tensorboard/webapp/settings:__subpackages__"])
+
+tf_ng_module(
+    name = "_redux",
+    srcs = [
+        "settings_actions.ts",
+        "settings_data_module.ts",
+        "settings_reducers.ts",
+        "settings_types.ts",
+    ],
+    deps = [
+        "//tensorboard/webapp/metrics:types",
+        "@npm//@angular/core",
+        "@npm//@ngrx/store",
+        "@npm//rxjs",
+    ],
+)
+
+tf_ng_module(
+    name = "test_lib",
+    testonly = True,
+    srcs = [
+        "settings_reducers_test.ts",
+        "settings_selectors_test.ts",
+        "testing.ts",
+    ],
+    deps = [
+        ":_redux",
+        "//tensorboard/webapp/angular:expect_ngrx_store_testing",
+        "@npm//@angular/common",
+        "@npm//@angular/compiler",
+        "@npm//@angular/core",
+        "@npm//@angular/forms",
+        "@npm//@angular/platform-browser",
+        "@npm//@ngrx/store",
+        "@npm//@types/jasmine",
+    ],
+)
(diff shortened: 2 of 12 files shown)


psybuzz:
> For consumer of the selector, it is quite hard to now mix the
globalSetting + per-experiment override.

Firstly, why do we want this complicated mixing?  Can we get away with a much simpler "metrics settings always persist globally across all experiments"?  I don't think there's any need to save and restore a different set of settings for each experimentId.  The UX seems to become more confusing, to me.  If we hear strong user feedback requesting this, we can always consider ways to make experimentId-specific settings later.

> In fact, you really need three selectors that purportedly do the similar things:
one for getting the global setting (needed for rendering right nav).
one for getting and rendering per-experiment setting (needed for
settings view component).
one for mixing the two and just getting the, for example, smoothing.

It is mentioned at the end of the PR description that global settings are changed in the settings dialog, so presumably the first case was meant to say "one for getting the global setting (needed for rendering the settings dialog)"?

> Introduce UI affordance to change the global settings in the settings dialog
(conditionally rendered based on plugins listing)

I'm interpreting this statement as a proposal to have 2 smoothing sliders, 1 in the right nav and another 1 in the settings dialog.  Why are we adding redundant metrics settings in the settings dialog, when the settings pane already shows them?

Can we instead keep things simple and focus on the persistence behavior, without changing introducing new settings UI?

IIUC, we do not need to be introducing redundant actions/selectors/reducing functions for each metrics setting, if we care only about persisting the existing ones.

stephanwlee:
Before addressing other parts, I think we should first nail down on your first point.

>  why do we want this complicated mixing? Can we get away with a much simpler "metrics settings always persist globally across all experiments"? I don't think there's any need to save and restore a different set of settings for each experimentId. The UX seems to become more confusing, to me. If we hear strong user feedback requesting this, we can always consider ways to make experimentId-specific settings later.

I am a bit worried that your simplified flow will change current behavior of the app. As I proposed it, it would augment current behavior and gives user an added behavior while your proposal would change the behavior completely. At the expense of a bit more complex selector, I am willing to do something that is additive. 

The bigger picture is, we want to change the default values in our current timeseries (smoothing to 0 and tooltip sorting to DESCENDING). However, we are afraid that some people will dislike this change and want to provide to revert to the older defaults via the global settings. We also want to know whether our choice to change the default was the right one by analyzing events in our hosted services. 

WDYT?

psybuzz:
> I am a bit worried that your simplified flow will change current behavior of the app.

Could you please elaborate on your concern?  When navigating from one experiment to another, why is it wrong to restore the most recent smoothing, tooltip sort, etc?

> As I proposed it, it would augment current behavior and gives user an added behavior while your proposal would change the behavior completely. At the expense of a bit more complex selector, I am willing to do something that is additive.

The added behavior creates confusion for users, imo.  In a multi-experiment world, if I open a specific experiment, how does one know whether the current smoothing value comes from experiment-specific setting overrides, or whether it is using the global setting?  If one changes the smoothing via the settings dialog, how does the UI communicate that this is a global change that affects other experimentIds?  If one updates the smoothing by changing the URL query param, does that affect the global setting or the override?  If a experiment-specific override is active, does a change to the setting in the settings dialog affect also update the override, or does it count as a surprising no-op?  How does one clear/unset their experiment-specific overrides?

> The bigger picture is, we want to change the default values in our current timeseries (smoothing to 0 and tooltip sorting to DESCENDING). However, we are afraid that some people will dislike this change and want to provide to revert to the older defaults via the global settings.

IIUC, making the existing setting controls global settings is consistent with that goal.

> We also want to know whether our choice to change the default was the right one by analyzing events in our hosted services.

Could you help me understand why experiment-specific overrides is needed for understanding whether the defaults are correct?  It seems to me that in the simplified flow, with only global settings, tracking the occurrence of how many times the global setting is changed still addresses the goal.

stephanwlee:
There were offline discussions about this as comment thread prompted many clarifications.

As planned, we do have a way to introduce this feature, then eventually move to a model where right pane settings are globally persisted. However, each UX change would cause (potentially minor) user confusions. While the suggested change does change existing behavior (with or without experiments), we did agree to move forward with that instead. 

As such, the change is now obsolete.

PR #4986, opened 2021-05-18
runs: fully implement color group by UI

stephanwlee:
This change introduces:
- checkbox for displaying the selected groupBy setting
- dispatch actions when you use a menu item

image

With this change, now, colorBy "runs" and "experiments" are functional.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ts
+++ b/tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ts
@@ -12,7 +12,15 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-import {ChangeDetectionStrategy, Component} from '@angular/core';
+import {
+  ChangeDetectionStrategy,
+  Component,
+  EventEmitter,
+  Input,
+  Output,
+} from '@angular/core';
+
+import {GroupBy, GroupByKey} from '../../types';
 
 @Component({
   selector: 'runs-group-menu-button-component',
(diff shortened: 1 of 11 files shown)


tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ts
```diff
@@ -24,18 +32,58 @@ import {ChangeDetectionStrategy, Component} from '@angular/core';
...
+        role="menuitemradio"
+        (click)="onGroupByChange.emit({key: GroupByKey.EXPERIMENT})"
+      >
+        <span>
+          <mat-icon
+            *ngIf="selectedGroupBy.key === GroupByKey.EXPERIMENT"
+            svgIcon="done_24px"
+          ></mat-icon>
+        </span>
+        <label>Experiments</label>
```
japie1235813, line 48: Just to clarify. Do we want `s` at the end or not?
`Experiment` or `Experiments`
(In the doc we used singular not plural to communicate, but I don't object to change it to plural)
I don't know which makes more sense but just want to be consistent so let me know your decision and I'll follow.

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ts
```diff
@@ -24,18 +32,58 @@ import {ChangeDetectionStrategy, Component} from '@angular/core';
...
+        role="menuitemradio"
+        (click)="onGroupByChange.emit({key: GroupByKey.RUN})"
+      >
+        <span>
+          <mat-icon
+            *ngIf="selectedGroupBy.key === GroupByKey.RUN"
+            svgIcon="done_24px"
+          ></mat-icon>
+        </span>
+        <label>Runs</label>
```
japie1235813, line 61: same as `Experiments` 
Want to see if we want to use `runs` instead of `run` from now on

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -572,11 +575,86 @@ describe('runs_table', () => {
...
-          'Experiments',
-          'Runs',
-          'Regex',
-        ]);
+        expect(
+          items.map((element) => element.querySelector('label')!.textContent)
+        ).toEqual(['Experiments', 'Runs', 'Regex']);
+      });
+
+      it('renders a check for a menu item that has groupBy', () => {
```
japie1235813, line 583: Can we add `icon`? like `a check icon` 
I was confused and think about other check

`renders a check icon for the current groupBy menu item`

japie1235813 (approved):

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
stephanwlee: Done. Yeah, my text does look confusing :). Thanks.

PR #4987, opened 2021-05-19
timeseries: make image support conditional

psybuzz:
Adds a new feature flag to allow disabling support for showing
images in the Time Series dashboard.

Manually checked that running `bazel tensorboard:dev` still
shows the Image cards and settings, and that setting 
`metricsImageSupportEnabled: false` in
'feature_flag_store_config_provider.ts' results in Image cards
and settings being hidden from the UI.

image


bmd3k:
> and that setting
> `metricsImageSupportEnabled: true` in
> 'feature_flag_store_config_provider.ts' results in Image cards
> and settings being hidden from the UI.

Just to double check, did you mean `metricsImageSupportEnabled: false`?

tensorboard/webapp/metrics/data_source/metrics_data_source.ts
```diff
@@ -179,8 +188,26 @@ export class TBMetricsDataSource implements MetricsDataSource {
         })
       );
     });
+    const isImagesSupported$ = this.store.select(getIsFeatureFlagsLoaded).pipe(
```
bmd3k, line 191: I think what you've written here using both getIsFeatureFlagsLoaded and getIsMetricsImageSupportEnabled is most correct and future proof, but a few thoughts:

* The way feature flag values are currently loaded, we have been able to assume elsewhere in the code that it is safe to access feature flag selectors without first checking getIsFeatureFlagsLoaded. But this only holds true if (1) FeatureFlagModule is properly imported before your code runs and (2) fetch of feature flag values occurs synchronously when FeatureFlagEffects is loaded (which it does right now). 
  * https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts;l=39;drc=9b29bb2706e5443052ac5fefacab742dfda11192).
* I hadn't really thought before about why this worked and why it might not work in the future. Again I think what you've written here is probably the most correct and most future proof.
* On the other hand, what you've written here is kind of verbose and its a shame that accessing a feature flag requires this sort of verbosity.
* Would it be worth adding some sort of helper function that we can use to access FeatureFlag values that hides this complexity?

  ```
  export function getFeatureFlag(store : State<FeatureFlagState>, featureFlagSelector : ??) 
      : Observable<boolean>
  ```

  The observable it returns would only emit a single value.

I'm not suggesting that you do that in this PR or that you do it at all. I mostly want to know if you think this chain of thought is agreeable. If so I may volunteer to do this cleanup, time permitting.

tensorboard/webapp/metrics/data_source/metrics_data_source_test.ts
```diff
@@ -28,15 +34,23 @@ import {MetricsDataSource, PluginType} from './types';
...
+      providers: [
+        {provide: MetricsDataSource, useClass: TBMetricsDataSource},
+        provideMockStore(),
+      ],
     }).compileComponents();
 
     httpMock = TestBed.inject(HttpTestingController);
     dataSource = TestBed.inject(MetricsDataSource);
+    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
+    store.overrideSelector(selectors.getFeatureFlags, buildFeatureFlag());
```
bmd3k, line 51: Do you need to override this selector given that you've overriden the two selectors you are directly using?

tensorboard/webapp/metrics/metrics_module.ts
```diff
@@ -77,6 +77,8 @@ export function alertActionProvider() {
     ),
     EffectsModule.forFeature([MetricsEffects]),
     AlertActionModule.registerAlertActions(alertActionProvider),
+    AppRoutingModule,
```
bmd3k, line 80: My intuition is that it is safer to import these modules before creating your MetricsEffects. We've seen at least one subtle bug related to the order in which effects are registered (#4952). By importing the dependent modules, first, you allow them to assume that they will be the first to handle any actions that they fire.

tensorboard/webapp/metrics/views/metrics_views_module.ts
```diff
@@ -24,6 +25,12 @@ import {RightPaneModule} from './right_pane/right_pane_module';
 @NgModule({
   declarations: [MetricsDashboardContainer],
   exports: [MetricsDashboardContainer],
-  imports: [CommonModule, RightPaneModule, RunsSelectorModule, MainViewModule],
+  imports: [
+    CommonModule,
+    RightPaneModule,
+    RunsSelectorModule,
+    MainViewModule,
+    FeatureFlagModule,
```
bmd3k, line 33: Given that the only view code that seems to use feature flags is right-pane-related and that RightPaneModule imports FeatureFlagModule, should we actually modify MetricsViewsModule to also import FeatureFlagsModule?

tensorboard/webapp/metrics/views/right_pane/settings_view_container.ts
```diff
@@ -66,6 +67,10 @@ import {HistogramMode, TooltipSort, XAxisType} from '../../types';
 export class SettingsViewContainer {
   constructor(private readonly store: Store<State>) {}
 
+  readonly isImageSupportEnabled$ = this.store.select(
```
bmd3k, line 70: After writing that long comment about how you are correct to use getFeatureFlagsIsLoaded in tandem with feature flag selectors, I only now see you've done the opposite here.

Should we also use getFeatureFlagsIsLoaded here? Or is there a reason not to do it here that doesn't apply to MetricsDataSource?

bmd3k (approved):

psybuzz (commented):
> Just to double check, did you mean metricsImageSupportEnabled: false?

Whoops, yes, that is what I intended.  Updated the description.
Thanks for the review!

PR #4988, opened 2021-05-19
hparams: fix the parallel coordinate layout

### LINKED ISSUE #3414: hparams parallel coordinate view should adjust width for number of axes
### Issue

If we have a large number of hparams and metrics (= a large number of axes) in the parallel coordinate view we obviously need a wide plot pane to show all the axes in a meaningful way. Note that here we have more than ten axes that are still very well visible.

grafik

But if you then disable some of them, the spacing between the remaining axes just gets stretched, and if there are sufficiently many they get stretched beyond a usable distance. Compared to the previoius screenshot we now only have five axes selected, but only three of them are visible. But as we saw there is room for at least ten of them.

grafik

### Suggestion
My suggestion is that that the spacing between the axes should remain constant no matter how many parameters there are. Optionally one could introduce a slider/spinbox to adjust the spacing to your needs.

To replicate this issue you can use following python code to generate a summary file (needs pytorch),

```
from torch.utils.tensorboard import SummaryWriter
tb = SummaryWriter()
tb.add_hparams({"hparam"+str(k): k for k in range(10)}, {"metric"+str(k): k for k in range(10)})
```

then start tensorboard with `tensorboard --logdir=runs` and open `localhost:6006` in your browser.

stephanwlee:
Bug
---
Previously, we were setting the width of the container based on count of
all hparams and metrics without considering the visibility. This had
caused the UI to be in very awkward state where each visible hparams and
metrics are spaced a lot even if the selection count is quite low.

Fix
---
We layout the parallel coordinates based on number of visible
{hparam,metric}s. The change is a bit larger because:

- I had to add type information and refactor it a little for
  understanding the code (minimally added the types).
- Undo the performance optimization where we do not mutate the layout
  unless important piece of information in the configuration has
  changed. To detect the old configuration vs. new configuration, we
  have to now remember an additional state in the component.

Fixes #3414.

Previous:
image

New:
image

### DIFF
--- a/tensorboard/plugins/hparams/tf_hparams_parallel_coords_plot/BUILD
+++ b/tensorboard/plugins/hparams/tf_hparams_parallel_coords_plot/BUILD
@@ -17,8 +17,8 @@ tf_ts_library(
     deps = [
         "//tensorboard/components/polymer:legacy_element_mixin",
         "//tensorboard/plugins/hparams:types",
-        "//tensorboard/plugins/hparams/tf_hparams_query_pane:schema_d_ts",
         "//tensorboard/plugins/hparams/tf_hparams_session_group_values",
+        "//tensorboard/plugins/hparams/tf_hparams_types",
         "//tensorboard/plugins/hparams/tf_hparams_utils",
         "@npm//@polymer/decorators",
         "@npm//@polymer/polymer",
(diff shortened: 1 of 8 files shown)


psybuzz (approved):
Thanks for adding type interfaces too!

mhamilton723:
@stephanwlee @psybuzz is this released yet im having these issues with 2.5.0. Thanks!

psybuzz:
Thanks for the ping.  The fix will be part of the upcoming TensorBoard 2.6.0 (and TensorFlow 2.6.0) release, so please stay tuned.  The TB cadence has roughly matched the TensorFlow release cadence, so the ETA is soon (on the order of days/weeks from now, not months).

If you're in need of this fix immediately, it should be available in https://pypi.org/project/tb-nightly/ right now.

PR #4989, opened 2021-05-19
Move constant creation in debugger_v2_plugin_test

allenlavoie:
Prepares for a TensorFlow change which will make tf.constant run an op, which would otherwise have upset the assertions in this test.

### DIFF
--- a/tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
+++ b/tensorboard/plugins/debugger_v2/debugger_v2_plugin_test.py
@@ -57,6 +57,7 @@ def _generate_tfdbg_v2_data(
       logarithm_times: Optionally take logarithm of the final `x` file _ times
         iteratively, in order to produce nans.
     """
+    x = tf.constant([1, 3, 3, 7], dtype=tf.float32)
     writer = tf.debugging.experimental.enable_dump_debug_info(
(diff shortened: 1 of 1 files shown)


stephanwlee (commented):
Thanks for the PR Allen. Few dumb questions (code looks innocuous, just want to understand it more for my sake):

AFAICT, the `x = tf.constant` was declared outside of `@tf.function` already and is run in pure Eager context. Was it an issue because it was created after a writer/Resource was created? If so, what is the nuance?

allenlavoie:
Right, it's just a question of eager `tf.constant` running before or after the `enable_dump_debug_info` line. Right now tf.constant doesn't run an op, but I have a change pending which does run an op. And if that op ends up in the debug info, the existing assertions which check things like the number of executions break (arguably that means they're too fragile, but it'd be pretty hard to write thorough tests for debug output without relying on that implementation detail).

Moving the constant before the debug info dump starts means that the tests are unaffected by `tf.constant` running an op. So for now it's a no-op, but in the future it will prevent the test from breaking (assuming my tf.constant change goes in).

The TF change is only for eager tf.constant; inside a tf.function constants are already accompanied by an op.

stephanwlee (approved):
Got it; I forgot that `enable_dump_debug_info` alters behavior with its `op_callback` and `function_callback`. Thanks for the explanation, Allen.

PR #4991, opened 2021-05-19
timeseries: add data source for persisting setting

stephanwlee:
Background
----------
Currently, TensorBoard forgets settings (at least on the Angular side)
when browser refreshes and even for the ones that are persisted in the
URLs do not carry over to other experiments or other TensorBoard browser
tabs.

This change
-----------
This change is a part of the greater effort to mitigate the usability
issue by persisted few settings (for now) to LocalStorage. Do note that
this is not a panacea as users can change port number and the states
will be forgotten.

The DataSource API is created with hosted services in mind where we may
add a HTTP API or DataProvider API for persisting user settings.

### DIFF
--- a/tensorboard/webapp/metrics/data_source/metrics_data_source.ts
+++ b/tensorboard/webapp/metrics/data_source/metrics_data_source.ts
@@ -372,9 +372,16 @@ export class TBMetricsDataSource implements MetricsDataSource {
           value = TooltipSort.NEAREST;
           break;
         default:
-        // Deliberately fallthrough; may have TooltipSort from a newer version
-        // of TensorBoard where there is an enum that this version is unaware
-        // of.
+          if (TooltipSort[unsanitizedObject.tooltipSort]) {
+            const _ = unsanitizedObject.tooltipSort as never;
+            throw new RangeError(
+              `TooltipSort persistence case not handled properly: ${unsanitizedObject.tooltipSort}`
+            );
+          } else {
+            // Deliberately fallthrough; may have TooltipSort from a newer version
+            // of TensorBoard where there is an enum that this version is unaware
+            // of.
+          }
       }
       if (value !== null) {
         settings.tooltipSort = value;
--- a/tensorboard/webapp/metrics/data_source/metrics_data_source.ts
+++ b/tensorboard/webapp/metrics/data_source/metrics_data_source.ts
@@ -13,10 +13,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {Injectable} from '@angular/core';
-import {TBHttpClient} from '../../webapp_data_source/tb_http_client';
-import {forkJoin, Observable} from 'rxjs';
-import {map} from 'rxjs/operators';
+import {forkJoin, Observable, of} from 'rxjs';
+import {map, tap} from 'rxjs/operators';
 
+import {LocalStorage} from '../../util/local_storage';
+import {TBHttpClient} from '../../webapp_data_source/tb_http_client';
+import {TooltipSort} from '../internal_types';
 import {
   BackendTagMetadata,
   BackendTimeSeriesRequest,
@@ -37,9 +39,21 @@ import {
   TagToRunSampledInfo,
   TimeSeriesRequest,
   TimeSeriesResponse,
+  PersistableSettings,
 } from './types';
 
 const HTTP_PATH_PREFIX = 'data/plugin/timeseries';
+const LOCAL_STORAGE_KEY = '_tb_global_settings';
+
+/**
+ * `declare` so it does not get mangled or mangled differently when
+ * compiler changes.
+ */
+declare interface SerializableSettings {
+  scalarSmoothing?: number;
+  tooltipSort?: string;
+  ignoreOutliers?: boolean;
+}
 
 function parseRunId(runId: string): {run: string; experimentId: string} {
   const slashIndex = runId.indexOf('/');
@@ -168,7 +182,10 @@ function buildCombinedTagMetadata(results: TagMetadata[]): TagMetadata {
  */
 @Injectable()
 export class TBMetricsDataSource implements MetricsDataSource {
-  constructor(private readonly http: TBHttpClient) {}
+  constructor(
+    private readonly http: TBHttpClient,
+    private readonly localStorage: LocalStorage
+  ) {}
 
   fetchTagMetadata(experimentIds: string[]) {
     const fetches = experimentIds.map((experimentId) => {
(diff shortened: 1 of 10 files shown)


tensorboard/webapp/metrics/data_source/metrics_data_source.ts
```diff
@@ -296,4 +313,95 @@ export class TBMetricsDataSource implements MetricsDataSource {
...
+        case TooltipSort.DEFAULT:
+          value = TooltipSort.DEFAULT;
+          break;
+        case TooltipSort.NEAREST:
+          value = TooltipSort.NEAREST;
+          break;
+        default:
+        // Deliberately fallthrough; may have TooltipSort from a newer version
+        // of TensorBoard where there is an enum that this version is unaware
+        // of.
```
psybuzz, line 377: When a new version of TB adds onto the TooltipSort enum, the author might not realize they need to update this function.  If keeping sanitization is desired, let's add a comment at the `TooltipSort` definition, mentioning that if it changes, then we should also change this file's function.

tensorboard/webapp/metrics/internal_types.ts
```diff
@@ -14,10 +14,14 @@ limitations under the License.
 ==============================================================================*/
 import {HistogramMode} from '../widgets/histogram/histogram_types';
 
-import {PluginType} from './data_source/types';
-
 export {HistogramMode};
 
+export enum PluginType {
```
psybuzz, line 19: Just checking, was there a reason why the `PluginType` definition moved from data_source/types to internal_types?  I don't see anything in this PR that requires it to move.

tensorboard/webapp/metrics/data_source/metrics_data_source.ts
```diff
@@ -37,9 +39,21 @@ import {
...
 } from './types';
 
 const HTTP_PATH_PREFIX = 'data/plugin/timeseries';
+const LOCAL_STORAGE_KEY = '_tb_global_settings';
+
+/**
+ * `declare` so it does not get mangled or mangled differently when
+ * compiler changes.
+ */
+declare interface SerializableSettings {
```
psybuzz, line 52: Not sure if this TS idea works out, but if the entire purpose of having these is to make TS check serializability, could we drop `SerializableSettings` and `serializeSettings` and just ensure `PersistableSettings` is serializable via `PersistableSettings extend Record<string, string|number|boolean>`?

tensorboard/webapp/metrics/data_source/metrics_data_source_test.ts
```diff
@@ -336,4 +343,131 @@ describe('TBMetricsDataSource test', () => {
...
+        TEST_ONLY.LOCAL_STORAGE_KEY,
+        '{"ignoreOutliers":false,"scalarSmoothing":0.3,' +
+          '"tooltipSort":"ascending"}'
+      );
+    });
+
+    it('sets NaN as `null` even if that is in the settings', () => {
+      dataSource
+        .setSettings({
+          scalarSmoothing: NaN,
```
psybuzz, line 416: Interesting, I didn't know this about the JSON serializer.

psybuzz (commented):
One purely non-blocking high level question: if we want any other feature outside of metrics/ to save settings, e.g. autoReload, autoReloadDuration, it looks like we will have to duplicate much of the prod code and tests in `metrics_data_source.ts`.  There seems to be nothing metrics-specific besides the `PersistableSettings` type properties and sanitization.  Is making a reusable SettingsDataSource a non-goal?

tensorboard/webapp/metrics/internal_types.ts
stephanwlee: It creates a circular dep before.

- InternalTypes requires to read PluginType from `data_source/types`
- `data_source/types` require to read `TooltipSort` enum from `InternalTypes`.

stephanwlee:
> One purely non-blocking high level question: if we want any other feature outside of metrics/ to save settings, e.g. autoReload, autoReloadDuration, it looks like we will have to duplicate much of the prod code and tests in metrics_data_source.ts. There seems to be nothing metrics-specific besides the PersistableSettings type properties and sanitization. Is making a reusable SettingsDataSource a non-goal?

While that's true, I am weary of it for... one reason: multiple TB services. If we move it to SettingsModule, it would be reading states from other features and create Data dependency between `Metrics -> Settings`. If we continue down this path, it would create more coupling and I am generally weary of that (at least for now since not all services has TimeSeries). 

All in all, yes, DataSource _can_ move and we can make the `DataSource` shareable while having each respective features' effect use the aforementioned DataSource, but we did not yet have a case when a data source from other features injected into others (with exception of `tensorboard/webapp/data_sources`). This may be something we can consider when we truly want to apply it to other features.

tensorboard/webapp/metrics/data_source/metrics_data_source.ts
stephanwlee: > just ensure PersistableSettings is serializable via PersistableSettings extend Record<string, string|number|boolean>

I prefer it this way; something that is to be persisted should have its own interface while the data structure that weaves through internals have another one. Sure, we can make it congruent but, imo, that's exactly in the realm of "let's dedup just because they happen to be the same".

tensorboard/webapp/util/local_storage.ts
```diff
@@ -0,0 +1,35 @@
...
+import {Injectable, NgModule} from '@angular/core';
+
+@Injectable()
+export class LocalStorage {
+  setItem(key: string, value: string): void {
+    localStorage.setItem(key, value);
+  }
+
+  getItem(key: string): string {
+    return localStorage.getItem(key) ?? '';
```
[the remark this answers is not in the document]
stephanwlee: Done.

tensorboard/webapp/metrics/data_source/metrics_data_source.ts
```diff
@@ -37,9 +39,21 @@ import {
   TagToRunSampledInfo,
   TimeSeriesRequest,
   TimeSeriesResponse,
+  PersistableSettings,
 } from './types';
 
 const HTTP_PATH_PREFIX = 'data/plugin/timeseries';
+const LOCAL_STORAGE_KEY = '_tb_global_settings';
```
psybuzz, line 46: Given the stated concern about coupling, should we namespace the key so that it is specific to timeseries?  e.g. `LOCAL_STORAGE_KEY = '_tb_global_settings.timeseries'` so that in the future, another feature's setting named `ignoreOutliers` with a very different meaning does not collide.  Alternatively, we could make the setting names specific enough that they should never collide e.g. "timeseriesScalarSmoothing", "timeseriesTooltipSort".

psybuzz (approved):
Exciting!

PR #4994, opened 2021-05-20
timeseries: persist and read settings

stephanwlee:
This is a continuation of #4991 where we are attempting to persist
several settings from TimeSeries to improve frustrating user experience
of having to reset settings over and over.

This change integrates the DataSource with the MetricsEffects so persist
new settings when it changes and bootstrap the application state from
the state in the LocalStorage.

Product choice: this change persist the smoothing value user may provide
in the URL since it felt more natural that way.

### DIFF
--- a/tensorboard/webapp/metrics/effects/metrics_effects_test.ts
+++ b/tensorboard/webapp/metrics/effects/metrics_effects_test.ts
@@ -21,7 +21,7 @@ import {getActivePlugin} from '../../core/store';
 import * as coreTesting from '../../core/testing';
 import {DataLoadState} from '../../types/data';
 import {TBHttpClientTestingModule} from '../../webapp_data_source/tb_http_client_testing';
-import {of, Subject} from 'rxjs';
+import {EMPTY, of, Subject} from 'rxjs';
 
 import {buildNavigatedAction} from '../../app_routing/testing';
 import {State} from '../../app_state';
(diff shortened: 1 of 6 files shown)


stephanwlee:
To review: there was a large amount of white space change (indentation) so I would recommend turning on "Hide whitespace changes" option.

image

tensorboard/webapp/metrics/effects/metrics_effects_test.ts
```diff
@@ -21,7 +21,7 @@ import {getActivePlugin} from '../../core/store';
 import * as coreTesting from '../../core/testing';
 import {DataLoadState} from '../../types/data';
 import {TBHttpClientTestingModule} from '../../webapp_data_source/tb_http_client_testing';
-import {of, Subject} from 'rxjs';
+import {EMPTY, of, Subject} from 'rxjs';
```
bmd3k, line 24: There appears to be some significant refactoring in this CL mixed in with the new tests. Can we split the preliminary refactoring to a different PR?

stephanwlee: Would splitting them into two separate commits help you review the PR? Also, did you turn on the "Hide whitespace changes" option on the GitHub? It is not a "refactor" and more of an indentation change :\

bmd3k: Thanks for the tip. Indeed it was just whitespace. I totally missed your earlier comment about it, too.

bmd3k (approved):

PR #4996, opened 2021-05-20
docs: update audio demo to TF2

psybuzz:
Updates the audio_demo
- TF1 to TF2 style
- More minimal

Test plan
Manually ran the demo with `bazel run //tensorboard/plugins/audio:audio_demo`, opened TensorBoard and saw that the audios appeared properly.

image

### DIFF
--- a/tensorboard/plugins/audio/audio_demo.py
+++ b/tensorboard/plugins/audio/audio_demo.py
@@ -14,7 +14,6 @@
 # ==============================================================================
 """Sample data exhibiting audio summaries, via a waveform generator."""
 
-
 import inspect
 import math
 import os.path
@@ -22,7 +21,6 @@ import os.path
 from absl import app
 from absl import flags
 import tensorflow as tf
-from tensorboard.plugins.audio import summary
 
 FLAGS = flags.FLAGS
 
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/audio/audio_demo.py
```diff
@@ -166,116 +102,33 @@ def square_wave(frequency):
...
-    #
-    # This is precisely the desired result, so we're done!
-    return amplitude
-
-
-# If we want to get fancy, we can use our above waves as primitives to
-# build more interesting waves.
-
-
-def bisine_wave(frequency):
```
nfelt, line 207: I might consider keeping this one around too just since it demonstrates audio with 2 channels (i.e. stereo), and it's not very complicated.  (We should be able to drop the name scopes.)

Maybe also worth having this one return two different clips (or as the UI shows, "samples" - in the `max_outputs` sense, not the `sample_rate` sense, which is rather confusing for the audio context...) just to show that this is also possible, i.e. you could return something like this:

```python
tf.concat([tf.concat([sine_lo, sine_hi], axis=2), tf.concat([sine_hi, sine_lo], axis=2)], axis=0)
```

It looks like the bisine_wahwah_wave one does this, so that way we can still show that aspect of the functionality with fewer lines of code to maintain.

nfelt (approved):
Thanks!

tensorboard/plugins/audio/audio_demo.py
psybuzz: Done.

Good idea to have an example showing 2 clips.

psybuzz (commented):
Thanks!

PR #4997, opened 2021-05-20
docs: update text demo

psybuzz:
Minor update to the text demo to match how users would
write text summaries.

Manually tested by running `bazel run //tensorboard/plugins/text:text_demo` and checked that the demo still works.

image


bileschi (approved):
LGTM!

PR #4999, opened 2021-05-20
docs: update mesh demo

psybuzz:
Merges `mesh_demo.py` and `mesh_demo_v2.py` into a single demo,
dropping the TF1 version.

Manually tested by running `bazel run //tensorboard/plugins/mesh:mesh_demo`
and opening the Mesh dashboard and seeing that it looks unchanged.
image


tensorboard/plugins/mesh/mesh_demo.py
```diff
@@ -20,27 +20,38 @@
...
+    )
+
+
 def run():
-    """Runs session with a mesh summary."""
-    # Mesh summaries only work on TensorFlow 1.x.
-    if int(tf.__version__.split(".")[0]) > 1:
-        raise ImportError("TensorFlow 1.x is required to run this demo.")
+    """Runs training steps with a mesh summary."""
     # Flag mesh_path is required.
```
nfelt, line 55: Optional since this isn't really TF 2.0 related, but it would be nice if this was self-contained. As reported by @wchargin (oh hai) on #2578 there is a .ply file in-repo that we could bundle to make this simpler to run:

https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/mesh/test_data/icosphere.ply

nfelt (approved):
Thanks! CI failed due to needing a `BUILD` file change too I believe.

I guess this means #2578 is officially stale, huh :P

tensorboard/plugins/mesh/mesh_demo.py
psybuzz: Done.

PR #5004, opened 2021-05-21
Update images_demo to be TF2-friendly, and simpler.

ericdnielsen:
Update the images demos to ensure they are TF2 (rather than TF1) APIs.  

Additionally simplify the demo so its easier to understand and only highlight how to use the summary writer.  For a deeper investigation/instruction in how to use images in an ML system, refer to the images tutorial: https://www.tensorflow.org/tensorboard/image_summaries

The embedded plugins/<plugin_name>_demo.py are intended primary to create representative data for testing dashboards rather than for being canonical ML pipelines.

Reference image of the new demo output:
image

### DIFF
--- a/tensorboard/plugins/image/BUILD
+++ b/tensorboard/plugins/image/BUILD
@@ -3,20 +3,20 @@
 
 load("//tensorboard/defs:protos.bzl", "tb_proto_library")
 
-package(default_visibility = ["//tensorboard:internal"])
+package(default_visibility=["//tensorboard:internal"])
 
 licenses(["notice"])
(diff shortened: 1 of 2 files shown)


ericdnielsen:
Running black over the modifeid files appears to have completely reformatted the BUILD file.  Let me know if you want that reverted.

nfelt (approved):
Overall LGTM, but just for reference would you mind including a screenshot of the resulting images (e.g. in the PR description)?

ericdnielsen:
Added the image.

PR #5006, opened 2021-05-21
docs: update graphs_demo.py

yatbear:
Update *graphs_demo* as part of the TF 2.0 migration doc fixit efforts.
- Replace the `tf.compat.v1.summary.TaggedRunMetadata` demo with a graph writing demo. 
- Suppress deprecation warnings.
- Add more comments and instruction logs.

Screenshots of UI changes:
- `tagged` is changed to `write_graph`:
image

- `tagged` run is now replaced with `write_graph` demo that visualizes a power function graph.
image


tensorboard/plugins/graph/graphs_demo.py
```diff
@@ -129,5 +128,18 @@ def g(i):
             tf.summary.trace_export("prof_g", step=step)
 
 
+def main():
+    # Create three demo graphs.
+    with _silence_deprecation_warnings():
```
nfelt, line 133: Hmm, could we scope the silence a little more narrowly to just the lines that require it?  That way we can still notice if there are new deprecation warnings on other symbols - the goal here should just be silence warnings that are unavoidable and known (like the ones about the profiler that come from the `tf.summary.trace_*` lines specifically).

tensorboard/plugins/graph/graphs_demo.py
```diff
@@ -19,47 +19,46 @@
...
+    logdir = os.path.join(LOGDIR, "write_graph")
+
+    @tf.function
+    def f():
+        x = tf.constant(2)
+        y = tf.constant(3)
+        return x ** y
+
+    with tf.summary.create_file_writer(logdir).as_default():
+        if hasattr(tf.summary, "graph"):
```
nfelt, line 59: Let's add some kind of comment or explanation for the `hasattr` - e.g. `# tf.summary.graph requires TF 2.5.0+` - or maybe even better, we can add an else condition like

```python
else:
    print("Could not find tf.summary.graph(); use TF 2.5.0+ to run full demo")
```

nfelt (approved):

nfelt (approved):
Thanks! LGTM

PR #5009, opened 2021-05-21
docs: update scalars demo

bileschi:
Updates scalars demo to use TF2 conventions.

Somewhat of a major update.

Manually tested by running `bazel run tensorboard/plugins/scalar:scalars_demo` and checked that the demo works and produces standard visual output.
Screen Shot 2021-05-21 at 6 11 34 PM

### DIFF
--- a/tensorboard/plugins/scalar/scalars_demo.py
+++ b/tensorboard/plugins/scalar/scalars_demo.py
@@ -42,24 +42,25 @@ def one_step(temperature, ambient_temperature, heat_coefficient, step):
       step: tf.int64 value of the current step number.  Used in the
         `summary.scalar` API.
     """
-    tf.summary.scalar(
-        name="current",
-        data=temperature,
-        step=step,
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/scalar/scalars_demo.py
```diff
@@ -133,27 +114,46 @@ def run_all(logdir, verbose=False):
...
-                    run_name,
-                    initial_temperature,
-                    final_temperature,
-                    heat_coefficient,
+                writer = tf.summary.create_file_writer(
+                    os.path.join(logdir, run_name)
                 )
+                with writer.as_default():
+                    run(
+                        tf.constant(initial_temperature),
```
nfelt, line 131: Maybe worth a comment about why the `tf.constant()`s are here? I.e. to prevent unwanted retracing. We could also just specify `input_signature` on the `@tf.function` decorator which might be a little more foolproof.

nfelt (approved):
Thanks!

PR #5014, opened 2021-05-24
docs: update histogram demo to TF2

japie1235813:
Update TF1 to TF2 style

Test plan:
Manually ran the demo with `bazel run //tensorboard/plugins/histrogram:histrogram_demo`
Screen Shot 2021-05-24 at 1 29 26 PM

### DIFF
--- a/tensorboard/plugins/histogram/histograms_demo.py
+++ b/tensorboard/plugins/histogram/histograms_demo.py
@@ -18,7 +18,6 @@
 
 from absl import app
 import tensorflow as tf
-import os
 
 # Directory into which to write tensorboard data.
 LOGDIR = "/tmp/histograms_demo"
--- a/tensorboard/plugins/histogram/histograms_demo.py
+++ b/tensorboard/plugins/histogram/histograms_demo.py
@@ -23,6 +23,7 @@ import os
 # Directory into which to write tensorboard data.
 LOGDIR = "/tmp/histograms_demo"
 
+
 def run(k):
     # Make a normal distribution, with a shifting mean
     mean_moving_normal = tf.random.normal(shape=[1000], mean=(5 * k), stddev=1)
@@ -99,16 +100,18 @@ def run(k):
         "two normal distributions.",
     )
 
+
 def run_all(logdir, num_summaries=400):
     """Generate a bunch of histogram data, and write it to logdir."""
     tf.random.set_seed(0)
-    writer = tf.summary.create_file_writer(os.path.join(logdir, 'my_run_name'))
+    writer = tf.summary.create_file_writer(os.path.join(logdir, "my_run_name"))
     with writer.as_default():
         for step in range(num_summaries):
             tf.summary.experimental.set_step(step)
             k = step / float(num_summaries)
             run(k)
             writer.flush()
+
 
 def main(unused_argv):
     print("Running histograms demo. Output saving to %s." % LOGDIR)
--- a/tensorboard/plugins/histogram/histograms_demo.py
+++ b/tensorboard/plugins/histogram/histograms_demo.py
@@ -18,12 +18,12 @@
 
 from absl import app
 import tensorflow as tf
+import os
 
 # Directory into which to write tensorboard data.
 LOGDIR = "/tmp/histograms_demo"
 
 def run(k):
-    tf.random.set_seed(0)
     # Make a normal distribution, with a shifting mean
     mean_moving_normal = tf.random.normal(shape=[1000], mean=(5 * k), stddev=1)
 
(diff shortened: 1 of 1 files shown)


bileschi (approved):
LGTM, thanks!

tensorboard/plugins/histogram/histograms_demo.py
```diff
@@ -107,18 +100,17 @@ def run_all(logdir, verbose=False, num_summaries=400):
...
+    writer = tf.summary.create_file_writer(os.path.join(logdir, "my_run_name"))
+    with writer.as_default():
+        for step in range(num_summaries):
+            tf.summary.experimental.set_step(step)
+            k = step / float(num_summaries)
+            run(k)
+            writer.flush()
 
 
 def main(unused_argv):
```
psybuzz, line 116: Could we add a print statement that informs users about TensorBoard / TensorBoard.dev?

```
print(
        "To view results in your browser, run `tensorboard --logdir %s`"
        % LOGDIR
    )
    print(
        "Logs can be uploaded publicly to TensorBoard.dev via "
        + "`tensorboard dev upload --logdir %s`" % LOGDIR
    )
```

tensorboard/plugins/histogram/histograms_demo.py
```diff
@@ -107,18 +100,17 @@ def run_all(logdir, verbose=False, num_summaries=400):
...
-        k_val = step / float(N)
-        summ = sess.run(summaries, feed_dict={k: k_val})
-        writer.add_summary(summ, global_step=step)
+def run_all(logdir, num_summaries=400):
+    """Generate a bunch of histogram data, and write it to logdir."""
+    tf.random.set_seed(0)
+    writer = tf.summary.create_file_writer(os.path.join(logdir, "my_run_name"))
+    with writer.as_default():
+        for step in range(num_summaries):
+            tf.summary.experimental.set_step(step)
```
psybuzz, line 110: I think we are able to avoid using the "experimental" pieces of the API in this demo case.  Could we drop this experimental.set_step, and instead pass the `step` into the `run(k, step)`?  Then, each `tf.summary.histogram(..., step=step)` can specify the step without relying on the global `set_step`.

tensorboard/plugins/histogram/histograms_demo.py
```diff
@@ -18,25 +18,18 @@
...
 
-def run_all(logdir, verbose=False, num_summaries=400):
-    """Generate a bunch of histogram data, and write it to logdir."""
-    del verbose
-
-    tf.compat.v1.set_random_seed(0)
-
-    k = tf.compat.v1.placeholder(tf.float32)
-
+def run(k):
```
psybuzz, line 27: nit: could we add a docstring that explains what the `k` arg refers to?

psybuzz (changes requested):

tensorboard/plugins/histogram/histograms_demo.py
nfelt: +1 that it's a bit better to avoid `experimental` in a demo.  What Erik suggests here should work fine.

japie1235813: Thanks for the review!

tensorboard/plugins/histogram/histograms_demo.py
japie1235813: Done. For this demo I addd `/my_run_name` to align with above settings

tensorboard/plugins/histogram/histograms_demo.py
japie1235813: Done. Thanks for the help!

psybuzz (approved):
Looks good!

tensorboard/plugins/histogram/histograms_demo.py
nfelt: Maybe we can just leave off `my_run_name` entirely, both here and in the actual writer construction?  I don't think it's really necessary since we have only a single run here, and omitting it would be more consistent with how this demo worked previously, as well as with the other demos (where it's just `/tmp/scalars_demo` for example).

tensorboard/plugins/histogram/histograms_demo.py
```diff
@@ -24,7 +24,11 @@
 LOGDIR = "/tmp/histograms_demo"
 
 
-def run(k):
+def run(k, step):
+    """
+    Arguments:
+        k: a float in the range [0, 1] that affects the histogram values written by the run.
```
nfelt, line 30: Minor, but could we document `step` here too?  If we're documenting any argument, I think it's best to just document all of them to avoid being inconsistent between the docs and the signature.

nfelt (approved):

PR #5016, opened 2021-05-25
color grouping: group runs by regex

japie1235813:
Color groups the runs by regex. No UI/UX involved.

- Match regex with name (not id)
- Construct the group key by the capture groups connected with '_'
- If there are no capture group, use the matched string as the group key
- if there are no capture group, use keyword `matches` to group all matched runs
- check invalid regex
- check empty input

### DIFF
--- a/tensorboard/webapp/runs/store/utils.ts
+++ b/tensorboard/webapp/runs/store/utils.ts
@@ -24,7 +24,7 @@ export function groupRuns(
   runs: Run[],
   runIdToExpId: Readonly<Record<RunId, ExperimentId>>
 ): {[groupId: string]: Run[]} {
-  const runGroups: {[groupId: string]: Run[]} = {};
+  let runGroups: {[groupId: string]: Run[]} = {};
   switch (groupBy.key) {
     case GroupByKey.RUN:
       for (const run of runs) {
@@ -41,7 +41,58 @@ export function groupRuns(
       break;
 
     case GroupByKey.REGEX:
-      throw new Error('Not implemented');
+      if (!groupBy.regexString) {
+        //TODO(japie1235813): props users when the input is invalid
+        throw new Error('Invalid regex');
+      }
+
+      // TODO(japie1235813): add additonal `\` to convert string to regex, which
+      // makes `new RegExp()` construct properly
+      // For example, convert `foo\d+bar` to `foo\\d+bar`
+
+      try {
+        new RegExp(groupBy.regexString);
+      } catch(e) {
+        //TODO(japie1235813): props users when the input is invalid
+        throw new Error('Invalid regex');
+      }
+
+      const regExp = new RegExp(groupBy.regexString);
+      // Checks if there is capture group in regex.
+      let isCaptureGroup = false;
+
+      for (const run of runs) {
+        let matches = (run.name).match(regExp)
+          if (matches) {
+          if(matches.length > 1) {
+            matches = matches.slice(1)
+            isCaptureGroup = true;
+          }
+          const id = matches.length === 1 ? matches[0] : matches.join('_');
+          const runs = runGroups[id] || [];
+          runs.push(run);
+          runGroups[id] = runs;
+        } else {
+          runGroups[run.id] = [run];
+        }
+      }
+
+      if (!isCaptureGroup) {
+        // No capture group in regex string. Groups all the matched runs together.
+        const matchedRuns = [];
+        runGroups = {};
+        for (const run of runs) {
+          let matches = (run.name).match(regExp)
+          if (matches) {
+            matchedRuns.push(run);
+            delete runGroups[run.id];
+          } else {
+            runGroups[run.id] = [run];
+          }
+        }
+        runGroups['matches'] = matchedRuns;
+      }
+      break;
     default:
   }
   return runGroups;
--- a/tensorboard/webapp/runs/store/utils.ts
+++ b/tensorboard/webapp/runs/store/utils.ts
@@ -43,7 +43,7 @@ export function groupRuns(
     case GroupByKey.REGEX:
       if (!groupBy.regexString) {
         //TODO(japie1235813): props users when the input is invalid
-        throw new Error('Invalid regex');
+        throw new Error('Empty regex string.');
       }
 
       // TODO(japie1235813): add additonal `\` to convert string to regex, which
@@ -52,9 +52,9 @@ export function groupRuns(
 
       try {
         new RegExp(groupBy.regexString);
-      } catch(e) {
+      } catch (e) {
         //TODO(japie1235813): props users when the input is invalid
-        throw new Error('Invalid regex');
+        throw new Error('Invalid regex.');
       }
 
       const regExp = new RegExp(groupBy.regexString);
@@ -62,10 +62,10 @@ export function groupRuns(
       let isCaptureGroup = false;
 
       for (const run of runs) {
-        let matches = (run.name).match(regExp)
-          if (matches) {
-          if(matches.length > 1) {
-            matches = matches.slice(1)
+        let matches = run.name.match(regExp);
+        if (matches) {
+          if (matches.length > 1) {
+            matches = matches.slice(1);
             isCaptureGroup = true;
           }
           const id = matches.length === 1 ? matches[0] : matches.join('_');
@@ -82,7 +82,7 @@ export function groupRuns(
         const matchedRuns = [];
         runGroups = {};
         for (const run of runs) {
-          let matches = (run.name).match(regExp)
+          let matches = run.name.match(regExp);
           if (matches) {
             matchedRuns.push(run);
             delete runGroups[run.id];
--- a/tensorboard/webapp/runs/store/utils.ts
+++ b/tensorboard/webapp/runs/store/utils.ts
@@ -42,8 +42,8 @@ export function groupRuns(
 
     case GroupByKey.REGEX:
       if (!groupBy.regexString) {
-        //TODO(japie1235813): props users when the input is invalid
-        throw new Error('Empty regex string.');
+        //TODO(japie1235813): props users the input is invalid
+        break;
       }
 
       // TODO(japie1235813): add additonal `\` to convert string to regex, which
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/runs/store/utils_test.ts
```diff
@@ -90,5 +90,138 @@ describe('run store utils test', () => {
...
+          matches: [
+            buildRun({id: 'eid1/alpha', name: 'foo1bar1'}),
+            buildRun({id: 'eid1/beta', name: 'foo1bar2'}),
+            buildRun({id: 'eid2/beta', name: 'foo2bar1'}),
+          ],
+          'eid2/gamma': [buildRun({id: 'eid2/gamma', name: 'gamma'})],
+        });
+      });
+
+      it('groups runs by regex with one capture group', () => {
```
stephanwlee, line 170: What do we do with non-matches?

tensorboard/webapp/runs/store/utils_test.ts
```diff
@@ -90,5 +90,138 @@ describe('run store utils test', () => {
...
+          {
+            'eid1/alpha': 'eid1',
+            'eid1/beta': 'eid1',
+            'eid2/beta': 'eid2',
+            'eid2/gamma': 'eid2',
+          }
+        );
+
+        expect(actual).toEqual({
+          matches: [
```
stephanwlee, line 161: What would happen if my tag name magically is `matches`? I think it is better to change the interface of `groupRuns` and return list of `matches` and `nonMatches`. For example,

```ts
export interface RunGroup {
  matches: Record<string, Run[]>;
  nonMatches: Run[];
}

export function groupRuns(...): RunGroup;
```

tensorboard/webapp/runs/store/utils.ts
```diff
@@ -41,7 +41,58 @@ export function groupRuns(
...
+      // TODO(japie1235813): add additonal `\` to convert string to regex, which
+      // makes `new RegExp()` construct properly
+      // For example, convert `foo\d+bar` to `foo\\d+bar`
+
+      try {
+        new RegExp(groupBy.regexString);
+      } catch (e) {
+        //TODO(japie1235813): props users when the input is invalid
+        throw new Error('Invalid regex.');
+      }
```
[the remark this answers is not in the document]
japie1235813: updated with `break;` in this ways the `group=groupRuns(..)`(line 224 `run_reducers,ts`) would be empty object and no keys to iterate, which then makes the state unchanged.

tensorboard/webapp/runs/store/utils.ts
```diff
@@ -41,7 +41,58 @@ export function groupRuns(
...
+
+      const regExp = new RegExp(groupBy.regexString);
```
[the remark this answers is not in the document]
japie1235813: looks good. updated.

tensorboard/webapp/runs/store/utils.ts
```diff
@@ -41,7 +41,58 @@ export function groupRuns(
...
     case GroupByKey.REGEX:
-      throw new Error('Not implemented');
+      if (!groupBy.regexString) {
+        //TODO(japie1235813): props users when the input is invalid
+        throw new Error('Empty regex string.');
+      }
+
+      // TODO(japie1235813): add additonal `\` to convert string to regex, which
+      // makes `new RegExp()` construct properly
+      // For example, convert `foo\d+bar` to `foo\\d+bar`
```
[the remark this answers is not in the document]
japie1235813: thanks for the reference!

tensorboard/webapp/runs/store/utils.ts
```diff
@@ -41,7 +41,58 @@ export function groupRuns(
...
+      let isCaptureGroup = false;
+
+      for (const run of runs) {
+        let matches = run.name.match(regExp);
+        if (matches) {
+          if (matches.length > 1) {
+            matches = matches.slice(1);
+            isCaptureGroup = true;
+          }
+          const id = matches.length === 1 ? matches[0] : matches.join('_');
```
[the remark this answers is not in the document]
japie1235813: thanks for pointing out `matches.join` case
although I don't see why it's a bad matches when the matches includes `'_'`, I agree that `JSON.stringify` is a better option. updated.

tensorboard/webapp/runs/store/utils_test.ts
```diff
@@ -90,5 +90,138 @@ describe('run store utils test', () => {
...
+              'eid1/alpha': 'eid1',
+              'eid1/beta': 'eid1',
+              'eid2/beta': 'eid2',
+              'eid2/gamma': 'eid2',
+            }
+          );
+        } catch (error) {
+          errorMessage = error.message;
+        }
+        expect(errorMessage).toBe('Empty regex string.');
```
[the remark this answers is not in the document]
japie1235813: noted. thanks! (although throw error is removed)

tensorboard/webapp/runs/store/utils_test.ts
japie1235813: I think you're talking about the case when the run.name is `matches` but the run is not a match to the regex?
in this case this run would be assigned to the group where  `run.id` is the key, not `matches`
anything I missed?

but anyway I updated the return type and, although feel it's more complicated, please let me know what do you think. thanks.

tensorboard/webapp/runs/store/utils.ts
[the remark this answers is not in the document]
stephanwlee: In case the regex string is something like `foo(_|)bar(_|)`, the matches for `foo_bar` and `foobar_` would be both:
- `foo_bar`: `["foo_bar", "_", ""]`
- `foobar_`: `["foo_bar", "", "_"]`

and,
`["_", ""].join("_") === "__"` and `["", "_"].join("_") === "__"` which would be faulty.

tensorboard/webapp/runs/store/utils_test.ts
```diff
@@ -90,5 +90,141 @@ describe('run store utils test', () => {
...
+            buildRun({id: 'eid2/gamma', name: 'gamma'}),
+          ],
+          {
+            'eid1/alpha': 'eid1',
+            'eid1/beta': 'eid1',
+            'eid2/beta': 'eid2',
+            'eid2/gamma': 'eid2',
+          }
+        );
+        expect(actual.matches).toEqual({});
```
stephanwlee, line 111: Change this to:

```ts
expect(actual).toEqual({
  matches: {},
  nonMatches: [],
});
```

and remove the commented line below.

tensorboard/webapp/runs/store/utils_test.ts
```diff
@@ -90,5 +90,141 @@ describe('run store utils test', () => {
         });
       });
     });
+
+    describe('by regex', () => {
+      it('do not change the group when the regex is empty', () => {
```
stephanwlee, line 95: `does not group when the regex is empty` (this method does not really change anything)

tensorboard/webapp/runs/store/utils.ts
```diff
@@ -41,8 +47,59 @@ export function groupRuns(
...
+            delete runGroups[run.id];
+          }
+        }
+        runGroups['matches'] = matchedRuns;
+      }
+      break;
     default:
   }
-  return runGroups;
+  runGroup.matches = runGroups;
```
stephanwlee, line 103: Remove this code by doing one of two things:

Option 1.
```ts
const matches: {[groupId: string]: Run[]} = {};
const nonMatches: Run[] = [];
const runGroup: RunGroup = {matches, nonMatches};
```

Option 2
```ts
...
// In L98
runGroups.matches['matches'] = matchedRuns;
...
```

tensorboard/webapp/runs/store/utils.ts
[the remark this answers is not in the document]
japie1235813: I see. thanks for the explanation.

tensorboard/webapp/runs/store/utils.ts
```diff
@@ -41,8 +47,59 @@ export function groupRuns(
...
+        for (const run of runs) {
+          let matches = run.name.match(regExp);
+          if (matches) {
+            matchedRuns.push(run);
+            delete runGroups[run.id];
+          }
+        }
+        runGroups['matches'] = matchedRuns;
+      }
+      break;
```
stephanwlee, line 100: Currently there are two for-loops for `runs` in case the regex string does not have the capture group. How about instead, we do:
```ts
for (const run of runs) {
  const matches = run.name.match(regExp);
  if (matches) {
    const hasCapturingGroup = matches.length > 1;
    // In case regex string does not have a capturing group, we use pseudo group id of `pseudo_group`.
    const id = hasCapturingGroup ? JSON.stringify(matches.slice(1)) : 'pseudo_group';
    const runs = runGroups[id] || [];
    runs.push(run);
    runGroups[id] = runs;
  } else {
    nonMatches.push(run);
  }
}
```

stephanwlee (commented):
Cool, looking better!

tensorboard/webapp/runs/store/utils.ts
```diff
@@ -23,26 +23,61 @@ export function groupRuns(
...
+        break;
+      }
+
+      for (const run of runs) {
+        const matchesList = run.name.match(regExp);
+        if (matchesList) {
+          const hasCapturingGroup = matchesList.length > 1;
+          // In case regex string does not have a capturing group, we use pseudo group id of `pseudo_group`.
+          const id = hasCapturingGroup
+            ? JSON.stringify(matchesList)
```
stephanwlee, line 70: You want to omit the matched string from the group id so we can only group based on content inside the capturing group. Do `JSON.stringify(matchesList.slice(1))`

tensorboard/webapp/runs/store/utils_test.ts
```diff
@@ -90,5 +90,150 @@ describe('run store utils test', () => {
...
+              buildRun({id: 'eid2/gamma', name: 'foo2bar3'}),
+            ],
+          },
+          nonMatches: [buildRun({id: 'eid2/alpha', name: 'alpha'})],
+        });
+      });
+
+      it('groups runs by regex with multiple capture group', () => {
+        const actual = groupRuns(
+          {key: GroupByKey.REGEX, regexString: 'foo(\\d+)bar(\\d+)'},
```
stephanwlee, line 204: Could you modify the test by changing this regex string to `foo(\\d+)bar(\\d+).*`

stephanwlee (approved):

PR #5017, opened 2021-05-25
runs: move html to its own file

japie1235813:
The html content is too long for embedded template. Separate them as another ng.html file


tensorboard/webapp/runs/views/runs_table/runs_group_menu_button.ng.html
```diff
@@ -0,0 +1,68 @@
+<!--
```
stephanwlee, line 1: Change the file name to `runs_group_menu_button_component.ng.html` to mirror that of TypeScript component file.

stephanwlee (approved):

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button.ng.html
japie1235813: Thanks for the catch. Updated.

PR #5018, opened 2021-05-25
runs: create groupBy override vs. initial value

stephanwlee:
When we work on our setting persistence, we want to be able to change
our default behavior. However, if we always persist the groupBy value in
the store to the URL, current business logic will be ingrained into
every dashboard URLs and its bookmark can lead to a case where we cannot
change our default behavior.

this change simply refactors so initial value of `groupBy` is separated
from the overridden value.


japie1235813 (approved):
Thanks for the pr. It's really clear and easy to review.

PR #5022, opened 2021-05-27
runs: persist colorBy in URL

stephanwlee:
This change adds the feature where if user's colorBy option is different
from the default value, it will be persisted up in the URL as
`?colorBy`. Current valid values are: `experiment`, `run`, and
`regex:<str>`.


tensorboard/webapp/routes/dashboard_deeplink_provider.ts
```diff
@@ -132,6 +158,22 @@ export class DashboardDeepLinkProvider extends DeepLinkProvider {
...
+            case 'experiment':
+              groupBy = {key: GroupByKey.EXPERIMENT};
+              break;
+            case 'run':
+              groupBy = {key: GroupByKey.RUN};
+              break;
+          }
+
+          if (value.startsWith('regex:')) {
+            const [, regexString] = value.split(':', 2);
```
japie1235813, line 172: if the regex is `foo\:bar`, the value would be `regex:foo\:bar`
but the regexString here becomes `foo`
why adding `regex:` prefix then extract here?

tensorboard/webapp/routes/dashboard_deeplink_provider_test.ts
```diff
@@ -104,6 +106,23 @@ describe('core deeplink provider', () => {
...
+        function assert(value: string, expectedGroupBy: GroupBy | null) {
+          const state = provider.deserializeQueryParams([
+            {key: 'colorGroup', value},
+          ]);
+          expect(state.runs.groupBy).toEqual(expectedGroupBy);
+        }
+        assert('experiment', {key: GroupByKey.EXPERIMENT});
+        assert('run', {key: GroupByKey.RUN});
+        assert('regex:', {key: GroupByKey.REGEX, regexString: ''});
+        assert('regex:hello', {key: GroupByKey.REGEX, regexString: 'hello'});
```
japie1235813, line 120: How about `'regex:hello:kitty'` ?

tensorboard/webapp/runs/types.ts
```diff
@@ -66,3 +66,12 @@ export interface RegexGroupBy {
 }
 
 export type GroupBy = BaseGroupBy | RegexGroupBy;
+
+/**
+ * The runs-related state created by deserializing a URL.
+ */
+export interface URLDeserializedState {
+  runs: {
+    groupBy: GroupBy | null;
```
japie1235813, line 75: Is this needed to be changed once we separate gorupBy with "groupByKey" and "regexString"?

stephanwlee: No, it does not need to change; public interface outside of the redux still can remain `GroupBy` without destructing it.

tensorboard/webapp/routes/dashboard_deeplink_provider.ts
stephanwlee: Ah, I misunderstood `split` API (thought it would behave like Python's `split`). 

> why adding regex: prefix then extract here?

We want to be able to distinguish different modes with the same query parameter, `?colorBy`. For example, 
```
?colorBy=experiment -> {key: GroupByKey.EXPERIMENT}
?colorBy=regex:<string> -> {key: GroupByKey.REGEX, regexString: <string>}
```

tensorboard/webapp/routes/dashboard_deeplink_provider_test.ts
stephanwlee: Thanks, the test case is added.

tensorboard/webapp/runs/store/runs_reducers_test.ts
```diff
@@ -831,4 +833,143 @@ describe('runs_reducers', () => {
...
+        key: GroupByKey.RUN,
+      });
+    });
+
+    // Color grouping potentially is an expensive operation and assigning colors
+    // on route changes may not actually be effective at all. Because we are
+    // using RouteContextState, color assignment and groupBy information should
+    // not go out of sync. It would only be used to assign initial groupBy
+    // information from URL that will often trigger a run fetch which will be
+    // handled separately by the fetchRuns flow.
```
psybuzz, line 908: iiuc, the gist of this comment is that stateRehydratedFromUrl intentionally does not update color assignment, because it is expensive, so we rely on the fetchRunsSucceeded to handle it.

Suggestion: can we put this comment in the runs_reducers's stateRehydratedFromUrl handler?

> assigning colors on route changes may not actually be effective at all
Could you please help me understand what this means?

> color assignment and groupBy information should not go out of sync
Could you elaborate on what this means?  In this PR, the reducer now updates groupBy without updating color assignments, which suggests that color assignment and groupBy do not need to be kept in sync.

nit: RouteContextState --> RouteContextedState

psybuzz (commented):
Assuming that the plan is to reuse the existing run text input filter for the "regex mode", one thing to consider is how we might also persist the "run filter" text in non-regex mode, some day.  The existing scalars dashboard uses something like "?tagFilter=foo", so it would be a bit strange if we ended up having both "?tagFilter=myfilterisnowduplicated&colorBy:regex:myfilterisnowduplicated" in the URL.

tensorboard/webapp/runs/store/runs_reducers_test.ts
stephanwlee: Left a comment in the code instead of leaving the context on GitHub. PTAL.

stephanwlee:
> The existing scalars dashboard uses something like "?tagFilter=foo", so it would be a bit strange if we ended up having both "?tagFilter=myfilterisnowduplicated&colorBy=regex:myfilterisnowduplicated" in the URL.

Forgot to address this. Right, depending on which UX we decide on, if `runRegexFilter` (not tag but run) becomes the source of truth for the regex group by, we would modify the serialization format to drop `:` and just have `?runColorBy=regex`.

psybuzz (approved):
> we would modify the serialization format to drop : and just have ?runColorBy=regex.

Thanks, sgtm!

PR #5024, opened 2021-05-27
timeseries: keep runs table header sticky

stephanwlee:
This change makes the runs table header sticky to the scrollable
container.

Tested sync and it exhibits no discernible visual changes at all.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
@@ -60,6 +60,16 @@ $_border-color: #0000001f;
   .header {
     color: mat-color($tb-foreground, text);
     white-space: nowrap;
+
+    // Allows table header to remain sticky to the scrollable container.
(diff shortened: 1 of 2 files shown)


psybuzz (approved):
I was hoping we might be able to make the header row itself sticky, rather than the individual children inside it.  However, it didn't work when I tried it, so this lgtm.

Would it be easy and make sense to have the "Select all runs" button and "All runs in this page are selected but not all runs..." sticky as well, since it is semantically similar to the header?

stephanwlee:
> I was hoping we might be able to make the header row itself sticky, rather than the individual children inside it. However, it didn't work when I tried it, so this lgtm.

LOL, you actually tried what I wrestled with. I, too, found that weird.

stephanwlee:
Unfortunately, the select-all flow is a lot harder to support with `position: sticky` and CSS table (as opposed to `<table>`). I may just reorder the elements and will require different kind of changes. Will do that in a follow up.

PR #5025, opened 2021-05-27
color grouping: move regexfilter attribute from runUi to runData

japie1235813:
One of the two color grouping by regex ui option is to align the filter run regex and color grouping regex.
Move the regexFilter attribute from ui state to data state.

### DIFF
--- a/tensorboard/webapp/runs/store/runs_types.ts
+++ b/tensorboard/webapp/runs/store/runs_types.ts
@@ -52,6 +52,8 @@ export interface RunsDataRoutefulState {
   groupKeyToColorString: Map<string, string>;
   userSetGroupBy?: GroupBy;
   initialGroupBy: GroupBy;
+  colorGroupRegexString: string;
+  regexFilter: string;
 }
 
 export interface RunsDataRoutelessState {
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/runs/store/runs_types.ts
```diff
@@ -52,6 +52,8 @@ export interface RunsDataRoutefulState {
   groupKeyToColorString: Map<string, string>;
   userSetGroupBy?: GroupBy;
```
stephanwlee, line 53: You may want to change this to `GroupByKey` so there is no redundant information in the `colorGroupRegexString`.

stephanwlee (commented):
Looking great!

tensorboard/webapp/runs/store/runs_types.ts
japie1235813: Agree. Yet this changes includes `groupRuns` method changes and a bunch of utils(group by regex) and tests changes. Would like to separate them into another pr.

stephanwlee: Another PR sounds good but do note that we don't want to change API of utils, actions, and selectors.

stephanwlee (approved):

PR #5026, opened 2021-05-28
infra: add customization component

yatbear:
CustomizableComponent enables **dynamic customization** of components without having every single layer above it being aware of the change (see [Angular InjectionToken](https://angular.io/api/core/InjectionToken) for more details).

This is the first break-down step of @bmd3k's prototype for the customization approach. 

Example usage:

1. Define a customizable Component, for example adding a button:
    ```
    const CustomizableButton = new InjectionToken<Type<unknown>>('Customizable Button');
    ```

2. Where the customization point is desired, use `tb-customization` component to wrap some default behavior with the `[customizableComponent]` attribute:
     ```
     <tb-customization [customizableComponent]="customButtonIfProvided">
       <div>There is no button by default.</dev>
     </tb-customization>
    ```

3. In the constructor of the component containing the customization point, optionally inject a custom button:
    ```
    constructor(
      @Optional() @Inject(CustomizableButton) readonly customButtonIfProvided: Type<unknown>)
    ```

If you do not want any button, you're done. The TensorBoard service will use the default option (in this case displaying a message). However, if you need to add a button:

4. Define a Component with your customization:
    ```
    @Injectable()
    @Component({
       selector: 'my-custom-button-component',
       template: '<button mat-button>I am a special button!</button>'
    })
    export class MyCustomButtonComponent {}
    ```

5. Provide the customized Component in an NgModule:
    ```
    @NgModule({
      declarations: [MyCustomButtonComponent],
      entryComponents: [MyCustomButtonComponent],
      providers: [{
        provide: CustomizableButton,
        useClass: MyCustomButtonComponent,
      }]
    })
    ```

### DIFF
--- a/tensorboard/webapp/customization/customizable_component.ts
+++ b/tensorboard/webapp/customization/customizable_component.ts
@@ -29,49 +29,48 @@ import {
  *
  * Cookbook:
  *
- * 1. Define an empty marker class that a Component will have to implement in
- *    order to be recognized as the customization:
+ * 1. Define a customizable Component, for example a button:
  *
- *    export class SomeCustomComponentType {};
+ *    const CustomizableButton = new InjectionToken<Type<unknown>>('Customizable Button');
  *
  * 2. Where the customization point is desired, use this Component to wrap some
  *    default behavior. Bind to some possibly-empty variable with the
  *    [customizableComponent] attribute:
  *
- *    <tb-customization [customizableComponent]="customComponentIfProvided">
- *      <div>This is the default content.</div>
+ *    <tb-customization [customizableComponent]="customButtonIfProvided">
+ *      <button mat-button>I am the default button.</button>
  *    </tb-customization>
  *
  * 3. In the constructor of the component containing the customization point,
- *    optionally inject an instance of SomeCustomComponentType:
+ *    optionally inject a custom button:
  *
  *    constructor(
- *        @Optional readonly customComponentIfProvided: SomeCustomComponentType)
+ *      @Optional() @Inject(CustomizableButton)
+ *      readonly customButtonIfProvided: Type<unknown>)
  *
  * If you do not wish to customize the behavior for a certain TensorBoard
- * service, you're done. The TensorBoard service will receive the default
- * behavior.
+ * service (in this case, a button), you're done. The TensorBoard service
+ * will receive the default behavior.
  *
- * However, if you need to define a customization for a TensorBoard service:
+ * However, if you need to define a customization for the button:
  *
- * 4. Define a Component implementation of SomeCustomComponentType with the
- *    customized behavior:
+ * 4. Define a Component with your customization:
  *
  *    @Injectable()
  *    @Component({
- *      selector: 'my-custom-component',
- *      template: '<div>This is the customized content.</div>'
- *    });
- *    export class MyCustomComponent implements SomeCustomComponentType {}
+ *      selector: 'my-custom-button-component',
+ *      template: '<button mat-button>I am a special button!</button>'
+ *    })
+ *    export class MyCustomButtonComponent {}
  *
  * 5. Provide the customized Component in an NgModule:
  *
  *    @NgModule({
- *      declarations: [MyCustomComponent],
- *      entryComponents: [MyCustomComponent],
+ *      declarations: [MyCustomButtonComponent],
+ *      entryComponents: [MyCustomButtonComponent],
  *      providers: [{
- *        provide: SomeCustomComponentType,
- *        useClass: MyCustomComponent,
+ *        provide: CustomizableButton,
+ *        useClass: MyCustomButtonComponent,
  *      }]
  *    })
  */
--- a/tensorboard/webapp/customization/customizable_component.ts
+++ b/tensorboard/webapp/customization/customizable_component.ts
@@ -83,7 +83,7 @@ import {
   `,
 })
 export class CustomizableComponent implements OnInit {
-  @Input() customizableComponent!: {constructor: Type<unknown>} | null;
+  @Input() customizableComponent!: {constructor: Type<unknown>} | undefined;
 
   constructor(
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/customization/customizable_component.ts
```diff
@@ -0,0 +1,102 @@
...
+ * surgical customizations deep within the component hierarchy. The
+ * customizations can be provided dynamically via Angular dependency injection
+ * and providers.
+ *
+ * Cookbook:
+ *
+ * 1. Define an empty marker class that a Component will have to implement in
+ *    order to be recognized as the customization:
+ *
+ *    export class SomeCustomComponentType {};
```
stephanwlee, line 35: Perhaps update the cookbook to add below instead of declaring a class:

```ts
 const CustomizableButton = new InjectionToken<Type<unknown>>('Customizable Button');
```

tensorboard/webapp/customization/customization_test.ts
```diff
@@ -0,0 +1,113 @@
...
+describe('tb-customization', () => {
+  it('renders default if no customization provided', async () => {
+    await TestBed.configureTestingModule({
+      imports: [
+        CustomizationModule,
+        // Note: Notably not importing the module that provides the
+        // implementation of CustomizableComponentType.
+      ],
+      declarations: [ParentComponent],
+    }).compileComponents();
```
stephanwlee, line 87: Another way to avoid this repetition is to factor this out to:

```ts
async function setUp(extraImports: NgModule[] = []) {
      imports: [
        CustomizationModule,
        ...extraImports,
      ],
      declarations: [ParentComponent],
    }).compileComponents();
}
```

tensorboard/webapp/customization/customization_test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * An implementation of CustomizableComponentType to be provided and injected
+ * into the ParentComponent for some tests.
+ */
+@Component({
+  selector: 'customizable-component',
+  template: ` <div>Showing Customized Text!</div> `,
+})
+export class CustomizableComponent implements CustomizableComponentType {}
```
stephanwlee, line 61: `implements CustomizableComponentType` part is a bit odd even that it is not an interface and we really don't want to support invoking method on a component. Just omit it to prevent others from getting a wrong idea?

tensorboard/webapp/customization/customization_test.ts
```diff
@@ -0,0 +1,113 @@
...
+export class CustomizableComponentType {}
+
+/**
+ * Parent class that uses the <tb-customization> component.
+ */
+@Component({
+  selector: 'parent-component',
+  template: `
+    <tb-customization [customizableComponent]="customizableComponent">
+      <div class="default">Showing Default Text!</div>
```
stephanwlee, line 33: Would you mind altering this example a little bit to below?

```html
<tb-customization [customizableComponent]="customizableComponent">
  <span>Showing </span>
  <span>Default Text!</span>
</tb-customization>
```

I want to make sure this can properly support multiple child nodes.

tensorboard/webapp/customization/customizable_component.ts
yatbear: Done.

tensorboard/webapp/customization/customization_test.ts
yatbear: Thanks! Ended up declaring `extraImports` as type `any[]`, otherwise there was a type error.

tensorboard/webapp/customization/customization_test.ts
yatbear: Removed.

tensorboard/webapp/customization/customization_test.ts
yatbear: Thanks for the suggestion!

tensorboard/webapp/customization/customizable_component.ts
```diff
@@ -0,0 +1,101 @@
...
+@Component({
+  selector: 'tb-customization',
+  template: `
+    <ng-container *ngIf="!customizableComponent">
+      <ng-content></ng-content>
+    </ng-container>
+  `,
+})
+export class CustomizableComponent implements OnInit {
+  @Input() customizableComponent!: {constructor: Type<unknown>} | null;
```
stephanwlee, line 86: `null` -> `undefined`. JavaScript/TypeScript has `undefined` vs. `null` difference and, in this case, it would be undefined.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
```diff
@@ -104,7 +104,7 @@
       [seriesMetadataMap]="chartMetadataMap"
       [xScaleType]="xScaleType"
       [yScaleType]="yScaleType"
-      [customXFormatter]="xAxisType === XAxisType.RELATIVE ? relativeXFormatter : undefined"
+      [customXFormatter]="getCustomXFormatter()"
```
stephanwlee, line 107: I think your PR now contains part of mine too. Rebase and make sure your history is near on top of master's.

stephanwlee (approved):
Looks good but consider force pushing with history fixed for posterity.

Also, consider changing the title of the PR to something like "infra: add customization component" and what problem this component is meant to fix. Since there is a nice JSDoc on the source, I will do not think you need to duplicate that in the PR comment but some kind of code snippet of example usage would definitely be :cherries:  on :cake: .

tensorboard/webapp/customization/customizable_component.ts
yatbear: Thanks for the explanation!

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
yatbear: Done.

Issue #1277, opened 2018-07-11
[Feature Request] Night Mode Toggle Option


datlife:
It would be super helpful for people working at night like me.

Thanks,

nfelt:
As a crude approximation of this, you can just invert the colors in the page by running `document.documentElement.style = "filter: invert(100%)"` in the JS console.

Doing this in a more sophisticated way seems tricky, because the various visualizations do different things in terms of color, background, etc.  For example, would you expect that a night mode would change how images on the Image dashboard get displayed (e.g. by tinting them orange) or would you expect it to preserve the original pixel values?  It could also be a bit tricky to define color palettes for the runs that show up well against a dark background (I'm not sure if the current ones would all work).

datlife:
Hi @nfelt,

Thank you for asking. 
* Inspired from recent night mode updates of Youtube / Reddit, I think it is okay to keep the images as-is. 
* I am currently using "Dark Mode" Extension on Google Chrome to switch to night mode. However, for Tensorboard, the scalar tab is blacked out whenever I enable the extension.
* Yeah, true that the color palettes might be messed up. I do not have experience in web development so I am not sure how to comment on this.

Thanks.
Dat

nfelt:
Hmm, if you get a chance could you post a screenshot of what Tensorboard looks like with the extension?  Even if it doesn't look quite right.

datlife:
Sure!

Here is the example of before and after I toggle the Dark Mode extension

### Before
![selection_009](https://user-images.githubusercontent.com/13214917/42587912-06998a7a-84f1-11e8-842d-0c2e3dd78719.png)

### After
![selection_008](https://user-images.githubusercontent.com/13214917/42587918-0963e4d0-84f1-11e8-87ee-c53dfad71851.png)

vedal:
Hi :)
Any updates on this feature?

mhamilton723:
+1 on this feature req!

stephanwlee:
I am happy to report that we now have the dark theme of TensorBoard. 

![image](https://user-images.githubusercontent.com/2547313/124030454-0e733b80-d9ab-11eb-9354-819525ebbd4d.png)

![image](https://user-images.githubusercontent.com/2547313/124030498-192dd080-d9ab-11eb-9584-107321d66e73.png)

It should default apply the dark mode based on your browser settings (or your OS settings; we use the media query) or you can manually set to use a specific theme by interesting with the menu here:

[...]

PR #5031, opened 2021-06-02
dark mode: add feature flag and UI supporter

stephanwlee:
This change is the first of many to support dark mode. There is the
blessed way of supporting dark mode in Angular; their theming system
supports it quite well. However, there are large gap between what
Angular and Angular TensorBoard can support vs. what Polymer based
components can support. Until we figure it out, we want to keep it under
wraps by disabling it for our users.

Also, by leveraging the feature flagging with the URL deserialization,
we can use query parameter to force enable the dark mode for testing
easily.

In subsequent changes, we will piece by piece support dark mode
manually.

### DIFF
--- a/tensorboard/webapp/core/views/dark_mode_supporter_test.ts
+++ b/tensorboard/webapp/core/views/dark_mode_supporter_test.ts
@@ -34,6 +34,10 @@ describe('dark mode supporter test', () => {
     store.overrideSelector(getDarkModeEnabled, false);
   });
 
+  afterEach(() => {
+    document.body.classList.remove('dark-mode');
+  });
+
   it('sets class name on body when dark mode is enabled', () => {
     store.overrideSelector(getDarkModeEnabled, false);
     const fixture = TestBed.createComponent(DarkModeSupportContainer);
--- a/tensorboard/webapp/core/views/dark_mode_supporter_test.ts
+++ b/tensorboard/webapp/core/views/dark_mode_supporter_test.ts
@@ -1,0 +1,58 @@
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {TestBed} from '@angular/core/testing';
+import {NoopAnimationsModule} from '@angular/platform-browser/animations';
+import {Store} from '@ngrx/store';
+import {MockStore, provideMockStore} from '@ngrx/store/testing';
+
+import {State} from '../../app_state';
+import {getDarkModeEnabled} from '../../selectors';
+import {DarkModeSupportContainer} from './dark_mode_supporter_container';
+
+describe('dark mode supporter test', () => {
+  let store: MockStore<State>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [NoopAnimationsModule],
+      providers: [provideMockStore(), DarkModeSupportContainer],
+      declarations: [DarkModeSupportContainer],
+    }).compileComponents();
+    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
+    store.overrideSelector(getDarkModeEnabled, false);
+  });
+
+  it('sets class name on body when dark mode is enabled', () => {
+    store.overrideSelector(getDarkModeEnabled, false);
+    const fixture = TestBed.createComponent(DarkModeSupportContainer);
+    fixture.detectChanges();
+    expect(document.body.classList.contains('dark-mode')).toBe(false);
+
+    store.overrideSelector(getDarkModeEnabled, true);
+    store.refreshState();
+    expect(document.body.classList.contains('dark-mode')).toBe(true);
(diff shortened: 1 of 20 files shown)


tensorboard/webapp/core/views/dark_mode_supporter_test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [NoopAnimationsModule],
+      providers: [provideMockStore(), DarkModeSupportContainer],
+      declarations: [DarkModeSupportContainer],
+    }).compileComponents();
+    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
+    store.overrideSelector(getDarkModeEnabled, false);
+  });
```
psybuzz, line 35: Should we need to reset the `document.body.classList` to default in an `afterEach` to not leak state?

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
```diff
@@ -108,3 +148,6 @@ describe('tb_feature_flag_data_source', () => {
     });
   });
 });
+function DARK_MODE_MEDIA_QUERY(DARK_MODE_MEDIA_QUERY: any) {
```
psybuzz, line 151: unused?

tensorboard/webapp/webapp_data_source/tb_feature_flag_module.ts
```diff
@@ -15,15 +15,35 @@ limitations under the License.
 
 import {NgModule} from '@angular/core';
 
-import {TBFeatureFlagDataSource} from './tb_feature_flag_data_source_types';
-import {QueryParamsFeatureFlagDataSource} from './tb_feature_flag_data_source';
+import {
+  TBFeatureFlagDataSource,
+  TbFeatureFlagDataSources,
```
psybuzz, line 20: Can we use consistent capitalization of `TB` vs `Tb`?

psybuzz (commented):
Overall, things look good, except I'm opposed to the introduction of MediaQueryFeatureFlagDataSource.  I think the specialization of FeatureFlagDataSource into a MediaQuery* and QueryParam* variant adds too much complexity at the moment.  A multi-provider requires introduces a more ambiguity about where a particular feature flag is provided, and all TB surfaces need to include this "QueryParameter provider should appear the last" comment.

imo, the only thing we gain by having this split is that other TB surfaces can opt to include only queryParam flags without including the mediaQuery ones, in a way that does not require composition or redefining all the mediaQuery flags.  This benefit only applies if the other TB surface has no other need to compose/redefine feature flag data source.

If we renamed the existing `QueryParamFeatureFlagDataSource` to be `OSSFeatureFlagDataSource` or something, it would be reasonable for it to read from both query params and media queries.  Other TB surfaces might have to compose/redefine their own feature flag data source, but this is something that other TB surfaces will do already.

Alternatively, wdyt about having DarkModeSupportContainer be responsible for calling `window.matchMedia`?  I don't know if the media query result can change over the lifetime of the page session (if user changes the OS setting without refreshing the tab), but if so, maybe we don't need MediaQueryFeatureFlagDataSource at all?

stephanwlee:
> Alternatively, wdyt about having DarkModeSupportContainer be responsible for calling window.matchMedia? I don't know if the media query result can change over the lifetime of the page session (if user changes the OS setting without refreshing the tab), but if so, maybe we don't need MediaQueryFeatureFlagDataSource at all?

It can change over time (OS nowadays changes modes based on time of the day) which I will not support for now but I want a user to be able to toggle it with a button so the state will not only come from the media query. For a stateful things, I really wish we'd use the standard pattern instead of doing some hybrid where convenient.

tensorboard/webapp/webapp_data_source/tb_feature_flag_testing.ts
```diff
@@ -18,6 +18,8 @@ import {Injectable, NgModule} from '@angular/core';
 import {buildFeatureFlag} from '../feature_flag/testing';
 import {TBFeatureFlagDataSource} from './tb_feature_flag_data_source_types';
 
+export {TBFeatureFlagDataSource} from './tb_feature_flag_data_source_types';
```
psybuzz, line 21: unneeded/unused?

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -59,12 +65,34 @@ export class QueryParamsFeatureFlagDataSource extends TBFeatureFlagDataSource {
...
   }
 
   protected getParams() {
     return util.getParams();
   }
+
+  protected getPartialFeaturesFromMediaQuery(): {enableDarkMode?: boolean} {
+    // Set feature flag value for query parameters that are explicitly
+    // specified. Feature flags for unspecified query parameters remain unset so
+    // their values in the underlying state are not inadvertently changed.
```
psybuzz, line 84: Let's update this comment still refers to "query parameters" although the code below reads media queries.  Optionally, I think the comment could also be dropped, given the comment below on L88

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
```diff
@@ -12,7 +12,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-import {Injectable} from '@angular/core';
+import {Injectable, InjectionToken} from '@angular/core';
```
psybuzz, line 15: unused `InjectionToken`

psybuzz (approved):
Thank you for updating the approach, looking forward to this well requested feature!  https://github.com/tensorflow/tensorboard/issues/1277

PR #5036, opened 2021-06-02
docs: Graph tutorial link to tb.dev

bileschi:
Adds instructions to push to tb.dev to graph.ipynb.  Mirrors getting_started.ipynb.

* Tested
Ran in colab

### DIFF
--- a/docs/graphs.ipynb
+++ b/docs/graphs.ipynb
@@ -295,6 +295,26 @@
     },
     {
       "cell_type": "markdown",
+      "source": [
+        "## TensorBoard.dev\n",
+        "\n",
+        "You can also optionally use TensorBoard.dev, to create a hosted, shareable experiment. "
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "outputs": [],
+      "source": [
+        "!tensorboard dev upload \\\n",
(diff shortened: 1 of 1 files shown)


docs/graphs.ipynb
```diff
@@ -293,6 +293,31 @@
         "%tensorboard --logdir logs"
       ]
     },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "9PFgFjlPEqXb"
+      },
+      "source": [
+        "## TensorBoard.dev\n",
```
nfelt, line 302: It seems too prominent to make this a `##` header, since then it's at the same level as the headers for Op-level Graph and Conceptual Graph.  The result is that the following content which is supposed to be about the Op-level graph actually shows up under the header for TensorBoard.dev instead:

image

I'd either skip the TensorBoard.dev header entirely (since it's not really a full "section" about TB.dev; it's just a single code example) or at least make it `###` or `####` so it doesn't disrupt the flow.

nfelt (changes requested):

docs/graphs.ipynb
bileschi: Good call.  Removed the header line entirely.

nfelt (approved):

PR #5037, opened 2021-06-03
dark mode: disable media query based feature

stephanwlee:
We are not actually ready to turn on the feature in OSS for people who
are using media query and is using tb-nightly regularly. Until we are
more ready with integration tests, we will disable the media query based
automatic enablement of the dark mode feature.
mode feature

### DIFF
--- a/tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
+++ b/tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
@@ -84,7 +84,9 @@ export class QueryParamsFeatureFlagDataSource extends TBFeatureFlagDataSource {
 
     // When media query matches positively, it certainly means user wants it but
     // it is not definitive otherwise (i.e., query params can override it).
-    if (enableDarkMode) {
+    // TODO(stephanwlee): enable the feature when most of the UI is actually
+    // ready for usage.
+    if (enableDarkMode && false) {
       featureFlags.enableDarkMode = true;
     }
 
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -84,7 +84,9 @@ export class QueryParamsFeatureFlagDataSource extends TBFeatureFlagDataSource {
 
     // When media query matches positively, it certainly means user wants it but
     // it is not definitive otherwise (i.e., query params can override it).
-    if (enableDarkMode) {
+    // TODO(stephanwlee): enable the feature when most of the UI is actually
+    // ready for usage.
+    if (enableDarkMode && false) {
```
psybuzz, line 89: Optional: alternatively, would it be preferred to set featureFlags.enableDarkMode to true only when both the queryParam asks for it AND the media query prefers it?  That way, it remains off by default but anyone can still test out dark theme in TB nightly as well.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
```diff
@@ -121,7 +121,8 @@ describe('tb_feature_flag_data_source', () => {
             .and.returnValue({matches: matchDarkMode} as MediaQueryList);
         }
 
-        it('returns enableDarkMode when media query matches dark mode', () => {
+        // TODO(stephanwlee): disabled while working on the feature.
+        xit('returns enableDarkMode when media query matches dark mode', () => {
```
psybuzz, line 125: fyi: I think internal presubmit will complain about committing code with 'xit'.  An optional alternative is to change the test expectation to false, if desired.

psybuzz (approved):

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
stephanwlee: It would be impossible for me to try out and debug the feature myself :) Sure, I can use CDT's feature to mock it but it has been a bit annoying.

PR #5038, opened 2021-06-03
dark mode: enable dark mode at PluginComponent

stephanwlee:
This change adds SASS variable for dark theme and conditionally applies
the dark modified styles on Angular Material components. This change
also adds imperfect theme supports on the Plugin stamper.

To exemplify its functionality, we have manually fixed the scalars
plugin as an experiment and added its changes accordingly.
After this change, scalars plugin look okay.

Do note that we had to adjust the colors to use one from variable and it
will cause certain colors to adjust in a minor way that would be caught
by the screenshot tests.

### DIFF
--- a/tensorboard/webapp/plugins/plugins_component.scss
+++ b/tensorboard/webapp/plugins/plugins_component.scss
@@ -1,4 +1,4 @@
-/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


tensorboard/webapp/plugins/plugins_component.scss
```diff
@@ -0,0 +1,63 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
psybuzz, line 1: nit: 2021

tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
```diff
@@ -19,7 +19,6 @@ $_table-content-row-min-height: 43px;
 $_font-size: 13px;
 
 :host {
-  background-color: #fff;
```
psybuzz, line 22: Looks like this change affects light-theme.  Let's preserve light-theme's run table colors:

image

psybuzz (approved):

PR #5039, opened 2021-06-04
refactor: move inline_images to tensorboard/defs/

yatbear:
For tensorboard.dev to reuse the inline_images tool.

Tested Debugger V2 image loading:
image


tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive/BUILD
```diff
@@ -4,19 +4,12 @@ package(default_visibility = ["//tensorboard:internal"])
...
-
 genrule(
     name = "gen_inactive_component.ng.html",
     srcs = ["inactive_component.uninlined.ng.html"] + glob(["images/*.png"]),
     outs = ["inactive_component.ng.html"],
-    cmd = "$(execpath //tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive:inline_images) $(SRCS) >'$@'",
-    exec_tools = [":inline_images"],
+    cmd = "$(execpath //tensorboard/defs:inline_images) $(SRCS) >'$@'",
+    exec_tools = ["//tensorboard/defs:inline_images"],
 )
```
stephanwlee, line 13: Great but we don't want all consumers of the tool to use genrule and this boilerplatey syntax or API of the tool.

Instead, make a macro that goes something like:
```py
# In appropriate .bzl file
def tf_inline_images(name, html_template, images, out):
    native.genrule(
        name=name,
        srcs=[html_template] + images,
        outs=[out],
        cmd="$(execpath //tensorboard/defs:inline_images) $(SRCS) >'$@'",
        exec_tools=["//tensorboard/defs:inline_images"],
    ) 
```

which you will use it like:
```py
load(<path to the bzl>, "tf_inline_images")
...


tf_inline_images(
    name="gen_inactive_component.ng.html",
    html_template = "inactive_component.uninlined.ng.html",
    images = glob(["images/*.png"]),
    out = "inactive_component.ng.html",
)
```

yatbear: Added a macro in to `defs.bzl`. Thanks for the informative comment!

tensorboard/defs/defs.bzl
```diff
@@ -200,3 +200,24 @@ def tf_ng_module(assets = [], **kwargs):
         angular_assets = assets,
         **kwargs
     )
+
+def tf_inline_images(name, html_template, images, out):
+    """Inline png images in html.
+
+    Replaces %*.png% in the input .html file with a data URI containing the
+    base64-encoded image content of the corresopnding .png files in the images/
+    subdirectory.
```
stephanwlee, line 209: We won't be getting an `images` directory per say but rather we would get their file names as input to the rule. So, it would be more correct to say,
```
Replaces %<file_basename>.png% in the input `html_template` with a data URI containing the
base64-encoded image content of the corresopnding .png files in the `images` input.

In case there is a collision in the base file name, the first instance will take precedence over the
others.

Example:
# In html_template
<img src="%my_file.png%" />

# in BUILD
tf_inline_pngs(
  name = "my_rule",
  html_template = "path_to_my_template.html",
  images = [
    "path_to/my_file.png",
  ] + glob("some_folder/*.png"),
  out = "my_filename.html",
)
```

tensorboard/defs/defs.bzl
```diff
@@ -200,3 +200,24 @@ def tf_ng_module(assets = [], **kwargs):
         angular_assets = assets,
         **kwargs
     )
+
+def tf_inline_images(name, html_template, images, out):
```
stephanwlee, line 204: I forgot that the utility literally only inlines PNG. We can generalize that and keep current name (by using mimetypes Python module to ascertain the MIME information) and replace these lines (`MIME_TYPE` part) or rename this rule to be `tf_inline_pngs`. WDYT?

https://github.com/tensorflow/tensorboard/blob/ba778d70baf2754932914d2715dfce6e5eb80dde/tensorboard/plugins/debugger_v2/tf_debugger_v2_plugin/views/inactive/inline_images.py#L27-L38

stephanwlee (approved):
It looks great! Thanks.

My comments would be good to include in this changeset but we can definitely improve to afterwards.

tensorboard/defs/defs.bzl
yatbear: Good point! Renaming to `tf_inline_pngs` for now and will generalize it later (added a TODO).

tensorboard/defs/defs.bzl
yatbear: Done. Thanks!

stephanwlee (approved):

PR #5040, opened 2021-06-04
dark mode: fix hparams, graph, mesh, and profile

stephanwlee:
This change fixes the visuals of hparams, graph, mesh and profile
redirection plugin under the dark mode.

Do note that we did not attempt to have a generic solution for all
iframe based plugin since it becomes wildly more difficult to reliably
apply dark styles. Instead, we will show the default light mode which
will be quite jarring but that's the best we can do as of right now.

### DIFF
--- a/tensorboard/plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-redirect-dashboard.ts
+++ b/tensorboard/plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-redirect-dashboard.ts
@@ -49,6 +49,15 @@ class TfProfileRedirectDashboard extends LegacyElementMixin(PolymerElement) {
     </div>
 
     <style>
+      :host {
+        display: block;
+        height: 100%;
+        left: 0;
+        position: absolute;
+        top: 0;
+        width: 100%;
+      }
+
       .message {
         margin: 80px auto 0 auto;
         max-width: 540px;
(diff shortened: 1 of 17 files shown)


tensorboard/plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-redirect-dashboard.ts
```diff
@@ -49,6 +49,15 @@ class TfProfileRedirectDashboard extends LegacyElementMixin(PolymerElement) {
     </div>
 
     <style>
+      :host {
+        display: block;
+        height: 100%;
+        left: 0;
+        position: absolute;
+        top: 0;
+        width: 100%;
```
psybuzz, line 58: top/left can be dropped.

Optional: replacing this entire block with 
```
:host {
  display: flex;
}
```
seems like it would have the same effect, if you prefer avoiding pos: abs.

psybuzz (approved):
Going from another dashboard to the "profile redirect" dashboard, it looks a bit strange.

image

Can we make the text white and give it a solid background?

psybuzz:
> Going from another dashboard to the "profile redirect" dashboard, it looks a bit strange.

Whoops just realized I was looking at the "dashboard not found" page, not the "profile redirect".  Feel free to ignore this comment, since it is out of scope of the PR

tensorboard/plugins/profile_redirect/tf_profile_redirect_dashboard/tf-profile-redirect-dashboard.ts
stephanwlee: FWIW, they really are not equivalent. 

image

Without global background color set, with flex, it would look incorrect (which was what I used to have). However, since I now set the global background, they appear the same thus incorporated your suggestion.

PR #5042, opened 2021-06-04
dark mode: improve visuals of time series plugin

stephanwlee:
This change improves how TimeSeries dashboard looks under the dark mode.
It largely looks okay but the gpu line chart is a bit broken under it.

Do note that some of the color change will cause minor screenshot test failures.
For instance, we were using white + opacity for border color on TimeSeries
runs table which had to change to a opaque color.

### DIFF
--- a/tensorboard/webapp/metrics/views/metrics_container.scss
+++ b/tensorboard/webapp/metrics/views/metrics_container.scss
@@ -14,7 +14,7 @@ limitations under the License.
 ==============================================================================*/
 @import 'tensorboard/webapp/theme/tb_theme';
 
-:host {
+.plugin-view {
   display: flex;
   height: 100%;
 }
@@ -24,6 +24,11 @@ nav {
   border-right: 1px solid mat-color($tb-foreground, border);
   flex: none;
   width: 340px;
+
+  @include tb-dark-theme {
+    background-color: map-get($tb-dark-background, background);
+    border-right-color: mat-color($tb-dark-foreground, border);
+  }
 }
 
 metrics-main-view {
--- a/tensorboard/webapp/metrics/views/metrics_container.scss
+++ b/tensorboard/webapp/metrics/views/metrics_container.scss
@@ -14,7 +14,7 @@ limitations under the License.
 ==============================================================================*/
 @import 'tensorboard/webapp/theme/tb_theme';
 
-.plugin-view {
+:host {
   display: flex;
   height: 100%;
 }
--- a/tensorboard/webapp/metrics/views/metrics_container.ts
+++ b/tensorboard/webapp/metrics/views/metrics_container.ts
@@ -17,10 +17,12 @@ import {ChangeDetectionStrategy, Component} from '@angular/core';
 @Component({
   selector: 'metrics-dashboard',
   template: `
-    <nav>
-      <runs-selector></runs-selector>
-    </nav>
-    <metrics-main-view></metrics-main-view>
+    <div class="plugin-view">
+      <nav>
+        <runs-selector></runs-selector>
+      </nav>
+      <metrics-main-view></metrics-main-view>
+    </div>
   `,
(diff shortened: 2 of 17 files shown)


tensorboard/webapp/metrics/views/metrics_container.ts
```diff
@@ -17,10 +17,12 @@ import {ChangeDetectionStrategy, Component} from '@angular/core';
 @Component({
   selector: 'metrics-dashboard',
   template: `
-    <nav>
-      <runs-selector></runs-selector>
-    </nav>
-    <metrics-main-view></metrics-main-view>
+    <div class="plugin-view">
```
psybuzz, line 20: I'm not seeing why this extra wrapper was necessary.  Can we drop it, or if not, could you add a comment explaining why it's needed?

tensorboard/webapp/tbdev_upload/tbdev_upload_dialog_component.scss
```diff
@@ -51,6 +51,10 @@ p {
   display: flex;
   justify-content: space-between;
   padding: 2px 12px;
+
+  @include tb-dark-theme {
+    background-color: #f5f6f722;
```
psybuzz, line 56: This isn't supposed to be a transparent element.  If we want to make it clear from reading that it's the same hue/saturation with higher brightness, consider hsl()?

tensorboard/webapp/metrics/views/metrics_container.scss
```diff
@@ -24,6 +24,11 @@ nav {
   border-right: 1px solid mat-color($tb-foreground, border);
   flex: none;
   width: 340px;
+
+  @include tb-dark-theme {
+    background-color: map-get($tb-dark-background, background);
+    border-right-color: mat-color($tb-dark-foreground, border);
+  }
```
psybuzz, line 31: To check my understanding, is this equivalent to
```
nav {
  @include tb-theme-background-prop(background-color, background);
  @include tb-theme-foreground-prop(border-right, border, 1px solid);
```
?

tensorboard/webapp/theme/_tb_theme.template.scss
```diff
@@ -68,12 +92,101 @@ $tb-dark-background: map-get($tb-dark-theme, background);
...
+ *   // :host-context(body.dark-mode) div[ngcomponentweirdid] {
+ *   //   border-left: 1px solid map_get($tb-dark-foreground, border);
+ *   // }
+ *   //
+ *   // Third argument is optional. In the example below, `color` only takes
+ *   // color value so it is not needed.
+ *   @include tb-theme-foreground-prop(color, secondary-text);
+ * }
+ */
+@mixin tb-theme-foreground-prop($property, $name, $value: null) {
```
psybuzz, line 142: The example mentioned is great.

Upon first glance, I think there is a little bit of confusion between these param names.  Wdyt about giving these params more specific names or doing a JSDoc-like comment, something along the lines of:

- $property -> $css-property
- $name -> $tb-variable  (must be a key in $tb-theme)
- $value -> $css-value-prefix

?

psybuzz (approved):
Nice!

tensorboard/webapp/metrics/views/metrics_container.ts
stephanwlee: Removed. It was needed at one point but now can be removed.

tensorboard/webapp/metrics/views/metrics_container.scss
stephanwlee: Correct; I still cannot make up my mind on which pattern is better when there are more than N *-props call but this is the pattern that I wanted to include in the code base for reference.

tensorboard/webapp/theme/_tb_theme.template.scss
stephanwlee: Done.

PR #5043, opened 2021-06-04
dark mode: improve GPU line chart under dark mode

stephanwlee:
Since we had to enable the transparency mode on the GPU line chart for
the grid line, the task got a lot easier; we just have to set the
background and color to inherit from ancestor. Yes, GPU line chart is 
snappier with alpha channel turned off but gridding and what not is very
convenient with alpha channel.

Even with the opacity, we still have to propagate it since we calculate
a concrete opaque color value for WebGL renderer and it needs to 
know the color mode in order to properly calculate the concrete color.

image

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
@@ -26,13 +26,17 @@ import {
   TrianglePaintOption,
 } from './renderer_types';
 
-function createOpacityAdjustedColor(hex: string, opacity: number): THREE.Color {
+function createOpacityAdjustedColor(
+  baseColorHex: string,
+  hex: string,
+  opacity: number
+): THREE.Color {
   if (opacity === 1) return new THREE.Color(hex);
   const newD3Color = hsl(hex);
   if (!newD3Color) {
     throw new Error(`d3 failed to recognize the color: ${hex}`);
   }
-  return new THREE.Color(interpolateHsl(newD3Color, '#fff')(1 - opacity));
+  return new THREE.Color(interpolateHsl(newD3Color, baseColorHex)(1 - opacity));
 }
 
 enum CacheType {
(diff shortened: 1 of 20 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
```diff
@@ -294,6 +304,12 @@ export class ThreeRenderer implements ObjectRenderer<CacheValue> {
     }
   }
 
+  setUseDarkMode(useDarkMode: boolean): void {
+    this.backgroundColor = useDarkMode ? '#303030' : '#ffffff';
```
psybuzz, line 308: nit: let's be consistent with either '#fff' or '#ffffff' on L266 and L308

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.scss
```diff
@@ -28,6 +30,13 @@ limitations under the License.
     'dot xaxis';
   grid-template-columns: 50px 1fr;
   grid-auto-rows: 1fr 30px;
+
+  // Unlike other Angular components, we may use this component outside
+  // of the Angular app and want to control the dark mode styling based on
+  // the component prop alone.
```
psybuzz, line 36: I wasn't aware of any restriction that the LineChartComponent Angular component must be reusable outside of the Angular app.  I was expecting we could use one of the existing CSS patterns `@tb-theme-foreground-prop(color, text)` or the `@include tb-dark-theme` instead of having a 3rd way.

iiuc, TB surfaces that do not support dark theme will never add ".dark-mode" to the body, so ":host-context(.dark-mode)" would be false for them, and this line chart would behave as expected.

- Why wouldn't `@include tb-dark-theme` work outside of the Angular app?
- Are we setting a precedent that all components under widgets/ must not depend on `tensorboard/webapp/theme/tb_theme`?  (which is not the case today)

stephanwlee: > Why wouldn't @include tb-dark-theme work outside of the Angular app?

In a light themed app, I want to be able to provide a dark chart (since it is a prop anyways), so in that case, the `:host-context()` method is definitely not the way to go.

> Are we setting a precedent that all components under widgets/ must not depend on tensorboard/webapp/theme/tb_theme? (which is not the case today)

No, I am not setting such precedence. I would be using themed variable or mat-* variables for something that does not need to vary based on theme.

psybuzz (approved):

PR #5045, opened 2021-06-07
dark mode: improve app-header under dark mode

stephanwlee:
This change fixes the style of app-header under the dark mode.
Previously, it was a bit too bright for the dark mode and, as such, we
now use more darker material orange. This improves legibility of the
font as it increases the contrast with the white font.

### DIFF
--- a/tensorboard/webapp/theme/_tb_palette.scss
+++ b/tensorboard/webapp/theme/_tb_palette.scss
@@ -44,15 +44,15 @@ $tf-slate: (
    several orange values, but only one of them (--tb-orange-strong) is used
    widely. */
 $tf-orange: (
-  100: mat-color($mat-orange, 500),
-  200: mat-color($mat-orange, 500),
-  300: mat-color($mat-orange, 500),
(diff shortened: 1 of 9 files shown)


tensorboard/webapp/theme/_tb_palette.scss
```diff
@@ -44,15 +44,15 @@ $tf-slate: (
...
-  100: mat-color($mat-orange, 500),
-  200: mat-color($mat-orange, 500),
-  300: mat-color($mat-orange, 500),
-  400: mat-color($mat-orange, 500),
+  100: mat-color($mat-orange, 100),
+  200: mat-color($mat-orange, 200),
+  300: mat-color($mat-orange, 300),
+  400: mat-color($mat-orange, 400),
   500: mat-color($mat-orange, 500),
-  600: mat-color($mat-orange, 700),
```
psybuzz, line 52: Glad this is fixed!

psybuzz (approved):

PR #5048, opened 2021-06-08
color grouping: separate regexString from GroupBy 

japie1235813:
Motivation
Previously we embedded `regexString` into `GroupByRegex` as one type of `GroupBy`. This pr separate the string from the GroupByKey enum to prepare the regexString could be the same as `regexFilter` (used by filter out run) or `colorGroupRegexString` (regex only for color grouping, differentiated from `regexFilter`)
Also `userSetGroupBy`/`initialGroupBy` is adjusted to `GroupByKey` enum only. 

Technical details

update: remain the original GroupBy type structure; compose them _only_ in user selector 
specifically in `getRunUserSetGroupBy` method

Outdated:
* To not change the APIs of utils, selectors, and actions. We preserve the attributes of `GroupBy` which includes `key` and `regexString`. To be specific:
```
export type GroupBy = {
  key: GroupByKey | null;
  regexString?: string;
};
```
* The key is set to `GroupByKey` enum; regexString remains string type but should be updates with `colorGroupRegexString` from dataState.
* This pr set the regexString with the color group only regex (`colorGroupRegexString`); further prototype with `regexFilter` is TBC.

### DIFF
--- a/tensorboard/webapp/routes/dashboard_deeplink_provider.ts
+++ b/tensorboard/webapp/routes/dashboard_deeplink_provider.ts
@@ -124,7 +124,7 @@ export class DashboardDeepLinkProvider extends DeepLinkProvider {
       ),
       store.select(selectors.getRunUserSetGroupBy).pipe(
         map((groupBy) => {
-          if (!groupBy) return [];
+          if (!groupBy || groupBy.key === null) return [];
           let value: string;
 
           switch (groupBy.key) {
--- a/tensorboard/webapp/routes/dashboard_deeplink_provider.ts
+++ b/tensorboard/webapp/routes/dashboard_deeplink_provider.ts
@@ -124,7 +124,7 @@ export class DashboardDeepLinkProvider extends DeepLinkProvider {
       ),
       store.select(selectors.getRunUserSetGroupBy).pipe(
         map((groupBy) => {
-          if (!groupBy || groupBy.key === null) return [];
+          if (!groupBy) return [];
           let value: string;
 
           switch (groupBy.key) {
--- a/tensorboard/webapp/runs/store/runs_reducers.ts
+++ b/tensorboard/webapp/runs/store/runs_reducers.ts
@@ -225,7 +225,9 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
     const defaultRunColorForGroupBy = new Map(state.defaultRunColorForGroupBy);
 
     const groups = groupRuns(
-      state.userSetGroupByKey ? createGroupBy(state.userSetGroupByKey, state.colorGroupRegexString) : state.initialGroupBy,
+      state.userSetGroupByKey
+        ? createGroupBy(state.userSetGroupByKey, state.colorGroupRegexString)
+        : state.initialGroupBy,
       runsForAllExperiments,
       state.runIdToExpId
     );
--- a/tensorboard/webapp/runs/store/runs_reducers.ts
+++ b/tensorboard/webapp/runs/store/runs_reducers.ts
@@ -280,7 +280,7 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
 
       return {
         ...state,
-        userSetGroupByKey: groupBy.key ?? undefined,
+        userSetGroupByKey: groupBy.key,
         defaultRunColorForGroupBy,
         groupKeyToColorString,
         // Resets the color override when the groupBy changes.
--- a/tensorboard/webapp/runs/store/runs_reducers.ts
+++ b/tensorboard/webapp/runs/store/runs_reducers.ts
@@ -48,7 +48,7 @@ const {
     runColorOverrideForGroupBy: new Map(),
     defaultRunColorForGroupBy: new Map(),
     groupKeyToColorString: new Map(),
-    initialGroupBy: {key: GroupByKey.RUN},
+    initialGroupBy: GroupByKey.RUN,
     colorGroupRegexString: '',
     regexFilter: '',
   },
@@ -95,7 +95,7 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
 
     return {
       ...state,
-      userSetGroupBy: groupBy,
+      userSetGroupBy: groupBy.key ?? undefined,
     };
   }),
   on(runsActions.fetchRunsRequested, (state, action) => {
@@ -224,7 +224,7 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
     const defaultRunColorForGroupBy = new Map(state.defaultRunColorForGroupBy);
 
     const groups = groupRuns(
-      state.userSetGroupBy ?? state.initialGroupBy,
+      {key: state.userSetGroupBy ?? state.initialGroupBy, regexString: ''},
       runsForAllExperiments,
       state.runIdToExpId
     );
@@ -277,7 +277,7 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
 
       return {
         ...state,
-        userSetGroupBy: groupBy,
+        userSetGroupBy: groupBy.key ?? undefined,
         defaultRunColorForGroupBy,
         groupKeyToColorString,
         // Resets the color override when the groupBy changes.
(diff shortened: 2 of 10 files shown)


tensorboard/webapp/runs/store/runs_types.ts
```diff
@@ -50,8 +50,8 @@ export interface RunsDataRoutefulState {
   defaultRunColorForGroupBy: Map<RunId, string>;
   runColorOverrideForGroupBy: Map<RunId, string>;
   groupKeyToColorString: Map<string, string>;
-  userSetGroupBy?: GroupBy;
-  initialGroupBy: GroupBy;
+  userSetGroupBy?: GroupByKey;
+  initialGroupBy: GroupByKey;
```
stephanwlee, line 54: This should not change to `GroupByKey` for the stronger typing reasons. It should remain as `GroupBy`. You should only change ` userSetGroupBy` to something like below:

```ts
userSetGroupByKey?: GroupByKey;
colorGroupRegexString?: string;
```

tensorboard/webapp/runs/types.ts
```diff
@@ -61,16 +61,10 @@ export enum GroupByKey {
...
-}
-
-export interface RegexGroupBy {
-  key: GroupByKey.REGEX;
-  regexString: string;
-}
-
-export type GroupBy = BaseGroupBy | RegexGroupBy;
+export type GroupBy = {
+  key: GroupByKey | null;
```
stephanwlee, line 65: `key` should not be nullable. Please consider reverting back this change. It was strongly typed as is.

tensorboard/webapp/runs/store/runs_selectors.ts
```diff
@@ -137,7 +142,19 @@ export const getRunGroupBy = createSelector(
   getRunUserSetGroupBy,
   getDataState,
   (userSetGroupBy: GroupBy | null, dataState: RunsDataState): GroupBy => {
-    return userSetGroupBy ?? dataState.initialGroupBy;
+    const regexString = dataState.colorGroupRegexString;
+    if (
+      userSetGroupBy &&
+      userSetGroupBy.hasOwnProperty('key') &&
```
stephanwlee, line 148: Should avoid using `hasOwnProperty` as it will not do the proper thing when JSCompiler mangles property name. Please revert back to the original types and it will be fine.

stephanwlee (changes requested):

tensorboard/webapp/routes/dashboard_deeplink_provider.ts
```diff
@@ -124,7 +124,7 @@ export class DashboardDeepLinkProvider extends DeepLinkProvider {
       ),
       store.select(selectors.getRunUserSetGroupBy).pipe(
         map((groupBy) => {
-          if (!groupBy) return [];
+          if (!groupBy || groupBy.key === null) return [];
```
stephanwlee, line 127: This now can be reverted as `groupBy.key` cannot be a null.

tensorboard/webapp/runs/store/runs_reducers.ts
```diff
@@ -277,7 +280,7 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
 
       return {
         ...state,
-        userSetGroupBy: groupBy,
+        userSetGroupByKey: groupBy.key ?? undefined,
```
stephanwlee, line 283: `userSetGroupByKey: groupBy.key ?? undefined` -> `userSetGroupByKey: groupBy.key`

tensorboard/webapp/runs/store/utils.ts
```diff
@@ -81,3 +81,21 @@ export function groupRuns(
   }
   return runGroup;
 }
+
+/**
+ * Util function for composing `GroupBy` key and regex string.
+ */
+export function createGroupBy(
```
stephanwlee, line 88: Sorry for the trouble but can you write a small test that tests this utility?

tensorboard/webapp/runs/store/utils.ts
```diff
@@ -81,3 +81,21 @@ export function groupRuns(
...
+  groupByKey: GroupByKey,
+  regexString: string | null
```
stephanwlee, line 90: Can `regexString` be `null` in Reducer? If not, is it important that it is `null`able in this utility?

tensorboard/webapp/runs/store/runs_selectors_test.ts
```diff
@@ -415,13 +411,9 @@ describe('runs_selectors', () => {
...
-          userSetGroupBy: {
-            key: GroupByKey.REGEX,
-            regexString: 'hello',
-          },
-          initialGroupBy: {
-            key: GroupByKey.RUN,
-          },
+          colorGroupRegexString: 'hello',
+          initialGroupBy: {key: GroupByKey.RUN},
+          userSetGroupByKey: GroupByKey.REGEX,
```
stephanwlee, line 416: Could you add a case for:
```
colorGroupRegexString: '',
initialGroupBy: {key: GroupByKey.REGEX, regexString: 'hi'},
userSetGroupByKey: GroupByKey.REGEX,
```

I honestly do not know what the right behavior should be but we should codify our behavior in this test spec.

stephanwlee (commented):
Almost there except for test! :D

tensorboard/webapp/runs/store/runs_selectors_test.ts
japie1235813: thanks for pointing this out. I think this means the user set groupBykey with '' (empty) string so the the result is 
`{key: GroupByKey.REGEX, regexString: ''}`

tensorboard/webapp/routes/dashboard_deeplink_provider.ts
japie1235813: Ah, thanks for catching this. Done

tensorboard/webapp/runs/store/utils.ts
japie1235813: good call. added them.
except for not sure how to test `default:` any suggestions? I think it should be covered by the first two cases?

tensorboard/webapp/runs/store/utils.ts
japie1235813: removed `null` type.
Yet I think `regexString` could be an optional input. add `?` and replaced with empty string if it does not exist.

tensorboard/webapp/runs/store/runs_reducers.ts
```diff
@@ -95,7 +96,7 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
 
     return {
       ...state,
-      userSetGroupBy: groupBy,
+      userSetGroupByKey: groupBy.key ?? undefined,
```
stephanwlee, line 99: > `userSetGroupByKey: groupBy.key ?? undefined,`

Can `groupBy.key` be `undefined`? If so, change this to `userSetGroupByKey: groupBy.key ?? null` instead.

tensorboard/webapp/runs/store/runs_selectors.ts
```diff
@@ -126,7 +126,12 @@ export const getRunSelectionMap = createSelector(
 export const getRunUserSetGroupBy = createSelector(
   getDataState,
   (dataState: RunsDataState): GroupBy | null => {
-    return dataState.userSetGroupBy ?? null;
+    return dataState.userSetGroupByKey
```
stephanwlee, line 129: This should be a null-check. Make sure there is a test that would have caught this. If the `userSetGroupByKey === GroupByKey.RUN`, it will hit this error (I think `GroupByKey.RUN == 0`).

tensorboard/webapp/runs/store/utils_test.ts
```diff
@@ -232,4 +232,30 @@ describe('run store utils test', () => {
       });
     });
   });
+
+  describe('creatGroupBy', () => {
+    it('group by run', () => {
```
stephanwlee, line 237: s/'group by/'groups by/gc (here and below)

stephanwlee (approved):
Thanks for the change!

PR #5052, opened 2021-06-09
docs: add experiment list information in tensorboard.dev tutorial colab

yatbear:
For Googlers, please refer to issue [b/190221215](https://b.corp.google.com/issues/190221215).

Preview: https://github.com/yatbear/tensorboard/blob/master/docs/tbdev_getting_started.ipynb


bileschi (approved):
Looks great!

bileschi:
Looks like perhaps a flake in CI?  Don't see why it would fail with this change.  Also could be an issue w/ tf-nightly?

yatbear:
> Looks like perhaps a flake in CI? Don't see why it would fail with this change. Also could be an issue w/ tf-nightly?



> Looks like perhaps a flake in CI? Don't see why it would fail with this change. Also could be an issue w/ tf-nightly?

Looks like there's been some fix to the tf-nightly and now everything looks ok.

PR #5058, opened 2021-06-09
dark mode: specify app-bar color via `default`

stephanwlee:
Color palette has a notion of `default`, `ligher`, `darker`, and `text`
and we can access the said color via name.

This change makes use of the `default` when getting the primary color.


psybuzz (approved):
Thanks for this!

PR #5065, opened 2021-06-16
feature flag: add flag for controlling dark mode

stephanwlee:
We want to slowly enable the dark mode based on media query on
app-by-app basis. This change allows each application to configure 
whether the media query based dark mode is possible.

### DIFF
--- a/tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
+++ b/tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
@@ -14,12 +14,14 @@ limitations under the License.
 ==============================================================================*/
 
 import {Injectable} from '@angular/core';
-import {createAction, Action} from '@ngrx/store';
+import {createAction, Action, Store} from '@ngrx/store';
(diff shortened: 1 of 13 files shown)


tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
```diff
@@ -34,15 +36,17 @@ export class FeatureFlagEffects {
   readonly getFeatureFlags$ = createEffect(() =>
     this.actions$.pipe(
       ofType(effectsInitialized),
-      map(() => {
-        const features = this.dataSource.getFeatures();
+      combineLatestWith(this.store.select(getIsAutoDarkModeAllowed)),
```
bmd3k, line 39: Feature flags that control feature flags. We need to go deeper.

bmd3k (approved):
> We want to slowly enable the dark mode based on media query on
> app-by-app basis. This change allows each application to configure so
> that the media query based dark mode is possible.

"so that" => "whether"?

PR #5068, opened 2021-06-16
ci: pin tf-nightly to 20210614

stephanwlee:
With the latest nightly (20210616), TensorBoard CI fails because hparams
API fails to load properly when importing TensorFlow inside TensorBoard
project.

Specifically, below fails.

```sh
# cd into TensorBoard repository
cd tensorboard
python
>>> import tensorflow as tf
```

This may have to do with summaries are exported under TensorFlow and
requires a bit more precise diagnosis. Until then, we will pin the
tf-nightly to the last good version.

Suspected breaking change: https://github.com/tensorflow/tensorflow/commit/2e6acf753381cb7902017bcab03e204bb41b7091

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tf_version_id: ['tf-nightly', 'notf']
+        tf_version_id: ['tf-nightly==2.6.0.dev20210614', 'notf']
         python_version: ['3.7']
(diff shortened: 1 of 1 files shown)


stephanwlee:
When the CI passes, we will make changes to the repository configuration to require "build (tf, 3.7)" instead.

nfelt (approved):
LGTM. I think we just need the branch protection rule change now.

stephanwlee:
Thanks a lot!

PR #5071, opened 2021-06-16
infra: update officially supported Python version

### LINKED ISSUE #5069: Which Python versions are currently supported?
Hi,

I'm one of the maintainers for [google-auth](https://pypi.org/project/google-auth/). We're planning on officially removing 2.7 support soon (requiring python >= 3.6). In preparation for that, I looked for packages that declare direct dependencies on `google-auth` and have 2.7 in the trove classifiers.

I noticed that the `setup.py` lists 2.7, 3.4, 3.5, and 3.6 and the `python_requires` is more permissive:

https://github.com/tensorflow/tensorboard/blob/13c6ecf76443e3bae084a4d6e3de8254b0e39ab3/tensorboard/pip_package/setup.py#L76-L81

https://github.com/tensorflow/tensorboard/blob/13c6ecf76443e3bae084a4d6e3de8254b0e39ab3/tensorboard/pip_package/setup.py#L65-L66

The pyproject.toml has 3.6, 3.7, and 3.8.

https://github.com/tensorflow/tensorboard/blob/13c6ecf76443e3bae084a4d6e3de8254b0e39ab3/pyproject.toml#L1-L3

Could someone clarify which Python versions are currently supported? If 2.7 is currently supported, what is the timeline for removing support?

Thanks!

stephanwlee:
Currently, `setup.py` was very out of date and claimed that TensorBoard
supported Python  2 which is dropped in EOY 2020. This updates the list of
supported Python to match that of TensorFlow's.

https://github.com/tensorflow/tensorflow/blob/5809916f1ea8954b88da0b3c66922f4e5e5e9b47/tensorflow/tools/pip_package/setup.py#L346-L351

Fixes #5069.


tensorboard/pip_package/setup.py
```diff
@@ -62,8 +62,6 @@ def get_readme():
             "tf_projector_plugin/projector_binary.js",
         ],
     },
-    # Disallow python 3.0 and 3.1 which lack a 'futures' module (see above).
```
nfelt, line 65: Optional but maybe kill the `futures` and `wheel` deps that were py2-only too?

https://github.com/tensorflow/tensorboard/blob/b53c723980a5429d81e0443772b41a26334ea288/tensorboard/pip_package/requirements.txt#L19-L20

nfelt (approved):
> Currently, `setup.py` was very out of date and claimed that TensorBoard
> supported Python which is dropped in EOY 2020.

...supported Python **2** which was dropped... <-- I assume you mean?

stephanwlee:
> ...supported Python 2 which was dropped... <-- I assume you mean?

Yup, updated the description and will make sure the commit message contains the right message.

tensorboard/pip_package/setup.py
stephanwlee: Done.

busunkim96: It may be nice to re-add python_requires with `python_requires=">=3.6"`, it will block folks with older version Python version from installing the library through pip https://packaging.python.org/guides/dropping-older-python-versions/

stephanwlee: Adding one in #5083.

PR #5073, opened 2021-06-17
dark mode: improve style of range_input & run table

stephanwlee:
Runs table was sparingly use tf-slate which was looking quite odd in the
dark mode. Instead, we now use the lighter mat-grey while the main
content is the same as the default background color.

This change also includes dark-themification of range-input which was
forgotten before.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_component.scss
@@ -193,10 +193,8 @@ $_font-size: 13px;
   }
 }
 
-@include tb-dark-theme {
-  :host.flex-layout {
-    mat-paginator {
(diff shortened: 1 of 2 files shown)


japie1235813 (approved):
No objection to this pr but I think it's better to include some screenshots in the future for those color changes. (especially the description indicates the color was 'odd', it's hard to imagine the oddness without looking at it) Thanks!

stephanwlee:
> No objection to this pr but I think it's better to include some screenshots in the future for those color changes. (especially the description indicates the color was 'odd', it's hard to imagine the oddness without looking at it) Thanks!

Sadly there are no flow that are actually using range_inputs or mat_paginator in runs table in OSS. I can supply said screenshot offline if you desire. Otherwise, our screenshot updates will be sufficient to capture the difference.

PR #5075, opened 2021-06-18
projector: allow lib to be side loaded

stephanwlee:
Internally, libraries like UMAP and threejs are loaded a bit differently
and the proxy genrule allows us to precisely add and side load the
runtime dependencies.


bmd3k (approved):

stephanwlee:
Thank you!

PR #5090, opened 2021-06-24
Move settings redux to settings feature.

bmd3k:
There are several properties that are global user settings that are currently located in the "core" feature:
* reloadEnabled
* reloadPeriodInMs
* pageSize

We move these and all associated redux to the "settings" feature. We also migrate some references to them elsewhere in the code.

We get a bit lucky with google-internal code. This change is written so that we don't have to modify any google-internal code immediately. We:
1. Reexport some settings actions and selectors from core. 
2. Hack CoreState type so that its declaration includes SettingsState.

Everything seems to build and work fine with these changes. (Googlers, see cl/381226386). After changes here are imported into Google, we will update internal references and then another PR will remove some of the hacks introduced in this change.

Testing:
* Migrated all relevant tests from core to settings.
* Ran tensorboard and played with settings and see that they work.
* Ran internal tensorboards (tbdev and tbcorp), see that they work, and see that their analytics code also still works.


psybuzz (approved):
Thanks for testing this out on internal surfaces!

I think in the long term, there might be opportunity to rethink settings, but for the immediate use case this seems a reasonable step.

Issue #2885, opened 2019-11-04
Change histogram bucketing logic to avoid dynamic shapes


hongjunChoi:
Currently TPU does not support dynamic shapes. As so, bucketing logic in tf.summary.histogram() should be outside compiled. Once dynamic shapes are supported, remove this special handling.

_Edited: rather than waiting for dynamic shape support, it would be better if we can remove the special-casing by just changing the histogram format to avoid having dynamic shapes at all._

PR #5093, opened 2021-06-24
histogram: avoid data transfer overhead for TPU outside compilation

nfelt:
This updates the histogram summary TPU outside compilation workaround logic (needed until #2885 is fixed) so that we only invoke it when summaries will actually be written, to avoid the overhead of transferring data to the CPU only to be discarded.

cc @kenfranko 

Note: this is a modified roll-forward of #5066, now that https://github.com/tensorflow/tensorflow/commit/08bdda50bc308f41fcf7ef8fdae3524476130f52 has landed in TensorFlow (to fix the issue with [`AutoOutsideCompilationWithKerasTest`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/tests/automatic_outside_compilation_test.py) that required reverting in #5070).  We also adjust the patch itself to prevent the addition of the `tf.cond` from affecting the resulting tag name; see code comments for details.

For googlers, internal tests: cl/381296270

### DIFF
--- a/tensorboard/plugins/histogram/summary_v2.py
+++ b/tensorboard/plugins/histogram/summary_v2.py
@@ -131,16 +131,22 @@ def histogram(name, data, step=None, buckets=None, description=None):
                 metadata=summary_metadata,
             )
 
-    # `_buckets()` has dynamic output shapes which is not supported on TPU's. As so, place
-    # the bucketing ops on outside compilation cluster so that the function in executed on CPU.
-    # TODO(https://github.com/tensorflow/tensorboard/issues/2885): Remove this special
-    # handling once dynamic shapes are supported on TPU's.
+    # `_buckets()` has dynamic output shapes which is not supported on TPU's.
+    # To address this, explicitly mark this logic for outside compilation so it
+    # will be executed on the CPU, and skip it entirely if we aren't actually
+    # recording summaries to avoid overhead of transferring data.
+    # TODO(https://github.com/tensorflow/tensorboard/issues/2885): Remove this
+    # special handling once the format no longer requires dynamic output shapes.
     if isinstance(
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -113,34 +114,59 @@ def histogram(name, data, step=None, buckets=None, description=None):
         or tf.summary.summary_scope
     )
 
+    # Try to capture current name scope so we can re-enter it below within our
+    # histogram_summary helper. We do this to avoid having the `tf.cond` below
+    # insert an extra `cond` into the tag name.
```
stephanwlee, line 119: This is the part I understand the least in the PR. Is it correct that without this, our summary would have tag like `some/preexsting/tag/cond/histogram_summary` when we want `some/preexsting/tag/histogram_summary` and achieve this with `tf.name_scope("") + current_scope`? If this is true, wow, `name_scope("")` is interesting.

stephanwlee (approved):

tensorboard/plugins/histogram/summary_v2.py
nfelt: Yeah, that's exactly the issue. Using `tf.name_scope("")` will "reset" the active scope to the root level, which is why this works. I guess technically that's not documented in the eager API, but it is still documented for [`Graph.name_scope`](https://www.tensorflow.org/api_docs/python/tf/Graph#name_scope) and it's how `name_scope` has always handled empty strings:

https://github.com/tensorflow/tensorflow/blob/a4dfb8d1a71385bd6d122e4f27f86dcebb96712d/tensorflow/python/framework/ops.py#L6669-L6670

PR #5096, opened 2021-06-25
dark mode: button for toggling theme in header 

stephanwlee:
This change adds a button in the header that toggles different color
theme modes in TensorBoard.

image


tensorboard/webapp/header/dark_mode_toggle_component.ts
```diff
@@ -0,0 +1,76 @@
...
+        *ngSwitchCase="DarkModeOverride.DARK_MODE_ON"
+        svgIcon="dark_mode_24px"
+      ></mat-icon>
+    </button>
+    <mat-menu #menu="matMenu">
+      <button
+        mat-menu-item
+        (click)="onOverrideChanged.emit(DarkModeOverride.DEFAULT)"
+      >
+        <label>Browser default</label>
```
japie1235813, line 51: Is the text `Default` or `Browser default`? (I'm seeing the former in the screenshot)

japie1235813 (approved):

tensorboard/webapp/header/dark_mode_toggle_component.ts
stephanwlee: It is latter now. I had changed it after taking the screenshot to make it more verbose and correct.

PR #5098, opened 2021-06-28
Refactor settings dialog so it has both container and component layers.

bmd3k:
I am about to begin some work on settings dialog and noticed that, unlike our other view components, it does not separate redux considerations from view considerations.

This change refactors redux-specific things out of dialog_component and into a new wrapping dialog_container.


tensorboard/webapp/settings/_views/dialog_component_test.ts
```diff
@@ -1,233 +0,0 @@
-/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
bmd3k, line 1: Note: I deleted this file. It seems to be a leftover from a previous refactor and is not actually run as part of testing. settings_test.ts is the actual test file now.

tensorboard/webapp/settings/_views/dialog_container.ts
```diff
@@ -0,0 +1,63 @@
...
+      [reloadPeriodInMs]="reloadPeriodInMs$ | async"
+      [pageSize]="pageSize$ | async"
+      (reloadToggled)="onReloadToggled()"
+      (reloadPeriodInMsChanged)="onReloadPeriodInMsChanged($event)"
+      (pageSizeChanged)="onPageSizeChanged($event)"
+    ></settings-dialog-component>
+  `,
+})
+export class SettingsDialogContainer {
+  readonly reloadEnabled$ = this.store.pipe(select(getReloadEnabled));
```
japie1235813, line 46: can we just use `... = this.store.select(getReloadEnabled)`?

tensorboard/webapp/settings/_views/dialog_component.ts
```diff
@@ -177,7 +159,29 @@ export class SettingsDialogComponent implements OnInit, OnDestroy {
...
+      if (change.currentValue) this.reloadPeriodControl.enable();
+      else this.reloadPeriodControl.disable();
+    }
+
+    if (changes['pageSize']) {
+      const change = changes['pageSize'];
+      if (change.previousValue !== change.currentValue) {
+        this.paginationControl.setValue(change.currentValue);
+      }
+    }
```
japie1235813, line 181: is it possible to have multiple changes at a time?
i.e., line 163 and 170 are both true

if not, can we say
```
const change  = changes['reloadPeriodInMs'] ?? (changes['reloadEnabled'] ?? (changes['pageSize']) ?? null));
if (!change) return;
```

japie1235813 (commented):
overall looks good.
yet I prefer adding some prefix on naming.
like `setting_dialog_component.ts` instead of just `dialog_component` cause it's aligned with the class name and there are other places also have dialog (e.g. color by regex)

tensorboard/webapp/settings/_views/dialog_component.ts
bmd3k: In theory it could happen that you have multiple changes at the same time.

tensorboard/webapp/settings/_views/dialog_container.ts
bmd3k: Done.

bmd3k (commented):
Prefixed file names with 'settings_', as requested.

japie1235813 (approved):
Thanks for making the changes!

PR #5101, opened 2021-06-29
router: allow onRouteIdChange to reason with route

stephanwlee:
Previously, the optional third parameter to the route contexted reducer
helper only allowed the utility to specify a state [re]setter that
reasons with the state alone. However, there are cases when we need to
set the initial value of a store based on the routeKind. For instance,
we would like our colorGroup settings to be `GroupBy.EXPERIMENT` when
user is navigating to an `COMPARE_EXPERIMENT` routeKind while keeping
it `GroupBy.RUN` in other cases. Since the initial state is a function
of the new route (`routeKind`, really), this change expands the
API signature to allow for reasoning with the route.

### DIFF
--- a/tensorboard/webapp/app_routing/BUILD
+++ b/tensorboard/webapp/app_routing/BUILD
@@ -130,6 +130,7 @@ tf_ts_library(
     ],
     deps = [
         ":internal_utils",
+        ":types",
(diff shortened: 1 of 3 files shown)


japie1235813 (approved):
review and approve commit 4df51ec

PR #5103, opened 2021-06-29
line-chart: set attribute when data is loaded once

stephanwlee:
For better WebDriver testing, we would like to distinguish not_loaded,
loading, and loaded_at_least_once. By using a separate attribute on the
vz-line-chart, we were able to discern these differing states and
ultimately made our tests a bit more stable.

This change introduces another internal state in the
data-loader-behavior with which we can set the attribute onto the
vz-line-chart.

### DIFF
--- a/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
+++ b/tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
@@ -229,13 +229,13 @@ export function DataLoaderBehavior<Item, Data>(
                   this.onLoadFinish();
                 }
                 this._loadDataAsync = null;
+                this.dataLoadedAtLeastOnce = true;
               }
               const isDataFetchPending = Array.from(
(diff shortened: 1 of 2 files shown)


tensorboard/components/tf_dashboard_common/data-loader-behavior.ts
```diff
@@ -234,6 +235,7 @@ export function DataLoaderBehavior<Item, Data>(
               ).includes(LoadState.LOADING);
               if (!isDataFetchPending) {
                 this.dataLoading = false;
+                this.dataLoadedAtLeastOnce = true;
```
psybuzz, line 238: iiuc, L238 is called in cases when the request was cancelled and there was no followup request, which is not desired.  Shouldn't this go inside the `if (!result.cancelled)` block?

psybuzz (approved):

PR #5110, opened 2021-07-01
runs: filter based on experiment name and classic name

stephanwlee:
In older TensorBoard, our run names, in experiment aware world, used to
have run names prefixed with experiment name to uniquely identify a run
in an experiment. Our users may expect that behavior in our run selector
so we implement to match the older behavior.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
@@ -1278,6 +1278,75 @@ describe('runs_table', () => {
       });
     });
 
+    it('filters by original experiment name and legacy run name', () => {
+      selectSpy
+        .withArgs(getExperiment, {experimentId: 'rowling'})
+        .and.returnValue(
+          of(
+            buildExperiment({
+              name: 'Harry Potter',
+            })
+          )
+        );
+
+      selectSpy
+        .withArgs(TEST_ONLY.getRunsLoading, jasmine.any)
+        .and.returnValue(of(false));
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -1278,6 +1278,75 @@ describe('runs_table', () => {
...
+
+      fixture.detectChanges();
+      expect(getTableRowTextContent(fixture)).toEqual([
+        ['LotR', 'The Fellowship of the Ring'],
+        ['LotR', 'The Silmarillion'],
+      ]);
+
+      // Alias for Harry Potter contains "z". If we were matching against
+      // alias + run name, Harry Potter should not match below.
+      store.overrideSelector(getRunSelectorRegexFilter, 'o[^z]+/.+S[ei]');
```
psybuzz, line 1341: Despite the comment, I found myself having to jump around a bunch to realize that the reason "The Chamber Of Secrets" matches because this regex here matches "otter/The Chamber Of Se".  imo this would be more readable if we didn't use control characters in this regex.

psybuzz (approved):

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
stephanwlee: For now, I left a comment on what the regex is expecting instead of changing the regex itself. Hopefully no one will get too confused by this test.

PR #5113, opened 2021-07-08
color grouping: group by regex, set non-matched runs to inactive

japie1235813:
We want to remove the color assignments for the non-matched runs; this will mark them as "inactive" and apply gray borders on them.
This improves the ux of the group by regex functionality. 

* Screenshots of UI changes
Before: 
Screen Shot 2021-07-07 at 5 56 20 PM
After: 
Screen Shot 2021-07-07 at 5 56 40 PM

### DIFF
--- a/tensorboard/webapp/runs/store/runs_reducers.ts
+++ b/tensorboard/webapp/runs/store/runs_reducers.ts
@@ -261,6 +261,10 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
       }
     });
 
+    for (const run of groups.nonMatches) {
+      defaultRunColorForGroupBy.set(run.id, '');
+    }
+
     return {
       ...state,
       defaultRunColorForGroupBy,
@@ -294,6 +298,10 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
           defaultRunColorForGroupBy.set(run.id, color);
         }
       });
+
+      for (const run of groups.nonMatches) {
+        defaultRunColorForGroupBy.set(run.id, '');
+      }
 
       return {
         ...state,
--- a/tensorboard/webapp/runs/store/runs_reducers.ts
+++ b/tensorboard/webapp/runs/store/runs_reducers.ts
@@ -261,8 +261,9 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
       }
     });
 
+    // unassign color for nonmatched runs to apply default unassigned style
     for (const run of groups.nonMatches) {
-      defaultRunColorForGroupBy.set(run.id, '');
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/runs/store/runs_reducers.ts
```diff
@@ -295,6 +299,10 @@ const dataReducer: ActionReducer<RunsDataState, Action> = createReducer(
         }
       });
 
+      for (const run of groups.nonMatches) {
+        defaultRunColorForGroupBy.set(run.id, '');
```
psybuzz, line 303: Allowing views to choose a 'default color' for non-matching runs is reasonable, however I would suggest making the state explicitly hold either a default color or `null` instead of empty strings.

Using an empty string means that all consumers of `getRunColorMap` **must** prefer doing `colorMap[runId] || defaultColor` instead of `colorMap[runId] ?? defaultColor`, if they want to show a default color for non-matching runs.  Making it a requirement for views to handle a `null` is slightly more obvious than an empty string I think, because TS will hopefully complain when a view tries to do `let x: string = colorMap[runId]` without handling the `null` case.

If we'd like to proceed with empty strings, could we
- Check that all consumers [1] handle the empty string, and update cases like ScalarCardContainer [2] to use `colorMap[runId] || defaultColor`
- Add comments in https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/runs/store/runs_types.ts#L50, explaining that the `defaultRunColorForGroupBy` field may contain empty string values for non-matching runs, and that `runColorOverrideForGroupBy` may not contain empty string values
- Update the comment on `getRunColorMap` [3], so that future consumers may realize that the map may contain empty string values

Or if you prefer using `null`, could we update the selector's comment to describe what a `null` value means?

Or if you prefer keeping a default color in the state, then it guarantees runs to have the same color across all views, and no further action needed.

Wdyt?

[1] https://cs.opensource.google/search?q=%22getRunColorMap)%22&sq=&ss=tensorflow%2Ftensorboard
[2] https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts#L380
[3] https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/runs/store/runs_selectors.ts#L189

japie1235813: I agree with empty string might cause unforeseen issues; null is not applicable cause `defaultRunColorForGroupBy` is `Map()` (type string is needed)
I modified it to delete the `run.id` from `defaultRunColorForGroupBy` thus all null check will work properly as before.

psybuzz: Deleting sounds good, thanks!

psybuzz (approved):

PR #5117, opened 2021-07-08
color grouping: add group by regex under feature flag

japie1235813:
We want to bring up color group by regex gradually by adding a feature flag `enableColorGroupByRegex`
The feature flag is added in global (also shown in OSS) and we want to flip the flag in tb.corp only later (no concrete plan to bring color grouping in OSS yet)

  - show "Regex" under feature flag
  - the edit row is not yet added. thus click the regex won't do any effect (cause regexString is empty and no way to edit it now)

* Screenshots of UI changes
with `?enableColorGroupByRegex=true` we can see third item on the dropdown. Click it should emit;
Screen Shot 2021-07-08 at 4 30 53 PM

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_container.ts
+++ b/tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_container.ts
@@ -37,7 +37,7 @@ import {GroupBy} from '../../types';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class RunsGroupMenuButtonContainer {
-  showGroupByRegex$?: Observable<boolean> = this.store.select(
+  showGroupByRegex$: Observable<boolean> = this.store.select(
     getEnabledColorGroupByRegex
   );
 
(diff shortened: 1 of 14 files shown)


tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_container.ts
```diff
@@ -29,12 +30,17 @@ import {GroupBy} from '../../types';
...
     <runs-group-menu-button-component
       [selectedGroupBy]="selectedGroupBy$ | async"
+      [showGroupByRegex]="showGroupByRegex$ | async"
       (onGroupByChange)="onGroupByChange($event)"
     ></runs-group-menu-button-component>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class RunsGroupMenuButtonContainer {
+  showGroupByRegex$?: Observable<boolean> = this.store.select(
```
psybuzz, line 40: Can the '?' be dropped, since this field is initialized and never set to null/undefined?

psybuzz (approved):
Nice

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_container.ts
japie1235813: ah you're right. thanks for catching this. removed

PR #5119, opened 2021-07-09
Move settings tests into top-level tf_ng_web_test_suite.

bmd3k:
Move settings tests into top-level tf_ng_web_test_suite. I either forget to run the settings tests or, if I'm working on settings features, I forget to run the top-level tests.

### DIFF
--- a/tensorboard/webapp/BUILD
+++ b/tensorboard/webapp/BUILD
@@ -324,6 +324,8 @@ tf_ng_web_test_suite(
         "//tensorboard/webapp/runs/store:testing",
         "//tensorboard/webapp/runs/views/runs_selector:runs_selector_test",
         "//tensorboard/webapp/runs/views/runs_table:runs_table_test",
+        "//tensorboard/webapp/settings/_redux:_redux_test",
+        "//tensorboard/webapp/settings/_views:test_lib",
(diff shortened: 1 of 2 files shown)


japie1235813 (approved):

japie1235813:
Not an objection but I think it's okay to separate the tests. Especially now we have too many tests in the same set.

stephanwlee:
I want to raise an objection.

CI runs both so if you don't feel safe about making a change, you can rely on GitHub actions to report that error. 

If you want all tests to run, just do bazel test webapp/....  The point of having a test target for a feature is to make bazel cachability better and make the unit test iteration faster. It is getting a bit unbearably slow and it adds to our CI time whose environment is more challenged than our machines.

bmd3k:
> I want to raise an objection.
> 
> CI runs both so if you don't feel safe about making a change, you can rely on GitHub actions to report that error.
> 
> If you want all tests to run, just do bazel test webapp/.... The point of having a test target for a feature is to make bazel cachability better and make the unit test iteration faster. It is getting a bit unbearably slow and it adds to our CI time whose environment is more challenged than our machines.

Makes sense. I didn't realize there was any motivation to it. I just saw that `notification_center` and `settings` deviated from the norm and assumed this was an error (partially introduced by me). Will abandon this PR.

PR #5121, opened 2021-07-10
core: extend page title module functionalities

yatbear:
Extend PageTitleModule to be used by TensorBoard.corp and TensorBoard.dev surfaces. 

- Uses `TensorBoard` as default page/tab title;

- If specified (and non-empty), flag `window_title` will overwrite page title in all cases; 

- If specified (and non-empty), uses the given Tensorboard brand name as page title, e.g. `TensorBoard.corp`;

- Uses non-empty experiment name to name the single experiment `${experimentName} - ${tensorboardBrandName}`, example:

Screen Shot 2021-07-12 at 2 05 23 PM

For Googlers, please refer to issue [b/189246978](https://b.corp.google.com/issues/189246978).


tensorboard/webapp/core/views/page_title_container.ts
```diff
@@ -36,12 +58,44 @@ import {State} from '../state';
...
+  );
+
+  private readonly experimentName$ = this.experiment$.pipe(
+    map((experiment) => (experiment ? experiment.name : null))
+  );
+
+  readonly title$ = this.store.select(getEnvironment).pipe(
+    combineLatestWith(this.experimentName$),
+    map(([env, experimentName]) => {
+      if (env.window_title && env.window_title.length > 0) {
```
psybuzz, line 83: We can drop the `&& env.window_title.length > 0` and the `&& experimentName.length > 0` on L86, since an empty JS string will be falsy.

tensorboard/webapp/core/views/page_title_container.ts
```diff
@@ -36,12 +58,44 @@ import {State} from '../state';
...
+        // (it's an empty string when the `--window_title` flag is not set)
+        return env.window_title;
+      } else if (experimentName && experimentName.length > 0) {
+        return experimentName;
+      } else {
+        return this.tbServiceName
+          ? `TensorBoard.${this.tbServiceName}`
```
psybuzz, line 90: Instead of making the service name only the part after "TensorBoard.", could we make it the entire string?  As a hypothetical example, something like "Vertex TensorBoard" might be desired one day, which does not begin with the "TensorBoard." prefix.

tensorboard/webapp/core/views/page_title_container.ts
```diff
@@ -36,12 +58,44 @@ import {State} from '../state';
...
+  private readonly getExperimentId$ = this.store.select(getRouteKind).pipe(
+    filter((routeKind) => routeKind === RouteKind.EXPERIMENT),
+    withLatestFrom(this.store.select(getExperimentIdsFromRoute)),
+    map(([, experimentIds]) =>
+      experimentIds && experimentIds.length === 1 ? experimentIds[0] : null
+    )
+  );
+
+  private readonly experiment$ = this.getExperimentId$.pipe(
+    filter(Boolean),
```
psybuzz, line 70: I haven't checked myself on a non-dashboard page (e.g. exp list, not found).  However, with this `filter(Boolean)`, I think the `experimentName$` observable would never emit, and the `title$` would never emit (I think combineLatestWith will wait for the first emission), so I think the PageTitleComponent just gets an initial `null`.

To ensure that an experiment list would get the serviceName initially, can we add a `startWith()` on the `title$` observable?

https://www.learnrxjs.io/learn-rxjs/operators/combination/startwith

tensorboard/webapp/core/views/page_title_container.ts
```diff
@@ -36,12 +58,44 @@ import {State} from '../state';
...
+    mergeMap((experimentId) => {
+      return this.store.select(getExperiment, {experimentId});
+    })
+  );
+
+  private readonly experimentName$ = this.experiment$.pipe(
```
psybuzz, line 76: optional: I would slightly prefer inlining this `map` into the earlier chain, and rename `experiment$` to `experimentName$`, since it is only used in 1 place.

tensorboard/webapp/core/views/page_title_test.ts
```diff
@@ -0,0 +1,132 @@
...
+      })
+    );
+    store.overrideSelector(getEnvironment, {
+      data_location: 'my-location',
+      window_title: 'I am the real title',
+    });
+    const fixture = TestBed.createComponent(PageTitleContainer);
+    fixture.detectChanges();
+
+    expect(document.title).toBe('I am the real title');
```
psybuzz, line 69: Although probably not a practical concern right now, these tests may be subject to interference from a test runner / testing framework that also updates the document.title, and if these test cases are run in the same window, nothing resets the `document.title` before the next test case.

It's a bit more work, but could we replace the prod calls to `document.title = ...` with `utils.setDocumentTitle(...)`, a utility that we can spyOn?  Then, we can make these expectations here into something like
```ts
const spy = spyOn(utils, 'setDocumentTitle');
expect(spy).toHaveBeenCalledWith('I am the real title');
```

a similar pattern can be found here:
https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/app_routing/location.ts#L126

tensorboard/webapp/core/views/page_title_container.ts
nfelt: +1 - also, for consistency maybe nice to refer to this as the "brand" (like `tbBrandName`) since that's the term we've used for the header logo text, historically:

https://github.com/tensorflow/tensorboard/blob/adb892c2dec51346ecaae69982b7c4162f5eddae/tensorboard/webapp/header/header_component.ts#L21

tensorboard/webapp/core/views/page_title_container.ts
```diff
@@ -36,12 +58,44 @@ import {State} from '../state';
...
+  );
+
+  readonly title$ = this.store.select(getEnvironment).pipe(
+    combineLatestWith(this.experimentName$),
+    map(([env, experimentName]) => {
+      if (env.window_title && env.window_title.length > 0) {
+        // (it's an empty string when the `--window_title` flag is not set)
+        return env.window_title;
+      } else if (experimentName && experimentName.length > 0) {
+        return experimentName;
```
nfelt, line 87: I'd consider always using the brand name (as discussed below) as a suffix to the title, so that the full title clearly indicates this is TensorBoard no matter what strange user-provided experiment name there might be.  This is pretty common in various other places where the primary title is user-controlled content, e.g. github pull request pages are suffixed with the repo name, google docs pages end in ` - Google Docs`, google search results end in ` - Google Search`.

So maybe `${experimentName} - ${this.tbBrandName}` here?

tensorboard/webapp/core/views/page_title_container.ts
yatbear: Removed the unnecessary checks.

tensorboard/webapp/core/views/page_title_container.ts
yatbear: Good point!

tensorboard/webapp/core/views/page_title_container.ts
yatbear: Good catch! Also changed the tests to not initialize with EXPERIMENT route to catch the scenario where the observables would not emit.

tensorboard/webapp/core/views/page_title_container.ts
yatbear: Merged into one.

tensorboard/webapp/core/views/page_title_test.ts
yatbear: Thanks for the suggestion and knowledge pass-down!

tensorboard/webapp/core/views/page_title_container.ts
yatbear: Thanks for the suggestion!

tensorboard/webapp/core/views/page_title_container.ts
```diff
@@ -36,12 +59,43 @@ import {State} from '../state';
...
+    }),
+    map((experiment) => (experiment ? experiment.name : null))
+  );
+
+  readonly title$ = this.store.select(getEnvironment).pipe(
+    combineLatestWith(this.experimentName$),
+    map(([env, experimentName]) => {
+      const tbBrandName = this.customBrandName
+        ? this.customBrandName
+        : 'TensorBoard';
```
psybuzz, line 83: Since we're using this hardcoded string again on L93, could we extract it into something like `const DEFAULT_BRAND_NAME = 'TensorBoard'` near the top of the file?

optional: `this.customBrandName || DEFAULT_BRAND_NAME` also works and is shorter, if you prefer

tensorboard/webapp/core/views/page_title_container.ts
```diff
@@ -36,12 +59,43 @@ import {State} from '../state';
...
+      }
+    }),
+    startWith(this.customBrandName ? this.customBrandName : 'TensorBoard'),
     distinctUntilChanged()
   );
 
-  constructor(private readonly store: Store<State>) {}
+  constructor(
+    private readonly store: Store<State>,
+    @Optional() @Inject(TB_BRAND_NAME) readonly customBrandName: string
```
psybuzz, line 99: nit: is it possible to make `private readonly`?

psybuzz (approved):
Nice work

tensorboard/webapp/core/views/page_title_container.ts
yatbear: Done.

tensorboard/webapp/core/views/page_title_container.ts
yatbear: The concise version is much better, thanks! Also removed the magic constant `TensorBoard` as suggested.

tensorboard/webapp/core/views/page_title_test.ts
```diff
@@ -0,0 +1,158 @@
...
+    );
+    const fixture = TestBed.createComponent(PageTitleContainer);
+    fixture.detectChanges();
+
+    expect(spy).toHaveBeenCalledWith(
+      'All you need is TensorBoard - TensorBoard'
+    );
+  });
+
+  it('uses `Tensorboard` as default tab title', () => {
```
psybuzz, line 93: Also, could we also add a test case covering the comparison route, specifically?  Right now it just shows the brand name, but might be worth considering something fancier later.

yatbear: SG, added a test case :)

PR #5123, opened 2021-07-12
color grouping: Regex edit

japie1235813:
* Motivation for features / changes!
add regex edit row to the dropdown
click to pop out the dialog
if there is no regexString set, display "(none set)" text

* Screenshots of UI changes
Bm3KCvG5tqprpzY

1. apply `?enableColorGroupByRegex=true` in url
2. navigate to TIME SERIES and click palette icon
3. click the last row (pencil icon) and text (none set)

### DIFF
--- a/tensorboard/webapp/runs/store/runs_selectors_test.ts
+++ b/tensorboard/webapp/runs/store/runs_selectors_test.ts
@@ -469,9 +469,7 @@ describe('runs_selectors', () => {
     });
 
     it('returns default empty string if user never has set one', () => {
-      const state = buildStateFromRunsState(
-        buildRunsState({})
-      );
+      const state = buildStateFromRunsState(buildRunsState({}));
 
       expect(selectors.getColorGroupRegexString(state)).toEqual('');
     });
--- a/tensorboard/webapp/runs/store/runs_selectors_test.ts
+++ b/tensorboard/webapp/runs/store/runs_selectors_test.ts
@@ -460,8 +460,6 @@ describe('runs_selectors', () => {
       const state = buildStateFromRunsState(
         buildRunsState({
           colorGroupRegexString: 'foo(\\d+)',
-          initialGroupBy: {key: GroupByKey.RUN},
-          userSetGroupByKey: GroupByKey.REGEX,
         })
       );
 
@@ -477,8 +475,6 @@ describe('runs_selectors', () => {
     it('returns regex string even if it is not user set groupby', () => {
       const state = buildStateFromRunsState(
         buildRunsState({
-          initialGroupBy: {key: GroupByKey.RUN},
-          userSetGroupByKey: GroupByKey.RUN,
           colorGroupRegexString: 'foo(\\d+)',
         })
       );
--- a/tensorboard/webapp/runs/store/runs_selectors_test.ts
+++ b/tensorboard/webapp/runs/store/runs_selectors_test.ts
@@ -449,4 +449,43 @@ describe('runs_selectors', () => {
       });
     });
   });
+
+  describe('#getColorGroupRegexString', () => {
+    beforeEach(() => {
+      // Clear the memoization.
+      selectors.getColorGroupRegexString.release();
+    });
+
+    it('returns regex string when it is group by regex', () => {
+      const state = buildStateFromRunsState(
+        buildRunsState({
+          colorGroupRegexString: 'foo(\\d+)',
+          initialGroupBy: {key: GroupByKey.RUN},
+          userSetGroupByKey: GroupByKey.REGEX,
+        })
+      );
+
+      expect(selectors.getColorGroupRegexString(state)).toEqual('foo(\\d+)');
+    });
+
+    it('returns default empty string if user never has set one', () => {
+      const state = buildStateFromRunsState(
+        buildRunsState({})
+      );
+
+      expect(selectors.getColorGroupRegexString(state)).toEqual('');
+    });
+
+    it('returns regex string even if it is not user set groupby', () => {
+      const state = buildStateFromRunsState(
+        buildRunsState({
+          initialGroupBy: {key: GroupByKey.RUN},
+          userSetGroupByKey: GroupByKey.RUN,
+          colorGroupRegexString: 'foo(\\d+)',
+        })
+      );
+
+      expect(selectors.getColorGroupRegexString(state)).toEqual('foo(\\d+)');
+    });
+  });
 });
--- a/tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ng.html
+++ b/tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ng.html
@@ -75,7 +75,7 @@ limitations under the License.
     <span>
       <mat-icon svgIcon="edit_24px"></mat-icon>
     </span>
-    <label *ngIf="regexString" >{{regexString}}</label>
+    <label *ngIf="regexString">{{regexString}}</label>
     <label *ngIf="!regexString" class="none-set-string">(none set)</label>
   </button>
 </mat-menu>
--- a/tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ng.html
+++ b/tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ng.html
@@ -68,9 +68,11 @@ limitations under the License.
   </button>
   <button
     mat-menu-item
-    role="menuitemradio"
+    role="menuitem"
+    *ngIf="showGroupByRegex"
+    [attr.aria-checked]="selectedGroupBy.key === GroupByKey.REGEX"
     (click)="onRegexStringEdit()"
-    class="displayRegexString"
+    class="display-regex-string"
   >
     <span>
       <mat-icon svgIcon="edit_24px"></mat-icon>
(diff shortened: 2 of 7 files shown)


tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ng.html
```diff
@@ -66,4 +66,16 @@
     </span>
     <label>Regex</label>
   </button>
+  <button
+    mat-menu-item
+    role="menuitemradio"
+    (click)="onRegexStringEdit()"
+    class="displayRegexString"
+  >
```
stephanwlee, line 74: Should this be only rendered under a flag?

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ts
```diff
@@ -31,9 +33,21 @@ import {GroupBy, GroupByKey} from '../../types';
...
 
+  @Input() experimentIds!: string[];
+  @Input() regexString!: string;
   @Input() selectedGroupBy!: GroupBy;
   @Input() showGroupByRegex!: boolean;
 
   @Output()
   onGroupByChange = new EventEmitter<GroupBy>();
+
+  constructor(public dialog: MatDialog) {}
```
stephanwlee, line 44: `public` -> `private readonly`

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ts
```diff
@@ -31,9 +33,21 @@ import {GroupBy, GroupByKey} from '../../types';
...
+
+  onRegexStringEdit() {
+    // data pass in the experiment id
+    const dialogRef = this.dialog.open(RegexEditDialogContainer, {
+      width: '1050px',
```
stephanwlee, line 49: Is there a way to alter the width and height? Seems like it is possible to set `maxHeight` and `maxWidth`: https://material.angular.io/components/dialog/api#MatDialogConfig

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.scss
```diff
@@ -36,4 +36,11 @@ limitations under the License.
     height: $_icon-size;
     width: $_icon-size;
   }
+
+  .displayRegexString {
```
stephanwlee, line 40: https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Name_Delimiters

`.display-regex-string`.

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.scss
```diff
@@ -36,4 +36,11 @@ limitations under the License.
...
+    padding-left: 40px;
+    .none-set-string {
+      color: gray;
```
stephanwlee, line 43: Please use one of the material color. Better yet, you should use one of the tb-theme method.

tensorboard/webapp/runs/store/runs_selectors_test.ts
```diff
@@ -449,4 +449,41 @@ describe('runs_selectors', () => {
...
+      // Clear the memoization.
+      selectors.getColorGroupRegexString.release();
+    });
+
+    it('returns regex string when it is group by regex', () => {
+      const state = buildStateFromRunsState(
+        buildRunsState({
+          colorGroupRegexString: 'foo(\\d+)',
+          initialGroupBy: {key: GroupByKey.RUN},
+          userSetGroupByKey: GroupByKey.REGEX,
```
stephanwlee, line 464: If these are irrelevant to the spec, omit it. It makes the test harder to maintain when specs have too many unnecessary setup code. Here and below.

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -193,6 +195,7 @@ describe('runs_table', () => {
         NoopAnimationsModule,
         FilterInputModule,
         RangeInputModule,
+        MatDialogModule,
```
stephanwlee, line 198: Sort for consistency/readability.

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -604,10 +609,10 @@ describe('runs_table', () => {
 
           expect(
             items.map((element) => element.getAttribute('aria-checked'))
-          ).toEqual(['true', 'false', 'false']);
+          ).toEqual(['true', 'false', 'false', null]);
```
stephanwlee, line 612: Shouldn't the last one be `aria-checked: false` and not `null`? If possible, I think we should check a11y spec to see what type of roles we are supposed to use for this type of UI.

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -669,6 +674,14 @@ describe('runs_table', () => {
...
           })
         );
+
+        regexEdit.click();
+        const dialogContainer = overlayContainer
+          .getContainerElement()
+          .querySelector('mat-dialog-container');
+        expect(dialogContainer).toBeTruthy();
+        expect(dialogContainer!.textContent).toContain('Save');
+        expect(dialogContainer!.textContent).toContain('Cancel');
```
stephanwlee, line 684:
```ts
const dialogContainer = overlayContainer
          .getContainerElement()
          .querySelector('mat-dialog-container');
const cancelButton, saveButton = dialogContainer.querySelectorAll('button');
expect(cancelButton!.textContent).toBe('Cancel');
expect(saveButton!.textContent).toBe('Save');
```

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -27,6 +27,7 @@ import {
   TestBed,
 } from '@angular/core/testing';
 import {MatCheckboxModule} from '@angular/material/checkbox';
+import {MatDialogModule} from '@angular/material/dialog';
```
stephanwlee, line 30: Make sure BUILD file has dependency on expect_mat_dialog.

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ng.html
```diff
@@ -66,4 +66,16 @@
     </span>
     <label>Regex</label>
   </button>
+  <button
+    mat-menu-item
+    role="menuitemradio"
```
stephanwlee, line 71: This should be `menutem` since this is not a radio button item.

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
japie1235813: yes it does have  `"//tensorboard/webapp/angular:expect_angular_material_dialog",`

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ts
japie1235813: ah, sorry I forget about this. updated.

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ng.html
japie1235813: yes it should be. overlooked. added

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.scss
```diff
@@ -36,4 +36,11 @@ limitations under the License.
     height: $_icon-size;
     width: $_icon-size;
   }
+
+  .display-regex-string {
+    padding-left: 40px;
+    .none-set-string {
+      color: map-get($tb-foreground, secondary-text);
```
stephanwlee, line 43: use `@include tb-theme-foreground-prop(color, secondary-text);` instead.

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
japie1235813: this looks better. thanks!

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
japie1235813: add `aria-checked` in html

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ng.html
```diff
@@ -66,4 +66,18 @@
     </span>
     <label>Regex</label>
   </button>
+  <button
+    mat-menu-item
+    role="menuitem"
+    *ngIf="showGroupByRegex"
+    [attr.aria-checked]="selectedGroupBy.key === GroupByKey.REGEX"
```
stephanwlee, line 73: This should be removed; multi-select radio button does not make sense and it does not make sense for `menuitem` to have `aria-checked` property.

stephanwlee (approved):

tensorboard/webapp/runs/views/runs_table/runs_group_menu_button_component.ng.html
japie1235813: so then the test `renders a check icon and aria-checked for the current groupBy menu item` the last one is expected to be `null`

stephanwlee: Ack. We should be `querySelectorAll`ing `menuitemradio` instead in the test.

japie1235813: updated with "menuitemradio" and remove `null` etc in the test check

PR #5124, opened 2021-07-13
fix build and lint errors in google3

yatbear:
Fixes for:
- Deprecation warnings for selectors with props (GetExperiment)
- Type error for `customBrandName` (should accept `null`)
- `TB_BRAND_NAME` initialization error ([details](https://fusion2.corp.google.com/invocations/dfcc9be9-250e-4bf8-8182-b7588ea9267b/targets/%2F%2Fthird_party%2Ftensorboard%2Fwebapp%2Fcore%2Fviews:test_lib;config=b7b0423d5e2c3613de291426e5174dedca769b7600ec4129989b00744534613e/log))

### DIFF
--- a/tensorboard/webapp/core/views/page_title_container.ts
+++ b/tensorboard/webapp/core/views/page_title_container.ts
@@ -72,6 +72,9 @@ export class PageTitleContainer {
   private readonly experimentName$ = this.getExperimentId$.pipe(
     filter(Boolean),
     mergeMap((experimentId) => {
+      // Selectors with props are deprecated (getExperiment):
+      // https://github.com/ngrx/platform/issues/2980
+      // tslint:disable-next-line:deprecation
       return this.store.select(getExperiment, {experimentId});
     }),
     map((experiment) => (experiment ? experiment.name : null))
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/core/views/page_title_container.ts
```diff
@@ -96,6 +99,8 @@ export class PageTitleContainer {
 
   constructor(
     private readonly store: Store<State>,
-    @Optional() @Inject(TB_BRAND_NAME) private readonly customBrandName: string
+    @Optional()
+    @Inject(TB_BRAND_NAME)
+    private readonly customBrandName: string | null
```
stephanwlee, line 104: Just wondering, for an `Optional` param, it is `undefined` and not `null`, right?

stephanwlee (approved):

tensorboard/webapp/core/views/page_title_container.ts
yatbear: Fixed. Thanks for the lightning-fast response!

psybuzz: Just jumping in to double check, does `@Optional` actually return `undefined` when the dependency is not provided?  These official Angular docs suggest it returns `null`, so not sure if the Angular site is just outdated:

"The @Optional property decorator tells Angular to return null when it can't find the dependency."
https://angular.io/guide/dependency-injection-in-action#make-a-dependency-optional-and-limit-search-with-host
https://angular.io/api/core/Optional

yatbear: This is fixed in #5126, thanks Stephan!

PR #5125, opened 2021-07-13
theme: style mat-button not as hyperlink

stephanwlee:
There are some cases when a hyperlink is visually styled as a material
button. Due to our recent changes that globally style all hyperlinks,
those hyperlink-mat-buttons were also styled like one and it has caused
visual jank. This change undoes that by styling hyperlinks without
mat-*-button classes as hyperlink.

Prior to this change, "Learn More" below were styled as purple.
image


psybuzz (approved):
Thanks!

PR #5127, opened 2021-07-13
color grouping: fix expid pass bug

japie1235813:
When passing down the experiment id, it was overlooked in rund_group_menu_componet, which results in `OnRegexEdit` fails to function. Added it back and attach tests about regex editing.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
@@ -95,6 +95,9 @@ import {RunsGroupMenuButtonContainer} from './runs_group_menu_button_container';
 import {RunsTableComponent} from './runs_table_component';
 import {RunsTableContainer, TEST_ONLY} from './runs_table_container';
 import {HparamSpec, MetricSpec, RunsTableColumn} from './types';
+enum REGEXEDIT {
+  REGEXEDIT = 4,
+}
 
 @Injectable()
 class ColorPickerTestHelper {
@@ -181,6 +184,23 @@ describe('runs_table', () => {
     return Array.from(
       overlayContainer.getContainerElement().querySelectorAll('[mat-menu-item]')
     );
+  }
+
+  function getColorGroupByHTMLElement(KEY: GroupByKey | REGEXEDIT) {
+    const items = getOverlayMenuItems();
+    const [experiment, run, regex, regexEdit] = items as HTMLElement[];
+    switch (KEY) {
+      case GroupByKey.RUN:
+        return run;
+      case GroupByKey.EXPERIMENT:
+        return experiment;
+      case GroupByKey.REGEX:
+        return regex;
+      case REGEXEDIT.REGEXEDIT:
+        return regexEdit;
+      default:
+        return null;
+    }
   }
 
   beforeEach(async () => {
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -691,6 +691,93 @@ describe('runs_table', () => {
...
+            experimentIds: ['book'],
+            groupBy: {key: GroupByKey.REGEX, regexString: ''},
+          })
+        );
+
+        (overlayContainer
+          .getContainerElement()
+          .querySelector(
+            'mat-dialog-container input'
+          ) as HTMLInputElement).value = 'foo(\\d+)';
```
stephanwlee, line 731: Use `sendKeys` on webapp/testing/dom.ts

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -691,6 +691,93 @@ describe('runs_table', () => {
...
+          [RunsTableColumn.RUN_NAME, RunsTableColumn.RUN_COLOR]
+        );
+
+        const menuButton = fixture.debugElement
+          .query(By.directive(RunsGroupMenuButtonContainer))
+          .query(By.css('button'));
+        menuButton.nativeElement.click();
+
+        const items = getOverlayMenuItems();
+        const [experiments, runs, regex, regexEdit] = items as HTMLElement[];
```
stephanwlee, line 756: Create a little utility method with semantic name for this instead.

japie1235813: Please take a look. thanks!

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -182,6 +186,23 @@ describe('runs_table', () => {
     );
   }
 
+  function getColorGroupByHTMLElement(KEY: GroupByKey | REGEXEDIT) {
```
stephanwlee, line 189: KEY -> key. It does not have to be ALL_CAPS

stephanwlee: Please specify the return type, too.

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -94,6 +95,9 @@ import {RunsGroupMenuButtonContainer} from './runs_group_menu_button_container';
 import {RunsTableComponent} from './runs_table_component';
 import {RunsTableContainer, TEST_ONLY} from './runs_table_container';
 import {HparamSpec, MetricSpec, RunsTableColumn} from './types';
+enum REGEXEDIT {
+  REGEXEDIT = 4,
```
stephanwlee, line 99: since `4` can collide with future enum value, put something ridiculous or just use a string.

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
stephanwlee: `REGEXEDIT` here is an object, not a literal `4`. IDK how it is matching against the switch case L199 but please change it to `key: GroupByKey | 'regexEdit'` or something like this.

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -691,6 +708,90 @@ describe('runs_table', () => {
...
+        store.overrideSelector(getRunGroupBy, {key: GroupByKey.EXPERIMENT});
+        const fixture = createComponent(
+          ['book'],
+          [RunsTableColumn.RUN_NAME, RunsTableColumn.RUN_COLOR]
+        );
+
+        const menuButton = fixture.debugElement
+          .query(By.directive(RunsGroupMenuButtonContainer))
+          .query(By.css('button'));
+        menuButton.nativeElement.click();
```
stephanwlee, line 769: No action required: well, I wanted this part to be within the utility function but it is fine as is.

stephanwlee (approved):

PR #5128, opened 2021-07-14
core: make title$ also listen to route kind observable

yatbear:
To ensure that title$ emits even when experimentName$ doesn't, so that when navigating away from single experiment page, page title would change accordingly.


psybuzz (commented):
Thanks, since the "experiment -> experiment list" case was caught later, could we add a test case that covers this?

yatbear:
> Thanks, since the "experiment -> experiment list" case was caught later, could we add a test case that covers this?

Added a test case.

psybuzz (approved):
Thanks!

PR #5129, opened 2021-07-14
color grouping: group by regex, fill example and some diaglog wording

japie1235813:
- move cancel/save buttons to the right bottom
- add title and intro paragraph (see doc)
- click `Try (...)` to fill the example

Screen Shot 2021-07-16 at 11 38 25 PM

Note:
- UI finalization WIP and screenshot tests come later
- live grouping result updates not included

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
+++ b/tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
@@ -14,12 +14,14 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
+<h1 mat-dialog-title>Color runs by regex</h1>
+
 <mat-dialog-content>
+  <p>Enter a regex with capturing groups to match against run names:</p>
   <mat-form-field>
     <input
       matInput
       #regexStringInput
-      placeholder="(\w+/).+"
       value="{{ regexString }}"
       (keydown.enter)="onEnter($event.target.value)"
       cdkFocusInitial
@@ -27,10 +29,23 @@ limitations under the License.
   </mat-form-field>
 </mat-dialog-content>
 
-<div mat-dialog-actions>
+<div class="example-details">
+  <p>
+    Each matching run will be assigned a color based on the "key" formed by its
+    matches to the capturing groups. <br />
+    <button mat-stroked-button (click)="fillExample()">
+      Try <code>(train|eval)</code>
+    </button>
+    to assign all runs containing <code>train</code> to one color and all runs
+    containing <code>eval</code> to another color.
+  </p>
+</div>
+
+<div mat-dialog-actions align="end">
   <button mat-button mat-dialog-close>Cancel</button>
   <button
-    mat-button
+    mat-raised-button
+    color="primary"
     mat-dialog-close
     (click)="onSaveClick(regexStringInput.value)"
   >
--- a/tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
+++ b/tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
@@ -33,7 +33,7 @@ limitations under the License.
   <p>
     Each matching run will be assigned a color based on the "key" formed by its
     matches to the capturing groups. <br />
-    <button mat-stroked-button (click)="fillExample()">
+    <button (click)="fillExample('(train|eval)')">
       Try <code>(train|eval)</code>
     </button>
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
```diff
@@ -14,23 +14,38 @@
...
     />
   </mat-form-field>
 </mat-dialog-content>
 
-<div mat-dialog-actions>
+<div class="example-details">
+  <p>
+    Each matching run will be assigned a color based on the "key" formed by its
+    matches to the capturing groups. <br />
+    <button mat-stroked-button (click)="fillExample()">
```
stephanwlee, line 36: Perhaps fill the argument and write `"fillExample('(train|eval)')"` so the example regex all in HTML.

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
```diff
@@ -14,23 +14,38 @@
...
+      Try <code>(train|eval)</code>
+    </button>
+    to assign all runs containing <code>train</code> to one color and all runs
+    containing <code>eval</code> to another color.
```
stephanwlee, line 40: As you said earlier, we may want to have a theme variable for styling the `code` so it looks more like GitHub `code`.

stephanwlee (approved):

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
nfelt: We might want to consider just using a link (i.e. underline and maybe also blue coloring) for the live example?  The inline button is a little cramped-looked IMO, and I think a link would still be fairly obviously a click target.

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
japie1235813: defer the changes to next pr

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
japie1235813: - updated with '(train|eval)' in HTML
- replaced button with plain text link

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
```diff
@@ -14,23 +14,38 @@
...
     />
   </mat-form-field>
 </mat-dialog-content>
 
-<div mat-dialog-actions>
+<div class="example-details">
+  <p>
+    Each matching run will be assigned a color based on the "key" formed by its
+    matches to the capturing groups. <br />
+    <a href="#" (click)="fillExample('(train|eval)')">
```
stephanwlee, line 36: This is semantically a button, not an anchor element that takes you to a different page. Please use semantic element, `<button>` here. You don't have to use mat-button for this.

nfelt: Do you mean use `<button>` but it doesn't have to look like a button?  I suggested making it appear like a link earlier since I think something more text-oriented would fit better in an inline/paragraph context.  But I guess it could still use `<button>` like these link-lookalikes here on Github do:

Screen Shot 2021-07-16 at 12 23 23 PM

If you mean something else maybe we should come to some consensus first before asking Jie-Wei to change it back and forth.

stephanwlee: > Do you mean use <button> but it doesn't have to look like a button? I suggested making it appear like a link earlier since I think something more text-oriented would fit better in an inline/paragraph context. But I guess it could still use <button> like these link-lookalikes here on Github do:

Correct. For accessibility and other purposes, you want button to use `<button>`. 

Yes, I am not objecting to your feedback of wanting to see a link-like UI for the button.

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_component.scss
```diff
@@ -0,0 +1,29 @@
...
+  button {
+    background-color: transparent;
+    padding: 0;
+    border: none;
+    cursor: pointer;
+    text-decoration: underline;
+    color: map-get($tb-foreground, link);
+    &:visited {
+      color: map-get($tb-foreground, link-visited);
+    }
```
stephanwlee, line 27: You should use
`@include tb-theme-foreground-prop(link, color);` here and make changes for visited.

stephanwlee (approved):

PR #5130, opened 2021-07-14
core: introduce persistent_settings

stephanwlee:
TensorBoard generally is a forgetful application; when user changes
settings, TensorBoard does not often remember it that a new browser tab
would forget the setting. Users will painstakingly reset the settings to
their liking every time one opens TensorBoard. There were several
efforts to make that better by remembering in local storage. Instead of
fractured support for it all over the code base, this change attempts to
introduce a new core feature, Persistent Settings.

Persistent settings is built with eventual future where we have a proper
settings storage in the backend (not just local storage). As such, the
feature grabs settings from multiple ngrx features to interact with the
storage layer efficiently.

Do note that this change distinguishes a global settings vs.
per-experiment settings. There are settings that should be remembered on
per experiment basis while others are more experiment global. This
change only addresses the latter kind as a start.


tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source.ts
```diff
@@ -0,0 +1,121 @@
...
+ * An implementation of PersistentSettingsDataSource that stores global settings
+ * in browser local storage.
+ */
+@Injectable()
+export class PersistentSettingsDataSourceImpl
+  implements PersistentSettingsDataSource {
+  constructor(private readonly localStorage: LocalStorage) {}
+
+  private serializeSettings(settings: Partial<PersistableSettings>): string {
+    const serializableSettings: BackendSettings = {
```
psybuzz, line 42: A lot of neat stuff going on in this PR!  I have a couple high level questions:
- How would an internal-only TensorBoard, or something like TensorBoard.dev define persistent settings?  Wdyt about losing type safety of `BackendSettings`, and instead of hardcoding metrics-specific settings here, just treating it as a bag of key/values?
- There is already a 'Settings' feature/module.  Would it make sense to eventually merge the 'PersistentSettings' feature/module, or is the plan to maintain both?

Also, thanks for migrating the legacy settings to the new scheme in this PR!

stephanwlee: > How would an internal-only TensorBoard, or something like TensorBoard.dev define persistent settings? Wdyt about losing type safety of BackendSettings, and instead of hardcoding metrics-specific settings here, just treating it as a bag of key/values?

Losing type safety is not great so I took an approach to make use of generics and custom converter instead. Simple key-value will surely bite us when properties get mangled since internal types in metrics will get mangled and will not map nicely onto unmangled key-value settings (and at some point/layer, we need to do a translation between unmangled <-> mangled).

> There is already a 'Settings' feature/module. Would it make sense to eventually merge the 'PersistentSettings' feature/module, or is the plan to maintain both?

Plan is to maintain both. Putting persistence logic inside the settings made it quite busy and it does not pain me much to have a separate folder for "featureless" feature.

japie1235813: Thanks for the explanation. `Persisted settings` and `Settings` are both called "settings". How do you plan to make it clearly that they're different if the plan is to maintain both? 
Can we change `Persisted settings` or `Setting` to `experiment configuration`, which sounds more intuitive to me?

tensorboard/webapp/persistent_settings/_redux/persistent_settings_actions.ts
```diff
@@ -0,0 +1,30 @@
...
+import {PersistableSettings} from '../_data_source/types';
+
+/** @typehack */ import * as _typeHackModels from '@ngrx/store/src/models';
+
+/**
+ * Describes settings loaded from a global settings storage. Dispatched once
+ * when the application bootstraps.
+ */
+export const globalSettingsLoaded = createAction(
+  '[Persistent Settings] Global Settings Loaded',
```
japie1235813, line 26: I'm confused.  I thought this pr is not for any global settings (across experiments) but for per experiment?

japie1235813 (commented):
TBH this pr is too large to review.  Is there any way to split them into small PRs?

tensorboard/webapp/persistent_settings/_redux/persistent_settings_actions.ts
stephanwlee: It is exactly the opposite.

stephanwlee:
@japie1235813 reverted bulk of changes and left the essentials that I consider as a unit that is required to understand some part of critical logic. Although _redux folder would be required to understand the glue between config_module and _data_source, I hope you are okay with seeing that implementation in a follow up PR.

tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source.ts
stephanwlee: > Can we change Persisted settings or Setting to experiment configuration, which sounds more intuitive to me?

The persisted_settings module is only responsible with persisting the data (I was thinking about calling it "User preference" but an obvious question someone will ask is, '"Preference" vs. "Settings": they are synonym! why?'. This is why I stuck with "settings". I can leave some module doc to explain the difference for now.

tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source.ts
```diff
@@ -0,0 +1,146 @@
...
+  abstract uiToBackend(
+    uiSettings: Partial<UiSettings>
+  ): Partial<StorageSettings>;
+  abstract backendToUi(
+    backendSettings: Partial<StorageSettings>
+  ): Partial<UiSettings>;
+}
+
+@Injectable()
+export class OSSSettingsConverter extends SettingsCoverter<
```
japie1235813, line 42: the naming`OSSSettingsConverter` makes me wonder: Is there any other converter in your mind that it also extends `SettingsCoverter`?

tensorboard/webapp/persistent_settings/_data_source/types.ts
```diff
@@ -0,0 +1,35 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+/**
+ * Global settings that the backend remembers. `declare`d so it does not get
+ * mangled or mangled differently when compiler changes.
+ */
```
japie1235813, line 19: the "compiler" here means it is either localstorage or (future backend support?)

what are the "keys" in your mind that these two interfaces could be different?
(then we need a converter for it)

tensorboard/webapp/persistent_settings/persistent_settings_module.ts
```diff
@@ -0,0 +1,30 @@
...
+import {NgModule} from '@angular/core';
+
+import {PersistentSettingsConfigModule} from './persistent_settings_config_module';
+import {PersistentSettingsDataSourceModule} from './_data_source/persistent_settings_data_source_module';
+
+/**
+ * Persistent Settings module is responsible for persisting and loading settings
+ * from other features.
+ *
+ * For settings of the app, please refer to "settings" feature instead.
```
japie1235813, line 24: I cannot understand at once from this docString. what are "other features" and what are "settings of the app"
Can we add some examples on the description?

tensorboard/webapp/persistent_settings/persistent_settings_module.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+@NgModule({
+  imports: [PersistentSettingsDataSourceModule],
+  providers: [PersistentSettingsConfigModule],
```
japie1235813, line 28: is `PersistentSettingsModule` just a wrap for `PersistentSettingsConfigModule`?
but the usage is still using `PersistentSettingsConfigModule`? (e.g. `PersistentSettingsConfigModule.defineGlobalSetting` in `metrics_module.ts` in earlier pr version)

tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source_test.ts
```diff
@@ -0,0 +1,233 @@
...
+  let getItemSpy: jasmine.Spy;
+
+  async function configure<UiSettings, StorageSettings>(
+    Converter: new () => SettingsCoverter<UiSettings, StorageSettings>
+  ) {
+    await TestBed.configureTestingModule({
+      imports: [LocalStorageTestingModule],
+      providers: [
+        PersistentSettingsDataSourceImpl,
+        {provide: SettingsCoverter, useClass: Converter},
```
japie1235813, line 42: why provider here only "PersistentSettingsDataSourceImpl" but not
```
   {
      provide: PersistentSettingsDataSource,
      useClass: PersistentSettingsDataSourceImpl,
    },
```

tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source.ts
stephanwlee: Yes, as Erik asked, internal versions like TB.dev will require more settings to be persisted on top of OSS ones and this is recognizing that.

tensorboard/webapp/persistent_settings/persistent_settings_module.ts
stephanwlee: `PersistentSettingsConfigModule` and `PersistentSettingsModule` are different. Former is a module that only provides settings _configuration_ and `PersistentSettingsModule` (really, effects that is exposed as part of `PersistentSettingsModule`) uses ConfigModule. I hope this helps.

tensorboard/webapp/persistent_settings/persistent_settings_module.ts
stephanwlee: Left a comment in the code. PTAL.

tensorboard/webapp/persistent_settings/_data_source/types.ts
stephanwlee: Clarified the "compiler" bit. PTAL.

tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source_test.ts
stephanwlee: Great question. Below, we `TestBed.inject`  the `PersistentSettingsDataSourceImpl` and only test the implementation of `PersistentSettingsDataSourceImpl`. Whether we provide it under `PersistentSettingsDataSource` InjectionToken or not does not matter for this test.

japie1235813 (approved):

tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source.ts
```diff
@@ -0,0 +1,146 @@
...
+const GLOBAL_LOCAL_STORAGE_KEY = '_tb_global_settings';
+
+@Injectable()
+export abstract class PersistentSettingsDataSource<UiSettings> {
+  abstract setSettings(partialSetting: Partial<UiSettings>): Observable<void>;
+  abstract getSettings(): Observable<Partial<UiSettings>>;
+}
+
+@Injectable()
+export abstract class SettingsCoverter<UiSettings, StorageSettings> {
```
psybuzz, line 32: nit: Coverter -> Converter

here and below

tensorboard/webapp/persistent_settings/persistent_settings_config_module.ts
```diff
@@ -0,0 +1,82 @@
...
+  getGlobalSettingSelectors(): SettingSelector<State, Settings>[] {
+    return this.globalSettingSelectors ?? [];
+  }
+
+  /**
+   * Registers a global setting that is to be persisted when a store emits a
+   * change. For per-experiment settings, please contact TensorBoard team
+   * member.
+   *
+   * Example usage:
```
psybuzz, line 55: This example is nice.  It seems like after `defineGlobalSetting`, adding a new converter is always required.  If so, could we also mention in this usage comment that a converter is also necessary and why?

PR #5132, opened 2021-07-15
Add optional repository_id to Experiment

ericdnielsen:
Extends the Experiment interface with an optional repository_id field.

### DIFF
--- a/tensorboard/webapp/experiments/types.ts
+++ b/tensorboard/webapp/experiments/types.ts
@@ -14,9 +14,9 @@ limitations under the License.
 ==============================================================================*/
 export declare interface Experiment {
   id: string;
+  repository_id?: string;
   name: string;
   start_time: number;
-  repository_id?: string;
   owner?: string;
   description?: string;
   hparams?: string;
--- a/tensorboard/webapp/experiments/types.ts
+++ b/tensorboard/webapp/experiments/types.ts
@@ -16,6 +16,7 @@ export declare interface Experiment {
   id: string;
   name: string;
   start_time: number;
+  repository_id?: string;
   owner?: string;
   description?: string;
   hparams?: string;


psybuzz (approved):

tensorboard/webapp/experiments/types.ts
```diff
@@ -16,6 +16,7 @@ export declare interface Experiment {
   id: string;
   name: string;
   start_time: number;
+  repository_id?: string;
```
nfelt, line 19: Can we match the order of properties from the internal interface (i.e. as approached in cl/384754058)?  I realize they aren't entirely 1:1 identical but they're close enough that we might as well not diverge.

ericdnielsen: Do we care about keeping all the non-optional fields first, followed by optional?  That was my reason for the divergence here.

nfelt: I'm not sure how critical that is (maybe @stephanwlee or @psybuzz have a style opinion).  That said, the status quo we have here around which fields are and aren't optional is a bit odd to me in the first place.  It's not particularly obvious why `name` and `start_time` are always required but `owner` and `description` are not (especially given that in practice for our internal service both name and description _can_ be empty).

Semantically speaking, it seems appropriate to me that `repository_id` would ultimately be mandatory as a field (even if sometimes the value it takes is the empty string), even if we don't choose to make that the case now due to having to refactor existing code.  So I think my inclination would still be to stick with the same field order.  But since I'm not that familiar with TypeScript conventions in this respect I'd be happy to hear if there are particular reasons why we might want the semantic "required-ness" or precedence of fields to be different between here and the Python interface.

stephanwlee: >  why name and start_time are always required but owner and description are not (especially given that in practice for our internal service both name and description can be empty).

The idea is that the Experiment can one day be used in OSS context where strategies for `owner` (guess we can get it from `$USER`) and description are not clear. This type definition is not made to match that of one backend implementation but it encodes assumption that frontend is making of the backend system*s*.

ericdnielsen: Reordered as requested.  http_api.md doesn't have an exhaustive list of fields at present, so I don't think we need to change anything there.

psybuzz: my 2 cents:
> Do we care about keeping all the non-optional fields first, followed by optional?

I am not aware of any style mandate, so a mixed order of optional/non-optional that matches the field order of the backend works for me.

> Semantically speaking, it seems appropriate to me that repository_id would ultimately be mandatory as a field

Things are clearer, less surprising to me if the requiredness of fields matches the "backend".  Specifically, the common requiredness across all backends.  If some internal TensorBoard.X sets the field, while, say, TensorBoard.dev's backend intends to never set it for the forseeable future, an optional field here in TS makes most sense to me.  That said, it's unclear whether all TensorBoard variants intend to support repository_id eventually, so practically keeping it optional is very reasonable imo.

nfelt: Thanks everyone for the insight, that makes sense to me.  I think Erik's suggestion around what's common across backends sounds reasonable, maybe it's just that since the hypothetical OSS "Experiment" backend isn't fully specified there's some ambiguity about which fields it would support.

PR #5134, opened 2021-07-17
persistent_settings: glue config and data_source with Redux

stephanwlee:
This change includes Redux changes that glues the Settings configuration
with DataSource. When there are configurations for global settings, the effects
now subscribe to the store changes, and when value changes, we persist the
setting in the local storage using the DataSource.

Note that this change does not yet migrate Metrics or notification_center to
use the new persistent_settings.

### DIFF
--- a/tensorboard/webapp/persistent_settings/_redux/persistent_settings_effects_test.ts
+++ b/tensorboard/webapp/persistent_settings/_redux/persistent_settings_effects_test.ts
@@ -139,7 +139,6 @@ describe('persistent_settings effects test', () => {
         getSettingsSpy.and.returnValue(of({}));
         action.next(TEST_ONLY.initAction());
 
-        tick();
         tick(TEST_ONLY.DEBOUNCE_PERIOD_IN_MS * 2);
         expect(setSettingsSpy).not.toHaveBeenCalled();
       }));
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/persistent_settings/_redux/persistent_settings_effects_test.ts
```diff
@@ -0,0 +1,211 @@
...
+      }));
+
+      it('ignores value emitted from initial subscription', fakeAsync(() => {
+        store.overrideSelector(setSmoothingSelector, {scalarSmoothing: 0.1});
+
+        getSettingsSpy.and.returnValue(of({}));
+        action.next(TEST_ONLY.initAction());
+
+        tick();
+        tick(TEST_ONLY.DEBOUNCE_PERIOD_IN_MS * 2);
```
japie1235813, line 143: why we need two `tick()` here? omit the first one?

japie1235813 (approved):

tensorboard/webapp/persistent_settings/_redux/persistent_settings_effects_test.ts
stephanwlee: It was unnecessary. Removed. Thanks for catching that.

PR #5135, opened 2021-07-19
metric: use persistent setting storage

stephanwlee:
Metrics dashboard, for storing smoothing override values, used local
storage directly and, instead, now it uses the common globla persistent
settings infrastructure.


japie1235813 (approved):
LGTM! thanks for splitting the previous pr into smaller ones. This is easier to follow.

PR #5136, opened 2021-07-19
feature_flag: dark mode in persistent storage

stephanwlee:
TensorBoard previously forgot user's setting on theme. This change uses
the newly introduced Persistent Settings storage system to store the
value under the local storage.

~Not ready: the change currently includes changes from #5135.~


tensorboard/webapp/feature_flag/store/feature_flag_reducers_test.ts
```diff
@@ -106,4 +107,47 @@ describe('feature_flag_reducers', () => {
...
+        }),
+      });
+
+      const state1 = reducers(
+        prevState,
+        globalSettingsLoaded({
+          partialSettings: {},
+        })
+      );
+      expect(state1.flagOverrides!.enableDarkModeOverride).toBe(false);
```
japie1235813, line 126: why this is `false` but not default value `null`?
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/feature_flag/store/feature_flag_store_config_provider.ts;l=25

japie1235813 (approved):

tensorboard/webapp/feature_flag/store/feature_flag_reducers_test.ts
stephanwlee: It is because the initial value, in the test, is set to `false` in L116.
```
        flagOverrides: buildFeatureFlag({
          enableDarkModeOverride: false,
        }),
```

PR #5140, opened 2021-07-20
router: allow programmatical route definition

stephanwlee:
Previously, we only allowed very static route definition. Concrete route
aside, even the redirection route definition took path for "from" and
"to".

However, there now are the cases when we want to redirect, depending on
incomign path parts, redirect to different paths or route kinds. Now,
with this change, you can specify a definition that would be invoked on
a path match by invoking app specified `redirector`.

* Screenshots of UI changes

### DIFF
--- a/tensorboard/webapp/app_routing/route_config.ts
+++ b/tensorboard/webapp/app_routing/route_config.ts
@@ -12,7 +12,14 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-import {ConcreteRouteDef, RouteDef} from './route_config_types';
+import {
+  ConcreteRouteDef,
+  isConcreteRouteDef,
+  isStaticRedirectionRouteDef,
+  ProgrammticRedirectionRouteDef,
+  RedirectionRouteDef,
+  RouteDef,
+} from './route_config_types';
 import {Route, RouteKind, RouteParams} from './types';
 
 interface NegativeMatch {
@@ -70,19 +77,66 @@ function getPathFragments(path: string): PathFragment[] {
   });
 }
 
+enum RouteType {
+  CONCRETE,
+  STATIC_REDIRECTION,
+  PROGRAMMATIC_REDIRECTION,
+}
+
+interface ConcreteRouteConfigMatcher {
+  routeType: RouteType.CONCRETE;
+  definition: ConcreteRouteDef;
+}
+
+interface StaticRedirectionRouteConfigMatcher {
+  routeType: RouteType.STATIC_REDIRECTION;
+  definition: RedirectionRouteDef;
+}
+
+interface ProgrammaticalRouteConfigMatcher {
+  routeType: RouteType.PROGRAMMATIC_REDIRECTION;
+  definition: ProgrammticRedirectionRouteDef;
+}
+
+type MatcherConfiguration =
+  | ConcreteRouteConfigMatcher
+  | StaticRedirectionRouteConfigMatcher
+  | ProgrammaticalRouteConfigMatcher;
+
 class RouteConfigMatcher {
   private readonly pathFragments: PathFragment[];
   private readonly pathMatchers: PathMatcher[];
   private readonly redirectionFragments: PathFragment[] | null;
+  readonly config: MatcherConfiguration;
 
-  constructor(readonly config: RouteDef) {
-    this.validateConfig(config);
-    this.pathFragments = getPathFragments(config.path);
+  constructor(routeDef: RouteDef) {
+    this.validateConfig(routeDef);
+    this.config = this.getConfig(routeDef);
+    this.pathFragments = getPathFragments(routeDef.path);
     this.pathMatchers = this.getPathMatchers(this.pathFragments);
     this.redirectionFragments =
-      config.routeKind === null
-        ? getPathFragments(config.redirectionPath)
+      this.config.routeType === RouteType.STATIC_REDIRECTION
+        ? getPathFragments(this.config.definition.redirectionPath)
         : null;
+  }
+
+  private getConfig(routeDef: RouteDef): MatcherConfiguration {
+    if (isConcreteRouteDef(routeDef)) {
+      return {
+        routeType: RouteType.CONCRETE,
+        definition: routeDef,
+      };
+    }
+    if (isStaticRedirectionRouteDef(routeDef)) {
+      return {
+        routeType: RouteType.STATIC_REDIRECTION,
+        definition: routeDef,
+      };
+    }
+    return {
+      routeType: RouteType.PROGRAMMATIC_REDIRECTION,
+      definition: routeDef,
+    };
   }
 
   private validateConfig({path}: RouteDef) {
@@ -158,6 +212,16 @@ class RouteConfigMatcher {
         this.redirectionFragments,
         combinedParams
       );
+      return {
+        result: true,
+        params: combinedParams,
+        pathParts: newPathParts,
+        isRedirection: true,
+      };
+    }
+
+    if (this.config.routeType === RouteType.PROGRAMMATIC_REDIRECTION) {
+      const newPathParts = this.config.definition.redirector(pathParts);
       return {
         result: true,
         params: combinedParams,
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/app_routing/route_config_types.ts
```diff
@@ -47,9 +47,39 @@ export interface ConcreteRouteDef {
 }
 
 export interface RedirectionRouteDef {
-  routeKind: null;
+  routeKind?: null;
```
stephanwlee, line 50: Do note that this is to be dropped in internal TensorBoard when no redirection specifies routeKind.

tensorboard/webapp/app_routing/route_config_types.ts
```diff
@@ -47,9 +47,39 @@ export interface ConcreteRouteDef {
...
+   * definition in `ConcreteRouteDef`, is only used as a wild card and its value
+   * is not used meaningfully as all raw `pathParts` are passed to the
+   * `redirector`.
+   */
+  path: string;
+  /**
+   * A function that returns, from a pathParts, a new pathParts it should
+   * redirect to.
+   */
+  redirector: (pathParts: string[]) => string[];
```
psybuzz, line 67: Since we don't have a dedicated type for `PathParts` and the concept is mainly used in route_config.ts which is not exposed outside of AppRouting, could we make the comment for this field explain what a `pathPart` is?

tensorboard/webapp/app_routing/route_config_test.ts
```diff
@@ -503,4 +502,103 @@ describe('route config', () => {
...
+
+      expect(config.match(buildNavigation({pathname: '/tb/redirect'}))).toEqual(
+        buildRouteMatch({
+          pathname: '/tensorboard',
+          params: {},
+        })
+      );
+    });
+
+    it('handles parameter is opaque string in path parts', () => {
```
psybuzz, line 565: nit: 'handles an opaque parameter in path parts'

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -70,21 +77,68 @@ function getPathFragments(path: string): PathFragment[] {
   });
 }
 
+enum RouteType {
```
psybuzz, line 80: Since this is not the 'type of a `Route`', in the sense of the `Route` type used throughout AppRouting, would it make sense to rename this to 'RouteDefType'?

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -249,20 +313,24 @@ export class RouteConfigs {
...
       }
     }
   }
 
   private validateRouteConfigs(configs: RouteDef[]) {
-    const defaultRoutes = configs.filter((routeDef) => {
-      return Boolean(routeDef.routeKind !== null && routeDef.defaultRoute);
-    });
+    const concreteDefinitions = configs.filter((routeDef) => {
+      return Boolean(isConcreteRouteDef(routeDef));
```
psybuzz, line 331: Can the `Boolean` be dropped, assuming TS knows `isConcreteRouteDef` returns a boolean?

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -249,20 +313,24 @@ export class RouteConfigs {
 
       this.configMatchers.push(matcher);
 
-      if (matcher.config.routeKind !== null) {
-        this.routeKindToConfigMatchers.set(matcher.config.routeKind, matcher);
+      if (matcher.config.routeType === RouteType.CONCRETE) {
+        this.routeKindToConfigMatchers.set(
```
psybuzz, line 317: `routekindtoconfigmatchers` only ever stores CONCRETE config matchers, and L385 checks at runtime whether a value from the map is type CONCRETE.  Now that this PR introduces a convenient `ConcreteRouteConfigMatcher`, can we drop part of the L385 check and type the map as `routekindtoconfigmatchers: Map<RouteKind, ConcreteRouteConfigMatcher>`?

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -249,20 +313,24 @@ export class RouteConfigs {
...
+          matcher.config.definition.routeKind,
+          matcher
+        );
 
-        if (matcher.config.defaultRoute) {
+        if (matcher.config.definition.defaultRoute) {
```
psybuzz, line 322: Now that we have `ConcreteRouteConfigMatcher`, would it make sense to drop L419, the part that checks at runtime whether `defaultRouteConfig` is a CONCRETE type, and instead do a check here `isConcreteRoute*`?  Then, we can more strictly type the field `defaultRouteConfig: ConcreteRouteConfigMatcher | null;`

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -166,6 +220,16 @@ class RouteConfigMatcher {
       };
     }
 
+    if (this.config.routeType === RouteType.PROGRAMMATIC_REDIRECTION) {
```
psybuzz, line 223: optional: now that we have 3 types of route configs, L210 might read better as `if (routeDefType === REDIRECTION && this.redirectionFragments)`, in case readers don't realize that `redirectionFragments` are specific to only the REDIRECTION route def type.

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -317,12 +381,19 @@ export class RouteConfigs {
             break;
           }
 
-          const config = matcher.config as ConcreteRouteDef;
+          if (matcher.config.routeType !== RouteType.CONCRETE) {
+            throw new RangeError(
+              'None concrete route definition cannot match for concrete ' +
```
psybuzz, line 386: nit: "No concrete route definition can match..."

psybuzz (commented):
Exciting!

tensorboard/webapp/app_routing/route_config_test.ts
stephanwlee: Wow, that was a crappy wording. Thanks for catching that and suggesting a fix.

tensorboard/webapp/app_routing/route_config.ts
stephanwlee: Definitely refactored and improved.

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -198,19 +212,56 @@ class RouteConfigMatcher {
...
+}
 
-  /**
-   * In case of parameter projection failure (cannot create pathname), it throws
-   * a RangeError.
-   */
-  matchByParams(params: RouteParams): PositiveMatch {
-    const pathParts = this.reprojectPathByParams(this.pathFragments, params);
+class ConcreteRouteConfigMatcher extends RouteConfigMatcher {
+  constructor(readonly definition: ConcreteRouteDef) {
```
psybuzz, line 218: Can we drop the entire constructor?

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -198,19 +212,56 @@ class RouteConfigMatcher {
...
+    super(definition);
+  }
+}
+
+class StaticRedirectionRouteConfigMatcher extends RouteConfigMatcher {
+  private readonly redirectionFragments: PathFragment[];
+
+  constructor(readonly definition: RedirectionRouteDef) {
```
psybuzz, line 226: Can we drop the `readonly`?  The `definition` does not need to be stored as a property

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -70,19 +77,25 @@ function getPathFragments(path: string): PathFragment[] {
...
 }
 
-class RouteConfigMatcher {
-  private readonly pathFragments: PathFragment[];
-  private readonly pathMatchers: PathMatcher[];
-  private readonly redirectionFragments: PathFragment[] | null;
+abstract class RouteConfigMatcher {
+  protected readonly pathFragments: PathFragment[];
+  protected readonly pathMatchers: PathMatcher[];
+  abstract readonly definition: RouteDef;
```
psybuzz, line 83: Can we drop the `definition` member, since it isn't used by the abstract base class at all, and only 2 of the 3 subclasses?

psybuzz (approved):
Nice

tensorboard/webapp/app_routing/route_config.ts
stephanwlee: Notice that this constructor _requires_ `ConcreteRouteDef` as opposed to any `RouteDef`. I don't think I can drop it.

tensorboard/webapp/app_routing/route_config.ts
stephanwlee: It is read in the downstream `RouteConfigMatchers`.

stephanwlee: While this specific `definition is not read below, having some congruent interface in all `RouteConfigMatcher` definitely is something I was going for so I would like to keep it as is.

tensorboard/webapp/app_routing/route_config.ts
stephanwlee: Ditto.

PR #5145, opened 2021-07-21
router: redirect with query parameter

stephanwlee:
This change adds advanced feature to the router where a redirector can
redirect to a path with query parameter.

Imagine following use case. We want to add a nice redirection route
where, depending on path part value, we can redirect to EXPERIMENT or
EXPERIMENTS and, in the latter case, from the path part value, we'd like
to also pre-populate the user token. This would require
DeepLinkProvider-like module that would also reason and pre-populate
state from a path part (as opposed to value in the query parameter) and
that's farther in the realm of possible designs than what is proposed
here.

Here, we are proposing that the redirector can create almost
`Navigation` like payload where it can specify both pathname and query
parameter to a next route. Do note that this feature is clamped down to
only programmatical navigation and we have no plans to expand scope of
other route definitions.

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -201,8 +201,19 @@ export class AppRoutingEffects {
     const dispatchNavigating$ = this.validatedRoute$.pipe(
       tap(({routeMatch, options}) => {
         if (options.browserInitiated && routeMatch.deepLinkProvider) {
+          // Query paramter formed by the redirector is passed to the
+          // deserializer instead of one from Location.getSearch(). This
+          // behavior emulates redirected URL to be on the URL bar such as
+          // "/compare?foo=bar" based on information provided by redirector (do
+          // note that location.getSearch() will return current query parameter
+          // which is pre-redirection URL).
+          const queryParams =
+            routeMatch.originateFromRedirection &&
+            routeMatch.redirectionOnlyQueryParams
+              ? routeMatch.redirectionOnlyQueryParams
+              : this.location.getSearch();
           const rehydratingState = routeMatch.deepLinkProvider.deserializeQueryParams(
-            this.location.getSearch()
+            queryParams
           );
           this.store.dispatch(
             stateRehydratedFromUrl({
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
@@ -89,6 +89,23 @@ describe('app_routing_effects', () => {
             deserializeQueryParams: deserializeQueryParamsSpy,
           },
         },
+        {
+          path: '/redirect_no_query/:wildcard',
+          redirector: (pathParts: string[]) => {
+            return {
+              pathParts: pathParts.slice(1),
+            };
+          },
+        },
+        {
+          path: '/redirect_query/:wildcard',
+          redirector: (pathParts: string[]) => {
+            return {
+              pathParts: pathParts.slice(1),
+              queryParams: [{key: 'hard', value: 'coded'}],
+            };
+          },
+        },
       ];
     }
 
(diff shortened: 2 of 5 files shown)


tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -517,6 +534,103 @@ describe('app_routing_effects', () => {
...
+          // deserializer instead of one from Location.getSearch(). This
+          // behavior emulates redirected URL to be on the URL bar. That is,
+          // when passing through the redirection flow, the actual URL on the
+          // location bar is pre-redirection, "/redirect_query/compare". If
+          // redirector wants to "set" href to "/compare?foo=bar", the query
+          // parameter has to come from the redirector, not from the URL bar.
+          deserializeQueryParamsSpy
+            .withArgs([{key: 'hard', value: 'coded'}])
+            .and.returnValue({good: 'value'})
+            .and.throwError('Invalid');
```
psybuzz, line 581: I'm not familiar with what happens when a spy has both `.and.returnValue` and `.and.throwError`, does this return a value on the first call, and throw an error on subsequent calls?

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -517,6 +534,103 @@ describe('app_routing_effects', () => {
...
+                navigationOptions: {
+                  replaceState: true,
+                },
+              }),
+            }),
+            jasmine.any(Object),
+          ]);
+        }));
+
+        it('redirects with query parameter to no deepLinkProvider', fakeAsync(() => {
```
psybuzz, line 608: optional: consider adding a comment on L81 noting that some test cases depend on the /experiments route's absence of a deepLinkProvider, or alternatively configure another route named something like /no-deeplink-provider.

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -517,6 +534,103 @@ describe('app_routing_effects', () => {
...
+            actions.stateRehydratedFromUrl({
+              routeKind: RouteKind.COMPARE_EXPERIMENT,
+              partialState: {good: 'value'},
+            }),
+            actions.navigating({
+              after: buildRoute({
+                routeKind: RouteKind.COMPARE_EXPERIMENT,
+                params: {},
+                pathname: '/compare',
+                queryParams: [],
```
psybuzz, line 598: I was initially surprised that the `queryParams` field in the route is empty, when the initial redirector provides queryParams.  Just checking, is this empty because of L582 `serializeStateToQueryParamsSpy.and.returnValue(of([]));`?  If so, would it make sense to add a case where the serializer produces queryParams as well?

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -201,8 +201,19 @@ export class AppRoutingEffects {
     const dispatchNavigating$ = this.validatedRoute$.pipe(
       tap(({routeMatch, options}) => {
         if (options.browserInitiated && routeMatch.deepLinkProvider) {
+          // Query paramter formed by the redirector is passed to the
+          // deserializer instead of one from Location.getSearch(). This
+          // behavior emulates redirected URL to be on the URL bar such as
+          // "/compare?foo=bar" based on information provided by redirector (do
+          // note that location.getSearch() will return current query parameter
+          // which is pre-redirection URL).
```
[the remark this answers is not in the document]
stephanwlee: > original query params?

Not sure what a original query parameter would be but effect would replace query parameter with:
- if deeplink provider exists, whatever DeepLinkProvider returns
- else, drop everything

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
stephanwlee: PTAL at https://stackblitz.com/edit/jasmine-testing-v98jbt?file=src%2Fidentity.spec.ts.

In words, if the method is called with argument that is not specified `withArgs(X).and.returnValue()`, it will throw.

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
stephanwlee: There already are tests for deeplink provider scheme under "deeplinks writes" spec sections.

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -375,19 +398,37 @@ export class RouteConfigs {
...
+          return {
+            ...base,
+            originateFromRedirection: true,
+            redirectionOnlyQueryParams,
+          };
         }
       }
+
+      if (!hasMatch) {
+        redirectionOnlyQueryParams = undefined;
```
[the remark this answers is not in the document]
stephanwlee: Removed. Explicitly added a test case that I thought I was guarding against. 

When you go through, RedirectWithQ -> Redirect -> Concrete, I did not want the query parameter from the first route to be propagated to Concrete but that definitely is not the case without this statement.

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -375,19 +398,37 @@ export class RouteConfigs {
...
+        break;
+      }
+
       if (wasRedirected) {
         redirectionCount++;
       }
-      if (redirectionCount > this.maxRedirection || !wasRedirected) {
+      if (redirectionCount > this.maxRedirection) {
```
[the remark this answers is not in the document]
stephanwlee: Ack. Cannot think of an easy way to make this much readable ATM.

psybuzz (approved):

PR #5146, opened 2021-07-21
notification center: persist setting via common storage

stephanwlee:
Now with the persistent settings storage feature, we do not need to
manually interact with the LocalStorage. This change adopts the new
infrastructure instead.

Have tested this end-to-end with the internal application: ~cl/385883667~ cl/386121119


tensorboard/webapp/notification_center/_notification_center_module.ts
```diff
@@ -14,20 +14,40 @@ limitations under the License.
...
+}
+
 @NgModule({
   imports: [
     StoreModule.forFeature(NOTIFICATION_FEATURE_KEY, reducers),
     EffectsModule.forFeature([NotificationCenterEffects]),
     NotificationCenterDataSourceModule,
     NotificationCenterViewModule,
+    PersistentSettingsConfigModule.defineGlobalSetting<
+      State,
```
japie1235813, line 48: internal cl build fails, please consider fixing it before merging

japie1235813 (approved):

tensorboard/webapp/notification_center/_notification_center_module.ts
stephanwlee: The build was stale and it passes with the latest change. Thanks.

PR #5148, opened 2021-07-21
color grouping: grouping result preview

japie1235813:
Preview the color grouping result before users save them.

  - generate preview grouping results in `regex_edit_dialog` container, the colors might be different from what run table shows (same as state)
  - construct `colorRunsMap` (type `[string, Run[]][]`) to pass down color-runs mapping information 
for example
```
[
["#425066", [run1, run2, ...] ],
["#12b5cb", [run3, run4, ...] ],
...
]
```
The colors might be repeated (because we only have 7 colors in the palette) but the preview will not "merge" them (they are different items of  `colorRunsMap` array

* Screenshots of UI changes

https://user-images.githubusercontent.com/1131010/126570206-9d103550-0574-4656-bb41-e23dd080fdab.mov

updated ui (with truncated number of runs per group)
Screen Shot 2021-07-22 at 3 07 58 PM

Note: UI details (margin/color... etc) could be adjusted later. Please focus on functionality on this pr.


tensorboard/webapp/runs/views/runs_table/regex_edit_dialog.ng.html
```diff
@@ -41,6 +42,19 @@ <h1 mat-dialog-title>Color runs by regex</h1>
...
 
+<div *ngIf="colorRunsMap.length" class="group-container">
+  <div>Grouping result preview:</div>
+  <div class="grouping-preview">
+    <div
+      class="group"
+      *ngFor="let colorRunsPair of colorRunsMap"
+      style="background: {{colorRunsPair[0]}}"
+    >
+      <span *ngFor="let run of colorRunsPair[1]"> {{ run.name }}<br /> </span>
```
stephanwlee, line 53: consider using ul/li for list of things. Also consider truncating the result (and say "and N more runs matches") as seeing 100s of result is not that helpful anyways

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_container.ts
```diff
@@ -43,6 +56,49 @@ export class RegexEditDialogContainer {
...
+    const groupKeyToColorString = new Map<string, string>();
+    this.colorRunsMap = [];
+    Object.entries(groups.matches).forEach(([groupId, runs]) => {
+      const color =
+        groupKeyToColorString.get(groupId) ??
+        CHART_COLOR_PALLETE[
+          groupKeyToColorString.size % CHART_COLOR_PALLETE.length
+        ];
+      groupKeyToColorString.set(groupId, color);
+      this.colorRunsMap.push([color, runs as Run[]]);
```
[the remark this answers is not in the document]
japie1235813: thanks for the sample pr!

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
```diff
@@ -254,4 +264,74 @@ describe('regex_edit_dialog', () => {
     const input = fixture.debugElement.query(By.css('input'));
     expect(input.nativeElement.value).toBe('(train|eval)');
   });
+
+  describe('live grouping result preview', () => {
```
stephanwlee, line 268: I actually do not trust the code I wrote. Would you mind adding a small spec that tests whether the input is populated with regex string from the store initially? If you could add another spec that tests for its changes would be nice: like if store dispatches change and if the regex string in the store has changed, do we want to update the input? Probably not.

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
```diff
@@ -254,4 +264,74 @@ describe('regex_edit_dialog', () => {
...
+      const fixture = createComponent(['rose']);
+      fixture.detectChanges();
+
+      const groupingResult = fixture.debugElement.query(
+        By.css('.group-container')
+      );
+      expect(groupingResult).toBeNull();
+    });
+
+    it('renders grouping preview on regex query input', fakeAsync(() => {
```
stephanwlee, line 279: Would you mind adding a small test that tests what we do with an invalid regex string? Examples of "invalid" regex string are:
- empty string
- invalid regex
- no match

stephanwlee (approved):
Thank you for incorporating my suggestions!

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
japie1235813: > whether the input is populated with regex string from the store initially
added (not yet push)
> if the regex string in the store has changed
a couple of questions:
- why this might happen?
- at this timing (regex string in the store had changed) do we already have tentativeString?
`merge` takes the last value so if the store changed value after tentativeString changes, `groupByRegexString$` would be the store value (`this.store.select(getColorGroupRegexString)` then)?
my own interpretation of `groupByRegexString$`: get the value from store as the initial value, then always be `tentativeString$` afterward.

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
japie1235813: I'm not sure about the second item "invalid regex" do you mean if the input is just a string but not regex string?
for example `train` vs `train_(\d+)`
the first case I think it's okay to still render a group (and only one group)
if not, please provide some examples to make my test cases clear. Thanks!

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
stephanwlee: > why this might happen?

it does not happen but as a view component, you do not want to make too much assumptions but, instead, just follow the contract between a store and the view (i.e., store can emit change for whatever reason it wants). 

> my own interpretation of groupByRegexString$: get the value from store as the initial value, then always be tentativeString$ afterward.

Yes, that was my intention when I wrote that code. However, I don't want to trust brain so I would want a test that test exactly that assumption :)

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
stephanwlee: `train_(\d+` would be a broken regex. `\` would be another one.

PR #5150, opened 2021-07-23
Expose CompareExperimentStr parser as util

stephanwlee:
Ideally, we want to populate store with an information that exists in
the comparison string. However, router only provides the information in
an opaque `experimentIds` (typed as string) without an ability to reason
with the structure inside.

Ideally, RouteParameter should have a way to construct a structured data
structure and pass it through `navigated` action but that makes
AppRoutingEffects even more complex while making the RouteConfig more
elaborate.

For now, we are exposing the utility for stores to use instead.

### DIFF
--- a/tensorboard/webapp/app_routing/store_only_utils.ts
+++ b/tensorboard/webapp/app_routing/store_only_utils.ts
@@ -21,7 +21,8 @@ import {CompareRouteParams} from './types';
 /**
  * Returns experimentId to alias information encoded in CompareRouteParams.
  *
- * Do not use this outside of _redux store.
+ * This utility is used by only limited packages. Please refer to visiblity in
+ * BUILD.
  */
 export function getCompareExperimentIdAliasSpec(
   routeParams: CompareRouteParams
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/app_routing/store_only_utils.ts
```diff
@@ -0,0 +1,35 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {parseCompareExperimentStr} from './internal_utils';
+import {CompareRouteParams} from './types';
+
+// Ideally, we would not be exposing this utility method this, but there is no
+// way to share the structured information to other stores without changing
```
psybuzz, line 19: I'm under the assumption that we don't want any "other store" to use this fn to parse route params, only the Experiment reducer, specifically?

If so, could we make this comment explicit about that, something like "we want the Experiment reducer to be able to listen to `navigated` actions and extract the alias from the RouteParams in the action payload, so that the experiment store can save aliases in its state".

If the intent is actually to let any other store use this, then disregard this comment.

psybuzz (approved):

tensorboard/webapp/app_routing/store_only_utils.ts
stephanwlee: Left some comment and decided to use visibility to actually enforce this instead.

PR #5151, opened 2021-07-23
color grouping: on clicking `Regex`, open dialog if regex string is empty

japie1235813:
Currently we have four items on the dropdown menu. The forth one is editing the regex string. However while there is no regexString and user clicks `Regex` item, we fire `groupByChanges` with empty Regex string. Now we want to prop user to input the string by opening up the dialog.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
@@ -721,8 +721,6 @@ describe('runs_table', () => {
         expect(dispatchSpy).toHaveBeenCalledWith(
           runGroupByChanged({
             experimentIds: ['book'],
-            // TODO(japie1235813): regexString is hardcoded to '' for now;
-            // should be fixed when regex support is properly implemented.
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -711,6 +701,53 @@ describe('runs_table', () => {
...
+            experimentIds: ['book'],
+            groupBy: {key: GroupByKey.REGEX, regexString: 'run'},
+          })
+        );
+      });
+
+      it('opens edit dialog on clicking `Regex` when the regex string is not set', () => {
+        store.overrideSelector(getEnabledColorGroup, true);
+        store.overrideSelector(getEnabledColorGroupByRegex, true);
+        store.overrideSelector(getRunGroupBy, {key: GroupByKey.EXPERIMENT});
```
stephanwlee, line 732: Please add `store.overrideSelector(getColorGroupRegexString, '');` explicitly instead of relying on `beforeEach` or previous spec.

stephanwlee (approved):

PR #5155, opened 2021-07-27
histogram: actually enable SummaryV2OpGraphTest

nfelt:
The `write_histogram_event` method in the `SummaryV2OpGraphTest` test case is defined in order to override functionality in the parent class (`SummaryV2OpTest`, the non-graph-mode version), but the parent class never actually called that method, so the child class was essentially just a repeat of the parent test case rather than actually testing graph mode behavior.

This fixes that oversight. I spot-checked for other instances of this pattern and they look correctly structured:
https://cs.opensource.google/search?ss=tensorflow%2Ftensorboard&q=file:summary_test%5C.py%20write_%5Ba-z%5D%2B_event

### DIFF
--- a/tensorboard/plugins/histogram/summary_test.py
+++ b/tensorboard/plugins/histogram/summary_test.py
@@ -176,11 +176,7 @@ class SummaryV2OpTest(SummaryBaseTest, tf.test.TestCase):
         return self.histogram_event(*args, **kwargs).summary
 
     def histogram_event(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


nfelt:
Actually I just accidentally merged this commit as part of #5156 already, so I'll close this PR.

japie1235813:
Sorry for overlooking this.

nfelt:
No problem at all!  I just accidentally merged it already soooo 🤷

PR #5159, opened 2021-07-27
histogram: axis for Angular histogram widget

stephanwlee:
This change is a start of the rewrite for the histogram widget.
Currently, the component was written as a shallow wrapper around our
Polymer based vz-histogram-timeseries component. This wrapper was not
easy enough to make changes to address some of the user feedbacks and
implement a new feature.

To that end, we are rewriting the component so bulk of the logic resides
in the Angular land without Polymer. Do note that goal is _not_ to
remove Polymer but is to have more manageable visualization code with
much better types. Even the Polymer component is a thin wrapper around
d3 and the most interesting logic resides in the `draw` method. To be
clear, we are not completely gutting out d3 as it still serves useful
purposes for axis, formatting,

This change is the first of more to follow. This change only calculates
the axis and renders the same axis as existing histogram visualization.
No component actually make use of it and you will need to manually
modify `histogram_card_component.ng.html` to make use of the new
version.

### DIFF
--- a/tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
+++ b/tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
@@ -1,0 +1,193 @@
+/* Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {
+  ChangeDetectionStrategy,
+  Component,
+  ElementRef,
+  Input,
+  OnChanges,
+  ViewChild,
+} from '@angular/core';
+
+import * as d3 from '../../third_party/d3';
+import {
+  ColorScale,
+  HistogramData,
+  HistogramMode,
+  TimeProperty,
+} from './histogram_types';
+
+@Component({
+  selector: 'tb-histogram-v2',
+  templateUrl: 'histogram_v2_component.ng.html',
+  styleUrls: ['histogram_v2_component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class HistogramV2Component implements OnChanges {
+  @ViewChild('xAxis') private readonly xAxis!: ElementRef;
+  @ViewChild('yAxis') private readonly yAxis!: ElementRef;
+  @ViewChild('content') private readonly content!: ElementRef;
+
+  @Input() mode: HistogramMode = HistogramMode.OFFSET;
+
+  @Input() timeProperty: TimeProperty = TimeProperty.STEP;
+
+  @Input() colorScale?: ColorScale;
+
+  @Input() name!: string;
+
+  @Input() data!: HistogramData;
+
+  private yScale?:
+    | d3.ScaleTime<number, number>
+    | d3.ScaleLinear<number, number>;
+
+  constructor(private readonly host: ElementRef) {}
+
+  ngOnChanges() {
+    if (this.content) {
+      this.ngAfterViewInit();
+    }
+  }
+
+  ngAfterViewInit() {
+    this.updateChart();
+  }
+
+  private updateChart() {
+    this.renderXAxis(this.data);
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -0,0 +1,193 @@
...
+      domain = [0, countMax];
+    } else {
+      const yValues = data.map((datum) => {
+        switch (this.timeProperty) {
+          case TimeProperty.WALL_TIME:
+            return datum.wallTime;
+          case TimeProperty.STEP:
+            return datum.step;
+          case TimeProperty.RELATIVE:
+            return datum.wallTime - data[0]?.wallTime ?? 0;
```
psybuzz, line 113: Can the '?' and '?? 0's be dropped, since data[0] should always be truthy?

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -0,0 +1,193 @@
...
+      switch (this.timeProperty) {
+        case TimeProperty.WALL_TIME:
+          anyYAxis.tickFormat(d3.timeFormat('%m/%d %X'));
+          break;
+        case TimeProperty.STEP: {
+          anyYAxis.tickFormat(d3.format('.0f'));
+          break;
+        }
+        case TimeProperty.RELATIVE: {
+          anyYAxis.tickFormat((d: number): string => {
```
psybuzz, line 147: nit: consider renaming `d` to something like `timeInMs`?

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -0,0 +1,193 @@
...
+        }
+      });
+      const {min: yMin, max: yMax} = getMinMax(yValues, (val) => val);
+      domain = [yMin, yMax];
+    }
+    this.yScale.domain(domain);
+  }
+
+  private renderYAxis() {
+    const yScale = this.yScale!;
```
psybuzz, line 123: I'm not sure whether followups need the `yScale` to be a property, but would it make more sense to make `yScale` a parameter to `renderYAxis` instead of asserting truthiness here?

tensorboard/webapp/widgets/histogram/histogram_v2_test.ts
```diff
@@ -0,0 +1,249 @@
...
+  };
+}
+
+function buildHistogramDatum(
+  override: Partial<HistogramDatum> = {}
+): HistogramDatum {
+  return {
+    wallTime: 1000,
+    step: 0,
+    bins: new Array(10).fill(buildBin()),
```
psybuzz, line 45: Presumably this function should return bins whose elements are not exactly the same object reference?
e.g. something like
`new Array(10).fill(0).map(x => buildBin())`

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -0,0 +1,193 @@
...
+  @Input() data!: HistogramData;
+
+  private yScale?:
+    | d3.ScaleTime<number, number>
+    | d3.ScaleLinear<number, number>;
+
+  constructor(private readonly host: ElementRef) {}
+
+  ngOnChanges() {
+    if (this.content) {
```
psybuzz, line 60: Should we be skipping updateChart() when only the colorScale/name changes?

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -0,0 +1,193 @@
...
+  @ViewChild('yAxis') private readonly yAxis!: ElementRef;
+  @ViewChild('content') private readonly content!: ElementRef;
+
+  @Input() mode: HistogramMode = HistogramMode.OFFSET;
+
+  @Input() timeProperty: TimeProperty = TimeProperty.STEP;
+
+  @Input() colorScale?: ColorScale;
+
+  @Input() name!: string;
```
psybuzz, line 49: The "v1" histogram component mentions we would not need `name` as an input if we replaced `colorScale` with `color`.  If that still makes sense, should we either follow that, or copy over the same comment?

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
stephanwlee: That'd be an optimization to be done later.

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
stephanwlee: Will address that in a follow up that PR is about ready.

psybuzz (approved):

PR #5161, opened 2021-07-28
color grouping: rehydrated regexString from url

japie1235813:
Currently when users click buttons to save regexString, we preserve this information in the url. Yet when the browser refreshed afterward, we did not rehydrate the string back to our state and the information disappears. This pr aims to address that.

### DIFF
--- a/tensorboard/webapp/runs/store/runs_reducers_test.ts
+++ b/tensorboard/webapp/runs/store/runs_reducers_test.ts
@@ -1060,6 +1060,57 @@ describe('runs_reducers', () => {
         ])
       );
     });
+
+    it('sets regexString on groupBy REGEX', () => {
+      const state = buildRunsState({
+        initialGroupBy: {key: GroupByKey.EXPERIMENT},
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/runs/store/runs_reducers_test.ts
```diff
@@ -1060,6 +1060,57 @@ describe('runs_reducers', () => {
...
+          ['run3', '#ccc'],
+          ['run4', '#ddd'],
+        ]),
+        defaultRunColorForGroupBy: new Map([
+          ['run1', '#aaa'],
+          ['run2', '#bbb'],
+          ['run3', '#ccc'],
+          ['run4', '#ddd'],
+        ]),
+        runColorOverrideForGroupBy: new Map([['run1', '#ccc']]),
```
stephanwlee, line 1096: bulk of these are unrelated to your assertion. To keep your test spec focused, please drop it.

stephanwlee (approved):

PR #5162, opened 2021-07-28
histogram: render Overlay and Offset charts

stephanwlee:
This change builds on top of the Angular version of the histograms
component where we had axex rendering implemented. This change finally
draws the histogram like the Polymer counterpart. Most of the UX
considerations are mostly replica of that of Polymer part and I did not
intend to have much changes besides minor color changes (now use theme
variable instead of hardcoded hex value).

This change does not include:
- interaction like tooltips
- redraw on resize

image
image

### DIFF
--- a/tensorboard/webapp/widgets/histogram/histogram_v2_component.ng.html
+++ b/tensorboard/webapp/widgets/histogram/histogram_v2_component.ng.html
@@ -41,7 +41,6 @@ limitations under the License.
         <path
           [attr.d]="getHistogramPath(datum)"
           [style.color]="getHistogramFill(datum)"
-          stroke=""
         ></path>
       </g>
     </g>
--- a/tensorboard/webapp/widgets/histogram/histogram_v2_component.ng.html
+++ b/tensorboard/webapp/widgets/histogram/histogram_v2_component.ng.html
@@ -14,8 +14,36 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-<div class="main">
+<div [class]="'main ' + mode + ' ' + timeProperty">
   <svg #xAxis class="axis x-axis"></svg>
   <svg #yAxis class="axis y-axis"></svg>
-  <svg #content class="content"></svg>
+  <svg #content class="content">
+    <g class="grid">
+      <g
+        *ngFor="let tick of getGridTickYLocs()"
+        [attr.transform]="'translate(0, ' + tick + ')'"
+      >
+        <line class="tick" x2="100%"></line>
+      </g>
+    </g>
+
+    <g class="histograms">
+      <g
+        *ngFor="let datum of data; trackBy: trackByWallTime;"
+        [attr.transform]="getGroupTransform(datum)"
+        class="histogram"
+      >
+        <line
+          *ngIf="mode === HistogramMode.OFFSET"
+          class="baseline"
+          x2="100%"
+        ></line>
+        <path
+          [attr.d]="getHistogramPath(datum)"
+          [style.color]="getHistogramFill(datum)"
+          stroke=""
+        ></path>
+      </g>
+    </g>
+  </svg>
 </div>
--- a/tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
+++ b/tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
@@ -80,8 +80,8 @@ export class HistogramV2Component implements OnChanges {
   readonly HistogramMode = HistogramMode;
 
   private layout: Layout = {
-    histogramHeight: 1,
-    contentClientRect: {height: 1, width: 1},
+    histogramHeight: 0,
+    contentClientRect: {height: 0, width: 0},
   };
   private scales: Scales;
   private domInitialized = false;
@@ -90,11 +90,11 @@ export class HistogramV2Component implements OnChanges {
     // `data` may not be available at the constructor time. Since we recalculate
     // the scales on the `ngAfterViewInit`, let's just initialize `scales` with
     // their default values.
-    this.scales = this.updateScales(this.layout, []);
+    this.scales = this.computeScales([]);
   }
 
   ngOnChanges() {
-    if (this.content) {
+    if (this.domInitialized) {
       this.updateChart();
     }
   }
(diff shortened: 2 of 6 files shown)


tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -55,83 +77,178 @@ export class HistogramV2Component implements OnChanges {
...
 
-  private yScale?:
-    | d3.ScaleTime<number, number>
-    | d3.ScaleLinear<number, number>;
+  readonly HistogramMode = HistogramMode;
 
-  constructor(private readonly host: ElementRef) {}
+  private layout: Layout = {
+    histogramHeight: 1,
+    contentClientRect: {height: 1, width: 1},
```
psybuzz, line 84: Just to confirm: is it important for these to be nonzero, or is `1` just an arbitrary default value?  I often use `0` as default myself, and I'm not sure from reading whether there might be something wrong about choosing `0` here.

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -55,83 +77,178 @@ export class HistogramV2Component implements OnChanges {
...
+
+  constructor(private readonly changeDetector: ChangeDetectorRef) {
+    // `data` may not be available at the constructor time. Since we recalculate
+    // the scales on the `ngAfterViewInit`, let's just initialize `scales` with
+    // their default values.
+    this.scales = this.updateScales(this.layout, []);
+  }
 
   ngOnChanges() {
     if (this.content) {
```
psybuzz, line 97: Can replace all `this.content` checks with `this.domInitialized`?

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -55,83 +77,178 @@ export class HistogramV2Component implements OnChanges {
...
+    this.changeDetector.detectChanges();
+  }
+
+  getHistogramPath(datum: HistogramDatum): string {
+    if (!this.domInitialized || !datum.bins.length) return '';
+    const xScale = this.scales.binScale;
+    const yScale = this.scales.countScale;
+
+    const firstBin = datum.bins[0];
+    const lastBin = datum.bins.slice(-1)[0];
```
psybuzz, line 115: Interesting, I've never seen this before myself, but it's quite concise!

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -55,83 +77,178 @@ export class HistogramV2Component implements OnChanges {
...
+  }
+
+  trackByWallTime(datum: HistogramDatum): number {
+    return datum.wallTime;
+  }
+
+  // translates container for histogram so we can have more sensible coordinate
+  // system for reasoning with the coordinate system of a histogram.
+  getGroupTransform(datum: HistogramDatum): string {
+    if (!this.domInitialized || this.mode === HistogramMode.OVERLAY) return '';
```
psybuzz, line 136: I'm not sure why some methods like `getGroupTransform`, `getHistogramPath` check `this.domInitialized` while others like `getHistogramFill`, `getGridTickYLocs` do not.  All 4 are used in the template.  How did we decide which ones require the `domInitialized` check?

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -55,83 +77,178 @@ export class HistogramV2Component implements OnChanges {
...
+      this.layout.contentClientRect = this.content.nativeElement.getBoundingClientRect();
+      this.layout.histogramHeight = this.layout.contentClientRect.height / 2.5;
+    }
   }
 
   private updateChart() {
-    this.renderXAxis(this.data);
-    this.updateYScaleAndDomain(this.data);
-    this.renderYAxis();
+    this.scales = this.updateScales(this.layout, this.data);
```
psybuzz, line 171: I'm inclined to suggest dropping the params to these 3 functions, since they all come from member properties that those functions already have access to.  This call to `this.updateScales()` suggests that the only class members this function requires looking at are the `layout` and `data`, however that is not the case as

- L177 directly reads `this.layout`, ignoring the param
- L204 reads `this.colorScale`, and L248 reads `this.mode`, both properties that weren't passed in

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -55,83 +77,178 @@ export class HistogramV2Component implements OnChanges {
...
+      const offsetAxisHeight =
+        this.mode === HistogramMode.OFFSET
+          ? height - layout.histogramHeight
+          : 0;
+      temporalScale.range([height - offsetAxisHeight, height]);
+      countScale.range([0, -layout.histogramHeight]);
     }
-    this.yScale.domain(domain);
+
+    return {binScale, d3ColorScale, countScale, temporalScale};
```
psybuzz, line 236: Since `updateScales` does not actually update the property, consider either renaming to `computeScales` or making the method actually update the scales.

tensorboard/webapp/widgets/histogram/histogram_v2_test.ts
```diff
@@ -242,7 +237,131 @@ describe('histogram v2 test', () => {
...
+      });
+
+      it('moves the group position depending on the timeProperty', () => {
+        const fixture = createComponent('foo', [
+          buildHistogramDatum({step: 0, wallTime: 100}),
+          buildHistogramDatum({step: 5, wallTime: -200}),
+          buildHistogramDatum({step: 10, wallTime: 400}),
+        ]);
+        fixture.componentInstance.mode = HistogramMode.OFFSET;
+        fixture.componentInstance.timeProperty = TimeProperty.STEP;
```
psybuzz, line 293: Is it intentional to set the timeProperty to STEP when L296 changes it again, without a test expectation in between?

psybuzz (approved):
Nice!  fyi I have not manually tested, so presuming that some eventual followup PR that actually enables this will have a reviewer look at it manually.

tensorboard/webapp/widgets/histogram/histogram_v2_test.ts
stephanwlee: Yup,k I wanted to assert whether it updates correctly. We already have test spec for asserting on initial render with `STEP` so I omitted that.

tensorboard/webapp/widgets/histogram/histogram_v2_component.ng.html
```diff
@@ -14,8 +14,36 @@
...
+      >
+        <line
+          *ngIf="mode === HistogramMode.OFFSET"
+          class="baseline"
+          x2="100%"
+        ></line>
+        <path
+          [attr.d]="getHistogramPath(datum)"
+          [style.color]="getHistogramFill(datum)"
+          stroke=""
```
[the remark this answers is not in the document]
stephanwlee: Nope, a simple mistake.

PR #5163, opened 2021-07-29
Update README.md to include no-data example

MalcolmSlaney:
Added an example of what good and bad event files might contain.

As a less than full-time user of TB, I see this problem a lot, and I would really love to see good and bad examples so I more easily know what I'm seeing.

Alternative suggestions welcome.

* Motivation for features / changes: It wasn't clear to this user what an empty event file looked like.  Let alone a good one.

* Technical description of changes:  Added the --inspect output from a good and a bad event file.

* Screenshots of UI changes: N/A

* Detailed steps to verify changes work correctly (as executed by you): Documentation change only.

* Alternate designs / implementations considered: Would love to see better examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,8 +254,9 @@ tensorboard in inspect mode to inspect the contents of your event files.
 
 `tensorboard --inspect --logdir DIRECTORY_PATH`
 
-An empty event file (from a Keras job that starts output at batch 2) might look like this:
-
+The output for an event file corresponding to a blank TensorBoard may still sometimes 
+show a few steps, representing a few initial events that aren't shown by TensorBoard 
+(for example, when using the Keras TensorBoard callback):
 ```
 tensor
    first_step           0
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -254,6 +254,30 @@ tensorboard in inspect mode to inspect the contents of your event files.
 
 `tensorboard --inspect --logdir DIRECTORY_PATH`
 
+An empty event file (from a Keras job that starts output at batch 2) might look like this:
```
nfelt, line 257: It might be worth a little clarification here that this event file isn't exactly "empty" (in that it does contain some events, as indicated by the non-zero steps) but that it doesn't show any summary data.

E.g. perhaps word as something like:

> The output for an event file corresponding to a blank TensorBoard may still sometimes show a few steps, representing a few initial events that aren't shown by TensorBoard (for example, when using the Keras TensorBoard callback):
>
> [output]
>
> In contrast, the output for an event file with more data might look like this:
>
> [output]

nfelt (commented):
Thanks for the contribution!  One clarification suggested

README.md
MalcolmSlaney: Thanks.  Much better now.

nfelt (approved):

PR #5168, opened 2021-07-30
sync: fix tslint 

japie1235813:
cl/387191760


stephanwlee (commented):
It is NOT a readonly property so this change is incorrect. Sorry.

stephanwlee (approved):

stephanwlee:
I had mistaken. We are manually modifying a property in the object when we resize the browser. Sorry for the confusion.

PR #5170, opened 2021-07-30
router: expose getExperimentIdsFromRouteParams and getRouteId

yatbear:
Expose `getExperimentIdsFromRouteParams` and `getRouteId` as external utils, `experimentIds` and `routeId` will be able to be retrieved from navigation.

### DIFF
--- a/tensorboard/webapp/app_routing/store_only_utils.ts
+++ b/tensorboard/webapp/app_routing/store_only_utils.ts
@@ -18,8 +18,7 @@ import {
   parseCompareExperimentStr,
   serializeCompareExperimentParams,
 } from './internal_utils';
-import {CompareRouteParams, RouteKind, RouteParams} from './types';
-import {ProgrammaticalNavigation} from './programmatical_navigation_types';
+import {CompareRouteParams, Route} from './types';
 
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/app_routing/store_only_utils.ts
```diff
@@ -34,3 +40,45 @@ export function getCompareExperimentIdAliasSpec(
...
+}
+
+/**
+ * Returns routeId from navigation.
+ *
+ * This utility is used by only limited packages. Please refer to visiblity in
+ * BUILD.
+ */
+export function getRouteIdFromNavigation(
+  navigation: ProgrammaticalNavigation
```
stephanwlee, line 78: Instead of `ProgrammaticalNavigation`, you want to use `Route`. Action payload for https://github.com/tensorflow/tensorboard/blob/bc216f499ec934113fdfd0f92a055c448b6daa8e/tensorboard/webapp/app_routing/actions/app_routing_actions.ts#L56-L57.

yatbear: Done, thanks!

stephanwlee (approved):

PR #5171, opened 2021-07-30
feature flag: add "enableLinkTime"

stephanwlee:
We are now working on a new feature that we call "Linked time
visualization". This work is to be done under a feature flag until it is
ready.


tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -21,6 +21,7 @@ import {
   EXPERIMENTAL_PLUGIN_QUERY_PARAM_KEY,
   SCALARS_BATCH_SIZE_PARAM_KEY,
   TBFeatureFlagDataSource,
+  ENABLE_LINK_TIME_PARAM_KEY,
```
japie1235813, line 24: nit: move this up to make it sort alphabetically

japie1235813 (approved):

PR #5172, opened 2021-07-31
widget: intersection observer directive

stephanwlee:
TensorBoard had use IntersectionObserver so optimize for rendering.
Since we have done that manually several times, we are now widgetizing
it so it is easier to create an element that would use the relatively
new browser API.


psybuzz (approved):
looks nice!

PR #5174, opened 2021-08-02
histogram: prevent render while invisible.

stephanwlee:
This change integrates Intersection Observer to the new Angular based
histogram so we prevent needless updates while invisible. Do note that
this solution is not perfect because any prop changes will be reflected
onto the DOM created via template; could not yet find an Angular
equivalent to `shouldComponentUpdate` in React.


tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -139,8 +140,8 @@ export class HistogramV2Component
   }
 
   ngOnChanges() {
-    if (this.domInitialized) {
-      this.updateChart();
+    if (this.domVisible) {
```
psybuzz, line 143: Let's drop the domVisible check here, now that `updateChartIfVisible` is clearly doing the check internally.

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -130,6 +130,7 @@ export class HistogramV2Component
     },
   };
   private domInitialized = false;
+  private domVisible = false;
 
   constructor(private readonly changeDetector: ChangeDetectorRef) {
     // `data` may not be available at the constructor time. Since we recalculate
```
psybuzz, line 136: Now that we no longer have `ngAfterViewInit`, let's update this comment that refers to it.

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -239,8 +231,25 @@ export class HistogramV2Component
...
+  }
+
+  onVisibilityChange({visible}: {visible: boolean}) {
+    this.domVisible = visible;
+    if (!visible) return;
+
+    if (!this.domInitialized) {
+      this.domInitialized = true;
+      this.changeDetector.detectChanges();
+      fromEvent<MouseEvent>(this.main.nativeElement, 'mousemove', {
```
psybuzz, line 244: By moving the `fromEvent` into `onVisibilityChange`, would this be triggering a new `addEventListener` every time a histogram is scrolled into view?  I think the `ngAfterViewInit` might be a preferred place for this, if we don't want unbounded # of listeners.

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -239,8 +231,25 @@ export class HistogramV2Component
...
+    this.updateChartIfVisible();
+  }
+
+  onVisibilityChange({visible}: {visible: boolean}) {
+    this.domVisible = visible;
+    if (!visible) return;
+
+    if (!this.domInitialized) {
+      this.domInitialized = true;
+      this.changeDetector.detectChanges();
```
psybuzz, line 243: Is the ordering of these lines intentionally set to put detectChanges before calling `updateClientRects` below?  If so, could we add a comment?

tensorboard/webapp/widgets/histogram/histogram_v2_test.ts
```diff
@@ -155,14 +175,77 @@ describe('histogram v2 test', () => {
...
+      fixture.detectChanges();
+
+      expect(
+        getAxisLabelText(fixture.debugElement.query(byCss.X_AXIS))
+      ).toEqual(['0', '50', '100']);
+    });
   });
 
   describe('y axis render', () => {
+    it('does not render until component becomes visible', () => {
```
psybuzz, line 237: This test case is named similarly to the one on L143 for the x axis.  However, the x axis one asserts the axis label text is empty before changing visibility to false, while this case does not.  Should they be consistent?

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
```diff
@@ -194,7 +184,7 @@ export class HistogramV2Component
     // Unlike other methods used in Angular template, if we return non-empty
     // value before the DOM and everything is initialized, this method can emit
     // junk (path with NaN) values causing browser to noisily print warnings.
-    if (!this.domInitialized || !datum.bins.length) return '';
+    if (!this.scales.initialized || !datum.bins.length) return '';
```
psybuzz, line 187: optional: currently, `computeScales` is called both when domVisible=true and false (from the constructor only).  Would things be simpler if we just had `scales: null` by default, and have `if (!this.scales)` checks here?

psybuzz (approved):

tensorboard/webapp/widgets/histogram/histogram_v2_component.ts
stephanwlee: To prevent constant subscription, I am doing this once using `domInitialized`.

psybuzz: d'oh that's pretty clear upon re-read, sorry

PR #5176, opened 2021-08-02
linked time: add basic data model

stephanwlee:
This change adds basic reducers for setting and changing linked time
value. The states and actions in this changeset is highly experimental
and is prone for changes in the future when we implement UI features
along with it.

### DIFF
--- a/tensorboard/webapp/metrics/actions/index.ts
+++ b/tensorboard/webapp/metrics/actions/index.ts
@@ -134,3 +134,17 @@ export const cardPinStateToggled = createAction(
   '[Metrics] Card Pin State Toggled',
   props<{cardId: CardId; canCreateNewPins: boolean; wasPinned: boolean}>()
 );
+
+export const timeSelectionChanged = createAction(
+  '[Metrics] Linked Time Selection Changed',
+  props<{
+    startStep: number;
+    startWallTime: number;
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/metrics/actions/index.ts
```diff
@@ -134,3 +134,17 @@ export const cardPinStateToggled = createAction(
...
   props<{cardId: CardId; canCreateNewPins: boolean; wasPinned: boolean}>()
 );
+
+export const timeSelectionChanged = createAction(
+  '[Metrics] Linked Time Selection Changed',
+  props<{
+    startStep: number;
+    startWallTime: number;
+    endStep?: number;
+    endWallTime?: number;
```
japie1235813, line 144: curious why it's named 'wall' time?

japie1235813 (approved):

tensorboard/webapp/metrics/actions/index.ts
stephanwlee: The term originates from https://en.wikipedia.org/wiki/Elapsed_real_time#:~:text=Elapsed%20real%20time%2C%20real%20time,at%20which%20the%20task%20started.

and in context of TensorBoard, the "Wall" option.
image

PR #5177, opened 2021-08-03
histogram: migrate to Angular histogram

stephanwlee:
This changes TimeSeries to use the new Angular based histogram
visualization instead of the Polymer one. This will be used to build
more elaborate features on top of.


stephanwlee:
@psybuzz in case email was not sent out, :ping_pong: :)

psybuzz (commented):
Sorry for the late reply, code in this PR looks good.

After trying this out manually, there's 1 difference in behavior I see.  In the old histogram v1, hovering the cursor over the 'blank space' above the histogram (but still within the bounding rect) will automatically highlight the histogram datum with the smallest step and still shows the cursor tooltip with the value at the nearest x point.

The new histogram v2 does not seem to show that tooltip.  Would this be worth fixing before enabling the transition?

image
image

stephanwlee:
I am aware of the differences and I think it is a reasonable to not make the number stick with the tooltip. If you look at the first screenshot, you don't see any black dot associated with the "selected" or "hovered" slice so the number above feels very arbitrary. Instead, showing the slice value on cursor ONLY when black dot moves felt like an okay substitute to the subtly broken UX.

psybuzz (approved):
lgtm, since we are aware of this difference.

> If you look at the first screenshot, you don't see any black dot associated with the "selected" or "hovered" slice so the number above feels very arbitrary.

I think there is a black dot shown in v1 histograms, but it happens to be occluded by the other colored dots.  For that aspect, histogram v2 is better, since the black dot is always 'on top' of other colored dots.

In this screenshot, we can see the v1 histogram still has a black dot, even when the cursor is over the blank space.

image

stephanwlee:
ACK about my misunderstanding what is happening. Still, I think this is not the most desirable feature of Polymer based histogram.

PR #5182, opened 2021-08-03
linked time: scalar single time step visuals

stephanwlee:
This change adds a vertical bar when there is a single time step
selected by a user. To accomplish this change, we are expanding the
APIs of the `line_chart_component` to allow for visual customization
which `scalar_card` uses to render the vertical bar.

This is first of many to follow so it, by no means, is ready for
consumption.

image


stephanwlee:
Friendly ping, @bmd3k :)

bmd3k (approved):

PR #5183, opened 2021-08-03
router: add a registry module for unsaved experiment updates 

yatbear:
- Defines `DirtyUpdates` interface that currently identifies which experiments have unsaved metadata edits.
- Defines `DirtyUpdatesRegistryModule` that can be used to register unsaved updates.

Googlers, see [b/195041247](https://b.corp.google.com/issues/195041247) for more context.

### DIFF
--- a/tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
+++ b/tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
@@ -30,15 +30,13 @@ export class DirtyUpdatesRegistryModule<State, Updates> {
     @Optional()
     @Inject(DIRTY_UPDATES_TOKEN)
     dirtyUpdatesSelectorFactories: Array<
-      () => DirtyUpdatesSelector<State, Updates>
+      DirtyUpdatesSelector<State, Updates>
     > | null
   ) {
     if (!dirtyUpdatesSelectorFactories) {
       return;
     }
-    this.dirtyUpdatesSelectors = dirtyUpdatesSelectorFactories.map((factory) =>
-      factory()
-    );
+    this.dirtyUpdatesSelectors = dirtyUpdatesSelectorFactories;
   }
 
   /**
@@ -65,7 +63,7 @@ export class DirtyUpdatesRegistryModule<State, Updates> {
    * export class MyModule {}
    */
   static registerDirtyUpdates<State, Updates>(
-    dirtyUpdateSelectorFactory: () => DirtyUpdatesSelector<State, Updates>
+    dirtyUpdateSelectorFactory: DirtyUpdatesSelector<State, Updates>
   ): ModuleWithProviders<DirtyUpdatesRegistryModule<any, {}>> {
     return {
       ngModule: DirtyUpdatesRegistryModule,
--- a/tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
+++ b/tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
@@ -20,30 +20,20 @@ import {
 } from './dirty_updates_registry_types';
 
 @NgModule()
-export class DirtyUpdatesRegistryModule<State, Updates> {
-  private readonly dirtyUpdatesSelectors: DirtyUpdatesSelector<
-    State,
-    Updates
-  >[] = [];
-
+export class DirtyUpdatesRegistryModule<State> {
   constructor(
     @Optional()
     @Inject(DIRTY_UPDATES_TOKEN)
-    dirtyUpdatesSelectorFactories: Array<
-      DirtyUpdatesSelector<State, Updates>
+    private readonly dirtyUpdatesSelectorFactories: Array<
+      DirtyUpdatesSelector<State>
     > | null
-  ) {
-    if (!dirtyUpdatesSelectorFactories) {
-      return;
-    }
-    this.dirtyUpdatesSelectors = dirtyUpdatesSelectorFactories;
-  }
+  ) {}
 
   /**
    * Returns Ngrx selectors for getting dirty updates.
    */
-  getDirtyUpdatesSelectors(): DirtyUpdatesSelector<State, Updates>[] {
-    return this.dirtyUpdatesSelectors ?? [];
+  getDirtyUpdatesSelectors(): DirtyUpdatesSelector<State>[] {
+    return this.dirtyUpdatesSelectorFactories ?? [];
   }
 
   /**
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
```diff
@@ -0,0 +1,78 @@
...
+   *
+   * @NgModule({
+   *   imports: [
+   *     DirtyUpdatesRegistryModule.registerDirtyUpdates(getDirtyExperiments),
+   *   ],
+   *   declarations: [ExperimentsView]
+   * })
+   */
+  static registerDirtyUpdates(
+    dirtyUpdatesConfigProvider: () => string[]
```
stephanwlee, line 65: Instead of list of strings that would be attained at the constructor time, you want user of the API to provide a factory that would return _selectors_ so you can use it against the store to grab the latest value from the Store.

Similar to this: https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/persistent_settings/persistent_settings_config_module.ts#L75 

Do note that the comment is quite dated so take it with a grain of salt ;) 

I suspect changing the contract here will change tests and API of `getDirtyUpdates`.

tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
```diff
@@ -0,0 +1,78 @@
...
+  ) {
+    if (!eidsList) {
+      return;
+    }
+    const experimentIds: string[] = [];
+    for (const eids of eidsList) {
+      for (const eid of eids) {
+        experimentIds.push(eid);
+      }
+    }
```
stephanwlee, line 35: I suspect you will not require this code block after the change but in the future, try `[[1], 2].flat()` instead :)

tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
yatbear: Thanks (once again) for the helpful code pointers. And the suspicion is correct.

tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
yatbear: Second suspicion is also correct!

tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
```diff
@@ -14,63 +14,66 @@ limitations under the License.
...
+    Updates
+  >[] = [];
 
   constructor(
-    @Optional() @Inject(EXPS_WITH_DIRTY_UPDATES_TOKEN) eidsList: string[][]
+    @Optional()
+    @Inject(DIRTY_UPDATES_TOKEN)
+    dirtyUpdatesSelectorFactories: Array<
+      () => DirtyUpdatesSelector<State, Updates>
+    > | null
```
stephanwlee, line 34: Do note that Angular dependency injector is the one that would be invoking the factory and inject only `DirtyUpdatesSelector<State, Updates>`.  That said, I am surprised the test is passing.

tensorboard/webapp/app_routing/dirty_updates_registry_module_test.ts
```diff
@@ -12,42 +12,43 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    return TestBed.inject(DirtyUpdatesRegistryModule);
+  }
+
+  describe('#getDirtyUpdatesSelectors', () => {
+    it('returns selectors for getting the dirty updates', async () => {
+      const getUpdatesInExpList = createSelector(
+        (s) => s,
+        (state: any) => {
+          return {
+            experimentIds: ['otter', 'penguine'],
```
stephanwlee, line 41: Cute animals! :heart:

tensorboard/webapp/app_routing/dirty_updates_registry_types.ts
```diff
@@ -13,7 +13,15 @@ See the License for the specific language governing permissions and
...
 import {InjectionToken} from '@angular/core';
+import {Selector} from '@ngrx/store';
 
-export const EXPS_WITH_DIRTY_UPDATES_TOKEN = new InjectionToken<string[]>(
-  '[App Routing] Experiments with Dirty Updates'
+import {DirtyUpdates} from './types';
+
+export type DirtyUpdatesSelector<
+  State,
+  Updates extends DirtyUpdates
```
stephanwlee, line 22: Actually, I don't think we would want to let user provide a `Partial<Updates>`. Remove this parameter from generics and require all Selectors to return `DirtyUpdates`.

stephanwlee (commented):
Almost there!

tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
yatbear: Fixed. I think the test was problematic as well and is now fixed.

tensorboard/webapp/app_routing/dirty_updates_registry_types.ts
yatbear: Removed partial.

tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
```diff
@@ -0,0 +1,79 @@
...
+    State,
+    Updates
+  >[] = [];
+
+  constructor(
+    @Optional()
+    @Inject(DIRTY_UPDATES_TOKEN)
+    dirtyUpdatesSelectorFactories: Array<
+      DirtyUpdatesSelector<State, Updates>
+    > | null
```
stephanwlee, line 34: If you write
```ts
private readonly dirtyUpdatesSelectorFactories: Array<
      DirtyUpdatesSelector<State, Updates>
    > | null
```

you can get rid of L36-39 as well as L24-27.

tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
```diff
@@ -0,0 +1,79 @@
...
+   *       createSelector(baseSelector, (values) => {
+   *         return {experimentIds: values};
+   *       }),
+   *     ),
+   *   ],
+   * })
+   * export class MyModule {}
+   */
+  static registerDirtyUpdates<State, Updates>(
+    dirtyUpdateSelectorFactory: DirtyUpdatesSelector<State, Updates>
```
stephanwlee, line 66: Since we use `useFactory`, this needs to be typed as a function that returns a selector, i.e., `() => DirtyUpdatesSelector<State>`.

stephanwlee (approved):

tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
yatbear: Thanks!

tensorboard/webapp/app_routing/dirty_updates_registry_module.ts
yatbear: Thanks for the suggestion!

PR #5186, opened 2021-08-04
time series: ignore case when matching tags

stephanwlee:
This change ignores casing when matching tags with regex or strings.


tensorboard/webapp/metrics/views/main_view/main_view_test.ts
```diff
@@ -938,6 +938,18 @@ describe('metrics main view', () => {
       ]);
     });
 
+    fit('ignores case when matching the regex', () => {
```
psybuzz, line 941: `fit` --> `it`!

psybuzz (approved):

Issue #4660, opened 2021-02-06
Add missing docs for the Time Series dashboard


psybuzz:
Tracking issue for adding documentation that describes how to use the
Time Series dashboard. While some of its features are new (run color
customization, pinning), it notably does not require any extra setup
beyond what is already required for the Scalars/Histograms/Images
dashboards.

pindinagesh:
@psybuzz 

Could you please look at this [PR](https://github.com/tensorflow/tensorboard/pull/5193) and confirm if this can be closed?

psybuzz:
Thanks for checking in on this.  Seems reasonable to close.

PR #5193, opened 2021-08-04
docs: document the Time Series dashboard

psybuzz:
This adds documentation that is both user-facing and contributor-facing,
for the Time Series dashboard. This does not touch any of the guides or
tutorials hosted at tensorflow.org.

See https://github.com/tensorflow/tensorboard/issues/4660


docs/timeseries_dashboard.md
```diff
@@ -0,0 +1,44 @@
...
+* Pinned cards: click the 'pin' icon on any card to add it to the pinned section
+at the top for quick comparison.
+
+* Settings: the right pane offers settings for charts and other visualizations.
+Important settings will persist across TensorBoard sessions, when hosted at the
+same URL origin.
+
+* Autocomplete in tag filter: search for specific charts more easily.
+
+## Implementation notes
```
nfelt, line 24: The content here feels like it's a little more on the side of a design doc or a README for plugin developers - it's not really something users would need to know.  That seems a little at odds with the location of this doc, since currently this page (`docs/timeseries_dashboard.md`) is alongside our user-facing plugin docs that are actually published on tensorflow.org/tensorboard.  (This one would also end up published there, although since it's not in the YAML file it wouldn't be very discoverable.)

Conventionally, the more developer-oriented docs have gone in `plugins/foo/README.md` - sometimes that content is also more user-facing but if there's developer-facing content, that's where we put it, e.g.

- https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/plugins/pr_curve/README.md
- https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/plugins/graph/README.md

So I'd recommend putting this content there, and then either omitting this file, or including only the user-facing aspects there (and in that case also adding it to the YAML file so it's linked from the site nav).

README.md
```diff
@@ -235,6 +235,23 @@ projector tutorial](https://www.tensorflow.org/tutorials/text/word_embeddings).
 The Text Dashboard displays text snippets saved via `tf.summary.text`. Markdown
 features including hyperlinks, lists, and tables are all supported.
 
+### Time Series Dashboard (new)
```
nfelt, line 238: I might omit `(new)` - it is new-*er* yes, but not *that* new (released in 2.4.0 last November), and this README is already so stale that I foresee this still saying `(new)` 10 years from now if left in ;)

nfelt (commented):
Thanks for getting to this!

docs/timeseries_dashboard.md
psybuzz: Omitted this file for now, and moved contents to plugins/metrics/README.md.

Makes sense.  I think omitting the user-facing one in this PR would be better, so a separate PR could address it and add to the YAML later.

docs/timeseries_dashboard.md
```diff
@@ -0,0 +1,44 @@
...
+specified in TypeScript at [webapp/metrics/data_source/metrics_backend_types.ts](https://github.com/tensorflow/tensorboard/blob/449273b7c3124283c0837a2caa2f887b1dc6235f/tensorboard/webapp/metrics/data_source/metrics_backend_types.ts).
+
+### Frontend
+
+Notably, the frontend is aware of a few concepts
+
+* Experiment: the currently selected logdir of a TensorBoard instance represents a fixed scope of runs. In this sense, an 'experiment' represents the collection of all the runs in the logdir.
+* RunId: users currently provided a run name during summary writing, for example 'run_directory_1' in `tf.summary.create_file_writer('run_directory_1')`. To support a future with multiple experiments, the RunId concept is a unique identifier of a single run in a specific experiment. Multiple experiments can have a run with the same name 'run_directory_1', but each run will have its own RunId.
```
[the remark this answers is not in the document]
psybuzz: Done.

README.md
psybuzz: Done.

nfelt (approved):

PR #5202, opened 2021-08-05
color grouping: set non-matched runs to light gray

japie1235813:
Currently we do not set any color to non-matched runs, which makes the chart cannot be read at all. Now we set light gray (#ececec) on those non-matched ones.
Screen Shot 2021-08-05 at 2 18 52 PM

### DIFF
--- a/tensorboard/webapp/util/colors.ts
+++ b/tensorboard/webapp/util/colors.ts
@@ -22,4 +22,4 @@ export const CHART_COLOR_PALLETE = [
   '#e8710a', // Orange 600
 ];
 
-export const NONMATCHED_COLOR = '#ececec';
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/util/colors.ts
```diff
@@ -21,3 +21,5 @@ export const CHART_COLOR_PALLETE = [
   '#7cb342', // Lt green 600
   '#e8710a', // Orange 600
 ];
+
+export const NONMATCHED_COLOR = '#ececec';
```
stephanwlee, line 25: `NONMATCHED_COLOR` ->  `NON_MATCHED_COLOR`

stephanwlee (approved):

PR #5203, opened 2021-08-05
Improve error message for proto_test

bmd3k:
Improve error message for proto_test

* Add step for generating protos for RustBoard.
* Add instructions on how to handle new files that must be copied over
to TB for the BUILD to succeed.


bileschi (approved):
Thanks for improving with clear instructions!

PR #5206, opened 2021-08-06
cherry-pick: core: make reload dialog functional (#5200)

bmd3k:
Cherry-picking cf438df778c800113da065e4cbe7a4db2eef38ac 

From original commit:

Because of the discrepancy between store and view on the minimum reload
period, when you set the reload period to be less than 30s, it appeared
correct while it functionally was not true. This change fixes that.

This change also allows user to set 30s sharp which was previouisly
disallowed (this was a bug).

### DIFF
--- a/tensorboard/webapp/settings/_redux/settings_reducers.ts
+++ b/tensorboard/webapp/settings/_redux/settings_reducers.ts
@@ -31,7 +31,7 @@ function settingsReady(state: SettingsState): boolean {
 }
 
 // Auto reload period cannot be lower than 30s to prevent server load.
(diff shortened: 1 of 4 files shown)


bileschi (approved):

bmd3k:
Cherry-picking cf438df778c800113da065e4cbe7a4db2eef38ac 

From original commit:

Because of the discrepancy between store and view on the minimum reload
period, when you set the reload period to be less than 30s, it appeared
correct while it functionally was not true. This change fixes that.

This change also allows user to set 30s sharp which was previouisly
disallowed (this was a bug).

PR #5212, opened 2021-08-06
color group: show message when regex does not match anything

stephanwlee:
We show preview of what runs a regex string would match. When regex
string does not match any runs, we showed nothing in the preview
potentially confusing our users. This change now prints a warning when
non-empty regex matches no entries.

image


japie1235813 (approved):
Thanks for making this change!

PR #5217, opened 2021-08-09
color grouping: trap focus in the dialog

japie1235813:
Under some scenario (the filter run regex has been edited), when users hit ctrl+z, the focus goes back to the filter run input and the dialog input lost the focus. Although we do not prevent the ctrl+z happen, we can exam what is the new activeElement and move the focus back if the new activeElement is not within the regex dialog.

Listen to regex dialog component `focusout` event firing 
use timeout to wait for the `activeElement` to move from the input to new one (Step 2 to 3 in the below flow)

Focus change event sequence: (credit to @stephanwlee)
1. user tabs from input
2. foucsout (or blur); activeElement is still the input
3. focus actually moves and updates activeElement to the next focusable element.
4. focusin (or focus on that specific focusable element)

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_component.ts
+++ b/tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_component.ts
@@ -13,11 +13,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 import {
+  AfterViewInit,
   ChangeDetectionStrategy,
   Component,
+  ElementRef,
   EventEmitter,
   Input,
   Output,
+  ViewChild,
 } from '@angular/core';
 import {MatDialogRef} from '@angular/material/dialog';
 import {Run} from '../../types';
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_component.ts
```diff
@@ -61,4 +69,16 @@ export class RegexEditDialogComponent {
   regexInputChange(regexString: string) {
     this.regexInputOnChange.emit(regexString);
   }
+
+  resetFocus() {
```
stephanwlee, line 73: Make this a private method.

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
```diff
@@ -511,4 +511,19 @@ describe('regex_edit_dialog', () => {
...
+    fixture.detectChanges();
+    const input = fixture.debugElement.query(By.css('input'));
+    // Works around for input element focus initial (cdkFocusInitial)
+    input.nativeElement.focus();
+
+    const div = document.createElement('div');
+    div.focus();
+    fixture.detectChanges();
+
+    expect(document.activeElement).toBe(input.nativeElement);
```
stephanwlee, line 526: I am surprised you don't have to wait a microtask or `tick` to assert on `activeElement`. Could you console.log test and component code to make sure this is actually working as you intended?

For example, before `div.focus`, you can do one `console.log`, put one in `handleFocusOut`, `resetFocus`, and before `expect(document.activeElement).toBe(input.nativeElement);` to see in which order things are being printed.

stephanwlee (approved):

PR #5218, opened 2021-08-09
color grouping: generate preview for example filling

japie1235813:
Previously we listen to `input` of the input, it does not get triggered when user click `try (train|eval)` button, which modify the input value. Now we manually trigger `regexInputChange` method when the `try` button is clicked.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
+++ b/tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
@@ -265,7 +265,7 @@ describe('regex_edit_dialog', () => {
     );
   });
 
-  it('fills example', () => {
+  it('fills example and generates warning', () => {
     const fixture = createComponent(['rose']);
     fixture.detectChanges();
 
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
```diff
@@ -277,8 +277,36 @@ describe('regex_edit_dialog', () => {
...
+    button.nativeElement.click();
+    fixture.detectChanges();
+
+    const groupingResult = fixture.debugElement.query(
+      By.css('.group-container')
+    );
+    expect(groupingResult).not.toBeNull();
+    const groups = fixture.debugElement.queryAll(By.css('.group'));
+    expect(groups.length).toBe(1);
+    tick(500);
```
stephanwlee, line 307: `discardPeriodicTasks` might be better.

stephanwlee (approved):

tensorboard/webapp/runs/views/runs_table/regex_edit_dialog_test.ts
japie1235813: Adjusted. Thanks for the advice.

PR #5219, opened 2021-08-10
core: make sidebar resizable

stephanwlee:
This change introduces new layout component that you can use to have
consistent layout in a plugin. This layout component is responsible for
interactions and remembering the last set width.

image


tensorboard/webapp/core/views/layout_container.ts
```diff
@@ -0,0 +1,117 @@
...
+import {fromEvent, Observable, Subject} from 'rxjs';
+import {filter, takeUntil} from 'rxjs/operators';
+
+import {sideBarWidthChanged} from '../actions';
+import {State} from '../state';
+import {getSideBarWidthInPercent} from '../store/core_selectors';
+import {MouseEventButtons} from '../../util/dom';
+
+@Component({
+  selector: 'tb-layout',
```
psybuzz, line 31: Consider including "resizable" in the name, e.g. "tb-resizable-layout", for more descriptiveness.

tensorboard/webapp/core/views/layout_container.scss
```diff
@@ -0,0 +1,72 @@
...
+  display: flex;
+  flex-direction: row;
+  height: 100%;
+  width: 100%;
+  position: relative;
+}
+
+.sidebar {
+  max-width: 80vw;
+  min-width: 75px;
```
psybuzz, line 27: Could we make this min-width set via something like `[style.minWidth.px]=MIN_SIDEBAR_WIDTH` in the layout_container.ts, so that we only have this constant value '75px' defined in 1 place?

tensorboard/webapp/core/core_module.ts
```diff
@@ -13,25 +13,44 @@ See the License for the specific language governing permissions and
...
-import {DeepLinkerInterface} from '../deeplink';
+import {CORE_FEATURE_KEY} from './store/core_types';
+
+/** @typehack */ import * as _typeHackNgrxStore from '@ngrx/store';
+
+export function getSideBarWidthSetting() {
+  return createSelector<State, number, PersistableSettings>(
+    getSideBarWidthInPercent,
+    (sideBarWidthInPercent) => {
+      return {sideBarWidthInPercent};
```
psybuzz, line 40: If we plan to ever have another resizable layout in the UI later on, which also needs its width persisted, it would be reasonable to separately store their widths.  To prevent future ambiguity as to which sidebar this setting refers to, would it reasonable to name this property something specific like `metricsRunSidebarWidth`?

tensorboard/webapp/core/store/core_types.ts
```diff
@@ -37,6 +37,8 @@ export interface CoreState {
   // For now, we want them here for Polymer interop states reasons, too.
   polymerInteropRuns: Run[];
   polymerInteropRunSelection: Set<RunId>;
+  // Number between 0 and 100.
+  sideBarWidthInPercent: number;
```
psybuzz, line 41: I originally expected to see the value stored under metrics/, since it is specific to Time Series UI.  Having it under core/ makes sense when under the assumption that there will not be other resizable layouts.  Would it be better to move this field, the reducer, selector, etc to metrics/ as "(get)metricsRunSidebarWidth", if we want other features besides metrics/ to have persisted sidebar widths?  e.g. maybe the Hparams dashboard remembers its own sidebar width separately in the future.

If not, could we at least add comments to here and in the selector definition that mention it is specific to Time Series' run sidebar?

psybuzz (commented):
optional: In some products, the sidebar width is set using an absolute unit (e.g. px) rather than as a percentage, so that resizing the browser window doesn't affect all the sidebar widths inside.  Would it make more sense to set 'px' instead?

tensorboard/webapp/core/store/core_types.ts
stephanwlee: > This change introduces new layout component that you can use to have
consistent layout in a plugin.

This component is expected to be used by other plugins as the description indicates.

For instance, I want this to be usable in npmi plugin which I have in a follow up change. I think it makes sense for all sidebar widths to be linked thus this state remains in the Core plugin.

tensorboard/webapp/core/core_module.ts
stephanwlee: I think my comment below makes this obsolete.

stephanwlee:
> optional: In some products, the sidebar width is set using an absolute unit (e.g. px) rather than as a percentage, so that resizing the browser window doesn't affect all the sidebar widths inside. Would it make more sense to set 'px' instead?

The opposite was what I was looking for. I wanted the width of the sidebar to be relative to the size of the window which often is what I am looking for in a UI.

tensorboard/webapp/core/views/layout_container.ts
stephanwlee: I am iffy about that. I want this to be used in all pages' layout when possible like https://github.com/tensorflow/tensorboard/blob/master/tensorboard/components/tf_dashboard_common/tf-dashboard-layout.ts#L23 and whether it is resizable or not is not the part I want plugin authors to care/know.

psybuzz: Hm, then perhaps something along the lines of `tb-dashboard-layout` or `tb-plugin-dashboard-layout`?  With `tb-layout`, it's unclear to me when it should be used.

tensorboard/webapp/core/store/core_types.ts
psybuzz: Thanks for clarifying, I see where this is going now.  This setting and the tb-layout component isn't to be used to make any resizable sidebar, (e.g. not to be used for cases like the Time Series settings pane), but specifically only for replacing the top-level layout for each plugin dashboard.

psybuzz (approved):

PR #5223, opened 2021-08-10
router: warn user if there are unsaved edits when navigating away 

yatbear:
Warn user with `window.confirm` if there are unsaved edits when navigating away from the current route:
- If user confirms, discards the edits and navigate to the new route;
- Otherwise discontinue the `dispatchNavigating$`.

Tested locally in TensorBoard.corp: https://screencast.googleplex.com/cast/NjQ0NTg4OTcxNzUzNDcyMHxhNjFjMTViOS0zZQ

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
@@ -16,7 +16,7 @@ limitations under the License.
 import {Component} from '@angular/core';
 import {fakeAsync, flush, TestBed, tick} from '@angular/core/testing';
 import {provideMockActions} from '@ngrx/effects/testing';
-import {Action, createAction, props, Store} from '@ngrx/store';
+import {Action, createAction, createSelector, props, Store} from '@ngrx/store';
 import {MockStore, provideMockStore} from '@ngrx/store/testing';
 import {of, ReplaySubject} from 'rxjs';
 
@@ -25,6 +25,7 @@ import * as actions from '../actions';
 import {navigationRequested} from '../actions';
 import {AppRootProvider, TestableAppRootProvider} from '../app_root';
 import {Location} from '../location';
+import {DirtyUpdatesRegistryModule} from '../dirty_updates_registry_module';
 import {
   NavigateToCompare,
   NavigateToExperiments,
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/app_routing/store/app_routing_reducers.ts
```diff
@@ -23,6 +23,9 @@ const initialState: AppRoutingState = {
 
 const reducer = createReducer(
   initialState,
+  on(actions.discardDirtyUpdates, (state) => {
+    return {...state};
+  }),
```
stephanwlee, line 28: Please undo this change; this is basically noop and we do not need to make changes to the app_routing state based on this action at all.

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -255,6 +277,94 @@ describe('app_routing_effects', () => {
...
+              pathname: '/experiment/meow',
+            })
+          );
+
+          expect(window.confirm).toHaveBeenCalledTimes(1);
+          expect(actualActions).toEqual([actions.discardDirtyUpdates()]);
+
+          tick();
+          expect(actualActions).toEqual([
+            actions.navigating({
```
stephanwlee, line 341: I am surprised that the test is passing here. Shouldn't the `actualActions` be an array with two elements where the first element in the `actions.dsicardDirtyUpdates()`? Similarly, I  expected the L352 to have three elements.  What am I missing?

yatbear:
Updated the local test screencast, clicking on back button seems to only work when the navigation is `in app`.

tensorboard/webapp/app_routing/store/app_routing_reducers.ts
yatbear: Done. Thanks!

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
yatbear: I used `fakeAsync` incorrectly before and it didn't capture the errors. Fixed.

stephanwlee (approved):

PR #5237, opened 2021-08-12
line chart: introduce onContextLost renderer callback

psybuzz:
Each line chart component in the Time Series dashboard gets its own canvas and
webgl rendering context. After scrolling through too many line charts (for my
device after ~30), the page issues warnings in the console like "Too many active
WebGL contexts. Oldest context will be lost".

When the browser decides there are too many contexts, it disposes old ones
causing blank line charts. This change adds an event dispatched by the line chart's
internal `ChartImpl` to notify clients when the rendering context is lost.

This only applies to the WebGL renderer for now.  A followup change addresses the
remediation of blank scalar cards.

Googlers, see b/196294346.

Diffbase: none.
Followup: https://github.com/tensorflow/tensorboard/pull/5239

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/chart.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/chart.ts
@@ -66,7 +66,8 @@ export class ChartImpl implements Chart {
         this.renderer = new ThreeRenderer(
           option.container,
           coordinator,
-          option.devicePixelRatio
+          option.devicePixelRatio,
+          option.callbacks.onContextLost
         );
         break;
       }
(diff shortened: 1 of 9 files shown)


japie1235813 (approved):
The structure lgtm. 
What thing is not clear to me: what do you think the app should do when the context lost happens?

psybuzz:
> what do you think the app should do when the context lost happens?

Upon losing webgl context, my intention is to have the LineChartComponent listen for that renderer event and tell its parent component that a "fatal renderer error" has occurred.  In that scenario, a ScalarCardComponent, which owns a LineChartComponent, can decide to destroy and re-create the LineChartComponent upon a fatal error.

Sorry, I realize it wasn't simple to view the followup PR's changes independently.  I will rebase the followup before requesting review, and remember to use branches in the main repo for easier diffing in the future.

PR #5239, opened 2021-08-13
line chart: recreate charts upon fatal renderer errors

psybuzz:
The line chart's internal `ChartImpl` notifies the line chart when it is a WebGL
renderer that loses its context due to the page having too many active WebGL
contexts at once (>30 for my device).

This change adds logic to treat those errors as fatal when the line chart needs
to be updated.  Scalar cards now handle the fatal error by destroying and
recreating the entire line chart component, fixing the blank chart issue.

Manually checked that charts are not blank after scrolling through lots of line
charts (used the tagFilter to view 2x the line charts, since elements are
duplicated).

Alternatives considered: calling restoreContext() [1] on the context does not
consistently fix the blank charts.  When it does, it appears to lose the original
context configuration (antialias: true, precision: 'highp', alpha: true), leading to
low quality visuals [2].

Googlers, see b/196294346.

[1] https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_lose_context/restoreContext
[2] https://imgur.com/a/kKnhqYJ

Diffbase: https://github.com/tensorflow/tensorboard/pull/5237
Followup: none

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
@@ -18,10 +18,12 @@ import {
   ChangeDetectorRef,
   Component,
   ElementRef,
+  EventEmitter,
   Input,
   OnChanges,
   OnDestroy,
   OnInit,
+  Output,
   SimpleChanges,
   TemplateRef,
   ViewChild,
@@ -150,6 +152,9 @@ export class LineChartComponent
   @Input()
   ignoreYOutliers: boolean = false;
 
+  @Output()
+  onRendererFatalError = new EventEmitter<void>();
+
   readonly Y_GRID_COUNT = 6;
   readonly X_GRID_COUNT = 10;
 
@@ -173,6 +178,8 @@ export class LineChartComponent
   // onChanges.
   private isViewBoxChanged = true;
   private scaleUpdated = true;
+  private isRenderingContextLost = false;
+  private didEmitRendererFatalError = false;
 
   constructor(private readonly changeDetector: ChangeDetectorRef) {}
 
@@ -182,6 +189,10 @@ export class LineChartComponent
   }
 
   ngOnChanges(changes: SimpleChanges) {
+    if (changes['disableUpdate'] && !this.disableUpdate) {
+      this.checkRendererFatalError();
+    }
+
     // OnChanges only decides whether props need to be updated and do not directly update
     // the line chart.
 
@@ -230,6 +241,14 @@ export class LineChartComponent
     // After view is initialized, if we ever change the Angular prop that should propagate
     // to children, we need to retrigger the Angular change.
     this.changeDetector.detectChanges();
+  }
+
+  checkRendererFatalError() {
+    if (this.didEmitRendererFatalError || !this.isRenderingContextLost) {
+      return;
+    }
+    this.didEmitRendererFatalError = true;
+    this.onRendererFatalError.emit();
   }
 
   onViewResize() {
@@ -289,16 +308,27 @@ export class LineChartComponent
     return false;
   }
 
+  private onContextLost() {
+    // After context is lost, we intentionally wait until updates are enabled or
+    // requested before notifying clients that a fatal error occurred. Errors
+    // occurring when the component is hidden are not fatal enough to notify.
+    this.isRenderingContextLost = true;
+  }
+
+  triggerContextLostForTest() {
+    this.onContextLost();
+  }
+
   private initializeChart() {
     if (this.lineChart) {
       throw new Error('LineChart should not be initialized multiple times.');
     }
 
     const rendererType = this.getRendererType();
-    // Do not yet need to subscribe to the `onDrawEnd`.
     const callbacks: ChartCallbacks = {
+      // Do not yet need to subscribe to the `onDrawEnd`.
       onDrawEnd: () => {},
-      onContextLost: () => {},
+      onContextLost: this.onContextLost.bind(this),
     };
     let params: ChartOptions | null = null;
 
@@ -368,6 +398,7 @@ export class LineChartComponent
    */
   private updateLineChart() {
     if (!this.lineChart || this.disableUpdate) return;
+    this.checkRendererFatalError();
 
     if (this.scaleUpdated) {
       this.scaleUpdated = false;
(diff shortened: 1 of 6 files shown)


stephanwlee:
Would you have a material you can share on whether this is happening because of too many contexts? Casual Google search tells me that this can happen when a canvas becomes invisible while first creating/drawing it. Also there was some conversation about the library attempting to recover from the event automatically <https://github.com/mrdoob/three.js/issues/5857#issuecomment-69156171>.  Even with the remediation, it is believable that Three.js will fail to recover when the canvas leaves the viewport but, again, I do not get an impression that this happens because of # of contexts (unless your memory is running out and browser is eagerly removing it). 

If this truly happens only when becoming invisible while drawing the first frame, I think we can do away without propagating the information all the way up to scalar_card_component but instead have internal state in line_chart_component.ts and based on the `disableUpdate` value, we can decide whether to reattempt to bootstrap or not. WDYT?

psybuzz:
From what I've seen manually, blank charts can be consistently triggered well after the canvas is initially drawn and visible.  In the attached screencast, TensorBoard loads the dashboard containing 100s of scalar cards, then after scrolling through about 30-40, we scroll 3 scalar cards into view, and observe 3 "too many contexts" warnings in the console.  After scrolling up, we can hide and reshow all the original cards (via entering a filter and removing the tag filter).  Doing so produces exactly 3 blank cards, suggesting that for all these cards that were originally drawn with a visible DOM element, the only difference between a blank one and a non-blank one after the repro is the blank ones were the ones 'evicted' by the browser.

Based on that, I think it's safe to say that listening to some 'context lost' signal is necessary.  It's unclear to me what the exact max limit on webgl contexts is or whether it's dependent on GPU memory, but [this crbug](https://bugs.chromium.org/p/chromium/issues/detail?id=771792) suggests it might be device dependent in the future.

[webglContextLossBlankChart.webm.zip](https://github.com/tensorflow/tensorboard/files/6995674/webglContextLossBlankChart.webm.zip)

After scrolling through more and more cards, the warning just keep piling on.

image


> I think we can do away without propagating the information all the way up to scalar_card_component but instead have internal state in line_chart_component.ts and based on the disableUpdate value, we can decide whether to reattempt to bootstrap or not. WDYT?

It might be possible to re-bootstrap the line chart from within line_chart_component without making scalar_card_component aware of things.  However, given how stateful the LineChartComponent is, I opted to do the 'full re-creation' in ScalarCardComponent, as it is much less complexity.

stephanwlee (commented):
At the very least, can we make it so that the context loss issue is contained in the LineChartComponent (and it "blinks" using its own `ngIf`) instead of propagating it all the way up to the ScalarCardComponent? 

Also, if your empirical and crbugs are correct, then depending on screen, this bug will happen no matter what if there are so many charts above the fold, right? Perhaps we should be using half SVG and half Canvas in the future.

psybuzz:
Ready for another look!

> At the very least, can we make it so that the context loss issue is contained in the LineChartComponent (and it "blinks" using its own ngIf) instead of propagating it all the way up to the ScalarCardComponent?

Done.

> Also, if your empirical and crbugs are correct, then depending on screen, this bug will happen no matter what if there are so many charts above the fold, right?

Yes.  After hiding and showing cards (e.g. enter the filtered view then clear the filter), the early cards were blank.

> Perhaps we should be using half SVG and half Canvas in the future.

Yeah, something to consider in the future.  This webgl limitation was quite surprising to me, and I think TensorBoard's use case is fairly unique among webapps.

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
```diff
@@ -289,16 +322,30 @@ export class LineChartComponent
     return false;
   }
 
+  private onContextLost() {
+    // Since context may be lost when the component is hidden or does not need
+    // updates, the teardown and recreation happens lazily.
+    this.isRenderingContextLost = true;
```
stephanwlee, line 328: For some reason, I thought you'd dispose `lineChart` here and set it to `this.lineChart = null`. Would that make the code more readable in anyway?

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
```diff
@@ -364,10 +411,13 @@ export class LineChartComponent
   }
 
   /**
-   * Minimally and imperatively updates the chart library depending on prop changed.
+   * Minimally and imperatively updates the chart library depending on prop
+   * changed. When adding new `lineChart.*()` calls, keep this in sync with
```
stephanwlee, line 415: I do not see any `lineChart.*` API invoked in `recoverRendererIfNeeded`. Do you truly need them to be synced?

tensorboard/webapp/widgets/line_chart_v2/line_chart_component_test.ts
```diff
@@ -648,4 +648,162 @@ describe('line_chart_v2/line_chart test', () => {
...
+      expectChartUpdateSpiesToHaveBeenCalledTimes(1); // Initial render.
+
+      fixture.componentInstance.chart.triggerContextLostForTest();
+      fixture.detectChanges();
+
+      expect(fixture.componentInstance.chart.getLineChartForTest()).toBe(
+        chartBefore
+      );
+      expect(disposeSpy).toHaveBeenCalledTimes(0);
+      expectChartUpdateSpiesToHaveBeenCalledTimes(1);
```
stephanwlee, line 679: Unclear whether these assertions are truly testing whether the line chart is attempting to recover. I guess assertion of `lineChartBefore === lineChartAfter` is the one? If so, put a semantic name on the assertion method so we can change the assertion easily.

stephanwlee (approved):

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
psybuzz: Done, yes a bit more expected.

tensorboard/webapp/widgets/line_chart_v2/line_chart_component.ts
psybuzz: Updated the comment to be `this.lineChart.set*()`.  If we added a `this.lineChart.setZMode(...)` inside `updateLineChart` one day, we'd need the corresponding flag `this.zModeUpdated = true` inside `recoverRendererIfNeeded` so that any changes to initial values of this hypothetical "zMode" would persist after recovery.

tensorboard/webapp/widgets/line_chart_v2/line_chart_component_test.ts
psybuzz: Good idea, done.

psybuzz (commented):
Thanks for the review!

PR #5240, opened 2021-08-13
notification: clip the date to today

stephanwlee:
There were cases when the notification payload we handcrafted is in the
future and since we only dismiss the notification by recording the last
read time (NOW), there were cases when new notifications in the future
are never dismissable. To prevent developer mistakes, we can clip the
date to start of the day so a notification can never been in the future.

### DIFF
--- a/tensorboard/webapp/notification_center/_data_source/notification_center_data_source.ts
+++ b/tensorboard/webapp/notification_center/_data_source/notification_center_data_source.ts
@@ -41,6 +41,7 @@ export class TBNotificationCenterDataSource
           return {
             ...response,
             notifications: response.notifications.map((notification) => {
+              if (!notification.hasOwnProperty('dateInMs')) return notification;
               return {
                 ...notification,
                 // `dateInMs` can never exceed local machine's today at 12:00 AM.
--- a/tensorboard/webapp/notification_center/_data_source/notification_center_data_source.ts
+++ b/tensorboard/webapp/notification_center/_data_source/notification_center_data_source.ts
@@ -14,6 +14,7 @@ limitations under the License.
 ==============================================================================*/
 import {Injectable} from '@angular/core';
 import {Observable} from 'rxjs';
+import {map} from 'rxjs/operators';
 
 import {TBHttpClient} from '../../webapp_data_source/tb_http_client';
 import {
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/notification_center/_data_source/notification_center_data_source_test.ts
```diff
@@ -84,4 +84,30 @@ describe('TBNotificationCenterDataSource test', () => {
     expect(resultSpy).not.toHaveBeenCalled();
     expect(errorSpy).toHaveBeenCalledWith(httpErrorResponse);
   });
+
+  fit(
```
japie1235813: nit: remove `fit`

tensorboard/webapp/notification_center/_data_source/notification_center_data_source.ts
```diff
@@ -30,6 +31,24 @@ export class TBNotificationCenterDataSource
...
 
   fetchNotifications(): Observable<NotificationCenterResponse> {
-    return this.http.get<NotificationCenterResponse>(`/data/notifications`);
+    return this.http
+      .get<NotificationCenterResponse>(`/data/notifications`)
+      .pipe(
+        map((response) => {
+          const todayInMs: number = new Date(
+            new Date().toDateString()
+          ).getTime();
```
japie1235813, line 40: can we use `(new Date()).getTime()`?
(which returns the current time)
Or do you feel strongly about 12AM?

japie1235813 (approved):

tensorboard/webapp/notification_center/_data_source/notification_center_data_source.ts
stephanwlee: I think so. Imagine this:
1. there is a notification at 9pm today
2. Now is 1pm so when fetched data, it gets clipped to 1pm.
3. you first load TensorBoard and read notification at 1pm today
4. Wait 2hrs and now reload TensorBoard
5. Now notifications are clipped at 3pm while you last read notification at 1pm so red dot shows up.

nfelt: Doesn't clipping to 12am still result in problems if the notification is more than 1 day in the future?

1. it's Monday at 9am
2. notification scheduled for Wednesday at 9am
3. notification timestamp is clipped to 12am Monday
4. user successfully marks notification as read
5. Tuesday morning: notification timestamp is clipped to 12am Tuesday
6. user sees unread notification again

IMO it seems better if we fix this further downstream (not in the fetching phase) by just not showing notifications at all if they're still in the future.  That would let us use future-timestamped notifications to deliberately roll something out at a precise time if we needed that for some reason.

Or to put it another way, we'd have 3 possible states for notifications, A) in the future, B) in the past but more recent than last read timestamp, or C) in the past and older than last read timestamp.

The unread count / red dot only considers notifications in state B, and the actual notification panel only shows notifications in states B and C, so notifications in state A are entirely hidden.

stephanwlee: Luckily, we have been good with at least getting the date right so I am not very worried about that. 

>  deliberately roll something out at a precise time if we needed that for some reason.

Rolling something out at certain date makes sense but I am not sure if we want to have a precise control over the millisecond precision here. Moreover, releases often is accompanied by a feature flag flip which is not time controlled so you will need to have a release anyways.

PR #5242, opened 2021-08-13
ui: fix wrapping of dropdown items

psybuzz:
Long items within tb-dropdown are supposed to wrap onto multiple
lines. Our CSS was incorrectly using a selector without enough
specificity, and Angular was overriding our classname.

Before
image

After
image

### DIFF
--- a/tensorboard/webapp/widgets/dropdown/dropdown_component.scss
+++ b/tensorboard/webapp/widgets/dropdown/dropdown_component.scss
@@ -32,8 +32,9 @@ mat-select:focus {
   max-width: 70vw;
 }
 
-.tb-mat-option {
-  // Overrides Angular Material's fixed height.
+// This selector needs relatively more specificity than Angular Material's
+// to override the fixed height.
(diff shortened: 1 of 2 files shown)


stephanwlee:
Would it be at all possible to use native select instead of wrapping content? I have not seen many instances where there are multi-line formatting on `<select>`

psybuzz:
> Would it be at all possible to use native select instead of wrapping content? I have not seen many instances where there are multi-line formatting on `<select>`

The original motivation for making items wrap is so that very long experiment names in an experiment dropdown would be at least readable in their entirety.  The longest tag names are usually not as long as the longest experiment names, but they happen to share the same component.  The tradeoff is we either can show the full option text on multiple lines, or avoid wrapping entirely and lose the full name, which may help distinguish similar long experiment names.  Could you confirm which is preferable?

`<tb-dropdown>` in the tag filter updates as the user types, so I think opening a native `<select>` would steal focus from the text input field, interrupting typing.

bmd3k (approved):

PR #5243, opened 2021-08-13
infra: allow enabling the ngrx debug log at runtime

psybuzz:
While investigating performance issues in environments with a
release build of TensorBoard, it can be convenient to see the ngrx log
of actions and states in the devtools console.

This introduces a flag configurable at runtime on the window object to
turn on or off this debug logging for TensorBoard developers.


psybuzz:
I forgot that ?_IntegrationTest also enables the logging, so closing this.

Issue #5165, opened 2021-07-29
Support fsspec for reading remote directories


f4hy:
A common gotcha on my team is using `hdfs` or other remote storage systems as the location to store the tensorboard logs. If `tensorflow>2.0` is installed this works just fine, but if tf is not installed at all then writing the output works only for local or s3 locations.

When using tensorboard to write logs installing tensorflow just to get remote directory access is a pretty heavy dependency to add.  Would the team be open to replacing the remote file access with something like `fsspec` (what pandas uses for remote file support) instead of gfile? If so I would be happy to take on the project.

stephanwlee:
Hi @f4hy, sorry for a late reply. 

I would like to understand your set up a little bit more. TensorBoard today does not let you write any summaries without TensorFlow. Are you asking for a possibility where you can emit logs without dependency on TensorFlow? If so, have you ever looked into [tensorboardX](https://github.com/lanpa/tensorboardX) (not our project)? tensorboardX does not support hdfs but if that's what you are using today, I would ask you to file an issue against them instead.

Above, you were explicitly asking about the writer parts of TensorBoard but are you at all hindered by the read path? Without TensorFlow installed, both rustboard and our TF shims do not support hdfs.
- https://github.com/tensorflow/tensorboard/blob/b6e8baf0e334e9e16c7157503a704cca511029b2/tensorboard/data/server/cli/dynamic_logdir.rs#L29-L32
- https://github.com/tensorflow/tensorboard/blob/b6e8baf0e334e9e16c7157503a704cca511029b2/tensorboard/compat/tensorflow_stub/io/gfile.py#L404-L406

f4hy:
I am writing to tensorboard through pytorch, https://pytorch.org/docs/stable/tensorboard.html 
[...]

PR #5248, opened 2021-08-17
gfile: add support for fsspec filesystems

d4l3k:

This adds support for fsspec filesystem backends to the gfile compat library. This allows for tensorboard to be used with a wider range of file system backends.

See: https://github.com/tensorflow/tensorboard/issues/5165

This adds a new GFile filesystem backend that's used if no other registered filesystems are present. In that case fsspec is checked to see if it has an available one and then returns a wrapped version.

fsspec provides similar semantics to normal usage however, it doesn't support seeking for files opened in text mode which can be problematic for longer text files. However, this seems to work regardless since most of tensorboards larger files are in a binary format.

I've copied the existing local filesystem unit tests and changed the paths to be prefixed with `file://` so they go through the fsspec backend.

```
$ bazel run //tensorboard -- --logdir file:///tmp/torchxrun/lightning_logs/version_1/
$ bazel test //tensorboard/compat/tensorflow_stub:gfile_fsspec_test //tensorboard/summary/writer:event_file_writer_fsspec_test --test_output=errors
```

### DIFF
--- a/tensorboard/compat/tensorflow_stub/BUILD
+++ b/tensorboard/compat/tensorflow_stub/BUILD
@@ -16,6 +16,7 @@ py_library(
     srcs_version = "PY3",
     deps = [
         "//tensorboard:expect_absl_flags_installed",
+        "//tensorboard:expect_fsspec_installed",
         "//tensorboard:expect_numpy_installed",
         "//tensorboard/compat/proto:protos_all_py_pb2",
     ],
@@ -56,6 +57,19 @@ py_test(
     ],
     deps = [
         ":tensorflow_stub",
+        "//tensorboard:test",
+    ],
+)
+
+py_test(
+    name = "gfile_fsspec_test",
+    size = "small",
+    srcs = ["io/gfile_fsspec_test.py"],
+    srcs_version = "PY3",
+    tags = ["support_notf"],
+    deps = [
+        ":tensorflow_stub",
+        "//tensorboard:expect_fsspec_installed",
         "//tensorboard:test",
     ],
 )
--- a/tensorboard/compat/tensorflow_stub/io/gfile.py
+++ b/tensorboard/compat/tensorflow_stub/io/gfile.py
@@ -24,6 +24,7 @@ from collections import namedtuple
 import glob as py_glob
 import io
 import os
+import os.path
 import sys
 import tempfile
 
@@ -34,6 +35,13 @@ try:
     S3_ENABLED = True
 except ImportError:
     S3_ENABLED = False
+
+try:
+    import fsspec
+
+    FSSPEC_ENABLED = True
+except ImportError:
+    FSSPEC_ENABLED = False
 
 if sys.version_info < (3, 0):
     # In Python 2 FileExistsError is not defined and the
@@ -69,6 +77,8 @@ def get_filesystem(filename):
     if index >= 0:
         prefix = filename[:index]
     fs = _REGISTERED_FILESYSTEMS.get(prefix, None)
+    if fs is None:
+        fs = _get_fsspec_filesystem(filename)
     if fs is None:
         raise ValueError("No recognized filesystem for prefix %s" % prefix)
     return fs
@@ -401,6 +411,242 @@ class S3FileSystem(object):
                 raise
 
 
+class FSSpecFileSystem(object):
+    """Provides filesystem access via fsspec.
+
+    The current gfile interface doesn't map perfectly to the fsspec interface
+    leading to some notable inefficiencies.
+
+    * Reads and writes to files cause the file to be reopened each time which
+      can cause a performance hit when accessing local file systems.
+    * walk doesn't use the native fsspec walk function so performance may be
+      slower.
+
+    See https://github.com/tensorflow/tensorboard/issues/5286 for more info on
+    limitations.
+    """
+
+    SEPARATOR = "://"
+    CHAIN_SEPARATOR = "::"
+
+    def _validate_path(self, path):
+        parts = path.split(self.CHAIN_SEPARATOR)
+        for part in parts[:-1]:
+            if self.SEPARATOR in part:
+                raise errors.InvalidArgumentError(
+                    None,
+                    None,
+                    "fsspec URL must only have paths in the last chained filesystem, got {}".format(
+                        path
+                    ),
+                )
+
+    def _translate_errors(func):
+        def func_wrapper(self, *args, **kwargs):
+            try:
+                return func(self, *args, **kwargs)
+            except FileNotFoundError as e:
+                raise errors.NotFoundError(None, None, str(e))
+
+        return func_wrapper
+
+    def _fs_path(self, filename):
+        if isinstance(filename, bytes):
+            filename = filename.decode("utf-8")
+        self._validate_path(filename)
+
+        fs, path = fsspec.core.url_to_fs(filename)
+        return fs, path
+
+    @_translate_errors
+    def exists(self, filename):
+        """Determines whether a path exists or not."""
+        fs, path = self._fs_path(filename)
+        return fs.exists(path)
+
+    def _join(self, sep, paths):
+        """
+        _join joins the paths with the given separator.
+        """
+        result = []
+        for part in paths:
+            if part.startswith(sep):
+                result = []
+            if result and result[-1] and not result[-1].endswith(sep):
+                result.append(sep)
+            result.append(part)
+        return "".join(result)
+
+    @_translate_errors
+    def join(self, path, *paths):
+        """Join paths with a slash."""
+        self._validate_path(path)
+
+        before, sep, last_path = path.rpartition(self.CHAIN_SEPARATOR)
+        chain_prefix = before + sep
+        protocol, path = fsspec.core.split_protocol(last_path)
+        fs = fsspec.get_filesystem_class(protocol)
+        if protocol:
+            chain_prefix += protocol + self.SEPARATOR
+        return chain_prefix + self._join(fs.sep, ((path,) + paths))
+
+    @_translate_errors
+    def read(self, filename, binary_mode=False, size=None, continue_from=None):
+        """Reads contents of a file to a string.
+
+        Args:
+            filename: string, a path
+            binary_mode: bool, read as binary if True, otherwise text
+            size: int, number of bytes or characters to read, otherwise
+                read all the contents of the file (from the continuation
+                marker, if present).
+            continue_from: An opaque value returned from a prior invocation of
+                `read(...)` marking the last read position, so that reading
+                may continue from there.  Otherwise read from the beginning.
+
+        Returns:
+            A tuple of `(data, continuation_token)` where `data' provides either
+            bytes read from the file (if `binary_mode == true`) or the decoded
+            string representation thereof (otherwise), and `continuation_token`
+            is an opaque value that can be passed to the next invocation of
+            `read(...) ' in order to continue from the last read position.
+        """
+        fs, path = self._fs_path(filename)
+
+        mode = "rb" if binary_mode else "r"
+        encoding = None if binary_mode else "utf8"
+        if not exists(filename):
+            raise errors.NotFoundError(
+                None, None, "Not Found: " + compat.as_text(filename)
+            )
+        with fs.open(path, mode, encoding=encoding) as f:
+            if continue_from is not None:
+                if not f.seekable():
+                    raise errors.InvalidArgumentError(
+                        None,
+                        None,
+                        "{} is not seekable".format(filename),
+                    )
+                offset = continue_from.get("opaque_offset", None)
+                if offset is not None:
+                    f.seek(offset)
+
+            data = f.read(size)
+            # The new offset may not be `offset + len(data)`, due to decoding
+            # and newline translation.
+            # So, just measure it in whatever terms the underlying stream uses.
+            continuation_token = (
+                {"opaque_offset": f.tell()} if f.seekable() else {}
+            )
+            return (data, continuation_token)
+
+    @_translate_errors
+    def write(self, filename, file_content, binary_mode=False):
+        """Writes string file contents to a file.
+
+        Args:
+            filename: string, a path
+            file_content: string, the contents
+            binary_mode: bool, write as binary if True, otherwise text
+        """
+        self._write(filename, file_content, "wb" if binary_mode else "w")
+
+    @_translate_errors
+    def append(self, filename, file_content, binary_mode=False):
+        """Append string file contents to a file.
+
+        Args:
+            filename: string, a path
+            file_content: string, the contents to append
+            binary_mode: bool, write as binary if True, otherwise text
+        """
+        self._write(filename, file_content, "ab" if binary_mode else "a")
+
+    def _write(self, filename, file_content, mode):
+        fs, path = self._fs_path(filename)
+        encoding = None if "b" in mode else "utf8"
+        with fs.open(path, mode, encoding=encoding) as f:
+            compatify = compat.as_bytes if "b" in mode else compat.as_text
+            f.write(compatify(file_content))
+
+    def _get_chain_protocol_prefix(self, filename):
+        chain_prefix, chain_sep, last_path = filename.rpartition(
+            self.CHAIN_SEPARATOR
+        )
+        protocol, sep, _ = last_path.rpartition(self.SEPARATOR)
+        return chain_prefix + chain_sep + protocol + sep
+
+    @_translate_errors
+    def glob(self, filename):
+        """Returns a list of files that match the given pattern(s)."""
+        if isinstance(filename, bytes):
+            filename = filename.decode("utf-8")
+
+        fs, path = self._fs_path(filename)
+        files = fs.glob(path)
+
+        # check if applying the original chaining is required.
+        if (
+            self.SEPARATOR not in filename
+            and self.CHAIN_SEPARATOR not in filename
+        ):
+            return files
+
+        prefix = self._get_chain_protocol_prefix(filename)
+
+        return [
+            file
+            if (self.SEPARATOR in file or self.CHAIN_SEPARATOR in file)
+            else prefix + file
+            for file in files
+        ]
+
+    @_translate_errors
+    def isdir(self, dirname):
+        """Returns whether the path is a directory or not."""
+        fs, path = self._fs_path(dirname)
+        return fs.isdir(path)
+
+    @_translate_errors
+    def listdir(self, dirname):
+        """Returns a list of entries contained within a directory."""
+        fs, path = self._fs_path(dirname)
+        files = fs.listdir(path, detail=False)
+        files = [os.path.basename(fname) for fname in files]
+        return files
+
+    @_translate_errors
+    def makedirs(self, dirname):
+        """Creates a directory and all parent/intermediate directories."""
+        fs, path = self._fs_path(dirname)
+        return fs.makedirs(path, exist_ok=True)
+
+    @_translate_errors
+    def stat(self, filename):
+        """Returns file statistics for a given path."""
+        fs, path = self._fs_path(filename)
+        return StatData(fs.size(path))
+
+
+_FSSPEC_FILESYSTEM = FSSpecFileSystem()
+
+
+def _get_fsspec_filesystem(filename):
+    """
+    _get_fsspec_filesystem checks if the provided protocol is known to fsspec
+    and if so returns the filesystem wrapper for it.
+    """
+    if not FSSPEC_ENABLED:
+        return None
+
+    segment = filename.partition(FSSpecFileSystem.CHAIN_SEPARATOR)[0]
+    protocol = segment.partition(FSSpecFileSystem.SEPARATOR)[0]
+    if fsspec.get_filesystem_class(protocol):
+        return _FSSPEC_FILESYSTEM
+    else:
+        return None
+
+
 register_filesystem("", LocalFileSystem())
 if S3_ENABLED:
     register_filesystem("s3", S3FileSystem())
@@ -514,6 +760,7 @@ class GFile(object):
                 # write the first chunk to truncate file if it already exists
                 self.fs.write(self.filename, file_content, self.binary_mode)
                 self.write_started = True
+
             else:
                 # append the later chunks
                 self.fs.append(self.filename, file_content, self.binary_mode)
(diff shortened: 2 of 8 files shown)


tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,189 @@ def stat(self, filename):
...
+            f.write(compatify(file_content))
+
+    @_translate_errors
+    def glob(self, filename):
+        """Returns a list of files that match the given pattern(s)."""
+        if isinstance(filename, bytes):
+            filename = filename.decode("utf-8")
+        fs, _, _ = fsspec.get_fs_token_paths(
+            filename,
+        )
```
Reubend, line 538: Quick question: are you repeating the logic from `_fs_path` here to avoid running the validation at the end of that function?

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,189 @@ def stat(self, filename):
...
+        files = fs.glob(filename)
+        SEPARATOR = "://"
+
+        prefixed_files = []
+        if SEPARATOR in filename:
+            for file in files:
+                if SEPARATOR not in file:
+                    file = fs.protocol + SEPARATOR + file
+                prefixed_files.append(file)
+        return prefixed_files
```
Reubend, line 548: I'm not certain this will work in all cases... there are chained systems (ex: `dask::s3://bucket/key`) which fsspec supports, local systems where paths don't have a protocol (ex: `C:\Users\reubend\file.txt`) and a few other edge cases.

Have you tested this with those? My initial guess is that putting the protocol and then the separator might break those cases.

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
```diff
@@ -0,0 +1,391 @@
...
+            "waldo",
+        ]
+        expected_listing = [os.path.join(temp_dir, f) for f in expected]
+        gotten_listing = gfile.glob(os.path.join(temp_dir, "*"))
+        self.assertCountEqual(
+            expected_listing,
+            gotten_listing,
+            "Files must match. Expected %r. Got %r."
+            % (expected_listing, gotten_listing),
+        )
```
Reubend, line 56: Rather than checking that the _number_ of entires is the same, can we check that the actual _contents_ of each entry is the same? I want to make sure that each path is well formed.

Reubend (changes requested):

tensorboard/compat/tensorflow_stub/io/gfile.py
d4l3k: yeah, get_fs_token_paths expands the filename globs and returns the files. I should probably change this to just grab the filesystem and not do any of the filepath resolution

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
d4l3k: I copied almost all of these test cases from gfile_test.py

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,189 @@ def stat(self, filename):
...
+            if not binary_mode:
+                raise errors.InvalidArgumentError(
+                    None,
+                    None,
+                    "continue_from not supported with fsspec text mode",
+                )
+            offset = continue_from.get("opaque_offset", None)
+        with fs.open(path, mode, encoding=encoding) as f:
+            if offset is not None:
+                f.seek(offset)
```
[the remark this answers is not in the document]
d4l3k: I do the check for this a few lines up though using seekable is probably the better approach here

tensorboard/compat/tensorflow_stub/io/gfile.py
d4l3k: updated the diff to handle chained file systems and added test cases for them

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
d4l3k: assertCountEqual counts the occurences of the elements and compares them. So it does check the entries are the same but ignores ordering.

d4l3k (commented):
updated after feedback + now supports chained file systems such as `simplecache::zip://foo.txt::s3://bucket/some/path`.

d4l3k:
added file system chaining support to join() -- chained file systems is pretty tricky to get working right

d4l3k:
made filesystem existence checking much cheaper

Reubend (approved):

tensorboard/summary/writer/BUILD
```diff
@@ -56,6 +56,20 @@ py_test(
...
+    main = "event_file_writer_fsspec_test.py",
+    srcs_version = "PY3",
+    tags = ["support_notf"],
+    deps = [
+        ":writer",
+        "//tensorboard:test",
+    ],
+)
+
+
```
nfelt, line 72: Could you remove the extra newline to satisfy buildifier?

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
```diff
@@ -0,0 +1,443 @@
...
+        ckpt_path = os.path.join(temp_dir, "model.ckpt")
+
+        # Write out newlines as given (i.e., \r\n) regardless of OS, so as to
+        # test translation on read.
+        with fsspec.open(ckpt_path, "w", newline="") as f:
+            data = "asdf\nasdf\nasdf\n"
+            f.write(data)
+        with gfile.GFile(ckpt_path, "r") as f:
+            f.buff_chunk_size = 4  # Test buffering by reducing chunk size
+            with self.assertRaises(errors.InvalidArgumentError):
```
nfelt, line 203: Why is it expected that this would raise InvalidArgumentError?

tensorboard/compat/tensorflow_stub/BUILD
```diff
@@ -59,3 +59,15 @@ py_test(
...
+
+py_test(
+    name = "gfile_fsspec_test",
+    size = "small",
+    srcs = ["io/gfile_fsspec_test.py"],
+    srcs_version = "PY3",
+    tags = ["support_notf"],
+    deps = [
+        ":tensorflow_stub",
+        "//tensorboard:test",
```
nfelt, line 71: Could you add a line here with `"//tensorboard:expect_fsspec_installed",` similar to the one for numpy earlier in the file?  And then we'll need to create the corresponding target, which you can base off of this one for pandas:

https://github.com/tensorflow/tensorboard/blob/31e48c8616e75be4fffadb1efa720c9446de0460/tensorboard/BUILD#L484-L489

This looks odd on GitHub since it's just a placeholder, but it allows us to flag that we need the `fsspec` dependency in a way that scales for the internal build.

We'll also need the same dep added to the test in tensorboard/summary/writer/BUILD.

nfelt (changes requested):
Thank you @d4l3k for contributing this and @Ruebend for the review so far!  Apologies for the wait.  I've tested it out a bit with local files and GCS and it looks like it's working great, just a few questions plus some minor stuff to fix the presubmit checks etc.

> So in the future, we should refactor this to allow custom walk implementations, and add a `walk` function for fsspec. Then if/when we replace `gfile` with fsspec entirely, we can just use that walk function instead.

This is a great point (although FWIW, most regular users of TensorBoard who aren't doing anything special should now be going through our data server for the fast load path, which is in Rust and has its own filesystem abstraction separate from these changes).  I've made note of it as part of a larger issue I filed: #5286.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,222 @@ def stat(self, filename):
                 raise
 
 
+class FSSpecFileSystem(object):
+    """Provides filesystem access via fsspec."""
+
+    SEPARATOR = "://"
+    CHAIN_SEPARATOR = "::"
```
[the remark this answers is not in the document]
Reubend: I definitely agree that this complicates the implementation a lot, but I wanted to chime in with a quick note that chaining is often used for wrapping a filesystem within a cache, i.e. `filecache::s3://bucket/key`. That's super useful when using remote filesystems with high latency, since it lets you use the cached copies for repeated reads/writes rather than making additional network requests. 

I wonder if there's some simple way to handle joining and globs that doesn't break with chaining?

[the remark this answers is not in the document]
nfelt: Ok, good to know that's desirable.

Maybe as a compromise it would be possible to restrict the chaining then to only allow the protocol separator in the final component of the chain?  So that would allow `filecache::s3://bucket/key` but not the more complicated cases from https://filesystem-spec.readthedocs.io/en/latest/features.html#url-chaining like `simplecache::zip://*.csv::gcs://bucket/afile.zip`.

That would resolve what is IMO the strangest thing, which is how joining and globbing inserts new path components into the middle of the path rather than appending them.  Effectively then we're just modeling it as though there's still only the single protocol separator, but one which potentially has a "procotol" composed of several filesystems chained together, which seems simpler to reason about.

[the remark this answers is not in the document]
Reubend: +1, I think that's a really good compromise as long as we can throw an exception with a nice message in cases that aren't handled by that.

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
d4l3k: the fsspec implementation here doesn't support seekable streams for text, though digging deeper into it I'm actually not quite sure why here. It uses io.TextIOWrapper which does support them, need to take a closer look at this

tensorboard/compat/tensorflow_stub/io/gfile.py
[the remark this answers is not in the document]
d4l3k: I think having chained file systems might be beneficial down the line but for now it seems unnecessarily complex. Always possible to relax the constraint later. Checking the URLs are valid w/ a nice exception seems doable

tensorboard/compat/tensorflow_stub/BUILD
d4l3k: done

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,222 @@ def stat(self, filename):
                 raise
 
 
+class FSSpecFileSystem(object):
+    """Provides filesystem access via fsspec."""
```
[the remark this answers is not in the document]
d4l3k: done

tensorboard/compat/tensorflow_stub/io/gfile.py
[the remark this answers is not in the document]
d4l3k: added validation that only the last FS can have paths

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,222 @@ def stat(self, filename):
...
+
+    SEPARATOR = "://"
+    CHAIN_SEPARATOR = "::"
+
+    def __init__(self):
+        pass
```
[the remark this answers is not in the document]
d4l3k: done

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,222 @@ def stat(self, filename):
...
+    """
+    if not FSSPEC_ENABLED:
+        return None
+
+    try:
+        segment = filename.split(FSSpecFileSystem.CHAIN_SEPARATOR)[0]
+        protocol = segment.split(FSSpecFileSystem.SEPARATOR)[0]
+        if fsspec.get_filesystem_class(protocol):
+            return _FSSPEC_FILESYSTEM
+    except (ImportError, ValueError):
```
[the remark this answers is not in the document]
d4l3k: yeah, I was looking at https://github.com/intake/filesystem_spec/blob/master/fsspec/registry.py#L208-L213 and trying to match the value semantics of the existing code but I think bubbling up the actual ValueError/ImportError is best here

updated

tensorboard/pip_package/requirements_dev.txt
```diff
@@ -21,6 +21,8 @@ pandas~=1.0
 # For gfile S3 test
 boto3==1.9.86
 moto==1.3.7
+# For gfile fsspec test
+fsspec=2021.7.0
```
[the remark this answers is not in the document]
d4l3k: done

tensorboard/summary/writer/BUILD
d4l3k: done

d4l3k (commented):
updated after feedback

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
d4l3k: https://github.com/intake/filesystem_spec/pull/743 seems to be a bug with fsspec local filesystem in text mode.

d4l3k:
@nfelt can you approve running workflows? want to make sure CI is happy but it says "First-time contributors need a maintainer to approve running workflows."

d4l3k:
buildifier + flake8 should be happy now

nfelt (changes requested):

tensorboard/summary/writer/BUILD
```diff
@@ -56,6 +56,19 @@ py_test(
...
+py_test(
+    name = "event_file_writer_fsspec_test",
+    size = "small",
+    srcs = ["event_file_writer_fsspec_test.py"],
+    main = "event_file_writer_fsspec_test.py",
+    srcs_version = "PY3",
+    tags = ["support_notf"],
+    deps = [
+        ":writer",
+        "//tensorboard:test",
```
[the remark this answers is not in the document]
d4l3k: done

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,245 @@ def stat(self, filename):
...
+        return fs, paths[0]
+
+    @_translate_errors
+    def exists(self, filename):
+        """Determines whether a path exists or not."""
+        fs, path = self._fs_path(filename)
+        return fs.exists(path)
+
+    @_translate_errors
+    def join(self, path, *paths):
```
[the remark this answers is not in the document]
d4l3k: done

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,245 @@ def stat(self, filename):
...
+        self._write(filename, file_content, "ab" if binary_mode else "a")
+
+    def _write(self, filename, file_content, mode):
+        fs, path = self._fs_path(filename)
+        encoding = None if "b" in mode else "utf8"
+        with fs.open(path, mode, encoding=encoding) as f:
+            compatify = compat.as_bytes if "b" in mode else compat.as_text
+            f.write(compatify(file_content))
+
+    def _apply_prefix(self, filename, files):
```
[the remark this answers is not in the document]
d4l3k: done

tensorboard/compat/tensorflow_stub/io/gfile.py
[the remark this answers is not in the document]
d4l3k: updated to handle `simplecache::/tmp/foo`

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
d4l3k: I just removed that check for now. We can add it once that's fixed. Don't want to add a check for the bad behavior only for it to fail on dep update.

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
```diff
@@ -0,0 +1,449 @@
...
+                % (subdir, expected_listing, gotten_listing),
+            )
+
+    def testNonExistentFilesystem(self):
+        with self.assertRaises(ValueError):
+            gfile.get_filesystem("nonexistent::blah://filesystem")
+
+    def testExistance(self):
+        self.assertIsInstance(
+            gfile.get_filesystem("simplecache::nonexistent::s3://blah/blah"),
```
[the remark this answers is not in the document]
d4l3k: done

d4l3k (commented):
updated to simplify path logic + other feedback

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,240 @@ def stat(self, filename):
...
+                raise errors.NotFoundError(None, None, str(e))
+
+        return func_wrapper
+
+    def _fs_path(self, filename):
+        if isinstance(filename, bytes):
+            filename = filename.decode("utf-8")
+        self._validate_path(filename)
+
+        fs, _, paths = fsspec.get_fs_token_paths(
```
nfelt, line 458: It turns out that internally we're on an older version of fsspec (0.7.4) that doesn't have https://github.com/intake/filesystem_spec/commit/8f96ee5b92eeb7511986092553beed1637cd3a1f, which fixes `get_fs_token_paths()` to handle unchaining itself.

Could we instead use `fsspec.core.url_to_fs()` here?  That is present in 0.7.4 onwards and does handle unchaining, and it actually returns both fs and path, so it even simplifies this code to just `return fsspec.core.url_to_fs(filename)`.  With that change I can get the tests to pass internally.

nfelt (changes requested):
Thank you for making the changes!  Everything LGTM except a few last things, and aside from the one mentioned issue I've confirmed that we can land this internally as well.

I'll be away on vacation starting today through the end of next week, but I'll leave some instructions with my teammates re: trying to get this merged assuming it's ready before I'm back.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,240 @@ def stat(self, filename):
...
+        if (
+            self.SEPARATOR not in filename
+            and self.CHAIN_SEPARATOR not in filename
+        ):
+            return files
+
+        return [
+            file
+            if (self.SEPARATOR in file or self.CHAIN_SEPARATOR in file)
+            else self.join(filename, file)
```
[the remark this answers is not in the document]
d4l3k: ahh yeah you're right about that, the underspecified glob output is causing problems. I was wondering about the absolute vs relative case.

I think this is now correct and I added a test case with the in memory file system which returns relative paths for glob similar to gcs

tensorboard/compat/tensorflow_stub/io/gfile.py
d4l3k: done

d4l3k (commented):
addressed feedback + expanded unit tests to cover those cases

I did file an issue on the upstream on the pain points when handling these paths so hopefully we can get a first party maintained solution for this. https://github.com/intake/filesystem_spec/issues/747

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
```diff
@@ -0,0 +1,541 @@
...
+        with self.assertRaises(ValueError):
+            gfile.get_filesystem("nonexistent::blah://filesystem")
+
+    def testExistance(self):
+        self.assertIsInstance(
+            gfile.get_filesystem("simplecache::nonexistent::file://blah/blah"),
+            gfile.FSSpecFileSystem,
+        ),
+
+    def testJoin(self):
```
stephanwlee, line 401: For more completeness, would you mind adding test cases for these?
```
# Trailing slash on the last element
>>> os.path.join("hello", "world/")
'hello/world/'

# empty path at the end
>>> os.path.join("hello", "world", "")
'hello/world/'

# absolute path in the middle
>>> os.path.join("hello", "", "world", "", "/wow")
'/wow'
```

tensorboard/compat/tensorflow_stub/io/gfile_fsspec_test.py
```diff
@@ -0,0 +1,541 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
stephanwlee, line 1: nit: 2018 -> 2021.

tensorboard/compat/tensorflow_stub/io/gfile.py
```diff
@@ -401,6 +411,242 @@ def stat(self, filename):
...
+
+def _get_fsspec_filesystem(filename):
+    """
+    _get_fsspec_filesystem checks if the provided protocol is known to fsspec
+    and if so returns the filesystem wrapper for it.
+    """
+    if not FSSPEC_ENABLED:
+        return None
+
+    segment = filename.split(FSSpecFileSystem.CHAIN_SEPARATOR)[0]
```
stephanwlee, line 642: Perhaps pass a second argument to `split`. e.g., `segment = filename.split(FSSpecFileSystem.CHAIN_SEPARATOR, 1)[0]`. Here and below.

tensorboard/pip_package/requirements_dev.txt
```diff
@@ -21,6 +21,8 @@ pandas~=1.0
 # For gfile S3 test
 boto3==1.9.86
 moto==1.3.7
+# For gfile fsspec test
+fsspec==2021.8.1
```
stephanwlee, line 25: Would you mind changing this to 0.7.4 for now? As Nick indicated, we internally are using older version of fsspec and don't want to inadvertently use a newer APIs.

stephanwlee (commented):
Hello! Nick is now on long vacation so I will be taking over the review and sync. After CI passes, I will take some time to test sync again and merge.

I don't mean to do the review from the top but I have spotted minor things like typos which you may want to address. Also, the failure in CI is real.
```
line 531, in testGlobNonAbsolute
    self.assertCountEqual(
AssertionError: Element counts were not equal:
First has 1, Second has 0:  'memory:///dir/bar.txt'
First has 1, Second has 0:  'memory:///dir/foo.txt'
First has 0, Second has 1:  'memory://dir/foo.txt'
First has 0, Second has 1:  'memory://dir/bar.txt'

```

d4l3k (commented):
test failure with memory paths seems to be a fsspec difference, memory uses absolute paths in 2021.8.1 but not 0.7.4

stephanwlee (approved):

stephanwlee:
Thank you for the contribution!

PR #5250, opened 2021-08-17
time series: show sample on image

stephanwlee:
Image plugin has a notion of sample which was previously not visible to
user. This change shows sample, when it is relevant, in the image card.

image

When step number is large and the card is tiny:
image

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_component.scss
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_component.scss
@@ -16,7 +16,6 @@ limitations under the License.
 @import '../common';
 
 $_title-to-heading-gap: 12px;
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/metrics/views/card_renderer/image_card_component.scss
```diff
@@ -76,29 +85,17 @@ $_second-row-height: 15px;
...
-  }
-}
-
 .run,
+.sample,
 .step {
   @include tb-theme-foreground-prop(color, secondary-text);
   font-size: 13px;
-  height: $_second-row-height;
+  // height: $_second-row-height;
```
japie1235813, line 93: remove it

japie1235813 (approved):

PR #5254, opened 2021-08-18
timeseries: fix memory leaks involving CardObserver and shareReplay

psybuzz:
This addresses 2 non-trivial sources of memory leaks observed when
filtering tags in the Time Series dashboard.

a) CardObserver used to track Elements that were destroyed via
ngOnDestroy and stored them in a `Set<Element>`. Now with a WeakSet
instead, we cover cases when the browser's IntersectionObserverEntry
is not handled for some reason.

b) The rxjs operator "shareReplay" has some known gotchas [1], [2].
When more observables subscribe to a source containing
`shareReplay(1)` in its operator chain, the source remains subscribed
even after the other observables unsubscribe. This change uses the
`takeUntil(ngUnsubscribe$)` pattern to ensure any component using a
`shareReplay` will not keep it subscribed after the component is gone.
Alternatively we could use `refCount`, but this opts for the `takeUntil`
pattern already used widely in the codebase.

Manually checked that the dashboard still operates properly, and in
Angular's dev mode, using a specific demo logdir, adding and clearing a
"." tag filter has these characteristics:

Before this change: 5.8 MB, 5.5K DOM nodes added each cycle
After this change: 0.05 MB, <10 DOM nodes added each cycle

Googlers, see b/196996936

[1] https://github.com/ReactiveX/rxjs/issues/5034
[2] https://github.com/ReactiveX/rxjs/issues/5931

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
@@ -297,6 +297,7 @@ export class RunsTableContainer implements OnInit, OnDestroy {
       })
     );
     this.allUnsortedRunTableItems$ = rawAllUnsortedRunTableItems$.pipe(
+      takeUntil(this.ngUnsubscribe),
       shareReplay(1)
     );
     this.allItemsLength$ = this.allUnsortedRunTableItems$.pipe(
@@ -305,7 +306,7 @@ export class RunsTableContainer implements OnInit, OnDestroy {
 
     const getFilteredItems$ = this.getFilteredItems$(
       this.allUnsortedRunTableItems$
-    ).pipe(shareReplay(1));
+    ).pipe(takeUntil(this.ngUnsubscribe), shareReplay(1));
 
     this.filteredItemsLength$ = getFilteredItems$.pipe(
       map((items) => items.length)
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
```diff
@@ -305,7 +306,7 @@ export class RunsTableContainer implements OnInit, OnDestroy {
 
     const getFilteredItems$ = this.getFilteredItems$(
       this.allUnsortedRunTableItems$
-    ).pipe(shareReplay(1));
+    ).pipe(takeUntil(this.ngUnsubscribe), shareReplay(1));
```
bmd3k, line 309: Do you mind putting a comma after shareReplay(1) here so it formats on multiple lines?

bmd3k (approved):

tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
psybuzz: I just tried, but unfortunately Prettier seems to undo the comma and remove the multiline, I think because `this.getFilteredItems$(this.allUnsortedRunTableItems$)` did not fit onto one line.

psybuzz:
Thanks for the review!

PR #5260, opened 2021-08-20
histogram: pass color instead of scale

stephanwlee:
This change knocks off one of the TODO on the histograms. Instead of
a function that would return a color, this change pipes color to
histograms component instead.


japie1235813 (approved):
LGTM
Just to clarify: is `runColorScale` implemented yet? I am tracing the code but got lost and not sure where it's passed down

stephanwlee:
Yes, it is. CardViewContainer -> CardViewComponent -> HistogramCardContainer ->  HistogramCardComponent -> HistogramV2Component

PR #5261, opened 2021-08-20
histogram: use `transform` to prevent repaint

stephanwlee:
Paint is one of the most expensive operations when you have too many
items painting at the same time. One technique to prevent that is to use
`transform` and `will-change` which often put the DOM in the GPU context
instead. All in all, using `attr.transform` still causes repaint while
`style.transform` does not so we chose to use latter instead.

### DIFF
--- a/tensorboard/webapp/widgets/histogram/histogram_component.ng.html
+++ b/tensorboard/webapp/widgets/histogram/histogram_component.ng.html
@@ -28,7 +28,7 @@ limitations under the License.
       <!-- d3 places axis label at 9px below from the top edge. -->
       <g
         *ngIf="tooltipData"
-        [attr.transform]="getCssTranslate(tooltipData.xAxis.position, 9)"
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/widgets/histogram/histogram_v2_component.scss
```diff
@@ -125,15 +129,19 @@ svg {
     stroke-dasharray: 2;
   }
 
+  $_path-circle-opacity: 0.6;
```
japie1235813, line 132: Why increase the opacity from .5 to .6? Not an objection but just in case this is a typo.

japie1235813 (approved):

tensorboard/webapp/widgets/histogram/histogram_v2_component.scss
stephanwlee: Yes, it was intentional. With some improvement here, it did not look ideal so I had bumped up the number by 0.1.

PR #5263, opened 2021-08-21
timeseries: improve tag filtering

stephanwlee:
In performance profiling, there were two medium sized opportunities for
optimization:
1. we were spending significant CPU cycles for rendering all tag matches
in the autocomplete and it was taking non-trivial amount of time
(~100ms) when there were 3000 tags. This definitely contributed to
slowness when every keystroke is expected to be very responsive.
2. `compareTagNames` were taking a significant amount of time. Given
that the list of tags/cards does not change unless the underlying data
from backend changes, we really do not need to sort on every keystroke
that changes tagFilter. We can pre-sort the tag list before we apply the
filter.

These two changes, combined, gave significantly subjectively snappier
experience.

### DIFF
--- a/tensorboard/webapp/metrics/views/main_view/filter_input_container.ts
+++ b/tensorboard/webapp/metrics/views/main_view/filter_input_container.ts
@@ -71,6 +71,7 @@ export class MetricsFilterInputContainer {
       // De-duplicate using Set since Image cards has a notion of Sample and
       // the same `run` and `tag` can appear more than once.
       map((tags) => [...new Set(tags)]),
+      map((tags) => tags.sort(compareTagNames)),
       combineLatestWith(this.store.select(getMetricsTagFilter)),
       map<[string[], string], [string[], RegExp | null]>(
         ([tags, tagFilter]) => {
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/metrics/views/main_view/filter_input_container.ts
```diff
@@ -84,9 +85,7 @@ export class MetricsFilterInputContainer {
       ),
       filter(([, tagFilterRegex]) => tagFilterRegex !== null),
       map(([tags, tagFilterRegex]) => {
-        return tags
-          .filter((tag: string) => tagFilterRegex!.test(tag))
-          .sort(compareTagNames);
+        return tags.filter((tag: string) => tagFilterRegex!.test(tag));
```
japie1235813, line 88: I can tell that "sorting `compareTagNames` earlier" part is in `filtered_view_container.ts` 
does it also the same case in this `filter_input_container.ts`? what are the relationship between these two components? 
I searched where this `metrics-tag-filter` is applied but not see anywhere is using this 
https://cs.opensource.google/search?q=%22%27metrics-tag-filter%22&sq=&ss=tensorflow%2Ftensorboard

japie1235813 (approved):

tensorboard/webapp/metrics/views/main_view/filter_input_container.ts
stephanwlee: Do not know why it is not picking up but it is used here: https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/views/main_view/main_view_component.ng.html;l=18

Two containers are related but are orthogonal. Input one renders autocomplete based on tag matches while the View one renders cards based on matches. They technically can share a common container parent that passes down list of tag matches but, currently, they are implemented as two separate containers with their own Observable streams both requiring similar optimizations.

PR #5265, opened 2021-08-23
timeseries: debounce tag filter changes

stephanwlee:
This puts debounce on the tag filter changes. Tag view does not have to
change on every keystroke so we put some debounce period so we can be
more performant by not updating too much.

This change also moves the plugin type filtering up since it does not change
too rapidly.

### DIFF
--- a/tensorboard/webapp/metrics/views/main_view/filtered_view_container.ts
+++ b/tensorboard/webapp/metrics/views/main_view/filtered_view_container.ts
@@ -17,6 +17,7 @@ import {createSelector, Store} from '@ngrx/store';
 import {Observable} from 'rxjs';
 import {
   combineLatestWith,
+  debounceTime,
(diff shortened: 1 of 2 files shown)


nfelt (approved):
Just wondering, is the `DeepReadonly` actually needed for this change for some reason? Or is it just a drive-by cleanup? Or a separate perf improvement?

stephanwlee:
> Just wondering, is the DeepReadonly actually needed for this change for some reason? Or is it just a drive-by cleanup? Or a separate perf improvement?

Because TypeScript compiler started to complain about the inferred type of the Observable, we had to strongly type it throughout and `DeepReadonly` is more correct type of the Observable as the Selector returns immutable version of the object. So, in short, yes, cleanup and was kind of warranted since we had to explicitly type the Observable throughout.

PR #5275, opened 2021-08-27
router: do not warn users about dirty updates on the same page

yatbear:
When users change tabs in the same dashboard or edit query parameters in the same list view, unsaved updates should be kept without warnings.

Googlers, see [issue](https://b.corp.google.com/issues/197321281) for more details.

#edit_exp

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -204,14 +204,18 @@ export class AppRoutingEffects {
    */
   navigate$ = createEffect(() => {
     const dispatchNavigating$ = this.validatedRoute$.pipe(
-      mergeMap((routes) => {
-        const samePathname =
-          routes.routeMatch.pathname === this.location.getPath();
+      withLatestFrom(this.store.select(getActiveRoute)),
+      mergeMap(([routes, oldRoute]) => {
+        const sameRouteId =
+          oldRoute !== null &&
+          getRouteId(routes.routeMatch.routeKind, routes.routeMatch.params) ===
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -205,9 +205,13 @@ export class AppRoutingEffects {
   navigate$ = createEffect(() => {
     const dispatchNavigating$ = this.validatedRoute$.pipe(
       mergeMap((routes) => {
+        const samePathname =
+          routes.routeMatch.pathname === this.location.getPath();
```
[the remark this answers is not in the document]
yatbear: Thanks for the informative comment! Changed to check route IDs.

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -334,6 +334,45 @@ describe('app_routing_effects', () => {
...
+        spyOn(window, 'confirm');
+        getPathSpy.and.returnValue('/experiments');
+        // Changing tab.
+        getHashSpy.and.returnValue('#foo');
+        action.next(
+          actions.navigationRequested({
+            pathname: '/experiments',
+          })
+        );
+        tick();
```
stephanwlee, line 347: Looking at this spec alone, it is unclear whether this is supposed to have the same routeId as current one. Can you overrideSelector or setup so that is is more explicit here?

stephanwlee (approved):

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
yatbear: Done.

PR #5276, opened 2021-08-30
infra: upgrade TypeScript to 4.4.2.

stephanwlee:
Internally, they are meaning to upgrade TypeScript and this is the
pre-work for the upgrade.

When we upgraded TypeScript, some type checks changed and this change
adapts to the change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "@types/lodash": "^4.14.172",
     "@types/marked": "^2.0.4",
     "@types/node": "^16.4.13",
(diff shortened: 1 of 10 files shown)


yatbear (approved):
Thanks!!

yatbear (commented):
Does `@bazel/typescript` version need to be updated?

stephanwlee:
> Does @bazel/typescript version need to be updated?

Not yet. That requires newer version of rules_nodejs which we are not doing today.

PR #5283, opened 2021-09-01
color group: omit Experiments if comparison is impossible

stephanwlee:
This change removes "Experiments" option from the color group menu when
there is no comparison route registered in the application. Color group
by experiment is useless when seeing dashboard of one experiment.


japie1235813 (approved):
Thanks for making this change and also re-structure some tests

PR #5288, opened 2021-09-02
util: separate out IsCloudPath from io_wrapper

yatbear:
To be replaced with an internal implementation.

Googlers, see [issue](https://buganizer.corp.google.com/issues/198329196) for more details.

#onduty

### DIFF
--- a/tensorboard/util/io_util.py
+++ b/tensorboard/util/io_util.py
@@ -15,6 +15,6 @@
 """TensorBoard IO helpers."""
 
 
-def IsCloudPath(unused_path):
-    """Not implemented; simply returns False."""
-    return False
+def IsCloudPath(path):
+    """Checks whether a given path is Cloud filesystem path."""
+    return path.startswith("gs://") or path.startswith("s3://")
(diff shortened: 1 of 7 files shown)


tensorboard/util/io_util.py
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""TensorBoard IO helpers."""
+
+
+def IsCloudPath(unused_path):
+    """Not implemented; simply returns False."""
```
nfelt, line 19: Let's preserve the open-source-relevant behavior of the existing helper, which has:

```
    return (
        path.startswith("gs://")
        or path.startswith("s3://")
        or path.startswith("/cns/")
    )
```

So we could drop `/cns/` but I would definitely keep the other two, since that benefits open source usage and we don't want to discard that benefit.

We can also preserve the corresponding tests (not that they are particularly complex tests :))

nfelt (approved):
Thanks for doing this, let's just make sure we preserve the existing support for gs and s3 though?

tensorboard/util/io_util.py
yatbear: Done. Thanks!

PR #5290, opened 2021-09-03
timeseries: adjust wording of empty match warning

nfelt:
Minor tweak since the existing wording for the empty match warning was a bit confusing to me.  In particular it felt hard to scan with the tag count in between the two different filters.  I'm assuming the idea was to distinguish that the tag filter filters "tags" and the plugin type filter instead filters "cards", but practically speaking I think we can treat them as equivalent here, and "tag filtering" is the term used in the section header and in the search field placeholder.

Here's the proposed wording:
 
Screen Shot 2021-09-02 at 5 46 35 PM

### DIFF
--- a/tensorboard/webapp/metrics/views/main_view/empty_tag_match_message_component.ts
+++ b/tensorboard/webapp/metrics/views/main_view/empty_tag_match_message_component.ts
@@ -32,13 +32,12 @@ declare namespace Intl {
 
 @Component({
   selector: 'metrics-empty-tag-match-component',
-  template: `No cards matches tag filter
-    <code>/{{ tagFilterRegex }}/</code> among {{ tagCounts | number }} tags<span
-      *ngIf="pluginTypes.size"
-    >
+  template: `No matches for tag filter <code>/{{ tagFilterRegex }}/</code
+    ><span *ngIf="pluginTypes.size">
       and {{ getPluginTypeFilterString(pluginTypes) }} visualization
       filter</span
-    >.`,
+    >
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/views/main_view/empty_tag_match_message_component.ts
```diff
@@ -32,13 +32,12 @@ declare namespace Intl {
...
-    <code>/{{ tagFilterRegex }}/</code> among {{ tagCounts | number }} tags<span
-      *ngIf="pluginTypes.size"
-    >
+  template: `No matches for tag filter <code>/{{ tagFilterRegex }}/</code
+    ><span *ngIf="pluginTypes.size">
       and {{ getPluginTypeFilterString(pluginTypes) }} visualization
       filter</span
-    >.`,
+    >
+    out of {{ tagCounts | number }} tags.`,
```
stephanwlee, line 40: FYI, the choice to put tagCount before the visualization time was deliberate since viz filter does not really filter on tags.

stephanwlee (approved):

tensorboard/webapp/metrics/views/main_view/empty_tag_match_message_component.ts
nfelt: Yeah, I can see that (mentioned in PR description), but I think as a user the distinction between tags vs cards is not all that helpful, especially since everything else here is referring to filtering "tags", and just having the tag count at the end reads a little better.

PR #5291, opened 2021-09-03
timeseries: second draft of TS promotion

stephanwlee:
This change adds some messages and placeholders for the TimeSeries
promotion. It also adds two buttons users can use to navigate to
scalars, old experience, or dismiss the promotion.

Screenshot from 2021-09-02 22-55-19

### DIFF
--- a/tensorboard/webapp/metrics/views/metrics_container.scss
+++ b/tensorboard/webapp/metrics/views/metrics_container.scss
@@ -25,7 +25,7 @@ limitations under the License.
 .notice {
   background-color: rgba(mat-color($mat-yellow, 200), 0.85);
   border-bottom: 1px solid mat-color($mat-yellow, 500);
-  color: #000;
+  color: map-get($tb-foreground, text);
   display: block;
   flex: 0 0;
 }
--- a/tensorboard/webapp/metrics/views/metrics_container.scss
+++ b/tensorboard/webapp/metrics/views/metrics_container.scss
@@ -16,12 +16,17 @@ limitations under the License.
 
 :host {
   display: flex;
-  height: 100%;
   flex-direction: column;
+  height: 100%;
+  justify-content: stretch;
   overflow: hidden;
 }
 
 .notice {
+  background-color: rgba(mat-color($mat-yellow, 200), 0.85);
+  border-bottom: 1px solid mat-color($mat-yellow, 500);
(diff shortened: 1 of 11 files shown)


tensorboard/webapp/metrics/views/metrics_promo_notice_component.ng.html
```diff
@@ -14,4 +14,14 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-Temporary butter bar content.
+<tb-customization [customizableComponent]="customPromoMessage" class="message">
+  Welcome to new default experience of TensorBoard. Time Series lets you view
+  all visualizations at once, put them side-by-side with pins, and customize
+  colors. Scalars and other plugins are still available.
+  <button class="go-to-scalars" (click)="onGoToScalars.emit()">
+    Go to Scalars plugin</button
```
japie1235813, line 22: Can we make this button looks like more clickable?
```
    text-decoration: underline;
    @include tb-theme-foreground-prop(color, link);
```

tensorboard/webapp/metrics/views/metrics_container.scss
```diff
@@ -16,12 +16,17 @@ limitations under the License.
...
   flex-direction: column;
+  height: 100%;
+  justify-content: stretch;
   overflow: hidden;
 }
 
 .notice {
+  background-color: rgba(mat-color($mat-yellow, 200), 0.85);
+  border-bottom: 1px solid mat-color($mat-yellow, 500);
+  color: #000;
```
japie1235813, line 28: The text color and background contrast is a bit high to me. Can we change it to "rgb(33,33,33)" as other black in dashboard?
perhaps `color: inherit;`

japie1235813 (approved):
Thanks for attaching the screenshot!

tensorboard/webapp/metrics/views/metrics_promo_notice_component.ng.html
stephanwlee: Done.

tensorboard/webapp/metrics/views/metrics_container.scss
stephanwlee: Cannot inherit since I don't want colors to change in the dark mode.

PR #5296, opened 2021-09-07
Use overflow-wrap on location data text.

JamesHollyer:
When the file path to the logdir has a long title without any delimiting characters([/=_,-]) the path is truncated. This line of css stops that truncation by moving the truncated text to a new line.

Googlers, see b/181583629.


nfelt (approved):
LGTM, thanks!  Sometimes it's handy to include before/after screenshots for changes that have a visual effect, but this is pretty minor so totally optional here.

PR #5298, opened 2021-09-07
timeseries: theme group control

stephanwlee:
Pagination controls were improperly themed and this change fixes that.
Now, the secondary-text colors are correctly picked based on current
theme settings and the `<input>` is hard themed (instead of relying on
OS settings and browser's default dark styling). It also styles disabled
button so when the "Previous" or "Next" button is disabled, we have a
visual cue.

## Before
old

## After
Screenshot from 2021-09-07 14-15-02

### DIFF
--- a/tensorboard/webapp/metrics/views/main_view/card_grid_component.scss
+++ b/tensorboard/webapp/metrics/views/main_view/card_grid_component.scss
@@ -52,7 +52,7 @@ card-view {
 }
 
 .group-controls {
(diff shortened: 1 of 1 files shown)


JamesHollyer (approved):
LGTM. I am glad you added a disabled color I think it needed it. Maybe include that addition in the description?

stephanwlee:
Edited the description and the commit message. Thanks for the review.

PR #5299, opened 2021-09-08
timeseries: partially implement linked time on histogram

stephanwlee:
This change implements linked time fob on the histograms visualization.
Unlike the scalar card case where we really want to make a generic line
chart component, this change embeds the linked time feature in the
histogram visualization because, that way, we do not have to contort
ourselves with its complex coordinate system. This makes the code a lot
more manageable and easy to understand.

While implementing the feature, this change refactored the visual
component for the linked time fob so we can reuse its style in other
places too.

Screenshot from 2021-09-07 20-29-42


tensorboard/webapp/widgets/histogram/histogram_component.ng.html
```diff
@@ -34,20 +34,39 @@
...
+        </g>
       </g>
-    </g>
-  </svg>
+    </svg>
+    <!-- Disable the feature when in non-offset and non-step mode. -->
+    <ng-container
+      *ngIf="
+        mode === HistogramMode.OFFSET &&
+        timeProperty=== TimeProperty.STEP &&
```
japie1235813, line 56: nit: a space before `===`

tensorboard/webapp/widgets/histogram/histogram_component.ts
```diff
@@ -21,6 +21,7 @@ import {
   Input,
   OnChanges,
   OnDestroy,
+  TemplateRef,
```
japie1235813, line 24: unused import?

japie1235813 (approved):

PR #5301, opened 2021-09-08
timeseries: Adjust scroll position after pagination

JamesHollyer:
Previously when you changed pages to grouping that had less cards it could cause a jarring and confusing effect as the size of the grouping changed. Many times this would case the graphs you were looking for to go completely off your screen and you had to scroll to see them. 

Now after the new page appears Tensorboard will automatically scroll to the same position relative to the new size of the grouping.

Googlers see b/160888098.

### DIFF
--- a/tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
+++ b/tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
@@ -60,29 +60,25 @@ export class CardGridComponent {
     return isBottomControl;
   }
 
-  PaginationCallback(pageIndex: number, target: HTMLElement) {
+  handlePageChange(pageIndex: number, target: HTMLElement) {
     // Clear call stack to allow dom update before updating scroll to keep
     // relative position.
-    setTimeout(
-      this.scrollToKeepTargetPosition.bind(
-        this,
-        target,
-        target.getBoundingClientRect().top
-      ),
-      0
-    );
+    const topBeforeChange = target.getBoundingClientRect().top;
+    setTimeout(() => {
+      this.scrollToKeepTargetPosition(target, topBeforeChange);
+    }, 0);
 
     this.pageIndexChanged.emit(pageIndex);
   }
 
   scrollToKeepTargetPosition(target: HTMLElement, previousTop: number) {
-    const ScrollingElement = this.cdkScrollable?.getElementRef().nativeElement;
-    if (ScrollingElement) {
-      ScrollingElement.scrollTo(
+    const scrollingElement = this.cdkScrollable?.getElementRef().nativeElement;
+    if (scrollingElement) {
+      scrollingElement.scrollTo(
         0,
         target.getBoundingClientRect().top -
           previousTop +
-          ScrollingElement.scrollTop
+          scrollingElement.scrollTop
       );
     }
   }
@@ -111,6 +107,6 @@ export class CardGridComponent {
       input.value = String(nextValue + 1);
     }
 
-    this.PaginationCallback(nextValue, input);
+    this.handlePageChange(nextValue, input);
   }
 }
--- a/tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
+++ b/tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
@@ -16,12 +16,10 @@ import {CdkScrollable} from '@angular/cdk/scrolling';
 import {
   ChangeDetectionStrategy,
   Component,
-  ElementRef,
   EventEmitter,
   Input,
   Optional,
   Output,
-  ViewChild,
 } from '@angular/core';
 
 import {PluginType} from '../../data_source';
@@ -50,10 +48,6 @@ export class CardGridComponent {
   @Output() pageIndexChanged = new EventEmitter<number>();
   @Output() groupExpansionToggled = new EventEmitter<void>();
 
-  @ViewChild('cardGrid', {static: true}) cardGridElement!: ElementRef<
-    HTMLDivElement
-  >;
-
   constructor(
     @Optional() private readonly cdkScrollable: CdkScrollable | null
   ) {}
@@ -66,19 +60,17 @@ export class CardGridComponent {
     return isBottomControl;
   }
 
-  testCallback(pageIndex: number, target: HTMLElement) {
-    const ScrollingElement = this.cdkScrollable?.getElementRef().nativeElement;
-    if (ScrollingElement) {
-      const distanceToTop =
-        target.getBoundingClientRect().top - ScrollingElement.scrollTop;
-
-      // Clear call stack to allow dom update before updating scroll to keep
-      // relative position.
-      setTimeout(
-        this.scrollToKeepTargetPosition.bind(this, target, distanceToTop),
-        0
-      );
-    }
+  PaginationCallback(pageIndex: number, target: HTMLElement) {
+    // Clear call stack to allow dom update before updating scroll to keep
+    // relative position.
+    setTimeout(
+      this.scrollToKeepTargetPosition.bind(
+        this,
+        target,
+        target.getBoundingClientRect().top
+      ),
+      0
+    );
 
     this.pageIndexChanged.emit(pageIndex);
   }
@@ -88,7 +80,9 @@ export class CardGridComponent {
     if (ScrollingElement) {
       ScrollingElement.scrollTo(
         0,
-        target.getBoundingClientRect().top - previousTop
+        target.getBoundingClientRect().top -
+          previousTop +
+          ScrollingElement.scrollTop
       );
     }
   }
@@ -117,6 +111,6 @@ export class CardGridComponent {
       input.value = String(nextValue + 1);
     }
 
-    this.pageIndexChanged.emit(nextValue);
+    this.PaginationCallback(nextValue, input);
   }
 }
--- a/tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
+++ b/tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
@@ -12,12 +12,16 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+import {CdkScrollable} from '@angular/cdk/scrolling';
 import {
   ChangeDetectionStrategy,
   Component,
+  ElementRef,
   EventEmitter,
   Input,
+  Optional,
   Output,
+  ViewChild,
 } from '@angular/core';
 
 import {PluginType} from '../../data_source';
@@ -46,12 +50,47 @@ export class CardGridComponent {
   @Output() pageIndexChanged = new EventEmitter<number>();
   @Output() groupExpansionToggled = new EventEmitter<void>();
 
-  showExpand(isBottomControl: boolean) {
+  @ViewChild('cardGrid', {static: true}) cardGridElement!: ElementRef<
+    HTMLDivElement
+  >;
+
+  constructor(
+    @Optional() private readonly cdkScrollable: CdkScrollable | null
+  ) {}
+
+  showExpand(isBottomControl: boolean): boolean {
     return isBottomControl ? this.isGroupExpandable : false;
   }
 
   showPaginationInput(isBottomControl: boolean) {
     return isBottomControl;
+  }
+
+  testCallback(pageIndex: number, target: HTMLElement) {
+    const ScrollingElement = this.cdkScrollable?.getElementRef().nativeElement;
+    if (ScrollingElement) {
+      const distanceToTop =
+        target.getBoundingClientRect().top - ScrollingElement.scrollTop;
+
+      // Clear call stack to allow dom update before updating scroll to keep
+      // relative position.
+      setTimeout(
+        this.scrollToKeepTargetPosition.bind(this, target, distanceToTop),
+        0
+      );
+    }
+
+    this.pageIndexChanged.emit(pageIndex);
+  }
+
+  scrollToKeepTargetPosition(target: HTMLElement, previousTop: number) {
+    const ScrollingElement = this.cdkScrollable?.getElementRef().nativeElement;
+    if (ScrollingElement) {
+      ScrollingElement.scrollTo(
+        0,
+        target.getBoundingClientRect().top - previousTop
+      );
+    }
   }
 
   trackByCards(index: number, cardIdWithMetadata: CardIdWithMetadata) {
--- a/tensorboard/webapp/metrics/views/main_view/card_grid_container.ts
+++ b/tensorboard/webapp/metrics/views/main_view/card_grid_container.ts
@@ -12,12 +12,14 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+import {CdkScrollable} from '@angular/cdk/scrolling';
 import {
   ChangeDetectionStrategy,
   Component,
   Input,
   OnChanges,
   OnDestroy,
+  Optional,
   SimpleChanges,
 } from '@angular/core';
 import {Store} from '@ngrx/store';
@@ -149,7 +151,9 @@ export class CardGridContainer implements OnChanges, OnDestroy {
     })
   );
 
-  constructor(private readonly store: Store<State>) {}
+  constructor(
+    private readonly store: Store<State>,
+    @Optional() private readonly cdkScrollable: CdkScrollable | null) {}
 
   ngOnChanges(changes: SimpleChanges) {
     if (changes['cardIdsWithMetadata']) {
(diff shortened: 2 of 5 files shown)


tensorboard/webapp/metrics/views/main_view/card_grid_container.ts
```diff
@@ -168,6 +173,25 @@ export class CardGridContainer implements OnChanges, OnDestroy {
...
+    if (ScrollingElement) {
+      let relativePosition =
+        ScrollingElement.scrollHeight - ScrollingElement.scrollTop;
+      // Clear call stack to allow dom update before updating scroll to keep
+      // relative position.
+      setTimeout(this.scrollToRelativePosition.bind(this, relativePosition), 0);
+    }
+  }
+
+  scrollToRelativePosition(relativePosition: number) {
```
bmd3k, line 187: Is there any possibility to add tests for the new functionality? (and, likewise, a test to ensure it doesn't get applied to presses on the top row of next/previous buttons?)

bmd3k (commented):
Would you mind adding some clarifying text or tag to the PR title saying that this is for Time Series plugin?

nfelt:
> Would you mind adding some clarifying text or tag to the PR title saying that this is for Time Series plugin?

FYI, an easy way to do this is just adding a "scope" prefix to the start of the commit message like `timeseries: adjust scroll position after pagination`.  (Sort of loosely in the style of https://conventionalcommits.org, not that we follow that strictly at all.)  This is particularly useful later when we assemble release notes since it gives a high-level hint as to what part of the codebase/app is being modified; you can see some prior PRs for examples.

tensorboard/webapp/metrics/views/main_view/card_grid_container.ts
```diff
@@ -168,6 +173,25 @@ export class CardGridContainer implements OnChanges, OnDestroy {
 
   onPageIndexChanged(newIndex: number) {
     this.pageIndex$.next(newIndex);
+
+    let ScrollingElement = this.cdkScrollable?.getElementRef().nativeElement;
+    if (ScrollingElement) {
+      let relativePosition =
```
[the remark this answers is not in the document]
JamesHollyer: done.

tensorboard/webapp/metrics/views/main_view/card_grid_container.ts
```diff
@@ -149,7 +151,10 @@ export class CardGridContainer implements OnChanges, OnDestroy {
     })
   );
 
-  constructor(private readonly store: Store<State>) {}
+  constructor(
```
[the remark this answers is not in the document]
JamesHollyer: I rewrote this entire thing. It works for top buttons, bottom buttons, and the pagination input controller.

tensorboard/webapp/metrics/views/main_view/card_grid_container.ts
JamesHollyer: Looking into testing the new functionality.

tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
```diff
@@ -46,14 +48,45 @@ export class CardGridComponent {
...
 
   showPaginationInput(isBottomControl: boolean) {
     return isBottomControl;
   }
 
+  PaginationCallback(pageIndex: number, target: HTMLElement) {
+    // Clear call stack to allow dom update before updating scroll to keep
+    // relative position.
+    setTimeout(
+      this.scrollToKeepTargetPosition.bind(
```
bmd3k, line 67: bind() is a bit of an old fashioned way to bind a function to the correct `this`. arrow functions are our conventional way to do this now. I think this should work:

```
setTimeout(() => {
  this.scrollToKeepTargetPosition(target, target.getBoundingClientRect().top);
}, 0);
```

scrollToKeepTargetPosition(...) will have correct `this` scope when it is called.

tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
```diff
@@ -46,14 +48,45 @@ export class CardGridComponent {
...
+
+  showExpand(isBottomControl: boolean): boolean {
     return isBottomControl ? this.isGroupExpandable : false;
   }
 
   showPaginationInput(isBottomControl: boolean) {
     return isBottomControl;
   }
 
+  PaginationCallback(pageIndex: number, target: HTMLElement) {
```
bmd3k, line 63: I think it would be more conventional to start this with a verb. `setPaginationCallback` would be an improvement. Perhaps even better is handlePageChange() since (1) it also emits `pageIndexChanged` and (2) I don't know what you mean by `Callback`.

bmd3k (changes requested):

tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
JamesHollyer: Well that is exciting that I have been coding long enough to have my ways become old fashioned :P. 

Changed.

tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
JamesHollyer: I like handlePageChange.

tensorboard/webapp/metrics/views/main_view/card_grid_container.ts
JamesHollyer: Tests added. PTAL.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,188 @@
...
+        [groupName]="cardObserver"
+      ></metrics-card-grid>
+      <div class="placeholder">placeholder</div>
+    </div>
+  `,
+  styles: [
+    `
+      div {
+        position: fixed;
+        display: block;
```
stephanwlee, line 45: `div` is already a block element so  you do not need this; here and below.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,188 @@
...
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [NoopAnimationsModule, ScrollingModule],
+      declarations: [
+        CardGridComponent,
+        CardGridContainer,
+        TestableComponent,
+        CardViewComponent,
+        CardViewContainer,
+        CardLazyLoader,
```
stephanwlee, line 95: If these are not important part of the test, omit this and use `NO_ERROR_SCHEMA`: https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/header/header_test.ts;l=80;drc=d9d46c833b20113664a9290b53bacfe3b841894e.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,188 @@
...
+      ],
+      providers: [
+        provideMockStore({
+          initialState: appStateFromMetricsState(),
+        }),
```
stephanwlee, line 100: preferential thing: some test specify the mockStore but, in actually, you don't need this at all. Since we overrideSelector and make sure all states are mocked away (we don't want to deal with real reducer structure; i.e., reducer shape is an implementation detail that views should not be aware of and selectors should be the public API between reducer <--> view. In this world, view test knowing the shape is meh).

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,188 @@
...
+      ],
+    }).compileComponents();
+
+    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
+    store.overrideSelector(selectors.getRunColorMap, {});
+  });
+
+  fit('keeps pagination button position when page size changes', async () => {
```
stephanwlee, line 108: FYI: `fit` here.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,188 @@
...
+      .queryAll(By.css('.prev'))
+      .map((nextDebugElements) => {
+        return nextDebugElements.nativeElement!;
+      });
+    const PaginationInput: HTMLInputElement = fixture.debugElement.query(
+      By.css('input')
+    ).nativeElement;
+    const scrollingElement = fixture.nativeElement.children[0];
+
+    // Test scrolling adjustements on bottom next button.
```
stephanwlee, line 128: typo: adjustements -> adjustments throughout.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,188 @@
...
+        }),
+      ],
+    }).compileComponents();
+
+    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
+    store.overrideSelector(selectors.getRunColorMap, {});
+  });
+
+  fit('keeps pagination button position when page size changes', async () => {
+    store.overrideSelector(settingsSelectors.getPageSize, 2);
```
stephanwlee, line 109: If I were you, I would specify the `cardIdsWithMetadata` here. Currently, it has three elements and because the pageSize is 2, we should expect next page to fix with less content. Reading this spec alone, this fact is not clear and I need to bounce around in this file, which is a bit unfortunate. You will notice that some tests are needlessly long and what not but this often is the key consideration.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,188 @@
...
+    // Test scrolling adjustements on bottom next button.
+    scrollingElement.scrollTo(0, nextButtons[1].offsetTop - scrollOffset);
+    fixture.detectChanges();
+    nextButtons[1].click();
+    fixture.detectChanges();
+    // Clear call stack to invoke the scroll adjustement logic.
+    await timeout(0);
+    expect(nextButtons[1].offsetTop - scrollingElement.scrollTop).toEqual(
+      scrollOffset
+    );
```
stephanwlee, line 137: Great; would you be able to assert and make sure that the content size has changed after clicking on the next button? That is, this test would pass as is if content in the grid never changed.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,188 @@
...
+    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
+    store.overrideSelector(selectors.getRunColorMap, {});
+  });
+
+  fit('keeps pagination button position when page size changes', async () => {
+    store.overrideSelector(settingsSelectors.getPageSize, 2);
+    let scrollOffset = Math.floor(Math.random() * scrollElementHeight);
+    const fixture = TestBed.createComponent(TestableComponent);
+    fixture.detectChanges();
+    const nextButtons = fixture.debugElement
```
stephanwlee, line 113: might be more readable to say `const [topNextButton, bottomNextButton] = fixture.debugElement...`

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: I see that being done in other tests with `store.overrideSelector(selectors.getNonEmptyCardIdsWithMetadata,[...])`.
However, they have elements that use that selector to get the cards. My TestingComponent has to inject the cards into the `metrics-card-grid` directly. 

I injected these after creating the component. But that should help the readability and also allow other tests in the future to have a different number of cards. Let me know what you think.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: I didn't like the idea of asserting something that we actually do not really care much about. However, I see your point that we are not sure the test is working properly if we do not have it.

I added expect calls for this where appropriate.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: Done

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: I was thinking along those lines at one point. All selectors should be overridden so we don't actually need a state object. However, without this I get `NullInjectorError: No provider for Store!` is there a different Provider I can put here to resolve this issue? (feel free to message me offline to discuss.)

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: Brilliant. Done.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: Removed.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: Done.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
stephanwlee: You can provide a mock store but without the initialState.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: Clean. I like it.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: Ah I see what you mean. Works great. I had to override one more selector that I didn't even realize was being called but, all is working now. Thanks!

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: done.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,199 @@
...
+      .placeholder {
+        position: relative;
+        height: 700px;
+      }
+    `,
+  ],
+})
+class TestableComponent {
+  @Input() groupName: string = 'test group name';
+  @Input() cardIdsWithMetadata: CardIdWithMetadata[] = [];
```
bmd3k, line 56: Do groupName and cardObserver actually need to be inputs? I don't see them used in the tests.

bmd3k (approved):
Nice test!

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,199 @@
...
+
+@Component({
+  selector: 'testing-component',
+  template: `
+    <div cdkScrollable>
+      <div class="placeholder">placeholder</div>
+      <metrics-card-grid
+        [cardIdsWithMetadata]="cardIdsWithMetadata"
+        [cardObserver]="cardObserver"
+        [groupName]="cardObserver"
```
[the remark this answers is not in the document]
JamesHollyer: Yep!

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,199 @@
...
+        return nextDebugElements.nativeElement!;
+      });
+    const PaginationInput: HTMLInputElement = fixture.debugElement.query(
+      By.css('input')
+    ).nativeElement;
+    const scrollingElement = fixture.nativeElement.children[0];
+
+    // Test scrolling adjustments on bottom next button.
+    scrollingElement.scrollTo(0, bottomNextButtons.offsetTop - scrollOffset);
+    fixture.detectChanges();
```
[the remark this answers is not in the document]
JamesHollyer: You are correct. Removed.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,199 @@
...
+    }).compileComponents();
+
+    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
+    store.overrideSelector(selectors.getRunColorMap, {});
+    store.overrideSelector(getMetricsTagGroupExpansionState, true);
+  });
+
+  it('keeps pagination button position when page size changes', fakeAsync(() => {
+    store.overrideSelector(settingsSelectors.getPageSize, 2);
+    let scrollOffset = Math.floor(Math.random() * scrollElementHeight);
```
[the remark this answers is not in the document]
JamesHollyer: Yea, I was not sure about doing this. I wanted to simulate the real use case but, I was worried about the principle of having randomness in a test. We don't have fuzzer tests do we?

I made this a fixed value.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,199 @@
...
+import {appStateFromMetricsState} from '../../testing';
+import {State} from '../../../app_state';
+import {selectors as settingsSelectors} from '../../../settings';
+import * as selectors from '../../../selectors';
+import {getMetricsTagGroupExpansionState} from '../../../selectors';
+
+const scrollElementHeight = 100;
+
+@Component({
+  selector: 'testing-component',
```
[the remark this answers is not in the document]
JamesHollyer: I like that. Done.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
```diff
@@ -0,0 +1,199 @@
...
+    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
+    store.overrideSelector(selectors.getRunColorMap, {});
+    store.overrideSelector(getMetricsTagGroupExpansionState, true);
+  });
+
+  it('keeps pagination button position when page size changes', fakeAsync(() => {
+    store.overrideSelector(settingsSelectors.getPageSize, 2);
+    let scrollOffset = Math.floor(Math.random() * scrollElementHeight);
+    const fixture = TestBed.createComponent(TestableComponent);
+    fixture.componentInstance.cardIdsWithMetadata = [
```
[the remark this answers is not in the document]
JamesHollyer: Added.

tensorboard/webapp/metrics/views/main_view/card_grid_test.ts
JamesHollyer: Not needed. Removed.

PR #5302, opened 2021-09-09
linked time: implement histogram color for linked time

stephanwlee:
This change implements fill/color for the linked time feature. When in a
single step mode, if there exists a histogram data with the matching
step, we will highlight it by putting subdued color for non-matching
ones while showing a color for the match. For the range selected time
mode, this change does the same for match/non-match creating a range of
colored histograms.

For single:
Screenshot from 2021-09-08 17-08-05

For range:
Screenshot from 2021-09-08 17-06-53

### DIFF
--- a/tensorboard/webapp/widgets/histogram/histogram_component.ng.html
+++ b/tensorboard/webapp/widgets/histogram/histogram_component.ng.html
@@ -50,14 +50,7 @@ limitations under the License.
       </g>
     </svg>
     <!-- Disable the feature when in non-offset and non-step mode. -->
-    <ng-container
-      *ngIf="
-        mode === HistogramMode.OFFSET &&
-        timeProperty === TimeProperty.STEP &&
-        linkedTime &&
-        scales
-      "
-    >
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/widgets/histogram/histogram_component.ng.html
```diff
@@ -77,11 +70,17 @@
       </g>
     </g>
 
-    <g #histograms class="histograms">
+    <g
+      #histograms
+      [class.histograms]="true"
+      [class.linked-time-enabled]="linkedTime"
```
japie1235813, line 76: is this class name used?

japie1235813 (approved):

tensorboard/webapp/widgets/histogram/histogram_component.ng.html
stephanwlee: It is unused in test or CSS but I thought it would be useful to have this class in general. One use case would be to use it in webdriver test.

PR #5303, opened 2021-09-09
linked time: fix fob positioning on scalar

stephanwlee:
Previously, when using range selection, the fob was incorrectly placed.
This change fixes that.

Before:
Screenshot from 2021-09-08 17-17-59

After:
Screenshot from 2021-09-08 17-18-42


JamesHollyer (approved):
LGTM

Issue #5246, opened 2021-08-15
Error bar color does not change after switching the dark/light mode; Color 


twni2016:
Hi Tensorboard team,

Thank you for the latest update v2.6.0. I love the dark mode feature, but I found it has some issue on the front end:

If I switch the mode from light to dark (or reverse), the background color changes, but the colors of the curves and their error bars (in shadows) does not change. E.g. 

Switching from light to dark:
<img width="251" alt="Screen Shot 2021-08-14 at 8 38 35 PM" src="https://user-images.githubusercontent.com/20842215/129463568-74d7ef2c-c256-449e-89c9-5e98eda41971.png">

Switching from dark to light:
<img width="233" alt="Screen Shot 2021-08-14 at 8 38 57 PM" src="https://user-images.githubusercontent.com/20842215/129463581-3ca61f7f-8c25-4a22-b844-dfd391850991.png">

You can see the error bars remain the previous colors. Now my temporary solutions are to tune the scalar smoothing value to force change. But I hope the color change can be done automatically.

By the way, in the dark mode, the curves of dark colors such as black, cannot be viewed clearly. Maybe a brighter set of colors in dark mode?

## System
```
INFO: installed: tensorboard==2.6.0                                                                                                                                                                       
WARNING: no installation among: ['tensorflow', 'tensorflow-gpu', 'tf-nightly', 'tf-nightly-2.0-preview', 'tf-nightly-gpu', 'tf-nightly-gpu-2.0-preview']                                                  
WARNING: no installation among: ['tensorflow-estimator', 'tensorflow-estimator-2.0-preview', 'tf-estimator-nightly']                                                                                      
INFO: installed: tensorboard-data-server==0.6.1       
```
Ubuntu 20.04.2 LTS
Chrome Version 91.0.4472.114 (Official Build) (x86_64)

stephanwlee:
This is an oversight on our part.

PR #5305, opened 2021-09-09
Changing dark mode makes paint dirty

JamesHollyer:
This is a fix for [issue 5246](https://github.com/tensorflow/tensorboard/issues/5246).

Upon initial change from light to dark mode and dark to light mode the graphs used to look like this:
Screen Shot 2021-09-08 at 9 57 35 PM
Screen Shot 2021-09-08 at 9 57 53 PM

and now they look like this:
Screen Shot 2021-09-08 at 9 54 17 PM
Screen Shot 2021-09-08 at 9 54 10 PM

The problem was that that change from dark to light mode did lot indicate the the SeriesLineView that the lines were "dirty" so upon update it did not realize they needed to be redrawn.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/chart.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/chart.ts
@@ -145,6 +145,7 @@ export class ChartImpl implements Chart {
 
   setUseDarkMode(useDarkMode: boolean) {
     this.renderer.setUseDarkMode(useDarkMode);
+    this.seriesLineView.markAsPaintDirty();
     this.scheduleRepaint();
   }
 


stephanwlee (commented):
Ah, great. I did not think this would literally be one line fix. Would you mind taking a look at the test and meaningfully prevent regression or detect a bug? (A possible answer is, it is hard to test this in a way that can prevent any regressions and require more integrational style testing such as one that uses WebDriver).

Also, pro-tip: if you type "Fixes #<ISSUE_NUMBER>.", GitHub will automatically close the issue when you merge the PR (https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests).

stephanwlee (approved):
In case testing this is more challenging than adding a few LOC, I am approving the change.

PR #5313, opened 2021-09-15
hash: fix issues with the app router.

stephanwlee:
For hash storage, we are using "onhashchange" to listen to the hash
changes but it apparently is quite faulty.

Repro case:
```js
window.onhashchange = () => console.log("hash changed!");
window.location.hash = "yo"
// prints "hash changed!"
window.history.back();
// prints "hash changed!"
window.history.forward();
// prints "hash changed!"

const url = new URL(window.location.toString());
url.search = "random"
window.history.replaceState(null, '', url.toString());

window.history.back();
// prints nothing
window.history.forward();
// prints nothing
```

To fix this, instead of listening to onhashchange, we now listen to
onpopstate which gets triggered in all cases (hash + search + pathname).

### DIFF
--- a/tensorboard/webapp/core/views/hash_storage_component.ts
+++ b/tensorboard/webapp/core/views/hash_storage_component.ts
@@ -54,12 +54,15 @@ export class HashStorageComponent implements OnInit, OnChanges, OnDestroy {
   ).pipe(takeUntil(this.ngUnsubscribe));
 
   ngOnInit() {
-    // Cannot use the tf_storage hash listener because it binds to event before
-    // the zone.js patch. According to [1], zone.js patches various asynchronos
-    // calls and event listeners to detect "changes" and mark components as
-    // dirty for re-render. When using tf_storage hash listener, it causes bad
-    // renders in Angular due to missing dirtiness detection. [1]:
-    // https://blog.angular-university.io/how-does-angular-2-change-detection-really-work/
+    // Note: A couple alternative implementations to using 'popstate' event that
+    // turn out to be buggy:
+    // 1. tf_storage hash listener: It binds to events before zone.js patches
+    //    event listeners for change detection ([1]).
+    // 2. 'hashchange' event: We observed that window.history.back() and
+    //    window.history.forward() do not trigger 'hashchange' events after some
+    //    calls to replaceState which AppRouting uses.
+    //
+    // [1]: https://blog.angular-university.io/how-does-angular-2-change-detection-really-work/
     this.onHashChange.subscribe(() => {
       const activePluginId = this.deepLinker.getPluginId();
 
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/core/views/hash_storage_component.ts
```diff
@@ -45,29 +46,35 @@ export class HashStorageComponent implements OnInit, OnChanges, OnDestroy {
...
 
   ngOnInit() {
-    // Cannot use the tf_storage hash listener because it binds to event before the
-    // zone.js patch. According to [1], zone.js patches various asynchronos calls and
-    // event listeners to detect "changes" and mark components as dirty for re-render.
-    // When using tf_storage hash listener, it causes bad renders in Angular due to
-    // missing dirtiness detection.
-    // [1]: https://blog.angular-university.io/how-does-angular-2-change-detection-really-work/
-    window.addEventListener('hashchange', this.onHashChange);
+    // Cannot use the tf_storage hash listener because it binds to event before
```
bmd3k, line 57: It could be helpful to modify the comment here to briefly mention both alternative implementations that you have tried.

For example:

// Note: A couple alternative implementations to using 'popstate' event that turn out to be buggy:
//   1. tf_storage hash listener: It binds to events before zone.js patches event listeners for change detection ([1])
//   2. 'hashchange' event: We observed that window.history.back() and window.history.forward() do not trigger 'hashchange' events after some calls to replaceState
//
// [1]: https://blog.angular-university.io/how-does-angular-2-change-detection-really-work/

tensorboard/webapp/core/views/hash_storage_test.ts
```diff
@@ -154,13 +152,13 @@ describe('hash storage test', () => {
     expect(setPluginIdSpy).toHaveBeenCalledWith('bar', jasmine.any(Object));
   });
 
-  it('dispatches plugin changed event when hash changes', () => {
+  it('dispatches plugin changed event when popstate (hash) changes', () => {
```
bmd3k, line 155: Is it possible to modify these tests so that you actually change the URL hash, including in the way that we know doesn't work for 'hashchange'?

bmd3k (approved):

tensorboard/webapp/core/views/hash_storage_test.ts
stephanwlee: Considered it. Unfortunately, Karma uses actual browsers and using history API breaks the test runner but also becomes very stateful. I think this would be better tested with integrational screenshot tests.

tensorboard/webapp/core/views/hash_storage_component.ts
stephanwlee: Done.

PR #5314, opened 2021-09-16
timeseries: show indication when there is a new pin

stephanwlee:
The reason why we decided to show pinned section during search is
because we wanted to make pinning more intuitive while you are filter
searching (previously, when you pinned a card, there was no visual
indication of an action besides a small fill on the pin icon which is
occluded by the cursor). This change makes the section a bit smaller but
compensate for the small UI with an indicator for new pinned card.

Screenshot from 2021-09-15 19-31-43

### DIFF
--- a/tensorboard/webapp/metrics/views/main_view/pinned_view_component.ts
+++ b/tensorboard/webapp/metrics/views/main_view/pinned_view_component.ts
@@ -28,6 +28,12 @@ import {CardIdWithMetadata} from '../metrics_view_types';
         <span *ngIf="cardIdsWithMetadata.length > 1" class="group-card-count"
           >{{ cardIdsWithMetadata.length }} cards</span
         >
+        <span
+          *ngFor="let id of newCardPinnedIds"
+          class="new-card-pinned"
+          [data-id]="id"
+          >New card pinned</span
+        >
       </span>
     </div>
     <metrics-card-grid
@@ -45,4 +51,5 @@ import {CardIdWithMetadata} from '../metrics_view_types';
 export class PinnedViewComponent {
   @Input() cardObserver!: CardObserver;
   @Input() cardIdsWithMetadata!: CardIdWithMetadata[];
+  @Input() newCardPinnedIds!: number[];
 }
--- a/tensorboard/webapp/metrics/views/main_view/pinned_view_component.ts
+++ b/tensorboard/webapp/metrics/views/main_view/pinned_view_component.ts
@@ -28,10 +28,7 @@ import {CardIdWithMetadata} from '../metrics_view_types';
         <span *ngIf="cardIdsWithMetadata.length > 1" class="group-card-count"
           >{{ cardIdsWithMetadata.length }} cards</span
         >
-        <span
-          *ngFor="let id of newCardPinnedIds"
-          class="new-card-pinned"
-          [data-id]="id"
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/metrics/views/main_view/pinned_view_container.ts
```diff
@@ -41,4 +42,33 @@ export class PinnedViewContainer {
   readonly cardIdsWithMetadata$: Observable<
     DeepReadonly<CardIdWithMetadata[]>
   > = this.store.select(getPinnedCardsWithMetadata).pipe(startWith([]));
+
+  readonly newCardPinnedIds$: Observable<number[]> = this.store
```
JamesHollyer, line 46: This name confused me. AFAICT it does not describe what is actually resulting from the code below. Maybe something like TimestampOfLatestPin or TimeOfLatestPinning?

JamesHollyer (commented):
This seems needlessly abstract complicated. Would it be possible to check the length of the array returned from getPinnedCardsWithMetadata and if it just increased then restart the animation with *ngIf?

Also, the screenshot added does not reflect what I am seeing. The screenshot shows the animation at the bottom of the pinned section but, it is actually at the top.

tensorboard/webapp/metrics/views/main_view/pinned_view_container.ts
stephanwlee: Left a little comment on what the field is supposed to be. It is supposed to be an opaque id.

stephanwlee:
> This seems needlessly abstract complicated. Would it be possible to check the length of the array returned from getPinnedCardsWithMetadata and if it just increased then restart the animation with *ngIf?

I don't think so. Two things:
- First, AFAIK, there is no nice way to trigger CSS animation so we are unmounting and mounting a new DOM that would trigger the animation. Think of `newCardPinnedIds$` as returning something that would be used in `key` prop in React.
- I don't think we can rely only on length. TensorBoard does allow user to use browser back button to go back far. If user goes from `pin=A,B,C` to `A,D,E`, the length remains but two new cards are pinned.

> Also, the screenshot added does not reflect what I am seeing. The screenshot shows the animation at the bottom of the pinned section but, it is actually at the top.

That's because the screenshot was taken in the tag filter mode and not the default view.

tensorboard/webapp/metrics/views/main_view/pinned_view_component.ts
```diff
@@ -28,6 +28,12 @@ import {CardIdWithMetadata} from '../metrics_view_types';
         <span *ngIf="cardIdsWithMetadata.length > 1" class="group-card-count"
           >{{ cardIdsWithMetadata.length }} cards</span
         >
+        <span
+          *ngFor="let id of newCardPinnedIds"
+          class="new-card-pinned"
+          [data-id]="id"
```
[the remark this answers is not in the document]
stephanwlee: Actually, it is used in the test: 
```
   expect(card2IndicatorBefore.properties['data-id']).not.toBe(
          card2IndicatorAfter.properties['data-id']
        );
```

tensorboard/webapp/metrics/views/main_view/pinned_view_container.ts
JamesHollyer: Ok thanks for the comment. That helps.

JamesHollyer (approved):
Fair enough. LGTM

PR #5315, opened 2021-09-16
Update migrate.ipynb

jvishnuvardhan:
Updated `2.0` to `2.x` for better understanding.

### DIFF
--- a/docs/migrate.ipynb
+++ b/docs/migrate.ipynb
@@ -37,7 +37,7 @@
         "id": "HEYuO5NFwDK9"
       },
       "source": [
-        "# Migrating tf.summary usage to TF 2.0\n",
+        "# Migrating tf.summary usage to TF 2.X\n",
         "\n",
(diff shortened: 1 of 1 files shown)


docs/migrate.ipynb
```diff
@@ -37,7 +37,7 @@
         "id": "HEYuO5NFwDK9"
       },
       "source": [
-        "# Migrating tf.summary usage to TF 2.0\n",
+        "# Migrating tf.summary usage to TF 2.X\n",
```
stephanwlee, line 40: Could you make this consistent with lower "x" instead? It is written as `2.x` elsewhere.

stephanwlee (changes requested):

docs/migrate.ipynb
jvishnuvardhan: Updated based o your suggestion. Thanks

stephanwlee (approved):
Thank you!

PR #5316, opened 2021-09-16
histogram: change histogram format for singular data case

yatbear:
When input data contains only one unique value, generate user specified number of buckets (let `N` denote this number), where the first `N-1` buckets are closed-open (`[v, v)`) with zero counts, and the last bucket is closed (`[v, v]`) with a nonzero count.

This is a POC.

Below are the visualization of the histograms for an empty distribution and a distribution of a singular value:
- Before: image
- After: 
image

#histogram #tpu

### DIFF
--- a/tensorboard/plugins/histogram/summary_test.py
+++ b/tensorboard/plugins/histogram/summary_test.py
@@ -70,16 +70,6 @@ class SummaryBaseTest(object):
         buckets = tensor_util.make_ndarray(pb.value[0].tensor)
         np.testing.assert_allclose(buckets, np.array([]).reshape((0, 3)))
 
-    def test_singleton_input(self):
-        pb = self.histogram("twelve", [12])
-        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
-        np.testing.assert_allclose(buckets, np.array([[11.5, 12.5, 1]]))
-
-    def test_input_with_all_same_values(self):
-        pb = self.histogram("twelven", [12, 12, 12])
-        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
-        np.testing.assert_allclose(buckets, np.array([[11.5, 12.5, 3]]))
-
     def test_fixed_input(self):
         pass  # TODO: test a small fixed input
 
@@ -135,6 +125,16 @@ class SummaryV1PbTest(SummaryBaseTest, tf.test.TestCase):
             "a/b/histogram_summary", self.histogram("a/b", []).value[0].tag
         )
 
+    def test_singleton_input(self):
+        pb = self.histogram("twelve", [12])
+        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
+        np.testing.assert_allclose(buckets, np.array([[11.5, 12.5, 1]]))
+
+    def test_input_with_all_same_values(self):
+        pb = self.histogram("twelven", [12, 12, 12])
+        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
+        np.testing.assert_allclose(buckets, np.array([[11.5, 12.5, 3]]))
+
 
 class SummaryV1OpTest(SummaryBaseTest, tf.test.TestCase):
     def histogram(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


tensorboard/plugins/histogram/summary_test.py
```diff
@@ -70,16 +70,6 @@ def test_empty_input_of_high_rank(self):
         buckets = tensor_util.make_ndarray(pb.value[0].tensor)
         np.testing.assert_allclose(buckets, np.array([]).reshape((0, 3)))
 
-    def test_singleton_input(self):
```
nfelt, line 73: Ok for a proof of concept, but FWIW, if we remove these test cases from the base test case, we should ensure they get pushed down into all the various test cases - right now this PR only moves them into SummaryV1PbTest and SummaryV2PbTest but that leaves the 3 Op tests uncovered.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -196,53 +196,50 @@ def _buckets(data, bucket_count=None):
         tf.debugging.assert_type(bucket_count, tf.int32)
         data = tf.reshape(data, shape=[-1])  # flatten
         data = tf.cast(data, tf.float64)
-        is_empty = tf.equal(tf.size(input=data), 0)
-
-        def when_empty():
+        data_size = tf.size(input=data)
+        if tf.equal(data_size, 0):
```
nfelt, line 200: So one issue here is that we can't really use normal Python control flow (it works in some contexts, like eager mode and like a tf.function with autograph enabled, but these functions also need to work in tf.functions without autograph, and in legacy graph mode - some context at https://www.tensorflow.org/guide/function#autograph_transformations).

Instead, we can use tf.cond() like the original code had in order to achieve the effect of an if statement.  Same goes for the indices below.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -196,53 +196,50 @@ def _buckets(data, bucket_count=None):
...
-                    a=tf.stack([bucket_starts, bucket_ends, bucket_counts])
-                )
+        min_ = tf.reduce_min(input_tensor=data)
+        max_ = tf.reduce_max(input_tensor=data)
+        range_ = max_ - min_
+        is_singular = tf.equal(range_, 0)
+
+        bucket_width = range_ / tf.cast(bucket_count, tf.float64)
+        offsets = data - min_
+        # For singular data, indices should simply be
```
nfelt, line 210: This is certainly true, but if we're going to use a tf.cond() anyway, we can probably push this a bit further downstream to avoid the overhead of generating all the indices only to reduce_sum() everything later.  I.e. we can just directly construct something like `single_value_bucket_counts` as `bucket_count - 1` zeros followed by a single count equal to `data_size`, and use that instead of doing the computation of the bucket counts.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -196,53 +196,50 @@ def _buckets(data, bucket_count=None):
...
+        non_singular_bucket_indices = tf.cast(
+            tf.floor(offsets / bucket_width), dtype=tf.int32
+        )
+        bucket_indices = (
+            singular_bucket_indices
+            if is_singular
+            else non_singular_bucket_indices
+        )
+        clamped_indices = tf.minimum(bucket_indices, bucket_count - 1)
+        # Cast to float64 to prevent overflow beyond limt 2^24 in reduce_sum below.
```
nfelt, line 223: On further thought (which I should have suggested earlier), I think it's probably best if we break this fix out separately and do it first, since it's a small fix to the existing code, rather than mingling it with the v3 format changes.  Do you mind taking that on?  It'd involve just the approximately-one-line change here, plus adding a test to confirm it stays fixed.

nfelt (commented):
Nice, thanks for looking into this!

One high-level thing I'd suggest is that we could actually create a new function in summary_v2.py that's called `histogram_v3()`, and then we can stage these changes in a way that affects that function only, which would make it easier to get them landed in smaller PRs without worrying about it being a breaking change yet for existing `tf.summary.histogram()` users (since there are some other places in TensorBoard that we probably should update before releasing the format change itself, as the design doc describes).

My suggestion would be to start out with a separate change in which we create `histogram_v3()` as basically a copy of `histogram()` but without all the special handling for name scopes and TPU outside compilation.  Essentially `histogram_v3` would look like rolling back `histogram` to the simple version we had before all the TPU issues: https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/plugins/histogram/summary_v2.py;l=69-79;drc=fdeeb395deaa953d2e5e6375c418a84feb05800b

I'd then re-apply this one later change to use LazyTensorCreator, since that's useful even aside from the TPU issues: https://github.com/tensorflow/tensorboard/pull/2899

I'd also create a copy of the bucketing helper function, `_buckets_v3()` which would be used by `histogram_v3()`.

Once we have that set up, we can easily land changes like this that affect only `_buckets_v3()` and have a separate set of tests, etc.  Once it's working and we've vetted it on TPU, we can cut over the public API symbol from the v2 to the v3 function.  How does that sound?

PR #5317, opened 2021-09-17
line chart: skip axis label render based on visibility

stephanwlee:
This change improves line chart axis by measuring how big the label is
so a label does not overlap with another.

Instead of using DOM to measure the dimension, this uses 2D Canvas to
measure the text which is a lot performant as it would not have to cause
reflow.

Do note that we can be smarter with the tick filtering but this
iteration is quite dumb and only filters from the start.

Screenshot from 2021-09-16 17-34-21

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_utils.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_utils.ts
@@ -168,3 +168,48 @@ export function getTicksForLinearScale(
 
   return {major: Array.from(majorTickValMap.values()), minor};
 }
+
+const canvasForMeasure = document.createElement('canvas').getContext('2d');
+
+export function filterTicksByVisibility(
+  minorTicks: MinorTick[],
+  getDomPos: (tick: MinorTick) => number,
+  axis: 'x' | 'y',
+  axisFont: string,
+  marginBetweenAxis = 5
+): MinorTick[] {
+  if (!minorTicks.length || !canvasForMeasure) return minorTicks;
+  // While tick is in data coordinate system, DOM is on the opposite system;
+  // while pixels go from top=0 to down, data goes from bottom=0 to up.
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_utils.ts
```diff
@@ -168,3 +168,48 @@ export function getTicksForLinearScale(
 
   return {major: Array.from(majorTickValMap.values()), minor};
 }
+
+const canvasForMeasure = document.createElement('canvas').getContext('2d');
+
+export function filterTicksByVisibility(
```
japie1235813, line 174: Could you add docstring to explain this function along with the parameters? 
I do not fully understand them until read the tests. (especially `getDomPos` and `marginBetweenAxis`)

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
```diff
@@ -26,20 +26,25 @@ import {
...
+const AXIS_FONT = '11px Roboto, sans-serif';
+
 @Component({
   selector: 'line-chart-axis',
   templateUrl: 'line_chart_axis_view.ng.html',
   styleUrls: ['line_chart_axis_view.css'],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class LineChartAxisComponent {
+  readonly axisFont = AXIS_FONT;
```
japie1235813, line 46: is this variable used anywhere? If you are not going to change it, line 107 is good enough?

japie1235813 (approved):

tensorboard/webapp/widgets/line_chart_v2/sub_view/line_chart_axis_view.ts
stephanwlee: Thanks, it was a mistake.

PR #5323, opened 2021-09-20
cleanup: log as a warning instead of an error when HParams-plugin experiment data is not found

yatbear:
Reduce non-essential errors in the logs for easier debugging.

Googlers, see [issue](https://b.corp.google.com/issues/199306961) for more context.

#onduty

### DIFF
--- a/tensorboard/plugins/hparams/get_experiment.py
+++ b/tensorboard/plugins/hparams/get_experiment.py
@@ -15,7 +15,7 @@
 """Classes and functions for handling the GetExperiment API call."""
 
 
-from tensorboard.plugins.hparams import error
+import logging
 
 
 class Handler(object):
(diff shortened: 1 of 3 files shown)


stephanwlee (commented):
Great but I believe you need to make some change here: https://github.com/tensorflow/tensorboard/blob/06d5a5bd9afb63db82c57a55ddf4e58eb0dcd207/tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.ts#L530-L544

We are kind of relying on the promise failing in few frontend components.

yatbear:
> Great but I believe you need to make some change here:
> 
> https://github.com/tensorflow/tensorboard/blob/06d5a5bd9afb63db82c57a55ddf4e58eb0dcd207/tensorboard/plugins/hparams/tf_hparams_query_pane/tf-hparams-query-pane.ts#L530-L544
> 
> We are kind of relying on the promise failing in few frontend components.

Thanks! Change applied.

stephanwlee (approved):

PR #5325, opened 2021-09-21
timeseries: keep all digits for step values in scalar cards

yatbear:
To make the step values more meaningful, e.g. display `1,001,001` instead of just `1M`.

Visualization:

Screen Shot 2021-09-22 at 2 12 29 PM

Googlers, see [issue](https://b.corp.google.com/issues/200663497).

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -29,6 +29,7 @@ import {
   Formatter,
   numberFormatter,
   relativeTimeFormatter,
+  intlNumberFormatter,
   siNumberFormatter,
 } from '../../../widgets/line_chart_v2/lib/formatter';
 import {LineChartComponent} from '../../../widgets/line_chart_v2/line_chart_component';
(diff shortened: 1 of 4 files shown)


stephanwlee (commented):
This definitely is not the scalable way so I think it would be much better to display their actual value.

Since TensorBoard is used in a scientific setting, even when very long, I think it is much nicer to show their actual value in full length especially in a tooltip. 

As such, we want to use something like Intl.NumberFormat.
```js
> new Intl.NumberFormat().format(10010000)
< // At the very least, ","s in between number lets you read the number much easily.
< '10,010,000'
```

yatbear:
> This definitely is not the scalable way so I think it would be much better to display their actual value.
> 
> Since TensorBoard is used in a scientific setting, even when very long, I think it is much nicer to show their actual value in full length especially in a tooltip.
> 
> As such, we want to use something like Intl.NumberFormat.
> 
> ```js
> > new Intl.NumberFormat().format(10010000)
> < // At the very least, ","s in between number lets you read the number much easily.
> < '10,010,000'
> ```

Done. Thanks!

stephanwlee (approved):

PR #5331, opened 2021-09-23
cleanup: raise an exception when no experiment is returned

yatbear:
This fixes the failures introduced in https://github.com/tensorflow/tensorboard/pull/5323.

#onduty

### DIFF
--- a/tensorboard/plugins/hparams/hparams_plugin.py
+++ b/tensorboard/plugins/hparams/hparams_plugin.py
@@ -96,8 +96,8 @@ class HParamsPlugin(base_plugin.TBPlugin):
                 ctx, self._context, experiment_id
             ).run()
             if not experiment:
-                return http_util.Respond(
-                    request, "Experiment not found", "text/plain", 204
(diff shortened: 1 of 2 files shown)


tensorboard/plugins/hparams/hparams_plugin.py
```diff
@@ -96,8 +96,8 @@ def download_data_route(self, request):
                 ctx, self._context, experiment_id
             ).run()
             if not experiment:
-                return http_util.Respond(
-                    request, "Experiment not found", "text/plain", 204
+                raise werkzeug.exceptions.BadRequest(
```
stephanwlee, line 99: Please use TensorBoard errors instead. https://github.com/tensorflow/tensorboard/blob/0411fc1feebb9224a94c94fd5b8f4fa3fc5d3388/tensorboard/errors.py#L62

stephanwlee (commented):
It is correct that we are raising errors again and we will basically go back to pre-#5323? If so, would it be correct to undo the JavaScript changes?

PR #5337, opened 2021-09-24
histogram: cast tf.reduce_sum input to float64

yatbear:
This is to avoid the accumulating floating point errors in `tf.reduce_sum` when using `float32` where large sums are truncated to `2^24` as described in https://github.com/tensorflow/tensorflow/issues/36128.

#histogram #tpu

### DIFF
--- a/tensorboard/plugins/histogram/summary_test.py
+++ b/tensorboard/plugins/histogram/summary_test.py
@@ -120,14 +120,14 @@ class SummaryBaseTest(object):
         self.assertEqual(buckets.shape, (bucket_count, 3))
 
     def test_with_large_counts(self):
-        # Check for overflow with large count (2^24) of data.
-        large_count = 2 ** 24
+        # Check for accumulating floating point errors with large counts (> 2^24).
+        # See https://github.com/tensorflow/tensorflow/issues/51419 for details.
+        large_count = 20_000_000
         data = [0] + [1] * large_count
         pb = self.histogram("large_count", data=data, buckets=2)
         buckets = tensor_util.make_ndarray(pb.value[0].tensor)
-        np.testing.assert_allclose(
-            buckets, np.array([[0, 0.5, 1], [0.5, 1, large_count]])
-        )
+        self.assertEqual(buckets[0][2], 1)
+        self.assertEqual(buckets[1][2], large_count)
 
 
 class SummaryV1PbTest(SummaryBaseTest, tf.test.TestCase):
--- a/tensorboard/plugins/histogram/summary_test.py
+++ b/tensorboard/plugins/histogram/summary_test.py
@@ -119,6 +119,17 @@ class SummaryBaseTest(object):
         buckets = tensor_util.make_ndarray(pb.value[0].tensor)
         self.assertEqual(buckets.shape, (bucket_count, 3))
 
+    def test_with_large_counts(self):
+        # Check for overflow with large count (2^24) of data in float16.
+        large_count = 2 ** 24
+        data = [0] + [1] * large_count
+
+        pb = self.histogram("large_count", data=data, buckets=2)
+        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
+        np.testing.assert_allclose(
+            buckets, np.array([[0, 0.5, 1], [0.5, 1, large_count]])
+        )
+
 
 class SummaryV1PbTest(SummaryBaseTest, tf.test.TestCase):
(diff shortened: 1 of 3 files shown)


tensorboard/plugins/histogram/summary_test.py
```diff
@@ -119,6 +119,16 @@ def test_when_bucket_count_not_statically_known(self):
         buckets = tensor_util.make_ndarray(pb.value[0].tensor)
         self.assertEqual(buckets.shape, (bucket_count, 3))
 
+    def test_with_large_counts(self):
+        # Check for overflow with large count (2^24) of data.
+        large_count = 2 ** 24
```
nfelt, line 124: It's not actually enough to test 2^24, since that's the highest number that still works - we need to test a number greater than 2^24 (e.g. 2^24 + 1) in order to get the failure case.

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -119,6 +119,16 @@ def test_when_bucket_count_not_statically_known(self):
...
+        data = [0] + [1] * large_count
+        pb = self.histogram("large_count", data=data, buckets=2)
+        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
+        np.testing.assert_allclose(
```
nfelt, line 128: We shouldn't be using allclose() here, because that allows some error in the values - specifically, it allows a relative difference of `rtol=1e-07` by default.  And the difference we want to measure could be as small as between 16777216 and 16777217, which is an error of only 1/16777216 = 5.96e-8, so it's small enough that `allclose()` says this is close enough to count as equal.

I'd recommend just doing `self.assertEqual(buckets[1][2], large_count)`, since we expect this result to be exactly correct.

nfelt (changes requested):
> For the particular test case added where the total count of `1` is `2^24`, without converting to `float64`, the incorrect result would be `2048` instead of `16777216`.

I'm not sure I follow this, can you explain a bit more?  The failure mode that I'd expect to see in this test is that we would get `16777216` as the incorrect result, when trying to sum more than that number of `1.0` values (basically, it's as though the count is truncated to `16777216`).

Right now, when I remove the changes to summary_v2.py, the added test still passes.  I think it's due to a couple things that should be fixed in the test (see other comments).

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -214,7 +214,11 @@ def when_nonsingular():
                     tf.floor(offsets / bucket_width), dtype=tf.int32
                 )
                 clamped_indices = tf.minimum(bucket_indices, bucket_count - 1)
-                one_hots = tf.one_hot(clamped_indices, depth=bucket_count)
+                # Cast to float64 to prevent overflow beyond limt 2^24 in reduce_sum below.
+                one_hots = tf.cast(
+                    tf.one_hot(clamped_indices, depth=bucket_count),
```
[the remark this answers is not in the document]

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -214,7 +214,11 @@ def when_nonsingular():
                     tf.floor(offsets / bucket_width), dtype=tf.int32
                 )
                 clamped_indices = tf.minimum(bucket_indices, bucket_count - 1)
-                one_hots = tf.one_hot(clamped_indices, depth=bucket_count)
+                # Cast to float64 to prevent overflow beyond limt 2^24 in reduce_sum below.
```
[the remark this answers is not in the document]
yatbear: Done. Thanks!

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -119,6 +119,16 @@ def test_when_bucket_count_not_statically_known(self):
         buckets = tensor_util.make_ndarray(pb.value[0].tensor)
         self.assertEqual(buckets.shape, (bucket_count, 3))
 
+    def test_with_large_counts(self):
+        # Check for overflow with large count (2^24) of data.
```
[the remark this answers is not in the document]
yatbear: Done.

tensorboard/plugins/histogram/summary_test.py
yatbear: Done.

tensorboard/plugins/histogram/summary_test.py
yatbear: Done.

nfelt (approved):
Looks great, thanks for fixing this!

PR #5341, opened 2021-09-27
cleanup: log as info when no hparams experiment data is found

yatbear:
Reduce non-essential errors in the logs for easier debugging. 

`"HParams error: Can't find an HParams-plugin experiment data in the log directory. Note that it takes some time to scan the log directory; if you just started Tensorboard it could be that we haven't finished scanning it yet. Consider trying again in a few seconds."` should be logged at `INFO` level.

Tested sync locally. 

Googlers, see [issue](https://b.corp.google.com/issues/199306961) for more context.

#onduty

### DIFF
--- a/tensorboard/plugins/hparams/BUILD
+++ b/tensorboard/plugins/hparams/BUILD
@@ -38,6 +38,7 @@ py_library(
         ":error",
         ":metadata",
         ":protos_all_py_pb2",
+        "//tensorboard:errors",
         "//tensorboard:plugin_util",
         "//tensorboard/backend:http_util",
         "//tensorboard/data:provider",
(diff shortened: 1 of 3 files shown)


stephanwlee:
Are Werkzeug errors emitted as `logging.info` and thus this change is correct? 

If possible, I really do think we really should be using tensorboard.errors (https://github.com/tensorflow/tensorboard/blob/master/tensorboard/errors.py#L62). I would like to know how our log ingestor handle werkzeug errors and do not contribute to the log spam.

yatbear:
> Are Werkzeug errors emitted as `logging.info` and thus this change is correct?
> 
> If possible, I really do think we really should be using tensorboard.errors (https://github.com/tensorflow/tensorboard/blob/master/tensorboard/errors.py#L62). I would like to know how our log ingestor handle werkzeug errors and do not contribute to the log spam.
> 
> While preventing the log spam is good, ideally, we don't want to raise exceptions on such empty responses.

Changed to raise `tensorboard.errors.NotFoundError` but log as info, and still raise werkzeug BadRequest (returns 400).

stephanwlee (approved):

PR #5344, opened 2021-09-27
run selector: subdue violent exception when hparams are invalid

stephanwlee:
When hparams specs in comparison mode are not compatible (e.g., an
experiment has Discrete one while for the same name another experiment
has Interval one), we currently throw a RangeError. That error manifests
as very violent error that breaks entire app and make the TimeSeries
plugin not functional. Since TimeSeries does not really render hparams,
it does not make sense for this error to so violently throw error. Now,
with this change, we do not even read the hparams value when in
TimeSeries.

Also, as a bonus, we have relaxed the constraint a bit and try to
reconcile with incompatible filters if any of the compatible ones are
empty.

### DIFF
--- a/tensorboard/webapp/hparams/_redux/utils_test.ts
+++ b/tensorboard/webapp/hparams/_redux/utils_test.ts
@@ -232,7 +232,7 @@ describe('hparams/_redux/utils test', () => {
       ).toThrow();
     });
 
-    it('does not throw there exists a conflicting hparam that only has empty values', () => {
+    it('does not throw if there exists a conflicting hparam that only has empty values', () => {
       expect(() =>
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/hparams/_redux/utils_test.ts
```diff
@@ -231,6 +231,55 @@ describe('hparams/_redux/utils test', () => {
         ])
       ).toThrow();
     });
+
+    it('does not throw there exists a conflicting hparam that only has empty values', () => {
```
bmd3k, line 235: 'does not throw <b>if</b> there..."

bmd3k (approved):

PR #5348, opened 2021-09-28
histogram: add back the test for large counts

yatbear:
This reverts commit 9529cf7dce77c5928ee1f80e841e3703ba3fe70d.

Also increase the summary_test size from `small` to `medium` to get more RAM. 

Tested the sync locally.

#histogram #tpu

### DIFF
--- a/tensorboard/plugins/histogram/BUILD
+++ b/tensorboard/plugins/histogram/BUILD
@@ -112,6 +112,7 @@ py_library(
 
 py_test(
     name = "summary_test",
+    # Ensure enough RAM for test_with_large_counts.
     size = "medium",
     srcs = ["summary_test.py"],
     main = "summary_test.py",
--- a/tensorboard/plugins/histogram/BUILD
+++ b/tensorboard/plugins/histogram/BUILD
@@ -112,7 +112,7 @@ py_library(
 
 py_test(
     name = "summary_test",
-    size = "small",
+    size = "medium",
     srcs = ["summary_test.py"],
     main = "summary_test.py",
     srcs_version = "PY3",
--- a/tensorboard/plugins/histogram/summary_test.py
+++ b/tensorboard/plugins/histogram/summary_test.py
@@ -122,10 +122,12 @@ class SummaryBaseTest(object):
     def test_with_large_counts(self):
         # Check for accumulating floating point errors with large counts (> 2^24).
         # See https://github.com/tensorflow/tensorflow/issues/51419 for details.
-        # TODO(ytjing): add a test for counts > 2^24.
-        self.skipTest(
-            "TODO(#36128): figure out how to test without segmentation fault in sync"
-        )
+        large_count = 20_000_000
+        data = [0] + [1] * large_count
+        pb = self.histogram("large_count", data=data, buckets=2)
+        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
+        self.assertEqual(buckets[0][2], 1)
+        self.assertEqual(buckets[1][2], large_count)
 
 
 class SummaryV1PbTest(SummaryBaseTest, tf.test.TestCase):


tensorboard/plugins/histogram/summary_test.py
```diff
@@ -122,10 +122,12 @@ def test_when_bucket_count_not_statically_known(self):
     def test_with_large_counts(self):
         # Check for accumulating floating point errors with large counts (> 2^24).
         # See https://github.com/tensorflow/tensorflow/issues/51419 for details.
-        # TODO(ytjing): add a test for counts > 2^24.
-        self.skipTest(
```
nfelt, line 126: FYI but the reason I suggested skipTest() is that it skips the rest of the test body whether it exists or not, so more typically, rather than removing the body of the test at the same time, you can just add the one line and then remove it later.  That way the blame/history for the other lines isn't affected by the temporary disabling.

tensorboard/plugins/histogram/BUILD
```diff
@@ -112,7 +112,7 @@ py_library(
 
 py_test(
     name = "summary_test",
-    size = "small",
+    size = "medium",
```
nfelt, line 115: Can you add a note about why this needs to be medium?  E.g. `# Ensure enough RAM for test_with_large_counts`

nfelt (approved):
LGTM, but I'm surprised that running out of memory would segfault TF. Can we file a bug upstream with them or something?

tensorboard/plugins/histogram/summary_test.py
yatbear: Got it!

tensorboard/plugins/histogram/BUILD
yatbear: Done.

yatbear:
> LGTM, but I'm surprised that running out of memory would segfault TF. Can we file a bug upstream with them or something?

I don't think this is a TF issue, I commented out all the tf code in `test_with_large_counts`, it ran of out memory with just these two lines:
```
large_count = 20_000_000
data = [0] + [1] * large_count
```
Setting the test size to `small` only gives us 20MB RAM.

nfelt:
> > LGTM, but I'm surprised that running out of memory would segfault TF. Can we file a bug upstream with them or something?
> 
> I don't think this is a TF issue, I commented out all the tf code in `test_with_large_counts`, it ran of out memory with just these two lines:
> 
> ```
> large_count = 20_000_000
> data = [0] + [1] * large_count
> ```
> 
> Setting the test size to `small` only gives us 20MB RAM, and the size of `data` created above is ~152MB.

Hmm, that makes since, but was it running out of RAM in OSS though?  If 20MB is the Bazel size I'd expect CI to have failed too.

And the fact that TF segfaulted still seems like a TF issue?  If those 2 lines are enough to OOM, wouldn't the internal test have crashed before the TF segfault?  Unless I guess it was running multiple tests in different threads or something and the OOM caused a different thread to segfault?

yatbear:
> > > LGTM, but I'm surprised that running out of memory would segfault TF. Can we file a bug upstream with them or something?
> > 
> > 
> > I don't think this is a TF issue, I commented out all the tf code in `test_with_large_counts`, it ran of out memory with just these two lines:
> > ```
> > large_count = 20_000_000
> > data = [0] + [1] * large_count
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Setting the test size to `small` only gives us 20MB RAM, and the size of `data` created above is ~152MB.
> 
> Hmm, that makes since, but was it running out of RAM in OSS though? If 20MB is the Bazel size I'd expect CI to have failed too.
> 
> And the fact that TF segfaulted still seems like a TF issue? If those 2 lines are enough to OOM, wouldn't the internal test have crashed before the TF segfault? Unless I guess it was running multiple tests in different threads or something and the OOM caused a different thread to segfault?

The OSS ran fine without any segfault before this change.

PR #5349, opened 2021-09-28
runs table: fix color-picker position

stephanwlee:
ngx-color-picker is capable of automatically position w.r.t. the view
port. However, because:
1. we hard coded "bottom-right" position, and
2. we had flicker of color-picker positioning incorrectly in the
viewport because of lack of `overflow: hidden`

it miscalculated the dialog's position and free space w.r.t. to the view
port. By putting the `overflow`, it knows to only take available pixels
in the view port (instead of making the page scrollable and think that
there exists more space below the view port) into account.

Previous:
image

New:
image


stephanwlee:
@JamesHollyer ping!

JamesHollyer (approved):
LGTM

PR #5356, opened 2021-10-01
histogram: implement buckets_v3

yatbear:
Main changes:
- For `single value input`, the histogram output will no longer be reshaped to length 1, instead, there will be `{bucket_count}` buckets where the first `{bucket_count - 1}` buckets have `zero` counts and the last bucket has count `{data_size}`, since only the last bucket is closed-closed and the other buckets are closed-open.
- Tests for input data containing a single unique value are pushed down to each test case, and more v3 pb op tests will be added once `histogram_pb_v3` is implemented. 
- Naming `singular` is replaced by `single_value` in v3 to avoid confusion with the mathematical term `singularity`.

#histogram #tpu


nfelt (changes requested):

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -465,17 +466,24 @@ def when_nonsingular():
...
+                # Left and right edges are the same for single value input.
+                edges = tf.fill([bucket_count], max_)
+                # Counts for the first {bucket_count - 1} buckets [v, v) are 0.
+                zero_bucket_counts = tf.constant(
+                    [0] * (bucket_count - 1), dtype=tf.int32
                 )
-                return tf.transpose(
-                    a=tf.stack([bucket_starts, bucket_ends, bucket_counts])
+                # Count for last bucket [v, v] is {data_size}.
+                last_bucket_count = tf.expand_dims(data_size, 0)
```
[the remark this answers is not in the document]

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -465,17 +466,24 @@ def when_nonsingular():
...
-                bucket_ends = tf.stack([center + 0.5])
-                bucket_counts = tf.stack(
-                    [tf.cast(tf.size(input=data), tf.float64)]
+            def when_single_value():
+                """When input data contains a single unique value."""
+                # Left and right edges are the same for single value input.
+                edges = tf.fill([bucket_count], max_)
+                # Counts for the first {bucket_count - 1} buckets [v, v) are 0.
+                zero_bucket_counts = tf.constant(
+                    [0] * (bucket_count - 1), dtype=tf.int32
```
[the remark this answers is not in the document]
yatbear: Done. For bucket_count == 0 case, used an `logical_or` instead.

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -70,16 +70,6 @@ def test_empty_input_of_high_rank(self):
         buckets = tensor_util.make_ndarray(pb.value[0].tensor)
         np.testing.assert_allclose(buckets, np.array([]).reshape((0, 3)))
 
-    def test_singleton_input(self):
```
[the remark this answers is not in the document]
yatbear: Done.

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -263,6 +293,99 @@ def graph_fn():
...
+        # Delete the event file to reset to an empty directory for later calls.
+        # TODO(nickfelt): use a unique subdirectory per writer instead.
+        os.remove(event_files[0])
+        return events[1]
+
+    def write_histogram_event(self, *args, **kwargs):
+        kwargs.setdefault("step", 1)
+        writer = tf2.summary.create_file_writer(self.get_temp_dir())
+        with writer.as_default():
+            summary.histogram(*args, **kwargs)
```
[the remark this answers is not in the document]
yatbear: Done. Thanks!

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -263,6 +293,99 @@ def graph_fn():
...
+        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
+        np.testing.assert_allclose(buckets, np.array([[11.5, 12.5, 1]]))
+
+    def test_input_with_all_same_values(self):
+        pb = self.histogram("twelven", [12, 12, 12])
+        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
+        np.testing.assert_allclose(buckets, np.array([[11.5, 12.5, 3]]))
+
+
+class SummaryV3OpTest(SummaryBaseTest, tf.test.TestCase):
```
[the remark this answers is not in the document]
yatbear: Done. Thanks for the suggestions!

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -263,6 +293,99 @@ def graph_fn():
...
+        pb = self.histogram("twelve", [12])
+        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
+        # By default there will be 30 buckets.
+        expected_buckets = np.array(
+            [[12, 12, 0] for _ in range(29)] + [[12, 12, 1]]
+        )
+        np.testing.assert_allclose(buckets, expected_buckets)
+
+    def test_input_with_all_same_values(self):
+        pb = self.histogram("twelven", [12, 12, 12], buckets=2)
```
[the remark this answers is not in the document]
yatbear: I picked a random number earlier. Changed `buckets` to `None` like the other test cases.

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -381,11 +318,19 @@ def test_singleton_input(self):
...
+        # By default there will be 30 buckets.
+        expected_buckets = np.array(
+            [[12, 12, 0] for _ in range(29)] + [[12, 12, 3]]
+        )
         np.testing.assert_allclose(buckets, expected_buckets)
 
+    def test_zero_bucket_count(self):
+        pb = self.histogram("zero_bucket_count", [1, 1, 1], buckets=0)
+        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
+        np.testing.assert_allclose(buckets, np.array([]).reshape((0, 3)))
```
nfelt, line 332: assert_array_equal makes a bit more sense here IMO than assert_allclose, since there are no actual values to compare, so this is really just a shape comparison.  In general, I'd try to use stricter assertions except in cases where we really need to allow for floating point error.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -471,13 +475,10 @@ def when_single_value():
                 # Left and right edges are the same for single value input.
                 edges = tf.fill([bucket_count], max_)
                 # Counts for the first {bucket_count - 1} buckets [v, v) are 0.
-                zero_bucket_counts = tf.constant(
-                    [0] * (bucket_count - 1), dtype=tf.int32
-                )
+                zero_bucket_counts = tf.repeat([0], bucket_count - 1)
```
nfelt, line 478: I know that `tf.repeat()` was my recommendation... but now that I'm looking at the code again, it's a bit odd that we're constructing two arrays here and using `tf.fill([length], value)` for one, and then `tf.repeat([value], length)` for the other - exactly opposite approaches!

Optional, but probably more consistent to do `tf.fill([bucket_count - 1], 0)` instead here.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -427,7 +427,11 @@ def _buckets_v3(data, bucket_count=None):
         data = tf.reshape(data, shape=[-1])  # flatten
         data = tf.cast(data, tf.float64)
         data_size = tf.size(input=data)
-        is_empty = tf.equal(data_size, 0)
+        # If bucket_count is specified as zero, an empty tensor of shape
```
nfelt, line 430: So based on the design spec, there is actually a distinction in these two cases, where "empty" in terms of no buckets should be the `(0, 3)` tensor, but "empty" in terms of no data is supposed be the `(num_buckets, 3)` tensor of all zero values.  (This is a change from the v2 behavior of also returning the empty `(0, 3)` tensor in that case, which means that the `test_empty_input` tests in the base case would need overrides for the v3 format.)  It's ok to not implement that part in this PR, we could do it in a later one, but just wanted to highlight that in case it wasn't clear.

That said, the two cases _can_ still both be handled as one condition, like you have it here, if we change the `when_empty()` implementation to:

```
def when_empty():
    return tf.zeros((bucket_count, 3), dtype=tf.float64)
```

That will naturally handle the case of 0 buckets by emitting the empty tensor of shape (0, 3).

tensorboard/plugins/histogram/summary_v2.py

tensorboard/plugins/histogram/summary_test.py
yatbear: Done.

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -293,47 +266,11 @@ def graph_fn():
...
-        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
-        np.testing.assert_allclose(buckets, np.array([[11.5, 12.5, 1]]))
-
-    def test_input_with_all_same_values(self):
-        pb = self.histogram("twelven", [12, 12, 12])
-        buckets = tensor_util.make_ndarray(pb.value[0].tensor)
-        np.testing.assert_allclose(buckets, np.array([[11.5, 12.5, 3]]))
-
-
-class SummaryV3OpTest(SummaryBaseTest, tf.test.TestCase):
```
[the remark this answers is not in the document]
yatbear: Done.

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -263,6 +347,14 @@ def graph_fn():
         with writer.as_default():
             graph_fn.get_concrete_function()
 
+    def test_zero_bucket_count(self):
+        self.skipTest(
+            "TODO: figure out why this doesn't work in graph test case"
```
nfelt, line 352: So I think the reason this was failing is that in graph mode, when it compiles the TensorFlow code into a graph, all of the branches of the `tf.cond()`s are inserted as parts of the graph (even if those branches won't be taken when the graph is actually executed).  When doing that, TensorFlow does shape inference, described a little bit here: https://www.tensorflow.org/guide/create_op#shape_functions_in_c

Basically, it tries to compute what the resulting shape will be for all of the ops in all of the branches.  At the line where we currently do `tf.fill([bucket_count - 1], 0)`, because `bucket_count` is a constant, it is able to determine at compile time that this will result in `tf.fill([-1], 0)` and the shape function for `tf.fill` emits an error, because this would result in an output shape that doesn't make sense.

Hence the error (at least, the one that I was seeing when trying this PR locally):

> ValueError: Fill dimensions must be >= 0 for '{{node zero_bucket_count/write_summary/buckets/cond/cond/Fill_1}} = Fill[T=DT_INT32, index_type=DT_INT32](zero_bucket_count/write_summary/buckets/cond/cond/Fill_1/dims, zero_bucket_count/write_summary/buckets/cond/cond/Fill_1/value)' with input shapes: [1], [] and with input tensors computed as partial shapes: input[0] = [?].

In other words, it doesn't actually mean it's picking the wrong branch of the conditional to execute; it's that during graph compilation, it visits all the branches regardless of what will be ultimately executed.

(Aside: you could argue that when `bucket_count` is a constant 0, we could actually optimize away all the `tf.cond` ops and all the branches besides the empty case, at which point we wouldn't hit this error.  I'm guessing that the compilation's constant-folding isn't quite smart enough to do that.  Also, if you change the test to pass `bucket_count=tf.Variable(0)` and replace the use of `max()` with `tf.math.maximum()` as mentioned in the other comment, it will also pass, since now that the bucket count is a variable, it doesn't do as much shape inference and can't raise the false positive error.)

Ultimately, the best way to fix this is to just ensure that all the branches are still creating the correctly shaped tensor of dimensions `(bucket_count, 3)`, even when we know that branch will never be executed. Here's a patch that gets the test to pass:

```diff
diff --git i/tensorboard/plugins/histogram/summary_v2.py w/tensorboard/plugins/histogram/summary_v2.py
index bff980492..e1cd01846 100644
--- i/tensorboard/plugins/histogram/summary_v2.py
+++ w/tensorboard/plugins/histogram/summary_v2.py
@@ -425,6 +425,8 @@ def _buckets_v3(data, bucket_count=None):
     with tf.name_scope("buckets"):
         tf.debugging.assert_scalar(bucket_count)
         tf.debugging.assert_type(bucket_count, tf.int32)
+        # Treat a negative bucket count as zero.
+        bucket_count = tf.math.maximum(0, bucket_count)
         data = tf.reshape(data, shape=[-1])  # flatten
         data = tf.cast(data, tf.float64)
         data_size = tf.size(input=data)
@@ -440,7 +442,7 @@ def _buckets_v3(data, bucket_count=None):
             2. If the input data is empty, a tensor of shape (bucket_count, 3)
               of all zero values will be returned.
             """
-            return tf.zeros((max(0, bucket_count), 3), dtype=tf.float64)
+            return tf.zeros((bucket_count, 3), dtype=tf.float64)

         def when_nonempty():
             min_ = tf.reduce_min(input_tensor=data)
@@ -480,11 +482,12 @@ def _buckets_v3(data, bucket_count=None):
                 """When input data contains a single unique value."""
                 # Left and right edges are the same for single value input.
                 edges = tf.fill([bucket_count], max_)
-                # Counts for the first {bucket_count - 1} buckets [v, v) are 0.
-                zero_bucket_counts = tf.fill([bucket_count - 1], 0)
-                # Count for last bucket [v, v] is {data_size}.
+                # Bucket counts are 0 except the last bucket (if bucket_count > 0),
+                # which is `data_size`. Ensure that the resulting counts vector has
+                # length `bucket_count` always, including the bucket_count==0 case.
+                zeroes = tf.fill([bucket_count], 0)
                 bucket_counts = tf.cast(
-                    tf.concat([zero_bucket_counts, [data_size]], 0),
+                    tf.concat([zeroes[:-1], [data_size]], 0)[:bucket_count],
                     dtype=tf.float64,
                 )
                 return tf.transpose(a=tf.stack([edges, edges, bucket_counts]))
```

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -293,47 +266,11 @@ def graph_fn():
...
-
-    def write_histogram_event(self, *args, **kwargs):
-        kwargs.setdefault("step", 1)
-        writer = tf2.summary.create_file_writer(self.get_temp_dir())
-        with writer.as_default():
-            summary.histogram(*args, **kwargs)
-        writer.close()
 
+class SummaryV3OpGraphTest(SummaryV2OpTest, tf.test.TestCase):
+    def call_histogram_op(self, *args, **kwargs):
```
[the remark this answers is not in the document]
nfelt: Ack, fair enough to inherit that way instead.

tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -426,19 +427,29 @@ def _buckets_v3(data, bucket_count=None):
...
         def when_empty():
-            return tf.constant([], shape=(0, 3), dtype=tf.float64)
+            """When input data is empty or bucket_count is zero.
+
+            1. If bucket_count is specified as zero, an empty tensor of shape
+              (0, 3) will be returned.
+            2. If the input data is empty, a tensor of shape (bucket_count, 3)
+              of all zero values will be returned.
+            """
+            return tf.zeros((max(0, bucket_count), 3), dtype=tf.float64)
```
nfelt, line 443: We should use `tf.math.maximum()` here, rather than Python's `max()`, in order to be compatible with `bucket_count` values that aren't native Python numbers (e.g. a `tf.constant()` instead).

Also I'd recommend just adding a conversion up at the top that does `bucket_count = tf.math.maximum(0, bucket_count)` so we don't have to worry about negative bucket counts in any of the later logic.

tensorboard/plugins/histogram/summary_test.py
yatbear: Thanks so much for the clear explanation and the patch!

tensorboard/plugins/histogram/summary_test.py
```diff
@@ -316,14 +316,15 @@ def write_histogram_event(self, *args, **kwargs):
...
-        @tf2.function
+        @tf2.function(autograph=True)
         def graph_fn():
             # Recreate the active scope inside the defun since it won't propagate.
             with tf.name_scope(scope):
                 self.call_histogram_op(*args, **kwargs)
 
         writer = tf2.summary.create_file_writer(self.get_temp_dir())
         with writer.as_default():
+            # print(tf2.autograph.to_code(graph_fn.python_function))
```
nfelt, line 327: Unintentional?  Also the `autograph=True` addition above seems like it shouldn't be necessary?  (I believe it's true by default.)

nfelt (approved):
LGTM modulo what looks like unintentional additions

PR #5357, opened 2021-10-01
hparams: treat no data as an empty experiment rather than an error

### LINKED ISSUE #5347: Having no hparams data should return 200 instead of 400
Redefine following endpoints:
- `hparams/experiment`
- `hparams/session_groups`
- `hparams/download_data`

So that having no hparams data is still a 200 OK response, and just returns either A) an entirely empty response `{}` or B) a placeholder response for an empty experiment (with no hparams or metrics, and placeholder values for the name, description, etc).

Googlers, see b/156776593 for more context.

nfelt:
Fixes #5347.

This updates the HParams plugin's `Context.experiment_from_metadata()` to always return an `Experiment` proto, regardless of whether there are any HParams summaries.  The experiment computation already has logic that handles the case where no experiment-level HParams summary was logged by synthesizing an experiment from the existing session-level summaries (which results in an experiment with hparam and metric infos, but no name, description, user, or started time).  This PR simply adjusts that logic so that we return the synthetic experiment even when there are no session-level summaries, in which case we return empty hparam and metric info repeated fields as well.

This is done to address #5347 and ensure that the HParams endpoints don't return 400 errors for the common and normal situation in which there are no hparams summaries logged; instead they will now return 200 OK and an appropriately empty response.  Note that the HParams plugin `is_active()` will already have prevented the HParams plugin from showing up as active in this case, but this ensures those errors aren't thrown if the user manually navigates to it in the INACTIVE menu, or if other parts of the app were to invoke those endpoints unconditionally.

Test plan:

- Added a unit test to backend_context_test.py
- Tested internally, see cl/395956617
- Tested running TB with this PR on a logdir with no hparams summaries. UI still shows the expected `No hparams data was found.` message. The only visible difference is that now that it's not an error, we do load session groups, and thus show that 0 session groups were loaded, making this page "1 / 0", and we show the download as CSV/JSON/TEX links (which result in empty files, see below). If we really care, we could update the HParams frontend in `tf-hparams-query-pane.ts` to check for this case, but it doesn't seem all that important to me.
- Confirmed endpoints don't raise errors
    - `/experiment` returns
        ```
        {
          "name": "",
          "description": "",
          "user": "",
          "timeCreatedSecs": 0.0,
          "hparamInfos": [],
          "metricInfos": []
        }
        ```
    - `/session_groups` returns
        ```
        {
          "sessionGroups": [],
          "totalSize": 0
        }
        ```
    - `/download_data` CSV returns `\r\n`
    - `/download_data` JSON returns `{"header": [], "rows": []}`
    - `/download_data` TEX returns
        ```
        \begin{table}[tbp]
        \begin{tabular}{}
         \\ \hline
        \hline
        \end{tabular}
        \end{table}
        ```

### DIFF
--- a/tensorboard/plugins/hparams/BUILD
+++ b/tensorboard/plugins/hparams/BUILD
@@ -38,7 +38,6 @@ py_library(
         ":error",
         ":metadata",
         ":protos_all_py_pb2",
-        "//tensorboard:errors",
         "//tensorboard:plugin_util",
         "//tensorboard/backend:http_util",
         "//tensorboard/data:provider",
--- a/tensorboard/plugins/hparams/backend_context.py
+++ b/tensorboard/plugins/hparams/backend_context.py
@@ -70,7 +70,7 @@ class Context(object):
         Returns:
           The experiment protobuffer. If no tags are found from which an experiment
           protobuffer can be built (possibly, because the event data has not been
-          completely loaded yet), returns None.
+          completely loaded yet), returns an entirely empty experiment.
         """
         experiment = self._find_experiment_tag(hparams_run_to_tag_to_content)
         if experiment:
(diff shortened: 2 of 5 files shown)


yatbear (approved):
Thanks for fixing this!

PR #5361, opened 2021-10-05
settings: customizable color palette

stephanwlee:
TensorBoard currently ships with a set of colors for visualization. One
of the key limitations and consideration was that we wanted the default
experience to be respectful to a person with color blindness. As a
result, our color set only contained seven colors and we have received
many feedbacks that the set is too repetitive.

This change mitigates the issue by allowing users to customize and
select more expanded set of colors. However, this change does not yet
remember the selection and will be reset every time the application
opens. This problem will be solved in a subsequent change.

image


tensorboard/webapp/util/colors.ts
```diff
@@ -73,3 +73,77 @@ export const DEFAULT_PALETTE: ColorPalette = {
...
+  inactive: {
+    name: 'Light gray',
+    lightHex: '#e0e0e0', // Gray 300
+    darkHex: '#3b3b3b',
+  },
+};
+
+export const colorBlindAssist3: ColorPalette = {
+  id: 'color_blind_assist_3',
+  name: 'Color Blind Assist',
```
nfelt, line 122: This name is kind of confusing now that it's not in the context of the others in palettes.ts.  It's actually less-colorblind-friendly than the classic scheme probably is, since the tradeoff made here is lower contrast between colors but a higher number of colors.  Maybe it would be better to call this "Classic Extended" or just "Extended" or something?

tensorboard/webapp/settings/_views/settings_dialog_component.ng.html
```diff
@@ -55,3 +55,97 @@ <h3>Settings</h3>
     Page size has to be a positive integer.
   </mat-error>
 </div>
+
+<div>
+  <h4>Visualization colors</h4>
+  <div>
+    Customize colors that visualzation such as line chart will use. When there
```
nfelt, line 62: typo: visualization, missing `i`

nfelt, line 62: typo: visualization, missing `i`  Also should probably be "visualizations"?

FWIW, I think this might read a bit better as something like

> Choose the color palette to use for visualizations (currently only works for the Time Series plugin). When there are more items than the number of colors, it will cycle repeatedly.

In particular - "customize" seems odd since you can't actually specify a custom palette, and specifying "runs" seems a little too specific, because Time Series lets you assign colors to groups that aren't necessarily just 1 run.

PR #5362, opened 2021-10-05
time series: remember settings pane open state

stephanwlee:
This change puts the settings pane opened state in the global settings
storage so the view remain the same as how user has left it.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -400,9 +400,13 @@ const reducer = createReducer(
         ? !partialSettings.timeSeriesPromotionDismissed
         : state.promoteTimeSeries;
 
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -399,9 +399,13 @@ const reducer = createReducer(
         ? !partialSettings.timeSeriesPromotionDismissed
         : state.promoteTimeSeries;
 
+    const isSettingsPaneOpen =
+      partialSettings.timeSeriesSettingsPaneOpened ?? state.isSettingsPaneOpen;
```
japie1235813, line 403: somehow I did not see `isSettingsPaneOpen: true,` around line 265 in this pr. probably need rebase before merge

japie1235813 (approved):

PR #5364, opened 2021-10-07
Add infrastructure for numbering experiments

JamesHollyer:
Changes experiment alias to be an object with includes the normal text alias plus a number to help disambiguate them. This also adds the component that knows how to print out the new ExperimentAlias object.


tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -39,9 +42,10 @@ export function getDisplayNameForRun(
 export function partitionSeries(series: PartialSeries[]): PartitionedSeries[] {
   const partitionedSeries: PartitionedSeries[] = [];
   for (const datum of series) {
-    const currentPartition: Array<
-      Omit<PartitionedSeries, 'partitionSize' | 'partitionIndex'>
-    > = [];
+    const currentPartition: Array<Omit<
+      PartitionedSeries,
+      'partitionSize' | 'partitionIndex'
+    >> = [];
```
stephanwlee, line 48: If this is just formatting change, please omit this as it is distractive.

tensorboard/webapp/widgets/experiment_alias/experiment_alias_module.ts
```diff
@@ -0,0 +1,25 @@
+/* Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
stephanwlee, line 1: 2021

tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -12,22 +12,25 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
     return runId;
   }
 
-  const displayName = [experimentName, run ? run.name : '...']
-    .filter(Boolean)
-    .join('/');
+  let displayName = [run ? run.name : '...'].filter(Boolean).join('/');
+
+  if (experimentAlias) {
+    displayName = `[${experimentAlias.aliasNumber}] ${experimentAlias.aliasText}/${displayName}`;
```
stephanwlee, line 32: No action required: to make this format consistent, this is the type of UX that might benefit from the Angular Pipe.

tensorboard/webapp/metrics/views/card_renderer/run_name_component.ts
```diff
@@ -14,12 +14,20 @@ limitations under the License.
...
+import {ExperimentAlias} from '../../../experiments/types';
+
 @Component({
   selector: 'card-run-name-component',
-  template: '{{ name }}',
+  template: `<tb-experiment-alias
+      *ngIf="experimentAlias != null"
+      [alias]="experimentAlias!"
+    ></tb-experiment-alias>
+    <span *ngIf="experimentAlias != null">/</span>
```
stephanwlee, line 25: Normally, in an inline element, every whitespace counts so Please try to remove whitespace and newlines. 
```
</tb-experiment-alias><span
  *ngIf="experimentAlias != null">/</span>
```

Generally, Prettier is pretty good at doing the right thing.

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.scss
```diff
@@ -0,0 +1,37 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+@import 'tensorboard/webapp/theme/tb_theme';
+
+:host {
+  display: inline-flex;
+  align-items: baseline;
```
stephanwlee, line 19: Try to sort CSS a-z.

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+span {
+  background-color: inherit;
+  caret-color: currentColor;
```
stephanwlee, line 24: `span` would never show caret. Omit this? Same with outline and border.

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.scss
```diff
@@ -0,0 +1,37 @@
...
+  color: currentColor;
+  font: inherit;
+  border: none;
+  outline: none;
+  padding: 0;
+  flex-grow: 1;
```
stephanwlee, line 30: Is there a reason for both spans to be using `flex-grow: 1`?

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.ts
```diff
@@ -0,0 +1,34 @@
...
+  selector: 'tb-experiment-alias',
+  template: `
+    <span class="alias-number">{{ alias.aliasNumber }}</span>
+    <span>{{ alias.aliasText }}</span>
+  `,
+  styleUrls: [`experiment_alias_component.css`],
+})
+export class ExperimentAliasComponent {
+  @Input()
+  alias: ExperimentAlias = {aliasText: '', aliasNumber: -1};
```
stephanwlee, line 33: Omit default inputs to widgets. This component show throw or not be used when its primary input, `alias` is missing.

tensorboard/webapp/metrics/views/card_renderer/run_name_component.ts
```diff
@@ -14,12 +14,20 @@ limitations under the License.
...
+    <span>{{ name }}</span>`,
   styleUrls: ['run_name_component.css'],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class RunNameComponent {
   @Input() name!: string;
```
stephanwlee, line 31: Instead of `name` and `experimentAlias` as Inputs, I am thinking whether we should just pass `Experiment` and `Run` as Input. WDYT?

tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -12,22 +12,25 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+import {ExperimentAlias} from '../../../experiments/types';
 import {Run} from '../../../runs/store/runs_types';
 import {LinkedTime} from '../../types';
 import {PartialSeries, PartitionedSeries} from './scalar_card_types';
 
 export function getDisplayNameForRun(
   runId: string,
   run: Run | null,
-  experimentName: string | null | undefined
+  experimentAlias: ExperimentAlias | null | undefined
```
stephanwlee, line 23: AFAICT, scalar_card_container is another place that uses this utility. Do we need to change anything there for this change? I don't see that module in the changelist. Especially since we have set it to `null` in one of two places that this utility is being used, I suspect we can probably get rid of this utility in favor of common widget or something like that. WDYT?

tensorboard/webapp/util/matcher.ts
```diff
@@ -51,8 +53,9 @@ export function matchRunToRegex(
   const matchables = [runMatchable.runName];
   if (shouldMatchExperiment) {
     matchables.push(
-      runMatchable.experimentAlias,
-      `${runMatchable.experimentAlias}/${runMatchable.runName}`
+      runMatchable.experimentName,
```
stephanwlee, line 56: Please omit experimentName. That was a deliberate choice.

stephanwlee (commented):
Looking great. I will look forward to seeing the tests soon :)

tensorboard/webapp/app_routing/store/app_routing_selectors.ts
```diff
@@ -90,13 +91,17 @@ export const getRouteId = createSelector(
...
     }
 
     const compareParams = routeParams as CompareRouteParams;
-    const map = getCompareExperimentIdAliasSpec(compareParams);
-    return Object.fromEntries(map.entries());
+    const userDefinedAliasMap = getCompareExperimentIdAliasSpec(compareParams);
+    let aliasMap = new Map<string, ExperimentAlias>();
+    for (let [experimentId, userDefinedAlias] of userDefinedAliasMap) {
+      aliasMap.set(experimentId, userDefinedAlias);
+    }
```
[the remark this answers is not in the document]
JamesHollyer: Oh good catch. At one point this was where I was adding the numbers in and that was useful. It does nothing now.

tensorboard/webapp/experiments/types.ts
```diff
@@ -23,3 +23,8 @@ export declare interface Experiment {
   tags?: string[];
   related_links?: Array<{name: string; url: string}>;
 }
+
+export declare interface ExperimentAlias {
```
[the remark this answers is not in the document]
JamesHollyer: Removed. Why do we have declare on Experiment? I don't know where else that is defined.

JamesHollyer (commented):
Thanks for the review! PTAL at the changes and tests.

tensorboard/webapp/app_routing/store_only_utils.ts
```diff
@@ -30,13 +31,16 @@ import {CompareRouteParams, Route} from './types';
...
-): Map<string, string> {
-  const idToDisplayName = new Map<string, string>();
+): Map<string, ExperimentAlias> {
+  const idToDisplayName = new Map<string, ExperimentAlias>();
   const nameAndIds = parseCompareExperimentStr(routeParams.experimentIds);
+  let aliasNumber = 0;
   for (const {id, name} of nameAndIds) {
-    if (name) {
-      idToDisplayName.set(id, name);
+    aliasNumber++;
```
stephanwlee, line 39: Is it contract of the utility that the aliasNumber starts from 1 instead of 0? That looks a lot like presentational logic and not data logic which is weird to find in the data util. 

Anyways, if it should start from 1, please document it on jsdoc since that contract is not documentable with types alone.

tensorboard/webapp/app_routing/store_only_utils_test.ts
```diff
@@ -39,14 +39,37 @@ describe('app_routing store_only_utils test', () => {
...
     });
 
     it('does not include empty aliases', () => {
       const map = getCompareExperimentIdAliasSpec({
         experimentIds: 'a:123,:345',
       });
-      expect(map).toEqual(new Map([['123', 'a']]));
+      expect(map).toEqual(new Map([['123', {aliasText: 'a', aliasNumber: 1}]]));
+    });
+    it('does not include empty aliases', () => {
```
stephanwlee, line 63: For consistency, please put a line break in between.

tensorboard/webapp/metrics/views/card_renderer/run_name_component.ts
stephanwlee: No action required: since it is literally "RunName" component, I think knowing about `Run` is not too wild. The suggestion was not to make it so that the component gets more information but rather I did not want the consumer of the widgets to carefully destruct information readily available on Run and Experiment to conform to this Input.

tensorboard/webapp/metrics/views/card_renderer/run_name_component.ts
```diff
@@ -14,12 +14,20 @@ limitations under the License.
...
+  @Input() experimentAlias: ExperimentAlias | null | undefined;
```
stephanwlee, line 32: Checking: is it important for the widget to support `undefined` alias? If not, can we omit it? Unlike `null`, `undefined` generally feels like a fault of a user.

tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -12,22 +12,25 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 ): string {
-  if (!run && !experimentName) {
+  if (!run && !experimentAlias) {
     return runId;
   }
 
-  const displayName = [experimentName, run ? run.name : '...']
-    .filter(Boolean)
-    .join('/');
+  let displayName = [run ? run.name : '...'].filter(Boolean).join('/');
```
stephanwlee, line 29: This array only has one item so this feels incorrect. Just do `displayName = run ? run.name : '...'` or `displayName = run?.name ?? '...'` (which functionally is different but run.name should not be empty as it can break other UI so default showing "..." feels good.)

tensorboard/webapp/util/matcher.ts
```diff
@@ -12,14 +12,16 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
+import {ExperimentAlias} from '../experiments/types';
+
```
stephanwlee, line 16: This should appear below `@fileoverview`

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.scss
```diff
@@ -0,0 +1,33 @@
...
+  display: inline-flex;
+  align-items: baseline;
+}
+
+span {
+  background-color: inherit;
+  color: currentColor;
+  font: inherit;
+  padding: 0;
+}
```
stephanwlee, line 33: All of these properties/values are default values for `span`. Maybe drop it.

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.ts
```diff
@@ -0,0 +1,34 @@
...
+  selector: 'tb-experiment-alias',
+  template: `
+    <span *ngIf="alias" class="alias-number">{{ alias.aliasNumber }}</span>
+    <span *ngIf="alias">{{ alias.aliasText }}</span>
+  `,
+  styleUrls: [`experiment_alias_component.css`],
+})
+export class ExperimentAliasComponent {
+  @Input()
+  alias?: ExperimentAlias;
```
stephanwlee, line 33: Similar to the default value argument, this module does not make too much sense when `alias` is undefined. Please make it `alias!: ExperimentAlias`. If empty value is explicitly allowed, maybe type this as `alias!: ExperimentAlias | null` or, in very rare case, `alias!: ExperimentAlias | undefined` (to be very explicit).

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * The component used to display experiment alias and help users distinguish
+ * this experiment from others.
+ */
+@Component({
+  selector: 'tb-experiment-alias',
+  template: `
+    <span *ngIf="alias" class="alias-number">{{ alias.aliasNumber }}</span>
+    <span *ngIf="alias">{{ alias.aliasText }}</span>
```
stephanwlee, line 27: Can be rewritten as
```html
<ng-container *ngIf="alias">
  <span  class="alias-number">{{ alias.aliasNumber }}</span>
  <span>{{ alias.aliasText }}</span>
</ng-container>
```

tensorboard/webapp/widgets/experiment_alias/experiment_alias_test.ts
```diff
@@ -0,0 +1,47 @@
...
+    const numberSpan = fixture.debugElement.queryAll(By.css('.alias-number'));
+    expect(numberSpan[0].nativeElement.textContent).toBe('1');
+  });
+
+  it('renders nothing no value', () => {
+    const fixture = TestBed.createComponent(ExperimentAliasComponent);
+    fixture.detectChanges();
+
+    const input = fixture.debugElement.queryAll(By.css('span'));
+    expect(input.length).toBe(0);
```
stephanwlee, line 45: Might be better to assert on its emptiness with `expect(fixture.debugElement.nativeElement.textContent).toBe('')`.

tensorboard/webapp/metrics/views/card_renderer/utils.ts
JamesHollyer: done.

tensorboard/webapp/app_routing/store_only_utils_test.ts
JamesHollyer: done.

tensorboard/webapp/util/matcher.ts
JamesHollyer: Done.

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.scss
JamesHollyer: Done.

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.ts
JamesHollyer: Same hesitation as before. I will not be hesitant to use the bang in the future.

tensorboard/webapp/widgets/experiment_alias/experiment_alias_component.ts
JamesHollyer: I am removing the ngIf completely now because we are not allowing undefined. But, that is good to know.

tensorboard/webapp/widgets/experiment_alias/experiment_alias_test.ts
JamesHollyer: This test has been removed.

tensorboard/webapp/app_routing/store_only_utils.ts
JamesHollyer: Documented.

tensorboard/webapp/core/effects/core_effects_test.ts
```diff
@@ -462,6 +462,20 @@ describe('core_effects', () => {
...
+      store.refreshState();
+      tick(TEST_ONLY.ALIAS_CHANGE_RUNS_RELOAD_THROTTLE_IN_MS * 2);
+      expect(recordedActions).toEqual([
+        polymerRunsFetchRequested(),
+        polymerRunsFetchRequested(),
+        polymerRunsFetchRequested(),
+        polymerRunsFetchRequested(),
+      ]);
+
       discardPeriodicTasks();
```
stephanwlee, line 479: If it is not troublesome, would you mind factoring this test out into its own behavioral test spec? This spec was already scary long and it is even longer now.

tensorboard/webapp/metrics/views/card_renderer/utils.ts
stephanwlee: I forget, was the new line in the displayName between the number and alias text intentional? That seems counter intuitive.

tensorboard/webapp/app_routing/store/app_routing_selectors.ts
```diff
@@ -100,8 +103,12 @@ export const getExperimentIdToAliasMap = createSelector(
     }
 
     const compareParams = routeParams as CompareRouteParams;
-    const map = getCompareExperimentIdAliasSpec(compareParams);
-    return Object.fromEntries(map.entries());
+    // We still need to use this internally until while switching to the new
+    // method.
+    // tslint:disable:deprecation
+    const userDefinedAliasMap = getCompareExperimentIdAliasSpec(compareParams);
+    // tslint:enable:deprecation
```
stephanwlee, line 110: OSS TB does not use tslint so this is vestigial; we do not use something that is not enforceable.

tensorboard/webapp/core/effects/core_effects_test.ts
JamesHollyer: No problem. I decided to copy the logic instead of making a separate describe/before each. Seemed unnecessarily abstract for just 2 tests.

tensorboard/webapp/app_routing/store/app_routing_selectors.ts
JamesHollyer: Removed.

tensorboard/webapp/metrics/views/card_renderer/utils.ts
JamesHollyer: There is no newline there. Github is just wrapping it so it fits on the screen.

tensorboard/webapp/core/effects/core_effects_test.ts
```diff
@@ -465,15 +465,52 @@ describe('core_effects', () => {
...
     }));
 
+    fit('fetches polymer runs when alias numbers change when in comparison', fakeAsync(() => {
```
stephanwlee, line 468: Sync fails because of `fit`.

JamesHollyer: doh!

PR #5366, opened 2021-10-07
demo: fix example plugin

japie1235813:
Replace old multiplexer with `MultiplexerDataProvider` and make related changes.

### DIFF
--- a/tensorboard/examples/plugins/example_basic/tensorboard_plugin_example/plugin.py
+++ b/tensorboard/examples/plugins/example_basic/tensorboard_plugin_example/plugin.py
@@ -20,6 +20,7 @@ import os
 
 from tensorboard.plugins import base_plugin
 from tensorboard.util import tensor_util
+from tensorboard import plugin_util
 import werkzeug
 from werkzeug import wrappers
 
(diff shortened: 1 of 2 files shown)


tensorboard/examples/plugins/example_basic/tensorboard_plugin_example/plugin.py
```diff
@@ -92,16 +93,23 @@ def _serve_greetings(self, request):
...
-            ]
+            read_result = self.data_provider.read_tensors(
+                ctx,
+                downsample=1,
+                plugin_name=metadata.PLUGIN_NAME,
+                experiment_id=experiment,
+                run_tag_filter=provider.RunTagFilter(runs=[run], tags=[tag]),
+            )
+
+            data = read_result.get(run, {}).get(tag, {})
```
nfelt, line 110: If there is no actual data for this run and tag, the result will be an empty dictionary, but then on the next line `data[0]` will fail.  It's also not actually the right type - to match `read_tensors()`, that type should be a list.

I'd just change it to `.get(tag, [])` and then add an explicit check, like

```py
if not data:
    raise werkzeug.exceptions.BadRequest("Invalid run or tag")
event_data = [data[0].numpy.item().decode("utf-8")]
```

The `except KeyError` below only worked with multiplexer.Tensors(), it won't work with the code you have now since `.get()` on a dict won't actually raise a KeyError.

Issue #5369, opened 2021-10-09
Text plugin markdown encoding is slow


stephanwlee:
TensorBoard currently, in its Python process, encode a text tensor into a markdown (and sometimes to a table).

Recently, we have noticed that it is notably slower on large tensors which we have attributed largely to the markdown parser. While debugging, there was one particular hotspot which we have noticed--hint: it is not the markdown parser!

<details>
<summary>Test script</summary>

### Faster
```py
import time
from tensorboard import plugin_util

for i in range(5000, 50000, 5000):
    start = time.time()
    plugin_util.markdown_to_safe_html(
        ("[" + " ".join([str(i) for i in range(i)]) + "]")
    )
    end = time.time()
    print("Converting %d length string took %f" % (i, end - start))

# 
# Converting 5000 length string took 0.016349
# Converting 10000 length string took 0.033403
# Converting 15000 length string took 0.053684
# Converting 20000 length string took 0.070949
# Converting 25000 length string took 0.090817
# Converting 30000 length string took 0.108568
# Converting 35000 length string took 0.129561
# Converting 40000 length string took 0.147728
# Converting 45000 length string took 0.169784
```

### Slower
```py
import time
from tensorboard import plugin_util

for i in range(5000, 50000, 5000):
    start = time.time()
    plugin_util.markdown_to_safe_html(
        ("[" + " ".join([str(i) for i in range(i)]) + "]").encode("utf8")
    )
    end = time.time()
    print("Converting %d length string took %f" % (i, end - start))

# 
# Converting 5000 length string took 0.110732
# Converting 10000 length string took 0.428144
# Converting 15000 length string took 0.969050
# Converting 20000 length string took 1.694072
# Converting 25000 length string took 2.639599
# Converting 30000 length string took 3.787034
# Converting 35000 length string took 5.172249
# Converting 40000 length string took 6.771216
# Converting 45000 length string took 8.655072
```

</details>

When expected further, it looks like one innocuous line is contributing to the most of the time spent in the binary mode.

https://github.com/tensorflow/tensorboard/blob/65371efb6b0e84aabcb3bf31247d58616466fc8c/tensorboard/plugin_util.py#L138

For 45000 length string, specifically, we have spent about 7.8s on just replacement which is quite surprising to me.
```
Decode: 0.140072,
Plus Replace: 7.920270
Plus Convert: 7.945538
```

In fact, when removed the `replace` line, I was able to restore the comparable performance.

[...]

PR #832, opened 2017-12-21
Add text dashboard toggle for enabling markdown

chihuahua:
Added a checkbox to the text dashboard that lets the user control
whether to enable markdown view of text widgets. The
`markdownEnabled` boolean property is stored within local storage.
The property defaults to true.

If markdown is enabled, we parse text as markdown and render it.
This matches current behavior.

image

If markdown is disabled, text is rendered raw, and line numbers are
provided.

image

Fixes #830.

PR #5378, opened 2021-10-14
text: add checkbox to control server-side markdown conversion

### LINKED ISSUE #830: Feature request: option to show raw text instead of Markdown
Most of the time I need to view the raw text rather than markdown rendered. Currently I wrap the generated text with `<pre>` and `</pre>` to avoid symbols being interpreted as style markers, but because the text is generated by a RNN, it sometimes contains `</pre>` itself to break the whole thing.

I believe a checkbox on Tensorboard that turns off Markdown would be really helpful to many people, and it shouldn't be much work, since displaying text without formatting seems a much easier job than markdown rendering.

nfelt:
Workaround for #5369. Fixes #830.

This adds a simple `[√] Enable Markdown` checkbox to the Polymer text plugin, which if deselected (it's selected by default to preserve existing behavior) will skip the markdown conversion step on the server. To do this, we introduce a `markdown` query parameter to the `/data/plugin/text/text` endpoint, and add a `plugin_util.safe_html()` function that does only the sanitization step of `markdown_to_safe_html`.

I've made the checkbox persist to localstorage (like a few other Polymer properties) since I expect most people will either want it always on or always off (and for working around the slow rendering, having to re-disable it on load each time would make this basically worthless as a workaround).

There are some limitations to this approach (it's Polymer-only; client-side markdown rendering would still be nicer). Ultimately, it might make more sense to add a summary API option to specify at write time whether the data should be treated as markdown or not (as suggested in https://github.com/tensorflow/tensorboard/pull/832#issuecomment-353653142). But that requires more in-depth thought since it's persisted forever, and wouldn't help users on older versions of TensorFlow or users who might want to occasionally disable the Markdown rendering for debugging, etc. So even if we do add that option, I think we might still want to keep a UI control, and have it just become a try-state of `auto` (use summary metadata) in addition to `true` and `false`.

Googlers, see cl/402948742 for internal test sync and test case.

### DIFF
--- a/tensorboard/plugins/text/text_plugin.py
+++ b/tensorboard/plugins/text/text_plugin.py
@@ -187,7 +187,13 @@ def text_array_to_html(text_arr, enable_markdown):
             lambda xs: make_table(np.array(xs).reshape(text_arr.shape)),
         )
     else:
-        table = plugin_util.safe_html(make_table(text_arr.astype(str)))
+        # Convert utf-8 bytes to str. The built-in np.char.decode doesn't work on
+        # object arrays, and converting to an numpy chararray is lossy.
+        decode = lambda bs: bs.decode("utf-8") if isinstance(bs, bytes) else bs
+        text_arr_str = np.array(
+            [decode(bs) for bs in text_arr.reshape(-1)]
+        ).reshape(text_arr.shape)
+        table = plugin_util.safe_html(make_table(text_arr_str))
     return warning + table
 
 
--- a/tensorboard/plugins/text/text_plugin.py
+++ b/tensorboard/plugins/text/text_plugin.py
@@ -152,7 +152,7 @@ def reduce_to_2d(arr):
     return arr[slices]
 
 
-def text_array_to_html(text_arr):
+def text_array_to_html(text_arr, enable_markdown):
     """Take a numpy.ndarray containing strings, and convert it into html.
 
     If the ndarray contains a single scalar string, that string is converted to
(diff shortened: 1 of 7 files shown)


tensorboard/plugins/text/text_plugin.py
```diff
@@ -164,29 +164,42 @@ def text_array_to_html(text_arr):
...
+            text_arr.reshape(-1),
+            lambda xs: make_table(np.array(xs).reshape(text_arr.shape)),
+        )
+    else:
+        # Convert utf-8 bytes to str. The built-in np.char.decode doesn't work on
+        # object arrays, and converting to an numpy chararray is lossy.
+        decode = lambda bs: bs.decode("utf-8") if isinstance(bs, bytes) else bs
+        text_arr_str = np.array(
+            [decode(bs) for bs in text_arr.reshape(-1)]
+        ).reshape(text_arr.shape)
```
stephanwlee, line 195: Do note that the markdown version makes sure it accounts for null characters resulting from decoding mismatch. 

In tensorboard/plugin_util.py.
```py
            source_decoded = source.decode("utf-8")
            # Remove null bytes and warn if there were any, since it probably means
            # we were given a bad encoding.
            source = source_decoded.replace("\x00", "")
            total_null_bytes += len(source_decoded) - len(source)
...

    warning = ""
    if total_null_bytes:
        warning = (
            "<!-- WARNING: discarded %d null bytes in markdown string "
            "after UTF-8 decoding -->\n"
        ) % total_null_bytes
```

stephanwlee (approved):
Another compromise we are making is that we cannot turn on/off markdown for a specific card since this is more of a global setting.

tensorboard/plugins/text/text_plugin.py
nfelt: Yeah, I considered adding that to `safe_html` but wasn't sure it was warranted since the sanitizer strips the null bytes anyway (confirmed with a [unit test](https://github.com/tensorflow/tensorboard/blob/ee9d15e15cf195eb4312dc4d3679c32cc1b9f934/tensorboard/plugin_util_test.py#L66-L74)). The reason given in the markdown version for doing the stripping is that otherwise, the markdown conversion gets confused by the null bytes: https://github.com/tensorflow/tensorboard/blob/ee9d15e15cf195eb4312dc4d3679c32cc1b9f934/tensorboard/plugin_util_test.py#L175-L181

So the only difference really is whether we insert the diagnostic HTML comment or not. That didn't quite seem worth any potential overhead to me, but I'm happy to add it back if you prefer?

PR #5391, opened 2021-10-28
histogram: make legacy histogram visualization compatible with v3 data format

yatbear:
Changes:
- Add special handling for single value data case in legacy histogram visualization logic to support v3 format.
- Also add tests for `histogramCore.ts`.

Visualization with **v3** histogram data input:
- overlay mode: 
image
- offset mode: 
image

Visualization with **v2** histogram data input:
image

#histogram #tpu

### DIFF
--- a/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
+++ b/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
@@ -90,6 +90,12 @@ export function intermediateToD3(
     min = 0;
     max = 0;
   }
+  if (max === min) {
+    // If the output range is 0 width, use a default non 0 range for
+    // visualization purpose.
+    max = min * 1.1 + 1;
+    min = min / 1.1 - 1;
+  }
   // Terminology note: _buckets_ are the input to this function,
   // while _bins_ are our output.
   const binWidth = (max - min) / numBins;
@@ -140,12 +146,6 @@ export function backendToVz(histograms: BackendHistogram[]): VzHistogram[] {
   const intermediateHistograms = histograms.map(backendToIntermediate);
   let minmin = d3.min(intermediateHistograms, (h) => h.min);
   let maxmax = d3.max(intermediateHistograms, (h) => h.max);
-  // If the output range is 0 width, use a default non 0 range for
-  // visualization purpose.
-  if (minmin !== undefined && maxmax !== undefined && minmin === maxmax) {
-    maxmax = maxmax * 1.1 + 1;
-    minmin = minmin / 1.1 - 1;
-  }
   return intermediateHistograms.map((h) => ({
     wall_time: h.wall_time,
     step: h.step,
--- a/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
+++ b/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
@@ -100,7 +100,7 @@ export function intermediateToD3(
   // while _bins_ are our output.
   const binWidth = (max - min) / numBins;
   let bucketIndex = 0;
-  let d3HistogramBins: D3HistogramBin[] = [];
+  const d3HistogramBins: D3HistogramBin[] = [];
   for (let i = 0; i < numBins; i++) {
     const binLeft = min + i * binWidth;
     const binRight = binLeft + binWidth;
--- a/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
+++ b/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
@@ -28,8 +28,8 @@ export type BackendHistogram = [
 export type IntermediateHistogram = {
   wall_time: number; // in seconds
   step: number;
-  min: number;
-  max: number;
+  min: number | undefined;
+  max: number | undefined;
   buckets: {
     left: number;
     right: number;
@@ -86,10 +86,22 @@ export function intermediateToD3(
   max: number,
   numBins = 30
 ): D3HistogramBin[] {
+  // Empty case.
+  if (!histogram.buckets || histogram.buckets.length === 0) {
+    return [];
+  }
+  // Single value case.
   if (max === min) {
-    // Create bins even if all the data has a single value.
-    max = min * 1.1 + 1;
-    min = min / 1.1 - 1;
+    let count = 0;
+    for (let bucket of histogram.buckets) {
+      count += bucket.count;
+    }
+    let bins: D3HistogramBin[] = [];
+    for (let i = 0; i < numBins - 1; i++) {
+      bins.push({x: min, dx: 0, y: 0});
+    }
+    bins.push({x: min, dx: 0, y: count});
+    return bins;
   }
   // Terminology note: _buckets_ are the input to this function,
   // while _bins_ are our output.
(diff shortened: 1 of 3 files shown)


tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
```diff
@@ -86,10 +86,22 @@ export function intermediateToD3(
   max: number,
   numBins = 30
 ): D3HistogramBin[] {
+  // Empty case.
+  if (!histogram.buckets || histogram.buckets.length === 0) {
+    return [];
+  }
+  // Single value case.
```
nfelt, line 93: So, but while this does render something, which is better than nothing at all... it's a little hard to interpret (especially in the offset view, where it's just kind of a blank white vertical line until you hover over it).

Kind of like the empty buckets case above, it might produce a cleaner result (and more like Time Series) if instead of special-casing the single value case here, we leave in place the existing logic to compute a spaced-out min and max.  (It's ok that this doesn't match the histogram v3 requirements; those are only about the canonical "persisted" form of the data, and at this point in the code, we're optimizing for the visualization itself, so the reformatting is only about presentation.)  Instead, we can modify how the intersection is computed below so that we still count input buckets with a zero width as contributing to the output bins.  This is basically the same as what the newer Time Series logic does to calculate the bin's "contribution".

I haven't fully tested it but here's a sample patch to the body of the while loop to show the suggestion:

```patch
@@ -119,12 +112,20 @@ export function intermediateToD3(
       // infinite-sized.
       const bucketRight = Math.min(max, histogram.buckets[bucketIndex].right);
       const bucketLeft = Math.max(min, histogram.buckets[bucketIndex].left);
-      const intersect =
-        Math.min(bucketRight, binRight) - Math.max(bucketLeft, binLeft);
-      const count =
-        (intersect / (bucketRight - bucketLeft)) *
-        histogram.buckets[bucketIndex].count;
-      binY += intersect > 0 ? count : 0;
+      const bucketWidth = bucketRight - bucketLeft;
+      if (bucketWidth > 0) {
+        const intersect =
+          Math.min(bucketRight, binRight) - Math.max(bucketLeft, binLeft);
+        const count =
+          (intersect / (bucketRight - bucketLeft)) *
+          histogram.buckets[bucketIndex].count;
+        binY += intersect > 0 ? count : 0;
+      } else {
+        const isFinalBin = binRight >= max;
+        const singleValueOverlap = (binLeft <= bucketLeft
+            && (isFinalBin ? bucketRight <= binRight : bucketRight < binRight));
+        binY += singleValueOverlap ? histogram.buckets[bucketIndex].count : 0;
+      }
       // If `bucketRight` is bigger than `binRight`, then this bin is
       // finished and there is data for the next bin, so don't increment
       // `bucketIndex`.```

File: tensorboard/plugins/histogram/tf_histogram_dashboard/histogram_core_test.ts [hunk 2]
```diff
@@ -0,0 +1,141 @@
...
+        step: 0,
+        min: undefined,
+        max: undefined,
+        buckets: [],
+      });
+    });
+
+    it('converts backend histogram to intermediate', () => {
+      const bins: BackendHistogramBin[] = [
+        [2, 3, 1],
```
nfelt, line 37: This order is a little odd, since in real data it would presumably be `[1, 2, 2], [2, 3, 1], [3, 4, 1]` so the bucket boundaries are in ascending order?

nfelt (commented):
Thank you for adding the tests!  Main suggestion is just to have the logic look more like the Time Series logic (that you're already adjusting in https://github.com/tensorflow/tensorboard/pull/5392).

tensorboard/plugins/histogram/tf_histogram_dashboard/histogram_core_test.ts
yatbear: Fixed.

tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
yatbear: Thanks for the further explanation and clarifying that we should focus on the visualization purpose here!

- For the single data case, `min` would equal `max`, and it will never enter the `d3.range(min, max, binWidth)` loop, replaced `d3.range` with a for loop.
- Also applied an offset to `minmin` and `maxmax` in `backendToVz()` if they are equal like the timeseries approach.

tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
```diff
@@ -86,10 +86,22 @@ export function intermediateToD3(
   max: number,
   numBins = 30
 ): D3HistogramBin[] {
+  // Empty case.
```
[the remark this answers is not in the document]
yatbear: Done, also updated the empty test cases.

tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
```diff
@@ -86,10 +86,22 @@ export function intermediateToD3(
   max: number,
```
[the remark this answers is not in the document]
yatbear: Done.

tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
nfelt: Just curious, is there a reason to apply the offsetting to minmin and maxmax in `backendToVz()` instead of here in `intermediateToD3()`, which is where it was before?  It seems simpler to just revert to what's currently implemented, which matches timeseries:

https://github.com/tensorflow/tensorboard/blob/165f2cbd2a44cc74d041223653cc447666e07ec8/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts#L89-L93

If we put this logic below the `if (min === undefined || max == undefined) {` block that you added, then we will also handle that case, rather than leaving one edge case in which `min === max` (since they're both remapped from undefined to 0).  I suppose if you prefer, we could move the undefined handling into `backendToVz()` and then re-tighten the arguments here so that they can't be `undefined` after all - but then we still kind of need some check here to confirm that the arguments aren't equal, which is why doing it here makes a little more sense to me.

That way, we can ensure that `intermediateToD3()` is consistently always returning bins that aren't zero-width, which is easier for the visualization to meaningfully interpret.

yatbear: Thanks! For some reason after the last iteration I forgot this `[*|/] 1.1 [+|-] 1` existed here and thought it only existed in the Timeseries one.

nfelt (approved):

tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
```diff
@@ -121,14 +137,15 @@ export function intermediateToD3(
...
-    return {x: binLeft, dx: binWidth, y: binY};
-  });
+    d3HistogramBins.push({x: binLeft, dx: binWidth, y: binY});
+  }
+  return d3HistogramBins;
 }
 
 export function backendToVz(histograms: BackendHistogram[]): VzHistogram[] {
   const intermediateHistograms = histograms.map(backendToIntermediate);
-  const minmin = d3.min(intermediateHistograms, (h) => h.min);
```
nfelt, line 130: Just noticed, but this change could be reverted now (they can stay `const`).

yatbear: Changed back, thanks!

PR #5393, opened 2021-11-01
bug: prevent violent error when Experiment metadata is missing

stephanwlee:
ui_selector which provides a convenience selector for getting the run
selection was throwing a JavaScript exception violently when experiment
metadata did not load but runs list was loaded.

Above case can happen when the backend implements /data/experiments
incorrectly and inadvertently apply `limit` on the list of experiments
passed. As such, TensorBoard may not fetch an experiment metadata ever
when list of experiment ids go above `N`.

In theory, there should be a fix to the backend, but we also do not want
the web code to violently throw and henceforth break the app completely.
Especially since the web code never makes use of the experiment name
anymore when filtering, we can easily guard against the case, hence the
change.

### DIFF
--- a/tensorboard/webapp/util/ui_selectors_test.ts
+++ b/tensorboard/webapp/util/ui_selectors_test.ts
@@ -266,7 +266,7 @@ describe('ui_selectors test', () => {
         );
       });
 
-      it('does not violently throw when an experiment metadata DNE', () => {
+      it('does not violently throw when an experiment metadata is null', () => {
         const state = {
           ...buildStateFromAppRoutingState(
             buildAppRoutingState({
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/util/ui_selectors_test.ts
```diff
@@ -265,6 +265,63 @@ describe('ui_selectors test', () => {
           ])
         );
       });
+
+      it('does not violently throw when an experiment metadata DNE', () => {
```
bmd3k, line 269: What is DNE?

bmd3k (approved):

tensorboard/webapp/util/ui_selectors_test.ts
stephanwlee: Does not exist. It was terminology I used when doing math. If it is too obscure, I will change it. Let me know what you prefer.

bmd3k: Yes it seems a little obscure.

stephanwlee: Done.

PR #5396, opened 2021-11-03
DataFrame API: move ExperimentFromDev._pivot_dataframe to a separate utils.py

andrewluchen:
`_pivot_dataframe` doesn't depend on `ExperimentFromDev`. Moving it to a separate module allows it to be re-used outside of `ExperimentFromDev`.

### DIFF
--- a/tensorboard/data/experimental/experiment_from_dev.py
+++ b/tensorboard/data/experimental/experiment_from_dev.py
@@ -19,9 +19,9 @@ import sys
 import time
 
 import grpc
-import numpy as np
 
 from tensorboard.data.experimental import base_experiment
+from tensorboard.data.experimental import utils as experimental_utils
 from tensorboard.uploader import auth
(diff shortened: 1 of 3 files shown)


tensorboard/data/experimental/experiment_from_dev.py
```diff
@@ -22,6 +22,7 @@
 import numpy as np
```
nfelt, line 22: This import is unused now; mind removing it and the placeholder `expect_numpy_installed` dep from the BUILD target?  That should fix the presubmit check.  If you re-request my review after that's updated, I can re-trigger the presubmits.

nfelt (approved):
LGTM

tensorboard/data/experimental/experiment_from_dev.py
andrewluchen: Done!

PR #5399, opened 2021-11-04
fix build - - with dirty commits

yatbear:
Check for deep equality in the test.

Failed run: https://github.com/tensorflow/tensorboard/runs/4106578999?check_suite_focus=true
```
Chrome Headless 84.0.4147.0 (Linux x86_64) metrics main view pinned view new pinned indicator shows the indicator when the same card gets pinned toggled FAILED
	Error: Expected <span _ngcontent-a-c320 class="new-card-pinned" data-id="1000">...</span> not to be <span _ngcontent-a-c320 class="new-card-pinned" data-id="1000">...</span>. Tip: To check for deep equality, use .toEqual() instead of .toBe().
	    at <Jasmine>
	    at UserContext.<anonymous> (tensorboard/webapp/metrics/views/main_view/main_view_test.ts:1615:56 <- org_tensorflow_tensorboard/tensorboard/webapp/metrics/views/main_view/main_view_test.js:1199:68)
	    at ZoneDelegate.invoke (node_modules/zone.js/dist/zone-testing-bundle.js:407:30)
	    at ProxyZoneSpec.onInvoke (node_modules/zone.js/dist/zone-testing-bundle.js:3765:43)

	Error: Expected '1000' not to be '1000'.
	    at <Jasmine>
	    at UserContext.<anonymous> (tensorboard/webapp/metrics/views/main_view/main_view_test.ts:1618:64 <- org_tensorflow_tensorboard/tensorboard/webapp/metrics/views/main_view/main_view_test.js:1200:76)
	    at ZoneDelegate.invoke (node_modules/zone.js/dist/zone-testing-bundle.js:407:30)
	    at ProxyZoneSpec.onInvoke (node_modules/zone.js/dist/zone-testing-bundle.js:3765:43)
```

#oncall

### DIFF
--- a/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
+++ b/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
@@ -90,6 +90,12 @@ export function intermediateToD3(
     min = 0;
     max = 0;
   }
+  if (max === min) {
+    // If the output range is 0 width, use a default non 0 range for
+    // visualization purpose.
+    max = min * 1.1 + 1;
+    min = min / 1.1 - 1;
+  }
   // Terminology note: _buckets_ are the input to this function,
   // while _bins_ are our output.
   const binWidth = (max - min) / numBins;
@@ -140,12 +146,6 @@ export function backendToVz(histograms: BackendHistogram[]): VzHistogram[] {
   const intermediateHistograms = histograms.map(backendToIntermediate);
   let minmin = d3.min(intermediateHistograms, (h) => h.min);
   let maxmax = d3.max(intermediateHistograms, (h) => h.max);
-  // If the output range is 0 width, use a default non 0 range for
-  // visualization purpose.
-  if (minmin !== undefined && maxmax !== undefined && minmin === maxmax) {
-    maxmax = maxmax * 1.1 + 1;
-    minmin = minmin / 1.1 - 1;
-  }
   return intermediateHistograms.map((h) => ({
     wall_time: h.wall_time,
     step: h.step,
--- a/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
+++ b/tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
@@ -28,8 +28,8 @@ export type BackendHistogram = [
 export type IntermediateHistogram = {
   wall_time: number; // in seconds
   step: number;
-  min: number;
-  max: number;
+  min: number | undefined;
+  max: number | undefined;
   buckets: {
     left: number;
     right: number;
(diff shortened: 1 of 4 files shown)


tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
```diff
@@ -82,20 +82,27 @@ export function backendToIntermediate(
...
+    // visualization purpose.
     max = min * 1.1 + 1;
     min = min / 1.1 - 1;
   }
   // Terminology note: _buckets_ are the input to this function,
   // while _bins_ are our output.
   const binWidth = (max - min) / numBins;
   let bucketIndex = 0;
-  return d3.range(min, max, binWidth).map((binLeft) => {
+  let d3HistogramBins: D3HistogramBin[] = [];
```
stephanwlee, line 103: let -> const. You are not changing the reference to the array but rather pushing value into the same array instance.

tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
```diff
@@ -107,12 +114,21 @@ export function intermediateToD3(
...
+          (intersect / (bucketRight - bucketLeft)) *
+          histogram.buckets[bucketIndex].count;
+        binY += intersect > 0 ? count : 0;
+      } else {
+        const isFinalBin = binRight >= max;
+        const singleValueOverlap =
+          binLeft <= bucketLeft &&
+          (isFinalBin ? bucketRight <= binRight : bucketRight < binRight);
+        binY += singleValueOverlap ? histogram.buckets[bucketIndex].count : 0;
+      }
```
stephanwlee, line 131: Is this change important for fixing the build breakage or is this something else? If it is related, would you mind telling me how this is related?

tensorboard/webapp/metrics/views/main_view/main_view_test.ts
```diff
@@ -1612,10 +1612,10 @@ describe('metrics main view', () => {
         ]);
         const card2IndicatorAfter = fixture.debugElement.query(byCss.INDICATOR);
 
-        expect(card2IndicatorBefore.nativeElement).not.toBe(
+        expect(card2IndicatorBefore.nativeElement).not.toEqual(
           card2IndicatorAfter.nativeElement
         );
-        expect(card2IndicatorBefore.attributes['data-id']).not.toBe(
+        expect(card2IndicatorBefore.attributes['data-id']).not.toEqual(
```
stephanwlee, line 1618: I think the first `not.toBe` was more equal since, while the deep values can be equal, I did not want two DOM nodes to be of the same instance. How often does this test fail?

tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
yatbear: Thanks for this comment! I will apply it in https://github.com/tensorflow/tensorboard/pull/5391.

tensorboard/plugins/histogram/tf_histogram_dashboard/histogramCore.ts
yatbear: This was mixed with another branch for some reason. Sorry for the confusion!

PR #5404, opened 2021-11-05
histogram: remove `backwards` buckets in v1 histogram migration

yatbear:
Remove empty buckets on both ends to avoid having `backwards` (left edge > right edge) buckets.

Ran the change internally: cl/407885499

Visualization change:
- before: 
image
- after: 
image

#histogram #tpu

### DIFF
--- a/tensorboard/data_compat.py
+++ b/tensorboard/data_compat.py
@@ -83,33 +83,42 @@ def make_summary(tag, metadata, data):
 def _migrate_histogram_value(value):
     """Convert `old-style` histogram value to `new-style`.
 
-    Since by default min value is DBL_MAX and max value is -DBL_MAX, empty
-    buckets on the left and right ends are removed to prevent the case of
-    `backwards` (left edge > right edge) buckets.
+    The "old-style" format can have outermost bucket limits of -DBL_MAX and
+    DBL_MAX, which are problematic for visualization. We replace those here
+    with the actual min and max values seen in the input data, but then in
+    order to avoid introducing "backwards" buckets (where left edge > right
+    edge), we first must drop all empty buckets on the left and right ends.
     """
+    summary_metadata = histogram_metadata.create_summary_metadata(
+        display_name=value.metadata.display_name or value.tag,
+        description=value.metadata.summary_description,
+    )
+
     histogram_value = value.histo
     bucket_counts = histogram_value.bucket
     # Find the indices of the leftmost and rightmost non-empty buckets.
     n = len(bucket_counts)
     start = next((i for i in range(n) if bucket_counts[i] > 0), n)
-    end = next((i for i in range(n - 1, -1, -1) if bucket_counts[i] > 0), -1)
-    # Discard empty buckets on both ends.
-    bucket_lefts = histogram_value.bucket_limit[start:end]
-    bucket_rights = histogram_value.bucket_limit[start:end]
+    end = next((i for i in reversed(range(n)) if bucket_counts[i] > 0), -1)
+    if start > end:
+        # If all input buckets were empty, treat it as a zero-bucket
+        # new-style histogram.
+        return make_summary(
+            value.tag, summary_metadata, np.zeros([0, 3], dtype=np.float32)
+        )
+    # Discard empty buckets on both ends, and keep only the "inner" edges
+    # from the remaining buckets. Note that bucket indices range from `start`
+    # to `end` inclusive, but bucket_limit indices are exclusive of `end` -
+    # this is because bucket_limit[i] is the right-hand edge for bucket[i].
     bucket_counts = bucket_counts[start : end + 1]
-    if start <= end:
-        # Use min as the left-hand limit for the first non-empty bucket.
-        bucket_lefts = [histogram_value.min] + bucket_lefts
-        # Use max as the right-hand limit for the last non-empty bucket.
-        bucket_rights = bucket_rights + [histogram_value.max]
+    inner_edges = histogram_value.bucket_limit[start:end]
+    # Use min as the left-hand limit for the first non-empty bucket.
+    bucket_lefts = [histogram_value.min] + inner_edges
+    # Use max as the right-hand limit for the last non-empty bucket.
+    bucket_rights = inner_edges + [histogram_value.max]
     buckets = np.array(
         [bucket_lefts, bucket_rights, bucket_counts], dtype=np.float32
     ).transpose()
-
-    summary_metadata = histogram_metadata.create_summary_metadata(
-        display_name=value.metadata.display_name or value.tag,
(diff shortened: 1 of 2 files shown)


tensorboard/data_compat_test.py
```diff
@@ -201,10 +201,76 @@ def test_histogram(self):
...
+        self.assertEqual(expected_metadata, new_value.metadata)
+        self.assertTrue(new_value.HasField("tensor"))
+        buckets = tensor_util.make_ndarray(new_value.tensor)
+        # Only one bucket is kept.
+        self.assertEqual((1, 3), buckets.shape)
+        self.assertEqual(1, buckets[0][0])
+        self.assertEqual(1, buckets[-1][1])
+        self.assertEqual(1024, buckets[0][2])
+
+    def test_histogram_with_empty_buckets_on_both_ends(self):
```
nfelt, line 251: It might be worth an additional test case using a histogram with data that has extremal values, e.g. [-1e20, 1e20], which should produce counts in the "farthest out" buckets that the legacy histogram format can generate.  In particular, in that case the final bucket in the legacy histogram format is non-empty (whereas usually it's empty).

tensorboard/data_compat.py
```diff
@@ -81,10 +81,27 @@ def make_summary(tag, metadata, data):
 
 
 def _migrate_histogram_value(value):
+    """Convert `old-style` histogram value to `new-style`.
+
+    Since by default min value is DBL_MAX and max value is -DBL_MAX, empty
```
nfelt, line 86: I might phrase this a little differently - in particular, "min" and "max" here might be easy to confuse with the legacy histogram format's actual min and max fields (which are derived from the data), but I think you're actually referring to the minimum and maximum bucket limits?  (Also, I think you have -DBL_MAX and DBL_MAX swapped.)

Maybe something like:

> The "old-style" format can have outermost bucket limits of -DBL_MAX and DBL_MAX,
> which are problematic for visualization. We replace those here with the actual min and
> max values seen in the input data, but then in order to avoid introducing "backwards"
> buckets (where left edge > right edge), we first must drop all empty buckets on the
> left and right ends.

nfelt (commented):
Thanks for taking this on!  The logic itself looks good to me, just some suggestions to make it clearer.

tensorboard/data_compat.py
```diff
@@ -81,10 +81,27 @@ def make_summary(tag, metadata, data):
...
+    `backwards` (left edge > right edge) buckets.
+    """
     histogram_value = value.histo
-    bucket_lefts = [histogram_value.min] + histogram_value.bucket_limit[:-1]
-    bucket_rights = histogram_value.bucket_limit[:-1] + [histogram_value.max]
     bucket_counts = histogram_value.bucket
+    # Find the indices of the leftmost and rightmost non-empty buckets.
+    n = len(bucket_counts)
+    start = next((i for i in range(n) if bucket_counts[i] > 0), n)
+    end = next((i for i in range(n - 1, -1, -1) if bucket_counts[i] > 0), -1)
```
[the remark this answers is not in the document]
yatbear: Done. Thanks!

tensorboard/data_compat.py
```diff
@@ -81,10 +81,27 @@ def make_summary(tag, metadata, data):
...
+    # Discard empty buckets on both ends.
+    bucket_lefts = histogram_value.bucket_limit[start:end]
+    bucket_rights = histogram_value.bucket_limit[start:end]
+    bucket_counts = bucket_counts[start : end + 1]
+    if start <= end:
```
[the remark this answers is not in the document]
yatbear: Done.

tensorboard/data_compat.py
```diff
@@ -81,10 +81,27 @@ def make_summary(tag, metadata, data):
...
+    """
     histogram_value = value.histo
-    bucket_lefts = [histogram_value.min] + histogram_value.bucket_limit[:-1]
-    bucket_rights = histogram_value.bucket_limit[:-1] + [histogram_value.max]
     bucket_counts = histogram_value.bucket
+    # Find the indices of the leftmost and rightmost non-empty buckets.
+    n = len(bucket_counts)
+    start = next((i for i in range(n) if bucket_counts[i] > 0), n)
+    end = next((i for i in range(n - 1, -1, -1) if bucket_counts[i] > 0), -1)
+    # Discard empty buckets on both ends.
```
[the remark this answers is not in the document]
yatbear: Done, thanks for improving the readability!

tensorboard/data_compat_test.py
yatbear: Added.

tensorboard/data_compat.py
yatbear: This is referring to https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/lib/histogram/histogram.cc#L93, when `_min` and `_max` aren't set given empty buckets, I think the limits could end up being [DBL_MAX, -DBL_MAX], but I agree that it's very confusing, adopted the suggested doc string, thanks!

tensorboard/data_compat.py
```diff
@@ -81,19 +81,45 @@ def make_summary(tag, metadata, data):
...
+    histogram_value = value.histo
+    bucket_counts = histogram_value.bucket
+    # Find the indices of the leftmost and rightmost non-empty buckets.
+    n = len(bucket_counts)
+    start = next((i for i in range(n) if bucket_counts[i] > 0), n)
+    end = next((i for i in reversed(range(n)) if bucket_counts[i] > 0), -1)
+    if start > end:
+        # If all input buckets were empty, treat it as a zero-bucket
+        # new-style histogram.
+        return make_summary(
```
nfelt, line 106: Ah right, I forgot that we need to wrap this in `make_summary()` in two places if we do an early return here. Alternatively we could do something like:

```py
if start > end:
    buckets = np.zeros([0, 3], dtype=np.float32)
else:
    # rest of normal codepath goes here
    buckets = np.array([...], dtype=np.float32).transpose()
return make_summary(value.tag, summary_metadata, buckets)
```

Up to you!

tensorboard/data_compat.py
nfelt: Ohhh right I understand now, sorry for the confusion! But yes, since in the all-empty histogram case now, we should be ignoring the legacy format's min and max anyway (since we just produce the [0, 3] empty tensor), I think the fact that those values might be [DBL_MAX, -DBL_MAX] hopefully shouldn't be an issue.

nfelt (approved):

tensorboard/data_compat.py
yatbear: Changed the sequence to avoid moving the code blocks around.

PR #5412, opened 2021-11-09
Time Series: preserve run regex filter string in URL

japie1235813:
Time Series does not remember regex query put in the run filter. We remember this by adding it to the URL.

Screen Shot 2021-11-10 at 10 28 58 AM

### DIFF
--- a/tensorboard/webapp/routes/dashboard_deeplink_provider_types.ts
+++ b/tensorboard/webapp/routes/dashboard_deeplink_provider_types.ts
@@ -27,3 +27,5 @@ export const PINNED_CARDS_KEY = 'pinnedCards';
 export const RUN_COLOR_GROUP_KEY = 'runColorGroup';
 
 export const TAG_FILTER_KEY = 'tagFilter';
+
+export const RUN_REGEX_FILTER_KEY = 'runRegexFilter'
--- a/tensorboard/webapp/routes/dashboard_deeplink_provider_types.ts
+++ b/tensorboard/webapp/routes/dashboard_deeplink_provider_types.ts
@@ -28,4 +28,4 @@ export const RUN_COLOR_GROUP_KEY = 'runColorGroup';
 
 export const TAG_FILTER_KEY = 'tagFilter';
 
-export const RUN_REGEX_FILTER_KEY = 'runRegexFilter';
+export const RUN_FILTER_KEY = 'runFilter';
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/routes/dashboard_deeplink_provider_types.ts
```diff
@@ -27,3 +27,5 @@ export const PINNED_CARDS_KEY = 'pinnedCards';
 export const RUN_COLOR_GROUP_KEY = 'runColorGroup';
 
 export const TAG_FILTER_KEY = 'tagFilter';
+
+export const RUN_REGEX_FILTER_KEY = 'runRegexFilter';
```
stephanwlee, line 31: Perhaps simpler, `runFilter` to match `tagFilter`? They are both regexes.

tensorboard/webapp/runs/store/runs_reducers_test.ts
```diff
@@ -1130,6 +1134,77 @@ describe('runs_reducers', () => {
...
+      };
+      const nextState = runsReducers.reducers(
+        state,
+        stateRehydratedFromUrl({
+          routeKind: RouteKind.EXPERIMENT,
+          partialState,
+        })
+      );
+
+      expect(nextState.data.regexFilter).toEqual('hello');
```
stephanwlee, line 1157: toEqual -> toBe. Here and below.

stephanwlee (approved):

tensorboard/webapp/routes/dashboard_deeplink_provider_types.ts
japie1235813: Isn't `tagFilter ` related to another filter? (where the arrow points at)
Screen Shot 2021-11-10 at 10 53 29 AM
t)

stephanwlee: So, in L29, we have "tagFilter" instead of "tagRegexFilter" even though tag filter also let you specify a regex.

Likewise, I am asking  if we can drop word" regex" from "runRegexFilter" like we did for tag one.

japie1235813: ah I see. I thought you want to rename `runRegexFilter` to `tag..`

PR #5415, opened 2021-11-11
histogram: set v3 as default

yatbear:
Set V3 implementation as default, dynamic shapes will be eliminated.

Ran internally: cl/408885623

#histogram #tpu

### DIFF
--- a/tensorboard/plugins/histogram/summary.py
+++ b/tensorboard/plugins/histogram/summary.py
@@ -16,20 +16,13 @@
 
 A histogram summary stores a list of buckets. Each bucket is encoded as
 a triple `[left_edge, right_edge, count]`. Thus, a full histogram is
-encoded as a tensor of dimension `[k, 3]`, where the first `k - 1` buckets
-are closed-open and the last bucket is closed-closed.
+encoded as a tensor of dimension `[k, 3]`.
 
 In general, the value of `k` (the number of buckets) will be a constant,
-like 30. For V2 format (deprecated), there are two edge cases: if there
-is no data, then there are no buckets (the shape is `[0, 3]`); and if
-there is data but all points have the same value, then there is one
-bucket whose left and right endpoints are the same (the shape is `[1, 3]`).
-
-For V3 format, the shape of the output histogram is always constant (`[k, 3]`).
-In the case of empty data, the output will be an all-zero histogram of shape
-`[k, 3]`, where all edges and counts are zeros. If there is data but all points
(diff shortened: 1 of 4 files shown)


tensorboard/plugins/histogram/summary.py
```diff
@@ -16,13 +16,20 @@
 
 A histogram summary stores a list of buckets. Each bucket is encoded as
 a triple `[left_edge, right_edge, count]`. Thus, a full histogram is
-encoded as a tensor of dimension `[k, 3]`.
+encoded as a tensor of dimension `[k, 3]`, where the first `k - 1` buckets
```
nfelt, line 19: This module docstring is all referring to the legacy "tb.summary" APIs (for TensorFlow 1.x) that were defined in this file, which we aren't changing, so I think it's ok to just leave this docstring as-is (and only have the new format documented in summary_v2.py's docstring).  At some point we should rearrange and clean this up so that's more obvious but probably best to do that for all the summary APIs vs piecemeal.

nfelt (commented):
Actually, if you don't mind, I'd recommend that in this PR we just do _only_ the cutover - i.e. let's not remove the v2 codepath yet.  Just in case there are issues and we have to roll back, that makes the rollback as minimal as possible and avoids extra churn in the code.

For example, I think it would be enough to just do the following:
- in summary_v2.py
  - rename `histogram` to `histogram_v2`
  - add `histogram = histogram_v3` at the bottom
- in summary_test.py
  - update SummaryV2OpTest.call_histogram_op to invoke `histogram_v2` instead of just `histogram`

tensorboard/plugins/histogram/summary.py
yatbear: Done.

tensorboard/plugins/histogram/summary.py
```diff
@@ -36,12 +36,12 @@
...
 
 # Export V2 versions.
+histogram_v2 = summary_v2.histogram_v2
+
+# Export the default versions.
 histogram = summary_v2.histogram
 histogram_pb = summary_v2.histogram_pb
 
-# Export V3 versions.
-histogram_v3 = summary_v2.histogram_v3
```
nfelt, line 43: Let's keep this one line for now, so that anyone who was testing this out internally won't be broken by this change?

nfelt (approved):

tensorboard/plugins/histogram/summary.py
yatbear: Done.

PR #5426, opened 2021-11-17
Time Series: fix dark mode color

japie1235813:
Fix dark mode color for collapsible groups

before 
Screen Shot 2021-11-16 at 11 22 33 PM

after 
Screen Shot 2021-11-16 at 11 49 26 PM
light mode
Screen Shot 2021-11-17 at 9 22 08 AM

### DIFF
--- a/tensorboard/webapp/metrics/views/main_view/card_group_toolbar_component.scss
+++ b/tensorboard/webapp/metrics/views/main_view/card_group_toolbar_component.scss
@@ -34,11 +34,16 @@ limitations under the License.
       cursor: pointer;
     }
   }
+
+  @include tb-dark-theme {
+    .group-toolbar {
+      color: map-get($tb-dark-foreground, text);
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/metrics/views/main_view/card_group_toolbar_component.scss
```diff
@@ -34,11 +34,16 @@ limitations under the License.
       cursor: pointer;
     }
   }
+
+  @include tb-dark-theme {
+    .group-toolbar {
+      color: map-get($tb-dark-foreground, text);
+    }
+  }
 }
```
stephanwlee, line 43: First, you are no longer allowed to use `map-get`. 
Second: can you just do `@include tb-theme-foreground-prop(color, text);` in L24?

stephanwlee (approved):

PR #5427, opened 2021-11-17
timeseries: define a card width slider

yatbear:
This UI change is currently hidden behind a flag. Note that:

- Currently `cardMaxWidthInVW` is hardcoded to `30` in `settings_view_component`, this value should be obtained from the corresponding metrics selector.
- The global persistent settings need to be defined in `metrics_module` using the selector for `cardMaxWidthInVW`.

Preview: 
image

image


tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
```diff
@@ -302,6 +311,53 @@ describe('metrics right_pane', () => {
...
+          TEST_ONLY.MIN_CARD_WIDTH_SLIDER_VALUE.toString()
+        );
+      });
+
+      it('dispatches action when adjusting the slider', fakeAsync(() => {
+        const fixture = TestBed.createComponent(SettingsViewContainer);
+        fixture.detectChanges();
+        const slider = select(fixture, CARD_WIDTH_SLIDER);
+
+        // Adjust slider.
```
japie1235813, line 339: `Adjusts` and I think the code should speak for itself so no need to comment here.

tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
```diff
@@ -302,6 +311,53 @@ describe('metrics right_pane', () => {
...
+        slider.triggerEventHandler('input', {value: 90});
+        tick(TEST_ONLY.SLIDER_AUDIT_TIME_MS);
+
+        expect(dispatchSpy).toHaveBeenCalledOnceWith(
+          actions.metricsChangeCardWidth({cardMaxWidthInVW: 90})
+        );
+      }));
+
+      it('dispatches action when clicking reset', () => {
```
japie1235813, line 348: "dispatches `metricsResetCardWidth` action.."

tensorboard/webapp/metrics/views/right_pane/settings_view_component.scss
```diff
@@ -39,6 +39,7 @@ section .control-row:not(:last-child) {
 }
 
 .control-name {
+  margin-top: 16px;
```
japie1235813, line 42: Why the margin top is needed? Isn't the current css setting not enough?

nfelt:
Cool!  Just my $0.02, but I would perhaps use a slider that doesn't show the actual numerical value? Since it won't be very obvious to the user what the meaning of the number is, and leaving it unspecified suggests that the setting is more about the overall "feel" versus a precise value, which gives us more flexibility to adjust the behavior later if need be.

yatbear:
> Cool! Just my $0.02, but I would perhaps use a slider that doesn't show the actual numerical value? Since it won't be very obvious to the user what the meaning of the number is, and leaving it unspecified suggests that the setting is more about the overall "feel" versus a precise value, which gives us more flexibility to adjust the behavior later if need be.

Sounds good, removed the labels, and thanks for your $maxint!

tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
```diff
@@ -89,6 +89,7 @@ describe('metrics right_pane', () => {
       );
       store.overrideSelector(selectors.getIsFeatureFlagsLoaded, true);
       store.overrideSelector(selectors.getIsMetricsImageSupportEnabled, true);
+      store.overrideSelector(selectors.getEnabledCardWidthSetting, false);
```
japie1235813, line 92: Could you move this down a line (line 93) to make it more organized? (I would prefer alphabetically sorting but that's already not the case)

tensorboard/webapp/metrics/actions/index.ts
```diff
@@ -70,6 +70,15 @@ export const metricsChangeXAxisType = createAction(
   props<{xAxisType: XAxisType}>()
 );
 
+export const metricsChangeCardWidth = createAction(
+  '[Metrics] Metrics Setting Change Card Width',
+  props<{cardMaxWidthInVW: number}>()
+);
+
+export const metricsResetCardWidth = createAction(
+  '[Metrics] Card Width Setting Reset'
```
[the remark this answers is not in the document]
yatbear: Got it, changed to `[Metrics] Metrics Setting Reset Card Width`.

tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
yatbear: Done.

tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
yatbear: Done.

tensorboard/webapp/metrics/views/right_pane/settings_view_component.scss
yatbear: For some reason, earlier without this there was no space between the slider and axis drop-down, it must be some unintentional bug from the initial version. Removed.

tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
yatbear: Done.

japie1235813 (approved):

PR #5428, opened 2021-11-17
Time Series: adjust card width (UI)

japie1235813:
Some tag name on cards might be too long and be truncated. We want to provide users a setting to adjust the card width. This pr only includes the part of metrics/ and the card with would be set based on the state value.

Only update the max width of the card in css. The range of the max width is 30-100vw.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_selectors.ts
+++ b/tensorboard/webapp/metrics/store/metrics_selectors.ts
@@ -280,9 +280,9 @@ export const getMetricsSettingOverrides = createSelector(
   }
 );
 
-export const getMetricCardMaxWidthMultiplier = createSelector(
+export const getMetricCardMaxWidth = createSelector(
   selectSettings,
-  (settings): number => settings.cardMaxWidthMultiplier
+  (settings): number => settings.cardMaxWidthInVW
 );
 
 export const getMetricsTooltipSort = createSelector(
--- a/tensorboard/webapp/metrics/store/metrics_selectors.ts
+++ b/tensorboard/webapp/metrics/store/metrics_selectors.ts
@@ -280,6 +280,11 @@ export const getMetricsSettingOverrides = createSelector(
   }
 );
 
+export const getMetricCardMaxWidthMultiplier = createSelector(
+  selectSettings,
+  (settings): number => settings.cardMaxWidthMultiplier
+);
+
 export const getMetricsTooltipSort = createSelector(
   selectSettings,
   (settings): TooltipSort => settings.tooltipSort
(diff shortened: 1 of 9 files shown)


tensorboard/webapp/metrics/store/metrics_selectors.ts
```diff
@@ -280,6 +280,11 @@ export const getMetricsSettingOverrides = createSelector(
   }
 );
 
+export const getMetricCardMaxWidth = createSelector(
```
yatbear, line 283: Should this be getMetric[s]CardMaxWidth to be consistent with the other ones?

japie1235813: good catch. done

tensorboard/webapp/metrics/store/metrics_types.ts
```diff
@@ -221,6 +222,7 @@ export interface State {
 }
 
 export const METRICS_SETTINGS_DEFAULT: MetricsSettings = {
+  cardMaxWidthInVW: 30,
```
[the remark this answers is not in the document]
japie1235813: Thanks for catching this. It should be null by default

tensorboard/webapp/metrics/views/main_view/card_grid_component.ts
```diff
@@ -49,6 +55,16 @@ export class CardGridComponent {
     @Optional() private readonly cdkScrollable: CdkScrollable | null
   ) {}
 
+  ngOnInit() {
+    if (
+      this.cardMaxWidthInVW &&
+      this.cardMaxWidthInVW >= MIN_CARD_MAX_WIDTH_IN_VW &&
```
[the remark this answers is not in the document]
japie1235813: Short answer is no.  The range of the slider should be 30 to 100 (unit vw). And we will not show the numbers to preserve interpretation (instead of explicitly explaining what the numbers are)

tensorboard/webapp/metrics/views/main_view/main_view_test.ts
```diff
@@ -894,6 +895,75 @@ describe('metrics main view', () => {
...
+        fixture.detectChanges();
+
+        expect(
+          fixture.debugElement.query(By.css('.card-grid')).styles[
+            'grid-template-columns'
+          ]
+        ).toBe('repeat(auto-fill, minmax(335px, 50vw))');
+      });
+
+      it('does not set the max width with invalid width value', () => {
```
bmd3k, line 935: Just for completeness, and given that it will be the default behavior, would you mind adding a test where getEnabledCardWidthSetting is true and getMetricsCardWidth is null?

bmd3k (approved):

PR #5430, opened 2021-11-17
chore: explicitly depend on Angular Material style dep

andrewseguin:
Removes the _angular_material_theming.scss since it is unused. All stylesheets directly import the Angular Material styles through "@use".

This will keep stylesheets from transitively accessing the old Angular Material theming API and instead help enforce "@use". This ended up catching another couple uses of the old API.


andrewseguin:
Woo passed =D

stephanwlee (approved):
Change looks good but I suspect this would require some copybara script change. I will test the sync before merging so others are not blocked. Sorry about that!

andrewseguin:
Sounds good, I figured there might be some tweaks needed for the copybara.

PR #5432, opened 2021-11-18
Time Series: Card width adjustment, connect settings and ui together

japie1235813:
Last piece of our card width adjustment. Take the updated value from setting panel and apply on the card css to change the width.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -646,6 +646,24 @@ const reducer = createReducer(
       },
     };
   }),
+  on(actions.metricsChangeCardWidth, (state, {cardMaxWidthInVW}) => {
+    return {
+      ...state,
+      settingOverrides: {
+        ...state.settingOverrides,
+        cardMaxWidthInVW,
+      },
+    };
+  }),
+  on(actions.metricsResetCardWidth, (state) => {
+    return {
+      ...state,
+      settingOverrides: {
+        ...state.settingOverrides,
+        cardMaxWidthInVW: null,
+      },
+    };
+  }),
   on(
     actions.multipleTimeSeriesRequested,
     (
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -646,6 +646,24 @@ const reducer = createReducer(
       },
     };
   }),
+  on(actions.metricsChangeCardWidth, (state, {cardMaxWidthInVW}) => {
```
yatbear, line 649: (Dunno how to comment lines above, so adding it here) On line 398, do we need to have
```
if (typeof partialSettings.timeSeriesCardMaxWidthInVW === 'number') {
      metricsSettings.timeSeriesCardMaxWidthInVW = partialSettings.timeSeriesCardMaxWidthInVW;
    }
```
like the other ones?

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -773,6 +773,34 @@ describe('metrics reducers', () => {
         HistogramMode.OVERLAY
       );
     });
+
+    it('changes metricsChangeCardWidth on metricsChangeCardWidth', () => {
```
yatbear, line 777: Should the name be `changes  *cardMaxWidthInVW* on metricsChangeCardWidth`?

tensorboard/webapp/metrics/store/metrics_reducers.ts
japie1235813: Unfortunately we didn't preserve the card width setting (yet) so it's meaningless to add them here.

yatbear (approved):

japie1235813:
The last piece of our card width adjustment (almost). Take the updated slider value from setting panel and apply on the card css to change the width.

PR #5434, opened 2021-11-19
timeseries: preserve card width setting in local storage

yatbear:
Tested locally:

Screen Shot 2021-11-19 at 12 06 28 PM

#timeseries


japie1235813:
Thanks for pr!

japie1235813 (approved):

PR #5435, opened 2021-11-22
[time-namespaced state] Generate namespace ids in app_routing_effects.

bmd3k:
As part of the work to migrate to "time-namespaced state" (Googlers, see http://go/tb-timenamespaced-state) we need to refactor namespace id generation from route_contexted_reducer_helper.ts to app_routing_effects.ts.

This requires changing the navigated() action to include the namespace information - beforeNamespaceId and afterNamespaceId. So there is a somewhat high blast radius. I've tried to ensure that the changes will not break the build. I'll also comment on the PR to try to make sense of the changes.

There is also a new property in app_routing state: activeNamespaceId. We use this to populate the beforeNamespaceId value.

### DIFF
--- a/tensorboard/webapp/app_routing/actions/app_routing_actions.ts
+++ b/tensorboard/webapp/app_routing/actions/app_routing_actions.ts
@@ -70,8 +70,10 @@ export const navigating = createAction(
 export interface NavigatedPayload {
   before: Route | null;
   after: Route;
-  beforeNamespaceId: string | null;
-  afterNamespaceId: string;
+  // TODO(bdubois): Remove optionality of these properties when all internal
+  // code is migrated to specify them.
+  beforeNamespaceId?: string | null;
+  afterNamespaceId?: string;
 }
 
 /**
--- a/tensorboard/webapp/app_routing/actions/app_routing_actions.ts
+++ b/tensorboard/webapp/app_routing/actions/app_routing_actions.ts
@@ -70,6 +70,8 @@ export const navigating = createAction(
 export interface NavigatedPayload {
   before: Route | null;
   after: Route;
+  beforeNamespaceId: string | null;
+  afterNamespaceId: string;
 }
 
 /**
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -48,7 +48,10 @@ import {Location} from '../location';
 import {ProgrammaticalNavigationModule} from '../programmatical_navigation_module';
 import {RouteConfigs} from '../route_config';
 import {RouteRegistryModule} from '../route_registry_module';
-import {getActiveRoute} from '../store/app_routing_selectors';
+import {
+  getActiveRoute,
+  getActiveNamespaceId,
+} from '../store/app_routing_selectors';
 import {Navigation, Route, RouteKind, RouteParams} from '../types';
 
 /** @typehack */ import * as _typeHackNgrxEffects from '@ngrx/effects';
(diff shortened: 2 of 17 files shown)


tensorboard/webapp/app_routing/actions/app_routing_actions.ts
```diff
@@ -70,6 +70,10 @@ export const navigating = createAction(
 export interface NavigatedPayload {
   before: Route | null;
   after: Route;
+  // TODO(bdubois): Remove optionality of these properties when all internal
+  // code is migrated to specify them.
+  beforeNamespaceId?: string | null;
```
bmd3k, line 75: Two new properties in navigated() payload. Temporarily they are optional to allow internal build to pass.

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -384,9 +387,17 @@ export class AppRoutingEffects {
...
-        return navigated({before: oldRoute, after: route});
+      withLatestFrom(
+        this.store.select(getActiveRoute),
+        this.store.select(getActiveNamespaceId)
+      ),
+      map(([{route}, oldRoute, oldNamespaceId]) => {
+        return navigated({
+          before: oldRoute,
+          after: route,
+          beforeNamespaceId: oldNamespaceId,
```
bmd3k, line 398: app_routing_effects gets beforeNamespaceId from state and generates the afterNamespaceId from routeId.

tensorboard/webapp/app_routing/store/app_routing_reducers.ts
```diff
@@ -27,8 +28,15 @@ const reducer = createReducer(
...
-  on(actions.navigated, (state, {after}) => {
-    return {...state, activeRoute: after, nextRoute: null};
+  on(actions.navigated, (state, {after, afterNamespaceId}) => {
+    return {
+      ...state,
+      activeRoute: after,
+      nextRoute: null,
+      // TODO(bdubois): Remove the null fallback when afterNamespaceId is no
+      // longer optional.
+      activeNamespaceId: afterNamespaceId || null,
```
bmd3k, line 38: There is new app_routing state: activeNamespaceId that we can use to populate beforeNamespaceId on navigated() events.

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -96,7 +100,7 @@ describe('app_routing_effects', () => {
         },
         {
           routeKind: RouteKind.COMPARE_EXPERIMENT,
-          path: '/compare',
+          path: '/compare/:experimentIds',
```
bmd3k, line 103: Because getRouteId is not called in app_routing_effects, we need to ensure the route ids it is expecting are well-formed. This has a bit of a big blast radius in this test because we now have to generate proper compare route ids.

tensorboard/webapp/app_routing/route_contexted_reducer_helper.ts
```diff
@@ -110,62 +110,96 @@ export function createRouteContextedState<
...
+      // Note: We don't reset to initial state on initial load because we
+      // assume the state already has values from bootstrapping deeplinks and
+      // we should not overwrite the values.
+      nextRoutefulState = routefulInitialState;
+    }
+
+    return {
+      ...state,
+      ...nextRoutefulState,
+      privateRouteContextedState: nextContextedStateCache,
```
stephanwlee, line 164: Would you mind renaming `privateRouteContextedState` to `privateNamespacedState` in this PR? Since it is a private property, it should be safe to do it here.

tensorboard/webapp/app_routing/route_contexted_reducer_helper.ts
```diff
@@ -110,62 +110,96 @@ export function createRouteContextedState<
...
+    let nextRoutefulState = {};
+    // Note: state.privateRouteContextedState always exists in practice except
+    // for in tests.
+    if (
+      afterNamespaceId &&
+      state.privateRouteContextedState?.[afterNamespaceId]
+    ) {
+      // Swap in existing state since it already exists in the cache.
+      nextRoutefulState = state.privateRouteContextedState[afterNamespaceId];
+    } else if (beforeRouteId) {
```
stephanwlee, line 151: I know this is largely a refactor but I want to do a quick audit on where we use routeId. Can this safely change to `beforeNamespaceId` or does that break certain assumptions?

tensorboard/webapp/app_routing/store/app_routing_selectors.ts
```diff
@@ -51,6 +51,13 @@ export const getNextRouteForRouterOutletOnly = createSelector(
   }
 );
 
+export const getActiveNamespaceId = createSelector(
+  getAppRoutingState,
+  (state) => {
```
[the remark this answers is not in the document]
bmd3k: Done.

stephanwlee (approved):

PR #5440, opened 2021-11-29
[time-namespaced state] Add feature flag for time-namespaced state.

bmd3k:
Add a feature flag for the new behavior to namespace state sequentially by time rather than by id.

Googlers, see http://go/tb-timespaced-state.

### DIFF
--- a/tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
+++ b/tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
@@ -45,3 +45,5 @@ export const ENABLE_COLOR_GROUP_BY_REGEX_QUERY_PARAM_KEY =
 export const ENABLE_DARK_MODE_QUERY_PARAM_KEY = 'darkMode';
 
 export const ENABLE_LINK_TIME_PARAM_KEY = 'enableLinkTime';
+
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
```diff
@@ -45,3 +45,5 @@ export const ENABLE_COLOR_GROUP_BY_REGEX_QUERY_PARAM_KEY =
 export const ENABLE_DARK_MODE_QUERY_PARAM_KEY = 'darkMode';
 
 export const ENABLE_LINK_TIME_PARAM_KEY = 'enableLinkTime';
+
+export const ENABLE_TIME_NAMESPACED_STATE = 'enableTimespace';
```
japie1235813, line 49: Could we change it to `enableTimeNamespaced` to be consistent with other names?

japie1235813 (approved):

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_types.ts
bmd3k: Whoops, I did this locally but forgot to push here before merging. I will send out a new PR soon to fix this.

PR #5442, opened 2021-11-30
Time Series: tooltip sort by alphabetical order

japie1235813:
In the list of tooltip sorting method, we remove "default" and replace with "alphabetical order". The alphabetical order is based on the display name.

Screen Shot 2021-11-30 at 1 44 54 PM
Screen Shot 2021-11-30 at 10 38 22 AM

### DIFF
--- a/tensorboard/webapp/metrics/internal_types.ts
+++ b/tensorboard/webapp/metrics/internal_types.ts
@@ -31,6 +31,7 @@ export enum TooltipSort {
   ASCENDING = 'ascending',
   DESCENDING = 'descending',
   NEAREST = 'nearest',
+  ALPHABETICAL = 'Alphabetical',
 }
 
 export enum XAxisType {
--- a/tensorboard/webapp/metrics/internal_types.ts
+++ b/tensorboard/webapp/metrics/internal_types.ts
@@ -27,7 +27,7 @@ export enum PluginType {
 // When editing a value of the enum, please write a backward compatible
 // deserializer in data_source/metrics_data_source.ts.
 export enum TooltipSort {
-  ALPHABETICAL = 'Alphabetical',
+  ALPHABETICAL = 'alphabetical',
   ASCENDING = 'ascending',
   DESCENDING = 'descending',
   NEAREST = 'nearest',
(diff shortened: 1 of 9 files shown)


tensorboard/webapp/metrics/internal_types.ts
```diff
@@ -27,7 +27,7 @@ export enum PluginType {
 // When editing a value of the enum, please write a backward compatible
 // deserializer in data_source/metrics_data_source.ts.
 export enum TooltipSort {
-  DEFAULT = 'default',
+  ALPHABETICAL = 'Alphabetical',
```
stephanwlee, line 30: `ALPHABETICAL = 'alphabetical',` for consistency please.

stephanwlee (commented):
Cool, could you tell me whether changing the setting is "sticky" and is remembered? Also, we may need some migration in case `{tooltipSort: "default"}` is set on local storage to `{tooltipSort: "alphabetical"}`.

japie1235813:
Yes, it's preserved in local storage
Screen Shot 2021-11-30 at 5 20 00 PM
.

for `{tooltipSort: "default"}` under that case the sorting method would be the default (which is `alphabetical`).
Do you feel we need to explicitly change the value in local storage to ` {tooltipSort: "alphabetical"}`? 
(I'm not quite sure how to manually update the local storage when loading, try a couple of things but does not work out)

In new commit I add DEFAULT back and fall to alphabetic when loading the global settings.

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -374,15 +374,18 @@ const reducer = createReducer(
     const metricsSettings: Partial<MetricsSettings> = {};
     if (partialSettings.tooltipSortString) {
       switch (partialSettings.tooltipSortString) {
+        case TooltipSort.DEFAULT:
+          metricsSettings.tooltipSort = TooltipSort.ALPHABETICAL;
+          break;
```
stephanwlee, line 379: alternative: remove these two lines.

stephanwlee (approved):
Tested manually by altering the local storage. Looks like because of the switch statement in the metrics_reducers, everything work as intended. LGTM.

PR #5443, opened 2021-11-30
histogram: clean up v2 code

yatbear:
Remove V2 code.

Ran internally: cl/413690112

#histogram #tpu

### DIFF
--- a/tensorboard/plugins/histogram/summary.py
+++ b/tensorboard/plugins/histogram/summary.py
@@ -35,7 +35,7 @@ from tensorboard.plugins.histogram import metadata
 from tensorboard.plugins.histogram import summary_v2
 
 
-# Export the default versions.
+# Export the V3 versions.
 histogram = summary_v2.histogram
 histogram_pb = summary_v2.histogram_pb
 
--- a/tensorboard/plugins/histogram/summary.py
+++ b/tensorboard/plugins/histogram/summary.py
@@ -35,7 +35,7 @@ from tensorboard.plugins.histogram import metadata
 from tensorboard.plugins.histogram import summary_v2
 
 
-# Export the V3 versions.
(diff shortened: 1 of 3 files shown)


tensorboard/plugins/histogram/summary_v2.py
```diff
@@ -46,217 +38,6 @@
 DEFAULT_BUCKET_COUNT = 30
 
 
-def histogram_v2(name, data, step=None, buckets=None, description=None):
```
nfelt, line 49: Now that we're removing `histogram_v2`, I think it'd make sense to rename `histogram_v3` to just `histogram` (what the v2 one was called prior to the migration). We can then remove the `histogram = histogram_v3` line at the bottom.

When we do that, it might make sense to add a `#`-style comment (not in the docstring, which is public, but just for our own reference) that mentions that this new `histogram` is actually an updated v3 format, not the v2 format, as the filename would imply.

nfelt (approved):
Assumed this was ready for review since it was requested, but could you also mark it non-draft if that's the case?

tensorboard/plugins/histogram/summary_v2.py
yatbear: Renamed and added the comment.

tensorboard/plugins/histogram/summary.py
```diff
@@ -35,12 +35,6 @@
 from tensorboard.plugins.histogram import summary_v2
 
 
-# Export V2 versions.
-histogram_v2 = summary_v2.histogram_v2
-
-# Export V3 versions.
-histogram_v3 = summary_v2.histogram_v3
-
 # Export the default versions.
```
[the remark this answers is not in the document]
yatbear: Done.

PR #5447, opened 2021-12-03
[time-namespaced state] Support generation of time-based namespace ids.

bmd3k:
If enableTimeNamespacedState feature flag is on, generate new time-based namespace ids on any navigation where application declares resetNamespacedState = true.
    
Add support in the routerLink directive for the resetNavigationState property.
    
Googlers, see http://go/tb-time-namespaced-state.


tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -159,6 +160,7 @@ export class AppRoutingEffects {
         options: {
           replaceState: navigationWithAbsolutePath.replaceState,
           browserInitiated: navigationWithAbsolutePath.browserInitiated,
+          resetNamespacedState: navigationWithAbsolutePath.resetNamespacedState,
```
japie1235813, line 163: Given onPopState is not updated yet, would this value be null?

japie1235813, line 163: Given onPopState is not updated yet, is the value null here?

japie1235813 (approved):

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -519,6 +521,144 @@ describe('app_routing_effects', () => {
...
+          actions.navigated({
+            before: null,
+            after: buildRoute({
+              routeKind: RouteKind.EXPERIMENTS,
+              pathname: '/experiments',
+            }),
+            // From getActiveNamespaceId().
+            beforeNamespaceId: null,
+            // Value of before + the 5 milliseconds from tick(5000).
+            afterNamespaceId: Date.now().toString(),
```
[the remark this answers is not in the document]
bmd3k: Sorry for the confusion. The comments were out of date. I updated them.

1. In the first test the value is just the value of the mock clock (Date.now()) without any additional time added.
2. In the second test is the initial value of the mock clock + 5000 milliseconds.

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
bmd3k: Actually this handles merge(onNavigationRequested$, onInit$ and onPopState()) so it might not be null because of onNavigationRequested$.

PR #5449, opened 2021-12-06
[npmi] Added Embedding Projection Container

Sparkier:
Added the header for the embedding projection visualization.

* Screenshots of UI changes
image

`bazel test //tensorboard/webapp:karma_test_chromium_local`


japie1235813:
Thanks for the pr. A couple of questions:
1. Could you state at a higher level why creating a new module "embedding-projection" rather than reuse of the <violin-filter> which looks almost identical except for the metric-filter.
2. What is the role/functionality of the projection component? Is it served as filter or for other usages? 

I'm a bit lack of the context could you help to fill me in?

Sparkier:
Sure, in the end this should look something like:
image
Then one can filter annotations based on this UMAP projection. In my screenshot you see a clothing cluster.

I just started with this very small PR to get back into it again and to make it easy for you. If you want to look into the finished code for this (though without tests and not cleaned up), you can find that [here](https://github.com/Sparkier/tensorboard/tree/embeddings_view).

tensorboard/webapp/plugins/npmi/views/embeddings/embeddings_component.scss
```diff
@@ -65,6 +70,14 @@ limitations under the License.
   justify-content: center;
 }
 
+.run-divider {
+  content: '';
+  width: 100%;
+  height: 3px;
+  overflow: hidden;
+  background-color: mat.get-color-from-palette($tb-foreground, divider);
```
japie1235813, line 78: a-z sorted

tensorboard/webapp/plugins/npmi/views/embeddings/embeddings_component.scss
```diff
@@ -65,6 +70,14 @@ limitations under the License.
   justify-content: center;
 }
 
+.run-divider {
+  content: '';
```
japie1235813, line 74: What is the content you want to put here later?
It's not clear to me how you want to use this element. (From the naming I thought it is just a line to divide the run list below.  If my guess is correct, why not just use border attribute?)

japie1235813 (commented):
Thanks for the content and the repo.
 
Not needed for this pr but if you have better data for me to test your future pr, please let me know. (Currently I'm using the demo data from [readme](https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/npmi/README.md#demo-data)

Also it would be helpful if you can state more about the purpose in the description. (for example, if you add some new files, what is the context and rationale) It would be great and more efficient to review the code with less assumption and guessing. 
For example, I was wondering why not reuse violit-filter because it looks almost identical to your embedding-projection. Only after checking a bit on your repo (still not thorough) then I think it's good to keep them separated.

Thanks!

tensorboard/webapp/plugins/npmi/views/embeddings/embeddings_component.scss
Sparkier: Yes, this is just a line for dividing things. This has been used like that in `main_component` as well. I guess either change it everywhere or do it the same, no? What do you suggest?

Sparkier:
> Thanks for the content and the repo.
> 
> Not needed for this pr but if you have better data for me to test your future pr, please let me know. (Currently I'm using the demo data from [readme](https://github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/npmi/README.md#demo-data)
> 
> Also it would be helpful if you can state more about the purpose in the description. (for example, if you add some new files, what is the context and rationale) It would be great and more efficient to review the code with less assumption and guessing. For example, I was wondering why not reuse violit-filter because it looks almost identical to your embedding-projection. Only after checking a bit on your repo (still not thorough) then I think it's good to keep them separated.
> 
> Thanks!

I do have data I could send you. How can I get that to you?

Yep, will provide more context in the future. I assumed Stephan would review these changes and he would have known more about the project. Now that I know I can provide more context, no problem.

tensorboard/webapp/plugins/npmi/views/embeddings/embeddings_component.scss
japie1235813: I searched a bit but can't find it in `main_component`. Do you mean `grabber`? [link](https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/plugins/npmi/views/main/main_component.scss#L73)

Sparkier:
FYI if I run prettier myself I don't see what the problem is:
image
So it is very hard for me to fix this linter issue.

japie1235813 (approved):
For the data needed to test your future pr, can you simulate the representative data in `npmi_demo_data.py` with instructions on how to test the pr with the data?

Also I think the prettier version is probably different from ci. Could you try running `yarn fix-lint`?

Sparkier:
I will change the demo data in a separate PR. Changed the divider as you suggested.

Sparkier:
@japie1235813, I cannot merge this PR as I do not have the rights to do so. You would need to do that for me.

bmd3k:
I reran the CI and looks good. The approver is OOO so I will merge. Sorry for the delay - thanks for your patience.

PR #5452, opened 2021-12-07
Allow camera to have initial position and look-at vectors.

podlipensky:
Currently in mesh plugin camera is positioned automatically which may not be convenient for applications where the model predicts both scene mesh/point cloud as well as camera position and orientation. This PR will allow to specify initial camera position and orientation as following:
```
const plugin_config = {
  'camera': {
    'position': [0, 0, 0],
    'look_at': [1, 1, 1]
  }
}
```

### DIFF
--- a/tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
+++ b/tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
@@ -50,6 +50,8 @@ export class MeshViewer extends THREE.EventDispatcher {
   private _animationFrameIndex?: number;
 
   private _camera?: THREE.PerspectiveCamera;
+  private _init_camera_position?: THREE.Vector3;
+  private _init_camera_look_at?: THREE.Vector3;
   private _cameraControls?: OrbitControls;
   private _renderer: THREE.WebGLRenderer;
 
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
```diff
@@ -50,6 +50,8 @@ export class MeshViewer extends THREE.EventDispatcher {
   private _animationFrameIndex?: number;
 
   private _camera?: THREE.PerspectiveCamera;
+  private _init_camera_position?: THREE.Vector3;
+  private _init_camera_look_at?: THREE.Vector3;
```
stephanwlee, line 54: JavaScript/TypeScript use camelCase instead of snake_case. Here and below. Please fix.

tensorboard/plugins/mesh/tf_mesh_dashboard/mesh-viewer.ts
```diff
@@ -411,11 +428,15 @@ export class MeshViewer extends THREE.EventDispatcher {
...
     // Set camera position and orientation.
-    this.setCameraViewpoint(
-      {x: center.x, y: center.y, z: camera_z},
-      camera_to_far_edge * 3,
-      center
-    );
+    let camera_position = new THREE.Vector3(center.x, center.y, camera_z);
+    if (this._init_camera_position) {
+      camera_position = this._init_camera_position;
+    }
```
stephanwlee, line 434: FYI, this can be rewritten as: 
```ts
const cameraPosition = this._init_camera_position ?? new THREE.Vector3(center.x, center.y, camera_z);
```
This syntactic sugar is not required.

stephanwlee (approved):

PR #5454, opened 2021-12-09
[time-namespaced state] Fix tags and pinned cards on experiments adding/removal

japie1235813:
Previously within a namespace, when experiments are added or removed the load tag metadata is not fired. Now we fire the loading on changes of route id. The key change for this pr is cardMetadataMap is reset. The map contains data from previous route and, because the namespace does not change, the map is outdated (includes card id of removed experiments). It is safe to reset the map because the metadata loaded action is fired after receiving complete tag metadata.

Other corresponding state updated include nextCardToPinnedCopy nextPinnedCardToOriginal.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -53,7 +53,7 @@ import {
   getCardId,
   getRunIds,
   getTimeSeriesLoadable,
-  updateCardMap,
+  updateCardMaps,
 } from './metrics_store_internal_utils';
 import {
   CardMetadataMap,
@@ -486,8 +486,6 @@ const reducer = createReducer(
         images: tagMetadata[PluginType.IMAGES],
       };
 
-      // Reset cardMetadataMap because metadata loaded action fired after
-      // complete tag metadata is received.
       const newCardMetadataMap = {} as CardMetadataMap;
       const nextCardMetadataList = buildCardMetadataList(nextTagMetadata);
       const nextCardList = [];
@@ -514,13 +512,16 @@ const reducer = createReducer(
         }
       }
 
-      const {nextCardToPinnedCopy,
-             nextPinnedCardToOriginal,
-             nextCardMetadataMap}
-          = updateCardMap(state.cardToPinnedCopy,
-                          state.pinnedCardToOriginal,
-                          newCardMetadataMap,
-                          nextCardList);
+      const {
+        nextCardToPinnedCopy,
+        nextPinnedCardToOriginal,
+        nextCardMetadataMap,
+      } = updateCardMaps(
+        state.cardToPinnedCopy,
+        state.pinnedCardToOriginal,
+        newCardMetadataMap,
+        nextCardList
+      );
 
       const resolvedResult = buildOrReturnStateWithUnresolvedImportedPins(
         state.unresolvedImportedPinnedCards,
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -50,10 +50,10 @@ import {
   canCreateNewPins,
   createPluginDataWithLoadable,
   createRunToLoadState,
+  generateNextPinnedCardMappings,
   getCardId,
   getRunIds,
   getTimeSeriesLoadable,
-  updatePinnedCardMappingsUnderNamespaceUnchanged,
 } from './metrics_store_internal_utils';
 import {
   CardMetadataMap,
@@ -280,8 +280,8 @@ const {initialState, reducers: routeContextReducer} = createRouteContextedState<
     return {
       ...state,
       // We want to trigger tag metadata to reload every time route id changed, which is
-      // needed wthin the same namespace.
-      // TODO(japie1235813):moves the reload trigging to a proper place.
+      // needed within the same namespace.
+      // TODO(japie1235813):moves the reload triggering to a proper place.
       tagMetadataLoadState: {
         state: DataLoadState.NOT_LOADED,
         lastLoadedTimeInMs: null,
@@ -513,20 +513,22 @@ const reducer = createReducer(
         }
       }
 
-      // Updates pinned/original card id mapping and cardmetadatamap because
-      // they are routeful state and remain unchanged under the same namespace.
-      // The mappings are outdated under experiments removal.
-      // The outputs remain the same on the namespace changed.
+      // Generates next pinned/original card id mapping because they are routeful
+      // state and remain unchanged under the same namespace.
       const {
         nextCardToPinnedCopy,
         nextPinnedCardToOriginal,
-        nextCardMetadataMap,
-      } = updatePinnedCardMappingsUnderNamespaceUnchanged(
+        pinnedCardMetadataMap,
+      } = generateNextPinnedCardMappings(
         state.cardToPinnedCopy,
         state.pinnedCardToOriginal,
         newCardMetadataMap,
         nextCardList
       );
+      const nextCardMetadataMap = {
+        ...newCardMetadataMap,
+        ...pinnedCardMetadataMap,
+      };
 
       const resolvedResult = buildOrReturnStateWithUnresolvedImportedPins(
         state.unresolvedImportedPinnedCards,
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -50,6 +50,7 @@ import {
   canCreateNewPins,
   createPluginDataWithLoadable,
   createRunToLoadState,
+  generateNextCardStepIndex,
   generateNextPinnedCardMappings,
   getCardId,
   getRunIds,
@@ -530,13 +531,18 @@ const reducer = createReducer(
         ...pinnedCardMetadataMap,
       };
 
+      const nextCardStepIndex = generateNextCardStepIndex(
+        state.cardStepIndex,
+        nextCardMetadataMap
+      );
+
       const resolvedResult = buildOrReturnStateWithUnresolvedImportedPins(
         state.unresolvedImportedPinnedCards,
         nextCardList,
         nextCardMetadataMap,
         nextCardToPinnedCopy,
         nextPinnedCardToOriginal,
-        state.cardStepIndex
+        nextCardStepIndex
       );
 
       return {
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -53,6 +53,7 @@ import {
   getCardId,
   getRunIds,
   getTimeSeriesLoadable,
+  updateCardMap,
 } from './metrics_store_internal_utils';
 import {
   CardMetadataMap,
@@ -278,6 +279,12 @@ const {initialState, reducers: routeContextReducer} = createRouteContextedState<
   (state) => {
     return {
       ...state,
+      // Forces tag metadata to reload every time route id changed. This is
+      // needed wthin the same namespace.
+      tagMetadataLoadState: {
+        state: DataLoadState.NOT_LOADED,
+        lastLoadedTimeInMs: null,
+      },
       // Reset visible cards in case we resume a route that was left dirty.
       // Since visibility tracking is async, the state may not have received
       // 'exited card' updates when it was cached by the router.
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -502,4 +504,104 @@ describe('metrics store utils', () => {
       expect(canCreateNewPins(state)).toBe(false);
     });
   });
+
+  describe('updateCardMaps', () => {
```
bmd3k, line 508: Although I appreciate the modularity of breaking this up into updateCardMaps(), I think at least some of the testing for this use case needs to be done at the reducer level. This new logic is a core but subtle behavior of a very complex reducer.

* Keeping it disjoint like this actually makes it more difficult to reason out what the behavior of the reducer is supposed to be.

* Keeping it disjoint makes the reducer more difficult to successfully refactor without introducing bugs. A key goal should be: Someone should be able to refactor the reducer logic and have strong certainty that these use cases haven't been broken. (I personally would like to hack around with this reducer code to see if we can greatly simplify it and having the tests at the reducer level would make this easier).

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -512,12 +513,23 @@ const reducer = createReducer(
...
       }
 
-      const resolvedResult = buildOrReturnStateWithUnresolvedImportedPins(
-        state.unresolvedImportedPinnedCards,
-        newCardIds,
+      const {
+        nextCardToPinnedCopy,
+        nextPinnedCardToOriginal,
         nextCardMetadataMap,
+      } = updateCardMaps(
```
bmd3k, line 520: Can you explain why we need to pass state.cardToPinnedCopy and state.pinnedCardToOriginal to this function, rather than just starting from scratch?

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
bmd3k: Just want to clarify: I think its fine to keep updateCardMaps() but I think some of the testing needs to be moved or duplicated in metrics_reducers_test.

japie1235813: Thanks for the clarification and suggestion. I was thinking of the duplicated tests too and thought maybe they should not be duplicated. On the other hand, it also does not hurt to add more tests. I will just add tests for the method for future easiness of refactoring.

tensorboard/webapp/metrics/store/metrics_reducers.ts
japie1235813: Without passing these functions we don't have information to construct the maps. CardList contains a list of ids, `cardMetadataMap`, which is built from scratch in this change, maps `CardId` to card data (tag, runId etc). 
(due to the same reason we need to put back the pinned-original card id mapping to `cardMetadataMap`)

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -278,6 +279,13 @@ const {initialState, reducers: routeContextReducer} = createRouteContextedState<
   (state) => {
     return {
       ...state,
+      // We want to trigger tag metadata to reload every time route id changed, which is
+      // needed wthin the same namespace.
```
bmd3k, line 283: wthin => within

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -278,6 +279,13 @@ const {initialState, reducers: routeContextReducer} = createRouteContextedState<
...
+      // TODO(japie1235813):moves the reload trigging to a proper place.
```
bmd3k, line 284: trigging => triggering

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -511,12 +513,27 @@ const reducer = createReducer(
         }
       }
 
-      const resolvedResult = buildOrReturnStateWithUnresolvedImportedPins(
-        state.unresolvedImportedPinnedCards,
-        newCardIds,
+      // Updates pinned/original card id mapping and cardmetadatamap because
```
bmd3k, line 516: On second thought, if you make changes to the documentation and API that I describe, then you probably don't need this comment.

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -302,6 +303,179 @@ describe('metrics reducers', () => {
...
+      expect(nextState.cardList).toEqual(expectedState.cardList);
+      expect(nextState.cardToPinnedCopy).toEqual(
+        expectedState.cardToPinnedCopy
+      );
+      expect(nextState.pinnedCardToOriginal).toEqual(
+        expectedState.pinnedCardToOriginal
+      );
+    });
+
+    it('updates cardMetadataMap and remains pinned/original cards mapping unchanged on non-pinned cards removal', () => {
```
bmd3k, line 368: remains => keeps?

tensorboard/webapp/metrics/store/metrics_reducers.ts
bmd3k: I understand better now - I thought we could build these maps without looking at previous state but I realize now that isn't possible.

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -423,13 +597,23 @@ describe('metrics reducers', () => {
...
+
       const action = actions.metricsTagMetadataLoaded({tagMetadata});
       const nextState = reducers(beforeState, action);
 
-      expect(nextState.cardMetadataMap['<cardId>']).toEqual(
-        createScalarCardMetadata()
+      const expectedCardMetadataMap: CardMetadataMap = {};
+      expectedCardMetadataMap[cardId] = cardMetadata;
+
+      expect(nextState.cardMetadataMap['<cardId>']).not.toBe(origCardMetadata);
```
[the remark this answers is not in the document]
japie1235813: undefined or null does not work out because it checks type. To be concrete the meesage is
```
 Argument of type 'undefined' is not assignable to parameter of type 'Expected<CardMetadata>'.
```

I replace it with `hasOwnProperty`
```
expect(nextState.cardMetadataMap.hasOwnProperty('<cardId>')).toBe(false);
```

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -295,6 +295,55 @@ export function buildOrReturnStateWithPinnedCopy(
...
+ * mapping of removed experiments. This function handles the mappings.
+ * In the reducer nextCardmetadatamap is created from empty instead of
+ * preserving the previous map, which might also contain outdated mapping due to
+ * the same reason. To preserve the pinned card mapping we need to adding the
+ * pinned card mappng to cardmetadatamap.
+ */
+export function updatePinnedCardMappingsUnderNamespaceUnchanged(
+  cardToPinnedCopy: CardToPinnedCard,
+  pinnedCardToOriginal: PinnedCardToCard,
+  nextCardMetadataMap: CardMetadataMap,
```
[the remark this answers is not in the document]
japie1235813: Thanks for the suggestion! I like the suggested flow of return all newly created mapping. Updated

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -302,6 +303,283 @@ describe('metrics reducers', () => {
...
+      const pinnedCopyId2 = getPinnedCardId(cardId2);
+      const beforeState = buildMetricsState({
+        cardMetadataMap: {
+          [cardId1]: cardMetadata1,
+          [cardId2]: cardMetadata2,
+          [pinnedCopyId1]: cardMetadata1,
+          [pinnedCopyId2]: cardMetadata2,
+        },
+        cardStepIndex: {
+          [pinnedCopyId1]: 1,
```
bmd3k, line 499: Could/should cardId1 also have an entry here?

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -302,6 +303,283 @@ describe('metrics reducers', () => {
...
+      });
+      const nextState = reducers(beforeState, action);
+
+      const expectedState = buildMetricsState({
+        cardMetadataMap: {
+          [cardId1]: cardMetadata1,
+          [pinnedCopyId1]: cardMetadata1,
+        },
+        cardStepIndex: {
+          [pinnedCopyId1]: 1,
```
bmd3k, line 529: Same: should cardid1 also be here?

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -295,6 +295,55 @@ export function buildOrReturnStateWithPinnedCopy(
   };
 }
 
+/**
+ * Returns updated cardToPinnedCopy, pinnedCardToOriginal, and cardMetadataMap.
```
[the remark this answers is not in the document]
bmd3k: I still think some adjustments to documentation would be helpful. Note that I'm also implicitly suggesting renaming cardToPinnedCopy to previousCardToPinnedCopy.

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -295,6 +295,55 @@ export function buildOrReturnStateWithPinnedCopy(
...
+  nextCardList: CardId[]
+) {
+  const nextCardToPinnedCopy = new Map() as CardToPinnedCard;
+  for (const cardId of nextCardList) {
+    if (cardToPinnedCopy.has(cardId)) {
+      nextCardToPinnedCopy.set(cardId, cardToPinnedCopy.get(cardId)!);
+    }
+  }
+
+  const nextPinnedCardToOriginal = new Map(pinnedCardToOriginal);
```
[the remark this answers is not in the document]
bmd3k: I still think this is good cleanup to do.

In fact, could you have just a single for loop for the entire function?

```
  const nextCardToPinnedCopy = new Map() as CardToPinnedCard;
  const nextPinnedCardToOriginal = new Map() as PinnedCardToCard; 
  const pinnedCardMetadataMap = {} as CardMetadataMap;
  for (const cardId of nextCardList) {
    if (cardToPinnedCopy.has(cardId)) {
      const pinnedCardId = cardToPinnedCopy.get(cardId)!
      nextCardToPinnedCopy.set(cardId, pinnedCardId);
      nextPinnedCardToOriginal.set(pinnedCardId, cardId);
      pinnedCardMetadataMap[pinnedCardId] = nextCardMetadataMap[pinnedCardId];
    }
  }

  return {nextCardToPinnedCopy, nextPinnedCardToOriginal, pinnedCardMetadataMap};
```

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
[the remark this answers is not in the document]
bmd3k: Thanks.

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -295,6 +295,55 @@ export function buildOrReturnStateWithPinnedCopy(
...
+ * preserving the previous map, which might also contain outdated mapping due to
+ * the same reason. To preserve the pinned card mapping we need to adding the
+ * pinned card mappng to cardmetadatamap.
+ */
+export function updatePinnedCardMappingsUnderNamespaceUnchanged(
+  cardToPinnedCopy: CardToPinnedCard,
+  pinnedCardToOriginal: PinnedCardToCard,
+  nextCardMetadataMap: CardMetadataMap,
+  nextCardList: CardId[]
+) {
```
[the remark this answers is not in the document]
bmd3k: Thanks for taking care of this.

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -295,6 +295,72 @@ export function buildOrReturnStateWithPinnedCopy(
...
+    nextPinnedCardToOriginal,
+    pinnedCardMetadataMap,
+  };
+}
+
+/**
+ * Removes cards not in cardMetadataMap from cardStepIndex mapping.
+ */
+export function generateNextCardStepIndex(
+  cardStepIndex: CardStepIndexMap,
```
bmd3k, line 352: For consistency with generateNextPinnedCardMappings function, I'd recommend renaming these arguments to previousCardStepIndex and nextCardMetadataMap.

bmd3k (approved):

PR #5456, opened 2021-12-10
histogram: define a separate large sized py_test rule for `test_with_large_counts`

yatbear:
`test_with_large_counts` tests will be skipped in the original `summary_test`.

Googlers, see [b/205170479](https://b.corp.google.com/issues/205170479) for more details.

#histogram #onduty


nfelt (changes requested):
Hmm, it's a bit too bad to have to replicate all of the setup and infrastructure for the various flavors of summary API in the separate test file.  Here's an idea - we could keep the test part of the single source file, but make it not run by default, and define a separate py_test() rule that _does_ run that test.

I think we could do this as follows:

- Add logic to the test case that skips the large count test unless the `TEST_SIZE` environment variable is "large" (https://docs.bazel.build/versions/main/test-encyclopedia.html#initial-conditions)
- In the separate `py_test()` rule, we make that one `size = "large"` and we set the `args` parameter (https://docs.bazel.build/versions/main/be/python.html#py_test) to the fully qualified name of the test method we want to run. This will get passed to the test-specific `main()` function and should result in only that test method being run: https://docs.python.org/3/library/unittest.html#unittest.main

yatbear:
> Hmm, it's a bit too bad to have to replicate all of the setup and infrastructure for the various flavors of summary API in the separate test file. Here's an idea - we could keep the test part of the single source file, but make it not run by default, and define a separate py_test() rule that _does_ run that test.
> 
> I think we could do this as follows:
> 
> * Add logic to the test case that skips the large count test unless the `TEST_SIZE` environment variable is "large" (https://docs.bazel.build/versions/main/test-encyclopedia.html#initial-conditions)
> * In the separate `py_test()` rule, we make that one `size = "large"` and we set the `args` parameter (https://docs.bazel.build/versions/main/be/python.html#py_test) to the fully qualified name of the test method we want to run. This will get passed to the test-specific `main()` function and should result in only that test method being run: https://docs.python.org/3/library/unittest.html#unittest.main

Thanks for the details!

tensorboard/plugins/histogram/BUILD
```diff
@@ -125,6 +125,28 @@ py_test(
     ],
 )
 
+py_test(
+    name = "large_count_summary_test",
+    # Ensure enough RAM for `test_with_large_counts`.
+    size = "large",
+    srcs = ["summary_test.py"],
+    args = [
+        "SummaryV1OpTest.test_with_large_counts",
```
nfelt, line 134: Oh right, I forgot this is actually a bunch of individual tests. Perhaps we can instead use `args = ["-k", "test_with_large_counts"]`?

https://docs.python.org/3.7/library/unittest.html#cmdoption-unittest-k

nfelt (approved):

tensorboard/plugins/histogram/BUILD
yatbear: Thanks for the suggestion. The first arg needs to be `"-- -k"` for some reason otherwise the flag won't be recognized.

PR #5459, opened 2021-12-11
type: enable `exactOptionalPropertyTypes` option 

stephanwlee:
Before this change, below held true:
```ts
interface Foo {
  bar?: number;
}

// Legal
const foo1: Foo = {bar: undefined};
// Legal
const foo2: Foo = {};
```

After this change,
```ts
interface Foo {
  bar?: number;
  baz: number|undefined;
}

// Illegal; `bar` cannot be `undefined` and must be number.
const foo1: Foo = {bar: undefined, baz: undefined};
// Illegal; `baz` property must exist. Must be one of `number` of
// `undefined`.
const foo1: Foo = {};
// Legal; `bar` property does not exist while `baz` is specified.
const foo2: Foo = {baz: undefined};
```

When a property is defined as `?:`, it does not mean that the property
can be `undefined` but it means a property can be omitted altogether or
it must exist with the value of the type as defined. As such, when a
property is written with `undefined`, the value type must explicitly
allow `undefined` like `baz` example above.

Note: this change currently includes #5458. Please review the second commit and beyond.

### DIFF
--- a/tensorboard/webapp/app_routing/route_config.ts
+++ b/tensorboard/webapp/app_routing/route_config.ts
@@ -31,7 +31,7 @@ interface PositiveMatch {
   params: RouteParams;
   pathParts: string[];
   isRedirection: boolean;
-  redirectionQueryParams?: SerializableQueryParams;
+  redirectionQueryParams?: SerializableQueryParams | undefined;
 }
 
 type Match = NegativeMatch | PositiveMatch;
@@ -285,7 +285,7 @@ export interface NonRedirectionRouteMatch extends BaseRedirectionRouteMatch {
 
 export interface RedirectionRouteMatch extends BaseRedirectionRouteMatch {
   originateFromRedirection: true;
-  redirectionOnlyQueryParams?: Route['queryParams'];
+  redirectionOnlyQueryParams: Route['queryParams'] | undefined;
 }
 
 export type RouteMatch = NonRedirectionRouteMatch | RedirectionRouteMatch;
@@ -441,7 +441,8 @@ export class RouteConfigs {
         deepLinkProvider: definition.deepLinkProvider ?? null,
         pathname: definition.path,
         params: {},
-        originateFromRedirection: wasRedirected,
+        originateFromRedirection: true,
+        redirectionOnlyQueryParams: undefined,
       };
     }
 
--- a/tensorboard/webapp/app_routing/route_config.ts
+++ b/tensorboard/webapp/app_routing/route_config.ts
@@ -31,7 +31,7 @@ interface PositiveMatch {
   params: RouteParams;
   pathParts: string[];
   isRedirection: boolean;
-  redirectionQueryParams?: SerializableQueryParams | undefined;
+  redirectionQueryParams: SerializableQueryParams | undefined;
 }
 
 type Match = NegativeMatch | PositiveMatch;
@@ -172,6 +172,7 @@ abstract class RouteConfigMatcher {
       params: combinedParams,
       pathParts,
       isRedirection: false,
+      redirectionQueryParams: undefined,
     };
   }
 
@@ -186,6 +187,7 @@ abstract class RouteConfigMatcher {
       params,
       pathParts,
       isRedirection: false,
+      redirectionQueryParams: undefined,
     };
   }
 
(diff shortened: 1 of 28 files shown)


tensorboard/webapp/metrics/effects/index.ts
```diff
@@ -199,9 +199,13 @@ export class MetricsEffects implements OnInitEffects {
      */
     const requests: TimeSeriesRequest[] = fetchInfos.map((fetchInfo) => {
       const {plugin, tag, runId, sample} = fetchInfo;
-      return isSingleRunPlugin(plugin)
-        ? {plugin, tag, sample, runId: runId!}
-        : {plugin, tag, sample, experimentIds};
+      const partialRequest: TimeSeriesRequest = isSingleRunPlugin(plugin)
+        ? {plugin, tag, runId: runId!}
+        : {plugin, tag, experimentIds};
+      if (sample !== undefined) {
```
bmd3k, line 205: It seems you have two major choices for fixing errors arising from this new compiler option:
1. Convert property from `property?: type` to `property: type | undefined`
2. Keep `property?: type` but guard setting the property only if we know there is a value to set it to -- for example: `if (valueToSet !== undefined) property.type = valueToSet`

It seems like you applied these options roughly equally in this change. Did you have some methodology to deciding whether to use option 1 or option 2?

tensorboard/webapp/webapp_data_source/tb_http_client.ts
```diff
@@ -84,7 +84,7 @@ export class TBHttpClient implements TBHttpClientInterface {
     path: string,
     // Angular's HttpClient is typed exactly this way.
     body: any | null,
-    options: PostOptions = {}
+    options: PostOptions | undefined = {}
```
bmd3k, line 87: What does it mean for this to have both undefined and a default value of {}? Does the default value mean it's still an optional property? But there are still some cases where you want to explicitly set it to undefined?

bmd3k (commented):
Worth noting that Google's TS team looked at enabling this property internally but decided against it: b/193869056.

tensorboard/webapp/webapp_data_source/tb_http_client.ts
stephanwlee: Unfortunately, TypeScript does not seem to understand that `options` cannot be undefined when having default value like this. `undefined` was introduce TypeScript compiler and that's all. 

No, there isn't a case where we would like `options` to ever be `undefined` but we still guard against that for now.

tensorboard/webapp/metrics/effects/index.ts
stephanwlee: Binary classification roughly happened like this:

1. is this an internal type where we can be more explicit about existence of a property? If so, I preferred `<T | undefined>` over `prop?: T` since conditionally forming a data structure that conditionally has a property is more difficult and is harder to validate. For instance, `foo.hasOwnProperty('propName')` forces `"propName"` to be understood by the compiler and mangle accordingly or becomes incompatible with mangler like Closure compiler. 
2. if server responses/requests can omit a property, requires a property to be omitted, or is more efficient with a property removed than having undefined value for each one of them, I preferred `<T | undefined>`. 

Metrics data source does not do much transformation to convert a backend type to an internal type so this module, specifically, has a lot of the mixed pattern.

tensorboard/webapp/app_routing/route_config.ts
```diff
@@ -441,7 +441,8 @@ export class RouteConfigs {
         deepLinkProvider: definition.deepLinkProvider ?? null,
         pathname: definition.path,
         params: {},
-        originateFromRedirection: wasRedirected,
+        originateFromRedirection: true,
```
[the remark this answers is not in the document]
stephanwlee: This was an intentional change since this is handling the case where no route definition matched current url and we are "redirecting" to the default route. Also, hard coding `originateFromRedirection: true` keeps type very simple and clear.

stephanwlee:
> Worth noting that Google's TS team looked at enabling this property internally but decided against it: b/193869056.

Understood but it looks like the task was scrapped not because it is wrong but because (1) it is difficult to do LSC and (2) unclear whether developers truly intended to use `prop?: T` and `prop: T | undefined` interchangeably. Because this distinction has recently bit us, I am proposing that we enable the flag so we are clearer about our intention.

PR #5464, opened 2021-12-14
[time-namespaced state] Remove most usage of getRouteId().

bmd3k:
We remove most usage of getRouteId() (both the internal util function and the selector).

getRouteId() has otherwise been used by the codebase to ask "are two routes the same route kind and the same set of experiments"? So we introduce a new "areSameRouteKindAndExperiments()" function to answer this question more clearly.

For the most part, getRouteId() the selector can be replaced with getActiveRoute() and possibly the new areSameRouteAndExperiments().

For the most part, getRouteId() the internal util function can be replaced with areSameRouteAndExperiments().

The finaly remaining usage of getRouteId() is for namespace id generation in app_routing_effects. As this form of namespace id is on the way out (googlers, see go/tb-timespaced-state), we expect to be able to remove this sometime in the new year.


tensorboard/webapp/app_routing/testing.ts
```diff
@@ -35,13 +34,26 @@ export function buildCompareRoute(
...
+    pathname: '/compare',
     params: {experimentIds: aliasAndExperimentIds.join(',')},
     queryParams: [],
     ...routeOverride,
   };
 }
 
+export function buildExperimentRoute(
+  experimentId: string,
+  routeOverride: Partial<Route> = {}
```
japie1235813, line 46: The input is a bit confusing to me. I won't know to pass experiment id from the name. Also the level of usage for the second param is different from the first one. Would you consider combining the two params to be more general input?
for example  `buildExperimentRoute({params: {experimentId: 'foo'}})`?

Or if the second param is not used, can we rename to buildExperimentRouteFromId('foo')?

tensorboard/webapp/runs/views/runs_table/runs_table_container.ts
```diff
@@ -389,8 +391,12 @@ export class RunsTableContainer implements OnInit, OnDestroy {
      * the store aware of the visibility of any run tables.
      */
     if (this.columns.includes(RunsTableColumn.CHECKBOX)) {
-      const runsExceedLimitForRoute$ = this.store.select(getRouteId).pipe(
+      const runsExceedLimitForRoute$ = this.store.select(getActiveRoute).pipe(
         takeUntil(this.ngUnsubscribe),
+        // BDTODO: This needs to be tested manually.
+        distinctUntilChanged((prevRoute, currRoute) => {
+          return areSameRouteAndExperiments(prevRoute, currRoute);
+        }),
```
japie1235813, line 399: What (or what timing) makes the route might be different that we want to add `distinctUntilChanged` here?

tensorboard/webapp/app_routing/internal_utils_test.ts
```diff
@@ -170,6 +171,110 @@ describe('app_routing/utils', () => {
...
+
+    it('returns false when one route is null', () => {
+      expect(utils.areSameRouteAndExperiments(buildRoute(), null)).toBeFalse();
+      expect(utils.areSameRouteAndExperiments(null, buildRoute())).toBeFalse();
+    });
+
+    it('returns false when different ', () => {
+      expect(utils.areSameRouteAndExperiments(buildRoute(), null)).toBeFalse();
+      expect(utils.areSameRouteAndExperiments(null, buildRoute())).toBeFalse();
+    });
```
japie1235813, line 187: What makes this test case different than the previous one?

japie1235813 (commented):
Overall the pr lgtm. I think it's clear to replace the comparison with a single function. 
Just want to clarify, in my mindset, "route" is composed of "route kind" and "params" (params contains experiment ids), both of which contributes the construction of "route Id" (which we want to deprecate)
I think it's clear to add route kind, `areSameRouteKindAndExperiments`. WDYT?

tensorboard/webapp/app_routing/testing.ts
bmd3k: Done. Changed to buildExperimentRouteFromId.

tensorboard/webapp/app_routing/views/router_outlet_test.ts
```diff
@@ -154,7 +154,7 @@ describe('router_outlet', () => {
     expect(fixture.nativeElement.innerText).toBe('I am a test');
   });
 
-  it('does not remove page when navigating to the same routeId', () => {
+  it('does not remove page when navigating to same route and experiments', () => {
```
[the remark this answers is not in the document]
bmd3k: Correct. Done.

tensorboard/webapp/app_routing/internal_utils_test.ts
bmd3k: I think I meant for this test to be 'returns false when different kind' but never got around to finishing it. Fixed it up now.

bmd3k:
> Overall the pr lgtm. I think it's clear to replace the comparison with a single function. Just want to clarify, in my mindset, "route" is composed of "route kind" and "params" (params contains experiment ids), both of which contributes the construction of "route Id" (which we want to deprecate) I think it's clear to add route kind, `areSameRouteKindAndExperiments`. WDYT?

Makes sense to me. Renamed.

japie1235813 (approved):
Thanks for making the changes!

PR #5465, opened 2021-12-15
line chart: better WebGL edge case handle

stephanwlee:
This change makes two improvements to WebGL support for the line chart.

1. Better feature detection. We now also listen to
   `webglcontextcreationerror` which gets emitted when browser fails to
   create webgl context from a canvas. This can be a good indication
   that user's system may not be able to handle WebGL renderer.
2. Better removal of JavaScript object. When an Angular components get
   scrapped, we get a chance to clean more stuff up. With the signal, we
   now instruct Three.js renderer to dispose all the objects more
   explicitly for better freeing of WebGL context + Three.js objects.

There is no great way to unit test to guarantee correctness of the
utility and the operation. Thus, we will leave such test to an
integration test that would exercise browser capabilities.  For now, we
are happy with a manual functional test.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
@@ -84,6 +84,11 @@ export interface ObjectRenderer<CacheValue = {}> {
   destroyObject(cachedValue: CacheValue): void;
 
   setUseDarkMode(useDarkMode: boolean): void;
+
+  /**
+   * Disposes rendering context. After invocation, renderer will never be used.
+   */
+  dispose(): void;
 }
 
 interface PaintOption {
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
```diff
@@ -84,6 +84,11 @@ export interface ObjectRenderer<CacheValue = {}> {
   destroyObject(cachedValue: CacheValue): void;
 
   setUseDarkMode(useDarkMode: boolean): void;
+
+  /**
+   * Disposes rendering context. After invocation, renderer will never be used.
```
bmd3k, line 89: "will never be used" or "should never be used"?

bmd3k (approved):

tensorboard/webapp/widgets/line_chart_v2/lib/renderer/renderer_types.ts
stephanwlee: This is indicating to the implementer of renderer that one should expect that the renderer will no longer be used ever after `dispose` is called. `should never be used` sounds like an instruction to a user of the renderer. :thinking: 

I could see this jsdoc being used for both cases. Adopted your phrasing instead.

PR #5466, opened 2021-12-15
fix: timeseries rendering issue on Firefox

stephanwlee:
On Firefox, IntersectionObserverInit's rootMargin property must have a
unit value and thus cannot be an empty string. As such, our recent
refactor which populated its default value as `""` threw in Firefox
(while Chrome was lenient) causing cards to not appear in TimeSeries
dashboard. This change optionally populates the `rootMargin` property in
`IntersectionObserverInit` instead of default setting values.

The bug was caught in an integration test. Added unit test to fail faster in
the future.


nfelt (approved):
Thanks!

PR #5470, opened 2021-12-20
(time-namespaced state) Reuse namespace id on page reload if it exists.

bmd3k:
This change ensures that we read and reuse namespace id from browser history if a page is reloaded.

Consider the situation where user:

1. Selects to view an experiment and sets tag filter to 'A'.
2. Adds an experiment to the comparison and then sets the tag filter to 'B'.
3. Refreshes the page.

All these actions are supposed to take place in the same namespace. But currently the page refresh breaks this - a page load always generates a new namespace and new namespace id. So if user navigates back to the single view they will see tag filter set to 'A' instead of to 'B'.

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -86,21 +86,20 @@ enum NamespaceUpdateOption {
   FROM_HISTORY,
 }
 
-type NavigationOptions =
+type NamespaceUpdate =
   | {
-      browserInitiated: boolean;
-      replaceState: boolean;
-      namespaceUpdateOption:
-        | NamespaceUpdateOption.NEW
-        | NamespaceUpdateOption.UNCHANGED;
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -181,12 +182,31 @@ export class AppRoutingEffects {
...
+                namespaceId: namespaceId,
+              };
+
         return {
           pathname: this.location.getPath(),
           options: {
             browserInitiated: true,
             replaceState: true,
-            namespaceUpdateOption: NamespaceUpdateOption.UNCHANGED,
+            namespaceUpdate,
```
bmd3k, line 209: Note that I had to modify the NavigationOption typing so that I could construct the object this way without the compiler complaining. (Where "this way" is having a variable namespaceUpdate while other options -- browserInitiated and replaceState -- are statically set).

japie1235813 (approved):

PR #5473, opened 2021-12-21
(time-namespaced state) Make core state route-contexted.

bmd3k:
This makes core state "route-contexted" in our OSS repo and, therefore, for all flavors of TensorBoard.

Up to now it has only been "route-contexted" for TBcorp:
http://go/tbcorp-core-route-contexted

There's really no reason this difference needs to continue. We can make it route-contexted for all flavors of TensorBoard with no or little impact:
* For OSS and TBgcp there should be no user-visible impact on behavior, since they are effectively single-page apps. TBgcp doesn't even have the TimeSeries dashboard.
* For TBdev, which is a multi-page app, there is some impact on header behavior. But it, too, does not yet have the TimeSeries dashboard so there is otherwise little impact.

FYI, the real reason I'm making this change: I will be able to delete the code at http://go/tbcorp-core-route-contexted (it will be redundant) and then all references to '/tensorboard/webapp/app_routing/route_contexted_reducer_helper' will be in the OSS code base. There will be none in the internal codebase. This will make it much easier to rename symbols for http://b/210998256.


japie1235813 (approved):
OSS does have Time Series Dashboard but it might not be affected due to it's a single page app .

bmd3k:
> OSS does have Time Series Dashboard but it might not be affected due to it's a single page app .

Whoops, yes. I'll ensure the PR and commit descriptions are updated.

PR #5475, opened 2021-12-21
(time-namespaced state) Complete removal of routeId.

bmd3k:
After removing most of routeId in #5464 and then following up with changes to internal code, we can complete the removal of the term "RouteId". Notable changes:

* `getRouteId` ngrx selector is removed.

* `route_contexted_reducer_helper`'s `onRouteIdChanged` argument is renamed to `onRouteKindOrExperimentsChanged`.

* test helper function `buildNavigatedToNewRouteIdAction` is renamed to `buildNavigatedToNewExperimentAction`.

* The one remaining usage of "RouteId" is for generating namespace ids for App Routing. We therefore rename the internal function `getRouteId()` to `getRouteNamespaceId()` to signal the narrowed purpose of the function. Note that we will remove `getRouteNamespacedId()` with the launch of go/tb-timespaced-state.

### DIFF
--- a/tensorboard/webapp/app_routing/route_contexted_reducer_helper.ts
+++ b/tensorboard/webapp/app_routing/route_contexted_reducer_helper.ts
@@ -187,7 +187,8 @@ export function createRouteContextedState<
           !areSameRouteKindAndExperiments(before, after) &&
           onRouteKindOrExperimentsChanged
         ) {
-          // Experiments have changed. Delegate additional changes to the caller.
+          // RouteKind or Experiments have changed. Delegate additional changes
+          // to the caller.
           nextFullState = onRouteKindOrExperimentsChanged(nextFullState, after);
         }
 
--- a/tensorboard/webapp/app_routing/route_contexted_reducer_helper.ts
+++ b/tensorboard/webapp/app_routing/route_contexted_reducer_helper.ts
@@ -73,8 +73,8 @@ export type RouteContextedState<
  * Utility for managing routeful states. It returns route contexted
  * `initialState` and `reducers` that help manage the routeful state.
  *
- * An optional `onRouteIdChanged` function will modify the state after it
- * is loaded from the cache.
+ * An optional `onRouteKindOrExperimentsChanged` function will modify the state
+ * after it is loaded from the cache.
  *
  * Example usage:
  *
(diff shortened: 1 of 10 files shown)


tensorboard/webapp/app_routing/route_contexted_reducer_helper.ts
```diff
@@ -185,10 +185,10 @@ export function createRouteContextedState<
 
         if (
           !areSameRouteKindAndExperiments(before, after) &&
-          onRouteIdChanged
+          onRouteKindOrExperimentsChanged
         ) {
-          // areSameRouteKindAndExperimentsperiments have changed. Delegate additional changes to the caller.
-          nextFullState = onRouteIdChanged(nextFullState, after);
+          // Experiments have changed. Delegate additional changes to the caller.
```
japie1235813, line 190: `Route kinds or experiments have changed`? Or only experiments will be changed?

japie1235813 (approved):
Thanks for listing the notable changes. It helps with the pr review.

Overall the pr lgtm but I'm not quite sure about two things:
1. at first glance renaming `getRouteId` to `getRouteNamespaceId` seems not so meaningful to me while the function implementation stays the same.
2. What is the plan to get ride of `getRouteId` eventually? Are you thinking about rewrite the `getRouteNamespaceId` function?

tensorboard/webapp/app_routing/route_contexted_reducer_helper.ts
bmd3k: Right, updated.

bmd3k:
> Thanks for listing the notable changes. It helps with the pr review.
> 
> Overall the pr lgtm but I'm not quite sure about two things:
> 
> 1. at first glance renaming `getRouteId` to `getRouteNamespaceId` seems not so meaningful to me while the function implementation stays the same.

The rename is a bit of theatre. I agree it has little logical impact. Basically I wanted to make sure the following command returned as little as possible: `grep -r -i "routeId" tensorboard`. Now it returns a single result and I plan to remove that result when I rename the symbols in route_contexted_reducer_helper.ts.

> 2. What is the plan to get ride of `getRouteId` eventually? Are you thinking about rewrite the `getRouteNamespaceId` function?

The plan is to launch go/tb-timespaced-state and delete the old code, including `getRouteNamespacedId`.

Issue #5478, opened 2021-12-24
"ImportError: cannot import name 'Mapping' from 'collections'" with Python 3.10


simonkrenger:
## Environment information (required)

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version e43767ef2b648d0d5d57c00f38ccbd38390e38da

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=10, micro=1, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): posix.uname_result(sysname='Linux', nodename='fedora-ml.example.com', release='5.15.10-200.fc35.x86_64', version='#1 SMP Fri Dec 17 14:46:39 UTC 2021', machine='x86_64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: False
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
INFO: installed: tensorboard==2.7.0
INFO: installed: tensorflow==2.8.0rc0
INFO: installed: tf-estimator-nightly==2.8.0.dev2021122109
INFO: installed: tensorboard-data-server==0.6.1

--- check: tensorboard_python_version
INFO: tensorboard.version.VERSION: '2.7.0'

--- check: tensorflow_python_version
2021-12-24 22:11:13.556272: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
2021-12-24 22:11:13.556660: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
INFO: tensorflow.__version__: '2.8.0-rc0'
INFO: tensorflow.__git_version__: 'v1.12.1-69099-g0d00ab2fd15'

--- check: tensorboard_data_server_version
INFO: data server binary: '/home/simon/.local/lib/python3.10/site-packages/tensorboard_data_server/bin/server'
INFO: data server binary version: b'rustboard 0.6.1'

--- check: tensorboard_binary_path
INFO: which tensorboard: b'/home/simon/.local/bin/tensorboard\n'

--- check: addrinfos
socket.has_ipv6 = True
socket.AF_UNSPEC = <AddressFamily.AF_UNSPEC: 0>
[...]

PR #5490, opened 2022-01-07
chore: Upgrade html5lib to version 1.1.

bmd3k:
As noted by https://github.com/tensorflow/tensorboard/issues/5478 , tensorboard fails when run in a python 3.10 environment. The reporter fortunately discovered that the problems would be fixed with an upgrade of our html5lib dependency.

I tested by building a python3.10 virtualenv on my machine (this required building python 3.10 from source as we don't have it natively).

```
$ virtualenv -p /usr/local/bin/python3.10 ~/virtualenv/tensorboard-python-3.10
$ source ~/virtualenv/tensorboard-python-3.10/bin/activate
$ python3.10 -m ensurepip --upgrade
$ python3.10 -m pip install tensorboard
$ tensorboard --version
  File "/usr/local/google/home/bdubois/virtualenv/tensorboard-python-3.10/bin/tensorboard", line 5, in <module>
    from tensorboard.main import run_main
<snip>
    from collections import Mapping
ImportError: cannot import name 'Mapping' from 'collections' (/usr/local/lib/python3.10/collections/__init__.py)
$ python3.10 -m pip install --upgrade /tmp/tb-with-html5lib-1.1/tensorboard*.whl
$ tensorboard --version
TensorFlow installation not found - running with reduced feature set.
2.8.0a0
```

I also ran the tensorboard with some logdirs and did some basic sanity. I specifically checked that markdown rendering seemed to be working for a couple text plugin logs that I have.

### DIFF
--- a/third_party/python.bzl
+++ b/third_party/python.bzl
@@ -47,10 +47,11 @@ def tensorboard_python_workspace():
     http_archive(
         name = "org_html5lib",
         urls = [
-            "https://files.pythonhosted.org/packages/ac/b6/b55c3f49042f1df3dcd422b7f224f939892ee94f22abcf503a9b7339eaf2/html5lib-1.1.tar.gz",
+            "http://mirror.tensorflow.org/github.com/html5lib/html5lib-python/archive/1.1.tar.gz",
+            "https://github.com/html5lib/html5lib-python/archive/1.1.tar.gz",
         ],
-        sha256 = "b2e5b40261e20f354d198eae92afc10d750afb487ed5e50f9c4eaf07c184146f",
-        strip_prefix = "html5lib-1.1",
+        sha256 = "66e9e24a53c10c27abb6be8a3cf2cf55824c6ea1cef8570a633cb223ec46e894",
+        strip_prefix = "html5lib-python-1.1",
         build_file = str(Label("//third_party:html5lib.BUILD")),
     )
 
--- a/third_party/python.bzl
+++ b/third_party/python.bzl
@@ -47,11 +47,10 @@ def tensorboard_python_workspace():
     http_archive(
         name = "org_html5lib",
         urls = [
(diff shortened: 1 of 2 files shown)


third_party/python.bzl
```diff
@@ -47,11 +47,10 @@ def tensorboard_python_workspace():
     http_archive(
         name = "org_html5lib",
         urls = [
-            "http://mirror.tensorflow.org/github.com/html5lib/html5lib-python/archive/1.0.1.tar.gz",
```
nfelt, line 50: Let's keep a mirror URL. We can add one for this if it doesn't exist yet (go/tensorboard-tf-mirror), but it looks like there is one for http://mirror.tensorflow.org/github.com/html5lib/html5lib-python/archive/1.1.tar.gz already so that's probably easiest.

nfelt (commented):
Thanks for taking care of this!

> I chose to use the PyPi URL for the download of the dependency as I had difficulties finding the sha256 hashes for the version of the library at mirror.tensorflow.org and github.com (they differ from the PyPi one).

It's fine to just generate the sha256 hash locally - i.e. download the tarball and compute the hash yourself.  In this case it might be easier to still do that, since the github one is already mirrored (see other comment).

FWIW, to justify that a little - the point of the hash isn't really to verify the integrity of the file that you're downloading today from the server, because in general we assume you're downloading it via HTTPS, on a reasonably secure machine and network, etc, and it doesn't really add much protection to separately get a hash from the same server rather than just computing it yourself.  The point is to ensure the server doesn't change the contents of the file in the future.  That way, we reduce the amount of trust we need to have from "I trust this server in perpetuity" to "I trust this server right now, today".

third_party/python.bzl
bmd3k: Makes sense. Updated to the familiar mirrors and calculated the checksum with `sha256sum`.

nfelt (approved):

PR #5492, opened 2022-01-07
Prefer the WebGL Renderer

JamesHollyer:
Changing the preferred renderer back to WebGL after crash mitigations.


bmd3k (approved):
Thanks! This change makes sense -- it appears to be the reverse of https://github.com/tensorflow/tensorboard/pull/5397.

Issue #5489, opened 2022-01-07
Tensorboard 2.7 seems to be broken on Safari


severo:
## Environment information (required)

Please run `diagnose_tensorboard.py` (link below) in the same
environment from which you normally run TensorFlow/TensorBoard, and
paste the output here:

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version e43767ef2b648d0d5d57c00f38ccbd38390e38da

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=8, micro=11, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): posix.uname_result(sysname='Linux', nodename='aws-dev-sylvain-lesage.local', release='5.11.0-1023-aws', version='#24~20.04.1-Ubuntu SMP Fri Dec 3 13:46:41 UTC 2021', machine='x86_64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: False
INFO: $VIRTUAL_ENV: '/tmp/test/.venv'

--- check: installed_packages
WARNING: no installation among: ['tb-nightly', 'tensorboard', 'tensorflow-tensorboard']
WARNING: no installation among: ['tensorflow', 'tensorflow-gpu', 'tf-nightly', 'tf-nightly-2.0-preview', 'tf-nightly-gpu', 'tf-nightly-gpu-2.0-preview']
WARNING: no installation among: ['tensorflow-estimator', 'tensorflow-estimator-2.0-preview', 'tf-estimator-nightly']

--- check: tensorboard_python_version
INFO: tensorboard.version.VERSION: '2.7.0'

--- check: tensorflow_python_version
Traceback (most recent call last):
  File "diagnose_tensorboard.py", line 522, in main
    suggestions.extend(check())
  File "diagnose_tensorboard.py", line 75, in wrapper
    result = fn()
  File "diagnose_tensorboard.py", line 278, in tensorflow_python_version
    import tensorflow as tf
ModuleNotFoundError: No module named 'tensorflow'

--- check: tensorboard_data_server_version
INFO: data server binary: '/tmp/test/.venv/lib/python3.8/site-packages/tensorboard_data_server/bin/server'
INFO: data server binary version: b'rustboard 0.6.1'

--- check: tensorboard_binary_path
INFO: which tensorboard: b'/tmp/test/.venv/bin/tensorboard\n'

[...]

PR #5495, opened 2022-01-11
Change terser configuration to workaround threejs/terser incompatibility.

bmd3k:
Users have reported problems loading TensorBoard in Safari: https://github.com/tensorflow/tensorboard/issues/5489

Thanks to @severo and @bileschi we were able to identify the problem to an incompatibility between threejs (which we use for WebGL rendering) and terser (which we use for minimizing/optimizing our code). 

Angular encountered this problem last year and worked around it by changing their terser config by reducing number of passes from 3 to 2: 
* https://github.com/angular/angular-cli/issues/21107
* https://github.com/angular/angular-cli/commit/2c2b499193fb319e1c9cb92318610353b7720e2b

In our case I have to reduce the number of terser passes from 3 to 1. For some reason reducing to 2 does not solve the problem.

This change surprisingly appears to reduce binary size. Before this change the size of index.js is 7672844 bytes. After this change the size of index.js is 7670992 bytes -- reducing binary size by 1852 bytes.

### DIFF
--- a/tensorboard/defs/defs.bzl
+++ b/tensorboard/defs/defs.bzl
@@ -77,6 +77,10 @@ def tf_js_binary(
         terser_minified(
             name = internal_result_name,
             src = internal_rollup_name,
+            # Notes about the terser config:
+            # compress.passes - this is set to '1' to workaround issue with
+            #   terser and threejs. In practice it (surprisingly) generates
+            #   smaller results than when it was previously set to '3'.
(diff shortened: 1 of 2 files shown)


JamesHollyer (approved):
You are saying that compressing the binary more times actually increases the size of the binary? Hmmm seems like a Terser bug to me. Might even be worth reporting it to them?

This change seems like a win, win, win. LGTM

PR #5503, opened 2022-01-14
layout: adjust contain style to show the expander

japie1235813:
When the sidebar got resized and the width is small enough, we collapsed it. The bug is the expander used to reopen the sidebar disappears and the sidebar can never be expanded again. (What's worse is we remember this setting in local storage)

before 
Screen Shot 2022-01-13 at 4 50 18 PM

after
Screen Shot 2022-01-13 at 4 43 25 PM

The contain style is used to optimize the rendering, indicating the element might be independent of the rest of the dom tree. But for the expander, the width can't be sized without its descendants' sizes (the icon). Updates the contain to remove size from the consideration. `contain: content;` is equivalent to `contain: layout paint`

### DIFF
--- a/tensorboard/webapp/core/views/layout_container.scss
+++ b/tensorboard/webapp/core/views/layout_container.scss
@@ -30,7 +30,6 @@ limitations under the License.
 .expand {
   @include tb-theme-foreground-prop(border-color, border);
   box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


bmd3k (approved):
Thanks! I'll cherry-pick this into the 2.8.0 release.

PR #5504, opened 2022-01-14
Fix errors in copybara import.

yatbear:
Current sync failure: cl/421703176

- Use `tensorboard.compat.tensorflow_stub.errors` instead of `tf.errors`
  - To fix: `module 'tensorflow' has no attribute 'errors'`
- Properly mock `import` errors
- Remove `autospec=True` for failed tests:
  -  `testCheckFilesystemSupport_importTFIO` and `testCheckFilesystemSupport_raiseError`

Tested internally:  cl/421719230

### DIFF
--- a/tensorboard/backend/event_processing/data_ingester.py
+++ b/tensorboard/backend/event_processing/data_ingester.py
@@ -23,6 +23,8 @@ import time
 from tensorboard.backend.event_processing import data_provider
 from tensorboard.backend.event_processing import plugin_event_multiplexer
 from tensorboard.backend.event_processing import tag_types
+from tensorboard.compat import tf
+from tensorboard.compat.tensorflow_stub import errors
 from tensorboard.data import ingester
 from tensorboard.plugins.audio import metadata as audio_metadata
 from tensorboard.plugins.histogram import metadata as histogram_metadata
@@ -30,7 +32,6 @@ from tensorboard.plugins.image import metadata as image_metadata
 from tensorboard.plugins.pr_curve import metadata as pr_curve_metadata
 from tensorboard.plugins.scalar import metadata as scalar_metadata
 from tensorboard.util import tb_logging
-from tensorboard.compat import tf
 
 
 DEFAULT_SIZE_GUIDANCE = {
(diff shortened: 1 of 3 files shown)


tensorboard/backend/event_processing/data_ingester.py
```diff
@@ -273,4 +273,4 @@ def _check_filesystem_support(paths):
                     + " filesystem support, consider installing TensorFlow I/O"
                     + " (https://www.tensorflow.org/io) via `pip install tensorflow-io`."
                 ).format(missing_scheme, supported_schemes_msg),
-            )
+            ) from None
```
nfelt, line 276: typo?

tensorboard/backend/event_processing/data_ingester_test.py
```diff
@@ -252,45 +255,71 @@ def testSingleLetterGroup(self):
             )
 
 
-class FileSystemSupport(tb_test.TestCase):
+class ImportMock(object):
```
nfelt, line 258: This approach is fine, but FWIW, it shouldn't really be necessary to create a persistent object (and store it on the test class) for this, since in Python it's quite easy to make nested functions and closures.  So you could consider an approach like this which creates a general utility:

```python
_ORIGINAL_IMPORT = __import__

def fake_import(affected_name, *, success=True):
    def fake(name, *args, **kwargs):
        if name == affected_name:
            if success:
                return
            else:
                raise ImportError("Pretending I'm missing ;)")
        return _ORIGINAL_IMPORT(name, *args, **kwargs)
    return fake
```

Then in your code you could use it like this:

```python
with mock.patch(
    "builtins.__import__",
    side_effect=fake_import("tensorflow_io", success=True),
) as mock_import:
    # do stuff
```

nfelt (approved):

tensorboard/backend/event_processing/data_ingester.py
nfelt: Ah never mind I misread this and didn't realize it was attached to the `raise` (it's hidden in the GH diff).

That said, you might actually want to raise it `from e` where `e` is the original ImportError, since it shows the full trace then but with a clearer message ("The above exception was the direct cause of the following exception:" versus "During handling...").  Up to you though.

yatbear: Done.

tensorboard/backend/event_processing/data_ingester_test.py
yatbear: Done. Thanks!

PR #5508, opened 2022-01-18
time-namespaced state: fix experiment data initial loading on timeNamespacedState enabled

japie1235813:
fix experiment data initial loading on timeNamespacedState enabled.

This only happens when the browser is initiated on experiment view and timeNamespacedState is enabled from the param. `NamespaceUpdateOption` is set to `NEW` initially and gets updated to `UNCHANGED` on any navigation afterwards.
However in loading experiment view initially there is no navigation actions therefore the option remains `NEW`, which generates a new namespaceId. 
In the fix we set the value to `UNCHANGED` after the first run of app navigation flow. However this is tircky to add the unit tests. InitAction perfectly ran only once of the changed part and trigging any follow up actions triggers navigation$.
Add webtests to catch this. (googlers, please see cl/421637297)

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -466,7 +466,6 @@ export class AppRoutingEffects {
                           option: NamespaceUpdateOption.UNCHANGED,
                         },
                         replaceState: true,
-                        resetNamespacedState: false,
                       },
               };
             })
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -462,6 +462,9 @@ export class AppRoutingEffects {
                     ? options
                     : {
                         ...options,
+                        namespaceUpdate: {
+                          option: NamespaceUpdateOption.UNCHANGED,
+                        },
                         replaceState: true,
                         resetNamespacedState: false,
```
bmd3k, line 469: If you don't mind doing it while you're here, I think we can also delete resetNamespacedState here. Seems like it is forgotten cleanup from a previous refactor. As far as I can tell it is not actually part of NavigationOptions and it is not used anywhere.

bmd3k (approved):
Thanks!

PR #5511, opened 2022-01-19
time-namespaced state: generalize navigated callback function on createNamespaceContextedState

japie1235813:
Instead of listening on navigated action in feature reducer, we should implement `onRouteKindOrExperimentsChanged` and pass into `createNamespaceContextedState` which then runs on route (rout kind or experiments) changed. 

However it might be unclear for reducers that the state it returns only takes place when route changed (if there is no route change, the new state is not effective)
We tgeb move this logic out of app_routing and let feature reducers decide what they want to do on `navigated` action.

For the current metrics/ and runs/, the states are updated on route changed.

### DIFF
--- a/tensorboard/webapp/app_routing/namespaced_state_reducer_helper.ts
+++ b/tensorboard/webapp/app_routing/namespaced_state_reducer_helper.ts
@@ -53,7 +53,6 @@ limitations under the License.
 
 import {ActionReducer, createReducer, on} from '@ngrx/store';
 import {navigated} from './actions';
-import {areSameRouteKindAndExperiments} from './internal_utils';
 import {Route} from './types';
 
 // `privateNamespacedState` loosely typed only for ease of writing tests.
@@ -85,10 +84,12 @@ export type NamespaceContextedState<
  * Utility for managing namespaced state. It returns namespace-contexted
  * `initialState` and `reducers` that help manage the namespaced state.
  *
- * An optional `onRouteKindOrExperimentsChanged` function allows more
- * fine-grained changes on state after a key change in Route - that is, after
- * navigating to a Route with either a change in route kind or in the set of
- * experiments. It guarantees that the change happens after namespaced state has
+ * An optional `onNavigated` function allows more fine-grained changes to state
+ * based on the routes information. This function gurantees in app navigation
+ * happens and has finished. Instead of listening to navigated action on feature
+ * reducers, we should implement this `OnNavigated` function, which guarantees
+ * that the implmentation runs after route completely navigated.
+ * It guarantees that the change happens after namespaced state has
  * been properly cached or reset, if appropriate. Note that this callback may be
  * called even when namespace is not changed -- not all navigations to Routes
  * lead to namespaces changes.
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/app_routing/namespaced_state_reducer_helper.ts
```diff
@@ -85,10 +84,12 @@ export type NamespaceContextedState<
  * Utility for managing namespaced state. It returns namespace-contexted
  * `initialState` and `reducers` that help manage the namespaced state.
  *
- * An optional `onRouteKindOrExperimentsChanged` function allows more
- * fine-grained changes on state after a key change in Route - that is, after
- * navigating to a Route with either a change in route kind or in the set of
- * experiments. It guarantees that the change happens after namespaced state has
+ * An optional `onNavigated` function allows more fine-grained changes to state
+ * based on the routes information. This function gurantees in app navigation
```
bmd3k, line 88: "This function gurantees in app navigation happens and has finished." - I think this sentence is redundant with the very next sentence. So probably could remove and keep the documentation a little shorter?

tensorboard/webapp/app_routing/namespaced_state_reducer_helper.ts
```diff
@@ -85,10 +84,12 @@ export type NamespaceContextedState<
...
+ * happens and has finished. Instead of listening to navigated action on feature
+ * reducers, we should implement this `OnNavigated` function, which guarantees
```
bmd3k, line 90: `OnNavigated` => `onNavigated`.

tensorboard/webapp/app_routing/namespaced_state_reducer_helper.ts
```diff
@@ -85,10 +84,12 @@ export type NamespaceContextedState<
  * Utility for managing namespaced state. It returns namespace-contexted
  * `initialState` and `reducers` that help manage the namespaced state.
  *
- * An optional `onRouteKindOrExperimentsChanged` function allows more
- * fine-grained changes on state after a key change in Route - that is, after
- * navigating to a Route with either a change in route kind or in the set of
- * experiments. It guarantees that the change happens after namespaced state has
+ * An optional `onNavigated` function allows more fine-grained changes to state
+ * based on the routes information. This function gurantees in app navigation
+ * happens and has finished. Instead of listening to navigated action on feature
```
bmd3k, line 89: I recommend wrapping navigated in backquotes. `navigated`.

tensorboard/webapp/app_routing/namespaced_state_reducer_helper.ts
```diff
@@ -85,10 +84,12 @@ export type NamespaceContextedState<
...
+ * reducers, we should implement this `OnNavigated` function, which guarantees
+ * that the implmentation runs after route completely navigated.
```
bmd3k, line 91: implmentation => implementation

tensorboard/webapp/app_routing/namespaced_state_reducer_helper.ts
```diff
@@ -99,10 +100,13 @@ export type NamespaceContextedState<
  *    createNamespaceContextedState(
  *        {myNamespacedState: 0},
  *        {nonNamespacedState: 'one'},
- *        (state) => {
+ *        (state, oldRoute, newRoute) => {
  *          // Perform more complex state transformations based on route kind
  *          // or the set of experiments.
- *          return {nonNamespacedState: 'one'};
+ *          if (!areSameRouteKindAndExperiments(oldRoute, newRoute)) {
+ *            return {nonNamespacedState: 'one'};
```
bmd3k, line 107: This is my fault: No reason it needs to be "nonNamespaced"State I guess. That's a bit misleading. Perhaps just rename to "myState"

bmd3k (approved):

PR #5527, opened 2022-01-24
program: centralize logic for when data server is used

nfelt:
This refactoring centralizes the logic we use to decide whether we should use the data server (Rustboard) when `--load_fast=auto` (the default), so that the decision is made entirely in one place.  This also simplifies the branching so that there only needs to be one place where we create the local data ingester.

This is motivated by a coming change that will add another branch to this logic.

While I was there, I also replaced "multiplexer" with "slower Python-only load path" since the term "multiplexer" won't really mean anything to a typical end user.

Tested: ran tensorboard with a few different sets of flag values, confirmed that data server was activated or not activated as intended, and that log messages were printed.

### DIFF
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -432,16 +432,7 @@ class TensorBoard(object):
                 sys.stderr.write(msg)
                 sys.exit(1)
 
-        if flags.load_fast == "auto" and _should_use_data_server(flags.logdir):
-            if flags.logdir_spec and not flags.logdir:
-                logger.info(
-                    "Warning: --logdir_spec is not supported with --load_fast "
-                    + "behavior; falling back to multiplexer. To use the data "
-                    + "server, replace --logdir_spec with --logdir."
-                )
-                ingester = local_ingester.LocalDataIngester(flags)
-                ingester.start()
-                return ingester
+        if flags.load_fast == "auto" and _should_use_data_server(flags):
             try:
                 ingester = self._start_subprocess_data_ingester()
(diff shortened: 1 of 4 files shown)


tensorboard/program.py
```diff
@@ -432,16 +432,7 @@ def _make_data_ingester(self):
...
-        if flags.load_fast == "auto" and _should_use_data_server(flags.logdir):
-            if flags.logdir_spec and not flags.logdir:
-                logger.info(
-                    "Warning: --logdir_spec is not supported with --load_fast "
-                    + "behavior; falling back to multiplexer. To use the data "
-                    + "server, replace --logdir_spec with --logdir."
-                )
-                ingester = local_ingester.LocalDataIngester(flags)
-                ingester.start()
-                return ingester
```
japie1235813, line 444: not quire sure what `ingester` is about but it seems missing on the changes. Is it the legacy mode mentioned in line 486?

nfelt: Yes, there used to be 2 copies of the logic to start LocalDataIngester (which is the old non-RustBoard Python logic that loads event files using EventMultiplexer).  There was one copy here (under "load_fast == auto"), if we couldn't actually use RustBoard, and then one copy as the final fallback down at the end of this function (lines 448-450 in the new version of the file).

The centralization of the logic that I'm doing in this PR consolidates those 2 copies into 1, so that the single fallback at the end of the function works for both cases.  That's because the expanded version of `_should_use_data_server()` now includes the check here (`if flags.logdir_spec and not flags.logdir`).

japie1235813 (approved):

Issue #349, opened 2017-08-12
Fails to show new data when event files are replaced (e.g. w/ rsync)


potiuk:
# Problem summary
We have a problem that tensorboard (1.2.1) does not show new run data when we switched to syncing the data from Google Cloud Storage to local directory.

It looks like when the tensorboard reads an event file from local directory - it will not notice that the event file was deleted and recreated (which is quite valid case when you are using gsutil rsync to sync the data from Google Cloud Storage). I have a short-term workaround (to be tested) but I think there is a need to fiix it permanently in tensorboard - it's quite a valid case to handle due to another bug - #158)  

# Some context and problem manifestation

Initially we stored data from our ML runs in Google Cloud Storage and pointed our Tensorboard to reead the the data directly from there (logdir=gs://BUCKET/FOLDER) . We run the Tensorboard in Kubernetes cluster and made it available to all our team members via oauth proxy. This is great as we can always see fresh data and any member of the team can access it there. We wanted the data to refresh quite frequently - every few minutes. This worked very well but we found that there is a cost downside to it.

Due to #158 we had a lot of GCS accesses generated by TB just scanning GCS (several millions of accesses/day). This was not only slow but also incurred relatively high GCS cost (20 USD/4 days!). 

Then we switched to another approach - we sync the data to a local folder from gcs using `gsutil -m rsync -d gs://BUCKET/FOLDER LOCAL_FOLDER`and then tensorboard points to that locally downloaded folder instead of pointing to gs:// . 

Additional information - we have the whole tree structure within that logdir directory so that we can see and compare all the different runs we've already produced. 

Unfortunately after switching to rsync we seem to have big problem. The new runs (in new subfolders) are showing up amongst the others after the runs are executed, but they never show any data for those runs until tensorboard gets restarted. It looks like tensorboard picks up some initially synced partial results from the run - but not enough to show the data -  and then it somehow blocks from reading any new data for that folder.

When I start an exact copy of Tensorboard at the same machine, it shows the data for those new runs properly - you can see the screenshot below showing the two tensorboards side-by-side. Tensorboard on the left is freshly started, tensorboard on the right has been run before the run has been started.
![tensorboards_left_good_right_stalled](https://user-images.githubusercontent.com/595491/29238598-92daec54-7f39-11e7-9078-07a46b3fe285.png)

# Result of investigation
[...]

PR #5529, opened 2022-01-25
backend: implement event file replacement detection

nfelt:
This implements an optional mode in our Python `EventFileLoader` that will attempt to detect and handle the case when an event file is replaced entirely with a new version of the file containing additional data (rather than the new data being appended to the existing file, the normal behavior we expect).

This situation occurs for example when using `rsync` as described in #349.  When run without the `--inplace` option, `rsync` will transfer data to a temporary file and then swap it in for the final file, but this means that TensorBoard will only ever see the first version of that file and never shows any updates to it until it's restarted.  A similar issue happens for internal users - Googlers, please see b/201113906 for context.

The new logic uses `stat()` to monitor the size of the file, and if it grows, we will re-open the underlying iterator for that filename, resuming at the prior read offset. Care has been taken to try to avoid adding too much filesystem I/O overhead, but it's possible that this new mode is more expensive (since it re-opens the file potentially many times), hence the reason to leave it off by default, at least for now.

This functionality requires that the underlying iterator supports re-opening, which is under development for `tf.compat.v1.summary.tf_record_iterator()` (Googlers, please see cl/423871645).  It is _not_ currently supported for the no-TF mode stub iterator implementation (although I suspect, but have not confirmed, that the stub implementation is not affected by the original issue since it has a simpler albeit likely less efficient implementation that just re-opens files each time it reads).  Also, since this code change is currently only in our Python event loading code, it has no effect on RustBoard (our rust data server).  If there's demand, however, it should be possible to port this logic.

Note that there is currently no way to activate this mode - #5530 is the followup PR with the CLI flag and plumbing.

Tested: ran TensorBoard with these changes enabled (see followup PR) and confirmed that it now picks up new data when replacing an event file entirely rather than appending to it, writes the appropriate log messages, etc.

### DIFF
--- a/tensorboard/backend/event_processing/BUILD
+++ b/tensorboard/backend/event_processing/BUILD
@@ -186,6 +186,7 @@ py_test(
         ":event_file_loader",
         "//tensorboard:expect_tensorflow_installed",
         "//tensorboard:test",
+        "//tensorboard/compat",
         "//tensorboard/compat/proto:protos_all_py_pb2",
         "//tensorboard/summary/writer",
     ],
@@ -200,6 +201,7 @@ py_test(
     deps = [
         ":event_file_loader",
         "//tensorboard:test",
+        "//tensorboard/compat",
         "//tensorboard/compat:no_tensorflow",
         "//tensorboard/compat/proto:protos_all_py_pb2",
         "//tensorboard/summary/writer",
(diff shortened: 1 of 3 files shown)


nfelt:
FYI @bileschi - the CI test failures are expected until tf-nightly contains the changes to `tf_record_iterator` (see cl/423871645), but I thought I'd send it now anyway in case you have questions about the approach.  I can re-ping when CI is fully passing.

bileschi (approved):
Very nice tests

PR #5534, opened 2022-01-26
chore: Enable useUnknownInCatchVariables for polymer code.

bmd3k:
The internal languages team wants to turn on the useUnknownInCatchVariables compiler option and asked us to upstream some changes. Googlers, see: http://b/215141918 and http://cl/420753218.

This impacts the polymer codebase only, which compiles in a non-strict mode (whereas the Angular codebase is compiled in strict mode and, so, was already subject to useUnknownInCatchVariables).

I turned on the useUnknownInCatchVariables and tried to run the app. I fixed each error by typing the caught exception as `: any`. I then cross referenced with http://cl/420753218 to ensure I made the correct changes.


JamesHollyer (approved):
LGTM

PR #5535, opened 2022-01-26
time-namespaced: adding unique index on generating name space id

japie1235813:
To avoid repeated timestamp causing name space id collision, we add a random generated prefix.

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -105,10 +105,15 @@ type NavigationOptions = {
   namespaceUpdate: NamespaceUpdate;
 };
 
-function generate32bitRandomId() {
-  const arr = new Uint8Array(32);
-
+function getRandomValues(arr: Uint8Array) {
   crypto.getRandomValues(arr);
+  return arr;
+}
+
+function generate32bitRandomId() {
+  let arr = new Uint8Array(32);
+
+  arr = getRandomValues(arr);
 
   let ret = '';
   for (const el of arr) {
@@ -643,4 +648,4 @@ function getAfterNamespaceId(
   }
 }
 
-export const TEST_ONLY = {initAction, generate32bitRandomId};
+export const TEST_ONLY = {initAction, generate32bitRandomId, getRandomValues};
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -643,4 +643,4 @@ function getAfterNamespaceId(
   }
 }
 
-export const TEST_ONLY = {initAction};
+export const TEST_ONLY = {initAction, generate32bitRandomId};
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -43,6 +43,7 @@ import {
   arePathsAndQueryParamsEqual,
   areSameRouteKindAndExperiments,
   canRehydrateDeepLink,
+  generateRandomIdForNamespace,
   getRouteNamespaceId,
   serializeCompareExperimentParams,
 } from '../internal_utils';
@@ -103,21 +104,6 @@ type NavigationOptions = {
   browserInitiated: boolean;
   replaceState: boolean;
   namespaceUpdate: NamespaceUpdate;
-};
-
-const generate32bitRandomId = () => {
-  let arr = new Uint8Array(32);
-
-  // The usage of Crypto `getRandomValues` does not require return an array.
-  // This is for the convenience of mock the funciton in testing.
-  arr = crypto.getRandomValues(arr);
-
-  let ret = '';
-  for (const el of arr) {
-    ret += (el >> 4).toString(16);
-  }
-
-  return ret;
 };
 
 /**
@@ -638,11 +624,11 @@ function getAfterNamespaceId(
       beforeNamespaceId == null ||
       options.namespaceUpdate.option === NamespaceUpdateOption.NEW
     ) {
-      return `${generate32bitRandomId()}:${Date.now().toString()}`;
+      return `${Date.now().toString()}:${generateRandomIdForNamespace()}`;
     } else {
       return beforeNamespaceId;
     }
   }
 }
 
-export const TEST_ONLY = {initAction, generate32bitRandomId};
+export const TEST_ONLY = {initAction};
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -105,17 +105,13 @@ type NavigationOptions = {
   namespaceUpdate: NamespaceUpdate;
 };
 
-const getRandomValues = (arr: Uint8Array) => {
-  // The usage of Crypto `getRandomValues` does not require return array.
-  // This is for the convenience of mock the funciton in testing.
-  arr = crypto.getRandomValues(arr);
-  return arr;
-};
 
 const generate32bitRandomId = () => {
   let arr = new Uint8Array(32);
 
-  arr = getRandomValues(arr);
+  // The usage of Crypto `getRandomValues` does not require return an array.
+  // This is for the convenience of mock the funciton in testing.
+  arr = crypto.getRandomValues(arr);
 
   let ret = '';
   for (const el of arr) {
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -655,16 +665,17 @@ describe('app_routing_effects', () => {
...
+            afterNamespaceId: `023c:${beforeDate + 5000}`,
           }),
         ]);
       }));
 
       it('are reused on navigationRequested when resetNamespacedState is false', fakeAsync(() => {
         // Record initial time for use later.
-        const before = Date.now();
+        // '023c' is the string with base 16 convertd from TEST_UNIT8ARRAY
+        const before = `023c:${Date.now()}`;
```
bmd3k, line 678: I prefer the type of change you made in other places, where this was renamed beforeDate and then the random id and beforeDate were combined later in the expect() call.

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -700,7 +711,8 @@ describe('app_routing_effects', () => {
 
       it('are reused on navigationRequested when resetNamespacedState is unspecified', fakeAsync(() => {
         // Record initial time for use later.
-        const before = Date.now();
+        // '023c' is the string with base 16 convertd from TEST_UNIT8ARRAY
+        const before = `023c:${Date.now()}`;
```
bmd3k, line 715: Same: Just define beforeDate here and combine in the expect()?

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -787,16 +799,18 @@ describe('app_routing_effects', () => {
...
+            afterNamespaceId: `023c:${beforeDate + 5000}`,
           }),
         ]);
       }));
 
       it('are reused on programmatical navigation when resetNamespacedState is false', fakeAsync(() => {
         // Record initial time for use later.
-        const before = Date.now();
+        // '023c' is the string with base 16 convertd from TEST_UNIT8ARRAY
+        const before = `023c:${Date.now()}`;
```
bmd3k, line 813: Same as above.

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -828,7 +842,8 @@ describe('app_routing_effects', () => {
 
       it('are reused on programmatical navigation when resetNamespacedState is unspecified', fakeAsync(() => {
         // Record initial time for use later.
-        const before = Date.now();
+        // '023c' is the string with base 16 convertd from TEST_UNIT8ARRAY
+        const before = `023c:${Date.now()}`;
```
bmd3k, line 846: Same as above.

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -102,6 +102,21 @@ type NavigationOptions = {
   namespaceUpdate: NamespaceUpdate;
 };
 
+const generate32bitRandomId = () => {
+  let arr = new Uint8Array(32);
+
+  // The usage of Crypto `getRandomValues` does not require return an array.
+  // This is for the convenience of mock the funciton in testing.
+  arr = crypto.getRandomValues(arr);
```
bmd3k, line 110: Did you consider using callFake instead?

For example:

        spyOn(window.crypto, 'getRandomValues').and.callFake((arr) => {
          if (!(arr instanceof Uint8Array)) {
            return arr;
          }
          arr[0] = TEST_UINT8ARRAY[0];
          arr[1] = TEST_UINT8ARRAY[1];
          arr[2] = TEST_UINT8ARRAY[2];
          arr[3] = TEST_UINT8ARRAY[3];
          return arr;
        });

It would remove the awkwardness here at the cost of making your tests more verbose (although you could probably minimize the verbosity with a better callFake() implementation)

japie1235813: callFake was my first attempt and makes more sense to me. Unfortunately I was stock at weird typescript error.
`getRandomValues` is a generic function type `<T extends ArrayBufferView | null >(arr: T) => T`
see the first commit of this pr. both `set`, `fill`, `index` are not compilable.

bmd3k: These lines should fix that for you by telling the compiler to narrow the type:

```
      if (!(arr instanceof Uint8Array)) {
        return arr;
      }
```

tensorboard/webapp/app_routing/internal_utils.ts
```diff
@@ -192,3 +192,27 @@ export function areRoutesEqual(
     return paramA.key === paramB.key && paramA.value === paramB.value;
   });
 }
+
+/**
+ * Generates a 32-character long random id for name space suffix to avoid
```
bmd3k, line 197: "name space" => "namespace"

tensorboard/webapp/app_routing/internal_utils.ts
```diff
@@ -192,3 +192,27 @@ export function areRoutesEqual(
...
+ * collision of timestamp.
+ */
+export function generateRandomIdForNamespace() {
+  // Generates id with 32 numbers/charaters.
```
bmd3k, line 201: Is this comment duplicate of the function doc? Or is it supposed to describe specifically the next line? (If the latter, then it is incorrect).

tensorboard/webapp/app_routing/internal_utils.ts
```diff
@@ -192,3 +192,27 @@ export function areRoutesEqual(
...
+  const arr = new Uint8Array(32);
+
+  crypto.getRandomValues(arr);
+
+  let ret = '';
+  for (const el of arr) {
+    // We select base 16 which means the character of id is [0-9a-d].
+    // We only need four bit to convert to base 16 string (2^4) from Uint8array.
```
bmd3k, line 209: Nits:
"bit" => "bits";
"ride" => "rid";
"covert" => "convert";
"woudl" => "would";

tensorboard/webapp/app_routing/internal_utils_test.ts
```diff
@@ -21,6 +21,15 @@ import {
 } from './testing';
 import {RouteKind} from './types';
 
+function getMockReturnValuesFuntion(testUint8array: Uint8Array) {
+  return function <Uint8Array>(arr: Uint8Array): Uint8Array {
+    if (arr instanceof Uint8Array) {
+      arr.set(testUint8array);
+    }
+    return arr;
```
bmd3k, line 29: You could consider throwing an Error here instead of silently doing nothing. It would make clearer that something unexpected has happened.

tensorboard/webapp/app_routing/internal_utils_test.ts
```diff
@@ -406,4 +415,53 @@ describe('app_routing/utils', () => {
...
+  describe('#generateRandomIdForNamespace', () => {
+    let cryptoGetRandomValuesSpy: jasmine.Spy;
+
+    it('returns 32-long id', () => {
+      const result = utils.generateRandomIdForNamespace();
+
+      expect(result.length).toEqual(32);
+    });
+
+    it('returns id witd base 16', () => {
```
bmd3k, line 428: nit: witd => with

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -68,6 +68,10 @@ const testDirtyExperimentsSelector = createSelector(
     } as DirtyUpdates;
   }
 );
+const TEST_UINT8ARRAY = new Uint8Array([12, 34, 50, 160, 200]);
+// '023ac' is the string with base 16 convertd from TEST_UNIT8ARRAY
```
bmd3k, line 72: Nits:
convertd => converted
thr => the

tensorboard/webapp/app_routing/effects/app_routing_effects_test.ts
```diff
@@ -234,6 +238,18 @@ describe('app_routing_effects', () => {
...
       actualActions.push(action);
     });
+
+    const mockRandomValuesFunction = function <Uint8Array>(
+      arr: Uint8Array
+    ): Uint8Array {
+      if (arr instanceof Uint8Array) {
+        arr.set(TEST_UINT8ARRAY);
+      }
+      return arr;
```
bmd3k, line 248: You could consider throwing an Error here instead of silently doing nothing. It would make clearer that something unexpected has happened.

bmd3k (approved):

PR #5536, opened 2022-01-26
Move Feature Flag Data source to feature_flag folder

JamesHollyer:
This is part of a general effort to move these DataSource files to their feature folders.


bmd3k (commented):
Summarising what we discussed offline: We will have to make these changes in a way that doesn't break the internal builds when we import these changes. Will likely require multiple commits.

JamesHollyer:
This PR is too far out of date. If this move happens it will be in a new PR.

Issue #267, opened 2017-07-25
authentication and authorization support


anfeng:
Is there any plan to introduce authentication and authorization support into TensorBoard?
Some simple examples of authorization, 
- Only users belong to a white list could access this TB server.
- Client must within a particular network sub-domain.

jart:
Yes indeed. We're planning to build a TensorBoard service with a new scalable storage layer. It will be multi-tenant with ACLs and all that good stuff. See https://github.com/tensorflow/tensorboard/issues/92

kcgthb:
Mmmh, so now that #92 has been closed because it's "too ambitious", what's the plan to enable authentication?

The lack of any kind of authentication/authorization mechanism makes Tensorboard pretty much unusable outside of a developer laptop. It can't be reasonably used on any shared, multi-tenant environment.

nfelt:
We don't have any near-term plans to add auth mechanisms directly to TensorBoard, but it's possible today to run TensorBoard behind an authenticating reverse proxy like [nginx](https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/
), or since it's a WSGI app under the hood, run a slightly modified TensorBoard with some WSGI auth middleware.

Putting auth directly in TensorBoard would mean taking on a significant extra responsibility (since it's something that would be crucial to get right) and it would be very challenging to build something that would satisfy all use cases, even for example supporting authentication mechanisms for web (e.g. openID) vs offline/private (e.g. dedicated username/pw stored in TensorBoard or something like LDAP) requires substantially different work.  So I suspect that even if we continue with various parts of #92, we would be more likely to support auth via the underlying server platform rather than directly inside TensorBoard.

jdhayes:
How about just a long hash at the end of the URL for a uniqe address, similar to what jupyter-lab does?
This would at least allow HPC facilities to support this software. Using an nginx would not be practical in a shared environment where compute nodes have several users on them at a time.

yoshipon:
Hi @jdhayes, I just found that you can manually set a unique address with `--path_prefix` option.
How about launching TensorBoard with `tensorboard --path_prefix /$(openssl rand -hex 24)` ?

I'm not sure this is secure enough.

PR #5541, opened 2022-01-31
add path_prefix_from_file option, adding security by obscurity. This …

joveh:
…option is helpful on multi-user systems.

https://github.com/tensorflow/tensorboard/issues/267#issuecomment-671820015

We noticed open tensorboards on our shared systems 

* Screenshots of UI changes

path_prefix_from_file is an alternativ to path_prefix, if both are given path_prefix will be used

It would be nice to have a config-file implementation with all options for tensorboard, mybe using json or yaml format


tz-rrze (approved):
A simple but really helpful addition to make TensorBoard usable on a shared multi-user system without severe security concerns! Solves a long standing security issue.

tz-rrze:
@yoshipon: you correctly mentioned in https://github.com/tensorflow/tensorboard/issues/267#issuecomment-671820015 on a suggestion form @jdhayes that `--path_prefix` itself is still insecure on a multi-user system; I hope you like the new approach in this PR to read the hash from a file :-)

ericdnielsen:
We believe the comment https://github.com/tensorflow/tensorboard/issues/267#issuecomment-415829957 is still governing on changes/features requests along these lines.

mahendrapaipuri:
We are facing a similar issue at our HPC center. This is already a good step forward to run TensorBoard on multi-user environment. Is it possible to use the approach of JupyterLab here by providing the path prefix as a token and user needs that token to access TensorBoard?

jdhayes:
Yes, I agree. HPC facilities cannot use NGINX, as the localhost is still exposed to all users on the node. Just a note, that if anyone was going to make it as a service, then NGINX would be a good option to add real authentication.

The Jupyter hash/token is a good method for HPC facilities to obscure the URL such that users do not accidentally intrude. Malicious users are less of an issue, but even they would have difficulty finding the address if a hash/token was added.

If reading from a file is not acceptable, could the `--path_prefix` take the value from an environment variables if not provided on the command line?

joveh:
> Yes, I agree. HPC facilities cannot use NGINX, as the localhost is still exposed to all users on the node. Just a note, that if anyone was going to make it as a service, then NGINX would be a good option to add real authentication.
> 
> The Jupyter hash/token is a good method for HPC facilities to obscure the URL such that users do not accidentally intrude. Malicious users are less of an issue, but even they would have difficulty finding the address if a hash/token was added.
> 
> If reading from a file is not acceptable, could the `--path_prefix` take the value from an environment variables if not provided on the command line?

Please have a look at https://github.com/tensorflow/tensorboard/pull/5570

yoshipon:
@tz-rrze @joveh Thank you for making effort for the problems with HPC systems, and I'm sorry for my late reply.
Although it is unfortunate that this kind of PR does not seem to be merged, I agree that the main developers do not want to increase the security risk of this OSS.

I was using the following `tensorboard-hash` command, which makes no change on the `tensorboard` itself.
```python
#! /usr/bin/env python3

import re
import sys
import uuid
from tensorboard.main import run_main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.argv += ["--path_prefix", f"/{uuid.uuid4().hex}"]

    sys.exit(run_main())
```

However, now I'm not using this hack because (at least some old versions of) RustBoard, which makes awesome performance improvements of TensorBoard, seemed to communicate with the main process through an unauthenticated TCP/IP connection.
I respect for the main developer's policy, and now I'm running TensorBoard on my local PC by mounting the HPC's filesystem with `sshfs`.

PR #5549, opened 2022-02-02
time-namespaced state: Improve how we decide when to rehydrate state from URL.

bmd3k:
There are some browser history navigations where we were not appropriately rehydrating URL state. The root cause was that we were only rehydrating URL state at most one time for each namespace. But, in reality, we might need to do it multiple times: one time for the experiment list view and another time for the dashboard view.

(Googlers, see internal bug b/216829997).

This change fixes the issue by adding another dimension when determining whether we can rehydrate URL state. The dimension is based on `RouteKind` but actually represented by a new `DeepLinkGroup` enum.

The mapping of `RouteKind` to `DeepLinkGroup` is explicitly defined in the app routing code:
* `DeepLinkGroup.EXPERIMENTS` = `[RouteKind.EXPERIMENTS]`
* `DeepLinkGroup.DASHBOARD` = `[RouteKind.EXPERIMENT, RouteKind.COMPARE_EXPERIMENT]`

I considered making `DeepLinkGroup` (or some equivalent) a per-app configuration, either in the route configurations or in the `DeepLinkProvider` interface. But after playing around with these ideas, I realized the API and code would be quite messy for little gain. Besides, we don't currently see a world where the `DeepLinkGroup` definition differs between different versions of Tensorboard.

I also renamed `knownNamespaceIds` to `rehydatedDeepLinks` to make the purpose clearer and narrower.

### DIFF
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -430,6 +430,8 @@ export class AppRoutingEffects {
             rehydratedDeepLinks
           )
         ) {
+          // A deeplink has already been rehydrated for this RouteKind/Namespace
+          // combination so don't do it again.
           return;
         }
 
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -42,7 +42,7 @@ import {DirtyUpdatesRegistryModule} from '../dirty_updates_registry_module';
 import {
   areRoutesEqual,
   areSameRouteKindAndExperiments,
-  canBeRehydrated,
+  canRehydrateDeepLink,
   getRouteNamespaceId,
   serializeCompareExperimentParams,
 } from '../internal_utils';
@@ -422,7 +422,7 @@ export class AppRoutingEffects {
         if (
           options.namespaceUpdate.option ===
             NamespaceUpdateOption.FROM_HISTORY &&
-          !canBeRehydrated(
+          !canRehydrateDeepLink(
             routeMatch.routeKind,
             options.namespaceUpdate.namespaceId,
             rehydratedDeepLinks
--- a/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
+++ b/tensorboard/webapp/app_routing/effects/app_routing_effects.ts
@@ -42,6 +42,7 @@ import {DirtyUpdatesRegistryModule} from '../dirty_updates_registry_module';
 import {
   areRoutesEqual,
   areSameRouteKindAndExperiments,
+  canBeRehydrated,
   getRouteNamespaceId,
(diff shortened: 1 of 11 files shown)


tensorboard/webapp/app_routing/effects/app_routing_effects.ts
```diff
@@ -410,23 +411,22 @@ export class AppRoutingEffects {
...
 
-        const isKnownNamespace =
-          options.namespaceUpdate.option ===
-            NamespaceUpdateOption.FROM_HISTORY &&
-          knownNamespaceIds.has(options.namespaceUpdate.namespaceId);
+        if (!options.browserInitiated || !routeMatch.deepLinkProvider) {
+          return;
+        }
 
         if (
```
japie1235813, line 422: Can we have some comment explaining here? For example, the params in this namespace has been rehydrated once, so we don't do anything here.

tensorboard/webapp/app_routing/internal_utils_test.ts
```diff
@@ -406,4 +406,67 @@ describe('app_routing/utils', () => {
...
+            deepLinkGroup: DeepLinkGroup.EXPERIMENTS,
+            namespaceId: 'namespaceA',
+          },
+          {
+            deepLinkGroup: DeepLinkGroup.EXPERIMENTS,
+            namespaceId: 'namespaceB',
+          },
+        ])
+      ).toBeFalse();
+    });
```
japie1235813, line 470: add a test for routeKind is null?

japie1235813 (approved):
Overall the concept of adding rehydratedDeepLinks is clear and lgtm. Just some small things. I left some comments but leave it to you to make the call.

tensorboard/webapp/app_routing/internal_utils.ts
```diff
@@ -192,3 +194,55 @@ export function areRoutesEqual(
...
+      return DeepLinkGroup.DASHBOARD;
+    case RouteKind.UNKNOWN:
+    case RouteKind.NOT_SET:
+      return null;
+  }
+}
+
+/**
+ * Determines whether a combination of route kind and namespace id should have
+ * a URL deep link rehydrated into state.
```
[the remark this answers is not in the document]
bmd3k: Unfortunately there are some edge cases (for example, when getDeepLinkGroup() returns null) where a deep link has not been rehydrated but you still don't want to rehydrate it.  In such cases hadBeenDeepLinkRehydrated would be misleading name.

tensorboard/webapp/app_routing/internal_utils_test.ts
bmd3k: I added a test for when DeepLinkGroup is null (RouteKind cannot be null).

tensorboard/webapp/app_routing/effects/app_routing_effects.ts
bmd3k: Done.

Issue #5550, opened 2022-02-02
`--samples_per_plugin "images=0"` is documented to show all images, but shows no images at all


bersbersbers:
The TB 2.8.0 help says I can pass `--samples_per_plugin "images=0"` to show all images, but that actually shows none at all.

```shell
tensorboard --help
```
>   `--samples_per_plugin` `SAMPLES_PER_PLUGIN`.
> An optional comma separated list of plugin_name=num_samples pairs to explicitly specify how many samples to keep per tag for that plugin. For unspecified plugins, TensorBoard randomly downsamples logged summaries to reasonable values to prevent out-of-memory errors for long running jobs. This flag allows fine control over that downsampling. **Note that 0 means keep all samples of that type.** For instance "scalars=500,images=0" keeps 500 scalars and all images. Most users should not need to set this flag.

```python
import tensorflow as tf

with tf.summary.create_file_writer("tmp").as_default():
    tf.summary.image("Random image", tf.random.normal((1, 28, 28, 1)), step=0)
```

```shell
tensorboard --logdir tmp # shows the image
tensorboard --logdir tmp --samples_per_plugin "images=1" # shows the image
tensorboard --logdir tmp --samples_per_plugin "images=0" # does not show the image - this is the bug.
```

![image](https://user-images.githubusercontent.com/12128514/152210369-a2b992ff-8771-4f7e-bbe2-5b170a961a5a.png)

japie1235813:
Thanks for filing the bug. Have verified that the bug is reproducible.

nfelt:
Hmm, I think this issue arises because in the new DataProvider world, there isn't really a way to specify "no downsampling at all" since it's a required parameter and intentionally doesn't special-case 0: https://github.com/tensorflow/tensorboard/blob/595dd948d7623ea947d7df16bc48c9652cccbf2c/tensorboard/data/provider.py#L203-L205  So the documentation has gotten stale and out of sync with the implementation.

To truly fix this, we'd need to revisit that constraint and relax it to allow `None` there (presumably with implementation-defined behavior as to how many points it then returns).

It's also worth noting that both the Python-backed read path (using EventMultiplexer) and the RustBoard read path both require being given a "samples per plugin" configuration up-front, before they read the event files, which allows them to control how large a buffer they retain.  So even if we relax the DataProvider contract so you can ask for `downsample=None` or "as many points as are available", it won't really be "all the data" unless the read path was correctly configured to buffer unlimited data for that plugin.

[...]

PR #5551, opened 2022-02-03
images: reset sampling size to default on "images=0"

japie1235813:
When running TensorBoard with param `--samples_per_plugin "image=0"`, the down sampling size is 0 instead of "all samples" as suggested in readme. (https://github.com/tensorflow/tensorboard/issues/5550)
This pr reset it to `_DEFAULT_DOWNSAMPLING` which is still slightly from "keep all samples of that type". Please suggest whether increasing the number of sample makes sense.


nfelt:
So I don't think this solution is right - as you've already noted, resetting to the default is pretty different from making it unlimited.  I added some thoughts to #5550 but basically I think for now it might be easiest to just update the flag documentation to suggest using a very large number (like 1000000000) instead of 0.

japie1235813:
Thanks for replying the issue. Closed this pr.

Issue #5548, opened 2022-02-02
Keras graph visualization fails for models using mixed_float16 policy


bersbersbers:
### Diagnostics

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version e43767ef2b648d0d5d57c00f38ccbd38390e38da

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=9, micro=10, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): posix.uname_result(sysname='Linux', nodename='xxx', release='xxx', version='xxx', machine='x86_64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: False
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
INFO: installed: tensorboard==2.8.0
INFO: installed: tensorflow==2.8.0rc1
INFO: installed: tf-estimator-nightly==2.8.0.dev2021122109
INFO: installed: tensorboard-data-server==0.6.1

--- check: tensorboard_python_version
INFO: tensorboard.version.VERSION: '2.8.0'

--- check: tensorflow_python_version
INFO: tensorflow.__version__: '2.8.0-rc1'
INFO: tensorflow.__git_version__: 'v2.8.0-rc0-49-g244b9d77fd4'

--- check: tensorboard_data_server_version
INFO: data server binary: '/opt/pyenv/versions/3.9.10/lib/python3.9/site-packages/tensorboard_data_server/bin/server'
INFO: data server binary version: b'rustboard 0.6.1'

--- check: tensorboard_binary_path
INFO: which tensorboard: b'/opt/pyenv/versions/3.9.10/bin/tensorboard\n'

--- check: addrinfos
socket.has_ipv6 = True
socket.AF_UNSPEC = <AddressFamily.AF_UNSPEC: 0>
socket.SOCK_STREAM = <SocketKind.SOCK_STREAM: 1>
socket.AI_ADDRCONFIG = <AddressInfo.AI_ADDRCONFIG: 32>
socket.AI_PASSIVE = <AddressInfo.AI_PASSIVE: 1>
Loopback flags: <AddressInfo.AI_ADDRCONFIG: 32>
[...]

PR #5553, opened 2022-02-03
keras: skip config if it's a policy obj

japie1235813:
`layer_config` could be a keras Policy object. Gracefully fall with checking the type. #5548


nfelt (approved):
Thanks!

PR #5554, opened 2022-02-03
Linked Time: Adds Fob dragging feature to histogram.

JamesHollyer:
This change uses mouse events to implement a dragging feature on the linked time fob. This allows users to select and drag the fob to change the selected time. This implementation jumps the fob to the next step with data in that histogram which makes it easier for users to see the actual step data.


tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
```diff
@@ -0,0 +1,190 @@
...
+} from '@angular/core';
+import {LinkedTime} from '../../metrics/types';
+import {ScaleLinear, ScaleTime} from '../../third_party/d3';
+
+export enum AxisDirection {
+  horizontal = 0,
+  vertical,
+}
+
+export enum DraggingFob {
```
japie1235813, line 33: The "Dragging" wording was confusing to me. I think this is used to distinguish which fob the user is interacting with. Could we get ride of the Dragging part and just `Fobs`?

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
```diff
@@ -0,0 +1,190 @@
...
+  none = 0,
+  startFob,
+  endFob,
```
japie1235813, line 36: I think for enum it is better to be all upper case?

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
japie1235813: then we can also drop Fob inside
```
enum Fobs {
NONE,
START,
END,
}
```

japie1235813 (commented):
thanks for making the changes! I have not finished my review. I will have another run of review ideally tomorrow

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_test.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    newLinkedTimeActionSpy = jasmine.createSpy();
+    fixture.componentInstance.onSelectedTimeChangedMock =
+      newLinkedTimeActionSpy;
+    newLinkedTimeActionSpy.and.stub();
+
+    return fixture;
+  }
+
+  it('moves the start fob down to mouse when mouse is dragging down and is below fob', () => {
```
japie1235813, line 102: Does this mean the mouse is not dragging the fob "in a correct way"? What looks to me is, the mouse is below the fob a bit but still touches the fob and the mouse is moving up. This should also move the fob. If not, could you explain more on what this test is about. Under what scenario it happens?

JamesHollyer (commented):
Thanks Jie-Wei! I made some changes and tried explaining some things. Let me know what you think.

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_test.ts
```diff
@@ -0,0 +1,319 @@
...
+  @ViewChild('FobController')
+  fobController!: LinkedTimeFobControllerComponent;
+
+  steps!: number[];
+  axisDirection!: AxisDirection;
+  linkedTime!: LinkedTime;
+  temporalScale!: TemporalScale;
+
+  onSelectedTimeChangedMock(newLinkedTime: LinkedTime) {
+    this.linkedTime = newLinkedTime;
```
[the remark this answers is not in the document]
japie1235813: Please remove the implementation as it is not needed.
`onSelectTimeChanged: (linkedTime: LinkedTime) => void`

Could you add `@Input()` to the params as the example here? 
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/widgets/histogram/histogram_test.ts;l=85-90?q=%22class%20TestableComponent%20%7B%22&ss=tensorflow%2Ftensorboard

japie1235813 (commented):
This is looking good to me.

Could you check again the `onSelectTimeChanged` implementation of TestableComponent?

JamesHollyer (commented):
Thanks for the Review!

japie1235813 (approved):

PR #5556, opened 2022-02-04
core: updates `--samples_per_plugin` help command

japie1235813:
The help doc is stale. "Setting `"images=0"` to show all samples" is not supported anymore. Updates it to match the actual implementation. #5550

### DIFF
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -642,10 +642,10 @@ An optional comma separated list of plugin_name=num_samples pairs to
 explicitly specify how many samples to keep per tag for that plugin. For
 unspecified plugins, TensorBoard randomly downsamples logged summaries
 to reasonable values to prevent out-of-memory errors for long running
-jobs. This flag allows fine control over that downsampling. Note that 0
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/core/core_plugin.py
```diff
@@ -642,10 +642,10 @@ def define_flags(self, parser):
 explicitly specify how many samples to keep per tag for that plugin. For
 unspecified plugins, TensorBoard randomly downsamples logged summaries
 to reasonable values to prevent out-of-memory errors for long running
-jobs. This flag allows fine control over that downsampling. Note that 0
-means keep all samples of that type. For instance "scalars=500,images=0"
-keeps 500 scalars and all images. Most users should not need to set this
-flag.\
+jobs. This flag allows fine control over that downsampling. Note that
+downsampling is required for all plugins. Without specifying the numbers
```
nfelt, line 646: Hmm, I might word the sentence here slightly differently, like "If a plugin is not specified in this list, a plugin-specific default number of samples will be enforced."

Also I think it might be better to leave out the actual default numbers (since if we change them, we might forget to update this), or if you do include them at least mention something like "for example".

nfelt (approved):
Thanks!

PR #5560, opened 2022-02-08
backend: load the lazy-loaded TF module before mocking them

yatbear:
This fixes an internal order-dependent error (b/218420006). Mocking lazy-loaded TF symbols before actually loading them is problematic, the mocked modules won't unmock properly. In the future we should avoid mocking such symbols.

Googlers, see b/150299895 for more details.


nfelt (approved):
Nice, thanks!

PR #5561, opened 2022-02-08
infra: set protobuf version to 3.9.2 and downgrade gRPC

yatbear:
[tb-nightly 2.9.0a20220208](https://pypi.org/project/tb-nightly/2.9.0a20220208/) broke Keras Github CI, possibly due to the change https://github.com/tensorflow/tensorboard/pull/5528. We are depending on `gRPC v1.42.0` which is currently on Protobuf 3.18.1 ([commit](https://github.com/protocolbuffers/protobuf/commit/0dab03ba7bc438d7ba3eac2b2c1eb39ed520f928)). 

We are hoping to be able to explicitly define our own `com_google_protobuf` dep at `v3.9.2` and overwrite the gRPC one, to match with TF's protobuf version. Since gRPC release after `1.30.0` require protobuf version `3.12.0+` (https://github.com/grpc/grpc/pull/23324), gRPC version is downgraded to commit `b54a5b338637f92bfcf4b0bc05e0f57a5fd8fadd`, and `rules_apple` is added as a result.


WORKSPACE
```diff
@@ -97,6 +97,16 @@ load("@io_bazel_rules_sass//:defs.bzl", "sass_repositories")
 
 sass_repositories()
 
+http_archive(
```
nfelt, line 100: Let's add a note here about updating this, for future reference:

- We should ensure we always bump the requirements.txt `protobuf` dep to be `>=` the version here
- We probably should keep this in sync with TF, even if we're no longer using the org_tensorflow workspace to build, since there are other projects (like Keras) that do, and generating protobuf code at a more recent version than the `protobuf` runtime supplied by TF's bazel build tooling can lead to test failures for those projects.

tensorboard/pip_package/requirements.txt
```diff
@@ -24,7 +24,7 @@ google-auth >= 1.6.3, < 3
 google-auth-oauthlib >= 0.4.1, < 0.5
 markdown >= 2.6.8
 numpy >= 1.12.0
-protobuf >= 3.6.0
+protobuf >= 3.12.0
```
nfelt, line 27: Might as well just make this 3.9.2 - no reason to be more strict than necessary.

nfelt (approved):
Thanks for doing this!!

WORKSPACE
yatbear: Thanks for the note!

tensorboard/pip_package/requirements.txt
yatbear: Done.

nfelt (commented):
LGTM :)

WORKSPACE
```diff
@@ -97,14 +106,31 @@ load("@io_bazel_rules_sass//:defs.bzl", "sass_repositories")
...
 http_archive(
     name = "com_github_grpc_grpc",
-    sha256 = "b2f2620c762427bfeeef96a68c1924319f384e877bc0e084487601e4cc6e434c",
-    strip_prefix = "grpc-1.42.0",
+    sha256 = "b956598d8cbe168b5ee717b5dafa56563eb5201a947856a6688bbeac9cac4e1f",
+    strip_prefix = "grpc-b54a5b338637f92bfcf4b0bc05e0f57a5fd8fadd",
     urls = [
-        "http://mirror.tensorflow.org/github.com/grpc/grpc/archive/v1.42.0.tar.gz",
-        "https://github.com/grpc/grpc/archive/v1.42.0.tar.gz",  # 2021-11-17
+        # Same as TF: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace2.bzl#L492
```
nfelt, line 131: Maybe worth a note that we can't upgrade GRPC past 1.30.0 without also bumping protobuf?

(It's not clear that this commit that TF uses is any particular version, but it looks like it's roughly 1.27.0)

nfelt (approved):
Thanks for continuing to chip away at this!

WORKSPACE
yatbear: Done.

PR #5566, opened 2022-02-14
docs: update migration guide to include TF2 invocation in TF1 summary APIs

yatbear:
Preview: https://github.com/yatbear/tensorboard/blob/summary_docs/docs/migrate.ipynb

#tf_summary #v1_to_v2

### DIFF
--- a/docs/migrate.ipynb
+++ b/docs/migrate.ipynb
@@ -269,7 +269,7 @@
         "\n",
         "Note that when TF2 summary implementation is invoked, the return value will be an empty bytestring tensor, to avoid duplicate summary writing. Additionally, the input argument forwarding is best-effort and not all arguments will be preserved (`family` argument will be supported whereas `collections` will be removed).\n",
         "\n",
-        "Example to invoke <a href=\"https://www.tensorflow.org/api_docs/python/tf/summary/scalar\"><code>tf.summary.scalar</code></a> behaviors in <a href=\"https://www.tensorflow.org/api_docs/python/tf/compat/v1/summary\"><code>tf.compat.v1.summary</code></a>:"
+        "Example to invoke <a href=\"https://www.tensorflow.org/api_docs/python/tf/summary/scalar\"><code>tf.summary.scalar</code></a> behaviors in <a href=\"https://www.tensorflow.org/api_docs/python/tf/compat/v1/summary/scalar\"><code>tf.compat.v1.summary.scalar</code></a>:"
       ],
       "cell_type": "markdown",
       "metadata": {}
--- a/docs/migrate.ipynb
+++ b/docs/migrate.ipynb
@@ -267,8 +267,9 @@
         " - A suitable TF2 summary writer exists\n",
         " - A non-empty value for step is set for the writer (using <a href=\"https://www.tensorflow.org/api_docs/python/tf/summary/SummaryWriter#as_default\"><code>tf.summary.SummaryWriter.as_default</code></a>, or alternatively <a href=\"https://www.tensorflow.org/api_docs/python/tf/summary/experimental/set_step\"><code>tf.summary.experimental.set_step</code></a>)\n",
         "\n",
-        "Example to invoke <a href=\"https://www.tensorflow.org/api_docs/python/tf/summary/scalar\"><code>tf.summary.scalar\n",
-        "</code></a> behaviors in <a href=\"https://www.tensorflow.org/api_docs/python/tf/compat/v1/summary\"><code>tf.compat.v1.summary</code></a>:"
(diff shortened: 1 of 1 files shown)


docs/migrate.ipynb
```diff
@@ -254,7 +254,45 @@
...
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "1972f8ff0073"
+      },
+      "source": [
+        "### Partial Migration\n",
+        "\n",
+        "It is possible to invoke TF 2.x behaviors within <a href=\"https://www.tensorflow.org/api_docs/python/tf/compat/v1/summary\"><code>tf.compat.v1.summary</code></a> APIs under the following conditions:\n",
```
nfelt, line 268: I might add one or two extra sentences here to explain why the partial migration would be useful, and also to still encourage the full migration.  E.g. something like:

> To make migration to TF 2.x easier for users of model code that still depends heavily on the TF 1.x summary API logging ops like `tf.compat.v1.summary.scalar()`, it is possible to migrate only the writer APIs first, allowing for individual TF 1.x summary ops inside your model code to be fully migrated at a later point.
>
> To support this style of migration, tf.compat.v1.summary logging APIs will automatically forward to their TF 2.x equivalents under the following conditions:

docs/migrate.ipynb
```diff
@@ -254,7 +254,45 @@
...
+        "\n",
+        " - The outermost context is eager mode\n",
+        " - A suitable TF2 summary writer exists\n",
```
nfelt, line 271: I realize "suitable" was a word I suggested in the API docs, but since we're using it slightly differently here, it might be more precise to say something like "A default TF 2.x summary writer has been set"?

nfelt, line 271: nit: looking over this doc, it seems like we use "TF 2.x" everywhere versus the slightly more casual "TF2". I know we used TF2 in the API docs already, but maybe nicer for consistency to stick w/ TF 2.x throughout here.

docs/migrate.ipynb
```diff
@@ -254,7 +254,45 @@
...
+        " - A non-empty value for step is set for the writer (using <a href=\"https://www.tensorflow.org/api_docs/python/tf/summary/SummaryWriter#as_default\"><code>tf.summary.SummaryWriter.as_default</code></a>, or alternatively <a href=\"https://www.tensorflow.org/api_docs/python/tf/summary/experimental/set_step\"><code>tf.summary.experimental.set_step</code></a>)\n",
+        "\n",
+        "Note that when TF2 summary implementation is invoked, the return value will be an empty bytestring tensor, to avoid duplicate summary writing. Additionally, the input argument forwarding is best-effort and not all arguments will be preserved (`family` argument will be supported whereas `collections` will be removed).\n",
```
nfelt, line 274: I might add "for example" at the start of the parenthetical, just to be clear it's an example and that the family argument isn't the only one being preserved.

docs/migrate.ipynb
```diff
@@ -254,7 +254,45 @@
...
+        "id": "1972f8ff0073"
+      },
+      "source": [
+        "### Partial Migration\n",
+        "\n",
+        "It is possible to invoke TF 2.x behaviors within <a href=\"https://www.tensorflow.org/api_docs/python/tf/compat/v1/summary\"><code>tf.compat.v1.summary</code></a> APIs under the following conditions:\n",
+        "\n",
+        " - The outermost context is eager mode\n",
+        " - A suitable TF2 summary writer exists\n",
+        " - A non-empty value for step is set for the writer (using <a href=\"https://www.tensorflow.org/api_docs/python/tf/summary/SummaryWriter#as_default\"><code>tf.summary.SummaryWriter.as_default</code></a>, or alternatively <a href=\"https://www.tensorflow.org/api_docs/python/tf/summary/experimental/set_step\"><code>tf.summary.experimental.set_step</code></a>)\n",
```
nfelt, line 272: We could also mention `tf.compat.v1.train.*_global_step()` here too, since it does also work. Up to you though (I realize we left it out for the API docs).

nfelt (approved):
Very nice, thanks!

docs/migrate.ipynb
yatbear: Done. Thanks!

docs/migrate.ipynb
yatbear: Done.

docs/migrate.ipynb
yatbear: Done.

docs/migrate.ipynb
yatbear: Thanks for the reminder. In my mind 💭 for some reason I felt that it was updated in the API doc string, but it's not the case in reality. Will also update the API docs later.

PR #5567, opened 2022-02-15
scalar: manually set background to transparent

japie1235813:
The background of charts in scalar card have black backgrounds. It is because of the threejs upgrades https://github.com/tensorflow/tensorboard/commit/420916a6e4c80740432f1e2bd0c83d6dd145da10. Before we find a long-term solution, we can set the background opacity to zero as the temporary fix 

before 
Screen Shot 2022-02-14 at 4 21 45 PM

after
Screen Shot 2022-02-14 at 4 22 58 PM

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
@@ -283,14 +283,10 @@ export class ThreeRenderer implements ObjectRenderer<CacheValue> {
 
     this.renderer = new THREE.WebGLRenderer({
       canvas: canvas as HTMLCanvasElement,
-      context: canvas.getContext('webgl2', {
-        antialias: true,
-        precision: 'highp',
-        alpha: true,
-      } as WebGLContextAttributes) as WebGLRenderingContext,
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/renderer/threejs_renderer.ts
```diff
@@ -289,6 +289,8 @@ export class ThreeRenderer implements ObjectRenderer<CacheValue> {
         alpha: true,
       } as WebGLContextAttributes) as WebGLRenderingContext,
     });
+    // Workaround to fix background transparency is not set propoerly in threejs WebGLRenderer.
```
bmd3k, line 292: Nite: propoerly => properly.

bmd3k (approved):

japie1235813:
From the reply of three.js developers. https://github.com/mrdoob/three.js/issues/23495
The `alpha` and `antialias` params should be moved out of context. 
https://github.com/mrdoob/three.js/blob/766b95bad6ae2c98c5bf1c4c775c7c9e8f72f2d2/src/renderers/WebGLRenderer.js#L62

Also remove `precision` cause it is not WebGLContextAttributes
https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getContextAttributes

PR #5568, opened 2022-02-15
linked time: updates range input to contain single selection

japie1235813:
Now on single selection we only provides a slider to drag while under range selections there are two input area to set the exact selected steps. We want to enable users to set exact number in single slider too. 
By doing this we extend `tb-range-input` component to also cable of rendering single slider. The output of the component now contains `singleValue` in addition to `rangeValues`. In addition, an action is dispatched on upperValue removed. This is for syncing the `useRange` to decide it is single or range selection across the app. (for example, we need to update the fob(s) from two to one or vice versa)

Summary of changes:
* The "use range" checkbox is removed.
* `<mat-slider>` is moved from`settings_view_component` to inside `<tb-range-input>`
* `handleInputChange()` handles the logics shown as below

for min=50, max=1000,
single  [100, X]     → add upper value     → range [100, 500]
range  [100, 500]  → delete upper value → single [100, X]
single [100, X]    → delete lower value → single [50, X]
range  [100, 500]  → delete lower value → range  [50, 500]

demo
https://user-images.githubusercontent.com/1131010/154117923-0ad99b83-b001-4586-a9b9-5fb685be33c9.mov

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -936,10 +936,7 @@ const reducer = createReducer(
 
     // When end step value is provided but useRangeSelectTime is still false,
     // this means the current selection state is single. We flip it to range selection.
-    let useRangeSelectTime = state.useRangeSelectTime;
-    if (!state.useRangeSelectTime && change.endStep) {
-      useRangeSelectTime = true;
-    }
+    const useRangeSelectTime = change.endStep !== undefined;
 
     return {
       ...state,
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -934,8 +934,8 @@ const reducer = createReducer(
       nextEndStep = nextStartStep;
     }
 
-    // When end step value is provided but useRangeSelectTime is still false,
-    // this means the current selection state is single. We flip it to range selection.
+    // If there is no endStep then current selection state is single.
+    // Otherwise selection state is range.
     const useRangeSelectTime = change.endStep !== undefined;
 
     return {
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -934,6 +934,13 @@ const reducer = createReducer(
       nextEndStep = nextStartStep;
     }
 
+    // When end step value is provided but useRangeSelectTime is still false,
+    // this means the current selection state is single. We flip it to range selection.
+    let useRangeSelectTime = state.useRangeSelectTime;
+    if (!state.useRangeSelectTime && change.endStep) {
+      useRangeSelectTime = true;
+    }
+
     return {
       ...state,
       selectTimeEnabled: true,
@@ -945,6 +952,7 @@ const reducer = createReducer(
           step: nextEndStep,
         },
       },
+      useRangeSelectTime,
     };
   }),
   on(actions.useRangeSelectTimeToggled, (state) => {
--- a/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
@@ -2412,10 +2412,33 @@ describe('metrics reducers', () => {
         }
       );
 
-      it('sets `useRangeSelectTime` to true when `endStep` is present', () => {
+      it('sets `useRangeSelectTime` to true when `endStep` is present and selected time is null', () => {
         const beforeState = buildMetricsState({
           useRangeSelectTime: false,
           selectedTime: null,
+        });
+
+        const nextState = reducers(
+          beforeState,
+          actions.timeSelectionChanged({
+            startStep: 2,
+            endStep: 5,
+          })
+        );
+
+        expect(nextState.useRangeSelectTime).toEqual(true);
+      });
+
+      // This test case represents <tb-range-input> renders range slider from single slider.
+      it('sets `useRangeSelectTime` to true when `endStep` is present selected time is not null', () => {
+        const beforeState = buildMetricsState({
+          useRangeSelectTime: false,
+          selectedTime: {
+            start: {step: 0},
+            // When single slider is rendered, the end step is set to step max.
+            // Here set it as an arbitrary number.
+            end: {step: 100},
+          },
         });
 
         const nextState = reducers(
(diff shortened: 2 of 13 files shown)


tensorboard/webapp/metrics/views/right_pane/settings_view_component.ts
```diff
@@ -187,6 +187,11 @@ export class SettingsViewComponent {
     lowerValue: number;
     upperValue: number;
   }) {
+    // When range values are updated but useRange is still false, this means the
+    // current selection state is single. We flip it to range selection.
+    if (!this.useRangeSelectTime) {
+      this.useRangeSelectTimeToggled.emit();
```
bmd3k, line 193: Invoking the spirit of Stephan: You are introducing composite actions (two actions fired in quick succession asking for two different states to be mutated) which is a bit of a code smell that we want to avoid. 

I haven't had enough time to think about it but one idea that comes to mind is to move this logic (or some equivalent) the reducer layer.

tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
```diff
@@ -65,35 +65,19 @@ <h3 class="section-title">General</h3>
...
-            [lowerValue]="selectedTime?.start.step"
-            [upperValue]="selectedTime?.end?.step "
-            (value)="onStepRangeChanged($event)"
-          ></tb-range-input>
-        </ng-template>
+          [lowerValue]="selectedTime?.start.step"
+          [upperValue]="selectedTime?.end?.step"
+          [useRange]="useRangeSelectTime"
+          [inputEnabled]="selectTimeEnabled"
+          (singleValue)="onStepStartChanged($event)"
```
bmd3k, line 77: It seems like you're using two different sets of names to describe the same things:
Is it "singleValueChanged" or "stepStartChanged"?
Is it "rangeValueChanged" or "stepRangeChanged"?
Is it "upperValueRemoved" or "useRangeSelectTimeToggled"?

bmd3k (commented):
Sorry I didn't get enough time to devote to an entire review. I'm mostly focused on the actions/reducers/components layer and will have to look at markup, styling, and tests in a followup.

tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
```diff
@@ -144,7 +144,9 @@
                   [max]="column.filter.maxValue"
                   [lowerValue]="column.filter.filterLowerValue"
                   [upperValue]="column.filter.filterUpperValue"
-                  (value)="handleHparamIntervalChanged(column, $event)"
+                  [useRange]="true"
+                  [inputEnabled]="true"
```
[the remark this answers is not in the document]
japie1235813: They are indeed default true. I just thought it might be clear here but I can remove them

bmd3k (commented):
Thanks. Looking good. Just needs the tests for the metrics reducer (as you already suggested).

tensorboard/webapp/widgets/range_input/range_input_test.ts
```diff
@@ -167,198 +225,329 @@ describe('range input test', () => {
...
 
   describe('input control', () => {
     let fixture: ComponentFixture<TestableComponent>;
-    let onValue: jasmine.Spy;
+    let onRangeValuesChanged: jasmine.Spy;
 
-    beforeEach(() => {
-      const fixtureAndOnValue = createComponent({
+    function createRangeComponent() {
+      const fixtureAndonRangeValuesChanged = createComponent({
```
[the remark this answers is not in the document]
japie1235813: yap I wasn't sure keep it this way or directly use createComponent. Good thing about having a 
"createRangeComponent" function name indicates the component is range selection.
But I can switch it and let the param (useRange: true) speak for itself.

tensorboard/webapp/widgets/range_input/range_input_test.ts
```diff
@@ -167,198 +225,329 @@ describe('range input test', () => {
...
+      minInput.value = '';
+      minInput.dispatchEvent(new InputEvent('change'));
+
+      expect(onRangeValuesChanged).toHaveBeenCalledWith({
+        lowerValue: 0,
+        upperValue: 5,
+      });
+    });
+
+    it('renders range slider on adding upper value to single slider', () => {
```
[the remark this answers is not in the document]
japie1235813: good call. updated.
I was thinking about testing render sliders but then realize rendering which slider is controlled by 'useRange' and it is tested above. There is no way, or no easy way, of testing the logic of switching slider here.

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -934,6 +934,10 @@ const reducer = createReducer(
       nextEndStep = nextStartStep;
     }
 
+    // When end step value is provided but useRangeSelectTime is still false,
```
bmd3k, line 937: Perhaps clearer to avoid mentioning "flipping" (because this line doesn't necessarily flip anything). Something like: If there is no endStep then current selection state is single. Otherwise selection state is range.

tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
```diff
@@ -376,6 +376,15 @@ describe('metrics right_pane', () => {
       });
     });
 
+    it('does not display link time setting on link time disnabled', () => {
```
bmd3k, line 379: disnabled => disabled.

tensorboard/webapp/widgets/range_input/range_input_component.scss
```diff
@@ -48,6 +49,19 @@ input {
   justify-self: flex-end;
 }
 
+.single-slider {
```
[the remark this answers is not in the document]
bmd3k: Ah sorry I forgot that settings_view_component.ts still has its own sliders (they weren't all moved into tb-range-slider) so still needs to style those. Sorry for the confusion.

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -2411,6 +2411,103 @@ describe('metrics reducers', () => {
...
+            startStep: 2,
+            endStep: undefined,
+          })
+        );
+
+        expect(nextState1.useRangeSelectTime).toEqual(false);
+      });
+    });
+
+    it('sets `useRangeSelectTime` to false when `endStep` is undefinded', () => {
```
bmd3k, line 2493: undefinded => undefined.

bmd3k (approved):

PR #5569, opened 2022-02-15
linked time: moves Image thumb slider on single selection

japie1235813:
Previously we have those gray ticks to show the selected step(s) in linked time. This pr moves the thumb to the selected step, the image also changes correspondingly. There is still one tick rendered but it will be covered by orange thumb.

Note that this pr only moves the thumb (controlled by `stepIndex`) when selected step matches. Not move to the closest step. Also because of the nature of image cards (which shows one image at a time) we do not move the thumb on range selection for now before more complicated decision is made. 

demo
https://user-images.githubusercontent.com/1131010/154165453-efe655ff-864a-4af1-90f3-52ab9fa162fa.mov

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
@@ -286,9 +286,9 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
       map(([stepIndex, selectedTime, selectedSteps, stepValues]) => {
         if (!selectedTime || selectedTime.endStep) return stepIndex;
 
-        const nextStepInex = stepValues.indexOf(selectedSteps[0]);
-        if (nextStepInex === -1) return stepIndex;
-        return nextStepInex;
+        const nextStepIndex = stepValues.indexOf(selectedSteps[0]);
+        if (nextStepIndex === -1) return stepIndex;
+        return nextStepIndex;
       })
     );
 
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
@@ -281,13 +281,13 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
       this.store.select(getCardStepIndex, this.cardId),
       this.selectedTime$,
       this.selectedSteps$,
-      this.stepValues$,
+      this.stepValues$
     ).pipe(
       map(([stepIndex, selectedTime, selectedSteps, stepValues]) => {
         if (!selectedTime || selectedTime.endStep) return stepIndex;
 
         const nextStepInex = stepValues.indexOf(selectedSteps[0]);
-        if (nextStepInex === -1 ) return stepIndex;
+        if (nextStepInex === -1) return stepIndex;
         return nextStepInex;
       })
     );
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -299,6 +276,43 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
...
+      map(([stepIndex, selectedTime, selectedSteps, stepValues]) => {
+        if (!selectedTime || selectedTime.endStep) return stepIndex;
+
+        const nextStepInex = stepValues.indexOf(selectedSteps[0]);
+        if (nextStepInex === -1) return stepIndex;
+        return nextStepInex;
+      })
+    );
+
+    const timeSeriesAndStepIndex$ = combineLatest([
```
JamesHollyer, line 295: All of this has not changed. Is there a reason you moved it all? I would rather not mess up the git blame history on these.

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -661,5 +661,76 @@ describe('image card', () => {
...
+        end: null,
+      });
+      store.refreshState();
+      fixture.detectChanges();
+
+      slider = fixture.debugElement.query(By.css('mat-slider'));
+      expect(slider.nativeElement.getAttribute('aria-valuenow')).toBe('2');
+    });
+
+    it('dose not move slider thumb to selected time on single selection', () => {
```
JamesHollyer, line 700: on range selection*?

JamesHollyer (commented):
Looking good. I would also like to have a test for the case when we have a linked time selection that has no image associated with it.

tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
japie1235813: That's a fair question. Variables can only be used after they're defined.  (or they become undefined) 
`timeSeriesAndStepIndex$` needs `stepIndex$` which now needs `selectedSteps$`(and others, but the critical one is selectedSteps$)
that's why `stepIndex$` needed moved down and all other observables followed.

the chain is
selectedSteps$ -> stepIndex$ -> timeSeriesAndStepIndex$ -> stepDatum$ -> imageUrl$

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
japie1235813: Good catch. Thanks!

tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -299,6 +276,43 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
...
+    this.stepIndex$ = combineLatest(
+      this.store.select(getCardStepIndex, this.cardId),
+      this.selectedTime$,
+      this.selectedSteps$,
+      this.stepValues$
+    ).pipe(
+      map(([stepIndex, selectedTime, selectedSteps, stepValues]) => {
+        if (!selectedTime || selectedTime.endStep) return stepIndex;
+
+        const nextStepInex = stepValues.indexOf(selectedSteps[0]);
```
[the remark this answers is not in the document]
japie1235813: oops. updated

tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
JamesHollyer: I see. Well I wish it were not the case but, so be it.

JamesHollyer (approved):
LGTM

PR #5570, opened 2022-02-16
Gen path prefix

joveh:
Following the idea of @[jdhayes](https://github.com/jdhayes) in https://github.com/tensorflow/tensorboard/pull/5541 we implemented the option gen_path_prefix

new option in tensorboard call to improve security on multi-user systems.
* Screenshots of UI changes


bileschi:
Hello @joveh.

We are reluctant to add data-access-security focused features unless we were sure we could handle security well.

We are unlikely to accept this PR as it adds complexity to TensorBoard with little value beyond to serve the security use case, and putting data access security burden on TensorBoard expands our the mandate beyond our ability to execute.  

Is it possible to manage this use case using existing external practices and not depending on OSS TensorBoard to provide load-bearing security guarantees?   For instance, I understand that your particular use case is in a multi-user HPC system.  Is it possible to mount the HPC filesystem using a form of secure remote file system and then run TensorBoard locally (or hosted behind e.g., NGINX)

PR #5581, opened 2022-02-23
linked time: fix range input param rename

japie1235813:
I was renaming the param of range input and overlooked the one in settings.  This pr should fix it.
This was not caught due to unused attr of tb-range-input. This pr also removed useless tests to avoid false tests pass. The elements inside tb-range-input is not testable in settings_view component.

before 
Screen Shot 2022-02-23 at 9 47 07 AM

after (without checked enabled the input stays disabled)
Screen Shot 2022-02-23 at 9 46 04 AM

### DIFF
--- a/tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
+++ b/tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
@@ -448,34 +448,6 @@ describe('metrics right_pane', () => {
           expect(el.query(By.css('tb-range-input'))).toBeTruthy();
         });
 
-        it('disables tb-range-input on select time disabled', () => {
-          store.overrideSelector(selectors.getIsLinkedTimeEnabled, true);
-          store.overrideSelector(selectors.getMetricsXAxisType, XAxisType.STEP);
-          store.overrideSelector(selectors.getMetricsSelectTimeEnabled, false);
-          store.overrideSelector(selectors.getMetricsUseRangeSelectTime, false);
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
```diff
@@ -448,34 +448,6 @@ describe('metrics right_pane', () => {
...
-          store.overrideSelector(selectors.getMetricsXAxisType, XAxisType.STEP);
-          store.overrideSelector(selectors.getMetricsSelectTimeEnabled, false);
-          store.overrideSelector(selectors.getMetricsUseRangeSelectTime, false);
-          const fixture = TestBed.createComponent(SettingsViewContainer);
-          fixture.detectChanges();
-
-          const el = fixture.debugElement.query(
-            By.css('.linked-time tb-range-input')
-          );
-          expect(el.nativeElement.getAttribute('disabled')).toBe('true');
```
bmd3k, line 462: You should be able to do:

```
          expect(el.properties.enabled).toBe(false);
```

bmd3k (commented):
I think the changes to the component make sense but I think the tests can be salvaged.

tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
japie1235813: Ah I did not know this. Updated the task. Thanks!

bmd3k (approved):

PR #5582, opened 2022-02-23
linked time: color image slider in range selection

japie1235813:
Previously we have gray ticks indicate selected steps in range. This pr further colors the range. The coloring also supports selected steps does not have data (image here).

Current Image slider places ticks with the number of steps not the step values makes us cannot directly use end step and start step to calculate the tracker width. I choose to check the unit where start/end step exists and make the width proportional to the value. For example, we have steps [10,20,30,100] and range [20,65]. The range slider looks like:

Screen Shot 2022-02-23 at 1 10 10 PM

The track takes 50% of the unit between step 30 and 100.

Screenshots
Screen Shot 2022-02-23 at 1 02 24 PM

start step and end step in between two steps: track with no gray ticks
Screen Shot 2022-02-23 at 1 01 36 PM

steps are patricianly in range
Screen Shot 2022-02-23 at 1 03 43 PM

Summary of technical changes:
* Hide the slider of `<mat-slider>` (only keeps the thumb) when it is in range selection
* Render slider on range selection1
* Calculate the position and width of 'track-fill' based on start/end steps

What is also included in this pr but not directly related:
the tests of `ticks` in image_card_test is cleaned up by moving the card generation to `beforeEach`

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ts
@@ -83,11 +83,14 @@ export class ImageCardComponent {
 
   ngOnChanges(changes: SimpleChanges) {
     if (
-      !this.changeDistinct(changes['selectedSteps']) ||
-      !this.changeDistinct(changes['selectedTime'])
+      this.changeDistinct(changes['selectedSteps']) ||
+      this.changeDistinct(changes['selectedTime'])
     ) {
-      return;
+      this.renderRangeSlider();
     }
+  }
+
+  renderRangeSlider() {
     if (
       !this.selectedTime ||
       !this.selectedTime.endStep ||
@@ -161,7 +164,6 @@ export class ImageCardComponent {
     this.sliderStartPosition = `${startPosition * 100}%`;
     this.sliderTrackWidth = `${width * 100}%`;
   }
-
   getLinkedTimeTickLeftStyle(step: number) {
     if (this.stepValues.indexOf(step) == -1) {
       throw new Error(
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ts
@@ -101,7 +101,6 @@ export class ImageCardComponent {
     }
 
     const boundSize = this.stepValues.length - 1;
-    const sliderUnit = 1 / boundSize;
     const startStep =
       this.selectedTime.startStep < this.stepValues[0]
         ? this.stepValues[0]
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/metrics/views/card_renderer/image_card_component.ts
```diff
@@ -73,6 +77,94 @@ export class ImageCardComponent {
...
+        : this.selectedTime.startStep;
+    const endStep =
+      this.selectedTime.endStep > this.stepValues[boundSize]
+        ? this.stepValues[boundSize]
+        : this.selectedTime.endStep;
+    let startPosition = 0;
+    let width = 0;
+
+    let i = 0;
+    for (; i < this.stepValues.length - 1; i++) {
```
JamesHollyer, line 117: As discussed in person: I would like to see this logic encapsulated in a GetStartPosition function for readability.  I see that you reuse the i variable so maybe encapsulate it all in a GetStartPositionAndWidth function or just recalculate the i value in the GetSliderWidth function.

japie1235813: Thanks! It's a good suggestion. I prefer not recalculate the width so I put it as `getStartPositionAndWidth()` and add comments before each for loop chunk. PTAL.

tensorboard/webapp/metrics/views/card_renderer/image_card_component.ts
```diff
@@ -73,6 +77,94 @@ export class ImageCardComponent {
...
+      const currentStep = this.stepValues[i];
+      const nextStep = this.stepValues[i + 1];
+      if (startStep > nextStep) {
+        startPosition += sliderUnit;
```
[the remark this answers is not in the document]
japie1235813: Good point! Updated.

tensorboard/webapp/metrics/views/card_renderer/image_card_component.ts
```diff
@@ -73,6 +77,94 @@ export class ImageCardComponent {
...
+      } else if (currentStep <= startStep && startStep <= nextStep) {
+        startPosition +=
+          ((startStep - currentStep) / (nextStep - currentStep)) * sliderUnit;
+        break;
+      } else {
```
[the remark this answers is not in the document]
japie1235813: You're right it's unused. Removed.

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -490,176 +490,287 @@ describe('image card', () => {
...
+        // The third and fourth ticks are selected.
+        expect(dots.length).toBe(0);
+      });
     });
 
-    it('does not render ticks on slected range wrapped between steps ', () => {
-      store.overrideSelector(selectors.getMetricsSelectedTime, {
-        start: {step: 11},
-        end: {step: 14},
+    describe('sliders', () => {
```
JamesHollyer, line 619: Could you add a test for start.step being on a tick and end not on a tick and vice versa.

JamesHollyer (approved):
I think that is more readable now. Thanks for making the changes. I would like to see 2 more tests added. However, I am going to approve this now so you can submit once they are added. They should be very straight forward tests to add.

PR #5583, opened 2022-02-24
linked time: remove useRange input in <tb-range-input>

japie1235813:
When extending `<tb-range-input>` it took `useRange` to determine rendering of single or range slider. This pr further removes the explicit usage of `useRange` param but put upperValue to determine the rendering. That is to say, it is range selection when upperValue is null, otherwise it's single selection.

Summary of key changes:
* `useRange` is removed. The logic is replaced with `upperValue !== null` (equivalent to useRange is true)
* The interface`StoreInternalLinkedTime`, which does not allow end step to be null, is replaced with `LinkedTime`. 
* In reducer, on `timeSelectionChanged` action, we set `nextEndStep` to be null if the change does not have end step instead of step max.


tensorboard/webapp/widgets/range_input/range_input_component.ts
```diff
@@ -345,14 +345,21 @@ export class RangeInputComponent implements OnInit, OnDestroy {
     // range  [100, 500]  → delete upper value → single [100, X]
     // single [100, X]    → delete lower value → single [50, X]
     // range  [100, 500]  → delete lower value → range  [50, 500]
+    // single [100, X]    → add    upper value → range  [100, 500]
     if (this.isUpperValueRemoved(input, position)) {
+      this.upperValue = null;
```
bmd3k, line 350: I don't think we should set this, if it can be avoided. It's an Input parameter - we should rely on ngrx processing the action and the selectors emiting updated values.

tensorboard/webapp/widgets/range_input/range_input_component.ts
```diff
@@ -345,14 +345,21 @@ export class RangeInputComponent implements OnInit, OnDestroy {
...
+      this.singleValueChanged.emit(this.lowerValue);
```
bmd3k, line 351: This introduces composite actions - two actions being fired in quick succession. Can you model the underlying actions differently?

japie1235813: Thanks for pointing this out. That's a great point. Here I then removed the output `upperValueRemoved`.
The output is originally designed for "toggle useRange" and cause `useRangeSelectTime` to change. But then I realized `useRangeSelectTime` is set based on `upperValue !== null`. so on removing upper value, all I need to do is emit the `singleValueChanged` (which set `selectedTme.end` to null, which means upperValue is null and sets `useRangeSelectTime` to false)

To sum,
- upperValueRemoved output is removed
- `this.upperValue = null` is removed

PTAL and let me know what do you think. Thanks!

bmd3k (approved):

tensorboard/webapp/widgets/range_input/range_input_component.ts
```diff
@@ -345,14 +342,19 @@ export class RangeInputComponent implements OnInit, OnDestroy {
...
       return;
     }
-    if (this.isLowerValueUpdatedOnSingleSelection(position, this.useRange)) {
+    if (this.isLowerValueUpdatedOnSingleSelection(position)) {
       this.singleValueChanged.emit(numValue);
       return;
     }
+    if (this.upperValue === null) {
+      this.maybeNotifyNextRangeValues([this.lowerValue, numValue]);
```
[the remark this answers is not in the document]
japie1235813: Thanks! I was a bit struggling at adding on top of the logic in line 359-365. Tearing them apart is actually a good idea. Also thanks for the clear example code. I updated this part.

PR #5586, opened 2022-02-24
sync: fix index signature accessed 

japie1235813:
Property 'enabled' comes from an index signature, so it must be accessed with ['enabled'].
Googlers, please see cl/363052284 for sync result.


bmd3k (approved):
Fun.

PR #5587, opened 2022-02-25
linked time: moves Image thumb on range selection

japie1235813:
The slider thumb of image card on range selection was only moved on single selection (and there is a matched step). This pr moves the thumb on range selection based on the thumb's current position.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
@@ -284,9 +284,32 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
       this.stepValues$
     ).pipe(
       map(([stepIndex, selectedTime, selectedSteps, stepValues]) => {
-        if (!selectedTime || selectedTime.endStep) return stepIndex;
+        const firstSelectedStep =  selectedSteps[0];
+        const lastSelectedStep = selectedSteps[selectedSteps.length - 1];
+        if (!selectedTime || selectedTime.startStep > lastSelectedStep) return stepIndex;
 
-        const nextStepIndex = stepValues.indexOf(selectedSteps[0]);
+        // Moves thumb to the closest stepIndex on range selection.
+        if (selectedTime.endStep !== null && stepIndex !== null) {
+          const step = stepValues[stepIndex];
+
+          if (selectedTime.endStep < firstSelectedStep) {
+            return stepIndex;
+          }
+          // Does not move index when it is already in selected range.
+          if (selectedTime.startStep <= step && step <= selectedTime.endStep) {
+            return stepIndex;
+          }
+
+          if (step >= lastSelectedStep) {
+            return stepValues.indexOf(lastSelectedStep);
+          }
+          if (step <= firstSelectedStep) {
+            return stepValues.indexOf(firstSelectedStep);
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -284,9 +284,30 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
...
     ).pipe(
       map(([stepIndex, selectedTime, selectedSteps, stepValues]) => {
-        if (!selectedTime || selectedTime.endStep) return stepIndex;
+        const firstSelectedStep = selectedSteps[0];
+        const lastSelectedStep = selectedSteps[selectedSteps.length - 1];
+        if (!selectedTime || selectedTime.clipped) return stepIndex;
 
-        const nextStepIndex = stepValues.indexOf(selectedSteps[0]);
+        // Range selection
+        if (selectedTime.endStep !== null && stepIndex !== null) {
```
JamesHollyer, line 292: What does a null stepIndex indicate? When will this happen?

tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -284,9 +284,30 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
       this.stepValues$
     ).pipe(
       map(([stepIndex, selectedTime, selectedSteps, stepValues]) => {
-        if (!selectedTime || selectedTime.endStep) return stepIndex;
+        const firstSelectedStep = selectedSteps[0];
```
JamesHollyer, line 287: AFAICT this could be empty. Do we need to handle that case?

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -757,5 +757,83 @@ describe('image card', () => {
       slider = fixture.debugElement.query(By.css('mat-slider'));
       expect(slider.nativeElement.getAttribute('aria-valuenow')).toBe('2');
     });
+
+    describe('range selection', () => {
```
JamesHollyer, line 761: Can you add a test for when the entire range is outside the range of available images? 
Ex:
Start step = 50
End step = 60

JamesHollyer (commented):
AFAICT there is an edge case that is not handled here. If I am wrong just let me know and add the test.

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
japie1235813: it's covered by `'dose not moves slider thumb when selected time is clipped',` I set start step 55 and end step 65

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -834,6 +834,16 @@ describe('image card', () => {
 
         expect(getSliderThumbPosition(fixture2)).toBe('2');
       });
+
+      it('dose not moves slider thumb when selected range had no image', () => {
```
JamesHollyer, line 838: nit: does*

JamesHollyer (approved):
LGTM after nit.

PR #5589, opened 2022-03-01
Linked Time: Draggable Fob in Scalar Cards

JamesHollyer:
Utilizes the new LinkedTimeFobControllerComponent to implement the fob dragging functionality in the scalar cards.  Since the Scalar card has different types of data I had to create an interface(FobCardData) to handle the 2 different sets of data.


tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
```diff
@@ -112,26 +122,62 @@ export class LinkedTimeFobControllerComponent {
...
+      return minMax[0] < minMax[1] ? minMax[0] : minMax[1];
+    }
+
+    return this.fobCardData.histograms!.steps[0];
+  }
+
+  getDraggingFobCenter(): number {
+    // the fob is centered around the top when in the vertical direction and
+    // around the left when in the horizontal.
+    if (this.getAxisDirection() === AxisDirection.VERTICAL) {
```
japie1235813, line 172: so overall I think using Axis or this.fobCardData.historgram servers the same purpose. Could we stick to one way as the condition?

tensorboard/webapp/widgets/linked_time_fob/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  NONE,
+  START,
+  END,
+}
+
+export type TemporalScale =
+  | ScaleLinear<number, number>
+  | ScaleTime<number, number>;
+
+export interface FobCardData {
```
japie1235813, line 37: Now that I am wondering do we want to allow both keys existing in `FobCardData`?  Probably not.
Sorry I did not think about this clearer. Now what I think is better is
```
interface HistogramCardFobData {
    scale: TemporalScale;
    steps: number[];
}

interface ScalarCardFobData {
    scale: Scale;
    minMax: [number, number];
}

type FobCardData =  HistogramCardFobData | ScalarCardFobData
```

so then instead of checking this.fobCardData.histogram,
we could have an attribute serves as the fobCard type
some pseudo code could be like 
```
private pluginType: PluginType = PluginType.NONE;

if (typeof(fobCardData) === HistogramCardFobData) {
pluginType =  PluginType.HISTOGRAM
}
if (typeof(fobCardData) === ScalarCardFobData) {
pluginType =  PluginType.SCALAR
}
```

then instead of checking `this.fobCardData.historgram` we check `pluginType === PluginType.HISTOGRAM`

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
```diff
@@ -112,26 +122,62 @@ export class LinkedTimeFobControllerComponent {
...
+    if (this.fobCardData.scalars) {
+      let minMax = this.fobCardData.scalars.minMax;
+      return minMax[0] < minMax[1] ? minMax[0] : minMax[1];
+    }
+
+    return this.fobCardData.histograms!.steps[0];
+  }
+
+  getDraggingFobCenter(): number {
+    // the fob is centered around the top when in the vertical direction and
```
japie1235813, line 170: I'm actually a bit confused. Top and center don't sound the same to me. If you mean the center and the top/left is roughly at the same position. Could we just say `getDraggingFobTopOrLeft`?

japie1235813 (commented):
Just reviewed fob_controller component since I assume changes for scalar card and historgram card will follow afterward.

There are a lot of checking this.fobCardData.histogram I am wondering how we could make is more readable.

It would be great if the PR description could provide more context like screenshots for styling change, summary of technical changes and so on either for reviewers not actively in development and for anyone in the future to better understand. This would also help with the review to be faster and easier.

tensorboard/webapp/widgets/linked_time_fob/types.ts
japie1235813: so after rethink this, I think define the axis is better than define the plugin. Using `if(Axis.VERTICAL)` is more meaningful then `if(.. HISTOGRAM)`. Please ignore the second part of the comment above. Instead,
```
private axis: AxisDirection;

switch(typeof(fobCardData)) {
    case HistogramCardFobData:
         axis = AxisDirection.VERTICAL;
    case ScalarCardFobData:
         axis = AxisDirection.HORIZONTAL
}
```
You can put it in `ngOnInit`

JamesHollyer (commented):
I am liking our new FobCardData structure. Let me know what you think!

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
```diff
@@ -140,58 +186,113 @@ export class LinkedTimeFobControllerComponent {
...
+    if (this.fobCardData.scalars) {
+      let minMax = this.fobCardData.scalars.minMax;
+      return minMax[0] < minMax[1] ? minMax[1] : minMax[0];
+    }
+
+    let steps = this.fobCardData.histograms!.steps;
+    return steps[steps.length - 1];
   }
 
   // Gets the index of smallest step that the currentDraggingFob is allowed to go.
```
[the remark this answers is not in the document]
JamesHollyer: Updated. Good catch

tensorboard/webapp/widgets/linked_time_fob/types.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface HistogramFobData {
+  type: PluginType.HISTOGRAMS;
+  scale: TemporalScale;
+  steps: number[];
+}
+
+export interface ScalarFobData {
+  type: PluginType.SCALARS;
+  scale: Scale;
+  minMax: [number, number];
```
bmd3k, line 42: Can ScalarFobData also be modeled with `steps: number[]`? The Scalar charts are also a finite set of integer values. And in some environments the steps might be sampled so they aren't even a complete set of integers in the range [min,max].

tensorboard/webapp/widgets/linked_time_fob/types.ts
```diff
@@ -0,0 +1,45 @@
...
+limitations under the License.
+==============================================================================*/
+
+import {PluginType} from '../../metrics/types';
+import {TemporalScale} from '../histogram/histogram_types';
+import {Scale} from '../line_chart_v2/lib/public_types';
+
+export {Scale} from '../line_chart_v2/lib/public_types';
+
+export enum AxisDirection {
```
bmd3k, line 22: There seems to be a 1:1 relationship with Axis and PluginType at the moment - so maybe FobData should just contain AxisDirection?

tensorboard/webapp/widgets/linked_time_fob/types.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export enum Fob {
+  NONE,
+  START,
+  END,
+}
+
+export interface HistogramFobData {
+  type: PluginType.HISTOGRAMS;
```
bmd3k, line 34: I find it suspicious that we would want or need to know the PluginType. My hope with the other suggestions is that we can eliminate this from the Fob data model and code.

bmd3k (commented):
This is a really large PR. It will require several iterations of back and forth. Even for the small number of comments I am leaving I had to spend quite a bit of time sifting through the code.

Would it have been possible, for example, to at least have broken this PR up like this?
* One PR using linkable_fob in scalars card, addressing only the AxisDirection issue.
* A second PR addressing the issues with the steps/minMax data model.

tensorboard/webapp/widgets/linked_time_fob/types.ts
JamesHollyer: We thought about using the steps from each line. That seemed strange as each line can have different sets of steps. We could use a union of all the steps for each line. However, since there is no highlighting individual steps the continuous smooth dragging that we have implemented here makes the most sense. Lastly, the scale used in the Scalar cards requires this minMax(it is actually not the min and max steps that we have data for but, the min and max steps that are shown on the axis which is slightly different).

tensorboard/webapp/widgets/linked_time_fob/types.ts
JamesHollyer: I am trying to keep things generalized so that this can be used for other cards(like the image card) if we decide to. Note that the data from each card is different so we still need a type that can encompass both sets of data and ideally one that enforces that we have all of one or all of the other.

tensorboard/webapp/widgets/linked_time_fob/types.ts
JamesHollyer: This type attribute is special and very useful. When we do a check like fobData.type === PluginType.HISTOGRAM the compiler now know that the fobData(which could be either HistogramFobData or ScalarFobData) is of the type HistogramFobData and will enforce that.

JamesHollyer (commented):
Now that you say it I think that would have been a reasonable split. I would have had to add some extra logic in there but, it probably would have been worth it.  I will try harder to find ways to split these things in the future.

tensorboard/webapp/widgets/linked_time_fob/types.ts
bmd3k: I think what I'm saying is that you have not generalized it sufficiently.  You have this:

```
  public getAxisDirection() {
    return this.fobCardData.type === PluginType.SCALARS
      ? AxisDirection.HORIZONTAL
      : AxisDirection.VERTICAL;
  }
```

To be truly "generalized" I would expect AxisDirection to be a field in FobCardData and the value of AxisDirection would not depend on PluginType so explicitly.

tensorboard/webapp/widgets/linked_time_fob/types.ts
bmd3k: I understand that you are using the field to help the compiler narrow the types. You use it to narrow FobCardData to one of ScalarFobData or HistogramFobData. It's a nice feature of the TypeScript compiler and we use it elsewhere in the code base.

I think, though, this is a misuse of type narrowing.

The end result in your controller code is that every function effectively looks like this:

```
someFunction() {
  if (this.fobCardData.type === PluginType.SCALARS) {
    // Do one thing for SCALARS.
    return;
  }

  // Do something completely different for HISTOGRAMS.
  return;
}
```

This is a anti-pattern. I think we can do better. My ideal is that SCALARS and HISTOGRAMS actually end up being similar enough that we don't need to treat then so differently. But if we do need to treat them so differently then there are better ways to do it, which we can discuss.

tensorboard/webapp/widgets/linked_time_fob/types.ts
bmd3k: > We could use a union of all the steps for each line.

I  mentioned offline that I think this particular idea is intriguing and possibly a nice compromise between behavior and complexity. I'd like us to try it out.

bmd3k (commented):
So to summarize our offline discussion: We are going to break this up into smaller PRs and address these changes in smaller parts. The first one is [5595](https://github.com/tensorflow/tensorboard/pull/5595).

JamesHollyer:

PR #5594, opened 2022-03-04
linked time: moves thumb to closest step on single selection when they're "close enough"

japie1235813:
Previously on single selection we only move the thumb to steps when it is exactly matched with selected step. Now we want to move the thumb (to show the corresponding image) when the selected step is "close enough".

We define the "close enough" to be 10% of the distance between two step values. For example we have steps [20, 50, 200, 400]. We move the thumb to 50 when its range is within 47-65. Calculated by 10% of (50-20) and 10% of (200-50).

Screen Shot 2022-03-03 at 4 42 02 PM

screenshot shows the step is 91800 and the thumb moved to the step when selected time is 918100.
Screen Shot 2022-03-03 at 3 57 12 PM

Note: this pr also includes some refactor of tests hierarchy, most of the tests are not intended to be re-reviewed. Only four new tests are added into "single selection"

Test hierarchy
```
linked time
    |-ticks
    |-render sliders
    |-thumb movement
          |-single selection
          |-range selection
```

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
@@ -291,21 +291,22 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
         // When there is no selected steps. We check if start step is
         // "close" enough to a step value and move it.
         if (selectedSteps.length === 0) {
+          if (stepValues.length === 1) return stepIndex;
+
           const step = selectedTime.startStep;
           for (let i = 0; i < stepValues.length - 2; i++) {
             const currentSelectedStep = stepValues[i];
             const nextSelectedStep = stepValues[i + 1];
             const distance =
               (nextSelectedStep - currentSelectedStep) * DISTANCE_RATIO;
+            if (step < currentSelectedStep) return stepIndex;
 
-            if (Math.abs(step - currentSelectedStep) <= distance) {
+            if (step - currentSelectedStep <= distance) {
               return stepValues.indexOf(currentSelectedStep);
             }
-            if (Math.abs(step - nextSelectedStep) <= distance) {
+            if (nextSelectedStep - step <= distance) {
               return stepValues.indexOf(nextSelectedStep);
             }
-
-            if (step < currentSelectedStep) return stepIndex;
           }
         }
 
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
@@ -57,6 +57,8 @@ import {CardRenderer} from '../metrics_view_types';
 import {getTagDisplayName} from '../utils';
 import {maybeClipSelectedTime, ViewSelectedTime} from './utils';
 
+const DISTANCE_RATIO = 0.1;
+
 type ImageCardMetadata = CardMetadata & {
   plugin: PluginType.IMAGES;
   sample: number;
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -57,6 +57,8 @@ import {CardRenderer} from '../metrics_view_types';
 import {getTagDisplayName} from '../utils';
 import {maybeClipSelectedTime, ViewSelectedTime} from './utils';
 
+const DISTANCE_RATIO = 0.1;
```
bmd3k, line 60: I'm fine with this choice and I like that the code is written in a way that lets us play with ratio. However, I'm curious: Why not something closer to 0.5?

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -809,187 +809,275 @@ describe('image card', () => {
...
+          let slider = fixture.debugElement.query(By.css('mat-slider'));
+          expect(slider.nativeElement.getAttribute('aria-valuenow')).toBe('2');
+        });
 
-      slider = fixture.debugElement.query(By.css('mat-slider'));
-      expect(slider.nativeElement.getAttribute('aria-valuenow')).toBe('2');
-    });
+        it('moves slider thumb to smaller closest step when they are close enough', () => {
+          store.overrideSelector(selectors.getMetricsSelectedTime, {
+            start: {step: 11},
```
bmd3k, line 876: Maybe add a note somewhere explaining explicitly that 11 is close enough to 10, so we move to the corresponding step 0?

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -809,187 +809,275 @@ describe('image card', () => {
...
-        2 /* stepIndex */
-      );
-      const fixture = createImageCardContainer('card1');
-      fixture.detectChanges();
-      let slider = fixture.debugElement.query(By.css('mat-slider'));
-      expect(slider.nativeElement.getAttribute('aria-valuenow')).toBe('2');
-    });
+        it('moves slider thumb to larger closest step when they are close enough', () => {
+          store.overrideSelector(selectors.getMetricsSelectedTime, {
+            start: {step: 19},
```
bmd3k, line 901: Same, include a brief explanation here about the numbers (19 close enough to 20 so choose step 1)?

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -809,187 +809,275 @@ describe('image card', () => {
...
+          fixture.detectChanges();
+          let slider = fixture.debugElement.query(By.css('mat-slider'));
+          expect(slider.nativeElement.getAttribute('aria-valuenow')).toBe('2');
+        });
 
-      let slider = fixture.debugElement.query(By.css('mat-slider'));
-      expect(slider.nativeElement.getAttribute('aria-valuenow')).toBe('2');
+        it('does not move slider thumb to smaller closest step when it is clipped', () => {
+          store.overrideSelector(selectors.getMetricsSelectedTime, {
+            start: {step: 41},
```
bmd3k, line 951: Can you add an explanation of why this is clipped (I can't figure it out)? (And perhaps the additional information of "if this were not clipped, 41 is close enough to 40 so would otherwise choose step 3)

tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -284,8 +286,30 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
...
+          const step = selectedTime.startStep;
+          for (let i = 0; i < stepValues.length - 2; i++) {
+            const currentSelectedStep = stepValues[i];
+            const nextSelectedStep = stepValues[i + 1];
+            const distance =
+              (nextSelectedStep - currentSelectedStep) * DISTANCE_RATIO;
+            if (step < currentSelectedStep) return stepIndex;
+
+            if (step - currentSelectedStep <= distance) {
+              return stepValues.indexOf(currentSelectedStep);
```
bmd3k, line 305: Could you: `return i;`

tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -284,8 +286,30 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
...
+            }
+            if (nextSelectedStep - step <= distance) {
+              return stepValues.indexOf(nextSelectedStep);
```
bmd3k, line 308: Could you `return i + 1;`

tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -284,8 +286,30 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
...
+        if (!selectedTime || selectedTime.clipped) return stepIndex;
+
+        // When there is no selected steps. We check if start step is
+        // "close" enough to a step value and move it.
+        if (selectedSteps.length === 0) {
+          if (stepValues.length === 1) return stepIndex;
+
+          const step = selectedTime.startStep;
+          for (let i = 0; i < stepValues.length - 2; i++) {
+            const currentSelectedStep = stepValues[i];
```
bmd3k, line 298: Maybe a little clearer to rename these to `currentStepValue` and `nextStepValue`.

(It might also be worth incorporating `value` into the currently-named `step` variable -- but I'm not certain how best to incorporate that - `stepValue` still seems insufficient)

bmd3k (approved):

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
bmd3k, line 876: Oh, also, would it be worth adding a test where selected time is step: 12 and assert that no steps are selected. (Ie show that there is an actual threshold of 10%)

bmd3k (approved):

PR #5595, opened 2022-03-04
Use new types in fob controller

JamesHollyer:
The purpose of this PR is to prepare the fob controller to be used in other cards. We are doing that by encapsulating Histogram specific data into a specific interface. Histogram is the only card that uses a temporal scale so that is included. The steps array are used to skip all steps that do not have data on that histogram. The current plan for Scalar cards is to have dragging be continuous so the array of steps will not be needed.

There is no functional change in this PR.

### DIFF
--- a/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
@@ -23,20 +23,8 @@ import {
   ViewChild,
 } from '@angular/core';
 import {LinkedTime} from '../../metrics/types';
-import {ScaleLinear, ScaleTime} from '../../third_party/d3';
+import {AxisDirection, Fob, HistogramFobData} from './types';
 
-export enum AxisDirection {
-  HORIZONTAL,
-  VERTICAL,
-}
-
-export enum Fob {
-  NONE,
-  START,
-  END,
-}
-
-type TemporalScale = ScaleLinear<number, number> | ScaleTime<number, number>;
 @Component({
   selector: 'linked-time-fob-controller',
   templateUrl: 'linked_time_fob_controller_component.ng.html',
@@ -48,9 +36,8 @@ export class LinkedTimeFobControllerComponent {
   @ViewChild('startFobWrapper') readonly startFobWrapper!: ElementRef;
   @ViewChild('endFobWrapper') readonly endFobWrapper!: ElementRef;
   @Input() axisDirection!: AxisDirection;
-  @Input() steps!: number[];
   @Input() linkedTime!: LinkedTime;
-  @Input() temporalScale!: TemporalScale;
+  @Input() histogramFobData!: HistogramFobData;
   @Output() onSelectTimeChanged = new EventEmitter<LinkedTime>();
 
   private currentDraggingFob: Fob = Fob.NONE;
@@ -65,10 +52,10 @@ export class LinkedTimeFobControllerComponent {
 
   getCssTranslatePx(step: number): string {
     if (this.axisDirection === AxisDirection.VERTICAL) {
-      return `translate(0px, ${this.temporalScale(step)}px)`;
+      return `translate(0px, ${this.histogramFobData.scale(step)}px)`;
     }
 
-    return `translate(${this.temporalScale(step)}px, 0px)`;
+    return `translate(${this.histogramFobData.scale(step)}px, 0px)`;
   }
 
   startDrag(fob: Fob) {
@@ -91,7 +78,9 @@ export class LinkedTimeFobControllerComponent {
     let newLinkedTime = this.linkedTime;
     if (this.isDraggingHigher(event.clientY, event.movementY)) {
       const stepAbove =
-        this.steps[this.getStepIndexHigherThanMousePosition(event.clientY)];
+        this.histogramFobData.steps[
+          this.getStepIndexHigherThanMousePosition(event.clientY)
+        ];
       if (this.currentDraggingFob === Fob.END) {
         newLinkedTime.end!.step = stepAbove;
       } else {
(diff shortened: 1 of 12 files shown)


tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
```diff
@@ -48,9 +36,8 @@ export class LinkedTimeFobControllerComponent {
   @ViewChild('startFobWrapper') readonly startFobWrapper!: ElementRef;
   @ViewChild('endFobWrapper') readonly endFobWrapper!: ElementRef;
   @Input() axisDirection!: AxisDirection;
-  @Input() steps!: number[];
   @Input() linkedTime!: LinkedTime;
-  @Input() temporalScale!: TemporalScale;
+  @Input() histogramFobData!: HistogramFobData;
```
japie1235813, line 40: so would there another input for, say, `ScalarFobData`, or do you plan to use `FobCardData `
which is a combination for this two?

japie1235813 (approved):

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
JamesHollyer: I plan on combining them and replacing this with CardFobData.

tensorboard/webapp/widgets/linked_time_fob/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  VERTICAL,
+}
+
+export enum Fob {
+  NONE,
+  START,
+  END,
+}
+
+export interface HistogramFobData {
```
bmd3k, line 26: As I mentioned in https://github.com/tensorflow/tensorboard/pull/5589, this does not seem like the direction we want to go.  Type narrowing does not seem like the right tool for the job.

This is my understanding of the problem you are trying to solve here:

1. linked_time_fob requires some notion of `Scale`.
2. scalars and histograms have their own separate notions/implementations of Scale that are not related to each other.

If this is the case, I think an important goal is: linked_time_fob should not be aware of both notions of Scale. It should only be aware of, at most, a single notion of Scale. If there are multiple notions of Scale then they should be hidden behind an abstraction.

So, some possibilities:
1) Refactor Scalar and Histograms so that they use the same notion of Scale. Reuse this one notion in linked_time_fob. I honestly have no idea if this is reasonable or even possible. But it should be the first thing considered. Have you given this any thought?

2) If there needs to be multiple notions of Scale, then hide that behind an abstraction and make linked_time_fob only use that abstraction. An input to linked-time-fob-controller could be some interface like "ScaleHandler" and Scalars and Histograms pass in their own implementations of this.

More generally I think an important goal for linked_time_fob should be: It should not understand or refer to "Scalars" or "Histograms" nor should it understand any concept specific to Scalars or Histograms that cannot somehow be abstracted and/or generalized.

japie1235813: Some details for refactoring Scalar and Histogram use the same notion of Scale: 
This is most unlikely and it gets things pretty complicated. The scalar and histogram data type was designed (by stephanlee) differently for a reason. The runs on scalar have different set of steps while there is only one set of steps on histogram. 
In our demo data, all demo runs have the same set of steps, they all have data on step, say, 100 and non on step 101 etc. Yet in reality, users data could be more like run1&2 have data on step 100,103 while run3 have data on step 101,102,103. 
Screen Shot 2022-02-24 at 3 36 52 PM

The types (minMax or number[]) then cause the differences on the implementation (iterate the steps or set a number directly). 

I think passing abstraction function is a great idea to make fob controller more generalize. To better confirm how it looks like in code:

```
//fob_controller.ts
translateStepToPixel(step) {
  return this.scaleHandler(step);
}

// histogram/scalar.ng.html
<fob-controller
  scaleHandler="scalaHandler" 

// histogram_compoent.ts
scalaHandler(step) { 
    return this.scale.forward(...)
}

// scalar_compoent.ts
scalaHandler(step) { 
    return this.scale.(step)
}
```


yet there are some other methods, for example, `getStepHigherThanMousePosition`, also depends on if it's minMax or array of steps. Would you also say we should we should pass this handler to <fob-controller>? Note there might be quite some elements in fob-controller needed to calculate the position.

```
//fob_controller.ts
getStepHigherThanMousePosition(step) {
  return this.getStepHigherThanMousePosition(position, bound, axisOverlay);
}

//scalar.ng.html
<fob-controller
  getStepHigherThanMousePosition="getStepHigherThanMousePosition" 

// scalar_compoent.ts
getStepHigherThanMousePosition(position, bound, axisOverlay) { 
    let stepAtMouse = Math.round(
         this.scale.reverse(
         ... 
    
    return stepAtMouse;
}

```

JamesHollyer: I have put some thought into combining the scales. That was my first thought actually. I think it might be possible with some clever type script things. However, the scales are quite different and the histogram one is generated from d3. Here is [the histogram's scale definition](https://source.corp.google.com/piper///depot/google3/third_party/javascript/typings/d3_scale/index.d.ts;rcl=432080676;l=37). We only use the function attribute(which I didn't even know was a possible attribute until I saw this). Then here is the [Scalar's scale](https://source.corp.google.com/piper///depot/google3/third_party/tensorboard/webapp/widgets/line_chart_v2/lib/scale_types.ts;rcl=432080676;l=33). Where we use both the forward and reverse functions. I could theoretically see a way bind the domain and range to a function to just take in "x". Then we could maybe histogram scale's function and the forward function could be combined and the inverse and reverse functions could be combined. This is all theoretical and I have no idea how to do that or if Typescript even allows for that.

However, there is a more fundamental problem of functionality. We want the way this dragging works to be different on histogram and scalar cards. Histogram dragging jumps to steps while scalar is a smooth continuous drag. @bmd3k I understand that you don't want the fob to know it is in histogram or scalar but maybe we can have a distinction between discrete dragging and continuous dragging. As Jie-Wei said discrete dragging needs the list of steps while continuous dragging only needs the min and max step.

So if we can somehow combine these two scales then we still need the separation of discrete and continuous.

tensorboard/webapp/widgets/linked_time_fob/types.ts
```diff
@@ -0,0 +1,29 @@
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
```
japie1235813, line 1: Just a note that I intend to create types.ts here too. #5602 
For those types used only in linked_time_fob, I'm thinking about put them as "internal_types.ts"

JamesHollyer:

PR #5597, opened 2022-03-04
linked time: Show histogram with color on mouse hovering

japie1235813:
On single selection currently histogram card only show data with the matched step. We want to enable color displaying on those hovered steps.

This could not be done in css (using `:hover`) is because the color comes from d3 (with `d3ColorScale`). Thus we add/remove `.no-color` class on mouse enter/leave.

Screenshot: selected step 0, hover over step 235
Screen Shot 2022-03-03 at 8 25 16 PM

### DIFF
--- a/tensorboard/webapp/widgets/histogram/histogram_component.ts
+++ b/tensorboard/webapp/widgets/histogram/histogram_component.ts
@@ -261,6 +261,7 @@ export class HistogramComponent implements AfterViewInit, OnChanges, OnDestroy {
   }
 
   updateColorOnHover(event: MouseEvent, isHover: boolean) {
+    // When link time is disabled all histogram should be colored.
     if (!this.isLinkedTimeEnabled(this.linkedTime)) {
       return;
     }
--- a/tensorboard/webapp/widgets/histogram/histogram_component.ts
+++ b/tensorboard/webapp/widgets/histogram/histogram_component.ts
@@ -260,6 +260,17 @@ export class HistogramComponent implements AfterViewInit, OnChanges, OnDestroy {
       : '';
   }
 
+  updateColorOnHover(event: MouseEvent, isHover: boolean) {
+    if (!this.isLinkedTimeEnabled(this.linkedTime)) {
+      return;
+    }
+    if (isHover) {
+      (event.target! as HTMLElement).classList.remove('no-color');
+    } else {
+      (event.target! as HTMLElement).classList.add('no-color');
+    }
+  }
+
   getGridTickYLocs(): number[] {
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/histogram/histogram_component.ts
```diff
@@ -260,6 +260,17 @@ export class HistogramComponent implements AfterViewInit, OnChanges, OnDestroy {
       : '';
   }
 
+  updateColorOnHover(event: MouseEvent, isHover: boolean) {
+    if (!this.isLinkedTimeEnabled(this.linkedTime)) {
```
bmd3k, line 264: Maybe comment that when link time is disabled all histogram lines are colored so there is no need to update?

tensorboard/webapp/widgets/histogram/histogram_test.ts
```diff
@@ -906,6 +906,35 @@ describe('histogram test', () => {
         fixture.detectChanges();
         expect(doHistogramsHaveColor(fixture)).toEqual([false, false, false]);
       });
+
+      it('puts color on histogram that mouse hovers over', () => {
```
bmd3k, line 910: Do you need a test to ensure that mouseleave does not disable a histograms color when linked time is disabled?

tensorboard/webapp/widgets/histogram/histogram_test.ts
```diff
@@ -906,6 +906,35 @@ describe('histogram test', () => {
...
+
+        const firstHistogram = fixture.debugElement.queryAll(
+          By.css('g.histogram')
+        )[0];
+
+        firstHistogram.triggerEventHandler('mouseenter', {
+          target: firstHistogram.nativeElement,
+        });
+        fixture.detectChanges();
+        expect(doHistogramsHaveColor(fixture)).toEqual([true, true, false]);
```
bmd3k, line 930: Maybe add comment explaining why stepIndexes 0 and 1 are highlighted?

tensorboard/webapp/widgets/histogram/histogram_test.ts
```diff
@@ -906,6 +906,35 @@ describe('histogram test', () => {
...
+
+        firstHistogram.triggerEventHandler('mouseleave', {
+          target: firstHistogram.nativeElement,
+        });
+        fixture.detectChanges();
+        expect(doHistogramsHaveColor(fixture)).toEqual([false, true, false]);
```
bmd3k, line 936: Similar, add a comment?

bmd3k (approved):

PR #5598, opened 2022-03-04
alias: add displayAlias to DropdownOption

yatbear:
This is the first step for b/222085838. The additional argument is needed due to the formatting (bold alias prefix) for each dropdown option.

### DIFF
--- a/tensorboard/webapp/widgets/dropdown/dropdown_component.ts
+++ b/tensorboard/webapp/widgets/dropdown/dropdown_component.ts
@@ -42,13 +42,8 @@ export interface DropdownOption {
         [value]="option.value"
         [disabled]="option.disabled"
       >
-        <p *ngIf="option.displayAlias; else noAlias">
-          <b>{{ option.displayAlias }}</b
-          >: {{ option.displayText }}
-        </p>
-        <ng-template #noAlias>
-          {{ option.displayText }}
-        </ng-template>
+        <b *ngIf="option.displayAlias">{{ option.displayAlias }}: </b>
+        {{ option.displayText }}
       </mat-option>
     </mat-select>
   `,
--- a/tensorboard/webapp/widgets/dropdown/dropdown_component.ts
+++ b/tensorboard/webapp/widgets/dropdown/dropdown_component.ts
@@ -21,6 +21,10 @@ export interface DropdownOption {
    * Whether the option should appear non-interactable. 'enabled' by default.
    */
   disabled?: boolean;
+  /**
+   * Optional alias for the displayText.
+   */
+  displayAlias?: string;
 }
 
 /**
@@ -38,7 +42,13 @@ export interface DropdownOption {
         [value]="option.value"
         [disabled]="option.disabled"
       >
-        {{ option.displayText }}
+        <p *ngIf="option.displayAlias; else noAlias">
+          <b>{{ option.displayAlias }}</b
+          >: {{ option.displayText }}
+        </p>
+        <ng-template #noAlias>
+          {{ option.displayText }}
+        </ng-template>
       </mat-option>
     </mat-select>
   `,
--- a/tensorboard/webapp/widgets/dropdown/dropdown_test.ts
+++ b/tensorboard/webapp/widgets/dropdown/dropdown_test.ts
@@ -34,15 +34,16 @@ import {DropdownComponent, DropdownOption} from './dropdown_component';
 class TestableComponent {
   @Input() expId!: string;
   @Input() configOptions!: DropdownOption[];
-  @Input() expChanged!: (event: {value: string}) => void;
+
+  expChanged(value: string) {
+    this.expId = value;
+  }
 }
 
 describe('tb-dropdown', () => {
-  let expChangedSpy: jasmine.Spy;
-
   const byCss = {
     DROPDOWN: By.directive(DropdownComponent),
-    OPTION: By.css('.mat-option-text'),
+    MAT_SELECT: By.css('mat-select'),
   };
 
   beforeEach(async () => {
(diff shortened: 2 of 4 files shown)


tensorboard/webapp/widgets/dropdown/dropdown_component.ts
```diff
@@ -38,7 +42,13 @@ export interface DropdownOption {
...
         [disabled]="option.disabled"
       >
-        {{ option.displayText }}
+        <p *ngIf="option.displayAlias; else noAlias">
+          <b>{{ option.displayAlias }}</b
+          >: {{ option.displayText }}
+        </p>
+        <ng-template #noAlias>
+          {{ option.displayText }}
+        </ng-template>
```
japie1235813, line 51: Could this be more simplified?
```
<b *ngIf="option.displayAlias">{{ option.displayAlias }}</b> {{...displayText}}
```

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
```diff
@@ -0,0 +1,124 @@
...
+      ]);
+    });
+
+    it(' with valid aliases', () => {
+      let fixture = createTestableComponent({
+        expId: '2',
+        configOptions: [
+          {
+            value: '2',
+            displayText: 'loooonnnnngggggg',
```
japie1235813, line 105: It does not affect the test but, I think it's better to have the mockup text to be something more readable. For example, 'test experiment name'

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
```diff
@@ -0,0 +1,124 @@
...
+            disabled: false,
+            displayAlias: 'long',
```
japie1235813, line 107: ditto. Something like "test alias"

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
```diff
@@ -0,0 +1,124 @@
...
+      const dropdown = fixture.debugElement.query(byCss.DROPDOWN);
+      expect(dropdown.componentInstance.value).toEqual('1');
+      expect(dropdown.componentInstance.options).toEqual([
+        jasmine.objectContaining({
+          value: '',
+          displayText: 'none',
+          disabled: true,
+        }),
+        jasmine.objectContaining({value: '1', displayText: 'abc'}),
+      ]);
```
japie1235813, line 96: Could you actually test the dom element? Isn't that why you create `byCss.OPTION` for? Now the selector is not used.

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  describe('renders options', () => {
+    it(' with no aliases', () => {
+      let fixture = createTestableComponent({
+        expId: '1',
+        configOptions: [
+          {
+            value: '',
+            displayText: 'none',
```
japie1235813, line 74: what it would look like on displayText to be "none"?  Do we still show it with literally "none"?

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
```diff
@@ -0,0 +1,124 @@
...
+  }): ComponentFixture<TestableComponent> {
+    const fixture = TestBed.createComponent(TestableComponent);
+    fixture.componentInstance.expId = input.expId || '';
+    fixture.componentInstance.configOptions = input.configOptions || [];
+    expChangedSpy = jasmine.createSpy();
+    fixture.componentInstance.expChanged = expChangedSpy;
+    return fixture;
+  }
+
+  describe('renders options', () => {
```
japie1235813, line 67: Thanks for adding the test. One thing I found missing from the test is `expChanged` output is not tested. Could you add more tests on this? (and that's why you created `expChangedSpy`, isn't it?)

tensorboard/webapp/widgets/dropdown/dropdown_component.ts
yatbear: Done, thanks for the suggestions!

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
yatbear: Done.

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
yatbear: Added TODOs.

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
yatbear: Yes, currently in our config dropdown, the first option is "none".

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
yatbear: Added a test for option change.

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
```diff
@@ -0,0 +1,150 @@
...
+      [options]="configOptions"
+      (selectionChange)="expChanged($event)"
+    ></tb-dropdown>
+  `,
+})
+class TestableComponent {
+  @Input() expId!: string;
+  @Input() configOptions!: DropdownOption[];
+
+  expChanged(value: string) {
```
japie1235813, line 38: I saw the implementation is used in link 148 but what we want to do is have a spy and check if the spy is called with desired param. To be more specific, you can replace this with `@Input() selectionChange!: (value: string) => void`

later in line 62. you create a spy and assign to this input

```
selectionChangeSpy = jasmine.createSpy();
...
fixture.componentInstance.selectionChange = selectionChangeSpy;
```

then in tests we do 
```
expect(selectionChangeSpy).toHaveBeenCalledWith('2')
```

you can check the example `onSingleValueChanged` here
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/widgets/range_input/range_input_test.ts;l=66?q=range_input&ss=tensorflow%2Ftensorboard

japie1235813: I saw the implementation is used in link 148 but what we want to do is have a spy and check if the spy is called with desired param. To be more specific, you can replace this with `@Input() selectionChange!: (value: string) => void`

later in line 62. you create a spy and assign to this input

```
selectionChangeSpy = jasmine.createSpy();
...
fixture.componentInstance.selectionChange = selectionChangeSpy;
```

then in tests we do 
```
expect(selectionChangeSpy).toHaveBeenCalledWith('2')
```

you can check the example `onSingleValueChanged` here
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/widgets/range_input/range_input_test.ts;l=66?q=range_input&ss=tensorflow%2Ftensorboard

yatbear: Changed, thanks for the example!

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
```diff
@@ -0,0 +1,148 @@
...
+      [value]="expId"
+      [options]="configOptions"
+      (selectionChange)="expChanged($event)"
+    ></tb-dropdown>
+  `,
+})
+class TestableComponent {
+  @Input() expId!: string;
+  @Input() configOptions!: DropdownOption[];
+  @Input() expChanged!: (event: {value: string}) => void;
```
japie1235813, line 37: Also it's not obvious is test cases but I would also suggest rename `selectionChange` or `onSelectionChange` to be meaningful.

japie1235813 (approved):

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
japie1235813: Sorry this comment is cached and sent again. Please ignore it

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
yatbear: Done. Thanks for the readability suggestion!

tensorboard/webapp/widgets/dropdown/dropdown_test.ts
```diff
@@ -0,0 +1,148 @@
...
+      ],
+    });
+    fixture.detectChanges();
+
+    // Triggers selection change event.
+    const matSelect = fixture.debugElement.query(byCss.MAT_SELECT);
+    matSelect.triggerEventHandler('selectionChange', {value: '2'});
+    fixture.detectChanges();
+
+    expect(selectionChangeSpy).toHaveBeenCalledOnceWith('2');
```
japie1235813, line 146: I think this should be called with `{value: '2;}` instead of '2'. Please check again.

yatbear: After checking the call arg appeared to be `'2'`, similar to the example: https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/widgets/range_input/range_input_test.ts;l=228?q=range_input&ss=tensorflow%2Ftensorboard

PR #5605, opened 2022-03-10
Create FobCardAdapter interface and use it for Histograms

JamesHollyer:
This change adds an interface that will be used to encapsulate card specific logic. This interface is implemented in the HistogramFobAdapter and used in the Fob Controller.

There is no functional change in this PR.

This change is to prepare for adding fob dragging functionality to Scalar cards which is bring done [here](https://github.com/tensorflow/tensorboard/pull/5607).

### DIFF
--- a/tensorboard/webapp/widgets/histogram/histogram_fob_adapter_test.ts
+++ b/tensorboard/webapp/widgets/histogram/histogram_fob_adapter_test.ts
@@ -1,0 +1,143 @@
+import {Component} from '@angular/core';
+import {ComponentFixture, TestBed} from '@angular/core/testing';
+import {ScaleLinear} from '../../third_party/d3';
+import {HistogramFobAdapter} from './histogram_fob_adapter';
+
+@Component({
+  selector: 'axis-overlay',
+  template: ` <div></div> `,
+})
+class AxisOverlayComponent {}
+
+describe('HistogramFobAdapter', () => {
+  let temporalScaleSpy: jasmine.Spy;
+  beforeEach;
+  function createInterface(customSteps?: number[]) {
+    temporalScaleSpy = jasmine.createSpy();
+    let scale = temporalScaleSpy as unknown as ScaleLinear<number, number>;
+    let steps = customSteps || [0, 1, 2, 3];
+    temporalScaleSpy.and.callFake((step: number) => {
+      return step;
+    });
+    return new HistogramFobAdapter(scale, steps);
+  }
+  describe('setBounds', () => {
+    it('sets bounds properly when given no overrides', () => {
+      let fobInterface = createInterface([100, 200, 5000]);
+      fobInterface.setBounds({});
+      expect(fobInterface.lowerBound).toEqual(100);
+      expect(fobInterface.upperBound).toEqual(5000);
+    });
+    it('sets bounds properly when given lower bound override', () => {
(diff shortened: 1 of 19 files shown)


tensorboard/webapp/widgets/linked_time_fob/types.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export enum Fob {
+  NONE,
+  START,
+  END,
+}
+
+export interface FobCardAdapter {
+  upperBound: number;
```
bmd3k, line 27: This (upperBound and lowerBound) is internal state that you expect each implementation to have. It suggests that the state should be moved to LinkedTimeFobControllerComponent. So, in this particular case, upperBound and lowerBound would be moved back to LinkedTimeFobControllerComponent. (Aside: upperBound and lowerBound are pretty ambiguous names in this case - I think in LinkedTimeFobControllerComponent these should be named something like upperDraggableBound and lowerDraggableBound?)

What might this mean for the FobCardAdapter API? A possibility:  Instead of setBound() you could have `getCardLowerBound(): number` and `getCardUpperBound(): number`. For other API calls you could pass an argument draggableBounds: [number, number].

tensorboard/webapp/widgets/linked_time_fob/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  lowerBound: number;
+  // Set the upper and lower bounds. Implementation can determine lower and
+  // upper bounds as it needs to. However, if the overrides are set those must
+  // be used.
+  setBounds(overrides: {lowerOverride?: number; higherOverride?: number}): void;
+  // Uses whatever underlying scale is need to translate the proper pixel offset
+  stepToPixel(step: number, domain: [number, number]): number;
```
bmd3k, line 34: Can you give `domain` a better name and/or explain what it is in the jsdoc for the function?

tensorboard/webapp/widgets/histogram/histogram_fob_adapter_test.ts
```diff
@@ -0,0 +1,158 @@
...
+  function createInterface(customSteps?: number[]) {
+    temporalScaleSpy = jasmine.createSpy();
+    let scale = temporalScaleSpy as unknown as ScaleLinear<number, number>;
+    let steps = customSteps || [0, 1, 2, 3];
+    temporalScaleSpy.and.callFake((step: number) => {
+      return step;
+    });
+    return new HistogramFobAdapter(scale, steps);
+  }
+  describe('setBounds', () => {
```
bmd3k, line 39: These types of tests are ok -- in a lot of cases testing the interface is best practice. But I think it is important to also test the combination of LinkedTimeFobControllerComponent + HistogramFobAdapter since there are some important interactions there.  So you could either:

1) Supplement these tests with webtests (If you go this route I'd suggest trying to write those webtests side x side with this PR so you can get a sense if it will actually work)
2) Supplement these tests with unit tests that combine LinkedTimeFobControllerComponent + HistogramFobAdapter.

I suspect option (2) is easier for histograms only because these tests probably already exist in some form.

If you do option (2) you might find it a good idea to create a first-class histogram_fob_controller_component that does all the wiring for you.

JamesHollyer:

PR #5606, opened 2022-03-10
alias: test mat-option DOM using Angular test harnesses

yatbear:
The list of `mat-option` elements are rendered in a separate overlay container outside the DOM instead of as a descendent of `mat-select`, therefore we cannot directly query by css to check the text contents. Fortunately we can use the Angular component test harnesses ([MatSelectHarness](https://material.angular.io/components/select/api#MatSelectHarness)) to properly interact with `mat-select` and `mat-option` in the test environment.

Tested internally: cl/433235493

### DIFF
--- a/tensorboard/webapp/angular/BUILD
+++ b/tensorboard/webapp/angular/BUILD
@@ -211,6 +211,15 @@ tf_ts_library(
     ],
 )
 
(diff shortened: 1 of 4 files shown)


japie1235813 (approved):
Thanks for finding the solution of testing mat-option in dom!

PR #5612, opened 2022-03-15
linked time: adds clipped text hint on image card

japie1235813:
As the escalation mark shown in histogram and scalar card when linked time is clipped, we want to add the same for image card by reusing module `vis_selected_time_warning` The logic follows what already used by histogram/scalar card. 

screenshot
Screen Shot 2022-03-14 at 8 21 45 PM

Screen Shot 2022-03-14 at 8 23 11 PM

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
@@ -21,9 +21,9 @@ limitations under the License.
       title="{{ tag }}"
       value="{{ title }}"
     ></tb-truncated-path>
-    <vis-selected-time-clipped
+    <vis-selected-time-warning
       *ngIf="selectedTime && selectedTime.clipped"
-    ></vis-selected-time-clipped>
+    ></vis-selected-time-warning>
     <span class="controls">
       <button
         mat-icon-button
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
@@ -21,9 +21,9 @@ limitations under the License.
       title="{{ tag }}"
       value="{{ title }}"
     ></tb-truncated-path>
-    <vis-selected-time-warning
+    <vis-selected-time-clipped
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
```diff
@@ -21,6 +21,9 @@
       title="{{ tag }}"
       value="{{ title }}"
     ></tb-truncated-path>
+    <vis-selected-time-warning
```
bmd3k, line 24: I'm a little confused: Isn't this still called vis-selected-time-clipped?

http://google3/third_party/tensorboard/webapp/metrics/views/card_renderer/vis_selected_time_warning_component.ts;l=21;rcl=435097075

tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
```diff
@@ -21,6 +21,9 @@
...
+      *ngIf="selectedTime && selectedTime.clipped"
```
bmd3k, line 25: Would it be worth setting "isClipped" property?

https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/views/card_renderer/vis_selected_time_warning_component.ts;l=40;drc=ea35c6034141ae4c0fbdd23716539a57d5447dc2

Your TODO makes it seem like you don't want the default to be true.

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -1141,5 +1143,111 @@ describe('image card', () => {
         });
       });
     });
+
+    describe('selectedTime beyond range of data', () => {
+      it('clips the selectedTime to max step', () => {
```
bmd3k, line 1148: Also a little confused here: Are the first 2 of these 3 tests just testing existing functionality? I think only the 3rd one is a test for what is being added in this PR?

bmd3k (commented):
It seems fine but there are a couple things that confuse me - maybe the PR is a bit stale?

bmd3k (approved):

PR #5615, opened 2022-03-16
linked time: move stepIndex change to reducer

japie1235813:
Previously for linked time in image card we move slider thumb on different conditions (single/range and where the thumb position is.. etc). #5569 #5587 #5594 
Yet by moving thumb in view instead of setting the stepIndex in ngrx step cause bugs like "flashing", which you can see the thumb jumps back and forth pretty fast. This is because when no "moving conditions" are satisfied, we do not move the thumb, at this time if stepIndex state is null, it was set to maximum step.
For example, think about a case where the stepIndex is null, and steps are [10, 20, 30, 40, 50], while the selected step slowly increase from [18,19,20,21,22,23], the index of thumb would be [4,1,1,1,4,4]. This causes pretty bad user experiences.

To handle this bug, we moved the set stepIndex logic to metrics/reducer. In the above example, the thumb stays at index 1 once moved. Then when selected step is increased to 22, the thumb stays at index 1 instead of jumping back to index 4.

This pr was large due to it moves the result merged from 3 PRs. Most of the logics are not intended to be re-reviewed. I suggest the review flow below:

0. run it and play around. see how thumb no longer "flash". Goolgers, please see [demo server](https://20220316-233622z-jieweiwu-dot-tensorboard-corp-dev.googleplex.com)
1. metrics_reducer.ts:  `actions.timeSelectionChanged` method
2. metrucs_store_internal_utils.ts: `generateNexCardStepIndexFromSelectTime` method
This method contains all logics moved from `image_card_container.ts` with further breakdown subsets into multiple funcitons
3. Subfunctions in  `generateNexCardStepIndexFromSelectTime`
4. tests:
I have clear commit for some of the subfunctions tests
#getSelectedSteps & #getImageCardStepValues: unfortunately they're merged into the first [commit](https://github.com/tensorflow/tensorboard/commit/4e05202f15644db9554987e295b4019e13ce3bdd)
#updateNextCardStepIndexOnSingleSelection: [commit](https://github.com/tensorflow/tensorboard/commit/8d75d37a5e554fd870e034c9f35b71058ee8c548)
#updateNextCardStepIndexOnRangeSelection: [commit](https://github.com/tensorflow/tensorboard/commit/9782feeea9e7204890ef3d08f4e3a985e1d38613)
#generateNexCardStepIndexFromSelectTime: this is relatively trivial. Most functionalities have been covered in above test cases. The only left is to iterate the card ids. It would be better tested with ability to mock customized global function. [commit](https://github.com/tensorflow/tensorboard/commit/041bd8ec3e650c4117d185e9401ca11c1e85e9f1)

Next step: tests for reducers. To reduce the size of this PR. The tests are addressed later.


japie1235813:
Close this PR as all break-down changes have been merged.

PR #5616, opened 2022-03-17
linked time: dummy method to prepare stepIndex moving

japie1235813:
This is the first step to move setting stepIndex in image card from view (image container) to ngrx state (reducer). Full context please see #5615

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
+++ b/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
@@ -378,10 +378,6 @@ export function generateNextCardStepIndexFromSelectTime(
 ): CardStepIndexMap {
   const nextCardStepIndex = {...previousCardStepIndex};
 
-  Object.keys(previousCardStepIndex).forEach((cardId) => {
-    if (!cardId.includes('"plugin":"images"')) return;
-  });
-
   return nextCardStepIndex;
 }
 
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -370,5 +370,19 @@ export function canCreateNewPins(state: MetricsState) {
...
 }
+/**
+ * Sets cardStepIndex for image card based on selected time.
+ */
+export function generateNextCardStepIndexFromSelectTime(
+  previousCardStepIndex: CardStepIndexMap
+): CardStepIndexMap {
+  const nextCardStepIndex = {...previousCardStepIndex};
+
+  Object.keys(previousCardStepIndex).forEach((cardId) => {
```
bmd3k, line 381: I think you have been a bit too aggressive in breaking this up into smaller CLs - would be nice to have at least something worth testing :). I think this does nothing. If I'm right, please delete it. If I'm wrong, please correct me and add a comment.

Having said that, if this logic does nothing and you remove it then I approve of this PR.

bmd3k (approved):

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
japie1235813: I added this to make the method meaning. As stated in the jsdoc, it's meant to be image card only. 
Let me know if you still prefer to remove this (which will leave the jsdoc a bit unexplained, but if you're okay with this I'm okay too)

PR #5617, opened 2022-03-17
[WIP]Implement and use new FobCardAdapter interface

JamesHollyer:


JamesHollyer:

PR #5620, opened 2022-03-18
[WIP]Implement and Use new FobCardAdapter Interface in HistogramComponent

JamesHollyer:
Similar to PR #5617 this is changes the dragging fob logic to be implemented using a new interface that can be used by implemented by multiple cards. Currently it is only used in histogram. This PR has no functional change.

In contrast to PR #5617 this uses the HistogramComponent class itself as the FobCardAdapter instead of a separate class.


JamesHollyer:

PR #5624, opened 2022-03-21
linked time: stepIndex moving: add function to get selected step(s)

japie1235813:
Creates a function to get selected step(s) within selected time. This prepares for stepIndex moving. Motivation and full context please see https://github.com/tensorflow/tensorboard/pull/5615.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
+++ b/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
@@ -406,22 +406,24 @@ export function getImageCardStepValues(
 }
 
 /**
- * Returns selected steps that are within selected time given a list of step
+ * Returns the subset of steps that are within selected time given a list of steps
  */
 function getSelectedSteps(
   selectedTime: LinkedTime | null,
-  stepValues: number[]
+  steps: number[]
 ) {
   if (!selectedTime) return [];
 
+  // Single selection: returns start step if matching any step in the list, otherwise returns nothing.
   if (selectedTime.end === null) {
-    if (stepValues.indexOf(selectedTime.start.step) !== -1)
+    if (steps.indexOf(selectedTime.start.step) !== -1)
       return [selectedTime.start.step];
     return [];
   }
 
+  // Range selection.
   const selectedStepsInRange = [];
-  for (const step of stepValues) {
+  for (const step of steps) {
     if (step >= selectedTime.start.step && step <= selectedTime.end.step) {
       selectedStepsInRange.push(step);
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -404,4 +405,28 @@ export function getImageCardStepValues(
   return loadable.runToSeries[runId].map((stepDatum) => stepDatum.step);
 }
 
-export const TEST_ONLY = {getImageCardStepValues, util};
+/**
+ * Returns selected steps that are within selected time given a list of step
+ */
+function getSelectedSteps(
+  selectedTime: LinkedTime | null,
+  stepValues: number[]
```
bmd3k, line 413: Sorry, this request is a bit belated: But consider not using the term "stepValues" since the "values" doesn't really mean anything  to me - they are just "steps" and there are no additional "values", right?

Same advice for getImageCardStepValues. Maybe just call it getImageCardSteps.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -742,4 +742,41 @@ describe('metrics store utils', () => {
...
+      const stepValues = [10];
+
+      expect(getSelectedSteps(selectedTime, stepValues)).toEqual([10]);
+    });
+
+    it(`gets selected steps on range selection`, () => {
+      const selectedTime = {start: {step: 10}, end: {step: 40}};
+      const stepValues = [5, 10, 20, 40];
+
+      expect(getSelectedSteps(selectedTime, stepValues)).toEqual([10, 20, 40]);
```
bmd3k, line 758: This test gives me confidence that we only select steps above the bottom of the selectedTime but NOT that we only select steps below the top of the selectedTime. Maybe add one more case?

bmd3k (approved):

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
japie1235813: Change this case to steps=[5..50], time=[10,40]
and added another one for steps=[5..40], time=[10,50]

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
japie1235813: Sure! let me make rename `getImageCardSteps` a separate PR

PR #5625, opened 2022-03-21
linked time: reuse getImageCardStepValues from store unit in selector

japie1235813:
In #5619 we have created a method `getImageCardStepValues` in store to return all step values in an image card. This duplicates what had been used in `image_card_container.ts`. To reduce the duplicated logic, we create a selector and reuse this method.

Note: we should also check if cardId is valid in `getImageCardStepValues`. That will come in following PR with appropriate tests.


tensorboard/webapp/metrics/store/metrics_selectors.ts
```diff
@@ -189,6 +189,23 @@ export const getCardStepIndex = createSelector(
   }
 );
 
+/**
+ * Returns step values of an image card.
+ */
+export const getMetricsImageCardStepValues = createSelector(
+  selectMetricsState,
+  (state: MetricsState, cardId: CardId): number[] => {
+    if (!state.cardMetadataMap.hasOwnProperty(cardId)) {
```
JamesHollyer, line 198: Isn't this already checked in getImageCardStepValue. Why duplicate?

tensorboard/webapp/metrics/store/metrics_selectors_test.ts
```diff
@@ -402,6 +402,168 @@ describe('metrics selectors', () => {
     });
   });
 
+  describe('getMetricsImageCardStepValues', () => {
```
JamesHollyer, line 405: All of these tests seems to just be duplicates of the tests on getImageCardStepValues. Why duplicate it? Maybe just test that it calls that function or something? Would be annoying to have to change these tests in 2 places if we want to implement some change in the logic.

JamesHollyer (approved):
LGTM

PR #5626, opened 2022-03-22
Clear tag and chart metrics data on navigated action.

bmd3k:

For Google-internal TensorBoards with more complex navigation, users are running into bugs where loading new experiments is not loading any chart data or loading new comparisons is not loading new chart data. (see b/225162725).

The root cause is stale tag and chart metadata that is being used to render the charts before new tag metadata is retrieved. We replace this data with the new tag metadata, anyway, so the change is to clear this data earlier, on navigated event whenever the sets of experiments in a route changes.

### DIFF
--- a/tensorboard/webapp/metrics/effects/index.ts
+++ b/tensorboard/webapp/metrics/effects/index.ts
@@ -224,8 +224,8 @@ export class MetricsEffects implements OnInitEffects {
 
   private readonly visibleCardsWithoutDataChanged$ = this.actions$.pipe(
     ofType(actions.cardVisibilityChanged),
-    switchMap(() => this.getVisibleCardFetchInfos().pipe(take(1))),
-    map((fetchInfos) => {
+    withLatestFrom(this.getVisibleCardFetchInfos()),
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/metrics/effects/index.ts
```diff
@@ -224,17 +224,17 @@ export class MetricsEffects implements OnInitEffects {
 
   private readonly visibleCardsWithoutDataChanged$ = this.actions$.pipe(
     ofType(actions.cardVisibilityChanged),
-    switchMap(() => this.getVisibleCardFetchInfos().pipe(take(1))),
-    map((fetchInfos) => {
+    withLatestFrom(this.getVisibleCardFetchInfos()),
```
bmd3k, line 227: Also switching to use more idiomatic rxjs - getVisibleCardFetchInfos is derived from ngrx state so it will always have a value ready. Using withLatestFrom instead of take(1) shouldn't change behavior.

This is a side effect of my investigation into the problem. I could move it into a different PR if you'd like.

japie1235813 (approved):

PR #5627, opened 2022-03-22
time namespaced: add pinned card back when experiments are added back

japie1235813:
In comparison view, when an experiment is removed and added back, if there are pinned cards from these re-added experiments, we lost name even within the same namespace. We want to bring these pinned cards back when the associated experiments are added back.
This PR introduce a new state, `cardToPinnedCopyCache`, to remember all card ids that have ever been pinned and not unpinned. That is to say, we update `cardToPinnedCopy` state to know what are the pinned cards that we want to show under the current card lists. But `cardToPinnedCopyCache` is only updated on pinned/unpinned actions along with `cardToPinnedCopy`, we do not update it on `metricsTagMetadataLoaded`.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
+++ b/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
@@ -300,14 +300,13 @@ export function buildOrReturnStateWithPinnedCopy(
  * state after an update to the set of experiments/tags/cards. Pinned cards are
  * removed if their corresponding card is no longer present in the state (perhaps
  * because the corresponding experiment and tag have been removed from the
- * comparison). Pinned cards are added if they are pinned before. It generates new
- * cardToPinnedCopy, pinnedCardToOriginal maps to reflect the new set of pinned
- * cards. It generates a cardMetadataMap object that is a combination of the input
- * nextCardMetadataMap as well as metadata for the new set of pinned cards.
- * @param previousCardToPinnedCopyCache The set of pinned cards that were present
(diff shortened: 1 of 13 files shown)


tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -296,35 +296,34 @@ export function buildOrReturnStateWithPinnedCopy(
 }
 
 /**
- * preserving the previous map, which might also contain outdated mapping due to
  * Determines which pinned cards should continue to be present in the metrics
```
bmd3k, line 299: After submission, if it is feasible, do you mind adding webtests for this functionality. Also perhaps for the use case where the user unpins a card and returns to the experiment later?

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -296,35 +296,34 @@ export function buildOrReturnStateWithPinnedCopy(
...
  * state after an update to the set of experiments/tags/cards. Pinned cards are
  * removed if their corresponding card is no longer present in the state (perhaps
  * because the corresponding experiment and tag have been removed from the
- * comparison). No new pinned cards are added by this function. It generates new
+ * comparison). Pinned cards are added if they are pinned before. It generates new
```
bmd3k, line 303: "pinned before" (and "present before") => just to be a little clearer about the use case, perhaps word as "previously pinned by the user"

bmd3k (approved):
Thanks! I was wondering about this just the other day.

PR #5629, opened 2022-03-22
linked time: stepIndex moving: moving thumb movement of single selection to stepIndex update in reducer

japie1235813:
Motivation and full context please see https://github.com/tensorflow/tensorboard/pull/5615.

This PR updates stepIndex on single selection:
* Moves the logic of thumb position to `getNextCardStepIndexOnSingleSelection` in store_internal_util
* The tests for `generateNextCardStepIndexFromSelectedTime` look similar to `getNextCardStepIndexOnSingleSelection` but it includes larger usage of `getImageCardStepValues`
* Removes corresponding tests in `image_card_test` cause the thumb movements are no longer updated in the view

Note: this PR does not have the change to rename `getImageCardStepValues` to `getImageCardSteps`


tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -428,4 +451,50 @@ function getSelectedSteps(selectedTime: LinkedTime | null, steps: number[]) {
...
+/**
+ * Gets cardStepIndex updated based on single selection.
+ */
+function getNextCardStepIndexOnSingleSelection(
+  cardId: string,
+  startStep: number,
+  steps: number[],
+  previousCardStepIndex: CardStepIndexMap
+) {
+  const nextCardStepIndex = {...previousCardStepIndex};
```
bmd3k, line 463: I find it odd that you're regenerating nextCardStepIndex over and over - it doesn't seem like it should be the responsibility of this function - is it necessary?  I think the only thing that should be generating it is generateNextCardStepIndexFromSelectedTime. Consider having this function return the index value and have it updated by generateNextCardStepIndexFromSelectedTime.

In that case you would also have two choices: When you can't calculate a new card index, do you return the old one or do you return null? If null, the call to this function could look something like:

```
    if (selectedTime.end === null) {
      const cardStep = getImageCardStepFromSingleSelection(
        selectedTime.start.step,
        steps
      );
      if (nextCardStep !== null) {
          nextCardStepIndex[cardId] = nextCardStep;
      }
    }
```

If this all makes sense then I'd also recommend changing the name. Add 'Image', remove 'Index', use same verb 'From' as the parent function. 

Something like: getImageCardStepFromSingleSelection.

japie1235813: Thanks for the comment.  Updated to `getImageCardStepIndexFromSingleSelection`
Still prefer reserving the 'index' part. Without index it sound like stepValues (which all renamed to steps).

What do you think?

japie1235813: Thanks for the comment. Updated the method to return step index.
Still prefer keep 'index' otherwise it's confusing with stepValues, which all renames to steps.
steps could be [193,000, 283,000, 989,000] while the step index is [0, 1, 2]

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -646,6 +647,173 @@ describe('metrics store utils', () => {
...
+    it(`does not update cardStepIndex on selected step with no image`, () => {
+      const selectedTime = {start: {step: 15}, end: null};
+      const nextCardStepIndex = generateNextCardStepIndexFromSelectedTime(
+        previousCardStepIndex,
+        cardMetadataMap,
+        timeSeriesData,
+        selectedTime
+      );
+
+      expect(nextCardStepIndex).toEqual({[imageCardId]: null});
```
bmd3k, line 751: I'd find this test much more reassuring if the original value for imageCardId were anything other than null. Otherwise, I might believe that you are taking the null value from getNextImageCardStepIndexFromSingleSelection and using it.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -646,6 +647,173 @@ describe('metrics store utils', () => {
...
+    });
+    it('updates cardStepIndex to smaller closest stepIndexIndex when they are close enough', () => {
```
bmd3k, line 753: I think this test is redundant - doesn't really add any existing coverage.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -646,6 +647,173 @@ describe('metrics store utils', () => {
...
+        previousCardStepIndex,
+        cardMetadataMap,
+        timeSeriesData,
+        selectedTime
+      );
+
+      expect(nextCardStepIndex).toEqual({[imageCardId]: 0});
+    });
+
+    it('dose not update cardStepIndex when selected step is not close to any step values', () => {
```
bmd3k, line 765: Also redundant.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -646,6 +647,173 @@ describe('metrics store utils', () => {
...
+        previousCardStepIndex,
+        cardMetadataMap,
+        timeSeriesData,
+        selectedTime
+      );
+
+      expect(nextCardStepIndex).toEqual({[imageCardId]: null});
+    });
+
+    it('updates cardStepIndex to larger closest stepIndex when they are close enough', () => {
```
bmd3k, line 777: Also redundant.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -646,6 +647,173 @@ describe('metrics store utils', () => {
...
+        previousCardStepIndex,
+        cardMetadataMap,
+        timeSeriesData,
+        selectedTime
+      );
+
+      expect(nextCardStepIndex).toEqual({[imageCardId]: 1});
+    });
+
+    it('dose not update cardStepIndex when there is only one unmatched step', () => {
```
bmd3k, line 789: Also redundant.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -646,6 +647,173 @@ describe('metrics store utils', () => {
...
+    const previousCardStepIndex = {[imageCardId]: null};
+    const cardMetadataMap = {
+      [imageCardId]: {
+        runId: 'test run Id',
+        plugin: PluginType.IMAGES,
+        tag: 'tagC',
+        sample: 111,
+      },
+    };
+    let timeSeriesData: TimeSeriesData = {
```
bmd3k, line 661: Could you use buildTimeSeriesData() from webapp/metrcis/testing.ts?  It might also be worth adding an `overrides` argument to that function so you can also use it in other places in these tests.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -646,6 +647,173 @@ describe('metrics store utils', () => {
     });
   });
 
+  describe('generateNextCardStepIndexFromSelectedTime', () => {
+    const imageCardId = 'test imagee card id "plugin":"images"';
```
bmd3k, line 651: It's just an id but I imagine you meant image instead of imagee.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -951,71 +946,63 @@ describe('metrics store utils', () => {
...
 
       expect(nextStepIndex).toEqual(1);
     });
 
-    it('dose not update nextStepIndex when there is only one unmatched step', () => {
-      const nextStartStep = 15;
-      const unmatchedSteps = [10];
+    it('does not return step Index when there is only one unmatched step', () => {
+      const nextStepIndex = getNextImageCardStepIndexFromSingleSelection(15, [
+        10,
```
bmd3k, line 996: nit: removing the trailing comma will format this slightly more sanely, I think.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -951,71 +946,63 @@ describe('metrics store utils', () => {
...
+      ]);
+
+      expect(nextStepIndex).toEqual(null);
+    });
+
+    it('does not return step Index when there is no selected step', () => {
```
bmd3k, line 1002: "when there is no selected step" => "when there are no steps"? Looks like selected step is 15 but there are no steps

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
```diff
@@ -786,4 +949,76 @@ describe('metrics store utils', () => {
...
+      const nextStepIndex = getNextImageCardStepIndexFromSingleSelection(
+        cardId,
+        nextStartStep,
+        steps
+      );
+
+      expect(nextStepIndex).toEqual(1);
+    });
+
+    it('dose not update nextStepIndex when there is only one unmatched step', () => {
```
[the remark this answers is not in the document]
bmd3k: Yes, that makes more sense. I was indeed thinking about the line `if (steps.length === 1) return null;`

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
bmd3k: I think you should just reuse buildTimeSeriesData() rather than introducing new createCustomizedTimeSeriesData(). We do this commonly elsewhere in the codebase. You can make override an optional argument and none of the existing usages should notice the change.

tensorboard/webapp/metrics/store/metrics_store_internal_utils_test.ts
bmd3k: > But this one tests the combination of getSelectedSteps and getNextImageCardStepIndexFromSingleSelection and also test the logic selectedTime.end === null.

Is that combination already covered by both `updates cardStepIndex to matched selected time` and `does not update cardStepIndex on selected step with no image`? Together those also test the conditions when getNextImageCardStepIndexFromSingleSelection returns non-null and returns null, respectively.

Just to be clear, I don't think you should mock. I'm saying that I believe your first three tests for 'generateNextCardStepIndexFromSelectedTime' cover all the code and interactions in that function (well, except for when `selectedTime.end !== null`, but I understand that will come later). The remaining tests in this group don't test anything that isn't covered elsewhere and are more complex than their alternatives, so I would recommend removing them.

bmd3k (approved):

japie1235813 (commented):
Removed tests to only keep the first three;
Removed createCutomizedTimeSeriesData method but customized the data in test with buildTimeSeriesData()

PR #5634, opened 2022-03-29
Support for Chinese and other UTF8 characters as folder names

Littleor:

When the folder is named Chinese and other UTF8 characters, it is not possible to toggle Runs on the front end.

As shown below, the toggle will not work because of the `btoa` function.

Screen Shot 2022-03-29 at 19 45 41

The encoding method of the characters before `atob` has been added to avoid the above problem.

* Screenshots of UI changes

None

Create a folder called `中文字符` in the log path and store ri in it, then start tensorboard, toggle Runs in the lower left corner of SCALARS does not work and the console reports an error (see above)

None

### DIFF
--- a/tensorboard/components/tf_storage/storage.ts
+++ b/tensorboard/components/tf_storage/storage.ts
@@ -87,7 +87,11 @@ export const {
   getObserver: getObjectObserver,
   disposeBinding: disposeObjectBinding,
 } = makeBindings(
-  (s) => JSON.parse(decodeURIComponent(escape(window.atob(s)))) as Record<string, string>,
+  (s) =>
+    JSON.parse(decodeURIComponent(escape(window.atob(s)))) as Record<
+      string,
+      string
+    >,
   (o) => btoa(unescape(encodeURIComponent(JSON.stringify(o))))
 );
 export interface StorageOptions<T> {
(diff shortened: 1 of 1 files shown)


bmd3k:
First off, apologies for taking so long to respond. This has been on the back of my mind but I am just getting to it. Thanks for your contribution to TensorBoard!

Have you used the "Time Series" dashboard? It is a new, better-supported dashboard that includes Scalar charts. We have recently elevated it to the first dashboard in the nightly builds (See https://github.com/tensorflow/tensorboard/issues/5704 for more details). It does not have the problem you are seeing in the Scalars dashboard but has the downside that run selection is not currently stored in the URL. Would it be sufficient for you to start using the "Time Series" dashboard instead?

As for this change, there are a couple concerns that I have that I want to think about or explore further:

1. We would want to ensure backward compatibility with the existing implementation to make sure users' bookmarked URLs still work. I have only done a little bit of testing here so far.
2. I hesitate to rely on unescape() to solve the problem given its semi-deprecated state and warnings of undesirable behavior (as descrbied in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape).

So I await your response to the suggestion to use the Time Series dashboard and then we can decide if we need to continue to consider this change.

PR #5636, opened 2022-03-29
linked time: stepIndex moving: moving thumb movement of range selection to stepIndex update in reducer

japie1235813:
Motivation and full context please see https://github.com/tensorflow/tensorboard/pull/5615.

This PR updates stepIndex on range selection:
* Removes the stepIndex in image_container entirely
* Moves the logics (update within/outside range) to `metrics_store_internal_utils`
* Moves corresponding tests in `image_card_test` to `metrics_store_internal_utils_test` cause the thumb movements are no longer updated in the view

This PR also includes some clean up in `metrics_store_internal_utils_test.ts`

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
+++ b/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
@@ -394,8 +394,8 @@ export function generateNextCardStepIndexFromSelectedTime(
     );
 
     let nextStepIndex = null;
-    // Single Selection
     if (selectedTime.end === null) {
+      // Single Selection
       nextStepIndex = getNextImageCardStepIndexFromSingleSelection(
         selectedTime.start.step,
         steps
@@ -506,7 +506,7 @@ function getNextImageCardStepIndexFromSingleSelection(
 
 /**
  * Gets next stepIndex for an image card based on range selection. Returns null if nothing should change.
- * @param selectedSteps The selected steps from selected time. It should contain one or mutliple steps.
+ * @param selectedSteps The selected steps from selected time. It should contain empty or one to multple steps.
  * @param steps The steps in an image card.
  * @param step The step where the current step index locates.
  */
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -493,9 +504,40 @@ function getNextImageCardStepIndexFromSingleSelection(
...
+ * @param selectedSteps The selected steps from selected time. It should contain one or mutliple steps.
+ * @param steps The steps in an image card.
+ * @param step The step where the current step index locates.
+ */
+function getNextImageCardStepIndexFromRangeSelection(
+  selectedSteps: number[],
+  steps: number[],
+  step: number
+): number | null {
+  const firstSelectedStep = selectedSteps[0];
```
bmd3k, line 518: If selectedSteps is [] then this will return undefined. What happens then?

japie1235813: Good question, In practice all check will not pass and it returns null at the end. One test case is when selectedSteps is [].
Since you brought up I think it is not good to let it work as intented from side effect, I added a condition to return early.

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -400,6 +400,17 @@ export function generateNextCardStepIndexFromSelectedTime(
         selectedTime.start.step,
         steps
       );
+    } else {
+      // Range Selection
```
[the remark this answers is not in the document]
japie1235813: Done

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -493,9 +504,38 @@ function getNextImageCardStepIndexFromSingleSelection(
   return null;
 }
 
+/**
+ * Gets next stepIndex for an image card based on range selection. Returns null if nothing should change.
+ * @param selectedSteps The selected steps from selected time. It should contain empty or one to multple steps.
```
bmd3k, line 509: "empty or one to multple steps". That's all possible states of an array so don't know if its just confusing to mention :).

bmd3k (approved):

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
japie1235813: That's true. Replacing it with just mention that it could be empty

PR #5637, opened 2022-03-29
pip: updates werkzeug to use Response in werkzeug 2.1

japie1235813:
werkzeug package is updated to 2.1 which `BaseResponse` is no longer supported. 
https://werkzeug.palletsprojects.com/en/2.0.x/wrappers/
This PR updates all `BaseResponse/BaseRequest` usage to the newer `Response/Request` and removes the constraint of werkzeug version

### DIFF
--- a/tensorboard/pip_package/requirements.txt
+++ b/tensorboard/pip_package/requirements.txt
@@ -29,5 +29,5 @@ requests >= 2.21.0, < 3
 setuptools >= 41.0.0
 tensorboard-data-server >= 0.6.0, < 0.7.0
 tensorboard-plugin-wit >= 1.6.0
-werkzeug >= 0.11.15
+werkzeug >= 1.0.1
 wheel >= 0.26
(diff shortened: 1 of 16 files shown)


tensorboard/pip_package/requirements.txt
```diff
@@ -29,5 +29,5 @@ requests >= 2.21.0, < 3
 setuptools >= 41.0.0
 tensorboard-data-server >= 0.6.0, < 0.7.0
 tensorboard-plugin-wit >= 1.6.0
-werkzeug >= 0.11.15, < 2.1.0
+werkzeug >= 0.11.15
```
nfelt, line 32: While we're modifying this, IMO we might as well update the requirement to `werkzeug >= 1.0.1`. That's the version that internal third_party werkzeug is using, which is generally nice to align with to avoid unexpected feature gaps.  The 0.11.15 version is very old (from 2016, compared to 2020 for 1.0.1) and I don't know of any reason we should still support it.

nfelt (approved):
Thanks for doing this!  Would you mind confirming that this also works when synced internally?

tensorboard/pip_package/requirements.txt
japie1235813: Thanks! Updated.
For testing the internal sync, it passed yesterday but not now because of some other PR got in since that.

PR #5640, opened 2022-03-30
Make summary_minimal target publicly visible

bartvm:
This allows usage of `tensorboard.summary.Writer` while satisfying strict dependency requirements in google3.


nfelt:
This was handled internally (Googlers: see b/228037639), so going to close out this PR if you don't mind.

PR #5641, opened 2022-03-30
linked time: updates cardStepIndex on selected time toggled

japie1235813:
We've updated cardStepIndex on and only on select time changed. This causes issue when select time is enabled the very first time, there is no linked time set yet and the card step index remain unchanged. However, in settings panel we still show the min step and single selection (while the select time is null in ngrx state). This causes a bad ux when the image steps actually matches the min step: logically we should move the step index but we did not. 

This PR sets card step index when linked time is enabled. `generateNextCardStepIndexFromSelectedTime` is reused. If select time is not set, we take the min step and apply single selection to update the image card step index.

Full context please see #5615 

Note: we did not test `generateNextCardStepIndexFromSelectedTime` on `timeSelectionChanged` action. Will add in a follow-up PR.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -942,8 +942,10 @@ const reducer = createReducer(
   }),
   on(actions.selectTimeEnableToggled, (state) => {
     let nextCardStepIndexMap = {...state.cardStepIndex};
+    const nextSelectTimeEnabled = !state.selectTimeEnabled;
+
     // Updates cardStepIndex only when toggle to enable selectedTime.
-    if (!state.selectTimeEnabled) {
+    if (nextSelectTimeEnabled) {
       const {min} = state.stepMinMax;
       const startStep = min === Infinity ? 0 : min;
       nextCardStepIndexMap = generateNextCardStepIndexFromSelectedTime(
@@ -958,7 +960,7 @@ const reducer = createReducer(
     return {
       ...state,
       cardStepIndex: nextCardStepIndexMap,
-      selectTimeEnabled: !state.selectTimeEnabled,
+      selectTimeEnabled: nextSelectTimeEnabled,
     };
   }),
   on(actions.timeSelectionChanged, (state, change) => {
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -943,17 +943,20 @@ const reducer = createReducer(
   on(actions.selectTimeEnableToggled, (state) => {
     let nextCardStepIndexMap = {...state.cardStepIndex};
     const nextSelectTimeEnabled = !state.selectTimeEnabled;
+    const {min} = state.stepMinMax;
+    const startStep = min === Infinity ? 0 : min;
+    const nextSelectedTime = state.selectedTime ?? {
+      start: {step: startStep},
+      end: null,
+    };
 
     // Updates cardStepIndex only when toggle to enable selectedTime.
     if (nextSelectTimeEnabled) {
-      const {min} = state.stepMinMax;
-      const startStep = min === Infinity ? 0 : min;
       nextCardStepIndexMap = generateNextCardStepIndexFromSelectedTime(
         state.cardStepIndex,
         state.cardMetadataMap,
         state.timeSeriesData,
-        // Sets start step to min step when select time is null
-        state.selectedTime ?? {start: {step: startStep}, end: null}
+        nextSelectedTime
       );
     }
 
@@ -961,6 +964,7 @@ const reducer = createReducer(
       ...state,
       cardStepIndex: nextCardStepIndexMap,
       selectTimeEnabled: nextSelectTimeEnabled,
+      selectedTime: nextSelectedTime,
     };
   }),
   on(actions.timeSelectionChanged, (state, change) => {
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -941,17 +941,18 @@ const reducer = createReducer(
     };
   }),
   on(actions.selectTimeEnableToggled, (state) => {
-    let nextCardStepIndexMap = {...state.cardStepIndex};
     const nextSelectTimeEnabled = !state.selectTimeEnabled;
-    const {min} = state.stepMinMax;
-    const startStep = min === Infinity ? 0 : min;
-    const nextSelectedTime = state.selectedTime ?? {
-      start: {step: startStep},
-      end: null,
-    };
+    let nextCardStepIndexMap = {...state.cardStepIndex};
+    let nextSelectedTime = state.selectedTime;
 
     // Updates cardStepIndex only when toggle to enable selectedTime.
     if (nextSelectTimeEnabled) {
+      const {min} = state.stepMinMax;
+      const startStep = min === Infinity ? 0 : min;
+      nextSelectedTime = state.selectedTime ?? {
+        start: {step: startStep},
+        end: null,
+      };
       nextCardStepIndexMap = generateNextCardStepIndexFromSelectedTime(
         state.cardStepIndex,
         state.cardMetadataMap,
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -941,8 +941,23 @@ const reducer = createReducer(
     };
   }),
   on(actions.selectTimeEnableToggled, (state) => {
+    let nextCardStepIndexMap = {...state.cardStepIndex};
+    // Updates cardStepIndex only when toggle to enable selectedTime.
+    if (!state.selectTimeEnabled) {
+      const {min} = state.stepMinMax;
+      const startStep = min === Infinity ? 0 : min;
+      nextCardStepIndexMap = generateNextCardStepIndexFromSelectedTime(
+        state.cardStepIndex,
+        state.cardMetadataMap,
+        state.timeSeriesData,
+        // Sets start step to min step when select time is null
+        state.selectedTime ?? {start: {step: startStep}, end: null}
+      );
+    }
+
     return {
       ...state,
+      cardStepIndex: nextCardStepIndexMap,
       selectTimeEnabled: !state.selectTimeEnabled,
     };
   }),
--- a/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
@@ -2511,7 +2511,7 @@ describe('metrics reducers', () => {
         expect(state3.selectTimeEnabled).toBe(false);
       });
 
-      it('sets cardStepIndex to step 0 when there is no selectedTime in state', () => {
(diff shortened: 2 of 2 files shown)


tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -941,8 +941,23 @@ const reducer = createReducer(
     };
   }),
   on(actions.selectTimeEnableToggled, (state) => {
+    let nextCardStepIndexMap = {...state.cardStepIndex};
+    // Updates cardStepIndex only when toggle to enable selectedTime.
+    if (!state.selectTimeEnabled) {
```
bmd3k, line 946: This is a little confusing from a readability perspective: It reads like "if select time is not enabled" but in practice it actually means "if select time is enabled".

Maybe define: `const selectTimeEnabled = !state.selectTimeEnabled;` and use it here and on L961.

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -941,8 +941,23 @@ const reducer = createReducer(
...
+      const {min} = state.stepMinMax;
+      const startStep = min === Infinity ? 0 : min;
+      nextCardStepIndexMap = generateNextCardStepIndexFromSelectedTime(
+        state.cardStepIndex,
+        state.cardMetadataMap,
+        state.timeSeriesData,
+        // Sets start step to min step when select time is null
+        state.selectedTime ?? {start: {step: startStep}, end: null}
```
bmd3k, line 954: Should we just set state.selectedTime to be {start: {step: startStep}, end: null} if it is null? Rather than the UI having special case for selectedTime is null then just set it to what the UI is pretending it to be.

const nextSelectedTime = state.selectedTime ?? {start: {step: startStep}, end: null};

And use  nextSelectedTime here and in the thing you are returning on L958.

bmd3k (commented):
Looks good but if you follow my advice to set state.selectedTime when linked time is first enabled then I would also expect a bunch of changes to how your new tests are written/organized - I want to see how that looks before approving.

tensorboard/webapp/metrics/store/metrics_reducers.ts
japie1235813: That's interesting idea. I think fundamentally it's not a bad idea to separate the ngrx state and handle what we want to show in UI. 
For example, if user click enable and then toggle again, we then set the selectTime while users are not set it really. It does not feel quite right but I tried to think about caveats but no really strong reasons.

On the other hand, I can also see the convenience of set the selected time here and we do not need to handle cases where select time is null. It is always something when it is enabled. (I can foreseen there are a number of cleanup work, including `timeSelectionCleared` action removal)

What do you think?

for the sake of time I go ahead and made those changes in second commit https://github.com/tensorflow/tensorboard/pull/5641/commits/c75b375d9ad7867abeb09dbd3ed08f14f141cb12

tensorboard/webapp/metrics/store/metrics_reducers.ts
japie1235813: Good suggestion. Updated.

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -941,9 +941,30 @@ const reducer = createReducer(
     };
   }),
   on(actions.selectTimeEnableToggled, (state) => {
+    let nextCardStepIndexMap = {...state.cardStepIndex};
+    const nextSelectTimeEnabled = !state.selectTimeEnabled;
+    const {min} = state.stepMinMax;
+    const startStep = min === Infinity ? 0 : min;
+    const nextSelectedTime = state.selectedTime ?? {
```
bmd3k, line 948: Would it be wise to only initialize nextSelectedTime if nextSelectTimeEnabled? I realize that, right now, in practice, the first time selectTimeEnableToggled is fired it is because the user is turning on linked time BUT it would be good to be certain.

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -2500,6 +2510,189 @@ describe('metrics reducers', () => {
...
+          },
+          cardStepIndex: {[imageCardId]: 2},
+        });
+
+        const state2 = reducers(state1, actions.selectTimeEnableToggled());
+
+        expect(state2.cardStepIndex).toEqual({[imageCardId]: 0});
+      });
+
+      it('sets cardStepIndex to min step when selectedTime is null before toggling', () => {
```
bmd3k, line 2545: I'm even more confident that you don't need this test case. Its definitely covered by `sets cardStepIndex to step 0 when selectedTime is null before toggling` and `sets selectedTime to min step when selectedTime is null before toggling`

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -2500,6 +2510,189 @@ describe('metrics reducers', () => {
...
+            },
+          },
+          cardStepIndex: {[imageCardId]: 2},
+        });
+
+        const state2 = reducers(state1, actions.selectTimeEnableToggled());
+        expect(state2.cardStepIndex).toEqual({[imageCardId]: 2});
+      });
+
+      it('updates step index to pre-existed selectedTime', () => {
```
bmd3k, line 2603: Maybe I would tweak the name a bit: 'updates step index using pre-existing selectedTime'

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -2500,6 +2510,189 @@ describe('metrics reducers', () => {
...
+          },
+          selectedTime: {start: {step: 20}, end: null},
+          cardStepIndex: {[imageCardId]: 2},
+        });
+
+        const state2 = reducers(state1, actions.selectTimeEnableToggled());
+        expect(state2.cardStepIndex).toEqual({[imageCardId]: 1});
+      });
+
+      it('dose not update step index when toggle to disable selectedTime', () => {
```
bmd3k, line 2632: nit: dose: does.

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -2500,6 +2510,189 @@ describe('metrics reducers', () => {
         const state3 = reducers(state2, actions.selectTimeEnableToggled());
         expect(state3.selectTimeEnabled).toBe(false);
       });
+
+      it('sets cardStepIndex to step 0 when selectedTime is null before toggling', () => {
```
bmd3k, line 2514: Up to you but I would feel comfortable without this test. I think its verbosity-to-value ratio is a bit too high. 

It is mostly covered by `updates step index to pre-existed selectedTime` and `sets selectedTime to min step when selectedTime is null before toggling`. The only additional coverage is that you assure that you are passing the newly-initialized nextSelectedTime into generateNextCardStepIndexFromSelectedTime instead of, say, the old state.selectedTime. I would be comfortable assuming that is working without a test.

Again, up to you.

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -2500,6 +2510,189 @@ describe('metrics reducers', () => {
...
+            },
+          },
+          cardStepIndex: {[imageCardId]: 2},
+        });
+
+        const state2 = reducers(state1, actions.selectTimeEnableToggled());
+        expect(state2.cardStepIndex).toEqual({[imageCardId]: 1});
+      });
+
+      it('does not update step index when selectedTime is null before toggling and no matched min step', () => {
```
bmd3k, line 2575: I also don't think you need this test any longer. Also covered by other tests.

bmd3k (approved):

tensorboard/webapp/metrics/store/metrics_reducers.ts
bmd3k: Thinking about it a bit more, there is one caveat: What if the "min" changes over time as new runs are loaded? Under the old style (where selectedTime is null) then all the UI would automatically adjust to this new min when linked time is (re)enabled. Now, with the new code, it will be stuck on the old min.

However, I think it's such a small caveat I wouldn't worry about spending time or code on it. Any defaults we choose when user turns on linked time is very likely wrong anyways.

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
japie1235813: This way it does feel like the whole direction for test cardStepIndex is tweak toward test selectedTime.
I think there are two groups:
1.
sets cardStepIndex to step 0
sets cardStepIndex to min step
2.
sets selectedTime to step 0
sets selectedTime to min step

1-1 and 2-1 should happen together, same for 1-2 and 2-2
but somehow you pick 1-1 and 2-2. Which cause me confused and feel unstructured.

I also thought about combining them to 
sets cardStepIndex and selectedTime to step 0
sets cardStepIndex and selectedTime to min step

which can result longer description but would that satisfy your concern of repeated tests?

bmd3k: I don't think your cardStepIndex tests have to care about either of these edge cases ( "step is 0", "step is min") - it doesn't really have anything to do with cardStepIndex. 

This is my methodology:

I start by looking at the code like I'm a code coverage tool: I look at all the code in the reducer and make sure that each line of code is covered by at least one test. I make sure that each branch/decision (if statements evaluating true and false, trinary operators evaluating true and false, for loops for multiple elements, etc...) is covered by a test.

When I do that, these are the test cases I see:

* selectedTime is not reinitialized if it doesn't have to be
* selectedTime is initialized with start step 0 when appropriate
* selectedTime is initialized with start step min when appropriate
* we call generateNextCardStepIndexFromSelectedTime when selectTimeEnabled and do something with the result
* we don't call generateNextCardStepIndexFromSelectedTime when !selectTimeEnabled (and so you effectively ignore the result it would return if you did call it)

It's not a perfect mechanism but it usually identifies 80 or 90% of the tests you want to write. To supplement, I also look for places where there might be risk of subtle bugs: Using the wrong variable in place of another, for example. When I do that, this is what I see:

* we pass the new  selectedTime value to generateNextCardStepIndexFromSelectedTime instead of the old null one (nextSelectedTime instead of state.selectedTime). 

What I don't see when I do this sort of analysis is:
* cardStepIndex cares about whether selectedTime.start.step = 0
* cardStepIndex cares about whether selectedTime.start.min = 0

Both those cases are just side-effects of initializing selectedTime and have nothing to do with cardStepIndex in my mind.

To think about it yet another way: If you modified the lines that initialize selectedTime step to 0 or min, do you have at least one test that fails? If you already have one, do you really need another?

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
bmd3k: By the way, another way to think about this: What we really care about is that we don't pass state.selectedTime when it is null. One way to guarantee this without writing a test is to change the types of generateNextCardStepIndexFromSelectedTime() so that selectedTime is not nullable. If that can be done, that's probably better than a test.

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
japie1235813: Thanks for the comment.
Updated `cardStepIndex` related test to only test "selectedTime.start.step = 0".

PR #5644, opened 2022-03-31
Move AxisDirection definition to linked-time-types file

JamesHollyer:
This change moves the AxisDirection definition from the linked_time_fob_controller_component.ts file to the linked_time_types.ts file.

The AxisDirection is used in multiple places and going to be used in more places in the future. It makes more sense to have it's type definitions in the file specifically made for types.

Also, I realized the ScalarCardComponent was still using AxisDirection when it did not need to so I simply removed it instead of updating the path.

### DIFF
--- a/tensorboard/webapp/widgets/linked_time_fob/linked_time_types.ts
+++ b/tensorboard/webapp/widgets/linked_time_fob/linked_time_types.ts
@@ -21,9 +21,7 @@ export interface LinkedTime {
 }
 
 /**
- * The linked time fob controller can be used inside both horizontal and
- * vertical axis. This enum is a utility used to inform to the controller of
- * that direction.
+ * The direction of the axis used to control the fob movements.
  */
 export enum AxisDirection {
   HORIZONTAL,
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/widgets/linked_time_fob/linked_time_types.ts
```diff
@@ -20,6 +20,16 @@ export interface LinkedTime {
   end: {step: number} | null;
 }
 
+/**
+ * The linked time fob controller can be used inside both horizontal and
+ * vertical axis. This enum is a utility used to inform to the controller of
```
japie1235813, line 25: nit: the description is a bit confusing to me. It sound like "linked time fob controller" is used in "axis" but I think you mean the other way?
And for type definition, I think it is okay to just say what it is first, and add more about how it might be used later if it's not sufficient to just say "what it is"

for example,
The direction of axis used to control the fob movements.

japie1235813 (approved):

tensorboard/webapp/widgets/linked_time_fob/linked_time_types.ts
JamesHollyer: Changed.

The Controller does actually live "inside" the axis. But that is mostly academic.

PR #5649, opened 2022-04-04
linked time: stepIndex moving: tests for card step updated on timeSelectionChanged

japie1235813:
Adds tests for `generateNextCardStepIndexFromSelectedTime` usage in `metrics_reducer`. Full context please see #5615 

The changes of `selectedTime` in this reducer have been tested. The logics of the method also have been tested in `stor_internal_utils` and not intent to repeat here. The tests on this higher reducer level is just to test the method returns `cardStepIndex`, which used to update (or not update) the next `cardStepIndex` state.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
@@ -2495,7 +2495,7 @@ describe('metrics reducers', () => {
         expect(nextState1.useRangeSelectTime).toEqual(false);
       });
 
-      it('sets `cardStepIndexMap` on selected time changed', () => {
+      it('sets `cardStepIndex` when step matches selected time', () => {
         const beforeState = buildMetricsState({
           selectTimeEnabled: false,
           cardMetadataMap,
@@ -2530,7 +2530,7 @@ describe('metrics reducers', () => {
         expect(nextState.cardStepIndex).toEqual({[imageCardId]: 0});
       });
 
-      it('does not set `cardStepIndexMap` on selected time changed', () => {
+      it('does not set `cardStepIndex` when steps do not match selected time', () => {
         const beforeState = buildMetricsState({
           selectTimeEnabled: false,
           cardMetadataMap,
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -2484,6 +2494,77 @@ describe('metrics reducers', () => {
 
         expect(nextState1.useRangeSelectTime).toEqual(false);
       });
+
+      it('sets `cardStepIndexMap` on selected time changed', () => {
```
JamesHollyer, line 2498: should 'cardStepIndexMap' just be cardStepIndex?

JamesHollyer: Also, maybe say something like "when new step has an image".

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -2484,6 +2494,77 @@ describe('metrics reducers', () => {
...
+          actions.timeSelectionChanged({
+            startStep: 10,
+            endStep: undefined,
+          })
+        );
+
+        expect(nextState.cardStepIndex).toEqual({[imageCardId]: 0});
+      });
+
+      it('does not set `cardStepIndexMap` on selected time changed', () => {
```
JamesHollyer, line 2533: cardStepIndex? Also, state why this one does not set cardStepIndex while above test does set it.

JamesHollyer (approved):
Am I correct in assuming all the edge case logic here is tested in the generateNextCardStepIndexFromSelectedTime tests? So this is just ensuring that the results from generateNextCardStepIndexFromSelectedTime get used correctly?

If so, LGTM mod test name changes.

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
japie1235813: Good point. Updated the description to make these two tests meaningful on why are they different.
also remove "on selected time changed" since all tests are under "#timeSelectionChanged" already

japie1235813 (commented):
Thanks for the review. Yes, you're correct. All edge cases have been tested on the function itself. The newly added tests are just to see cardStepIndex got updated and not updated.

PR #5650, opened 2022-04-04
linked time: cleanup: rename stepValues to steps

japie1235813:
As in previous PR comment #5624 from @bmd3k, the "value" in "stepValues" are meaningless. This PR simply renames all "stepValues" to "steps" for image card.

No additional logic are addressed.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
+++ b/tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
@@ -385,11 +385,7 @@ export function generateNextCardStepIndexFromSelectedTime(
   Object.keys(previousCardStepIndex).forEach((cardId) => {
     if (!cardId.includes('"plugin":"images"')) return;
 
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
```diff
@@ -422,7 +418,7 @@ export function generateNextCardStepIndexFromSelectedTime(
 /**
  * Returns step values of an image card.
```
JamesHollyer, line 419: "Returns steps of an image card."

JamesHollyer (approved):
LGTM after comment update.

tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts
japie1235813: Good catch! Thanks!

PR #5651, opened 2022-04-05
Fix option wrapping within tb-toolbar

ericdnielsen:
Long options within a tb-toolbar would wrap in an overlapping way.

Added a span around the option, with the non-truncated/full value as the title and the suppressed the wrapping.


yatbear:
Could you add a screenshot of the end result in the description? Thanks!

ericdnielsen:
I'll send you a screen shot internally.  I don't want to put the user data from the reproducible example I've been working from here.

yatbear (approved):
Thanks!

PR #5653, opened 2022-04-05
Horizontal fobs should be centered on X axis not Y axis

JamesHollyer:
This fixes a bug that centers the fob in the wrong direction when the axisDirection is set to horizontal.

The horizontal-fob class is currently not used but, it will be used when the LinkedTimeFobController is used in ScalarCards. Therefore there is currently no functional change.

### DIFF
--- a/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.scss
+++ b/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.scss
@@ -29,10 +29,10 @@ limitations under the License.
   }
 }
 
-.verticalFob {
+.vertical-fob {
   transform: translateY(-50%);
 }
 
-.horizontalFob {
+.horizontal-fob {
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.scss
```diff
@@ -34,5 +34,5 @@ limitations under the License.
 }
 
 .horizontalFob {
```
japie1235813, line 36: the css name should be dash-case instead of CamelCase

japie1235813 (approved):
What is not clear to me is how

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.scss
JamesHollyer: Oh yea definitely. I changed the case.

PR #5659, opened 2022-04-07
build: upgrade rules_nodejs and npm deps to 4.6.1

JamesHollyer:
This updates the npm Bazel packages as well as the rules_nodejs package we download to version 4.6.1. 

The hope is that this will allow us to also upgrade Angular. However, staying as up to date with Bazel as we can is just good practice on it's own.


nfelt (approved):
Code changes LGTM, but PR title/description could be a little clearer:

> This updates Bazel to version 4.6.1. 

This is a little misleadingly phrased, since it sounds like we're updating the actual Bazel binary version we use to build, i.e. this value in our CI config, which isn't changed in this PR and is still 4.0.0: https://github.com/tensorflow/tensorboard/blob/0fc9c6c0a3de06e9e60a48fefa3c229a41155403/.github/workflows/ci.yml#L30

I'd suggest a PR title (and resulting commit message) more like:

> build: upgrade rules_nodejs and npm deps to 4.6.1

PR #5660, opened 2022-04-08
linked time: utilize ScalarCardLinkedTimeFobController in the ScalarCardComponent

### LINKED ISSUE #5656: linked time fob is not being centered properly in Scalar Cards
On the Scalar Card the linked time fob has it's "left" aligned with the step associated with it. This fob should be centered around that step.

Screen Shot 2022-04-06 at 11 07 37 AM

JamesHollyer:
This PR replaces the linked time fob and all the logic for that fob inside the ScalarCardComponent with the ScalarCardLinkedTimeController. Currently the ScalarCardLinkedTimeController only has logic for fob placement. Therefore the only functional change here is that the fob will now be centered properly(this resolves #5656 )

The goal is to eventually implement all the necessary logic to add the fob dragging feature in the Scalar Card. Currently pointer events are set to none(this is done [here](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/components/vz_line_chart2/vz-line-chart2.ts;l=572)) so clicking and dragging does nothing. Once the logic for dragging is added that will be turned on.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -186,7 +186,10 @@ export class ScalarCardComponent<Downloader> {
     });
   }
 
-  getLinkedTime(): LinkedTime {
+  getLinkedTime(): LinkedTime | null {
+    if (this.selectedTime === null) {
+      return null;
+    }
     return {
       start: {
         step: this.selectedTime!.startStep,
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -37,6 +37,7 @@ import {
   ScaleType,
   TooltipDatum,
 } from '../../../widgets/line_chart_v2/types';
+import {LinkedTime} from '../../../widgets/linked_time_fob/linked_time_types';
 import {TooltipSort, XAxisType} from '../../types';
 import {
   ScalarCardDataSeries,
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -184,4 +185,15 @@ export class ScalarCardComponent<Downloader> {
       data: {cardId: this.cardId},
     });
   }
+
+  getLinkedTime(): LinkedTime {
+    return {
```
japie1235813, line 190: nit: I understand there is a ngIf* to check if selected time is null.
but the linkedTime input of `scalar-card-linked-time-fob-controller` could still be null
I think it's better to add the check and then dropped the escalation later 
```
    if (selectedTime === null) {
      return null;
    }
```

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2124,12 +2127,12 @@ describe('scalar card', () => {
         const fixture = createComponent('card1');
         fixture.detectChanges();
 
-        const fobs = fixture.debugElement.queryAll(
-          Selector.LINKED_TIME_AXIS_FOB
-        );
+        let testController = fixture.debugElement.query(
```
japie1235813, line 2130: nit: const

japie1235813 (approved):

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
JamesHollyer: Done.

PR #5662, opened 2022-04-11
angular: update FormControl usages to be forward compatible with typed forms.

dylhunn:
Tensorboard has a couple `FormControl`s which will be broken when typed forms lands in v14 inside google3. Ordinarily, you'd want to use the opt-out symbol (`UntypedFormControl`). However, since Tensorboard needs to be simultaneously compatible with both v12 (on Github) and v14 (inside google3), we can use this cast as a trick to disable the new types.

Forward compatibility with upcoming changes in google3.

Add casts to `FormControl` in order to prevent type inference.

* Screenshots of UI changes

N/A -- types only change. No runtime effect is possible.

`blaze build //third_party/tensorboard/webapp/settings/_views:_views`

go/typed-forms-lsc


JamesHollyer (approved):
LGTM

PR #5663, opened 2022-04-11
time namespaced: fix reload operation after pinned card

japie1235813:
Issue:
Pinned cards disappear on clicking auto reload button on the top right cause. Googlers please see b/228606608.

Fix
After #5627 we use a new state, "cardToPinnedCopyCache", to remember all cards that have been previous pinned within the same namespace. The state intends to be updated whenever a card is pinned/unpinned (action "cardPinStateToggled"). Yet the update is somehow dropped at the reducer level. This PR puts it back. Alone with other "refresh" (metadata loaded, state rehydrate from url)

Test:
- This failure is better to be caught on webtest, which I just started work on and noticed something went wrong. User noticed this earlier. Will continue work on the webtest later.
- Updated a couple of places to also test the new state "cardToPinnedCopyCache"

TODO cleanup:
integrate this state into `unresolvedImportedPinnedCards ` and `buildOrReturnStateWithUnresolvedImportedPins`.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
@@ -362,10 +362,7 @@ describe('metrics reducers', () => {
         },
         cardList: [cardId1],
         cardToPinnedCopy: new Map([[cardId1, pinnedCopyId1]]),
-        cardToPinnedCopyCache: new Map([
-          [cardId1, pinnedCopyId1],
-          [cardId2, pinnedCopyId2],
-        ]),
+        cardToPinnedCopyCache: new Map([[cardId1, pinnedCopyId1]]),
         pinnedCardToOriginal: new Map([[pinnedCopyId1, cardId1]]),
       });
       expect(nextState.cardMetadataMap).toEqual(expectedState.cardMetadataMap);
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -362,10 +362,7 @@ describe('metrics reducers', () => {
         },
         cardList: [cardId1],
         cardToPinnedCopy: new Map([[cardId1, pinnedCopyId1]]),
-        cardToPinnedCopyCache: new Map([
-          [cardId1, pinnedCopyId1],
-          [cardId2, pinnedCopyId2],
-        ]),
+        cardToPinnedCopyCache: new Map([[cardId1, pinnedCopyId1]]),
```
JamesHollyer, line 365: This does not seem in scope of this change but, I guess it is ok.

JamesHollyer (approved):

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
japie1235813: so `cardToPinnedCopyCache` is designed to be aligned with any pin/unpin actions towards "cardToPinnedCopy".
Only when experiment removal (not the actual cards pin/unpin) this two maps could be different (and `cardToPinnedCopyCache` is a superset of `cardToPinnedCopy`)

this test checks "pinned cards removal", which is an unpin action, so we should also check here.
this bug should be caught earlier with this test only if I misunderstood removal of pinned cards to removal of experiments

japie1235813: Correct myself here, this is _not_ unpin action but remove the cards from card list. The fix in in #5678

PR #5664, opened 2022-04-12
linked time: highlight the closest step in histogram

japie1235813:
When there is no selected step on histogram card in single selection, it's easy to get confused (no color). We want to highlight the closest step with data on this case.

Summary: update `this.selectedTime$` in `histogram_card_container.ts`
On histogram, to determine whether this card is "clipped" we have created "ViewSelectedTime" type and make the decision in histogram_container. Similarly, we adjust the step (to be specific, `startStep`) of `selectedTime` to the closest step accordingly. This makes sense to me because "selectedTime" is type "ViewSelectedTime". This type is to use in view only. 
Also all these changes are in view.  We don't change the LinkedTime-type state. Adjust the step here, and propagates it to  widget/historgram and histogram fob, both of which are passively rendering the view.

* Screenshots of UI changes
selected step=129; closest step=133
Screen Shot 2022-04-12 at 3 43 16 PM

* TODO:
1. Show icon indicator when we highlight the closest step instead of selected step.
2.  Consider change the input of <tb-histogram [linkedTime]=".."> to be other names. We can use slectedTime (type ViewSelectedTime) or change the name. To avoid confusion in the widget that we actually _not_ use state-level linked time.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/histogram_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/histogram_card_container.ts
@@ -22,7 +22,7 @@ import {
 } from '@angular/core';
 import {Store} from '@ngrx/store';
 import {combineLatest, Observable} from 'rxjs';
-import {combineLatestWith, filter, map} from 'rxjs/operators';
+import {filter, map} from 'rxjs/operators';
 import {State} from '../../../app_state';
 import {DataLoadState} from '../../../types/data';
 import {RunColorScale} from '../../../types/ui';
@@ -42,7 +42,12 @@ import {
 import {CardId, CardMetadata, LinkedTime} from '../../types';
 import {CardRenderer} from '../metrics_view_types';
 import {getTagDisplayName} from '../utils';
-import {maybeClipSelectedTime, ViewSelectedTime} from './utils';
+import {
+  getClosestNonTargetStep,
+  maybeClipSelectedTime,
+  maybeSetClosestStartStep,
+  ViewSelectedTime,
+} from './utils';
 
 type HistogramCardMetadata = CardMetadata & {
   plugin: PluginType.HISTOGRAMS;
@@ -100,6 +105,8 @@ export class HistogramCardContainer implements CardRenderer, OnInit {
   isPinned$?: Observable<boolean>;
   selectedTime$?: Observable<ViewSelectedTime | null>;
   isSelectedTimeClipped$?: Observable<boolean>;
+  steps$?: Observable<number[]>;
+  closestStep$?: Observable<number | null>;
 
   private isHistogramCardMetadata(
     cardMetadata: CardMetadata
@@ -150,18 +157,43 @@ export class HistogramCardContainer implements CardRenderer, OnInit {
       })
     );
 
-    this.selectedTime$ = this.store.select(getMetricsSelectedTime).pipe(
-      combineLatestWith(this.data$),
-      map(([selectedTime, data]) => {
+    this.steps$ = this.data$.pipe(
+      map((data) => data.map((datum) => datum.step))
+    );
+
+    this.closestStep$ = combineLatest([
+      this.store.select(getMetricsSelectedTime),
+      this.steps$,
+    ]).pipe(
+      map(([selectedTime, steps]) => {
+        // Only calculates the closest step in single selection.
+        if (!selectedTime || !selectedTime.start || selectedTime.end !== null)
+          return null;
+
+        return getClosestNonTargetStep(selectedTime.start.step, steps);
+      })
+    );
+
+    this.selectedTime$ = combineLatest([
+      this.store.select(getMetricsSelectedTime),
+      this.steps$,
+      this.closestStep$,
+    ]).pipe(
+      map(([selectedTime, steps, closestStep]) => {
         if (!selectedTime) return null;
 
         let minStep = Infinity;
         let maxStep = -Infinity;
-        for (const datum of data) {
-          minStep = Math.min(datum.step, minStep);
-          maxStep = Math.max(datum.step, maxStep);
+        for (const step of steps) {
+          minStep = Math.min(step, minStep);
+          maxStep = Math.max(step, maxStep);
         }
-        return maybeClipSelectedTime(selectedTime, minStep, maxStep);
+        const viewSelectedTime = maybeClipSelectedTime(
+          selectedTime,
+          minStep,
+          maxStep
+        );
+        return maybeSetClosestStartStep(viewSelectedTime, closestStep);
       })
     );
 
--- a/tensorboard/webapp/metrics/views/card_renderer/histogram_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/histogram_card_container.ts
@@ -43,7 +43,6 @@ import {CardId, CardMetadata, LinkedTime} from '../../types';
 import {CardRenderer} from '../metrics_view_types';
 import {getTagDisplayName} from '../utils';
 import {
-  getClosestNonSelectedStep,
   maybeClipSelectedTime,
   maybeSetClosestStartStep,
   ViewSelectedTime,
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/metrics/views/card_renderer/histogram_card_container.ts
```diff
@@ -150,18 +157,43 @@ export class HistogramCardContainer implements CardRenderer, OnInit {
...
+          minStep = Math.min(step, minStep);
+          maxStep = Math.max(step, maxStep);
         }
-        return maybeClipSelectedTime(selectedTime, minStep, maxStep);
+        const viewSelectedTime = maybeClipSelectedTime(
+          selectedTime,
+          minStep,
+          maxStep
+        );
+        return maybeSetClosestStartStep(viewSelectedTime, closestStep);
```
JamesHollyer, line 196: This can be greatly simplified.

`if (viewSelectedTime.endStep === null) {
          viewSelectedTime.startStep = getClosestStep(
            viewSelectedTime.startStep,
            steps
          );
        }`

getClosestStep is like getClosestNonSelectedStep except if the selected step exists it just return it.

`export function getClosestStep(selectedStep: number, steps: number[]): number {
  let minDistance = Infinity;
  let closestStep = null;
  for (const step of steps) {
    const distance = Math.abs(selectedStep - step);
    if (distance < minDistance) {
      minDistance = distance;
      closestStep = step;
    }
  }

  return closestStep || selectedStep;
}`

Then you can get rid of the closestNonSelectedStep$ variable entirely and both of those utility functions.

Let me know what you think.

JamesHollyer (commented):
I had a pretty big refactor/simplification idea. I tried it out and it seems to work just as well. Let me know if my comment makes sense. Happy to explain more or have discuss any pros and cons.

tensorboard/webapp/metrics/views/card_renderer/histogram_card_container.ts
japie1235813: 1.I considered similar idea of your "getClosestStep" but I didn't go with it. It could be even simpler that `closestStep` could be the same as `selectedStep` (distance is 0)

The reason why I need "closestNonSelectedStep" instead of "closestStep" is I need to know if the step is "adjusted" at all. (To show the indicator) 

concept like: "isClosestStepHighLighted$ =true when closestNonSelectedStep$ !== null"

2. the reason why I prefer not put condition in your first code block is that it is less readable. What does it mean that "endStep===null"? (believe me, Brian asked the same question) Why do I set startStep when the condition is true? then we ended up adding more comments while I think the function name and docstring could be a better place to explain.

japie1235813: 1. Oh yap, that sounds a good idea. Thanks!
2. TBH I still prefer wrapping that part in a method to keep this block shorter and readable. I also want to update "maybeSetClosestStartStep" a bit to avoid in-place modification.

JamesHollyer: 2. I personally think that it is more readable to say if not range selection set start step to closest step.  However, that sounds like personal preference so I am ok with you doing it your way. Carry on.

japie1235813: 2. Thanks. I also want to avoid some confusion, like, if "the closest step" is start step itself, technically we don't need to anything. Assigning it to startStep no matter what looks confusing. (Although practically the value does not get changed). So wrapping it in a function with a "maybe" statement makes more sense to me.

japie1235813 (commented):
Summary of revision:
1. I update method to getClosestStep to define it as: "find the closest step in this steps array" 
which means it can still returns null if it cannot find any closest step (only happens when steps is empty array)
return "closestStep ?? selectedStep" does not make sense to me to fulfill the definition

2. the above method is used in "maybeSetClosestStartStep".
removes unnecessary variable in "historgram_card_container", which leaves the only effective change is line 181.

3.adjust maybeSetClosestStartStep to reuse getClosestStep:
a.still guarantee we don't calculate closest step when it's not needed on range selection
b.but we waste time recalculate it if startStep is in steps (makes it the closest one)
I can add a check but the complexity still the same (using `indexOf()` is still O(n)) so I leave it for now and added a comment stating they're equivalent.

tensorboard/webapp/metrics/views/card_renderer/histogram_card_container.ts
```diff
@@ -150,18 +155,30 @@ export class HistogramCardContainer implements CardRenderer, OnInit {
...
     );
 
-    this.selectedTime$ = this.store.select(getMetricsSelectedTime).pipe(
-      combineLatestWith(this.data$),
-      map(([selectedTime, data]) => {
+    this.steps$ = this.data$.pipe(
+      map((data) => data.map((datum) => datum.step))
+    );
+
+    this.selectedTime$ = combineLatest([
```
JamesHollyer, line 162: I thought we had decided to have selectedTime be separate and then that was going to be used to get viewSelectedTime. If you have decided that is not needed or if you plan on doing that in another PR can you please rename this to viewSelectedTime to help clarify that this is not the true selectedTime?

tensorboard/webapp/metrics/views/card_renderer/histogram_card_test.ts
```diff
@@ -298,7 +298,7 @@ describe('histogram card', () => {
         By.directive(TestableHistogramWidget)
       );
       expect(viz.componentInstance.linkedTime).toEqual({
-        start: {step: 5},
+        start: {step: 1},
```
JamesHollyer, line 301: I am confused why this is 1. I am assuming that the closes step is 1 but, I do not see where the steps are set at all. Can we make that more obvious? Ideally we set the steps right here in this step and maybe make the steps around step 5 so this is more interesting. ex: [1, 2, 4, 8] and expect it to give 4.

tensorboard/webapp/metrics/views/card_renderer/utils_test.ts
```diff
@@ -243,4 +248,62 @@ describe('metrics card_renderer utils test', () => {
...
+      };
+
+      expect(maybeSetClosestStartStep(viewSelectedTime, [10, 20, 30])).toEqual({
+        startStep: 10,
+        endStep: null,
+        clipped: false,
+      });
+    });
+
+    it('does not set startStep when closest step is null', () => {
```
JamesHollyer, line 267: optional nit: I would rather this be called 'does not set startStep when passed an empty array of steps'

JamesHollyer (commented):
I think this version is much cleaner. I still have a couple comments. Let me know what you think.

tensorboard/webapp/metrics/views/card_renderer/histogram_card_container.ts
japie1235813: Yap I plan to rename it in next PR when doing the comparison. Please wait a bit.

tensorboard/webapp/metrics/views/card_renderer/histogram_card_test.ts
japie1235813: the data is set in "provideMockCardSeriesData" in testing.ts and the histogram data provided is [0, 1, 99] 
(same for scalar and image data, but we don't care here)
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/testing.ts;l=182?q=provideMockCardSeriesData&ss=tensorflow%2Ftensorboard

added a comment to help clarify

tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -137,3 +137,53 @@ export function maybeClipSelectedTime(
...
+
+/**
+ * Given an array of steps, returns the closest step to the selected step. Returns null
+ * if there is no closest step, which only happens with empty steps array.
+ */
+export function getClosestStep(
+  selectedStep: number,
+  steps: number[]
+): number | null {
+  if (steps.length === 0) {
```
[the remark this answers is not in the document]
japie1235813: Removed.

tensorboard/webapp/metrics/views/card_renderer/utils_test.ts
japie1235813: make sense to me. thanks!

JamesHollyer (approved):
LGTM

PR #5665, opened 2022-04-13
linked time: Wire up timeSelectionChange action

JamesHollyer:
In order to allow the LinkedTimeFobController to implement dragging it must be able to dispatch the timeSelectionChanged action. This pulls that action into the ScalarCardContainer and passes it down to give the controller access.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
@@ -51,6 +51,8 @@ import {
 import {DataLoadState} from '../../../types/data';
 import {classicSmoothing} from '../../../widgets/line_chart_v2/data_transformer';
 import {ScaleType} from '../../../widgets/line_chart_v2/types';
+import {LinkedTime} from '../../../widgets/linked_time_fob/linked_time_types';
+import {timeSelectionChanged} from '../../actions';
 import {PluginType, ScalarStepDatum} from '../../data_source';
 import {
   getCardLoadState,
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -517,4 +520,13 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
       };
     });
   }
+
+  onSelectTimeChanged(linkedTime: LinkedTime) {
```
japie1235813, line 524: nit: consider naming the input to "newLinkedTime" or "newValue" to make it more meaningful

see 
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/views/right_pane/settings_view_container.ts;l=195?q=%22onSelectTimeChanged%22&ss=tensorflow%2Ftensorboard

japie1235813 (commented):
Any tests for this change?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
JamesHollyer: Yea that makes sense I renamed it newLinkedTime. This should probably be done in the histogram_card_container as well but, out of scope of this PR.

JamesHollyer (commented):
The only meaningful tests I can think of will come in the integration tests once this is actually being called.

japie1235813 (approved):
so if there is nothing meaningful to call the output onSelectTimeChanged in scalar_card for this PR, please consider putting down something meaningful so then it is testable.
"onSelectTimeChanged" dose trigger an action, so I would like to see where you test the action is triggered.

JamesHollyer (commented):
The meaningful tests will be coming soon in a future PR.

PR #5669, opened 2022-04-15
Suppress external/com_google_protobuf warning in CI.

bileschi:
Bazel build CI is getting clogged with many warnings of the form:

`external/com_google_protobuf/python: warning: directory does not exist.`

The correct fix for this is to incorporate a version of proto which includes https://github.com/protocolbuffers/protobuf/issues/6049.   In the meantime, this kludge will make it easier to debug genuine issues.

Googlers: see b/222706811

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
           ( set -o pipefail;
             bazel build //tensorboard/... |&\
             grep -v 'external/com_google_protobuf/python: warning: directory' |&\
-            grep -v 'INFO: From ProtoCompile external/com_google_protobuf/python/google/protobuf/' )
+            grep -v 'INFO: From ProtoCompile ' )
       - name: 'Bazel: test (with TensorFlow support)'
         run: bazel test //tensorboard/...
         if: matrix.tf_version_id != 'notf'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,8 +91,9 @@ jobs:
         run: bazel fetch //tensorboard/...
       - name: 'Bazel: build'
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -90,7 +90,10 @@ jobs:
       - name: 'Bazel: fetch'
         run: bazel fetch //tensorboard/...
       - name: 'Bazel: build'
-        run: bazel build //tensorboard/...
+        run: |
```
bmd3k, line 93: Is there a way to comment in a yml file? Can you comment about why you're supressing the error message?

Similarly, is there a place we can leave a comment near where the real fix should go? Then we can comment that this kludge should be removed?

.github/workflows/ci.yml
```diff
@@ -90,7 +90,10 @@ jobs:
...
+          ( set -o pipefail;
+            bazel build //tensorboard/... |&\
+            grep -v 'external/com_google_protobuf/python: warning: directory' )
```
bmd3k, line 96: This does leave a bunch of dangling messages like:

INFO: From ProtoCompile external/com_google_protobuf/python/google/protobuf/any_pb2.py [for host]:
INFO: From ProtoCompile external/com_google_protobuf/python/google/protobuf/source_context_pb2.py [for host]:

Which might be confusing given that we're not acctually printing out the associated warnings from these files. Is there a way to also suppress these lines when the only associated error message is the one you are currently trying to suppress?

bmd3k (approved):
Wow, it seems like you eliminated > 95% of the logs for this step.

New: https://github.com/tensorflow/tensorboard/runs/6042319231?check_suite_focus=true

Old: https://github.com/tensorflow/tensorboard/runs/6041780709?check_suite_focus=true

PR #5672, opened 2022-04-18
linked time: wire selectTimeEnableToggled action down to LinkedTimeFobController

JamesHollyer:
Our goal is to eventually add a delete button on the fobs. This is the first change towards that goal.  When in range selection the deletion of a fob will put us into single selection with the remaining fob as the selected step. When we are in single selection the delete button will disable the selected time entirely.

Since we will need the ability to disable selected time from a fob we need to be able to dispatch the selectTimeEnableToggled action from the fob controller. In order to follow our structure this action is imported into the containers(HistogramCardContainer and ScalarCardContainer) and is passed down as an event until it reaches the LinkedTimeFobController.

I change the fob click event to call the event and ensured it did properly disable the selectedTime

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/histogram_card_component.ng.html
+++ b/tensorboard/webapp/metrics/views/card_renderer/histogram_card_component.ng.html
@@ -65,6 +65,7 @@ limitations under the License.
   [color]="runColorScale(runId)"
   [linkedTime]="convertToLinkedTime(selectedTime)"
   (onSelectTimeChanged)="onSelectTimeChanged.emit($event)"
+  (onSelectTimeToggle)="onSelectTimeToggle.emit()"
 ></tb-histogram>
 <ng-template #noData>
   <div *ngIf="loadState === DataLoadState.FAILED" class="empty-message">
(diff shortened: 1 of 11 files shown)


japie1235813 (commented):
I think a better breakdown would be implementing the fob dismiss button and then test "onSelectTimeToggle" is emitted. (I would image more cases should be handled because we don't want to emit `selectTimeEnableToggled` action when it is range selection. instead it is `timeSelectionChanged`.
For this PR, it's hard to say a pass for many emit actions and eventually dispatch in store without any tests.

Another part is, I think it's different between "toggle selection" and "deselect a fob in single selection". We still "preserve" the start step for the former (so then when it is enabled, we have the previous start step set). Yet what should we do for the latter? In setting panel, when the lower input is deleted, it is set to min step and is not disabled. I feel we should align them.

Another possible change I've not thought clearly is, could we make the input of timeSelectionChanged in metrics_reducer to take empty start step too? and then when it is null start step, we disable the linked time.

JamesHollyer:
I am willing to go the other way around if you would prefer. Yes, I am aware timeSelectionChanged is what is used when deleting a fob while in range selection. I have written up that logic locally.

The tests were going to be in another PR this was just a preliminary one to do some wiring. 

I personally think toggling selected time is the most logical and expected way to deselect a single selection. I see it as different from deleting the step in the typed step settings. It would be very confusing for them to deselect a fob and instead of being removed it just jumped to the first step. I don't think they need to be aligned. The alignment of the typed step in settings is with the typed step in the fob not with the deselection of the fob.

I thought about that too and decided against it. The LinkedTime and SelectedTime types currently do not allow for start step to be null and it would cause a lot of problems if they did.

I still think this is a decent first step towards what we need. Maybe we can discuss further offline? I am willing to add tests to ensure the wiring is done correctly.

PR #5678, opened 2022-04-20
time namespaced: fix preserving pinned card of previous experiments 

japie1235813:
When fixing a previous bug where pinned cards disappear after hitting reload button (#5663), I also assign "" for `stateRehydratedFromUrl` action. Yet the assignment for `metricsTagMetadataLoaded` action was incorrect, which make `cardToPinnedCopyCache` identical to `cardToPinnedCopy`, which makes the initial purpose (#5627) not working.
This PR removes the assignment.

The unit tests are "unpins a pinned copy", which we _should_ update `cardToPinnedCopyCache` and "updates pinned/original cards mapping on pinned cards removal" which we should _not_ update `cardToPinnedCopyCache` but `cardToPinnedCopy`

TODO: add webtests after sync (actually, this bug was caught in the process of adding the tests)

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -573,7 +573,6 @@ const reducer = createReducer(
       return {
         ...state,
         ...resolvedResult,
-        cardToPinnedCopyCache: resolvedResult.cardToPinnedCopy,
         tagGroupExpanded,
         tagMetadataLoadState: {
           state: DataLoadState.LOADED,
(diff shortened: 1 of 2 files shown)


bmd3k (approved):
I don't really understand what you mean by `I also assign "" for stateRehydratedFromUrl action` in the description but the change looks fine.

japie1235813 (commented):
I mean I should also handle the cache update when we read pinned cards from url. Anyway it is not directly relevant to this PR. Sorry if I was over explaining things and make it sound complex.

PR #5684, opened 2022-04-22
Unsubscribes ReloaderComponent from state changes when destroyed.

bmd3k:
In Google-internal TensorBoard instances, the ReloaderComponent might be loaded and destroyed multiple times. Unfortunately ReloaderComponent does not properly unsubscribe from ngrx Store changes when being destroyed. As described in b/230111530 this leads to strange behavior where we can end up having multiple auto-reload cycles in parallel because we have multiple ReloaderComponent instances (all but one is stale) trying to create these cycles. 

ReloaderCompoment manually subscribes to the Store in ngOnInit(). As we've seen in other Components in our codebase this requires a special mechanism to unsubscribe (this is the takeUntil(ngUnsubscribe) line). We reuse the pattern used in the other Components and apply it here.

### DIFF
--- a/tensorboard/webapp/reloader/reloader_component.ts
+++ b/tensorboard/webapp/reloader/reloader_component.ts
@@ -15,8 +15,8 @@ limitations under the License.
 import {DOCUMENT} from '@angular/common';
 import {ChangeDetectionStrategy, Component, Inject} from '@angular/core';
 import {select, Store} from '@ngrx/store';
-import {combineLatest} from 'rxjs';
-import {distinctUntilChanged} from 'rxjs/operators';
+import {combineLatest, Subject} from 'rxjs';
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/reloader/reloader_component.ts
```diff
@@ -85,5 +88,7 @@ export class ReloaderComponent {
       'visibilitychange',
       this.onVisibilityChange
     );
+    this.ngUnsubscribe.next();
```
JamesHollyer, line 91: I see this is done in the other examples of this pattern. However, I do not understand why we need to call next before we call complete.

JamesHollyer (approved):
LGTM. The comment is only for my own understanding.

PR #5681, opened 2022-04-21
linked time: add deselect button

JamesHollyer:
* Motivation for features / changes
This changes adds the button for deselecting a fob. That button currently does not work(feature is currently behind a flag so a useless button is ok). Logic will be added later to remove the fob when clicked.

* Technical description of changes
These changes are all for style. I tried my best to make it look as I would expect a delete button to look. 

* Screenshots of UI changes
The button looks like this:
Screen Shot 2022-04-22 at 1 36 31 PM

When the mouse is hovering over the button it looks like this(screenshot did not show mouse):
Screen Shot 2022-04-22 at 1 36 36 PM

PR #5685, opened 2022-04-26
linked time: Implement fob deselect logic

JamesHollyer:
This is part of a chain of PRs([the first PR](https://github.com/tensorflow/tensorboard/pull/5681) in this change added the button) that is intended to implement a feature to deselect the fob for the linked time feature. This change adds the logic which decides what to do when the button is clicked. Since there could be 1 or 2 fobs at any point there needs to be logic that handles each situation.

When in range selection(which means we have a start and an end fob) deselecting a fob will leave the remaining fob in place. This means we switch to single selection. If the end fob is deselected we simply remove it. However, if the start fob is deselected we must change the end fob to the start fob before removing the end fob. This gives the effect that the start fob was remove. Lastly when in single selection deselecting the fob toggles the feature entirely. In the PR we only emit an event. The next PR in this change will wire up that event to the selectTimeEnableToggled action.

* Screenshots of UI changes
This is the button that was added in the first PR linked above. I am including here to help in understanding what I am trying to accomplish. 
Screen Shot 2022-04-22 at 1 36 36 PM

### DIFF
--- a/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
@@ -192,6 +192,15 @@ export class LinkedTimeFobControllerComponent {
     this.onSelectTimeChanged.emit(newLinkedTime);
   }
 
+  /**
+   * When in range selection(which means we have a start and an end
+   * fob) deselecting a fob will leave the remaining fob in place. This means we
+   * switch to single selection. If the end fob is deselected we simply remove
+   * it. However, if the start fob is deselected we must change the end fob to
+   * the start fob before removing the end fob. This gives the effect that the
+   * start fob was remove. Lastly when in single selection deselecting the fob
+   * toggles the feature entirely.
+   */
   deselectFob(fob: Fob) {
     if (fob === Fob.END) {
       this.onSelectTimeChanged.emit({...this.linkedTime, end: null});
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
```diff
@@ -189,4 +191,18 @@ export class LinkedTimeFobControllerComponent {
...
   }
+
+  deselectFob(fob: Fob) {
+    if (fob === Fob.END) {
+      this.onSelectTimeChanged.emit({...this.linkedTime, end: null});
+      return;
+    }
+
+    if (this.linkedTime.end !== null) {
+      this.onSelectTimeChanged.emit({start: this.linkedTime.end, end: null});
```
bmd3k, line 202: Could you transfer some of the helpful documents from the PR description to here? Either as function-level comment or just comments sprinkled throughout the function implementation:

> When in range selection(which means we have a start and an end fob) deselecting a fob will leave the remaining fob in place. This means we switch to single selection. If the end fob is deselected we simply remove it. However, if the start fob is deselected we must change the end fob to the start fob before removing the end fob. This gives the effect that the start fob was remove. Lastly when in single selection deselecting the fob toggles the feature entirely.

bmd3k (approved):

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
JamesHollyer: Added.

PR #5687, opened 2022-04-27
linked time: refactor fob step editing with contenteditable

japie1235813:
Previously we can edit the step within the fob with the help of adding an input element. The actions were double clicking the div and then click to focus the input. This PR refactor removes input and directly use contenteditable in span element, which allows user to click the number and edit it right away.

In `linked_time_fob_component.ng.html`, input element is removed; add `contenteditable ` attribute to span element. 
Since editing the content makes `{{ step | number}}` rendering buggy. (It will not render anything once it's cleared). We change it to set innerHtml once step input changed in `ngOnChanges`

  Naming changes: 
  removeStep --> fobRemoved; stepChange --> stepChanged

* Screenshots of UI changes
before

Screen Shot 2022-04-26 at 10 47 56 PM

after  

https://user-images.githubusercontent.com/1131010/166061959-ea21e4e0-5090-4c5d-98dd-7f83c569f5a8.mov

The actions include:
 * click the span to edit
 *  hit enter to change selected step
 *  handle non-number input, including space and shift+enter
 *  enter empty string on range selection changes it to single selection 

TODOs: 
* [FIXED] only show blue boarder when editing
* empty string on single selection sets start step to min step of the experiment

### DIFF
--- a/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
+++ b/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
@@ -15,16 +15,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div class="fob">
+<div class="fob" aria-label="Drag fob to change selected step" role="button">
   <span
     contenteditable
     #stepSpan
+    role="textbox"
+    aria-label="Edit step"
     (blur)="stepTyped($event)"
     (keypress)="validateStep($event)"
     (keydown.enter)="stepTyped($event)"
     (keydown.shift.enter)="$event.preventDefault()"
   ></span>
-  <button aria-label="deselect fob" (click)="fobRemoved.emit()">
+  <button aria-label="Deselect fob" (click)="fobRemoved.emit()">
     <mat-icon svgIcon="close_24px"></mat-icon>
   </button>
 </div>
--- a/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
+++ b/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
@@ -15,7 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div class="fob" aria-label="Drag fob to change selected step" role="button">
+<div class="fob" aria-label="Drag fob to change selected step">
   <span
     contenteditable
     #stepSpan
--- a/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
+++ b/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
@@ -15,7 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div class="fob" aria-label="Drag fob to change selected step">
+<div class="fob">
   <span
     contenteditable
     #stepSpan
--- a/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
+++ b/tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
@@ -15,20 +15,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-<div class="fob" (dblclick)="typeStepRequested()">
-  <span *ngIf="!isTyping"> {{ step | number }} </span>
-  <button
-    aria-label="deselect fob"
-    *ngIf="!isTyping"
-    (click)="removeStep.emit()"
-  >
+<div class="fob">
+  <span
+    contenteditable
+    #stepSpan
+    (blur)="stepTyped($event)"
+    (keypress)="validateStep($event)"
+    (keydown.enter)="stepTyped($event)"
+    (keydown.shift.enter)="$event.preventDefault()"
+  ></span>
+  <button aria-label="deselect fob" (click)="fobRemoved.emit()">
     <mat-icon svgIcon="close_24px"></mat-icon>
   </button>
-  <input
-    *ngIf="isTyping"
-    type="number"
-    [value]="step"
-    (change)="stepTyped($event)"
-    (focusout)="lostFocus()"
-  />
 </div>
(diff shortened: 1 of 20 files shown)


tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
```diff
@@ -15,20 +15,16 @@
...
-<div class="fob" (dblclick)="typeStepRequested()">
-  <span *ngIf="!isTyping"> {{ step | number }} </span>
-  <button
-    aria-label="deselect fob"
-    *ngIf="!isTyping"
-    (click)="removeStep.emit()"
-  >
+<div class="fob">
+  <span
+    contenteditable
```
bmd3k, line 20: Why contenteditable instead of just an input with type="number"?

A couple reasons I ask:
1. My sense from reading about contenteditable is that it is probably overkill for this job. It's meant for richer text editing. We just want a numeric input. (you wouldn't need to write your own input validator, for example). I get the sense that if you can just use an input then you should just use an input.
2. From an accessibility/markup perspective, input elements are more recommended if you can: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role. If you use contenteditable then you have to remember to put roll="textbox".

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
```diff
@@ -15,20 +15,16 @@
...
 -->
 
-<div class="fob" (dblclick)="typeStepRequested()">
-  <span *ngIf="!isTyping"> {{ step | number }} </span>
-  <button
-    aria-label="deselect fob"
-    *ngIf="!isTyping"
-    (click)="removeStep.emit()"
-  >
+<div class="fob">
```
bmd3k, line 18: Since we've broached the topic of accessibility/markup, probably worth noting again that the fob is not properly marked up as an interactable element. My browser doesn't give me the nice cursor indicating I can drag it. And accessibility readers will have no idea what this is supposed to be.

bmd3k (commented):
Overall the change in behavior is nice but I don't know if contenteditable is the right tool for the job.

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
japie1235813: 1. In current implementation, we need `<input>` only when it's in "editing mode". To render the html elements(.fob, input, and X button), we need to keep a variable ("isTyping") to decide what elements to render and keeps the variable updated on user actions. With contenteditable, all of these can be removed.

2. To make it look better, we want to have the length of input "fit" nicely in the fob. The length is set by css and we need to update it with js  on key binding and do some pixel calculation; with contenteditable the span element is adjusted automatically and real-time to fit the length of digits

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
japie1235813: Added cursor pointer and aria-lable. Please let me know if anything is missing. Other things I'm looking at is "aria-grabbed"  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-grabbed 
but that requires event handler and a bit out of scope for this PR.

bmd3k: I had to think about this some more. I put some thoughts in https://docs.google.com/document/d/1VGY17H9UkF1pAUzhUuVnfALebFEepKbs-QwFYMtjbh0/edit#heading=h.c4dvmcp8ltnq (Sorry, Google-internal only).

I think we should do the simpler suggestion which requires that we decide the fob is not part of a slider:

1. Make the input tabable/focusable. Allow user to enter text. Mark with aria attributes (role, aria-label might be sufficient?). In which case I think you covered this.

2. The fob is not tabable/focusable. It reacts to changes to the input. It should not be marked with aria attributes. It should use `cursor: grab` and `cursor: grabbing` depending on the state.

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
JamesHollyer: Another issue with this being contentEditable is that the cursor shows up when you are dragging. The only solution to this that I see is that we would need to dynamically make it contentEditable which means we would have the isTyping variable or something similar back in.

I do think this looks WAY better than the input box but maybe the best fix there is some css clean up? Just thinking out loud here. This solution does seem cleaner in a lot of ways.

japie1235813:
> Usually it tries to do both
yes that's what I see now. How the element becomes editable is on click "down". so when mouse does the dragging (aks click down and move) It also becomes editable. I was thinking about changing that (maybe change it to click "up") later. The fob looks fine to me when dragging or actually that's terrible, I'm just too used to it, I can figure this out in this PR

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_controller_component.ts
```diff
@@ -170,7 +170,19 @@ export class LinkedTimeFobControllerComponent {
           this.axisOverlay.nativeElement.getBoundingClientRect().left;
   }
 
-  stepTyped(fob: Fob, step: number) {
+  stepTyped(fob: Fob, step: number | null) {
+    // Types empty string in fob.
+    if (step === null) {
+      // Removes fob on range selection and sets step to minimum on single selection.
+      if (this.linkedTime.end !== null) {
```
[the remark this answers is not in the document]
bmd3k: My opinion: It seems unlikely to me that a user would want this action to set the step to minimum step.

I leave it up to you, James, and Jesse to think about it more but I think one of the following results makes the most sense:

1) User wants to remove the fob.
2) It's a mistake. Don't allow the input to be saved and just keep the fob where it is.

bmd3k (approved):
Can you fix the PR title: contentediable => contenteditable.

tensorboard/webapp/widgets/linked_time_fob/linked_time_fob_component.ng.html
japie1235813: For now I want to go with (2) as the temporary solution. The downside I can think about is the text clicking and editing is not obvious at all. But from the study it seems unlikely users will edit the fob directly so I leave it for now.

japie1235813 (commented):
Somehow it was overlooked. Updated the PR title.

PR #5689, opened 2022-04-28
linked time: Wire up onSelectTimeEnableToggle action for deselecting fob

JamesHollyer:
This PR finishes the deselect fob feature. When in single selection(only 1 fob) deselecting the fob toggles the entire selected time. The logic for when this happens was added in [my previous pr](https://github.com/tensorflow/tensorboard/pull/5685). However it emitted an event that was not linked to anything. This PR gets the correct action from the ScalarCardContainer and the HistogramCardContainer and passes it as an event all the way down to the LinkedTimeFobControllerComponent.

This change is pretty straight forward. The tests however had to be done differently for the ScalarCard and the HistogramCard. In the ScalarCardTest file we are able to do a full integration test all the way down to the LinkedTimeFobComponent. However, the HistogramCardTest uses a dummy HistogramComponent so we cannot test it with elements lower in the stack. For this test I simply ensured correct action is fired when the dummy component fires the given action and then tested that the actual HistogramComponent fired that same event when the LinkedTimeFobComponent fires it's removeStep event.

* Screenshots of UI changes
This UI change was added in [this PR](https://github.com/tensorflow/tensorboard/pull/5681).
However, I will add the screenshots here to help understand the feature this is helping to add.
Screen Shot 2022-04-22 at 1 36 36 PM

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -2177,14 +2177,16 @@ describe('scalar card', () => {
         ).toEqual('30');
       }));
     });
+
     describe('fob controls', () => {
-      it('dispatches timeSelectionChanged action when fob is dragged', fakeAsync(() => {
+      let dispatchedActions: Action[] = [];
+      beforeEach(() => {
+        dispatchedActions = [];
         const runToSeries = {
           run1: [buildScalarStepData({step: 10})],
           run2: [buildScalarStepData({step: 20})],
           run3: [buildScalarStepData({step: 30})],
         };
-        const dispatchedActions: Action[] = [];
         spyOn(store, 'dispatch').and.callFake((action: Action) => {
           dispatchedActions.push(action);
         });
@@ -2195,6 +2197,9 @@ describe('scalar card', () => {
           null /* metadataOverride */,
           runToSeries
         );
+      });
+
+      it('dispatches timeSelectionChanged action when fob is dragged', fakeAsync(() => {
         store.overrideSelector(getMetricsSelectedTime, {
           start: {step: 20},
           end: null,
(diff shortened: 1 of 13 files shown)


tensorboard/webapp/widgets/histogram/histogram_test.ts
```diff
@@ -1198,5 +1209,32 @@ describe('histogram test', () => {
...
+          end: null,
+        };
+        fixture.componentInstance.onSelectTimeToggle = onSelectTimeToggleSpy;
+        fixture.detectChanges();
+        intersectionObserver.simulateVisibilityChange(fixture, true);
+
+        const fobComponent = fixture.debugElement.query(
+          By.directive(LinkedTimeFobComponent)
+        ).componentInstance;
+        fobComponent.removeStep.emit();
```
japie1235813, line 1234: "removeStep" is renamed to "fobRemoved" in #5687 Could you wait a bit and rebase this PR?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2222,6 +2223,37 @@ describe('scalar card', () => {
           }),
         ]);
       }));
+
+      it('toggles selected time when single fob is deselected', fakeAsync(() => {
```
japie1235813, line 2227: Should we also add a test for range selection, which emits `timeSelectionChanged`?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2222,6 +2223,37 @@ describe('scalar card', () => {
...
+        spyOn(store, 'dispatch').and.callFake((action: Action) => {
+          dispatchedActions.push(action);
+        });
+        provideMockCardRunToSeriesData(
+          selectSpy,
+          PluginType.SCALARS,
+          'card1',
+          null /* metadataOverride */,
+          runToSeries
+        );
```
japie1235813, line 2243: Feel like line 2228-2243 is the same for every test in this "describe" and is going to be the same for future tests in this block, considering move them to beforeEach?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2176,7 +2176,8 @@ describe('scalar card', () => {
           fobs[0].query(By.css('span')).nativeElement.textContent.trim()
         ).toEqual('30');
       }));
-
+    });
+    describe('fob controls', () => {
```
japie1235813, line 2180: Thanks for separating the tests. nit: a newline beforehand?

japie1235813 (commented):
Overall wiring lgtm. Just a couple of details. And could you wait to rebase #5678? Which has naming change and it will cause conflict.

japie1235813 (approved):

PR #5692, opened 2022-05-02
Disable IfChange / ThenChange Lint clause from vended protos

bileschi:
These lint clauses make sense in the original repo these files are vended from.  They are not useful here as the linked file does not exist in our repo, and we should not be manually changing these files anyway.

The change is innocuous in open source but the lint clauses trigger warning within Google's internal CI.

A subsequent change to `tensorboard/compat/proto/update.sh` will disable these lints automatically to prevent their reintroduction.

### DIFF
--- a/tensorboard/compat/proto/op_def.proto
+++ b/tensorboard/compat/proto/op_def.proto
@@ -15,7 +15,7 @@ option go_package = "github.com/tensorflow/tensorflow/tensorflow/go/core/framewo
 
 // Defines an operation. A NodeDef in a GraphDef specifies an Op by
 // using the "op" field which should match the name of a OpDef.
-// LINT.IfChange
+// DISABLED.IfChange
 message OpDef {
   // Op names starting with an underscore are reserved for internal use.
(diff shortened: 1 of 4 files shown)


JamesHollyer (approved):

tensorboard/compat/proto/op_def.proto
```diff
@@ -175,7 +174,6 @@ message OpDef {
   // trigger TF network failure handling logics.
   bool is_distributed_communication = 21;
 }
-// LINT.ThenChange(
 //     https://www.tensorflow.org/code/tensorboard/compat/proto/op_def_util.cc)
```
nfelt, line 177: Kind of a nit, but for cases like this, this leaves behind a somewhat confusing fragment with the latter half of the LINT directive. Would be a bit nicer if we either were to strip the entire thing, or leave some explanatory replacement like `// Disabled ThenChange block: (`

JamesHollyer:
I wrote the script to make these updates automatically. I think it makes sense to just merge these changes in the same PR.

https://github.com/tensorflow/tensorboard/pull/5693

PR #5693, opened 2022-05-03
Remove LINT check from copied proto files

JamesHollyer:
These lint checks are needed to ensure the Tensorflow proto files stay in sync with each other. However, in Tensorboard we do not make changes to these files so they do not need to stay synced. Furthermore the checks are looking for files that we do not copy over to Tensorboard so they actually fail.

This change updates the script we use to copy the proto files so that we disable the lint checks automatically every time we copy them. I also ran the script as done in [this PR](https://github.com/tensorflow/tensorboard/pull/5686) which was our latest update to these proto files for the 2.9 release.


tensorboard/compat/proto/update.sh
```diff
@@ -39,8 +39,10 @@ find tensorboard/compat/proto/ -type f  -name '*.proto' -exec perl -pi \
   -e 's|tensorflow\.DataType|tensorboard.DataType|g;' \
   -e 's|tensorflow\.TensorProto|tensorboard.TensorProto|g;' \
   -e 's|tensorflow\.TensorShapeProto|tensorboard.TensorShapeProto|g;' \
+  -e 's|LINT\.|DISABLED.|g;' \
```
nfelt, line 42: Could we narrow this matching a little bit, for example like this?

```
s|^// LINT\.|// DISABLED.|g;
```

That way it will only match `LINT.` if it's at the start of the line and preceded by `// `.

nfelt (approved):

PR #5699, opened 2022-05-09
docs: clarify the default number of histogram buckets in summary API doc

yatbear:
Currently it is not mentioned anywhere in the API documentation that the number of histogram buckets defaults to 30 if not set: https://www.tensorflow.org/api_docs/python/tf/summary/histogram

#tf_summary_v2 #histogram


nfelt (approved):
Nice, thanks!

PR #5705, opened 2022-05-12
linked time: Remove axis-overlay from linked_time_fob_controller.

bmd3k:
Remove "axis-overlay" from linked_time_fob_controller and make scalars and histograms both responsible for properly positioning the linked_time_fob_controller in the page layout.

The histograms card was relying on axis-overlay to properly position the linked_time_fob_controller rather than doing itself. Scalars card, however, was already properly positioning the linked_time_fob_controller itself. Since axis-overlay was becomign a small maintenance burden we just remove it and change histograms to behave more like scalars in this response.

The changes have no practical impact on the feature but it does require some adjustments in test.

Googlers, see cl/448303779 for results of internal sync.


japie1235813 (approved):
Thanks for helping with this cleanup. Patched and run it. Looks pretty good

PR #5715, opened 2022-05-19
tensorboard-data-server: add Apple M1 support

### LINKED ISSUE #5713: tensorboard-data-server: build issues on Apple M1 (arm64)
### Diagnostics

<details>
<summary>Diagnostics output</summary>

``````
--- check: autoidentify
INFO: diagnose_tensorboard.py version e43767ef2b648d0d5d57c00f38ccbd38390e38da

--- check: general
INFO: sys.version_info: sys.version_info(major=3, minor=8, micro=9, releaselevel='final', serial=0)
INFO: os.name: posix
INFO: os.uname(): posix.uname_result(sysname='Darwin', nodename='ajstewart-mbp', release='21.4.0', version='Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000', machine='arm64')
INFO: sys.getwindowsversion(): N/A

--- check: package_management
INFO: has conda-meta: False
INFO: $VIRTUAL_ENV: None

--- check: installed_packages
WARNING: no installation among: ['tb-nightly', 'tensorboard', 'tensorflow-tensorboard']
WARNING: no installation among: ['tensorflow', 'tensorflow-gpu', 'tf-nightly', 'tf-nightly-2.0-preview', 'tf-nightly-gpu', 'tf-nightly-gpu-2.0-preview']
WARNING: no installation among: ['tensorflow-estimator', 'tensorflow-estimator-2.0-preview', 'tf-estimator-nightly']

--- check: tensorboard_python_version
  File "diagnose_tensorboard.py", line 522, in main
    suggestions.extend(check())
  File "diagnose_tensorboard.py", line 75, in wrapper
    result = fn()
  File "diagnose_tensorboard.py", line 271, in tensorboard_python_version
    from tensorboard import version
ModuleNotFoundError: No module named 'tensorboard'

--- check: tensorflow_python_version
  File "diagnose_tensorboard.py", line 522, in main
    suggestions.extend(check())
  File "diagnose_tensorboard.py", line 75, in wrapper
    result = fn()
  File "diagnose_tensorboard.py", line 278, in tensorflow_python_version
    import tensorflow as tf
ModuleNotFoundError: No module named 'tensorflow'

--- check: tensorboard_data_server_version
INFO: no data server installed

--- check: tensorboard_binary_path
INFO: which tensorboard: None

--- check: addrinfos
socket.has_ipv6 = True
socket.AF_UNSPEC = <AddressFamily.AF_UNSPEC: 0>
socket.SOCK_STREAM = <SocketKind.SOCK_STREAM: 1>
socket.AI_ADDRCONFIG = <AddressInfo.AI_ADDRCONFIG: 1024>
socket.AI_PASSIVE = <AddressInfo.AI_PASSIVE: 1>
Loopback flags: <AddressInfo.AI_ADDRCONFIG: 1024>
Loopback infos: [(<AddressFamily.AF_INET6: 30>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('::1', 0, 0, 0)), (<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('127.0.0.1', 0))]
Wildcard flags: <AddressInfo.AI_PASSIVE: 1>
Wildcard infos: [(<AddressFamily.AF_INET6: 30>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('::', 0, 0, 0)), (<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('0.0.0.0', 0))]

--- check: readable_fqdn
INFO: socket.getfqdn(): 'ajstewart-mbp.dhcp.thefacebook.com'

--- check: stat_tensorboardinfo
INFO: directory: /var/folders/j1/68dlgpr91vlgs26vty2c8xk80000gn/T/.tensorboard-info
INFO: .tensorboard-info directory does not exist

--- check: source_trees_without_genfiles
INFO: tensorboard_roots (0): []; bad_roots (0): []

--- check: full_pip_freeze
INFO: pip freeze --all:
pip==20.2.3
setuptools==49.2.1
six @ file:///AppleInternal/BuildRoot/Library/Caches/com.apple.xbs/Sources/python3/python3-110/six-1.15.0-py2.py3-none-any.whl
wheel @ file:///AppleInternal/BuildRoot/Library/Caches/com.apple.xbs/Sources/python3/python3-110/wheel-0.36.2-py2.py3-none-any.whl

``````

</details>

## Steps to reproduce (required)

I'm trying to build tensorboard-data-server from source using the [Spack](https://spack.io) package manager. The build steps basically involve:
```console
$ cd tensorboard/data/server
$ cargo build --release
$ cd pip_package
$ python build.py --out-dir=... --server-binary=...
```
It seems like `build.py` is hardcoded to use x86_64 even when on arm64? The following build log illustrates this issue, it builds for arm64 but then stuffs it into an x86_64 wheel: [spack-build-out.txt](https://github.com/tensorflow/tensorboard/files/8720997/spack-build-out.txt). Is there a better approach to build a wheel than using `pip_package/build.py`?

@glennpj

adamjstewart:

Fixes #5713.

The `cpu_name` should match the hardware you are building on, it shouldn't be hardcoded to `x86_64`. Also, the macOS version depends on the hardware as well. `arm64` seems to require `macosx_11_0` or higher. See https://pypi.org/project/numpy/#files for an example of this.

* Screenshots of UI changes

N/A

Successfully built and installed a wheel on macOS 12.3.1 and Apple M1 (arm64).

An alternative would be to add flags to manually specify these things. If you want me to do that instead let me know, but that seemed like more work since we would need to change it in multiple places.


adamjstewart:
@japie1235813

tensorboard/data/server/pip_package/build.py
```diff
@@ -75,16 +75,19 @@ def main():
     if args.universal:
         platform_tag = "any"
     else:
+        cpu_name = platform.machine()
+        mac_platform = "macosx_10_9"
+        if cpu_name == "arm64":
```
japie1235813, line 80: Instead of checking this cpu automatically, could you add a flag to switch the cpu_name and platform? We did not have intent for the binary to work on arm64, it just happens to. With the flag passing in, it's pretty much at your own control on the command running level to do so.

adamjstewart: My only hesitation with doing this is that the binary you build likely only works on the platform/cpu you build it on. For instance, if someone is running on Linux x86_64, changing the wheel name to "arm64_maxosx_11_0" likely isn't going to produce a wheel that will work on M1, and vice versa. So it may seem misleading if the user can choose the platform/cpu themselves.

But if you still want me to do this, I'm happy to do it. Or if I'm misunderstanding something and these settings somehow control what gets built, let me know.

adamjstewart: Hmm, but if that were the case, then why would we be passing `-p <platform_tag` to `python setup.py bdist_wheel`? Maybe it does allow for cross compiling? Or maybe that only controls how old the macOS release version that gets supported is.

Before this PR, platform was dynamically chosen by which platform you're running on. Are you suggesting to make that a flag too?

adamjstewart: Out of curiosity, why does the package have cpu/platform-specific wheels AND universal wheels on PyPI? Does the package actually depend on the OS or is it pure-Python? If I'm only trying to build for myself, can I just build a universal wheel to avoid the issues I reported in #5713?

adamjstewart: Looks like the only differences between the wheels is whether or not they come with a `tensorboard_data_server/bin/server` binary.

adamjstewart: @japie1235813 any comments on the above?

japie1235813: We might want to cross-compile building the binary (e.g. building arm64 binary on x86 env). We don't want to have build.py make the assumption of cpu/os of where the binary comes from. build.py also does not check it and has no control of how the binary is built. Adding a flag provides the needs of not assuming anything and makes it clear that it's on the user's will to build an arm64 wheel.

adamjstewart: Okay, so if I add those flags and run `python build.py --platform=macosx_11_0 --cpu=arm64` on a Linux x86_64 machine, it will produce a wheel that works on macOS arm64, not a wheel that works on Linux x86_64? I'm still confused as to whether or not `build.py` actually supports cross-compiling. If it does, then great, I'll add those flags. If it doesn't, then adding flags is useless since you can only build a wheel for your current system.

adamjstewart: @japie1235813 did you ever hear back from the rest of the team? If `build.py` doesn't support cross-compiling, then there's no reason to add a flag that lets users shoot themselves in the foot. If you specify a different `--platform` and `--cpu` than the one you are running on, `build.py` will happily build a wheel for that platform/cpu, but it won't run on that platform/cpu.

japie1235813: The `build.py` checks if the env is supported (in which case officially we only support x86) but does not check anything about `server_binary` and have no idea if the `server_binary` is compatible with the supported cup/os.
In current scenario (hardcode x86) on your machine, it builds a wheel claiming supporting the x86 but actually only supports arm64. 

What you are suggesting will work if you're building an arm64 binary on arm64 machine, but still does not support cross-compilation. Technically you can bazel build arm64 Rustboard binary on x86 machine. Then in the build.py, adding the flag to specify the cpu that the binary is built for creates a wheel could be run on arm64.

adamjstewart:
Okay, latest version adds cross-compilation support. The argparse isn't perfect, ideally I think we would want `--platform/--cpu/--server-binary` to be mutually exclusive with `--universal`, but there isn't an easy way to do that in argparse: https://stackoverflow.com/questions/17909294/python-argparse-mutual-exclusive-group

adamjstewart:
Confirmed that the current commit builds properly on my system.

adamjstewart:
@japie1235813 is there anything else this PR still needs?

adamjstewart:
Ping @japie1235813

tensorboard/data/server/pip_package/build.py
```diff
@@ -35,12 +32,32 @@
...
 
 def main():
+    cpu_name = platform.machine()
+    mac_platform = "macosx_10_9"
+    if cpu_name == "arm64":
+        mac_platform = "macosx_11_0"
+    platform_name = {
+        # using platform tag values from TensorFlow releases
+        "Linux": "manylinux2010",
+        "Darwin": mac_platform
```
japie1235813, line 42: add comma at the end for lint.
`"Darwin": mac_platform,`

japie1235813 (approved):
Sorry for the late reply. Thanks for this contribution! LGTM

wookayin:
Is this has been released as a part of pre-built wheel? Neither tensorboard==2.10.0 nor nightly (2.11) seem to have a rustboard binary shipped in the wheel.

```
Option --load_fast=true not available: TensorBoard data server not supported on this platform.  
```

In the CI jobs https://github.com/tensorflow/tensorboard/actions/runs/3397568456/jobs/5649844791 have the macos-universal matrix disabled (running `native` only). Can we have it enabled and make TB's officially-released wheel for macos-arm64 support tensorboard-data-server? It'd be great if it can be included in the 2.11 release.

PR #5717, opened 2022-05-20
Stop removing experiment related query parameters

rileyajones:
Navigating to TensorBoard used to result in all provided query parameters where being removed when the application state was persisted to the query parameters. This would lead to refreshed of the page not having the experiment value set correctly.

Unfortunately there is neither a one to one mapping of query parameter overrides to experiment nor do the overrides exactly reflect the experiment they are overriding. i.e. `enabledLinkTime` is overridden by the parameter `enableLinkTime`. To resolve this issue I manually created a mapping experiment to their query parameter overrides which I then checked when the application state is persisted to the query parameters.

* Screenshots of UI changes
No UI change

Visit the dashboard with an experiment overridden via a query parameter and note that it is not removed.

Note that prior to this `enableColorGroupByRegex` and `enableColorGroup` where both explicitly handled in a similar way.

I considered two other approaches

1) The manual mapping could be avoided by changing the query parameter overrides (or else adding additional parameters to overrides). This could result in unexpected behavior for existing users unaware of the change but would be easier to maintain as more experiments are added.

2) Instead of replacing the all the query parameters, we could just update or remove those that had changed. This would be a bigger change to the application though and would risk introducing other issues.


bmd3k (commented):
For posterity, some of the things we discussed offline:
* There is a more general need to associate metadata with individual feature flags. This is a good opportunity to build the foundation for that.
* Riley suggested some changes to FeatureFlags type in general and will play around with that.
* There is the consideration that specific TensorBoard instances (like TensorBoard.dev, TensorBoard.corp) may extend the FeatureFlags type to include additional feature flags.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -40,62 +34,51 @@ export class QueryParamsFeatureFlagDataSource
...
-    if (params.has(ENABLE_LINK_TIME_PARAM_KEY)) {
-      featureFlags.enabledLinkedTime =
-        params.get(ENABLE_LINK_TIME_PARAM_KEY) !== 'false';
-    }
-
-    if (params.has(ENABLE_CARD_WIDTH_SETTING_PARAM_KEY)) {
-      featureFlags.enabledCardWidthSetting =
-        params.get(ENABLE_CARD_WIDTH_SETTING_PARAM_KEY) !== 'false';
-    }
+    // DO_NOT_SUBMIT check if enableDarkModeOverride performs correctly when null is set
```
rileyajones, line 43: Blocking based on this

tensorboard/webapp/routes/feature_flag_serializer.ts
```diff
@@ -0,0 +1,57 @@
...
+import {map} from 'rxjs/operators';
+
+import {Location} from '../app_routing/location';
+import {SerializableQueryParams} from '../app_routing/types';
+import {State} from '../app_state';
+import * as selectors from '../selectors';
+import {EXPERIMENTAL_PLUGIN_QUERY_PARAM_KEY} from '../webapp_data_source/tb_feature_flag_data_source_types';
+import {FeatureFlagMetadata, FeatureFlagQueryParameters,} from '../webapp_data_source/tb_feature_flag_query_parameters';
+
+export function getFeatureFlagStates(store: Store<State>):
```
rileyajones, line 26: Should this be a selector?

tensorboard/webapp/routes/feature_flag_serializer.ts
```diff
@@ -0,0 +1,57 @@
...
+    Observable<SerializableQueryParams> {
+  return store.select(selectors.getEnabledExperimentalPlugins)
+      .pipe(map((experimentalPlugins) => {
+        const queryParams = experimentalPlugins.map((pluginId) => {
+          return {key: EXPERIMENTAL_PLUGIN_QUERY_PARAM_KEY, value: pluginId};
+        });
```
rileyajones, line 32: Is this necessary

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -40,62 +35,39 @@ export class QueryParamsFeatureFlagDataSource
...
+    if (!queryParamOverride || !params.has(queryParamOverride)) {
+      return null;
     }
-
-    if (params.has(ENABLE_DATA_TABLE_PARAM_KEY)) {
-      featureFlags.enabledScalarDataTable =
-        params.get(ENABLE_DATA_TABLE_PARAM_KEY) !== 'false';
+    const paramValues: BaseFeatureFlagType[] = this.queryParams
+      .getParams()
+      .getAll(queryParamOverride)
```
rileyajones, line 65: If a user has more than one instance of a query param it will always be parsed as an array. This can lead to broken states. Consider adding another optional property to the `FeatureFlagMetadata`

tensorboard/webapp/webapp_data_source/tb_feature_flag_query_parameters.ts
```diff
@@ -0,0 +1,117 @@
...
+  ENABLE_COLOR_GROUP_QUERY_PARAM_KEY,
+  ENABLE_DARK_MODE_QUERY_PARAM_KEY,
+  ENABLE_DATA_TABLE_PARAM_KEY,
+  ENABLE_LINKED_TIME_PARAM_KEY,
+  EXPERIMENTAL_PLUGIN_QUERY_PARAM_KEY,
+  FORCE_SVG_RENDERER,
+  SCALARS_BATCH_SIZE_PARAM_KEY,
+} from './tb_feature_flag_data_source_types';
+
+export type BaseFeatureFlagType = Boolean | Number | String | null | undefined;
```
bmd3k, line 28: Instead of BaseFeatureFlagType and FeatureFlagType, my instinct is to instead make FeatureFlagMetadata a generic type:

export type FeatureFlagMetadata<T> = {
  displayName: string;
  queryParamOverride?: string;
  parseValue: (str: string) => T;
};

And then sort of adjust things from there. I've done this here: 

https://github.com/bmd3k/tensorboard/commit/c1060ca992afee047628c118a22d85d6ff11a9d2

It compiles and runs but it is incomplete -- it doesn't handle the experimentalPlugins case and and there are some parts of the code that we have to cast things to, sort of, bypass the compiler. I think it is more readable, though and still largely gives good compiler coverage.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -40,62 +35,39 @@ export class QueryParamsFeatureFlagDataSource
...
-        params.get(ENABLE_LINKED_TIME_PARAM_KEY) !== 'false';
-    }
-
-    if (params.has(ENABLE_CARD_WIDTH_SETTING_PARAM_KEY)) {
-      featureFlags.enabledCardWidthSetting =
-        params.get(ENABLE_CARD_WIDTH_SETTING_PARAM_KEY) !== 'false';
-    }
+    const featureFlags: Partial<Record<keyof FeatureFlags, FeatureFlagType>> =
+      enableMediaQuery ? this.getPartialFeaturesFromMediaQuery() : {};
+    Object.entries(FeatureFlagQueryParameters).forEach(
```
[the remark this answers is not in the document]
rileyajones: Yes, I plan change `FeatureFlagQueryParameters` to be passed in as a parameter from the `serializeStateToQueryParams` method of `DashboardDeepLinkProvider`. This will then be overridden by the relevant DeepLinkProviders for the other flavors of TensorBoard.

tensorboard/webapp/webapp_data_source/tb_feature_flag_query_parameters.ts
```diff
@@ -0,0 +1,117 @@
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
```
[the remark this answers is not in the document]
rileyajones: That's a good idea! I'll go ahead with a move and rename.

tensorboard/webapp/webapp_data_source/tb_feature_flag_query_parameters.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+export function parseBooleanOrNull(str: string): Boolean | null {
+  if (str === 'null') {
+    return null;
+  }
+  return parseBoolean(str);
+}
+
+export const FeatureFlagQueryParameters: {
```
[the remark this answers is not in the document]
rileyajones: Yeah, that does make sense, will do.

tensorboard/webapp/webapp_data_source/tb_feature_flag_query_parameters.ts
rileyajones: I originally had done this, then someplace in the middle of a refactor couldn't get it to work. Thanks for bringing this back up.

tensorboard/webapp/routes/feature_flag_serializer.ts
rileyajones: This dependency on the store can actually just be removed after refactoring the `experimentalPlugins` logic!

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
rileyajones: Yeah... this is easy enough to fix that it seems worth fixing

tensorboard/webapp/routes/feature_flag_serializer.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        Object.values(FeatureFlagQueryParameters)
+            .forEach((overriddenFeatureFlag: FeatureFlagMetadata) => {
+              const queryParamOverride =
+                  overriddenFeatureFlag.queryParamOverride;
+              if (queryParamOverride &&
+                  queryParamOverride in currentQueryParams) {
+                queryParams.push({
+                  key: queryParamOverride,
+                  value: currentQueryParams[queryParamOverride],
```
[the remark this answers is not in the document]
rileyajones: Simply making the URL a reflection of the feature flag state would be easier to do and thus would simplify the code. I was previously under the impression that reflecting `localStorage` was considered undesirable.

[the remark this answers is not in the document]
bmd3k: Having the URL reflect the localStorage overrides would be a good reminder to the user that they are in some special state.

It will help us quickly troubleshoot problems like "why can't I see feature A after feature A has launched?" when user forgot they set enableFeatureA=false in local storage.  I am certain this will happen all the time.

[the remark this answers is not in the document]
rileyajones: This makes lots of sense to me. That also ensures that sharing a link will result in the other person seeing the same thing. I've gone ahead and made this change and will be pushing a new commit quite soon.

tensorboard/webapp/app_routing/location.ts
```diff
@@ -57,7 +60,7 @@ export class Location implements LocationInterface {
   }
 
   getSearch(): SerializableQueryParams {
-    const searchParams = new URLSearchParams(window.location.search);
+    const searchParams = new URLSearchParams(utils.getSearch());
```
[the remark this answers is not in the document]
rileyajones: Originally I had changed this to for testing.

tensorboard/webapp/routes/feature_flag_serializer.ts
```diff
@@ -0,0 +1,41 @@
...
+        : [featureValue];
+      return values.map((value) => ({
+        key,
+        value: value?.toString(),
+      }));
+    })
+    .flat()
+    .filter(
+      ({key, value}) => key && value !== undefined
+    ) as SerializableQueryParams;
```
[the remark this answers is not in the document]
rileyajones: For some reason TypeScript wouldn't believe that `key` and `value` were always defined even after the filter.

tensorboard/webapp/routes/dashboard_deeplink_provider_test.ts
```diff
@@ -378,9 +378,9 @@ describe('core deeplink provider', () => {
       });
```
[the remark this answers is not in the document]
rileyajones: I'm assuming you just mean a code comment? I could also imagine you implying adding another `describe` here though. I'm adding a comment for now, lmk if you'd prefer something else.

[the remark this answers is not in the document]
bmd3k: Yup, just meant a comment!

tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
```diff
@@ -0,0 +1,122 @@
...
+} from '../../webapp_data_source/tb_feature_flag_data_source_types';
+import {FeatureFlags} from '../types';
+
+export type BaseFeatureFlagType = boolean | number | string | null | undefined;
+
+export type FeatureFlagType = BaseFeatureFlagType | Array<BaseFeatureFlagType>;
+
+export type FeatureFlagMetadata<T> = {
+  displayName: string;
+  defaultValue?: T;
```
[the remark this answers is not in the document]
bmd3k: Yes, I'm fine to change the behavior of inColab. I did a quick survey of usage internally and it looks like usage is always "tensorboardColab=true" and not just "tensorBoardColab".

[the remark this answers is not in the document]
rileyajones: I'm going to update `inColab`. I've noticed that the url can get a bit cluttered if you start enabling lots of feature flags in the UI so I'm going to preempt the issue and only serialize feature flags with non default values.

tensorboard/webapp/feature_flag/store/feature_flag_store_config_provider.ts
```diff
@@ -14,26 +14,21 @@ limitations under the License.
...
+import {FeatureFlagMetadataMap} from './feature_flag_metadata';
 import {FeatureFlagState} from './feature_flag_types';
 
+const defaultFlags = Object.entries(FeatureFlagMetadataMap).reduce(
+  (map, [key, {defaultValue}]) => {
+    map[key] = defaultValue;
+    return map;
+  },
+  {} as Record<string, any>
+) as FeatureFlags;
```
rileyajones, line 27: This is a pretty gross cast... The issue is that `Object.entries` (and all the other `Object` methods) drops the typing. This cast seems preferable to casting the values but I am open to alternatives.

I COULD just enumerate the feature flags below but my current approach seems to scale better as more feature flags  are added.

tensorboard/webapp/routes/dashboard_deeplink_provider_test.ts
```diff
@@ -369,9 +371,9 @@ describe('core deeplink provider', () => {
...
       store.refreshState();
 
-      expect(queryParamsSerialized[queryParamsSerialized.length - 1]).toEqual([
-        {key: 'enableColorGroup', value: 'true'},
-      ]);
+      expect(queryParamsSerialized[queryParamsSerialized.length - 1]).toEqual(
+        []
```
rileyajones, line 375: `enableColorGroup` defaults to `true` so persisting it is redundant.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
```diff
@@ -215,7 +289,7 @@ describe('tb_feature_flag_data_source', () => {
         );
         expect(dataSource.getFeatures()).toEqual({
           enabledExperimentalPlugins: ['a'],
-          inColab: false,
+          inColab: true,
```
rileyajones, line 292: Calling out that this behavior IS changing

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -40,62 +34,39 @@ export class QueryParamsFeatureFlagDataSource
...
-        params.get(ENABLE_LINKED_TIME_PARAM_KEY) !== 'false';
-    }
-
-    if (params.has(ENABLE_CARD_WIDTH_SETTING_PARAM_KEY)) {
-      featureFlags.enabledCardWidthSetting =
-        params.get(ENABLE_CARD_WIDTH_SETTING_PARAM_KEY) !== 'false';
-    }
+    const featureFlags: Partial<Record<keyof FeatureFlags, FeatureFlagType>> =
+      enableMediaQuery ? this.getPartialFeaturesFromMediaQuery() : {};
+    Object.entries(FeatureFlagMetadataMap).forEach(
```
bmd3k, line 42: Can this also be refactored into something reusable? I think you were already intending on reusing getFeatureValue() but maybe we don't have to stop there. Especially given how this mangles all type information, would be nice to just have to write that code once.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -40,62 +34,39 @@ export class QueryParamsFeatureFlagDataSource
...
+      ([flagName, flagMetadata]) => {
+        const featureValue = this.getFeatureValue(
+          flagMetadata as FeatureFlagMetadata<boolean>
```
bmd3k, line 45: Probably doesn't matter since type information is mangled here: But it might be more honest to type as `FeatureFlagMetadata<any>`.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source.ts
```diff
@@ -40,62 +34,39 @@ export class QueryParamsFeatureFlagDataSource
...
-        params.get(ENABLE_DATA_TABLE_PARAM_KEY) !== 'false';
+    const paramValues: T[] = params.getAll(queryParamOverride).map((value) => {
+      return flagMetadata.parseValue(value);
+    });
+    if (!paramValues.length) {
+      return null;
     }
 
-    return featureFlags;
+    return flagMetadata.isArray ? paramValues.flat() : paramValues[0];
```
bmd3k, line 69: Could be worth describing what case you are trying to support with isArray and flat() here, similarly to how you documented the use case where you implemented the inverse functionality.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
bmd3k: Yes, that's fine. Thanks.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
```diff
@@ -84,123 +84,197 @@ describe('tb_feature_flag_data_source', () => {
...
-          new URLSearchParams('enableColorGroup=false'),
-          new URLSearchParams('enableColorGroup='),
-          new URLSearchParams('enableColorGroup=true'),
-          new URLSearchParams('enableColorGroup=foo')
-        );
-        expect(dataSource.getFeatures()).toEqual({enabledColorGroup: false});
-        expect(dataSource.getFeatures()).toEqual({enabledColorGroup: true});
-        expect(dataSource.getFeatures()).toEqual({enabledColorGroup: true});
-        expect(dataSource.getFeatures()).toEqual({enabledColorGroup: true});
+      describe('returns enableColorGroup from the query params', () => {
```
bmd3k, line 87: To consider for a future PR: Can we greatly reduce the number of tests here? We're effectively testing the same logic over and over again.

bmd3k (approved):

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
rileyajones: +1
When I refactored these I was thinking how how screamed to be parameterized.

tensorboard/webapp/feature_flag/store/feature_flag_store_config_provider.ts
```diff
@@ -14,26 +14,21 @@ limitations under the License.
 ==============================================================================*/
 import {InjectionToken} from '@angular/core';
 import {StoreConfig} from '@ngrx/store';
+import {FeatureFlags} from '../types';
+import {FeatureFlagMetadataMap} from './feature_flag_metadata';
 import {FeatureFlagState} from './feature_flag_types';
 
+const defaultFlags = Object.entries(FeatureFlagMetadataMap).reduce(
```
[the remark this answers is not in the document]
rileyajones: That's a great idea!

bmd3k (approved):
Thanks!

PR #5718, opened 2022-05-20
Adds mechanism to include feature flags in Angular HTTP requests.

bmd3k:
This adds a mechanism that dumps the client-side FeatureFlag object into each HTTP request that originates from the TB Angular code base.

(Googlers, see: http://go/tb-client-feature-flags-in-data-provider for the motivation and high level design.)

The data is dumpted into a custom header. The format is effectively:
`X-TensorBoard-Feature-Flags: JSON.stringify(currentFeatureFlags)`

We implement this by introducing an HttpInterceptor. It allows us to intercept each request in the Angular HTTP pipeline and inject the new header. The HttpInterceptor is provided in the existing FeatureFlagModule so all instances of TensorBoard will receive this behavior without any additional configuration.

Additional things worth noting:

* This will send ALL feature flags in the request header rather than a curated subset of them. In a subsequent PR, we'll add support to mark certain feature flags as "sendToServer" and only send that subset. This PR unblocks the main use case at the cost of there being some temporary overhead in HTTP requests.

* In a subsequent PR We will add similar functionality to the TB Polymer code base.

* In a subsequent PR we will add support on the TB HTTP Server to consume these feature flags and trigger logic based on the values.

* I moved feature_flags tests into their own "karma_test" target separate from the big top-level karma_test target. stephanwlee advocated for breaking up the top-level karma_test target in the months before he left the team.

### DIFF
--- a/tensorboard/webapp/feature_flag/http/feature_flag_http_interceptor.ts
+++ b/tensorboard/webapp/feature_flag/http/feature_flag_http_interceptor.ts
@@ -1,0 +1,58 @@
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {
+  HttpEvent,
+  HttpHandler,
+  HttpInterceptor,
+  HttpRequest,
+} from '@angular/common/http';
+import {Injectable} from '@angular/core';
+import {select, Store} from '@ngrx/store';
+import {Observable} from 'rxjs';
+import {first, switchMap} from 'rxjs/operators';
+import {getFeatureFlags} from '../store/feature_flag_selectors';
+import {State as FeatureFlagState} from '../store/feature_flag_types';
+
+export const FEATURE_FLAGS_HEADER_NAME = 'X-TensorBoard-Feature-Flags';
+
+/**
+ * HttpInterceptor for injecting feature flags into each HTTP request
+ * originating from the Angular TensorBoard code base.
+ */
+@Injectable({providedIn: 'root'})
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/feature_flag/http/feature_flag_http_interceptor.ts
```diff
@@ -0,0 +1,58 @@
...
+import {getFeatureFlags} from '../store/feature_flag_selectors';
+import {State as FeatureFlagState} from '../store/feature_flag_types';
+
+export const FEATURE_FLAGS_HEADER_NAME = 'X-TensorBoard-Feature-Flags';
+
+/**
+ * HttpInterceptor for injecting feature flags into each HTTP request
+ * originating from the Angular TensorBoard code base.
+ */
+@Injectable({providedIn: 'root'})
```
JamesHollyer, line 34: Why do we have providedIn: 'root' for this interceptor but the AuthHttpIntercepter just has @Injectable. TBH from my quick scan of the HttpInterceptor documentation this seems to make more sense but, I just don't understand the discrepancy.

tensorboard/webapp/feature_flag/http/feature_flag_http_interceptor.ts
```diff
@@ -0,0 +1,58 @@
...
+  intercept(
+    request: HttpRequest<unknown>,
+    next: HttpHandler
+  ): Observable<HttpEvent<unknown>> {
+    return this.store.pipe(
+      select(getFeatureFlags),
+      first(),
+      switchMap((featureFlags) => {
+        // Add feature flags to the headers.
+        request = request.clone({
```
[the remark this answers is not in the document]
bmd3k: The HttpRequest here is immutable. This isn't specified in the HttpInterceptor documentation but it is mentioned in various tutorials/introductions. I verified that just calling the following does not work and tests fail:

```
          request.headers.set(
            FEATURE_FLAGS_HEADER_NAME,
            JSON.stringify(featureFlags)
          )
```

tensorboard/webapp/feature_flag/http/feature_flag_http_interceptor.ts
bmd3k: Good catch. At one point I was trying to follow the HttpInterceptor's "Usage Notes", including some of the other suggestions in there. Using `providedIn: 'root'` was one of the things. Ulimately I abandoned those guidelines so I may as well remove `providedIn: 'root'`, too.

Future me may regret this decision but it hasn't been a problem, as far as we know, for AuthHttpInterceptor in tbdev.

JamesHollyer (approved):
LGTM

PR #5719, opened 2022-05-23
[FOR REFERENCE-DO NOT REVIEW]First Rough Draft of Data Table Feature

JamesHollyer:
* Motivation for features / changes
The motivation for this change is to help provide context to future PRs that implement each piece.

* Technical description of changes
This PR creates a data_table widget and uses it when linked time is enabled. The widget is dynamic in which headers it shows. Currently the headers are hard coded into the ScalarCardComponent. However, we plan on having those be changeable by the user soon. 

* Screenshots of UI changes
Screen Shot 2022-05-23 at 4 29 44 PM

* Detailed steps to verify changes work correctly (as executed by you)

* Alternate designs / implementations considered
Adding a ScalarCardDataTableContainer and component that sat beside the ScalarCardContainer was an option I tried. The amount of work that the ScalarCardContainer does to consolidate and format the data from the ngrx state was too much to recreate. I decided it was better to allow that container to continue maintaining that logic and passing that down to a dumber component which is the widget.

PR #5720, opened 2022-05-23
Add DataTable Basic structure

JamesHollyer:
This is the first change towards adding a dedicated table to hold the data for a selected time in the Scalar Card. This change simply adds the basic structure of the DataTable widget which will be a component which displays the data passed to it. For more context on where this is headed see https://github.com/tensorflow/tensorboard/pull/5719. 

This just adds the structure to stand up a widget which will be implemented in later CLs.

* Screenshots of UI changes
This PR makes no UI changes but we eventually plan to add a table that will probably look something like this:

The look of that table is still subject to change.

I considered adding a ScalarCardDataTableContainer and component that sat beside the ScalarCardContainer. However, I decided the amount of work that the ScalarCardContainer does to consolidate and format the data from the ngrx state was too much to copy. I decided it was better to allow the ScalarCardContainer to continue maintaining that logic and passing the consolidated data down to a dumber component which is the widget.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
@@ -168,7 +168,11 @@ limitations under the License.
     </table>
   </ng-template>
 </div>
-
+<ng-container *ngIf="selectedTime">
+  <div>
+    <tb-data-table></tb-data-table>
+  </div>
+</ng-container>
 <ng-template
   #lineChartCustomVis
   let-viewExtent="viewExtent"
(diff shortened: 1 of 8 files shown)


japie1235813 (approved):

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ng.html
```diff
@@ -168,7 +168,11 @@
     </table>
   </ng-template>
 </div>
-
+<ng-container *ngIf="selectedTime">
```
japie1235813, line 171: This is fine now but later we want to decouple the table and linked time. Might worth adding TODO or comment here

PR #5732, opened 2022-06-01
linked time: rename linked time fob to card fob

japie1235813:
We want to decouple fob from linked time to be reused it in "Step Selector", which is data table in scalar card and displaying the value of step selected by the fob. This is the first step of rename all linked-time related fob and components to be just fob. We now use fob on the cards (to be specific, scalar card and histogram card). Therefore rename it to be "card fob" to make it less generic.

TODO: move type "LinkedTime" out of `card_fob_types`


JamesHollyer (approved):
LGTM

PR #5733, opened 2022-06-02
step selector: gather data to be passed to data table

JamesHollyer:
Motivation for features / changes

This is one pr in a chain intended to add a data table feature. This change gathers the data needed for the table. For more context on where this is headed see https://github.com/tensorflow/tensorboard/pull/5719.

Technical description of changes
The ScalarCard now decides which columns we will have on the table(currently this is hard coded but, we plan to allow for the user to select which columns they want so this is built with that in mind). It also uses the selectedTime to decide which points of data will be displayed.  Eventually we will replace selected time with a more narrow step selection that is specific for that card. 

Screenshots of UI changes
This PR makes no UI changes but we event
ually plan to add a table that will probably look something like this:

The look of that table is still subject to change.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -32,6 +32,7 @@ import {
   siNumberFormatter,
 } from '../../../widgets/line_chart_v2/lib/formatter';
 import {LineChartComponent} from '../../../widgets/line_chart_v2/line_chart_component';
+import {findClosestIndex} from '../../../widgets/line_chart_v2/sub_view/line_chart_interactive_utils';
 import {
   RendererType,
   ScaleType,
@@ -100,6 +101,8 @@ export class ScalarCardComponent<Downloader> {
 
   yScaleType = ScaleType.LINEAR;
   isViewBoxOverridden: boolean = false;
+  // array of columns that will be displayed in the Data Table when enabled. The
+  // order of this array will be the order displayed on the table.
   dataHeaders: ColumnHeaders[] = [
     ColumnHeaders.RUN,
     ColumnHeaders.VALUE,
@@ -199,20 +202,19 @@ export class ScalarCardComponent<Downloader> {
     const dataTableData = this.dataSeries
       .map((datum) => {
         const metadata = this.chartMetadataMap[datum.id];
-        const closestStartPoint = this.getClosestPoint(
-          datum.points,
-          this.selectedTime!.startStep
-        );
-        const closestEndPoint = null;
+        const closestStartPoint =
+          datum.points[
+            findClosestIndex(datum.points, this.selectedTime!.startStep)
+          ];
         return {
           metadata,
           closestStartPoint,
-          closestEndPoint,
         };
       })
       .filter(({metadata}) => {
         return metadata && metadata.visible && !Boolean(metadata.aux);
       });
+
     return dataTableData;
   }
 
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -40,7 +40,10 @@ import {
 import {LinkedTime} from '../../../widgets/linked_time_fob/linked_time_types';
 import {TooltipSort, XAxisType} from '../../types';
 import {
+  ColumnHeaders,
+  RunStepData,
   ScalarCardDataSeries,
+  ScalarCardPoint,
   ScalarCardSeriesMetadata,
   ScalarCardSeriesMetadataMap,
 } from './scalar_card_types';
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -183,6 +192,48 @@ export class ScalarCardComponent<Downloader> {
...
+        return {
+          metadata,
+          closestStartPoint,
+          closestEndPoint,
+        };
+      })
+      .filter(({metadata}) => {
+        return metadata && metadata.visible && !Boolean(metadata.aux);
+      });
+    return dataTableData;
```
japie1235813, line 216: Now the input type looks like: `{metadata: {...}, closestStartPoint: {...}}`
I'm thinking about cleaning this before feeding to data table

Maybe a more fundamental question is, how/where do you plan to clean up things? In my opinion, the data table takes a list of columnHeaders and an object. It "prints" the value from the object if there are the keys matching the columnHeader.

which means, the input [data]=.. is something like
```
[{
run: ..
step: ...
}, {
run: ..
step: ...
}]
```
each item represents a row in the data table. What do you think?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2246,4 +2246,229 @@ describe('scalar card', () => {
...
+            step: 2,
+            x: 2,
+            y: 10,
+            relativeTimeInMs: 1000,
+          },
+          closestEndPoint: null,
+        },
+      ]);
+    }));
+    it('Selects closest points to selectedTime', fakeAsync(() => {
```
japie1235813, line 2338: nit: could you put a new line between each `it('...`?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -183,6 +192,48 @@ export class ScalarCardComponent<Downloader> {
...
+  }
+
+  getClosestPoint(points: ScalarCardPoint[], step: number): ScalarCardPoint {
```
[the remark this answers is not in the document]
JamesHollyer: findClosestIndex will not work. That takes in a value and finds the point with the closest value. We want the point with the closest step. I could move this to the utils.ts file which provides utility functions to all card_renderers. However, it is currently only used here and I have no plans to use it elsewhere.

JamesHollyer (commented):
Thanks Jie-Wei!

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -183,6 +195,47 @@ export class ScalarCardComponent<Downloader> {
...
+        };
+      })
+      .filter(({metadata}) => {
+        return metadata && metadata.visible && !Boolean(metadata.aux);
+      });
+
+    return dataTableData;
+  }
+
+  getClosestPoint(points: ScalarCardPoint[], step: number): ScalarCardPoint {
```
japie1235813, line 221: Can we remove this method `getClosestPoint` then?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_types.ts
```diff
@@ -56,6 +66,11 @@ export interface ScalarCardPoint extends Point {
 
 export type ScalarCardDataSeries = DataSeries<ScalarCardPoint>;
 
+export interface RunStepData {
+  metadata: ScalarCardSeriesMetadata;
+  closestStartPoint: ScalarCardPoint;
```
japie1235813, line 71: Can we then make this to 
```
{
displayName: string,
color: .. // do we need color?
smoothed: number, // not sure where to get this
value: number,
step: number,
}
```
The idea of "giving a bunch of data and show what headers have" looks okay to me, but it is just confusing to me why we need to have two layers of that data (i.e., separate them to metadata and closestStartPoint)

Also if we want to show min/max later, do you plan to calculate this and passed to the component or do thin inside the component?
If the former, we need to extend this interface (to have min/max etc); if the latter, we then need to passed in all the data points.
I might prefer to do this as in the first option considering recalculating on updating the range.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
japie1235813: Can we add this ("get the required piece of information for each header") to the jsdoc of `tensorboard/webapp/widgets/data_table/data_table_component.ts`
Or added the comment above `interface RunStepData`?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -32,6 +32,7 @@ import {
   siNumberFormatter,
 } from '../../../widgets/line_chart_v2/lib/formatter';
 import {LineChartComponent} from '../../../widgets/line_chart_v2/line_chart_component';
+import {findClosestIndex} from '../../../widgets/line_chart_v2/sub_view/line_chart_interactive_utils';
```
japie1235813, line 35: I might be hesitant to import the interactive_utils widget here but let's keep this right now and refactor later.

japie1235813 (commented):
It is looking better just some final picky comments. Thanks for your patience!

PR #5744, opened 2022-06-06
step selector: gather data for data table

JamesHollyer:
This is one pr in a chain intended to add a data table feature. This change adds data types and gathers the data needed for the table. For more context on where this is headed see https://github.com/tensorflow/tensorboard/pull/5743

The ScalarCard now decides which columns we will have on the table(currently this is hard coded but, we plan to allow for the user to select which columns they want so this is built with that in mind). It also uses the selectedTime to decide which points of data will be displayed.  Eventually we will replace selected time with a more narrow step selection that is specific for that card. 

* Screenshots of UI changes
This PR makes no UI changes but, for reference we are aiming to make a table like this:

There was another design to pass the point data and run metadata down to the table and have that table understand how to get the correct data per column. We plan to one day allow for ranges of data and there are some calculations that would prove very difficult in the data table(for instance the average of all values within the range) so we went with this solution where the values are "calculated" in the ScalarCardComponent and then passed to the data table.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -33,6 +33,7 @@ import {
   siNumberFormatter,
 } from '../../../widgets/line_chart_v2/lib/formatter';
 import {LineChartComponent} from '../../../widgets/line_chart_v2/line_chart_component';
+import {findClosestIndex} from '../../../widgets/line_chart_v2/sub_view/line_chart_interactive_utils';
 import {
   RendererType,
   ScaleType,
@@ -40,9 +41,11 @@ import {
 } from '../../../widgets/line_chart_v2/types';
 import {TooltipSort, XAxisType} from '../../types';
 import {
+  ColumnHeaders,
   ScalarCardDataSeries,
   ScalarCardSeriesMetadata,
   ScalarCardSeriesMetadataMap,
+  SelectedStepRunData,
 } from './scalar_card_types';
 import {ViewSelectedTime} from './utils';
 
@@ -97,6 +100,12 @@ export class ScalarCardComponent<Downloader> {
 
   yScaleType = ScaleType.LINEAR;
   isViewBoxOverridden: boolean = false;
+  dataHeaders: ColumnHeaders[] = [
+    ColumnHeaders.RUN,
+    ColumnHeaders.VALUE,
+    ColumnHeaders.STEP,
+    ColumnHeaders.RELATIVE_TIME,
+  ];
 
   toggleYScaleType() {
     this.yScaleType =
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/widgets/data_table/data_table_component.ts
```diff
@@ -13,12 +13,19 @@ See the License for the specific language governing permissions and
...
 
 @Component({
   selector: 'tb-data-table',
   templateUrl: 'data_table_component.ng.html',
   styleUrls: ['data_table_component.css'],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class DataTableComponent {}
+export class DataTableComponent {
+  @Input() headers!: ColumnHeaders[];
```
japie1235813, line 29: The order of `ColumnHeaders[]` matters on component rendering but it's not documented anywhere, can we make a note here?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -97,6 +100,12 @@ export class ScalarCardComponent<Downloader> {
 
   yScaleType = ScaleType.LINEAR;
   isViewBoxOverridden: boolean = false;
+  dataHeaders: ColumnHeaders[] = [
+    ColumnHeaders.RUN,
+    ColumnHeaders.VALUE,
+    ColumnHeaders.STEP,
+    ColumnHeaders.RELATIVE_TIME,
+  ];
```
japie1235813, line 108: Just a thought, do you expect this array to be dynamic changing or a constant? 

Not required in this PR but I might prefer is as as a constant and put this outside of the scalar_card_component. Later in range selection, we put more attributes in the array but the data_table_component is smart enough to not render the column if there is no such key provided in data (type SelectedStepRunData). That is to say, in single selection  we have 'VALUE_CHANGE' in 'headers' but not in keys of SelectedStepRunData type object so the data_component will not render this column.

japie1235813 (approved):
woo-hoo I like this structure! Great work. Just some nits.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_types.ts
```diff
@@ -70,3 +84,7 @@ export interface PartitionedSeries {
   runId: string;
   points: ScalarCardPoint[];
 }
+
+export type SelectedStepRunData = {
```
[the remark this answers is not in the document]
JamesHollyer: Done.

tensorboard/webapp/widgets/data_table/data_table_component.ts
JamesHollyer: Done.

JamesHollyer (commented):
Thanks Jie-Wei

PR #5747, opened 2022-06-07
Change mimetype mapping of js files to 'text/javascript'.

bmd3k:
While investigating the test failure that lead to #5746, we realized that RFC 9239 (https://www.rfc-editor.org/rfc/rfc9239) has changed the preferred js mimetype to 'text/javascript' from 'application/javascript'.

To be a good citizen of the internet, this change modifies the logic introduced in #3128 to force the mapping of js files to mimetype 'text/javascript'. We must also modify http_util.py to ensure that the proper charset is included in the Content-Type header. And, finally, we modify the example plugin to use 'text/javascript' as mimetype.

To Test:
* I ran local tensorboard and used the chrome developer tools to verify that the Content-Type for js files fetched by the browser was 'text/javascript; charset=utf-8'.  I did a basic sanity test ensuring both Angular and Polymer portions of the application are running.
* I ran a tensorboard with the example plugin installed and verified that index.js is returned with Content-Type 'text/javascript'.

### DIFF
--- a/tensorboard/backend/http_util.py
+++ b/tensorboard/backend/http_util.py
@@ -63,6 +63,7 @@ _TEXTUAL_MIMETYPES = set(
         "text/css",
         "text/csv",
         "text/html",
+        "text/javascript",
         "text/plain",
         "text/tab-separated-values",
         "text/x-protobuf",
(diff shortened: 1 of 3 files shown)


nfelt (approved):
Thanks!

PR #5748, opened 2022-06-07
step selector: add internalSelectedTime to scalar card

japie1235813:
This is part of the work to disconnect fob from linked time. When linked time is disabled, we want to updates the fob location (and corresponding data table) for this scalar card only. This PR adds a new attribute "internalSelectedTime" which exists in component only.

   * We build "internalSelectedTime" inside scalar card. The initial value comes from the min step in this card.
   * We only dispatch timeSelectionChanged when "LinkedTime" is enabled. Whether linkedTime is enabled based on selectedTime inside scalar_card_component, aka `selectedTime===null` if linkedTime is not enabled.
   * Remove feature flag `isDataTableEnabled` and eventually replace with "internalSelectedTime" (The current data table still using linkedTime start step, you will need to enable linked time to see it, later "internalSelectedTime" should be added as OR condition with proper input handling.)

* This initial PR implementation has a couple of known TODOs:
  * render data table when internalSelectedTime is not null
  * consider move range input inside Step Selector
  * apply range visual effect (i.e., faded out out of range area) with internalSelectedTime
  * remember previous selected step after toggle off Step Selector

* Screenshots of UI changes

https://user-images.githubusercontent.com/1131010/172483602-752c836e-e750-4aeb-ba9b-8385482800e7.mov

Considered building "internalSelectedTime" in container as observable then using Subject to trigger this observable. Eventually abandon that direction cause it gets complicated to check if selectedTime$ observable has value (aka linked time is enabled)

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -84,8 +84,8 @@ export class ScalarCardComponent<Downloader> {
   @Input() xScaleType!: ScaleType;
   @Input() useDarkMode!: boolean;
   @Input() forceSvg!: boolean;
-  @Input() isDataTableEnabled!: boolean;
   @Input() selectedTime!: ViewSelectedTime | null;
+  @Input() internalSelectedTime!: LinkedTime;
 
   @Output() onFullSizeToggle = new EventEmitter<void>();
   @Output() onPinClicked = new EventEmitter<boolean>();
@@ -200,8 +200,9 @@ export class ScalarCardComponent<Downloader> {
 
   getLinkedTime(): LinkedTime | null {
     if (this.selectedTime === null) {
-      return null;
+      return this.internalSelectedTime;
     }
+
     return {
       start: {
         step: this.selectedTime!.startStep,
@@ -252,5 +253,13 @@ export class ScalarCardComponent<Downloader> {
       });
 
     return dataTableData;
+  }
+
+  onFobSelectTimeChanged(newSelectedTime: LinkedTime) {
+    this.internalSelectedTime = newSelectedTime;
+
+    if (this.selectedTime !== null) {
+      this.onSelectTimeChanged.emit(newSelectedTime);
+    }
   }
 }
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -316,7 +316,7 @@ describe('scalar card', () => {
     store.overrideSelector(selectors.getRunColorMap, {});
     store.overrideSelector(selectors.getDarkModeEnabled, false);
     store.overrideSelector(selectors.getForceSvgFeatureFlag, false);
-    store.overrideSelector(selectors.getIsDataTableEnabled, false);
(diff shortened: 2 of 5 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -253,4 +254,12 @@ export class ScalarCardComponent<Downloader> {
 
     return dataTableData;
   }
+
+  onFobSelectTimeChanged(newSelectedTime: LinkedTime) {
+    this.internalSelectedTime = newSelectedTime;
```
JamesHollyer, line 259: This is interesting. So the internalSelectedTime updates when updating linked time...but only on the card that is actually being dragged? Was that intentional? 

This could probably lead to some strange behavior. I think this might be a good design decision but, it is hard to tell without playing around with it.

I just wanted to make sure this was intentional and I am ok with it.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2445,4 +2445,66 @@ describe('scalar card', () => {
...
+        const fixture = createComponent('card1');
+        fixture.detectChanges();
+        const testController = fixture.debugElement.query(
+          By.directive(CardFobControllerComponent)
+        ).componentInstance;
+
+        expect(testController).toBeTruthy();
+      }));
+
+      it('dose not dispatch timeSelectionChanged action when fob is dragged', fakeAsync(() => {
```
JamesHollyer, line 2482: does* not

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2445,4 +2445,66 @@ describe('scalar card', () => {
...
+        testController.mouseMove(fakeEvent);
+        fixture.detectChanges();
+
+        const fobs = fixture.debugElement.queryAll(
+          By.directive(CardFobComponent)
+        );
+        expect(
+          fobs[0].query(By.css('span')).nativeElement.textContent.trim()
+        ).toEqual('25');
+        expect(dispatchedActions).toEqual([]);
```
JamesHollyer, line 2506: Can we also test that the internalSelectedTime IS updated.

JamesHollyer (approved):
I am going to go ahead and approve this. Please update the mistypes I have pointed out before submitting.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
japie1235813: Here's what I think: dragging the fob will always updates internal selected time. Only when it is "linked" we dispatch timeSelectionChange and update linkedTime in ngrx state to also update fob/data table in other cards.
This is intentional to keep the internal selected time up to date when we uncheck "LinkedTime" in setting panel. Without this there will be a jump because mismatch between internal selected time and linked selected time.

I think the input type "LinkedTime" is where causes the confusing, if we think about it as "a pair of start step and end step", would that be less confusing? (that's also why I change rename "newLinkedTime" to "newSelectedTime" but change the name of type involves a lot of other things)

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
japie1235813: That makes sense to me. Done

PR #5750, opened 2022-06-08
TensorBoard 2.9.1

bmd3k:
Cherrypicks:
* 174d8a6d35d Force protobuf dependency < 3.20.
* 28201b437ef Recognize text/javascript as valid js mimetype for caching purposes


nfelt (approved):
Thanks for doing this!

PR #5754, opened 2022-06-09
step selector: updates selected time to single selection only

japie1235813:
To cope with launch phases we only apply single selection to step selector for now. Therefore when dragging fob to new selected time, we update start step and sets end to null. This prevents having two fobs after linked time launch, which enables users to set range selection. The type of internalSelectedTime is still LinkedTime, which reserves future extension to range selection of step selector.

On fob changed, we set the start step instead of LinkedTime object.

* Screenshots of UI changes

https://user-images.githubusercontent.com/1131010/172928846-f19ab1e5-0582-4526-8a57-e80a4ba7f43b.mov

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -2530,7 +2530,7 @@ describe('scalar card', () => {
 
         // Simulate dragging start fob to step 25
         testController.startDrag(Fob.START);
-        const fakeEvent = new MouseEvent('mousemove', {
+        let fakeEvent = new MouseEvent('mousemove', {
           clientX: 25 + controllerStartPosition,
           movementX: 1,
         });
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2514,6 +2514,51 @@ describe('scalar card', () => {
...
+        const fixture = createComponent('card1');
+        fixture.detectChanges();
+        const testController = fixture.debugElement.query(
+          By.directive(CardFobControllerComponent)
+        ).componentInstance;
+        const controllerStartPosition =
+          testController.root.nativeElement.getBoundingClientRect().left;
+
+        // Simulate dragging fob to step 25.
+        testController.startDrag(Fob.START);
```
JamesHollyer, line 2532: I think it would be best to also add an End Fob drag here as well.

JamesHollyer (approved):
I added a suggestion for the test but, this LGTM.

PR #5755, opened 2022-06-10
build(deps): bump prost to 0.9.0, tonic to 0.6.2, tonic-reflection to 0.3.0

bmd3k:
Note: This subsumes #5738, which just attempted to update prost-types to 0.8.0.

For Reviewers: The only file I modified by hand is Cargo.toml. Everything else is auto-generated from subsequent commands (`cargo update`, `cargo raze`, `bazel run //tensorboard/data/server:update_protos`)

#5738 attempted to update prost-types from 0.7.0 to 0.8.0. This did not work out-of-the-box for several reasons, the most important of which is that we could not successfully build and run the server with mixed versions of prost (0.7.0), prost-types (0.8.0), and prost-build (0.7.0).

This PR attempts to upgrade all three of prost, prost-types, and prost-build to the same version at the same time. We also upgrade tonic, tonic-reflection, and tonic-build so that they, too, depend on the same prost version and we don't have multiple prost crates of different versions.

* https://crates.io/crates/prost/0.9.0 
* https://crates.io/crates/prost-types/0.9.0 
* https://crates.io/crates/prost-build/0.9.0 
* https://crates.io/crates/tonic/0.6.2 
* https://crates.io/crates/tonic-build/0.6.2 
* https://crates.io/crates/tonic-reflection/0.3.0 

We choose to upgrade prost to 0.9.0 because, why not? We attempted to upgrade to 0.10.* (the latest version) but these versions of prost have a dependency on cmake, which seems to assume cmake is installed on your machine, which it typically is not within Google. There seems to be some work to further refine the dependency on cmake (https://github.com/tokio-rs/prost/pull/620, https://github.com/tokio-rs/prost/pull/657) so hopefully this gets resolved in a future release of prost and a future upgrade to, say, 0.11 will not have the cmake problems.

The upgrade was not straightforward for me, a rust and rustboard newb. Several other edits had to be made and commands had to be run to get it to work. This was the incantation that finally worked for me:
* Edit Cargo.toml to update versions of prost, prost-types, prost-build, tonic, tonic-reflection, tonic-build and to update any related raze metadata entries.
* Run `cargo update -p prost -p prost-types -p prost-build -p tonic -p tonic-reflection -p tonic-build`
* Inspect Cargo.lock and compare to raze metadata entries in Cargo.toml. If any of the versions changed (indexmap) then update the associated raze metadata entry. If any of the crates were removed (none were) then remove the associated raze metadata entry. Also, libc version didn't change but I updated its raze metadata anyway since its version seems to have changed in the past.
* Run `cargo raze`
* Run `bazel test //tensorboard/data/server:update_protos_test` (especially since this is updating proto-related libraries) and witness it FAIL and then run `bazel run //tensorboard/data/server:update_protos`.
* Confirm server works with `bazel run -c opt //tensorboard/data/server -- --logdir <some logdir>`


nfelt (approved):
Thanks for taking this on, and for the detailed explanation in the PR description!

> The upgrade was not straightforward for me, a rust and rustboard newb. Several other edits had to be made and commands had to be run to get it to work. This was the incantation that finally worked for me:

Just wondering, were you starting from the process here? https://github.com/tensorflow/tensorboard/blob/master/tensorboard/data/server/DEVELOPMENT.md#adding-third-party-dependencies

If there are additional steps that you discovered were needed (e.g. `cargo update` rather than `cargo fetch`), maybe it would be worthwhile to update that doc so it matches what you needed to do?

PR #5766, opened 2022-06-16
ci: build tb-nightly pip package and upload it as an artifact

yatbear:
Add steps to the existing workflow to build the `tb-nightly` pip package and upload it as an artifact.

Tested: 
Checked that CI properly built a package containing a `.whl` file [here](https://github.com/tensorflow/tensorboard/actions/runs/2518130900):
image

Downloaded the file, unzipped and installed in a virtualenv (`pip install tb_nightly-2.10.0a20220616-py3-none-any.whl`), and ran `tensorboard` locally.

#jubilee #ci #tb-nightly


.github/workflows/ci.yml
```diff
@@ -120,6 +120,22 @@ jobs:
           bazel test //tensorboard/summary/writer:event_file_writer_s3_test &&
           bazel test //tensorboard/compat/tensorflow_stub:gfile_fsspec_test &&
           bazel test //tensorboard/summary/writer:event_file_writer_fsspec_test
+      - name: 'Bazel: build Pip package (with TensorFlow only)'
+        if: matrix.tf_version_id == 'tf'
+        run: |
+          version="$(grep -Po "(?<=VERSION = ['\"]).*?(?=['\"])" tensorboard/version.py)"
```
nfelt, line 126: Optional, but we might consider factoring out the `grep` and `sed` commands here (the first 5 lines) into a separate shell script (e.g. like `tensorboard/tools/update_version.sh`) so that they're a bit easier to test and invoke manually?

In particular, it might be nice to be able to extend these commands so that we could have this same build step work for doing actual releases (not just nightly), in which case we'd need to make the logic a bit more complicated, and that would probably be best to do in a separate script vs inline here.  However, fine to leave as-is for now if you prefer.

nfelt (approved):
Nice work!

.github/workflows/ci.yml
yatbear: Good point! Done.

.github/workflows/ci.yml
```diff
@@ -120,6 +120,18 @@ jobs:
           bazel test //tensorboard/summary/writer:event_file_writer_s3_test &&
           bazel test //tensorboard/compat/tensorflow_stub:gfile_fsspec_test &&
           bazel test //tensorboard/summary/writer:event_file_writer_fsspec_test
+      - name: 'Bazel: build Pip package (with TensorFlow only)'
+        if: matrix.tf_version_id == 'tf'
+        run: |
+          ./tensorboard/tools/mirror_urls_test.sh
```
nfelt, line 126: Should be update_version.sh?

yatbear: oops, forgot to commit the latest change, thanks!

yatbear: Also had to add `chmod a+x` for the script.

PR #5768, opened 2022-06-21
Remove transitive dep from tensorboard polymer on Angular.

bmd3k:
Change `tensorboard/webapp/feature_flag/http:const` target to use `tf_ts_library` instead of `tf_ng_module`. const.ts does not have any dependency on Angular so using `tf_ng_module` was unnecessary and created a transitive dependency from our polymer codebase to Angular, which we don't like (this was caught by internal tests).


JamesHollyer (approved):
LGTM

PR #5771, opened 2022-06-23
Remediate setup-node GH Action warning

groszewn:
The `setup-node` GH Action currently specifies `version` (which has been
deprecated since October 1, 2019) instead of `node-version`.

Current action output:
image

After this change:
image


nfelt (approved):
Thanks!

PR #5779, opened 2022-06-28
step selector: add missing dep for @angular/core/testing

nfelt:
The file data_table_test.ts as added in https://github.com/tensorflow/tensorboard/pull/5767 uses `@angular/core/testing`, and for that import, we require this placeholder dep in order to sync properly internal to Google.


japie1235813 (approved):
Thanks for helping with the missing deps!

PR #5781, opened 2022-06-28
cleanup: Remove tomservo comment that no longer applies

groszewn:
As we've migrated out nightly build to GH Actions, the tomservo comment
no longer applies.


WORKSPACE
```diff
@@ -25,7 +25,6 @@ load("@bazel_skylib//lib:versions.bzl", "versions")
 
 # Keep this version in sync with:
 #  * The BAZEL environment variable defined in .github/workflows/ci.yml.
-#  * The Bazel version installed for nightly builds (Googlers, see go/tb-tomservo).
```
yatbear, line 28: Maybe we could keep the first half of the comment `The Bazel version installed for nightly builds.`?

groszewn: The comment above it (referencing .github/workflows/ci.yml) is the version installed for nightly builds (since the nightly workflow invokes the CI job). How about:

`The Bazel environment version defined in .github/workflows/ci.yml, which is used for CI and nightly builds.`?

yatbear: Makes sense!

yatbear (approved):

PR #5782, opened 2022-06-28
data: add `default_prefix` arg for DispatchingDataProvider

yatbear:
This is a prerequisite step for b/237101984.

Googlers, see the discussion in: cl/457725357

#refactor


nfelt (approved):
LGTM! Probably worth confirming this will pass internal tests via `copybara_presubmit.sh`, if you haven't checked that already.

yatbear:
> LGTM! Probably worth confirming this will pass internal tests via `copybara_presubmit.sh`, if you haven't checked that already.

Thanks for the suggestion! There are indeed bugs! I had to set the `_default_provider` to unprefixed provider if no `default_prefix` directly, rather than deriving from the unprefixed provider, since in internal tests the `unprefixed_provider` might be a pre-existing one included in the `_providers`. See cl/458001580 for the test.

nfelt:
> > LGTM! Probably worth confirming this will pass internal tests via `copybara_presubmit.sh`, if you haven't checked that already.
> 
> Thanks for the suggestion! There are indeed bugs! I had to set the `_default_provider` to unprefixed provider if no `default_prefix` directly, rather than deriving from the unprefixed provider, since in internal tests the `unprefixed_provider` might be a pre-existing one included in the `_providers`. See [cl/458001580](http://cl/458001580) for the test.

Ah I see, I didn't look closely at the tests (I assume you mean it might _not_ be a pre-existing one included in the providers). Yeah, I think what you have here works and is simpler anyway.  It won't set default_prefix to something useful, but that's ok because we don't actually use default_prefix in this PR other than to set the default_provider, and we won't use default_prefix in the internal version except after we already start passing it in as an argument.  So yep, LGTM!

PR #5784, opened 2022-06-29
step selector: deselect fob in step selector

japie1235813:
The behavior of deselect fob still connects to linked time behavior (update "selectTimeEnableToggled", which is for linked time). We want to make the fob disappear when we deselect it in step selector single selection. This PR checks whether linked time is enabled and fires `stepSelectorEnableToggled` or `selectTimeEnableToggled` actions accordingly.

1. fire `stepSelectorEnableToggled` action when linked time is not enabled. The "onSelectTimeToggle" is for linked time change selected time in ngrx state. When it's step selector, we do not have selected step in state (storing in each card level)
2. fix step selector checking bug: it's probably misunderstanding and misuse. `isScalarStepSelectorFeatureEnabled` is about wheteh the feature flag is turned on. `isScalarStepSelectorEnabled` is when users check/uncheck step selector.

* Demo

https://user-images.githubusercontent.com/1131010/176325540-24f330b2-37fa-4f57-9c55-4a5445940d14.mov

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -2621,7 +2621,7 @@ describe('scalar card', () => {
         });
       }));
 
-      it('toggles selected time when single fob is deselected in step selector', fakeAsync(() => {
+      it('toggles when single fob is deselected', fakeAsync(() => {
         const fixture = createComponent('card1');
         fixture.detectChanges();
         const fobComponent = fixture.debugElement.query(
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -69,7 +69,11 @@ import {
 } from '../../../widgets/line_chart_v2/types';
 import {ResizeDetectorTestingModule} from '../../../widgets/resize_detector_testing_module';
 import {TruncatedPathModule} from '../../../widgets/text/truncated_path_module';
-import {selectTimeEnableToggled, timeSelectionChanged} from '../../actions';
+import {
+  selectTimeEnableToggled,
+  stepSelectorEnableToggled,
+  timeSelectionChanged,
+} from '../../actions';
 import {PluginType} from '../../data_source';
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2616,6 +2620,17 @@ describe('scalar card', () => {
           end: null,
         });
       }));
+
+      it('toggles selected time when single fob is deselected in step selector', fakeAsync(() => {
```
bmd3k, line 2624: maybe: toggles step selector when single fob is deselected in step selector

tensorboard/webapp/metrics/views/right_pane/settings_view_component.ts
```diff
@@ -65,6 +65,7 @@ export class SettingsViewComponent {
 
   @Input() isCardWidthSettingEnabled!: boolean;
   @Input() isLinkedTimeFeatureEnabled!: boolean;
+  @Input() isScalarStepSelectorFeatureEnabled!: boolean;
   @Input() isScalarStepSelectorEnabled!: boolean;
   @Input() selectTimeEnabled!: boolean;
```
bmd3k, line 70: selectTimeEnabled => isLinkedTimeEnabled?

So you have isLinkedTimeFeatureEnabled and isLinkedTimeEnabled just like for ScalarStepSelector. This one might be worth doing in this PR but up to you.

bmd3k (approved):
I've called out a couple symbols that you might want to rename as part of the bigger rename PR we discussed offline; one symbol you could probably change in this PR; and then one suggestion to slightly rename a test.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
japie1235813: removed "step selector" cause this is nested in the "step selector feature integration" test

tensorboard/webapp/metrics/views/right_pane/settings_view_component.ts
japie1235813: already started on the giant renaming PR before seeing this comment. Will put in following PR

PR #5793, opened 2022-07-06
Patch protobuf and grpc dependencies to be Python 3.10 compatible.

bmd3k:
This PR allows TensorBoard to build and run with Python 3.10.

There are incompatibilities in the protobuf and grpc dependencies, which are quite old but we are unable to upgrade (see: #5561 , Googlers: http://b/219030239). We patch the necessary fixes from those projects into our older versions using the `patches` property of the `http_archive` rule.

For grpc we patch https://github.com/grpc/grpc/commit/dbe73c9004e483d24168c220cd589fe1824e72bc to avoid the following compile-time error:

```
Error in fail: Python Configuration Error: Problem getting python include path for /usr/local/google/home/bdubois/virtualenv/tensorboard/bin/python.
<string>:1: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
<string>:1: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
Is the Python binary path set up right? (See ./configure or /usr/local/google/home/bdubois/virtualenv/tensorboard/bin/python.) Is distutils installed? Are Python headers installed? Try installing python-dev or python3-dev on Debian-based systems. Try python-devel or python3-devel on Redhat-based systems.
```

For protobuf we patch https://github.com/protocolbuffers/protobuf/commit/9d61eada0f47d7be793983638c4a29707b192d0c to avoid the following runtime error at startup:

```
 File "/usr/local/google/home/bdubois/.cache/bazel/_bazel_bdubois/079646a57be11faea0b2bfefccb2a81a/execroot/org_tensorflow_tensorboard/bazel-out/k8-fastbuild/bin/tensorboard/dev.runfiles/com_google_protobuf/python/google/protobuf/descriptor.py", line 47, in <module>
    from google.protobuf.pyext import _message
AttributeError: module 'collections' has no attribute 'MutableSequence'
```

The changes patched into message.cc are not the latest. I considered also patching https://github.com/protocolbuffers/protobuf/commit/624d29d83306f3ce2c7e092dd44a891e04215e67 to remove the references to PY_MAJOR_VERSION but that change is wide in scope.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -123,6 +123,8 @@ http_archive(
         # $ git checkout tags/v3.9.2 -b my-patch
         # $ git cherry-pick 9d61eada0f47d7be793983638c4a29707b192d0c
         # $ git diff HEAD~1 > protobuf.patch
+        # # Remove trailing whitespace to satisify whitespace_hygiene_test.py.
+        # $ sed -i 's/[[:space:]]*$//' protobuf.patch
         # ```
         "//third_party:protobuf.patch",
     ],
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -110,33 +110,53 @@ sass_repositories()
 # might lead to test failures.
 http_archive(
     name = "com_google_protobuf",
+    patch_args = ["-p1"],
+    patches = [
+        # To maintain compatibility with python 3.10 and greater, we need to patch
+        # in the following protobuf change:
+        # https://github.com/grpc/grpc/commit/9d61eada0f47d7be793983638c4a29707b192d0c
+        #
+        # To reproduce the patch:
+        # ```
+        # $ git clone https://github.com/protocolbuffers/protobuf.git
+        # $ cd protobuf
+        # $ git checkout tags/v3.9.2 -b my-patch
+        # $ git cherry-pick 9d61eada0f47d7be793983638c4a29707b192d0c
+        # $ git diff HEAD~1 > protobuf.patch
+        # ```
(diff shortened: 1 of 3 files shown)


yatbear (approved):
Thanks!

WORKSPACE
```diff
@@ -121,12 +139,34 @@ http_archive(
...
+        # ```
+        # $ git clone https://github.com/grpc/grpc.git
+        # $ cd grpc
+        # $ git checkout tags/v1.27.0-pre1 -b my-patch
+        # $ git cherry-pick dbe73c9004e483d24168c220cd589fe1824e72bc
+        # $ git diff HEAD~1 > grpc.patch
+        # ```
+        #
+        # Note that we choose tags/v1.27.0-pre1 as the base since the version in
+        # the archive is 1.27.0-dev. It's not necessarily an exact match but is
```
nfelt, line 158: For future reference, the prefix of the archive that we strip (`grpc-b54a5b338637f92bfcf4b0bc05e0f57a5fd8fadd`) includes the commit hash, so we can always check out specifically that commit if you want an exact match.

nfelt (commented):
Thanks for taking care of this!

WORKSPACE
bmd3k: Thanks. I'm adjusting the instructions in https://github.com/tensorflow/tensorboard/pull/5795.

PR #5794, opened 2022-07-07
Remove unnecessary properties in tsconfig.json and tsconfig-lax.json.

bmd3k:
I've audited our tsconfig.json (used for Angular codebase) and tsconfig-lax.json (used for Polymer codebase) and found some properties that no longer seem to be necessary.

In tsconfig.json:
* skipLibCheck: This property seems to generally be used to resolve some conflicts in types defined by dependencies. But removing this does not leave to build errors so I assume that whatever motivation there was for including it has been eliminated.

In tconfig-lax.json:
* noEmitOnError: We set this to false but default is false.
* noErrorTruncation: We set this to false but default is false.
* skipLibCheck: Same reasoning as for tsconfig.json. Doesn't seem necessary any longer.


JamesHollyer (approved):
LGTM

PR #5796, opened 2022-07-07
Serve TensorBoard at /flags

rileyajones:
There is an existing page available at the route "/flags" on the frontend, however, the backend does support this route so it is functionally inaccessible.

Rather than just serving that app at "/" I added a list of routes for the app to be served at.
The meta tag content attribute is now generated using parameter which is closured during app startup.
get_resource_apps now generates resource paths for ALL supported routes.
Because of the logic in the `clean_path` function found in `application.py` which prevents paths from ending in "/" it is necessary to support routes which DO and DO NOT end with "/" slightly convoluting the logic.

* Screenshots of UI changes 
image

Start the application and manually navigate to the "/flags" route. Assert that the page is rendered.

### DIFF
--- a/tensorboard/plugins/core/core_plugin.py
+++ b/tensorboard/plugins/core/core_plugin.py
@@ -45,6 +45,10 @@ DEFAULT_PORT = 6006
 # transition to 'text/javascript'.
 JS_MIMETYPES = ["text/javascript", "application/javascript"]
 JS_CACHE_EXPIRATION_IN_SECS = 86400
+# Th paths that the application should be served at
+# Note that paths other than "/" should NOT end in '/' because trailing '/'
+# characters are removed by the serving logic in tensorboard/backend/application.py
+APP_PATHS = ["/", "/flags"]
 
 
 class CorePlugin(base_plugin.TBPlugin):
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/app_routing/programmatical_navigation_types.ts
```diff
@@ -40,10 +40,17 @@ export interface NavigateToExperiments {
...
+  routeParams: {},
+  resetNamespacedState?: boolean;
+}
+
 export type ProgrammaticalNavigation =
   | NavigateToExperiment
   | NavigateToCompare
-  | NavigateToExperiments;
+  | NavigateToExperiments
+  | NavigateToFlags;
```
rileyajones, line 53: This is not technically necessary but it does make the typing more accurate.

tensorboard/plugins/core/core_plugin.py
```diff
@@ -109,17 +113,34 @@ def get_resource_apps(self):
...
+                            if app_path.endswith("/")
+                            else app_path + "/"
+                        )
+                        apps[path_with_slash + path] = wsgi_app
+        for app_path in APP_PATHS:
+            if app_path.endswith("/"):
+                apps[app_path] = apps[app_path + "index.html"]
+            else:
+                apps[app_path] = apps[app_path + "/index.html"]
+                apps[app_path] = apps[app_path]
```
rileyajones, line 143: This DEFINITELY needs a unit test

JamesHollyer: Unit test here would be great. I don't understand what line 143 does.

tensorboard/webapp/app_routing/programmatical_navigation_types.ts
JamesHollyer: I don't have strong feelings about this either way. However, I would have left this out because we do not want to programmatically navigate. 

I can definitely see this being useful one day if we decide to have a secret button that navigates here or something. But, I think we can add it then.

JamesHollyer (commented):
This LGTM. However, I am not familiar with this code enough to sign off on it. I am going to add Brian as a reviewer so sign off on the python piece.

rileyajones:
@JamesHollyer I spoke with @bmd3k and @nfelt about this on Friday and we decided to use a query param and a modal instead to prevent the python routing code from diverging. That is also why I didn't bother adding a test btw.
I'm converting this to draft rather than closing in case we decide to come back to this.

PR #5760, opened 2022-06-15
Feature Helper: add route and build page

JamesHollyer:
* Motivation for features / changes
Adding a page which helps to see which flags are enabled, allow flag changes, and enable more persistent flag settings. This change adds the page and displays which flags are enabled. The check boxes are currently not functional.

* Technical description of changes
This PR adds the route in the angular app. The python web server still does not recognize this path so it is only accessible with in app routing. When the angular app is loaded the routing works. The page simply uses the getFeatureFlags selector to get which flags are enabled and loops through each to display their status.

* Screenshots of UI changes
Currently the page looks like this: 
Screen Shot 2022-06-15 at 11 30 42 AM

It is subject to change.

* Detailed steps to verify changes work correctly (as executed by you)
I added a link which takes me to the page for testing. This link was not added into the PR.

* Alternate designs / implementations considered

PR #5800, opened 2022-07-11
Show flags page as modal when showFlags query parameter is set

rileyajones:
I'm breaking this change into two parts and blocking this with #5868
# Description of Changes
Add new feature flag to control showing the feature flag modal, add modal to wrapper component to show flags page when the showFlags query parameter is set.

~Currently blocked by [5717](https://github.com/tensorflow/tensorboard/pull/5717)~

As of [5760](https://github.com/tensorflow/tensorboard/pull/5760) we have a `/flags` page which is not accessible due to the backend routing code.

Last week I suggested updating the routing code to enable `/flags` to be shown ([5796](https://github.com/tensorflow/tensorboard/pull/5796)). And it was decided that it was undesirable to complicate the routing code and thus decided to show the page as a modal instead.

I've chosen to show the `FlagsPageContainer` as the contents of a `MatDialog` modal based on the status of the `enableShowFlags` feature flag which is controlled by the query parameter `showFlags`.

Given that we want to use the modal as an alternative to a route, it should appear on all pages and that the modal is unrelated to any existing part of the page, I have elected to create and manage it as part of the page wrapper.

#### `showFlags=true`
image

image

 #### `showFlags` not set
image

1) Launch and navigate to TensorBoard.
2) Note that the flags modal does not appear.
3) Add the query param `showFlags=true` or `showFlags`.
4) Assert that the modal is shown.
5) Assert that clicking behind the modal results in it being hidden.
6) Remove the query param or set `showFlags=false` and assert the modal no longer appears.


tensorboard/webapp/tb_wrapper/tb_wrapper_component.ts
```diff
@@ -37,4 +46,27 @@ import {ChangeDetectionStrategy, Component} from '@angular/core';
   ],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class TensorBoardWrapperComponent {}
+export class TensorBoardWrapperComponent {
```
rileyajones, line 49: It's unclear to me that this is the best spot to put this when considering other flavors of TensorBoard but I'm not sure where would be better and thus am open to suggestions.

tensorboard/webapp/feature_flag/views/feature_flag_page.ng.html
```diff
@@ -16,16 +16,31 @@
...
+            <mat-option value="enabled">Enabled</mat-option>
+            <mat-option value="disabled">Disabled</mat-option>
+          </mat-select>
+        </td>
+        <td>
+          <div>{{state.flag}}</div>
+        </td>
       </tr>
     </ng-container>
+    <button mat-button (click)="resetFlags()">Reset</button>
```
JamesHollyer, line 44: This should be "Reset All"

tensorboard/webapp/feature_flag/views/feature_flag_page_component.ts
```diff
@@ -12,20 +12,31 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+} from './types';
+
 @Component({
   selector: 'feature-flag-page-component',
+  styleUrls: ['feature_flag_page_style.css'],
   templateUrl: `feature_flag_page.ng.html`,
 })
 export class FeatureFlagPageComponent {
-  @Input() featureFlags!: FeatureFlagState;
+  @Input() featureFlagStatuses!: Array<FeatureFlagState<keyof FeatureFlags>>;
```
JamesHollyer, line 29: I think we generally prefer FeatureFlagState\<keyof FeatureFlags\>[]

tensorboard/webapp/feature_flag/views/feature_flag_page_component.ts
```diff
@@ -12,20 +12,31 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+  @Output() flagChanged = new EventEmitter<FeatureFlagStatusEvent>();
+
+  @Output() flagsReset = new EventEmitter();
 
-  getFlagKeys(): string[] {
-    return Object.keys(this.featureFlags);
+  setFlag(flag: keyof FeatureFlags, status: FeatureFlagStatus) {
```
JamesHollyer, line 35: Why not just emit the event directly from the html?

tensorboard/webapp/feature_flag/views/feature_flag_page_component.ts
```diff
@@ -12,20 +12,31 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    this.flagChanged.emit({flag, status});
   }
 
-  flagChanged(flag: string) {
-    // TODO: dispatch action which stores new flag states and updates state.
+  resetFlags() {
```
JamesHollyer, line 39: Same comment amount emitting directly.

tensorboard/webapp/feature_flag/views/feature_flag_page_container.ts
```diff
@@ -14,15 +14,90 @@ limitations under the License.
...
 })
 export class FeatureFlagPageContainer {
   constructor(private readonly store: Store<State>) {}
-  readonly featureFlags$ = this.store.select(getFeatureFlags);
+
+  readonly featureFlags$: Observable<FeatureFlagState<keyof FeatureFlags>[]> =
+    this.store.select(getDefaultFeatureFlags).pipe(
+      withLatestFrom(this.store.select(getOverriddenFeatureFlags)),
+      map(([defaultFeatureFlags, overriddenFeatureFlags]) => {
+        console.log(defaultFeatureFlags);
```
JamesHollyer, line 51: remove console log.

tensorboard/webapp/tb_wrapper/tb_wrapper_component.ts
JamesHollyer: I am also not sure this is the best spot. I would have to do some exploring to see if I can find a better spot. But, maybe someone else will have a good suggestion so I will wait on other reviews before digging further.

bmd3k: How about:
* Create new FeatureFlagModalTriggerComponent that contains this logic.
* Put <feature-flag-modal-trigger> in AppContainer: https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/app_container.ng.html

By putting it in AppContainer you would get the modal for free in all routes (in theory -- if Standalone TensorBoard actually contained other routes). But, more importantly, it would be a model for both TensorBoard.corp and TensorBoard.dev to also follow (they have their own AppContainers) and they do have multiple routes.

tensorboard/webapp/feature_flag/views/feature_flag_module.ts
```diff
@@ -23,7 +24,7 @@ import {FeatureFlagPageContainer} from './feature_flag_page_container';
  */
 @NgModule({
   declarations: [FeatureFlagPageComponent, FeatureFlagPageContainer],
-  imports: [CommonModule, MatCheckboxModule],
+  imports: [CommonModule, MatButtonModule, MatSelectModule],
```
bmd3k, line 27: You should chat with James about this change. It will require getting the internal version of this source allowlisted to use these modules. I believe James had a TODO to get blanket allowlisted for all these Angular component-related modules. Now (ie before this PR is submitted) could be a good time to get that process started.

tensorboard/webapp/tb_wrapper/tb_wrapper_component.ts
bmd3k: And, to be clear, by "Put in AppContainer" I mean modify app_container.ng.html to refer to <feature-flag-modal-trigger> component.

tensorboard/webapp/feature_flag/views/feature_flag_module.ts
JamesHollyer: I do remember this. I never followed up. I will submit that request tomorrow after syncing it with the latest.
(for googlers see cl/456268297).

tensorboard/webapp/feature_flag/views/feature_flag_page_component.ts
rileyajones: Internal vs external lint rules 🤷‍♂️

tensorboard/webapp/tb_wrapper/tb_wrapper_component.ts
rileyajones: I believe I've made the change you suggested.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_component.ts
```diff
@@ -0,0 +1,54 @@
...
+import {State} from '../../app_state';
+import {featureFlagOverrideChanged} from '../actions/feature_flag_actions';
+import {
+  getFeatureFlags,
+  getShowFlagsEnabled,
+} from '../store/feature_flag_selectors';
+import {FeatureFlagPageContainer} from '../views/feature_flag_page_container';
+
+@Component({
+  selector: 'feature-flag-modal-trigger-component',
```
bmd3k, line 27: Naming convention: Remove 'component' from the selector name.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_component.ts
```diff
@@ -0,0 +1,54 @@
...
+  template: ``,
+  styles: [],
+})
+export class FeatureFlagModalTriggerComponent {
```
bmd3k, line 31: Name this "FeatureFlagModalTriggerContainer". 

Our naming convention: Typically any Angular Component that gets data from ngrx is named with "Container" rather than with "Component". Angular Components with "Component" in the name typically are ngrx-free - if they need data from ngrx they then are wrapped by a "Container" layer to do the ngrx stuff.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_component.ts
```diff
@@ -0,0 +1,54 @@
...
+  readonly featureFlags$ = this.store.select(getFeatureFlags);
+  private featureFlagsDialog?: MatDialogRef<FeatureFlagPageContainer>;
+
+  constructor(private readonly store: Store<State>, private dialog: MatDialog) {
+    this.showFeatureFlags$.subscribe((showFeatureFlags: boolean) => {
+      if (showFeatureFlags) {
+        this.featureFlagsDialog = this.dialog.open(FeatureFlagPageContainer);
+        this.featureFlagsDialog.afterClosed().subscribe(() => {
+          this.store.dispatch(
+            featureFlagOverrideChanged({
```
bmd3k, line 42: Interesting. Can you add a comment explaining why you do this? (I guess you are just trying to keep the dialog from reopening?)

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_component.ts
```diff
@@ -0,0 +1,54 @@
...
+  selector: 'feature-flag-modal-trigger-component',
+  template: ``,
+  styles: [],
+})
+export class FeatureFlagModalTriggerComponent {
+  readonly showFeatureFlags$ = this.store.select(getShowFlagsEnabled);
+  readonly featureFlags$ = this.store.select(getFeatureFlags);
+  private featureFlagsDialog?: MatDialogRef<FeatureFlagPageContainer>;
+
+  constructor(private readonly store: Store<State>, private dialog: MatDialog) {
```
bmd3k, line 36: Can you add tests? Especially important for this layer to have some tests since we don't want to accidentally open the dialog for users.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_module.ts
```diff
@@ -0,0 +1,29 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+import {CommonModule} from '@angular/common';
+import {NgModule} from '@angular/core';
+import {FeatureFlagModalTriggerComponent} from './feature_flag_modal_trigger_component';
+import {FeatureFlagPageModule} from './feature_flag_module';
+
+/**
+ * Provides the wrapper component that renders the main dashboard page.
```
bmd3k, line 21: This comment appears to have been copy and pasted and not updated.

tensorboard/webapp/feature_flag/views/feature_flag_page_container.ts
```diff
@@ -14,15 +14,89 @@ limitations under the License.
...
+    [featureFlagStatuses]="featureFlags$ | async"
+    (flagChanged)="onFlagChanged($event)"
+    (allFlagsReset)="onAllFlagsReset()"
   ></feature-flag-page-component>`,
 })
 export class FeatureFlagPageContainer {
   constructor(private readonly store: Store<State>) {}
-  readonly featureFlags$ = this.store.select(getFeatureFlags);
+
+  readonly featureFlags$: Observable<FeatureFlagState<keyof FeatureFlags>[]> =
```
bmd3k, line 47: Can you add some tests for FeatureFlagPageContainer/FeatureFlagPageComponent?

tensorboard/webapp/feature_flag/views/types.ts
```diff
@@ -0,0 +1,32 @@
...
+==============================================================================*/
+import {FeatureFlags} from '../types';
+
+export const enum FeatureFlagStatus {
+  ENABLED = 'enabled',
+  DISABLED = 'disabled',
+  DEFAULT = 'default',
+}
+
+export type FeatureFlagState<K extends keyof FeatureFlags> = {
```
bmd3k, line 23: "State" has a specific meaning in our application - it refers to data stored in ngrx. Can you rename?

tensorboard/webapp/feature_flag/views/feature_flag_page.ng.html
```diff
@@ -16,16 +16,31 @@
...
-          >Enabled</mat-checkbox
-        >
-        {{flag}}
+        <td>
+          <mat-select
+            [value]="state.status"
+            (selectionChange)="flagChanged.emit({flag: state.flag, status: $event.value})"
+          >
+            <mat-option value="default"
+              >Default - {{state.defaultValue}}</mat-option
```
bmd3k, line 33: Is it possible to use same wording? 
Rather than:
"Default - true", "Default - false", "Enabled", "Disabled"
Have:
"Default - Enabled", "Default - Disabled", "Enabled", "Disabled".

tensorboard/webapp/feature_flag/views/feature_flag_page_container.ts
```diff
@@ -14,15 +14,89 @@ limitations under the License.
...
+    this.store.select(getDefaultFeatureFlags).pipe(
+      withLatestFrom(this.store.select(getOverriddenFeatureFlags)),
+      map(([defaultFeatureFlags, overriddenFeatureFlags]) => {
+        return Object.entries(defaultFeatureFlags).map(
```
bmd3k, line 51: To consider for a followup PR (in order to keep this one from growing any larger): Can we disable the UI for any non-boolean feature flags until we support them? Like, render a line for those properties but don't allow the user to interact with the line and don't show "Enabled"/"Disabled" but some other text like "Not Supported in this UI" or something?

Or, if that's too difficult, just don't render them at all on the page.

Either way, I'd prefer not to be able to "Enable/Disable" these non-boolean feature flags in the UI.

tensorboard/webapp/feature_flag/views/feature_flag_module.ts
JamesHollyer: Trying to sync that CL was a pain because the allowlists were moved. (New cl for googlers cl/468741117)

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_component.ts
rileyajones: Thanks for the detailed explanation, that makes a lot of sense.

tensorboard/webapp/feature_flag/views/feature_flag_page.ng.html
rileyajones: I'm not sure that a direct mapping scales well if we decide to support non boolean flags. I'm going to instead create a formatter that will convert `true`/`false` values to `Enabled`/`Disabled` while also allowing us to extend formatting in the future.

tensorboard/webapp/feature_flag/views/feature_flag_module.ts
JamesHollyer: The Angular team suggests we update to use the new "MDC materials". I am inclined to agree with them. I suggest we keep this as is and Riley add this directory to the allow list. Then we discuss a large scale refactor  to all materials to use the new system. Once we are using the new system we no longer need to update allow lists.

tensorboard/webapp/feature_flag/views/feature_flag_page_container.ts
rileyajones: I've gone ahead and disabled these for now

Screen Shot 2022-08-22 at 1 14 17 PM
.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_component.ts
rileyajones: I'll go ahead and add a comment. This is to stop the dialog from opening again after the page is refreshed as I imagine people do not want this to always appear.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_component.ts
rileyajones: I've gone ahead and added some

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_container_test.ts
```diff
@@ -0,0 +1,102 @@
...
+      actualActions.push(action);
+    });
+  });
+
+  it('creates modal when enableShowFlags is true', () => {
+    store.overrideSelector(getDefaultFeatureFlags, {} as FeatureFlags);
+    store.overrideSelector(getOverriddenFeatureFlags, {});
+    store.overrideSelector(getShowFlagsEnabled, true);
+    const dialogMock = new MatDialogMock();
+    new FeatureFlagModalTriggerContainer(store, dialogMock as any);
```
bmd3k, line 90: Use Angular's built-in TestBed.createComponent() instead of constructing the components directly.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_container_test.ts
```diff
@@ -0,0 +1,102 @@
...
+import {
+  getDefaultFeatureFlags,
+  getOverriddenFeatureFlags,
+  getShowFlagsEnabled,
+} from '../store/feature_flag_selectors';
+import {FeatureFlags} from '../types';
+import {FeatureFlagModalTriggerContainer} from './feature_flag_modal_trigger_container';
+import {FeatureFlagPageContainer} from './feature_flag_page_container';
+
+class MatDialogMock {
```
bmd3k, line 36: https://material.angular.io/cdk/test-harnesses/overview

Sometime in the last 6 months we decided we should use test harnesses when appropriate. This seems like a possible great candidate, given that their first example is about using a harness for MatDialog. Would you mind trying to use that?

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_container.ts
```diff
@@ -0,0 +1,55 @@
...
+  readonly featureFlags$ = this.store.select(getFeatureFlags);
+  private featureFlagsDialog?: MatDialogRef<FeatureFlagPageContainer>;
+
+  constructor(private readonly store: Store<State>, private dialog: MatDialog) {
+    this.showFeatureFlags$.subscribe((showFeatureFlags: boolean) => {
+      if (showFeatureFlags) {
+        this.featureFlagsDialog = this.dialog.open(FeatureFlagPageContainer);
+        this.featureFlagsDialog.afterClosed().subscribe(() => {
+          // By disabling the flag when the dialog is closed to prevent it from appearing again after the page is refreshed.
+          this.store.dispatch(
```
bmd3k, line 42: Thinking about this after playing with the code locally: I think when the dialog closes the page should be refreshed.

My reasoning: This is the first time a user can change feature flag state AFTER the app has been loaded. This probably opens up a ton of subtle bugs that we never had to consider in the past. We shouldn't allow the user to play with the app in this state. Refreshing the page after the dialog is closed would put the app back into a proper state.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_container.ts
```diff
@@ -0,0 +1,55 @@
...
+            featureFlagOverrideChanged({
+              flags: {enableShowFlags: false},
+            })
+          );
+        });
+        return;
+      }
+      if (this.featureFlagsDialog) {
+        this.featureFlagsDialog.close();
```
bmd3k, line 51: This doesn't seem right. The only times I can imagine this code path being invoked are:

* User sets showFlags to false in the dialog.
* User presses "Reset All" in the dialog.

In neither case would I expect the dialog to be immediately closed.

bmd3k (commented):
It's a big change so I can only give you a partial review. I've reviewed everything in-depth except for the modal code and tests.

tensorboard/webapp/feature_flag/views/feature_flag_module.ts
JamesHollyer: The angular team allowed us to change the legacy allowlist to allow the entire webapp directory. Therefore you no longer need to add this directory explicitly to the allowlist.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_container.ts
rileyajones: I think it depends on how you think about `showFlags`. If you really think it is a feature flag then it makes sense that it should close, but I can definitely see why you would view that as confusing so I'll go ahead and remove this.

tensorboard/webapp/BUILD
```diff
@@ -266,6 +267,7 @@ tf_ng_web_test_suite(
         "//tensorboard/webapp/customization:customization_test_lib",
         "//tensorboard/webapp/deeplink:deeplink_test_lib",
         "//tensorboard/webapp/feature_flag/effects:effects_test_lib",
+        "//tensorboard/webapp/feature_flag/views:views_test",
```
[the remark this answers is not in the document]
rileyajones: That was in fact confusing but I'll go ahead and consolidate these. Out of curiosity do you have any idea as to why it has its own test suite? Looking deeper it seems there are actually 13 test suites?

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_container.ts
bmd3k: Sure. Let's address in a followup PR. The key is that users should not be able to use the app after changing the flag state.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_container_test.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+  });
+
+  function createComponent() {
+    fixture = TestBed.createComponent(FeatureFlagModalTriggerContainer);
+    loader = TestbedHarnessEnvironment.loader(fixture);
+    rootLoader = TestbedHarnessEnvironment.documentRootLoader(fixture);
+  }
+
+  it('creates modal when enableShowFlags is true', async () => {
```
bmd3k, line 74: Nice, this seems cleaner than having to create our own Fake.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_container_test.ts
```diff
@@ -0,0 +1,96 @@
...
+  getDefaultFeatureFlags,
+  getOverriddenFeatureFlags,
+  getShowFlagsEnabled,
+} from '../store/feature_flag_selectors';
+import {FeatureFlags} from '../types';
+import {FeatureFlagModalTriggerContainer} from './feature_flag_modal_trigger_container';
+import {FeatureFlagPageContainer} from './feature_flag_page_container';
+
+describe('feature_flag_modal_trigger_container', () => {
+  let actualActions: Action[];
```
bmd3k, line 40: actualActions and dispatchSpy now seem to be effectively unused.

tensorboard/webapp/feature_flag/views/feature_flag_page.ng.html
```diff
@@ -16,16 +16,37 @@
...
-          [checked]="featureFlags[flag]"
-          (change)="flagChanged(flag)"
-          >Enabled</mat-checkbox
-        >
-        {{flag}}
+        <td>
+          <div>{{flagStatus.flag}}</div>
+        </td>
+        <ng-container
+          *ngIf="flagStatus.defaultValue === true || flagStatus.defaultValue === false; then selectBlock else staticBlock"
```
bmd3k, line 31: I found `flagStatus.defaultValue === true || flagStatus.defaultValue === false` to be a bit amusing but maybe it would be clearer if we instead checked `typeof flagStatus.defaultValue === 'boolean'`

tensorboard/webapp/feature_flag/views/feature_flag_page.ng.html
```diff
@@ -16,16 +16,37 @@
...
+            (selectionChange)="flagChanged.emit({flag: flagStatus.flag, status: $event.value})"
+          >
+            <mat-option value="default"
+              >Default {{formatFlagValue(flagStatus.defaultValue)}}</mat-option
+            >
+            <mat-option value="enabled">Enabled</mat-option>
+            <mat-option value="disabled">Disabled</mat-option>
+          </mat-select>
+        </ng-template>
+        <ng-template #staticBlock>
```
bmd3k, line 45: It might be clearer to rename "staticBlock" to "unsupportedBlock"?

tensorboard/webapp/feature_flag/views/feature_flag_page_component_test.ts
```diff
@@ -0,0 +1,121 @@
...
+      imports: [CommonModule],
+      schemas: [NO_ERRORS_SCHEMA],
+    }).compileComponents();
+
+    const fixture = TestBed.createComponent(TestableComponent);
+    fixture.detectChanges();
+    return fixture.nativeElement.querySelector('feature-flag-page-component');
+  }
+
+  it('creates rows for each feature flag', async () => {
```
bmd3k, line 50: Normally we prefer to have a single test file for each container/component combination. We think it strikes a good balance between writing isolated tests but also ensuring a particular container/component combination works as expected when integrated together. The fact that they are split into a container/component is (mostly) an implementation detail that we hide from the tests.

In this case we would have a feature_flag_page_test instead of both feature_flag_page_component_test and feature_flag_page_container_test.

I mention it here specifically because for these tests where you introduce TestableComponent, you could instead test by constructing a FeatureFlagPageContainer and specifying appropriate values for your ngrx selectors and then checking that the UI renders as expected.

tensorboard/webapp/feature_flag/views/feature_flag_page_container_test.ts
```diff
@@ -0,0 +1,130 @@
...
+  function getComponent() {
+    return TestBed.createComponent(FeatureFlagPageContainer).componentInstance;
+  }
+
+  describe('onFlagChanged', () => {
+    it('creates override status is set to enabled or disabled not set', () => {
+      store.overrideSelector(getDefaultFeatureFlags, {} as FeatureFlags);
+      store.overrideSelector(getOverriddenFeatureFlags, {});
+      const component = getComponent();
+      component.onFlagChanged({
```
bmd3k, line 60: Again in the context of a single test file for each container/component combination, we generally prefer to test dispatching actions starting at the UI. That is: setup your ngrx selectors as appropriate, render the container/component/UI, interact with an element, verify the action is fired.

tensorboard/webapp/feature_flag/views/feature_flag_page_component_test.ts
```diff
@@ -0,0 +1,121 @@
...
+      },
+    ]);
+
+    const dataCells = component.querySelectorAll('td');
+    expect(dataCells.length).toEqual(3);
+    const selectors = component.querySelectorAll('mat-select');
+    expect(selectors.length).toEqual(1);
+  });
+
+  describe('formatFlagValue', () => {
```
bmd3k, line 88: Just to mention: These tests as written make sense to me. I would still consider moving them into feature_flag_page_test, if possible. But you could also keep these in feature_flag_page_component_test.

tensorboard/webapp/feature_flag/views/feature_flag_page_component.ts
```diff
@@ -12,20 +12,48 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-import {FeatureFlagState} from '../store/feature_flag_types';
+import {Component, EventEmitter, Input, Output} from '@angular/core';
+import {FeatureFlagType} from '../store/feature_flag_metadata';
+import {FeatureFlags} from '../types';
+import {FeatureFlagStatus, FeatureFlagStatusEvent} from './types';
+
 @Component({
   selector: 'feature-flag-page-component',
+  styleUrls: ['feature_flag_page_style.css'],
   templateUrl: `feature_flag_page.ng.html`,
```
bmd3k, line 23: style and template files should be named feature_flag_page_component.css and feature_flag_page_component.ng.html. "style" is redundant (in our opinion). we include "component" because it differentiates them from possible "container" css and html files, too.

bmd3k (commented):
Thanks, I think the rest looks pretty good. I think the biggest thing I'd recommend is to reorganize/refactor the feature_flag_page tests a bit to adhere to team norms. Otherwise just some nits.

tensorboard/webapp/feature_flag/views/feature_flag_modal_trigger_container_test.ts
rileyajones: Definitely, thanks for pointing me in the direction of the official solution.

tensorboard/webapp/feature_flag/views/feature_flag_page_component_test.ts
rileyajones: Done.

tensorboard/webapp/feature_flag/views/feature_flag_page_container_test.ts
rileyajones: I believe I have done this correctly.

rileyajones: I've merged the files and refactored the tests to move away from `componentInstance` except in the instance of `formatFlagValue` tests which I believe are significantly simpler as is.

tensorboard/webapp/feature_flag/views/feature_flag_page_component.ts
```diff
@@ -12,20 +12,52 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 
-  getFlagKeys(): string[] {
-    return Object.keys(this.featureFlags);
+  isEditable(flagStatus: FeatureFlagStatus<keyof FeatureFlags>) {
+    return typeof flagStatus.defaultValue === 'boolean';
   }
 
-  flagChanged(flag: string) {
-    // TODO: dispatch action which stores new flag states and updates state.
+  formatFlagValue(value: FeatureFlagType): string {
```
JamesHollyer, line 56: What is the logic in having formatFlagValue and getFormattedFlagValue as separate functions? Seems pretty straight forward to just have a single function.

tensorboard/webapp/feature_flag/views/feature_flag_page_test.ts
```diff
@@ -0,0 +1,232 @@
...
+  let dispatchSpy: jasmine.Spy;
+  let fixture: ComponentFixture<FeatureFlagPageContainer>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      declarations: [FeatureFlagPageContainer],
+      imports: [CommonModule, FeatureFlagPageModule],
+      providers: [
+        provideMockStore({
+          initialState: buildState(
```
bmd3k, line 52: An alternative to constructing initialState like this that we usually use is to just override all the selectors that the components rely on. I'm fine with what you have since its simple but just wanted to let you know in case you find that simpler in this case.

JamesHollyer (commented):
I think this is getting pretty close. In the future try to split up PRs into smaller chunks for review.

tensorboard/webapp/feature_flag/views/feature_flag_page_test.ts
```diff
@@ -0,0 +1,232 @@
...
+    store.overrideSelector(getOverriddenFeatureFlags, {
+      inColab: true,
+    });
+    createComponent();
+    const component = getComponent();
+
+    const dataCells = component.querySelectorAll('td');
+    expect(dataCells.length).toEqual(3);
+    const selectors = component.querySelectorAll('mat-select');
+    expect(selectors.length).toEqual(1);
```
bmd3k, line 184: Would it be worth to make the tests a litlte stricter - ie actually verify that the "inColab" row has a mat-select while the "enabledExperimentalPlugins" is marked as unsupported? The test is unusually vague about this.

bmd3k (approved):

tensorboard/webapp/feature_flag/views/feature_flag_page_test.ts
rileyajones: I tried that initially and I was getting errors from the selectors indicating that state was undefined.

tensorboard/webapp/feature_flag/views/feature_flag_page_test.ts
rileyajones: You're right, this is pretty vague. I've gone ahead and added another `expect` checking the value of the `td` which I think should be specific enough.

tensorboard/webapp/feature_flag/views/feature_flag_page_component.ts
rileyajones: My logic was that `getFormattedFlag` essentially serializes the flag value while `formatFlagValue` decorates it. The naming does seem kinda confusing though so I'll rename them to hopefully make more sense.

JamesHollyer (approved):

PR #5801, opened 2022-07-12
Cleanup code for limiting step selector to min and max values of series.

bmd3k:
In https://github.com/tensorflow/tensorboard/pull/5799 we fixed a bug, limiting the step selector to be in the range of the min and max values of the chart data.

This change cleans up some of that code based on the comments in the review. We:

1) Limit `minMaxSteps$` to be `Observable<MinMaxStep>` instead of `Observable<MinMaxStep | null>`.
2) Refactor some code to rely directly on `minMaxSteps$` rather than recalculating the data.


JamesHollyer (approved):
That is much cleaner. Thanks for doing this!

PR #5807, opened 2022-07-15
Step Selector: Fix fob position rendering bug and remove CardFobAdaptor

japie1235813:
Current we define a `CardFobController` and implements the methods at CardFobController level (e.g. `ScalarCardFobController`) and pass "this" to child component. However,  Angular will not detech the data changes within `ScalarCardFobController` (in our case, minMax and axisSize) because as far as it concerns, "this" does not change.
This causes the bug when minMax and axisSize are updated (changing window size, expand card group..) it does not cause card_fob_component to be re-rendered and the fob position is off.

To handle this, we need to modify the method `getAxisPositionFromStep` which transform step to position, to two separate methods, getAxisPositionFromStartStep and getAxisPositionFromEndStep, and execute these two methods on `ScalarCardFobController` level, and feed the result to CardFobController. This way we are sure the input data is changed and Angular will detect the changes then re-render the CardFobController.

For the rest of CardFobAdaptor methods, they are used to get step and involving mouse movements. They are either do not have the same re-render requirements or the changes will cause timeSelection changes, which for sure will trigger child component re-rendering.
Therefore, I preserve this four methods and rename it to CardFobGetStepHelper and pass it to child component.
This PR does not include test changes and is used for discussion.

Update: we only keep getStepHigherThanPosition and getStepLowerThanPosition in a helper object and pass it as functions. The other 4 methods are used as passing values.

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -47,8 +47,8 @@ export class CardFobControllerComponent {
   @Input() axisDirection!: AxisDirection;
   @Input() timeSelection!: TimeSelection;
   @Input() cardFobHelper!: CardFobGetStepFromPositionHelper;
-  @Input() axisPositionFromStartStep!: number;
-  @Input() axisPositionFromEndStep!: number;
+  @Input() startStepAxisPosition!: number;
+  @Input() endStepAxisPosition!: number | null;
   @Input() highestStep!: number;
   @Input() lowestStep!: number;
   @Input() showExtendedLine?: Boolean = false;
@@ -69,16 +69,19 @@ export class CardFobControllerComponent {
 
   getCssTranslatePxForStartFob() {
     if (this.axisDirection === AxisDirection.VERTICAL) {
-      return `translate(0px, ${this.axisPositionFromStartStep}px)`;
+      return `translate(0px, ${this.startStepAxisPosition}px)`;
     }
-    return `translate(${this.axisPositionFromStartStep}px, 0px)`;
+    return `translate(${this.startStepAxisPosition}px, 0px)`;
   }
 
   getCssTranslatePxForEndFob() {
-    if (this.axisDirection === AxisDirection.VERTICAL) {
-      return `translate(0px, ${this.axisPositionFromEndStep}px)`;
+    if (this.endStepAxisPosition === null) {
+      return '';
     }
-    return `translate(${this.axisPositionFromEndStep}px, 0px)`;
+    if (this.axisDirection === AxisDirection.VERTICAL) {
+      return `translate(0px, ${this.endStepAxisPosition}px)`;
+    }
+    return `translate(${this.endStepAxisPosition}px, 0px)`;
   }
 
   startDrag(fob: Fob, affordance: TimeSelectionAffordance) {
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -15,18 +15,16 @@ limitations under the License.
 
 import {
   ChangeDetectionStrategy,
-  ChangeDetectorRef,
   Component,
   ElementRef,
   EventEmitter,
   Input,
   Output,
-  SimpleChanges,
   ViewChild,
 } from '@angular/core';
 import {
   AxisDirection,
-  CardFobAdapter,
+  CardFobGetStepFromPositionHelper,
   TimeSelection,
   TimeSelectionAffordance,
 } from './card_fob_types';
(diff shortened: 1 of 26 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_types.ts
```diff
@@ -29,35 +29,24 @@ export enum AxisDirection {
 }
 
 /**
- * This class is intended to be implemented by the card that has a
+ * These helper functions are intended to be used in the card that has a
```
bmd3k, line 32: "These helper functions are intended to be defined by the card that is a parent of CardFobControllerComponent." ?

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -41,9 +45,12 @@ export class CardFobControllerComponent {
   @ViewChild('endFobWrapper') readonly endFobWrapper!: ElementRef;
   @Input() axisDirection!: AxisDirection;
   @Input() timeSelection!: TimeSelection;
-  @Input() cardAdapter!: CardFobAdapter;
+  @Input() cardFobHelper!: CardFobGetStepFromPositionHelper;
+  @Input() axisPositionFromStartStep!: number;
```
bmd3k, line 49: Consider this wordy but more accurate name: timeSelectionStartStepAxisPosition.
Similarly: timeSelectionEndStepAxisPosition.

or combine them into a single object TimeSelectionAxisPositions: {
  start: number;
  end: number;
}

(I started wondering what the difference is supposed to be between lowestStep/startStep and highestStep/endStep, then realized startStep and endStep were specifically related to TimeSelection).

tensorboard/webapp/widgets/histogram/histogram_card_fob_controller.ts
```diff
@@ -24,30 +30,50 @@ import {TemporalScale} from './histogram_component';
...
+    getStepHigherThanAxisPosition:
+      this.getStepHigherThanAxisPosition.bind(this),
+    getStepLowerThanAxisPosition: this.getStepLowerThanAxisPosition.bind(this),
+  };
+
+  getAxisPositionFromStartStep() {
+    return this.temporalScale(this.timeSelection.start.step);
+  }
+
+  getAxisPositionFromEndStep() {
```
bmd3k, line 62: Note:

@Input() axisPositionFromEndStep!: number;

... does not include `null` in the type. Unfortunately our Angular compiler does not catch this in the OSS code base but we should consider updating the return type to better reflect the possibility.

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
```diff
@@ -40,22 +48,27 @@ class TestableComponent {
 
   @Input() axisDirection!: AxisDirection;
   @Input() timeSelection!: TimeSelection;
-  @Input() cardFobAdapter!: CardFobAdapter;
+  @Input() cardFobHelper!: CardFobGetStepFromPositionHelper;
   @Input() showExtendedLine?: Boolean;
 
   @Input() onTimeSelectionChanged!: (newTimeSelection: TimeSelection) => void;
   @Input() onTimeSelectionToggled!: () => void;
+  @Input() getAxisPositionFromStartStep!: () => number;
```
bmd3k, line 56: Can you change the types of these four new properties from function to plain-old number? And refactor the tests to reflect this?

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
```diff
@@ -83,29 +96,37 @@ describe('card_fob_controller', () => {
...
     });
-    fixture.componentInstance.cardFobAdapter = cardFobAdapter;
+    fixture.componentInstance.cardFobHelper = cardFobHelper;
+    fixture.componentInstance.getAxisPositionFromStartStep =
+      getAxisPositionFromStartStepSpy;
+    fixture.componentInstance.getAxisPositionFromEndStep =
+      getAxisPositionFromEndStepSpy;
+    fixture.componentInstance.getHighestStep = getHighestStepSpy;
+    fixture.componentInstance.getLowestStep = getLowestStepSpy;
+    fixture.componentInstance.cardFobHelper = cardFobHelper;
```
bmd3k, line 129: Note: Duplicate of L122.

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
bmd3k: I assume you could do the following on L123 to 128:
```

    fixture.componentInstance.axisPositionFromStartStep = input.timeSelection.start.step;
    fixture.componentInstance.axisPositionFromEndStep = input.timeSelection.end ? input.timeSelection.end.step : null;
    fixture.componentInstance.highestStep = 4;
    fixture.componentInstance.lowestStep = 0;
```

bmd3k (approved):

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
japie1235813: With assigning to axisPositionFromStartStep directly, the tests are failing. (for example the first test below "vertical dragging" at line 166 and line 182 fails)
After debugging I found the reason is that although the initial value is taken from input.timeSelection, the timeSelection variable got updated from mousemove, but it never updates "`startStepAxisPosition`" and "`getCssTranslatePxForStartFob`" still returns the same initial number.

I think highestStep and lowestStep can be assigned from input but `axisPositionFromStartStep` and `axisPositionFromEndStep` should be kept as function and check the spy. 

Let me know if you have any other suggestions.

bmd3k: Oh I see. Yes, that makes sense! Thanks for explaining.

PR #5809, opened 2022-07-18
Step Selector: Allow mouse to be anywhere while dragging.

JamesHollyer:
It was not a good experience to force users to keep mouse on axis while dragging the fob.  This allows users to mouse mouse anywhere while the fob is being dragged and it will continue dragging until a mouse up event.

* Technical description
The main changes for this are in the card_fob_controller_component.ts file. It is difficult to register for document level mouse events from the html page so we have to do it programatically. Also, the reference for this listener is held on the document object. This causes some issues as it is never removed even when the component is destroyed. Also since the call comes from that object it does not hold a reference to "this". To keep the reference I bound "this" to the call. I also remove the listener when dragging stops so we do not keep adding more and more listeners to the body. This turned out to be a bit tricky as we had to keep a reference to the EventListener and calls to .bind always create a new reference. I had to make that reference of type any to avoid issues with [--strictFunctionTypes](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#strict-function-types) as described nicely in [this stack overflow answer](https://stackoverflow.com/a/47171216/4254870).

Also of note: mouse events registered to the document will continue firing even when the pointer is outside the browser window. Therefore there is no longer a need for mouseleave events as we will continue dragging even off the page and the mouseup will always fire no matter where the mouse is when the user releases their click.

googlers see b/239453022

This also solves the issue that the dragging does not work when steps are too far apart. Googlers see b/238666730

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -2238,7 +2238,6 @@ describe('scalar card', () => {
         let fakeEvent = new MouseEvent('mousemove', {
           clientX: 25 + controllerStartPosition,
           movementX: 1,
-          buttons: 1,
         });
         testController.mouseMove(fakeEvent);
         testController.stopDrag();
@@ -2663,7 +2662,6 @@ describe('scalar card', () => {
         const fakeEvent = new MouseEvent('mousemove', {
           clientX: 25 + controllerStartPosition,
           movementX: 1,
-          buttons: 1,
         });
         testController.mouseMove(fakeEvent);
         testController.stopDrag();
@@ -2721,7 +2719,6 @@ describe('scalar card', () => {
         let fakeEvent = new MouseEvent('mousemove', {
           clientX: 25 + controllerStartPosition,
           movementX: 1,
-          buttons: 1,
         });
         testController.mouseMove(fakeEvent);
         testController.stopDrag();
@@ -2760,7 +2757,6 @@ describe('scalar card', () => {
         fakeEvent = new MouseEvent('mousemove', {
           clientX: 28 + controllerStartPosition,
           movementX: -1,
-          buttons: 1,
         });
         testController.mouseMove(fakeEvent);
         testController.stopDrag();
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -2238,6 +2238,7 @@ describe('scalar card', () => {
         let fakeEvent = new MouseEvent('mousemove', {
           clientX: 25 + controllerStartPosition,
           movementX: 1,
+          buttons: 1,
         });
         testController.mouseMove(fakeEvent);
         testController.stopDrag();
@@ -2662,6 +2663,7 @@ describe('scalar card', () => {
         const fakeEvent = new MouseEvent('mousemove', {
           clientX: 25 + controllerStartPosition,
           movementX: 1,
+          buttons: 1,
         });
         testController.mouseMove(fakeEvent);
         testController.stopDrag();
@@ -2719,6 +2721,7 @@ describe('scalar card', () => {
         let fakeEvent = new MouseEvent('mousemove', {
           clientX: 25 + controllerStartPosition,
           movementX: 1,
+          buttons: 1,
         });
         testController.mouseMove(fakeEvent);
         testController.stopDrag();
@@ -2757,6 +2760,7 @@ describe('scalar card', () => {
         fakeEvent = new MouseEvent('mousemove', {
           clientX: 28 + controllerStartPosition,
           movementX: -1,
+          buttons: 1,
         });
         testController.mouseMove(fakeEvent);
         testController.stopDrag();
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -62,6 +62,11 @@ export class CardFobControllerComponent {
   private currentDraggingFob: Fob = Fob.NONE;
   private affordance: TimeSelectionAffordance = TimeSelectionAffordance.NONE;
 
+  // mouseListener is used to keep a reference to the EventListener used to
+  // track mouse movement in order to remove that listener.
```
bmd3k, line 66: Should stopListener be incorporated into the comment?

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -85,20 +90,15 @@ export class CardFobControllerComponent {
   }
 
   startDrag(fob: Fob, affordance: TimeSelectionAffordance) {
+    document.addEventListener('mousemove', this.mouseListener);
```
bmd3k, line 93: I see RxJs `fromEvent()` used sometimes in the code base. Did you try using it? I don't know if it will make your life easier or not but it could be worth exploring. You can see an example usage here:

https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/widgets/histogram/histogram_component.ts;l=153;drc=c52d1385a5f4b0877b548bb398754adda50cf409

What you have here otherwise seems fine with me.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2229,6 +2229,7 @@ describe('scalar card', () => {
         let fakeEvent = new MouseEvent('mousemove', {
           clientX: 25 + controllerStartPosition,
           movementX: 1,
+          buttons: 1,
```
bmd3k, line 2232: A bunch of references to "buttons" here that I guess you don't need any longer.

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
```diff
@@ -166,7 +166,11 @@ describe('card_fob_controller', () => {
       ).toEqual(1);
 
       fobController.startDrag(Fob.START, TimeSelectionAffordance.FOB);
-      const fakeEvent = new MouseEvent('mousemove', {clientY: 3, movementY: 1});
+      const fakeEvent = new MouseEvent('mousemove', {
+        clientY: 3,
+        movementY: 1,
+        buttons: 1,
```
bmd3k, line 172: Also many references to "buttons" in this file you can get rid of.

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
```diff
@@ -398,6 +418,51 @@ describe('card_fob_controller', () => {
...
+      addListenerSpy.and.callFake((type, listener) => {
+        if (type === 'mousemove') {
+          mouseMoveListener = listener;
+        }
+
+        if (type === 'mouseup') {
+          mouseUpListener = listener;
+        }
+      });
+      document.removeEventListener = removeListenerSpy;
```
bmd3k, line 440: This test has side effects that could impact other tests - modifications to document.addEventListener and document.removeEventListener. This could cause subtle problems in other tests. My recommendation would be to modify document.addEventListener and document.removeEventListener in the beforeEach() method and then clean them back up in the afterEach() method.

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
```diff
@@ -668,6 +757,52 @@ describe('card_fob_controller', () => {
       ).toEqual(3);
       expect(onTimeSelectionChanged).toHaveBeenCalledTimes(0);
     });
+
+    it('adds and removes event listener', () => {
```
bmd3k, line 761: I feel like your first test was sufficient to cover everything. The relevant code isn't different for horizontal vs vertical. If you agree with this feeling, maybe just add a `describe('dragging')` as a sibling to `describe('vertical dragging')` and `describe('horizontal dragging')` and put the one and only copy of this test there.

tensorboard/webapp/widgets/histogram/histogram_card_fob_test.ts
```diff
@@ -163,6 +163,7 @@ describe('HistogramCardFobController', () => {
       const fakeEvent = new MouseEvent('mousemove', {
```
bmd3k, line 163: also some buttons references here that can be removed.

bmd3k (approved):

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
JamesHollyer: Not sure how to clean up in afterEach. However, I started using spyOn which from what I understand automatically removes the spy after the test.

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
JamesHollyer: Done.

tensorboard/webapp/widgets/histogram/histogram_card_fob_test.ts
JamesHollyer: Done.

JamesHollyer (commented):
Thanks Brian

PR #5816, opened 2022-07-21
step selector: add alias in run name if any

japie1235813:
Previously when displaying the run column we display the run name. Yet when comparing multiple experiments we missed the alias number and text. This PR adds it to the table.

Extract alias info from metadata and combine with Run name. A space is added between alias number and text. We do have alias component, however introducing the component adds too much unnecessary complexity (the data_table widget needs to treat it "special" as color run and also customize the background color of alias number). I decide to no involve this component and just simply add alias as part of RUN column.

* Screenshots of UI changes
Before 
Screen Shot 2022-07-21 at 4 15 39 PM
After
Screen Shot 2022-07-21 at 3 32 24 PM

Screen Shot 2022-07-21 at 4 17 35 PM

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -237,10 +237,14 @@ export class ScalarCardComponent<Downloader> {
           datum.points[findClosestIndex(datum.points, startStep)];
         const selectedStepData: SelectedStepRunData = {};
         selectedStepData.COLOR = metadata.color;
+        let alias = '';
+        if (metadata.alias) {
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -237,10 +237,14 @@ export class ScalarCardComponent<Downloader> {
           datum.points[findClosestIndex(datum.points, startStep)];
         const selectedStepData: SelectedStepRunData = {};
         selectedStepData.COLOR = metadata.color;
+        let alias = '';
```
bmd3k, line 240: alias is only used by ColumnHeaders.RUN case, so seems like you could move this code (L240 to L243) into that case block. Keep the variable closer to where it is used.

bmd3k (approved):

PR #5818, opened 2022-07-22
Adds specialized formatter to histogram component.

bileschi:
Using SI notation for histogram bins smaller than 1 causes confusion.  '500m' can mean many things, not just '0.5'.  Googlers see b/238492397

Replaces tick formatter with a 3 way switch.  SI notation for small values, Exponential notation for large values, and natural notation for middle values.

* Screenshots of UI changes

before:  
----

image
image

after:
----

image
image

Ran TensorBoard standalone locally and compared across a very wide range of histogram bin bucket sizes.

It may be possible to update tensorboard/webapp/widgets/histogram/histogram_component.ts and re-use formatters from there.   Did not pursue as it would require extra customization there, and this was deemed sufficiently clear.

### DIFF
--- a/tensorboard/webapp/widgets/histogram/histogram_component.ts
+++ b/tensorboard/webapp/widgets/histogram/histogram_component.ts
@@ -30,6 +30,7 @@ import {takeUntil} from 'rxjs/operators';
 import * as d3 from '../../third_party/d3';
 import {HCLColor} from '../../third_party/d3';
 import {TimeSelection} from '../card_fob/card_fob_types';
+import {formatTickNumber} from './formatter';
 import {
   Bin,
   HistogramData,
@@ -37,7 +38,6 @@ import {
   HistogramMode,
   TimeProperty,
 } from './histogram_types';
-import {formatTickNumber} from './formatter';
 
 type BinScale = d3.ScaleLinear<number, number>;
 type CountScale = d3.ScaleLinear<number, number>;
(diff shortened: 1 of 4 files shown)


rileyajones (approved):
LGTM

tensorboard/webapp/widgets/histogram/histogram_component.ts
```diff
@@ -76,6 +76,34 @@ export interface TooltipData {
...
+        return '0';
+    }
+    const absNum = Math.abs(x as number);
+    if (absNum >= LARGE_NUMBER){
+        return d3LargeFormatter(x);
+    }
+    if (absNum < SMALL_NUMBER){
+        return d3SmallFormatter(x);
+    }
+    return d3MiddleFormatter(x);
```
rileyajones, line 102: This is very nitpicky but it's a little strange that this is ordered `LARGE_NUMBER`, `SMALL_NUMBER`, `MEDIUM_NUMBER`.

Perhaps consider gating around a series minimums.
```suggestion
    if (absNum >= LARGE_NUMBER){
        return d3LargeFormatter(x);
    }
    if (absNum >= SMALL_NUMBER){
        return d3MiddleFormatter(x);
    }
    return d3SmallFormatter(x);
```

If taking this suggestion you might also consider renaming `SMALL_NUMBER` to `MEDIUM_NUMBER`.

bmd3k (commented):
Please add some tests if it is reasonable to do.

bileschi:
Factored out the formatter and added tests.

bmd3k (commented):
A little late but LGTM. Thanks!

PR #5822, opened 2022-07-26
linked time: add affordance for toggling action

japie1235813:
We are interested in learning how often deselect is trigged. However, when deselecting fob in single selection, we triggered toggle action instead of time selection change. We do not know if the toggle action is trigged by deselecting fob or check/uncheck box in settings pane. This PR adds an affordance to the action.

For now toggle action from scalar/histogram only comes from deselecting fob. For simplicity we set the affordance to FOB_DESELECT directly.
When testing [metrics_reducers_test.ts](https://github.com/tensorflow/tensorboard/compare/master...japie1235813:affordanceRefineDeselect?expand=1#diff-25fad7392d0f0699cf58a78c9f214958c60499791a89e4510e2f588c8ea26e67) we don't care about affordance, so just use placeholder `{}`.

* Event fired shown in developer console
Screen Shot 2022-07-26 at 3 33 43 PM
Screen Shot 2022-07-26 at 3 33 33 PM


tensorboard/webapp/metrics/actions/index.ts
```diff
@@ -188,11 +191,17 @@ export const timeSelectionCleared = createAction(
 );
 
 export const linkedTimeToggled = createAction(
-  '[Metrics] Linked Time Enable Toggle'
+  '[Metrics] Linked Time Enable Toggle',
+  props<{
+    affordance?: TimeSelectionToggleAffordance;
```
bmd3k, line 196: Add similar comments as L183/184?

bmd3k (approved):

PR #5823, opened 2022-07-27
ci: update to macos 11

japie1235813:
According to https://github.com/actions/virtual-environments/issues/5583, macos-10.15 is no longer supported and fails our ci flow. Upgrade t o macos-11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,7 @@ jobs:
       fail-fast: false
       matrix:
         mode: ['native']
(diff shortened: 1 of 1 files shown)


bmd3k (commented):
Should we consider an upgrade to macos-12 or even macos-latest?

bileschi (approved):
Approving as I made the exact same proposed change independently.   

Either 11 or 12 is probably fine.  We can upgrade to 12 later if we decide that's valuable.

PR #5829, opened 2022-07-29
Use esbuild for javascript bundling.

bmd3k:

  The upcoming Angular 13 upgrade breaks our custom tf_dev_js_binary rule, which we used for fast, unminified bundling of the tensorboard:dev target.

  One straighforward solution is to move all of our javascript bundling to the esbuild bundler. It is just as fast as our custom bundling in tf_dev_js_binary and much faster than the rollup+terser combination we've been using for bundling our production build. The bundles it generates also tend to be smaller than the bundles we are currently generating.

  By using esbuild in our tf_js_binary rule we can migrate all bundling to this single rule and eliminate tf_dev_js_binary rule. The only difference between a "prod" and "dev" target is whether we choose to minify or not. We can also remove rollup and terser from our project.

  See: https://esbuild.github.io/

  Note that bazel documents how to use esbuild with ts_project but not with ts_library. We managed to get it to work but
  its long-term support is unknown.

  https://www.npmjs.com/package/@bazel/esbuild

  For core TensorBoard app:
  * `bazel run tensorboard` and `bazel run tensorboard:dev`.
  * Simple check of 'Time Series', 'Scalars', 'Images', 'Graphs', 'Distributions', 'Histograms', 'Text', and 'Hparams'. 
  * Set 'pagination limit' in the settings and ensure it is honored by the 'Time Series' and 'Images' plugins - this ensures the angular/polymer bridge is working.
  * Set 'ignore outliers in chart scaling' in both 'Time Series' and 'Scalars' plugins and refresh and verify the changes are honored. - this ensures integration with local storage is still working.
  * Check the 'Debugger' plugin and the Monaco code editor is working within it.
  
  I also verified that other binaries using tf_js_binary rule still seem to work:
  * chart_worker.js
  * tf_graph_app
  * vz_projector
  * tensor_widget (could only verify it builds, couldn't figure out how to render it)

  I checked that bundle times are noticeably faster.
  * for the prod build of tb_webapp_binary.js, the rollup+terser bundling could be 20s to 30s. but with esbuild it is just 1s.
  * for the dev build of tb_webapp_binary.js, there is not really any noticeable performance gain.

  I checked bundle sizes. They actually all end up being smaller:
  * At commit 0f4d6e9761a1cb3d5188ab56e82d2eca85c656a6
  * For the production app with minified binaries:
    * bazel-bin/tensorboard/webapp/tb_webapp_binary.js 4378795 (vs 4464436)
    * bazel-bin/tensorboard/components/polymer3_lib_binary.js 3166751 (vs 3177827)
    * bazel-bin/tensorboard/webapp/index.js 7631949 (vs 7728765)
  * For the dev app with unminified binaries:
    * bazel-bin/tensorboard/webapp/dev_assets/tb_webapp_binary.js 7190752 (vs 26059122)
    * bazel-bin/tensorboard/webapp/dev_assets/polymer3_lib_binary_dev.js 6124163 (vs 8121074)


yarn.lock
```diff
@@ -3588,7 +3480,7 @@ magic-string@0.25.7:
   dependencies:
     sourcemap-codec "^1.4.4"
 
-magic-string@^0.25.0, magic-string@^0.25.7:
```
JamesHollyer, line 3591: What did "magic-string@^0.25.7" even do?

tensorboard/defs/defs.bzl
```diff
@@ -39,61 +35,50 @@ def tf_js_binary(
...
-
-    # For some reason, terser_minified is not visible from other targets. Copy
-    # or re-export seems to work okay.
-    native.genrule(
+    # esbuild is a fast JavaScript bundler[1] appropriate for both production
+    # and development builds.
+    #
+    # Bazel documents[2] how to use esbuild bundling with ts_project but we use
+    # the not-quite-deprecated ts_library rule instead of ts_project. We've
+    # managed to get esbuild working with ts_library but its long-term support
```
JamesHollyer, line 53: Did you have to do anything besides specifying .mjs extensions to get ts_library to work?

JamesHollyer (approved):
+64 lines -442 lines!!! This is awesome! Thanks for figuring this out! 

I had a few questions and one minor nit suggestion. It looks great!

tensorboard/defs/defs.bzl
```diff
@@ -13,14 +13,11 @@
 # limitations under the License.
 """External-only delegates for various BUILD rules."""
 
-load("@npm//@bazel/rollup:index.bzl", "rollup_bundle")
+load("@io_bazel_rules_sass//:defs.bzl", "npm_sass_library", "sass_binary", "sass_library")
```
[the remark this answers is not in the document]
bmd3k: I moved it up to sort the imports alphabetically.

yarn.lock
bmd3k: I don't know but I guess rollup or terser depended on it.

tensorboard/defs/defs.bzl
bmd3k: That's all I was certain was ts_library-specific.

PR #5831, opened 2022-07-29
Support selecting single run using double click.

bmd3k:

  Users have asked for a way to select one and only one item in the Runs list in the Time Series dashboard. That is, select one run while deselecting all the others. This is supported in Scalars dashboad with a radio button but was not ported to Time Series.

  We decided to support this differently in Time Series -- with a double click. Double clicking on the checkbox for any run will select that run while deselecting the remaining runs.

  We listen for dblclick events on the mat-checkbox. Events and actions are triggered up to the reducer layer where just the one run is selected while the other runs are deselected.

  We recognize that this might be a bit difficult to discover and some might consider it an anti-pattern but we think it is the best option for our situation. Some users might even try this out, having learned the pattern from other products.

### DIFF
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
@@ -233,6 +233,7 @@ limitations under the License.
           <mat-checkbox
             [checked]="item.selected"
             (change)="onSelectionToggle.emit(item)"
+            (dblclick)="onSelectionDblClick.emit(item)"
           ></mat-checkbox>
         </span>
         <tb-experiment-alias
--- a/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
+++ b/tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
@@ -92,6 +92,7 @@ import {
   runSelectorRegexFilterChanged,
   runSelectorSortChanged,
   runTableShown,
+  singleRunSelected,
 } from '../../actions';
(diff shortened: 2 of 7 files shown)


tensorboard/webapp/runs/views/runs_table/runs_table_component.ng.html
```diff
@@ -233,6 +233,7 @@
           <mat-checkbox
             [checked]="item.selected"
             (change)="onSelectionToggle.emit(item)"
+            (dblclick)="onSelectionDblClick.emit(item)"
```
japie1235813, line 236: Can we also add a note here that `(dblclick)` will also fire `(change)`?

tensorboard/webapp/runs/views/runs_table/runs_table_test.ts
```diff
@@ -1828,6 +1829,32 @@ describe('runs_table', () => {
...
+      const rows = fixture.nativeElement.querySelectorAll(Selector.ITEM_ROW);
+      const book2 = rows[1].querySelector('mat-checkbox');
+      book2.dispatchEvent(new MouseEvent('dblclick', {relatedTarget: book2}));
+
+      expect(dispatchSpy).toHaveBeenCalledWith(
+        singleRunSelected({
+          runId: 'book2',
+        })
+      );
+    });
```
japie1235813, line 1856: Would it be worth to add a test that click and dblclick event are fired in such order and we get the single selection result?

japie1235813 (approved):

PR #5832, opened 2022-07-29
Feature: Add route specific configuration for replaceState

rileyajones:
Programatic routing should be able to amend the route rather than redirect the user. This comes up in situations like updating route params.

### DIFF
--- a/tensorboard/webapp/app_routing/programmatical_navigation_types.ts
+++ b/tensorboard/webapp/app_routing/programmatical_navigation_types.ts
@@ -24,9 +24,11 @@ export interface NavigateToExperiment {
   routeKind: RouteKind.EXPERIMENT;
   routeParams: ExperimentRouteParams;
   resetNamespacedState?: boolean;
+  replaceState?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/app_routing/programmatical_navigation_types.ts
```diff
@@ -21,12 +21,14 @@ export const NAVIGATION_PROVIDER = new InjectionToken<NavigationLambda[]>(
 );
 
 export interface NavigateToExperiment {
+  replaceState?: boolean;
```
bmd3k, line 24: Mind putting this either directly above or directly below resetNamespacedState?. It's about equal in importance. I would prefer to keep routeKind and routeParams at the top.

bmd3k (approved):

tensorboard/webapp/app_routing/programmatical_navigation_types.ts
rileyajones: Sure, I wasn't sure what the ordering convention was so I was going for alphabetical.

PR #5833, opened 2022-08-01
Bug Fix: Updating typing surrounding `feature_flag_serializer_test.ts`

rileyajones:
Apparently TBCorp has different (and sometimes conflicting) lint rules so `feature_flag_serializer_test` was unable to be imported.

While "updating" the typing there I also cleaned up some of the typing in the related files.

#5717


bmd3k (approved):
Nit: Description says "change variable names" but this is more like "adjust types and change property access".

PR #5834, opened 2022-08-01
Bug Fix: Only override forceSvg from localStorage when the key is present

rileyajones:
We always override the value of `forceSvg` regardless of whether a value is set in localStorage. This is an existing bug that is only coming to light now that we persist feature flags to the queryParams.
#5717 

1) Clear localStorage
2) Visit tensorboard and note that the `forceSVG` query parameter is not set.


JamesHollyer (approved):
LGTM

PR #5837, opened 2022-08-02
Update cardToPinnedCopyCache when resolving unresolved pins.

bmd3k:
Googlers, see b/240338417 "Clicking the refresh UI button unpins pinned graphs".

When a user loads a TensorBoard with pinned cards in the URL then clicks on the TensorBoard reload button, the  pinned cards disappear!

Explanation:
* Soon after page load, when handing tag metadata, unresolved pins are resolved into actual pinned cards. Unfortunately, the cardToPinnedCopyCache property is not provided with the information:
  * https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/store/metrics_reducers.ts;l=565;drc=2f5661647c264a3cd5758a04a7de0beb0c604747
* When user then presses the TensorBoard reload button and then tag metadata is reloaded, the cache is considered empty and the actual set of pinned cards is also cleared:
  * https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/store/metrics_reducers.ts;l=550;drc=2f5661647c264a3cd5758a04a7de0beb0c604747

The solution is to ensure that cardToPinnedCopyCache is also populated when resolving unresolved pins.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -372,6 +372,7 @@ const reducer = createReducer(
       state.cardList,
       state.cardMetadataMap,
       state.cardToPinnedCopy,
+      state.cardToPinnedCopyCache,
       state.pinnedCardToOriginal,
       state.cardStepIndex
     );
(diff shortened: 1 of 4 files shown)


japie1235813 (approved):
The bug was my fault. Thanks for fixing this!

PR #5838, opened 2022-08-02
Update `DashboardDeeplinkProvider` to get feature flag metadata dynamically

rileyajones:
This would make it easier to extend the set of feature flags in TBcorp.

Go to tensorboard with a feature flag overridden by a query parameter.
Assert that the query parameter is not removed.

While changes to TBcorp would be possible without this (by extending and then deduplicating query parameters) this would be much cleaner.

### DIFF
--- a/tensorboard/webapp/routes/dashboard_deeplink_provider.ts
+++ b/tensorboard/webapp/routes/dashboard_deeplink_provider.ts
@@ -19,8 +19,9 @@ import {map} from 'rxjs/operators';
 import {DeepLinkProvider} from '../app_routing/deep_link_provider';
 import {SerializableQueryParams} from '../app_routing/types';
 import {State} from '../app_state';
-import {FeatureFlagMetadataMap} from '../feature_flag/store/feature_flag_metadata';
+import {FeatureFlagMetadataMapType} from '../feature_flag/store/feature_flag_metadata';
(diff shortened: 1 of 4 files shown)


bmd3k (approved):

tensorboard/webapp/routes/dashboard_deeplink_provider.ts
```diff
@@ -45,7 +46,12 @@ const COLOR_GROUP_REGEX_VALUE_PREFIX = 'regex:';
  * Provides deeplinking for the core dashboards page.
  */
 @Injectable()
-export class DashboardDeepLinkProvider extends DeepLinkProvider {
+export class DashboardDeepLinkProvider<
+  T extends FeatureFlags = FeatureFlags
```
JamesHollyer, line 50: what does this do for you? The idea is to be able to pass a new FeatureFlag type that has extra flags in TB Corp? Then update FeatureFlags in OSS afterwards?

JamesHollyer (commented):
Still trying to understand this.

tensorboard/webapp/routes/dashboard_deeplink_provider.ts
rileyajones: This says that it needs to be an object that contains all the keys of OSS FeatureFlags but can also have more (i.e. TB Corp feature flags)

PR #5839, opened 2022-08-03
Step Selector: Add smoothed column in data table when enabled

JamesHollyer:
This change adds the "Smoothed" column to the data table which we expect users to want. It also helps align the data table more closely with the tooltip.

Since this column is more dynamic and should only appear when smoothing is enabled I moved the dataHeaders construction to the container where I made it dependent on the smoothingEnabled$ Observable.

* Screenshots of UI changes
Screen Shot 2022-08-03 at 3 27 56 PM

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -2692,7 +2692,7 @@ describe('scalar card', () => {
       expect(
         scalarCardComponent.componentInstance.getTimeSelectionTableData()[0]
           .SMOOTHED
-      ).toBe(scalarCardComponent.componentInstance.dataSeries[1].points[1].y);
+      ).toBe(6.000000000000001);
     }));
 
     it('does not add smoothed column header when smoothed is disabled', fakeAsync(() => {
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -1123,34 +1123,6 @@ describe('scalar card', () => {
     });
   }));
 
-  it('adds smoothed column header when smoothed is enabled', fakeAsync(() => {
-    store.overrideSelector(selectors.getMetricsScalarSmoothing, 0.1);
-
-    const fixture = createComponent('card1');
-    fixture.detectChanges();
-    const scalarCardComponent = fixture.debugElement.query(
-      By.directive(ScalarCardComponent)
-    );
-
-    expect(scalarCardComponent.componentInstance.dataHeaders).toContain(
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -1122,6 +1123,34 @@ describe('scalar card', () => {
     });
   }));
 
+  it('adds smoothed column header when smoothed is enabled', fakeAsync(() => {
```
japie1235813, line 1126: Should this moved into "getTimeSelectionTableData" section?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -1122,6 +1123,34 @@ describe('scalar card', () => {
...
+
+    const fixture = createComponent('card1');
+    fixture.detectChanges();
+    const scalarCardComponent = fixture.debugElement.query(
+      By.directive(ScalarCardComponent)
+    );
+
+    expect(scalarCardComponent.componentInstance.dataHeaders).toContain(
+      ColumnHeaders.SMOOTHED
+    );
```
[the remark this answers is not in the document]
JamesHollyer: Added test to ensure we use the y value.

JamesHollyer (commented):
Thanks Jie-Wei. Let me know if you see anything else you want to add/change.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2651,6 +2652,62 @@ describe('scalar card', () => {
...
+      );
+
+      expect(scalarCardComponent.componentInstance.dataHeaders).toContain(
+        ColumnHeaders.SMOOTHED
+      );
+
+      expect(
+        scalarCardComponent.componentInstance.getTimeSelectionTableData()[0]
+          .SMOOTHED
+      ).toBe(scalarCardComponent.componentInstance.dataSeries[1].points[1].y);
```
japie1235813, line 2695: To make it clear and simpler in testing, I think it's a good practice to make the input be a simple number or string instead of variable. (constant variable is fine)
Could you replace the variable here with a number (or string)?

JamesHollyer: Yea it was just a weird number so I thought this might make more sense. I am cool using it though. Changed.

japie1235813: If the number is not intuitive, please add comment above to explain. I guess this is caused by values of 1, 20, 30. Yet I cannot recall how is the smooth calculated.

japie1235813 (approved):

PR #5840, opened 2022-08-04
Replace d3.selection.html with .text.

vrana:
Its usage could lead to XSS and it's not compatible with Trusted Types.


bmd3k (approved):
Thanks Jakub. I audited current usage of this function and it looks like we only use text for these menu items.  So the change shouldn't unexpectedly break anything.

PR #5841, opened 2022-08-04
Fix "Card Width" property reset button.

bmd3k:
The reset button for "Card Width" does not effectively reset the property. Pressing it resets it to the value that it was set to when the page was loaded. (So, for example, if the value is 535px on page load, pressing reset will reset it to 535px and not reset the property entirely).

The reasoning is:
* Pressing the reset button will remove the property from state, effectively making it undefined:
  * https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/store/metrics_reducers.ts;l=709;drc=2059b53ef0f6875dbaa8bca4362653c07d0069fd
* The PersistentSettingsDataSource does not update the field in local storage if the value is undefined:
  * https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source.ts;l=87;drc=54be1df836817d1a5a6378902eec23a9fe5cb40d
* And, so, the value will continue to be whatever was stored in local storage on page load. 

We fix the problem by setting the state value to null rather than removing it entirely. This means it will also be set to null in local storage and effectively reset the property.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers.ts
@@ -712,7 +712,7 @@ const reducer = createReducer(
       settingOverrides: {
         ...state.settingOverrides,
         cardMinWidth: null,
-      }
+      },
     };
   }),
   on(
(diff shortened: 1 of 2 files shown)


japie1235813 (approved):
Thanks for the fix. In addition to this PR, do we also need to clean up the value stored in local storage?

bmd3k:
> Thanks for the fix. In addition to this PR, do we also need to clean up the value stored in local storage?

I don't think so. We can't identify or cleanup users who've tried to reset in the past. But the next time they press reset it should just work.

PR #5843, opened 2022-08-04
Fix tests to be compatible with numpy>=1.23.0 and python>=3.9.0.

bmd3k:
Fix tests to be compatible with numpy>=1.23.0 and python>=3.9.0.

It's possible nobody has run these tests locally in a while because I have numpy 1.23.1 and python 3.10.5 installed on my machine and these tests failed for me.

Note that the numpy changes, specifically, were originally proposed internall in cl/465149745, which states:

> NumPy 1.23 removes support for non-tuple indexing of NumPy arrays ([https://numpy.org/devdocs/release/1.23.0-notes.html#expired-deprecations](https://www.google.com/url?sa=D&q=https%3A%2F%2Fnumpy.org%2Fdevdocs%2Frelease%2F1.23.0-notes.html%23expired-deprecations)). The workaround is to convert multidimensional indices to a tuple.

Python 3.9.0 removes collections.Callable. We should now use collections.abc.Callable.

### DIFF
--- a/tensorboard/plugins/metrics/metrics_plugin_test.py
+++ b/tensorboard/plugins/metrics/metrics_plugin_test.py
@@ -16,7 +16,7 @@
 
 
 import argparse
-import collections
+import collections.abc
 import os.path
 
 import tensorflow.compat.v1 as tf1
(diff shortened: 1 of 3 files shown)


yatbear (approved):
Thank you so much!

PR #5844, opened 2022-08-05
Update 2.10 RC with fixes.

yatbear:
Googlers, see b/241285964.


bmd3k (approved):
Thanks! Don't forget to "Rebase and Merge" this one instead of "Squash and Merge".

Note that https://github.com/tensorflow/tensorboard/pull/5839 is probably not necessary given that data table is not yet enabled but including it isn't harmful.

PR #5849, opened 2022-08-05
Update generics in feature_flag_metadata and feature_flag_serializer

rileyajones:
In TBcorp and TBdev it is useful to build feature flag metadata maps based on unions of types. This typing was unnecessarily restrictive and prevented the creation of clean unions.

Make some ugly type casts in internal code?

### DIFF
--- a/tensorboard/webapp/routes/feature_flag_serializer.ts
+++ b/tensorboard/webapp/routes/feature_flag_serializer.ts
@@ -18,10 +18,9 @@ import {
   FeatureFlagMetadataMapType,
   FeatureFlagType,
 } from '../feature_flag/store/feature_flag_metadata';
-import {FeatureFlags} from '../feature_flag/types';
 
-export function featureFlagsToSerializableQueryParams<T extends FeatureFlags>(
-  overriddenFeatureFlags: Partial<FeatureFlags>,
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/routes/feature_flag_serializer.ts
```diff
@@ -18,24 +18,23 @@ import {
...
+        featureFlagMetadataMap[featureFlag as keyof T];
       if (!featureFlagMetadata) {
         // No metadata for this feature flag. Shouldn't happen but we must
         // include the check for the compiler.
         // Return empty item. Will be filtered out.
         return {};
       }
       const key = featureFlagMetadata.queryParamOverride;
-      if (!key || featureValue === undefined) {
+      if (!key || (featureValue as FeatureFlagType) === undefined) {
```
bmd3k, line 37: Out of curiosity, does the compiler actually complain about "featureValue === undefined" without the cast?

bmd3k (approved):

tensorboard/webapp/routes/feature_flag_serializer.ts
rileyajones: No and I'm not sure how this got to be this way, last Friday is a bit of a haze

PR #5851, opened 2022-08-08
linked time: Add value change to data table in range selection

JamesHollyer:
This is the first change which adds a column in the data table for range selection. We plan to have many columns which deal with a range for the launch of linked time. 

This change sets the structure for changing the column headers based on the linked time selection and the structure for having different styles in certain columns of the data table. I used ngSwitch for the style in anticipation for more styles in other columns.

* Screenshots of UI changes
Screen Shot 2022-08-08 at 4 46 41 PM

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
@@ -356,21 +356,6 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
       })
     );
 
-    // this.smoothingEnabled$.pipe(
-    //   map((smoothingEnabled) => {
-    //     const headers = [
-    //       ColumnHeaders.RUN,
-    //       ColumnHeaders.VALUE,
-    //       ColumnHeaders.STEP,
-    //       ColumnHeaders.RELATIVE_TIME,
-    //     ];
-    //     if (smoothingEnabled) {
-    //       headers.splice(1, 0, ColumnHeaders.SMOOTHED);
-    //     }
-    //     return headers;
-    //   })
-    // );
-
     this.dataSeries$ = partitionedSeries$.pipe(
       // Smooth
       combineLatestWith(this.store.select(getMetricsScalarSmoothing)),
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
@@ -416,15 +416,20 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
       this.linkedTimeSelection$,
     ]).pipe(
       map(([smoothingEnabled, timeSelection]) => {
-        const headers = [ColumnHeaders.RUN, ColumnHeaders.STEP];
+        const headers: ColumnHeaders[] = [];
         if (timeSelection === null || timeSelection.endStep === null) {
-          headers.splice(1, 0, ColumnHeaders.VALUE);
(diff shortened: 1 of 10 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -426,6 +411,25 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
       )
     );
 
+    this.columnHeaders$ = combineLatest([
+      this.smoothingEnabled$,
+      this.linkedTimeSelection$,
+    ]).pipe(
+      map(([smoothingEnabled, timeSelection]) => {
+        const headers = [ColumnHeaders.RUN, ColumnHeaders.STEP];
```
bmd3k, line 419: I think it would be clearer and easier to reason out if the code were a bit more verbose and repetitive:

```
const headers = [];
if (timeSelection === null || timeSelection.endStep == = null) {
  // No step selected or single step selected.
  headers.push(ColumnHeaders.RUN);
  if (smoothingEnabled) {
    headers.push(ColumnHeaders.SMOOTHED);
  }
  headers.push(ColumnHeaders.VALUE);
  headers.push(ColumnHeaders.STEP);
  headers.push(ColumnHeaders.RELATIVE_TIME);
} else {
  // Range selected.
  headers.push(ColumnHeaders.RUN);
  headers.push(ColumnHeaders.VALUE_CHANGE);
  headers.push(ColumnHeaders.STEP);
}
return headers;
```

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2456,6 +2456,64 @@ describe('scalar card', () => {
...
+        {
+          COLOR: '#fff',
+          RUN: 'run2',
+          STEP: 1,
+          VALUE_CHANGE: 24,
+        },
+      ]);
+    }));
+
     it('selects closest points to time selection', fakeAsync(() => {
```
bmd3k, line 2517: Should endStep be integrated into this test since there is also logic to handle finding the closets point? Or perhaps a new test for that? (I don't think this is covered in the new test you've already added).

bmd3k (approved):

tensorboard/webapp/widgets/data_table/data_table_component.ng.html
```diff
@@ -31,8 +31,23 @@
             <span [style.backgroundColor]="runData.COLOR"></span>
           </td>
           <ng-container *ngFor="let header of headers;">
-            <td>
-              <span>{{ getFormattedDataForColumn(header, runData) }}</span>
+            <td [ngSwitch]="header">
+              <span *ngSwitchCase="'VALUE_CHANGE'">
```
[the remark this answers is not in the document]
JamesHollyer: Done.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
JamesHollyer: You are right. I was getting too fancy. Done.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
JamesHollyer: Yea, the problem is that currently we do not add that step into the data. So I could test it by ensuring that the VALUE_CHANGE is correct when the selected step is off. But, I think it will be cleaner to add this test when I add the END_STEP value. Which I will do next.

JamesHollyer (commented):
Thanks Brian. I will be sure to add that test in my next PR.

PR #5855, opened 2022-08-09
🍒: Cherry-pick fix that elevates Time Series dashboard for image only events to 2.10 branch

yatbear:
Cherry-picks the fix in #5852.


bmd3k (approved):
Thanks!

PR #5857, opened 2022-08-09
Improve FeatureFlagMetadata type so parse only specified when required.

bmd3k:
In the newly-added FeatureFlagMetadataMap, we are specifying parseValue functions for a number of feature flags that don't need them.

This change improves the typing of FeatureFlagMetadata so we only specify parseValue functions when they are required - that is, when the metadata includes a queryParamOverride value.

This makes the type a bit more strict. This allows us to get rid of the parseBooleanOrNull parser.

### DIFF
--- a/tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
+++ b/tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
@@ -22,14 +22,22 @@ export type FeatureFlagType =
   | null
   | undefined;
 
-export type FeatureFlagMetadata<T> = {
-  defaultValue: T;
-  // The name of the query param users can use to override the feature flag
-  // value. If unspecified then users cannot override the feature flag value.
-  queryParamOverride?: string;
-  // Function that translates a query param value into the feature flag value.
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
```diff
@@ -22,14 +22,22 @@ export type FeatureFlagType =
...
+    }
+  | {
+      defaultValue: T;
+      // Some feature flags can be overridden by query params in the URL. They
+      // should specify the name of the query param here.
+      queryParamOverride: string;
+      // Additionally they should specify a way to parse the query param string
+      // values into the feature flag value.
+      parseValue: (str: string) => T;
+    };
```
rileyajones, line 40: 👏 
This makes sense to me and you're right that this does a good job of removing the need for `parse` functions handling each different type. I shied away from this initially as I thought the complicated typing could be off-putting but seeing it actually implemented it looks cleaner than I had expected.

rileyajones (approved):

PR #5858, opened 2022-08-09
Stop event propagation when starting to drag step selector

JamesHollyer:
Some strange behavior was observed when trying to drag the step selector. The root cause of this behavior was that when clicking and dragging in certain places the mouse was triggering other events(such as trying to move an entire graphic or highlighting text. This PR stops the propagation of those mouse clicks so that only the dragging that we want to occur occurs.

Googlers see b/241921321

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -90,7 +90,14 @@ export class CardFobControllerComponent {
     return `translate(${this.endStepAxisPosition}px, 0px)`;
   }
 
-  startDrag(fob: Fob, affordance: TimeSelectionAffordance) {
+  stopEventPropagation(e: Event) {
+    if (e.stopPropagation) e.stopPropagation();
+    if (e.preventDefault) e.preventDefault();
+    e.cancelBubble = true;
+  }
+
+  startDrag(fob: Fob, affordance: TimeSelectionAffordance, event: MouseEvent) {
+    this.stopEventPropagation(event);
     document.addEventListener('mousemove', this.mouseListener);
     document.addEventListener('mouseup', this.stopListener);
     this.currentDraggingFob = fob;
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -90,7 +90,14 @@ export class CardFobControllerComponent {
     return `translate(${this.endStepAxisPosition}px, 0px)`;
   }
 
-  startDrag(fob: Fob, affordance: TimeSelectionAffordance) {
+  stopEventPropagation(e: Event) {
+    if (e.stopPropagation) e.stopPropagation();
+    if (e.preventDefault) e.preventDefault();
+    e.cancelBubble = true;
```
japie1235813, line 96: Is cancelBubble needed? I saw it is marked as deprecated. What is the reason for adding this in addition to stopPropagation and preventDefault?
https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelBubble

I'm having hard time reproducing the first glitch so not totally sure what would happen if we remove line 96. I will leave this decision to you.

japie1235813 (approved):
Thanks for the investigation and quick fix!

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
JamesHollyer: From what I was reading some old browsers do not have preventDefault or stopPropagation and cancelBubble is the only way to do it. Honestly I think it is only IE. I was back and forth on having it in there or not I do not feel strongly either way.

bmd3k: You can search the rest of the TensorBoard codebase to see the assumptions we make about these methods and properties. You don't need cancelBubble. You can also assume stopPropagation and preventDefault exist. So the code could just be:

```
e.stopPropagation();
e.preventDefault();
```

It would be worth sending a cleanup PR to avoid confusion in the future.

JamesHollyer: Will do. Thanks Brian.

PR #5859, opened 2022-08-10
Update version to 2.10.0

yatbear:


bmd3k (approved):
Nit: Can you change the title of the PR and the resulting commit to be "TensorBoard 2.10.0", like the sample commit from the documentation?

yatbear:
> Nit: Can you change the title of the PR and the resulting commit to be "TensorBoard 2.10.0", like the sample commit from the documentation?

Done.

PR #5861, opened 2022-08-10
Update notebook docs with Time Series as default dashboard. 

bmd3k:
With the TensorBoard 2.10 release the "Time Series" dashboard will be the default dashboard for logs with scalars, histograms, or images events. We are updating external-facing documentation to reflect this change.

This change updates notebook documentation (ie documentation runnable in Colab and Jupyter) to use "Time Series" dashboard where previously it used "Scalars" or "Images" dashboards. I could not find any documentation that uses the "Histograms" dashboard.

Most notable changes:
* Updated a number of images. The images being replaced have a variety of sizes but I've mostly standardized these images on 1600x800.
  * This is partially because I don't have a monitor that can take screenshots that are >3000 pixels in width and partially because the documentation that shows the images doesn't show the images wider than 851 pixels, as far as I can tell (example: https://www.tensorflow.org/tensorboard/get_started).
  * The one exception to the standard size is notebooks_tensorboard_tall.png, which has to be 1000 pixels in height to match the example notebook configuration.
* Some of the models being created now give explicit names to the layers. This is so that the tags displayed in the Time Series dashboard maintain a logical ordering of scalars ahead of histograms and images.


bileschi (approved):
LGTM - Thanks!

PR #5864, opened 2022-08-10
Linked Time: Add Start and End Step values to data table

JamesHollyer:
When we launch linked time we will be allowing a range selection. The data table needs to be able to handle range values when in that mode. This change adds the Start Step and End Step columns so the user knows which step is selected for each run.

* Screenshots of UI changes
Screen Shot 2022-08-10 at 4 18 13 PM

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -281,19 +281,16 @@ export class ScalarCardComponent<Downloader> {
                 continue;
               }
               selectedStepData.VALUE_CHANGE =
-                closestEndPoint!.value - closestStartPoint.value;
+                closestEndPoint.value - closestStartPoint.value;
               continue;
             case ColumnHeaders.START_STEP:
-              if (!closestEndPoint) {
-                continue;
-              }
               selectedStepData.START_STEP = closestStartPoint.step;
               continue;
             case ColumnHeaders.END_STEP:
               if (!closestEndPoint) {
                 continue;
               }
-              selectedStepData.END_STEP = closestEndPoint?.step;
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -283,6 +283,18 @@ export class ScalarCardComponent<Downloader> {
               selectedStepData.VALUE_CHANGE =
                 closestEndPoint!.value - closestStartPoint.value;
               continue;
+            case ColumnHeaders.START_STEP:
+              if (!closestEndPoint) {
```
bmd3k, line 287: closestEndPoint is not used in the calculation for START_STEP. Is it necessary to check it here?

The only reason I could imagine it being here is because its shorthand for "This is a range selection property only" but the responsibility for deciding that is actually scalar_card_container's when it is calculating columnHeaders$.

bmd3k (approved):

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
JamesHollyer: I agree this should not be there.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
```diff
@@ -283,6 +283,18 @@ export class ScalarCardComponent<Downloader> {
...
+                continue;
+              }
+              selectedStepData.START_STEP = closestStartPoint.step;
+              continue;
+            case ColumnHeaders.END_STEP:
+              if (!closestEndPoint) {
+                continue;
+              }
+              selectedStepData.END_STEP = closestEndPoint?.step;
```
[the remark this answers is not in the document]
JamesHollyer: Removed.

PR #5866, opened 2022-08-11
scalar: fix scalar_demos

yatbear:
Resets the default writer every time before writing scalar summaries, to fix the error: https://github.com/tensorflow/tensorflow/issues/25707#issuecomment-1212190906
```
Node: 'temperature/current/write_summary'
Resource localhost/_AnonymousVar0/N10tensorflow22SummaryWriterInterfaceE does not exist.
	 [[{{node temperature/current/write_summary}}]] [Op:__inference_one_step_62]
```

Googlers, see b/239742608.

Tested locally:
image

Also tested generating demo data internally in cl/466983681.

#onduty

### DIFF
--- a/tensorboard/plugins/scalar/scalars_demo.py
+++ b/tensorboard/plugins/scalar/scalars_demo.py
@@ -28,12 +28,13 @@ STEPS = 1000
 
 
 @tf.function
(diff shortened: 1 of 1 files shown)


nfelt (approved):
Thanks!

PR #5868, opened 2022-08-11
Feature: Add NgRx Actions, Reducers, Selectors for the Feature Flag Modal

rileyajones:
### Motivation for features / changes
We want to have a modal to control the state of feature flags and persist their state to localStorage.
Creating such a modal is a large change and so, I'm breaking its creation into two PRs, this one and #5800.

This PR is just intended to prepare the NgRx and data source to be used the ui in #5800.

### Technical description of changes
There are existing actions to handle getting and setting feature flag overrides, but none to remove an override. To resolve this, I'm adding two actions, one to reset a set number of overrides, and another to reset all overrides. 

I'm also adding a selector to get the default state of all the feature flags. This should allow the modal to work in other flavors of TensorBoard which have different sets of feature flags without worrying about their specifics.

* Screenshots of UI changes
No ui changes yet

Here's an example of what it may look like
image

image


tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
```diff
@@ -43,3 +43,16 @@ export const featureFlagOverrideChanged = createAction(
     flags: Partial<FeatureFlags>;
   }>()
 );
+
+export const resetFeatureFlagOverrides = createAction(
```
JamesHollyer, line 47: I have been told the name of the action should state what action has been taken as opposed to the effect that will take place. This is a subtle difference. Maybe change this to FlagSetToDefault or something along those lines.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
```diff
@@ -43,3 +43,16 @@ export const featureFlagOverrideChanged = createAction(
...
+  '[FEATURE FLAG] Resetting feature flag overrides',
+  props<{
+    flags: Array<keyof FeatureFlags>;
+  }>()
+);
+
+export const resetAllFeatureFlagOverrides = createAction(
```
JamesHollyer, line 54: Same naming issue here. Maybe FeatureFlagResetClicked?

tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
```diff
@@ -95,6 +98,32 @@ export class FeatureFlagEffects {
...
   );
 
+  readonly resetFeatureFlagOverrides$ = createEffect(
+    () =>
+      this.actions$.pipe(
+        ofType(resetFeatureFlagOverrides),
+        tap(({flags}) => {
+          flags.forEach((flag) => {
+            this.dataSource.resetPersistedFeatureFlag(
+              flag as keyof FeatureFlags
```
JamesHollyer, line 108: This type casting does not seem necessary. The action defines this already.  Obviously if the compiler is complaining then I guess we have to leave it. I just wanted to make sure this was the case.

JamesHollyer (commented):
I still need to review the tests. Here is what I have so far.

tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
rileyajones: Good call, I think this was necessary in an older version.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
rileyajones: What do you think of `featureFlagOverridesReset`?

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
```diff
@@ -455,5 +455,33 @@ describe('tb_feature_flag_data_source', () => {
         );
       });
     });
+
+    describe('resetAllPersistedFeatureFlags', () => {
+      it('removes entry from localStorage', () => {
+        const mockStorage: Record<string, string> = {};
+        spyOn(localStorage, 'getItem').and.callFake((key: string) => {
```
bmd3k, line 462: If James also prefers this test pattern (essentially using a fake instead of a mock) then it could be worth writing a cleanup PR to rewrite the existing tests to use this pattern, too.

tensorboard/webapp/feature_flag/views/types.ts
```diff
@@ -0,0 +1,32 @@
+/* Copyright 2022 The TensorFlow Authors. All Rights Reserved.
```
bmd3k, line 1: I'd recommend deferring this particular file to a future PR where you actually introduce the view. Doesn't seem like the types in this file are used by any of the other files in this PR. Its difficult to judge them without any additional context.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
```diff
@@ -43,3 +43,16 @@ export const featureFlagOverrideChanged = createAction(
     flags: Partial<FeatureFlags>;
   }>()
 );
+
+export const featureFlagOverridesReset = createAction(
+  '[FEATURE FLAG] Resetting feature flag overrides',
```
bmd3k, line 48: Nit: Match tense of other actions "Resetting" => "Reset".

And the other one, too.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
```diff
@@ -43,3 +43,16 @@ export const featureFlagOverrideChanged = createAction(
...
+  props<{
+    flags: Array<keyof FeatureFlags>;
+  }>()
+);
+
+export const featureFlagsOverridesAllSetToDefault = createAction(
+  '[FEATURE FLAG] Resetting all feature flag overrides',
+  // NgRx does not allow actions without props.
```
bmd3k, line 56: I don't think that's true.

Here are some actions without props:
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/core/actions/core_actions.ts;l=39;drc=367839f30de3968eff5fecb8b38e1d5782434b6b

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
```diff
@@ -43,3 +43,16 @@ export const featureFlagOverrideChanged = createAction(
     flags: Partial<FeatureFlags>;
   }>()
 );
+
+export const featureFlagOverridesReset = createAction(
```
bmd3k, line 47: I'm a little suspicious about featureFlagOverrideChanged and featureFlagOverridesReset living side x side. It either points to something wrong with your UI or something wrong with how you're modeling actions.

I am guessing that:
* "featureFlagOverrideChanged" is fired by clicking on some sort of "Save" button at the bottom of the dialog.
* "featureFlagsOverridesAllSetToDefault" is fired by clickign on some sort of "Reset All" button at the bottom of the dialog.

Is that right? If so, where does "featureFlagOverridesReset" fit in? Is there a third "Reset Selected" button or similar? That doesn't seem quite right.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
```diff
@@ -43,3 +43,16 @@ export const featureFlagOverrideChanged = createAction(
...
+  '[FEATURE FLAG] Resetting feature flag overrides',
+  props<{
+    flags: Array<keyof FeatureFlags>;
+  }>()
+);
+
+export const featureFlagsOverridesAllSetToDefault = createAction(
```
bmd3k, line 54: What is the difference between "Reset" and "SetToDefault"? If they are the same thing, can we pick one name and use it consistently?

tensorboard/webapp/feature_flag/effects/feature_flag_effects.ts
```diff
@@ -95,6 +97,30 @@ export class FeatureFlagEffects {
     {dispatch: false}
   );
 
+  readonly resetFeatureFlagOverrides$ = createEffect(
```
bmd3k, line 100: Can you add tests for the new effects?

tensorboard/webapp/feature_flag/store/feature_flag_reducers.ts
```diff
@@ -43,6 +44,34 @@ const reducer = createReducer<FeatureFlagState>(
       },
     };
   }),
+  on(actions.featureFlagOverrideChanged, (state, newOverrides) => {
```
bmd3k, line 47: Could you add a test for featureFlagOverrideChanged?

bmd3k (commented):
Looks pretty good. I feel like there is something slightly off with how you've modeled your Actions (or I misunderstand the UI). And there are some missing tests.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
JamesHollyer: I would prefer to not have a bunch of test infrastructure which simulates local storage and just test that the expected calls are made.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
JamesHollyer: featureFlagOverrideChanged is used when the user sets a flag to override as true or as false. However, the Partial<FeatureFlags> is not able to indicate that a flag is set to default. So featureFlagOverridesReset handles the setting to default.

I was under the impression we would save these individually as the user changes the setting(ie there is no "save" button). However, if we do have a save button then featureFlagOverrideChanged(which should be named featureFlagOverrideSaved in this case) should be all that is necessary and the flags that are missing from the Partial<FeatureFlags> object can all be set to default.

I think I prefer the option where we call these actions on individual changes and we do not require a save button.

JamesHollyer: For a bit of background I wanted to make featureFlagOverrideChanged have a single key and value pair but I could not get the type to work the way I wanted. We decided to use Partial<FeatureFlags> as a compromise.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
JamesHollyer: +1

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
bmd3k: Thanks for the explanation, that makes sense. Yes agreed that it would be fine/preferable to omit the "Save" button.

I assumed there is a "Save" button because these actions specify lists of flags. Should they instead specify a single flag?

```
props<{
  flag: keyof FeatureFlags
}>()
```

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
rileyajones: 🤯 that seems really obvious seeing it now but I couldn't figure it out.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
rileyajones: Consistency in inconsistence

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
rileyajones: I was trying to keep my new actions consistent with the existing ones. I believe the logic for using `Partial<FeatureFlags>` originally was to persist the key and value of the flag which lead to otherwise complicated typing.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
rileyajones: I felt like my version was a bit more thorough, but I've gone ahead and refactored it to make it consistent with the existing tests.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
```diff
@@ -45,14 +45,12 @@ export const featureFlagOverrideChanged = createAction(
...
   props<{
     flags: Array<keyof FeatureFlags>;
   }>()
 );
 
-export const featureFlagsOverridesAllSetToDefault = createAction(
-  '[FEATURE FLAG] Resetting all feature flag overrides',
-  // NgRx does not allow actions without props.
-  (props = undefined) => props
+export const featureFlagOverridesAllReset = createAction(
```
JamesHollyer, line 54: nit: this reads wrong to me. Slight preference to allFeatureFlagOverridesReset

JamesHollyer (approved):
LGTM

tensorboard/webapp/feature_flag/store/feature_flag_reducers_test.ts
```diff
@@ -108,6 +108,149 @@ describe('feature_flag_reducers', () => {
...
+      const prevState = buildFeatureFlagState();
+      const nextState = reducers(
+        prevState,
+        actions.featureFlagOverridesReset({flags: []})
+      );
+      // Intentionally using toBe rather than toEqual to ensure the object is the SAME object (not a copy).
+      expect(nextState).toBe(prevState);
+    });
+
+    it('does not modify previous state', () => {
```
bmd3k, line 163: FWIW we shouldn't be modifying previous state in any of our reducers but we never test for it. You could consider dropping this test.

tensorboard/webapp/feature_flag/store/feature_flag_reducers_test.ts
```diff
@@ -108,6 +108,149 @@ describe('feature_flag_reducers', () => {
...
+        prevState,
+        actions.featureFlagOverridesReset({flags: ['inColab']})
+      );
+      expect(nextState.flagOverrides).toEqual({});
+      expect(nextState.defaultFlags).toBe(prevState.defaultFlags);
+    });
+  });
+
+  describe('#featureFlagsOverridesAllSetToDefault', () => {
+    it('does not modify previous state', () => {
```
bmd3k, line 215: Same as above: we don't usually test for this.

tensorboard/webapp/webapp_data_source/tb_feature_flag_data_source_test.ts
bmd3k: FWIW I preferred the use of fakes but +1 for consistency.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
JamesHollyer: I am ok with this change if you feel strongly about it. We had discussed this as an option. The problem is that there is no way to guarantee the FeatureFlagType(which is basically a big OR statement with all the possible types) matches the type associated with the given key. It was all a lot cleaner and more strongly typed when we just used the Partial. I see the confusion this causes but, I still have a slight preference to the way it is now.

tensorboard/webapp/feature_flag/store/feature_flag_reducers_test.ts
rileyajones: I'm aware that the previous state should never be modified but I figured it was worth testing. I guess if that's opposed to convention though I'll remove the test.

tensorboard/webapp/feature_flag/actions/feature_flag_actions.ts
bmd3k: Yes, FeatureFlagType is unfortunately necessary to workaround some (relatively minor) limitations in the TypeScript type system.

I think in this case we should prefer the more-accurate data model over more-accurate type checks.

Issue #1803, opened 2019-02-04
Histogram binning and plotting useless and deceptive in some cases


mkillinger:
First of all, I think Tensorboard is really great and important UI for real-time feedback on model training, but I run into a few issues with accuracy of the displayed information.

There are two related "sub-issues", the first is much more important:
1.  It seems there are cases of distribtutions for which the resulting histogram bins don't show any discriminating information any more. Example:
True distribution, computed offline:
![true_dist](https://user-images.githubusercontent.com/47332575/52228926-63f19f80-2868-11e9-9919-145a714958c7.png)
Tensorboad histogram:
![tensorboard_hist](https://user-images.githubusercontent.com/47332575/52229189-2e998180-2869-11e9-94af-430eaaf54a27.png)
The visualization is exactly flat/uniform which would indicate a serious issue with training the model. However, the actual distribution is more or less Gaussian around 0.97, so in that sense the tensorboard histogram is deceptive and not showing any useful information. I could imagine that this issue is caused by the fact that the variance is small compared to the mean - but a reasonable scaling/binning method should be robust against that, it's not like the absolute values in this example are completely ill-posed.

2. This issue concerns the "style" of the visualization. If it's supposed to be a histogram, why is it shown as piece-wise linear connected lines and not as a bar plot?
If a continuous representation is preferred for aesthetic or other reasons, the right thing would be a kernel density estimation (which for scalar data should not be super expensive). But a piece-wise linear visualization is confusing and introducing strange artifacts. One of this artifacts can be seen when using the "overlay" mode for showing the time evolution of histograms:
![overlay](https://user-images.githubusercontent.com/47332575/52229826-b2a03900-286a-11e9-8b7f-bbd3c5a122c7.png)
The alternation of constant/horizontal sections and oblique interpolation lines creates a lot of visual clutter that makes it hard to see the actual time evolution of from one histogram to the next. A smooth KDE visualization would presumably be much easier to visually parse, but a plain bar plot might also be easier to understand due to it's simplicity.

A side note, unrelated to histograms:
The moving average smoothing function for scalar data should convert the user-provided raw smoothingWeight to per-time-series parameter that takes into account the sampling rate of each time-series displayed. The training loss sampling rate is often orders of magnitude larger than the sampling rate of evaluator jobs. It is absolutely counter-intuitive to use the "point-to-point" moving average with the same smoothingWeight for those different time-series.
[...]

PR #5871, opened 2022-08-12
docs: add a note about histogram rebinning and "smoothing"

yatbear:
See issue https://github.com/tensorflow/tensorboard/issues/1803 for details.

#onduty

### DIFF
--- a/tensorboard/plugins/histogram/README.md
+++ b/tensorboard/plugins/histogram/README.md
@@ -4,6 +4,8 @@ The TensorBoard Histogram Dashboard displays how the distribution of some
 `Tensor` in your TensorFlow graph has changed over time. It does this by showing
 many histograms visualizations of your tensor at different points in time.
 
+Note that due to frontend limitation and for visualization purpose, by default the histogram data would be rebinned into 30 bins using linear binning, where `y` counts from the original input bins are distributed among the new (increasing and non-overlapping) bins based on the amount of overlap (see existing [issue](https://github.com/tensorflow/tensorboard/issues/1803)).
+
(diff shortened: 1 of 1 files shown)


tensorboard/plugins/histogram/README.md
```diff
@@ -4,6 +4,8 @@ The TensorBoard Histogram Dashboard displays how the distribution of some
 `Tensor` in your TensorFlow graph has changed over time. It does this by showing
 many histograms visualizations of your tensor at different points in time.
 
+Note that due to frontend limitation and for visualization purpose, by default the histogram data would be rebinned into 30 bins using linear binning, where `y` counts from the original input bins are distributed among the new (increasing and non-overlapping) bins based on the amount of overlap (see existing [issue](https://github.com/tensorflow/tensorboard/issues/1803)).
```
nfelt, line 7: Hmm, I'd skip "frontend limitation" (since we _could_ change this code) and "by default" (since there's no way to disable it), and maybe add some explicit acknowledgement that this can lead to distortions or inaccuracy?  E.g. something like:

> Note that for visualization purposes, histogram data is rebinned into 30 linearly spaced bins, where `y` counts from the original input bins at each step are linearly distributed into the new bins, which are shared across all steps. For some distributions, this can result in significant distortions (see existing [issue]).

nfelt (approved):
Thanks!

tensorboard/plugins/histogram/README.md
yatbear: Modified. Thanks!

PR #5875, opened 2022-08-15
Remove Time Series Promotion logic and feature flag.

bmd3k:
Time Series promotion box is only used internally but Time Series has been launched internally for over a year (maybe two?). Remove the promotion butter bar, the backing logic, and the feature flag.


tensorboard/webapp/metrics/store/metrics_selectors_test.ts
```diff
@@ -863,29 +863,16 @@ describe('metrics selectors', () => {
...
-        buildMetricsState({promoteTimeSeries: false})
-      );
-      expect(selectors.getPromoteTimeSeries(state)).toEqual(false);
-    });
-  });
-
   describe('#isMetricsSettingsPaneOpen', () => {
     beforeEach(() => {
-      selectors.getPromoteTimeSeries.release();
+      selectors.isMetricsSettingsPaneOpen.release();
```
JamesHollyer, line 868: So was this test incorrect before?

tensorboard/webapp/metrics/views/metrics_container_test.ts
```diff
@@ -12,87 +12,25 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-    store = TestBed.inject<Store<State>>(Store) as MockStore<State>;
-    store.overrideSelector(getIsTimeSeriesPromotionEnabled, false);
-    store.overrideSelector(getPromoteTimeSeries, false);
   });
 
   it('renders', () => {
     const fixture = TestBed.createComponent(MetricsDashboardContainer);
     fixture.detectChanges();
-
     expect(fixture.debugElement.query(By.css('runs-selector'))).not.toBeNull();
```
JamesHollyer, line 34: This may not be in the scope of this PR but, should we also test that the metrics-main-view renders as well.

JamesHollyer (approved):
LGTM. I had a couple comments but, it all seems in order.

tensorboard/webapp/metrics/views/metrics_container_test.ts
bmd3k: To be honest I was considering deleting this test since there isn't really any logic remaining. I kept this test just as a sanity that it renders anything at all. I don't think it really matters what it looks up - just that it can look up anything.

tensorboard/webapp/metrics/store/metrics_selectors_test.ts
bmd3k: That's right. Looks like a copy & paste error.

Issue #5725, opened 2022-05-26
cleanup: drop Python 3.6 and adopt dataclasses


nfelt:
We should consider dropping Python 3.6 support (i.e. bump [this line](https://github.com/tensorflow/tensorboard/blob/28ff87e12c60fe1daafb9f17c0cc86d23a04e79d/tensorboard/pip_package/setup.py#L67)).

Internally, we've been on 3.7 for a while and recently moved to 3.9. In open source, 3.6 is already end-of-lifed as of the end of 2021: https://peps.python.org/pep-0494/#lifespan

Dropping 3.6 will let us start using 3.7-only features, of which the most useful would be dataclasses: https://docs.python.org/3.7/library/dataclasses.html

There are a number of classes in the codebase that could be converted to dataclasses for a more compact definition and less boilerplate, including:

- most uses of namedtuple: https://cs.opensource.google/search?q=namedtuple%5C(&sq=&ss=tensorflow%2Ftensorboard
- [`RequestContext`](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/context.py)
- [`FrontendMetadata`](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/plugins/base_plugin.py;l=104;drc=a8276a7682f1838994981c06ab6733f376e02663)
- [`TBContext`](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/plugins/base_plugin.py;l=229;drc=a8276a7682f1838994981c06ab6733f376e02663)

- DataProvider API helper classes, e.g. [ExperimentMetadata] and friends (https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/data/provider.py;l=362;drc=bd1f897f523cc7b02532ea9cadc90a938c8f952b)

One possible tradeoff is that dataclasses with default values for any properties don't support `__slots__` until 3.10: https://stackoverflow.com/questions/50180735/how-can-dataclasses-be-made-to-work-better-with-slots/69661861#69661861  But many of our existing classes don't use defaults anyway, and for those that do, they might not really need to use `__slots__`.

PR #5878, opened 2022-08-15
build: explicitly require Python 3.7+

groszewn:
Python 3.6 is well past its [EOL](https://peps.python.org/pep-0494/#lifespan), so we should require 3.7+.

Contributes to #5725

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,7 +214,7 @@ jobs:
         # flake8 should run on each Python version that we target,
         # because the errors and warnings can differ due to language
         # changes, and we want to catch them all.
-        python_version: ['3.6', '3.7']
+        python_version: ['3.7', '3.8']
     steps:
       - uses: actions/checkout@50fbc622fc4ef5163becd7fab6573eac35f8462e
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -214,7 +214,7 @@ jobs:
         # flake8 should run on each Python version that we target,
         # because the errors and warnings can differ due to language
         # changes, and we want to catch them all.
-        python_version: ['3.6', '3.7']
+        python_version: ['3.7', '3.8']
```
nfelt, line 217: Want to add 3.9 and 3.10 in addition to 3.8?  We should be 3.10 compatible as of https://github.com/tensorflow/tensorboard/pull/5793, and while obviously we haven't been good about keeping this line up to date, the comment above has a rationale for running at least flake8 for every supported Python version.

If you do that we could also add 3.9 and 3.10 tags in both setup.py files as well.

nfelt (approved):
Thanks for doing this!

.github/workflows/ci.yml
groszewn: Good catch, updated!

PR #5880, opened 2022-08-16
Bug Fix: Fix centering of linked time fob dismiss button

rileyajones:
The dismiss button for the linked time fob was off center when the chrome font size was greater than medium.

The reason it was off center was because the size of the container was hard coded to 11px and the font size property was not set.

* Screenshots of UI changes
Before
Screen Shot 2022-08-16 at 1 31 50 PM

After
Screen Shot 2022-08-16 at 1 32 10 PM

1) go to Chrome settings(chrome://settings/)
2) select appearance on the left side menu
3) Change the "Font size" option to "large"
4) open tensorboard at http://localhost:6006/?enableLinkedTime=true&enableDataTable=true#timeseries
5) observe the fob

Allow the size of the fob to expand along with browser font size by replacing the hard coded pixel values.

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_component.scss
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_component.scss
@@ -49,6 +49,7 @@ button {
   padding: 0;
   border: 0;
   border-radius: 50%;
+  font-size: 11px;
   width: 11px;
   height: 11px;
   background-color: inherit;


JamesHollyer (approved):
LGTM

rileyajones:
We don't really scale anything else so it doesn't look good. Someday it would be nice to have a larger refactor of font sizes to make this sort of thing work.

PR #5881, opened 2022-08-17
Step Selector: Stop unnecessarily rerendering data table.

JamesHollyer:
The launch of the data table caused a significant performance reduction when scrolling the mouse a crossed the scalar card. This is because we created a new object for the data being passed to the table. Therefore even if the information did not change(which is the case while mousing over the card) the table re-rendered.  This change ensures the same object is passed when nothing changes.

Using the "Paint flashing" option in Rendering in the chrome dev tools I proved that the re-rendering no longer occurs. Also, even without the tool you can tell the performance is much better.

googlers see b/242572017

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_component.ts
@@ -72,6 +72,7 @@ export class ScalarCardComponent<Downloader> {
   readonly DataLoadState = DataLoadState;
   readonly RendererType = RendererType;
   readonly ScaleType = ScaleType;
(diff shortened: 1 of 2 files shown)


bmd3k (commented):
I believe Angular has a natural way of handling this kind of thing: refactor the bits you don't want rerendered into their own components and pass them the subset of data that they need. The refactored components are isolated from unrelated data changes and will be rerendered less frequently.

I tried it out here:
https://github.com/tensorflow/tensorboard/commit/a34a951e7efc5415ae2c757f72bd7d8a9e75b1c9

JamesHollyer:
closing in favor of the solution created in https://github.com/tensorflow/tensorboard/pull/5887

PR #5882, opened 2022-08-17
Limit feature flags sent to server.

bmd3k:
In all requests from the TensorBoard application to the HTTP server we currently populate the "X-TensorBoard-Feature-Flags" header with the entire set of feature flags and values. We decided we only want to send the set of feature flags that are relevant to the HTTP Server and whichever DataProvider implementation backs it.

This change limits the feature flags in two ways:
* Introduce the "sendToServerWhenOverridden" metadata field. Only flags that are marked as "sendToServerWhenOverridden" can be sent to the HTTP Server.
* Only send flags which have been overriden by the user.

So for a flag value to be included in "X-TensorBoard-Feature-Flags" it must both have "sendToServerWhenOverridden" set to true and it must be overridden by the user.

Note that none of the flags in Standalone TensorBoard are marked as sendToServerWhenOverridden - the use case is currently internal.

### DIFF
--- a/tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
+++ b/tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
@@ -37,6 +37,9 @@ export type FeatureFlagMetadata<T> =
       // Additionally they should specify a way to parse the query param string
       // values into the feature flag value.
       parseValue: (str: string) => T;
+      // Some overridable feature flags should be sent to the server if the user
+      // has specified an override value.
+      sendToServer?: boolean;
     };
 
 export type FeatureFlagMetadataMapType<T> = {
--- a/tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
+++ b/tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
@@ -37,9 +37,9 @@ export type FeatureFlagMetadata<T> =
       // Additionally they should specify a way to parse the query param string
       // values into the feature flag value.
       parseValue: (str: string) => T;
-      // Some overridable feature flags should be sent to the server if the user
-      // has specified an override value.
-      sendToServer?: boolean;
+      // Indicates that the feature flag and value should be sent to the server
+      // if the user has specified an override value.
+      sendToServerWhenOverridden?: boolean;
     };
 
 export type FeatureFlagMetadataMapType<T> = {
--- a/tensorboard/webapp/feature_flag/store/feature_flag_selectors.ts
+++ b/tensorboard/webapp/feature_flag/store/feature_flag_selectors.ts
@@ -58,6 +58,24 @@ export const getFeatureFlagsMetadata = createSelector(
   }
 );
 
+export const getFeatureFlagsToSendToServer = createSelector(
+  selectFeatureFlagState,
+  (state: FeatureFlagState): Partial<FeatureFlags> => {
+    let featureFlagsToSendToServer: Partial<FeatureFlags> = {};
+    for (const entry in state.flagOverrides) {
+      const entryMetadata = state.metadata[entry as keyof FeatureFlags];
+      if (!entryMetadata.queryParamOverride || !entryMetadata.sendToServer) {
+        continue;
+      }
+      featureFlagsToSendToServer = {
+        ...featureFlagsToSendToServer,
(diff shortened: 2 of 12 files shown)


tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
```diff
@@ -37,6 +37,9 @@ export type FeatureFlagMetadata<T> =
       // Additionally they should specify a way to parse the query param string
       // values into the feature flag value.
       parseValue: (str: string) => T;
+      // Some overridable feature flags should be sent to the server if the user
+      // has specified an override value.
+      sendToServer?: boolean;
```
rileyajones, line 42: nitpick: `sendToServer` makes me think it will always be sent to the server, not just when it is overridden.
```suggestion
      sendToServerWhenOverridden?: boolean;
```

tensorboard/webapp/feature_flag/store/feature_flag_selectors.ts
```diff
@@ -58,6 +58,24 @@ export const getFeatureFlagsMetadata = createSelector(
...
+    let featureFlagsToSendToServer: Partial<FeatureFlags> = {};
+    for (const entry in state.flagOverrides) {
+      const entryMetadata = state.metadata[entry as keyof FeatureFlags];
+      if (!entryMetadata.queryParamOverride || !entryMetadata.sendToServer) {
+        continue;
+      }
+      featureFlagsToSendToServer = {
+        ...featureFlagsToSendToServer,
+        [entry]: state.flagOverrides[entry as keyof FeatureFlags],
+      };
```
rileyajones, line 73: This looks a lot like the `reduce` spread anti-pattern.
https://www.richsnapp.com/article/2019/06-09-reduce-spread-anti-pattern
```suggestion
      featureFlagsToSendToServer[entry] = state.flagOverrides[entry as keyof FeatureFlags];
```

tensorboard/webapp/feature_flag/store/feature_flag_metadata.ts
bmd3k: Done.

tensorboard/webapp/feature_flag/store/feature_flag_selectors.ts
bmd3k: Unfortunately not totally that simple. Making that change leads to the following error:

Type 'number | boolean | string[] | null | undefined' is not assignable to type 'never'.

The anti-pattern you identifeid here was a workaround. But I agree from a runtime complexity standpoint its pretty ugly. So I've changed to the other workaround: type featureFlagsToSendToServer as any, initially, and then cast to Partial<FeatureFlags> when returning it. How do you feel about that?

tensorboard/webapp/feature_flag/store/feature_flag_selectors.ts
```diff
@@ -58,6 +58,24 @@ export const getFeatureFlagsMetadata = createSelector(
...
+    }
+    return featureFlagsToSendToServer;
```
rileyajones: > Unfortunately not totally that simple. Making that change leads to the following error:
> 
> Type 'number | boolean | string[] | null | undefined' is not assignable to type 'never'.
> 
> The anti-pattern you identifeid here was a workaround. But I agree from a runtime complexity standpoint its pretty ugly. So I've changed to the other workaround: type featureFlagsToSendToServer as any, initially, and then cast to Partial when returning it. How do you feel about that?

You can get away with it with some slightly looser typing.
```suggestion
    let featureFlagsToSendToServer: Partial<Record<keyof FeatureFlags, FeatureFlagType>> = {};
    for (const entry in state.flagOverrides) {
      const entryMetadata = state.metadata[entry as keyof FeatureFlags];
      if (
        entryMetadata.queryParamOverride &&
        entryMetadata.sendToServerWhenOverridden
      ) {
        featureFlagsToSendToServer[entry as keyof FeatureFlags] = state.flagOverrides[entry as keyof FeatureFlags];
      }
    }
    return featureFlagsToSendToServer as Partial<FeatureFlags>;
```

bmd3k: Done.

rileyajones (approved):

PR #5884, opened 2022-08-17
linked time: Add min, max, and percentage change columns

JamesHollyer:
When we launch linked time we will be allowing a range selection. The data table needs to be able to handle range values when in that mode. This change adds the Min, Max, and Percentage Change columns so the user knows which step is selected for each run.

There are two main changes between these columns and others that have been added. The first is that the min and max have some calculation that I included in a utility function that I put directly in the ScalarCardDataTable. The second is that the percentage change uses arrows similar to the value change column. For code reuse I created an ng-template for the arrow and used that template for both columns.

* Screenshots of UI changes
This is how the UI looks after this PR. In a future PR I will replace the word "Change" with a delta symbol and I plan to make a few other header clean up changes.

Screen Shot 2022-08-17 at 3 06 53 PM


tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -428,7 +428,10 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
           headers.push(ColumnHeaders.RELATIVE_TIME);
         } else {
           headers.push(ColumnHeaders.RUN);
+          headers.push(ColumnHeaders.MIN_VALUE);
```
bmd3k, line 431: I would be nice to have a comment saying this else block is for range selection.

tensorboard/webapp/widgets/data_table/data_table_component.ng.html
```diff
@@ -33,22 +33,33 @@
...
+                <ng-container
+                  *ngTemplateOutlet="arrow; context: {$implicit:runData.PERCENTAGE_CHANGE}"
+                ></ng-container>
                 {{ getFormattedDataForColumn(header, runData) }}
               </span>
               <span *ngSwitchDefault
                 >{{ getFormattedDataForColumn(header, runData) }}</span
               >
             </td>
+            <ng-template #arrow let-value>
```
bmd3k, line 51: This seems like a slightly odd place to locate your template: Can it be moved to the top-level or closer to the top-level of the dom hierarchy in this file? Maybe just move it to the end of the file?

tensorboard/webapp/widgets/data_table/data_table_component.ts
```diff
@@ -144,6 +150,29 @@ export class DataTableComponent {
         return intlNumberFormatter.formatShort(
           selectedStepRunData.END_VALUE as number
         );
+      case ColumnHeaders.MIN_VALUE:
+        if (selectedStepRunData.MIN_VALUE === undefined) {
+          return '';
+        }
+        return intlNumberFormatter.formatShort(
+          selectedStepRunData.MIN_VALUE as number
```
bmd3k, line 158: slightly odd that you have to cast this 'as number' given that you just checked whether it is undefined. What does the compiler think is the type of selectedStepRunData.MIN_VALUE?

(I guess comment applies multiple places in this switch statement)

bmd3k (approved):

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
JamesHollyer: Done.

tensorboard/webapp/widgets/data_table/data_table_component.ng.html
JamesHollyer: I thought it was convenient to have a it close to where it is being used. I moved it to the bottom of the file though.

tensorboard/webapp/widgets/data_table/data_table_component.ts
JamesHollyer: The type is number | string | undefined. So at this point the compiler thinks it is a number | string.

JamesHollyer (commented):
Thanks Brian.

PR #5888, opened 2022-08-18
Update affordance `enums` to use string values

rileyajones:
The existing enums using numeric values prevented them from being compared and made some internal logic much more complicated.

1) Use the change in internal tensorboard.
2) Perform actions to dispatch effected analytics events
3) Use the analytics dashboard to ensure the events are unchanged.

Using a `record` instead would potentially be more complicated but would also remove the need to map enums to strings.

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_types.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_types.ts
@@ -25,19 +25,19 @@ export interface TimeSelection {
  * Only used for internal analytics.
  */
 export enum TimeSelectionAffordance {
-  NONE,
+  NONE = 'NO_AFFORDANCE',
   // Dragging the extended line above a fob.
-  EXTENDED_LINE,
+  EXTENDED_LINE = 'EXTENDED_LINE',
   // Dragging the fob.
-  FOB,
+  FOB = 'FOB',
   // Clicking the deselect button.
-  FOB_REMOVED,
+  FOB_REMOVED = 'FOB_REMOVED',
   // Typing the step in fob.
-  FOB_TEXT,
+  FOB_TEXT = 'FOB_TEXT',
   // Typing the step in setting pane.
-  SETTINGS_TEXT,
+  SETTINGS_TEXT = 'SETTINGS_TEXT',
   // Dragging the slider in setting pane.
-  SETTINGS_SLIDER,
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_types.ts
```diff
@@ -25,31 +25,31 @@ export interface TimeSelection {
  * Only used for internal analytics.
  */
 export enum TimeSelectionAffordance {
-  NONE,
+  NONE = 'NO_AFFORDANCE',
```
rileyajones, line 28: Note that the string values of `TimeSelectionAffordance.NONE` and `TimeSelectionToggleAffordance.NONE` are intentionally not set to `"NONE"` so that they are different in case we wanted to handle them differently in the future.

tensorboard/webapp/widgets/card_fob/card_fob_types.ts
```diff
@@ -25,31 +25,31 @@ export interface TimeSelection {
...
+  FOB_REMOVED = 'FOB_REMOVED',
   // Typing the step in fob.
-  FOB_TEXT,
+  FOB_TEXT = 'FOB_TEXT',
   // Typing the step in setting pane.
-  SETTINGS_TEXT,
+  SETTINGS_TEXT = 'SETTINGS_TEXT',
   // Dragging the slider in setting pane.
-  SETTINGS_SLIDER,
+  SETTINGS_SLIDER = 'SETTINGS_SLIDER',
```
japie1235813, line 40: How about making the string lower case and separated with space? i.e. "settings slider" Then probably the extra conversion could be saved. What do you think?

japie1235813 (approved):

tensorboard/webapp/widgets/card_fob/card_fob_types.ts
rileyajones: It mostly could except for the situation with `NONE`. I'll go ahead and change them all though.

rileyajones: I'm going to change to use camel case as that is what they are converted to internally.

japie1235813: nice. I misremembered the converted string type for internal analysis. lgtm

japie1235813 (approved):

PR #5889, opened 2022-08-19
Change data table header background based on theme

JamesHollyer:
Currently the background color of the header is hard coded to white. This is a problem when users are in dark mode. This PR sets the background the the correct theme color based and changes it for dark mode.

Before:
Screen Shot 2022-08-22 at 11 36 08 AM

After:
Screen Shot 2022-08-22 at 11 58 49 AM

### DIFF
--- a/tensorboard/webapp/widgets/data_table/data_table_component.scss
+++ b/tensorboard/webapp/widgets/data_table/data_table_component.scss
@@ -24,6 +24,7 @@ limitations under the License.
     position: sticky;
     text-align: left;
     top: 0;
+
     @include tb-dark-theme {
       background-color: map-get($tb-dark-background, background);
     }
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/widgets/data_table/data_table_component.scss
```diff
@@ -12,16 +12,21 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
   border-spacing: 4px;
   font-size: 13px;
 
   th {
-    background-color: white;
+    background-color: mat.get-color-from-palette($tb-background, background);
     position: sticky;
     text-align: left;
     top: 0;
+    @include tb-dark-theme {
```
japie1235813, line 27: nit: could you add a new line between 26-27?
I think it's cleaner to separate them.
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/plugins/plugins_component.scss;l=23?q=%22@include%20tb-dark-theme%20%7B%22&ss=tensorflow%2Ftensorboard

japie1235813 (approved):
Could you attach the before/after screenshots?
Please also update/add webtests for this PR, thanks!

PR #5893, opened 2022-08-23
Fix brittle Zlib-dependent test

groszewn:
Tests that expect zlib to produce a particular bitstream for some input
data are brittle, because the output of zlib is variable for different
implementations and different tunings.

We trust the underlying zlib `encode_png` op does the right thing if it
has a valid PNG header and is of a reasonable size.

### DIFF
--- a/tensorboard/util/encoder_test.py
+++ b/tensorboard/util/encoder_test.py
@@ -33,7 +33,7 @@ class TensorFlowPngEncoderTest(tf.test.TestCase):
         # assume it did the right thing. We trust the underlying
         # `encode_png` op.
         self.assertEqual(b"\x89PNG", data[:4])
-        self.assertGreater(len(data), 128)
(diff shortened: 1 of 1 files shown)


bmd3k (approved):
This seems like a fine temporary solution to unbreak TAP. Do you think there is an opportunity to followup with a less-fragile solution?

bileschi (approved):
LGTM - thanks

groszewn:
We could decode the compressed PNG and test equality against the source data to avoid the current fragility. I can follow up with another PR if that sounds good.

nfelt:
FWIW, I think we could just drop the length check entirely. This code is ~mostly obsolete, and we are already checking for the PNG magic number, which seems sufficient IMO.

groszewn:
Noted! See #5896.

PR #5896, opened 2022-08-23
Remove extraneous check on encoded data length

groszewn:
The encoded data length is something that can change over time with
improvements to the underlying compression. We trust the underlying
`encode_op`, so just checking that that we have a valid PNG header is
sufficient.


nfelt (approved):

nfelt:
Thanks!

PR #5897, opened 2022-08-23
Update range input widget to output the source of the change

rileyajones:
There are two people added on this but I'm only looking for a single person to review it.

We would like to log the source of range changes internally for analytics purposes but it is not currently possible to differentiate between events created from text changes and events created from slider changes. 

I added another property to each of the events which describes how the range change was created (text input vs slider).

1) Start tensorboard
2) Navigate to http://localhost:6006/?enableLinkedTime=true&enableDataTable=true#timeseries
3) Enable `Link By Step` in the right pane
4) Use the range input component and ensure it works as expected.

The range_input component could emit a second event describing the source of the change which could then be correlated based on timing. Unfortunately that it isn't a strong binding and leaves room for error, however, it would avoid potential breakage and allow complete separation between analytics and normal usage..


tensorboard/webapp/metrics/views/right_pane/settings_view_container.ts
```diff
@@ -44,6 +48,14 @@ import {
   XAxisType,
 } from '../../types';
 
+const RANGE_INPUT_SOURCE_TO_AFFORDANCE: Record<
```
JamesHollyer, line 51: This object seems needlessly complicated. Why not just use source === RangeInputSource.TEXT ? TimeSelectionAffordance.SETTINGS_TEXT | TimeSelectionAffordance.SETTINGS_SLIDER.  That seems easier to read to me but maybe that is just my opinion.

That being said I do not see anything wrong with this way of converting.

tensorboard/webapp/metrics/views/right_pane/settings_view_component.ts
```diff
@@ -71,7 +76,10 @@ export class SettingsViewComponent {
   @Input() stepMinMax!: {min: number; max: number};
 
   @Output() linkedTimeToggled = new EventEmitter<void>();
-  @Output() linkedTimeSelectionChanged = new EventEmitter<TimeSelection>();
+  @Output()
+  linkedTimeSelectionChanged = new EventEmitter<
+    TimeSelection & {source: RangeInputSource}
```
JamesHollyer, line 81: This combined type and the RangeValuesEvent type are very similar. Could we use the RangeValuesEvent here? Maybe even change the RangeValuesEvent type to extend TimeSelection?

rileyajones: I would rather the `range_input` component not know about linked time so I would prefer them use different objects.

tensorboard/webapp/metrics/views/right_pane/settings_view_container.ts
rileyajones: This will lead to a loud failure if we add to `RangeInputSource` which I view as desirable.

tensorboard/webapp/metrics/views/right_pane/settings_view_component.ts
```diff
@@ -174,27 +182,23 @@ export class SettingsViewComponent {
     );
   }
 
-  onSingleValueChanged(stepValue: number) {
+  onSingleValueChanged({value, source}: SingleValueEvent) {
     this.linkedTimeSelectionChanged.emit({
-      start: {step: stepValue},
+      start: {step: value},
       end: null,
+      source,
```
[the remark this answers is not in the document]
rileyajones: Done.

tensorboard/webapp/widgets/range_input/range_input_test.ts
```diff
@@ -426,7 +444,10 @@ describe('range input test', () => {
       minInput.value = '2';
       minInput.dispatchEvent(new InputEvent('change'));
 
-      expect(onSingleValueChanged).toHaveBeenCalledWith(2);
+      expect(onSingleValueChanged).toHaveBeenCalledWith({
+        value: 2,
+        source: RangeInputSource.TEXT,
```
JamesHollyer, line 449: There should also be tests for onSingleValueChanged with RangeInputSource.SLIDER.

JamesHollyer (commented):
I would think the single slider event should also have a source. It is emitted from directly from the html file. Also update the test to include that case.

tensorboard/webapp/widgets/range_input/range_input_test.ts
rileyajones: > I would think the single slider event should also have a source. It is emitted from directly from the html file. Also update the test to include that case.

I didn't catch the event emitted from the dom, thanks! I'm moving that to a function as I believe the enum is clearer that way.

rileyajones: I've gone ahead and added a test.

JamesHollyer (approved):

PR #5898, opened 2022-08-24
linked time: displays expected step for link by step in settings

japie1235813:
Before user check "Link by step" we show them what is the step we are going to link across cards.

This only works on 
  * single selection (we can think about extend to range selection later)
  * when linked time is disabled: we don't want to have two identical numbers shown up close and updated at the same time

* Screenshots of UI changes
https://user-images.githubusercontent.com/1131010/186510747-88776ec2-7d86-42f8-8011-9345658c519d.mov

### DIFF
--- a/tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
+++ b/tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
@@ -550,6 +550,53 @@ describe('metrics right_pane', () => {
             })
           );
         });
+
+        it('displays expected linked time start step when linked time is not enabled', () => {
+          store.overrideSelector(
+            selectors.getMetricsLinkedTimeSelectionSetting,
+            {
+              start: {step: 20},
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/metrics/views/right_pane/right_pane_test.ts
```diff
@@ -550,6 +550,53 @@ describe('metrics right_pane', () => {
...
+          store.overrideSelector(
+            selectors.getMetricsLinkedTimeSelectionSetting,
+            {
+              start: {step: 40},
+              end: null,
+            }
+          );
+          store.refreshState();
+          fixture.detectChanges();
+          expect(el.nativeElement.textContent).toContain('40');
```
bmd3k, line 580: I'd be slightly more comfortable if you could check that they are 'Link by step 20' and 'Link by step 40'.

bmd3k (approved):

PR #5899, opened 2022-08-25
Linked Time: Clean data table headers

JamesHollyer:
In our next phase of launching linked time we introduce the range selection. This has a whole new set of columns in the data table. These new columns have already been added however, after discussions with the team and with users we want to change the ordering and text in the headers. This PR makes those changes to the headers.

This change is mostly text changes and some css. The only real logic added was to add an ngswitch inside the data_table_component.ng.html file which adds the triangle mat-icon(which is called "change_history" for some reason).  This is also the first time we added that icon so I had to add it to our build in the js.bzl file.

* Screenshots of UI changes
Screen Shot 2022-08-25 at 2 17 55 PM

### DIFF
--- a/tensorboard/webapp/widgets/data_table/data_table_component.ng.html
+++ b/tensorboard/webapp/widgets/data_table/data_table_component.ng.html
@@ -20,23 +20,21 @@ limitations under the License.
           <!-- This header is intentionally left blank for the color column -->
         </th>
         <ng-container *ngFor="let header of headers;">
-          <th [ngSwitch]="header">
-            <div *ngSwitchCase="ColumnHeaders.VALUE_CHANGE" class="cell">
+          <th>
+            <div [ngSwitch]="header" class="cell">
               <mat-icon
+                *ngSwitchCase="ColumnHeaders.VALUE_CHANGE"
                 class="delta-arrow"
                 svgIcon="change_history_24px"
               ></mat-icon>
-              <div>{{ getHeaderTextColumn(header) }}</div>
-            </div>
-            <div *ngSwitchCase="ColumnHeaders.PERCENTAGE_CHANGE" class="cell">
               <mat-icon
+                *ngSwitchCase="ColumnHeaders.PERCENTAGE_CHANGE"
                 class="delta-arrow"
                 svgIcon="change_history_24px"
               ></mat-icon>
-              <div>{{ getHeaderTextColumn(header) }}</div>
-            </div>
-            <div *ngSwitchDefault class="cell extra-right-padding">
-              {{ getHeaderTextColumn(header) }}
+              <div *ngSwitchDefault class="extra-right-padding"></div>
+
+              <span>{{ getHeaderTextColumn(header) }}</span>
             </div>
           </th>
         </ng-container>
--- a/tensorboard/webapp/widgets/data_table/data_table_component.ng.html
+++ b/tensorboard/webapp/widgets/data_table/data_table_component.ng.html
@@ -20,7 +20,25 @@ limitations under the License.
           <!-- This header is intentionally left blank for the color column -->
         </th>
         <ng-container *ngFor="let header of headers;">
-          <th>{{ getHeaderTextColumn(header) }}</th>
+          <th [ngSwitch]="header">
+            <div *ngSwitchCase="ColumnHeaders.VALUE_CHANGE" class="cell">
+              <mat-icon
+                class="delta-arrow"
+                svgIcon="change_history_24px"
+              ></mat-icon>
+              <div>{{ getHeaderTextColumn(header) }}</div>
+            </div>
+            <div *ngSwitchCase="ColumnHeaders.PERCENTAGE_CHANGE" class="cell">
+              <mat-icon
+                class="delta-arrow"
+                svgIcon="change_history_24px"
+              ></mat-icon>
+              <div>{{ getHeaderTextColumn(header) }}</div>
+            </div>
+            <div *ngSwitchDefault class="cell without-icon">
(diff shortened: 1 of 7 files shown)


tensorboard/webapp/widgets/data_table/data_table_component.ng.html
```diff
@@ -20,7 +20,25 @@
...
+              <mat-icon
+                class="delta-arrow"
+                svgIcon="change_history_24px"
+              ></mat-icon>
+              <div>{{ getHeaderTextColumn(header) }}</div>
+            </div>
+            <div *ngSwitchDefault class="cell without-icon">
+              {{ getHeaderTextColumn(header) }}
+            </div>
+          </th>
```
japie1235813, line 41: Can we move the header switch logic to only icon?
For example
```
          <th [ngSwitch]="header" class="cell">
              <mat-icon
                *ngSwitchCase="ColumnHeaders.VALUE_CHANGE"
                class="delta-arrow"
                svgIcon="change_history_24px"
              ></mat-icon>
              <mat-icon
                *ngSwitchCase="ColumnHeaders.PERCENTAGE_CHANGE"
                class="delta-arrow"
                svgIcon="change_history_24px"
              ></mat-icon>
              <div
                *ngSwitchDefault
                class="without-icon"
              >
              </div>
              <div>{{ getHeaderTextColumn(header) }}</div>
          </th>
```

I tested it a bit and it looks similar

japie1235813: Same, I would argue for the runData below. But that's not included in this PR.

tensorboard/webapp/widgets/data_table/data_table_component.scss
```diff
@@ -24,12 +24,23 @@ limitations under the License.
...
       background-color: map-get($tb-dark-background, background);
     }
   }
 
+  .cell {
+    display: flex;
+    align-items: center;
+  }
+
+  .without-icon {
```
japie1235813, line 39: Instead of "without-icon" I think an expressive naming is like "extra-right-padding". Especially there is only one css change here.
It's not clear to me when I see "without-icon" in the html

tensorboard/webapp/widgets/data_table/data_table_component.scss
```diff
@@ -24,12 +24,23 @@ limitations under the License.
...
     text-align: left;
     top: 0;
+    vertical-align: bottom;
 
     @include tb-dark-theme {
       background-color: map-get($tb-dark-background, background);
     }
   }
 
+  .cell {
```
japie1235813, line 34: Can we rename this to header or datatable-header?
Technically every cell in the table is a cell, but you only want this applied to the th header

tensorboard/webapp/widgets/data_table/data_table_component.scss
```diff
@@ -24,12 +24,23 @@ limitations under the License.
...
+    display: flex;
+    align-items: center;
```
japie1235813, line 36: nit: a-z sorted

JamesHollyer (commented):
Hey Jie-Wei. I left some replies. Happy to make the changes if you feel strongly.

tensorboard/webapp/widgets/data_table/data_table_component.ng.html
japie1235813: Except color (which you've handled differently) I don't feel anything might \not contain header text. Let's just make the switch cases for icons. (From my perspective, "no icon" is also one of the icon options)

tensorboard/webapp/widgets/data_table/data_table_component.ng.html
```diff
@@ -32,21 +50,21 @@
...
                 <ng-container
                   *ngTemplateOutlet="arrow; context: {$implicit:runData.PERCENTAGE_CHANGE}"
                 ></ng-container>
                 {{ getFormattedDataForColumn(header, runData) }}
-              </span>
-              <span *ngSwitchDefault
-                >{{ getFormattedDataForColumn(header, runData) }}</span
-              >
+              </div>
+              <div *ngSwitchDefault class="cell extra-right-padding">
```
japie1235813, line 65: I also want to reconstruct the html to remove the repeated cell class. Yet this is not the requirement for this PR.

tensorboard/webapp/widgets/data_table/data_table_component.scss
japie1235813: ok I saw it in runData section

tensorboard/webapp/widgets/data_table/data_table_test.ts
```diff
@@ -90,16 +90,16 @@ describe('data table', () => {
...
-    expect(headerElements[9].nativeElement.innerText).toBe('End');
+    expect(headerElements[8].nativeElement.innerText).toBe('Start Value');
+    expect(headerElements[9].nativeElement.innerText).toBe('End Value');
     expect(headerElements[10].nativeElement.innerText).toBe('Min');
     expect(headerElements[11].nativeElement.innerText).toBe('Max');
-    expect(headerElements[12].nativeElement.innerText).toBe(
-      'Percentage Change'
-    );
+    expect(headerElements[12].nativeElement.innerText).toBe('%');
+    expect(headerElements[12].queryAll(By.css('mat-icon')).length).toBe(1);
```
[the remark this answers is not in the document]
japie1235813: I would then remove the length check here. I think the length is not necessary/important now. but I'm also okay with leaving it. Up to you.

japie1235813 (approved):

PR #5900, opened 2022-08-25
Refactor TimeSeletionChanged events

rileyajones:
1) We had two different data types that were being used inconsistently.
2) Each place that the object was used redeclared the type which made it difficult to track and update.

There should be no functional changes.
I played with both the scalar and histogram cards to ensure they still worked, then imported the code internally and stepped through it with a debugger to ensure the desired effect was met.
Ignore the inconsistent data structures and just add the new property where needed.

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_types.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_types.ts
@@ -88,3 +88,13 @@ export interface CardFobGetStepFromPositionHelper {
    */
   getStepLowerThanAxisPosition(position: number): number;
 }
+
+/**
+ * Contents of the timeSelectionChanged event.
+ */
+export type TimeSelectionChanged = {
+  // Affordance for internal analytics purpose. When no affordance is specified or is
+  // undefined we do not want to log an analytics event.
+  affordance?: TimeSelectionAffordance;
+  timeSelection: TimeSelection;
+};
--- a/tensorboard/webapp/widgets/card_fob/card_fob_types.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_types.ts
@@ -90,9 +90,9 @@ export interface CardFobGetStepFromPositionHelper {
 }
 
 /**
- * Contents of the timeSelectionChanged event.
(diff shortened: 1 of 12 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_types.ts
```diff
@@ -88,3 +88,14 @@ export interface CardFobGetStepFromPositionHelper {
...
 }
+
+/**
+ * Contents of the timeSelectionChanged event.
+ */
+export type TimeSelectionChanged = {
+  // Affordance for internal analytics purpose. When no affordance is specified or is
+  // undefined we do not want to log an analytics event.
+  affordance?: TimeSelectionAffordance;
+  incomplete?: boolean;
```
rileyajones, line 99: Is this really the best attribute name?
It is used to indicate the user is still dragging the fob and thus the "Change" is not complete.

japie1235813 (commented):
I got the wrong impression that this is changing the ViewTimeSelection but actually it is about the metric/actions prop.
I might slightly toward to the refactoring. The undefined or null  for end/endStep was a pain to me previously. I would also want to know other people's thoughts.

rileyajones:
> I got the wrong impression that this is changing the ViewTimeSelection but actually it is about the metric/actions prop. I might slightly toward to the refactoring. The undefined or null for end/endStep was a pain to me previously. I would also want to know other people's thoughts.

My "2 cents" on this is that updating updating `TimeSelection` from
```
interface TimeSelection {
  start: {step: number};
  end: {step: number} | null;
}
```
to 
```
interface TimeSelection {
  start: {step: number};
  end?: {step: number};
}
```

would be significantly easier to work with as it removes the need to continually declare `end` as `null`.

I'm going to wait to make any change to that object until hearing from @JamesHollyer though as I suspect he has an opinion on it.

JamesHollyer (commented):
This change looks good. I personally like how explicit the null is. So much changes if that is null or not that I would rather it be explicitly set. That is a slight preference though. I would not be upset if you decide to make it an optional attribute.

JamesHollyer (approved):

PR #5901, opened 2022-08-26
Revert "linked time: displays expected step for link by step in settings"

japie1235813:
Reverts tensorflow/tensorboard#5898

Thanks for @rileyajones 's findings. The original PR will ruin our internal analysis.
We will end up in a case that the affordance check does not pass when catching stepSelectorTimeSelectionChange. I will investigate what's the cause.
To prevent this being released next Tue and to not block the release. Revert this PR until the analysis is handled.

Update: I cannot reproduce the bug


japie1235813:
Cannot reproduce the bug. Close this PR.

PR #5902, opened 2022-08-28
Add completes for shells by shtab, fix #5886

### LINKED ISSUE #5886: Add shell completion
How about adding shell completion for bash/zsh/fish/posh/etc?

```shell
❯ tensorboard --<Tab>
option
--bind_all                              Serve on all public interfaces. This will expose your TensorBoard instance to the network on both IPv4 and IPv6 (where available). Mutually exclusive with `--host`.
--db                                    (experimental) sets SQL database URI and enables DB backend mode, which is read-only unless --db_import is also passed.
--db_import                             (experimental) enables DB read-and-import mode, which in combination with --logdir imports event files into a DB backend on the fly. The backing DB is temporary unless --db is also passed to specify a DB path to use.
--detect_file_replacement               BOOL
...
```

Thanks!

Freed-Wu:

Fix #5886, Add completes for shell

* Screenshots of UI changes

Take bash as an example:

```bash
[root@desktop Desktop]# tensorboard dev upload --<TAB>
--description  --dry_run      --help         --helpfull     --logdir       --name         --one_shot     --plugins      --verbose
```

```
tensorboard --print-completion bash | sudo tee /usr/share/bash-completion/completions/tensorboard
tensorboard --print-completion tcsh | sudo tee /etc/profile.d/tensorboard.completion.csh
# tensorboard --print-completion zsh | sudo tee /usr/share/zsh/site-functions/_tensorboard
```

`tensorboard --print-completion zsh` has some bug which comes from shtab.

- Realizing a py script to do this work which is not dependent in shtab.
- click, etc, See <https://docs.iterative.ai/shtab/#alternatives>

### DIFF
--- a/tensorboard/_shtab.py
+++ b/tensorboard/_shtab.py
@@ -1,0 +1,8 @@
+FILE = None
+DIRECTORY = DIR = None
+
+
+def add_argument_to(parser, *args, **kwargs):
+    from argparse import Action
+    Action.complete = None
+    return parser
--- a/tensorboard/pip_package/setup.py
+++ b/tensorboard/pip_package/setup.py
@@ -63,6 +63,9 @@ setup(
         ],
     },
     install_requires=REQUIRED_PACKAGES,
+    extras_require={
+        "shtab": ["completion"],
+    },
     tests_require=REQUIRED_PACKAGES,
     python_requires=">=3.7",
     # PyPI package information.
--- a/tensorboard/program.py
+++ b/tensorboard/program.py
@@ -31,6 +31,7 @@ from abc import abstractmethod
 import argparse
 import atexit
 from collections import defaultdict
+from contextlib import suppress
 import errno
 import logging
 import mimetypes
@@ -43,6 +44,10 @@ import threading
 import time
 import urllib.parse
 
+try:
+    import shtab
+except ImportError:
+    from . import _shtab as shtab
 from absl import flags as absl_flags
 from absl.flags import argparse_flags
 from werkzeug import serving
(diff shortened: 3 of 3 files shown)


arcra (commented):
Thanks for the contribution!

We're a bit reluctant to introduce this dependency here. Especially because this gets synced to our internal repository, and having references to other third-party libraries comes with additional burden and maintenance considerations.

We were thinking that exposing the parser and using the [CLI mode](https://docs.iterative.ai/shtab/use/#cli-usage) might be a better approach that doesn't require us to include this dependency, but we're not sure this would work well with the logic we have for our subparser for the serve subcommand. (This might be what you were referring to in your latest comment on issue #5886)

An alternative is to use argcomplete, to which we do already have a reference in our internal repository (v.1.9.4).

You're welcome to give it a shot with either of these approaches, if you have time and interest, but this will probably have to wait until we have some time to look into it with more detail, and quite honestly, it's not at the top of our priority, sorry.

Freed-Wu:
How about make shtab as a internal module? Like pytorch's appdir <https://github.com/pytorch/pytorch/blob/master/torch/_appdirs.py>? Just add a new module `_shtab.py`.

Or make it become a optional dependency? e.g., `pip install tensorboard[completion]`.

Freed-Wu:
> introduce this dependency

It usually be a make depend not a run depend, see <https://github.com/archlinux/svntogit-community/blob/packages/trash-cli/trunk/PKGBUILD#L14>
You can refer

- <https://github.com/google/vimdoc/blob/main/setup.py#L36>
- <https://github.com/andreafrancia/trash-cli/blob/master/setup.cfg#L31-L33>
- <https://github.com/inducer/pudb/blob/main/setup.py#L28>
- <https://github.com/piccolomo/plotext/blob/master/setup.py#L25>
- <https://github.com/wookayin/gpustat/blob/master/setup.py#L114>

> An alternative is to use argcomplete

the weakness of argcomplete is listed in <https://docs.iterative.ai/shtab/#alternatives>

> we have some time to look into it with more detail

If you want, remember post me :smile: Now, let me close it.

PR #5903, opened 2022-08-29
GitHub Actions: Cache output of `bazel fetch`

rileyajones:
Our builds are taking an extremely long time due to http request timeouts. Hopefully by caching the external `bazel` dependencies we should see build times drop dramatically.
image

We could switch to using `docker` images in which we prefetch dependencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,8 +82,20 @@ jobs:
             ;
       - name: 'Check Pip state'
         run: pip freeze --all
+      - name: 'Restore Bazel Cache'
+        uses: actions/cache@1a9e2138d905efd099035b49d8b7a3888c653ca8
+        with:
+          path: |
+            bazel-bin
+          key: bazel-fetch-${{ runner.os }}-${{ hashFiles('.github/workflows/ci.yml') }}
       - name: 'Bazel: fetch'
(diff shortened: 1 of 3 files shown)


nfelt (commented):
So, we already do have read/write build caching set up for Bazel, using a GCS bucket: https://github.com/tensorflow/tensorboard/blob/6e192f35f9700cdf0a11645edd795b619db973d0/ci/bazelrc#L27-L34

(The cache is only written to for submitted code: https://github.com/tensorflow/tensorboard/blob/6e192f35f9700cdf0a11645edd795b619db973d0/.github/workflows/ci.yml#L53-L72) 

Could you elaborate a little about the current timeout errors you're seeing?

Bazel is pretty finicky about what you cache, and it's easy for over-broad caching to mess up the hermeticity guarantees, which is why we've tried to stick to the built-in caching support. To use `actions/cache` you'd need to set the `key` to something that avoids cache pollution. I'm also not sure `bazel-bin` is a path you'd want to cache, since AFAIK it contains built binaries, so a clean `bazel fetch` wouldn't actually write anything to `bazel-bin` anyway - it looks like fetching writes to this output path: https://bazel.build/docs/external#layout But again, if possible it's probably safer to try to tune the built-in caching rather than manually caching bits of the Bazel output tree.

PR #5904, opened 2022-08-29
Bug Fix: Update BUILD rules for feature flags modal

rileyajones:
I'm only looking for a single approval here but I'm adding multiple reviewers due to time concerns. Proof that this works for Googlers cl/470825370.

The current build rules do not work internally

1) Imported and ran presubmits internally.


bileschi (approved):
Thanks!

PR #5905, opened 2022-08-30
Added support for pathlib paths as a logdir using the recommended os.PathLike class.

mspils:
Most other Tensorflow functions accept Paths in other form than strings. 
This small change makes the plugin more consistent and users do not need to cast their Path objects to string manually.

- Imported os
- added another elif case for the type of writer:
elif isinstance(writer, os.PathLike):
            self._writer = tf.compat.v2.summary.create_file_writer(os.fsdecode(writer))
- Changed the documentation to fit this

Barely tested to be honest. Just ran the tutorial code here: https://www.tensorflow.org/tensorboard/hyperparameter_tuning_with_hparams
with my version

Once all Tensorflow functions accept pathlikes instead of just strings the distinction between str and PathLike will be unnecessary.

### DIFF
--- a/tensorboard/plugins/hparams/_keras.py
+++ b/tensorboard/plugins/hparams/_keras.py
@@ -17,7 +17,7 @@
 Use `tensorboard.plugins.hparams.api` to access this module's contents.
 """
 
-
+import os 
 import tensorflow as tf
 
 from tensorboard.plugins.hparams import api_pb2
@@ -39,7 +39,7 @@ class Callback(tf.keras.callbacks.Callback):
 
         Args:
           writer: The `SummaryWriter` object to which hparams should be
-            written, or a logdir (as a `str`) to be passed to
+            written, or a logdir (as a `str` or `PathLike`) to be passed to
             `tf.summary.create_file_writer` to create such a writer.
           hparams: A `dict` mapping hyperparameters to the values used in
             this session. Keys should be the names of `HParam` objects used
@@ -62,10 +62,12 @@ class Callback(tf.keras.callbacks.Callback):
         summary_v2.hparams_pb(self._hparams, trial_id=self._trial_id)
         if writer is None:
             raise TypeError(
-                "writer must be a `SummaryWriter` or `str`, not None"
+                "writer must be a `SummaryWriter`, `str` or `PathLike`, not None"
             )
         elif isinstance(writer, str):
             self._writer = tf.compat.v2.summary.create_file_writer(writer)
+        elif isinstance(writer, os.PathLike):
+            self._writer = tf.compat.v2.summary.create_file_writer(os.fsdecode(writer))
         else:
             self._writer = writer
 


arcra (changes requested):
This seems reasonable. I'll check with the team, but I don't think there would be objections to this. Would you mind adding a small test for it? There seems to already be a test for using an explicit writer:
github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/hparams/_keras_test.py#L133

mspils:
Sounds very reasonable. 
I added a minimal test, that just checks if nothing crashes and that a logfile is created.

tensorboard/plugins/hparams/_keras.py
```diff
@@ -62,10 +62,12 @@ def __init__(self, writer, hparams, trial_id=None):
         summary_v2.hparams_pb(self._hparams, trial_id=self._trial_id)
         if writer is None:
             raise TypeError(
-                "writer must be a `SummaryWriter` or `str`, not None"
+                "writer must be a `SummaryWriter`, `str` or `PathLike`, not None"
             )
         elif isinstance(writer, str):
             self._writer = tf.compat.v2.summary.create_file_writer(writer)
+        elif isinstance(writer, os.PathLike):
+            self._writer = tf.compat.v2.summary.create_file_writer(os.fsdecode(writer))
```
arcra, line 70: `os.fsdecode` seems to be related/useful for file names with specific encodings, but we'd prefer not to get into anything related to filesystem encodings for now. Can you please update it to use `os.fspath`?

TensorFlow does not use `os.fsdecode` anywhere, but it uses `os.fspath` in several places, so that would also be more consistent.

tensorboard/plugins/hparams/_keras_test.py
```diff
@@ -145,6 +146,15 @@ def test_explicit_writer(self):
         # We'll assume that the contents are correct, as in the case where
         # the file writer was constructed implicitly.
 
+    def test_pathlib_writer(self):
+
```
arcra, line 150: Please remove this empty line.

arcra (changes requested):

tensorboard/plugins/hparams/_keras.py
mspils: Good point, sorry for the additional work :)

tensorboard/plugins/hparams/_keras.py
```diff
@@ -62,10 +62,12 @@ def __init__(self, writer, hparams, trial_id=None):
         summary_v2.hparams_pb(self._hparams, trial_id=self._trial_id)
         if writer is None:
             raise TypeError(
-                "writer must be a `SummaryWriter` or `str`, not None"
+                "writer must be a `SummaryWriter`, `str` or `PathLike`, not None"
             )
         elif isinstance(writer, str):
             self._writer = tf.compat.v2.summary.create_file_writer(writer)
+        elif isinstance(writer, os.PathLike):
+            self._writer = tf.compat.v2.summary.create_file_writer(os.fspath(writer))
```
rileyajones, line 70:
```suggestion
            self._writer = tf.compat.v2.summary.create_file_writer(
                os.fspath(writer)
            )
```

tensorboard/plugins/hparams/_keras.py
```diff
@@ -17,7 +17,7 @@
 Use `tensorboard.plugins.hparams.api` to access this module's contents.
 """
 
-
+import os
```
rileyajones, line 20:
```suggestion
import os
```

arcra (changes requested):
In addition to the suggestions Riley left, there's a test failing, because the error message needs to be updated in this test:
github.com/tensorflow/tensorboard/blob/master/tensorboard/plugins/hparams/_keras_test.py#L168

mspils:
Thanks for the Change requests.
All those requirements take some time to get used to, but seem very reasonable :)

tensorboard/plugins/hparams/_keras.py
```diff
@@ -62,10 +62,14 @@ def __init__(self, writer, hparams, trial_id=None):
         summary_v2.hparams_pb(self._hparams, trial_id=self._trial_id)
         if writer is None:
             raise TypeError(
-                "writer must be a `SummaryWriter` or `str`, not None"
+                "writer must be a `SummaryWriter`, `str` or `PathLike`, not None"
```
rileyajones, line 65: Just a nitpick here but it would be nice to keep the formatting consistent
```suggestion
                "writer must be a `SummaryWriter`, `str`, or `PathLike`, not None"
```

tensorboard/plugins/hparams/_keras_test.py
```diff
@@ -165,7 +174,7 @@ def test_reuse_failure(self):
     def test_invalid_writer(self):
         with self.assertRaisesRegex(
             TypeError,
-            "writer must be a `SummaryWriter` or `str`, not None",
+            "writer must be a `SummaryWriter`, `str` or `PathLike`, not None",
```
rileyajones, line 177: Same comment as above
```suggestion
            "writer must be a `SummaryWriter`, `str`, or `PathLike`, not None"
```

rileyajones (approved):

PR #5907, opened 2022-08-30
Allow event propogation on Fob clicks

JamesHollyer:
Event propagation was causing some when dragging the line above the fob. This prompted us to stop propagation for startDrag events. However, that propagation is needed on the fob to allow users to type in the step. This change adds logic to allow propagation of mouse events on fob clicks.

* Screenshots of UI changes
Before change:

After change:

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -95,6 +95,9 @@ export class CardFobControllerComponent {
   }
 
   startDrag(fob: Fob, affordance: TimeSelectionAffordance, event: MouseEvent) {
+    // The Fob has the contenteditable attribute. This needs the event to
+    // propagate. The contenteditable attribute also stops MouseEvent from
+    // causing the strange behavior we are avoiding by stopping propagation.
     if (affordance !== TimeSelectionAffordance.FOB) {
       this.stopEventPropagation(event);
     }
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -95,9 +95,10 @@ export class CardFobControllerComponent {
   }
 
   startDrag(fob: Fob, affordance: TimeSelectionAffordance, event: MouseEvent) {
-    // The Fob has the contenteditable attribute. This needs the event to
-    // propagate. The contenteditable attribute also stops MouseEvent from
-    // causing the strange behavior we are avoiding by stopping propagation.
+    // When clicking on the fob, we needs to allow MouseEvent propagate to the
+    // text span in the fob for editing. The contenteditable attribute also
+    // stops the event from propagating which satisfies our need to call
+    // stopEventPropagation regardless of affordance.
     if (affordance !== TimeSelectionAffordance.FOB) {
       this.stopEventPropagation(event);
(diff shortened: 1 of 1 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -95,7 +95,9 @@ export class CardFobControllerComponent {
   }
 
   startDrag(fob: Fob, affordance: TimeSelectionAffordance, event: MouseEvent) {
-    this.stopEventPropagation(event);
+    if (affordance !== TimeSelectionAffordance.FOB) {
```
JamesHollyer, line 98: What are your thoughts on using affordance for logic like this?

japie1235813: One thing I cannot figure out from this logic is, why selecting other content like text and graph will not happen if I click on the fob? Logically this condition is not satisfied and stopEventPropagation is not trigged.

japie1235813 (commented):
Thanks for the before/after video! It's really clear on what this PR has done.

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
JamesHollyer: After some testing and playing around I determined the contenteditable attribute stops the strange behavior from the event propagation. I added a comment trying to explain.

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -95,7 +95,12 @@ export class CardFobControllerComponent {
   }
 
   startDrag(fob: Fob, affordance: TimeSelectionAffordance, event: MouseEvent) {
-    this.stopEventPropagation(event);
+    // The Fob has the contenteditable attribute. This needs the event to
+    // propagate. The contenteditable attribute also stops MouseEvent from
+    // causing the strange behavior we are avoiding by stopping propagation.
```
japie1235813, line 100: I would start with saying the need and then how the behavior is handled, 
"When clicking on the fob, we needs to allow MouseEvent propagate to the text span in the fob for editing. The conteneditable atrribute in the text span also calls stopPropagation() and preventDefault(), which satisfies our need to call stopEventPropagation regardless of affordance."

It would be great to have some official document to back up the claim. Yet this is the closest thing I can find https://gist.github.com/bwindels/bef8cbddea30f4b77b50
I can see stopPropagation is called on keypress

japie1235813 (approved):
nit: PR title `propagation` typo; add category for better search later (i.e., "linked time: allows ...")

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
JamesHollyer: There is no official documentation about this. The spec only has language like "This could be triggered as the default action of keydown events with various key identifiers and as the default action of mousedown events."(the spec is here https://www.w3.org/TR/2008/WD-html5-20080610/editing.html). I know empirically that it works and that is what I was going off of.

JamesHollyer: I do not think it is correct to say that is calls stopPropagation() and preventDefault(). I believe it has similar behavior but it is not a javascript call under the hood.

PR #5908, opened 2022-08-30
Linked Time: Round to nearest integer when dragging range slider

JamesHollyer:
Steps are discrete. It does not make sense to have a slider which chooses steps to use floating point. This forces that slider to round to the nearest integer.

Since the slider is a widget I wanted to allow the user of the widget to decide if floating point was something that they wanted. Therefore I made it an optional input.

* Screenshots of UI changes

This is what is looks like in the edge case with only 2 steps.

### DIFF
--- a/tensorboard/webapp/widgets/range_input/range_input_component.ts
+++ b/tensorboard/webapp/widgets/range_input/range_input_component.ts
@@ -135,6 +135,11 @@ export class RangeInputComponent implements OnInit, OnDestroy {
    */
   @Input() enabled: boolean = true;
 
+  /**
+   * Whether the slider returns discrete integers or allows for floating points.
+   */
+  @Input() returnIntegers: boolean = false;
+
   @Output()
   rangeValuesChanged = new EventEmitter<RangeValues>();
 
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/range_input/range_input_test.ts
```diff
@@ -271,6 +277,26 @@ describe('range input test', () => {
...
+        });
+        const [leftThumb] = getThumbsOnRange(fixture);
+        startMovingThumb(leftThumb);
+        moveThumb(leftThumb, 170);
+        // Without rounding, lowerValue would be 2.4.
+        expect(onRangeValuesChanged).toHaveBeenCalledWith({
+          lowerValue: 2,
+          upperValue: 5,
+          source: 'SLIDER',
+        });
```
japie1235813, line 297: Can you add the test case when the lowerValue is 2.6 and round to 3?
Also lowerValue is 2 and the return is 2 just to be thorough?

tensorboard/webapp/widgets/range_input/range_input_component.ts
```diff
@@ -245,10 +250,14 @@ export class RangeInputComponent implements OnInit, OnDestroy {
       xPositionInPercent = compensatedRelativeXInPx / (right - left);
     }
 
-    const newValue = this.getClippedValue(
+    let newValue = this.getClippedValue(
       this.min + (this.max - this.min) * xPositionInPercent
     );
 
+    if (this.returnIntegers) {
+      newValue = Math.round(newValue);
```
rileyajones, line 258: Since we have just rounded the value there should be no concern around floating point math so could we just return here?
```suggestion
      return Math.round(newValue);
```

rileyajones (approved):
Thanks for testing the edge case with two data points. Assuming that FEELS good, I'm generally okay with this but +1 to the comment about the test.

tensorboard/webapp/widgets/range_input/range_input_component.ts
JamesHollyer: Yea that makes sense. Then I can change that back to a const.

tensorboard/webapp/widgets/range_input/range_input_test.ts
JamesHollyer: Added.

PR #5911, opened 2022-09-01
Linked Time: Sort Data Table based on header clicks

JamesHollyer:
With the introduction of the data table in Scalar Cards many more useful features become available. One of these features is sorting that data table. This change adds the ability for users to sort by any column by clicking on the header to that column.

The sortingInfo object, which holds the information about which column and what order we are sorting, is stored in the ScalarCardComponent. It is currently not put into the ngrx state. It is piped through the ScalarCardDataTable and into the DataTableComponent. The ScalarCardDataTable uses this information to sort the data passed to the DataTableComponent and the DataTableComponent itself uses the information to update the header UI and inform the information emitted when the header is clicked.

* Screenshots of UI changes

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
@@ -12,7 +12,13 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
-import {ChangeDetectionStrategy, Component, Input} from '@angular/core';
+import {
+  ChangeDetectionStrategy,
+  Component,
+  EventEmitter,
+  Input,
+  Output,
+} from '@angular/core';
 import {TimeSelection} from '../../../widgets/card_fob/card_fob_types';
 import {findClosestIndex} from '../../../widgets/line_chart_v2/sub_view/line_chart_interactive_utils';
 import {
@@ -21,6 +27,8 @@ import {
   ScalarCardPoint,
   ScalarCardSeriesMetadataMap,
   SelectedStepRunData,
+  SortingInfo,
+  SortingOrder,
 } from './scalar_card_types';
 import {TimeSelectionView} from './utils';
 
@@ -30,6 +38,8 @@ import {TimeSelectionView} from './utils';
     <tb-data-table
       [headers]="dataHeaders"
       [data]="getTimeSelectionTableData()"
+      [sortingInfo]="sortingInfo"
+      (sortDataBy)="sortDataBy.emit($event)"
     ></tb-data-table>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush,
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/widgets/data_table/data_table_component.ts
```diff
@@ -177,4 +189,15 @@ export class DataTableComponent {
...
+  headerClicked(header: ColumnHeaders) {
+    let sortingOrder = SortingOrder.ASCENDING;
+    if (
+      this.sortingInfo.header === header &&
+      this.sortingInfo.order === SortingOrder.ASCENDING
+    ) {
+      sortingOrder = SortingOrder.DESCENDING;
+    }
+    this.sortDataBy.emit({header, order: sortingOrder});
+  }
```
rileyajones, line 202: This is a nitpick, but storing and mutating the variable seems unnecessary.

```suggestion
  headerClicked(header: ColumnHeaders) {
    if (
      this.sortingInfo.header === header &&
      this.sortingInfo.order === SortingOrder.ASCENDING
    ) {
      this.sortDataBy.emit({header, order: SortingOrder.DESCENDING});
      return;
    }
    this.sortDataBy.emit({header, order: SortingOrder.ASCENDING});
  }
```

tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
```diff
@@ -183,7 +196,22 @@ export class ScalarCardDataTable {
...
+        ) {
+          return this.sortingInfo.order === SortingOrder.ASCENDING ? -1 : 1;
+        }
+        if (
+          point1[this.sortingInfo.header]! > point2[this.sortingInfo.header]!
+        ) {
+          return this.sortingInfo.order === SortingOrder.ASCENDING ? 1 : -1;
+        }
 
+        return 0;
```
rileyajones, line 212: nitpick: This if -> ternary could really just be a ternary. Sorting is always a little funny so I MIGHT have these values backwards.
```suggestion
        return this.sortingInfo.order === SortingOrder.ASCENDING ? point2[this.sortingInfo.header]! - point1[this.sortingInfo.header]! : point1[this.sortingInfo.header]! - point2[this.sortingInfo.header]!;
```

rileyajones (approved):
Generally lgtm, just a few nitpicks.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
JamesHollyer: For posterity: As we discussed the problem here is that this is sometimes text and we cannot subtract it.

japie1235813:
This is super cool! Thanks for working on the feature. I like it.

A couple of UI/UX suggestions (could be in follow-up cls as discussed):

1.Show arrow on hovering. As our hparam table under experiment list. The header does not seem clickable.
https://screenshot.googleplex.com/9bmo9i9mkXbJN3L
https://screenshot.googleplex.com/6u9BXnkY65pofG9

2. The space between columns are "flashing". Can we reserve some space for arrows?
3. The icons v and ^ means more like expansion/collapsing as in our TS card grouping. Let's switch to icons the same as our hparam table. You can find the exp by searching "test with new logger stuff prod"

These are my preliminary suggestions. Could have more thoughts after trying it. Yet I think they would be relatively smaller.

PR #5912, opened 2022-09-01
Linked Time: Removed old customVisTemplate that is no longer used

JamesHollyer:
The new implementation of Linked time moved away from this template but it has not been removed. This PR has no visible change in UI or features but removes a lot of unused code.

We have now switched over to using the lineChartCustomXAxisVis for all linked time needs.


japie1235813:
As offline discussion, please assign to me again when it's ready to review. Thanks!

japie1235813 (approved):
Thanks for cleaning this up!

PR #5913, opened 2022-09-02
Load Feature Flags from `localStorage`

rileyajones:
This makes the feature flags modal much more useful

1) Prior to making the change open TensorBoard with the `?showFlags` query param.
2) Change a flag to a non default value.
3) Open a fresh tab and note that it no longer appears and overridden in the modal.
4) Apply the change and note that it now appears as overridden.


JamesHollyer (approved):
lgtm

PR #5914, opened 2022-09-06
Swap card fobs when one is dragged past another

rileyajones:
This is how it behaves in the settings panel so it stands to reason that the fobs on the chart should behave the same.

Change the fob currently being dragged when they collide

* Screenshots of UI changes
e27da5a4-cac1-4c25-b400-3a3fccb10c46

1) Start TensorBoard
2) Navigate to [localhost:6006?enableLinkedTime&enableDataTable](http://localhost:6006?enableLinkedTime&enableDataTable)
3) Enable "Link by step" in the settings panel
4) Set a value for the end step
5) Validate you can drag the start fob past the end fob
6) Validate you can drag the end fob past the start fob

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -123,7 +123,7 @@ export class CardFobControllerComponent {
     return this.axisDirection === AxisDirection.VERTICAL;
   }
 
-  private moveCurrentFob(
+  private getNewTimeSelection(
     newStep: number,
     timeSelection: TimeSelection
   ): TimeSelection {
@@ -136,7 +136,7 @@ export class CardFobControllerComponent {
       if (newStep < this.timeSelection.start.step) {
         timeSelection.end!.step = this.timeSelection.start.step;
         this.currentDraggingFob = Fob.START;
-        return this.moveCurrentFob(newStep, timeSelection);
+        return this.getNewTimeSelection(newStep, timeSelection);
       }
       timeSelection.end!.step = newStep;
 
@@ -146,7 +146,7 @@ export class CardFobControllerComponent {
     if (this.timeSelection.end && newStep > this.timeSelection.end.step) {
       timeSelection.start.step = this.timeSelection.end.step;
       this.currentDraggingFob = Fob.END;
-      return this.moveCurrentFob(newStep, timeSelection);
+      return this.getNewTimeSelection(newStep, timeSelection);
     }
     timeSelection.start.step = newStep;
 
@@ -172,7 +172,10 @@ export class CardFobControllerComponent {
       return;
     }
 
-    const newTimeSelection = this.moveCurrentFob(newStep, this.timeSelection);
+    const newTimeSelection = this.getNewTimeSelection(
+      newStep,
+      this.timeSelection
+    );
     this.onTimeSelectionChanged.emit({
       timeSelection: newTimeSelection,
     });
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -123,33 +123,42 @@ export class CardFobControllerComponent {
     return this.axisDirection === AxisDirection.VERTICAL;
   }
 
+  private shouldSwapFobs(newStep: number) {
+    return this.currentDraggingFob === Fob.END
+      ? newStep < this.timeSelection.start.step
+      : this.timeSelection.end && newStep > this.timeSelection.end.step;
+  }
+
   private getNewTimeSelection(
     newStep: number,
     timeSelection: TimeSelection
   ): TimeSelection {
+    // Single Selection
     if (!this.timeSelection.end) {
       timeSelection.start.step = newStep;
       return timeSelection;
     }
 
-    if (this.currentDraggingFob === Fob.END) {
-      if (newStep < this.timeSelection.start.step) {
-        timeSelection.end!.step = this.timeSelection.start.step;
-        this.currentDraggingFob = Fob.START;
-        return this.getNewTimeSelection(newStep, timeSelection);
-      }
-      timeSelection.end!.step = newStep;
-
+    // Range Selection
+    // Swapping if fobs pass each other
+    if (this.shouldSwapFobs(newStep)) {
+      const [oldDraggingFob, newDraggingFob] =
+        this.currentDraggingFob === Fob.END
+          ? [Fob.END, Fob.START]
+          : [Fob.START, Fob.END];
+      timeSelection[oldDraggingFob].step =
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -123,10 +123,39 @@ export class CardFobControllerComponent {
...
+    if (!this.timeSelection.end) {
+      timeSelection.start.step = newStep;
+      return timeSelection;
+    }
+
+    if (this.currentDraggingFob === Fob.END) {
+      if (newStep < this.timeSelection.start.step) {
+        timeSelection.end!.step = this.timeSelection.start.step;
+        this.currentDraggingFob = Fob.START;
+        return this.moveCurrentFob(newStep, timeSelection);
```
JamesHollyer, line 139: Why not just set the timeSelection.start.step = newStep right here?

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -123,10 +123,39 @@ export class CardFobControllerComponent {
...
+      }
+      timeSelection.end!.step = newStep;
+
+      return timeSelection;
+    }
+
+    if (this.timeSelection.end && newStep > this.timeSelection.end.step) {
+      timeSelection.start.step = this.timeSelection.end.step;
+      this.currentDraggingFob = Fob.END;
+      return this.moveCurrentFob(newStep, timeSelection);
```
JamesHollyer, line 149: Same here. I think it would be easy to just set timeSelection.end.step = newStep and return timeSelection.

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -123,10 +123,39 @@ export class CardFobControllerComponent {
     return this.axisDirection === AxisDirection.VERTICAL;
   }
 
+  private moveCurrentFob(
```
JamesHollyer, line 126: This name is confusing as you may actually be moving the other fob. Also, it does not actually move the fob it just calculates the new timeSelection. I think we should do 1 of 2 things. The easiest it probably to just rename this to something like getNewTimeSelection. However, I also think it might be a reasonable idea to just have this function emit the event when the time is right.

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
```diff
@@ -207,6 +207,46 @@ describe('card_fob_controller', () => {
...
+      fobController.mouseMove(fakeEvent);
+      expect((fobController as any).currentDraggingFob).toEqual(Fob.END);
+
+      fobController.stopDrag();
+      fixture.detectChanges();
+
+      // expect(getStepHigherSpy).toHaveBeenCalledOnceWith(3);
+      // expect(
+      //   fobController.startFobWrapper.nativeElement.getBoundingClientRect().top
+      // ).toEqual(3);
```
JamesHollyer, line 241: remove.

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
```diff
@@ -207,6 +207,46 @@ describe('card_fob_controller', () => {
         mouseUpListener
       );
     });
+
+    it('swaps fobs being dragged when one fobs collide', () => {
```
JamesHollyer, line 211: I would like to add two more tests. 

1. A test that drags the end fob past the start fob
2. A test that starts with start.step === end.step then drags the end step down.

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
rileyajones: I mean sure... but recursion is cool? 
image

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
rileyajones: Yeah... this name is definitely a doozy, I'll change it to `getNewTimeSelection`. I'm trying to break the calculation apart from the event emitter.

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
rileyajones: Recursion :skull:

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
rileyajones: Recursion :skull:

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -36,6 +36,11 @@ export enum Fob {
   END,
 }
 
+const TIME_SELECTION_TO_FOB: Record<keyof TimeSelection, Fob> = {
+  start: Fob.START,
+  end: Fob.END,
+};
```
rileyajones, line 42: This is somewhat undesirable and should perhaps be a reason to abandon the `Fob` enum.

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -123,10 +128,48 @@ export class CardFobControllerComponent {
     return this.axisDirection === AxisDirection.VERTICAL;
   }
 
+  private shouldSwapFobs(newStep: number) {
+    return this.currentDraggingFob === Fob.END
```
JamesHollyer, line 132: IMO this does not read well. I would rather have 

(this.currentDraggingFob === Fob.END && newStep < this.timeSelection.start.step) || (this.currentDraggingFob === Fob.END && newStep > this.timeSelection.end?.step)

newStep > undefined will always return false so I do not think you need to check this.timeSelection.end

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
JamesHollyer: Abandon it in favor of what? What would this.currentDraggingFob be?

JamesHollyer (approved):

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
rileyajones: I would replace it with a type that naturally maps to `keyof TimeSelection` (maybe just use `keyof TimeSelection`). I looked at what that would take though and it is beyond the scope of this PR.

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
JamesHollyer: That way works for me.

JamesHollyer (approved):

PR #5915, opened 2022-09-06
Linked Time: Sorting Header css cleanup and hover indication

JamesHollyer:
This is a followup to PR #5911 which added the sorting functionality to the data table. This is purely a UI change to help indicate to users that you can click the header to sort and help make the feature look better.

* Screenshots of UI changes

### DIFF
--- a/tensorboard/webapp/widgets/data_table/data_table_component.ng.html
+++ b/tensorboard/webapp/widgets/data_table/data_table_component.ng.html
@@ -37,15 +37,15 @@ limitations under the License.
 
               <div
                 class="sorting-icon-container"
-                [ngClass]="header === sortingInfo.header ? 'opaque' : 'show-on-hover'"
+                [ngClass]="header === sortingInfo.header ? 'show' : 'show-on-hover'"
               >
                 <mat-icon
                   *ngIf="sortingInfo.order === SortingOrder.ASCENDING || header !== sortingInfo.header"
-                  svgIcon="arrow_downward_24px"
+                  svgIcon="arrow_upward_24px"
                 ></mat-icon>
                 <mat-icon
                   *ngIf="sortingInfo.order === SortingOrder.DESCENDING && header === sortingInfo.header"
-                  svgIcon="arrow_upward_24px"
+                  svgIcon="arrow_downward_24px"
                 ></mat-icon>
               </div>
             </div>
@@ -84,14 +84,6 @@ limitations under the License.
   </table>
 </div>
 <ng-template #arrow let-value>
-  <mat-icon
-    *ngIf="value >= 0"
-    class="delta-arrow"
-    svgIcon="arrow_upward_24px"
-  ></mat-icon>
-  <mat-icon
-    *ngIf="value < 0"
-    class="delta-arrow"
-    svgIcon="arrow_downward_24px"
-  ></mat-icon>
+  <mat-icon *ngIf="value >= 0" svgIcon="arrow_upward_24px"></mat-icon>
+  <mat-icon *ngIf="value < 0" svgIcon="arrow_downward_24px"></mat-icon>
 </ng-template>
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/widgets/data_table/data_table_component.ng.html
```diff
@@ -25,28 +25,27 @@
...
                 svgIcon="change_history_24px"
               ></mat-icon>
               <div *ngSwitchDefault class="extra-right-padding"></div>
 
               <span>{{ getHeaderTextColumn(header) }}</span>
 
-              <div *ngIf="header === sortingInfo.header">
+              <div
+                class="sorting-icon-container"
+                [ngClass]="header === sortingInfo.header ? 'opaque' : 'show-on-hover'"
```
japie1235813, line 40: nit: "opaque" -> "show" makes it more meaningful IMO

tensorboard/webapp/widgets/data_table/data_table_test.ts
```diff
@@ -255,4 +253,32 @@ describe('data table', () => {
...
+      window.getComputedStyle(
+        headerElements[2].queryAll(By.css('.sorting-icon-container'))[0]
+          .nativeElement
+      ).opacity
+    ).toBe('0');
+    expect(
+      window.getComputedStyle(
+        headerElements[3].queryAll(By.css('.sorting-icon-container'))[0]
+          .nativeElement
+      ).opacity
```
japie1235813, line 281: I probably will just check if the class "opaque" exits

tensorboard/webapp/widgets/data_table/data_table_test.ts
```diff
@@ -255,4 +253,32 @@ describe('data table', () => {
...
+    ).toBe('0');
+  });
```
japie1235813, line 283: Do we have tests for showing the icon when it is sorted on that header?

Another thing I'm wondering is, do we have tests for those icons? For example, when it's sorting on this header and it's ascending sorting, we should show down arrow_downward_24px (and we can expect getAttribute('svgIcon') to be that name)

tensorboard/webapp/widgets/data_table/data_table_test.ts
JamesHollyer: This way seems more accurate. For instance I just changed the class to "show" instead of "opaque" and this test still works. How strongly do you feel about this?

tensorboard/webapp/widgets/data_table/data_table_test.ts
JamesHollyer: I am not sure what all you are asking.

I added a check to ensure the upward arrow is there.  I also added a test for the downward arrow when order is DESCENDING. 

Does that satisfy this comment?

tensorboard/webapp/widgets/data_table/data_table_test.ts
japie1235813: In my understanding, those values related to visual factors will go to screenshot tests. `window.getComputedStyle` is making things more complex than needed. Instead of convinced by the example you provided, it actually concerns me more. The feeling of "something we are changing significantly does not fail the test" does not feel right to me.

I cannot see to find any usage of css and test that css attribute in current codebase. 
On the opposite I can find some examples for checking class not css value: instead of checking border-color this only checks the class:
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/widgets/content_wrapping_input/content_wrapping_input_test.ts;l=146?q=tensorboard%2Fwebapp%2Fwidgets%2Fcontent_wrapping_input%2Fcontent_wrapping_input_test.ts&ss=tensorflow%2Ftensorboard

I will be convinced if you find an example OR I can still live with it if you strongly prefer this.

japie1235813 (approved):

tensorboard/webapp/widgets/data_table/data_table_test.ts
JamesHollyer: The example was supposed to point out that the test I had tests what is actually important(that the user can or cannot see the icon). The name of the class is not what is important.  However since this is not tested this way elsewhere I am perfectly fine staying consistent with other tests. I have changed it.

PR #5921, opened 2022-09-09
Scalar Card Fob: Update time selection when zoom changes

rileyajones:
After zooming on the scalar card with step selection enabled the fob(s) are not guaranteed to be visible. This is because their position is dependent on the derived `timeSelection` property of the `card_fob_controller_component`.

I changed the `card_fob_controller_component` to modify its' `timeSelection` property (and thus fob positions) whenever a change to its `lowestStep` or `highestStep` is detected.

This was more complicated than expected as the component is rendered as a child of the `line_chart_component` which will always render the component at least twice due to the way it derives its `viewBox` property.

To get around duplicate logging I updated the `line_chart` to use a more identifiable default value for its `viewBox` property. This enabled me to detect whether the viewBox is in its default state and only render the child component once the chart is fully initialized.

### One Fob
e074ea87-29e1-49e8-a313-cfb4470199f5

### Two Fobs
a9c28d5b-e311-450e-8fa7-b680feb8b370

### Zooming With No Effect
89664387-969d-4615-88b8-09a2c554f643

1) Start TensorBoard.
2) Navigate to `localhost:6006?enableLinkedTime&enableDataTable#timeseries`.
3) Enable linked time in the side bar.
4) Zoom into the chart in a way the fob(s) are still within the range.
5) Assert the timeSelection is unchanged.
6) Zoom the chart to a section where the fob is outside the range.
7) Assert the fob is moved within the range.
8) Enable linked time in the side bar.
9) Repeat the steps 6-7 ensuring both fobs respond accordingly.

Going deeper and using developer console
1) Start TensorBoard.
2) Navigate to `localhost:6006?enableLinkedTime&enableDataTable#timeseries`.
3) Enable linked time in the side bar.
4) Open the developer console and clear the logs.
5) Collapse and then expand a section with a scalar card
6) Assert that no `Time Selection Changed` events were logged.

### 1) Update the line chart to use props directly rather than deriving the `viewBox` component.
I attempted to do this but quickly realized that it would be a significant refactor that is beyond the scope of this task. It might be worth looking into in the future as it would make the component more consistent with the rest of the code base and likely bring a decent performance boost.

### 2) Update the `card_fob_controller` to stop deriving its `timeSelected` property
This seems like a pretty strange thing to do as it he property originally received as an `Input` and the parent is not necessarily notified of changes.
This also turns out to be a pretty significant refactor as the `card_fob_controller` has a bunch of mouse tracking logic that relies on the object.

### 3) Update the `scalar_card_fob_controller` instead of the `card_fob_controller`
Despite the `timeSelection` originating in the `scalar_card_fob_controller` it is actually modified by the `card_fob_controller` (see option 2).

### 4) Change the line chart to use redux
I did a bunch of the work trying to get this to work but eventually came to the conclusion that the nonstandard way the components here are written makes this quite difficult. If the line chart were to be refactored to be more inline with the rest of the codebase this would become more viable.

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -19,7 +19,10 @@ import {
   ElementRef,
   EventEmitter,
   Input,
+  OnChanges,
   Output,
+  SimpleChange,
+  SimpleChanges,
   ViewChild,
 } from '@angular/core';
 import {
@@ -47,7 +50,7 @@ const TIME_SELECTION_TO_FOB: Record<keyof TimeSelection, Fob> = {
   styleUrls: ['card_fob_controller_component.css'],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class CardFobControllerComponent {
+export class CardFobControllerComponent implements OnChanges {
   @ViewChild('startFobWrapper') readonly startFobWrapper!: ElementRef;
   @ViewChild('endFobWrapper') readonly endFobWrapper!: ElementRef;
   @Input() axisDirection!: AxisDirection;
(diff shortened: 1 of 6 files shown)


rileyajones:
This could reasonably be split into a change to the line chart and a change to the fob controller, I bundled them to clarify the cause of the change but I'm happy to break them apart if anyone desires.

JamesHollyer (commented):
I am not sure we want to actually change the selected time onChange. I was thinking we should "clip" the selection just like we currently do in linked time when the selected value is outside the range of a chart. Then show the warning icon to inform users what is happening.

for reference on clipping: https://source.corp.google.com/piper///depot/google3/third_party/tensorboard/webapp/metrics/views/card_renderer/utils.ts;l=94;bpv=1;bpt=1;rcl=472866013

Issue #5708, opened 2022-05-17
tracking issue for upgrading `protobuf` to 3.19.0+ (and gRPC as well)


nfelt:
_Googlers, see also b/219030239._

This is a tracking issue for updating our `protobuf` dependency to 3.19.0 or greater. That would simplify our lives in several respects: it would solve https://github.com/tensorflow/tensorboard/issues/5703, and it would allow us to upgrade gRPC past 1.30.0 (blocked by https://github.com/grpc/grpc/issues/23311).

That update would involve both updating the `protoc` compiler we use to generate Python bindings, as well as updating the Python protobuf runtime version we depend on (see below for what that means).

This is currently blocked on TensorFlow's own upgrade (which is blocked by https://github.com/tensorflow/tensorflow/issues/53234; Googlers, see b/182876485).  We're blocked on TF because a [previous attempt](https://github.com/tensorflow/tensorboard/pull/5561) to update protobuf to 3.18.1 resulted in build breakages for other TF-ecosystem projects that rely on us, since our generated protos were incompatible with the older protobuf runtimes required by those projects.

---

Explainer: the way we depend on protobuf is a bit subtle.  We build our `.proto` files into python libraries (`py_pb2` targets) using this `tb_proto_library` build rule: https://github.com/tensorflow/tensorboard/blob/master/tensorboard/defs/protos.bzl

(Example of `tb_proto_library` usage [here](https://github.com/tensorflow/tensorboard/blob/454cfcaa3ec0ad2b2228e094f016d593d2c3ea58/tensorboard/compat/proto/BUILD#L293-L299))

The build rule depends on 2 components that are provided by the `protobuf` dependency:

- `protoc` - the protocol buffer compiler: https://github.com/tensorflow/tensorboard/blob/454cfcaa3ec0ad2b2228e094f016d593d2c3ea58/tensorboard/defs/protos.bzl#L35

- the "protobuf Python runtime": https://github.com/tensorflow/tensorboard/blob/454cfcaa3ec0ad2b2228e094f016d593d2c3ea58/tensorboard/defs/protos.bzl#L28

The `protoc` compiler is the tool that at build time actually generates `_pb2.py` files from the `.proto` files. Note however that we check the generated pb2.py files into our pip package, so that users don't need to run `protoc` themselves to execute TensorBoard locally, but when building TensorBoard with Bazel we do run `protoc` for this reason (which has its [own issues](https://github.com/tensorflow/tensorboard/issues/3213)).

[...]

PR #5922, opened 2022-09-10
Support protobuf 3.20

adamjstewart:

The conda-forge feedstock for tensorboard is attempting to move forward with protobuf support: https://github.com/conda-forge/tensorboard-feedstock/pull/64
@ngam @h-vetinari @hmaarrfk

This PR bumps the upper bound on the protobuf dependency to `< 4`. According to https://github.com/protocolbuffers/protobuf/issues/10051, the only backwards incompatible change that was introduced was between 3.X and 4.X, not between 3.19 and 3.20, hence the change in major version number.

* Screenshots of UI changes

N/A

The conda-forge feedstock for tensorboard now uses protobuf 3.20 without issue, as far as I can tell.

In the long run, we'll eventually need to support protobuf 4.X, but that's a bigger change that is being tracked in #5708.
@nfelt

### DIFF
--- a/tensorboard/pip_package/requirements.txt
+++ b/tensorboard/pip_package/requirements.txt
@@ -24,11 +24,11 @@ google-auth >= 1.6.3, < 3
 google-auth-oauthlib >= 0.4.1, < 0.5
 markdown >= 2.6.8
 numpy >= 1.12.0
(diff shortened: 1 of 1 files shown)


rileyajones (approved):

smartdolphin:
Great job!

PR #5923, opened 2022-09-10
Update requirements.txt

SMH17:
The requirements force you to use google-auth-oauthlib <0.5. However the change introduced in  google-auth-oauthlib 0.5.x (deprecation of OAuth out-of-band flow) doesn't affect any function, therefore the requirement google-auth-oauthlib<0.5 should be removed.

Removal of version constraint to restrict google-auth-oauthlib to versions < 0.5

To verify that the 0.5 changes don't affect any function it's enough to check the presence of one of these lines:
```
redirect_uri=urn:ietf:wg:oauth:2.0:oob
redirect_uri=urn:ietf:wg:oauth:2.0:oob:auto
redirect_uri=oob
```
[reference to Oauth documentation](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration)

### DIFF
--- a/tensorboard/pip_package/requirements.txt
+++ b/tensorboard/pip_package/requirements.txt
@@ -21,7 +21,7 @@ grpcio >= 1.24.3
 # Until this issue is closed
 # https://github.com/googleapis/google-cloud-python/issues/10566
 google-auth >= 1.6.3, < 3
-google-auth-oauthlib >= 0.4.1, < 0.5
+google-auth-oauthlib >= 0.4.1
 markdown >= 2.6.8
 numpy >= 1.12.0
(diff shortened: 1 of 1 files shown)


rileyajones (dismissed):

bmd3k:
We do actually use Oauth out-of-band:

https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/uploader/auth.py;l=58;drc=b306651ab5baa6d362326464a703463b4683b265

The search you are using does not seem to be flexible enough to catch all usage.

SMH17:
@bmd3k You are right. I have missed that call in uploader.

PR #5925, opened 2022-09-13
Bug Fix: Update Histogram YAxis Values to match those shown in the chart

rileyajones:
There is a bug where the lines on the histogram chart do not align with the ticks in the YAxis. This because of some fault logic around the maximum number ticks to be shown there. The issue will occur when the number of lines (steps) is fewer than the height of the chart / 15 (a seemingly arbitrary value).

I updated the logic surrounding the number of ticks to be shown so that the number of ticks <= number steps.

* Screenshots of UI changes
Before
image
image

After
image
image

1) Start tensorboard with a log directory containing histogram data. Below is some python code. which should generate valid logs.
2) Navigate to [localhost:6006?enableDataTable&enableLinkedTime](http://localhost:6006?enableDataTable&enableLinkedTime)
3) Note that the histogram chart has the same number of YAxis ticks as lines on the chart.
4) Enable linked time and set a value for the upper value to enable range selection.
Screen Shot 2022-09-13 at 4 36 40 PM
5) Note that the upper fob is placed correctly on the chart.

```
# https://www.tensorflow.org/tensorboard/get_started
import tensorflow as tf
import datetime
mnist = tf.keras.datasets.mnist

(x_train, y_train),(x_test, y_test) = mnist.load_data()
x_train, x_test = x_train / 255.0, x_test / 255.0

def create_model():
  return tf.keras.models.Sequential([
    tf.keras.layers.Flatten(input_shape=(28, 28)),
    tf.keras.layers.Dense(512, activation='relu'),
    tf.keras.layers.Dropout(0.2),
    tf.keras.layers.Dense(10, activation='softmax')
  ])

model = create_model()
model.compile(optimizer='adam',
              loss='sparse_categorical_crossentropy',
              metrics=['accuracy'])

log_dir = "logs/fit/" + datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
tensorboard_callback = tf.keras.callbacks.TensorBoard(log_dir=log_dir, histogram_freq=1)

model.fit(x=x_train, 
          y=y_train, 
          epochs=5, 
          validation_data=(x_test, y_test), 
          callbacks=[tensorboard_callback])

```

### DIFF
--- a/tensorboard/webapp/widgets/histogram/histogram_component.ts
+++ b/tensorboard/webapp/widgets/histogram/histogram_component.ts
@@ -469,17 +469,28 @@ export class HistogramComponent implements AfterViewInit, OnChanges, OnDestroy {
     }
   }
 
+  private getMaxTicks() {
+    const {height} = this.layout.contentClientRect;
+    const maxPerHeight = height / 15;
+
+    if (this.timeProperty === TimeProperty.STEP) {
+      return Math.min(this.data.length, maxPerHeight);
+    }
+
+    return maxPerHeight;
+  }
+
   private renderYAxis() {
     if (!this.scales) return;
     const yScale =
       this.mode === HistogramMode.OVERLAY
         ? this.scales.countScale
         : this.scales.temporalScale;
-    const {height} = this.layout.contentClientRect;
-    const yAxis = d3.axisRight(yScale).ticks(Math.max(2, height / 15));
-    // d3 on DefinitelyTyped is typed incorrectly and it does not allow function
-    // that takes (d: Data) => string to be specified in the parameter unlike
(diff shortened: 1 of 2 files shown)


JamesHollyer (approved):
This will fix the issue when the number of steps is less than height/15.  In your description you state that the problem only occurs then. However, IIUC the problem actually occurs when the number of steps is not evenly divisible by height/15. At that point I believe it is much less noticeable so maybe we do not care but I think it worth noting. Maybe a follow up PR to fix that issue?

rileyajones:
@JamesHollyer I was wrong about the spacing issue when number of spaces is not evenly divisible, d3 seems to handle this correctly.

PR #5927, opened 2022-09-15
Step Selector: Add Range enabled boolean to state and update based checkbox

JamesHollyer:
This is part of the effort to enable range selection in an upcoming launch. 

This simply follows the same pattern used for other settings such as the step selector.

### DIFF
--- a/tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
+++ b/tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
@@ -40,7 +40,10 @@ limitations under the License.
     </mat-checkbox>
     <span class="indent">(Scalars only)</span>
     <div class="indent range-selection" *ngIf="isRangeSelectionAllowed">
-      <mat-checkbox [disabled]="!isAxisTypeStep()"
+      <mat-checkbox
+        [checked]="isScalarStepSelectorRangeEnabled"
+        (change)="stepSelectorRangeToggled.emit()"
+        [disabled]="!isAxisTypeStep()"
         >Enable Range Selection
       </mat-checkbox>
     </div>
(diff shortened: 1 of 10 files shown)


tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
```diff
@@ -40,7 +40,10 @@ <h3 class="section-title">General</h3>
     </mat-checkbox>
     <span class="indent">(Scalars only)</span>
     <div class="indent range-selection" *ngIf="isRangeSelectionAllowed">
-      <mat-checkbox [disabled]="!isAxisTypeStep()"
+      <mat-checkbox
+        [checked]="isScalarStepSelectorRangeEnabled"
+        (change)="stepSelectorRangeToggled.emit()"
+        [disabled]="!isAxisTypeStep()"
```
bmd3k, line 46: Is it worth adding a test that the checkbox is disabled when there is a different axis type?

tensorboard/webapp/metrics/views/right_pane/settings_view_container.ts
```diff
@@ -221,6 +228,14 @@ export class SettingsViewContainer {
     );
   }
 
+  onStepSelectorRangeToggled() {
+    this.store.dispatch(
+      stepSelectorRangeToggled({
+        affordance: TimeSelectionToggleAffordance.CHECK_BOX,
```
bmd3k, line 234: Would it be worth discssing with Jie-Wei whether it is ok to reuse the TimeSelectionToggleAffordance enum for range toggling or if we should be adding a new enum for Range toggling? It seems like we were adding different enums for different settings and its not clear to me that TimeSelectionToggleAffordance applies to range toggling.

Another possibility, if range can only be toggled on/off one way, is to not specific affordance at all.

bmd3k (approved):

tensorboard/webapp/metrics/views/right_pane/settings_view_component.ng.html
JamesHollyer: Yea good idea. Added.

PR #5928, opened 2022-09-15
Added section on logging batch-level metrics to get_started.ipynb

nkovela1:
This PR adds a new section to "Get Started with Tensorboard" focused on adding custom metrics for batch-level logging during training. This is related to an issue (b/206800675) with the functionality of the update_freq parameter that causes the example code in TF docs (https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/TensorBoard) to not work as intended. Adding this section to the notebook would provide a good workaround to allow custom batch-level summaries for both subclassed and Functional API Models.

* Screenshots of UI changes
Screen Shot 2022-09-15 at 1 41 20 PM (2)
Screen Shot 2022-09-15 at 1 45 00 PM 2

Results in Tensorboard are as intended. A full run-through was executed for the notebook to ensure existing and new functionality.

Alternates included moving the new section up to after Keras model.fit() section, but I believe new section should be last, as it should be after the introduction of using tf.summary in the "Other methods" section.


docs/get_started.ipynb
```diff
@@ -80,11 +80,41 @@
...
+            "build": "Built on Tue Sep 13 06:02:56 2022 (1663074176)\nBuilt by colaboratory-release-builder@lflt2.prod.google.com:/google/src/cloud/buildrabbit-username/buildrabbit-client/google3\nBuilt as //research/colab/notebook:notebook_backend_py3\nBuild ID: dc272e53-f9ff-4991-9d28-698b0397c5c7\nBuilt from changelist 473976297 in a mint client based on //depot/google3\nBuild label: colab-20220913-060100-RC00\nBuild platform: gcc-4.X.Y-crosstool-v18-llvm-grtev4-k8\nBuild tool: Blaze, release blaze-2022.08.25-4 (mainline @469755128)\nBuilt with par options [--compress, --strip]\nCurrently running under Python 3.9.14: embedded.\n",
+            "user": "nkovela"
+          },
+          "ename": "ModuleNotFoundError",
+          "evalue": "ignored",
+          "output_type": "error",
+          "traceback": [
+            "\u001b[0;31mModuleNotFoundError:\u001b[0m No module named 'tensorflow'",
+            "",
+            "\nConsider using a custom runtime: go/colab_binary",
```
rileyajones, line 110: Non Googlers won't be able to view this link, is there an OSS alternative you could point them towards?

docs/get_started.ipynb
```diff
@@ -80,11 +80,41 @@
...
+              "/tmp/ipy-be-hphvjaxl/profile_colab/security/kernel-4b9d0a95-ce46-4d79-9bcd-7217e06076ad.json",
+              "--profile-dir",
+              "/tmp/ipy-be-hphvjaxl/profile_colab",
+              "--profile=colab",
+              "--ipython-dir=/tmp/ipy-be-vrvxubir",
+              "--no-secure"
+            ],
+            "build": "Built on Tue Sep 13 06:02:56 2022 (1663074176)\nBuilt by colaboratory-release-builder@lflt2.prod.google.com:/google/src/cloud/buildrabbit-username/buildrabbit-client/google3\nBuilt as //research/colab/notebook:notebook_backend_py3\nBuild ID: dc272e53-f9ff-4991-9d28-698b0397c5c7\nBuilt from changelist 473976297 in a mint client based on //depot/google3\nBuild label: colab-20220913-060100-RC00\nBuild platform: gcc-4.X.Y-crosstool-v18-llvm-grtev4-k8\nBuild tool: Blaze, release blaze-2022.08.25-4 (mainline @469755128)\nBuilt with par options [--compress, --strip]\nCurrently running under Python 3.9.14: embedded.\n",
+            "user": "nkovela"
+          },
```
rileyajones, line 103: Delete this debug section
```suggestion
```

docs/get_started.ipynb
```diff
@@ -11,7 +11,7 @@
     },
     {
       "cell_type": "code",
-      "execution_count": null,
+      "execution_count": 5,
```
rileyajones, line 14: Please revert this
```suggestion
      "execution_count": null,
```

docs/get_started.ipynb
```diff
@@ -80,11 +80,41 @@
     },
     {
       "cell_type": "code",
-      "execution_count": null,
+      "execution_count": 3,
```
rileyajones, line 83: Please revert this
```suggestion
      "execution_count": null,
```

docs/get_started.ipynb
```diff
@@ -481,10 +511,166 @@
     {
```
rileyajones, line 511: It looks like this image was broken before you made this change but it would be nice if you fixed it rather than just deleting it.
```suggestion
    {
      "cell_type": "markdown",
      "metadata": {
        "id": "NVpnilhEgQXk"
      },
      "source": [
        "![Quickstart Gradient Tape](https://raw.githubusercontent.com/tensorflow/tensorboard/master/docs/images/quickstart_gradient_tape.png)"
      ]
    },
    {
```

docs/get_started.ipynb
nkovela1: This was simply part of an error message when running locally because I didn't have tensorflow installed (forgot to clear my output). Sorry, fixed.

rileyajones: I thought that might be the case 👍

rileyajones:
A brief summary of our offline conversation for posterity.

This new section seems like it is a better fit under the [Logging Custom Scalars Section](https://www.tensorflow.org/tensorboard/scalars_and_keras#logging_custom_scalars section).

There are also several references to batch level logging in [api docs callbacks section](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/TensorBoard) that are currently incorrect and will be updated to point here.

rileyajones (approved):

nfelt (changes requested):
Hi - apologies for the back-and-forth, but I'm not actually sure this section is a great fit for the Get Started guide, at least not as written.

In particular, right now the guide directs users down one of two paths - either the first section if they're using `Model.fit()`, or the second section "with other methods" if they're using a custom training loop.

But this section doesn't fit very cleanly into either section - it's using tf.summary, which makes it more "advanced" than the first section, but it's also using `Model.fit()`, which makes it confusing to insert into the "with other methods" section, because the whole point of that section is that it's for when you are _not_ using Model.fit().

The most similar content to this that we have today is in the Scalars notebook where we cover custom metrics (although not batch-level ones): https://www.tensorflow.org/tensorboard/scalars_and_keras#logging_custom_scalars

My suggestion would be that we try to merge what you have here with that content into one place, that would covers how to use the tf.summary API with Keras to log summaries that aren't available by default with the callback - whether that's logging a custom metric like the learning rate example, logging batch-level metrics, etc.

That could just all go at the bottom of the Scalars notebook, or we could also consider adding a new page for it, like "Custom summaries in Keras".

docs/scalars_and_keras.ipynb
```diff
@@ -440,28 +425,374 @@
...
+        "train_writer = tf.summary.create_file_writer(log_dir)"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "8wqqNaXP2bLc"
+      },
+      "source": [
+        "Define our Tensorboard callback to log both epoch-level and batch-level metrics to our log directory and call `model.fit()` with our selected `batch_size`: "
```
nfelt, line 592: nit: can you capitalize as "TensorBoard" (capital B) here and in the other markdown cells below?

docs/scalars_and_keras.ipynb
```diff
@@ -440,28 +425,374 @@
...
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "XlcafPNY2oUW"
+      },
+      "outputs": [],
+      "source": [
+        "%tensorboard --logdir logs/batch_level"
```
nfelt, line 630: Optional, but ideally, we would include a screenshot that shows what this command output would look like - this would be rendered in the HTML version of the notebook shown on TensorFlow.org.  There's examples above like this:

```
<!-- <img class=\"tfo-display-only-on-site\" src=\"https://github.com/tensorflow/tensorboard/blob/master/docs/images/scalars_custom_lr.png?raw=1\"/> -->
```

We'd need to add another image to https://github.com/tensorflow/tensorboard/tree/master/docs/images and then reference in here in a similar way.  This could also be done later, though.

docs/scalars_and_keras.ipynb
```diff
@@ -440,28 +425,374 @@
...
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "T0eE_yjt45TN"
+      },
+      "source": [
+        "Batch-level logging can also be implemented cumulatively, averaging each batch's metrics with those of previous batches in the same epoch and resulting in a smoother training curve when logging batch-level metrics.\n",
+        "\n",
+        "NOTE: Cumulative batch-level logging can show large discontinuities at epoch boundaries due to high batch variance in accuracy and loss."
```
nfelt, line 650: Running the notebook, this doesn't actually seem to happen - the curves are completely smooth across epochs. I think that's happening because the metrics aren't being reset on epoch boundaries. The reset behavior would happen automatically if they were added to the model explicitly (via `layer.add_metric()`), but as written now they're just free-floating metric objects - I think we probably want to add them so that the behavior is more like the built-in metrics.

nfelt (commented):
Thanks for making the changes - this basically LGTM aside from a couple of formatting issues and the question about resetting the cumulative metrics. FYI, I'll be OOO tomorrow traveling, but should be good for a final approval on Wednesday - I appreciate your bearing with me, I know it's been a long wait!

docs/scalars_and_keras.ipynb
```diff
@@ -440,28 +412,202 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 10,
+      "execution_count": null,
       "metadata": {
         "id": "97T4vT3QkQJH"
       },
-      "outputs": [
```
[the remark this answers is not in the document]
nkovela1: Fixed. Added notebook with all cell outputs except `%tensorboard` magic cells.

docs/scalars_and_keras.ipynb
```diff
@@ -440,28 +412,202 @@
...
+        "## Batch-level logging\n"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "p-ltiSuypLVE"
+      },
+      "source": [
+        "First let's load the MNIST dataset, normalize the data and write a function that creates a simple Keras model for classifying the images into 10 classes."
```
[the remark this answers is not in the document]
nkovela1: Yes, this can be done in the future. I kept it with MNIST to allow for the use of more diverse custom metrics used in multiclass classification, such as sparse categorical accuracy, but this can be simplified for consolidation in a later PR.

docs/scalars_and_keras.ipynb
nkovela1: Fixed.

docs/scalars_and_keras.ipynb
nkovela1: This can be done later, as publishing this PR sooner would help with a lot of recent YAQS questions on this issue. Thanks!

docs/scalars_and_keras.ipynb
nkovela1: Thanks for pointing this out! I realized this early today. After discussing with my manager about cumulative batch-level logging today, however, we actually decided to keep this as-is on purpose, as it's a more useful use case to increase granularity in the metric curve by accumulating across epoch boundaries than to have sudden discontinuities between epochs. Those who want batch-by-batch behavior within the epoch can refer to the instantaneous case. As such, I've edited the description and removed the note on discontinuities.

nfelt: Hmm. I still have significant concerns about showing cumulative averaging across all training steps - I'd like to discuss further in a follow-up. Not resetting at epoch boundaries (and using a non-decaying average) means that any initial outlier values at the start of training will bias the rest of the curve indefinitely, which could be very misleading. For example, see this screenshot:

image

By the end of 5 epochs, the `batch_level` instantaneous loss has pretty much plateaued around 0.1, but the `batch_avg` is still up at 0.2, even though the epoch-level average loss has been below 0.2 since the second epoch (as shown in the logging that `model.fit()` produces or the built-in Keras summaries that write epoch-level loss values).  If all the user wants is smoothing of the batch-level variation, TensorBoard's smoothing algorithm can provide that with greater accuracy, and in a way that can always be adjusted or disabled later to see the true data.

I'll approve to get this PR unblocked, but if we keep this behavior, I'd prefer that you at least reinstate some kind of warning or note in the text, to ensure users realize that this method of averaging will equally weight loss from all batches seen during training, and hence won't match the built-in per-epoch loss metric.

nfelt (approved):

docs/scalars_and_keras.ipynb
```diff
@@ -462,6 +474,424 @@
...
+        "              metrics=['accuracy'])"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "IlqPZ-h_1_0z"
+      },
+      "source": [
+        "Setting up a summary writer to a different log directory:"
```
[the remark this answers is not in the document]
nkovela1: This copy is intentional, I am setting the summary writer to a different log directory for instantaneous logging (logs/batch_level) and then setting it to a different log directory for cumulative logging (logs/batch_avg).

PR #5929, opened 2022-09-18
Use all theme configuration keys

andrewseguin:
Theme configurations used internal to Google will soon be required to define all three keys of the theme config: "color", "typography", and "density".

Setting `typography: null` and `density: 0` is a no-op that has no affect


rileyajones (approved):
Thanks for helping us keep up to date!

PR #5931, opened 2022-09-19
feature flags: Refresh page after feature flag modal closed.

bmd3k:
When the feature flag modal is closed, the application is possibly in a state where it was loaded with one set of feature flags but now the feature flags have possibly changed.  This is undefined and unpredictable - functionality that is gated by feature flags generally haven't needed to consider this possibility and we really don't want to have to think about this possibility in the future.

The simple solution is to refresh the page as soon as the modal is closed. This means the application is reloaded with a stable set of feature flags.

I also change the logic that attempts to remove 'showFlags' from the URL after the modal is closed. Use featureFlagOverrideReset instead of featuerFlagOverridesChanged. The practical result for the user is effectively the same but it keeps their local storage clean of 'showFlags' setting so is a little more hygienic.


rileyajones (approved):
This definitely needs to exist but the current state has been convenient for testing some UI only changes.

PR #5933, opened 2022-09-19
Scalar Card Fob: Move fob when zoom changes

rileyajones:
Tests will be added once the open question is answered

After zooming on the scalar card with step selection enabled the fob(s) are not guaranteed to be visible.

~### Open Questions:~
#### ~Because steps must be whole numbers it is quite easy for a user to get themselves into an invalid situation where no valid steps exist. What should we do in this situation?~
405ff102-f8a4-444a-8c53-c91f274b123b

While this is an awkward user experience the user will have to get themselves into this situation so we are electing to leave this as is.

* Updated the `line_chart_component` to emit an event when the zoom level is changed.
* This event is then bubbled up to the `scalar_card_container` which is responsible for deriving the `TimeSelectionView` (i.e. the clipped `MinMaxStep` used to to determine the fob position).
* Created an `observable` around this event which is then combined with the existing `observable` listening to the series data change to ensure the `TimeSelectionView` is always recomputed

## Screenshots of UI changes
Zooming on a valid range with a single fob
9e539d8b-c225-4f9c-aa25-c7fe89c79913

Resetting view box with a single fob
3d5d15d9-91bb-4a1e-9dae-7e41816fd715

#5938 will be required to fix this issue when range selection is enabled.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -97,6 +97,7 @@ import {ScalarCardDataTable} from './scalar_card_data_table';
 import {ScalarCardFobController} from './scalar_card_fob_controller';
 import {
   ColumnHeaders,
+  MinMaxStep,
   ScalarCardPoint,
   ScalarCardSeriesMetadata,
   SeriesType,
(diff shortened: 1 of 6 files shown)


rileyajones:
At the moment (until #5938) is complete and merged enabling linked time will cause zooming will trigger the "clipped" state. I think this is okay for now because linked time is hidden behind a feature flag that is off by default and the existing functionality is even more broken

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2394,6 +2395,41 @@ describe('scalar card', () => {
...
+          newSteps = minMaxStep;
+        });
+        fixture.componentInstance.onLineChartZoom({
+          x: [9.235, 30.4],
+          y: [0, 100],
+        });
+        expect(newSteps!).toEqual({
+          minStep: 10,
+          maxStep: 30,
+        });
```
rileyajones, line 2430: I'm aware the convention is to test by validating the ui but in this case the changed ui is part of a subcomponent that does not necessarily update based on calls to `detectChanges`. I'm sure I could get it to work but this seemed cleaner.

tensorboard/webapp/widgets/line_chart_v2/line_chart_component_test.ts
```diff
@@ -385,6 +396,42 @@ describe('line_chart_v2/line_chart test', () => {
...
+      ],
+      seriesMetadataMap: {foo: buildMetadata({id: 'foo', visible: true})},
+      yScaleType: ScaleType.LINEAR,
+    });
+    spyOn(fixture.componentInstance.viewBoxChanged, 'emit');
+    fixture.detectChanges();
+    expect(updateViewBoxSpy).toHaveBeenCalledTimes(1);
+    expect(updateViewBoxSpy.calls.argsFor(0)).toEqual([
+      {x: [-0.2, 2.2], y: [-1.2, 1.2]},
+    ]);
```
JamesHollyer, line 419: Why not use toHaveBeenCalledOnceWith?

JamesHollyer (approved):
LGTM. I like that this turned out to be a pretty simple change.

tensorboard/webapp/widgets/line_chart_v2/line_chart_component_test.ts
rileyajones: Good question!

PR #5935, opened 2022-09-20
[WIP]Step Selector: store card time selections in ngrx state

JamesHollyer:
Currently there is some strange behavior in the ScalarCardComponent where we update the timeSelection(which is an Input) directly in the component. This is because it cannot be updated in the ScalarCardContainer since it is only an Observable object. This strangeness is due to the fact that we strayed away from out typical pattern when deciding to store the timeSelection for each card in the ScalarCardComponent instead of in the NGRX state. This change remedies the situation by following the prescribed pattern of keeping all state in NGRX.

There are multiple place in the code where we change the timeSelection Input object. There are a couple extra changes in here to ensure that object never changes in the component.


JamesHollyer:
This was done in #6240

PR #5938, opened 2022-09-21
Bugfix: Fix zooming with linked time by changing the definition of "clipped"

rileyajones:
Change the definition of "clipped" to "A portion of the time selection cannot be shown"

Currently when a the entirety of the time selection is outside of a chart it is determined to be "clipped" and a section is chosen to be shown.

When a card is "clipped" an icon is shown
Screenshot from 2022-09-21 14-09-49

The current behavior is somewhat confusing as it can lead to states where the card does not show all the content but still does not register as "clipped"
Screenshot from 2022-09-21 14-15-28

The issue is then further compounded by "zooming"
Screenshot from 2022-09-21 14-15-59

I updated the `maybeClipTimeSelection` method to always return the maximum time selection available within the bounds of the `minStep` and `maxStep` provided and setting `clipped` to `true` if the `timeSelection` has been changed.

## Screenshots of UI changes
Zooming with linked time enabled without range to include current step
4c9091a2-ee78-4646-9266-548b67f9e6d1

Zooming with linked time enabled to not include current step
71900b0e-78ac-4f79-8a57-8d88d0b67134 (1)

Zooming to a subset of the time selection
b2196686-1ab9-4ae9-9307-3865a89bdb74

Zooming to a partial subset of the time selection
8a8afbf5-b221-4bed-91ca-0dc5ffacf638

Zooming to a range including only a single whole number
0f188df0-e82e-4c60-aae8-e1ad2716443c

Zooming to a range including zero whole numbers
f57f3f41-3f1b-4db5-82f5-01e86aba3928

See gifs above

Not allowing zoom to effect the clipped status of a card.
I found the existing definition of clipped to be confusing and unintuitive and zooming compounded that. Additionally it would have been harder to maintain it working that way. For reference see [this commit](https://github.com/rileyajones/tensorboard/commit/3b27f4e9803a3d1c5eceb0f65264118a9987d5da) implementing it this way.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
@@ -91,10 +91,7 @@ limitations under the License.
       [value]="stepIndex"
       (input)="onSliderInput($event)"
     ></mat-slider>
-    <div
-      class="linked-time-wrapper"
-      *ngIf="linkedTimeSelection"
-    >
+    <div class="linked-time-wrapper" *ngIf="linkedTimeSelection">
       <span *ngIf="linkedTimeSelection.endStep !== null">
         <span class="slider-track"></span>
         <span
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
@@ -82,7 +82,7 @@ limitations under the License.
     <mat-slider
       class="step-slider"
       color="primary"
-      [ngClass]="[linkedTimeSelection && !linkedTimeSelection.clipped && linkedTimeSelection.endStep !== null ? 'hide-slider' : '']"
+      [ngClass]="[linkedTimeSelection && linkedTimeSelection.endStep !== null ? 'hide-slider' : '']"
       [disabled]="steps.length <= 1"
       [min]="0"
       [max]="steps.length - 1"
@@ -93,7 +93,7 @@ limitations under the License.
     ></mat-slider>
     <div
       class="linked-time-wrapper"
-      *ngIf="linkedTimeSelection && !linkedTimeSelection.clipped"
+      *ngIf="linkedTimeSelection"
     >
       <span *ngIf="linkedTimeSelection.endStep !== null">
         <span class="slider-track"></span>
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
@@ -263,12 +263,11 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
 
           const minStep = Math.min(...steps);
           const maxStep = Math.max(...steps);
-          const potentiallyClippedTimeSelection = maybeClipLinkedTimeSelection(
+          return maybeClipLinkedTimeSelection(
             linkedTimeSelection,
             minStep,
             maxStep
           );
-          return potentiallyClippedTimeSelection;
         })
       );
 
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
@@ -261,17 +261,14 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
         map(([linkedTimeSelection, steps]) => {
           if (!linkedTimeSelection) return null;
 
-          let minStep = Infinity;
-          let maxStep = -Infinity;
-          for (const step of steps) {
-            minStep = Math.min(step, minStep);
-            maxStep = Math.max(step, maxStep);
-          }
-          return maybeClipLinkedTimeSelection(
+          const minStep = Math.min(...steps);
+          const maxStep = Math.max(...steps);
+          const potentiallyClippedTimeSelection = maybeClipLinkedTimeSelection(
             linkedTimeSelection,
             minStep,
             maxStep
           );
+          return potentiallyClippedTimeSelection;
         })
       );
 
@@ -287,16 +284,11 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
           return [];
         }
 
-        const selectedStepsInRange = [];
-        for (const step of steps) {
-          if (
+        return steps.filter(
+          (step) =>
             step >= linkedTimeSelection.startStep &&
-            step <= linkedTimeSelection.endStep
-          ) {
-            selectedStepsInRange.push(step);
-          }
-        }
-        return selectedStepsInRange;
+            step <= linkedTimeSelection.endStep!
+        );
       })
     );
 
--- a/tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
@@ -564,28 +564,7 @@ describe('image card', () => {
         );
       });
 
-      it('renders ticks on selected steps are particially in range', () => {
-        store.overrideSelector(selectors.getMetricsLinkedTimeSelection, {
-          start: {step: 25},
-          end: {step: 350},
-        });
-
-        const fixture = createImageCardContainer('card1');
-        fixture.detectChanges();
-
-        const dots = fixture.debugElement.queryAll(
-          By.css('.linked-time-wrapper .linked-time-tick')
-        );
-        // The third and fourth ticks are selected.
-        expect(dots[0].nativeElement.getAttribute('style')).toBe(
-          TICKS_STYLE[2]
-        );
-        expect(dots[1].nativeElement.getAttribute('style')).toBe(
-          TICKS_STYLE[3]
-        );
-      });
-
-      it('does not render ticks on slected range wrapped between steps ', () => {
+      it('does not render ticks on selected range wrapped between steps ', () => {
         store.overrideSelector(selectors.getMetricsLinkedTimeSelection, {
           start: {step: 11},
           end: {step: 14},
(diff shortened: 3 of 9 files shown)


tensorboard/webapp/metrics/views/card_renderer/utils_test.ts
```diff
@@ -310,4 +311,140 @@ describe('metrics card_renderer utils test', () => {
...
+          4
+        )
+      ).toEqual({
+        startStep: 1,
+        endStep: null,
+        clipped: true,
+      });
+    });
+
+    it('changes end when end.step < maxStep', () => {
```
japie1235813, line 333: ditto: something like "clips to maxStep when time selection end step is greater than view extend" ?

tensorboard/webapp/metrics/views/card_renderer/utils_test.ts
```diff
@@ -310,4 +311,140 @@ describe('metrics card_renderer utils test', () => {
...
+          3
+        )
+      ).toEqual({
+        startStep: 0,
+        endStep: 3,
+        clipped: true,
+      });
+    });
+
+    it('does not create an endstep when timeSelection does not contain one', () => {
```
japie1235813, line 350: "does not clip when time selection falls into the view extend"?

tensorboard/webapp/metrics/views/card_renderer/utils_test.ts
```diff
@@ -310,4 +311,140 @@ describe('metrics card_renderer utils test', () => {
...
+          50
+        )
+      ).toEqual({
+        startStep: 30,
+        endStep: 50,
+        clipped: true,
+      });
+    });
+
+    it('returns startStep === endStep === maxStep when timeSelection is above maxStep', () => {
```
japie1235813, line 384: "startStep === endStep === maxStep" is not meaningful to me. I think what you try to convey is "clips both fob to maxStep"?

tensorboard/webapp/metrics/views/card_renderer/utils_test.ts
```diff
@@ -310,4 +311,140 @@ describe('metrics card_renderer utils test', () => {
...
+          30
+        )
+      ).toEqual({
+        startStep: 20,
+        endStep: 20,
+        clipped: true,
+      });
+    });
+
+    it('returns clipped === false when provided timeSelection is a subspace of minStep - maxStep', () => {
```
japie1235813, line 418: nit: "does not clip when time selection falls within the view extend"? 
view extend, card step range.. etc

tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -96,45 +96,27 @@ export function maybeClipLinkedTimeSelection(
   minStep: number,
   maxStep: number
 ): TimeSelectionView {
+  const startStep = Math.min(
+    Math.max(minStep, timeSelection.start.step),
+    maxStep
+  );
+  const endStep = timeSelection.end
+    ? Math.max(Math.min(maxStep, timeSelection.end.step), minStep)
+    : null;
```
japie1235813, line 105: Would it be worth adding comments on how do you set startStep and endStep? It does take me a while to check the logic with max wrapped in min and verse vice .

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -564,28 +564,7 @@ describe('image card', () => {
         );
       });
 
-      it('renders ticks on selected steps are particially in range', () => {
```
japie1235813, line 567: Can you justify the reason for deleting this test? I don't understand.

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
```diff
@@ -726,42 +705,6 @@ describe('image card', () => {
         );
       });
 
-      it('renders range slider on selected steps which end step is out of range, ', () => {
```
japie1235813, line 729: Same here. What would be image range slider looks like in this PR?

I'm also wondering any screenshot changes. (We should have tests for image in range)

japie1235813 (commented):
Thanks for listing out the cases with gif! That makes the PR clear and easy to examine.

Question about zooming to outside range on range selection. Do we want to just show a single fob or two fobs overlapping? (From the code it looks like we will have two)

After thinking about this. I think it makes sense to have two. Then user can drag one and leave another at the maxStep/minStep.

tensorboard/webapp/metrics/views/card_renderer/image_card_test.ts
rileyajones: I think I misread this test originally.

tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
```diff
@@ -82,7 +82,7 @@
     <mat-slider
       class="step-slider"
       color="primary"
-      [ngClass]="[linkedTimeSelection && !linkedTimeSelection.clipped && linkedTimeSelection.endStep !== null ? 'hide-slider' : '']"
+      [ngClass]="[linkedTimeSelection && linkedTimeSelection.endStep !== null ? 'hide-slider' : '']"
```
rileyajones, line 85: Because I am changing the definition of "clipped" the image card now needs to render the slider when the time selection is clipped.

**Question**: Should We hide the slider when the `timeSelection` is fully outside of the `minMaxStep` instead?

japie1235813: I would prefer to use the double slider than single (i.e. still hide it). Although from UI perspective it's not making a lot of differences (the line is colored on single slider vs gray on double slider)
single 
=====o----
double
-------o----

tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
```diff
@@ -91,10 +91,7 @@
       [value]="stepIndex"
       (input)="onSliderInput($event)"
     ></mat-slider>
-    <div
-      class="linked-time-wrapper"
-      *ngIf="linkedTimeSelection && !linkedTimeSelection.clipped"
-    >
+    <div class="linked-time-wrapper" *ngIf="linkedTimeSelection">
```
japie1235813, line 94: Do we need to check if endStep === null here?

tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -261,12 +261,8 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
...
           if (!linkedTimeSelection) return null;
 
-          let minStep = Infinity;
-          let maxStep = -Infinity;
-          for (const step of steps) {
-            minStep = Math.min(step, minStep);
-            maxStep = Math.max(step, maxStep);
-          }
+          const minStep = Math.min(...steps);
+          const maxStep = Math.max(...steps);
```
japie1235813, line 265: Interesting. Didn't know you can do this.

tensorboard/webapp/metrics/views/card_renderer/image_card_container.ts
```diff
@@ -287,16 +283,11 @@ export class ImageCardContainer implements CardRenderer, OnInit, OnDestroy {
...
+          (step) =>
             step >= linkedTimeSelection.startStep &&
-            step <= linkedTimeSelection.endStep
-          ) {
-            selectedStepsInRange.push(step);
-          }
-        }
-        return selectedStepsInRange;
+            step <= linkedTimeSelection.endStep!
+        );
```
japie1235813, line 290: Thanks for the cleanup!

tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -96,45 +96,32 @@ export function maybeClipLinkedTimeSelection(
...
+   *      This ensures the result is <= maxStep.
+   */
+  const startStep = Math.min(
+    Math.max(minStep, timeSelection.start.step),
+    maxStep
+  );
+  /**
+   * 1) Take smallest between the maxStep and the current step.
+   *      This ensures that the current step is <= maxStep
+   * 2) Take the greatest between the output of step 1 and the minStep.
```
japie1235813, line 112: What is "step 1"?

tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -96,45 +96,32 @@ export function maybeClipLinkedTimeSelection(
...
+   *      This ensures the result is >= minStep.
+   */
+  const endStep = timeSelection.end
+    ? Math.max(Math.min(maxStep, timeSelection.end.step), minStep)
```
japie1235813, line 116: Does this equivalent to 
`Math.min(Math.max(minStep, timeSelection.end.step), maxStep)?`

japie1235813 (approved):
Overall lgtm. Thanks for the PR!
Some question regards to the comment and clipped logic but I can okay with whatever changes you make or choose not to make.

tensorboard/webapp/metrics/views/card_renderer/utils.ts
rileyajones: Would it make more sense if I called it "Bullet Point #1"

tensorboard/webapp/metrics/views/card_renderer/utils.ts
rileyajones: Yes. I'm sensing that these nested `Math.min` / `Math.max` calls are not as readable as I had hoped. I'm going to abstract this logic to a helper function to improve readability.

tensorboard/webapp/metrics/views/card_renderer/utils.ts
rileyajones: I'm refactoring the logic this refers to so I don't think this comment will be needed anymore.

tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
rileyajones: At the moment I'm not actually changing the behavior here. I personally don't have an opinion on the slider but I would be happy to change it in a future PR.

For reference:
Nothing Enabled (Also looks the same with just Step Selection enabled)
Before:
image

After:
image

Linked Time Enabled
Before:
983f8d79-70d3-4e84-8965-c766ee28cceb

After:
e46aaa25-b065-49c2-aa71-90f9f41c0094

tensorboard/webapp/metrics/views/card_renderer/image_card_component.ng.html
rileyajones: Are you suggesting removing the null check on line 98 -> 95?

The behavior is the same with and without it.

50b2833a-7856-4ea6-bffb-562c25393ca6

tensorboard/webapp/metrics/views/card_renderer/utils.ts
```diff
@@ -91,50 +91,32 @@ export interface TimeSelectionView {
...
-  if (maxStep <= timeSelection.start.step) {
-    return {
-      startStep: maxStep,
-      endStep: null,
-      clipped: true,
-    };
-  }
-  // Case when time selection is on the left of the minimum of the time
-  // series.
+  const startStep = boundNumber(timeSelection.start.step, minStep, maxStep);
```
japie1235813, line 109: As offline discussion, propose some name changes:

const startStep = timeSelection.start.step;
const maybeClippedStartStep = clipStepWithinMinMax(startStep, minStep, maxStep)
...
const clipped = maybeClippedStartStep !== startStep || ...
    
What do you think?

japie1235813 (approved):

PR #5940, opened 2022-09-21
chore: add strictNullChecks to tsconfig

japie1235813:
Internal code health to turn on strictNullChecks. Googlers, please see cl/474951173 and cl/474976805 for desired result and cl/469760530 for sync.

### DIFF
--- a/tensorboard/components/tf_storage/storage.ts
+++ b/tensorboard/components/tf_storage/storage.ts
@@ -17,6 +17,7 @@ import {
   addHashListener,
   addStorageListener,
   fireStorageChanged,
+  ListenKey,
   removeHashListenerByKey,
   removeStorageListenerByKey,
 } from './listeners';
@@ -112,8 +113,8 @@ export function makeBindings<T>(
   getObserver: (key: string, options: AutoStorageOptions<T>) => Function;
   disposeBinding: () => void;
 } {
-  const hashListeners = [];
-  const storageListeners = [];
+  const hashListeners: ListenKey[] = [];
+  const storageListeners: ListenKey[] = [];
   function get(key: string, options: StorageOptions<T> = {}): T {
     const {defaultValue, useLocalStorage = false} = options;
     const value = useLocalStorage
--- a/tensorboard/components/tf_storage/storage.ts
+++ b/tensorboard/components/tf_storage/storage.ts
@@ -120,7 +120,7 @@ export function makeBindings<T>(
     const value = useLocalStorage
       ? window.localStorage.getItem(key)
       : componentToDict(readComponent())[key];
-    return value == undefined ? _.cloneDeep(defaultValue) : fromString(value);
+    return value == undefined ? _.cloneDeep(defaultValue)! : fromString(value)!;
   }
   function set(key: string, value: T, options: SetterOptions<T> = {}): void {
     const {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,6 +14,7 @@
     "noImplicitOverride": true,
     "skipLibCheck": true,
     "strict": true,
+    "strictNullChecks": true,
     // Don't scan the node_modules/@types folder for ambient types.
     // This would force us to have all the types in the dependencies of
     // each library.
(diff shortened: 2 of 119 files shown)


tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "noImplicitOverride": true,
     "skipLibCheck": true,
     "strict": true,
+    "strictNullChecks": true,
```
bmd3k, line 17: I think this is the wrong place. I believe all the polymer code is built with strict_checks = False, which triggers:

https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/defs/defs.bzl;l=100;drc=0df21fcb82f92cec7549fb19e1db43e26012988f

And, so, uses this tsconfig instead:

https://cs.opensource.google/tensorflow/tensorboard/+/master:tsconfig-lax.json

Have you been able to configure tsconfig such that the build fails without the changes from cl/474951173 and cl/474976805?

You don't have to modify this file because it already has "strict": true, which I believe already enables strictNullChecks.

https://www.typescriptlang.org/tsconfig#strict

japie1235813: Thanks! Yap I was confused by not seeing build error from this tsconfig change. Didn't know the tsconfig files are separate. Will check again.

tensorboard/components/tf_storage/storage.ts
```diff
@@ -112,8 +113,8 @@ export function makeBindings<T>(
   getObserver: (key: string, options: AutoStorageOptions<T>) => Function;
   disposeBinding: () => void;
 } {
-  const hashListeners = [];
-  const storageListeners = [];
+  const hashListeners: ListenKey[] = [];
+  const storageListeners: ListenKey[] = [];
   function get(key: string, options: StorageOptions<T> = {}): T {
     const {defaultValue, useLocalStorage = false} = options;
     const value = useLocalStorage
```
gunan, line 120: For line 123, please change it to :

```
return value == undefined ? _.cloneDeep(defaultValue)! : fromString(value)!;
```

japie1235813: So this PR wasn't up-to-dated when you're reviewing it. I created the internal cl but didn't update here. Do you prefer me to update here or create internal cl? cl/469760530

gunan: however you prefer. I thought the PR had to go in first as tensorboard is github first?
but for me, updating google side is the important part.

gunan: Just wanted to check, the last build failures seem to be due to download errors.
Would you like me to inspect the build failures here, or the build logs on the internal cl?

japie1235813: In our process we developed in GH first. I just wonder if it's more convenient to you to have an internal cl (I guess you are using internal script or so to add those checks?)
So I would say which you prefer and/or more convenient for you to update those type syntax then.
If you're going to examine this in OSS and willing to build the env (steps provided in b/), you can also patch this PR and add fixes. However you prefer

Regards to the error, I'm not sure what you mean by download errors. They seem like type check to me 
Screen Shot 2022-09-29 at 9 26 21 AM

gunan: I have been confused by these errors:

```
Suppressed: com.google.devtools.build.lib.remote.http.DownloadTimeoutException: Download of '/tensorboard-build-cache/cas/15f51cafb9422a55951f104d456d2854ebfe0bbf4b2bfe9c63e395d14657866a' timed out. Received 0 bytes.
		at com.google.devtools.build.lib.remote.http.HttpDownloadHandler.exceptionCaught(HttpDownloadHandler.java:156)
		at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:302)
		at io.netty.channel.AbstractChannelHandlerContext.invokeExceptionCaught(AbstractChannelHandlerContext.java:281)
```

but now I see the errors.

I will setup the local development and patch your branch and build. I will comment the fixes in a bit, hopefully

japie1235813: Ah I see. That's an unimportant warning and you can ignore those. The critical part is shown in the screenshot. If that builds the DownloadTimeoutException error might go away. Please ping me if you need anything

bmd3k (approved):

PR #5941, opened 2022-09-22
Update `scalar_card_container` to use a `BehaviorSubject`

rileyajones:
The scalar card container holds state (`stepSelectorTimeSelection`) which is then read by scalar card component and scalar card fob controller. Both of these child components potentially need to be able to update this state so converting I am `stepSelectorTimeSelection` to a `BehaviorSubject` to allow it to easily broadcast changes.

* Screenshots of UI changes
N/A

There should be no changes, all the tests should pass.

This could have been another kind of `Subject` but in this case having an initial value was desirable.

We could have done this as a pure observable and maintained a list of subscribers but that seems unnecessarily messy.


tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -182,7 +189,7 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
   dataSeries$?: Observable<ScalarCardDataSeries[]>;
   chartMetadataMap$?: Observable<ScalarCardSeriesMetadataMap>;
   linkedTimeSelection$?: Observable<TimeSelectionView | null>;
-  stepSelectorTimeSelection$?: Observable<TimeSelection | null>;
+  stepSelectorTimeSelection$ = new BehaviorSubject<TimeSelection | null>(null);
```
JamesHollyer, line 192: I think this should be moved out of this list of declarations.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -533,14 +540,14 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
...
+    combineLatest([
       this.minMaxSteps$,
       this.store.select(getMetricsStepSelectorEnabled),
-    ]).pipe(
-      map(([{minStep}, enableStepSelector]) => {
-        return enableStepSelector ? {start: {step: minStep}, end: null} : null;
-      })
-    );
+    ]).subscribe(([{minStep}, enableStepSelector]) => {
+      this.stepSelectorTimeSelection$.next(
```
JamesHollyer, line 547: So, to be clear, this will reset the stepSelectorTimeSelection to the min step if minMaxSteps$ changes? I understand this already happens. I just want to make sure I understand. Also, now that this is a BehaviorSubject we can view the actual value right? So we could for instance check to see if the current value is less than the new min value or something along those lines?

JamesHollyer (commented):
This is awesome! thanks

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
rileyajones: Yes this will reset to `minStep` when `minMaxSteps$` changes.
Yes we can view the value with `getValue`

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
rileyajones: I moved this to be with the selectors and made it `readonly` as I think it is doing the same thing in spirit. LMK if you'd rather it live someplace else, there isn't a ton of precedent for this.

JamesHollyer (approved):

Issue #5924, opened 2022-09-13
Projector hangs on Computing PCA when using Tensorboard2.10 with Pytorch


XiangL-Repo:
it seems tensorboard 2.10 introduces this problem, I tried to revert to 2.9.1 version. and the hanging disappears.

rileyajones:
Could you please provide some clear steps on how to reproduce this?

Bomme:
I can reproduce the issue with `torch==1.12.1 tensorboard==2.10.0` by running the following code:

```python
import torch
from torch.utils.tensorboard import SummaryWriter

x = torch.rand((100, 25))

writer = SummaryWriter()
writer.add_embedding(x)
```

and then opening the projector view.

rileyajones:
This seems to have been caused by the bundler change in #5829

baridfor3:
Any solution? I got the same issue.

rileyajones:
@baridfor3 We are currently recommending using 2.9.1, this is a build issue and it's hard to say how long it will take to get this patched.

baridfor3:
@rileyajones Thanks for your answer.   2.9.1 works for me.

bmd3k:
Thanks for bringing this to our attention. The projector plugin should now be fixed in TensorBoard 2.10.1.

PR #5944, opened 2022-09-22
projector: Fix esbuild-bundled projector.

bmd3k:
The change to use the esbuild bundler (https://github.com/tensorflow/tensorboard/pull/5829) broke the projector plugin (https://github.com/tensorflow/tensorboard/issues/5924).

The root cause is in the 'numeric' library, which we use for calculating the PCA. The library requires that the symbol 'numeric' is available in the global scope when its operations are executed by other modules. See, for example, how the definition of some of its operations refer to the string 'numeric' in the Function definition, unmodifiable by the
bundler/minification code:

https://github.com/sloisel/numeric/blob/656fa1254be540f428710738ca9c1539625777f1/src/numeric.js#L696

The esbuild bundler does not keep 'numeric' in global scope and instead renames it as part of bundling/minification. We work around this by manually adding it to global scope.

We introduce the wrapper tensorboard/webapp/third_party/numeric.ts to add 'numeric' to the global scope and require that tensorboard code import that wrapper rather than importing numeric directly. We add a CI check to ensure that numeric is not imported directly.


tensorboard/webapp/third_party/numeric.ts
```diff
@@ -0,0 +1,32 @@
...
+// example, how the definition of some of its operations refer to the string
+// 'numeric' in the Function definition, unmodifiable by the
+// bundler/minification code:
+//
+// https://github.com/sloisel/numeric/blob/656fa1254be540f428710738ca9c1539625777f1/src/numeric.js#L696
+//
+// The esbuild bundler does not keep 'numeric' in global scope and instead
+// renames it as part of bundling/minification. We work around this by manually
+// adding it to global scope here.
+window['numeric'] = window['numeric'] ?? numeric;
```
rileyajones, line 28: This makes sense as a fix given the circumstance but it's still pretty gross. Looking at the situation though I'm not really sure how it was working before.
It seems like the `numeric` library hasn't been updated since 2012 so at some point (clearly not a priority) it would be nice to move away from it.

rileyajones (approved):

PR #5947, opened 2022-09-22
Step Selector: Stop modifying timeSelection Input

JamesHollyer:
We used to modify the timeSelection Input object in the ScalarCardComponent. Since #5941 This object is now updated in the ScalarCardContainer as it should be. However, the modifications in lower level components are still there. This change removes those modifications and completely relies on the container changes. 

I also took the opportunity to do some cleanup that I thought was nice. I moved the lineChartZoom$ declaration down to the other BehaviorSubject definitions and defined it as readonly. I also made the minMaxStep Observable a BehaviorSubject so that I can use the value to check the new timeSelections before modifying it. This check was previously done in the ScalarCardComponent since we did not have access to the minMaxStep values in the container. The logic makes more sense in the Container and allows us to remove that function entirely and just emit the event directly.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
@@ -99,6 +99,8 @@ import {
   TimeSelectionView,
 } from './utils';
 
+const DEFAULT_MIN = -Infinity;
+const DEFAULT_MAX = Infinity;
 type ScalarCardMetadata = CardMetadata & {
   plugin: PluginType.SCALARS;
 };
@@ -191,18 +193,20 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
   dataSeries$?: Observable<ScalarCardDataSeries[]>;
   chartMetadataMap$?: Observable<ScalarCardSeriesMetadataMap>;
   linkedTimeSelection$?: Observable<TimeSelectionView | null>;
-  minMaxSteps$?: Observable<MinMaxStep>;
   columnHeaders$?: Observable<ColumnHeaders[]>;
-
-  lineChartZoom$ = new BehaviorSubject<MinMaxStep>({
-    minStep: -Infinity,
-    maxStep: Infinity,
-  });
 
   onVisibilityChange({visible}: {visible: boolean}) {
     this.isVisible = visible;
   }
 
+  readonly minMaxSteps$ = new BehaviorSubject<MinMaxStep>({
+    minStep: DEFAULT_MIN,
+    maxStep: DEFAULT_MAX,
+  });
+  readonly lineChartZoom$ = new BehaviorSubject<MinMaxStep>({
+    minStep: DEFAULT_MIN,
+    maxStep: DEFAULT_MAX,
+  });
   readonly stepSelectorTimeSelection$ =
     new BehaviorSubject<TimeSelection | null>(null);
   readonly useDarkMode$ = this.store.select(getDarkModeEnabled);
@@ -348,11 +352,8 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
       shareReplay(1)
     );
 
-    this.minMaxSteps$ = combineLatest([
-      partitionedSeries$,
-      this.lineChartZoom$,
-    ]).pipe(
-      map(([series, viewPort]) => {
+    combineLatest([partitionedSeries$, this.lineChartZoom$]).subscribe(
+      ([series, viewPort]) => {
         const allPoints = series
           .map(({points}) => points.map(({x}) => x))
           .flat();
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -146,10 +146,12 @@ export class CardFobControllerComponent {
     newStep: number,
     timeSelection: TimeSelection
   ): TimeSelection {
+    let newTimeSelection = {start: timeSelection.start, end: timeSelection.end};
```
rileyajones, line 149: I'm super happy to see this cleanup.
I think it makes more sense to remove the argument and just reference the property directly. Also you never reassign `newTimeSelection` so it can be `const`
```suggestion
    const newTimeSelection = {...this.timeSelection};
```

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -191,18 +191,20 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
...
   }
 
+  readonly minMaxSteps$ = new BehaviorSubject<MinMaxStep>({
+    minStep: -Infinity,
+    maxStep: Infinity,
+  });
+  readonly lineChartZoom$ = new BehaviorSubject<MinMaxStep>({
+    minStep: -Infinity,
+    maxStep: Infinity,
+  });
```
rileyajones, line 207: Can you break this default value out into a constant like I am doing in go/tbpr/5938?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -611,10 +610,38 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
...
+        step:
+          newEnd.step < minStep
+            ? minStep
+            : newEnd.step > maxStep
+            ? maxStep
+            : newEnd.step,
+      };
+    }
+
+    const newTimeSelection = {start: newStart, end: newEnd};
```
rileyajones, line 636: Maybe it's just me but the two sets of nested ternaries are pretty hard to read.

```suggestion
    const {minStep, maxStep} = this.minMaxSteps$.getValue();
    function bound(value: number) {
      if (value > maxStep) return maxStep;
      if (value < minStep) return minStep;
      return value;
    }
    const start = bound(newTimeSelectionWithAffordance.timeSelection.start.step);
    const end = newTimeSelectionWithAffordance.timeSelection.end ? 
                   bound(newTimeSelectionWithAffordance.timeSelection.end.step) :
                   null;
    const newTimeSelection = {start, end};
```

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
JamesHollyer: Done.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
JamesHollyer: I am not sure what you are referring to in that PR. However, I made a constant these values.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -611,10 +612,28 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
...
+          : start.step,
+    };
+
+    // TODO: bound end step when range selection is implemented.
+
+    const newTimeSelection = {
+      start,
+      end: newTimeSelectionWithAffordance.timeSelection.end,
+    };
+
```
rileyajones, line 633: This does break range selection when linked time is enabled.

image

I was hopeful it wouldn't because of the existing bounding logic used to compute the `minMaxStep` however that logic is only used to compute the fob position. I'm not sure why this logic needs to be different though

```suggestion
    const {minStep, maxStep} = this.minMaxSteps$.getValue();
    const {startStep, endStep} = maybeClipLinkedTimeSelection(newTimeSelectionWithAffordance.timeSelection, minStep, maxStep);
    const newTimeSelection = {
      start: {step: startStep},
      end: endStep ? {step: endStep} : null
    };
```

My PR changing the clipping logic makes this experience a little bit better but I think it is fine even in its current form.

Before #5938
Step  Selection
84ee8f0d-e08b-4f25-ae1b-f925b9a0abec

Linked Time + Time Selection
02d4a32c-3f3e-4201-b075-5807845d79a7

After #5938
Step Selection
2750b40b-3745-46e7-9001-e735b4ca5fd9

Linked Time + Time Selection
60db7e06-48af-4e61-94d8-8c7544ed7c0e

JamesHollyer: I like it. Done.

rileyajones (approved):
Awesome, thanks for fixing this

PR #5951, opened 2022-09-27
Access colors through theme color config

andrewseguin:
Angular Material theme configs put color in a "color" map in the theme config. The colors stored on the first-level are deprecated and copies for backwards compatibility, and will soon be removed.

Also, explicitly include typography rather than including it via core. This exclusion will soon be required for all apps


andrewseguin:
An upload step failed and I'm not sure how to make it pass, is it a flake?

rileyajones (approved):

PR #5953, opened 2022-09-27
Bug Fix: Fix data table sorting of nans

rileyajones:
Sorting in a column which contains `NaN` in the data table used to break the sorting.

Sorting was broken because `NaN` values were actually being passed in as strings which lead to all the values in the column being sorted as though they were strings. To resolve this I translate all values into "sortable" versions which serves to additionally guard against other potentially problematic types.

1) Start tensorboard with a logdir which contains runs with `NaN` values.
2) Go to localhost:6006?enableDataTable
3) Enable step selection and select a step in a scalar card where some but not all values are `NaN`
4) Sort the table by the `value` column and ensure that sorting makes sense.

We could run all incoming series data through a map and convert the string `"NaN"` to `NaN` so sorting would work. That solution seems a bit hackier though and doesn't do as much to guard against other sorting anomalies.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
@@ -2972,6 +2972,70 @@ describe('scalar card', () => {
       expect(data[1].RUN).toEqual('run1');
       expect(data[2].RUN).toEqual('run2');
     }));
+
+    it('Correctly orders NaNs', fakeAsync(() => {
+      const runToSeries = {
+        run1: [{wallTime: 1, value: 1, step: 1}],
+        run2: [{wallTime: 1, value: 2, step: 1}],
+        run3: [{wallTime: 1, value: 3, step: 1}],
+        run4: [{wallTime: 1, value: NaN, step: 1}],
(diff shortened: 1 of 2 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2972,6 +2972,70 @@ describe('scalar card', () => {
...
+      const data =
+        scalarCardDataTable.componentInstance.getTimeSelectionTableData();
+
+      // expect(data[0].RUN).toEqual('run4');
+      // expect(data[1].RUN).toEqual('run5');
+      // expect(data[2].RUN).toEqual('run6');
+      // expect(data[3].RUN).toEqual('run7');
+      // expect(data[4].RUN).toEqual('run1');
+      // expect(data[5].RUN).toEqual('run2');
+      // expect(data[6].RUN).toEqual('run3');
```
JamesHollyer, line 3030: remove commented code.

JamesHollyer (approved):

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
rileyajones: Oops, done

PR #5956, opened 2022-09-29
Step Selector: Enable Range Selection for Scalar Card

rileyajones:
~Blocked by #5962~
~Blocked by #5965~

# Motivation for features / changes
We would like the feature to view data related to a range of steps.

# Technical description of changes
Now that both step selection and linked time selection both support range selection I am merging the object used to pass their values down through the scalar card component tree. This provides an opaque interface for the sub-components to utilize without needing to care about whether linked time is enabled.

# Screenshots of UI changes
## Enabling Range Selection
771710fb-bbcd-4914-ab26-7b8dd8a2af8f

## Disabling Range Selection
49ff19c9-915d-40fe-a4db-c1d71343d6d9
33b28465-0f0f-414a-8743-a69cf135916e
b98ac7bd-a8e5-4300-823e-3f1599cf5644

## Enabling Linked Time With Range Selection
eb6150e1-117c-4adc-b766-5f1f7cb5deee

## Enabling Range Selection With Linked Time
e7162159-9f4d-4ca3-ba7c-57d101c9664d

# Detailed steps to verify changes work correctly (as executed by you)
1) Start tensorboard with an experiment containing a scalar and histogram card
2) Navigate to http://localhost:6006?enableLinkedTime&enableRangeSelection&enableDataTable
3) Click the "Enable Range Selection" checkbox
image
4) Assert the the scalar card has two fobs
image
5) Assert the scalar card data table is in range mode
image
image
6) Disable "Enable Range Selection
7) Assert the second fob disappears from the scalar card
8) Enable "Link By Step"
9) Enable "Enable Range Selection"
10) Assert the fob reappears in the correct position.

PR #5962, opened 2022-10-04
Range Selection: Simplify range selection in redux state

rileyajones:
The redux state currently has two values used to represent whether range selection is enabled (`stepSelectorRangeEnabled` and `linkedTimeRangeEnabled`). Now that the UX around enabling range selection is being changed to simply be a checkbox in the settings panel this doesn't make a lot of sense.

This will enable #5956 will flush out the functionality for range selection without linked time enabled.

replace `stepSelectorRangeEnabled` and `linkedTimeRangeEnabled` with `rangeSelectionEnabled` in redux state.

* Screenshots of UI changes
N/A

1) Start tensorboard
2) Go to http://localhost:6006?allowRangeSelection&enableDataTable&enableLinkedTime
3) Enable "Link by Step" and add an end value
4) Assert that you now have two fobs (do not expect range selection to be fully functional in without "Link By Step" enabled the rest of the functionality is coming in #5956)
image
5) Assert that "Enable Range Selection" is now enabled
6) Disable "Enable Range Selection"
7) Assert that the end fob disappears
8) Re-enable "Enable Range Selection"
9) Assert the end fob reappears.

Do not replace the two keys and just modify the reducers and selectors to check both, this seems much messier.


tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -979,18 +978,22 @@ const reducer = createReducer(
...
+    let linkedTimeSelection = state.linkedTimeSelection
+      ? {...state.linkedTimeSelection}
+      : null;
+    if (nextRangeSelectionEnabled) {
+      nextStepSelectorEnabled = nextRangeSelectionEnabled;
     }
 
     return {
       ...state,
+      linkedTimeSelection,
```
JamesHollyer, line 994: I do not understand what you are trying to accomplish with by setting this here.

tensorboard/webapp/metrics/store/metrics_selectors.ts
```diff
@@ -371,14 +371,14 @@ export const getMetricsStepSelectorEnabled = createSelector(
 export const getMetricsStepSelectorRangeEnabled = createSelector(
```
JamesHollyer, line 371: I think we should combine getMetricsStepSelectorRangeEnabled and getMetricsLinkedTimeRangeEnabled into getMetricsRangeEnabled. Why do we want two selectors which return the same value?

tensorboard/webapp/metrics/store/metrics_reducers.ts
rileyajones: Looks like I missed pulling this when I split the step selector portion of out this pr.

tensorboard/webapp/metrics/store/metrics_selectors.ts
rileyajones: I definitely should have done that initially, done now.

JamesHollyer (approved):

PR #5963, opened 2022-10-04
Linked Time: Remove right panel widget

rileyajones:
~Blocked by #5965~

* Motivation for features / changes
We determined that the widget provided a bad user experience. The feature was mostly used to enable range selection which can now be enabled via the "Enable Range Selection" checkbox.

* Technical description of changes
Pure Deletion, just removing the widget.

* Screenshots of UI changes

* Detailed steps to verify changes work correctly (as executed by you)
1) Start tensorboard
2) Visit http://localhost:6006?enableLinkedTime&enableDataTable
3) Enable the "Link by Step" checkbox
4) Assert no additional widget is shown in the right panel.

PR #5965, opened 2022-10-06
Step Selector: Update Redux Logic around enabling Range Selection

rileyajones:
Currently "Enable Range Selection" DOES enable range selection, but it does not create another fob so it doesn't appear to be enabled. This PR handles the 1st part of the process, generating an end value in the redux state. #5963 will handle the majority of the ui implementation.
We are planning to remove the right panel widget currently used to enable range selection #5963.

When range selection is enabled if and there is no current `timeSelection` or `timeSelection` only has a single value generate a new time selection value by finding the min / max step in all the `timeSeriesData`

# Screenshots of UI changes
There is no ui effect to enabling range without linked time
## Disabling Step Selector Deselects "Enable Range Selection"
a59eb425-11f4-4d08-8ef6-0e2c32e75188

## Enabling range with linked time enabled
a1cc9fe4-2b70-4c38-9805-e62ffb121251

**TLDR**: See gifs above
1) Start tensorboard
2) Go to http://localhost:6006?enableLinkedTime=true&enableDataTable=true&allowRangeSelection=true#timeseries
3) Enable step selection and range selection
4) Assert only a single fob is shown on the scalar card and none on the histogram card
5) Enable linked time
6) Assert two fobs are immediately shown on the scalar and histogram cards
7) Disable range selection
8) Assert the second fob is removed from both cards.


tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -1076,6 +1092,43 @@ export function reducers(state: MetricsState | undefined, action: Action) {
   return composeReducers(reducer, namespaceContextedReducer)(state, action);
 }
 
+function getMinMaxTimeSelection(timeSeriesData: TimeSeriesData): TimeSelection {
```
japie1235813, line 1095: The name makes me think it's the min/max _of_ time selection. But it's "_of_ the experiment _for_ time selection)
Propose something like `getMinMaxStepForTimeSelection` or just `getStepMinMax` ?

Another more high level question, what is  the output different compared to the ngrx state? 

Updated here
[tensorboard/webapp/metrics/store/metrics_reducers.ts](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/store/metrics_reducers.ts;l=790?q=tensorboard%2Fwebapp%2Fmetrics%2Fstore%2Fmetrics_reducers.ts&ss=tensorflow%2Ftensorboard)

Note: I haven't check the implementation details yet but the output looks like what you want

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -951,6 +958,7 @@ const reducer = createReducer(
     let nextCardStepIndexMap = {...state.cardStepIndex};
     let nextLinkedTimeSelection = state.linkedTimeSelection;
     let nextStepSelectorEnabled = state.stepSelectorEnabled;
+    let linkedTimeRangeEnabled = state.rangeSelectionEnabled;
```
japie1235813, line 961: Can you give me a little bit more color on why we need to enable linked time when range is checked?

tensorboard/webapp/metrics/store/metrics_types.ts
```diff
@@ -92,17 +96,31 @@ export type TimeSeriesLoadables = {
...
-  [tag: string]: ScalarTimeSeriesLoadable;
-}
+export type TimeSeriesLoadableType =
+  | ScalarTimeSeriesLoadable
+  | HistogramTimeSeriesLoadable
+  | SampledImageTimeSeriesLoadable;
 
-export interface HistogramTimeSeriesData {
-  [tag: string]: HistogramTimeSeriesLoadable;
+export function isSampledImageTimeSeriesLoadable(
```
japie1235813, line 104: I don't think this is the appropriate place to put this function. (if we need it)

tensorboard/webapp/metrics/store/metrics_reducers.ts
rileyajones: This is actually a miss from when I split this apart from #5962. I'm surprised this didn't throw a type error :thinking:

tensorboard/webapp/metrics/store/metrics_reducers.ts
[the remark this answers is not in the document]
rileyajones: Sure, we don't generally add JSDoc to functions which aren't exported.

rileyajones: > The name makes me think it's the min/max of time selection. But it's "of the experiment for time selection)
Propose something like getMinMaxStepForTimeSelection or just getStepMinMax ?

`getStepMinMax` makes sense to me.
Looking at `getMinMaxTimeSelection` I would hope it was read as
"action" -> "get"
"noun" -> "Min Max TimeSelection"
i.e. get a time selection containing the min and max step.

> Another more high level question, what is the output different compared to the ngrx state?

This is a pure function, there is no relationship to the ngrx state.

tensorboard/webapp/metrics/store/metrics_types.ts
rileyajones: I don't love this being here either but it seemed to be the convention? i.e. [tensorboard/webapp/metrics/data_source/types.ts](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/data_source/types.ts;l=52)

We don't really have a great spot for utility functions in this part of the code. I could imagine moving it to either [tensorboard/webapp/metrics/store/metrics_reducers.ts](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/store/metrics_reducers.ts) or [tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/store/metrics_store_internal_utils.ts).
Do you have a preference?

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -981,15 +990,20 @@ const reducer = createReducer(
   on(actions.rangeSelectionToggled, (state) => {
     const nextRangeSelectionEnabled = !state.rangeSelectionEnabled;
     let nextStepSelectorEnabled = state.stepSelectorEnabled;
+    let linkedTimeSelection = state.linkedTimeSelection;
 
     if (nextRangeSelectionEnabled) {
       nextStepSelectorEnabled = nextRangeSelectionEnabled;
+      if (!linkedTimeSelection || !linkedTimeSelection.end) {
+        linkedTimeSelection = getMinMaxTimeSelection(state.timeSeriesData);
```
japie1235813: I mean, can I just reuse the value stepMinMax here with perhaps some infinity check? If not, why?
```
linkedTimeSelection = {
start: stepMinMax.min,
end: stepMinMax.max,
}
```

rileyajones: I didn't know that property was already computed. That would have saved me hours.

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -981,15 +981,28 @@ const reducer = createReducer(
...
     let nextStepSelectorEnabled = state.stepSelectorEnabled;
+    let linkedTimeSelection = state.linkedTimeSelection;
 
     if (nextRangeSelectionEnabled) {
       nextStepSelectorEnabled = nextRangeSelectionEnabled;
+      if (!linkedTimeSelection) {
+        linkedTimeSelection = {
+          start: {step: state.stepMinMax.min},
+          end: {step: state.stepMinMax.max},
+        };
```
japie1235813, line 992: For my own reference, this is because linked time selection will be set [here](https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/store/metrics_reducers.ts;l=959?q=tensorboard%2Fwebapp%2Fmetrics%2Fstore%2Fmetrics_reducers.ts&ss=tensorflow%2Ftensorboard) (to single selection) if we don't update it when range is enabled.

The clicking flow to not update here is: enable SS --> enable Range --> enable LT
(then we will find only one fob as the result)

Riley, feel free to add more and correct me.

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -1028,14 +1041,16 @@ const reducer = createReducer(
...
+    const nextLinkedTimeEnabled =
+      nextStepSelectorEnabled && state.linkedTimeEnabled;
+    const nextRangeSelectionEnabled =
+      nextStepSelectorEnabled && state.rangeSelectionEnabled;
 
     return {
       ...state,
       linkedTimeEnabled: nextLinkedTimeEnabled,
       stepSelectorEnabled: nextStepSelectorEnabled,
+      rangeSelectionEnabled: nextRangeSelectionEnabled,
```
japie1235813, line 1053: Test for this?

tensorboard/webapp/metrics/store/metrics_types.ts
japie1235813: Do we still need this function?

tensorboard/webapp/metrics/store/metrics_reducers.ts
```diff
@@ -981,15 +981,28 @@ const reducer = createReducer(
...
     const nextRangeSelectionEnabled = !state.rangeSelectionEnabled;
     let nextStepSelectorEnabled = state.stepSelectorEnabled;
+    let linkedTimeSelection = state.linkedTimeSelection;
 
     if (nextRangeSelectionEnabled) {
       nextStepSelectorEnabled = nextRangeSelectionEnabled;
+      if (!linkedTimeSelection) {
+        linkedTimeSelection = {
+          start: {step: state.stepMinMax.min},
+          end: {step: state.stepMinMax.max},
```
japie1235813, line 991: Do we want to check the infinity/-infinity? It does not feel right to set that. If we set the selection to be infinity/-infinity, where do we rely on the correct steps of fobs? Clipping logic? (Will any warning icon pop up then?)

japie1235813 (commented):
Generally lgtm. Still want to get some small things right before approval. Thanks for your patience and bearing with me.

tensorboard/webapp/metrics/store/metrics_reducers.ts
rileyajones: I would rather not check for Infinity/-Infinity here. There isn't really an obvious value for this to fall back to otherwise and I think incorrectly clipped fobs (yes the clipping logic will come up here) seems better than the feature not working.

japie1235813 (approved):

PR #5966, opened 2022-10-06
Feature: Add Dockerfile to build TensorBoard as a container

rileyajones:
Containerization can be helpful in many situations from deploying to the cloud, to running a task in CI, and even easily setting up a development environment.

## Noteworthy Callouts
Our OSS CI is often quite slow, this is likely due to an issue with network egress caps. Using a prebuilt image could dramatically reduce egress and potentially speed up CI runs.

Setting up a tensorboard development environment is somewhat difficult (getting the right versions of `bazel`, `python`, `node` etc are essential. With this image (and docker preinstalled) setup becomes two commands which are the same across platforms (OSX, Linux, maybe even Windows?).

I added a Dockerfile so that the project can easily be built as a Docker container.

# Detailed steps to verify changes work correctly (as executed by you)
## Building the image
Clone the project and run the following command from the root directory
`docker build -t tensorflow/tensorboard .`

## Example Usage
```
docker run \
    -p 6006:6006 \
    -v $PWD/tensorboard:/tensorboard/tensorboard \
    -v $MY_LOG_DIR:/logdir \
    tensorflow/tensorboard \
    ibazel run tensorboard:dev -- --logdir /logdir --bind_all
```


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+ENV BAZEL_VERSION='4.0.0'
+ENV BAZEL_SHA256SUM='7bee349a626281fc8b8d04a7a0b0358492712377400ab12533aeb39c2eb2b901'
+ENV BUILDTOOLS_VERSION='3.0.0'
+ENV BUILDIFIER_SHA256SUM='e92a6793c7134c5431c58fbc34700664f101e5c9b1c1fcd93b97978e8b7f88db'
+ENV BUILDOZER_SHA256SUM='3d58a0b6972e4535718cdd6c12778170ea7382de7c75bc3728f5719437ffb84d'
+ENV TENSORFLOW_VERSION='tf-nightly'
+
+# Get the code
+RUN git clone https://github.com/tensorflow/tensorboard /tensorboard
+WORKDIR /tensorboard
```
groszewn, line 17: Each [RUN](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers) directive creates a new image layer. Consider consolidating all of the build steps into a single `RUN` command.

At minimum, I'd suggest doing the more costly build steps that don't change often (i.e. download and installbazel,ibazel) before the things that change more frequently (pip dependencies, getting the most up-to-date code) to leverage the [build cache](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache) more effectively.

rileyajones: I'm not sure why consolidating the build steps would utilize the build cache more effectively, wouldn't more granular steps be better for caching?

groszewn: Apologies, those were separate suggestions, but that was definitely not clear.

After some offline discussion, it definitely doesn't make sense to have a monolithic `RUN` directive.

More granular steps would be good for caching as long as they're ordered properly (i.e. less frequently-updated steps come earlier).

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# Setup build environment
+ENV BAZEL_VERSION='4.0.0'
+ENV BAZEL_SHA256SUM='7bee349a626281fc8b8d04a7a0b0358492712377400ab12533aeb39c2eb2b901'
+ENV BUILDTOOLS_VERSION='3.0.0'
+ENV BUILDIFIER_SHA256SUM='e92a6793c7134c5431c58fbc34700664f101e5c9b1c1fcd93b97978e8b7f88db'
+ENV BUILDOZER_SHA256SUM='3d58a0b6972e4535718cdd6c12778170ea7382de7c75bc3728f5719437ffb84d'
+ENV TENSORFLOW_VERSION='tf-nightly'
+
+# Get the code
+COPY . /tensorboard
```
groszewn, line 16: Since you're using `COPY` instead of `git clone` here, you could split up retrieval/installation of your dependencies and pulling in the actual tensorboard application code. This should ensure you don't retrigger the dependency setup steps when you're modifying the rest of the code, which will speed up subsequent builds.

```
# Setup Bazel
COPY ./ci /tensorboard/ci
WORKDIR /tensorboard
RUN ci/download_bazel.sh "${BAZEL_VERSION}" "${BAZEL_SHA256SUM}" ~/bazel
RUN mv ~/bazel /usr/local/bin/bazel && chmod +x /usr/local/bin/bazel && cp ./ci/bazelrc ~/.bazelrc
RUN npm i -g @bazel/ibazel

# Install python dependencies
COPY ./tensorboard/pip_package /tensorboard/tensorboard/pip_package
RUN pip install -r ./tensorboard/pip_package/requirements.txt -r ./tensorboard/pip_package/requirements_dev.txt "$TENSORFLOW_VERSION" && pip freeze --all

# Get the code
COPY . /tensorboard
```

rileyajones: This is an interesting idea however installing dependencies then running `COPY . /tensorboard` will not necessarily work if the user has built the project locally and has build artifacts (i.e. `bazel-out`, `node_modules`, etc). I'm going to add a `.dockerignore` as well to try to get the best of both worlds.

groszewn (approved):

PR #5969, opened 2022-10-11
Bug Fix: Prevent toggling range from resetting time selection

rileyajones:
When toggling range selection the remaining fob was being set to the min value in the range.

* Screenshots of UI changes
Range Enabled:
image
Range Disabled:
image

See it in action
69353c11-7537-4b81-9bf7-441c6b34fee1

Zooming Still Works:
2ff41be9-598c-4bfb-8959-c4d7073b3870

1) Start tensorboard
2) Go to localhost:6006?enableRangeSelection
3) Enable range selection in the settings panel
4) Move both fobs on the scalar card
5) Disable range selection either via the checkbox or by removing a fob
6) Assert the remaining fob is not effected.
8) Renable range selection
7) Zoom in on the chart and assert both fobs are visible.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
@@ -604,39 +604,30 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
       this.minMaxSteps$,
       this.store.select(getMetricsStepSelectorEnabled),
       this.store.select(getMetricsRangeSelectionEnabled),
-    ]).subscribe(([minMaxStep, enableStepSelector, rangeSelectionEnabled]) => {
-      this.stepSelectorTimeSelection$.next(
-        this.computeStepSelectorTimeSelection(
-          minMaxStep,
-          enableStepSelector,
-          rangeSelectionEnabled
-        )
-      );
-    });
+    ]).subscribe(
+      ([{minStep, maxStep}, enableStepSelector, rangeSelectionEnabled]) => {
+        this.stepSelectorTimeSelection$.next(
+          enableStepSelector
+            ? {
+                start: {
+                  step: clipStepWithinMinMax(
+                    this.stepSelectorTimeSelection$.getValue()?.start.step ??
+                      minStep,
+                    minStep,
+                    maxStep
+                  ),
+                },
+                end: rangeSelectionEnabled ? {step: maxStep} : null,
+              }
+            : null
+        );
+      }
+    );
   }
 
   ngOnDestroy() {
     this.ngUnsubscribe.next();
     this.ngUnsubscribe.complete();
-  }
-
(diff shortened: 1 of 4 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
```diff
@@ -603,25 +604,41 @@ export class ScalarCardContainer implements CardRenderer, OnInit, OnDestroy {
...
-                start: {step: minStep},
-                end: stepSelectorRangeEnabled ? {step: maxStep} : null,
-              }
-            : null
-        );
-      }
-    );
+    ]).subscribe(([minMaxStep, enableStepSelector, rangeSelectionEnabled]) => {
+      this.stepSelectorTimeSelection$.next(
+        this.computeStepSelectorTimeSelection(
```
JamesHollyer, line 609: I do not think a utility function is needed here.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2154,6 +2154,112 @@ describe('scalar card', () => {
...
+                minStep: 0,
+                maxStep: 50,
+              },
+              false,
+              false
+            );
+          expect(timeSelection).toBeNull();
+        }));
+
+        it('time selection defaults to min/max', fakeAsync(() => {
```
JamesHollyer, line 2183: it('defaults to min/max'...

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -2154,6 +2154,112 @@ describe('scalar card', () => {
...
+              true,
+              true
+            );
+          expect(timeSelection).toEqual({
+            start: {step: 0},
+            end: {step: 50},
+          });
+        }));
+
+        it('uses existing min step when defined', fakeAsync(() => {
```
JamesHollyer, line 2200: it('uses existing start step when defined',...

JamesHollyer (commented):
It seems obvious to me that any logic inside that subscribe callback is computing the stepSelectorTimeSelection. I don't think it needs to be put into a function.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts
rileyajones: This makes it a lot easier to test

JamesHollyer: Not my favorite reason to add an abstraction. We are also now missing the logic that this is called when minMaxStep, enableStepSelector, or rangeSelectionEnabled are changed. And adding those tests are just as difficult as testing this without the function. I will accept it because I want this fix in and it is not that bad.

JamesHollyer (approved):

PR #5970, opened 2022-10-12
Handle the JSON parse error in HParams plugin

qihach64:
Currently, feeding badly formatted request data to `tb.corp/experiment/<id>/data/plugin/hparams/experiment` endpoint can result in 500s (see  [b/250791957](https://b.corp.google.com/issues/250791957)):

```
...
    raise ParseError('Failed to load JSON: {0}.'.format(str(e)))
google3.net.proto2.python.public.json_format.ParseError: Failed to load JSON: Expecting value: line 1 column 1 (char 0).
```
The 500s are confusing to the end users because they will see an `Internal Server Error` page without any useful debugging information. It also adds unnecessary noise to the logging/monitoring system which makes it hard to found the real alerts.

 This CL catches the JSON ParseError, logs the request data for debugging, and throws 400s.

## Screenshots of UI changes
   
image

Open the browser and enter the hparams plugin endpoint of an experiment that feeds the badly-formatted request data, e.g `/experiment/1811564056169732114/data/plugin/hparams/experiment?request=blah` and you should see the UI changes above rather than an `Internal Server Error` page.

### DIFF
--- a/tensorboard/plugins/hparams/hparams_plugin.py
+++ b/tensorboard/plugins/hparams/hparams_plugin.py
@@ -191,13 +191,18 @@ class HParamsPlugin(base_plugin.TBPlugin):
 
 
 def _parse_request_argument(request, proto_class):
-    if request.method == "POST":
-        return json_format.Parse(request.data, proto_class())
-
-    # args.get() returns the request URI-unescaped.
-    request_json = request.args.get("request")
-    if request_json is None:
+    request_json = (
+        request.data
+        if request.method == "POST"
+        else request.args.get("request")
+    )
+    try:
+        return json_format.Parse(request_json, proto_class())
+    # if request_json is None, json_format.Parse will throw an AttributeError:
+    # 'NoneType' object has no attribute 'decode'.
+    except (AttributeError, json_format.ParseError) as e:
         raise error.HParamsError(
-            "Expected a JSON-formatted 'request' arg of type: %s" % proto_class
-        )
-    return json_format.Parse(request_json, proto_class())
+            "Expected a JSON-formatted request data of type: {}, but got {} ".format(
+                proto_class, request_json
+            )
+        ) from e


tensorboard/plugins/hparams/hparams_plugin.py
```diff
@@ -191,13 +191,22 @@ def _get_scalars_plugin(self):
...
+        if request.method == "POST"
+        else request.args.get("request")
+    )
     if request_json is None:
         raise error.HParamsError(
-            "Expected a JSON-formatted 'request' arg of type: %s" % proto_class
+            "Expected a JSON-formatted request data of type: %s" % proto_class
+        )
+    try:
+        request_proto = json_format.Parse(request_json, proto_class())
```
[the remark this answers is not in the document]
qihach64: Done
Good catch.

tensorboard/plugins/hparams/hparams_plugin.py
```diff
@@ -191,13 +191,21 @@ def _get_scalars_plugin(self):
...
+    try:
+        return json_format.Parse(request_json, proto_class())
+    except json_format.ParseError as e:
+        err_desc = (
+            "Failed to parse the request json {} into a proto: {}".format(
+                request_json, str(e)
+            )
         )
-    return json_format.Parse(request_json, proto_class())
+        raise werkzeug.exceptions.BadRequest(description=err_desc)
```
arcra, line 211: I just realized that the code calling this method would catch an HParamsError and convert it to a BadRequest error, so let's be consistent and make this an HParamsError as well. Sorry for not catching this earlier.

nfelt, line 211: It's generally preferable in Python 3 to do a "raise from" in the format `raise <new exception> from e` when re-raising an exception, since the stacktrace will make more sense that way. See

https://stackoverflow.com/questions/24752395/python-raise-from-usage

nfelt (approved):

PR #5972, opened 2022-10-12
Avoid parsing Text Plugin metadata b"{}"

qihach64:
There is a library that emits b"{}" as the plugin's metadata, which triggers 500 errors in the Text Plugin server code ([b/250789176](https://b.corp.google.com/issues/250789176)):

```
google3.net.proto2.python.public.message.DecodeError: Error parsing message with type 'tensorboard.TextPluginData'
```

This commit handles the edge case by returning a TextPluginData default proto without parsing the metadata. 

## Screenshots of UI changes
- For the edge case that feeds in b"{}", users will see something like image rather than an "Internal server error" page. 

Open the browser and enter the TextPlugin metadata endpoint of an experiment that generates the invalid metadata, e.g. `experiment/4725997998739972179/data/plugin/text/tags` and you should see the resulting valid metadata.

### DIFF
--- a/tensorboard/plugins/text/metadata.py
+++ b/tensorboard/plugins/text/metadata.py
@@ -15,6 +15,8 @@
 """Internal information about the text plugin."""
 
 
+import werkzeug
+from google.protobuf import message
 from tensorboard.compat.proto import summary_pb2
 from tensorboard.plugins.text import plugin_data_pb2
 
@@ -53,6 +55,15 @@ def parse_plugin_metadata(content):
     """
     if not isinstance(content, bytes):
         raise TypeError("Content type must be bytes")
+    if content == b"{}":
+        # Old-style JSON format. Equivalent to an all-default proto.
+        return plugin_data_pb2.TextPluginData()
+    try:
+        result = plugin_data_pb2.TextPluginData.FromString(content)
+    except (message.DecodeError, TypeError) as e:
+        raise werkzeug.exceptions.InternalServerError(
+            "Failed to parse plugin metadata {}: {}".format(content, str(e))
+        )
     result = plugin_data_pb2.TextPluginData.FromString(content)
     if result.version == 0:
         return result
--- a/tensorboard/plugins/text/metadata.py
+++ b/tensorboard/plugins/text/metadata.py
@@ -15,8 +15,6 @@
 """Internal information about the text plugin."""
 
 
-import werkzeug
-from google.protobuf import message
 from tensorboard.compat.proto import summary_pb2
 from tensorboard.plugins.text import plugin_data_pb2
 
--- a/tensorboard/plugins/text/metadata.py
+++ b/tensorboard/plugins/text/metadata.py
@@ -58,12 +58,6 @@ def parse_plugin_metadata(content):
     if content == b"{}":
         # Old-style JSON format. Equivalent to an all-default proto.
         return plugin_data_pb2.TextPluginData()
-    try:
-        result = plugin_data_pb2.TextPluginData.FromString(content)
-    except (message.DecodeError, TypeError) as e:
-        raise werkzeug.exceptions.InternalServerError(
-            "Failed to parse plugin metadata {}: {}".format(content, str(e))
-        )
     result = plugin_data_pb2.TextPluginData.FromString(content)
     if result.version == 0:
         return result


tensorboard/plugins/text/metadata.py
```diff
@@ -53,6 +55,15 @@ def parse_plugin_metadata(content):
     """
     if not isinstance(content, bytes):
         raise TypeError("Content type must be bytes")
+    if content == b"{}":
+        # Old-style JSON format. Equivalent to an all-default proto.
+        return plugin_data_pb2.TextPluginData()
+    try:
+        result = plugin_data_pb2.TextPluginData.FromString(content)
+    except (message.DecodeError, TypeError) as e:
+        raise werkzeug.exceptions.BadRequest(
```
arcra, line 64: I don't know if bad request is the appropriate response here. Technically, all the user did is try to access an experiment, which is not precisely a "bad request". Maybe in this case we should just log something (that would help us debug), and let it fail the same way as before? Hopefully, the previous expected case would be enough to handle most of these cases.

qihach64: Good call! 

I picked 400 here because the library that generates the invalid metadata can be the one to blame. But since request data comes from dataprovider.list_tensors https://source.corp.google.com/piper///depot/google3/third_party/tensorboard/plugins/text_v2/text_v2_plugin.py;l=145;bpv=1;bpt=1;rcl=480228505?q=project:tensorboard%20%22parse_plugin_metadata%22 , which can be considered server-side code, I think 500 also makes sense when list_tensors has bugs that cause invalid metadata to be sent.

> Maybe in this case we should just log something (that would help us debug), and let it fail the same way as before?

Regardless of which status code we pick, I think we need to at least give users error details strings like this 
image rather than returning "Internal Server Error" without further info. Ideally, printing out error detail/stack traces should be done at the framework level but it's not in the scope of this PR. 

I can raise werkzeug.exceptions.InternalServerError with error details if that sounds good to you.

arcra: In general, we should be wary of externalizing any internal details (in error messages or otherwise), although in this case, being an OSS product, it might not matter too much.

I was expecting we would just log but keep the existing behavior and see if this is still an issue, but I guess this message could help users dig into what's wrong and potentially fix the code that writes these events, let's check quickly with Nick before merging, but it's fine by me.

nfelt: I'm not crazy about trying to return a "friendly" 500 error here - in general, we shouldn't ever really *intend* to return a 500 error, because a 500 should mean "the server encountered an unexpected failure condition" which typically isn't something a user would be expected to debug themselves. So returning a 500 deliberately for a particular case is kind of an antipattern to me.

IMO, the simplest resolution here is to fix the one known case (`b"{}"`) that is specific to the text plugin, and otherwise leave the behavior unchanged.

I agree it would be nice to avoid the 500 entirely when encountering invalid plugin metadata for the sake of robustness (although I don't think it's a particularly common issue).  If we do that, all plugins have basically the same codepath and ideally we'd fix them all, to avoid having the codepaths diverge. We can avoid the 500 by logging a warning and then just skipping this tag - we have similar logic already that does this if the metadata has a version field that's too new: https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/plugins/text/text_plugin.py;l=245-247;drc=519c6bb340294a45e17fad0353bdb5620b61abed  So we'd need to change parse_plugin_metadata() to do something like return None if it can't be parsed, and then also skip the tag in that case. (We might be able to update MetadataVersionChecker to handle this, to avoid needing the same boilerplate in every plugin.)

In terms of making 500s easier to debug in general, our current logic for turning exceptions into HTTP error responses only shows the exception error message if they are subclasses of our PublicError class: https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/backend/application.py;l=565-582;drc=e23de9dfca79048b570389c5b025ad8788197d7a This is for the reasons Adrian mentioned - we have to be a bit cautious not to disclose details in errors that weren't intended to be public (more of a concern for hosted services, but they also use this logic right now).  I also agree it'd be useful if we could lift that restriction for ourselves, but it'd be most effective to do that by adjusting this exception handler so it applies to all exceptions, not just specific known cases.  For example, we could consider customizing the handler so that we show more detail if the user is a TB developer, or if the binary is run with some kind of "dev mode" startup flag or environment variable (but we'd have to figure out what signals to use for those cases).

qihach64: Thanks for the discussion! The reasoning makes lots of sense and I picked the easiest resolution for now.

arcra: LGTM.

nfelt: LGTM as-is.  I'd suggest we try this first, and if we're still seeing errors we can add logging?  I expect that this will account for almost all of the occurrences in practice, and if we want to handle the general case better, then as mentioned above it'd feel more consistent to do it for all plugins, not just text.

nfelt (approved):
LGTM but can you update the PR title so it's more accurate to what it's doing now?

PR #5975, opened 2022-10-13
Step Selector: Persist step selection settings.

JamesHollyer:
The new step selection feature currently must be enabled every time you load Tensorboard. This is annoying for users who like this feature and have to turn it on constantly. This change persists the settings in LocalStorage so it can stay enabled on that machine until it is disabled.

Just followed the pattern of settings that are already persisted.

### DIFF
--- a/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
+++ b/tensorboard/webapp/metrics/store/metrics_reducers_test.ts
@@ -2414,6 +2414,29 @@ describe('metrics reducers', () => {
     });
   });
 
+  it('loads Step Selector Setting into the next state', () => {
+    const beforeState = buildMetricsState({
+      stepSelectorEnabled: false,
+      rangeSelectionEnabled: false,
+      linkedTimeEnabled: false,
+    });
+
+    const nextState = reducers(
+      beforeState,
+      globalSettingsLoaded({
+        partialSettings: {
+          stepSelectorEnabled: true,
+          rangeSelectionEnabled: true,
+          linkedTimeEnabled: true,
+        },
+      })
+    );
+
+    expect(nextState.stepSelectorEnabled).toBe(true);
+    expect(nextState.rangeSelectionEnabled).toBe(true);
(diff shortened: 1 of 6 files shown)


tensorboard/webapp/metrics/store/metrics_reducers_test.ts
```diff
@@ -2414,6 +2414,29 @@ describe('metrics reducers', () => {
     });
   });
 
+  it('loads Step Selector Setting into the next state', () => {
```
bmd3k, line 2417: I'd prefer three different tests so we can test each property in isolation. Partially to maintain consistency with the rest of the file. But also to make sure, for example, that stepSelectorEnabled is not enabled by the linkedTimeEnabled setting in local storage.

tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source.ts
```diff
@@ -84,6 +84,16 @@ export class OSSSettingsConverter extends SettingsConverter<
       serializableSettings.timeSeriesCardMinWidth =
         settings.timeSeriesCardMinWidth;
     }
+    if (settings.stepSelectorEnabled !== undefined) {
```
bmd3k, line 87: Is it possible to add tests for the new logic in this file?

tensorboard/webapp/metrics/store/metrics_reducers_test.ts
JamesHollyer: Separated.

tensorboard/webapp/persistent_settings/_data_source/persistent_settings_data_source.ts
JamesHollyer: The other ones are not really tested. But, I added test that hit all the new logic paths.

JamesHollyer (commented):
Sorry I had to force push. I only changed the test files.

bmd3k (approved):

Issue #5968, opened 2022-10-11
Removing a Fob triggers a Time Selection Changed Event


rileyajones:
## Issue description
This happens because there is click event is bound to the [card_fob_component](https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/widgets/card_fob/card_fob_component.ts)
AND a mousedown event bound to the [card_fob_controller_component](https://github.com/tensorflow/tensorboard/blob/master/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ng.html)

Before
![image](https://user-images.githubusercontent.com/78179109/195204357-2eec7c68-aaa7-4a28-b05b-2ede27f091fd.png)
After
![image](https://user-images.githubusercontent.com/78179109/195204454-e1032ff4-4966-48f7-ad6b-56a89c66b046.png)

Action
![07e8d019-69aa-434e-9e42-f05b489222ea](https://user-images.githubusercontent.com/78179109/195204413-6d6d27ba-1187-4224-b080-bc862259eb6f.gif)

[cast.webm](https://user-images.githubusercontent.com/78179109/195204211-b3223504-ea00-4dd7-81ec-aa37b94ebe23.webm)

PR #5976, opened 2022-10-13
Bug Fix: Prevent Card Fob from dispatching duplicate events when a fob is removed

rileyajones:
#5968

The card fob would dispatch a `timeSelectionChanged` event on mouse up even if the time selection had not changed...

1) Start tensorboard
2) Go to http://localhost:6006
3) Enable step selection
4) Open the console and clear it
5) Dismiss the fob by clicking this button
image

Add a `mousedown` event listener to the dismiss button which prevents propagation. Unfortunately this would prevent clicks starting on the dismiss button from dragging the fob.

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -63,6 +63,7 @@ export class CardFobControllerComponent {
     new EventEmitter<TimeSelectionWithAffordance>();
   @Output() onTimeSelectionToggled = new EventEmitter();
 
+  private dragTimeSelectionStart: TimeSelection | undefined;
   private currentDraggingFob: Fob = Fob.NONE;
   private affordance: TimeSelectionAffordance = TimeSelectionAffordance.NONE;
 
@@ -110,6 +111,7 @@ export class CardFobControllerComponent {
     document.addEventListener('mousemove', this.mouseListener);
     document.addEventListener('mouseup', this.stopListener);
     this.currentDraggingFob = fob;
+    this.dragTimeSelectionStart = {...this.timeSelection};
     this.affordance = affordance;
   }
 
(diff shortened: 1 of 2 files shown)


JamesHollyer (commented):
Weird edge case. If someone drags the fob around then goes back to original spot we will not register the change. I guess this is ok. Maybe a boolean like "didMove" which flips on movement and resets when stopDrag is called?

rileyajones:
> Weird edge case. If someone drags the fob around then goes back to original spot we will not register the change. I guess this is ok. Maybe a boolean like "didMove" which flips on movement and resets when stopDrag is called?

Do we actually want to fire a `timeSelectionChanged` event in that case though? In this case the time selection has not changed.

JamesHollyer:
> > Weird edge case. If someone drags the fob around then goes back to original spot we will not register the change. I guess this is ok. Maybe a boolean like "didMove" which flips on movement and resets when stopDrag is called?
> 
> Do we actually want to fire a `timeSelectionChanged` event in that case though? In this case the time selection has not changed.

Not a huge deal but, yea I think we would want to log the metric that they dragged the fob around.  If the user drags the fob around...checks out some values and then moves it back to starting position I would want to track that in our metrics as a change.

rileyajones:
> > > Weird edge case. If someone drags the fob around then goes back to original spot we will not register the change. I guess this is ok. Maybe a boolean like "didMove" which flips on movement and resets when stopDrag is called?
> > 
> > 
> > Do we actually want to fire a `timeSelectionChanged` event in that case though? In this case the time selection has not changed.
> 
> Not a huge deal but, yea I think we would want to log the metric that they dragged the fob around. If the user drags the fob around...checks out some values and then moves it back to starting position I would want to track that in our metrics as a change.

That makes sense, it should work that way now

JamesHollyer (approved):

PR #5977, opened 2022-10-14
Upgrade rules_nodejs to 5.7.0, rules_sass to 1.55, minimum bazel to 4.2.2.

bmd3k:
Upgrades rule_nodejs dependency to 5.7.0.

This also requires upgrading:
* npm @bazel dependencies to 5.7.0
* rules_sass to its latest version 1.55.0.
* minimum bazel version to 4.2.2

Following two sets of instructions:
* General instructions for migration to 5.X major release: https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0 
* Instructions for 5.7.0 specifically: https://github.com/bazelbuild/rules_nodejs/releases/tag/5.7.0

Other notable things:
* yarn_install's exports_directories_only property now defaults to True. We must set this to False in order to remain compatible with ts_library
  * See:  https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0#exports_directories_only
* No longer need to override node_version as rules_nodejs default version of node is now 16.12.0.
  * See: https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0#updated-defaults
  * We were overriding this only because rules_nodejs defaul version of node was too old.
* ts_library has moved from @bazel/typescript to @bazel/concatjs
  * See: https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0#ts_library
* rules_sass upgrade was necessary to avoid SassCompiler hanging indefinitely. Builds were not completing.
* bazel version was not necessary but does match the minimum version specified by rules_nodejs:
  * See: https://github.com/bazelbuild/rules_nodejs/blob/5.7.0/index.bzl#L96


JamesHollyer (approved):
Looks good! I am happy to see that swapping the ts_library over to coming from the new concatjs didn't break anything!

PR #5984, opened 2022-10-18
chore: Upgrade github actions to use ubuntu-20.04.

bmd3k:
ubuntu-18.04 is deprecated and experiences periodic brownouts.
See: https://github.com/actions/runner-images/issues/6002

I initially attempted to upgrade to ubuntu-22.04 but could not solve problems running karma and ChromeHeadless driver.
```
[1018/165541.955009:WARNING:gpu_process_host.cc(1228)] The GPU process has crashed 6 time(s)
[1018/165541.955035:FATAL:gpu_data_manager_impl_private.cc(439)] GPU process isn't usable. Goodbye.
```
This problem does not appear to exist on ubuntu-20.04.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,6 +245,7 @@ jobs:
           nbfmt_version="174c9a5c1cc51a3af1de98d84824c811ecd49029"
           pip install black yamllint -c ./tensorboard/pip_package/requirements_dev.txt
           pip install -U git+https://github.com/tensorflow/docs@${nbfmt_version}
+          pip install -U "protobuf<3.20"
       - run: pip freeze --all
       - name: 'Lint Python code for style with Black'
         # You can run `black .` to fix all Black complaints.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ env:
```
rileyajones, line 31: Rather than repeatedly change the platform in 12 places how do you feel about using an environment variable?

```suggestion
  PLATFORM: ubuntu-22.-4
```

rileyajones (commented):
I know this is still a draft but still wanted to leave a suggestion

.github/workflows/ci.yml
bmd3k: Oddly this does not seem possible? The documentation on contexts makes it seems that maybe this could be possible:

https://docs.github.com/en/actions/learn-github-actions/contexts

But I couldn't get it to work (see #5986 for one example attempt). Searching Google's internal code base using "f:yml runs-on.*\$", it seems like the only way others have gotten this to work is to use the `matrix` context but we already use `matrix` where appropriate.

rileyajones: That's a shame, thanks for looking into it. There is some merit to declaring the platform for each step anyway so I don't mind the current approach too much.

rileyajones (approved):

PR #5987, opened 2022-10-19
Data Table: Sort Run Names Without Id and Alias

rileyajones:
For Googlers [b/254089050](b/254089050)
Run names are currently sorted differently in the data table than in the tool tip. I'm changing the sorting to be consistent with the tooltips alphabetical sort.

* Screenshots of UI changes
51842c87-6bf0-4492-8c93-34535f845943

1) Start tensorboard
2) Visit localhost:6006?enableDataTable&enableRangeSelection
3) Click the "Enable Range Selection" checkbox in the settings panel
4) Sort the scalar card data table
5) Assert that the table is sorted by the run name only (not the alias or alias id)

I considered creating a column alias map but this approach seemed more flexible.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
@@ -104,11 +104,7 @@ export class ScalarCardDataTable {
           closestEndPointIndex = findClosestIndex(datum.points, endStep);
           closestEndPoint = datum.points[closestEndPointIndex];
         }
-        const alias = metadata.alias
-          ? `${metadata.alias.aliasNumber} ${metadata.alias.aliasText}/`
-          : '';
         const selectedStepData: SelectedStepRunData = {};
-        selectedStepData.ALIAS = alias;
         selectedStepData.COLOR = metadata.color;
         selectedStepData.DISPLAY_NAME = metadata.displayName;
         for (const header of this.dataHeaders) {
@@ -224,6 +220,8 @@ function makeValueSortable(value: number | string | null | undefined) {
 
 function getSortableValue(point: SelectedStepRunData, header: ColumnHeaders) {
   switch (header) {
+    // The value shown in the "RUN" column is a string concatenation of Alias Id + Alias + Run Name
+    // but we would actually prefer to sort by just the run name.
     case ColumnHeaders.RUN:
       return makeValueSortable(point[ColumnHeaders.DISPLAY_NAME]);
     default:
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
@@ -189,8 +189,8 @@ export class ScalarCardDataTable {
       });
     dataTableData.sort(
       (point1: SelectedStepRunData, point2: SelectedStepRunData) => {
-        const p1 = getSortableValue(point1[this.sortingInfo.header]);
-        const p2 = getSortableValue(point2[this.sortingInfo.header]);
+        const p1 = getSortableValue(point1, this.sortingInfo.header);
+        const p2 = getSortableValue(point2, this.sortingInfo.header);
         if (p1 < p2) {
           return this.sortingInfo.order === SortingOrder.ASCENDING ? -1 : 1;
         }
@@ -205,7 +205,7 @@ export class ScalarCardDataTable {
   }
 }
 
-function getSortableValue(value: number | string | undefined | null) {
+function makeValueSortable(value: number | string | null | undefined) {
   if (
     Number.isNaN(value) ||
     value === 'NaN' ||
@@ -215,4 +215,13 @@ function getSortableValue(value: number | string | undefined | null) {
     return -Infinity;
   }
   return value;
+}
+
+function getSortableValue(point: SelectedStepRunData, header: ColumnHeaders) {
+  switch (header) {
+    case ColumnHeaders.RUN:
+      return makeValueSortable(point[ColumnHeaders.DISPLAY_NAME]);
+    default:
+      return makeValueSortable(point[header]);
+  }
 }
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
@@ -104,9 +104,10 @@ export class ScalarCardDataTable {
           closestEndPointIndex = findClosestIndex(datum.points, endStep);
           closestEndPoint = datum.points[closestEndPointIndex];
         }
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
```diff
@@ -216,3 +216,12 @@ function getSortableValue(value: number | string | undefined | null) {
   }
   return value;
 }
+
+function getSortableValue(point: SelectedStepRunData, header: ColumnHeaders) {
+  switch (header) {
+    case ColumnHeaders.RUN:
+      return makeValueSortable(point[ColumnHeaders.DISPLAY_NAME]);
```
bmd3k, line 223: I don't think point[ColumnHeaders.DISPLAY_NAME] is populated at this point. Verified with a plain `tensorboard` that I can't sort by RUN.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_types.ts
```diff
@@ -77,7 +77,9 @@ export interface PartitionedSeries {
  * value and the DataTable widget must know how to display each value.
  */
 export enum ColumnHeaders {
+  ALIAS = 'ALIAS',
   COLOR = 'COLOR',
+  DISPLAY_NAME = 'DISPLAY_NAME',
```
bmd3k, line 82: The code does not seem to be working so it's hard to know what you had in mind. . . but I'm confused by the choice to represent ALIAS and DISPLAY_NAME as Columns since you are not displaying them as such.

Should the display name data be tracked in some way that does not intersect/confuse with the concept of "ColumnHeaders" that are being displayed to the user?

rileyajones: I tested this on TB.corp so that I could ensure it met your usecase then manually copied the code out to github. "Surprisingly" manually copying lines of code is error prone... apologies for not catching that earlier.

`ColumnHeaders` is simply all the POTENTIAL headers, not the headers being displayed.
These make sense to me as potentially useful in the future though maybe `DISPLAY_NAME` isn't the best name?

If you'd still prefer I avoid using the headers I could move to storing a `RUN -> SortableName` map on the `ScalarCardDataTable`.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
```diff
@@ -216,3 +221,12 @@ function getSortableValue(value: number | string | undefined | null) {
   }
   return value;
 }
+
+function getSortableValue(point: SelectedStepRunData, header: ColumnHeaders) {
+  switch (header) {
+    case ColumnHeaders.RUN:
```
bmd3k, line 227: Can you leave a comment explaining the reasoning?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
```diff
@@ -216,3 +221,12 @@ function getSortableValue(value: number | string | undefined | null) {
...
+      return makeValueSortable(point[ColumnHeaders.DISPLAY_NAME]);
```
bmd3k, line 228: Can you add some tests?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_types.ts
bmd3k: I suspect the real impetus here is the SelectedStepRunData type. I think ideally we would just add a `displayName` type there (and use `displayName` much like you are currently using `DISPLAY_NAME` for the calculation) but you can't do that since it's a mapped type.

Maybe the best compromise here is to add DISPLAY_NAME but don't add ALIAS. The ALIAS stuff seems likely to be dead code forever.

tensorboard/webapp/widgets/data_table/data_table_component.ts
```diff
@@ -90,6 +94,10 @@ export class DataTableComponent {
     selectedStepRunData: SelectedStepRunData
   ): string {
     switch (columnHeader) {
+      case ColumnHeaders.ALIAS:
```
bmd3k, line 97: Can you add tests?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
bmd3k: Looks fixed now, thanks!

tensorboard/webapp/metrics/views/card_renderer/scalar_card_types.ts
rileyajones: Not adding `ALIAS` makes sense to me.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
rileyajones: I've gone ahead and added tests to the newly refactored version.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_data_table.ts
rileyajones: I've added a comment.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_types.ts
rileyajones: I went ahead with a pretty significant refactor to get away from using headers at all.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_types.ts
```diff
@@ -111,7 +111,7 @@ export interface SortingInfo {
  */
 export type SelectedStepRunData = {
   [key in ColumnHeaders]?: string | number;
-};
+} & {id: string};
```
bmd3k, line 114: Nice.

bmd3k (approved):

tensorboard/webapp/metrics/views/card_renderer/scalar_card_test.ts
```diff
@@ -3093,6 +3097,95 @@ describe('scalar card', () => {
...
+      console.log(
+        'ChartMetadatMap: ',
+        scalarCardDataTable.componentInstance.chartMetadataMap
+      );
+      fixture.detectChanges();
+
+      const data =
+        scalarCardDataTable.componentInstance.getTimeSelectionTableData();
+
+      expect(data[0].RUN).toEqual('7 g/run1');
```
bmd3k, line 3181: One small suggestion: I was momentarily confused by this test as it appeared to be ordering by alias id descending. Maybe mix up the alias ids a bit so its clearly not ordering by alias id at all.

For example: swap the alias id and number for runs 2 and 3.

bmd3k (approved):

PR #5989, opened 2022-10-20
Patch @bazel/concatjs to run karma tests outside sandbox and fix test runtime errors.

bmd3k:
* Motivation

  Karma tests are failing in chrome in certain environments (including, most recently, some internal Google machines) with errors like:
  
  ```
  [1019/144854.100947:WARNING:gpu_process_host.cc(1228)] The GPU process has crashed 1 time(s)
  [1019/144854.140417:WARNING:gpu_process_host.cc(1228)] The GPU process has crashed 2 time(s)
  [1019/144854.174092:WARNING:gpu_process_host.cc(1228)] The GPU process has crashed 3 time(s)
  [1019/144854.241059:WARNING:gpu_process_host.cc(1228)] The GPU process has crashed 4 time(s)
  [1019/144854.253018:WARNING:gpu_process_host.cc(1228)] The GPU process has crashed 5 time(s)
  [1019/144854.265413:WARNING:gpu_process_host.cc(1228)] The GPU process has crashed 6 time(s)
  [1019/144854.265453:FATAL:gpu_data_manager_impl_private.cc(439)] GPU process isn't usable. Goodbye.
  ```

* Workaround:

  We haven't been able to root cause the issue (is it specific to more recent Chrome versions? Is it OS-specific?) but we do know that we can workaround the issue by launching the karma chrome browser with the '--no-sandbox' option.
  
  This option is widely used by other karma clients. Notably we find it in:
  * Angular's guide on testing with Karma: https://angular.io/guide/testing
  * Google's internal browser launcher definition for chrome-linux (Googlers see https://source.corp.google.com/piper///depot/google3/testing/web/browsers/chrome-linux.json)
  
  So it seems appropriate for us to use this option as well.

* Implementation:

  The most straightforward way to force the '--no-sandbox' option is not very straightforward at all: Patch @bazel/concatjs to use the option. We do this using the [patch-package](https://www.npmjs.com/package/patch-package) library. We update @bazel/concatjs' 'karma.conf.js' file to use '--no-sandbox' even when it is supposed to be launching a Chrome browser with sandbox:

  * https://github.com/bazelbuild/rules_nodejs/blob/5.7.0/packages/concatjs/web_test/karma.conf.js#L387

* Alternatives:

  I examined the Chrome codebase and played around with other options (especially gpu-specific options like '--disable-gpu') but did not find any other Chrome flag that seemed to solve the issue.

  I explored specifying our own karma.conf.js file to define our own browsers or custom launchers but @bazel/concatjs does not allow us to specify these ourselves. It resets these values here:
  * https://github.com/bazelbuild/rules_nodejs/blob/5.7.0/packages/concatjs/web_test/karma.conf.js#L342

  And forces us to use their own custom launchers.

### DIFF
--- a/patches/BUILD
+++ b/patches/BUILD
@@ -1,0 +1,1 @@
+licenses(["notice"])
--- a/patches/BUILD
+++ b/patches/BUILD
@@ -1,1 +1,0 @@
-licenses(["notice"])
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -79,13 +79,13 @@ load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
 
 yarn_install(
     name = "npm",
+    data = ["//patches:@bazel+concatjs+5.7.0.patch"],
     # "Some rules only work by referencing labels nested inside npm packages
     # and therefore require turning off exports_directories_only."
     # This includes "ts_library".
     # See: https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0#exports_directories_only
     exports_directories_only = False,
     package_json = "//:package.json",
-    data = ["//patches:@bazel+concatjs+5.7.0.patch"],
     yarn_lock = "//:yarn.lock",
 )
 
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -85,6 +85,7 @@ yarn_install(
     # See: https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0#exports_directories_only
     exports_directories_only = False,
     package_json = "//:package.json",
+    data = ["//patches:@bazel+concatjs+5.7.0.patch"],
     yarn_lock = "//:yarn.lock",
 )
 
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "doc": "docs"
   },
   "scripts": {
-    "postinstall": "ngcc; patch-package",
+    "postinstall": "ngcc && patch-package",
     "build": "bazel build //...",
     "test": "ibazel test //...",
     "lint": "prettier --check 'tensorboard/**/*.'{css,html,js,ts,scss} .github/**/*.yml",
@@ -122,6 +122,7 @@
     "ngx-color-picker": "^11.0.0",
     "numeric": "^1.2.6",
     "plottable": "^3.9.0",
+    "postinstall-postinstall": "^2.1.0",
     "rxjs": "^7.3.0",
     "three": "~0.137.0",
     "umap-js": "^1.3.2",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "doc": "docs"
   },
   "scripts": {
-    "postinstall": "ngcc",
+    "postinstall": "ngcc; patch-package",
     "build": "bazel build //...",
     "test": "ibazel test //...",
     "lint": "prettier --check 'tensorboard/**/*.'{css,html,js,ts,scss} .github/**/*.yml",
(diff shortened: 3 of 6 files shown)


patches/BUILD
```diff
@@ -0,0 +1 @@
+licenses(["notice"])
```
JamesHollyer, line 1: No idea what this does.

rileyajones (approved):
This generally seems reasonable especially as a short term fix.
As a long term solution have you considered making a PR on the [Concatjs repo](https://github.com/bazelbuild/rules_nodejs)?
[Here is their karma config](https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/concatjs/web_test/karma.conf.js#L382)

JamesHollyer (commented):
Why are there so many yarn.lock changes?

JamesHollyer:
> 

yea maybe this function which determines if the --no-sandbox flag is enabled or not is out of date. https://github.com/bazelbuild/rules_nodejs/blob/cd6cecfd6e923bbf782884b68e9842a7bac4d08e/packages/concatjs/web_test/karma.conf.js#L69

JamesHollyer:
> This generally seems reasonable especially as a short term fix. As a long term solution have you considered making a PR on the [Concatjs repo](https://github.com/bazelbuild/rules_nodejs)? [Here is their karma config](https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/concatjs/web_test/karma.conf.js#L382)

Yea maybe [this function](https://github.com/bazelbuild/rules_nodejs/blob/cd6cecfd6e923bbf782884b68e9842a7bac4d08e/packages/concatjs/web_test/karma.conf.js#L69) which determines if the --no-sandbox flag is enabled or not is out of date.

patches/BUILD
bmd3k: Same, no idea what it does. Removed. (But I had to keep the BUILD file)

bmd3k:
> This generally seems reasonable especially as a short term fix. As a long term solution have you considered making a PR on the [Concatjs repo](https://github.com/bazelbuild/rules_nodejs)? [Here is their karma config](https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/concatjs/web_test/karma.conf.js#L382)

I had considered that but, unfortunately, without knowing the root cause there isn't really a clean change to propose. I agree with James that `supportChromeSandboxing()` would be the ideal place to make a change but I don't have a clear idea what that change would be.

bmd3k:
> Why are there so many yarn.lock changes?

Because I added two new yarn dependencies: patch-package and postinstall-postinstall. I used the guide at https://www.npmjs.com/package/patch-package.

bmd3k:
> > This generally seems reasonable especially as a short term fix. As a long term solution have you considered making a PR on the [Concatjs repo](https://github.com/bazelbuild/rules_nodejs)? [Here is their karma config](https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/concatjs/web_test/karma.conf.js#L382)
> 
> I had considered that but, unfortunately, without knowing the root cause there isn't really a clean change to propose. I agree with James that `supportChromeSandboxing()` would be the ideal place to make a change but I don't have a clear idea what that change would be.

It's probably worth noting a couple more things:
* The patch being applied here is to that exact file in the concatjs repo (karma.conf.js).
* patch-package is recommended by rules_nodejs as a means of making these sorts of custom changes if a PR to the main repo isn't possible (https://bazelbuild.github.io/rules_nodejs/changing-rules.html#patching-the-npm-packages)

JamesHollyer (approved):

PR #5991, opened 2022-10-21
Fix most tests that break with es2020 output.

bmd3k:
The migration to spec_bundle() will require we configure the TypeScript compiler to output to JavaScript 'es2020'. For the most part, this will require very little changes to our test code. There is one glaring exception, though, that we mostly address in this PR.

The following pattern will break:

in my_utilities.ts:
```
export function myFuncToSpyOn() {
  // Some code.
}
```

in my_test.ts:
```
import * as my_utilities from './my_utilities'

spyOn(my_utilities, 'myFuncToSpyOn').and.returnValue(1);
```

The Jasmine library will complain: 
`Error: <spyOn> : myFuncToSpyOn is not declared writable or has no setter`

Jasmine is aware of the issue but can't provide a fix:
* https://github.com/jasmine/jasmine/issues/1942
* https://jasmine.github.io/pages/faq.html#module-spy

Instead they suggest we refactor our code so that the functions being spied on are wrapped in some Class or other Object.

This PR does exactly that, for the most part doing the most simple of wrapping:

now in my_utilities.ts:
```
// No longer exported.
function myFuncToSpyOn() {
  // Some code.
}

export const MyUtilities = {
  myFuncToSpyOn
}
```

now in my_test.ts:
```
import {MyUtilities} from './my_utilities'

spyOn(MyUtilities, 'myFuncToSpyOn').and.returnValue(1);
```

The change has a bit of bigger blast radius because not only do the tests need to be changed but so does every other file (test or not) that are using the utility functions.

### DIFF
--- a/tensorboard/webapp/widgets/line_chart_v2/lib/utils.ts
+++ b/tensorboard/webapp/widgets/line_chart_v2/lib/utils.ts
@@ -15,7 +15,7 @@ limitations under the License.
 
 import {Extent, Polyline, Rect} from './internal_types';
 
-export function convertRectToExtent(rect: Rect): Extent {
+function convertRectToExtent(rect: Rect): Extent {
   return {
     x: [rect.x, rect.x + rect.width],
     y: [rect.y, rect.y + rect.height],
@@ -39,15 +39,15 @@ let cachedIsWebGl2Supported = false;
   }
 }
 
-export function isWebGl2Supported(): boolean {
+function isWebGl2Supported(): boolean {
   return cachedIsWebGl2Supported;
 }
 
-export function isOffscreenCanvasSupported(): boolean {
+function isOffscreenCanvasSupported(): boolean {
   return self.hasOwnProperty('OffscreenCanvas');
 }
 
-export function arePolylinesEqual(lineA: Polyline, lineB: Polyline) {
+function arePolylinesEqual(lineA: Polyline, lineB: Polyline) {
   if (lineA.length !== lineB.length) {
     return false;
   }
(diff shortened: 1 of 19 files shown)


tensorboard/webapp/widgets/line_chart_v2/lib/utils.ts
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 
 import {Extent, Polyline, Rect} from './internal_types';
 
-export function convertRectToExtent(rect: Rect): Extent {
+function convertRectToExtent(rect: Rect): Extent {
```
JamesHollyer, line 18: Can we continue to export these so that we can import them directly in places other than the tests?

bmd3k: If code uses the functions directly then they won't be using the mocks. 

For example:

If the code you're writing does this:
```
function myFunction() {
  return convertRectToExtent(this.someRect);
}
```

But your test does this:
```
spyOn(ChartUtils, 'convertRectToExtent').and.returnValue(someExtent);
expect(myFunction()).toEqual(someExtent);
```

It won't work. Because you're replacing what ChartUtils.convertRectToExtent points to with a mock/spy but raw convertRectToExtent remains unmodified.

JamesHollyer: Ah yea that makes sense.  Dang ok I guess we have to do it this way.

JamesHollyer (approved):

tensorboard/webapp/widgets/source_code/source_code_container_test.ts
```diff
@@ -17,7 +17,7 @@ limitations under the License.
  */
 import {Component, Input, NO_ERRORS_SCHEMA} from '@angular/core';
 import {ComponentFixture, TestBed} from '@angular/core/testing';
-import * as loadMonacoShim from './load_monaco_shim';
+import {MonacoShim} from './load_monaco_shim';
```
japie1235813, line 20: This breaks the sync. Would you mind working on the fix? 
See cl/487331149. Thanks

PR #5992, opened 2022-10-21
chore: Upgrade github actions to use ubuntu-22.04.

bmd3k:
Some other changes to be compatible with the 22.04 image:
* Bump python version for lint-python-yaml-docs. 3.6 no longer supported.
* Force downgrade of protobuf in lint-python-yaml-docs.

Note: We recently updated github actions to use ubuntu-20.04 in https://github.com/tensorflow/tensorboard/pull/5984 but subsequently unlocked ubuntu-22.04 with https://github.com/tensorflow/tensorboard/pull/5989.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


rileyajones (approved):
Thanks for handling this chore.
Originally I thought the python upgrade cause some issues with the Dockerfile but upon closer examination I think it's actually fine.

PR #5993, opened 2022-10-24
linked time: update scalar to "real" min max step

japie1235813:
The "minMax" input of scalar fob controller is actually the viewExtend/viewBox, not the min max step, while it's got used as the min max step. Visually the extend is larger than the min/max step and it is what we need for forward and reverse function.
```
            A   B   (chart)   C   D
            |---|-------------|---|
```
axis size: A to D
minMax:  A to D
line chart area: B to C
min step: B
max step C

This wasn't noticed because we clip the time selection at the scalar card level to prevent the fob go out of the valid range (B to C)

This PR adds another minMaxStep input and rename the old minMax to minMaxHorizentalViewExtend. 
(Note: skip the rename in test since that matters less)

* Screenshots of UI changes
Everything should be the same.

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_controller.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_controller.ts
@@ -24,6 +24,7 @@ import {
   TimeSelectionAffordance,
 } from '../../../widgets/card_fob/card_fob_types';
 import {Scale} from '../../../widgets/line_chart_v2/lib/public_types';
+import {MinMaxStep} from './scalar_card_types';
 
 @Component({
   selector: 'scalar-card-fob-controller',
@@ -47,7 +48,8 @@ import {Scale} from '../../../widgets/line_chart_v2/lib/public_types';
 export class ScalarCardFobController {
   @Input() timeSelection!: TimeSelection;
   @Input() scale!: Scale;
-  @Input() minMax!: [number, number];
+  @Input() minMaxHorizontalViewExtend!: [number, number];
+  @Input() minMaxStep!: MinMaxStep;
   @Input() axisSize!: number;
 
   @Output() onTimeSelectionChanged = new EventEmitter<{
@@ -65,7 +67,7 @@ export class ScalarCardFobController {
 
   getAxisPositionFromStartStep() {
     return this.scale.forward(
-      this.minMax,
+      this.minMaxHorizontalViewExtend,
       [0, this.axisSize],
       this.timeSelection.start.step
     );
@@ -76,20 +78,18 @@ export class ScalarCardFobController {
       return null;
     }
     return this.scale.forward(
-      this.minMax,
+      this.minMaxHorizontalViewExtend,
       [0, this.axisSize],
       this.timeSelection.end.step
     );
   }
 
   getHighestStep(): number {
-    const minMax = this.minMax;
-    return Math.floor(minMax[0] < minMax[1] ? minMax[1] : minMax[0]);
+    return this.minMaxStep.maxStep;
   }
 
   getLowestStep(): number {
-    const minMax = this.minMax;
-    return Math.ceil(minMax[0] < minMax[1] ? minMax[0] : minMax[1]);
+    return this.minMaxStep.minStep;
   }
 
   getStepHigherThanAxisPosition(position: number): number {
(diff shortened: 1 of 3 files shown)


tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_test.ts
```diff
@@ -89,56 +97,14 @@ describe('ScalarFobController', () => {
...
-    });
-
-    it('returns the next lowest step when max is a floating point.', () => {
-      const fixture = createComponent({minMax: [4.5, 1]});
-
-      const lowestStep = fixture.componentInstance.getHighestStep();
-
-      expect(lowestStep).toBe(4);
-    });
-  });
```
rileyajones, line 116: Why is this test being deleted?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_test.ts
```diff
@@ -89,56 +97,14 @@ describe('ScalarFobController', () => {
...
-      expect(lowestStep).toBe(0);
-    });
-
-    it('gets the lowest step when minMax is not in order', () => {
-      const fixture = createComponent({minMax: [2, 0]});
-
-      const lowestStep = fixture.componentInstance.getLowestStep();
-
-      expect(lowestStep).toBe(0);
-    });
```
rileyajones, line 133: Why is this test being deleted?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_controller.ts
```diff
@@ -76,20 +78,18 @@ export class ScalarCardFobController {
...
-      this.minMax,
+      this.minMaxHorizontalViewExtend,
       [0, this.axisSize],
       this.timeSelection.end.step
     );
   }
 
   getHighestStep(): number {
-    const minMax = this.minMax;
-    return Math.floor(minMax[0] < minMax[1] ? minMax[1] : minMax[0]);
```
rileyajones, line 87: Why is this check no longer needed?

rileyajones (commented):
I'm generally okay with this change (minus the deleted tests) but I'm not sure I understand why it needs to happen? The changed functionality here doesn't seem to be used anywhere and there are no tests added verifying that it works differently.

Could you please add a new test?

tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_controller.ts
japie1235813: As stated in the description, the idea of using minMax to get the highest step is not correct.
The old minMax is actually view extend, which could be negative. For example, the step is [0, 2000] and the viewextend is [-200, 2200]  (see the console.log)

Screen Shot 2022-10-24 at 3 25 18 PM

I did not look into the math here 
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/widgets/line_chart_v2/line_chart_internal_utils.ts;l=31
but a couple of checks confirms minMax is _not_ min max steps we need.

tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_test.ts
japie1235813: Fair question, they are removed because they tests the old logic, which used minMax to get he highest/lostest step, which are tested in the new test.

The part of min max step being floating point was an incorrect concept in the new logic (which ideally should not happen) I can 
1. add Math.round function in getHighestStep/getLowestStep if that's necessary (floating point was not tested for the original minMax
2. add minMax being float to function getStepLowerThanAxisPosition but I think that does not matter for testing that function

rileyajones (approved):
Alright, that makes sense to me, thanks for the explanation. BTW the original reason for using the extent in more places is to handle issues with zooming. I already verified that this works fine but it's something you might want to keep an eye on when adding the prospective fob.

japie1235813:
Regards to the zoom that's an interesting point. From here I think minMax step is handled on zooming
https://cs.opensource.google/tensorflow/tensorboard/+/master:tensorboard/webapp/metrics/views/card_renderer/scalar_card_container.ts;l=374

We can chat more when we cross the bridge.

PR #5995, opened 2022-10-25
linked time: set up feature flag and fob rendering

japie1235813:
Introduce feature flag in the card fob and used to render a fob component.

Getting prospectiveFob feature and passing all the way down to card-fob.

Note: this part might need refactor once we have nested container or some kind of structure refactoring.

* Screenshots of UI changes
None

### DIFF
--- a/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
+++ b/tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
@@ -50,6 +50,8 @@ const TIME_SELECTION_TO_FOB: Record<keyof TimeSelection, Fob> = {
 export class CardFobControllerComponent {
   @ViewChild('startFobWrapper') readonly startFobWrapper!: ElementRef;
   @ViewChild('endFobWrapper') readonly endFobWrapper!: ElementRef;
+  @ViewChild('prospectiveFobWrapper')
+  readonly prospectiveFobWrapper!: ElementRef;
   @Input() axisDirection!: AxisDirection;
   @Input() timeSelection!: TimeSelection;
   @Input() cardFobHelper!: CardFobGetStepFromPositionHelper;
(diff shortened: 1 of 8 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
```diff
@@ -85,6 +89,8 @@ describe('card_fob_controller', () => {
     timeSelection: TimeSelection;
     showExtendedLine?: Boolean;
     steps?: number[];
+    isProspectiveFobFeatureEnabled?: Boolean;
+    prospectiveStep?: number | null;
```
rileyajones, line 93: nitpick: Why make this null rather than `undefined`?
```suggestion
    prospectiveStep?: number;
```

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
```diff
@@ -58,6 +60,8 @@ export class CardFobControllerComponent {
   @Input() highestStep!: number;
   @Input() lowestStep!: number;
   @Input() showExtendedLine?: Boolean = false;
+  @Input() isProspectiveFobFeatureEnabled: Boolean = false;
+  @Input() prospectiveStep: number | null = null;
```
rileyajones, line 64: nitpick: Why make this null rather than `undefined`?
```suggestion
  @Input() prospectiveStep?: number;
```

rileyajones (approved):

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.ts
japie1235813: prospectiveStep will be set to null on mouseleave event. While undefined means the variable is not defined, assigning it to null makes more sense to me.
Is this reasonable to you? Or any other strong reason to use undefined?

I noticed this two inputs should be optional. Adding question marks.

tensorboard/webapp/widgets/card_fob/card_fob_controller_test.ts
japie1235813: Different from the code, the test here makes sense to just use undefined, since it will be assigned null in line 144.

japie1235813: Ah my bad, I still need null here since I want to test element rendeing on step being null

PR #5997, opened 2022-10-25
Bring proto update script up to date to deal with new protos in TF2.11

JamesHollyer:
In TF 2.11 the tensorflow/tsl/protobuf directory is created(https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tsl/protobuf). This directory has a histogram.proto file which is referenced from summary.proto(https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/summary.proto#L5). Additionally that proto is imported with the public keyword which we previously were not looking for.

### DIFF
--- a/tensorboard/compat/proto/update.sh
+++ b/tensorboard/compat/proto/update.sh
@@ -26,6 +26,10 @@ rsync --existing "$1"/tensorflow/core/protobuf/*.proto tensorboard/compat/proto/
 rsync --existing "$1"/tensorflow/core/profiler/*.proto tensorboard/compat/proto/
 rsync --existing "$1"/tensorflow/core/util/*.proto tensorboard/compat/proto/
 rsync --existing "$1"/tensorflow/python/framework/*.proto tensorboard/compat/proto/
+if [ -d " "$1"/tensorflow/tsl/protobuf" ] # This directory first appears in TF 2.11
+then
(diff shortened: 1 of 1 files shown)


nfelt (changes requested):
The script changes look good, but I'm confused about the .proto file changes, since those don't actually seem to include any new proto files from the tsl directory (including histogram.proto). Are the .proto changes the result of running the script for TF 2.10 or something?

Ideally I think we'd either do the script change and the TF 2.11 proto update together, or we'd do 2 separate PRs (one just the script, and a follow up with the proto update), rather than having this PR include proto changes that aren't actually the ones that result from the script run against TF 2.11.

JamesHollyer:
> The script changes look good, but I'm confused about the .proto file changes, since those don't actually seem to include any new proto files from the tsl directory (including histogram.proto). Are the .proto changes the result of running the script for TF 2.10 or something?
> 
> Ideally I think we'd either do the script change and the TF 2.11 proto update together, or we'd do 2 separate PRs (one just the script, and a follow up with the proto update), rather than having this PR include proto changes that aren't actually the ones that result from the script run against TF 2.11.

These proto updates were simply the latest sync's. I can see that it is confusing to put them in the same PR though. I have removed them now.

nfelt (approved):

PR #6004, opened 2022-10-27
linked time: create listen area for fob and set step to null on mouseleave

japie1235813:
On mouse hovering the axis, we want to show the prospective fob. On mouse leaving, we want to remove the fob. This PR creates that element and implement the mouse leaving part.

A new div element to take the mouse event for displaying the prospective fob.
The element is placed beneath start/end fob then those events will not be triggered when start/end fob hovered and ready to be dragged. 

The range of this area is subjected to change. I made something seems reasonable to me for now. We can also think of extending the area or do something else about it (might be similar to the dragging flow) but that will come later if we feel the need to improve the flow.

* Screenshots of UI changes
Not visible still, but the area in scalar card kind of covers axis:
Screen Shot 2022-10-27 at 5 11 39 PM

### DIFF
--- a/tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_controller.ts
+++ b/tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_controller.ts
@@ -41,6 +41,7 @@ import {MinMaxStep} from './scalar_card_types';
       [isProspectiveFobFeatureEnabled]="isProspectiveFobFeatureEnabled"
       [cardFobHelper]="cardFobHelper"
       [showExtendedLine]="true"
+      (onPrespectiveStepChanged)="onPrespectiveStepChanged($event)"
       (onTimeSelectionChanged)="onTimeSelectionChanged.emit($event)"
       (onTimeSelectionToggled)="onTimeSelectionToggled.emit($event)"
     ></card-fob-controller>
@@ -97,6 +98,10 @@ export class ScalarCardFobController {
       [0, this.axisSize],
       this.prospectiveStep
     );
+  }
+
+  onPrespectiveStepChanged(step: number | null) {
+    this.prospectiveStep = step;
   }
 
   getHighestStep(): number {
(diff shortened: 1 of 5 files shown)


tensorboard/webapp/widgets/card_fob/card_fob_controller_component.scss
```diff
@@ -54,3 +55,13 @@ limitations under the License.
   opacity: 0.8;
   cursor: pointer;
 }
+
+.horizontal-prospective-area {
+  height: 25px;
+  transform: translateY(225px);
```
rileyajones, line 61: These values seem fairly arbitrary, especially the translate. Given the feature isn't fully implemented I can't test this to verify it works with different screen sizes. Can you please leave a comment explaining this?

rileyajones (commented):
The css seems fine but was removing the TS intentional?

tensorboard/webapp/widgets/card_fob/card_fob_controller_component.scss
japie1235813: Thanks for checking this. Yes I agree this is not a good css style. Updated to fit x-axis.

japie1235813:
What is TS here? Time Selection? and do you mean the name of css or which part about css?

rileyajones:
> What is TS here? Time Selection? and do you mean the name of css or which part about css?

TypeScript

rileyajones (approved):

tensorboard/webapp/metrics/views/card_renderer/scalar_card_fob_controller.ts
```diff
@@ -99,6 +100,10 @@ export class ScalarCardFobController {
     );
   }
 
+  onPrespectiveStepChanged(step: number | null) {
```
JamesHollyer, line 103: Ok, I am a little late here but, this should be onProspectiveStepChanged right? Note the o in Prospective.

japie1235813: ah, thanks for catching this. Yes that's typo and should be fixed in next PR

